PR #1515, opened 2019-10-30
Use device safety flags to show prompts before installing updates

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdmgr.in
+++ b/data/bash-completion/fwupdmgr.in
@@ -42,6 +42,7 @@ _fwupdmgr_opts=(
 	'--no-unreported-check'
 	'--no-metadata-check'
 	'--no-reboot-check'
+	'--no-safety-check'
 	'--show-all-devices'
 	'--sign'
 	'--filter'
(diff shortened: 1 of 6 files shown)


superm1:
Updated output with the changes:
```
No upgrades for Thunderbolt Controller, current is 44.00: 44.00=same
No upgrades for RTS5413 in Dell dock, current is 01.21: 01.21=same
No upgrades for RTS5487 in Dell dock, current is 01.47: 01.47=same
No upgrades for ColorHug2, current is 2.0.7: 2.0.7=same, 2.0.6=older, 2.0.5=older, 2.0.2=older
Upgrade available for System Firmware from 0.1.7.0 to 0.1.8.0
XPS 13 9380 must remain plugged into a power source for the duration of the update to avoid damage. Continue with update? [Y|n]: y
Downloading 0.1.8.0 for System Firmware...
Updating System Firmware…[  -                                    ]
Scheduling…              [***************************************]

An update requires a reboot to complete. Restart now? [Y|n]: n
```

PR #1517, opened 2019-10-31
Wip/hughsie/cleanups

hughsie:


superm1:
Looks like just that nit that clang caught for CI.

PR #1518, opened 2019-10-31
Move the version definition of a release into `.version`

superm1:
This is inspired by a change in flashrom to read the version string for meson
dynamically.

When making a release just put the version into `.version` to make the tarball
and tag the release as approriate.

No need for "post release version bump", this happens automatically from git
now by there being a dirty commit.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/RELEASE
+++ b/RELEASE
@@ -1,12 +1,16 @@
 fwupd Release Notes
 
-Write release entries:
+0. Create release version
 
-git log --format="%s" --cherry-pick --right-only 1.3.3... | grep -i -v trivial | grep -v Merge | sort | uniq
+export release_ver=
+
+1. Write release entries:
+
+git log --format="%s" --cherry-pick --right-only $(git describe --tags --abbrev=0)..HEAD | grep -i -v trivial | grep -v Merge | sort | uniq
 Add any user visible changes into ../data/org.freedesktop.fwupd.metainfo.xml
 appstream-util appdata-to-news ../data/org.freedesktop.fwupd.metainfo.xml > NEWS
 
-Update translations:
+2. Update translations:
 
 ninja-build fwupd-pot
 tx push --source
@@ -14,10 +18,7 @@ tx pull --all --force --minimum-perc=5
 ninja-build fix-translations
 git add ../po/*.po
 
-2. Commit changes to git:
-
-# MAKE SURE THIS IS CORRECT
-export release_ver="1.3.4"
+3. Commit changes to git:
 
 git commit -a -m "Release fwupd ${release_ver}"
 git tag -s -f -m "Release fwupd ${release_ver}" "${release_ver}"
@@ -25,21 +26,14 @@ git tag -s -f -m "Release fwupd ${release_ver}" "${release_ver}"
 git push --tags
 git push
 
-3. Generate the tarball:
+4. Generate the tarball:
 
 ninja dist
 
-3a. Generate the additional verification metadata
+4a. Generate the additional verification metadata
 
 gpg -b -a meson-dist/fwupd-${release_ver}.tar.xz
 
-4. Upload tarball:
+5. Upload tarball:
 
 scp meson-dist/fwupd-${release_ver}.tar.* hughsient@people.freedesktop.org:~/public_html/releases
-
-5. Do post release version bump in meson.build
-
-6. Commit changes:
-
-git commit -a -m "trivial: post release version bump"
-git push
--- a/contrib/PKGBUILD
+++ b/contrib/PKGBUILD
@@ -16,9 +16,7 @@ makedepends=('meson' 'valgrind' 'gobject-introspection' 'gtk-doc' 'python-pillow
 pkgver() {
     cd ${pkgname}
 
-    VERSION=$(git describe | sed 's/-/.r/;s/-/./')
-    [ -z $VERSION ] && VERSION=$(head meson.build | grep ' version :' | cut -d \' -f2)
-
+    VERSION=$(./contrib/get-version.py | sed 's/-/.r/;s/-/./')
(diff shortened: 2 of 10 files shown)


superm1:
Hmm, seems that it doesn't want to work for `ninja dist` though :/

RELEASE
```diff
@@ -1,45 +1,47 @@
 fwupd Release Notes
 
-Write release entries:
+0. Create release version
 
-git log --format="%s" --cherry-pick --right-only 1.3.2... | grep -i -v trivial | grep -v Merge | sort | uniq
+export release_ver=
+echo $release_ver > .version
```
superm1: probably would be better to automatically create `.version` from the env variable

superm1:
> Hmm, seems that it doesn't want to work for `ninja dist` though :/

OK this modification should work now with `ninja dist`.

Basically the way it works now:
* if you're running a dirty version, the dirty version gets put into place
* If you're running a checked out tag, that's put in place
* When creating a release just export a variable `release_version` and it fills out everywhere it's supposed to.
* When committing the new release, the updated `.version` file 

My _hope_ is that we might be able to move the entire release process into [Github actions](https://github.com/fwupd/fwupd/actions/new), but that might be wishful thinking.

hughsie:
I'm a bit confused why we need a .version checked into git. Can't we just generate that based on the last git tag? i.e. we still need a commit to change the 1.3.3 to 1.3.4, surely?

superm1:
> I'm a bit confused why we need a .version checked into git. Can't we just generate that based on the last git tag? i.e. we still need a commit to change the 1.3.3 to 1.3.4, surely?

I was thinking the same thing - but it's because `ninja dist` doesn't have the VCS info.  So you can't get the version into the tarball otherwise that I could find.

hughsie:
> because ninja dist doesn't have the VCS info

I guess it copies the tree. This is probably something we should ask the meson developers about, it seems the kind of useful thing we want in all my projects.

superm1:
> > because ninja dist doesn't have the VCS info
> 
> I guess it copies the tree. This is probably something we should ask the meson developers about, it seems the kind of useful thing we want in all my projects.

As I can tell it clones and then builds a tarball without `.git`

```
ninja: Entering directory `build'
[0/1] Creating source packages
Cloning into '/home/superm1/src/fwupd/build/meson-dist/fwupd-1.3.2-174-g26d3ad00'...
done.
Submodule 'contrib/flatpak' (https://github.com/flathub/org.freedesktop.fwupd) registered for path 'contrib/flatpak'
Cloning into '/home/superm1/src/fwupd/build/meson-dist/fwupd-1.3.2-174-g26d3ad00/contrib/flatpak'...
Submodule path 'contrib/flatpak': checked out '3e60bea4e74e4143bb2524762092e3d6e1e8930f'
Testing distribution package /home/superm1/src/fwupd/build/meson-dist/fwupd-1.3.2-174-g26d3ad00.tar.xz
```

So if you don't commit the `.version` before it's created `.version` doesn't end up in the tarball, which means for a user the version string won't be known.

superm1:
> So if you don't commit the `.version` before it's created `.version` doesn't end up in the tarball, which means for a user the version string won't be known.

I guess I've got another idea though.  How about parsing `data/org.freedesktop.fwupd.metainfo.xml` to get the latest release?

hughsie:
>  How about parsing data/org.freedesktop.fwupd.metainfo.xml to get the latest release

I think that's a great idea.

superm1:
> I'm a bit confused, why would this be missing, and what's it got to do with the tests?

It's missing on the system I'm currently using, so I couldn't actually run `ninja dist` since synapticsmst self tests fail.
```
# ls /proc/modules
ls: cannot access '/proc/modules': No such file or directory
```

hughsie:
> so I couldn't actually run ninja dist since synapticsmst self tests fail

Got a backtrace or something? The self tests shouldn't be parsing the modules...

superm1:
> > so I couldn't actually run ninja dist since synapticsmst self tests fail
> 
> Got a backtrace or something? The self tests shouldn't be parsing the modules...

It's because of this: https://github.com/fwupd/fwupd/blob/master/plugins/synapticsmst/fu-plugin-synapticsmst.c#L25 which we intentionally do.

hughsie:
So maybe catch the error_local of g_file_get_contents and return FALSE if the error is not-exists? I think we need that for correct runtime operation too.

superm1:
> So maybe catch the error_local of g_file_get_contents and return FALSE if the error is not-exists? I think we need that for correct runtime operation too.

On this system synapticsmst won't work anyway.  I think skipping the self test is fine.

hughsie:
> On this system synapticsmst won't work anyway

Right, but fixing it in the plugin means that it would work on a machine with everything compiled into the kernel image, but with a synapticsmst interface. It's just fixing a bug that nobody has hit yet.

superm1:
> > On this system synapticsmst won't work anyway
> 
> Right, but fixing it in the plugin means that it would work on a machine with everything compiled into the kernel image, but with a synapticsmst interface. It's just fixing a bug that nobody has hit yet.

Gotcha.  OK adjusted.

hughsie:
I think you're a CI pass away from LGTM :)

superm1:
> I think you're a CI pass away from LGTM :)


Looks like CI is finally working 🎈

superm1:
> I'm slightly wondering if ninja dist will come out correctly, but there's only one way to tell :)

I think it might be an order of events thing.  As long as you follow the order listed in `RELEASE` (IE you've tagged `1.3.4` first) it should work.  I guess we'll see though indeed 😜

PR #1519, opened 2019-11-01
Wip/hughsie/history

hughsie:


superm1:
With clients potentially being able to automatically report updates more frequently hopefully that last commit doesn't lose as much data as if potentially could.

PR #1523, opened 2019-11-01
Rename unifying plugin to logitech-hid

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/logitech-hid/fu-logitech-hid-hidpp-msg.c
```diff
@@ -8,17 +8,17 @@
...
 
-#include "fu-unifying-hidpp.h"
-#include "fu-unifying-hidpp-msg.h"
+#include "fu-logitech-hid-hidpp.h"
+#include "fu-logitech-hid-hidpp-msg.h"
 
-FuUnifyingHidppMsg *
-fu_unifying_hidpp_msg_new (void)
+FuLogitechHidHidppMsg *
+fu_logitech_hid_hidpp_msg_new (void)
```
hughsie: hid_hidpp looks kinda odd to me -- maybe `FuLogitechHidppMsg` instead?

hughsie: Aaaaactually, I wonder if the plugin name should be logitech-hidpp?

FFY00: It will also have the bootloader protocols, but I guess that's pretty small.

superm1: I just pushed a fixup to rename to logitech-hidpp instead then.

FFY00:
Will this require any change in the remote metadata?

superm1:
> Will this require any change in the remote metadata?

I don't think so.  It still identifies the protocol as `com.logitech.unifying` and `com.logitech.unifyingsigned` which was the only interaction I expected between the two.

PR #1524, opened 2019-11-02
trivial: Fix a compile error with older versions of gudev

hughsie:
Just move the G_DEFINE_AUTOPTR_CLEANUP_FUNC to the internal header to avoid
forgetting to define this in each plugin.


hughsie:
This fixes a compile error in the flatpak package.

PR #1527, opened 2019-11-04
Add a new contrib script for installing a BIOS from Dell EXE files

superm1:
The intended use of this is something to direct people at when a capsule isn't available on LVFS (for whatever reason) so they can at least get their systems updated.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


contrib/add-capsule-header.py
```diff
@@ -45,27 +45,31 @@ def main(args):
             bin_data = bin_data[hdrsz_old:]
 
     # set header flags
-    flags = CAPSULE_FLAGS_PERSIST_ACROSS_RESET | CAPSULE_FLAGS_INITIATE_RESET
-    if args.flags:
-        flags = int(args.flags, 16)
+    flags = CAPSULE_FLAGS_PERSIST_ACROSS_RESET | CAPSULE_FLAGS_INITIATE_RESET | CAPSULE_FLAGS_POPULATE_SYSTEM_TABLE
```
hughsie: It looks like Windows unconditionally adds this too, in https://docs.microsoft.com/en-us/windows-hardware/drivers/bringup/processing-updates -- right?

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -0,0 +1,81 @@
...
+        #match the device we were given
+        if deviceid:
+            if item.get_id() != deviceid:
+                continue
+        # internal
+        if (not item.has_flag(1 << 0)):
+            continue
+        # needs reboot
+        if (not item.has_flag(8 << 0)):
+            continue
```
hughsie: do we need the outer brackets in both cases?

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -0,0 +1,81 @@
...
+        result = item.to_string()
+        # return the first hit for UEFI plugin
+        if 'uefi' in result:
```
hughsie: why not just match the plugin name?

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -0,0 +1,81 @@
...
+    print("Couldn't find any UEFI devices")
+    sys.exit(1)
+
+if __name__ == '__main__':
+    ARGS = parse_args()
+    CLIENT = Fwupd.Client()
+    CLIENT.connect()
+    check_exists(ARGS.exe)
+    directory = tempfile.mkdtemp()
+    guid,deviceid,version=find_uefi_device(CLIENT, ARGS.deviceid)
```
hughsie: spacing

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -0,0 +1,81 @@
...
+    cab = generate_cab(ARGS.exe, directory, guid, version)
+    install(CLIENT, cab, deviceid, True, True)
+    print("Reboot your machine to finish the update.")
```
hughsie: don't we need to check flags?

hughsie:
Fedora filelist needs fixing too:

    RPM build errors:
    File not found: /root/rpmbuild/BUILDROOT/fwupd-1.3.4-0.1alpha.fc30.x86_64/usr/share/fwupd/firmware-packager

contrib/add-capsule-header.py
superm1: Well it does based on what was in the INF.  
* I think it's a better "default" to have it in at least (especially since it's definitely required for Dell system)
* add_capsule_header.py has an argument to override anyway.

contrib/firmware_packager/install_dell_bios_exe.py
superm1: I don't think so.

contrib/firmware_packager/install_dell_bios_exe.py
superm1: because I didn't realize there is a `get_plugin` available to introspection :smiley: 
Will fix.

contrib/firmware_packager/install_dell_bios_exe.py
superm1: Flags are checked earlier on when fetching the device to get one that needs reboot.

PR #1528, opened 2019-11-04
Bump the required gudev version

hughsie:
Flatpak shared-modules now ships gudev over 232.


superm1:
The top level `meson.build`:
* [ ] doesn't need `HAVE_GUDEV_232` defined anymore.
* [ ] should have min dep version bumped up to 232

Issue #1526, opened 2019-11-03
Give information about inability to load plugin to the user


FFY00:
Right now when a plugin can't be loaded (ex. missing dependency) we just fail silently. This works fine but causes some confusing among users. I think we should save the reason and during device enumeration we could say that we skip X device because the plugin could not be loaded and provide the reason the plugin wasn't loaded.

@hughsie @superm1 what do you think?

This would prevent issues such as #1454.

superm1:
Well the problem is that on Every single system some plugins are disabled. For example you would never have coreboot and UEFI on the same system. So you don't want to message to the user every single plugin that was disabled.

Additionally you don't know what kind or how many devices a plugin will create at the time it's loaded.

The UEFI case you linked above sounds like a packaging problem to me, that library should have been installed if the package was installed.

In Arch could you split fwupd into two binary packages maybe (fwupd and fwupd-efi)? Then make fwupd-efi optional but it requires a hard dependency on the TPM library that was missing.

FFY00:
But don't we have metadata to identify the possible compatible devices before loading the plugin? At some point we need to hand the device over to the correct plugin, how do we choose that plugin? From what I understand we do that based on the quirk files, can't just print a message saying that device X, Y and Z were skipped because the plugin couldn't be loaded?

> Well the problem is that on Every single system some plugins are disabled. For example you would never have coreboot and UEFI on the same system. So you don't want to message to the user every single plugin that was disabled.

It is fine if a plugin was disabled, we just print the message if we were trying to use the plugin but couldn't.

> Additionally you don't know what kind or how many devices a plugin will create at the time it's loaded.

So we have no clue about the supported devices before the plugin is loaded?

> The UEFI case you linked above sounds like a packaging problem to me, that library should have been installed if the package was installed.

Well, I don't think it's that clear. For me is kind of a gray area. Packaging every single plugin separately is not a great option.

> In Arch could you split fwupd into two binary packages maybe (fwupd and fwupd-efi)? Then make fwupd-efi optional but it requires a hard dependency on the TPM library that was missing.

True but that's not a long term solution. If we end up adding more plugins with external dependencies we wouldn't want for distributions to have the split all of them in different packages.

Detecting the dependencies at runtime is what most sense to me. I think for most cases it should be possible to identify the supported devices (or at least, potentially supported devices) without needing any external library.

hughsie:
> Then make fwupd-efi optional but it requires a hard dependency on the TPM library that was missing

[...]

PR #1531, opened 2019-11-04
Create a dummy device for failing to load plugins

superm1:
One way to address the stuff mentioned in https://github.com/fwupd/fwupd/issues/1526.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -4369,6 +4369,15 @@ fu_engine_load_plugins (FuEngine *self, GError **error)
 		/* if loaded from fu_engine_load() open the plugin */
 		if (self->usb_ctx != NULL) {
 			if (!fu_plugin_open (plugin, filename, &error_local)) {
+				g_autoptr(FuDevice) device = fu_device_new ();
+				g_autofree gchar *dev_name = g_strdup_printf ("%s device", name);
+				fu_device_set_name (device, dev_name);
+				fu_device_set_id (device, filename);
+				fu_device_set_quirks (device, self->quirks);
+				fu_device_add_instance_id (device, filename);
+				fu_device_set_update_error (device, error_local->message);
+				fu_device_setup (device, NULL);
+				fu_engine_add_device (self, device);
 				g_warning ("%s", error_local->message);
 				continue;
 			}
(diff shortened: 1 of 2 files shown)


superm1:
With this in place, here is what shows up when TPM library is missing.

```
├─dell device:
│     Device ID:           274acc02c0c999db4fa80c249403d11cc870280a
│     Update Error:        failed to open plugin /usr/local/lib/x86_64-linux-gnu/fwupd-plugins-3/libfu_plugin_dell.so: libtss2-esys.so.0: cannot open shared object file: No such file or directory
│     GUID:                048ff691-5ebb-5079-9817-e890ffa1ad30 ← /usr/local/lib/x86_64-linux-gnu/fwupd-plugins-3/libfu_plugin_dell.so
│   
└─uefi device:
      Device ID:           415773eab9fa918d3a8084745dba598450ff458d
      Update Error:        failed to open plugin /usr/local/lib/x86_64-linux-gnu/fwupd-plugins-3/libfu_plugin_uefi.so: libtss2-esys.so.0: cannot open shared object file: No such file or directory
      GUID:                4ce8a2db-5cb9-5b07-a3fe-ee09e12be224 ← /usr/local/lib/x86_64-linux-gnu/fwupd-plugins-3/libfu_plugin_uefi.so
```

hughsie:
Not a fan :/ This is going to be really confusing where plugins are intentionally not loaded. I think devices in general need to represent **devices**, not plugins. I think `fwupdmgr get-plugins` would make sense to show this kind of information, but not in the device list.

superm1:
> Not a fan :/ This is going to be really confusing where plugins are intentionally not loaded. I think devices in general need to represent **devices**, not plugins. 

Intentionally not loaded?  Could you give an example?
I was thinking that this should only show plugins that failed to load, not failed to startup.

> I think `fwupdmgr get-plugins` would make sense to show this kind of information, but not in the device list.

Like a debugging feature?  I guess from the way you describe it this is what comes to mind:

```
# fwupdmgr get-plugins:
[color green] dfu: enabled
[color green] colorhug: enabled
[color grey] coreboot: not a coreboot system
.
.
[color red] uefi: disabled: libtss2-esys.so.0: cannot open shared object file: No such file or directory
.
.

```

hughsie:
> guess from the way you describe it this is what comes to mind

Yes, something exactly like that. I think it's a slippery slope trying to represent everything as either a device or release.

superm1:
> Yes, something exactly like that. I think it's a slippery slope trying to represent everything as either a device or release.

OK.  That would require a lot more planning as it's exposing plugins to clients, which is something that doesn't really exist today in the library or dbus interface.  I think we're back at the point of does the effort required to make this happen make sense.

PR #1535, opened 2019-11-06
udev-device: fix offset of vendor id of hidraw devices

CragW:
The read property from hidraw HID_ID is 0018:00002D1F:00004946 where
the vendor id 2D1F starts from offset 9 from base address.

This change probes the hidraw vid/pid with a wanted plugin properly.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-udev-device.c
+++ b/src/fu-udev-device.c
@@ -152,7 +152,7 @@ fu_udev_device_probe (FuDevice *device, GError **error)
 	    g_strcmp0 (priv->subsystem, "hidraw") == 0) {
 		tmp = g_udev_device_get_property (udev_parent, "HID_ID");
 		if (tmp != NULL && strlen (tmp) == 22) {
-			priv->vendor = fu_udev_device_read_uint16 (tmp + 10);
+			priv->vendor = fu_udev_device_read_uint16 (tmp + 9);
 			priv->model = fu_udev_device_read_uint16 (tmp + 18);
 		}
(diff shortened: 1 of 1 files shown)


superm1:
I'm pretty sure this will break other hidraw devices like Synaptics RMI touchpad and Unifying dongle.
@bentiss any idea why `HID_ID` is a little off on this particular Wacom device?

hughsie:
Furthermore, is there a defined standard for HID_ID? Can we just split the string and parse each section? That seems less fragile.

superm1:
The kernel exports it as uevents and it looks like this code hasn't changed in 12 years.
https://github.com/torvalds/linux/blame/0365fb6baeb1ebefbbdad9e3f48bab9b3ccb8df3/drivers/hid/hid-core.c#L2286

superm1:
Since the kernel has fixed delimiters of `:` in the property, I would say instead we're better off splitting along those and then interpreting each of those strings as numbers rather than this fixed offset stuff.

superm1:
Well actually Crag's change for the offset stuff looks right - it has just conveniently "worked" for vendors so far because the other Wacom VID, the Unifying VID, and the Synaptics RMI VID all start with "0".

So I think we should probably take this as is, but the split might be a nice follow up to make it less fragile since VID/PID can be up to 8 characters for some subsystems according to the kernel.

PR #1537, opened 2019-11-06
fu-udev-device: Adjust bounds checking to be safer

superm1:
Right now different integer sizes are used willy nilly and could
conceivably lead to overflow issues.

Additionally the values that come out of the kernel for `HID_ID`
are actually 8 digits, so they should be represented by a larger
data type.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

This is a follow up for #1535 which is still correct, but this should hopefully scale better and be less fragile.

### DIFF
--- a/src/fu-udev-device.c
+++ b/src/fu-udev-device.c
@@ -33,8 +33,8 @@
 typedef struct
 {
 	GUdevDevice		*udev_device;
-	guint16			 vendor;
-	guint16			 model;
+	guint32			 vendor;
+	guint32			 model;
 	guint8			 revision;
 	gchar			*subsystem;
 	gchar			*device_file;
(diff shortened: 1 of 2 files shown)


src/fu-udev-device.c
```diff
@@ -133,27 +145,34 @@ fu_udev_device_probe (FuDevice *device, GError **error)
...
 		tmp = g_udev_device_get_property (udev_parent, "HID_ID");
-		if (tmp != NULL && strlen (tmp) == 22) {
-			priv->vendor = fu_udev_device_read_uint16 (tmp + 9);
-			priv->model = fu_udev_device_read_uint16 (tmp + 18);
+		if (tmp != NULL) {
+			g_auto(GStrv) split = g_strsplit (tmp, ":", -1);
+			if (g_strv_length (split) == 3) {
+				guint64 val = g_ascii_strtoull (split[1], NULL, 16);
+				if (val < G_MAXUINT32)
+					priv->vendor = val;
```
hughsie: I think we want a g_warning here, right?

superm1: for the inverse; yes. i'll add.

superm1:
Also FYI I backported the fix from #1535 to 1_2_X, but won't backport this one.

PR #1542, opened 2019-11-07
usb: make loading vendor/product/serial strings non-fatal

aleksander0m:
If for any reason loading these strings fail (e.g. the DW5821e/eSIM
module fails to load the product string when running the fastboot USB
layout), do not make the whole open operation fail. Instead, default
to an empty string, e.g. so that the same operation is no longer
attempted any more.

### DIFF
--- a/src/fu-usb-device.c
+++ b/src/fu-usb-device.c
@@ -129,11 +129,16 @@ fu_usb_device_open (FuDevice *device, GError **error)
 		idx = g_usb_device_get_manufacturer_index (priv->usb_device);
 		if (idx != 0x00) {
 			g_autofree gchar *tmp = NULL;
+			g_autoptr(GError) error_local = NULL;
 			tmp = g_usb_device_get_string_descriptor (priv->usb_device,
-								  idx, error);
-			if (tmp == NULL)
-				return FALSE;
-			fu_device_set_vendor (device, g_strchomp (tmp));
+								  idx, &error_local);
+			if (tmp != NULL)
+				fu_device_set_vendor (device, g_strchomp (tmp));
+			else
+				g_debug ("failed to load manufacturer string for usb device %u:%u: %s",
+					 g_usb_device_get_bus (priv->usb_device),
+					 g_usb_device_get_address (priv->usb_device),
+					 error_local->message);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/fu-usb-device.c
```diff
@@ -130,10 +130,8 @@ fu_usb_device_open (FuDevice *device, GError **error)
 		if (idx != 0x00) {
 			g_autofree gchar *tmp = NULL;
 			tmp = g_usb_device_get_string_descriptor (priv->usb_device,
-								  idx, error);
-			if (tmp == NULL)
-				return FALSE;
-			fu_device_set_vendor (device, g_strchomp (tmp));
+								  idx, NULL);
+			fu_device_set_vendor (device, tmp ? g_strchomp (tmp) : "");
```
hughsie: I don't think setting "" makes sense. Maybe:

    if (tmp != NULL)
        fu_device_set_vendor (device, g_strchomp (tmp));

?

aleksander0m: I added that empty string so that the same string loading operation is not attempted any more, but well, I guess that would not harm either, maybe it succeeds on a new attempt later on (not in the case of the DW5821e/eSIM, where it always fails consistently). Let me update the commit.

aleksander0m:
Repushed after rewording the commit message as well.

src/fu-usb-device.c
```diff
@@ -130,10 +130,9 @@ fu_usb_device_open (FuDevice *device, GError **error)
 		if (idx != 0x00) {
 			g_autofree gchar *tmp = NULL;
 			tmp = g_usb_device_get_string_descriptor (priv->usb_device,
-								  idx, error);
-			if (tmp == NULL)
-				return FALSE;
-			fu_device_set_vendor (device, g_strchomp (tmp));
+								  idx, NULL);
```
superm1: don't we want to do an error local dance here and at least notate to debug logs that the string failed to read and why?  It could be indicative of a hardware problem and this could help debugging later.

aleksander0m: That is a good idea, let me update the patch.

aleksander0m: Pushed an update, let me know what you think

aleksander0m: This is how it looks like in the logs:
```
17:24:52:0251 FuDevice             using 0703a244275ed772f8e276c4b31389afc296101d for usb:00:03
17:24:53:0251 FuUsbDevice          failed to load product string for usb device 1:30: USB error on device 413c:81e1 : Operation timed out [-7]
17:24:53:0275 FuPluginFastboot     fastboot version=0.5
17:24:53:0335 FuPlugin             emit added from fastboot: 0703a244275ed772f8e276c4b31389afc296101d
```

superm1: Looks great, thanks.

aleksander0m: @hughsie are you fine with these new changes?

PR #1543, opened 2019-11-07
ModemManager: added support for the DW5821e/eSIM variant

aleksander0m:
Type of pull request:
- [X ] Feature

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -15,6 +15,9 @@
 #include "fu-mm-utils.h"
 #include "fu-qmi-pdc-updater.h"
 
+/* Amount of time for the modem to boot in fastboot mode. */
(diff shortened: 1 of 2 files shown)


plugins/modem-manager/fu-mm-device.c
```diff
@@ -15,6 +15,9 @@
 #include "fu-mm-utils.h"
 #include "fu-qmi-pdc-updater.h"
 
+/* Amount of time for the modem to boot in fastboot mode. */
+#define FU_MM_DEVICE_REMOVE_DELAY_RE_ENUMERATE	20000	/* ms */
```
hughsie: I wonder if this should be a quirk -- @superm1 -- what do you think?

superm1: The code comments how it's explicitly set for the "worst" case scenario, and it will usually come back sooner.  So I think it doesn't really affect the other supported devices any differently.  Not a lot of value in a quirk for something that is a timeout in my mind.

PR #1544, opened 2019-11-07
Add versions formats for the Microsoft Surface devices

hughsie:
Apparently there are yet more ways to encode a 32 bit number...


src/fu-common-version.c
```diff
@@ -80,6 +80,20 @@ fu_common_version_from_uint32 (guint32 val, FwupdVersionFormat kind)
...
+	if (kind == FWUPD_VERSION_FORMAT_SURFACE_LEGACY) {
+		/* 10b.12b.10b */
+		return g_strdup_printf ("%u.%u.%u",
+					(val >> 22) & 0x3ff,
+					(val >> 10) & 0xfff,
+					val & 0x3ff);
+	}
+	if (kind == FWUPD_VERSION_FORMAT_SURFACE) {
+		/* 8b.16b.8b */
+		return g_strdup_printf ("%u.%u.%u",
```
superm1: Is this used for UEFI on a surface?  or is this used for something else?

hughsie: Exactly that.

Issue #1529, opened 2019-11-04
8BitDo SF30Pro testing upgrade v1.32 works but reports "USB error on device" 


herochao:
Hello,
I just enabled the Testing LVFS repo to update my new SF30Pro to version 1.32 (v1.31->v1.32).

The update proceeded as normal but at the very end, the device reported to be disconnected, the controller apparently restarting in "Switch Mode". When reconnecting it in "Android Mode" and checking for firmware, I was prompted to submit a debug report (which I did), but the controller afterwards reported to be on version 1.32 and seemingly worked as usual.

This is really not an issue for me, but it might be interesting for you and provides testing for that particular update path.

Thank you for the great project and have a nice day!

update-output:
```
Upgrade available for 8Bitdo from 1.31 to 1.32
Downloading 1.32 for 8Bitdo...
Entpacken …              [***************************************]
Authentifizierung …      [***************************************]
8Bitdo wird aktualisiert … -                                     ]
Schreiben …              [***************************************] Weniger als eine Minute verbleiben…
USB error on device 2dc8:5750 : No such device (it may have been disconnected) [-4]
```

To help us pinpoint your issue, please insert the output of the
following commands when ran on the system with the issue:

```shell
$ fwupdmgr --version
client version:	1.3.3
compile-time dependency versions
	gusb:	0.3.0
	efivar:	37
daemon version:	1.3.3
```

Note, the switch `--version` is only present since version 0.9.6. If you use an
earlier version, please use the package manager to find out the package
version. For example, `dpkg -l fwupd`.

```shell
$ fwupdmgr get-devices
System Product Name
│
├─8Bitdo SF30 Pro:
│     Device ID:           672c087de09848d9e7ee32aa1dea2fbeb8b81e6b
│     Zusammenfassung:     A redesigned classic game controller
│     Current version:     1.32
│     Vendor:              8Bitdo (USB:0x2DC8)
│     GUIDs:               043c7673-bd45-5e2e-b8b9-c22df2dfec2a
│                          b02f1319-c1c4-54f4-b7f2-5bb3552f6a9e
│                          e95bfb47-11f9-5afd-856e-fbc45730456d
│                          269b3121-097b-50d8-b9ba-d1f64f9cd241
│     Device Flags:        • Updatable
│                          • Supported on remote server
│                          • Requires a bootloader
│                          • Device will not re-appear after update completes
│
[... SSDs, they hopefully didn't cause anything?]
│
└─System Firmware:
[...]

PR #1550, opened 2019-11-10
Skip cleanup after device is done updating if `will-disappear` is set…

superm1:
… (Fixes: #1529)

This fixes a regression introduced by 2031ce3bf67885a4db247fa4088b36ae50011ca0
that leads to:
```
USB error on device 2dc8:5750 : No such device (it may have been disconnected) [-4]
```

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/ebitdo/fu-ebitdo-device.c
+++ b/plugins/ebitdo/fu-ebitdo-device.c
@@ -512,8 +512,14 @@ fu_ebitdo_device_write_firmware (FuDevice *device,
 	/* when doing a soft-reboot the device does not re-enumerate properly
 	 * so manually reboot the GUsbDevice */
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
-	if (!g_usb_device_reset (usb_device, error)) {
-		g_prefix_error (error, "failed to force-reset device: ");
+	if (!g_usb_device_reset (usb_device, &error_local)) {
+		g_prefix_error (&error_local, "failed to force-reset device: ");
+		if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_WILL_DISAPPEAR)) {
+			fu_device_set_remove_delay (device, 0);
+			g_debug ("%s", error_local->message);
+			return TRUE;
+		}
+		g_propagate_error (error, g_steal_pointer (&error_local));
 		return FALSE;
 	}
 
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1872,7 +1872,14 @@ fu_engine_device_cleanup (FuEngine *self,
 			  FwupdInstallFlags flags,
 			  GError **error)
 {
-	g_autoptr(FuDeviceLocker) locker = fu_device_locker_new (device, error);
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_WILL_DISAPPEAR)) {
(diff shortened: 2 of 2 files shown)


src/fu-engine.c
```diff
@@ -1928,6 +1928,12 @@ fu_engine_update_cleanup (FuEngine *self,
 		return FALSE;
 	str = fu_device_to_string (device);
 	g_debug ("performing cleanup on %s", str);
+
+	if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_WILL_DISAPPEAR)) {
+		g_debug ("skipping cleanup due to will-disappear flag");
+		return TRUE;
+	}
+
```
superm1: This way is more consistent with the rest of the vfuncs that are overridden when `FWUPD_DEVICE_FLAG_WILL_DISAPPEAR` is set, but the unfortunate side effect of doing it this way is `fu_plugin_runner_update_cleanup ` also won't be called.  So it's worth considering putting this block into `fu_engine_device_cleanup` instead.

plugins/ebitdo/fu-ebitdo-device.c
```diff
@@ -512,8 +512,14 @@ fu_ebitdo_device_write_firmware (FuDevice *device,
 	/* when doing a soft-reboot the device does not re-enumerate properly
 	 * so manually reboot the GUsbDevice */
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
-	if (!g_usb_device_reset (usb_device, error)) {
```
hughsie: In other plugins there are a few error codes that we ignore, i.e. we do the error_local thing. I think it would be wise to do this here, as even in the non-FWUPD_DEVICE_FLAG_WILL_DISAPPEAR case a reset failing is quite a common thing due to a hub in the chain doing slightly the wrong thing.

I think it also makes sense to move the `fu_device_set_remove_delay` bit out either before or after the reset too.

src/fu-engine.c
hughsie: I still think we want to run cleanup even for the FWUPD_DEVICE_FLAG_WILL_DISAPPEAR case -- else it just gets a bit magic. I think we can detect the FWUPD_DEVICE_FLAG_WILL_DISAPPEAR case in the ebitdo plugin cleanup and just return TRUE there.

superm1: The problem is that the action of opening/closing the device locker is what causes a problem in the first place.  You can't open or close a device that has already disappeared, so the device cleanup function fails.

hughsie: Aha, gotcha. I think just move this into fu_engine_device_cleanup then.

plugins/ebitdo/fu-ebitdo-device.c
superm1: > In other plugins there are a few error codes that we ignore, i.e. we do the error_local thing. I think it would be wise to do this here, as even in the non-FWUPD_DEVICE_FLAG_WILL_DISAPPEAR case a reset failing is quite a common thing due to a hub in the chain doing slightly the wrong thing.

Are you just saying that usb reset failing shouldn't be fatal no matter it disappears or not?

Can you give an example?  I looked through the tree some and didn't see anything.

> 
> I think it also makes sense to move the `fu_device_set_remove_delay` bit out either before or after the reset too.

This is explicitly only for the case of `will-disappear`.  So if we make usb reset non-fatal all the time, then it still needs that guard.

PR #1552, opened 2019-11-11
fu-util: Allow overriding daemon/client check with environment variable

superm1:
The check is mostly for making sure that if someone installs a snap that
they don't try to use a non-snapped command line client or vice versa.

However this makes development annoying because it is common to launch
the daemon manually in verbose mode without systemd when checking a
problem.  So allow setting an environment variable to make the client
ignore this check.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


superm1:
> Can't we just ask the daemon/systemd if fwupd is running in the foreground? I don't think we want the version check ever if we're running the daemon by hand.

Great idea, thanks!  Just pushed that.

PR #1553, opened 2019-11-11
Update dependencies.xml

Sean-StarLabs:
get-version.py depends on git

And, not sure where it should be put, but a dependency for libfwupd2 of libxmlb is needed.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/dependencies.xml
+++ b/contrib/ci/dependencies.xml
@@ -418,6 +418,7 @@
       <package />
     </distro>
     <distro id="debian">
+      <control />
       <package variant="x86_64" />
       <package variant="s390x" />
       <package variant="i386" />


superm1:
Hmm, well actually we don't want a hard dependency on `git`.  Counter-proposal: https://github.com/fwupd/fwupd/pull/1554

superm1:
> And, not sure where it should be put, but a dependency for libfwupd2 of libxmlb is needed.

This is supposed to happen automatically from `shlibs:Depends`: https://github.com/fwupd/fwupd/blob/master/contrib/debian/control.in#L19

Can you show more clearly what is going wrong?  It might be because you have an older libxmlb on your system already perhaps.

Sean-StarLabs:
```
sudo OS=debian-x86_64 contrib/ci/generate_docker.py
sudo docker run --privileged -t -v `pwd`:/build fwupd-debian-x86_64
sudo dpkg -i fwupd_**.deb libfwupd2**.deb
sudo apt -f install
```

Then using fwupdmgr, you get the `fwupdmgr: /usr/lib/x86_64-linux-gnu/libxmlb.so.1: version `LIBXMLB_0.1.13`. 

Resolved by installing libxmlb and `libxmlb - 0.1.13-1` is available.

Sean-StarLabs:
> Hmm, well actually we don't want a hard dependency on `git`. Counter-proposal: #1554

Wouldn't that just not return a version number?

superm1:
> Wouldn't that just not return a version number?

It returns the last released version number.  If you don't have `git` installed it falls back to parsing the last released changelog.

superm1:
> Then using fwupdmgr, you get the fwupdmgr: /usr/lib/x86_64-linux-gnu/libxmlb.so.1: version LIBXMLB_0.1.13`.

Are you running the binaries on Ubuntu or on Debian?   Something seems fishy here.

Sean-StarLabs:
> > Then using fwupdmgr, you get the fwupdmgr: /usr/lib/x86_64-linux-gnu/libxmlb.so.1: version LIBXMLB_0.1.13`.
> 
> Are you running the binaries on Ubuntu or on Debian? Something seems fishy here.

Built on Ubuntu. Then installing on a Mint and a Kali (just to make it nice and simple).

But surely it shouldn't make a difference, if it's a dependency, it _should be installed on any Debian distro?

superm1:
> > > Then using fwupdmgr, you get the fwupdmgr: /usr/lib/x86_64-linux-gnu/libxmlb.so.1: version LIBXMLB_0.1.13`.
> > 
> > 
> > Are you running the binaries on Ubuntu or on Debian? Something seems fishy here.
> 
> Built on Ubuntu. Then installing on a Mint and a Kali (just to make it nice and simple).
> 
> But surely it shouldn't make a difference, if it's a dependency, it _should be installed on any Debian distro?

OK I think I see what's happening.  the shlib dependency checking doesn't actually imply a minimum version - just that the library is installed.

So you end up with something like this:
```
Depends: .........., libxmlb1, shared-mime-info
```

I think this might be happening because of the lack of a symbols file on libxmlb.  So from a fwupd perspective not a lot can be done, but libxmlb debian package would need the symbols file added to it's packaging.

PR #1556, opened 2019-11-12
Allows confined snaps to activate fwupd via D-Bus

kenvandine:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
LGTM for the distro built packages.
The one question I would have is do we need to do something similar to the classic snapped daemon to let it work the same way?

kenvandine:
> LGTM for the distro built packages.
> The one question I would have is do we need to do something similar to the classic snapped daemon to let it work the same way?

I don't think so, @jhenstridge can you please confirm?

superm1:
The systemd unit is dynamically built there by snapd, so I would guess if anything was going to do it, that would be snapd.

PR #1557, opened 2019-11-13
Wip/hughsie/via pd child

hughsie:


plugins/vli-usbhub/fu-vli-usbhub-device.c
```diff
@@ -469,7 +471,7 @@ fu_vli_usbhub_device_erase_sector (FuVliUsbhubDevice *self, guint32 addr, GError
 	return TRUE;
 }
 
-static gboolean
+gboolean
 fu_vli_usbhub_device_erase_sectors (FuVliUsbhubDevice *self,
 				    guint32 addr,
 				    gsize sz,
```
superm1: Should this split out to a common file?

plugins/vli-usbhub/fu-vli-usbhub-device.c
```diff
@@ -653,14 +655,17 @@ fu_vli_usbhub_device_guess_kind (FuVliUsbhubDevice *self, GError **error)
 	return TRUE;
 }
 
-static GBytes *
-fu_vli_usbhub_device_dump_firmware (FuVliUsbhubDevice *self, gsize bufsz, GError **error)
+GBytes *
+fu_vli_usbhub_device_dump_firmware (FuVliUsbhubDevice *self,
+				    guint32 address,
+				    gsize bufsz,
+				    GError **error)
```
superm1: Likewise with this

plugins/vli-usbhub/fu-vli-usbhub-device.c
```diff
@@ -691,6 +696,65 @@ fu_vli_usbhub_device_probe (FuDevice *device, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_vli_usbhub_device_setup_children (FuVliUsbhubDevice *self, GError **error)
+{
+	FuVliUsbhubPdHdr hdr = { 0x0 };
+	g_autoptr(FuDevice) dev = fu_device_new ();
```
superm1: This looks like a memory leak that you set dev twice

plugins/vli-usbhub/fu-vli-usbhub-device.c
hughsie: I don't think it can be, we need access to `self` for the things it calls.

plugins/vli-usbhub/fu-vli-usbhub-device.c
hughsie: Good catch, thanks!

hughsie:
> So this is actually just a region on the SPI chip the hub uses which is shared with PD controller and you don't actually communicate directly with PD controller.

Correct. It's not how I would do it, but I suppose using the usbhub to flash other sections of the flash is just what Realtek is doing too.

> In this case how do you reset it after firmware flash?

That's a very good question. I'll ask VIA.

> Stop it for the duration of firmware flash so that you don't crash it?

I believe the flash is mirrored into SRAM at power-up.

plugins/vli-usbhub/fu-vli-usbhub-device.c
superm1: Ah Yeah I see.  Maybe rename the function to indicate it's operating on a common SPI in some way.

hughsie: Something like `fu_vli_usbhub_device_spi_erase`?

superm1:
> That's a very good question. I'll ask VIA.

If this is not possible and it needs a hard power cycle, then I suppose this should get a flag of `FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE`.  And a flag to notify the user they'll need to replug manually (ugh `FWUPD_DEVICE_FLAG_NEEDS_BOOTLOADER`). or set an error and `FWUPD_ERROR_NEEDS_USER_ACTION`

> I believe the flash is mirrored into SRAM at power-up.

OK, that's good if so.

plugins/vli-usbhub/fu-vli-usbhub-device.c
superm1: ya

hughsie:
> then I suppose this should get a flag of FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE

Agreed, thanks.

> or set an error and FWUPD_ERROR_NEEDS_USER_ACTION

According to VIA they are never going to issue a PD update without a corresponding usbhub update, which means I think that we'd update the child first (PD), and then the parent (usbhub) and then the parent would do `fu_vli_usbhub_device_attach()` which causes a hub reset and for the firmware to be reloaded. That's assuming `FWUPD_DEVICE_FLAG_INSTALL_PARENT_FIRST` is not set...

I guess the other way to do this is rely on `fu_plugin_composite_cleanup()` like you do in the dell plugin. That might be cleaner all round...

superm1:
I think I'd lean on composite cleanup.

plugins/vli-usbhub/fu-vli-usbhub-pd-device.c
```diff
@@ -0,0 +1,206 @@
...
+	return TRUE;
+}
+
+static void
+fu_vli_usbhub_pd_device_init (FuVliUsbhubPdDevice *self)
+{
+	fu_device_add_icon (FU_DEVICE (self), "audio-card");
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_CAN_VERIFY_IMAGE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE);
```
superm1: well now that you're doing a composite cleanup that will cause the reset I don't think this is true anymore

plugins/vli-usbhub/fu-plugin-vli-usbhub.c
```diff
@@ -18,4 +19,35 @@ fu_plugin_init (FuPlugin *plugin)
...
+			     GError **error)
+{
+	g_autoptr(FuDeviceLocker) locker = NULL;
+	g_autoptr(FuDevice) parent = fu_plugin_vli_usbhub_get_parent (devices);
+	if (parent == NULL)
+		return TRUE;
+	locker = fu_device_locker_new (parent, error);
+	if (locker == NULL)
+		return FALSE;
+	return fu_device_attach (parent, error);
```
superm1: What happens when both hub and PD are in the update?  Won't this call attach twice?  Once for the hub and then once for the plugin.

superm1: (So my point was going to be that the functionality should move out of attach and into composite cleanup)

hughsie: >  Won't this call attach twice

No, the device isn't marked as BOOTLOADER and so the attach only gets run manually by the composite cleanup. That's the reason `fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);` was originally included.

plugins/vli-usbhub/fu-vli-usbhub-pd-device.c
hughsie: Ack, got it, thanks.

plugins/vli-usbhub/fu-plugin-vli-usbhub.c
superm1: Got it thanks

PR #1561, opened 2019-11-14
Add support for matching <firmware> requirements on device parents

hughsie:
For composite devices you might want to restrict the child device in respect
to the parent, for instance requiring the parent to have greater than a specific
version number.

The other useful thing to use this for is checking if the parent has a specific
GUID (of any version) which allows us to match against the common VID&PID
instance IDs. This would allow us to restrict a generic child device update to
a specific OEM vendor parent.

This is specified as <firmware depth="1" ...> to match the parent device and
<firmware depth="2" ...> to match the grandparent device.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1079,11 +1079,31 @@ static gboolean
 fu_engine_check_requirement_firmware (FuEngine *self, XbNode *req,
 				      FuDevice *device, GError **error)
 {
+	guint64 depth;
+	g_autoptr(FuDevice) device_actual = g_object_ref (device);
 	g_autoptr(GError) error_local = NULL;
+
+	/* look at the parent device */
+	depth = xb_node_get_attr_as_uint (req, "depth");
+	if (depth != G_MAXUINT64) {
+		for (guint64 i = 0; i < depth; i++) {
+			FuDevice *device_tmp = fu_device_get_parent (device_actual);
+			if (device_actual == NULL) {
+				g_set_error (error,
+					     FWUPD_ERROR,
+					     FWUPD_ERROR_NOT_SUPPORTED,
+					     "No parent device for %s "
+					     "(%" G_GUINT64_FORMAT "/%" G_GUINT64_FORMAT ")",
+					     fu_device_get_name (device_actual), i, depth);
+				return FALSE;
+			}
+			g_set_object (&device_actual, device_tmp);
+		}
+	}
 
 	/* old firmware version */
 	if (xb_node_get_text (req) == NULL) {
-		const gchar *version = fu_device_get_version (device);
+		const gchar *version = fu_device_get_version (device_actual);
 		if (!fu_engine_require_vercmp (req, version, &error_local)) {
 			if (g_strcmp0 (xb_node_get_attr (req, "compare"), "ge") == 0) {
 				g_set_error (error,
@@ -1105,7 +1125,7 @@ fu_engine_check_requirement_firmware (FuEngine *self, XbNode *req,
 
 	/* bootloader version */
 	if (g_strcmp0 (xb_node_get_text (req), "bootloader") == 0) {
-		const gchar *version = fu_device_get_version_bootloader (device);
+		const gchar *version = fu_device_get_version_bootloader (device_actual);
 		if (!fu_engine_require_vercmp (req, version, &error_local)) {
 			if (g_strcmp0 (xb_node_get_attr (req, "compare"), "ge") == 0) {
 				g_set_error (error,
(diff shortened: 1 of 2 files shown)


src/fu-engine.c
```diff
@@ -1079,11 +1079,29 @@ static gboolean
...
+	/* look at the parent device */
+	depth = xb_node_get_attr_as_uint (req, "depth");
+	if (depth != G_MAXUINT64) {
+		for (guint i = 0; i < depth; i++) {
+			device_actual = fu_device_get_parent (device_actual);
+			if (device_actual == NULL) {
+				g_set_error (error,
+					     FWUPD_ERROR,
+					     FWUPD_ERROR_NOT_SUPPORTED,
+					     "No parent device for %s (%u/%" G_GUINT64_FORMAT ")",
```
superm1: I personally think that name is more useful than device id in an error message.

src/fu-engine.c
```diff
@@ -1079,11 +1079,29 @@ static gboolean
 fu_engine_check_requirement_firmware (FuEngine *self, XbNode *req,
 				      FuDevice *device, GError **error)
 {
+	FuDevice *device_actual = device;
+	guint64 depth;
 	g_autoptr(GError) error_local = NULL;
 
+	/* look at the parent device */
+	depth = xb_node_get_attr_as_uint (req, "depth");
+	if (depth != G_MAXUINT64) {
```
superm1: Is this the default return if unset or the extreme? If it's the extreme I think this should be an error condition that it's this value.

src/fu-engine.c
```diff
@@ -1079,11 +1079,29 @@ static gboolean
...
+		for (guint i = 0; i < depth; i++) {
```
superm1: Doesn't this need to be guint64 to make sure you compare same types?

src/fu-engine.c
hughsie: It's the value for unset.

src/fu-engine.c
hughsie: Good catch, although I think it's unlikely to have more than 4294967295 nested children :)

src/fu-engine.c
hughsie: Sure, that works too. Fixed.

src/fu-engine.c
superm1: I sure hope not 😆

hughsie:
Actually, this needs more work to do the GUID matching.

hughsie:
Okay, this one should work. Self tests FTW.

src/fu-engine.c
```diff
@@ -1149,37 +1169,50 @@ fu_engine_check_requirement_firmware (FuEngine *self, XbNode *req,
...
 		const gchar *guid = xb_node_get_text (req);
 		const gchar *version;
-		g_autoptr(FuDevice) device2 = NULL;
 
 		/* find if the other device exists */
-		device2 = fu_device_list_get_by_guid (self->device_list, guid, error);
-		if (device2 == NULL)
-			return FALSE;
+		if (depth == G_MAXUINT64) {
+			device_actual = fu_device_list_get_by_guid (self->device_list, guid, error);
```
superm1: doesn't this need to be a g_set_object call to since you took a reference above?

hughsie: Good catch! Fixed.

src/fu-engine.c
```diff
@@ -1079,11 +1079,31 @@ static gboolean
...
 {
+	guint64 depth;
+	g_autoptr(FuDevice) device_actual = g_object_ref (device);
 	g_autoptr(GError) error_local = NULL;
 
+	/* look at the parent device */
+	depth = xb_node_get_attr_as_uint (req, "depth");
+	if (depth != G_MAXUINT64) {
+		for (guint64 i = 0; i < depth; i++) {
+			FuDevice *device_tmp = fu_device_get_parent (device_actual);
```
superm1: I think this might need to be `g_autoptr` and take a reference too since you're assigning it to another `g_autoptr`

hughsie: Nah, `fu_device_get_parent()` is `transfer-none`

superm1: But then you do a `g_set_object` later, it's still OK?  I'm worried on cleanup you destroy the parent.

hughsie: `g_set_object()` clears the first parameter if non-NULL, then refs the second parameter, saving it into the first. It's like doing:

    if (device_actual != NULL)
        g_object_unref (device_actual);
    device_actual = g_object_ref (device_tmp)

...but as one atomic thing.

superm1: Ah yeah - thanks for clarifying.

PR #1565, opened 2019-11-15
trivial: Detect the USB version automatically

hughsie:
Unfortunately we can't remove the usb2 and usb3 custom flags as we don't yet
depend on a new enough GUsb version.

### DIFF
--- a/src/fu-usb-device.c
+++ b/src/fu-usb-device.c
@@ -373,6 +373,30 @@ fu_usb_device_get_platform_id (FuUsbDevice *self)
 }
 
 /**
+ * fu_usb_device_get_spec:
+ * @self: A #FuUsbDevice
+ *
+ * Gets the string USB revision for the device.
+ *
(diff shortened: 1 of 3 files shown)


src/fu-usb-device.c
```diff
@@ -372,6 +372,30 @@ fu_usb_device_get_platform_id (FuUsbDevice *self)
...
+ * Gets the string USB revision for the device.
+ *
+ * Return value: a specification revision in BCD format, or 0x0 if not supported
+ *
+ * Since: 1.3.4
+ **/
+guint16
+fu_usb_device_get_spec (FuUsbDevice *self)
+{
+#if G_USB_CHECK_VERSION(0,3,1)
```
superm1: I think it might be good for our future selves to leave a comment here that custom flags used in `fu-vli-usbhub-device.c` go away when bumping this version.

hughsie: I'm hoping to do a new release soonish and raise the fwupd dep.

PR #1567, opened 2019-11-18
uefi: fall back to `ID_LIKE` when the path for `ID` doesn't exist (Fi…

superm1:
…xes:# 1566)

Some derivative distributions re-use bootloader paths from their
upstream.  When this happens the current logic to look for the `ID`
key in `/etc/os-release` doesn't work properly.

Adjust the logic to:
1) Use `ID`
2) Test the path exists.  If so, use it.
3) If it doesn't use `ID_LIKE`.
4) Test if that path exists, if so use it.
5) If that path doesn't exist, return the key from `ID`
6) The plugin will make this path.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-uefi-common.c
+++ b/plugins/uefi/fu-uefi-common.c
@@ -242,12 +242,15 @@ fu_uefi_get_esrt_entry_paths (const gchar *esrt_path, GError **error)
 }
 
 gchar *
-fu_uefi_get_esp_path_for_os (const gchar *esp_path)
+fu_uefi_get_esp_path_for_os (const gchar *base)
 {
-	const gchar *os_release_id = NULL;
 #ifndef EFI_OS_DIR
+	const gchar *os_release_id = NULL;
+	const gchar *id_like_id;
+	g_autofree gchar *esp_path = NULL;
 	g_autoptr(GError) error_local = NULL;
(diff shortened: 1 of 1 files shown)


plugins/uefi/fu-uefi-common.c
```diff
@@ -255,10 +256,23 @@ fu_uefi_get_esp_path_for_os (const gchar *esp_path)
 	}
 	if (os_release_id == NULL)
 		os_release_id = "unknown";
+	esp_path = g_build_filename (base, "EFI", os_release_id, NULL);
+	if (!g_file_test (esp_path, G_FILE_TEST_IS_DIR)) {
+		const gchar *id_like_id = g_hash_table_lookup (os_release, "ID_LIKE");
+		if (id_like_id != NULL) {
+			g_autofree gchar* id_like_path = g_build_filename (base, "EFI", id_like_id, NULL);
+			if (g_file_test (esp_path, G_FILE_TEST_IS_DIR)) {
+				free (esp_path);
```
hughsie: `g_free`

hughsie: This all seems a bit clumsy tbh..

superm1: Yeah you're right.  I  think it was leaking memory too.  I just reworked it some.

PR #1570, opened 2019-11-19
Coreboot fixes

PatrickRudolph:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [x] Documentation

### DIFF
--- a/plugins/coreboot/coreboot.quirk
+++ b/plugins/coreboot/coreboot.quirk
@@ -1,0 +1,2 @@
+[SmbiosManufacturer=LENOVO]
+CorebootVersionQuirks = lenovo-cbet-prefix
--- a/plugins/coreboot/fu-coreboot-common.c
+++ b/plugins/coreboot/fu-coreboot-common.c
@@ -11,20 +11,37 @@
 
 #include "fu-plugin-coreboot.h"
 
+/**
+ * FU_QUIRKS_COREBOOT_VERSION:
+ * @key: The SMBIOS manufacturer name
+ * @value: One of the following: "lenovo-cbet-prefix"
+ *
+ * "lenovo-cbet-prefix" quirk:
+ * The thinkpad_acpi kernel module requires a specific pattern
+ * in the DMI version string. To satisfy those requirements
+ * coreboot adds the CBETxxxx prefix to the DMI version string
+ * on all Lenovo devices. The prefix isn't present in the
+ * version string found in coreboot tables, or on other
+ * coreboot enabled devices.
+ *
+ * Since: 1.3.5
+ */
+#define	FU_QUIRKS_COREBOOT_VERSION	"CorebootVersionQuirks"
+#define	FU_QUIRK_CBET_PREFIX		"lenovo-cbet-prefix"
+
 /* Tries to convert the coreboot version string to a triplet string.
  * Returns NULL on error. */
 gchar *
 fu_plugin_coreboot_version_string_to_triplet (const gchar *coreboot_version,
 					      GError **error)
 {
-	guint cb_version = 0;
 	guint cb_major = 0;
 	guint cb_minor = 0;
 	guint cb_build = 0;
 	gint rc;
 
-	rc = sscanf (coreboot_version, "CBET%u %u.%u-%u", &cb_version, &cb_major,
-			&cb_minor, &cb_build);
+	rc = sscanf (coreboot_version, "%u.%u-%u", &cb_major, &cb_minor, &cb_build);
+
 	if (rc < 0) {
 		g_set_error (error,
 			     G_IO_ERROR,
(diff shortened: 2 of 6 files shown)


plugins/coreboot/fu-coreboot-common.c
```diff
@@ -23,8 +23,20 @@ fu_plugin_coreboot_version_string_to_triplet (const gchar *coreboot_version,
 	guint cb_build = 0;
 	gint rc;
 
-	rc = sscanf (coreboot_version, "CBET%u %u.%u-%u", &cb_version, &cb_major,
-			&cb_minor, &cb_build);
+	if (g_str_has_prefix (coreboot_version, "CBET")) {
+		/*
+		 * The thinkpad_acpi kernel module requires a specifc pattern
```
hughsie: specific

plugins/coreboot/fu-coreboot-common.c
```diff
@@ -23,8 +23,20 @@ fu_plugin_coreboot_version_string_to_triplet (const gchar *coreboot_version,
...
+		 * in the DMI version string. To satisfy those requirements
+		 * coreboot adds the CBETxxxx prefix to the DMI version string
+		 * on all Lenovo devices. The prefix isn't present in the
+		 * version string found in coreboot tables, or on other
+		 * coreboot enabled devices.
+		 */
+		rc = sscanf (coreboot_version, "CBET%u %u.%u-%u", &cb_version, &cb_major,
+			     &cb_minor, &cb_build);
```
hughsie: could we just do:

    if (g_string_has_prefix (coreboot_version, "CBET"))
        coreboot_version += 4

?

plugins/coreboot/fu-coreboot-common.c
```diff
@@ -23,8 +23,20 @@ fu_plugin_coreboot_version_string_to_triplet (const gchar *coreboot_version,
 	guint cb_build = 0;
 	gint rc;
 
-	rc = sscanf (coreboot_version, "CBET%u %u.%u-%u", &cb_version, &cb_major,
-			&cb_minor, &cb_build);
+	if (g_str_has_prefix (coreboot_version, "CBET")) {
```
superm1: Should this whole block be guarded by a custom flag and paired with a quirk perhaps?

plugins/coreboot/fu-coreboot-common.c
superm1: That seems like a sane approach too

hughsie:
@PatrickRudolph I was planning another upstream release on Friday, although I don't think these changes should block the release.

PatrickRudolph:
I'm preparing a better solution, but having them ready till Friday might not work out for me.

hughsie:
No worries, releases are cheap. No stress dude.

plugins/coreboot/fu-coreboot-common.c
PatrickRudolph: done

plugins/coreboot/fu-coreboot-common.c
PatrickRudolph: done, I skipped 9 chars as the prefix is 8 + 1 whitespace

plugins/coreboot/fu-coreboot-common.c
PatrickRudolph: added a coreboot.quirk

plugins/coreboot/coreboot.quirk
```diff
@@ -0,0 +1,2 @@
+[SmbiosManufacturer=LENOVO]
+VersionFormat = lenovo-thinkpad-apci-prefixed
```
superm1: @hughsie  do you think this has any unexpected effects outside of the coreboot plugin?

hughsie: I think calling it VersionFormat might be asking for trouble. Perhaps `VersionPrefix`?

superm1: Specifically that version format that is used isn't defined.

Would this work as a custom `Flags` key instead?  I think then it would be more logical to me.

superm1: And if it doesn't work as a custom `Flags` key I think this should be a key that is coreboot specific (IE something like `CorebootVersionQuirks = lenovo-cbet-prefix`)

PatrickRudolph: I wasn't aware that quirks are shared across plugins.
Can I simply add custom key value pairs like this?:
> [SmbiosManufacturer=LENOVO]
> CorebootVersionQuirks = lenovo-cbet-prefix

hughsie: Yup, that works, thanks.

PR #1573, opened 2019-11-20
Wip/hughsie/uefi

hughsie:


plugins/uefi/fu-uefi-bootmgr.c
```diff
@@ -176,7 +176,8 @@ fu_uefi_setup_bootnext_with_dp (const guint8 *dp_buf, guint8 *opt, gssize opt_si
 	/* create a new one */
 	} else {
 		g_autofree gchar *boot_next_name = NULL;
-		for (guint16 value = 0; value < G_MAXUINT16; value++) {
+		/* some firmware doesn't expect Boot0000 to be futzed with */
+		for (guint16 value = 1; value < G_MAXUINT16; value++) {
```
superm1: ooh I don't know about this.  Think about this scenario:

1. Start with Windows at `Boot0000`
2. Install Linux, get the shim entry at `Boot0001`
3. You blew away the Windows partition and something cleans up stale entries (I think BDS is allowed to do this) - `Boot0000` goes away
4. fwupd runs a FW update.  Why shouldn't `Boot0000` be picked here?

hughsie: My suspicion is that Boot0000 is being treated differently by the legacy Lenovo firmware. In your example, it's perfectly fine that there's no Boot0000 -- "holes" are allowed, and we'd just choose Boot0002 for fwupdx64.efi

superm1: Well I happen to have a system here in that exact state I proposed where `Boot0000` is `Linux-Firmware-Updater` (not in my intentional creation; just from regular fwupd usage).  It is dual booted and this is how it turned out.

```
# sudo efibootmgr -v
BootCurrent: 0001
Timeout: 2 seconds
BootOrder: 0001,0000
Boot0000* Linux Firmware Updater	HD(1,GPT,e4f6a002-1bfd-4b3f-97ea-c30c8179b20d,0x800,0x400000)/File(\EFI\ubuntu\shimx64.efi)\.f.w.u.p.d.x.6.4...e.f.i...
Boot0001* ubuntu	HD(1,GPT,e4f6a002-1bfd-4b3f-97ea-c30c8179b20d,0x800,0x400000)/File(\EFI\ubuntu\shimx64.efi)
Boot0002* Windows Boot Manager	HD(1,GPT,e4f6a002-1bfd-4b3f-97ea-c30c8179b20d,0x800,0x400000)/File(\EFI\Microsoft\Boot\bootmgfw.efi)WINDOWS.........x...B.C.D.O.B.J.E.C.T.=.{.9.d.e.a.8.6.2.c.-.5.c.d.d.-.4.e.7.0.-.a.c.c.1.-.f.3.2.b.3.4.4.d.4.7.9.5.}....................
```

This would cause a new undefined behavior on any system that "already" has LFU  at `Boot0000` because it wouldn't get re-used now right?

hughsie: > This would cause a new undefined behavior on any system that "already" has LFU at Boot0000 because it wouldn't get re-used now right?

No, this "starting from 1" index check is only in the case where we didn't find an existing `Linux Firmware Updater` entry.

superm1: > My suspicion is that Boot0000 is being treated differently by the legacy Lenovo firmware. In your example, it's perfectly fine that there's no Boot0000 -- "holes" are allowed, and we'd just choose Boot0002 for fwupdx64.efi

I would feel a lot more comfortable if we changed this behavior only for a quirk on Lenovo systems (similar to how Lenovo gets Linux-Firmware-Updater and everyone else gets Linux FIrmware Updater)

hughsie: Yes, that makes sense too. The thing is, I'm not 100% sure the Boot000 blacklisting is the actual root of this bug. I'd be fine to leave that last patch out for the next release and see if the bitmap cleanup fixes the bug on it's own.

superm1: OK the other stuff looks fine.

PR #1578, opened 2019-11-22
fwupdtool: don't show a warning when calling `get-plugins`

superm1:
The plugins are not opened for this function so of course they
won't set their build hashes.

Fixes errors like this:
`uefi should call fu_plugin_set_build_hash()`

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-plugin.c
+++ b/src/fu-plugin.c
@@ -107,6 +107,23 @@ typedef gboolean	 (*FuPluginUdevDeviceAddedFunc)	(FuPlugin	*self,
 							 GError		**error);
 
 /**
+ * fu_plugin_is_open:
+ * @self: A #FuPlugin
+ *
+ * Determines if the plugin is opened
+ *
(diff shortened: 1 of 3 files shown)


src/fu-plugin.c
```diff
@@ -106,6 +106,23 @@ typedef gboolean	 (*FuPluginUdevDeviceAddedFunc)	(FuPlugin	*self,
...
+ * @self: A #FuPlugin
+ *
+ * Determines if the plugin is opened
+ *
+ * Returns: TRUE for opened, FALSE for not
+ *
+ * Since: 1.3.5
+ **/
+gboolean
+fu_plugin_get_opened (FuPlugin *self)
```
hughsie: isn't _is_open() the usual suffix?

superm1: Corrected

PR #1581, opened 2019-11-23
Ensure the _get_type() gets the same ABI version as the constructor

hughsie:
This was also causing the _get_type() symbol to be ignored for objects without
any class methods except constructors.


superm1:
Ah ha, thanks for tracking that one down! Lgtm.

PR #1582, opened 2019-11-23
Allow compiling GUsb as a subproject

hughsie:
This allows us to get closer to compiling on FreeBSD.


hughsie:
If this works, and makes sense, I'll do a tag of gusb we can use rather than a wip branch.

superm1:
Yeah no concerns to this if it works indeed.

PR #1583, opened 2019-11-23
Wip/superm1/fixups

superm1:
Just pull a few trivial things out of the win32 branch so they don't get lost in rebases

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -4373,6 +4373,7 @@ fu_engine_load_plugins (FuEngine *self, GError **error)
 	const gchar *fn;
 	g_autoptr(GDir) dir = NULL;
 	g_autofree gchar *plugin_path = NULL;
+	g_autofree gchar *suffix = g_strdup_printf (".%s", G_MODULE_SUFFIX);
 
 	/* search */
 	plugin_path = fu_common_get_path (FU_PATH_KIND_PLUGINDIR_PKG);
@@ -4386,7 +4387,7 @@ fu_engine_load_plugins (FuEngine *self, GError **error)
 		g_autoptr(GError) error_local = NULL;
 
 		/* ignore non-plugins */
-		if (!g_str_has_suffix (fn, ".so"))
+		if (!g_str_has_suffix (fn, suffix))
 			continue;
 
 		/* is blacklisted */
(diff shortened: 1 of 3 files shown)


src/fu-engine.c
```diff
@@ -4380,7 +4380,7 @@ fu_engine_load_plugins (FuEngine *self, GError **error)
 		g_autoptr(GError) error_local = NULL;
 
 		/* ignore non-plugins */
-		if (!g_str_has_suffix (fn, ".so"))
+		if (!g_str_has_suffix (fn, G_MODULE_SUFFIX))
```
hughsie: Actually, thinking about this -- I think we need to add back the "." -- e.g. `"." G_MODULE_SUFFIX" otherwise we might match directories with the name `gutso` for example.

src/fu-udev-device.c
```diff
@@ -103,32 +103,6 @@ fu_udev_device_get_sysfs_attr_as_uint8 (GUdevDevice *udev_device, const gchar *n
 	return tmp;
 }
 
-static void
-fu_udev_device_dump_internal (GUdevDevice *udev_device)
```
hughsie: This is kinda useful for plugin development. I wonder if we should put this into the default `to_string()` implementation, perhaps covered by a `g_getenv("FU_UDEV_DEVICE_DEBUG")` or something.

superm1: Yeah I figured it was useful for something like that, but I didn't want to export a symbol to the new library if nothing is using it.

src/fu-engine.c
superm1: Good point

src/fu-udev-device.c
```diff
@@ -108,29 +108,40 @@ fu_udev_device_get_sysfs_attr_as_uint8 (GUdevDevice *udev_device, const gchar *n
...
-		return;
-	fu_udev_device_dump_internal (priv->udev_device);
+	FuDeviceClass *klass = FU_DEVICE_GET_CLASS (device);
+
+	if (g_getenv ("FU_UDEV_DEVICE_DEBUG") != NULL)
+		fu_udev_device_verbose (device, idt, str);
+
+	/* subclassed */
+	if (klass->to_string != NULL)
+		klass->to_string (device, idt + 1, str);
```
hughsie: That's going to be an infinite loop. Just drop those three lines as the subclass won't need the default version.

superm1: Makes sense now that I think about it.  I was thinking that devices subclassing would be able to have their own to_string's but they just override this one.

Fixed.

PR #1586, opened 2019-11-23
Switch the plugin interface to a shared library

superm1:
This is mostly intended to allow easier usage with Windows (which doesn't like if you can't resolve symbols at link time).  However I think it also has a nice side effect on the Linux side that the interface of what plugins can use, what clients can use and how plugins talk to the daemon gets an ABI.  Implementing it required cleaning up a lot of the documentation too so that an accurate version script could be built.  Hopefully no mistakes there.

It is based on top of the stuff at `wip/superm1/fixups`.
 
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
I think we probably want to split up src/fu-self-test.c too -- and putting anything small and self contained into libfwupdplugin/fu-self-test.c

superm1:
> I think we probably want to split up src/fu-self-test.c too -- and putting anything small and self contained into libfwupdplugin/fu-self-test.c

OK, just moved what I thought made sense over.

libfwupdplugin/fwupdplugin.h
```diff
@@ -0,0 +1,47 @@
...
+#include <libfwupdplugin/fu-firmware.h>
+#include <libfwupdplugin/fu-firmware-common.h>
+#include <libfwupdplugin/fu-firmware-image.h>
+#include <libfwupdplugin/fu-hwids.h>
+#include <libfwupdplugin/fu-history.h>
+#include <libfwupdplugin/fu-ihex-firmware.h>
+#include <libfwupdplugin/fu-io-channel.h>
+#include <libfwupdplugin/fu-plugin.h>
+#include <libfwupdplugin/fu-plugin-vfuncs.h>
+#include <libfwupdplugin/fu-progressbar.h>
```
hughsie: I don't think this should be exported in fwupdplugin.so - I think it probably out to be a static library built in src and just link_with'd in the places that need it. Maybe fwupdnoinst.a or something.

libfwupdplugin/fwupdplugin.h
```diff
@@ -0,0 +1,47 @@
...
+#include <libfwupdplugin/fu-quirks.h>
+#include <libfwupdplugin/fu-smbios.h>
+#include <libfwupdplugin/fu-srec-firmware.h>
+#include <libfwupdplugin/fu-test.h>
```
hughsie: Same solution with fwupdnoinst.a I think for fu-test.h

libfwupdplugin/fwupdplugin.h
```diff
@@ -0,0 +1,47 @@
...
+#include <libfwupdplugin/fu-udev-device.h>
+#include <libfwupdplugin/fu-usb-device.h>
+
+#ifndef FWUPD_DISABLE_DEPRECATED
+#include <libfwupdplugin/fu-deprecated.h>
```
hughsie: do we need this? I'd be included to just mark everything in this library as unstable for now.

libfwupdplugin/fwupdplugin.h
superm1: What do you think about splitting it up to a device attribute for percentage and using the existing device status attribute? Then we can keep it only as part of daemon and tools.

hughsie: I think the status property makes a lot of sense, although I don't know how hard it is to have multiple FuDevices updating one progressbar. I think the issue is also that other stuff uses it, e.g. plugins/dfu/dfu-tool.c -- although we could just nuke it from there as the DFU tool is less and less important to me given all the recent changes.

libfwupdplugin/fwupdplugin.h
superm1: Well even if we treat as unstable we would still do so version bumps when it went up right?

hughsie: I think so. I think we have to commit to semi-stable plugin API at some point -- more than just me and you are writing plugins now.

PR #1587, opened 2019-11-24
trivial: Use prefixes for DATADIR in config.h for portability

hughsie:
DATADIR is an enumerated type in MinGW, and the other names are very generic.


superm1:
The commit message is a little inaccurate, I suppose you discovered the need for more than `DATADIR` or just wanted to be cautious.  Content looks fine.

hughsie:
Just being consistent and cautious really.

PR #1590, opened 2019-11-25
synaptics-rmi: Use the build ID as the version number to match the vendor tool

hughsie:


hughsie:
> does this have any impact to the protocol matching version format on LVFS?

Yes, I'll switch it to plan when we tag the next release.

superm1:
> Yes, I'll switch it to plan when we tag the next release.

OK hopefully no firmware is released to LVFS between now and then.

PR #1594, opened 2019-11-25
Set the protocol per-device not per-plugin

hughsie:
Some plugins have devices with more than one protocol. Logically the protocol
belongs to the device, not the plugin, and in the future we could use this to
further check firmware that's about to be deployed.

This is also not exported into libfwupd (yet?) as it's remains a debug-feature
only -- protocols are not actually required for devices to be added.

### DIFF
--- a/plugins/dell-dock/fu-dell-dock-i2c-tbt.c
+++ b/plugins/dell-dock/fu-dell-dock-i2c-tbt.c
@@ -280,8 +280,10 @@ fu_dell_dock_tbt_finalize (GObject *object)
 }
 
 static void
-fu_dell_dock_tbt_init (FuDellDockTbt *device)
-{}
+fu_dell_dock_tbt_init (FuDellDockTbt *self)
+{
+	fu_device_set_protocol (FU_DEVICE (self), "com.intel.thunderbolt");
+}
 
 static void
 fu_dell_dock_tbt_class_init (FuDellDockTbtClass *klass)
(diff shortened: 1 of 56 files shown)


plugins/dell-dock/fu-dell-dock-i2c-tbt.c
```diff
@@ -280,8 +280,10 @@ fu_dell_dock_tbt_finalize (GObject *object)
 }
 
 static void
-fu_dell_dock_tbt_init (FuDellDockTbt *device)
-{}
+fu_dell_dock_tbt_init (FuDellDockTbt *self)
+{
+	fu_device_set_protocol (FU_DEVICE (self), "com.dell.dock");
+}
```
superm1: This is thunderbolt protocol for this one actually

src/fu-device.c
```diff
@@ -1411,6 +1419,44 @@ fu_device_set_logical_id (FuDevice *self, const gchar *logical_id)
...
+	return priv->protocol;
+}
+
+/**
+ * fu_device_set_protocol:
+ * @self: A #FuDevice
+ * @protocol: a defined protocol ID, e.g. `org.hughski.colorhug`
+ *
+ * Sets the protocol ID on the device. This may be used in the future to verify
+ * firmware files.
```
superm1: I don't think future is a good thing to mention here. We'll prob forget to update this comment since the code won't be touched.

PR #1599, opened 2019-11-26
Always use the more portable 3-arg form for open()

hughsie:


plugins/thunderbolt-power/fu-plugin-thunderbolt-power.c
```diff
@@ -187,7 +188,7 @@ fu_plugin_thunderbolt_power_kernel_force_power (FuPlugin *plugin, gboolean enabl
 		return FALSE;
 	}
 	g_debug ("Setting force power to %d using kernel", enable);
-	fd = g_open (data->force_path, O_WRONLY);
+	fd = g_open (data->force_path, O_WRONLY, S_IWUSR);
```
gicmo: As long as you are not creating the file (`g_creat`, `flags | O_CREAT`), the `mode` gets [ignored](https://github.com/torvalds/linux/blob/9c7db5004280767566e91a33445bf93aa479ef02/fs/open.c#L945), so I think just passing `0` here would be better.

hughsie: Fixed, thanks.

src/fu-io-channel.c
```diff
@@ -16,6 +16,7 @@
 #include <poll.h>
 #endif
 #include <string.h>
+#include <sys/stat.h>
```
superm1: Unneeded include, right?

hughsie: Nah, actually needed for the S_IRWXU define. I can split out this hunk as another commit if that's better.

PR #1600, opened 2019-11-26
trivial: Allow compiling without <fnmatch.h>

hughsie:
This also allows us to add a Win32 implementation if required in the future.

### DIFF
--- a/src/fu-common.c
+++ b/src/fu-common.c
@@ -11,6 +11,10 @@
 #include <gio/gunixinputstream.h>
 #include <glib/gstdio.h>
 
+#ifdef HAVE_FNMATCH_H
+#include <fnmatch.h>
+#endif
+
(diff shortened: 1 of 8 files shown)


src/fu-common.c
```diff
@@ -1623,6 +1627,29 @@ fu_common_realpath (const gchar *filename, GError **error)
...
+/**
+ * fu_common_fnmatch:
+ * @pattern: a glob pattern, e.g. `*foo*`
+ * @str: a string to match against the pattern, e.g. `bazfoobar`
+ *
+ * Matches a string against a glob pattern.
+ *
+ * Return value: %TRUE if the string matched
+ *
+ * Since: 1.2.6
```
superm1: 1.3.5

hughsie: Fixed, thanks.

PR #1601, opened 2019-11-26
trivial: Untangle FuHistory from FuPlugin

hughsie:
This will allow us to export FuPlugin in the future without dragging in the
private fu-history.h too.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -19,6 +19,7 @@
 #ifdef HAVE_UTSNAME_H
 #include <sys/utsname.h>
 #endif
+#include <errno.h>
 
 #include "fwupd-common-private.h"
 #include "fwupd-enums-private.h"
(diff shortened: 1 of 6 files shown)


src/fu-engine.c
```diff
@@ -1613,6 +1614,144 @@ fu_engine_is_running_offline (FuEngine *self)
...
+
+	/* create symlink for the systemd-system-update-generator */
+	rc = symlink (symlink_target, trigger);
+	if (rc < 0) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Failed to create symlink %s to %s: %s",
+			     trigger,
+			     "/var/lib", strerror (errno));
```
gicmo: There is `g_strerror`, which compared to `strerror` is thread-safe. Not sure it is important though.

superm1: This line shouldn't hardcode /var/lib, should look up string to use, no?

src/fu-engine.c
```diff
@@ -1613,6 +1614,144 @@ fu_engine_is_running_offline (FuEngine *self)
...
+ * @release: A #FwupdRelease
+ * @blob_cab: A #GBytes
+ * @flags: #FwupdInstallFlags
+ * @error: A #GError or NULL
+ *
+ * Schedule an offline update for the device
+ *
+ * Returns: #TRUE for success, #FALSE for failure
+ *
+ * Since: 1.3.6
```
superm1: Planning to skip 1.3.5?

hughsie: No, just need coffee

src/fu-engine.c
hughsie: > There is g_strerror

good catch, thanks.

hughsie: > This line shouldn't hardcode /var/lib

I don't think it matters what the destination is, although I guess it makes sense to chose something sensible.

hughsie: Hah, I already did that, the error message was just wrong. Fixed.

PR #1602, opened 2019-11-26
trivial: Add some optional debugging when hotplugging devices

hughsie:


hughsie:
We've got FWUPD_PROBE_VERBOSE -- I can use that if you'd prefer. I guess it makes sense to avoid having too many knobs to turn.

superm1:
Yeah I think in general if you're looking at probe errorsvor info hotplug or not one variable that is consistent is better.

PR #1604, opened 2019-11-26
trivial: Allow building GCab from a subpackage

hughsie:


subprojects/gcab.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+directory = gcab
+url = https://gitlab.gnome.org/GNOME/gcab.git
+revision = b55268ac1020cd6c033acb52f2e6ae984bf5c9fd
```
superm1: Does this correspond to a tag? Or at least will soon be tagged?

hughsie: It's just a commit; I'm not the gcab maintainer and don't get to choose release dates. I think it's fine really, it's not like it's a WIP branch that could go away.

superm1: Ah okay, yes no problem then. That was my concern it was some sub branch that could go away .

PR #1607, opened 2019-11-26
trivial: Move FuTest out of libfwupdprivate

hughsie:


hughsie:
Actually this doesn't work with the planned `libfwupdplugin` changes. I'm wondering if we can just hardlink the files instead for the few places we need these.

PR #1608, opened 2019-11-26
trivial: Do not include fu-test.c in libfwupdprivate

hughsie:
Use symlinks -- the file is a trivial helper used by various different modules,
and it's overkill to create a private or shared library just for 4 functions.

It also means there are no ordering issues when we split out fwupdplugin.


hughsie:
Of course, the alternative is we move the sane bits of fu-test into fu-common and just export it and stop worrying about it...

superm1:
Is there a concern for clashing symbol name when the plugins load? I do think your other idea is cleaner.

PR #1611, opened 2019-11-27
[WIP]plugin/coreboot: Add sysfs parsers

PatrickRudolph:
Requires Linux to be build with branch:
https://github.com/9elements/linux/tree/google_firmware_generic

Send upstream here:
https://lkml.org/lkml/2019/11/20/500

Provides /dev/mem free access to coreboot tables and CBMEM.

Signed-off-by: Patrick Rudolph <patrick.rudolph@9elements.com>

Type of pull request:
- [ ] Code fix
- [x] Feature
- [x] Documentation

### DIFF
--- a/plugins/coreboot/fu-coreboot-sysfs.c
+++ b/plugins/coreboot/fu-coreboot-sysfs.c
@@ -1,0 +1,145 @@
+/*
+ * Copyright (C) 2019 9elements Agency GmbH <patrick.rudolph@9elements.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <string.h>
+#include <stdio.h>
+
+#include "fu-plugin-coreboot.h"
+
+#define SYSFS_BUS "/sys/bus/coreboot/"
+
+/* Tries to detect the 'coreboot' kernel module presence. */
+gboolean
+fu_plugin_coreboot_sysfs_probe (void)
+{
+	return g_file_test(SYSFS_BUS, G_FILE_TEST_EXISTS|
+				      G_FILE_TEST_IS_DIR);
+}
+
+/* iterates over sysfs directories until the given ID is found. */
+static gchar *
+fu_plugin_coreboot_find_sysfs (const guint id,
+			       const gchar *base_path,
+			       const gchar *extension_path,
+			       GError **error)
+{
+	GDir *dir;
+	const gchar *subdir = NULL;
+	gchar *tmp = NULL;
+	gboolean ret;
+
+	dir = g_dir_open (base_path, 0, error);
+	if (!dir)
+		return NULL;
+
+	while (1) {
+		gchar *fp;
+		guint64 i;
+
+		subdir = g_dir_read_name(dir);
+		if (!subdir)
+			break;
+
+		fp = g_strdup_printf ("%s/%s/%s/id", base_path, subdir,
+				      extension_path);
+
+		ret = g_file_get_contents (fp, &tmp, NULL, NULL);
+		g_free (fp);
+
+		if (!ret) {
+			continue;
+		}
+		i = g_ascii_strtoull (tmp, NULL, 16);
+		g_free (tmp);
+
+		if (id == i)
+			break;
+	};
+
+	g_dir_close (dir);
+
+	if (subdir == NULL) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_ARGUMENT,
+			     "Requested id not found");
+		return NULL;
+	}
+	return g_strdup_printf ("%s/%s/%s/", base_path, subdir, extension_path);
+}
+
+/* Returns the coreboot tables with given tag */
+gchar *
+fu_plugin_coreboot_find_cb_table (const guint tag, gsize *length,
+				  GError **error)
+{
+	gchar *fp;
+	gboolean ret;
+	gchar *path;
+	gchar *tmp;
+
+	path = fu_plugin_coreboot_find_sysfs (tag, SYSFS_BUS "/devices/",
+					      "attributes", error);
+	if (!path)
+		return NULL;
+
+	fp = g_strdup_printf ("%s/data", path);
+	g_free (path);
+
+	ret = g_file_get_contents (fp, &tmp, length, error);
+	g_free (fp);
+
+	if (!ret)
+		return NULL;
+
+	return tmp;
+}
+
+/* Returns the CBMEM buffer with given id */
+gchar *
+fu_plugin_coreboot_find_cbmem (const guint id, gsize *length, goffset *address,
+			       GError **error)
+{
+	gchar *fp;
+	gboolean ret;
+	gchar *tmp;
+	gchar *path;
+
+	path = fu_plugin_coreboot_find_sysfs (id, SYSFS_BUS "/drivers/cbmem/",
+					      "cbmem_attributes", error);
+	if (!path)
+		return NULL;
+
+	if (address != NULL) {
+		/* Store the physical CBMEM buffer address if requested */
+
+		fp = g_strdup_printf ("%s/address", path);
+
+		ret = g_file_get_contents (fp, &tmp, NULL, error);
+		g_free (fp);
+
+		if (!ret) {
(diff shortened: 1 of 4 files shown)


plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+ */
+
+#include "config.h"
+
+#include <string.h>
+#include <stdio.h>
+
+#include "fu-plugin-coreboot.h"
+
+#define SYSFS_BUS "/sys/bus/coreboot/"
```
superm1: This should go into fu-common as a directory to lookup, especially if it's used for unit tests down the road.

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+gboolean
+fu_plugin_coreboot_sysfs_probe (void)
+{
+	return g_file_test(SYSFS_BUS, G_FILE_TEST_EXISTS|
+				      G_FILE_TEST_IS_DIR);
+}
+
+/* iterates over sysfs directories until the given ID is found. */
+static gchar *
+fu_plugin_coreboot_find_sysfs (const guint id,
```
superm1: Is it not possible to rely upon the kernel creating a uevent and udev picking it up instead to find this?

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+
+#include "fu-plugin-coreboot.h"
+
+#define SYSFS_BUS "/sys/bus/coreboot/"
+
+/* Tries to detect the 'coreboot' kernel module presence. */
+gboolean
+fu_plugin_coreboot_sysfs_probe (void)
+{
+	return g_file_test(SYSFS_BUS, G_FILE_TEST_EXISTS|
```
hughsie: space before bracket and pipe please

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+	const gchar *subdir = NULL;
+	gchar *tmp = NULL;
+	gboolean ret;
+
+	dir = g_dir_open (base_path, 0, error);
+	if (!dir)
+		return NULL;
+
+	while (1) {
+		gchar *fp;
```
hughsie: g_autofree

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/* iterates over sysfs directories until the given ID is found. */
+static gchar *
+fu_plugin_coreboot_find_sysfs (const guint id,
+			       const gchar *base_path,
+			       const gchar *extension_path,
+			       GError **error)
+{
+	GDir *dir;
```
hughsie: g_autoptr

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/* Returns the coreboot tables with given tag */
+gchar *
+fu_plugin_coreboot_find_cb_table (const guint tag, gsize *length,
+				  GError **error)
+{
+	gchar *fp;
+	gboolean ret;
+	gchar *path;
```
hughsie: g_autofree

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return g_strdup_printf ("%s/%s/%s/", base_path, subdir, extension_path);
+}
+
+/* Returns the coreboot tables with given tag */
+gchar *
+fu_plugin_coreboot_find_cb_table (const guint tag, gsize *length,
+				  GError **error)
+{
+	gchar *fp;
```
hughsie: g_autofree

plugins/coreboot/fu-coreboot-sysfs.c
```diff
@@ -0,0 +1,145 @@
...
+	gboolean ret;
+	gchar *path;
+	gchar *tmp;
+
+	path = fu_plugin_coreboot_find_sysfs (tag, SYSFS_BUS "/devices/",
+					      "attributes", error);
+	if (!path)
```
hughsie: compare to NULL please

plugins/coreboot/fu-coreboot-sysfs.c
PatrickRudolph: The coreboot kernel module creates virtual devices for every coreboot table entry. Looking at the udev code in fwupd I'm not sure how to set up an event listener on a specific bus. As there's no device file, vendor or product id and the subsystem is none of pci, mmc, usb, hid, ....
Can you give an example how such a code would look like?

hughsie: Maybe. Can you show us a tree of what the devices look like. I don't know if the coreboot tree is going to be suitable for the FuUdevDevice model where each exposed device is backed just by one GUdevDevice.

superm1: We could certainly support a new subsystem for coreboot if we need to too

PatrickRudolph: The virtual devices have attributes that needs to be read:

> /sys/bus/coreboot/devices/.../attributes/data
> /sys/bus/coreboot/devices/.../attributes/id

The cbmem driver (if loaded) creates an additional folder on those virtual devices:

> /sys/bus/coreboot/devices/.../cbmem_attributes/data
> /sys/bus/coreboot/devices/.../cbmem_attributes/id
> /sys/bus/coreboot/devices/.../cbmem_attributes/address

The key is matched what is present in *id*, and then contents of *data* is loaded.

Besides the newly added attributes the virtual device folder is empty.

superm1: I would expect you should be able to pick uevents for those devices and then use the udev path to point to the `data`, `id`, and `address` directories.  I think it should be cleaner than trying to walk the filesystem manually.

PatrickRudolph: Do you have a code example how to pick up the uevents for those devices? I've never worked with uevents before.

hughsie: If there's a valid udev subsystem you can do https://github.com/fwupd/fwupd/blob/master/plugins/optionrom/fu-plugin-optionrom.c#L18 and then something like `fu_plugin_set_device_gtype (plugin, FU_TYPE_OPTIONROM_DEVICE);` -- this makes fwupd create an object of the correct type whenever that udev subsystem is added. https://github.com/fwupd/fwupd/blob/master/plugins/optionrom/fu-optionrom-device.c gives a good example of using `fu_udev_device_get_sysfs_path()` for example.

hughsie:
@PatrickRudolph is there any hardware I can get hold of so i can help with pushing this over the finish line?

PatrickRudolph:
It should work on any coreboot enabled device, however I'm only testing on Intel platforms right now. I currently have no time to update the Linux kernel patches, so this could take a while to be actually useful.

hughsie:
> It should work on any coreboot enabled device

Got any ideas for the cheapest platform I can find on Ebay?

superm1:
I would think modern Chromebooks are your cheapest most _developer friendly_ way to interact with coreboot.

PatrickRudolph:
The cheapest are probably Lenovo Thinkpads (X201/X220/X230), which can be retrofitted with coreboot and they are easy to flash, as the flash chip is right under the keyboard. Those are also the only ones that have some documentation in coreboot as well...

plugins/coreboot/fu-coreboot-sysfs.c
hughsie: I don't think uevents work here.

hughsie:
I've fixed a lot of the issues here and pushed it as wip/coreboot on this repo. I'll close this PR and we can work together on that branch. @PatrickRudolph could you add the things to `wip/coreboot` that actually use the new `fu_plugin_coreboot_find_X()` symbols please. Thanks!

PR #1613, opened 2019-11-28
Wip/hughsie/dlopen once

hughsie:


plugins/test/meson.build
```diff
@@ -26,3 +26,25 @@ shared_module('fu_plugin_test',
...
+  ],
+  include_directories : [
+    root_incdir,
+    fwupd_incdir,
+    fwupdplugin_incdir,
+  ],
+  install : install_dummy,
+  install_dir: plugin_dir,
+  link_with : [
+    fwupd,
```
superm1: Does this actually need to link with fwupd? I would think it just needs fwupdplugin

contrib/debian/fwupd-tests.install
```diff
@@ -4,5 +4,6 @@
 #https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=872458
 usr/share/installed-tests/*
 usr/lib/*/fwupd-plugins-3/libfu_plugin_test.so
+usr/lib/*/fwupd-plugins-3/libfu_plugin_test_invalid_build_hash.so
```
superm1: I'm fine with just making above line test*.so to cover this in the future if we add more

plugins/test/fu-plugin-test-invalid-build-hash.c
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (C) 2016 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-plugin-vfuncs.h"
+#include "fu-hash.h"
```
superm1: Probably don't need fu-hash.h since you're just using invalid

plugins/test/meson.build
```diff
@@ -26,3 +26,25 @@ shared_module('fu_plugin_test',
     plugin_deps,
   ],
 )
+
+shared_module('fu_plugin_test_invalid_build_hash',
+  fu_hash,
```
superm1: Don't think you need fu_hash

contrib/debian/fwupd-tests.install
hughsie: I renamed it `libfu_plugin_invalid.so` to keep the daemon blacklist sort.

plugins/test/fu-plugin-test-invalid-build-hash.c
hughsie: Agree.

superm1:
Squash then lgtm now

PR #1620, opened 2019-11-30
Export the device protocol to the client --verbose output

hughsie:
It turns out this is useful for debugging.

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -36,7 +36,6 @@ typedef struct {
 	gchar				*equivalent_id;
 	gchar				*physical_id;
 	gchar				*logical_id;
-	gchar				*protocol;
 	FuDevice			*alternate;
 	FuDevice			*parent;	/* noref */
 	FuQuirks			*quirks;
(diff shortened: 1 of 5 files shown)


superm1:
I think that this should be part of the `to_string` output for the fwupd device.

libfwupdplugin/fu-device.c
```diff
@@ -1493,9 +1485,8 @@ fu_device_set_logical_id (FuDevice *self, const gchar *logical_id)
 const gchar *
 fu_device_get_protocol (FuDevice *self)
 {
-	FuDevicePrivate *priv = GET_PRIVATE (self);
 	g_return_val_if_fail (FU_IS_DEVICE (self), NULL);
-	return priv->protocol;
+	return fwupd_device_get_protocol (FWUPD_DEVICE (self));
 }
```
superm1: Maybe this whole function should move to `fu-device.h` as a #define to `fwupd_device_get_protocol`?

hughsie: No can do, it's a symbol in fwupdplugin.map now. I also think it might be the kind of thing we want to validate in the future, e.g. having an enumerated list of protocols, or checking that the protocol doesn't change on the FuDevice, for instance.

hughsie:
> I think that this should be part of the to_string output for the fwupd device.

It is, isn't it?

	fwupd_pad_kv_str (str, FWUPD_RESULT_KEY_PROTOCOL, priv->protocol);

If you're testing this, you of course need to be running the new daemon version to export the property...

libfwupdplugin/fu-device.c
superm1: Ah right, okay.

Issue #1618, opened 2019-11-30
fwupdmgr upgrade gives a message that is hard to understand


tsdgeos:
```shell
$ fwupdmgr --version
client version: 1.3.5
compile-time dependency versions
        gusb:   0.3.1
        efivar: 37
daemon version: 1.3.5
```

```shell
$ fwupdmgr get-devices
LC_ALL=en_US.UTF-8 fwupdmgr get-devices 
20EM000QSP
│
├─LITEON LCH-256V2S:
│     Device ID:           5dbeb140337f610d54913c0f43d15fdaa0eafb51
│     Summary:             ATA Drive
│     Current version:     3C87902
│     Vendor:              Intel Corporation
│     Serial Number:       SD0F66157L2TH6110FN0
│     GUIDs:               ae7ea9e0-b497-5b7b-bce1-fc358bdf84ad ← IDE\LITEON_LCH-256V2S_______________________3C87902
│                          f5caef3f-05a8-5bb3-8e25-b03a18619f24 ← IDE\0LITEON_LCH-256V2S_______________________
│                          ad6b029a-e346-545e-a360-fa076902a208 ← LITEON LCH-256V2S
│     Device Flags:        • Internal device
│                          • Updatable
│                          • Requires AC power
│                          • Needs a reboot after installation
│                          • Device is usable for the duration of the update
│   
├─System Firmware:
│     Device ID:           8be4cbd5f33a57bd5ac69de9004c4e5ecc91325f
│     Current version:     0.0.49
│     Minimum Version:     0.0.49
│     Vendor:              LENOVO
│     GUID:                b8d73eb5-feba-42e1-a106-04540b6ae7d6
│     Device Flags:        • Internal device
│                          • Updatable
│                          • Requires AC power
│                          • Needs a reboot after installation
│                          • Cryptographic hash verification is available
│                          • Device is usable for the duration of the update
│   
└─UEFI Device Firmware:
      Device ID:           aff24b50c0a8c292c6c2a81d3827c36d7a86f62a
      Current version:     184.50.3425
      Minimum Version:     184.50.3425
      GUID:                03d3297b-3851-4379-95ad-12e21a96c80a
      Device Flags:        • Internal device
                           • Updatable
                           • Requires AC power
                           • Supported on remote server
                           • Needs a reboot after installation
                           • Device is usable for the duration of the update
    
```

Please answer the following questions:

- Operating system and version:
Archlinux
[...]

PR #1622, opened 2019-12-01
fu-util/fu-tool: Hide output for update commands when nothing to be done (Fixes: #1618)

superm1:
Avoid confusing messages like:
```
No releases found for device: no HWIDs matched 90706264-e399-575b-a9fb-077ead03b9b4
```

Which aren't actually useful to the user.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Other than CI failing, this also has the side effect that these types of messages don't make it through:

```
No upgrades for Integrated Webcam™, current is 1.2.4: 1.2.4=same, 1.2.3=older
```

So I think that the filtering needs to happen in the engine not the client.

PR #1623, opened 2019-12-01
fu-engine: clarify which device doesn't have matching FW (Fixes: #1618)

superm1:
Alternative approach to #1622 
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I think we should probably match the error code in the frontend and do something translated...

PR #1624, opened 2019-12-01
trivial: libfwupdplugin: clean up a build warning

superm1:
```
libfwupdplugin/fu-device.c:1960: Warning: FwupdPlugin: fu_device_write_firmware: unknown parameter 'fw' in documentation comment, should be 'firmware'
```

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I don't think this is right. The warning is due to the ->write_firmware and write_firmare() aliasing. I think we should instead somehow add the vfunc to the (ignore) list so we don't include it in the GIR.

PR #1625, opened 2019-12-01
Export the raw device version to the client --verbose output

hughsie:
It turns out this is really useful for debugging UEFI firmware updates which
typically use the 0x-prefixed version numbers without a version format.


superm1:
To future proof this what do you think about making it guint64. Hard to predict what future devices version string comes from.

hughsie:
> what do you think about making it guint64

I also pondered this. I didn't have any examples on where a version is represented as a uint64_t. "Enforcing" the uint32 type means we can always print the version with the right number of leading zeros without it looking uint64 crazy.

superm1:
Isn't EFI ESRT version 64 bit?

hughsie:
> Isn't EFI ESRT version 64 bit?

Nah, https://docs.microsoft.com/en-us/windows-hardware/drivers/bringup/esrt-table-definition says 4 bytes, so uint32

superm1:
Huh, I guess I never looked closely enough. Okay then I guess this is fine.

Issue #1336, opened 2019-09-05
Add a "Generic" TPM2 plugin


superm1:
Just as a follow up to #1331 so this information isn't lost.

> There is a documented upgrade process, see the TPM 2.0 specification, Part 3, Section 27. I don't have any experience with these commands, but it seems you need to supply a signature of the new firmware to TPM2_FieldUpgradeStart, after that is validated successfully you can upload the firmware with TPM2_FieldUpgradeData.

https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-3-Commands-01.38.pdf#page=321

For the basic enumeration stuff:
* Ideally GUID strings would be built like this:
`TPM-$MODE-$MANUFACTURER-$VENDORSTR`

* Version strings can be read from fixed properties and are 4 parts.

* Updatability should be determined by TPM Ownership

superm1:
{CC @diabonas}

I had some internal discussion and it's not likely that any TPM vendor will support using a generic fwupd plugin due to certification requirements.

At least based on that, I would say a generic TPM2 plugin would likely be "information only".

hughsie:
> I would say a generic TPM2 plugin would likely be "information only".

I still think that would be a useful thing to have...

superm1:
OK, i'll re-open.

I think if we land something like this it will need to have something similar to how synaptics mst and dell-dock work where they build the device ID's the same but dell-dock is preferred.

hughsie:
Fixed with addc0211352c2193d5d7b3fbb224bf978cbecba7

PR #1628, opened 2019-12-02
Add a new plugin that can parse the TPM event log

hughsie:
Some devices do not have a stable PCR0 for the same firmware version, and I'd
like to collect the TPM event log for affected machines to debug why.


superm1:
Doesn't this only work with TPM1.2?  I didn't think the kernel exported TPM2 event log that way.  You probably should use the tpm2-tss to get TPM2.0 (which is way more common today)

plugins/meson.build
```diff
@@ -12,6 +12,7 @@ subdir('solokey')
 subdir('synaptics-cxaudio')
 subdir('synaptics-prometheus')
 subdir('test')
+subdir('tpm-eventlog')
```
superm1: Don't forget to add a README.md to this directory.

hughsie:
It seemed to work with my TPM in 2.0 mode too... https://patchwork.kernel.org/patch/9509447/ -- although I don't actually see any SHA256 hashes...

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -0,0 +1,44 @@
...
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+	fu_plugin_set_device_gtype (plugin, FU_TYPE_TPM_EVENTLOG_DEVICE);
+}
+
+gboolean
+fu_plugin_coldplug (FuPlugin *plugin, GError **error)
+{
+	const gchar *fn = "/sys/kernel/security/tpm0/binary_bios_measurements";
```
superm1: Shouldn't this be udev based in case there are multiple tpms?    There should be uevents for subsystem `tpm` so you can at least match to correct character device.

plugins/tpm-eventlog/fu-tpm-eventlog-common.c
```diff
@@ -0,0 +1,91 @@
...
+#include "fu-tpm-eventlog-common.h"
+
+const gchar *
+fu_tpm_eventlog_hash_to_string (FuTpmEventlogHashKind hash_kind)
+{
+	if (hash_kind == FU_TPM_EVENTLOG_HASH_KIND_SHA1)
+		return "SHA-1";
+	if (hash_kind == FU_TPM_EVENTLOG_HASH_KIND_SHA256)
+		return "SHA-256";
+	return NULL;
```
superm1: 384 and 512 should be here too

superm1:
> It seemed to work with my TPM in 2.0 mode too... https://patchwork.kernel.org/patch/9509447/ -- although I don't actually see any SHA256 hashes...

Was that actually merged?  I've got a 5.3.0 kernel on my box that's running TPM2.0 and don't see that sysfs node.

superm1:
I would actually say to make this more useful you should make it a generic "tpm" plugin that shows the TPM firmware version, not just the event log.

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -0,0 +1,170 @@
...
+	guint32		 event_type;
+	guint8		 digest[20];
+	guint32		 event_data_sz;
+} TpmPcrEvent;
+
+typedef struct __attribute__ ((packed)) {
+	guint32		 pcr_index;
+	guint32		 event_type;
+	guint32		 digestCount;
+} TpmPcrEvent2Header;
```
superm1: These two come from somewhere exported?  If so, should try to include that header instead?

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -0,0 +1,170 @@
...
+	g_free (item->checksum);
+	g_free (item);
+}
+
+static void
+fu_tpm_eventlog_device_init (FuTpmEventlogDevice *self)
+{
+	self->items = g_ptr_array_new_with_free_func ((GDestroyNotify) fu_tpm_eventlog_device_item_free);
+	fu_device_set_name (FU_DEVICE (self), "TPM Event Log");
+	fu_device_set_physical_id (FU_DEVICE (self), "tpm");
```
superm1: Physical ID should correspond to TPM number

hughsie:
> you should make it a generic "tpm" plugin that shows the TPM firmware version

You're probably right. Give me a few hours.

superm1:
> You're probably right. Give me a few hours.

The exact same way that the model number is read here: https://github.com/fwupd/fwupd/blob/master/plugins/dell/fu-plugin-dell.c#L575 should work with 
https://github.com/tpm2-software/tpm2-tss/blob/7b2dc9e2e1f4b44df073643075c24f74a0187211/include/tss2/tss2_tpm2_types.h#L482

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
hughsie: Some defines are available, but alas not the header format structures. I've used what I can from tss2.

hughsie:
> Was that actually merged?

I have 5.3.12 and see it here. Not having the event log is now a soft failure, and the TPM FuDevice still should get added.

plugins/tpm-eventlog/fu-tpm-eventlog-common.c
hughsie: Added. It's a shame tss2 doesn't provide that. I've opened a bug about both helpers: https://github.com/tpm2-software/tpm2-tss/issues/1567

plugins/meson.build
hughsie: Added, thanks.

hughsie:
@superm1 I'm kinda stuck now with:

    The following packages have unmet dependencies:
     libtss2-dev:s390x : Depends: libtss2-esys0:s390x (= 2.3.1-3) but it is not going to be installed

Any ideas?

plugins/tpm/README.md
```diff
@@ -0,0 +1,26 @@
...
+GUID Generation
+---------------
+
+These devices use custom GUIDs:
+
+
+ * `TPM\VEN_$(vendor)&DEV_$(type)`
+ * `TPM\VEN_$(vendor)&DEV_$(type)_VER_$(family)`, where `family` is either `2.0` or `1.2`
+
+Example GUIDs from a real system containing a TPM from Intel:
```
superm1: You mean firmware TPM?  Intel doesn't make discrete TPMs

hughsie: I guess I might have a "soft" TPM then, but my vendor is clearly `INT` and the string is "Intel".

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,232 @@
...
+	}
+	if (capability->data.tpmProperties.tpmProperty[0].property != query) {
+		g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
+			     "wrong query returned (got %x expected %x)",
+			     capability->data.tpmProperties.tpmProperty[0].property,
+			     query);
+		return FALSE;
+	}
+
+	if (val != NULL)
```
superm1: I'd actually say this should be a `g_return_val_if_fail` for the function - is there actually a case that we want to be supporting sending a NULL pointer for `val`?

hughsie: Ya g_return_val_if_fail makes sense here.

superm1:
> @superm1 I'm kinda stuck now with:
> 
> ```
> The following packages have unmet dependencies:
>  libtss2-dev:s390x : Depends: libtss2-esys0:s390x (= 2.3.1-3) but it is not going to be installed
> ```
> 
> Any ideas?

This seems transient to me because s390x libtss does exist in the debian archive (I actually fixed a FTBFS recently for that architecture upstream)., I suspect it goes away on it's own with a future push soon.

superm1:
For me it doesn't seem to be working for at least two things:
1. I'm missing event logs for TPM2.0 in the kernel (kernel 5.3.0)
2. Version isn't getting populated (comparison to dell included)

```
# sudo /usr/local/libexec/fwupd/fwupdtool get-devices  --plugin-whitelist=dell --plugin-whitelist=uefi --plugin-whitelist=tpm --show-all-devices
Loading…                 [-                                      ]14:16:53:0625 FuPluginTpmEventlog  no event log: Failed to open file “/sys/kernel/security/tpm0/binary_bios_measurements”: No such file or directory
○
│
├─System Firmware:
│     Current version:     0.1.8.0
│     Minimum Version:     0.1.8.0
│     Vendor:              Dell Inc.
│                          • Supported on remote server
│   
├─TPM:
│     Device ID:           c6a80ac3a22083423992a3cb15018989f37834d6
│     Current version:     33556224.33554432
│     Vendor:              NPCT75xrls (TPM:NTC)
│     GUIDs:               fac1c8f3-73c8-5cd6-8330-07a3690b5140 ← TPM\VEN_NTC&DEV_0000
│                          e9ccc1dc-960a-5e09-afe9-e59a904b776d ← TPM\VEN_NTC&DEV_0000&VER_2.0
│   
└─TPM 2.0:
      Device ID:           2e59657d5e0beb043a16639968ca6178bcedff42
      Summary:             Platform TPM device
      Current version:     7.2.0.2
      Vendor:              Dell Inc.
      Update Error:        Updating disabled due to TPM ownership
      GUIDs:               d621e687-b00d-539a-aeac-ac3eb530cbc5 ← 08af-2.0
                           7d65b10b-bb24-552d-ade5-590b3b278188 ← DELL-TPM-2.0-NTC-NPCT
                           6f5ddd3a-8339-5b2a-b9a6-cf3b92f6c86d ← DELL-TPM-2.0-NTC-NPCT75x
                           fe462d4a-e48f-5069-9172-47330fc5e838 ← DELL-TPM-2.0-NTC-NPCT75xrls
                           • Requires AC power
```

Compare to `tpm2_getcap`:

```
sudo tpm2_getcap  properties-fixed
TPM2_PT_FAMILY_INDICATOR:
  raw: 0x322E3000
  value: "2.0"
TPM2_PT_LEVEL:
  raw: 0
TPM2_PT_REVISION:
  value: 1.16
TPM2_PT_DAY_OF_YEAR:
  raw: 0x109
TPM2_PT_YEAR:
  raw: 0x7E0
TPM2_PT_MANUFACTURER:
  raw: 0x4E544300
  value: "NTC"
TPM2_PT_VENDOR_STRING_1:
  raw: 0x4E504354
  value: "NPCT"
TPM2_PT_VENDOR_STRING_2:
  raw: 0x37357800
  value: "75x"
TPM2_PT_VENDOR_STRING_3:
  raw: 0x200003B
  value: ""
TPM2_PT_VENDOR_STRING_4:
  raw: 0x726C7300
  value: "rls"
TPM2_PT_VENDOR_TPM_TYPE:
  raw: 0x0
TPM2_PT_FIRMWARE_VERSION_1:
  raw: 0x70002
TPM2_PT_FIRMWARE_VERSION_2:
  raw: 0x2
```

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,232 @@
...
+		g_prefix_error (error, "failed to read TPM type");
+		return FALSE;
+	}
+
+	/* these are not guaranteed by spec and may be NULL */
+	vendor2 = fu_plugin_tpm_get_string (ctx, TPM2_PT_VENDOR_STRING_2, error);
+	vendor3 = fu_plugin_tpm_get_string (ctx, TPM2_PT_VENDOR_STRING_3, error);
+	vendor4 = fu_plugin_tpm_get_string (ctx, TPM2_PT_VENDOR_STRING_4, error);
+	vendor = g_strjoin ("", vendor1, vendor2, vendor3, vendor4, NULL);
+	fu_device_set_vendor (device, vendor);
```
superm1: This should correspond to the manufacturer string not the vendor strings (`TPM2_PT_MANUFACTURER`).

Vendor strings should only be used for building GUID to identify the TPM.

superm1:
Once this is fixed the commit can close #1336.

CC @diabonas for review ^

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,232 @@
...
+	vendor_id = g_strdup_printf ("TPM:%s", manufacturer);
+	fu_device_set_vendor_id (device, vendor_id);
+
+	/* get version */
+	if (!fu_plugin_tpm_get_uint32 (ctx, TPM2_PT_FIRMWARE_VERSION_1, &version1, error))
+		return FALSE;
+	if (!fu_plugin_tpm_get_uint32 (ctx, TPM2_PT_FIRMWARE_VERSION_2, &version2, error))
+		return FALSE;
+	version = g_strdup_printf ("%u.%u", version1, version2);
+	fu_device_set_version (device, version, FWUPD_VERSION_FORMAT_PAIR);
```
superm1: So at least for Nuvoton this isn't a pair - it should be quad.  The spec make allusions that this should be a vendor specific way to interpret this field, but helpfully the spec links to more details are missing.
https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf

superm1:
I think that ideally this parent device should have it's device built the same as `dell`, but Dell should be marked so it can supersede it (since Dell can flash).  The child event log device should persist though.

plugins/tpm/fu-tpm-device.c
hughsie: > it should be quad

A 64 bit number -> `quad`?

plugins/tpm/fu-tpm-device.c
hughsie: > Vendor strings should only be used for building GUID to identify the TPM.

So the vendor ID string is meant to be free form string that identifies the firmware stream for the device? Both examples I have here just have the TPM vendor name.

plugins/tpm/fu-tpm-device.c
superm1: Yeah you can see `TPM2_PT_FIRMWARE_VERSION_1` is 0x70002 and `TPM2_PT_FIRMWARE_VERSION_2` is 0x2, which should expand to 7.2.0.2 for the TPM in my system.

plugins/tpm/fu-tpm-device.c
superm1: That was my understanding of it; vendors can do whatever they want with those strings.  @flihp ,@diabonas, @AndreasFuchsSIT any other context would be helpful.

hughsie:
> I'm missing event logs for TPM2.0 in the kernel (kernel 5.3.0)

I think you need a newer kernel for that.

> but Dell should be marked so it can supersede it

I did wonder if we could use the fu_plugin_register() stuff here and then "hook" that in the dell plugin, add the extra GUIDs and mark it as upgradable.

plugins/tpm/fu-tpm-device.c
hughsie: So we convert each uint32 numbers to pairs, and then glue them together? Maybe you were right about version_raw being 64 bit afterall...

superm1:
> > I'm missing event logs for TPM2.0 in the kernel (kernel 5.3.0)
> 
> I think you need a newer kernel for that.

Hmm, I checked and that patch you linked was merged a long time ago - 4.11.  So something is wrong with it not populating on my discrete TPM + FW combo.
https://github.com/torvalds/linux/commit/4d23cc323cdbee1cbcd8a7f059fff9ef2b0c473d

> 
> > but Dell should be marked so it can supersede it
> 
> I did wonder if we could use the fu_plugin_register() stuff here and then "hook" that in the dell plugin, add the extra GUIDs and mark it as upgradable.

Keep in mind that the Dell plugin does a register which UEFI picks up.  So we're talking about a 3 chain relation (which is fine).

plugins/tpm/fu-tpm-device.c
superm1: Yeah, similarly to how the vendor strings are glued together.  i think it' because all these definitions need to be done in 4 bytes increments.

> Maybe you were right about version_raw being 64 bit afterall...

Hah, I guess maybe so :)  Good thing that's not in a released version yet.

superm1:
> @superm1 I'm kinda stuck now with:
> 
> ```
> The following packages have unmet dependencies:
>  libtss2-dev:s390x : Depends: libtss2-esys0:s390x (= 2.3.1-3) but it is not going to be installed
> ```
> 
> Any ideas?

Something funny is going on with dependencies.  I'll report it to Debian.

```
root@553c40bab08a:~# apt install libtss2-esys0:s390x 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libtss2-esys0:s390x : Depends: tpm-udev:s390x but it is not installable
E: Unable to correct problems, you have held broken packages.
root@553c40bab08a:~# apt policy tpm-udev
tpm-udev:
  Installed: 0.2
  Candidate: 0.2
  Version table:
 *** 0.2 500
        500 http://deb.debian.org/debian testing/main amd64 Packages
        500 http://deb.debian.org/debian testing/main s390x Packages
        100 /var/lib/dpkg/status
```

superm1:
> s390x

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=946072

plugins/tpm/fu-tpm-device.c
diabonas: I don't think the contents of `TPM2_PT_VENDOR_STRING` are specified by the TPM 2.0 specification - it might contain the model of the TPM (e.g. for my Infineon TPM, it reads [`SLB9670`](https://www.infineon.com/cms/de/product/security-smart-card-solutions/optiga-embedded-security-solutions/optiga-tpm/slb-9670vq2.0/)), seemingly random bytes or be completely empty. `TPM2_PT_MANUFACTURER` on the other hand should have one of the values defined in the [TCG TPM Vendor ID Registry](https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-Vendor-ID-Registry-Version-1.01-Revision-1.00.pdf).

plugins/tpm/fu-tpm-device.c
hughsie: Okay, fixed.

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,291 @@
...
+		return "Samsung";
+	if (g_strcmp0 (manufacturer, "SNS") == 0)
+		return "Sinosun";
+	if (g_strcmp0 (manufacturer, "TXN") == 0)
+		return "Texas Instruments";
+	if (g_strcmp0 (manufacturer, "WEC") == 0)
+		return "Winbond";
+	if (g_strcmp0 (manufacturer, "ROCC") == 0)
+		return "Fuzhou Rockchip";
+	return NULL;
```
superm1: Aren't you missing Google?  `GOOG`

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,291 @@
...
+		return FALSE;
+	if (!fu_plugin_tpm_get_uint32 (ctx, TPM2_PT_FIRMWARE_VERSION_2, &version2, error))
+		return FALSE;
+	version_raw = ((guint64) version1) << 32 | ((guint64) version2);
+	fu_device_set_version_raw (device, version_raw);
+
+	/* this has to be done after _add_instance_id() adds the quirks */
+	verfmt = fu_device_get_version_format (device);
+	version = fu_common_version_from_uint64 (version_raw, verfmt);
+	fu_device_set_version (device, version, verfmt);
```
superm1: Double checked on my system and the version still isn't rendering right.

```
├─TPM:
│     Device ID:           c6a80ac3a22083423992a3cb15018989f37834d6
│     Current version:     512.1792.512.0
│     Vendor:              Nuvoton Technology (TPM:NTC)
│     GUIDs:               fac1c8f3-73c8-5cd6-8330-07a3690b5140
│                          4f3a6ec5-d6a6-5912-a802-78eefaddd489
│                          e9ccc1dc-960a-5e09-afe9-e59a904b776d
│                          e7ca0c22-b861-5f58-be40-ed89447463d5
│   
├─TPM 2.0:
│     Device ID:           2e59657d5e0beb043a16639968ca6178bcedff42
│     Summary:             Platform TPM device
│     Current version:     7.2.0.2
│     Vendor:              Dell Inc.
│     Update Error:        Updating disabled due to TPM ownership
│     GUIDs:               d621e687-b00d-539a-aeac-ac3eb530cbc5
│                          7d65b10b-bb24-552d-ade5-590b3b278188
│                          6f5ddd3a-8339-5b2a-b9a6-cf3b92f6c86d
│                          fe462d4a-e48f-5069-9172-47330fc5e838
```

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,291 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_tpm_device_init (FuTpmDevice *self)
+{
+	fu_device_set_name (FU_DEVICE (self), "TPM");
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_QUAD);
```
superm1: I think you probably also want to add some flags for internal too right?

hughsie: Agree, thanks.

plugins/tpm/fu-tpm-device.c
hughsie: Ahh, I was using the old `Vendor_ID_Registry_0-8_clean.pdf` -- I've added HPE, Microsoft and Google to the list. Good catch!

plugins/tpm/fu-tpm-device.c
superm1: Fixed it; just swapped endian.  It's a separate commit if you don't agree.

superm1:
> > s390x
> 
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=946072

I've got a fix for the s390x failure into Debian.
https://tracker.debian.org/news/1084374/accepted-tpm-udev-04-source-into-unstable/

It will take some time to propagate through.  So if you want to merge soon i'm fine if you want to keep this off for s390x for now and re-enable it later, otherwise in a day or two I should be able to modify the CI Dockerfile to pull that package from unstable.

plugins/tpm/fu-tpm-device.c
flihp: > I don't think the contents of `TPM2_PT_VENDOR_STRING` are specified by the TPM 2.0 specification

Agree. The spec isn't very helpful here. It describes `TPM2_PT_VENDOR_STRING` as:

> the first four characters of the vendor ID string

which IMHO is misleading since `TPM_PT_MANUFACTURER` is also described as a `vendor ID`, though not explicitly a string:

> the vendor ID unique to each TPM manufacturer

Even then, the vendor id registry maps the hex IDs to string values: ex. `0x47 0x4F 0x4F 0x47` -> `GOOG`. The only thing to go on here is that `TPM_PT_MANUFACTURER` is guaranteed to be unique (from the vendor id registry) whereas the `TPM2_PT_VENDOR_STRING` (only one is required, up to 4) could be anything.

plugins/tpm/fu-tpm-device.c
```diff
@@ -249,7 +249,7 @@ fu_tpm_device_setup (FuDevice *device, GError **error)
 		return FALSE;
 	if (!fu_plugin_tpm_get_uint32 (ctx, TPM2_PT_FIRMWARE_VERSION_2, &version2, error))
 		return FALSE;
-	version_raw = ((guint64) version1) << 32 | ((guint64) version2);
+	version_raw = ((guint64) GUINT32_FROM_BE (version1)) << 32 | ((guint64) GUINT32_FROM_BE (version2));
```
hughsie: But we've already done `GUINT32_FROM_BE` in `fu_plugin_tpm_get_uint32()` so this means it's undoing the byte-swaps and operating on raw bytes. Is that expected? I think we just want to remove *both* `GUINT32_FROM_BE` if that's the case.

superm1: Ah yes; I agree.

superm1:
The dell plugin registering stuff doesn't appear to be working.  I think it's because Dell runs a coldplug routine to probe for TPM rather than a udev event.  I think that it might make sense to just remove that part of the routine and instead rely upon the registered signal and probe for the bits at that time.

This makes `tpm` a dependency of dell working, but I don't see a problem with that.

superm1:
I figured out why I wasn't seeing my event log.  It turns out that it's only exported when secure boot is enabled.  After I turned on secure boot I see it, but this plugin fails to read the event log.

```
18:32:24:0687 FuEngine             1 devices with subsystem tpm
18:32:24:0687 FuDevice             using c6a80ac3a22083423992a3cb15018989f37834d6 for DEVNAME=/dev/tpm0
18:32:24:0734 FuPluginTpmEventlog  Event Data:
53 70 65 63 20 49 44 20 45 76 65 6e 74 30 33 00 00 00 00 00 00 02 00 02 02 00 00 00 04 00 14 00 
0b 00 20 00 00 
18:32:24:0734 FuPluginTpmEventlog  no event log: attempted to read 0x398e14c5 bytes from buffer of 0xd2cb
```

I _think_ it might because of an assumption that only the SHA1 digest is exported in the event log because this TPM does definitely support SHA256.

superm1:
> I figured out why I wasn't seeing my event log. It turns out that it's only exported when secure boot is enabled. After I turned on secure boot I see it, but this plugin fails to read the event log.
> 
> ```
> 18:32:24:0687 FuEngine             1 devices with subsystem tpm
> 18:32:24:0687 FuDevice             using c6a80ac3a22083423992a3cb15018989f37834d6 for DEVNAME=/dev/tpm0
> 18:32:24:0734 FuPluginTpmEventlog  Event Data:
> 53 70 65 63 20 49 44 20 45 76 65 6e 74 30 33 00 00 00 00 00 00 02 00 02 02 00 00 00 04 00 14 00 
> 0b 00 20 00 00 
> 18:32:24:0734 FuPluginTpmEventlog  no event log: attempted to read 0x398e14c5 bytes from buffer of 0xd2cb
> ```
> 
> I _think_ it might because of an assumption that only the SHA1 digest is exported in the event log because this TPM does definitely support SHA256.

Specifically if you look at the way the kernel does it, it dynamically calculates the size of each event, so I think the simple header that exists right now for `TpmPcrEvent` probably won't be enough.
https://github.com/torvalds/linux/blob/328fefadd9cfa15cd6ab746553d9ef13303c11a6/include/linux/tpm_eventlog.h#L158

hughsie:
Can you attach your boot log file please, and I'll fix the parser.

superm1:
[measurements.zip](https://github.com/fwupd/fwupd/files/3923532/measurements.zip)

plugins/dell/fu-plugin-dell.c
```diff
@@ -716,71 +709,61 @@ fu_plugin_dell_detect_tpm (FuPlugin *plugin, GError **error)
...
 	}
 	/* build GUIDs from TSS strings */
-	if (!fu_plugin_dell_add_tpm_model (dev, &error_tss))
+	if (!fu_plugin_dell_add_tpm_model (data->tpm_device, &error_tss))
 		g_debug ("could not build instances: %s", error_tss->message);
 
-	if (!fu_device_setup (dev, error))
-		return FALSE;
-	if (data->tpm_device == NULL)
-		fu_plugin_device_register (plugin, dev);
```
superm1: So I think the problem with this flow is that the device is still owned by `tpm` not by `uefi`.  So the `uefi` registration won't pick it up.

hughsie: So maybe I'm making this all too complicated. Perhaps it's better is we just have:

* A tpm plugin, where the `dell` plugin is marked BETTER_THAN the `tpm` plugin, so that if the dell plugin creates the TPM device the tpm-created one gets ignored. Sure, there's code duplication, but much less risk we cause a regression somewhere.

* A tpm-eventlog plugin, which adds the eventlog child functionality to *either* tpm device.

superm1: > A tpm plugin, where the dell plugin is marked BETTER_THAN the tpm plugin, so that if the dell plugin creates the TPM device the tpm-created one gets ignored. Sure, there's code duplication, but much less risk we cause a regression somewhere.

Yeah that means a lot less complexity of this dance between 3rd plugins then.

> A tpm-eventlog plugin, which adds the eventlog child functionality to either tpm device.

Yeah that makes sense - it can listen for the registered signal and look for something indicating it's a TPM and then mark that as a parent.

hughsie: > it can listen for the registered signal

I don't even think it needs to do that. Other plugins just use `fu_device_add_parent_guid`

superm1: Oh and make a common GUID for both that is platform-tpm or something. Yeah

hughsie: I went with `system-tpm` -- is that better or worse than `platform-tpm`?

superm1: That's fine too. No strong opinion

hughsie:
> measurements.zip

I'm assuming you don't mind if I include that in the self tests?

superm1:
> I'm assuming you don't mind if I include that in the self tests?

No concern there.

plugins/tpm/fu-tpm-device.c
hughsie: I've done that now.

hughsie:
Okay, this one should be a lot simpler to review and test. Fingers crossed. The extra data I'm adding to the report looks like this:

    0x00000008 231f248f12ef9f38549f1bda7a859b781b5caab0 [N.1.E.E.T.8.6.W.....]
    0x80000008 d235e68f43ec4db54caea469e1e69a8fcd93610f
    0x80000008 a8d6cef644a13a29fac2e89975bdd1fe804c259a
    0x80000008 ef86cabef6a3e450737931ab300c590466816f88
    0x80000008 2ab15dc50ed8159843cfd9abb03ffab713df8123 [..........D.....]
    0x00000004 9069ca78e7450a285173431b3e52c5c25299e473

Which does look slightly scary if you're looking at the JSON upload, but I'm pretty sure nobody does. It's easy enough to explain I suppose.

superm1:
I have to wonder if the event log even makes sense as a fwupd device and if instead it should be a debugging tool installed into `libexec`.

Here's my thoughts on why:
* It doesn't have a version set
* You don't get any event log data in the device itself, even with `--verbose`

So to get this information you have to run the daemon in verbose, or `fwupdtool` in verbose. At that point, isn't a separate tool easier? :)

hughsie:
> isn't a separate tool easier

I think the value is having the event log as part of the report metadata, and you can't do that with a debug tool. It's not like we can just email the few people that send reports with funky PCR0s, we have to have all the data sent already.

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -0,0 +1,51 @@
...
+	dev = fu_tpm_eventlog_device_new (buf, bufsz, error);
+	if (dev == NULL)
+		return FALSE;
+	if (!fu_device_setup (FU_DEVICE (dev), error))
+		return FALSE;
+
+	/* add optional report metadata */
+	str = fu_tpm_eventlog_device_report_metadata (dev);
+	g_debug ("using TPM event log report data of:\n%s", str);
+	fu_plugin_add_report_metadata (plugin, "TpmEventLog", str);
```
superm1: So this is adding it to the tpm event plugin, but why would this ever flow up to fwupd?  Doesn't it need to come in the UEFI plugin?

superm1:
> I think the value is having the event log as part of the report metadata, and you can't do that with a debug tool. It's not like we can just email the few people that send reports with funky PCR0s, we have to have all the data sent already.

I guess see my other comment then.  I think that you need to merge this with the UEFI plugin to make it useful and the reports to come up.  Otherwise this metadata shouldn't be added to UEFI reports should it?

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -0,0 +1,355 @@
...
+	g_autoptr(GString) csum = g_string_new (NULL);
+	g_autofree guint8 *digest = g_malloc0 (length);
+	if (!fu_memcpy_safe (digest, length, 0x0,	/* dst */
+			     buf, bufsz, offset,	/* src */
+			     length, error))
+		return FALSE;
+	for (guint i = 0; i < length; i++)
+		g_string_append_printf (csum, "%02x", digest[i]);
+	return g_string_free (g_steal_pointer (&csum), FALSE);
+}
```
superm1: I guess you meant for this function to go into fu-common, but forgot to move it?

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
hughsie: Hmm. In my head I was thinking that all the plugin report metadata would get sent, but you're correct in that only the plugin that processed the update would be added. I guess I was thinking about `runtime_versions` which is "global".

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
hughsie: I wasn't sure. I kinda wanted to leave it there, and only export it if another plugin needed the same functionality.

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
superm1: Which I think is fine, this can just merge with the UEFI plugin.  I do think that it would make sense to get this stuff out from a `libexec` tool too though.  Another argument to `fwupdate` that calls out to the `get_string` for this device would be handy.  I don't know any other tools that are parsing those measurements.

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -0,0 +1,355 @@
...
+						     data, datasz, 20,
+						     FU_DUMP_FLAGS_SHOW_ASCII);
+			}
+			item = g_new0 (FuTpmEventlogDeviceItem, 1);
+			item->kind = event_type;
+			item->checksum = g_steal_pointer (&csum);
+			item->blob = g_bytes_new_take (g_steal_pointer (&data), datasz);
+			g_ptr_array_add (self->items, item);
+		}
+
```
superm1: At least when this was run from a debugging tool, I would really like to see the values that go into all the PCR's.  It's really interesting to see which measurements are impacting PCRs.

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -0,0 +1,355 @@
...
+{
+	gchar sig[] = FU_TPM_EVENTLOG_V2_HDR_SIGNATURE;
+
+	/* look for TCG v2 signature */
+	if (!fu_memcpy_safe ((guint8 *) sig, sizeof(sig), 0x0,		/* dst */
+			     buf, bufsz, FU_TPM_EVENTLOG_V1_SIZE,	/* src */
+			     sizeof(sig), error))
+		return FALSE;
+	if (g_strcmp0 (sig, FU_TPM_EVENTLOG_V2_HDR_SIGNATURE) == 0)
+		return fu_tpm_eventlog_device_parse_blob_v2 (self, buf, bufsz, error);
```
superm1: It might be useful to note in a `g_debug` here that the blob was a v2 for future debugging points when somethign goes wrong.

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -0,0 +1,355 @@
...
+		/* read data block */
+		if (!fu_common_read_uint32_safe	(buf, bufsz, idx,
+						 &datasz, G_LITTLE_ENDIAN, error))
+			return FALSE;
+
+		/* save blob if PCR=0 */
+		idx += sizeof(datasz);
+		if (pcr == ESYS_TR_PCR0) {
+			g_autofree guint8 *data = NULL;
+			FuTpmEventlogDeviceItem *item;
```
superm1: I think you want these previous two lines reversed

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
superm1: OK, I'd say probably rename it to something without `fu_common` prefix then

superm1:
I ended up fixing up my comments myself.  There is also a new `--pcr` argument in fwupdate now that will show you the full event log corresponding to that PCR :)

plugins/uefi/meson.build
```diff
@@ -15,6 +15,8 @@ shared_module('fu_plugin_uefi',
   fu_hash,
   sources : [
     'fu-plugin-uefi.c',
+    'fu-tpm-eventlog-common.c',
+    'fu-tpm-eventlog-device.c',
```
hughsie: I don't think stuffing it all in one plugin solves the architectural problem. There's nothing really update-capsule specific about the tpm event log, and this feels like working around the design rather than fixing the real issue. I also don't think the TPM event log is a logical child of the system firmware, it's a child of the TPM device itself. I'll have a think about a better way we can do this.

plugins/uefi/fu-tpm-eventlog-device.c
```diff
@@ -240,31 +264,28 @@ fu_tpm_eventlog_device_parse_blob_v2 (FuTpmEventlogDevice *self,
 						 &datasz, G_LITTLE_ENDIAN, error))
 			return FALSE;
 
-		/* save blob if PCR=0 */
+		/* save blob */
 		idx += sizeof(datasz);
-		if (pcr == ESYS_TR_PCR0) {
```
hughsie: Saving all the bootlog data for every instance is a good chunk of the total RSS, about 60kb from my testing here. I think we only want to collect the PCR0 by default, and parse the rest on demand. I've got an idea for that.

plugins/uefi/fu-uefi-tool.c
```diff
@@ -305,6 +310,28 @@ main (int argc, char *argv[])
 		g_print ("%s\n", _("Disabled fwupdate debugging"));
 	}
 
+	if (action_pcr) {
```
hughsie: I think fwupdate should really be a deprecated command. I think we can do a much smaller binary that's also suitable for fuzz testing.

plugins/uefi/meson.build
superm1: The log itself is coming from firmware though, doesn't that make it a child of the firmware?

If it does go back into it's own plugin I would like to still see the functionality I added to fwupdate of spitting out a random PCR log in a separate tool though.

plugins/uefi/fu-tpm-eventlog-device.c
superm1: Okay

plugins/uefi/meson.build
hughsie: I'm just prototyping exactly that :)

superm1:
Superceded by https://github.com/fwupd/fwupd/pull/1644

PR #1630, opened 2019-12-03
Wip/hughsie/verfmt dell

hughsie:


superm1:
So this certainly will fix the rendering of the display - but trying to install firmware becomes problematic.

```
# fwupdmgr reinstall
No releases found for device: Firwmare version formats were different, device was 'dell-esrt' and release is 'quad'
```

hughsie:
> but trying to install firmware becomes problematic

A prerequisite of this commit would be to change all the Dell UEFI firmware on the LVFS to be `dell-esrt;quad` rather than the current of `quad`. We'd then write out something like this:

    <custom>
      <value key="LVFS::VersionFormat">quad</value>
      <value key="LVFS::VersionFormats">dell-esrt;quad</value>
    </custom>

Old versions of fwupd would read just `LVFS::VersionFormat` and new versions would just read `LVFS::VersionFormats` and would match with `true` in either case.

Actually, now you've got me thinking. Old versions of fwupd are using `xb_node_query_text()` which just queries the first node that matches. Maybe a simpler way would be to have:

    <custom>
      <value key="LVFS::VersionFormat">quad</value>
      <value key="LVFS::VersionFormat">dell-esrt</value>
    </custom>

and then that's compatible with both....

superm1:
> and then that's compatible with both....

Sounds like a pretty novel idea.

hughsie:
> Sounds like a pretty novel idea.

It works! Peer review for the win.

src/fu-install-task.c
```diff
@@ -92,6 +92,59 @@ fu_install_task_get_is_downgrade (FuInstallTask *self)
...
+{
+	FwupdVersionFormat fmt_dev = fu_device_get_version_format (self->device);
+	g_autoptr(GString) version_format = g_string_new (NULL);
+
+	/* for the error message */
+	for (guint i = 0; i < verfmts->len; i++) {
+		XbNode *verfmt = g_ptr_array_index (verfmts, i);
+		const gchar *tmp = xb_node_get_text (verfmt);
+		g_string_append_printf (version_format, "%s;", tmp);
+	}
```
superm1: If this is only used for the error and warning messages I'd think it makes sense to split this off to it's own static function to build and allocate this string in those scenarios.

src/fu-install-task.c
```diff
@@ -92,6 +92,59 @@ fu_install_task_get_is_downgrade (FuInstallTask *self)
...
+	    (flags & FWUPD_INSTALL_FLAG_FORCE) == 0) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "release version format '%s' but no device version format",
+			     version_format->str);
+		return FALSE;
+	}
+
+	/* fall back to older version formats */
```
superm1: this text is wrong; it's not falling back - it's just checking all version formats.

libfwupd/fwupd-enums.h
```diff
@@ -241,12 +241,13 @@ typedef enum {
...
  * @FWUPD_VERSION_FORMAT_PAIR:			Two AABB.CCDD version numbers
  * @FWUPD_VERSION_FORMAT_TRIPLET:		Microsoft-style AA.BB.CCDD version numbers
- * @FWUPD_VERSION_FORMAT_QUAD:			Dell-style AA.BB.CC.DD version numbers
+ * @FWUPD_VERSION_FORMAT_QUAD:			UEFI-style AA.BB.CC.DD version numbers
  * @FWUPD_VERSION_FORMAT_BCD:			Binary coded decimal notation
  * @FWUPD_VERSION_FORMAT_INTEL_ME:		Intel ME-style bitshifted notation
  * @FWUPD_VERSION_FORMAT_INTEL_ME2:		Intel ME-style A.B.CC.DDDD notation notation
  * @FWUPD_VERSION_FORMAT_SURFACE_LEGACY:	Legacy Microsoft Surface 10b.12b.10b
  * @FWUPD_VERSION_FORMAT_SURFACE:		Microsoft Surface 8b.16b.8b
+ * @FWUPD_VERSION_FORMAT_DELL_ESRT:		Dell ESRT BB.CC.DD style
```
superm1: I think I would prefer this to be `DELL_BIOS` or `DELL_SYSTEM` as it's only valid for use with the system firmware.  Device firmware doesn't have the same definition necessarily.

hughsie: "system" is a really hand-wavey noun -- I think "bios" is much better understood in this context. I'll switch everything to `dell-bios` and then work on the LVFS side.

src/fu-install-task.c
hughsie: Fixed.

src/fu-install-task.c
hughsie: Agreed. Done.

superm1:
Something i'm wondering about is what happens with `get-upgrades` and `upgrade` commands.  Do these comparisons still DTRT?  I guess once the LVFS side is done they should..

src/fu-install-task.c
```diff
@@ -92,6 +92,67 @@ fu_install_task_get_is_downgrade (FuInstallTask *self)
...
+}
+
+static gboolean
+fu_install_task_check_verfmt (FuInstallTask *self,
+			      GPtrArray *verfmts,
+			      FwupdInstallFlags flags,
+			      GError **error)
+{
+	FwupdVersionFormat fmt_dev = fu_device_get_version_format (self->device);
+	g_autoptr(GString) version_format = g_string_new (NULL);
```
superm1: not needed (failing clang)

```
../src/fu-install-task.c:116:21: error: unused variable 'version_format' [-Werror,-Wunused-variable]

        g_autoptr(GString) version_format = g_string_new (NULL);
```

superm1:
commit message needs updating (references `dell-esrt` rather than `dell-bios`)

hughsie:
> commit message needs updating

Done. I'll work on the LVFS bits now.

hughsie:
> Something i'm wondering about is what happens with get-upgrades and upgrade commands. Do these comparisons still DTRT

I don't see why they wouldn't work the same way. Really it's all `install()` behind the scenes....

superm1:
I checked this morning... Yeah the version comparison is failing on reinstall.
```
#  fwupdmgr reinstall 
Selected device: System Firmware
XPS 13 7390 must remain plugged into a power source for the duration of the update to avoid damage. Continue with update? [Y|n]: y
Downloading 1.2.0 for System Firmware...
Fetching firmware https://fwupd.org/downloads/945e4e936f974e958ac0bc9c7129707e76cc5660-XPS_13_7390_1.2.0.cab
Downloading…             [***************************************] Less than one minute remaining…
Specified firmware is older than installed '0x10200 < 1.2.0'
```

hughsie:
> Specified firmware is older than installed

Ahh, so it's making it through `fu_install_task_check_verfmt()`. This is where we stray into some bad API design...

The `fu_common_vercmp()` function takes two things, and tries to convert them to semvers using 
`fu_common_version_parse()` which helpfully hardcodes `FWUPD_VERSION_FORMAT_TRIPLET`...

This is one of those cases where you wonder how it all actually worked in the first place... Lets analyse:

945e4e936f974e958ac0bc9c7129707e76cc5660-XPS_13_7390_1.2.0.cab has `version="66048"` in the LVFS metadata, and:

    <custom>
      <value key="LVFS::VersionFormat">quad</value>
      <value key="LVFS::VersionFormat">dell-bios</value>
    </custom>

This is as expected. I have to run now, but I'll continue in an hour or so.

libfwupdplugin/fu-common-version.h
```diff
@@ -15,7 +15,8 @@ gchar		*fu_common_version_from_uint32	(guint32	 val,
 						 FwupdVersionFormat kind);
 gchar		*fu_common_version_from_uint16	(guint16	 val,
 						 FwupdVersionFormat kind);
-gchar		*fu_common_version_parse	(const gchar	*version);
+gchar		*fu_common_version_parse	(const gchar	*version)
+G_DEPRECATED_FOR(fu_common_version_parse_from_format);
```
superm1: is it worth adding to `fu-deprecated.h` or anything?

hughsie: I don't think so -- it seems a bit harsh to move it to fu-common-version-private.h too, although nothing will be actually linking that symbol directly I hope.

hughsie: FYI: https://github.com/fwupd/fwupd/pull/1638

hughsie:
> but I'll continue in an hour or so

I think with that other PR applied it should work. with **new** metadata it allows me to reinstall the metadata on the XPS 13 with this branch now.

superm1:
LGTM now

PR #1636, opened 2019-12-04
trivial: Add fu_common_version_from_uint64()

hughsie:


superm1:
Added symbol to wrong plugin version, but otherwise lgtm

PR #1637, opened 2019-12-04
Wip/hughsie/udev fixes

hughsie:


libfwupdplugin/fu-udev-device.c
```diff
@@ -342,6 +358,16 @@ fu_udev_device_set_dev (FuUdevDevice *self, GUdevDevice *udev_device)
 #ifdef HAVE_GUDEV
 	priv->subsystem = g_strdup (g_udev_device_get_subsystem (priv->udev_device));
 	priv->device_file = g_strdup (g_udev_device_get_device_file (priv->udev_device));
+
+	/* try to get one linst summary */
+	summary = g_udev_device_get_sysfs_attr (priv->udev_device, "description");
+	if (summary == NULL) {
+		parent = g_udev_device_get_parent (udev_device);
```
superm1: just a nit for consistency; everywhere else in this function you're using `priv->udev_device`

hughsie: Fixed. Consistency is good, don't feel bad when nitpicking. :)

PR #1638, opened 2019-12-04
Wip/hughsie/fu common vercmp

hughsie:


libfwupdplugin/fu-common-version.c
```diff
@@ -439,8 +442,8 @@ fu_common_version_verify_format (const gchar *version,
 
 /**
  * fu_common_vercmp:
- * @version_a: the release version, e.g. 1.2.3
- * @version_b: the release version, e.g. 1.2.3.1
+ * @version_a: the semver release version, e.g. 1.2.3
+ * @version_b: the semver release version, e.g. 1.2.3.1
```
gicmo: I guess `1.2.3.1` is technically not "semver" according to the [grammar](https://semver.org/) but I am not sure it is important here.

hughsie: TIL.

PR #1641, opened 2019-12-05
trivial: Render large values of VersionRaw in a more portable way

hughsie:


libfwupd/fwupd-device.c
```diff
@@ -1959,7 +1959,14 @@ fwupd_device_to_string (FwupdDevice *device)
 	if (priv->flashes_left < 2)
 		fwupd_pad_kv_int (str, FWUPD_RESULT_KEY_FLASHES_LEFT, priv->flashes_left);
 	if (priv->version_raw > 0) {
-		g_autofree gchar *tmp = g_strdup_printf ("0x%x", (guint) priv->version_raw);
+		g_autofree gchar *tmp = g_strdup_printf ("0x%08x", (guint) priv->version_raw);
```
superm1: Isn't this a memory leak?

hughsie: Doh! that was meant to be `= NULL`.  Will fix, thanks.

hughsie: Fixed.

PR #1642, opened 2019-12-05
Add a new plugin that exposes the system TPM device firmware version

hughsie:
This plugin does not yet allow the device to be upgraded, and is provided for
information only.

### DIFF
--- a/plugins/tpm/fu-tpm-device.c
+++ b/plugins/tpm/fu-tpm-device.c
@@ -1,0 +1,258 @@
+/*
+ * Copyright (C) 2019 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <tss2/tss2_esys.h>
+
+#include "fu-tpm-device.h"
+
+struct _FuTpmDevice {
+	FuUdevDevice		 parent_instance;
+};
+
+G_DEFINE_TYPE (FuTpmDevice, fu_tpm_device, FU_TYPE_UDEV_DEVICE)
+
+static void Esys_Finalize_autoptr_cleanup (ESYS_CONTEXT *esys_context)
+{
+	Esys_Finalize (&esys_context);
+}
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (ESYS_CONTEXT, Esys_Finalize_autoptr_cleanup)
+
+static gboolean
+fu_tpm_device_probe (FuUdevDevice *device, GError **error)
+{
+	return fu_udev_device_set_physical_id (device, "tpm", error);
+}
+
+static gboolean
+fu_tpm_device_get_uint32 (ESYS_CONTEXT *ctx, guint32 query, guint32 *val, GError **error)
+{
+	TSS2_RC rc;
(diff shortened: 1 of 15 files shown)


plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,266 @@
...
+			result[i] = 0x20;
+	}
+
+	return fu_common_strstrip (result);
+}
+
+static const gchar *
+fu_tpm_device_convert_manufacturer (const gchar *manufacturer)
+{
+	if (g_strcmp0 (manufacturer, "AMD") == 0)
```
superm1: I think it would be useful to reference the spec this was last updated from in a comment so that when a new vendor enters it's more obvious why they're missing.

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,266 @@
...
+fu_tpm_device_probe (FuUdevDevice *device, GError **error)
+{
+	FuTpmDevice *self = FU_TPM_DEVICE (device);
+
+	/* set the physical ID */
+	if (!fu_udev_device_set_physical_id (device, "tpm", error))
+		return FALSE;
+
+	/* already open, so don't touch */
+	g_object_set (self, "device-file", NULL, NULL);
```
superm1: This feels a little hacky. Maybe it's better to have a proper flag.

plugins/tpm/fu-tpm-device.c
hughsie: Agree. I've added a link to the spec version.

plugins/tpm/fu-tpm-device.c
hughsie: Agree. Give me a couple of mins.

hughsie: https://github.com/fwupd/fwupd/pull/1643

hughsie: Fixed.

plugins/tpm/fu-tpm-device.c
```diff
@@ -0,0 +1,258 @@
...
+	FuUdevDevice		 parent_instance;
+};
+
+G_DEFINE_TYPE (FuTpmDevice, fu_tpm_device, FU_TYPE_UDEV_DEVICE)
+
+static void Esys_Finalize_autoptr_cleanup (ESYS_CONTEXT *esys_context)
+{
+	Esys_Finalize (&esys_context);
+}
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (ESYS_CONTEXT, Esys_Finalize_autoptr_cleanup)
```
superm1: I think if we end up having any other use for TPM outside of querying it in these 3 plugins (TPM,Dell,UEFI) I think all the `ESys_*` calls should move into daemon.

hughsie: I did ponder something like `fu_plugin_get_esys_context()` but then thought it all got a bit complicated when the plugins started reusing the context doing wildly different things. I guess ideally the dell tpm device would just go away, and we could use the standard TPM flashing functionality rather than a capsule, but I know there are lots of other things in play right now.

superm1:
This commit should close #1336

PR #1643, opened 2019-12-05
Add fu_udev_device_set_flags()

hughsie:
This allows us to specify with more detail how the device is opened.

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -43,7 +43,7 @@ typedef struct
 	gchar			*subsystem;
 	gchar			*device_file;
 	gint			 fd;
-	gboolean		 readonly;
+	FuUdevDeviceFlags	 flags;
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -728,7 +728,30 @@ fu_udev_device_set_readonly (FuUdevDevice *self, gboolean readonly)
...
+
+/**
+ * fu_udev_device_set_flags:
+ * @self: A #FuUdevDevice
+ * @flags: a #FuUdevDeviceFlags, e.g. %FU_UDEV_DEVICE_FLAG_OPEN_READ
+ *
+ * Sets the parameters to use when opening the device.
+ *
+ * For example %FU_UDEV_DEVICE_FLAG_OPEN_READ means that fu_device_open()
+ * =
```
superm1: extraneous `=`

PR #1644, opened 2019-12-06
Wip/hughsie/tpm events2

hughsie:


plugins/tpm-eventlog/fu-tpm-eventlog.c
```diff
@@ -0,0 +1,100 @@
...
+		return EXIT_FAILURE;
+	}
+
+	/* set verbose? */
+	if (verbose)
+		g_setenv ("G_MESSAGES_DEBUG", "all", FALSE);
+
+	/* allow user to chose a local file */
+	if (argc <= 1) {
+		g_autoptr(GError) error_local = NULL;
```
superm1: I don't think there is a need to use local error here, just use `error`

plugins/tpm-eventlog/fu-tpm-eventlog.c
```diff
@@ -0,0 +1,100 @@
...
+		if (!fu_tmp_eventlog_process ("/sys/kernel/security/tpm0/binary_bios_measurements", &error_local)) {
+			/* TRANSLATORS: failed to read measurements file */
+			g_print ("%s: %s\n", _("Failed to parse file"),
+				 error_local->message);
+			return EXIT_FAILURE;
+		}
+		return EXIT_SUCCESS;
+	}
+	for (gint i = 1; i < argc; i++) {
+		g_autoptr(GError) error_local = NULL;
```
superm1: likewise, same here, just use `error`

plugins/tpm-eventlog/fu-tpm-eventlog.c
```diff
@@ -0,0 +1,100 @@
...
+main (int argc, char *argv[])
+{
+	gboolean verbose = FALSE;
+	gint rc = EXIT_SUCCESS;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GOptionContext) context = g_option_context_new (NULL);
+	const GOptionEntry options[] = {
+		{ "verbose", 'v', 0, G_OPTION_ARG_NONE, &verbose,
+			/* TRANSLATORS: command line option */
+			_("Show extra debugging information"), NULL },
```
superm1: I think it's still useful to be able to only look at one PCR log at a time.  Showing the whole log gets confusing because depending on the order of extend events you might have 3 events for PCR 0 here and then some for PCR 7 and then another for PCR 0 etc.

hughsie: Could we just sort them by the PCR index?

plugins/tpm-eventlog/fu-tpm-eventlog.c
hughsie: Gotcha.

plugins/tpm-eventlog/fu-tpm-eventlog.c
hughsie: g_clear_error to the rescue.

plugins/tpm-eventlog/fu-tpm-eventlog.c
hughsie: I added the sorting, it helps a lot.

superm1: Yeah sorting certainly helps.  However an option to use is cheap, take a look at the fixup commit I just added.

plugins/tpm-eventlog/fu-tpm-eventlog.c
```diff
@@ -0,0 +1,119 @@
...
+	if (!g_option_context_parse (context, &argc, &argv, &error)) {
+		/* TRANSLATORS: the user didn't read the man page */
+		g_print ("%s: %s\n", _("Failed to parse arguments"),
+			 error->message);
+		return EXIT_FAILURE;
+	}
+
+	/* set verbose? */
+	if (verbose)
+		g_setenv ("G_MESSAGES_DEBUG", "all", FALSE);
```
superm1: It might make sense to set `FWUPD_TPM_EVENTLOG_VERBOSE` as well here.

plugins/tpm-eventlog/fu-tpm-eventlog.c
```diff
@@ -0,0 +1,119 @@
...
+	if (argc <= 1) {
+		if (!fu_tmp_eventlog_process ("/sys/kernel/security/tpm0/binary_bios_measurements", pcr, &error)) {
+			/* TRANSLATORS: failed to read measurements file */
+			g_print ("%s: %s\n", _("Failed to parse file"),
+				 error->message);
+			return EXIT_FAILURE;
+		}
+		return EXIT_SUCCESS;
+	}
+	for (gint i = 1; i < argc; i++) {
```
superm1: The multiple args is for simpler fuzzer support right?  Since it's such a simple app, why not just support one arg and let the fuzzer call it a bunch of times with mutations instead?  At least that keeps this code simpler.  IE:
```
const gchar *fn = argc <= 1 ? /sys/kernel/security/tpm0/binary_bios_measurements" : argv[1];
if (!fu_tmp_eventlog_process (fn, pcr, &error)) {
	/* TRANSLATORS: failed to read measurements file */
	g_print ("%s: %s\n", _("Failed to parse file"),  error->message);
	return EXIT_FAILURE;
}
```

hughsie: Yp, that works for me. I only did the argv thing to be able to run the binary on all the fuzzer-generated crashes -- but all of those were the parser trying to allocate huge amounts of memory and was easily fixed. Do you want to do it, or me?

superm1: I'll do it, I have one other fixup I'm going to add in a min too.

PR #1645, opened 2019-12-06
fu-install-task: Enforce that device protocol matches metadata

superm1:
Currently devices with a matching GUID will match any metadata that
has the same GUID.

For Dell in the future there will be a situation that the same GUID
is supported by two different FwupdDevice objects but the payload
required for the different objects will vary.

So start matching the protocol and the GUID both to make sure that
the right Fwupd device gets matched from the metadata.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-install-task.c
+++ b/src/fu-install-task.c
@@ -169,6 +169,7 @@ fu_install_task_check_requirements (FuInstallTask *self,
 				    FwupdInstallFlags flags,
 				    GError **error)
 {
+	const gchar *protocol;
 	const gchar *version;
 	const gchar *version_release_raw;
 	const gchar *version_lowest;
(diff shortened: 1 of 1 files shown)


src/fu-install-task.c
```diff
@@ -210,6 +211,21 @@ fu_install_task_check_requirements (FuInstallTask *self,
 		return FALSE;
 	}
 
+	/* does the protocol match */
+	protocol = xb_node_query_text (self->component,
+				       "custom/value[@key='LVFS::UpdateProtocol']",
+				       NULL);
+	if (fu_device_get_protocol (self->device) != NULL && protocol != NULL &&
+	    g_strcmp0 (fu_device_get_protocol (self->device), protocol) != 0) {
+		g_set_error (error,
```
hughsie: do we want a --force escape hatch?

superm1: sure, not a bad idea.  added.

PR #1646, opened 2019-12-09
Wip/hughsie/vendor

hughsie:


hughsie:
@superm1 I'm tempted to tag another release, any reservations?

PR #1649, opened 2019-12-10
fu-engine: Atomically reload configuration files w/ modify-remote

superm1:
Relying upon the inotify event is racy for installed tests and has
been leading to some failures.

Fixes: https://github.com/fwupd/fwupd/issues/1648

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -106,6 +106,9 @@ enum {
 	SIGNAL_LAST
 };
 
+static void
+fu_engine_config_changed_cb (FuConfig *config, FuEngine *self);
+
 static guint signals[SIGNAL_LAST] = { 0 };
 
 G_DEFINE_TYPE (FuEngine, fu_engine, G_TYPE_OBJECT)
@@ -621,7 +624,14 @@ fu_engine_modify_remote (FuEngine *self,
 		return FALSE;
 	}
 	g_key_file_set_string (keyfile, "fwupd Remote", key, value);
-	return g_key_file_save_to_file (keyfile, filename, error);
+	g_debug ("Reloading configuration files");
+	if (!g_key_file_save_to_file (keyfile, filename, error))
+		return FALSE;
+	if (!fu_config_load (self->config, FU_CONFIG_LOAD_FLAG_NONE, error))
+		return FALSE;
+	fu_engine_config_changed_cb (self->config, self);
+
+	return TRUE;
 }
 
 /**


src/fu-engine.c
```diff
@@ -621,7 +621,10 @@ fu_engine_modify_remote (FuEngine *self,
 		return FALSE;
 	}
 	g_key_file_set_string (keyfile, "fwupd Remote", key, value);
-	return g_key_file_save_to_file (keyfile, filename, error);
+	g_debug ("Reloading configuration files");
+	if (!g_key_file_save_to_file (keyfile, filename, error))
+		return FALSE;
+	return fu_config_load (self->config, FU_CONFIG_LOAD_FLAG_NONE, error);
```
hughsie: ohh, I don't think it's that easy. `fu_config_load()` sets up all kind of stuff that's deallocated if it's re-used. I think you need something exported like `fu_quirks_check_silo` which uses `xb_silo_is_valid()`...

superm1:
My concern with this change is might actually lead to conffiles reloading several times, so might need to have a way to turn off inotify and back on.

src/fu-engine.c
superm1: Well that's exactly what fu-config.c does on the `fu_config_monitor_changed_cb`....

hughsie: In other news, you just discovered a nice memory leak in the existing code :) Want me to split up fu_config_load() for you?

superm1: sure go ahead :)

superm1:
so interestingly enough Fedora seems to have failed CI the same way that Debian was `No updatable devices` in both of it's above CI runs even with this atomic reload.  So maybe that's not root cause, but maybe it's more readily reproducible now..

superm1:
Looks like specifically it's missing the `supported` flag both in the `get-devices` and `get-updates` calls.
```
                           ? Supported on remote server
```

superm1:
@hughsie OK so with some debugging commits and looking over a lot of logs, I think I've got a clean solution.  Memory leak aside; I think it's the right thing to do to make it an atomic transaction.

After your split up from `fu-config` -> `fu-config`, `fu-remote` minor changes of course.

if you agree on this, I'll backport it into Debian and we can monitor if it actually has helped the issue without the larger/riskier split up.

hughsie:
I think it still leaks. My gut instinct is we fix it properly and then just tag a new tarball...

superm1:
> I think it still leaks. My gut instinct is we fix it properly and then just tag a new tarball...

Yes it leaks still, no different there. I was just thinking lighter fix for the CI problem for backporting, but if you want to do both in one swoop and tag that's fine to me.

superm1:
superceded by #1651

PR #1650, opened 2019-12-10
Split out the remote loading from FuConfig

hughsie:
It's confusing to have FuConfig load both the daemon.conf file and also keep
track of the enabled remotes. It's also wasteful of memory to keep the GKeyFile
alive the entire time.

Logically these are different pools of information and should be managed by
different objects. This allows us to implement reload() in a sane way.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -46,6 +46,7 @@
 #include "fu-plugin-list.h"
 #include "fu-plugin-private.h"
 #include "fu-quirks.h"
+#include "fu-remote-list.h"
 #include "fu-smbios-private.h"
 #include "fu-udev-device-private.h"
 #include "fu-usb-device-private.h"
@@ -69,6 +70,7 @@ struct _FuEngine
 	GUdevClient		*gudev_client;
 #endif
 	FuConfig		*config;
+	FuRemoteList		*remote_list;
 	FuDeviceList		*device_list;
 	FwupdStatus		 status;
 	gboolean		 tainted;
@@ -400,7 +402,7 @@ fu_engine_set_release_from_appstream (FuEngine *self,
 	remote_id = xb_node_query_text (component, "../custom/value[@key='fwupd::RemoteId']", NULL);
 	if (remote_id != NULL) {
 		fwupd_release_set_remote_id (rel, remote_id);
-		remote = fu_config_get_remote_by_id (self->config, remote_id);
+		remote = fu_remote_list_get_by_id (self->remote_list, remote_id);
 		if (remote == NULL)
 			g_warning ("no remote found for release %s", version_rel);
 	}
@@ -566,7 +568,7 @@ fu_engine_modify_config (FuEngine *self, const gchar *key, const gchar *value, G
 	}
 
 	/* modify, effective next reboot */
-	return fu_config_modify_and_save (self->config, key, value, error);
+	return fu_config_set_key_value (self->config, key, value, error);
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/fu-remotes.c
```diff
@@ -0,0 +1,504 @@
...
+	XbSilo			*silo;
+	GHashTable		*os_release;
+};
+
+G_DEFINE_TYPE (FuRemotes, fu_remotes, G_TYPE_OBJECT)
+
+static void
+fu_remotes_emit_changed (FuRemotes *self)
+{
+	g_debug ("::remotesuration changed");
```
superm1: what exactly is a `remotesuration`?

superm1: I'm guessing a find and replace of `config` to `remote` that applied to `configuration`?

src/fu-remotes.c
```diff
@@ -0,0 +1,503 @@
...
+	g_signal_emit (self, signals[SIGNAL_CHANGED], 0);
+}
+
+static GPtrArray *
+fu_remotes_get_all_paths (void)
+{
+	GPtrArray *paths = g_ptr_array_new_with_free_func (g_free);
+	const gchar *remotes_dir;
+	const gchar *system_prefixlibdir = "/usr/lib/fwupd";
```
superm1: I know this was hardcoded before, but I think that it should actually be added in via meson define.

hughsie:
I'm not sure I love the `FuRemotes` name. Maybe `FuRemoteLoader` or `FuRemoteList` might be better.

src/fu-engine.c
```diff
@@ -594,7 +596,7 @@ fu_engine_modify_remote (FuEngine *self,
 	g_autoptr(GKeyFile) keyfile = g_key_file_new ();
 
 	/* check remote is valid */
-	remote = fu_config_get_remote_by_id (self->config, remote_id);
+	remote = fu_remotes_get_by_id (self->remotes, remote_id);
 	if (remote == NULL) {
 		g_set_error (error,
 			     FWUPD_ERROR,
```
superm1: So this is still going to rely upon the `inotify` for the remote change event.  How about calling `fu_remotes_reload` directly in this function and maybe destroying/recreating the `GFileMonitor`?

src/fu-remotes.c
```diff
@@ -0,0 +1,503 @@
...
+fu_remotes_finalize (GObject *obj)
+{
+	FuRemotes *self = FU_REMOTES (obj);
+
+	if (self->os_release != NULL)
+		g_hash_table_unref (self->os_release);
+	if (self->silo != NULL)
+		g_object_unref (self->silo);
+	g_ptr_array_unref (self->remotes);
+	g_ptr_array_unref (self->monitors);
```
superm1: Do the monitors need to be cancelled before unref, or is that implict?

superm1:
> I'm not sure I love the `FuRemotes` name. Maybe `FuRemoteLoader` or `FuRemoteList` might be better.

I like `FuRemoteList`, it matches more  closely to `FuDeviceList`

src/fu-remotes.c
hughsie: They're cancelled on unref.

src/fu-remotes.c
hughsie: It's an old 16 century English word.... I'm bullshitting. It's search and replace :) Fixed.

src/fu-remotes.c
hughsie: This is actually deliberate. We add the ones in FU_PATH_KIND_SYSCONFDIR_PKG first, and then fall back to the per-system ones. This means if you run fwupd in a development prefix you still pick up the system ones so that the daemon actually works properly.

superm1: Even with systemd these days? Systemd environment overrides work without this I think.

src/fu-engine.c
hughsie: I think we can just create a `fu_remote_list_set_key_value()` and make it simpler still. Fixed.

src/fu-remotes.c
hughsie: Lets keep the behaviour the same for this PR and then try to rip the manual path out in a follow up PR.

superm1:
This should close #1648

PR #1651, opened 2019-12-10
Wip/hughsie/vendor

hughsie:
We probably ought to get a lot more serious about setting the vendor ID on LVFS vendor accounts. Notable omissions are ATA and Redfish, both of which not supplying anything like a vendor ID. Ideas welcome for those two.

### DIFF
--- a/plugins/coreboot/README.md
+++ b/plugins/coreboot/README.md
@@ -52,3 +52,8 @@ The following HWIDs are added on coreboot enabled platforms:
 
 They do match the values provided by `fwupdtool hwids` or
 the `ComputerHardwareIds.exe` Windows utility.
+
+Vendor ID Security
+------------------
+
+The vendor ID is set from the BIOS vendor, in this instance `DMI:coreboot`
--- a/plugins/coreboot/fu-plugin-coreboot.c
+++ b/plugins/coreboot/fu-plugin-coreboot.c
@@ -91,6 +91,7 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 	}
 	fu_device_set_vendor (dev, fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_MANUFACTURER));
 	fu_device_add_instance_id (dev, "main-system-firmware");
+	fu_device_set_vendor_id (dev, "DMI:coreboot");
 
 	for (guint i = 0; i < G_N_ELEMENTS (hwids); i++) {
 		char *str;
--- a/plugins/dell/fu-plugin-dell.c
+++ b/plugins/dell/fu-plugin-dell.c
@@ -286,6 +286,7 @@ fu_plugin_dock_node (FuPlugin *plugin, const gchar *platform,
 		dock_name = g_strdup_printf ("Dell %s", dock_type);
 	}
 	fu_device_set_vendor (dev, "Dell Inc.");
+	fu_device_set_vendor_id (dev, "PCI:0x1028");
 	fu_device_set_name (dev, dock_name);
 	fu_device_set_metadata (dev, FU_DEVICE_METADATA_UEFI_DEVICE_KIND, "device-firmware");
 	if (type == DOCK_TYPE_TB16) {
@@ -734,6 +735,7 @@ fu_plugin_dell_detect_tpm (FuPlugin *plugin, GError **error)
 	fu_device_add_instance_id (dev, tpm_guid_raw);
 	fu_device_add_instance_id (dev, "system-tpm");
 	fu_device_set_vendor (dev, "Dell Inc.");
+	fu_device_set_vendor_id (dev, "PCI:0x1028");
 	fu_device_set_name (dev, pretty_tpm_name);
 	fu_device_set_summary (dev, "Platform TPM device");
 	fu_device_set_version (dev, version_str, FWUPD_VERSION_FORMAT_QUAD);
@@ -768,6 +770,7 @@ fu_plugin_dell_detect_tpm (FuPlugin *plugin, GError **error)
 		fu_device_set_id (dev_alt, tpm_id_alt);
 		fu_device_add_instance_id (dev_alt, tpm_guid_raw_alt);
 		fu_device_set_vendor (dev, "Dell Inc.");
+		fu_device_set_vendor_id (dev, "PCI:0x1028");
 		fu_device_set_name (dev_alt, pretty_tpm_name_alt);
 		fu_device_set_summary (dev_alt, "Alternate mode for platform TPM device");
 		fu_device_add_flag (dev_alt, FWUPD_DEVICE_FLAG_INTERNAL);
--- a/plugins/uefi-recovery/fu-plugin-uefi-recovery.c
+++ b/plugins/uefi-recovery/fu-plugin-uefi-recovery.c
@@ -35,6 +35,7 @@ gboolean
 fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 {
 	GPtrArray *hwids = fu_plugin_get_hwids (plugin);
+	const gchar *dmi_vendor;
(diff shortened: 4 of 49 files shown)


plugins/coreboot/fu-plugin-coreboot.c
```diff
@@ -92,6 +93,13 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 	fu_device_set_vendor (dev, fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_MANUFACTURER));
 	fu_device_add_instance_id (dev, "main-system-firmware");
 
+	/* set vendor ID as the BIOS vendor */
+	dmi_vendor = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_BIOS_VENDOR);
```
superm1: Isn't this hardcoded to "coreboot"`  (see `fu_plugin_startup`)?

plugins/dell/fu-plugin-dell.c
```diff
@@ -286,6 +286,7 @@ fu_plugin_dock_node (FuPlugin *plugin, const gchar *platform,
 		dock_name = g_strdup_printf ("Dell %s", dock_type);
 	}
 	fu_device_set_vendor (dev, "Dell Inc.");
+	fu_device_set_vendor_id (dev, "TPM:DELL");
```
superm1: Actually I think 0x1028 is fine to use here instead.
https://devicehunt.com/view/type/pci/vendor/1028

plugins/coreboot/README.md
```diff
@@ -52,3 +52,8 @@ The following HWIDs are added on coreboot enabled platforms:
 
 They do match the values provided by `fwupdtool hwids` or
 the `ComputerHardwareIds.exe` Windows utility.
+
+Vendor ID Security
+------------------
+
+The vendor ID is set from the BIOS vendor, for example `DMI:Google`
```
superm1: Not sure about that

superm1:
I think you should add this to the new plugin checklist too.

plugins/uefi-recovery/fu-plugin-uefi-recovery.c
```diff
@@ -49,6 +50,14 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 		const gchar *hwid = g_ptr_array_index (hwids, i);
 		fu_device_add_guid (device, hwid);
 	}
+
+	/* set vendor ID as the BIOS vendor */
+	dmi_vendor = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_BIOS_VENDOR);
+	if (dmi_vendor != NULL) {
+		g_autofree gchar *vendor_id = g_strdup_printf ("DMI:%s", dmi_vendor);
+		fu_device_set_vendor_id (FU_DEVICE (dev), vendor_id);
```
superm1: `FU_DEVICE (dev)` -> `device`

superm1:
> We probably ought to get a lot more serious about setting the vendor ID on LVFS vendor accounts.

Need to think about when the checks happen.  I would think upload time is the right time, especially with the ability to upload into from ODM or IHV into OEM accounts but OEM does the testing and promotion to stable.

I would think that this needs to be a crawl, walk, run approach, maybe start out enforcing one vendor checks on one protocol, then expand.

> Notable omissions are ATA and Redfish

Anyone actually using Redfish yet?  I think in most implementations it should match the system vendor, but I don't know if that's true when you worry about payloads for other devices in the chassis.

ATA I think you can pull vendor from the identify block and strip that (it's already done for GUID anyway).

plugins/coreboot/fu-plugin-coreboot.c
hughsie: Ahh, good catch. Thanks.

plugins/dell/fu-plugin-dell.c
hughsie: I pondered PCI:0x1028 like the other Dell device, but this is logically a TPM device, and it's a vendor ID of DELL. I know DELL isn't a TPM vendor according to the spec, but in this case I think it's valid. That said, it's probably your decision which you want to use. :) What do you prefer?

plugins/coreboot/README.md
hughsie: Right, fixed, thanks.

plugins/uefi-recovery/fu-plugin-uefi-recovery.c
hughsie: Fixed, thanks.

hughsie:
> Need to think about when the checks happen.

This has to be client side. The namespace checks are the serverside checks, but this is the extra security step that matches the vendor uploading the firmware to the physical hardware VID. Vendors don't actually get to specify the vendor ID in the metainfo file at all, it gets magically added when generating the metadata.

> I would think that this needs to be a crawl, walk, run approach

We're in _walk_ already. I'd say about 80% of the vendors on the LVFS already set the vendor ID, but notably absent from that list are any vendors that have to deploy UEFI capsules. I'm hoping with this patchset we can make all the vendors "locked down" because at the moment Wacom can't flash Logitech hardware, but Dell certainly can.

> Anyone actually using Redfish yet? I think in most implementations it should match the system vendor, but I don't know if that's true when you worry about payloads for other devices in the chassis.

I know HPE was using it, but they've gone very quiet recently. Maybe we don't care for now.

> ATA I think you can pull vendor from the identify block and strip that

How? I didn't see anything suitable in the IDENTIFY block at all.

hughsie:
> I think you should add this to the new plugin checklist too.

Good idea, done.

plugins/dell/fu-plugin-dell.c
superm1: Well the thing that is vendor ID Dell is for representing the "Capsule".  The TPM is a Nuvoton or Infineon or ST Micro, etc.

superm1: And "Dell" would be the one uploading said payload to LVFS not one of the TPM vendors.

superm1:
> This has to be client side. The namespace checks are the serverside checks, but this is the extra security step that matches the vendor uploading the firmware to the physical hardware VID. Vendors don't actually get to specify the vendor ID in the metainfo file at all, it gets magically added when generating the metadata.

Well this is fine for IHV's who work on a single component or subsystem but not for OEM's.  For example Dell uploads Thunderbolt FW, UEFI, Dell Dock etc.  It can't be a blanket Vendor = VID.

> because at the moment Wacom can't flash Logitech hardware, but Dell certainly can.

Which I fully expect.  If Dell had a co-branded part with Logitech and Dell distributed the EXE files for Windows and fwupd payloads for Linux, Dell might actually be uploading to LVFS.
(Don't read into this example; it's hypothetical) 

TLDR;  my point is OEM != IHV.

superm1:
>How? I didn't see anything suitable in the IDENTIFY block at all.

I thought that was how a GUID was built, from something in there.  but maybe that's just a model number.

plugins/dell/fu-plugin-dell.c
hughsie: > is for representing the "Capsule".

That's a compelling argument. I'll switch it round.

hughsie:
>  but maybe that's just a model number

Some vendors do put the vendor and model there (split by space, [0] etc), but the majority just include a model number. The vendor tools seem to "know" which model prefixes are which drive vendors.

hughsie:
>  It can't be a blanket Vendor = VID.

Ohh, I agree it's not a 1:1 mapping at all. I'd fully expect Dell to have half a dozen vendor IDs assigned -- including that of synaptics and wacom for instance. What Dell wouldn't have is the vendor ID for Lenovo or Star Labs etc, and I think any way we can limit the horror story of a malicious vendor or user on the LVFS is a good step.

Issue #1617, opened 2019-11-30
fwupdmgr asked me if i wanted to reboot, i said yes, and nothing happened


albert-astals-cid-kdab:
To help us pinpoint your issue, please insert the output of the
following commands when ran on the system with the issue:

```shell
$ fwupdmgr --version
client version: 1.3.5
compile-time dependency versions
        gusb:   0.3.1
        efivar: 37
daemon version: 1.3.5
```

```shell
$ fwupdmgr get-devices
$ LC_ALL=en_US.UTF-8 fwupdmgr get-devices
XPS 15 9570
│
├─Thunderbolt Controller:
│     Device ID:           6075dae7d92d8a13e157f179fe41d008c140c71a
│     Summary:             Unmatched performance for high-speed I/O
│     Current version:     33.00
│     Vendor:              Dell (TBT:0x00D4)
│     GUIDs:               560774f3-821b-5ebb-9f1a-97cf34ce266f
│                          d3826540-45c3-5f3c-89d9-b21b27c6ad77
│     Device Flags:        • Internal device
│                          • Updatable
│                          • Requires AC power
│                          • Device stages updates
│   
├─KXG50ZNV1T02 NVMe TOSHIBA 1024GB:
│     Device ID:           f2759da7fe8e0388c5f3601cb072f837b1070b03
│     Summary:             NVM Express Solid State Drive
│     Current version:     AADA4105
│     Vendor:              Toshiba America Info Systems (NVME:0x1179)
│     GUIDs:               af8d42da-167c-546a-9294-c1502f6ccbab
│                          4d0aed03-a30c-52c6-99e7-a8977797c3d9
│                          2ce5641c-516e-5c6b-908f-6279409807b5
│     Device Flags:        • Internal device
│                          • Updatable
│                          • Requires AC power
│                          • Needs a reboot after installation
│                          • Device is usable for the duration of the update
│   
├─System Firmware:
│     Device ID:           66352e92895dc6d018071ad4ecfae249b0b8af0d
│     Current version:     0.1.13.0
│     Minimum Version:     0.1.13.0
│     Vendor:              Dell Inc.
│     GUID:                e4d7bdc4-cc95-4aa0-a982-5e915fc04bf0
│     Device Flags:        • Internal device
│                          • Updatable
│                          • Requires AC power
│                          • Supported on remote server
│                          • Needs a reboot after installation
│                          • Cryptographic hash verification is available
│                          • Device is usable for the duration of the update
│   
[...]

PR #1652, opened 2019-12-11
fu-util-common: correct the default prompt for reboot/shutdown (Fixes…

superm1:
…: #1617)

Y/n means that Y is the default if you press enter.
y/N means that N is the default if you press enter.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


albert-astals-cid-kdab:
Being annoying, i would suggest to move the printing of the Y|n or y|N to be inside the fu_util_prompt_for_boolean function and thus it would always "be correct".

But not my code to maintain, so feel free to ignore me :)

Issue #1659, opened 2019-12-12
Unable to update the firmware of an logitech unify reciver modell  U8000


tuxmaster5000:
```shell
$ fwupdmgr --version
client version:	1.2.11
compile-time dependency versions
	gusb:	0.3.0
	efivar:	37
daemon version:	1.2.11
```

```shell
$ fwupdmgr get-devices
System Firmware
  DeviceId:             123fd4143619569d8ddb6ea47d1d3911eb5ef07a
  Guid:                 230c8b18-8d9b-53ec-838b-6cfc0383493a
  Plugin:               uefi
  Flags:                internal|require-ac|registered|needs-reboot
  Vendor:               
  Version:              MKQ7710H.86A.0074.2018.1025.1727
  VersionFormat:        plain
  Icon:                 computer
  Created:              2019-12-12
  UpdateError:          Firmware can not be updated in legacy mode, switch to UEFI mode.

Hitachi HDS723020BLA642
  DeviceId:             602b0a6cc821d155208724f0e22f8d111542b74c
  Guid:                 c5e0df58-c081-5c67-a67b-69cfc5a481f6
  Guid:                 461ac46c-5439-5b77-b1f2-a75d024ea620
  Guid:                 55243e86-0ca6-5247-8525-00e93fa160b9
  Summary:              ATA Drive
  Plugin:               ata
  Flags:                internal|updatable|require-ac|registered|needs-reboot
  Version:              MN6OAA10
  VersionFormat:        plain
  Icon:                 drive-harddisk
  Created:              2019-12-12

SAMSUNG MZ7TD256HAFV-000L9
  DeviceId:             0a8c36d4c09c803cd6b5861e443fb7a41a20cbe6
  Guid:                 543ddeb8-1d9f-5b7e-80d3-5aafdbe24534
  Guid:                 dabb7316-b923-527b-9c4c-9855198795f4
  Guid:                 ef0f5dcf-3f29-5580-9961-dabce82231fe
  Summary:              ATA Drive
  Plugin:               ata
  Flags:                internal|updatable|require-ac|registered|needs-reboot
  Version:              DXT02L5Q
  VersionFormat:        plain
  Icon:                 drive-harddisk
  Created:              2019-12-12

HDT722516DLA380
  DeviceId:             7466579651466dd8b56d7a19479bed768440026d
  Guid:                 4f9b4da4-014f-5ac3-838c-2e7dccf2cdd7
[...]

Issue #1261, opened 2019-07-31
unifying: need better messaging for unsupported hardware


meliot:
fwupdmgr update not working, according to the error message the bootloader is unsupported:
```shell
meliot@ubuntu:~/fwupd$ fwupdmgr update
No releases found for device: Not compatible with bootloader version: failed predicate [BOT01.0[0-3]_* regex BOT01.04_B0016]
```

Thanks for your amazing work.
meliot

---

Requested information:
```shell
$ fwupdmgr --version
client version:	1.0.9
daemon version:	1.2.10
compile-time dependency versions
	appstream-glib:	0.7.7
	gusb:	0.2.11
	fwupdate:	10
	efivar:	34

```

```shell
$ fwupdmgr get-devices
Unifying Receiver
[..]
  Summary:              A miniaturised USB wireless receiver (bootloader)
  Plugin:               unifying
  Flags:                updatable|supported|registered|is-bootloader
  Vendor:               Logitech
  VendorId:             USB:0x046D
  Version:              RQR00.00_B0000
  VersionBootloader:    BOT01.04_B0016
  Icon:                 preferences-desktop-keyboard
  Created:              2019-07-31
```

Please answer the following questions:

- Operating system and version: PRETTY_NAME="Ubuntu 18.04.2 LTS"
- How did you install fwupd (ex: `from source`, `pacman`, `apt-get`, etc): snap
- Have you tried rebooting? yep
- Are you using an NVMe disk?

FFY00:
We need to change the way this is implemented so that fwupdmgr doesn't spit out errors like that.

hughsie:
I think the issue here is that we have the same failure mode for "we can't install the component that matches the hardware because of some underlying problem" and "we can't install the component that matches the hardware because it doesn't really match the hardware". i.e. one is a known failure the user needs to be aware of, and the other is a known restriction that's perfectly normal.

meliot:
Thanks for the update. In few words this means that the message will be fixed to show "unsupported hardware" and that the hardware will not be supported in the next future? 
Thank you

superm1:
The exact specifics of how messaging will be fixed up I'll defer to @FFY00 and @hughsie.  I was just triaging the issues some.

rgpublic:
Sorry for the spam, but I have the same question as @meliot . What does it now actually mean for end-users? That you cannot update the firmware of newer Logitech Unifiying Receivers with fwupd and they should be updated with Windows for now, right?

[...]

PR #1660, opened 2019-12-12
engine: don't expose bootloader version errors to users

superm1:
These messages are useful for debugging but not to users.  Many issues
get filed because they're cryptic and users can't do anything about them.

Downgrade the actual messages to debugging in the engine, and just send
back a generic message for clients.

Fixes: #1659
Fixes: #1261

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1119,15 +1119,16 @@ fu_engine_check_requirement_firmware (FuEngine *self, XbNode *req,
 			if (g_strcmp0 (xb_node_get_attr (req, "compare"), "ge") == 0) {
 				g_set_error (error,
 					     FWUPD_ERROR,
-					     FWUPD_ERROR_INVALID_FILE,
+					     FWUPD_ERROR_NOT_SUPPORTED,
 					     "Not compatible with bootloader version %s, requires >= %s",
-					     version, xb_node_get_attr (req, "version"));
+                                             version, xb_node_get_attr (req, "version"));
+
 			} else {
-				g_set_error (error,
-					     FWUPD_ERROR,
-					     FWUPD_ERROR_INVALID_FILE,
-					     "Not compatible with bootloader version: %s",
-					     error_local->message);
+				g_debug ("Bootloader is not compatible: %s", error_local->message);
+				g_set_error_literal (error,
+						     FWUPD_ERROR,
+						     FWUPD_ERROR_NOT_SUPPORTED,
+						     "Bootloader is not compatible");
 			}
 			return FALSE;
 		}


superm1:
I expect with this change it should look something like this:

```
# fwupdmgr get-updates
No upgrades available for Logitech Unifying: No compatible devices
```

hughsie:
Lets go back a bit, if only for my own sanity. When we do `GetReleases(device_id)` we go through the AppStream components looking for a `provide/firmware` that matches any one of the device GUIDs.

1. If no GUIDs match in the AppStream, we return with `NOTHING_TO_DO`: `No releases found`

2. If the fwupd version is too old to apply the update, we return with `INVALID_FILE`: `No releases found for device: Not compatible with org.freedesktop.fwupd version 1.2.11, requires >= 1.3.1`

3. If the hardware bootloader (or existing) version is too old, then we return with `INVALID_FILE`: `No releases found for device: Not compatible with bootloader version 0.1, requires >= 1.0`

4. If the firmware is unsuitable, e.g. the firmware is doing a perfectly valid `requires` check on the bootloader version, we'd return with `INVALID_FILE`: `No releases found for device: 
Not compatible with bootloader version: failed predicate [BOT03.0[0-1]_* regex BOT03.02_B0009]`

So, going through each one-by-one.

1. The user might be able do something about this, they could refresh the metadata, or enable `lvfs-testing` -- but it's unlikely this would work.

2. The user can certainly update fwupd, either by upgrading their distro or by using a flatpak or snap package. It might be important to notify the user that a security update is pending, with the release information, even if they can't update to it right now. I don't think essentially hiding the release like we do now is a great solution.

3. The user might be able to fix this, for example booting into Windows 10 to use winflash to get the "old version" up to a version that means capsules and the LVFS can take over. It might also be important to tell the user about this security update even if we can't apply it.

4. There's nothing the user can do about this, it's just really the wrong type of hardware for the firmware, and fwupd should just ignore this and not report anything.

So there are a few bugs to unpack:

 * We need to show releases, and explain to the user why they can't be applied -- ***if*** the user can do something about it

 * We can't start returning non-valid `FwupdRelease`s from `GetUpgrades()` as client software might start thinking that a non-empty list == has_updates to apply. Maybe we can return them from `GetReleases()` instead although it's not super-discoverable.

 * We need to hide "valid" requirement problems the user can't do anything about.

My worry is we don't have enough metadata in the metainfo.xml file that tells us "is the user able to fix (or work around) this". From a fwupd point of view the "has the wrong bootloader" and "you need to update the bootloader using a different tool" has the same error code. Perhaps we just use a _heuristic_ of anything to do with the bootloader = silent failure, anything else: report like we do now.

I guess this is what your patch does; but what I'd rather see is a bootloader failure being reported as a failure (with a `GError` set) but with a different error code that we can filter in the caller.  Maybe even `FWUPD_ERROR_NOT_SUPPORTED`.

Sorry for the brain dump.

superm1:
Quite the brain dump there.  Glad you came to the same conclusions as the PR though :)
I've made the change for error type.

hughsie:
I think the `Not compatible with bootloader version %s, requires` is still useful to the caller -- can you just change the error codes with the same effect?

FFY00:
> If the hardware bootloader (or existing) version is too old, then we return with INVALID_FILE: No releases found for device: Not compatible with bootloader version 0.1, requires >= 1.0

> The user might be able to fix this, for example booting into Windows 10 to use winflash to get the "old version" up to a version that means capsules and the LVFS can take over. It might also be important to tell the user about this security update even if we can't apply it.

No, the bootloader is not updatable. That's the reason we have unsigned and signed versions of the firmware.

> If the firmware is unsuitable, e.g. the firmware is doing a perfectly valid requires check on the bootloader version, we'd return with INVALID_FILE: No releases found for device: Not compatible with bootloader version: failed predicate [BOT03.0[0-1]_* regex BOT03.02_B0009]

> There's nothing the user can do about this, it's just really the wrong type of hardware for the firmware, and fwupd should just ignore this and not report anything.

Yes, in this case we need to tell the user no updates are available.

PR #1663, opened 2019-12-17
Wip/hughsie/vli tiny steps

hughsie:


plugins/vli/fu-vli-usbhub-device.c
```diff
@@ -25,7 +25,6 @@
 struct _FuVliUsbhubDevice
 {
 	FuUsbDevice		 parent_instance;
```
hughsie: Needs to be FuVliDevice.,..

plugins/vli/fu-vli-pd-device.c
```diff
@@ -0,0 +1,466 @@
...
+#include "config.h"
+
+#include "fu-firmware.h"
+
+#include "fu-vli-pd-device.h"
+#include "fu-vli-pd-firmware.h"
+
+struct _FuVliPdDevice
+{
+	FuUsbDevice		 parent_instance;
```
hughsie: FuVliDevice, and needs manually registering...

plugins/vli/fu-vli-pd-device.c
```diff
@@ -241,6 +242,31 @@ fu_vli_pd_device_reset_vl103 (FuVliDevice *device, GError **error)
...
+	if (!fu_device_setup (dev, &error_local)) {
+		if (g_error_matches (error_local,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_FOUND)) {
+			g_debug ("%s", error_local->message);
+		} else {
+			g_warning ("cannot create I²C parade device: %s",
+				   error_local->message);
+		}
+		return TRUE;
```
superm1: Do you actually want to return true in this instance?  I think you've got the return for probe and setup mixed up.  If probing fails, return TRUE, but if setup fails return FALSE is what makes sense to me.

hughsie: The way (I think) this works is that the PD chip is the parent, i.e. FuVliPdDevice and the parade device would be an optional child device. If the child is not found, we still want to be able to update the parent device.

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,668 @@
...
+	g_debug ("getting FW%X version", buf[0]);
+	if (!fu_vli_pd_parade_device_set_offset (self, 0x5000 | buf[0], error))
+		return FALSE;
+	if (!fu_vli_pd_parade_device_i2c_read (self, self->page7_addr, 0x00, buf, sizeof(buf), error))
+		return FALSE;
+
+	/* format as a hex triplet */
+	version_str = g_strdup_printf ("%02X.%02X.%02X", buf[0], buf[1], buf[2]);
+	fu_device_set_version (FU_DEVICE (self), version_str, FWUPD_VERSION_FORMAT_TRIPLET);
+	return TRUE;
```
superm1: since you stopped MCU at start, you don't need to restart the MCU after reading FW version?

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,668 @@
...
+	}
+
+	g_debug ("getting FW%X version", buf[0]);
+	if (!fu_vli_pd_parade_device_set_offset (self, 0x5000 | buf[0], error))
+		return FALSE;
+	if (!fu_vli_pd_parade_device_i2c_read (self, self->page7_addr, 0x00, buf, sizeof(buf), error))
+		return FALSE;
+
+	/* format as a hex triplet */
+	version_str = g_strdup_printf ("%02X.%02X.%02X", buf[0], buf[1], buf[2]);
```
superm1: do you know if this is always BCD?  If not it might be worth sanity checking it.

hughsie: That's a good question. I'll find out from Parade.

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,668 @@
...
+		if ((buf & 0x0C) == 0) {
+			ret = TRUE;
+			break;
+		}
+	}
+	if (!ret) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Wait_9e_TOut");
```
superm1: Not really a user friendly string if this situation happens

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,668 @@
...
+			if ((buf & 0x01) == 0) {
+				ret2 = TRUE;
+				break;
+			}
+		}
+		if (!ret2) {
+			g_set_error (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "Wait_93_TOut");
```
superm1: likewise, not really a user friendly string

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,668 @@
...
+		if ((buf & 0x01) == 0) {
+			ret = TRUE;
+			break;
+		}
+	}
+	if (!ret) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Wait_91_TOut");
```
superm1: same, string

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,668 @@
...
+			return FALSE;
+		fu_device_set_progress_full (FU_DEVICE (self), i, blocks->len);
+	}
+
+	/*  load F/W to SPI ROM */
+	if (!fu_vli_pd_parade_device_enable_mapping (self, error))
+		return FALSE;
+	if (!fu_vli_pd_parade_device_i2c_write (self, self->page_addr, 0x82, 0x20, error))
+		return FALSE; /* Reset_CLT2SPI_Interface */
+	g_usleep (1000 *  100 );
```
superm1: whitespace

plugins/vli/fu-vli-pd-device.c
superm1: But why would you set the custom flag that it has the parade device if it wasn't true?   That seems like a hardcoded relationship.

hughsie: > I think you've got the return for probe and setup mixed up. 

Ohh I see what you mean, sorry. Fixed.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: It seems to be anything. I've switched this for `"%u.%u.%u"`

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Fixed.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Fixed.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Fixed.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Fixed.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Valid question: waiting for VIA about this.

joakim-tjernlund:
Is this ready to be merged now?

superm1:
> Is this ready to be merged now?

I think we're waiting for information about how the MCU gets restarted after it's been stopped and probed at startup.

hughsie:
> about how the MCU gets restarted

Agreed, VLI is awaiting information from Parade.

joakim-tjernlund:
After this is merged, will Lenovo docks work  or is there more to be done?

plugins/vli/fu-vli-pd-parade-device.c
hughsie: We now have a `fu_vli_pd_parade_device_start_mcu()` -- thanks.

hughsie:
> After this is merged, will Lenovo docks work or is there more to be done

I think all the component parts are done, (e.g. cxaudio, parade, synapticsmst, vli etc) but there's still a lot of testing and QA to be done. I'm not sure how much I can share, but I know there 16 dock devices on the roadmap, and all that is obviously taking time to test. Lenovo want to release one archive that updates all the components on a dock rather than lots of updates that update specific parts.

i.e. we're getting there.

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,689 @@
...
+
+#include "fu-vli-pd-device.h"
+#include "fu-vli-pd-parade-device.h"
+
+struct _FuVliPdParadeDevice
+{
+	FuDevice		 parent_instance;
+	FuVliDeviceKind		 device_kind;
+	guint8			 page_addr;
+	guint8			 page7_addr;
```
superm1: Are there better names you can use than page and page7?

plugins/vli/fu-vli-pd-parade-device.c
```diff
@@ -0,0 +1,689 @@
...
+	/* sanity check */
+	if (bufsz > 0x40) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INVALID_FILE,
+			     "request too large");
+		return FALSE;
+	}
+
+	/* page2_address = 0x14/0x94, page7_address = 0x1E/0x9E --
```
superm1: Page2 doesn't seem used anywhere?

superm1:
> i.e. we're getting there.

Is there no EC to worry about updating? It seems like something is missing for power sequencing to me. And if not furthermore I suspect you'll want at least a "dummy" plugin for providing a parent device for all these. Even if the version is never updated it will help show their logical ownership in the tools.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: I'm not quite sure what they are myself. The original code was somewhat, shall we say, _creative_. I'll do some research.

hughsie:
> Is there no EC to worry about updating

On the mini-docks, no. On the bigger ones the EC is being updated via the UsbHub (the MSP IC) so it's only really a blob of memory we send to the USB hub.

superm1:
> being updated via the UsbHub (the MSP IC) so it's only really a blob of memory we send to the USB hub

So on these is there actually a need to install USB hub last or use that to power cycle dock via a mailbox or something?

hughsie:
> So on these is there actually a need to install USB hub last

That is a very good question. I'll ask VIA.

> or use that to power cycle dock via a mailbox or something?

IIUC, you tell the UsbHub to reset and it power cycles the entire hub device. Also, for this code, think "inexpensive USB-C dock", not "$$ TBT dock".

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Fixed.

plugins/vli/fu-vli-pd-parade-device.c
hughsie: Okay, it's the start address of page2 and page7 on the device itself. I've renamed and added some comments.

hughsie:
> There's a lot of magic numbers used in the code for various arguments

I've split out the obvious ones, I think the others have to remain secret I'm afraid.

PR #1666, opened 2019-12-20
Add a runtime warning when adding a device without an vendor-id or pr…

hughsie:
…otocol set

Also, allow setting the protocol in the quirk file if required.


hughsie:
Debian CI seems to be failing; no idea on the warnings :(

PR #1677, opened 2019-12-28
Wip/hughsie/fu engine install all releases

hughsie:


hughsie:
>  Otherwise this falls apart no?

The idea here is that there are some devices (cough, Intel SSDs, cough) that need just one firmware releases to be installed together without rebooting. It's something like this:

1.2.3 -> 1.2.4 = fine
1.2.4 -> 1.2.5 -> 1.2.6 = fine
1.2.4 -> 1.2.5 = bad
1.2.5 -> 1.2.6 = fine
1.2.6 -> 1.2.7 = fine
0.1.1 -> 1.2.5 -> 1.2.6 = fine

So really there's just one "composite" release that needs to have both firmwares in the same cab file. e.g. 1.2.5+1.2.6 come as a pair, and 1.2.7 would be shipped as just one. I assume it's some kind of data migration issue, but Intel say solving this device-side is impossible, hence this patch.

superm1:
> > Otherwise this falls apart no?
> 
> The idea here is that there are some devices (cough, Intel SSDs, cough) that need just one firmware releases to be installed together without rebooting. It's something like this:
> 
> 1.2.3 -> 1.2.4 = fine
> 1.2.4 -> 1.2.5 -> 1.2.6 = fine
> 1.2.4 -> 1.2.5 = bad
> 1.2.5 -> 1.2.6 = fine
> 1.2.6 -> 1.2.7 = fine
> 0.1.1 -> 1.2.5 -> 1.2.6 = fine
> 
> So really there's just one "composite" release that needs to have both firmwares in the same cab file. e.g. 1.2.5+1.2.6 come as a pair, and 1.2.7 would be shipped as just one. I assume it's some kind of data migration issue, but Intel say solving this device-side is impossible, hence this patch.

If this situation is that really there is just one bad firmware release, can't that firmware release be excluded from being posted to LVFS?

hughsie:
> can't that firmware release be excluded from being posted to LVFS

No, as 1.2.5 -> 1.2.6 is doing some kind of block data migration :/

superm1:
> > can't that firmware release be excluded from being posted to LVFS
> 
> No, as 1.2.5 -> 1.2.6 is doing some kind of block data migration :/

OK, but so if you try to go 1.2.3 -> 1.2.6 you will have a failure then, right?   So that's back to my point that you can't safely support this approach unless you somehow guarantee that they go 1.2.3->1.2.5, don't reboot and go to 1.2.6.

hughsie:
> so if you try to go 1.2.3 -> 1.2.6 you will have a failure then, right

Yes, and we'd need some `firmware` requirements with `le=version` to enforce this.

superm1:
> Yes, and we'd need some firmware requirements with le=version to enforce this.

The whole thing feels very fragile and dangerous.  It's really easy to forget to put a requirement like that and mess it up.  The requirements need to get placed on all the releases too this means because you can't predict when the transitions happen.  When you're talking about storage that is a devastating failure.

I think there needs to be sanity checking added in a few places:
1. A check that a version requirement was used when this flag is in place on LVFS
2. A client side quirk for this SSD to force a requirement version checks on all releases.

hughsie:
> The whole thing feels very fragile and dangerous

I think that's harsh, but fair :)

> A check that a version requirement was used when this flag is in place on LVFS

That seems like a nice-to-have, although I think we ought to concentrate on the client side as that's what's actually going to "save" us from disaster.

> A client side quirk for this SSD to force a requirement version checks on all releases.

That's a very good idea, and would be useful for other existing "tricky" devices we have to treat carefully. Something like `FWUPD_DEVICE_FLAG_FW_VERSION_CHECK_REQUIRED` possibly?

superm1:
> FWUPD_DEVICE_FLAG_FW_VERSION_CHECK_REQUIRED 

Sounds good.

hughsie:
> Sounds good.

https://github.com/fwupd/fwupd/pull/1692

hughsie:
> The whole thing feels very fragile and dangerous

More sane now? It's now only settable using a quirk.

hughsie:
>  I guess you'll need another commit to add a quirk for that flag for the SSD

Yes, I'm waiting for hardware now. Thanks.

PR #1690, opened 2020-01-06
Shut down automatically when there is system memory pressure

hughsie:
We can just rescan hardware if required; near OOM it's just more important to
free what we can and get out of the way.

### DIFF
--- a/contrib/ci/Dockerfile-fedora.in
+++ b/contrib/ci/Dockerfile-fedora.in
@@ -1,4 +1,4 @@
-FROM fedora:30
+FROM fedora:31
 %%%OS%%%
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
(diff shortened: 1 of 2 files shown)


superm1:
Is something in CI compiling this code yet?  if so, LGTM.

hughsie:
> Is something in CI compiling this code yet

No, but I could grab the glib2 from F32 in the F31 Fedora build if you think that's a good idea.

superm1:
> > Is something in CI compiling this code yet
> 
> No, but I could grab the glib2 from F32 in the F31 Fedora build if you think that's a good idea.

Yeah I think so.

It's not yet in Debian unstable or Arch yet either.
Anytime we put in `#ifdef` that's based upon something new like that it's probably a good idea to make sure something is at least compiling the code.

PR #1693, opened 2020-01-07
trivial: synaptics-mst: skip self tests for systems with amdgpu

superm1:
No need to fail these self tests when using amdgpu, just skip them.
Fixes unrelated issue found in #1183

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

CC @NTMan


hughsie:
I've slightly unwound this and pushed it as 7442b7ba778f3166789a9d515374d2b4829f0f68 -- many thanks.

PR #1701, opened 2020-01-13
Allow getting the list of updates in JSON format from fwupdagent

hughsie:
Apparently this is wanted by the SOS team too.

### DIFF
--- a/data/bash-completion/fwupdagent
+++ b/data/bash-completion/fwupdagent
@@ -1,5 +1,7 @@
 _fwupdagent_cmd_list=(
 	'get-devices'
+	'get-updates'
+	'get-upgrades'
 )
 
 _fwupdagent_opts=(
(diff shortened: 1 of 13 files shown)


data/bash-completion/fwupdagent
```diff
@@ -1,5 +1,6 @@
 _fwupdagent_cmd_list=(
 	'get-devices'
+	'get-updates'
```
superm1: Missing get upgrades

src/fu-agent.c
```diff
@@ -113,6 +155,50 @@ fu_util_get_devices (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	}
+
+	/* create header */
+	builder = json_builder_new ();
+	json_builder_begin_object (builder);
+	if (!fu_util_add_updates_json (priv, builder, error))
+		return FALSE;
+	json_builder_end_object (builder);
+
+
```
superm1: Extra new line

data/bash-completion/fwupdagent
hughsie: Fixed.

src/fu-agent.c
hughsie: Fixed, thanks.

PR #1702, opened 2020-01-14
upower: determine if dbus upower presents along with a valid device

CragW:
Chrome OS doesn't use upower therefore upower plugin ultimately fallen
failed by visting properties, errors are seen as below.

  FuPluginUpower  failed to get OnBattery value, assume on AC power
  FuPluginUPower  Failed to query power type, assume AC power

Add support for checking the DBus in response to UPower.EnumerateDevices
at least one device presents.

Signed-off-by: Crag Wang <crag_wang@dell.com>

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm-eventlog/fu-tpm-eventlog-common.c
+++ b/plugins/tpm-eventlog/fu-tpm-eventlog-common.c
@@ -26,7 +26,7 @@ fu_tpm_eventlog_pcr_to_string (gint pcr)
 	if (pcr == 6)
 		return "State transitions and wake events";
 	if (pcr == 7)
-		return "Platform manufacturer specifc measurements";
+		return "Platform manufacturer specific measurements";
 	if (pcr >= 8 && pcr <= 15)
 		return "Static operating system";
 	if (pcr == 16)


plugins/upower/fu-plugin-upower.c
```diff
@@ -33,6 +33,42 @@ fu_plugin_destroy (FuPlugin *plugin)
...
+				      -1,
+				      NULL,
+				      error);
+
+	// Check if D-Bus proxy reponses
+	if (val == NULL || !error )
+		return FALSE;
+
+	// Check the type of return in response to EnumerateDevices()
+	if (!g_variant_is_of_type (val, G_VARIANT_TYPE("(ao)"))){
```
CragW: I tried to compare val with type [G_VARIANT_TYPE_OBJECT_PATH_ARRAY](https://developer.gnome.org/glib/stable/glib-GVariantType.html#G-VARIANT-TYPE-OBJECT-PATH-ARRAY:CAPS) but the return in fact is "(ao)" instead of "ao" that was pre-defined in glib2.

Given that inconsistency I use G_VARIANT_TYPE with manual format at this place.

hughsie:
I think what you're checking for is the runtime presence of upower, rather than the presence of devices (which could come and go). What about if we used `g_bus_watch_name()` to watch `org.freedesktop.UPower` and in `GBusNameVanishedCallback` we set a flag that would cause the check to be neutered?

CragW:
Yes the main focus here is to determine the presence of upower service, your suggestion sounds good to me. I will make the modifications in another PR.

superm1:
You can just force push this branch and this PR will update automatically. No need to close and open a new one.

PR #1703, opened 2020-01-15
Don't check the battery percentage if UPwer is unavailable

hughsie:
Based on a patch by Crag Wang <Crag.Wang@dell.com>, many thanks.

### DIFF
--- a/plugins/upower/fu-plugin-upower.c
+++ b/plugins/upower/fu-plugin-upower.c
@@ -50,6 +50,14 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 		g_prefix_error (error, "failed to connect to upower: ");
 		return FALSE;
 	}
+	if (g_dbus_proxy_get_name_owner (data->upower_proxy) == NULL) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "no owner for %s",
+			     g_dbus_proxy_get_name (data->upower_proxy));
+		return FALSE;
+	}
 	data->display_proxy =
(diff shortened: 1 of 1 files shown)


hughsie:
@Crag.Wang does this work for you?

superm1:
I think it's better to do this in startup and disable the plugin no?  Or are you concerned for race condition then?

hughsie:
>  Or are you concerned for race condition

Well, less a race condition, more worried about the startup order. I guess that's less of an issue here as we're checking the upower state on firmware deployment. Maybe we check the name owner in startup, and if it's NULL we disable the plugin like you suggest?

superm1:
> more worried about the startup order.

Like `fwupd` starts before `upowerd`?  And then `upowerd` doesn't have an owner yet?  This can be mitigated by making `fwupd.service` depend on `upower.service` I would think.

> Maybe we check the name owner in startup, and if it's NULL we disable the plugin like you suggest?

I think that makes more sense, yes.

hughsie:
> This can be mitigated by making fwupd.service depend on upower.service I would think.

I don't think we need to worry about this, as `g_dbus_proxy_new_for_bus_sync()` will start upower for us if required as we're **not** using `G_DBUS_PROXY_FLAGS_DO_NOT_AUTO_START`

> I think that makes more sense, yes.

Okay, fixed up with that -- and it seems to work too.

CragW:
It works. On Chrome the plugin is disabled as per the owner of UPower doesn't exist.

superm1:
> It works. On Chrome the plugin is disabled as per the owner of UPower doesn't exist.

Great.  I also backported to 1_2_X since chromeos is still on 1.2.x AFAIK.

PR #1706, opened 2020-01-15
Generate a win32 setup binary

hughsie:


contrib/setup-win32.nsi
```diff
@@ -0,0 +1,135 @@
+#!Nsis Installer Command Script
+#
+# To build an installer from the script you would normally do:
+#
+#   dnf install mingw32-nsis
+#   makensis setup-win32.nsi
+
+Name ""
+OutFile "fwupd-setup-86_64.exe"
```
superm1: x86_64 no?

contrib/setup-win32.nsi
```diff
@@ -0,0 +1,135 @@
...
+  File "share/fwupd/quirks.d/rts54hid.quirk"
+  File "share/fwupd/quirks.d/rts54hub.quirk"
+  File "share/fwupd/quirks.d/solokey.quirk"
+  File "share/fwupd/quirks.d/steelseries.quirk"
+  File "share/fwupd/quirks.d/synaptics-cxaudio.quirk"
+  File "share/fwupd/quirks.d/synaptics-prometheus.quirk"
+  File "share/fwupd/quirks.d/vli-pd.quirk"
+  File "share/fwupd/quirks.d/vli-usbhub-lenovo.quirk"
+  File "share/fwupd/quirks.d/vli-usbhub.quirk"
+  File "share/fwupd/quirks.d/wacom-usb.quirk"
```
superm1: This large list of files, can we build it dynamically by just looking at stuff in dist/ perhaps instead?  Otherwise it's another place to forget to add things..

contrib/setup-win32.nsi
```diff
@@ -0,0 +1,135 @@
...
+  ReadEnvStr $0 COMSPEC
+  SetOutPath "$INSTDIR"
+SectionEnd
+
+Section "Start Menu Shortcuts"
+  SetOutPath "$INSTDIR\bin"
+  CreateDirectory "$SMPROGRAMS\fwupd"
+  CreateShortCut "$SMPROGRAMS\fwupd\Uninstall fwupd.lnk" "$INSTDIR\Uninstall fwupd.exe" "" "$INSTDIR\Uninstall fwupd.exe" 0
+  SetOutPath "$INSTDIR"
+  CreateShortCut "$SMPROGRAMS\fwupd\fwupdtool.lnk" "$INSTDIR\bin\fwupdtool.exe" "" "$INSTDIR\bin\fwupdtool.exe" 0
```
superm1: I don't think there is value in this being a shortcut.  It needs command line arguments, it shouldn't show up in the start menu.

contrib/setup-win32.nsi
```diff
@@ -0,0 +1,135 @@
...
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libusb-1.0.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libwinpthread-1.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libxml2-2.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/zlib1.dll"
+
+  # fwupd
+  File "dfu-tool.exe"
+  File "fwupd/fwupdagent.exe"
+  File "fwupd/fwupdtool.exe"
+  File "gcab.exe"
```
superm1: I don't think we want gcab.exe installed

contrib/setup-win32.nsi
```diff
@@ -0,0 +1,135 @@
...
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libtasn1-6.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libunistring-2.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libusb-1.0.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libwinpthread-1.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libxml2-2.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/zlib1.dll"
+
+  # fwupd
+  File "dfu-tool.exe"
+  File "fwupd/fwupdagent.exe"
```
superm1: Nor do we want fwupdagent.exe installed

superm1:
I guess initially we want to also capture the artifact .EXE installer to CircleCI:
https://circleci.com/docs/2.0/artifacts/
I've pushed a commit that would hopefully do this.

And then if it's in good enough shape to actually attach to fwupd releases on Github:
https://circleci.com/blog/automate-your-nuxt-js-app-deployment/

Is what we probably need next.

contrib/setup-win32.nsi
superm1: Adjusted to `x86_64.exe` myself.

contrib/setup-win32.nsi
```diff
@@ -0,0 +1,135 @@
+#!Nsis Installer Command Script
+#
+# To build an installer from the script you would normally do:
+#
+#   dnf install mingw32-nsis
+#   makensis setup-win32.nsi
+
+Name ""
+OutFile "fwupd-setup-x86_64.exe"
```
superm1: In the same vain as other parts of this file that I think should be dynamic - this probably should be something like `fwupd-$(git describe)-setup-x86_64.exe`.

contrib/setup-win32.nsi
hughsie: Globs to the rescue, fixed.

contrib/setup-win32.nsi
hughsie: Agreed, fixed.

hughsie:
Screenshot_Windows10_2020-01-16_10:43:03

superm1:
I think with this assumption that etc is its own directory .. from bin in the installer some directory handling for win32 is wrong. It was done mostly for CI (everything is put in .)

contrib/setup-win32.nsi
superm1: If it is difficult to do it dynamically in this file, a simple `mv` afterwards is fine.

hughsie: How would that work with `store_artifacts`... Can we include a blob there?

superm1: Can certainly include a directory if globs don't work.

hughsie: /me tries...

contrib/setup-win32.nsi
```diff
@@ -0,0 +1,90 @@
...
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libpcre-1.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libpsl-5.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libsoup-2.4-1.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libsqlite3-0.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libtasn1-6.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libunistring-2.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libusb-1.0.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libwinpthread-1.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/libxml2-2.dll"
+  File "/usr/x86_64-w64-mingw32/sys-root/mingw/bin/zlib1.dll"
```
superm1: Can you use a glob on these dlls too?

hughsie: I could, but we only need those. Including all of them adds another ~40Mb to the bundle...

hughsie:
> some directory handling for win32 is wrong

Do we need /etc at all?

superm1:
Yeah for pki stuff or daemon.conf

contrib/setup-win32.nsi
superm1: Definitely makes sense to hardcode list then agree.

hughsie:
> daemon.conf

Hah, so we weren't including that as we are doing a standalone build. I've fixed that up and added it to the setup binary.

superm1:
Regarding 3e57014 would it make sense to rename it to `engine.conf`?

hughsie:
> would it make sense to rename it to engine.conf

I'm hesitant to rename a user-modified file without a very good reason. I'd err on no.

superm1:
> I'm hesitant to rename a user-modified file without a very good reason. I'd err on no.

OK.

superm1:
I've modified the CI job to hopefully publish artifacts to a Github release when a tag is made.  We'll see how if it works on the next release.  If it fails, at least the artifact will also be on CircleCI so it can be manually be done.

Assuming I didn't break CI In the process, this PR LGTM now.

btw, it seems that CircleCI lets you upload any random artifacts you want.  We should keep this in mind in the future when we encounter something hard to debug but CI specific..

PR #1708, opened 2020-01-17
Fixing the "Linux Firmware Updater" bootentry won't be deleted

Ivanhu5866:
Type of pull request:
- [x ] Code fix
- [ ] Feature
- [ ] Documentation

1. Fixing the fwup_delete_variable will get EFI_INVALID_PARAMETER returned.
2. Fixing the bootentry boot#### never been found and hang when checking the bootentry created by fwupd.
3. Get rid of _StrHasPrefix by using CompareMem
https://github.com/vathpela/gnu-efi/issues/7

### DIFF
--- a/plugins/uefi/efi/fwup-efi.c
+++ b/plugins/uefi/efi/fwup-efi.c
@@ -17,16 +17,28 @@ fwup_delete_variable(CHAR16 *name, EFI_GUID *guid)
 {
 	EFI_STATUS rc;
 	UINT32 attrs = 0;
+	UINTN size = 0;
+	_cleanup_free VOID *buf = NULL;
 
 	/* get the attrs so we can delete it */
-	rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, NULL, NULL);
+	rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, &size, NULL);
 	if (EFI_ERROR(rc)) {
-		if (rc == EFI_NOT_FOUND) {
+		if (rc == EFI_BUFFER_TOO_SMALL) {
+			buf = fwup_malloc(size);
+			if (buf == NULL)
+				return EFI_OUT_OF_RESOURCES;
+			rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, &size, buf);
+			if (EFI_ERROR(rc)) {
+				fwup_debug(L"Could not get attr: %r", rc);
+				return rc;
+			}
+		} else if (rc == EFI_NOT_FOUND) {
 			fwup_debug(L"Not deleting variable '%s' as not found", name);
 			return EFI_SUCCESS;
+		} else {
+			fwup_debug(L"Could not get variable '%s' for delete: %r", name, rc);
+			return rc;
 		}
-		fwup_debug(L"Could not get variable '%s' for delete: %r", name, rc);
-		return rc;
 	}
 	return uefi_call_wrapper(RT->SetVariable, 5, name, guid, attrs, 0, NULL);
 }
--- a/plugins/uefi/efi/fwupdate.c
+++ b/plugins/uefi/efi/fwupdate.c
@@ -408,7 +408,7 @@ fwup_delete_boot_entry(VOID)
 		/* check if the variable name is Boot#### */
 		if (CompareGuid(&vendor_guid, &global_variable_guid) != 0)
 			continue;
-		if (StrCmp(variable_name, L"Boot") != 0)
+		if ((StrLen(variable_name) != 8) || (StrnCmp(variable_name, L"Boot", 4) != 0))
 			continue;
 
 		UINTN info_size = 0;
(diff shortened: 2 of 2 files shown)


plugins/uefi/efi/fwup-efi.c
```diff
@@ -17,16 +17,28 @@ fwup_delete_variable(CHAR16 *name, EFI_GUID *guid)
...
+	_cleanup_free VOID *buf = NULL;
 
 	/* get the attrs so we can delete it */
-	rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, NULL, NULL);
+	rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, &size, NULL);
 	if (EFI_ERROR(rc)) {
-		if (rc == EFI_NOT_FOUND) {
+		if (rc == EFI_BUFFER_TOO_SMALL) {
+			buf = fwup_malloc(size);
+			if (buf == NULL)
```
hughsie: why should we allocate the buffer for buf when we don't actually use it?

plugins/uefi/efi/fwup-efi.c
```diff
@@ -17,16 +17,28 @@ fwup_delete_variable(CHAR16 *name, EFI_GUID *guid)
 {
 	EFI_STATUS rc;
 	UINT32 attrs = 0;
+	UINTN size = 0;
+	_cleanup_free VOID *buf = NULL;
 
 	/* get the attrs so we can delete it */
-	rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, NULL, NULL);
+	rc = uefi_call_wrapper(RT->GetVariable, 5, name, guid, &attrs, &size, NULL);
```
hughsie: Is this working around a BIOS bug somewhere? I'm pretty sure NULL works here...

plugins/uefi/efi/fwupdate.c
```diff
@@ -408,7 +408,7 @@ fwup_delete_boot_entry(VOID)
 		/* check if the variable name is Boot#### */
 		if (CompareGuid(&vendor_guid, &global_variable_guid) != 0)
 			continue;
-		if (StrCmp(variable_name, L"Boot") != 0)
+		if ((StrLen(variable_name) != 8) || (StrnCmp(variable_name, L"Boot", 4) != 0))
 			continue;
```
hughsie: I think I'd rather have this as:

    if (StrLen(variable_name) != 8)
      continue;
    if (StrnCmp(variable_name, L"Boot", 4) != 0)
      continue

plugins/uefi/efi/fwupdate.c
```diff
@@ -427,8 +427,9 @@ fwup_delete_boot_entry(VOID)
 		 * check with EFI_LOAD_OPTION description
 		 */
 		EFI_LOAD_OPTION *load_op = (EFI_LOAD_OPTION *) info_ptr;
-		if (_StrHasPrefix(load_op->description, L"Linux Firmware Updater") ||
-		    _StrHasPrefix(load_op->description, L"Linux-Firmware-Updater")) {
+
+		if (_StrHasPrefix(&load_op->description, L"Linux Firmware Updater") ||
+		    _StrHasPrefix(&load_op->description, L"Linux-Firmware-Updater")) {
```
hughsie: Are you sure? isn't this defined as `CHAR16 description[]`?

plugins/uefi/efi/fwup-efi.c
Ivanhu5866: If you have UEFI spec 2.8 you can check the GetVariable returned, p.233.
When DataSize is NULL, UEFI firmeare should return EFI_INVALID_PARAMETER. Means that DataSize cannot be NULL.

hughsie: Okay, thanks for checking.

plugins/uefi/efi/fwup-efi.c
Ivanhu5866: When datasize is 0, will get EFI_BUFFER_TOO_SMALL and UEFI firemware will return the  needed buffer with datasize, and call GetVariable again to get attribute.

hughsie: UEFI really is terrible. Thanks for the sanity check.

plugins/uefi/efi/fwupdate.c
superm1: I kind of think it doesn't matter - I think he should just squash 
1c67712 and 
c54cc1d together anyway rather than ping pong the patches

plugins/uefi/efi/fwupdate.c
Ivanhu5866: Looks good to me.

plugins/uefi/efi/fwupdate.c
Ivanhu5866: I had tested couple Dell machines and the Intel UEFI develop platform "Rainbowpass", with these patches can successfully do the firmware update and delete the Linux Firmware Updater bootentry. Without "&",  got hang, I think it might be from the way to get the variable.

hughsie: ../plugins/uefi/efi/fwupdate.c:430:21: warning: passing argument 1 of ‘_StrHasPrefix’ from incompatible pointer type [-Wincompatible-pointer-types]
  430 |   if (_StrHasPrefix(&load_op->description, L"Linux Firmware Updater") ||
      |                     ^~~~~~~~~~~~~~~~~~~~~
      |                     |
      |                     CHAR16 ** {aka short unsigned int **}
../plugins/uefi/efi/fwupdate.c:370:32: note: expected ‘const CHAR16 *’ {aka ‘const short unsigned int *’} but argument is of type ‘CHAR16 **’ {aka ‘short unsigned int **’}

hughsie:
I've pushed as much of this PR I'm happy with to master. Thanks for identifying both issues.

plugins/uefi/efi/fwupdate.c
Ivanhu5866: Thanks, looks like I used the older version of fwupd. Will check again the latest version fwupd.

PR #1709, opened 2020-01-17
Wip/hughsie/fwupdtool metadata

hughsie:


src/fu-tool.c
```diff
@@ -1599,6 +1599,35 @@ fu_util_get_history (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	if (!fu_util_start_engine (priv, FU_ENGINE_LOAD_FLAG_NONE, error))
+		return FALSE;
+
+	remotes = fu_engine_get_remotes (priv->engine, error);
+	if (remotes == NULL)
+		return FALSE;
+	if (remotes->len == 0) {
+		/* TRANSLATORS: no repositories to download from */
+		g_print ("%s\n", _("No remotes available"));
+		return TRUE;
```
superm1: Shouldn't this return nothing to do instead?

hughsie: Probably yes. Will fix.

src/fu-tool.c
```diff
@@ -1673,15 +1673,17 @@ fu_util_get_remotes (FuUtilPrivate *priv, gchar **values, GError **error)
...
 	if (remotes == NULL)
 		return FALSE;
 	if (remotes->len == 0) {
-		/* TRANSLATORS: no repositories to download from */
-		g_print ("%s\n", _("No remotes available"));
-		return TRUE;
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOTHING_TO_DO,
+				     "no remotes available");
```
superm1: it could still be a translated error though, right?

superm1:
I'd guess this is to make `fwupdtool.exe` more useful?
Other than that return code, nothing else stands out to me.

I think it's worth double checking that `fwupdtool.exe` is doing relative paths right though for `var/cache` and not dependent upon which directory you're in when you execute it.

superm1:
> /root/project/build-win32/../src/fu-engine.c:2976: undefined reference to `fu_engine_get_existing_keyring_result'

Oh, I guess that means this also need to compile Windows with GPG or PKCS as a result of this.

src/fu-tool.c
hughsie: It's not usual to translate the error domain. I think the best way to do this is provide a translation of the error code.

hughsie:
> Oh, I guess that means this also need to compile Windows with GPG or PKCS as a result of this.

That's because I forgot to remove some ifdefs in the `fu_engine_update_metadata_bytes` split -- fixed now. I think PKCS is being compiled in -- I really want to try and avoid GPG on windows if I can avoid that particular can of worms.

hughsie:
> I really want to try and avoid GPG on windows if I can avoid that particular can of worms

I'm going to enable PKCS signing on fwupd (as well as GPG) and they I think this should work.

superm1:
> I'm going to enable PKCS signing on fwupd (as well as GPG) and they I think this should work.

In general then move clients over to the PKCS as preferred then?

hughsie:
> In general then move clients over to the PKCS as preferred then?

I guess that's a question for the mailing list. I certainly have a lot more warm and fuzzy feelings about the quality of the PKCS-7 code than the GPG code. I'm tempted to just generate both detached signatures for firmware and metadata, and just change `Keyring=` in lvfs.conf locally for a few weeks to see if there are any problems.

superm1:
> I guess that's a question for the mailing list. I certainly have a lot more warm and fuzzy feelings about the quality of the PKCS-7 code than the GPG code. I'm tempted to just generate both detached signatures for firmware and metadata, and just change Keyring= in lvfs.conf locally for a few weeks to see if there are any problems.

Unfortunately that means redoing "all" the CAB files at LVFS though doesn't it?

hughsie:
> Unfortunately that means redoing "all" the CAB files at LVFS though doesn't it?

That's a good question. `fu_keyring_get_release_flags()` looks in the metainfo.xml file inside the archive, and adds "asc", "p7b", "p7c" as signatures in that order. So existing firmware with GPG signatures will continue to work, as it's choosing GPG over PKCS-7. Only people compiling without GPG will actually use the PKCS-7 signatures. Given that fwupd.org has always been signing with GPG that's probably a good default.

As to if it makes sense to resign the existing firmware, that's a very good question. We might have a scenario in the future where the signing key changes (e.g. private key leaked by hackers, or the more plausible, signing key expired) so it probably makes sense to at least have the machinery in place to resign everything with the new configured keys.

Where this gets complicated is that firmware would then have multiple `checksum_signed_sha1` values, and in theory reports could be uploaded with the old or new checksum which we'd need to match against the `Firmware` object. I guess it would be unusual to be using a locally downloaded .cab file rather than just letting fwupdmgr/gnome-software get the new file, but not completely mad. I'm not sure how much we care about this given that this isn't something that would happen very often.

superm1:
>That's a good question. fu_keyring_get_release_flags() looks in the metainfo.xml file inside the archive, and adds "asc", "p7b", "p7c" as signatures in that order. So existing firmware with GPG signatures will continue to work, as it's choosing GPG over PKCS-7. Only people compiling without GPG will actually use the PKCS-7 signatures. Given that fwupd.org has always been signing with GPG that's probably a good default.

So maybe the right action is something new for fwupd to set priority of which to use first in daemon.conf like you were suggesting.

> Where this gets complicated is that firmware would then have multiple checksum_signed_sha1 values, and in theory reports could be uploaded with the old or new checksum which we'd need to match against the Firmware object. I guess it would be unusual to be using a locally downloaded .cab file rather than just letting fwupdmgr/gnome-software get the new file, but not completely mad. I'm not sure how much we care about this given that this isn't something that would happen very often.


I sort of think that a short term action should just be to sign new stuff with both, but not resign everything unless we come across a situation that we have  to.  By the time signing key expired happens, people will probably not be accessing such old firmware anymore anyway.

Old clients will keep using gpg, new clients can use pkcs7.  On a new client if a firmware isn't signed with pkcs7 it will fall back to gpg anyway.

hughsie:
> So maybe the right action is something new for fwupd to set priority of which to use first in daemon.conf like you were suggesting.

Yes, something like that makes sense I guess.

> I sort of think that a short term action should just be to sign new stuff with both

I've enabled metadata and firmware signing of both on the LVFS instance now.

> but not resign everything unless we come across a situation that we have to

Agree. I don't know if the "it doesn't work in Windows" is a valid situation.

superm1:
> Agree. I don't know if the "it doesn't work in Windows" is a valid situation.

I wouldn't worry about it.  For the 5 people that will be using this initially they can get by without signature verification of the CAB file.

PR #1710, opened 2020-01-17
Correctly import PKCS-7 remote metadata

hughsie:
The PEM data is larger than 2kB, which was a crazy-small max size for a
signature. Make this a more respectable 1Mb so that we can load PKCS-7
signatures with full certificate info included.

This wasn't being tested in the self tests as it only affects data read from
the remote file descriptor.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
> Probably should backport this to 1_2_X too.

Done, good idea.

PR #1714, opened 2020-01-20
Wip/hughsie/fresco

hughsie:
This is still WIP and I'm waiting back from the vendor about the FIXMEs. I thought I'd get some early review anyway.


plugins/fresco-pd/fresco-pd.quirk
```diff
@@ -0,0 +1,7 @@
+# FL7102
+[Guid=USB\VID_1D5C&PID_7102]
+Plugin = fresco_pd
+
+# FL7112
+[Guid=USB\VID_1D5C&PID_7112]
+Plugin = fresco_pd
```
superm1: Any particular reason not to just match everything that is `1D5C` vendor?  Do they make other devices too?

plugins/fresco-pd/fu-fresco-pd-firmware.c
```diff
@@ -0,0 +1,81 @@
...
+
+struct _FuFrescoPdFirmware {
+	FuFirmwareClass		 parent_instance;
+	guint8			 customer_id;
+};
+
+G_DEFINE_TYPE (FuFrescoPdFirmware, fu_fresco_pd_firmware, FU_TYPE_FIRMWARE)
+
+guint8
+fu_fresco_pd_firmware_get_project_id (FuFrescoPdFirmware *self)
```
superm1: Shouldn't this be `customer_id`?

plugins/fresco-pd/fu-fresco-pd-firmware.c
```diff
@@ -0,0 +1,81 @@
...
+	g_autofree gchar *version = NULL;
+	g_autoptr(FuFirmwareImage) img = fu_firmware_image_new (fw);
+
+	/* read version block */
+	if (!fu_memcpy_safe (ver, sizeof(ver), 0x0,	/* dst */
+			     buf, bufsz, 0x1000,	/* src */
+			     sizeof(ver), error))
+		return FALSE;
+
+	/* project ID is always the 2nd byte */
```
superm1: again using project and custom interchangably

plugins/fresco-pd/fu-fresco-pd-common.c
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+#include "config.h"
+
+#include "fu-fresco-pd-common.h"
+
+gchar *
+fu_fresco_pd_version_from_buf (const guint8 ver[4])
+{
+	if (ver[3] == 1 || ver[3] == 2)
```
superm1: so the last byte is indicating order?  It seems like this version buffer is really an information blob that contains customer ID; endian order and a firmware version.  Perhaps this whole string shouldn't be exported to the `FuDevice`.  What normally shows up in release notes?  Just the 2 bytes for major and minor version?

plugins/fresco-pd/fresco-pd.quirk
hughsie: I believe other kinds of device from Fresco have a different protocol. Apparently a FrescoPd plugin was appropriate.

plugins/fresco-pd/fu-fresco-pd-firmware.c
hughsie: Good catch, thanks.

plugins/fresco-pd/fu-fresco-pd-firmware.c
hughsie: Fixed.

plugins/fresco-pd/fresco-pd.quirk
superm1: OK

plugins/fresco-pd/fu-fresco-pd-common.c
hughsie: Nah, they use it as a complete version -- e.g. `Panther_FL7102_TypeC2HDMI_DELL_V7.122.4.0_2019_1112_1652.bin` -- my guess would be it's a typo in the silicon VHDL somewhere.

hughsie: >  Do you know what device

I have literally no idea, sorry.

superm1: OK if you hear something from them about it let me know.  I'm not aware any PD controller that Dell uses from this company; especially a PD controller with a USB interface.

PR #1735, opened 2020-01-31
Wip/hughsie/tpm event log replay

hughsie:


superm1:
It's probably worth taking this the step further of comparing that the calculated hash _actually_ matches and making some noise if it doesn't.

hughsie:
> comparing that the calculated hash actually matches

I guess we should do that in `tpm-eventlog` rather than `tpm` as the latter might be neutered in favour of the dell on, right? The `uefi` plugin isn't logically enough connected to the TPM, although maybe we can use the `registered()` event? Ideas welcome.

superm1:
Looks like one of the CI problems is probably this fix:
https://gitlab.gnome.org/GNOME/glib/merge_requests/1307

We can probably turn off -Wtypedef-redefinition for the failing builds for it for now.

superm1:
> I guess we should do that in tpm-eventlog rather than tpm as the latter might be neutered in favour of the dell on, right? The uefi plugin isn't logically enough connected to the TPM, although maybe we can use the registered() event? Ideas welcome.

I think it's most interesting in the uefi plugin.  Let it get calculated in tpm event log plugin and let uefi pick up a signal and make noise when there's a problem then and they don't add up.  That means that if you can't calculate event log (such as on Dell system that has secure boot turned off) you just don't send the signal to uefi plugin.

hughsie:
> I think it's most interesting in the uefi plugin.

I think I implemented it the other way around, but I don't think it matters much.

hughsie:
> With how scary this situation actually is

I don't know how scary it is. I guess the TPM could be doing the wrong thing in a non-scary way, but @mjg59 is the expert here.

> I think it's worth also adding this into the daemon log too perhaps.

I don't think anyone looks at the daemon log in reality. If it's really that serious we ought to add a new `FWUPD_DEVICE_FLAG_IS_BROKEN` and provide a translation in the various UIs.

mjg59:
There are machines that only produce sha256 hashes, so I think you need to handle that case. I also haven't checked the backend code, so I don't know whether you handle TCG2 format event logs (which would be a prerequisite for the first point).

There are definitely firmwares out there that generate incomplete event logs and I think it's worth logging that, but right now I wouldn't recommend expressing that to the user until we have a reasonable idea of how common it is.

hughsie:
> There are machines that only produce sha256 hashes, so I think you need to handle that case.

Right, will fix.

>  so I don't know whether you handle TCG2 format event logs

We do!

> There are definitely firmwares out there that generate incomplete event logs

Do you know why? e.g. out of space etc?

mjg59:
> Do you know why? e.g. out of space etc?

The most common example I've seen is measuring the TXT ACM but failing to include it in the event log.

hughsie:
@superm1 do you want to re-review given i've rewritten a lot of the string handling?

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -11,15 +11,30 @@
...
 
+struct FuPluginData {
+	GPtrArray		*pcr0s;
+};
+
 void
 fu_plugin_init (FuPlugin *plugin)
 {
+	fu_plugin_alloc_data (plugin, sizeof (FuPluginData));
+	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_RUN_BEFORE, "uefi");
```
superm1: Is this something that we should concern turning into a circular dependency?
UEFI has this:
```
        fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "tpm_eventlog");
```

hughsie: > Is this something that we should concern turning into a circular dependency

Good question. The `METADATA_SOURCE` rule doesn't affect plugin startup order and so this should be fine.

superm1:
Just as an FYI, I did confirm on an XPS 7390 with secure boot enabled that this is working properly:
``
17:39:30:0409 FuPluginTpmEventlog  TPM reconstructed event log matched PCR0 reading
```

PR #1741, opened 2020-02-04
Detect kernel lockdown

superm1:
This should avoid at least one cause of this showing up in dmesg every time fwupd starts:

```
Lockdown: fwupd: /dev/mem,kmem,port is restricted; see man kernel_lockdown.7
```

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1044,6 +1044,12 @@ fu_common_get_path (FuPathKind path_kind)
 		if (tmp != NULL)
 			return g_strdup (tmp);
 		return g_strdup ("/sys/bus/platform/drivers");
+	/* /sys/kernel/security */
+	case FU_PATH_KIND_SYSFSDIR_SECURITY:
+		tmp = g_getenv ("FWUPD_SYSFSSECURITYDIR");
+		if (tmp != NULL)
+			return g_strdup (tmp);
+		return g_strdup ("/sys/kernel/security");
 	/* /etc */
 	case FU_PATH_KIND_SYSCONFDIR:
 		tmp = g_getenv ("FWUPD_SYSCONFDIR");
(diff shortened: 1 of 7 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -1923,3 +1929,37 @@ fu_byte_array_append_uint32 (GByteArray *array, guint32 data, FuEndianType endia
...
+	options = g_strsplit (data, " ", -1);
+	for (guint i = 0; options[i] != NULL; i++) {
+		if (g_strcmp0 (options[i], "[none]") == 0)
+			return FALSE;
+	}
+	return TRUE;
+#else
+	return FALSE;
+#endif
+}
```
hughsie: I think you want a new line here.

libfwupdplugin/fu-common.c
```diff
@@ -1923,3 +1929,37 @@ fu_byte_array_append_uint32 (GByteArray *array, guint32 data, FuEndianType endia
...
+
+	if (!g_file_test (fname, G_FILE_TEST_EXISTS))
+		return FALSE;
+	if (!g_file_get_contents (fname, &data, &len, NULL))
+		return FALSE;
+	if (len < 1)
+		return FALSE;
+	options = g_strsplit (data, " ", -1);
+	for (guint i = 0; options[i] != NULL; i++) {
+		if (g_strcmp0 (options[i], "[none]") == 0)
```
hughsie: interesting, are the square brackets meaning inverted?

libfwupdplugin/fu-common.c
```diff
@@ -1923,3 +1929,37 @@ fu_byte_array_append_uint32 (GByteArray *array, guint32 data, FuEndianType endia
...
+
+/**
+ * fu_common_kernel_locked_down:
+ *
+ * Determines if kernel lockdown in effect
+ *
+ * Since: 1.3.8
+ **/
+gboolean
+fu_common_kernel_locked_down (void)
```
hughsie: could we add some tests for this please?

libfwupdplugin/fu-common.c
superm1: Here is what it looks like:
```
$ sudo cat /sys/kernel/security/lockdown 
none integrity [confidentiality]
```

I was thinking that detecting the `none` case makes most sense in case more options are added later.

hughsie: I'll admit I'm a bit confused what value means what.

libfwupdplugin/fu-common.c
superm1: Added

libfwupdplugin/fu-common.c
superm1: Fixed

libfwupdplugin/fu-common.c
superm1: I've seen this nomenclature in the kernel sysfs files previously, and actually used it myself in a driver I wrote.

The thing in the brackets is what's picked.  The other things are possible options.  So rather than two sysfs files that are:
`current_value` and 
`possible_values`

You just read one and can tell all the possible values and which one is active.

hughsie: TIL, thanks.

PR #1747, opened 2020-02-06
Wip/hughsie/vli pd

hughsie:


plugins/vli/fu-vli-pd-device.c
```diff
@@ -453,8 +454,21 @@ fu_vli_pd_device_detach (FuDevice *device, GError **error)
 
 	/* reset from SPI_Code into ROM_Code */
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
-	if (!fu_vli_device_reset (FU_VLI_DEVICE (device), error))
-		return FALSE;
+	if (!fu_vli_device_reset (FU_VLI_DEVICE (device), &error_local)) {
+		if (g_error_matches (error_local,
+				     G_USB_DEVICE_ERROR,
+				     G_USB_DEVICE_ERROR_NO_DEVICE) ||
```
superm1: in this situation do you really want to wait for replug still?  Does that mean it was unplugged?

hughsie: Ohh, I suppose we just want FAILED. I guess it depends on how far the USB transaction the device got before it powered-down the USB bus.

superm1: Oh, hmm and it might vary from time to time.  Can you do some testing to see if it's relatively consistent?

hughsie: I've only seen FAILED. I'll try with a few hubs in between and see what happens.

PR #1757, opened 2020-02-10
Revert 7b15a4d646845496da18ddff6a48e83f6d50173a

superm1:
This code is obviously causing regressions and the algorithm in use for deleting the various boot variables needs some adjusting.  For now, revert that commit to unbreak systems and perhaps algorithm can be adjusted later.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/efi/fwup-debug.c
+++ b/plugins/uefi/efi/fwup-debug.c
@@ -58,12 +58,12 @@ fwup_log(FwupLogLevel level, const char *func, const char *file, const int line,
 
 	if (debugging) {
 		_cleanup_free CHAR16 *out1 = NULL;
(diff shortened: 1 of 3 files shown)


superm1:
> Ack, if we're breaking things. Is it the delete part of the commit that's the problem, or the checking for BootXXXX?

Yeah when I reproed it on my system it was the checking for BootXXXX that was taking forever (memory leak?).  It never actually got to the right BootXXXX that corresponded to `Linux Firmware Updater` on my system.  It rebooted before that.

I think a better algorithm might be to look at what entry `BootCurrent` is set to.  Then make sure that corresponds to a `Linux Firmware Updater` entry and if so remove it (and adjust `BootOrder`)

PR #1758, opened 2020-02-10
Wip/hughsie/vli hub same vid pid

hughsie:


libfwupdplugin/fu-usb-device.c
```diff
@@ -103,6 +103,48 @@ fu_usb_device_is_open (FuUsbDevice *device)
...
+fu_usb_device_query_hub (FuUsbDevice *self, GError **error)
+{
+	FuUsbDevicePrivate *priv = GET_PRIVATE (self);
+	gsize sz = 0;
+	guint16 value = 0x29;
+	guint8 data[0x0c] = {0x0};
+	guint8 removable = 0xff;
+
+	/* longer descriptor for SuperSpeed */
+	if (g_usb_device_get_spec (priv->usb_device) >= 0x0300)
```
superm1: CI seems to be borking on this.  Is it only fairly new gusb?

hughsie: I fixed it, thanks CI :)

PR #1767, opened 2020-02-13
Do not use fu_common_vercmp() when the device verfmt is PLAIN

hughsie:
This fixes the upgrade from 5101AALB to 51H0AALB, which fwupd was comparing as '5101' > '51'.

### DIFF
--- a/src/fu-install-task.c
+++ b/src/fu-install-task.c
@@ -356,9 +356,14 @@ fu_install_task_check_requirements (FuInstallTask *self,
 	}
 
 	/* check semver */
-	version_release = fu_common_version_parse_from_format (version_release_raw,
-							       fu_device_get_version_format (self->device));
-	vercmp = fu_common_vercmp (version, version_release);
+	if (fu_device_get_version_format (self->device) == FWUPD_VERSION_FORMAT_PLAIN) {
+		version_release = g_strdup (version_release_raw);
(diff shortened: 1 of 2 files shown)


src/fu-install-task.c
```diff
@@ -356,9 +356,14 @@ fu_install_task_check_requirements (FuInstallTask *self,
...
-							       fu_device_get_version_format (self->device));
-	vercmp = fu_common_vercmp (version, version_release);
+	if (fu_device_get_version_format (self->device) == FWUPD_VERSION_FORMAT_PLAIN) {
+		version_release = g_strdup (version_release_raw);
+		vercmp = g_strcmp0 (version, version_release);
+	} else {
+		version_release = fu_common_version_parse_from_format (version_release_raw,
+								       fu_device_get_version_format (self->device));
+		vercmp = fu_common_vercmp (version, version_release);
+	}
```
superm1: Why do this in install task rather than in fu-common? I would think you could just flag fu_common_vercmp to handle plain versions as a special case. Then anywhere else that ends up comparing versions (say a plugin) could use it too.

hughsie: fu_common_vercmp() just has two strings, it doesn't know anything about the verfmt or the device.

superm1: Oh right okay... Then yeah this approach LGTM.

PR #1768, opened 2020-02-13
Include a vendor ID for ATA hardware

hughsie:
Some vendors want to ship updates for ATA hardware, but there are currently no
lock-down restrictions in place for this kind of hardware.

There is no official provided attribute for the vendor ID in the ATA spec,
so we have to be a little creative. We can match 90% of hardware using the
vendor name prefix, and the last 10% can be detected with a heuristic
that was the result of comparing over 900 drive models.

I'm not including very old drive models, media converters, raid controllers,
or external 'portable' drives as I don't think it is useful. Also, if the drive
contains a Dell vendor block just hardcode this as Dell rather than trying to
be clever.


plugins/ata/fu-ata-device.c
```diff
@@ -166,6 +166,136 @@ fu_ata_device_parse_id_maybe_dell (FuAtaDevice *self, const guint16 *buf)
 	guid_efi = fu_ata_device_get_guid_safe (buf, 129);
 	if (guid_efi != NULL)
 		fu_device_add_guid (FU_DEVICE (self), guid_efi);
+
+	/* owned by Dell */
+	fu_device_set_vendor (FU_DEVICE (self), "Dell");
+	fu_device_set_vendor_id (FU_DEVICE (self), "ATA:0x1028");
+}
```
superm1: Does this need to call fu_device_set_name still since the call moved around?

hughsie: A good question, but I don't think so. The the name should be set by FuUdevDevice already.

plugins/ata/fu-ata-device.c
```diff
@@ -166,6 +166,136 @@ fu_ata_device_parse_id_maybe_dell (FuAtaDevice *self, const guint16 *buf)
...
+{
+	struct {
+		const gchar	*prefix;			/* in CAPS */
+		guint16		 vid;
+		const gchar	*name;
+	} map_name[] = {
+		/* vendor matches */
+		{ "FUJITSU*",	0x10cf,	"Fujitsu" },
+		{ "SEAGATE*",	0x1bb1,	"Seagate" },
+		{ "ADATA*",	0x1cc1,	"ADATA" },
```
superm1: This appears mostly sorted except for these

hughsie: Ohh yes, good catch. Thanks.

plugins/ata/fu-ata-device.c
superm1: OK.

dylanfairchild:
Have you considered using the IEEE OUI from within the WWN for devices that support it in ATA Identify Device? It will map to a device manufacturer, not necessarily a brand, but it would give you a fallback if your model string matching fails. Most modern drives I have worked with support the WWN field. I am not fully understanding if you are looking for device manufacturer or brand / OEM here as "vendor". Since you are using model string, I think you may be looking for brand, but I thought I would throw this out there in case it is helpful.

See ACS-4 section 9.11.5.8 for a description of world wide name.
See here for a list of OUIs http://standards-oui.ieee.org/oui.txt.

hughsie:
> Have you considered using the IEEE OUI from within the WWN for devices that support it

I did look at that, but neither of my ATA drives set anything.

> Most modern drives I have worked with support the WWN field

Interesting; what kind of drives are you testing? I'm using `LITEONIT LCS-256M6S` and a rebranded Phison thing for testing. If you have a IDENTIFY blob with the WWN set I'd be very interested.

dylanfairchild:
Here are 2 drives which I had handy, these are both SATA SSDs.

Model:    WDC PC SA530 SDASB8Y512G
IEEE OUI: 1b44h
Model:    Samsung SSD 860 EVO 500GB
IEEE OUI: 2538h

[binary_blobs.zip](https://github.com/fwupd/fwupd/files/4206003/binary_blobs.zip)

hughsie:
@dylanfairchild do you mind if I include those files in the public fwupd self tests?

dylanfairchild:
No, please go ahead and use the files.

plugins/ata/fu-ata-device.c
```diff
@@ -246,32 +381,62 @@ fu_ata_device_parse_id (FuAtaDevice *self, const guint8 *buf, gsize sz, GError *
...
+		has_oui_quirk = fu_device_get_vendor (FU_DEVICE (self)) != NULL;
+	}
+
+	/* if not already set using the vendor block or a OUI quirk */
+	name = fu_ata_device_get_string (id, 27, 46);
+	if (name != NULL && !has_oui_quirk)
+		fu_ata_device_parse_vendor_name (self, name);
+
+	/* ask user to report data */
+	if (self->oui != 0x0 && !has_oui_quirk &&
```
superm1: I would think you want to have this show up in `UpdateError` in some way.  And then you can point people to the wiki page, and the wiki page will point them to daemon journal logs with the details.

plugins/ata/ata.quirk
```diff
@@ -1,3 +1,7 @@
 # match all devices with this udev subsystem
 [DeviceInstanceId=BLOCK]
 Plugin = ata
+
+[DeviceInstanceId=OUI\002538]
+Vendor = Samsung
+VendorId = ATA:0x144D
```
superm1: Here's another one that shows up in daemon logs from one of my workstations.

```
[DeviceInstanceId=OUI\000c50]
Vendor = Seagate
VendorId = ATA:0x1BB1
```

plugins/ata/fu-ata-device.c
superm1: Also, if you're going all in on these, you probably want to turn off the updatable flag in these instances too.

plugins/ata/ata.quirk
hughsie: Added, thanks.

plugins/ata/fu-ata-device.c
hughsie: > Also, if you're going all in on these, you probably want to turn off the updatable flag in these instances too.

If we do this the device "drops out" from `fwupdmgr get-devices` as it then needs `--show-all-devices` -- is this what we want? It means less people see the update message...

hughsie: > turn off the updatable flag

My StarLabs drive doesn't have a WWN, but is updatable as it have firmware on the LVFS. Or would we only remove UPDATABLE if the vendor_id is NULL, i.e. the hueristic failed also? I guess it depends if we want UPDATABLE to mean "the hardware is updatable" or "the policy allows this device to be updated"

superm1: > If we do this the device "drops out" from fwupdmgr get-devices as it then needs --show-all-devices -- is this what we want? It means less people see the update message...

Not if you set `UpdateError`.  It's listed then still.

> My StarLabs drive doesn't have a WWN, but is updatable as it have firmware on the LVFS. Or would we only remove UPDATABLE if the vendor_id is NULL, i.e. the hueristic failed also? I guess it depends if we want UPDATABLE to mean "the hardware is updatable" or "the policy allows this device to be updated"

I think removing when vendor ID is NULL makes most sense.

hughsie: > Not if you set UpdateError.

Do you mind if I do that as a phase2? I'd like only technical users to see this "report" quirk for the first few weeks; when we have the ~12 most popular SSD vendors covered I can open it up a bit.

hughsie: > I think removing when vendor ID is NULL makes most sense.

Done.

superm1: > Do you mind if I do that as a phase2? I'd like only technical users to see this "report" quirk for the first few weeks; when we have the ~12 most popular SSD vendors covered I can open it up a bit.

Realistically I don't think anyone is going to see this if it's only in the journal output.  Setting `UpdateError` will at least tell you to go look at the journal.

The other thing you can do is automatically generate the quirk and drop it in `/var/cache`, and the wiki page tell someone to go pick it up.

hughsie: > automatically generate the quirk

In many cases we don't know the vendor ID or name. The lack of vendor-id is now put in the update message, but no link to the wiki. I'm really hesitant to put one there after seeming people email the flashrom mailing list with years-old versions of flashrom data.

superm1: OK.. I think the `UpdateError` was the most important part.  One more thing, rather than an environment variable can you move the toggle for this into a `ata.conf` plugin configuration file?

hughsie: I think that's a good idea, but would need something like https://github.com/fwupd/fwupd/pull/1790 first

hughsie: Okay, fixed I think.

plugins/ata/fu-plugin-ata.c
```diff
@@ -18,3 +18,11 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_udev_subsystem (plugin, "block");
 	fu_plugin_set_device_gtype (plugin, FU_TYPE_ATA_DEVICE);
 }
+
+gboolean
+fu_plugin_device_created (FuPlugin *plugin, FuDevice *dev, GError **error)
+{
+	g_autofree gchar *tmp = fu_plugin_get_config_value (plugin, "UnknownOuiReport");
+	fu_ata_device_set_unknown_oui_report (FU_ATA_DEVICE (dev), g_strcmp0 (tmp, "true") == 0);
```
superm1: I think you want to use `g_ascii_strcasecmp` no?

superm1: Also I think you need to check that you don't get back `NULL` from `fu_plugin_get_config_value`, which might happen if `ata.conf` is destroyed or mangled.

plugins/ata/meson.build
```diff
@@ -6,6 +6,10 @@ install_data([
   install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
 )
 
+install_data(['ata.conf'],
+  install_dir:  join_paths(sysconfdir, 'fwupd')
+)
```
superm1: I'd guess Fedora CI is going to fail since it explicitly documents individual conf files, and this needs to be added, right?

src/fu-engine.c
```diff
@@ -4335,6 +4335,13 @@ fu_engine_add_device (FuEngine *self, FuDevice *device)
 		fu_device_set_update_error (device, "VersionFormat is ambiguous for this device");
 	}
 
+	/* no vendor-id, and so no way to lock it down! */
+	if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_UPDATABLE) &&
+	    fu_device_get_vendor_id (device) == NULL) {
+		fu_device_remove_flag (device, FWUPD_DEVICE_FLAG_UPDATABLE);
+		fu_device_set_update_error (device, "Not updatable as no vendor ID set");
```
superm1: I think the `Not updatable` prefix is implied by this coming from an `UpdateError` key.  Just saying `No vendor ID set` should probably be enough to convey the message

hughsie: Ack, fixed.

plugins/ata/meson.build
hughsie: Ack, fixed.

plugins/ata/fu-plugin-ata.c
hughsie: > g_ascii_strcasecmp

This isn't NULL safe.

> you need to check that you don't get back NULL

This is NULL safe: `g_strcmp0`

Maybe we want `fu_plugin_get_config_value_boolean()`?

superm1: > fu_plugin_get_config_value_boolean

Yeah, can probably simplify some other uses elsewhere in the codebase too then.

hughsie: Something like https://github.com/fwupd/fwupd/pull/1791 ?

hughsie: Okay, fixed now.

PR #1778, opened 2020-02-14
Fix a critical warning when installing some firmware

hughsie:
If the device replugs, then we were doing `g_set_object(&device, device_tmp)`
which worked, but had the unfortunate effect of not unwinding properly as
`device` was a function parameter rather than a local variable.

The observed effect using `watch ((GObject*)$1)->ref_count` was the refcount was
double-decremented and thus we got a critical warning on the final valid unref.
Fix this by copying the function parameter to a local variable which we can
then change if the device replugs.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1767,7 +1767,7 @@ fu_engine_schedule_update (FuEngine *self,
 
 static gboolean
 fu_engine_install_release (FuEngine *self,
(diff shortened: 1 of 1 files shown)


superm1:
That must have been hard to track down! Lgtm

hughsie:
> That must have been hard to track down

**2 HOURS!!!**

PR #1779, opened 2020-02-14
Generate an additional UEFI quirk file using the stable LVFS metadata

hughsie:
The UEFI ESRT table just gives us a table of GUIDs with some basic flags, and
isn't very useful to end users. This is acceptable for Dell as there is only
typically one ESRT entry, which is for the system firmware. On typical Lenovo
hardware there might be half-a-dozen different 'Device' entries which all look
very similar.

As it's not possible to get a channel-of-data from the ODMs to fwupd, use the
existing LVFS metadata to generate some better names for these devices.
Although this looks like a lot of repeated data, libxmlb helpfully dedupes the
strings for us, making the quirk store only slightly larger.

Also, I've deliberately made this a manual step as we're not going to have
internet access on distro builders, and I'd also like the fwupd tarball output
to be deterministic and repeatable.

### DIFF
--- a/RELEASE
+++ b/RELEASE
@@ -10,6 +10,11 @@ git log --format="%s" --cherry-pick --right-only $(git describe --tags --abbrev=
 Add any user visible changes into ../data/org.freedesktop.fwupd.metainfo.xml
 appstream-util appdata-to-news ../data/org.freedesktop.fwupd.metainfo.xml > NEWS
 
+1.5. Update ESRT names:
+
+fwupdmgr refresh
+ninja-build uefi-update-quirk
+
 2. Update translations:
 
 ninja-build fwupd-pot
--- a/plugins/uefi/generate-quirk.py
+++ b/plugins/uefi/generate-quirk.py
@@ -1,0 +1,70 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 20202 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+import sys
+import gzip
+import xml.etree.ElementTree as etree
+
+def get_kind_from_component(md):
+
+    # ideally we would use <category> but that's not available in the XML yet
+    name = md.find('name').text
+    if name.find('Embedded Controller') != -1:
+        return 'Embedded Controller'
+    if name.find('Corporate ME') != -1:
+        return 'Intel Management Engine'
+    if name.find('Consumer ME') != -1:
+        return 'Intel Management Engine'
+    return None
+
+def generate_quirk_from_xml(infn, outfn):
+
+    # parse GZipped XML
+    with gzip.open(infn, 'rb') as xml:
+        tree = etree.parse(xml)
+
+    # look for any interesting devices
+    seen_guids = {}
+    with open(outfn, 'w') as out:
+        out.write('# generated using {} -- DO NOT EDIT\n'.format(os.path.basename(sys.argv[0])))
+        for md in tree.findall('component'):
+
+            # filter by protocol
+            protocol = None
+            for val in md.findall('custom/value'):
+                if val.attrib['key'] == 'LVFS::UpdateProtocol':
+                    protocol = val.text
+            if protocol != 'org.uefi.capsule':
+                continue
+
+            # it is unlikely, but there might be multiple GUIDs here
+            guids = []
+            for prov in md.findall('provides/firmware'):
+                if prov.text in seen_guids:
+                    continue
+                guids.append(prov.text)
+            if not guids:
+                continue
+
+            # map to a suitable name for fwupd
+            kind = get_kind_from_component(md)
+            if not kind:
+                continue
+
+            # output another chunk
+            appstream_id = md.find('id').text
+            out.write('\n# from {}\n'.format(appstream_id))
+            for guid in guids:
+                out.write('[Guid={}]\n'.format(guid))
+                out.write('Name = {}\n'.format(kind))
+                seen_guids[guid] = True
+
+if __name__ == '__main__':
+    if len(sys.argv) != 3:
+        print('invalid args, expected INPUT OUTPUT')
+        sys.exit(1)
+    generate_quirk_from_xml(sys.argv[1], sys.argv[2])
--- a/plugins/uefi/meson.build
+++ b/plugins/uefi/meson.build
@@ -7,7 +7,18 @@ if efi_os_dir != ''
   cargs += '-DEFI_OS_DIR="' + efi_os_dir + '"'
 endif
 
-install_data(['uefi.quirk'],
+run_target('uefi-update-quirk',
+  command: [
+    python3.path(),
+    join_paths(meson.current_source_dir(), 'generate-quirk.py'),
+    '/var/lib/fwupd/remotes.d/lvfs/metadata.xml.gz', 'uefi-lvfs-generated.quirk'
+  ]
+)
+
+install_data([
+    'uefi.quirk',
+    'uefi-lvfs-generated.quirk',
+  ],
   install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
 )
 
--- a/plugins/uefi/uefi-lvfs-generated.quirk
+++ b/plugins/uefi/uefi-lvfs-generated.quirk
@@ -1,0 +1,273 @@
+# generated using generate-quirk.py -- DO NOT EDIT
+
+# from com.lenovo.ThinkPadN1CHT.firmware
+[Guid=ffaf3e12-a6ea-4352-851b-76c6455913c6]
+Name = Embedded Controller
+
+# from com.lenovo.ThinkPadN1CRM.firmware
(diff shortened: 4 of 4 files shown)


RELEASE
```diff
@@ -13,6 +13,7 @@ appstream-util appdata-to-news ../data/org.freedesktop.fwupd.metainfo.xml > NEWS
 2. Update translations:
 
 ninja-build fwupd-pot
+ninja-build uefi-update-quirk
```
superm1: You should probably mention this as an extra step "Refresh quirks"

superm1:
> Also, I've deliberately made this a manual step as we're not going to have
internet access on distro builders, and I'd also like the fwupd tarball output
to be deterministic and repeatable.

Completely agree, glad it's manual.

plugins/uefi/meson.build
```diff
@@ -7,7 +7,18 @@ if efi_os_dir != ''
...
+  command: [
+    python3.path(),
+    join_paths(meson.current_source_dir(), 'generate-quirk.py'),
+    '/var/lib/fwupd/remotes.d/lvfs/metadata.xml.gz', 'uefi-generated.quirk'
+  ]
+)
+
+install_data([
+    'uefi.quirk',
+    'uefi-generated.quirk',
```
superm1: Since this is coming from lvfs data, perhaps `uefi-lvfs-generated.quirk`?

plugins/uefi/generate-quirk.py
```diff
@@ -0,0 +1,70 @@
...
+            kind = get_kind_from_component(md)
+            if not kind:
+                continue
+
+            # output another chunk
+            appstream_id = md.find('id').text
+            out.write('\n# from {}\n'.format(appstream_id))
+            for guid in guids:
+                out.write('[Guid={}]\n'.format(guid))
+                out.write('Name = {}\n'.format(kind))
```
superm1: Do we perhaps want to also match `Vendor` here?  Although I know it's unlikely, it would help prevent a collision across vendors for ESRT GUIDs

hughsie: So you mean _match_, or _set_? I was hesitant to do the latter (and VendorId) and that seems a bit magic and undefined.

superm1: Match, if it's possible.  I just want to avoid `95f9a8f4-9531-483c-afee-c3f12774ff4b` corresponds to an ME on Lenovo and SSD on Dell and wrong thing shows up.

hughsie: I don't think that's possible. If GUIDs did collide, I really would go and shout at some people pretty loud.

superm1: Well the problem is no one gets to claim they own a GUID.  There is no public registry, nothing saying you need to clear your usage of a GUID or prefix with a standards body or anything.  So who do you yell at?  The second one who put firmware out?

On LVFS CHID are an optional lever to mitigate fallout from this happening.  But if it did happen then CHID doesn't apply to metadata like this.

I guess there is another way to go about this than what your PR is doing..
You can let the metadata coming _from LVFS_ **directly** influence the name of the device.  So before you refresh you see "UEFI Device firmware".  After you refresh you see "Intel Management Engine"

hughsie: > Well the problem is no one gets to claim they own a GUID

There are more GUIDs than stars in the universe. The likelihood of generating the same random GUID is just impossible. The only time I've heard it ever happening is things like databases that generate billions of GUIDs every minute. If we do clash GUIDs it's because some vendor has copied and pasted it. In that scenario the wrong device name is the last of their problems, and they'd have to offer a new system update to fix it before it would work.

> You can let the metadata coming from LVFS directly influence the name of the device

So far I've resisted doing this, as it's a bit of a slippery slope that starts with names, moves to flags, and before you know it there's a huge complex glob of code that's a huge weird state machine.

RELEASE
hughsie: Fixed

plugins/uefi/generate-quirk.py
superm1: > f we do clash GUIDs it's because some vendor has copied and pasted it.

Well copied and pasted it or **chose** to re-use it across two different platforms.  At least on WU that's "OK"; albeit really confusing.  I really wish the spec was better in this area.

> So far I've resisted doing this, as it's a bit of a slippery slope that starts with names, moves to flags, and before you know it there's a huge complex glob of code that's a huge weird state machine.

I do see how it's a slippery slope.  This PR as its stands today is not harmful.

plugins/uefi/meson.build
hughsie: Sure, that works too, fixed.

hughsie:
> or chose to re-use it across two different platforms

I supposed I might underestimate the incompetence of some ODMs :)

plugins/uefi/meson.build
```diff
@@ -7,7 +7,18 @@ if efi_os_dir != ''
   cargs += '-DEFI_OS_DIR="' + efi_os_dir + '"'
 endif
 
-install_data(['uefi.quirk'],
+run_target('uefi-update-quirk',
+  command: [
+    python3.path(),
+    join_paths(meson.current_source_dir(), 'generate-quirk.py'),
+    '/var/lib/fwupd/remotes.d/lvfs/metadata.xml.gz', 'uefi-lvfs-generated.quirk'
```
superm1: If we ever get to the point that releases can be just "Click a button on Github", then this probably needs to be able to `curl`/`wget` the URL directly.

joakim-tjernlund:
I just tried this on top of 1.3.8 and there is one missing:
```
─UEFI Device Firmware:
│     Device ID:           b4afb730aec9550f9c9aedd4b082716ef1611d18
│     Current version:     0.73.4
│     Minimum Version:     0.73.4
│     Vendor:              DMI:LENOVO
│     GUID:                6d1cf038-74ab-49b5-bb5c-6a1e99f3aea4
```

hughsie:
> there is one missing

We have no idea what that is, and it's never had firmware updates on the LVFS :(

joakim-tjernlund:
I see, looked at the Windows update ISO and there is no mention there either.

joakim-tjernlund:
In ticket https://github.com/fwupd/fwupd/issues/1803 one can see:
```
20HMS1FC02
│
├─Lenovo ThinkPad Dock:
│     Device ID:           73ef80b60058b4f18549921520bfd94eaf18710a
│     Summary:             USB 3.x Hub
│     Current version:     50.41
│     Vendor:              LENOVO (USB:0x17EF)
│     Install Duration:    10 seconds
│     GUIDs:               3a5bdad1-2726-5194-aa7d-04e1daf9946b
│                          8d708000-7d3b-5e85-ac9d-8ef32fb2d445
│                          1df23e1b-0cb9-52e0-9b9d-f8ed96aac18e
│                          6201fecc-1641-51f6-a6d2-38a06d5476bf
│                          b435be36-7b59-5463-ab34-abfe22cf7a53
│                          06cdabc6-2059-5daa-87ab-2f6fc31fd74b
│                          8da160e7-8f9a-5eba-82c0-5f18ecaac31e
│     Device Flags:        • Updatable
```
Got  a name "Lenovo ThinkPad Dock", I just got:
```
├─USB2.0 Hub:
│     Device ID:           54c916aa848583256a9b6959446d0890818cbe18
│     Summary:             USB 2.x Hub
│     Current version:     10.74
│     Vendor:              VIA Labs, Inc. (USB:0x17EF)
│     Install Duration:    15 seconds
│     GUIDs:               112b3acf-d42a-53cc-91c8-458c3b58f359 ← USB\VID_17EF&PID_3071&REV_0A74
│                          e2c96871-8be3-530e-a796-5652cee64eeb ← USB\VID_17EF&PID_3071
│                          3545c1f3-4d81-59f7-b9d9-d25d5811e4be ← USB\VID_17EF&PID_3071&HUB_10
│                          9af0af2d-99ff-551b-9134-e94809cc8016 ← USB\VID_17EF&PID_3071&DEV_VL820Q7
│     Device Flags:        • Updatable
│                          • Device stages updates
│                          • Device can recover flash failures
│   
└─USB2.0 Hub:
      Device ID:           672c087de09848d9e7ee32aa1dea2fbeb8b81e6b
      Summary:             USB 2.x Hub
      Current version:     10.73
      Vendor:              VIA Labs, Inc. (USB:0x17EF)
      Install Duration:    15 seconds
      GUIDs:               fa8ac445-2c73-5bd7-99ab-6cfd94d1e40c ← USB\VID_17EF&PID_3071&REV_0A73
                           e2c96871-8be3-530e-a796-5652cee64eeb ← USB\VID_17EF&PID_3071
                           6f3f32bd-ac98-523e-9128-ef67941fb614 ← USB\VID_17EF&PID_3071&HUB_20
                           193e6679-e01f-5afc-a4b9-bcf31e14f09f ← USB\VID_17EF&PID_3071&DEV_VL820Q8
      Device Flags:        • Updatable
                           • Device stages updates
                           • Device can recover flash failures
```
No name, missing translation?
Using 1.3.8

PR #1781, opened 2020-02-14
trivial: meson.build: add explicit dependency for udisks2

superm1:
See #1526 for more information.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
It's a runtime dep, not a build dep surely?!

superm1:
Right; It's not a build-dep, but this should at least make sure it gets installed was my thought.

hughsie:
Adding runtime deps to the build server serves little to no purpose. Most people "consuming" fwupd do so using binary packages rather than building from source themselves. I'd be happy with an extra section in README.md or something, but expressing these as build deps is the wrong way, and actually makes the distro compilation (and CI) build process longer with more storage space required.

Issue #1678, opened 2019-12-31
"fwupdmgr get-updates" user visible strings are not translatable


tsdgeos:
```shell
$ fwupdmgr --version
$ fwupdmgr --version
client version: 1.3.6
compile-time dependency versions
        gusb:   0.3.1
        efivar: 37
daemon version: 1.3.6
```

For example
```
$ fwupdmgr get-updates 
No hi ha cap actualització per a Unifying Receiver: current version is RQR12.10_B0032: RQR12.10_B0032=same, RQR12.08_B0030=older, RQR12.07_B0029=older
```

The "current version is" and "same" and "older" strings are not marked with _() when looking at the code.

Also i just realized that it seems that "Unifying Receiver" is also part of the codebase, so it'd be great if that was translatable too :)

hughsie:
I thought this would be fixed by https://github.com/fwupd/fwupd/commit/7e4949c4f87f1f9957154f0f81e5626d3bacd7fa

tsdgeos:
I don't know much (or anything) about the g_stuff so I may be wrong but 
"current version is %s"
is not marked with _() so as far as i understand that's not extracted to the .po files so not translatable?

superm1:
The last part of that string comes from the daemon, not the client.  So it won't be aware of the locale in use.  The only way we can really help that is to let the client notify the daemon of the locale over the wire and then the daemon can build translated strings for stuff like this.

superm1:
> Also i just realized that it seems that "Unifying Receiver" is also part of the codebase, so it'd be great if that was translatable too :)

This is the same problem - although it's a hardcoded string it's built by the daemon.

hughsie:
> is to let the client notify the daemon of the locale over the wire

PackageKit did this, setting the locale for the request and the daemon would return results in the correct language. It was a lot of work, and didn't work very well in most cases.

superm1:
>  didn't work very well in most cases.

Curious, what cases did it break?  It seems like a straightforward idea, but if it's a lot of work I agree probably not worth it.

Would it help if we tried to make the systemd unit source `/etc/default/locale` if it exists and then the daemon could at least try to get better strings based upon the language for the owner of a system.

hughsie:
> what cases did it break

I guess PackageKit was unusual in that it was calling all kinds of weird package backends, some of which not UTF-8 safe. Some backends were playing games with setlocale too. Perhaps that's less of a problem with the libraries we use in fwupd.

[...]

PR #1782, opened 2020-02-14
trivial: fu-engine: remove some non-translatable strings (Fixes: #1678)

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -333,8 +333,16 @@ fu_util_get_updates (FuUtilPrivate *priv, gchar **values, GError **error)
 		GNode *child;
 
 		/* not going to have results, so save a engine round-trip */
-		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED))
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_UPDATABLE))
 			continue;
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED)) {
+			/* TRANSLATORS: message letting the user know no device upgrade available due to missing on LVFS
+			* %1 is the device name */
+			g_autofree gchar *tmp = g_strdup_printf (_("• %s has no available firmware updates"),
+								 fwupd_device_get_name (dev));
+			g_printerr ("%s\n", tmp);
+			continue;
+		}
 		if (!fu_util_filter_device (priv, dev))
 			continue;
 
@@ -345,9 +353,11 @@ fu_util_get_updates (FuUtilPrivate *priv, gchar **values, GError **error)
 		if (rels == NULL) {
 			/* TRANSLATORS: message letting the user know no device upgrade available
 			* %1 is the device name */
-			upgrade_str = g_strdup_printf (_("No upgrades for %s"),
-						      fwupd_device_get_name (dev));
-			g_printerr ("%s: %s\n", upgrade_str, error_local->message);
+			g_autofree gchar *tmp = g_strdup_printf (_("• %s has the latest available firmware version"),
+								 fwupd_device_get_name (dev));
+			g_printerr ("%s\n", tmp);
+			/* discard the actual reason from user, but leave for debugging */
+			g_debug ("%s", error_local->message);
 			continue;
 		}
 		child = g_node_append_data (root, dev);
@@ -976,15 +986,29 @@ fu_util_update_all (FuUtilPrivate *priv, GError **error)
 		if (!fu_util_is_interesting_device (dev))
 			continue;
 		/* only show stuff that has metadata available */
-		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED))
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_UPDATABLE))
 			continue;
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED)) {
+			/* TRANSLATORS: message letting the user know no device upgrade available due to missing on LVFS
+			* %1 is the device name */
+			g_autofree gchar *tmp = g_strdup_printf (_("• %s has no available firmware updates"),
+								 fwupd_device_get_name (dev));
+			g_printerr ("%s\n", tmp);
+			continue;
+		}
 		if (!fu_util_filter_device (priv, dev))
 			continue;
 
 		device_id = fu_device_get_id (dev);
 		rels = fu_engine_get_upgrades (priv->engine, device_id, &error_local);
 		if (rels == NULL) {
-			g_printerr ("%s\n", error_local->message);
+			/* TRANSLATORS: message letting the user know no device upgrade available
+			* %1 is the device name */
+			g_autofree gchar *tmp = g_strdup_printf (_("• %s has the latest available firmware version"),
+								 fwupd_device_get_name (dev));
+			g_printerr ("%s\n", tmp);
+			/* discard the actual reason from user, but leave for debugging */
+			g_debug ("%s", error_local->message);
 			continue;
 		}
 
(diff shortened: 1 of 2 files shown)


superm1:
image
@tsdgeos

hughsie:
I'm rally not sure the arrows are any clearer if I'm honest. Lets step back a bit; what are we trying to say to the user? Is it useful for them to know the same version is available to reinstall on the command line? Or that a version is downgradable? I'd argue that perhaps we can get rid of all the things after the `:` with no loss of real functionality.

superm1:
You actually wrote that stuff in initially, but I'll hypothesize what it was for.  I think it was trying to convey why there are no updates available.  I agree however that information usually isn't super useful.

hughsie:
> I think it was trying to convey why there are no updates available

I think it harks back to a time before we had `GetReleases` and we had no other way of knowing why a release wasn't being suggested.

superm1:
Yeah then tearing out the whole set of code seems fine to me too.

src/fu-tool.c
```diff
@@ -976,15 +982,25 @@ fu_util_update_all (FuUtilPrivate *priv, GError **error)
...
 		if (!fu_util_filter_device (priv, dev))
 			continue;
 
 		device_id = fu_device_get_id (dev);
 		rels = fu_engine_get_upgrades (priv->engine, device_id, &error_local);
 		if (rels == NULL) {
-			g_printerr ("%s\n", error_local->message);
+			/* TRANSLATORS: message letting the user know no device upgrade available
+			* %1 is the device name */
+			g_printerr (_("• %s is updated to the latest version available\n"), fwupd_device_get_name (dev));
```
hughsie: I'm not sure about the message, but can't think of a better one.

hughsie: My issue is that it sounds like we did the update, when it might have been updated in Windows, or in the factory.

src/fu-tool.c
```diff
@@ -976,15 +982,25 @@ fu_util_update_all (FuUtilPrivate *priv, GError **error)
 		if (!fu_util_is_interesting_device (dev))
 			continue;
 		/* only show stuff that has metadata available */
-		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED))
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_UPDATABLE))
+			continue;
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED)) {
+			/* TRANSLATORS: message letting the user know no device upgrade available due to missing on LVFS
+			* %1 is the device name */
+			g_printerr (_("• %s does not have firmware updates posted on any enabled remotes\n"), fwupd_device_get_name (dev));
```
hughsie: I don't like `posted` and `remotes` makes it all sound very technical. Maybe just `%s has no available firmware updates`?

src/fu-tool.c
superm1: How about

```
• %s is running the latest version available firmware

```

src/fu-tool.c
superm1: Yeah that's good.

src/fu-tool.c
hughsie: `%s has the latest available firmware version`

superm1: Adjusted

src/fu-tool.c
superm1: Adjusted

superm1:
I personally like this approach a lot better.  The details are still very easily available for debugging.  Adding `--verbose` you can get them.  Sample output:

```
# fwupdmgr get-updates --verbose
• Thunderbolt Controller has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.501: current version is 44.00: 44.00=same
• Thunderbolt controller in Dell dock has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.502: current version is 43.00: 40.00=older
• Package level of Dell dock has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.503: current version is 01.00.08.01: 01.00.04.01=older
• RTS5413 in Dell dock has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.504: current version is 01.21: 01.21=same
• RTS5487 in Dell dock has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.505: current version is 01.47: 01.47=same
• VMM5331 in Dell dock has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.506: current version is 05.04.00: 05.03.10=older
• WD19TB has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.508: current version is 01.00.00.02: 01.00.00.00=older
• BC501 NVMe SK hynix 128GB has no available firmware updates
• System Firmware has the latest available firmware version
(fwupdmgr:125404): FuMain-DEBUG: 08:26:08.509: current version is 1.9.1: 1.9.1=same, 1.8.0=older, 1.7.0=older, 1.6.0=older, 1.5.0=older
• Touchpad has no available firmware updates
```

src/fu-util.c
```diff
@@ -1848,8 +1853,14 @@ fu_util_update_all (FuUtilPrivate *priv, GError **error)
 		g_autoptr(GError) error_local = NULL;
 
 		/* not going to have results, so save a D-Bus round-trip */
-		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED))
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_UPDATABLE))
+			continue;
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED)) {
+			/* TRANSLATORS: message letting the user know no device upgrade available due to missing on LVFS
+			* %1 is the device name */
+			g_printerr (_("• %s has no available firmware updates\n"), fwupd_device_get_name (dev));
```
hughsie: I don't think the translators are happy when you include `\n` in the `_()` string. You might need to do a `g_autofree gchar *tmp = g_strdup_printf` thing...

superm1: fixed

src/fu-tool.c
```diff
@@ -345,9 +353,11 @@ fu_util_get_updates (FuUtilPrivate *priv, gchar **values, GError **error)
 		if (rels == NULL) {
 			/* TRANSLATORS: message letting the user know no device upgrade available
 			* %1 is the device name */
-			upgrade_str = g_strdup_printf (_("No upgrades for %s"),
-						      fwupd_device_get_name (dev));
-			g_printerr ("%s: %s\n", upgrade_str, error_local->message);
+			g_autofree gchar *tmp = g_strdup_printf (_("• %s has the latest available firmware version"),
+						fwupd_device_get_name (dev));
```
hughsie: indent weird

src/fu-tool.c
```diff
@@ -333,8 +333,16 @@ fu_util_get_updates (FuUtilPrivate *priv, gchar **values, GError **error)
...
 
 		/* not going to have results, so save a engine round-trip */
-		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED))
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_UPDATABLE))
 			continue;
+		if (!fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_SUPPORTED)) {
+			/* TRANSLATORS: message letting the user know no device upgrade available due to missing on LVFS
+			* %1 is the device name */
+			g_autofree gchar *tmp = g_strdup_printf (_("• %s has no available firmware updates"),
+						fwupd_device_get_name (dev));
```
hughsie: indent weird, maybe also split up onto a new line?

PR #1784, opened 2020-02-17
Use the recently released flashrom v1.2

hughsie:


superm1:
Lgtm. I wonder does this mean we should also turn it "on" by default? Or should we wait until some distros actually package it to do that?

hughsie:
> Or should we wait until some distros actually package it to do that?

I'm just waiting for the Fedora package to make its way into rawhide so I can build it on Fedora by default.

PR #1785, opened 2020-02-17
ci: use standalone script to generate build dependencies

tomasbw:
Sometimes it is desirable to create a build environment
outside of docker.
Move dependencies parser to a standalone python script
and call it from generate_docker.py

Signed-off-by: Tomas Winkler <tomas.winkler@intel.com>

Type of pull request:
- [ ] Code fix
- [v] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/README.md
+++ b/contrib/ci/README.md
@@ -104,6 +104,8 @@ Each distribution will have ***package*** elements and ***control*** elements.
 * ___inclusive___ elements represent an inclusive list of architectures to be installed on
 * ___exclusive___ elements represent an exclusive list of architectures to not be installed on
 
+For convenience there is also a standalone script __generate_dependencies.py__ that parses ___dependencies.xml___.
+
 Dockerfile.in
 -------------
 The ***Dockerfile.in*** file will be used as a template to build the container.  No hardcoded dependencies should be put in this file.  They should be stored in ***dependencies.xml***.
--- a/contrib/ci/generate_dependencies.py
+++ b/contrib/ci/generate_dependencies.py
@@ -1,0 +1,73 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2017-2018 Dell, Inc.
+# Copyright (C) 2020 Intel, Inc.
+#
+# SPDX-License-Identifier: LGPL-2.1+
+#
+import os
+import sys
+import argparse
+import xml.etree.ElementTree as etree
+
+def parse_dependencies(OS, SUBOS, requested_type):
+    deps = []
+    dep = ''
+    directory = os.path.dirname(sys.argv[0])
+    tree = etree.parse(os.path.join(directory, "dependencies.xml"))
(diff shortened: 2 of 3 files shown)


contrib/ci/dependencies.py
```diff
@@ -0,0 +1,60 @@
...
+    try:
+        import distro
+        target = distro.linux_distribution()[0]
+    except ModuleNotFoundError:
+        target = None
+
+    target = os.getenv('OS', target)
+    if target is None:
+        print("Missing OS environment variable")
+        sys.exit(1)
```
superm1: The main reason for using OS in the original script was that it was a very convenient way for Travis to pass to the job from the environment what it needs to be.

For the purpose of a standalone script I would think supporting argument parsing would be more convenient (and you can have a --help command too then)

tomasbw: I can do that, the question if it is desirable to retain the env variable, too.

superm1: Yeah I think retain it so that it can still work with the docker build script

contrib/ci/dependencies.py
```diff
@@ -0,0 +1,60 @@
...
+
+    _os = target.lower()
+    _sub_os = ''
```
FFY00: Normal variables like this shouldn't start with a `_`.

contrib/ci/dependencies.py
```diff
@@ -0,0 +1,60 @@
...
+#
+import os
+import sys
+import xml.etree.ElementTree as etree
+
+def parse_dependencies(OS, SUBOS, requested_type):
+    deps = []
+    dep = ''
+    directory = os.path.dirname(sys.argv[0])
+    tree = etree.parse(os.path.join(directory, "dependencies.xml"))
```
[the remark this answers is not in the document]
tomasbw: Let's do all the python improvements in a new MR, so far this is just a movement of an existing code.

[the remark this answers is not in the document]
FFY00: Sure :)

contrib/ci/dependencies.py
```diff
@@ -0,0 +1,73 @@
...
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--os",
+                        default=target,
+                        choices=["fedora",
+                                 "centos",
+                                 "flatpak",
+                                 "debian",
+                                 "ubuntu",
+                                 "arch"],
```
superm1: I'm not crazy about this being hardcoded; but admittedly this changes very rarely and there isn't a place to readily read the supported distros.

superm1:
OK so just two more nits:
1. Would you mind renaming it to `generate_dependencies.py`?  I think it would match the others (`generate_debian.py` and `generate_docker.py`) more nicely.
2. Can you add some mention about this to `contrib/ci/README.md` too?

Issue #1783, opened 2020-02-15
Some tools shouldn't be in libexec


superm1:
I was reading through FHS after a MR was submitted to fwupd packaging in Debian and came across `libexec` details.

> /usr/libexec includes internal binaries that are not intended to be executed directly by users or shell scripts. Applications may use a single subdirectory under /usr/libexec.
> 
> Applications which use /usr/libexec in this way must not also use /usr/lib to store internal binaries, though they may use /usr/lib for the other purposes documented here.
> 
> 

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html

So that makes me think we should actually not be storing the following items there (as users or shell scripts may choose to access them directly).
* `fwupdagent`
* `fwupdate`
* `fwupdoffline`
* `fwupdtool`
* `fwupdtpmevlog`

The EFI binary it makes sense to me to still store there however, and the daemon.

Thoughts?

FFY00:
> as users or shell scripts may choose to access them directly

If this is supported then move then to `/usr/lib`.

> The EFI binary it makes sense to me to still store there however, and the daemon.

Yes , I agree. It should only be used internaly.

hughsie:
I'm not sure any of the tools should be used by "end users" except fwupdmgr. I really don't think debugging tools or executables designed to be run from systemd services belong in usrbin.

superm1:
Not `usr/bin`, but `usr/lib/$(arch)` or `usr/lib/$(arch)/fwupd`

hughsie:
> Not usr/bin, but usr/lib/$(arch)

Ahh, gotcha, sorry. I'm sorta struggling with the difference between usr/lib/$(arch) and libexec if I'm honest....

FFY00:
I think the technical differences between `libdir` and `libexecdir` have already been established for now. But I'll do a quick summary:

`bin` is for files that are supposed to be run by users, architecture dependent or otherwise.
`libexec` is for architecture dependent files that can be run by an external program.
`lib` is for architecture dependent files that should not be run by an external program.

If anyone still has questions feel free to ask :)

What we need to discuss is which components should be supported to be run by external programs. I am not comfortable enough giving an opinion on the matter as I am not that involved in the project and probably lack a lot of background.

So @hughsie and @superm1 you two should discuss.

hughsie:
> bin is for files that are supposed to be run by users, architecture dependent or otherwise.

[...]

PR #1786, opened 2020-02-18
Move binaries around to better adhere to FHS 3.0

superm1:
Include some updates from packaging in Debian, and then also shuffle some binaries that users can execute around to `usr/lib`.
See #1783 for more details.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

CC:
@bigon
@martinezjavier

### DIFF
--- a/contrib/ci/dependencies.xml
+++ b/contrib/ci/dependencies.xml
@@ -426,6 +426,15 @@
       <package variant="x86_64" />
     </distro>
   </dependency>
+  <dependency type="build" id="libglib2.0-doc">
+    <distro id="debian">
+      <package variant="x86_64" />
+      <package variant="i386" />
+    </distro>
+    <distro id="ubuntu">
+      <package variant="x86_64" />
+    </distro>
+  </dependency>
   <dependency type="build" id="libglib2.0-dev">
     <distro id="centos">
       <package>glib2-devel</package>
--- a/contrib/debian/control.in
+++ b/contrib/debian/control.in
@@ -6,6 +6,7 @@ Uploaders: Steve McIntyre <93sam@debian.org>,
            Matthias Klumpp <mak@debian.org>,
            Mario Limonciello <mario.limonciello@dell.com>
 Build-Depends: %%%DYNAMIC%%%
+Build-Depends-Indep: libglib2.0-doc <!nodoc>
 Standards-Version: 4.5.0
 Section: admin
 Homepage: https://github.com/fwupd/fwupd
--- a/contrib/debian/fwupd.install
+++ b/contrib/debian/fwupd.install
@@ -8,11 +8,11 @@ usr/share/icons/*
 usr/share/polkit-1/*
 usr/share/locale
 usr/share/metainfo/*
-usr/lib/*/fwupd
-usr/lib/*/fwupdagent
-usr/lib/*/fwupdoffline
-usr/lib/*/fwupdtool
-usr/lib/*/fwupdtpmevlog
+usr/libexec/fwupd/fwupd
+usr/libexec/fwupd/fwupdagent
+usr/libexec/fwupd/fwupdoffline
+usr/libexec/fwupd/fwupdtool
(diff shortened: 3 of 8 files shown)


hughsie:
I'm still confused; what's the difference between libexec and libdir? Doing this kind of change breaks lots of documentation (e.g. VIA and Lenovo use fwupdtool for testing) and there needs to be a good reason for doing it.

superm1:
> I'm still confused; what's the difference between libexec and libdir? Doing this kind of change breaks lots of documentation (e.g. VIA and Lenovo use fwupdtool for testing) and there needs to be a good reason for doing it.

If I understand it right; the difference is stuff in `libexecdir` is supposed to be seen as _internal_ to the application.  Binaries that a user or script shouldn't ever be executing.  Executable stuff in `libdir` being stuff that can execute.  That's my interpretation though; I could be completely off base here.  Others can speak up if so.

superm1:
> Applications which use /usr/libexec in this way must not also use /usr/lib to store internal binaries, though they may use /usr/lib for the other purposes documented here.

Maybe this should actually be interpreted as stuff that you want internal either pick libexecdir or libdir, but not both.  Stuff that you want people running should be in usr/bin?

hughsie:
I don't think I've seen a package that uses both.

FFY00:
Yes, `lib*` holds architecture dependent binaries, `libexec` holds binaries that could be executed, `lib` holds binaries that shouldn't be executed (at least externally, think of it like a public API).

superm1:
>Yes, lib* holds architecture dependent binaries, libexec holds binaries that could be executed, lib holds binaries that shouldn't be executed (at least externally, think of it like a public API).

So there are packages you know of using both?

> I don't think I've seen a package that uses both.

If we don't split along that libexec/libdir separation then maybe we should discuss which of those tools should actually leave libexecdir to go into usrbin (if any).  I think the most logical one is `fwupdtool` and `fwupdagent`.

hughsie:
In which case fwupdtool should stay in libexec I guess.

FFY00:
> So there are packages you know of using both?

In arch we set `libdir` and `libexecdir` both to `lib` so I am not sure about anything from the top of my mind, but I am pretty sure that git should use both, at least conceptually.

FFY00:
That sounded weird -- I am not sure if it does, but I am pretty sure it should.

superm1:
OK it's clear we don't have agreement here, so I am leaning on we should just leave it as is.  I'll tear out that last patch that moved stuff around and just merge the packaging changes that unify Debian and Ubuntu to using `libexecdir` in the first place .

FFY00:
We can go back to the issue and discuss which components should be private and which should be public.

bigon:
> > I'm still confused; what's the difference between libexec and libdir? Doing this kind of change breaks lots of documentation (e.g. VIA and Lenovo use fwupdtool for testing) and there needs to be a good reason for doing it.
> 
> If I understand it right; the difference is stuff in `libexecdir` is supposed to be seen as _internal_ to the application. Binaries that a user or script shouldn't ever be executing. Executable stuff in `libdir` being stuff that can execute. That's my interpretation though; I could be completely off base here. Others can speak up if so.

`bindir` is the one with "stuff that can execute", `libdir` is just only for libraries if `libexecdir`exists:

http://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html:
>/usr/libexec includes internal binaries that are not intended to be executed directly by users or shell scripts. Applications may use a single subdirectory under /usr/libexec.
> Applications which use /usr/libexec in this way must not also use /usr/lib to store internal binaries, though they may use /usr/lib for the other purposes documented here.

superm1:
> bindir is the one with "stuff that can execute", libdir is just only for libraries if libexecdirexists:

OK, so  since fwupd uses `libexecdir`, for stuff that we want _people_ executing, moving it to `bindir` is what makes sense.  Thanks for clarifying.  That's what the new PR does: https://github.com/fwupd/fwupd/pull/1788

PR #1788, opened 2020-02-19
Move some binaries to usr/bin

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/build_windows.sh
+++ b/contrib/ci/build_windows.sh
@@ -52,9 +52,6 @@ makensis -NOCD $build/contrib/setup-win32.nsi
 #so that it's actually executable
 cp /usr/x86_64-w64-mingw32/sys-root/mingw/bin/*.dll .
 
-#move these so they work
-mv fwupd/*.exe .
-
 #remove static archives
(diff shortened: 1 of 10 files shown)


hughsie:
If they're in bindir, they need manpages too :(

superm1:
> If they're in bindir, they need manpages too :(

Yep, I made them.

superm1:
> What about the other tools?

* `fwupd` should definitely be in `libexecdir`
* `fwupdoffline` is not intended for user use.  It's only used by systemd.
* `fwupdtpmevlog` I'm not sure (I can go either way)

hughsie:
> fwupdtpmevlog I'm not sure

If we're going with the logic of the user running debugging things, I think it should probably treated the same way as fwupdtool and fwupdate.

superm1:
> > fwupdtpmevlog I'm not sure
> 
> If we're going with the logic of the user running debugging things, I think it should probably treated the same way as fwupdtool and fwupdate.

OK I've moved that over as well now.

PR #1790, opened 2020-02-19
Add a plugin vfunc to run after subclassed FuDevice creation

hughsie:
Sometimes the plugin will want to influence the subclassed device, for instance
by reading a per-plugin config file. At the moment there's no way to do this,
as even _device_registered() is explicitly designed for devices *not* created
by the plugin itself.

Even if _device_registered() was changed to include the plugin creating the
object it would still happen well after the device has done _probe() and/or
_setup() and would probably be too late to do anything useful.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -1678,6 +1678,8 @@ fu_plugin_usb_device_added (FuPlugin *self, FuUsbDevice *device, GError **error)
 	/* create new device and incorporate existing properties */
 	dev = g_object_new (device_gtype, NULL);
 	fu_device_incorporate (dev, FU_DEVICE (device));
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-plugin.c
```diff
@@ -1950,6 +1954,35 @@ fu_plugin_runner_device_register (FuPlugin *self, FuDevice *device)
 	}
 }
 
+/**
+ * fu_plugin_runner_device_created:
+ * @self: a #FuPlugin
+ * @device: a #FuDevice
+ *
```
superm1: missing error

hughsie: Fixed, thanks.

PR #1791, opened 2020-02-19
trivial: Add helper fu_plugin_get_config_value_boolean()

hughsie:


plugins/uefi/fu-plugin-uefi.c
```diff
@@ -413,10 +413,7 @@ fu_plugin_uefi_load_config (FuPlugin *plugin, FuDevice *device, GError **error)
 	fu_device_set_metadata_integer (device, "RequireESPFreeSpace", sz_reqd);
 
 	/* shim used for SB or not? */
-	require_shim_for_sb = fu_plugin_get_config_value (plugin, "RequireShimForSecureBoot");
-	if (require_shim_for_sb == NULL ||
-	    g_ascii_strcasecmp (require_shim_for_sb, "true") == 0)
-		shim_needed = TRUE;
+	shim_needed = fu_plugin_get_config_value_boolean (plugin, "RequireShimForSecureBoot");
```
superm1: I think this (unintentionally) inverted the default for this value.  If it's not set, or invalid it will be false - but previously it was true if not set or invalid.

hughsie: I think that's fine; I think we should standardize on the key being `FALSE` if missing or invalid -- returning true for a commented out key was a bit weird.

superm1: I don't really like that if someone deletes an unmodified `uefi.conf` the behavior changes.  Perhaps we should rename the key then?

hughsie: >  Perhaps we should rename the key then

That might be best indeed; something like `DisableShimForSecureBoot`?

superm1: Yes, that's good.

hughsie: Something like this?

PR #1796, opened 2020-02-20
Always check for PLAIN when doing vercmp() operations

hughsie:
In 1de7cc we checked the version format when checking for update, but there are
many other places that are doing verfmt-insensitive comparisons. For instance,
the predicates in <requires> all fail if the device version format is plain.
his breaks updating some NVMe drives where the `ne` requirements are not
semantic versions.

To avoid trying to catch all the bugs in different places, and in case we have
a future verfmt that should be treated another way, refactor this out in to a
common function and deprecate the old function.

### DIFF
--- a/libfwupdplugin/fu-common-version.c
+++ b/libfwupdplugin/fu-common-version.c
@@ -450,6 +450,29 @@ fu_common_version_verify_format (const gchar *version,
 }
 
 /**
+ * fu_common_vercmp_full:
+ * @version_a: the semver release version, e.g. 1.2.3
+ * @version_b: the semver release version, e.g. 1.2.3.1
+ * @fmt: a #FwupdVersionFormat, e.g. %FWUPD_VERSION_FORMAT_PLAIN
+ *
+ * Compares version numbers for sorting taking into account the version format
+ * if required.
+ *
+ * Returns: -1 if a < b, +1 if a > b, 0 if they are equal, and %G_MAXINT on error
+ *
+ * Since: 1.3.9
+ */
+gint
+fu_common_vercmp_full (const gchar *version_a,
+		       const gchar *version_b,
+		       FwupdVersionFormat fmt)
+{
+	if (fmt == FWUPD_VERSION_FORMAT_PLAIN)
+		return g_strcmp0 (version_a, version_b);
+	return fu_common_vercmp (version_a, version_b);
+}
+
+/**
  * fu_common_vercmp:
  * @version_a: the semver release version, e.g. 1.2.3
  * @version_b: the semver release version, e.g. 1.2.3.1
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -50,7 +50,9 @@ fu_plugin_thunderbolt_safe_kernel (FuPlugin *plugin, GError **error)
 		return TRUE;
 	}
 
-	if (fu_common_vercmp (name_tmp.release, minimum_kernel) < 0) {
+	if (fu_common_vercmp_full (name_tmp.release,
+				   minimum_kernel,
+				   FWUPD_VERSION_FORMAT_TRIPLET) < 0) {
 		g_set_error (error,
 			     FWUPD_ERROR,
 			     FWUPD_ERROR_INTERNAL,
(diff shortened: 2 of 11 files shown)


libfwupdplugin/fu-common-version.c
```diff
@@ -449,6 +449,29 @@ fu_common_version_verify_format (const gchar *version,
...
+ * Since: 1.3.9
+ */
+gint
+fu_common_vercmp_full (const gchar *version_a,
+		       const gchar *version_b,
+		       FwupdVersionFormat fmt)
+{
+	if (fmt == FWUPD_VERSION_FORMAT_PLAIN)
+		return g_strcmp0 (version_a, version_b);
+	return fu_common_vercmp (version_a, version_b);
```
superm1: What do you think about also absorbing a NULL check into this function?  Then plugins won't have to do it.

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -50,7 +50,9 @@ fu_plugin_thunderbolt_safe_kernel (FuPlugin *plugin, GError **error)
 		return TRUE;
 	}
 
-	if (fu_common_vercmp (name_tmp.release, minimum_kernel) < 0) {
+	if (fu_common_vercmp_full (name_tmp.release,
+				   minimum_kernel,
+				   FWUPD_VERSION_FORMAT_TRIPLET) < 0) {
```
superm1: I wonder if we need to count the number of dots now in the config value to make sure this stayed triplet.  Otherwise `thunderbolt.conf` having a bad value could be problematic?

libfwupdplugin/fu-common-version.c
hughsie: I don't know how to compare NULL to a version? What would it return?

plugins/thunderbolt/fu-plugin-thunderbolt.c
hughsie: It's not important really what the verfmt actually is. It's just non-`PLAIN`.

libfwupdplugin/fu-common-version.c
```diff
@@ -449,6 +449,29 @@ fu_common_version_verify_format (const gchar *version,
...
+ * if required.
+ *
+ * Returns: -1 if a < b, +1 if a > b, 0 if they are equal, and %G_MAXINT on error
+ *
+ * Since: 1.3.9
+ */
+gint
+fu_common_vercmp_full (const gchar *version_a,
+		       const gchar *version_b,
+		       FwupdVersionFormat fmt)
```
superm1: Another thought here; since you have the version format information, would it make sense to count the dots to validate a and b are (conceivably) the same format?

hughsie: I also wondered this; but would testing a kernel version of `2.6.1.1 > 2.6.1` be then invalid?

src/fu-engine.c
```diff
@@ -1239,7 +1252,9 @@ fu_engine_check_requirement_id (FuEngine *self, XbNode *req, GError **error)
 			     xb_node_get_text (req));
 		return FALSE;
 	}
-	if (!fu_engine_require_vercmp (req, version, &error_local)) {
+	if (!fu_engine_require_vercmp (req, version,
+				       FWUPD_VERSION_FORMAT_TRIPLET, /* anything !PLAIN */
```
superm1: would `FWUPD_VERSION_FORMAT_UNKNOWN` make more sense then?

hughsie: Yes, good idea!

libfwupdplugin/fu-common-version.c
superm1: I guess that's true, it is inventing a return value that doesn't make sense, never mind.

libfwupdplugin/fu-common-version.c
superm1: Do distros add another digit there for their builds?  On Ubuntu my `uname -r` has the extra number as a dash (eg 5.3.0-18)

hughsie: Fair point, mine is a triplet too.

hughsie: Lets add this kind of checking for now; I don't want to introduce a regression. There's also no `GError` (or any way to return an error!) so I'm no sure how we'd signify an "invalid triplet" for instance.

src/fu-engine.c
hughsie: Fixed.

PR #1800, opened 2020-02-21
uefi: For UEFI device firmware don't default to triplet

superm1:
By definition if we don't know the format that the vendor is using
we should be defaulting to a plain integer instead.

This matches what Windows device manager renders for
Firmware Resource version as well.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -472,7 +472,7 @@ fu_plugin_uefi_get_version_format_for_type (FuPlugin *plugin, FuUefiDeviceKind d
 
 	/* we have no information for devices */
 	if (device_kind == FU_UEFI_DEVICE_KIND_DEVICE_FIRMWARE)
-		return FWUPD_VERSION_FORMAT_TRIPLET;
+		return FWUPD_VERSION_FORMAT_PLAIN;
 
 	content = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_MANUFACTURER);
 	if (content == NULL)


hughsie:
Not a fan; that means that all Lenovo machines are going to have device version numbers like 0x123456 surely?

superm1:
> Not a fan; that means that all Lenovo machines are going to have device version numbers like 0x123456 surely?

Not system firmware - device firmware.  The comment in the code does say that we can't know what device firmware format has.  I think if this is going to regress stuff in the field, those quirks need to embed the versionformat.

hughsie:
> Not system firmware - device firmware. 

Right, but some vendors (like Lenovo) have a UEFI ESRT device entry for the ME, the EC, etc.

> I think if this is going to regress stuff in the field

I think this is going to cause regressions.

> those quirks need to embed the versionformat.

That's not practical; fwupd in RHEL might only be updated every 18 months.

Going back a bit; why do you want the fields to match windows? We have the VersionRaw tag we show in --verbose...

    VendorId:             DMI:LENOVO
    Version:              184.70.3626
    VersionLowest:        0.0.1
    VersionFormat:        triplet
  VersionRaw:           0xb8460e2a

superm1:
> Going back a bit; why do you want the fields to match windows? We have the VersionRaw tag we show in --verbose...

Introducing support for UEFI device firmware on new systems, trying to get versionformat right.  The strings are **not triplets**

hughsie:
> Introducing support for UEFI device firmware on new systems, trying to get versionformat right. The strings are not triplets

But I've got an impossible choice here.

1) set the unknown devices to "plain" and require Lenovo to include a .quirk-->triplet for every device they are using triplet for right now, and plan to use in the next ~18 months

2) set the unknown devices to "triplet" and require Dell to include a .quirk-->plain for all the new devices they're planning to build in the next 18 months

superm1:
It's unfortunate; it seems like this should have been plain from the get go.

hughsie:
Also, feedback from real users: they *hate* version numbers like `0x123456`...

superm1:
> But I've got an impossible choice here.

The other option is to follow the state machine idea you weren't a fan of where versionformat from the metadata could influence the device.  Maybe restrict it to only certain protocols that is allowed?

superm1:
> Also, feedback from real users: they hate version numbers like 0x123456...

Well sure, but if the vendor of the UEFI device doesn't actually use decimals anywhere in the version strings, don't you want it to match the same thing everywhere you can potentially see it (BIOS setup, Windows, Linux, support site)

hughsie:
> The other option is to follow the state machine idea you weren't a fan of where versionformat from the metadata could influence the device

I'm pondering this also. I think it would also allow us to get rid of the `if $vendor=Dell then QUAD` stuff too. We'd have to recalculate all the FuDevice `version` from version_raw everytime we refresh the metadata, and also presumably when the device is added with an UNKNOWN version format. I do think we should be setting `UNKNOWN` rather than `PLAIN` in this kind of case, and making the engine do the work to create a version string from the version_raw if no metadata exists.

superm1:
> I'm pondering this also. I think it would also allow us to get rid of the if $vendor=Dell then QUAD stuff too. 

The main downside that comes to mind is that `fwupdmgr install file.cab` on a manually downloaded CAB will behave differently until you have run `fwupdmgr refresh` once.

I think generally this isn't a major problem though

>We'd have to recalculate all the FuDevice version from version_raw everytime we refresh the metadata, and also presumably when the device is added with an UNKNOWN version format. I do think we should be setting UNKNOWN rather than PLAIN in this kind of case, and making the engine do the work to create a version string from the version_raw if no metadata exists.

Could also lose `uefi-lvfs.quirk`

hughsie:
> The main downside that comes to mind is that fwupdmgr install file.cab on a manually downloaded CAB will behave differently until you have run fwupdmgr refresh once.

I still think we have to support systems that have no internet access. I know of at least one major company just installing .cabs using something like ansible :/ In this case I guess we'd use the version format from the archive just for the install_task rather than for setting the device.

> Could also lose uefi-lvfs.quirk

Maybe; that's slight different in where we don't want `<name>` to be mirrored into `FuDevice:name` -- the autoquirk is being generated from the **category** rather than the name which is probably weird enough to be specific to UEFI devices.

hughsie:
> the autoquirk is being generated from the category

I can have a play with this and lay some of the foundation work for version formats today. It does require something like https://github.com/fwupd/lvfs-website/pull/518 to do properly...

hughsie:
This depends on https://github.com/fwupd/fwupd/pull/1804 I guess now.

superm1:
Closing in favor of working on cleaning up #1807

PR #1801, opened 2020-02-21
fu-device-list: Check protocol before de-duping devices

superm1:
Some Dell devices offer the exact same GUID in UEFI as well as in NVME
plugins.  These devices can be flashed in either way, but the payload
for them is different.

The UEFI payload includes relevant headers and metadata to be acceptable
from a capsule.  The NVME payload is raw data that the drive will process.

Since baa9e75ca86f29239de03f7ab86a3e1309c37ac0, fwupd should enforce that
the correct device is flashed with the correct protocol.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-device-list.c
+++ b/src/fu-device-list.c
@@ -569,9 +569,12 @@ fu_device_list_add (FuDeviceList *self, FuDevice *device)
 		return;
 	}
 
-	/* added the same device from a different plugin */
-	if (item != NULL && g_strcmp0 (fu_device_get_plugin (item->device),
-				       fu_device_get_plugin (device)) != 0) {
+	/* added the same device, supporting same protocol, from a different plugin */
+	if (item != NULL &&
+	    g_strcmp0 (fu_device_get_plugin (item->device),
+		       fu_device_get_plugin (device)) != 0 &&
+	    g_strcmp0 (fu_device_get_protocol (item->device),
+		       fu_device_get_protocol (device)) == 0) {
 		if (fu_device_get_priority (device) < fu_device_get_priority (item->device)) {
 			g_debug ("ignoring device %s [%s] as better device %s [%s] already exists",
 				 fu_device_get_id (device),
--- a/src/fu-self-test.c
+++ b/src/fu-self-test.c
@@ -663,6 +663,90 @@ fu_engine_requirements_other_device_func (gconstpointer user_data)
 }
 
 static void
+fu_engine_requirements_protocol_check_func (gconstpointer user_data)
+{
+	g_autoptr(FuDevice) device1 = fu_device_new ();
+	g_autoptr(FuDevice) device2 = fu_device_new ();
+	g_autoptr(FuEngine) engine = fu_engine_new (FU_APP_FLAGS_NONE);
+	g_autoptr(GPtrArray) devices = NULL;
+	g_autoptr(FuInstallTask) task1 = NULL;
+	g_autoptr(FuInstallTask) task2 = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(XbNode) component = NULL;
+	g_autoptr(XbSilo) silo = NULL;
+	g_autoptr(XbSilo) silo_empty = xb_silo_new ();
+	gboolean ret;
+
+	const gchar *xml =
+		"<component>"
+		"  <provides>"
+		"    <firmware type=\"flashed\">12345678-1234-1234-1234-123456789012</firmware>"
+		"  </provides>"
+		"  <releases>"
+		"    <release version=\"4.5.7\">"
+		"      <checksum type=\"sha1\" filename=\"bios.bin\" target=\"content\"/>"
+		"    </release>"
+		"  </releases>"
+		"  <custom>"
+		"    <value key=\"LVFS::UpdateProtocol\">org.bar</value>"
+		"  </custom>"
+
+		"</component>";
+
+	/* no metadata in daemon */
+	fu_engine_set_silo (engine, silo_empty);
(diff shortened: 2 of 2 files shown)


superm1:
> I'd feel a bit happier with a unit test for this scenario.

OK. Just added.  Hopefully I got it right.

hughsie:
Are you *sure* you want to use the same guid? I think the LVFS might need a related fix if so...

superm1:
> Are you _sure_ you want to use the same guid? I think the LVFS might need a related fix if so...

No, I don't want to, I would rather not have this GUID as part of the NVME device.  However, there are valid `$REASONS` other folks want to have same in both.

hughsie:
>  However, there are valid $REASONS other folks want to have same in both.

Fair enough. You might be on the hook for the LVFS fix then :)

superm1:
> > However, there are valid $REASONS other folks want to have same in both.
> 
> Fair enough. You might be on the hook for the LVFS fix then :)

I don't know the LVFS problem you're hypothesizing, so I might not be sure how to fix it 😄

src/fu-device-list.c
```diff
@@ -569,9 +569,12 @@ fu_device_list_add (FuDeviceList *self, FuDevice *device)
...
 
-	/* added the same device from a different plugin */
-	if (item != NULL && g_strcmp0 (fu_device_get_plugin (item->device),
-				       fu_device_get_plugin (device)) != 0) {
+	/* added the same device, supporting same protocol, from a different plugin */
+	if (item != NULL &&
+	    g_strcmp0 (fu_device_get_plugin (item->device),
+		       fu_device_get_plugin (device)) != 0 &&
+	    g_strcmp0 (fu_device_get_protocol (item->device),
+		       fu_device_get_protocol (device)) == 0) {
```
hughsie: So just a thought; we added this chunk to dedupe things like the TPM (Dell upgradable, fallback one not so) -- isn't this going to break that dedupe action?

hughsie:
> I don't know the LVFS problem you're hypothesizing,

Ohh, I suppose you're only going to be uploading one protocol type firmware to the LVFS, so perhaps my concern is misplaced.

src/fu-device-list.c
superm1: The original reason it was added was Synaptics MST flashing, so that the `dell-dock` plugin got preferred over the `synapticsmst` plugin for resiliency, although I think it gets used for the thunderbolt/dell-dock as well.  Let me double check all 3 of those aren't ruined by this.

superm1:
> Ohh, I suppose you're only going to be uploading one protocol type firmware to the LVFS, so perhaps my concern is misplaced.

Yeah, if we have the same firmware available in both variants using the same GUID I'll be scratching my head too.

src/fu-device-list.c
hughsie: If it helps, I'm not against a .quirk for ALLOW_DUPLICATE_GUIDS to disable this check. Putting that on UEFI **or** NVMe would be quite a nice workaround.

superm1: I don't have the HDD scenario handy anymore, but I did check it before I submitted the PR.

TBT is ok because there is an explicit check for whether you're connected a thunderbolt port now in the code.  The dedupe code doesn't run:
https://github.com/fwupd/fwupd/blob/9a68084dbc2c5886c927875f977e8f50f9161bc3/plugins/dell-dock/fu-dell-dock-i2c-ec.c#L185

Synaptics MST is ok because of the priority check matches identically in both plugins:

```
20:26:09:0011 FuDeviceList         ignoring device 585c346c009da3cc19a400fb59a4ed8d0cbd0151 [synaptics_mst] as better device 0060e00540acb5ded8f04c214fb157816bbbb6a6 [dell_dock] already exists
```

TPM is ok, but only because it hits an earlier check.
```
20:30:07:0884 FuDeviceList         device c6a80ac3a22083423992a3cb15018989f37834d6 already exists, ignoring
```

So it does work
```
$ sudo fwupdtool get-devices --plugin-whitelist=dell --plugin-whitelist=tpm --plugin-whitelist=uefi
XPS 13 9380
│
├─System Firmware:
│     Current version:     1.9.1
│     Minimum Version:     1.9.1
│   
└─TPM 2.0:
      Device ID:           c6a80ac3a22083423992a3cb15018989f37834d6
      Summary:             Platform TPM device
      Current version:     7.2.0.2
      Vendor:              Dell Inc. (PCI:0x1028)
      Update Error:        Updating disabled due to TPM ownership
      GUIDs:               d621e687-b00d-539a-aeac-ac3eb530cbc5 ← 08af-2.0
                           ff71992e-52f7-5eea-94ef-883e56e034c6 ← system-tpm
                           7d65b10b-bb24-552d-ade5-590b3b278188 ← DELL-TPM-2.0-NTC-NPCT
                           6f5ddd3a-8339-5b2a-b9a6-cf3b92f6c86d ← DELL-TPM-2.0-NTC-NPCT75x
                           fe462d4a-e48f-5069-9172-47330fc5e838 ← DELL-TPM-2.0-NTC-NPCT75xrls
                           • Requires AC power
```

superm1: > If it helps, I'm not against a .quirk for ALLOW_DUPLICATE_GUIDS to disable this check. Putting that on UEFI **or** NVMe would be quite a nice workaround.

I wouldn't rule out that this scenario happen in the future with a different device, so I would prefer a generic solution if we can keep one.  If this breaks something else in the future and we need to revert it a quirk for uefi/nvme will be fine too.

PR #1802, opened 2020-02-22
Added completion script for fish shell

exploide:
Type of pull request:
- [ ] New plugin
- [ ] Code fix
- [X] Feature
- [ ] Documentation

I saw this repository already contains a completion script for bash. This PR adds a completion script for fish shell.

In case you don't like including it here, I could also submit it to the fish repository. However, it usually is preferred to ship completion scripts by upstream projects and bash is already included here.

What I have not done so far is adding the script to the spec file. It would belong into `/usr/share/fish/vendor_completions.d`.

### DIFF
--- a/contrib/debian/fwupd.install
+++ b/contrib/debian/fwupd.install
@@ -1,6 +1,7 @@
 usr/bin/*
 etc/*
 usr/share/bash-completion
+usr/share/fish/vendor_completions.d
 usr/share/fwupd/*
 usr/share/dbus-1/*
 usr/share/icons/*
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -272,6 +272,7 @@ mkdir -p --mode=0700 $RPM_BUILD_ROOT%{_localstatedir}/lib/fwupd/gnupg
 %{_datadir}/bash-completion/completions/fwupdmgr
 %{_datadir}/bash-completion/completions/fwupdtool
 %{_datadir}/bash-completion/completions/fwupdagent
+%{_datadir}/fish/vendor_completions.d/fwupdmgr.fish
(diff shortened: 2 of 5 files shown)


hughsie:
Doesn't it need installing somewhere?

exploide:
> Doesn't it need installing somewhere?

Fish is autoloading functions and complete scripts depending on their name whenever they are located in one of the proper directories. For this that would be inside `/usr/share/fish/vendor_completions.d`. That could be done from within the spec file, at least for RPMs.

hughsie:
I think it should be installed by meson.build too :)

exploide:
Ok, that's possible.

If I should do that, can you give me a pointer? I have no experience with this build system.

superm1:
Unlike bash-completion it appears you're not modifying the file at all with data from the installation variables.  So in that case you just need to install it to the proper directory.

1. Add your subdirectory to `data/meson.build`
2.  Some pointers in an example that installs a flat file into the filesystem are here:
https://github.com/fwupd/fwupd/blob/master/data/meson.build#L33
So make a `meson.build` in your subdirectory to do that.
3. Confirm it actually gets installed.
4. Add references to `contrib/fwupd.spec.in` and `contrib/debian/fwupd.install` as necessary to update deb/rpm packaging.

exploide:
Ok, thank you! Since I need to do these changes and run a test build, I need to wait until I'm back home at the end of the week. No progress until then, hope that's fine.

hughsie:
> No progress until then, hope that's fine

No problem at all, we appreciate the help!

exploide:
Done!

It took me way too long to understand that I need to commit my changes in order to be recognized during the local test builds...

hughsie:
Can you rebase against master and squash the commits together please. Then force push to update this PR then we're good to merge. Thanks!

exploide:
Sure, done :)

PR #1804, opened 2020-02-24
Allow server metadata to set the device name on some devices

hughsie:
This does 'bleed' the metadata contents into areas previously covered by quirks,
but in this case may be pragmatic and more up to date than a build-time
generated quirk file, which increases the user-friendliness of fwupdmgr.


hughsie:
Untested until https://github.com/fwupd/lvfs-website/pull/518 lands...

libfwupd/fwupd-enums.c
```diff
@@ -267,6 +271,10 @@ fwupd_device_flag_from_string (const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_VERSION_CHECK_REQUIRED;
 	if (g_strcmp0 (device_flag, "install-all-releases") == 0)
 		return FWUPD_DEVICE_FLAG_INSTALL_ALL_RELEASES;
+	if (g_strcmp0 (device_flag, "md-set-name") == 0)
+		return FWUPD_DEVICE_FLAG_MD_SET_NAME;
```
superm1: What do you envision that this would be used for instead of the category one?

hughsie: Two things came to mind: DFU devices and also USB hubs that are actually docking stations.

hughsie:
It worked!

PR #1807, opened 2020-02-24
Allow server metadata to set the version format on some devices

hughsie:
It appears just choosing 'if Dell then QUAD else TRIPLET' isn't good enough.

### DIFF
--- a/plugins/uefi/fu-uefi-device.c
+++ b/plugins/uefi/fu-uefi-device.c
@@ -695,6 +695,7 @@ fu_uefi_device_probe (FuDevice *device, GError **error)
 	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_INTERNAL);
 	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
 	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_REQUIRE_AC);
+	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_MD_SET_VERFMT);
 
 	/* add icons */
 	if (self->kind == FU_UEFI_DEVICE_KIND_DEVICE_FIRMWARE) {
@@ -726,6 +727,7 @@ static void
 fu_uefi_device_init (FuUefiDevice *self)
 {
 	fu_device_set_protocol (FU_DEVICE (self), "org.uefi.capsule");
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_NUMBER);
 }
 
 static void
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -187,6 +187,8 @@ fwupd_device_flag_to_string (FwupdDeviceFlags device_flag)
 		return "md-set-name";
 	if (device_flag == FWUPD_DEVICE_FLAG_MD_SET_NAME_CATEGORY)
 		return "md-set-name-category";
+	if (device_flag == FWUPD_DEVICE_FLAG_MD_SET_VERFMT)
+		return "md-set-verfmt";
 	if (device_flag == FWUPD_DEVICE_FLAG_UNKNOWN)
 		return "unknown";
 	return NULL;
@@ -275,6 +277,8 @@ fwupd_device_flag_from_string (const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_MD_SET_NAME;
 	if (g_strcmp0 (device_flag, "md-set-name-category") == 0)
 		return FWUPD_DEVICE_FLAG_MD_SET_NAME_CATEGORY;
+	if (g_strcmp0 (device_flag, "md-set-verfmt") == 0)
+		return FWUPD_DEVICE_FLAG_MD_SET_VERFMT;
 	return FWUPD_DEVICE_FLAG_UNKNOWN;
 }
 
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -101,6 +101,7 @@ typedef enum {
  * @FWUPD_DEVICE_FLAG_INSTALL_ALL_RELEASES:	Install each intermediate release rather than jumping direct to newest
  * @FWUPD_DEVICE_FLAG_MD_SET_NAME:		Set the device name from the metadata <name> if available
  * @FWUPD_DEVICE_FLAG_MD_SET_NAME_CATEGORY:	Set the device name from the metadata <category> if available
+ * @FWUPD_DEVICE_FLAG_MD_SET_VERFMT:		Set the device version format from the metadata if available
  *
  * The device flags.
  **/
@@ -139,6 +140,7 @@ typedef enum {
 #define FWUPD_DEVICE_FLAG_INSTALL_ALL_RELEASES	(1u << 31)	/* Since: 1.3.7 */
 #define FWUPD_DEVICE_FLAG_MD_SET_NAME		(1llu << 32)	/* Since: 1.3.9 */
 #define FWUPD_DEVICE_FLAG_MD_SET_NAME_CATEGORY	(1llu << 33)	/* Since: 1.3.9 */
+#define FWUPD_DEVICE_FLAG_MD_SET_VERFMT		(1llu << 34)	/* Since: 1.3.9 */
 #define FWUPD_DEVICE_FLAG_UNKNOWN		G_MAXUINT64	/* Since: 0.7.3 */
 typedef guint64 FwupdDeviceFlags;
 
--- a/libfwupdplugin/fu-quirks.h
+++ b/libfwupdplugin/fu-quirks.h
@@ -44,8 +44,6 @@ gboolean	 fu_quirks_lookup_by_id_iter		(FuQuirks	*self,
 							 gpointer	 user_data);
 
 #define	FU_QUIRKS_PLUGIN			"Plugin"
-#define	FU_QUIRKS_UEFI_VERSION_FORMAT		"UefiVersionFormat"
-#define	FU_QUIRKS_DAEMON_VERSION_FORMAT		"ComponentIDs"
 #define	FU_QUIRKS_FLAGS				"Flags"
 #define	FU_QUIRKS_SUMMARY			"Summary"
 #define	FU_QUIRKS_ICON				"Icon"
--- a/plugins/dell/dell.quirk
+++ b/plugins/dell/dell.quirk
@@ -32,9 +32,3 @@ ParentGuid = e7ca1f36-bf73-4574-afe6-a4ccacabf479
 [Guid=MST-tb18-vmm3330-274]
 Plugin = synaptics_mst
 ParentGuid = e7ca1f36-bf73-4574-afe6-a4ccacabf479
-
-[SmbiosManufacturer=Dell Inc.]
-UefiVersionFormat = dell-bios
-
-[SmbiosManufacturer=Alienware]
-UefiVersionFormat = dell-bios
--- a/plugins/dell/fu-plugin-dell.c
+++ b/plugins/dell/fu-plugin-dell.c
@@ -231,26 +231,6 @@ fu_plugin_dell_inject_fake_data (FuPlugin *plugin,
 	data->can_switch_modes = TRUE;
 }
 
-static FwupdVersionFormat
-fu_plugin_dell_get_version_format (FuPlugin *plugin)
-{
-	const gchar *content;
-	const gchar *quirk;
-	g_autofree gchar *group = NULL;
-
-	content = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_MANUFACTURER);
-	if (content == NULL)
-		return FWUPD_VERSION_FORMAT_TRIPLET;
-
-	/* any quirks match */
-	group = g_strdup_printf ("SmbiosManufacturer=%s", content);
-	quirk = fu_plugin_lookup_quirk_by_id (plugin, group,
-					      FU_QUIRKS_UEFI_VERSION_FORMAT);
-	if (quirk == NULL)
-		return FWUPD_VERSION_FORMAT_TRIPLET;
-	return fwupd_version_format_from_string (quirk);
-}
-
 static gboolean
 fu_plugin_dell_capsule_supported (FuPlugin *plugin)
 {
@@ -319,7 +299,7 @@ fu_plugin_usb_device_added (FuPlugin *plugin,
 			    GError **error)
 {
 	FuPluginData *data = fu_plugin_get_data (plugin);
-	FwupdVersionFormat version_format;
+	FwupdVersionFormat version_format = FWUPD_VERSION_FORMAT_DELL_BIOS;
 	guint16 pid;
 	guint16 vid;
 	const gchar *query_str;
@@ -372,7 +352,6 @@ fu_plugin_usb_device_added (FuPlugin *plugin,
 	g_debug ("Dock cable type: %" G_GUINT32_FORMAT, dock_info->cable_type);
 	g_debug ("Dock location: %d", dock_info->location);
 	g_debug ("Dock component count: %d", dock_info->component_count);
-	version_format = fu_plugin_dell_get_version_format (plugin);
 
 	for (guint i = 0; i < dock_info->component_count; i++) {
 		g_autofree gchar *fw_str = NULL;
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -463,30 +463,6 @@ fu_plugin_device_registered (FuPlugin *plugin, FuDevice *device)
 	}
 }
 
-static FwupdVersionFormat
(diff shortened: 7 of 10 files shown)


hughsie:
100% untested, and probably needs unit tests too.

superm1:
> 100% untested, and probably needs unit tests too.

I think it needs to drop the the automatically set dell DMI to `dell-bios`  in the uefi plugin too.

hughsie:
> I think it needs to drop the the automatically set dell DMI

Okay, done; i might have broken the world. Any chance you can try with Dell hardware now? You might also get your PLAIN device working correctly too if you set the VersionFormat and upload it to a embargo remote...

plugins/uefi/fu-uefi-device.c
```diff
@@ -725,6 +725,7 @@ static void
 fu_uefi_device_init (FuUefiDevice *self)
 {
 	fu_device_set_protocol (FU_DEVICE (self), "org.uefi.capsule");
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_TRIPLET);
```
superm1: So as part of of this change, shouldn't the default move to PLAIN or UNKNOWN for UEFI?  And then all metadata updates for the right value?

superm1:
> > I think it needs to drop the the automatically set dell DMI
> 
> Okay, done; i might have broken the world. Any chance you can try with Dell hardware now? You might also get your PLAIN device working correctly too if you set the VersionFormat and upload it to a embargo remote...

OK with no metadata as expected the version looks crazy (I do think it should probably just render `VersionRaw` until metadata comes through instead of nonsense versions)
```
├─System Firmware:
│     Device ID:           b6c08fb9e5384d9d101853cc1ca20cf0ce2df2e2
│     Current version:     0.1.1024
│     Minimum Version:     0.1.1024
│     GUID:                cbe49cca-492b-93e8-b0f4-f693501f271b
```

Trying to refresh metadata seems it broke:
```
$ sudo gdb --args  fwupdtool refresh 
GNU gdb (Ubuntu 9.1-0ubuntu1) 9.1
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from fwupdtool...
(gdb) bt
No stack.
(gdb) run
Starting program: /usr/local/bin/fwupdtool refresh
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff44d9700 (LWP 44003)]
[New Thread 0x7ffff3cd8700 (LWP 44004)]
[New Thread 0x7ffff3382700 (LWP 44005)]
[New Thread 0x7ffff2b81700 (LWP 44006)]
Loading…                 [   -                                   ]
OUI quirk required, please see https://github.com/fwupd/fwupd/wiki/ATA-Disk:-OUI-Quirk-Required!
---
[DeviceInstanceId=OUI\0004cf]
Vendor = Seagate
VendorId = ATA:0x1BB1
---
bb
Thread 1 "fwupdtool" received signal SIGTRAP, Trace/breakpoint trap.
0x00007ffff79ccfe5 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
(gdb) bt
#0  0x00007ffff79ccfe5 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff79ce2c9 in g_logv () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff79ce493 in g_log () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x000055555557f94e in fu_engine_md_refresh_device_verfmt (self=0x5555555f2040, device=0x5555556e82c0, component=0x555555671a10) at ../src/fu-engine.c:2798
#4  0x000055555557fa7b in fu_engine_md_refresh_device (self=0x5555555f2040, device=0x5555556e82c0) at ../src/fu-engine.c:2818
#5  0x000055555557fb05 in fu_engine_md_refresh_devices (self=0x5555555f2040) at ../src/fu-engine.c:2827
#6  0x0000555555587491 in fu_engine_load (self=0x5555555f2040, flags=FU_ENGINE_LOAD_FLAG_NONE, error=0x7fffffffe248) at ../src/fu-engine.c:5374
#7  0x0000555555569c0d in fu_util_start_engine (priv=0x5555555e66c0, flags=FU_ENGINE_LOAD_FLAG_NONE, error=0x7fffffffe248) at ../src/fu-tool.c:129
#8  0x000055555556e6b3 in fu_util_refresh (priv=0x5555555e66c0, values=0x7fffffffe628, error=0x7fffffffe248) at ../src/fu-tool.c:1674
#9  0x0000555555593842 in fu_util_cmd_array_run (array=0x5555555e6800, priv=0x5555555e66c0, command=0x7fffffffe856 "refresh", values=0x7fffffffe628, error=0x7fffffffe248) at ../src/fu-util-common.c:545
#10 0x000055555556f6fd in main (argc=2, argv=0x7fffffffe618) at ../src/fu-tool.c:2051
```

hughsie:
> Trying to refresh metadata seems it broke:

It seems `g_error ("MO")` isn't production quality. Fixed.

superm1:
```
XPS 13 7390
│
└─System Firmware:
      Device ID:           b6c08fb9e5384d9d101853cc1ca20cf0ce2df2e2
      Current version:     0.1.4.0
      Minimum Version:     66560
      Vendor:              Dell Inc. (DMI:Dell Inc.)
      GUID:                cbe49cca-492b-93e8-b0f4-f693501f271b
                           • Updatable
                           • Requires AC power
                           • Supported on remote server
                           • Needs a reboot after installation
                           • Device is usable for the duration of the update
```

So that's working now, but also need to adjust minimum version at the same time

superm1:
Well working-ish.  I think that's supposed to show up as `dell-bios`, so it might have taken the wrong one from the metadata.

hughsie:
>so it might have taken the wrong one from the metadata

Ahh, indeed. I forgot we did the reverse order trick to maintain compat. I'll fix up when I get back home.

hughsie:
> I'll fix up when I get back home.

Fixed hopefully.

superm1:
> > I'll fix up when I get back home.
> 
> Fixed hopefully.

Yep, that looks right.
I started the stuff for the libraries to fixup version raw/bootloader as well but didn't finish connecting it yet.

hughsie:
> but didn't finish connecting it yet.

I appreciate that. I did wonder about removing the version_format setting from the `set_version*` functions as it's getting a bit unweldy. I agree with you the bootloader and firmware (and minimum) all prbably need to share a version format to preserve sanity.

hughsie:
> I did wonder about removing the version_format setting from the set_version* functions

Yes, this is what we should do; having 4 things set the same property is kinda crazy. Do you want me to fix up that or are you on a roll?

superm1:
Meetings for the afternoon, so probably can't touch it again until tomorrow

hughsie:
> so probably can't touch it again until tomorrow

I've untangled some stuff. Comments welcome.

libfwupdplugin/fu-device.c
```diff
@@ -1360,18 +1360,41 @@ fu_device_set_id (FuDevice *self, const gchar *id)
...
  * @version: (allow-none): a string, e.g. `1.2.3`
- * @fmt: a #FwupdVersionFormat, e.g. %FWUPD_VERSION_FORMAT_TRIPLET
  *
  * Sets the device version, sanitizing the string if required.
  *
  * Since: 1.2.9
  **/
 void
-fu_device_set_version (FuDevice *self, const gchar *version, FwupdVersionFormat fmt)
+fu_device_set_version (FuDevice *self, const gchar *version)
```
superm1: What do you think about making this return true/false?  Now we have an order of operations that version format must be set before versions, so I think it would be good to not hide silent errors in the logs.

libfwupdplugin/fu-device.c
```diff
@@ -1468,14 +1472,13 @@ fu_device_set_version_lowest (FuDevice *self, const gchar *version, FwupdVersion
...
  * @version: (allow-none): a string, e.g. `1.2.3`
- * @fmt: a #FwupdVersionFormat, e.g. %FWUPD_VERSION_FORMAT_TRIPLET
  *
  * Sets the device bootloader version, sanitizing the string if required.
  *
  * Since: 1.3.9
  **/
 void
-fu_device_set_version_bootloader (FuDevice *self, const gchar *version, FwupdVersionFormat fmt)
+fu_device_set_version_bootloader (FuDevice *self, const gchar *version)
```
superm1: I'm almost wondering if it's worth having a static function with an enum or something for the type of version string being passed so this isn't duplicated 90% 3 times.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -436,7 +437,7 @@ fu_dell_dock_ec_get_dock_info (FuDevice *device,
 		return FALSE;
 	}
 
-	fu_device_set_version_lowest (device, self->ec_minimum_version, FWUPD_VERSION_FORMAT_PAIR);
+	fu_device_set_version_lowest (device, self->ec_minimum_version);
```
superm1: Perfect example of why it makes sense to hardcode to same for all 3 versions, this was a bug waiting to happen.

plugins/fastboot/fu-fastboot-device.c
```diff
@@ -318,8 +318,10 @@ fu_fastboot_device_setup (FuDevice *device, GError **error)
 	/* bootloader version */
 	if (!fu_fastboot_device_getvar (device, "version-bootloader", &version_bootloader, error))
 		return FALSE;
-	if (version_bootloader != NULL && version_bootloader[0] != '\0')
-		fu_device_set_version_bootloader (device, version_bootloader, FWUPD_VERSION_FORMAT_PAIR);
+	if (version_bootloader != NULL && version_bootloader[0] != '\0') {
+		fu_device_set_version_format (device, FWUPD_VERSION_FORMAT_PAIR);
```
superm1: I'm not sure this is actually needed now - it inherits from USB no?

hughsie: Usb is BCD I think

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
```diff
@@ -578,9 +578,9 @@ fu_synaptics_rmi_device_setup (FuDevice *device, GError **error)
 				  f01_basic->data[2],
 				  f01_basic->data[3],
 				  priv->flash.build_id);
-	fu_device_set_version (device, fw_ver, FWUPD_VERSION_FORMAT_TRIPLET);
+	fu_device_set_version (device, fw_ver);
 	bl_ver = g_strdup_printf ("%u.0", priv->flash.bootloader_id[1]);
-	fu_device_set_version_bootloader (device, bl_ver, FWUPD_VERSION_FORMAT_TRIPLET);
+	fu_device_set_version_bootloader (device, bl_ver);
```
superm1: This is emitting `54.0 is not a valid triplet`....

hughsie: Good catch. This needs to be 0.0.%u I guess.

superm1:
OK so with branch as it stands today here is what fresh install sees:
```
└─System Firmware:
      Device ID:           6c24a747f97668873b761558e322398a91dbf394
      Current version:     67841
      Minimum Version:     67841
      Vendor:              Dell Inc. (DMI:Dell Inc.)
      GUID:                ce945437-7358-49f1-95d8-6b694a10a755
                           • Updatable
                           • Requires AC power
                           • Needs a reboot after installation
                           • Device is usable for the duration of the update
    
```

Here is after the update:
```
└─System Firmware:
      Device ID:           6c24a747f97668873b761558e322398a91dbf394
      Current version:     1.9.1
      Minimum Version:     1.9.1
      Vendor:              Dell Inc. (DMI:Dell Inc.)
      GUID:                ce945437-7358-49f1-95d8-6b694a10a755
                           • Updatable
                           • Requires AC power
                           • Supported on remote server
                           • Needs a reboot after installation
                           • Device is usable for the duration of the update
```

At least _looks_ to be working properly to me.

superm1:
Now running `get-details` on a CAB from a fresh install gets this:
```
$ fwupdmgr get-details ~/.cache/fwupd/4bb7c12728f69531b38ae85d0de80c69d48cfab2-XPS_9380_1.9.1.cab
version format unset and version 0x10901 ambiguous
```
Which I suppose isn't _too_ surprising.

But this is:
```
$ fwupdmgr install ~/.cache/fwupd/4bb7c12728f69531b38ae85d0de80c69d48cfab2-XPS_9380_1.9.1.cab
Specified firmware is older than installed '0x10901 < 67841'
$ fwupdmgr refresh 
Fetching metadata https://cdn.fwupd.org/downloads/firmware.xml.gz
Downloading…             [***************************************]
Fetching signature https://cdn.fwupd.org/downloads/firmware.xml.gz.asc

Successfully downloaded new metadata: 8 local devices supported
$ fwupdmgr install ~/.cache/fwupd/4bb7c12728f69531b38ae85d0de80c69d48cfab2-XPS_9380_1.9.1.cab
Specified firmware is already installed '1.9.1'
```

superm1:
Treating it as a number makes it behave more like what I expect.
```
$ fwupdmgr get-details ~/.cache/fwupd/4bb7c12728f69531b38ae85d0de80c69d48cfab2-XPS_9380_1.9.1.cab
version format unset and version 0x10901 ambiguous
$ fwupdmgr install ~/.cache/fwupd/4bb7c12728f69531b38ae85d0de80c69d48cfab2-XPS_9380_1.9.1.cab
Specified firmware is already installed '67841'
$ fwupdmgr refresh
Fetching metadata https://cdn.fwupd.org/downloads/firmware.xml.gz
Downloading…             [***************************************]
Fetching signature https://cdn.fwupd.org/downloads/firmware.xml.gz.asc

Successfully downloaded new metadata: 8 local devices supported
$ fwupdmgr install ~/.cache/fwupd/4bb7c12728f69531b38ae85d0de80c69d48cfab2-XPS_9380_1.9.1.cab
Specified firmware is already installed '1.9.1'
```

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
hughsie: Fixed.

libfwupdplugin/fu-device.c
hughsie: I think if it gets any more complicated I agree, but at the moment it's pretty self contained.

libfwupdplugin/fu-device.c
hughsie: if it returns with an error, it also needs a GError -- but this isn't anything the daemon or plugin can handle in a sane way. This is the kind of thing you can only really catch when developing the plugin.

hughsie:
I've split this out as https://github.com/fwupd/fwupd/pull/1812 and https://github.com/fwupd/fwupd/pull/1813 for easier review. When those are in master I'll rebase this branch on top and squash everything down.

hughsie:
@superm1 this needs some heavy testing, especially on stuff like docks.

superm1:
> @superm1 this needs some heavy testing, especially on stuff like docks.

I don't think that it influences anything on Dell's docks that I can tell.  It's probably more for non-Dell you mean?

I hammered out the various permutations of `get-devices` and sent up a handful of fixes.

The behavior matches what I expect now for the Dell UEFI system firmware device too.

hughsie:
>  don't think that it influences anything on Dell's docks that I can tell

I was hoping that would be a good workout for the usb version format BCD ignoring stuff...

superm1:
> I was hoping that would be a good workout for the usb version format BCD ignoring stuff...

I did see that the correct versions shown on the docks I checked, and given the delta from the previous commit, it seems low risk to me:

```
diff --git a/plugins/dell-dock/fu-dell-dock-hid.c b/plugins/dell-dock/fu-dell-dock-hid.c
index 267b3725..18e129cc 100644
--- a/plugins/dell-dock/fu-dell-dock-hid.c
+++ b/plugins/dell-dock/fu-dell-dock-hid.c
@@ -185,7 +185,8 @@ fu_dell_dock_hid_get_hub_version (FuDevice *self,
        version = g_strdup_printf ("%02x.%02x",
                                   cmd_buffer.data[10],
                                   cmd_buffer.data[11]);
-       fu_device_set_version (self, version, FWUPD_VERSION_FORMAT_PAIR);
+       fu_device_set_version_format (self, FWUPD_VERSION_FORMAT_PAIR);
+       fu_device_set_version (self, version);
        return TRUE;
 }
 
```

Issue #1814, opened 2020-02-25
Laptop went to suspend while installing thunderbolt controller update


allanday:
Deets:

- Operating system and version: Fedora 31 Silverblue
- Is secure boot enabled (only for the UEFI plugin)? Presumably.
- Device: Lenovo X1 Carbon 6th gen, connected to a dock by thunderbolt, lid closed with a single external display

What happened:

  - I was in the Software app, and saw that there was an update pending for my thunderbolt controller
 - I clicked update. The button became insensitive and said installing.
 - My external monitor suddenly went blank, there was a pause of a few seconds, then the power button on the dock started to pulse. After a few seconds I pressed the button and was able to log in.

Not a great experience.

hughsie:
I think what happened here is:

 * Laptop in docking station with lid closed
 * TBT update started
 * External monitor went away momentarily
 * GNOME did the laptop-lid-closed action

I think we need to do the same logind inihibit that PackageKit does, i.e. https://www.freedesktop.org/wiki/Software/systemd/inhibit/ for the duration of the update. I can work on this after lunch.

PR #1815, opened 2020-02-25
Inihbit all power management actions using logind when updating

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/1814

### DIFF
--- a/plugins/logind/fu-plugin-logind.c
+++ b/plugins/logind/fu-plugin-logind.c
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <gio/gunixfdlist.h>
+#include <glib/gstdio.h>
+
+#include "fu-plugin-vfuncs.h"
+#include "fu-hash.h"
+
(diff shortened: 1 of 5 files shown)


plugins/logind/fu-plugin-logind.c
```diff
@@ -0,0 +1,124 @@
...
+		g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
+					       G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS |
+					       G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES,
+					       NULL,
+					       "org.freedesktop.login1",
+					       "/org/freedesktop/login1",
+					       "org.freedesktop.login1.Manager",
+					       NULL,
+					       error);
+	if (data->logind_proxy == NULL) {
```
superm1: As part of startup do you want to do something similar to https://github.com/fwupd/fwupd/commit/6b8c2612251c8b19ec707d2fce13051e66bf73e2#diff-d70662bee1ff6feea3befa4f19a01b18 to make sure that it really is available?

hughsie: I guess it's a good idea. I don't think logind is ever autostarted, but it can't hurt.

PR #1816, opened 2020-02-25
Check secure boot in TPM event log plugin

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
The first commit lgtm, the second will review later.

plugins/uefi/fu-self-test.c
```diff
@@ -167,7 +167,7 @@ fu_uefi_device_func (void)
 }
 
 static void
-fu_uefi_vars_func (void)
```
hughsie: This should move to libfwupdplugin if it's shared code now.

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -31,6 +32,30 @@ fu_plugin_destroy (FuPlugin *plugin)
...
+fu_plugin_tpm_eventlog_create_dummy (FuPlugin *plugin, const gchar *reason, GError **error)
+{
+	const gchar *key;
+	g_autoptr(FuDevice) dev = fu_device_new ();
+
+	key = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_MANUFACTURER);
+	if (key != NULL)
+		fu_device_set_vendor (dev, key);
+	key = "TPM measurements";
+	fu_device_set_name (dev, key);
```
hughsie: not sure i see the reason for re-using `key`

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -31,6 +32,30 @@ fu_plugin_destroy (FuPlugin *plugin)
...
 
+static gboolean
+fu_plugin_tpm_eventlog_create_dummy (FuPlugin *plugin, const gchar *reason, GError **error)
+{
+	const gchar *key;
+	g_autoptr(FuDevice) dev = fu_device_new ();
+
+	key = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_MANUFACTURER);
+	if (key != NULL)
+		fu_device_set_vendor (dev, key);
```
hughsie: We need a vendor_id too, no?

superm1: The dummy device from UEFI it's modeled on didn't have one.  If you think so, then yes we should put it in both places.

superm1:
@hughsie re-review this one?

hughsie:
> re-review this one

Let me ponder, this all seems a bit magic. I'm not loving the dummy devices popping up everywhere, and I'm wondering if we want something more generic.

hughsie:
> put it in UpdateMessage for the UEFI device

I think I'd prefer that for now; I'm pondering another thing to represent the "platform" and it's not all crystallised in my brain yet.

superm1:
OK here's how it looks with that set of changes.

```
├─System Firmware:
│     Current version:     1.9.1
│     Minimum Version:     1.9.1
│     Update Message:      Platform firmware measurement unavailable. Secure boot is disabled in BIOS setup, enabling it may fix this issue
│                          • Supported on remote server
```

PR #1819, opened 2020-02-27
vli: Set the MSP430 version format to pair

hughsie:
Apparently the Windows tool ignores the first byte, so we should too.


hughsie:
I'm just waiting for the ODM to verify the changes are as expected.

PR #1820, opened 2020-02-27
Create FuCabinet and untangle a lot of legacy code

hughsie:
Replace fu_common_cab_build_silo() with an actual GObject that can hold parsing
state. This cleans up the code a lot, and means we can add additional
functionality in the future without breaking ABI or API.

The long term plan is to verify the metadata and payload signatures when
parsing FuCabinet, rather than much later in _check_requirements().
This of course requires passing in a keyring context (which we don't yet have)
and would mean we can stop setting the various confusing 'fwupd::ReleaseBlob'
XbNode extra data.

No logic changes for now, just a lot of moving things into sane places.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-cabinet.c
```diff
@@ -0,0 +1,574 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+/**
+ * fu_cabinet_parse:
+ * @self: A #FuCabinet
+ * @data: A #GBytes
+ * @fn: A #FuCabinetParseFlags, e.g. %FU_CABINET_PARSE_FLAG_NONE
```
superm1: wrong variable name

hughsie: Good catch, thanks!

PR #1822, opened 2020-02-27
Use xb_builder_source_add_simple_adapter

djcampello:
Type of pull request: [ X ] Code fix

    Use xb_builder_source_add_simple_adapter instead
    xb_builder_source_add_adapter to avoid sniffing MIME type for output xml
    contents

### DIFF
--- a/libfwupdplugin/fu-quirks.c
+++ b/libfwupdplugin/fu-quirks.c
@@ -174,9 +174,15 @@ fu_quirks_add_quirks_for_path (FuQuirks *self, XbBuilder *builder,
 		g_autoptr(XbBuilderSource) source = xb_builder_source_new ();
 
 		/* load from keyfile */
+#if LIBXMLB_CHECK_VERSION(0,1,15)
+		xb_builder_source_add_simple_adapter (source, "text/plain,.quirk",
+						      fu_quirks_convert_quirk_to_xml_cb,
+						      NULL, NULL);
+#else
 		xb_builder_source_add_adapter (source, "text/plain,.quirk",
 					       fu_quirks_convert_quirk_to_xml_cb,
 					       NULL, NULL);
+#endif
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-quirks.c
```diff
@@ -174,9 +174,9 @@ fu_quirks_add_quirks_for_path (FuQuirks *self, XbBuilder *builder,
 		g_autoptr(XbBuilderSource) source = xb_builder_source_new ();
 
 		/* load from keyfile */
-		xb_builder_source_add_adapter (source, "text/plain,.quirk",
-					       fu_quirks_convert_quirk_to_xml_cb,
-					       NULL, NULL);
```
hughsie: You need to use `LIBXMLB_CHECK_VERSION` to avoid hard-depending on the very new symbol. Just fall back to the non-simple version.

libfwupdplugin/fu-quirks.c
```diff
@@ -174,9 +174,9 @@ fu_quirks_add_quirks_for_path (FuQuirks *self, XbBuilder *builder,
...
+		xb_builder_source_add_simple_adapter (source, "text/plain,.quirk",
```
superm1: To avoid forcing a hard dependency on the updated libxmlb, can you please add a `#define` version guard?  If on the newer libxmlb use this, and if not then use the older method.

libfwupdplugin/fu-quirks.c
superm1: jinx, but you have a better proposed solution with the macro than the `#define`.

djcampello: It was like 12 seconds apart. Fixed. Thanks. I actually did not want the guard since I'm patching libxmlb 1.14 on my system and I overlooked this.

superm1:
Thanks, LGTM if CI passes now.

PR #1824, opened 2020-02-28
Do not fail loading in /etc/machine-id is not available

djcampello:
In Chrome OS /etc/machine-id is not available on the sandbox
environment. Chrome OS specifically rotate it constantly and is
considered PII.

Signed-off-by: Daniel Campello <campello@chromium.org>

Type of pull request:
- [ x ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5243,6 +5243,9 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 	FuRemoteListLoadFlags remote_list_flags = FU_REMOTE_LIST_LOAD_FLAG_NONE;
 	FuQuirksLoadFlags quirks_flags = FU_QUIRKS_LOAD_FLAG_NONE;
 	g_autoptr(GPtrArray) checksums = NULL;
+#ifndef _WIN32
+	g_autoptr(GError) error_local = NULL;
+#endif
 
 	g_return_val_if_fail (FU_IS_ENGINE (self), FALSE);
 	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
@@ -5254,9 +5257,9 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 /* TODO: Read registry key [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography] "MachineGuid" */
 #ifndef _WIN32
 	/* cache machine ID so we can use it from a sandboxed app */
-	self->host_machine_id = fwupd_build_machine_id ("fwupd", error);
+	self->host_machine_id = fwupd_build_machine_id ("fwupd", &error_local);
 	if (self->host_machine_id == NULL)
-		return FALSE;
+		g_debug ("%s", error_local->message);
 #endif
 	/* read config file */
 	if (!fu_config_load (self->config, error)) {


libfwupd/fwupd-common.c
```diff
@@ -348,10 +348,6 @@ fwupd_build_machine_id (const gchar *salt, GError **error)
 		}
 	}
 	if (fn == NULL) {
-		g_set_error_literal (error,
-				     FWUPD_ERROR,
-				     FWUPD_ERROR_READ,
-				     "The machine-id is not present");
 		return NULL;
```
hughsie: functions returning with failure in Glib always have to set a GError -- you can remove this part of the patch -- see below.

src/fu-engine.c
```diff
@@ -5255,8 +5255,6 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 #ifndef _WIN32
 	/* cache machine ID so we can use it from a sandboxed app */
 	self->host_machine_id = fwupd_build_machine_id ("fwupd", error);
```
hughsie: just change the `error` to `NULL` if you don't care about failure.

src/fu-engine.c
```diff
@@ -5255,8 +5255,6 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
...
-	if (self->host_machine_id == NULL)
-		return FALSE;
```
hughsie: you are correct to do this if `error` is NULL.

superm1:
I think my personal preference would be that rather than this error go away entirely, it be promoted to `g_warning` so it at least is logged to the journal on systems with systemd.

Outside of ChromeOS the `machine-id` is used for when users optionally choose to send reports about updates to LVFS.

We also need to think about how the client would handle a missing machine ID.

djcampello:
> I think my personal preference would be that rather than this error go away entirely, it be promoted to `g_warning` so it at least is logged to the journal on systems with systemd.

Are you sure this is needed? This adds noise into our logs for *every* execution of fwupdtool:
16:04:27:0627 FuEngine             the machine-id is not present

src/fu-engine.c
```diff
@@ -5254,9 +5254,9 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 /* TODO: Read registry key [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography] "MachineGuid" */
 #ifndef _WIN32
 	/* cache machine ID so we can use it from a sandboxed app */
-	self->host_machine_id = fwupd_build_machine_id ("fwupd", error);
+	self->host_machine_id = fwupd_build_machine_id ("fwupd", NULL);
```
superm1: If it does end up being shown as a warning a better way to do it is to declare a local error like this:

```
g_autoptr(GError) error_machine_id = NULL;
self->host_machine_id =fwupd_build_machine_id ("fwupd", &error_local);
if (self->host_machine_id == NULL)
      g_warning ("%s", error_local->message);
```

hughsie: Is there anything better we can do? e.g. have a "fake" machine ID for chrome-os?

superm1:
> > I think my personal preference would be that rather than this error go away entirely, it be promoted to `g_warning` so it at least is logged to the journal on systems with systemd.
> 
> Are you sure this is needed? This adds noise into our logs for _every_ execution of fwupdtool:
> 16:04:27:0627 FuEngine the machine-id is not present

OK, I guess put it as `g_debug` then.  If we find a need to show this more verbosely later you can carry a downstream patch to `s/g_debug/g_warning` that line.

src/fu-engine.c
superm1: Oh I like that; so maybe add a new path only present on chrome systems that it can check for and set that fake string.

djcampello: I proposed this yesterday to be handled outside fwupd code base and got this: crrev.com/c/2078258/12#message-e151fbe95c08c57c5228904c7d1ddeac151ab7dd

superm1: For some reason that's not showing an actual message.  I don't think it needs to be outside the fwupd code base.  I'm saying in the fwupd codebase we can add an extra thing it falls back on if `/etc/machine-id` is missing at all those above locations that can indicate it's a chrome system and to use a dummy value instead.

Another alternative to do is if nothing is found then instead use one of the existing helpers that parsed `/etc/os-release` and fill in machine-id with the `ID` key from there.

djcampello: Maybe the shortening magic... here again: https://chromium-review.googlesource.com/c/chromiumos/overlays/chromiumos-overlay/+/2078258/12#message-e151fbe95c08c57c5228904c7d1ddeac151ab7dd

This is a sample os-release in my device:
BUILD_ID=12735.0.2019_12_06_1141
NAME=Chrome OS
ID_LIKE=chromiumos
GOOGLE_CRASH_ID=ChromeOS
VERSION_ID=80
BUG_REPORT_URL=https://crbug.com/new
VERSION=80
HOME_URL=https://www.chromium.org/chromium-os
ID=chromeos

ID would just be chromeos... It will probably be meaningless to use this.

hughsie:
Assuming CI passes. LGTM.

superm1:
@djcampello I'll backport this to 1_3_X.
@hughsie could you maybe tag 1.3.9?

hughsie:
> could you maybe tag 1.3.9

Lets do a 1.3.9 on Monday after I've had a chance to test all the backported stuff. It's getting late in the day now.

superm1:
Sounds good.

PR #1825, opened 2020-02-28
Backports to 1_3_X

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
> If CI looks good, LGTM, thanks. Could you also do a new PR for master that fixes up the version of the symbols we've added and not backported? TA!

Sure.  So I didn't backport https://github.com/fwupd/fwupd/commit/f50ff2c27e37a2914703de09a39e1fa5a8296a7b, but given 0c0fada didn't work without it, I have a manual fixup.   Wanted to  check with you though; do you think that one should come too to 1.3.9, or do the fixup?

hughsie:
> do you think that one should come too to 1.3.9,

I don't think that should go into stable. Fixup please!

PR #1827, opened 2020-02-29
Use Jcat files in firmware archives and for metadata

hughsie:
A Jcat file can be used to store GPG, PKCS-7 and SHA-256 checksums for multiple
files. This allows us to sign a firmware or metadata multiple times (perhaps
by the OEM and also then the LVFS) which further decentralizes the trust model
of the LVFS.

The Jcat format was chosen as the Microsoft catalog format is nonfree and not
documented. We also don't want to modify an existing .cat file created from WU
as this may make it unsuitable to use on Windows.

More information can be found here: https://github.com/hughsie/libjcat


hughsie:
I guess we also need to rip out the pkcs7 and gpg deps too. WIP.

superm1:
So this makes jcat a hard dependency immediately.  Maybe that's fine, but I would wonder if that makes sense given lead time to get this kind of thing into distros?

hughsie:
> So this makes jcat a hard dependency immediately

I'm not in any rush to get this merged into master; the lvfs-website generator bit would be good as we can rev/change the file specification with instant effect and then get libjcat into a state where it's good to get a first release and into distros.

src/fu-engine.c
```diff
@@ -3045,32 +3117,33 @@ fu_engine_update_metadata_bytes (FuEngine *self, const gchar *remote_id,
 
 	/* verify file */
 	keyring_kind = fwupd_remote_get_keyring_kind (remote);
-	if (keyring_kind != FWUPD_KEYRING_KIND_NONE) {
-		g_autoptr(FuKeyring) kr = NULL;
-		g_autoptr(FuKeyringResult) kr_result = NULL;
-		g_autoptr(FuKeyringResult) kr_result_old = NULL;
+	if (keyring_kind == FWUPD_KEYRING_KIND_GPG ||
```
superm1: I have to wonder - do we really still need to specify the type of keyring in the remote?  Why not just let the libjcat code look at it and determine from the stream what we have?

libfwupdplugin/fu-cabinet.c
```diff
@@ -71,6 +78,23 @@ fu_cabinet_set_size_max (FuCabinet *self, guint64 size_max)
 	self->size_max = size_max;
 }
 
+/**
+ * fu_cabinet_set_jcat_context: (skip):
+ * @self: A #FuCabinet
+ * @jcat_context: (nullable): A #JcatContext
+ *
+ * Sets the maximum size of the decompressed cabinet file.
```
superm1: comment looks wrong

libfwupdplugin/fu-cabinet.c
```diff
@@ -185,28 +215,64 @@ fu_cabinet_parse_release (FuCabinet *self, XbNode *release, GError **error)
...
+		if (results == NULL) {
+			g_debug ("failed to verify payload %s: %s",
+				 basename, error_local->message);
+		} else {
+			g_debug ("verified payload %s: %u",
+				 basename, results->len);
+			release_flags |= FWUPD_RELEASE_FLAG_TRUSTED_PAYLOAD;
+		}
+
+	/* legacy GPG detached signature */
```
superm1: I guess no extra fallback to pkcs7 is intentional, right?

libfwupdplugin/fu-cabinet.c
```diff
@@ -519,6 +634,22 @@ fu_cabinet_parse (FuCabinet *self,
 	if (!fu_cabinet_decompress (self, data, error))
 		return FALSE;
 
+	/* load the firmware.jcat if included */
```
superm1: Is it always going to be `firmware.jcat`?  Or should it be better to refer to this from one of the XML files and then load that?

meson_options.txt
```diff
@@ -2,13 +2,11 @@ option('build', type : 'combo', choices : ['all', 'standalone', 'library'], valu
 option('agent', type : 'boolean', value : true, description : 'enable the fwupd agent')
 option('consolekit', type : 'boolean', value : true, description : 'enable ConsoleKit support')
 option('firmware-packager', type : 'boolean', value : true, description : 'enable firmware-packager installation')
-option('gpg', type : 'boolean', value : true, description : 'enable the GPG verification support')
 option('gtkdoc', type : 'boolean', value : false, description : 'enable developer documentation')
 option('introspection', type : 'boolean', value : true, description : 'generate GObject Introspection data')
 option('lvfs', type : 'boolean', value : true, description : 'enable LVFS remotes')
 option('man', type : 'boolean', value : true, description : 'enable man pages')
 option('gudev', type : 'boolean', value : true, description : 'enable GUdev support')
-option('pkcs7', type : 'boolean', value : true, description : 'enable the PKCS7 verification support')
```
superm1: So right now there is actually the ability to compile the stack without both GPG and PKCS7 support.  IIRC this is what Google actually does when they compile into ChromeOS.  I think by moving to jcat and making it a hard dep that means it's impossible to decouple the trust verification.

@djcampello would you be able to provide some comments on your thoughts here?  If `-ENOCONTEXT`, go and take a look at https://blogs.gnome.org/hughsie/2020/02/28/introducing-jcat/

hughsie: You can compile `jcat` without GPG an without PKCS, although I'll admit I've not actually tested if fwupd still works in that case. I guess it should do.

superm1: we could certainly add a specific CI unit test for that scenario to make sure nothing blows up when no trust libraries in place.

libfwupdplugin/fu-cabinet.c
hughsie: it's a good question. We only turned the PKCS-7 support on a few weeks ago on the LVFS, and so only perhaps 1% of all the firmware has both signatures. I guess we could fall back to both if you think it's worth it.

src/fu-engine.c
hughsie: That **is** a good question. Something like looking for `PKCS-7` or `GPG` in the first line of the signature?

libfwupdplugin/fu-cabinet.c
superm1: Oh it's only been a few weeks?  Yeah I agree it's probably not worth it.  And it's probably also worth mentioning in the code comment this reason to only fall back to GPG.

Maybe also a sunset timeframe/date to remove the (legacy) code for fallback.

hughsie: >  Yeah I agree it's probably not worth it

The counterpoint here is that I'm working on a patchset that can resign the entire archive, so if we do that we'd get the pkcs "for free" even for old firmware.

> Maybe also a sunset timeframe/date to remove the (legacy) code for fallback.

How many years? Heh. RHEL.

src/fu-engine.c
superm1: Yeah, I don't know for sure what kind of header there is for both if that's enough, but I would think it's entirely possible to tell from the stream what you have.

libfwupdplugin/fu-cabinet.c
hughsie: > from one of the XML files

Which one? It all gets a bit circular if you allow the XML to define the signature filename, and the signature filename verifies the XML... Perhaps we should just load `*.jcat`? I know some vendors rename their files to `firmware.*` but the majority don't.

libfwupdplugin/fu-cabinet.c
superm1: 5?  The thing to keep in mind is metadata should always be getting refreshed.  The whole point of this stack is to be pulling in new things.  So if you're running a remote that generates old metadata formats, you're going to be forced to upgrade it at _some_ point.

libfwupdplugin/fu-cabinet.c
superm1: *.jcat sounds like a good idea.

src/fu-engine.c
```diff
@@ -3937,28 +4049,28 @@ fu_engine_add_approved_firmware (FuEngine *self, const gchar *checksum)
...
 	g_autoptr(GBytes) payload = NULL;
 	g_autoptr(GBytes) signature = NULL;
 
 	/* create detached signature and verify */
-	kr = fu_keyring_create_for_kind (FWUPD_KEYRING_KIND_PKCS7, error);
-	if (kr == NULL)
-		return NULL;
-	if (!fu_keyring_setup (kr, error))
+	jcat_engine = jcat_context_get_engine (self->jcat_context,
+					       FWUPD_KEYRING_KIND_PKCS7,
```
superm1: Isn't this the wrong enum type?  It should be a jcat type.

meson_options.txt
hughsie: Ideally I'd like something that matches what Google are doing with ChromeOS 100%...

src/fu-engine.c
hughsie: Great catch! I deliberately gave them the same values for GPG and PKCS, but this is still a valid bug.

src/fu-keyring-utils.h
```diff
@@ -7,12 +7,12 @@
...
 
-#include "fu-keyring.h"
 #include "fwupd-enums.h"
 
-FuKeyring	*fu_keyring_create_for_kind		(FwupdKeyringKind kind,
-							 GError		**error);
 gboolean	 fu_keyring_get_release_flags		(XbNode		*release,
 							 FwupdReleaseFlags *flags,
 							 GError		**error);
+FwupdKeyringKind fu_keyring_kind_from_blob_kind		(JcatBlobKind	 kind);
```
superm1: I would also question if there still needs to be a concept of a `FwupdKeyring/fu_keyring` anymore.  Instead maybe this should rename to something about getting jcat_release flags?  And returning jcat types?

meson_options.txt
superm1: I don't know how realistic that actually is, they could potentially change `USE` flags from one board to another leading to differently built utilities and stack.  So I think a simple CI test that builds jcat as a subproject to fwupd but sets gpg/pkcs7 to disabled and runs the fwupd battery of tests on it would be what checks this best.

src/fu-keyring-utils.h
hughsie: If we can nuke it from the remote files, I agree. I'll have a play.

src/meson.build
```diff
@@ -141,8 +127,7 @@ fwupdtool = executable(
     'fu-history.c',
     'fu-idle.c',
     'fu-install-task.c',
-    'fu-keyring.c',
-    'fu-keyring-result.c',
+
```
superm1: I think you don't want a newline here, right?  Just delete both of these.

subprojects/libjcat.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+directory = libjcat
+url = https://github.com/hughsie/libjcat.git
+revision = wip/hughsie/subproject
```
superm1: bring this back to master again

meson_options.txt
djcampello: > So right now there is actually the ability to compile the stack without both GPG and PKCS7 support. IIRC this is what Google actually does when they compile into ChromeOS. I think by moving to jcat and making it a hard dep that means it's impossible to decouple the trust verification.
> 
> @djcampello would you be able to provide some comments on your thoughts here? If `-ENOCONTEXT`, go and take a look at https://blogs.gnome.org/hughsie/2020/02/28/introducing-jcat/

Here is our current ebuild:
https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/master/sys-apps/fwupd/fwupd-9999.ebuild

It seems that at least we have IUSE=+gpg

superm1: @djcampello I'm not super familiar with ebuilds, with `IUSE=+gpg` does that mean that by default the ebuild has gpg turned off, and a board would need to explicitly build with `-gpg` to not have it?

Can you confirm then for example that a sarien build ended up with gpg turned on in the end?

libfwupdplugin/fu-cabinet.c
hughsie: Fixed, thanks.

meson_options.txt
djcampello: Actually I was mistaken... Even though is IUSE=+gpg... gpg is not used... Here are the actual USE flags on sarien:

sys-apps/fwupd-9999::chromiumos was built with the following:
USE="minimal nvme -agent -amt -consolekit -cros_host -dell -elogind -gpg -gtk-doc -introspection -man -nls -pkcs7 -redfish -synaptics -systemd -test -thunderbolt -uefi"

None of gpg or pkcs7 are used

superm1: Well I experimented with trying to build a docker container that was basically this:
* Start out with gentoo stage3 amd64
* Pull in portage
* Clone chromiumos-overlay, eclass-overlay
* emerge fwupd-9999

But I kept getting caught up on stuff that was very chromium specific (references to portage-stable, boards etc).  I think if we're going to have a proper CI test that builds it exactly "like ChromeOS does" to catch this stuff early we'll need you to help put it together.

For now I pushed something into this branch (`contrib/ci/trust.sh`) that will build with at least the flags I gathered should be matching what you shared above with pkcs7 and gpg turned off.

libfwupdplugin/fu-cabinet.c
hughsie: Fixed.

src/fu-engine.c
hughsie: I guess we certainly still need to say whether the remote needs to be signed or not, still, right?

superm1: Yeah

hughsie: Also, depending on the answer to this next question, we could get rid of a lot of code: "_how much do we care about non-LVFS generated signed remotes_"? I believe `local.py` generates unsigned remotes, and I don't know of anyone that's deployed a private LVFS instance internally.

superm1: Maybe it's worth putting a call out on social media to check? My gut is that you're right though.

src/fu-keyring-utils.h
hughsie: Played; removed.

libfwupd/fwupd-remote.c
```diff
@@ -388,6 +388,11 @@ fwupd_remote_load_from_filename (FwupdRemote *self,
 				     keyring_kind);
 			return FALSE;
 		}
+		if (priv->keyring_kind == FWUPD_KEYRING_KIND_GPG ||
+		    priv->keyring_kind == FWUPD_KEYRING_KIND_PKCS7) {
+			g_debug ("converting Keyring value to Jcat");
```
superm1: Maybe it's worth mentioning this as a warning that there is an old conf file in use.

hughsie: I don't think I can; the LVFS has to generate `Keyring=gpg` for remotes so it supports all fwupd versions.

superm1: But you don't have to look at that one; you could instead try to find jcat ones specifically and ignore others.

hughsie: Not sure I follow, sorry.

superm1: Whoops.  I was looking at the wrong function here, my mistake.

src/fu-engine.c
hughsie: Quite a few people using complete LVFS mirrors, a a smattering of people using local.py it seems.

libfwupd/fwupd-remote.c
```diff
@@ -153,6 +153,8 @@ fwupd_remote_get_suffix_for_keyring_kind (FwupdKeyringKind keyring_kind)
 		return ".asc";
 	if (keyring_kind == FWUPD_KEYRING_KIND_PKCS7)
 		return ".p7b";
+	if (keyring_kind == FWUPD_KEYRING_KIND_JCAT)
+		return ".jcat";
 	return NULL;
 }
```
superm1: This unction isn't used anymore.  Maybe deprecate it now?

hughsie: Ahh, it's static. Nuked!

superm1:
@hughsie I rebased this to master if you want to merge it now.

hughsie:
I think I'll use the official fedora package (rather than the copr) and then agreed.

PR #1830, opened 2020-03-03
Improved fish shell completion

exploide:
Just a small addition to the fish shell completion script I previously contributed. When completing device IDs, it now shows the device name as a description. This makes it easier to select the correct hex value.

deviceid-completion

Though, it is of limited use when most of the devices are just "UEFI Device Firmware".

By the way, @superm1 I think your bash completion broke when output format for `get-devices` and `get-remotes` changed a while back. Completing device and remote IDs in bash is currently not working.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


superm1:
Thanks for the heads up, yeah it looks like stuff like `enable-remote` doesn't work properly anymore in bash completion.  It's too bad `jq` isn't in most distros by default.  I would say that is a better way to parse fwupdagent stuff so this doesn't break in future.

PR #1831, opened 2020-03-03
Fix bash completion

superm1:
Thanks @exploide for the notification it was broken.  Resolve `get-devices` and `get-remotes` output to work properly.  Hopefully by slicing and dicing `fwupdagent` output the `get-devices` is less likely to break in the future.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdmgr.in
+++ b/data/bash-completion/fwupdmgr.in
@@ -66,7 +66,7 @@ _show_device_ids()
 	local description
 	OLDIFS=$IFS
 	IFS=$'\n'
-	description="$(command fwupdmgr get-devices | command awk '!/DeviceId/ { line = $0 }; /DeviceId/ { print $2 " {" line "}"}')"
+	description="$(command fwupdagent get-devices 2>/dev/null | sed -e 's,"Name" :,,; s,",,g; s,\,,,g; s,[[:space:]]\+,,' | command awk '!/DeviceId/ { line = $0 }; /DeviceId/ { print $3 " { "line" }"}')"
 	COMPREPLY+=( $(compgen -W "${description}" -- "$cur") )
 	IFS=$OLDIFS
 }
@@ -74,7 +74,7 @@ _show_device_ids()
 _show_remotes()
 {
 	local remotes
-	remotes="$(command fwupdmgr get-remotes | command awk '/Remote ID/ { print $3 }')"
+	remotes="$(command fwupdmgr get-remotes | command awk '/Remote ID/ { print $4 }')"
 	COMPREPLY+=( $(compgen -W "${remotes}" -- "$cur") )
 }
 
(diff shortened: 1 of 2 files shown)


hughsie:
Why do we get the extra arguments? Seems a bit of a hack...

superm1:
> Why do we get the extra arguments? Seems a bit of a hack...

It's intentional - so you don't just get a list of `DeviceID`s and need to remember which one goes to which device.

Right now it's like this (paraphrased):
```
# fwupdgr install file.cab <tab>
432c5bc1-9a04-4cd8-aa20-d02f39410d4c { System Firmware }
5ebbaa00-2f49-4bf6-8dd9-66d1e753c314 { UEFI Device Firmware }
```

So if you don't print what they are, how will you know what to pick?

hughsie:
So can we handle the '{' argument detection in the values[] parsing rather than in each cmd?

superm1:
> So can we handle the '{' argument detection in the values[] parsing rather than in each cmd?

Sure, that should be fine.

superm1:
> > So can we handle the '{' argument detection in the values[] parsing rather than in each cmd?
> 
> Sure, that should be fine.

Although it doesn't land in the same point in the array for both `install` and `activate` commands, so maybe not.

hughsie:
Sure, but we could just set the first '{' to NULL, right?

superm1:
> Sure, but we could just set the first '{' to NULL, right?

Ah yeah, good point. Thanks.

exploide:
Are you sure that `command fwupdagent get-devices` will work out of the box? I get `bash: fwupdagent: command not found` because `fwupdagent` is in `/usr/libexec/fwupd/fwupdagent` which is not in `$PATH` by default.

exploide:
To reply to your comment from #1830

> It's too bad jq isn't in most distros by default. I would say that is a better way to parse fwupdagent stuff so this doesn't break in future.

True. But I have seen other completion scripts using python for processing JSON. Checking first if `python` is available and then doing something like `/usr/libexec/fwupd/fwupdagent get-devices | python -S -c 'import json, sys; data = json.load(sys.stdin); print("\n".join(d["DeviceId"] for d in data["Devices"]))'`. The assumption here is that Python is available on most Linux installations.

But your approach might be robust enough and relying on a external program also has its downsides...

superm1:
> Are you sure that `command fwupdagent get-devices` will work out of the box? I get `bash: fwupdagent: command not found` because `fwupdagent` is in `/usr/libexec/fwupd/fwupdagent` which is not in `$PATH` by default.

It changed in https://github.com/fwupd/fwupd/commit/bf6d4556877573641f63d1b957986d6927f045b5

> usr/libexec/fwupd/fwupdagent get-devices | python -S -c 'import json, sys; data = json.load(sys.stdin); print("\n".join(d["DeviceId"] for d in data["Devices"]))'

Lets see if my approach breaks anywhere.  If it does, something like what you proposed will work, but we need to print the `{ }` and `Name` attributes too.

superm1:
> I fixed up a tiny leak, hope that's okay. If you squash them down (to 1 commit?) then LGTM, thanks.

Thanks!  I squashed it all to 1.

PR #1837, opened 2020-03-05
ci: build Debian using libflashrom

superm1:
The package is now in unstable in the 1.2-4 package.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

I expect it will still fail in Ubuntu, but once that package syncs to the archive should be in better shape.


superm1:
FYI I also filed https://bugs.archlinux.org/task/65707 so we can probably flip the default pretty soon.

superm1:
> No approval required from me, but CI has to pass obviously :)

Right, looks like still a problem, uploading 1.2-5 which will hopefully solve the remaining issue.

PR #1839, opened 2020-03-05
trivial: case insensitive while explorering cabinet files

CragW:
The cabinet file's extension name disregards uppercase or lowercase

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2649,9 +2649,10 @@ fu_engine_create_metadata (FuEngine *self, XbBuilder *builder,
 		g_autoptr(XbBuilderSource) source = NULL;
 		g_autoptr(GError) error_local = NULL;
 		const gchar *fn = g_ptr_array_index (files, i);
+		g_autofree gchar *fn_lowercase = g_ascii_strdown (fn, -1);
 
 		/* check is cab file */
-		if (!g_str_has_suffix (fn, ".cab")) {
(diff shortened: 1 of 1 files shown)


src/fu-engine.c
```diff
@@ -2651,7 +2651,7 @@ fu_engine_create_metadata (FuEngine *self, XbBuilder *builder,
 		const gchar *fn = g_ptr_array_index (files, i);
 
 		/* check is cab file */
-		if (!g_str_has_suffix (fn, ".cab")) {
+		if (!g_str_has_suffix (g_ascii_strdown(fn, -1), ".cab")) {
```
hughsie: `g_ascii_strdown` returns allocated memory. Would this be suitable instead?

    if (!g_str_has_suffix (fn, ".cab") &&
        !g_str_has_suffix (fn, ".CAB")) {

superm1: That of course wouldn't cover Cab or cAB though too.  I think if those are important to cover you'll need to do something above like

`g_autofree gchar * fn_lowercase = g_ascii_strdown(fn, -1);`

and then use `fn_lowercase` in this check.

CragW: Thanks for comments, I've made the changes.

superm1:
Cherry picked back to 1_3_X and 1_2_X too.

PR #1842, opened 2020-03-05
Add hexadecimal version format

superm1:
CPU microcode is distributed with a hexadecimal version.

Without this, UEFI device firmware for microcode shows with master:
```
Current version:     200
```

Various formats to try to render it:
```
•	Plain: 200
•	BCD: 0.0.0.128
•	Quad:  0.0.0.128
```

With this PR (and a quirk to set `VersionFormat` in place) renders as
```
Current version:     0x000000c8
```

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -519,6 +519,8 @@ fwupd_version_format_from_string (const gchar *str)
 		return FWUPD_VERSION_FORMAT_SURFACE;
 	if (g_strcmp0 (str, "dell-bios") == 0)
 		return FWUPD_VERSION_FORMAT_DELL_BIOS;
+	if (g_strcmp0 (str, "hex") == 0)
+		return FWUPD_VERSION_FORMAT_HEX;
 	return FWUPD_VERSION_FORMAT_UNKNOWN;
 }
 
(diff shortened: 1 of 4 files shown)


libfwupd/fwupd-enums.c
```diff
@@ -519,6 +519,8 @@ fwupd_version_format_from_string (const gchar *str)
 		return FWUPD_VERSION_FORMAT_SURFACE;
 	if (g_strcmp0 (str, "dell-bios") == 0)
 		return FWUPD_VERSION_FORMAT_DELL_BIOS;
+	if (g_strcmp0 (str, "hexadecimal") == 0)
+		return FWUPD_VERSION_FORMAT_HEXADECIMAL;
```
hughsie: I think I'd prefer `hex` and `_HEX` given we have things like `quad` already.

libfwupdplugin/fu-common-version.c
```diff
@@ -143,6 +143,10 @@ fu_common_version_from_uint32 (guint32 val, FwupdVersionFormat kind)
 					(val >> 8) & 0xff,
 					val & 0xff);
 	}
+	if (kind == FWUPD_VERSION_FORMAT_HEXADECIMAL) {
+		/* 0xAABBCCDD */
```
hughsie: We need this in the other `fu_common_version_from_uintXX` functions too, right?

hughsie:
> CPU microcode

?! Pray tell..

I think we also need the corresponding LVFS `hex` verfmt too, although I'm guessing you want to get this in first.

superm1:
> ?! Pray tell..

Probably the only device that I think hex version makes sense for that I've seen recently.

> I think we also need the corresponding LVFS hex verfmt too, although I'm guessing you want to get this in first.

Yeah, I wanted to make sure you were amenable to this version format before going down the lvfs side of things.

PR #1846, opened 2020-03-06
Don't build/install fwupdagent man page if agent build is not requested

Polynomial-C:
Otherwise build fails with:

  src/meson.build:196:2: ERROR: Unknown variable "fwupdagent".

Gentoo-bug: https://bugs.gentoo.org/711682
Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -193,20 +193,22 @@ if build_daemon and get_option('man')
     install : true,
     install_dir : join_paths(mandir, 'man1'),
   )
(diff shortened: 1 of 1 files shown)


superm1:
Oh you know this also reminds me that bash-completion depends on the agent...

PR #1847, opened 2020-03-06
Add a read-only CPU microcode plugin

superm1:
Type of pull request:
- [x] New plugin (Please include [new plugin checklist](https://github.com/hughsie/fwupd/wiki/New-plugin-checklist))
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/cpu/README.md
+++ b/plugins/cpu/README.md
@@ -1,0 +1,9 @@
+CPU Microcode Support
+=================
+
+Introduction
+------------
+
+This plugin reads the sysfs attributes associated with CPU microcode.
+It displays a read-only value of the CPU microcode version loaded onto
+the physical CPU at fwupd startup.
--- a/plugins/cpu/fu-cpu-device.c
+++ b/plugins/cpu/fu-cpu-device.c
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2019 Mario Limonciello <mario.limonciello@dell.com>
+ *
+ * SPDX-License-Identifier: GPL-2+
+ */
+
+#include "config.h"
+
+#include "fu-cpu-device.h"
+
+struct _FuCpuDevice {
+	FuDevice		 parent_instance;
+};
+
+G_DEFINE_TYPE (FuCpuDevice, fu_cpu_device, FU_TYPE_DEVICE)
+
+static void fu_cpu_device_parse_section (FuDevice *dev, const gchar *data)
+{
+	g_auto(GStrv) lines = NULL;
+
+	lines = g_strsplit (data, "\n", 0);
+	for (guint i = 0; lines[i] != NULL; i++) {
+		if (g_str_has_prefix (lines[i], "vendor_id")) {
+			g_auto(GStrv) fields = g_strsplit (lines[i], ":", -1);
+			if (fields[1] != NULL)
+				fu_device_set_vendor (dev, g_strchug (fields[1]));
+		} else if (g_str_has_prefix (lines[i], "model name")) {
+			g_auto(GStrv) fields = g_strsplit (lines[i], ":", -1);
+			if (fields[1] != NULL)
+				fu_device_set_name (dev, g_strchug (fields[1]));
+		} else if (g_str_has_prefix (lines[i], "microcode")) {
+			g_auto(GStrv) fields = g_strsplit (lines[i], ":", -1);
+			if (fields[1] != NULL)
+				fu_device_set_version (dev, g_strchug (fields[1]));
+		} else if (g_str_has_prefix (lines[i], "physical id")) {
+			g_auto(GStrv) fields = g_strsplit (lines[i], ":", -1);
+			if (fields[1] != NULL) {
+				g_autofree gchar *tmp = g_strdup_printf ("cpu:%s", g_strchug (fields[1]));
+				fu_device_set_physical_id (dev, tmp);
+			}
+		}
+	}
+}
+
+static void
+fu_cpu_device_init (FuCpuDevice *self)
+{
+	fu_device_add_guid (FU_DEVICE (self), "cpu");
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_INTERNAL);
+	fu_device_add_icon (FU_DEVICE (self), "computer");
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_HEX);
+}
+
+static void
+fu_cpu_device_class_init (FuCpuDeviceClass *klass)
+{
+}
+
+FuCpuDevice *
+fu_cpu_device_new (const gchar *section)
+{
+	FuCpuDevice *device = NULL;
+	device = g_object_new (FU_TYPE_CPU_DEVICE, NULL);
+	fu_cpu_device_parse_section (FU_DEVICE (device), section);
+	return device;
+}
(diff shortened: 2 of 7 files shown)


plugins/cpu/README.md
```diff
@@ -0,0 +1,9 @@
+CPU Microcde Support
```
hughsie: Microcode

plugins/cpu/README.md
```diff
@@ -0,0 +1,9 @@
...
+=================
+
+Introduction
+------------
+
+This plugin reads the sysfs attributes associated with CPU microcode.
+It displays a read-only value of the CPU microcode version loaded onto
+the physical CPU at fwupd startup.
```
hughsie: maybe a stupid question; but does the microcode have to be the same version on all cores and for all CPUs on the system?

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -0,0 +1,100 @@
...
+	if (!g_file_get_contents ("/proc/cpuinfo", &data, &length, error))
+		return FALSE;
+
+	lines = g_strsplit (data, "\n", 0);
+	for (guint i = 0; lines[i] != NULL; i++) {
+		if (g_str_has_prefix (lines[i], "processor")) {
+			g_auto(GStrv) fields = g_strsplit (lines[i], ":", -1);
+			if (fields[1] != NULL) {
+				if (processor != NULL)
+					g_free (processor);
```
hughsie: are they going to be different? the "last" one wins? I think you could have `if (processor == NULL && g_str_has_prefix (lines[i], "processor")) {` and cut out a bit of code.

plugins/cpu/README.md
superm1: For the physical CPU package yes.  That's why I'm only making the device when physical ID changes.

plugins/cpu/fu-plugin-cpu.c
superm1: Well you can't tell for sure how many CPUs are on the system in one pass of the for loop (and how many are going to be identical).  So I'm hoping this code will work for multi CPU package systems (make a device for each one) as well as multi CPU core systems (make one device).

plugins/meson.build
```diff
@@ -1,3 +1,4 @@
+subdir('cpu')
```
hughsie: maybe `cpu-microcode`? I'm not sure. Maybe `cpu` is best?

superm1: I went back and forth and flipped a coin.  Whichever you prefer.

hughsie: Leave it as it, no worries.

plugins/cpu/fu-plugin-cpu.c
superm1: Although I guess I could reverse the loop from bottom of `/proc/cpuinfo` to get physical ID first.  If that didn't change don't bother reading all the other fields.

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -0,0 +1,100 @@
...
+	g_autofree gchar *processor = NULL;
+	g_autofree gchar *model_name = NULL;
+	g_autofree gchar *microcode = NULL;
+	g_autofree gchar *vendor = NULL;
+	guint64 physical_id = G_MAXUINT64;
+
+	if (!g_file_get_contents ("/proc/cpuinfo", &data, &length, error))
+		return FALSE;
+
+	lines = g_strsplit (data, "\n", 0);
```
hughsie: the purist in me would say split on `\n\n` and then pass each chunk into a fu_cpu_device_new(const gchar *section) where it gets parsed there. Then when you get the FuCpuDevice back you look at the physical ID and if it's the same as the `physical_id` then it gets unrefd and ignored. I don't think it's any more fragmentation compared to what you have here.

plugins/cpu/fu-cpu-device.c
```diff
@@ -0,0 +1,79 @@
...
+#include "fu-cpu-device.h"
+
+struct _FuCpuDevice {
+	FuDevice		 parent_instance;
+};
+
+G_DEFINE_TYPE (FuCpuDevice, fu_cpu_device, FU_TYPE_DEVICE)
+
+static
+void fu_cpu_device_parse_section (FuDevice *dev, const gchar *data)
```
hughsie: `static void` on same line pls

plugins/cpu/fu-cpu-device.c
```diff
@@ -0,0 +1,79 @@
...
+				g_strchug (fields[1]);
+				fu_device_set_vendor (dev, fields[1]);
+			}
+			continue;
+		}
+		if (g_str_has_prefix (lines[i], "model name")) {
+			g_auto(GStrv) fields = g_strsplit (lines[i], ":", -1);
+			if (fields[1] != NULL) {
+				g_strchug (fields[1]);
+				fu_device_set_name (dev, fields[1]);
```
hughsie: If it helps, I think you can do `fu_device_set_name (dev, g_strchug (fields[1]));`

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -0,0 +1,42 @@
...
+	g_autofree gchar *data = NULL;
+	g_auto(GStrv) lines = NULL;
+
+	if (!g_file_get_contents ("/proc/cpuinfo", &data, &length, error))
+		return FALSE;
+
+	lines = g_strsplit (data, "\n\n", 0);
+	for (guint i = 0; lines[i] != NULL; i++) {
+		g_autoptr(FuCpuDevice) dev = NULL;
+		g_strstrip(lines[i]);
```
hughsie: is this needed?

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -0,0 +1,42 @@
...
+		if (strlen (lines[i]) == 0)
+			continue;
+		dev = fu_cpu_device_new (lines[i]);
```
hughsie: I still think you need to keep track of the last physical ID -- otherwise I'm going to get 64 devices deduped by the daemon, no?

plugins/cpu/fu-plugin-cpu.c
superm1: Yeah; I was seeing an extra device being created with no data.  This was my way to avoid it.  might be a cleaner way though.

plugins/cpu/fu-plugin-cpu.c
superm1: Isn't that the point of daemon de-duping?  If you think it should be in plugin though, can do it there too again.

hughsie: True. Adding 63 extra devices to the daemon isnt free, but I wonder if it's just premature optimization on my part. How noisy is is?

superm1: With debug logs turned on we're looking at about 5 extra debugging lines per CPU and I guess between 0 and 1/1000 of a second time per CPU?  I don't know the unit after seconds in the debug logs.

plugins/cpu/fu-plugin-cpu.c
superm1: Not needed, will remove.

plugins/cpu/fu-plugin-cpu.c
superm1: Fixed up the other two now.  I guess try the branch on your side and see if you think it's too much.

PR #1852, opened 2020-03-09
Fix `get-details` and `install` for VersionFormat changes

superm1:
When these values are coming from the metadata in the CAB file, it's important to apply them to the checks used for these two commands.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-install-task.c
+++ b/src/fu-install-task.c
@@ -334,12 +334,14 @@ fu_install_task_check_requirements (FuInstallTask *self,
 	}
 
 	/* check the version formats match if set in the release */
-	verfmts = xb_node_query (self->component,
-				 "custom/value[@key='LVFS::VersionFormat']",
-				 0, NULL);
-	if (verfmts != NULL) {
-		if (!fu_install_task_check_verfmt (self, verfmts, flags, error))
-			return FALSE;
+	if ((flags & FWUPD_INSTALL_FLAG_FORCE) == 0) {
+		verfmts = xb_node_query (self->component,
+					"custom/value[@key='LVFS::VersionFormat']",
+					0, NULL);
+		if (verfmts != NULL) {
+			if (!fu_install_task_check_verfmt (self, verfmts, flags, error))
+				return FALSE;
+		}
 	}
 
 	/* compare to the lowest supported version, if it exists */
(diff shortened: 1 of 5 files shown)


src/fu-tool.c
```diff
@@ -878,10 +878,23 @@ fu_util_install (FuUtilPrivate *priv, gchar **values, GError **error)
 
 			/* is this component valid for the device */
 			task = fu_install_task_new (device, component);
+			if (!fu_engine_check_requirements (priv->engine,
+							   task, priv->flags | FWUPD_INSTALL_FLAG_IGNORE_VERFMT,
```
hughsie: would FORCE work here as well?

src/fu-install-task.c
```diff
@@ -334,12 +334,14 @@ fu_install_task_check_requirements (FuInstallTask *self,
 	}
 
 	/* check the version formats match if set in the release */
-	verfmts = xb_node_query (self->component,
-				 "custom/value[@key='LVFS::VersionFormat']",
-				 0, NULL);
-	if (verfmts != NULL) {
-		if (!fu_install_task_check_verfmt (self, verfmts, flags, error))
-			return FALSE;
+	if ((flags & FWUPD_INSTALL_FLAG_IGNORE_VERFMT) == 0) {
```
hughsie: Is this also something we want to allow using FORCE?

libfwupd/fwupd-enums.h
```diff
@@ -174,6 +174,7 @@ typedef guint64 FwupdReleaseFlags;
  * @FWUPD_INSTALL_FLAG_ALLOW_OLDER:		Allow downgrading firmware
  * @FWUPD_INSTALL_FLAG_FORCE:			Force the update even if not a good idea
  * @FWUPD_INSTALL_FLAG_NO_HISTORY:		Do not write to the history database
+ * @FWUPD_INSTALL_FLAG_IGNORE_VERFMT:	Ignore the version format when checking requirements
```
hughsie: indent, if we need this at all....

libfwupd/fwupd-client.c
```diff
@@ -1045,6 +1045,10 @@ fwupd_client_install (FwupdClient *client,
 		g_variant_builder_add (&builder, "{sv}",
 				       "no-history", g_variant_new_boolean (TRUE));
 	}
+	if (install_flags & FWUPD_INSTALL_FLAG_IGNORE_VERFMT) {
```
hughsie: i think this is internal use only, and no need to parse.

src/fu-tool.c
superm1: Yes it does, adjusted to that much better thanks.

PR #1854, opened 2020-03-09
trivial: apply the version format to release in `get-details` too

superm1:
Just caught one more instance it was off.

Otherwise the output looks funny when you have the proper version
on the device, but release looks like a big number.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


src/fu-engine.c
```diff
@@ -2792,7 +2792,16 @@ fu_engine_md_refresh_device_verfmt (FuEngine *self, FuDevice *device, XbNode *co
 	/* found and different to existing */
 	if (verfmt != FWUPD_VERSION_FORMAT_UNKNOWN &&
 	    fu_device_get_version_format (device) != verfmt) {
+		FwupdRelease *rel = fu_device_get_release_default (device);
```
hughsie: do we only ever have one release?

src/fu-engine.c
```diff
@@ -2792,7 +2792,16 @@ fu_engine_md_refresh_device_verfmt (FuEngine *self, FuDevice *device, XbNode *co
...
 		fu_device_set_version_format (device, verfmt);
+		if (rel != NULL) {
+			const gchar *version_old = fwupd_release_get_version (rel);
+			guint64 tmp = fu_common_strtoull (version_old);
+			if (tmp > 0) {
+				g_autofree gchar *version_new = fu_common_version_from_uint32 (tmp, verfmt);
```
hughsie: not convinced here. Shouldn't we use the same logic we use for the device?

superm1: Well releases don't have a `VersionFormat` attribute or a raw version right now.  So If you think we should go that far down, it's a little more plumbing.

src/fu-engine.c
superm1: In a CAB file, I suppose it's technically possible that there are more than one release, and this could also be used by `fwupdmgr get-releases` I would guess.  So yes it probably needs to recognize all the releases then (or this needs to be plumbed deeper into the release itself not during the metadata refresh).

src/fu-engine.c
hughsie: I do wonder if that's too far down the rabbithole. Can't we just use the same verfmt as the device in all cases?

src/fu-engine.c
superm1: applied to all releases

src/fu-engine.c
superm1: I did an experiment with setting it immediately when the releases were loaded in `fu_engine_set_release_from_appstream` but realized that this will possibly lead to inconsistent results for devices that don't actually have the `md-set-*` flags set.  So my thought is it's better to refresh the release versions when metadata changes instead.

If you think this is still not the right way, would you propose another approach?

superm1:
@hughsie take a look at new approach.  I think this scales better.

hughsie:
> take a look at new approach

That's exactly what I meant. LGTM, thanks.

PR #1855, opened 2020-03-10
Wip/hughsie/fu device get root

hughsie:


libfwupdplugin/fu-device.c
```diff
@@ -456,6 +456,31 @@ fu_device_get_parent (FuDevice *self)
 	return priv->parent;
 }
 
+/**
+ * fu_device_get_root:
```
superm1: get root makes me think this is returning the parent of a parent (like the very top chain), but it's just returning the parent of the device.

hughsie: It really is getting the parent of the parent.

plugins/vli/fu-vli-usbhub-i2c-device.c
```diff
@@ -246,6 +247,16 @@ fu_vli_usbhub_i2c_device_write_firmware (FuDevice *device,
 		fu_device_set_progress_full (device, (gsize) j, (gsize) records->len);
 	}
 
+	/* the device automatically reboots */
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
+	fu_device_set_progress (device, 0);
+
+	/* this is unusual, but the MSP device reboot takes down the entire hub
+	 * for ~60 seconds and we don't want the parent device removing us */
+	root = fu_device_get_root (device);
```
superm1: do you maybe want an assertion that root != self at the top of this function?

hughsie: Not sure. I don;t think it's ever possible.

superm1: If parent wasn't set somehow, that seems like a bad programming error, which is why I thought an assertion made sense.

hughsie:
Self tests FTW.

Issue #1794, opened 2020-02-20
UEFI updates fail to install on Ubuntu 20.04


alexmurray:
After scheduling a firmware upgrade for 2 devices, they never get installed - and journalctl for fwupd shows an error about missing or invalid capsule header - see below for full details (I mentioned this in passing in https://github.com/fwupd/fwupd/issues/1772#issuecomment-586909620 but figured I should file a separate bug about it).

```shell
$ fwupdmgr get-updates
No upgrades for Thunderbolt Controller: current version is 43.00: 43.00=same
No upgrades for UEFI Device Firmware: current version is 184.70.3626: 184.70.3626=same, 184.65.3590=older, 184.60.3561=older, 184.55.3510=older
20KHCTO1WW
│
├─System Firmware:
│ │   Device ID:           1c53551e7da69d896138fac1ae131c83ad46d923
│ │   Current version:     0.1.44
│ │   Minimum Version:     0.0.1
│ │   Vendor:              LENOVO (DMI:LENOVO)
│ │   Update State:        success
│ │   GUID:                a4b51dca-8f97-4310-8821-3330f83c9135
│ │   Device Flags:        • Internal device
│ │                        • Updatable
│ │                        • Requires AC power
│ │                        • Supported on remote server
│ │                        • Needs a reboot after installation
│ │                        • Cryptographic hash verification is available
│ │                        • Device is usable for the duration of the update
│ │ 
│ └─ThinkPad X1 Carbon 6th System Update:
│       New version:       0.1.45
│       Remote ID:         lvfs
│       Summary:           Lenovo ThinkPad X1 Carbon 6th System Firmware
│       License:           Proprietary
│       Size:              9.5 MB
│       Vendor:            Lenovo Ltd.
│       Flags:             is-upgrade
│       Description:       Lenovo ThinkPad X1 Carbon 6th System Firmware Version 1.45
│       
│       The computer will be restarted automatically after updating BIOS completely. Do NOT turn off your computer or remove the AC adaptor while update is in progress.
│       
│       This stable release fixes the following issues:
│       
│        • Fixed an issue where Thunderbolt controller device might disappear from
│       
│       the Device Manager at 2nd boot.
│       
│       (Note) ECP version 1.20 or later is required for complete solution.
│     
└─UEFI Device Firmware:
  │   Device ID:           9698faabddf0d7b18925cfbbda95f8b0d0dacc53
  │   Current version:     0.1.19
  │   Minimum Version:     0.1.19
[...]

PR #1867, opened 2020-03-15
uefi: Don't use shim for non-secure boot configurations

superm1:
Since the entry is dynamically created this removes one area that
can potentially cause failures.
Helps: #1794

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
I thought you were on PTO?

superm1:
I thought so too. But my original plans got cancelled outside my control. TBD the rest of the week.

PR #1868, opened 2020-03-16
Fix a difficult-to-trigger daemon hang when replugging devices

hughsie:
Use only one GMainLoop in FuDeviceList; we can only iterate one loop at a time
anyway, and having the mainloops per-item complicates the lifecycle of the
fu_device_list_wait_for_replug() functionality considerably.

### DIFF
--- a/src/fu-device-list.c
+++ b/src/fu-device-list.c
@@ -38,6 +38,8 @@ struct _FuDeviceList
 	GObject			 parent_instance;
 	GPtrArray		*devices;	/* of FuDeviceItem */
 	GRWLock			 devices_mutex;
+	GMainLoop		*replug_loop;	/* block waiting for replug */
+	guint			 replug_id;	/* timeout the loop */
 };
 
 enum {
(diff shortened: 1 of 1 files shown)


superm1:
Does this change the timing of device replug waits? I would think if anything it might increase it a little, but that's probably okay. I don't think any devices so far have needed to have time critical replug.

hughsie:
> Does this change the timing of device replug waits

I don't think so; did I miss something? I appreciate the sanity check.

superm1:
Well maybe timing isn't a problem - but changing this way could affect if two devices are in replug at same time. You might quit the replug loop prematurely if both devices didn't come back in time.

hughsie:
>two devices are in replug at same time

Yes, and this is the real reason for the cleanup. Blocking with a new GMainLoop isn't great, but for a single threaded daemon like fwupd it's the right thing to do. What you can't do is wait on a 2nd main loop for the 1st main loop -- and having the loop specific to the device meant that could happen if for instance we didn't use the same Item for the replugged device in bootloader mode.

My gut instinct is KISS, where we say that we wait for any of the devices that came back, on the logic that if we're setting WAIT_FOR_REPLUG on more than one device with the code now we're probably in for a Bad Time (tm) wrt. a daemon hang.

superm1:
Well what do you think about dropping the flag for any other devices in replug still, and showing a warning?

hughsie:
That's a good idea. Give me a couple of hours.

hughsie:
> Give me a couple of hours

18 hours later... :)

PR #1875, opened 2020-03-17
logitech: Correctly set the protocol

superm1:
Fixes updates on signed receivers.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-runtime.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-runtime.c
@@ -229,6 +229,9 @@ fu_logitech_hidpp_runtime_setup_internal (FuDevice *device, GError **error)
 		if ((self->version_bl_major == 0x01 && config[8] >= 0x04) ||
 		    (self->version_bl_major == 0x03 && config[8] >= 0x02)) {
 			self->signed_firmware = TRUE;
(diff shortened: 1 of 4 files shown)


plugins/logitech-hidpp/fu-logitech-hidpp-runtime.c
```diff
@@ -229,6 +229,9 @@ fu_logitech_hidpp_runtime_setup_internal (FuDevice *device, GError **error)
 		if ((self->version_bl_major == 0x01 && config[8] >= 0x04) ||
 		    (self->version_bl_major == 0x03 && config[8] >= 0x02)) {
 			self->signed_firmware = TRUE;
+			fu_device_set_protocol (device, "com.logitech.unifyingsigned");
+		} else {
+			fu_device_set_protocol (device, "com.logitech.unifying");
```
FFY00: With those changes, `set_protocol` can be removed from `fu_logitech_hidpp_peripheral_init` and `fu_logitech_hidpp_peripheral_setup`, right?

superm1: I don't think so.. those are different devices from the quirks with no relation.  Peripherals should still be treated separately.

PR #1880, opened 2020-03-18
ccgx: Add a firmare parser for cyacd files

hughsie:
These are visually similar to Intel hex files, but different enough to demand
their own parser. Multiple images can be stored in one firmware file, with the
`addr` set to the SiliconID and the `idx` set to the position in the file.


hughsie:
Thanks for the speedy review. I've added the metadata parsing too, can you cast your eye over that too please.

superm1:
Yep nothing else stands  out in the diff, looks great.

PR #1883, opened 2020-03-19
ccgx: Create devices when in HID and I2C modes

hughsie:


plugins/ccgx/fu-ccgx-hid-device.c
```diff
@@ -0,0 +1,52 @@
...
+	return TRUE;
+}
+
+static void
+fu_ccgx_hid_device_init (FuCcgxHidDevice *self)
+{
+	fu_device_set_protocol (FU_DEVICE (self), "com.cypress.ccgx");
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
+//	fu_device_add_flag (FU_DEVICE (self), FWUPD_I2C_DEVICE_FLAG_UPDATABLE);
+	fu_hid_device_set_interface (FU_HID_DEVICE (self), 0x01);
```
superm1: That's a little unusual; no?  Is this right?

hughsie: Unusual https://github.com/fwupd/fwupd/pull/1879/files#diff-91b79cd74a72843f7ec3c454bfd29b9aR27 -- but not unheard of https://github.com/fwupd/fwupd/blob/master/plugins/synaptics-cxaudio/fu-synaptics-cxaudio-common.h#L15

PR #1887, opened 2020-03-20
Add support for EP963x hardware

hughsie:
Draft as I don't have the hardware to test this yet...


plugins/ep963x/fu-ep963x-device.c
```diff
@@ -0,0 +1,347 @@
...
+		bufhw[0] = FU_EP963_FEATURE_ID1_SIZE;
+		bufhw[1] = cmd;
+		if (!fu_hid_device_set_report (FU_HID_DEVICE (self), 0x00,
+					       bufhw, sizeof(bufhw),
+					       FU_EP963_DEVICE_TIMEOUT,
+					       FU_HID_DEVICE_FLAG_IS_FEATURE,
+					       error)) {
+			return FALSE;
+		}
+		g_usleep (100 * 1000);
```
superm1: If all commands are going to have this settle/sleep after, maybe split to a separate function

plugins/ep963x/fu-ep963x-device.c
```diff
@@ -0,0 +1,347 @@
...
+					 buf, sizeof(buf),
+					 error)) {
+		return FALSE;
+	}
+	version = g_strndup ((const gchar *) buf, 5);
+	fu_device_set_version (device, version);
+
+	/* the VID and PID are unchanged between bootloader modes */
+	if (g_strcmp0 (version, "00000") == 0) {
+		fu_device_set_flags (device, FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
```
superm1: I would say you should add instead of set to prevent debugging an error later

hughsie: Doh! good catch!

plugins/ep963x/fu-ep963x-device.c
hughsie: Not sure I follow, can you elaborate a bit pls.

superm1: It seems with set commands you're sleeping each time after for a settle time period?

hughsie: I think I know what you mean, and I think it uncovered a missing sleep, so nice one there. Fixed.

superm1: Yeah you adjusted it how I was thinking.  Glad you could make sense of what I meant :)

hughsie:
I'm in two minds about merging this before testing it on actual hardware. Hardware is on the way, but it could be weeks, and I don't want this to bitrot. Opinion?

superm1:
> I'm in two minds about merging this before testing it on actual hardware. Hardware is on the way, but it could be weeks, and I don't want this to bitrot. Opinion?

I don't have any opposition.  It's pretty straightforward. And if you find some problems and we have a release before they're fixed you can always make sure that the CAB files have a minimum version set to 1.4.1.

Issue #1890, opened 2020-03-21
AIAIAI H05 invalid starting token: CSR-dfu2


Salt-Factory:
**Describe the bug**
When trying to update my [AIAIA H05 headband](https://fwupd.org/lvfs/devices/com.AIAIAI.H05.firmware) to 1.6, I get the following error:

`invalid starting token on line 1: CSR-dfu2`

Right now, the headband is unable to turn on.
I get the same error when trying to reinstall 1.5.
Am I missing the CSR plugin?

**Steps to Reproduce**
I plugged in the headband, set the headband in update mode, and ran `fwupdmgr update`.

**Expected behavior**
I would have expected the headband to update.

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version

client version:	1.3.9
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.3.9
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):

fwupd was installed via the Arch repository, using pacman.

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
$ fwupdmgr get-devices --show-all-devices

Z97X-UD5H
│
├─88SE9172 SATA 6Gb/s Controller:
│     Device ID:           0fba4a8bedba53076fc9b147655e59455bc25a87
│     Current version:     12
│     Vendor:              Marvell Technology Group Ltd. (PCI:0x1B4B)
│     GUIDs:               4a090d9b-362e-580e-b0bc-47003768f3e6
│                          76a209d6-e1dc-598a-b9e9-a8dd911da340
│     Device Flags:        • Internal device
│                          • Cryptographic hash verification is available
│   
├─BLUETOOTH USB +EDR ADAPTER v2.1 UHE:
│     Device ID:           5d00620651c7c80da4d93a876b5d3a6138b0da7a
│     Current version:     3.88
│     Vendor:              Broadcom Corp (USB:0x050D)
│     GUIDs:               ef2af201-4ca1-5b3d-b36a-c00e8aec666b
│                          ba1e3114-39b3-57a0-b44a-3b1b36cfcf95
│     Device Flags:        • Updatable
│   
├─GM204 [GeForce GTX 970]:
│     Device ID:           71b677ca0f1bc2c5b804fa1d59e52064ce589293
│     Current version:     a1
│     Vendor:              NVIDIA Corporation (PCI:0x10DE)
[...]

PR #1891, opened 2020-03-22
csr: Parse the firmware as a DFU file

hughsie:
This was changed to Intel hex in 7afd7cba0d327b155b2201c9fbb5414ced9cfe06,
probbaly due to a copy-paste mistake. Change it back to DFU.

Fixes https://github.com/fwupd/fwupd/issues/1890


superm1:
Cherry picked it to 1_3_X too.

PR #1905, opened 2020-03-23
trivial: Autodetect the HID interface to use

hughsie:


libfwupdplugin/fu-hid-device.c
```diff
@@ -81,6 +81,29 @@ fu_hid_device_open (FuUsbDevice *device, GError **error)
...
 
+	/* auto-detect */
+	if (priv->interface == 0xff) {
+		g_autoptr(GPtrArray) ifaces = NULL;
+		ifaces = g_usb_device_get_interfaces (usb_device, error);
+		if (ifaces == NULL)
+			return FALSE;
+		for (guint i = 0; i < ifaces->len; i++) {
+			GUsbInterface *iface = g_ptr_array_index (ifaces, i);
+			if (g_usb_interface_get_class (iface) == 0x03) {
```
superm1: What is 0x03? Probably should have a define somewhere?

FFY00: It's the USB bus it.

@hughsie you should probably also add bluetooth here, no?

hughsie: > Probably should have a define somewhere

Fixed.

> you should probably also add bluetooth here

No, bluetooth isn't a device class, it's a transport.

PR #1907, opened 2020-03-24
ccgx: Set up the HPI I²C config at startup

hughsie:


hughsie:
@superm1 CI seems to be failing with the debian build stuck at:

    debconf: unable to initialize frontend: Dialog
    debconf: (TERM is not set, so the dialog frontend is not usable.)
    debconf: falling back to frontend: Readline
    Configuring tzdata
    ------------------

    Please select the geographic area in which you live. Subsequent configuration
    questions will narrow this down by presenting a list of cities, representing
    the time zones in which they are located.

      1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
      2. America     5. Arctic     8. Europe    11. SystemV
      3. Antarctica  6. Asia       9. Indian    12. US
    Geographic area:

superm1:
@hughsie pushed something to master that should fix it

plugins/ccgx/fu-ccgx-hid-device.c
```diff
@@ -29,13 +29,30 @@ fu_ccgx_hid_device_detach (FuDevice *device, GError **error)
...
 	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
 	return TRUE;
 }
 
+static gboolean
+fu_ccgx_hid_device_setup (FuDevice *device, GError **error)
+{
+	/* This seems insane... but we need to switch the device from HID
+	 * mode to HPI mode at startup. The device continues to function
+	 * exactly as before and no user-visible effects are noted */
```
superm1: Is it basically enabling the I2C communication to the device?  If so, I think that this should be un-done after enumerating version and such.  Otherwise it's leaving a comms hole for other processes to abuse the end point more easily with timing no?

plugins/ccgx/fu-ccgx-hpi-common.h
```diff
@@ -0,0 +1,344 @@
...
+	HPI_EVENT_SOURCE_DISABLED,
+	HPI_EVENT_SENDER_TIMEOUT,
+	HPI_EVENT_VDM_NO_RESPONSE,
+	HPI_EVENT_UNEXPECTED_VOLTAGE,
+	HPI_EVENT_ERROR_RECOVERY,
+	HPI_EVENT_EMCA_DETECT			= 0xA6,
+	HPI_EVENT_RP_CHANGE_DETECT		= 0xAA,
+	HPI_EVENT_TB_ENTERED			= 0xB0,
+	HPI_EVENT_TB_EXITED
+} HPIResp;
```
superm1: With the big gaps in documented events, I would question whether this being an enum is even valuable.  Should it just be a bunch of #defines for the ones documented?

plugins/ccgx/fu-ccgx-hpi-device.c
```diff
@@ -84,12 +612,151 @@ fu_ccgx_hpi_write_firmware (FuDevice *device,
...
+	self->hpi_addrsz = mode & 0x80 ? 2 : 1;
+	self->num_ports = (mode >> 2) & 0x03 ? 2 : 1;
+	self->fw_mode = (FWMode) (mode & 0x03);
+
+	/* add extra instance ID */
+	instance_id = g_strdup_printf ("USB\\VID_%04X&PID_%04X&MODE_%s",
+				       fu_usb_device_get_vid (FU_USB_DEVICE (device)),
+				       fu_usb_device_get_pid (FU_USB_DEVICE (device)),
+				       fu_ccgx_fw_mode_to_string (self->fw_mode));
+	fu_device_add_instance_id (device, instance_id);
```
superm1: Realistically do you see these 3 IDs as the ones being used to update FW and not the other ones?  I wonder if the "generic" IDs should actually be nuked in such case.  We have precedent in doing that in NVME with the Dell GUID's too.

hughsie: It's a good question. I'm not 100% sure how we're going to do the matching yet, I added this instance ID just for the custom UpdateMessage string. On this case, I think still having the other "generic" VID/PID ones are useful, as we can also do the CounterpartGuid stuff where we don't care which firmware is running.

plugins/ccgx/fu-ccgx-hpi-common.h
hughsie: I was torn about this; on one hand it's very verbose, on the other hand you can diff from the Windows sources too.

plugins/ccgx/fu-ccgx-hid-device.c
hughsie: Apparently there's no drawback to leaving it in HPI mode. We still need to be root to use the HPI device, and I suppose it's even more secure than in HID mode which is open by default on most OS's. In the case of one of the Lenovo hubs the HID mode is kinda useless as the only thing you can use it for is to switch the device into HPI mode....

plugins/ccgx/fu-ccgx-hpi-common.h
superm1: Ah if it's like this in the Windows sources and more easily diffable for changes, it's better to keep it this way I'd say.  New events that are otherwise accidentally forgotten to add will stand out much more easily.

PR #1908, opened 2020-03-24
trivial: Export fu_device_add_instance_id_full() for plugins to use

hughsie:
Sometimes we only want to add the instance ID to get the quirk matches, and it
is confusing to see the "fake" IDs in the 'fwupdmgr get-devices' output.

### DIFF
--- a/libfwupdplugin/fu-device-private.h
+++ b/libfwupdplugin/fu-device-private.h
@@ -11,20 +11,6 @@
 
 #define fu_device_set_plugin(d,v)		fwupd_device_set_plugin(FWUPD_DEVICE(d),v)
 
-/**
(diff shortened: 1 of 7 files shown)


superm1:
I think there are some other ones like this to hide too actually then. Perhaps  `system-tpm` on the TPM2.0 device?

hughsie:
I actually think that's a different flag. We're not using it for quirk matching, we're just kinda using it like a "parent identifier". Better name welcome. Phase 2 I think.

superm1:
> I actually think that's a different flag. We're not using it for quirk matching, we're just kinda using it like a "parent identifier". Better name welcome. Phase 2 I think.

Oh true.  OK.

PR #1910, opened 2020-03-24
Adjust special thunderbolt GUID for multi controller systems

superm1:
Having seen the system with both controllers having same VID/DID in use, i'm not convinced that PCI slot name will be safe enough to be considered constant.  So instead rely upon order of enumeration.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Well apparently this breaks though across an upgrade, `data->multi_thunderbolt_host` will increment again and `*-host1` goes to `*-host2`.

PR #1911, opened 2020-03-25
Wip/hughsie/ccgx

hughsie:


hughsie:
I've tested this, and seems to work well.

plugins/ccgx/fu-ccgx-hpi-device.c
```diff
@@ -605,11 +1068,60 @@ fu_ccgx_hpi_write_firmware (FuDevice *device,
...
+
+	/* validate fw */
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_VERIFY);
+	if (!fu_ccgx_hpi_validate_fw (self, fw_mode_alt, error)) {
+		g_prefix_error (error, "fw validate error: ");
+		return FALSE;
+	}
+
+	/* this is a good time to leave the flash mode *before* rebooting */
+	g_clear_object (&locker);
```
superm1: why not directly call `fu_ccgx_hpi_leave_flash_mode` so you can do error handling in case of a failure here?

plugins/ccgx/fu-ccgx-hpi-device.c
```diff
@@ -562,41 +907,159 @@ fu_ccgx_hpi_device_prepare_firmware (FuDevice *device,
...
+	g_autofree guint8 *buf = NULL;
+
+	/* read flash at correct address */
+	if (!fu_ccgx_hpi_get_metadata_offset (self, fw_mode, &addr, &md_offset, error))
+		return FALSE;
+	buf = g_malloc0 (self->flash_row_size);
+	if (!fu_ccgx_hpi_read_flash (self, addr,
+				     buf, self->flash_row_size,
+				     error)) {
+		g_prefix_error (error, "fw metadata read error: ");
```
superm1: I think you should use a unique prefix to help with debugging in the future.

plugins/ccgx/fu-ccgx-hpi-device.c
```diff
@@ -562,41 +907,159 @@ fu_ccgx_hpi_device_prepare_firmware (FuDevice *device,
...
+				 FWMode fw_mode,
+				 guint32 *addr,
+				 guint32 *offset,
+				 GError **error)
+{
+	guint32 addr_max;
+
+	/* sanity check */
+	if (self->flash_row_size == 0x0) {
+		g_set_error (error,
```
superm1: seems this should be a literal error

plugins/ccgx/fu-ccgx-hpi-device.c
```diff
@@ -562,41 +907,159 @@ fu_ccgx_hpi_device_prepare_firmware (FuDevice *device,
...
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "unset support row size: 0x%x",
+			     self->flash_row_size);
+		return FALSE;
+	}
```
superm1: do you perhaps want another check comparing flash row size and flash size?

plugins/ccgx/fu-ccgx-hpi-device.c
```diff
@@ -306,6 +374,9 @@ fu_ccgx_hpi_device_reg_read (FuCcgxHpiDevice *self,
 		g_prefix_error (error, "read error: ");
 		return FALSE;
 	}
+
+	/* 5 ms */
+	g_usleep (5000);
```
superm1: Why is this one different than `HPI_CMD_FLASH_READ_WRITE_DELAY_US`

plugins/ccgx/README.md
```diff
@@ -11,6 +11,44 @@ Supported devices:
  * Lenovo Gen2 Dock
  * Lenovo Hybrid Dock
 
+Device Flash
+============
+
+There are three kinds of flash layout. Single image firmware is not currently
+supported in this plugin.
```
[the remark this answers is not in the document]
hughsie: Anything that has `GType = FuCcgxH**pi**Device`, yes. Maybe this should be a boolean quirk, something like `is-symmetric`? Or just have a proper setter like `ImageType=symmetric`?

plugins/ccgx/fu-ccgx-hpi-device.c
hughsie: The idea of doing it as a `FuDeviceLocker` was that we would disable the flash mode on all the error paths too. We could certainly do as you suggest, but we'd need to add something like `gboolean fu_device_locker_close(FuDeviceLocker *self, GError **error)` to not _close_ things twice. I'm happy to do that if you think it's a good idea. Better names welcome too.

plugins/ccgx/fu-ccgx-hpi-device.c
hughsie: Fixed.

plugins/ccgx/fu-ccgx-hpi-device.c
hughsie: Fixed.

plugins/ccgx/README.md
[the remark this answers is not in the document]
superm1: Either way sounds fine to me.

plugins/ccgx/fu-ccgx-hpi-device.c
superm1: Yes, I think that an explicit close makes more sense.  That name seems fine to me.

plugins/ccgx/fu-ccgx-hpi-device.c
hughsie: Good idea, added.

plugins/ccgx/fu-ccgx-hpi-device.c
hughsie: Yes, it's all a bit muddled. I've moved them up into the read/write functions to be super explicit.

plugins/ccgx/fu-ccgx-hpi-device.c
hughsie: https://github.com/fwupd/fwupd/pull/1916

plugins/ccgx/README.md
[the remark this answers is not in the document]
hughsie: Okay, fixed.

Issue #1877, opened 2020-03-18
Logitech peripherals not showing up in fwupd


superm1:
**Describe the bug**
I noticed with a new K830 keyboard hooked up through a unifying dongle that the keyboard is not actually enumerated by fwupd.

It is however enumerated by solaar.

**Steps to Reproduce**
`# fwupdtool get-devices --plugin-whitelist=logitech_hidpp`

**fwupd version information**
c0f1f1c7e3adbef4346807528b7b9d515118b059

I dug around a little bit, and see it's broken at least as far back as d9f52c75738939cc9a6edcf268cc139904523c98 which from commit history seems like the time it should have been working.

I think what is happening is that the quirks are tagging the GType for the peripheral as a `FuLogitechHidPpRuntime` since it happens to have the same VEN/DEV as the unifying receiver it's connected through.   This seems to follow with the device being added "twice" too.

```
04:46:08:0499 FuEngine             using plugins: logitech_hidpp
04:46:08:0500 FuEngine             failed to find specified plugin dell: no plugin dell found
04:46:08:0502 FuEngine             failed to find specified plugin dell_dock: no plugin dell_dock found
04:46:08:0503 FuEngine             failed to find specified plugin dell_dock: no plugin dell_dock found
04:46:08:0519 FuEngine             6 devices with subsystem hidraw
04:46:08:0520 FuEngine             failed to find specified plugin synaptics_rmi: no plugin synaptics_rmi found
04:46:08:0521 FuDevice             using 279ed287-3607-549e-bacc-f873bb9838c4 for HIDRAW\VEN_046D&DEV_C52B
04:46:08:0522 XbSilo               0.02ms      query on / with `components/component/provides/firmware[@type='flashed'][text()='d25c014a-6dea-58cf-ac0f-762daf8be6ab']` limit=1 -> 0 results
04:46:08:0522 XbSilo               0.02ms      query on / with `components/component/provides/firmware[@type='flashed'][text()='82b088b2-dc64-5a93-abb3-d9f2cfa14fc0']` limit=1 -> 0 results
04:46:08:0522 XbSilo               0.02ms      query on / with `components/component/provides/firmware[@type='flashed'][text()='63cf2859-c965-5a77-9353-b1047125d7d9']` limit=1 -> 0 results
[...]

PR #1917, opened 2020-03-26
logitech_hidpp: show devices for all peripherals not just supported ones

superm1:
At least to discuss.  This fixes all my Logitech peripherals not showing up.  I finally now understand the experience for stuff not currently connected too and why the DJ changes are important.

@FFY00 I think that the plugin will need to respond to change events to update version string (both connect and disconnect most likely).

Fixes: #1877

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/daemon.conf
+++ b/data/daemon.conf
@@ -25,3 +25,6 @@ VerboseDomains=
 
 # Update the message of the day (MOTD) on device and metadata changes
 UpdateMotd=true
+
+# For some plugins, enumerate only devices supported by metadata
+EnumerateAllDevices=true
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -501,6 +501,7 @@ fu_engine_modify_config (FuEngine *self, const gchar *key, const gchar *value, G
 		"IdleTimeout",
 		"VerboseDomains",
 		"UpdateMotd",
+		"EnumerateAllDevices",
 		NULL };
 
 	g_return_val_if_fail (FU_IS_ENGINE (self), FALSE);
@@ -4880,6 +4881,10 @@ fu_engine_plugin_check_supported_cb (FuPlugin *plugin, const gchar *guid, FuEngi
 {
 	g_autoptr(XbNode) n = NULL;
 	g_autofree gchar *xpath = NULL;
+
+	if (fu_config_get_enumerate_all_devices (self->config))
+		return TRUE;
+
 	xpath = g_strdup_printf ("components/component/"
 				 "provides/firmware[@type='flashed'][text()='%s']",
 				 guid);
--- a/plugins/dfu/dfu.quirk
+++ b/plugins/dfu/dfu.quirk
@@ -12,11 +12,6 @@ CounterpartGuid = USB\VID_0BDA&PID_5800
 [DeviceInstanceId=USB\VID_0BDA&PID_5800]
 Flags = detach-for-attach
 
-# on PC platforms the DW1820A firmware is loaded at runtime and can't
-# be stored on the device itself as the flash chip is unpopulated
-[DeviceInstanceId=USB\VID_0A5C&PID_6412]
-Flags = only-supported
-
 # Openmoko Freerunner / GTA02
 [DeviceInstanceId=USB\VID_1D50&PID_5119]
 Plugin = dfu
--- a/src/fu-config.c
+++ b/src/fu-config.c
@@ -56,6 +56,7 @@ fu_config_reload (FuConfig *self, GError **error)
 	g_auto(GStrv) plugins = NULL;
 	g_autofree gchar *domains = NULL;
 	g_autoptr(GKeyFile) keyfile = g_key_file_new ();
+	g_autoptr(GError) error_update_motd = NULL;
 	g_autoptr(GError) error_enumerate_all = NULL;
 
 	g_debug ("loading config values from %s", self->config_file);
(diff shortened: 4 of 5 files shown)


plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
```diff
@@ -1021,7 +1021,4 @@ fu_logitech_hidpp_peripheral_init (FuLogitechHidPpPeripheral *self)
 	fu_device_set_protocol (FU_DEVICE (self), "com.logitech.unifying");
 	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_PLAIN);
 
-	/* there are a lot of unifying peripherals, but not all respond
-	 * well to opening -- so limit to ones with issued updates */
```
hughsie: isn't this still valid? Some (early?) hidpp devices do not cope well being probed.

FFY00: I am not aware about any devices that do this, but I haven't dealt with that many early devices.

FFY00:
> @FFY00 I think that the plugin will need to respond to change events to update version string (both connect and disconnect most likely).

We can, but we don't know the DJ device ID. Since there's max of 6 connected devices I think it's sane to ping everyone when we receive a connection ID.

The best solution still is listening to udev change events once my patch gets upstreamed. Listening for DJ events can, however, be an alternative for older kernels.

plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
superm1: I don't know, I was hoping to spur discussion to find out what devices are like this.  If it's just a handful, maybe we should just blacklist them.

What does solaar do for them?

FFY00: I think we ping everyone.

@hughsie do you remember which devices do this?

superm1:
> > @FFY00 I think that the plugin will need to respond to change events to update version string (both connect and disconnect most likely).
> 
> We can, but we don't know the DJ device ID. Since there's max of 6 connected devices I think it's sane to ping everyone when we receive a connection ID.
> 
> The best solution still is listening to udev change events once my patch gets upstreamed. Listening for DJ events can, however, be an alternative for older kernels.

And maybe just cache the last seen version in case one of them is disconnected?

FFY00:
> And maybe just cache the last seen version in case one of them is disconnected?

Yes.

plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
hughsie: I believe it was on advice from Marc/Nestor. I'd be fine switching to a blacklist approach if required.

FFY00: I think that's best. I think it's safe to say the big majority of devices don't have this issue. We can still query Logitech about this.

superm1: OK if we merge this PR, i'm pretty sure logitech plugin was the only one using `FWUPD_DEVICE_FLAG_ONLY_SUPPORTED`.  I think we should leave it in the daemon for a release or two in case we gotta go back and then consider to tear it out later if not.

hughsie: Is this the kind of thing we want to put in a config file? e.g. a `logitech-hidpp.conf` with a `OnlySupported = false`? It does make me worry slightly about long term (read: difficult to update) distros like RHEL.

superm1: That sounds good to me.  I'll add it.

hughsie:
Per daemon? I thought it would be per-plugin?

superm1:
> Per daemon? I thought it would be per-plugin?

I was going to do it per-plugin originally, but the plumbing was feeling a lot more hacky.  So I went daemon instead.

hughsie:
>  So I went daemon instead

What are the big-picture repercussions of enabling it globally?

data/daemon.conf
```diff
@@ -25,3 +25,6 @@ VerboseDomains=
 
 # Update the message of the day (MOTD) on device and metadata changes
 UpdateMotd=true
+
+# For some plugins, enumerate only devices supported by metadata
+OnlySupported=false
```
hughsie: This makes it sound like we only show supported devices, when I really think it needs to convey "allow plugins to filter by supported" as only very few use the FWUPD_DEVICE_FLAG_ONLY_SUPPORTED filter. Maybe `AllowOnlySupported` or even something out there like `EnumerateAllDevices=true` which seems more semantically valid. Other ideas welcome.

src/fu-engine.c
```diff
@@ -4880,6 +4881,10 @@ fu_engine_plugin_check_supported_cb (FuPlugin *plugin, const gchar *guid, FuEngi
 {
 	g_autoptr(XbNode) n = NULL;
 	g_autofree gchar *xpath = NULL;
+
+	if (!fu_config_get_only_supported (self->config))
```
hughsie: I think this breaks the "blacklisting" of https://github.com/fwupd/fwupd/blob/master/plugins/dfu/dfu.quirk#L15 too -- although you could argue the dfu quirk is kinda abusing the feature.

superm1: Well that particular one it doesn't actually cause a failure to enumerate.  I think we could just as easily drop that quirk.  It's not like anyone will be able to do any damage to it.

data/daemon.conf
superm1: I think of those my preference is `EnumerateAllDevices`.  I can adjust if you think that's good.

superm1:
> > So I went daemon instead
> 
> What are the big-picture repercussions of enabling it globally?

Since only one plugin is using it, it's a little hard to envision.  I think that when we end up using it for another plugin we can certainly revisit at that point if the default is appropriate.

data/daemon.conf
hughsie: EnumerateAllDevices wfm.

src/fu-engine.c
hughsie: Fair point.

PR #1919, opened 2020-03-27
Add FuDeviceRetry helper object

hughsie:
Sometimes plugins need to retry various commands send to hardware, either due
to unreliable transfers (e.g. using USB bulk) or from slightly quirky hardware.

Between them they seem to get various things wrong; either the error messages
are repeated and thus difficult to parse, or they just get the memory handling
of `g_propagate_prefixed_error()` wrong.

Providing a helper object we can reduce the amount of boilerplate. Additionally
we can support a 'reset' function that can try to automatically recover the
hardware for specific error domains and codes.

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2018 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2015-2020 Richard Hughes <richard@hughsie.com>
  *
  * SPDX-License-Identifier: LGPL-2.1+
  */
@@ -57,7 +57,14 @@ typedef struct {
 	gint				 open_refcount;	/* atomic */
 	GType				 specialized_gtype;
 	GPtrArray			*possible_plugins;
+	GPtrArray			*retry_recs;	/* of FuDeviceRetryRecovery */
 } FuDevicePrivate;
+
+typedef struct {
+	GQuark				 domain;
+	gint				 code;
+	FuDeviceRetryFunc		 recovery_func;
+} FuDeviceRetryRecovery;
 
 enum {
 	PROP_0,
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-device-retry.c
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/**
+ * fu_device_retry_new:
+ * @device: A #FuDevice
+ * @count: The number of retries to attempt
+ *
+ * Creates a new retry object.
+ *
+ * No reference is taken on @device, and so it must remain valid for the
```
superm1: Why not take the reference?  It seems like it would help avoid forgetting that this acts funny.

hughsie: I originally had this, but in the common case the subclassed FuDevice (e.g. FuCcgxHpiDevice) would put the FuDeviceRetry in the private data. This means the device refcount is 2. If the device goes away and we unref it, we drop the refcount to 1, but that doesn't finalize the subclassed object as there's already a "self" reference to itself in the FuDeviceRetry. This means we effectively leak the object.

Of course, the other thing to do would be to jam all the FuDeviceRetry functionality into FuDevice itself... but I didn't know if we wanted to blow up an already large object. I guess it's only really a guint and gpointer. Should I try that instead? It certainly makes all the refcount stuff easier.

I think I've convinced myself merging is indeed the right thing to do...

hughsie:
> there is at least one example in tree

There's probably half a dozen :( Let's get the API settled in then convert them.

libfwupdplugin/fu-device-retry.c
superm1: Look across the tree and see what stuff uses retries.  Is it only USB?
Maybe it makes sense to let it live directly in the `FuUSbDevice` if it's only used there right now.  And if it needs to grow across buses add into FuDevice later.

hughsie: I want to use it in FuSynapticsMstDevice too, so I think FuDevice is the right place. Patch coming up....

hughsie: Okay, new much simpler version below.

libfwupdplugin/fu-device.c
```diff
@@ -168,6 +175,115 @@ fu_device_add_possible_plugin (FuDevice *self, const gchar *plugin)
 	g_ptr_array_add (priv->possible_plugins, g_strdup (plugin));
 }
 
+/**
+ * fu_device_retry_add_condition:
```
superm1: I think you should rename this to `add_reset_func`

superm1: Or adjust documentation for the other one to reference calling the conditions not reset_func

hughsie: It's not specifically a reset tho, it's just a way to recover the current action. How about `fu_device_retry_add_recovery` or something?

superm1: Sure, that's better than condition.

PR #1920, opened 2020-03-27
ccgx: Reset the i2c bridge on status failure

hughsie:


hughsie:
@superm1 this is how I see plugins being converted...

hughsie:
@superm1 the other two look sane too?

plugins/nitrokey/fu-nitrokey-device.c
```diff
@@ -13,26 +13,31 @@
 
 G_DEFINE_TYPE (FuNitrokeyDevice, fu_nitrokey_device, FU_TYPE_USB_DEVICE)
 
+typedef struct {
+	guint8		 command;
+	const guint8	*buf_in;
+	gsize		 buf_in_sz;
+	guint8		*buf_out;
+	gsize		 buf_out_sz;
+	GCancellable	*cancellable;
```
superm1: I'm not sure plumbing cancellable all the way through makes sense.  It's `NULL` at it's earliest pass in.  Maybe just drop it as part of this?

superm1:
Left a note about nitrokey.
VLI seems fine (although I will note that's a lot of sleep to go _before_ and _after_ which was already the case).

plugins/nitrokey/fu-nitrokey-device.c
hughsie: Fixed, thanks. I also noticed it was another SET_REPORT/GET_REPORT conversion, so I'll do that as a follow up.

hughsie:
@superm1 what do you think of the last commit `fu_dell_dock_hub_recovery_impossible_cb`? I did ponder:

 * Allowing `fu_device_retry_add_recovery` to have a NULL `func` to signify that recovery is not possible
* Automatically adding `G_USB_DEVICE_ERROR_NO_DEVICE` in `FuUsbDevice` automatically for all devices.

Better ideas?

superm1:
> @superm1 what do you think of the last commit `fu_dell_dock_hub_recovery_impossible_cb`? I did ponder:
> 
> * Allowing `fu_device_retry_add_recovery` to have a NULL `func` to signify that recovery is not possible

I like this idea.

> * Automatically adding `G_USB_DEVICE_ERROR_NO_DEVICE` in `FuUsbDevice` automatically for all devices.
> 
> Better ideas?

A simple wait callback like dell dock I wonder if we should have a callback error function for all plugins to use.  I would suspect this is going to be a common occurrence.

hughsie:
> A simple wait callback like dell dock

How would we choose the timeout? Or something like `fu_device_retry_set_recovery_delay(FuDevice *d, guint timeout_ms)` that is used for all loops in `fu_device_retry()`

hughsie:
> fu_device_retry_set_delay

This actually simplifies things quite a bit. Added, and rebased.

PR #1921, opened 2020-03-27
Modernize the thunderbolt plugin

superm1:
WIP, but I'm hoping this will be ready for 1.4.0.  If the controller GUID thing looks good, I probably want that in 1_3_X too.

There are various TODO left, so any thoughts on those welcome.  The biggest TODO left in my mind is force power handling.

Any other things that should be done to modernize also welcome.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


superm1:
CC @YehezkelShB

superm1:
(Oh and yes I know self tests are busted, need to figure that out still too..)

hughsie:
> Modernize the thunderbolt plugin

This has been on my "really ought to fix" list for a long while, thanks for finally sorting it. I think you and @gicmo understand this stuff much better than me anyway.

superm1:
Sure.

Do you think you might be able to help with conversion to use fufirmware? I think you understand that part better at least.

hughsie:
> Do you think you might be able to help with conversion to use fufirmware

I think I understand what's going on now, I'll work in the firmware parser on Monday.

hughsie:
Current status:

    sudo ./src/fwupdtool firmware-parse TR_*0x09E1.bin thunderbolt
    FuThunderboltFirmware:
    IsHost:                  true
    IsNative:                true
    DeviceId:                0x15e7
    VendorId:                0xd4
    ModelId:                 0x9e1
    Generation:              0x3
    Ports:                   0x1
    HasPd:                   true
    Section0:                0x4000
    Section1:                0x4210
    Section2:                0x4610
    Section3:                0x271f0
      FuFirmwareImage:
      Data:                  0x66000

plugins/thunderbolt/fu-thunderbolt-firmware.c
```diff
@@ -734,54 +749,59 @@ fu_thunderbolt_image_validate (GBytes  *controller_fw,
...
 	 * good enough validation.
 	 */
-	if (hw_info->id == 0)
-		return UNKNOWN_DEVICE;
+	if (hw_info->id == 0) {
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_NOT_SUPPORTED,
+				     "no supported");
+		return FALSE;
```
YehezkelShB: So we decided we don't want to support newer devices if the host support hasn't added yet?
At least let's remove the comment that says we do want to support it.

superm1:
> Just another comment (besides the one I spotted in the code): I see you dropped native and safe-mode. This means we drop support for old (i.e. just couple of years old) HW? Is this considered normal for fwupd?

Can you point out in the code where you notice this should be?  That wasn't the intention.  It was just to drop it from the reporting metadata to LVFS.

superm1:
> BTW, the last things I knew, the plan was to continue producing add-in cards, which means safe-mode keeps being relevant. I'm sure your knowledge is much more up-to-date, so I assume there are no new add-in cards since the integration to the CPU (and no add-in cards planned for USB4 too)?

I don't believe I can publicly talk about unreleased future stuff.  But I agree we should keep supporting all features that were previously present no matter what is happening for the future.

YehezkelShB:
> > Just another comment (besides the one I spotted in the code): I see you dropped native and safe-mode. This means we drop support for old (i.e. just couple of years old) HW? Is this considered normal for fwupd?
> 
> Can you point out in the code where you notice this should be? That wasn't the intention. It was just to drop it from the reporting metadata to LVFS.

Sorry, I misunderstood your commit message.
If by "reporting metadata to LVFS" you mean the GUID generation (to select the correct FW image file) still takes "native" flag into consideration, it's just the statistic data collection that drops it, I'm totally fine with it.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -0,0 +1,645 @@
...
+	}
+}
+
+static void
+fu_thunderbolt_device_to_string (FuDevice *device, guint idt, GString *str)
+{
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE (device);
+	fu_common_string_append_kb (str, idt, "Host Controller", self->host);
+	fu_common_string_append_kb (str, idt, "Safe Mode", self->safe_mode);
+}
```
superm1: Since the FW is being parsed at daemon startup now we may as well also add the following:
* "Is Native"
* Generation

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -0,0 +1,645 @@
...
+		return NULL;
+
+	/* get current NVMEM */
+	nvmem = fu_thunderbolt_device_find_nvmem (self, TRUE, error);
+	if (nvmem == NULL)
+		return NULL;
+	controller_fw = g_file_load_bytes (nvmem, NULL, NULL, error);
+	if (!fu_firmware_parse (FU_FIRMWARE (firmware_old), controller_fw, flags, error))
+		return NULL;
+	if (fu_thunderbolt_firmware_is_host (firmware) !=
```
superm1: Since there is a way to tell if it's host by reading the FW image itself now - maybe we should use that at startup rather than the examining of the udev parent?  This seems safer.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -0,0 +1,645 @@
...
+	    fu_thunderbolt_firmware_get_device_id (firmware_old)) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INVALID_FILE,
+			     "incorrect device type, got 0x%04x, expected 0x%04x",
+			     fu_thunderbolt_firmware_get_device_id (firmware),
+			     fu_thunderbolt_firmware_get_device_id (firmware_old));
+		return NULL;
+	}
+	if ((flags & FWUPD_INSTALL_FLAG_FORCE) == 0) {
```
superm1: I would say we may as well just move this to the top of the function and let force override anything, not just the model ID check?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -0,0 +1,645 @@
...
+			     FWUPD_ERROR_INTERNAL,
+			     "update failed (status %" G_GINT64_MODIFIER "x)", status);
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_write_data (FuThunderboltDevice	*self,
```
superm1: This function I wonder if we should replace the code with FuChunks instead

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -0,0 +1,645 @@
...
+		if (domain_id != NULL)
+			fu_device_add_instance_id (device, domain_id);
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_plugin_thunderbolt_trigger_update (FuThunderboltDevice	*self,
```
superm1: Something I'm considering here is if we should instead kick this into a delayed activation (or maybe make that behavior optional).

If we did that then potentially then thunderbolt devices could set a flag that they are usable during an update and once we reboot/shutdown apply the update.

The downside is unplugging a Thunderbolt or USB4 device without rebooting will cause firmware to not be updated.  So it is definitely a trade off.

plugins/thunderbolt/fu-thunderbolt-firmware.c
```diff
@@ -0,0 +1,517 @@
...
+							  &pd_pointer,
+							  error)) {
+			g_prefix_error (error, "failed to read pd-pointer: ");
+			return FALSE;
+		}
+		self->has_pd = fu_thunderbolt_firmware_valid_pd_pointer (pd_pointer);
+	}
+
+	if (!self->is_host) {
+		/* TODO: is this to be multiplied by a multiplier, e.g. 0x10000? */
```
superm1: @YehezkelShB do you know?

plugins/thunderbolt/fu-thunderbolt-firmware.c
```diff
@@ -0,0 +1,517 @@
...
+	return TRUE;
+}
+
+/* assumes sections[_SECTION_DIGITAL].offset is already set */
+static gboolean
+fu_thunderbolt_firmware_read_sections (FuThunderboltFirmware *self, GError **error)
+{
+	guint32 offset;
+
+	if (self->gen >= 3 || self->gen == 0) {
```
superm1: I don't know that we should check > 3 unless this can be checked on a real Intel USB4 payload.

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: I don't think we can, it's not writing in blocks of a set size, it's just writing until write() returns and then advancing a counter with the number of bytes we wrote.

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: do we ever want to allow installing a host firmware on a device? I'd argue we just let the user blow their feet off... Maybe VID and PID makes sense to ignore tho?

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: I think if we can apply it live we can -- rebooting is a real pain the neck for most users.

plugins/thunderbolt/fu-thunderbolt-firmware.c
hughsie: ¯\_(ツ)_/¯ I just copied it from the old source code. I don't really understand what it's doing...

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: The only circumstance that this was needed I know of was a mid-development model ID string change.  Maybe you're right - we shouldn't let people shoot themselves in the foot until we know another real need for it.

plugins/thunderbolt/fu-thunderbolt-firmware.c
superm1: oh, okay lets leave it as is for now then and revisit when we see some usb4 stuff in the wild.

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: Is we want to find out if it's host we have to read a *lot* more of the image. 

    failed to read is-host: location is outside of the given image: attempted to read 0x01 bytes at offset 0x7f014b from buffer of 0x80

I've pushed a commit that at least makes use of the native boolean.

superm1: Sounds good, thanks.

plugins/thunderbolt/fu-thunderbolt-firmware.c
YehezkelShB: There is a reason why it was just one another location in the comparisons done for the images. This is just an "enum"-like value. There is no point with trying to treat it as a "real" size.

plugins/thunderbolt/fu-thunderbolt-firmware.c
```diff
@@ -0,0 +1,521 @@
...
+							  0x10C,
+							  &pd_pointer,
+							  error)) {
+			g_prefix_error (error, "failed to read pd-pointer: ");
+			return FALSE;
+		}
+		self->has_pd = fu_thunderbolt_firmware_valid_pd_pointer (pd_pointer);
+	}
+
+	if (!self->is_host) {
```
YehezkelShB: In the original code, this was checked and compared only for host, and not for all of them (just from AR). This wasn't a mistake. As I said, it's hard to follow all the changes done here in this massive code move, but any change that is more than pure refactoring, e.g. checking more things than was checked before or less things, isn't guaranteed to work properly.

YehezkelShB:
Seems like you decided to get rid of all the FW comparisons made for host FW, I can't find them in the new code. What am I missing?

plugins/thunderbolt/fu-thunderbolt-firmware.c
YehezkelShB: I think the assumption was that there will be new gen after 3, but these locations were meant to be kept the same. I don't see any harm in keeping it like this. New gen will not be supported automatically anyway, we have to add it to the controller ID list first.

hughsie:
> Seems like you decided to get rid of all the FW comparisons made for host FW

This was deliberate. The FuFirmware validation in all the other plugins is really just an end-user sanity check, rather than a system integrator check. In all honesty checking all the other fields is way overkill for what we're doing, and also means we have to update a big table of data for every generation change.

plugins/thunderbolt/fu-thunderbolt-firmware.c
hughsie: Understood. @superm1 do you think we should just remove the flash_size completely? It doesn't seem super-useful.

YehezkelShB:
> > Seems like you decided to get rid of all the FW comparisons made for host FW
> 
> This was deliberate. The FuFirmware validation in all the other plugins is really just an end-user sanity check, rather than a system integrator check. In all honesty checking all the other fields is way overkill for what we're doing, and also means we have to update a big table of data for every generation change.

I hope you talked to FW team to hear of they are OK with it. One of the real issues is that not all host FW include DROM. Especially, pre-AR FWs didn't have it at all. So you can't trust the vendor and model IDs to save you from flashing incompatible FW. I recommend to keep all those comparisons, even if they don't look like the style you'd like for "modern" plugin.

hughsie:
> So you can't trust the vendor and model IDs to save you from flashing incompatible FW

These firmware validation tests are really designed for millions of people as a sanity check, rather than a check for the ODMs to properly validate an image.

YehezkelShB:
> > So you can't trust the vendor and model IDs to save you from flashing incompatible FW
> 
> These firmware validation tests are really designed for millions of people as a sanity check, rather than a check for the ODMs to properly validate an image.

OK, I understand your point. It makes sense when you look at it from this perspective.

BTW, sorry again for not having the time right now for deep diving into this review.

plugins/thunderbolt/fu-thunderbolt-firmware.c
superm1: I would prefer lets leave it, I think that this type of thing is potentially useful for pre-production hardware for integrators and developers to catch a problem.

superm1: (It of course should be added to the output when parsing to make it useful though)

hughsie: I've wired it in properly. @YehezkelShB good catch about it only being AR and TR.

YehezkelShB: Still missing the "it was checked only for host" part of my comment. The current code does the opposite

superm1:
@hughsie I think we'll probably want to slip this to later than 1.4.0.  Hopefully @gicmo can help with the testing framework changes and eventualy I'll work on updating the force power plugin to work with this.

My main challenge with force power is with WFH I don't have anything that uses force power anymore to test with.

hughsie:
> I think we'll probably want to slip this to later than 1.4.0

Agree.

libfwupdplugin/fu-udev-device.c
```diff
@@ -943,6 +943,83 @@ fu_udev_device_pwrite (FuUdevDevice *self, goffset port, guint8 data, GError **e
...
 
+/**
+ * fu_udev_device_get_parent_name
+ * @self: A #FuUdevDevice
+ *
+ * Returns the name of the direct ancestor of this device
+ *
+ * Returns: string or NULL if unset or invalid
+ *
+ * Since: 1.4.0
```
superm1, line 954: need to fix this to 1.4.2

libfwupdplugin/fu-udev-device.c
```diff
@@ -943,6 +943,83 @@ fu_udev_device_pwrite (FuUdevDevice *self, goffset port, guint8 data, GError **e
...
+ * fu_udev_device_get_sysfs_attr:
+ * @self: A #FuUdevDevice
+ * @attr: offset address
+ * @error: A #GError, or %NULL
+ *
+ * Reads an arbitrary sysfs attribute 'attr' associated with UDEV device
+ *
+ * Returns: string or NULL
+ *
+ * Since: 1.4.0
```
superm1, line 982: need to fix this to 1.4.2

libfwupdplugin/fwupdplugin.map
```diff
@@ -569,6 +569,8 @@ LIBFWUPDPLUGIN_1.4.0 {
     fu_hid_device_set_report;
     fu_plugin_get_config_value_boolean;
     fu_plugin_runner_device_created;
+    fu_udev_device_get_parent_name;
+    fu_udev_device_get_sysfs_attr;
```
superm1, line 573: need to fix this to 1.4.2

superm1:
@hughsie @gicmo so hypothetically I've got the test suite fixed up now to work with the plugin overhaul, can you review that part and make sure it's looking good?

The last thing left is fixing up force power to work again, which will have to be best effort since I don't have hardware using it anymore.

superm1:
Btw the most difficult part to solve was understanding why change events were flowing through.  Turns out that gudev caches sysfs attributes!  I was going crazy trying to follow in a debugger why they weren't refreshing until I just straced the binary and noticed nothing called out to `nvm_version` again.
https://github.com/lu-zero/udev/blob/28f04cff8a5b4e4dd979438b7e91732a309400e7/src/libudev-device.c#L1286

libfwupdplugin/fu-udev-device.c
```diff
@@ -943,6 +943,57 @@ fu_udev_device_pwrite (FuUdevDevice *self, goffset port, guint8 data, GError **e
 #endif
 }
 
+/**
+ * fu_udev_device_get_sysfs_attr:
+ * @self: A #FuUdevDevice
+ * @attr: offset address
```
hughsie, line 975: is this true?

libfwupdplugin/fu-udev-device.c
```diff
@@ -943,6 +943,32 @@ fu_udev_device_pwrite (FuUdevDevice *self, goffset port, guint8 data, GError **e
...
+fu_udev_device_get_parent_name (FuUdevDevice *self)
+{
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GUdevDevice) parent = NULL;
+
+	g_return_val_if_fail (FU_IS_UDEV_DEVICE (self), NULL);
+
+	parent = g_udev_device_get_parent (priv->udev_device);
+	return parent == NULL ? NULL :g_udev_device_get_name (parent);
```
hughsie, line 966: The lifecycle here looks fishy. The `transfer:none` name of the parent belongs to the parent, but we're unreffing it when this function returns. I think this needs to be `g_strdup`'d and a `gchar *` returned.

libfwupdplugin/fu-plugin.c
```diff
@@ -1694,6 +1694,19 @@ fu_plugin_usb_device_added (FuPlugin *self, FuUsbDevice *device, GError **error)
...
+fu_plugin_udev_device_changed (FuPlugin *self, FuUdevDevice *device, GError **error)
+{
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	/* open */
+	locker = fu_device_locker_new (FU_DEVICE (device), error);
+	if (locker == NULL)
+		return FALSE;
+	return fu_device_rescan (FU_DEVICE (device), error);
+
```
hughsie, line 1707: random blank line

libfwupdplugin/fu-plugin.c
```diff
@@ -1872,8 +1885,14 @@ fu_plugin_runner_udev_device_changed (FuPlugin *self, FuUdevDevice *device, GErr
 
 	/* optional */
 	g_module_symbol (priv->module, "fu_plugin_udev_device_changed", (gpointer *) &func);
-	if (func == NULL)
+	if (func == NULL) {
+		if (priv->device_gtype != G_TYPE_INVALID ||
+		    fu_device_get_specialized_gtype (FU_DEVICE (device)) != G_TYPE_INVALID) {
```
hughsie, line 1890: do we need both? I was surprised to see the device_gtype one.

hughsie:
> The last thing left is fixing up force power to work again

I guess my only concern is if there are any vendors shipping TBT updates on the LVFS that require force power. Obviously you know for Dell, but what about Lenovo for example? What's the end-user-visible effect of trying to deploy a tbt update on a tbt controller that needs force power? If I understand correctly the tbt controller will just be "missing" without the force power stuff, unless a tbt device is attached. If that's the only "regression" I think that's acceptable.

libfwupdplugin/fu-plugin.c
superm1: I was modeling off of the usb ones, which yes do.

superm1:
> > The last thing left is fixing up force power to work again
> 
> I guess my only concern is if there are any vendors shipping TBT updates on the LVFS that require force power. Obviously you know for Dell, but what about Lenovo for example? What's the end-user-visible effect of trying to deploy a tbt update on a tbt controller that needs force power? If I understand correctly the tbt controller will just be "missing" without the force power stuff, unless a tbt device is attached. If that's the only "regression" I think that's acceptable.

That's correct, the obvious regression is that you can't update the device unless it's powered by something being plugged in.  I was also thinking this is an acceptable regression.

I'm not aware of any other impact though.

libfwupdplugin/fu-udev-device.c
superm1: fixed

libfwupdplugin/fu-udev-device.c
superm1: fixed

plugins/thunderbolt/fu-thunderbolt-firmware.c
superm1: I've fixed this in the current commit and reversed it back to only checking if on a host controller, thanks for catching this.

plugins/thunderbolt/fu-thunderbolt-firmware.c
superm1: I've adjusted the comment to make it clear that we shouldn't actually use the value here but just match it between images.

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: Added these.

superm1:
Thanks, will merge and continue to test with it.

PR #1922, opened 2020-03-27
Swap out GUID for multiple controllers with same ID

superm1:
Here's the third try at a GUID that is unique but consistent enough when two controllers on the host system share the same vendor ID and device ID.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
(FYI - backported this to 1_3_X too)

PR #1923, opened 2020-03-28
Use a static version number

hughsie:
Getting the version string from git means the commit version changes each time
we commit any patch, which means we need to use --force to install firmware
when building fwupd against a version that should be compatible.

It is also very inconvenient not bumping the release version for git snapshots
as firmware can no longer depend on the "planned" release triplet.

tl;dr: A good idea for Flashrom, not so awesome for me.


superm1:
> Getting the version string from git means the commit version changes each time
we commit any patch, which means we need to use --force to install firmware
when building fwupd against a version that should be compatible.

Something else we could alternatively do is instead only compare major and minor versions when doing this check rather than major/minor/micro.

> It is also very inconvenient not bumping the release version for git snapshots
as firmware can no longer depend on the "planned" release triplet.

Yes this part I've encountered problems with too.

src/fu-util-common.c
```diff
@@ -269,18 +269,11 @@ gchar *
 fu_util_get_client_version (void)
 {
 	GString *string = g_string_new ("");
-
 	g_string_append_printf (string,
 				"%i.%i.%i",
 				FWUPD_MAJOR_VERSION,
 				FWUPD_MINOR_VERSION,
 				FWUPD_MICRO_VERSION);
-#ifdef FWUPD_DIRTY_VERSION
```
superm1: If you prefer to go this approach with release versions, I do think we still need to show the commit version in the --version.  It's very useful to know if someone is testing with a patch or without a patch.

hughsie: Can we wire in a git check into meson easier than using an external helper? I have to switch to daddy mode now, but can look again on Monday.

meson.build
```diff
@@ -9,20 +9,23 @@ fwupd_version = meson.project_version()
...
-  conf.set_quoted('FWUPD_COMMIT_VERSION', fwupd_commit)
-endif
+conf = configuration_data()
 conf.set('FWUPD_MAJOR_VERSION', fwupd_major_version)
 conf.set('FWUPD_MINOR_VERSION', fwupd_minor_version)
 conf.set('FWUPD_MICRO_VERSION', fwupd_micro_version)
 conf.set_quoted('PACKAGE_VERSION', fwupd_version)
 
+# non tagged builds
+git_version = run_command('git', 'describe').stdout().strip()
```
superm1: to avoid making a hard git dependency I think you need to check for if git is available to run

hughsie: Good catch, fixed.

src/fu-util-common.c
hughsie: Fixed.

PR #1931, opened 2020-03-31
Wip/hughsie/ccgx2

hughsie:


hughsie:
@chlee75

chlee75:
I have tested segmentation fault with this patch.
There is no error as far. I think segmentation fault has been fixed.
Could you please upload it to LVFS?
Ryan.

plugins/ccgx/fu-ccgx-hid-device.c
```diff
@@ -44,13 +45,47 @@ fu_ccgx_hid_device_detach (FuDevice *device, GError **error)
...
+	g_autoptr(GError) error = NULL;
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	/* do this in idle so that the original HID device gets added */
+	locker = fu_device_locker_new (self, &error);
+	if (locker == NULL) {
+		g_warning ("failed to open HID device: %s",
+			   error->message);
+	} else {
+		if (!fu_ccgx_hid_device_detach (FU_DEVICE (self), &error)) {
```
superm1: What actually happens?  Does the FuDevice disappear?  I worry this is a timing hack that might not behave the same reliably.

hughsie: Yes, it unplugs and replugs with a different VID PID.

hughsie: actually, maybe a cleaner thing to do might be to detach and then return FALSE...

hughsie: actually 2, we need to add the device sometimes as in future hardware the HID mode is updatable, so we can't return failure from setup.

superm1: Well you could also just have a dummy setup method and do the detach in probe perhaps?

hughsie: I've switched it to returning FALSE from setup. Setup already has the device open so it's the natural time to do it. If/when we have a device that can be flashing in HID mode we can just do a different thing in setup.

hughsie:
@chlee75 can you verify this fixed version still works please?

hughsie:
> this fixed version still works please

It just worked for me using the embargo remote.

PR #1932, opened 2020-03-31
Install meson from pip for LGTM

hughsie:


.lgtm.yml
```diff
@@ -9,9 +9,14 @@ extraction:
...
+      - python3-pip
+      - python3-setuptools
+      - python3-wheel
       - python3-cairo
       - libssl-dev
     after_prepare:
+    - python3 -m pip install --user "meson >= 0.52.0"
+    - export PATH="$HOME/.local/bin:$PATH"
     - "wget -O libxmlb.zip https://github.com/hughsie/libxmlb/archive/0.1.13.zip"
     - "mkdir -p subprojects/libxmlb"
```
superm1: Is this part still needed?  It seems that libjcat at least is getting downloaded by subproject now on it's own

hughsie: Lets find out! I'll open a new PR with this change. It kind sucks we can't test the LGTM until we change the base commit. Fingers crossed!

Issue #1914, opened 2020-03-26
Supported flag being applied to stuff it shouldn't be


FFY00:
**Is your feature request related to a problem? Please describe.**
When debugging firmware updates we need to force a firmware install. The only way I know how to do this is by forcing the install of a .cab file. This, however, is not optimal.

**Describe the solution you'd like**
I would like to have a flag to force an update to the latest version in the repos. At least in `fwupdtool`.

**Describe alternatives you've considered**
Update from a .cab file.

hughsie:
Why do you need to force it? Is there a requirements issue or something?

FFY00:
For eg. testing new changes for the logitech-hidpp plugin. There are 4 firmware variants for receivers, I need to get .cab files for all and update. If I were able to just pass a flag to force reinstalling the latest firmware from the repos it would save some time.

Or am I doing something wrong?

hughsie:
`fwupdmgr reinstall` ?

FFY00:
That should work. It would be great to also have that in `fwupdtool`.

hughsie:
Ahh! I'm covered in children at the moment but if you could do a PR I'd be glad to review it. It should be simple to port from fwupdmgr like the other commands. TIA.

FFY00:
Sure, I'll do it when I can find some time.

FFY00:
So, fwupd is not finding any available firmware for Logitech receivers, which are my primary targets with this.

You should be able to reproduce:
```
$ fwupdmgr reinstall
No releases found: Bootloader is not compatible
```

I also tried, as a hack, not stopping in the first firmware found ans it's still the same.

superm1:
Hmm, I had no problem with master just now (at fe4b3ead)

My current firmware:
```
└─Unifying Receiver:
      Device ID:           8ad41d663d3f4ea61a81cbb32a54411a101eb850
      Summary:             A miniaturised USB wireless receiver
      Current version:     RQR12.10_B0032
      Bootloader Version:  BOT01.02_B0014
      Vendor:              USB:0x046D
      Install Duration:    30 seconds
      GUIDs:               9d131a0c-a606-580f-8eda-80587250b8d6
                           fcf55bf5-767b-51ce-9c17-f6f538c4ee9f
                           279ed287-3607-549e-bacc-f873bb9838c4
      Device Flags:        • Updatable
                           • Supported on remote server
```

Reinstalling:
```
$ fwupdmgr reinstall 
Choose a device:
[...]

PR #1935, opened 2020-03-31
logitech_hidpp: don't set bootloader as updatable

superm1:
I noticed this when trying to replicate issues in https://github.com/fwupd/fwupd/issues/1914

Bootloader is not actually updatable and the protocol isn't actually set
leading to a daemon warning when installing replacement firmware for the runtime.

```
15:03:42:0124 FuEngine             device 14ee44ec8b5a38110f48482adeb389e97a706747 [Unifying Receiver] does not define an update protocol
```

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-bootloader.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-bootloader.c
@@ -436,7 +436,6 @@ fu_logitech_hidpp_bootloader_request (FuLogitechHidPpBootloader *self,
 static void
 fu_logitech_hidpp_bootloader_init (FuLogitechHidPpBootloader *self)
 {
-	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
 	fu_device_add_icon (FU_DEVICE (self), "preferences-desktop-keyboard");
 	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_PLAIN);


hughsie:
I think bootloader is updatable; otherwise how would users recover the receiver on failed flashing? I think the proper fix here is to set the protocol really.

FFY00:
The bootloader is updateable. In fact, the device can only receive firmware in bootloader mode. During the normal update process we put the device in bootloader mode and then send the firmware.

superm1:
> I think bootloader is updatable; otherwise how would users recover the receiver on failed flashing? I think the proper fix here is to set the protocol really.

Oh it is updatable?  OK...

FFY00:
> Oh it is updatable? OK...

Yes, so if it is not updating, something else went wrong.

superm1:
> Yes, so if it is not updating, something else went wrong.

It's unusual to update the bootloader in the field for production hardware. It's usually kept read only and you just upgrade the application image.

That's the reason you can typically recover without using an ISP to flash.

hughsie:
It's updating the fw *from* the bootloader.

superm1:
> It's updating the fw _from_ the bootloader.

Oh... I understand.

So you're saying basically if the device was already in bootloader mode from a failed flash you would see this fwupd device that has `is-bootloader` and expect it to be upgradable got get back to runtime mode not that you would actually flash a new bootloader image from the bootloader.  That's a lot more sensible.

I've pushed what I think should set the protocol then now to this branch.

FFY00:
> So you're saying basically if the device was already in bootloader mode from a failed flash you would see this fwupd device that has `is-bootloader` and expect it to be upgradable got get back to runtime mode not that you would actually flash a new bootloader image from the bootloader. That's a lot more sensible.

Yes.

PR #1936, opened 2020-03-31
trivial: Use subprojects when building using LGTM

hughsie:


superm1:
Can do same with flashrom right?

hughsie:
> Can do same with flashrom right?

Looks like it! I was just waiting to hear back from the CI, but looks good now.

Issue #1876, opened 2020-03-17
logitech-hidpp refactoring


FFY00:
So, I am looking at the code and there are a few thing I think could be improved. I would like to refactor it in order to make it easier to support different devices.

Some things I would like to be able to achieve in the future:
  - Easily add more receivers
  - Support wired HID++ 2.0 devices

I have some questions:

1) The receiver bootloader protocol identification routine ({texas,nordic}/{,un}signed) seems to be split between `fu_logitech_hidpp_runtime_setup_internal` (usigned/unsigned) and `fu_logitech_hidpp_runtime_probe` (texas/nordic). Is there any reason for this?

2) To prevent all this regex craziness, I think we should mark have 4 different protocols, not just `com.logitech.unifying{,signed}`. Would it be possible to make this change? As this requires changes to the both metadata and the code we would need a min and max deamon version field in the metadata? This way we could keep updates for older daemons in the server. Is there anything like this in place? If not, is it doable to add>

3) As said in #1384, we will make the kernel issue a udev update event when an unavailable device becomes online again. Does fwupd already listen to those events?

superm1:
2. I think I'd lean upon instead implementing support for a payload to be used for multiple protocols, particularly if the daemon on older versions is able to recognize the first one and use that.  We got lucky with version format working that way.
Then we can make a switch on LVFS side to add a second protocol to all payloads as needed.

3. Yeah, the plugin just needs to indicate which subsystem of events it needs to pick up.  You can see that a bunch of other plugins do this already too.

PR #1938, opened 2020-03-31
logitech_hidpp: set the bootloader protocol

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-bootloader.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-bootloader.c
@@ -218,6 +218,8 @@ static gboolean
 fu_logitech_hidpp_bootloader_set_bl_version (FuLogitechHidPpBootloader *self, GError **error)
 {
 	guint16 build;
+	guint8 major;
+	guint8 minor;
 	g_autofree gchar *version = NULL;
 	g_autoptr(FuLogitechHidPpBootloaderRequest) req = fu_logitech_hidpp_bootloader_request_new ();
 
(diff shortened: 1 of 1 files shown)


hughsie:
Looks good to me, but I'll defer to @FFY00 for the version checks.

plugins/logitech-hidpp/fu-logitech-hidpp-bootloader.c
```diff
@@ -232,15 +234,20 @@ fu_logitech_hidpp_bootloader_set_bl_version (FuLogitechHidPpBootloader *self, GE
...
+	version = fu_logitech_hidpp_format_version ("BOT", major, minor, build);
 	if (version == NULL) {
 		g_prefix_error (error, "failed to format firmware version: ");
 		return FALSE;
 	}
 	fu_device_set_version_bootloader (FU_DEVICE (self), version);
+
+	if ((major == 0x01 && minor >= 0x04) ||
+	    (major == 0x03 && minor >= 0x02))
+		fu_device_set_protocol (FU_DEVICE (self), "com.logitech.unifyingsigned");
```
FFY00: Wouldn't it be better to change the protocol to be defined as `com.logitech.{nordic,texas}.{un,}signed`? We have all the required information right here. I don't think there's any need to have the split this way.

superm1: Well right now there isn't actually any protocol prefixes defined as nordic or texas.  We also don't have fallback code to allow devices to have multiple protocols.  For now this matches the protocols in use by the stack and LVFS.

I think we should work out the details of introducing more protocols (and potentially fallback protocols) separately from this bug fix (as discussing in https://github.com/fwupd/fwupd/issues/1876)

PR #1942, opened 2020-04-01
Only order `fwupd-refresh.service` after `network.target`

paulmenzel:


paulmenzel:
The device can be offline any time (laptops).

    $ /usr/bin/fwupdmgr refresh --no-metadata-check --force
    Metadaten werden abgerufen https://cdn.fwupd.org/downloads/firmware.xml.gz

    Failed to download https://cdn.fwupd.org/downloads/firmware.xml.gz: Cannot resolve hostname

So probably, the ordering can be removed, and a policy should be decided for `fwupdmgr` how to gracefully deal with an offline device.

superm1:
> So probably, the ordering can be removed, and a policy should be decided for fwupdmgr how to gracefully deal with an offline device.

Can you define what's not graceful about it right now?  The metadata and refresh is not required.

superm1:
And thanks for your fix on the targets.  I'll backport that to 1_3_X.

paulmenzel:
> I'll backport that to 1_3_X.

In my opinion it’s not important enough to warrant a backport. But your call.

superm1:
> In my opinion it’s not important enough to warrant a backport. But your call.

It's small and strictly a fix.  When we tag unstable 1.4.0 with the new features and libjcat support I expect we'll also tag a 1.3.10 for distros tracking stabler stuff.

PR #1943, opened 2020-04-01
trivial: fix self tests for empty `/etc/machine-id` (LP: #1870051)

superm1:
This ports the fix https://github.com/hughsie/libjcat/commit/6ac46df8e754006b0e8900e16a8ee6412a8d7360
from libjcat to fwupd 1_3_X.  master is not affected.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-self-test.c
+++ b/src/fu-self-test.c
@@ -2708,6 +2708,7 @@ static void
 fu_keyring_pkcs7_self_signed_func (gconstpointer user_data)
 {
 #ifdef ENABLE_PKCS7
+	static const gchar payload_str[] = "Hello, world!";
 	gboolean ret;
 	g_autoptr(FuKeyring) kr = NULL;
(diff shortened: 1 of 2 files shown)


hughsie:
PSA: I "fixed" the travis integration by removing and then re-adding the authorisation to github. Apparently it's a know issue.

src/fu-self-test.c
```diff
@@ -2708,6 +2708,7 @@ static void
 fu_keyring_pkcs7_self_signed_func (gconstpointer user_data)
 {
 #ifdef ENABLE_PKCS7
+	static const char payload_str[] = "Hello, world!";
```
hughsie, line 2711: I'd really like `const gchar` if possible.

superm1: sure.  will adjust.

PR #1944, opened 2020-04-01
stm-dfu: fixes for flashing dfuse on STM32f745

amfern:

PR #1946, opened 2020-04-02
dfu: Add a way to set the timeout from a quirk

hughsie:


hughsie:
@amfern this is probably what we should do.

amfern:
tested it with #1944 and it works

Issue #391, opened 2018-02-09
`fwupdmgr refresh` fails due to invalid signature


cocreature:
Running `fwupdmgr refresh` fails on my system with the following error message:
```shell
$ fwupdmgr refresh
Fetching metadata https://s3.amazonaws.com/lvfsbucket/downloads/firmware.xml.gz
Downloading…           [***************************************]
Fetching signature https://s3.amazonaws.com/lvfsbucket/downloads/firmware.xml.gz.asc

Failed to update metadata for lvfs: '48A6D80E4538BAC2' is not a valid signature
```

This is a new issue, on wednesday (07.02.2018), refreshing worked just fine. I can reproduce this on another machine as well (also Archlinux).

```shell
$ fwupdmgr --version
client version: 1.0.4
daemon version: 1.0.4
compile-time dependency versions
        appstream-glib: 0.7.4
        gusb:   0.2.11
```

- Operating system and version:
Archlinux 64bit
- How did you install fwupd (ex: `from source`, `pacman`, `apt-get`, etc):
`pacman`
- Have you tried rebooting?
Yes
- Are you using an NVMe disk?
Yes
- Is secure boot enabled (only for the UEFI plugin)?
No

superm1:
Yep, I can also reproduce on Ubuntu with master.

hughsie:
I'll try regenerating the metadata first as a quick fix.

hughsie:
Right, that worked, I'll have a look at the old files and try to work out what went wrong. I have a suspicion it's my fault and the staging LVFS accidentally uploaded metadata with the test key.

hughsie:
Nope, it appears the right key was used:

    $ gpg2 --verify firmware.xml.gz.asc firmware.xml.gz
    gpg: Signature made Wed 07 Feb 2018 12:47:04 GMT
    gpg:                using RSA key 48A6D80E4538BAC2
    gpg: BAD signature from "Linux Vendor Firmware Service <sign@fwupd.org>" [unknown]

(I have the public key imported system wide on this machine)

cocreature:
Thanks for the fix, that was impressively fast! I’ll leave it up to you to decide when to close the issue.

hughsie:
In the event log:

    2018-02-07 12:47:03 | ???.???.???.??? | dellqa | Uploaded file 05e0d0906098a57c4d0aa71d5f5fc4e620e7ee03-Signed_1152921504627420111.cab to stable /lvfs/upload

[...]

PR #1950, opened 2020-04-03
Load the signature to get the aliased CDN-safe version of the metadata

hughsie:
Switch to downloading the signature first, which we can then load to get the
suffixed build-specific URL of the actual metadata file. You need to have
libjcat 0.1.1 installed and fwupd built against the new version for this to
work.

Fixes https://github.com/fwupd/fwupd/issues/391

### DIFF
--- a/libfwupd/fwupd-remote.c
+++ b/libfwupd/fwupd-remote.c
@@ -7,6 +7,7 @@
 #include "config.h"
 
 #include <libsoup/soup.h>
+#include <jcat.h>
 
 #include "fwupd-deprecated.h"
 #include "fwupd-enums-private.h"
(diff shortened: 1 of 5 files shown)


libfwupd/fwupd-remote.c
```diff
@@ -896,6 +897,68 @@ fwupd_remote_get_metadata_uri (FwupdRemote *self)
...
+	g_autofree gchar *basename = NULL;
+	g_autofree gchar *baseuri = NULL;
+	g_autofree gchar *metadata_uri = NULL;
+	g_autoptr(GFile) gfile = NULL;
+	g_autoptr(JcatFile) jcat_file = jcat_file_new ();
+	g_autoptr(JcatItem) jcat_item = NULL;
+
+	g_return_val_if_fail (FWUPD_IS_REMOTE (self), FALSE);
+	g_return_val_if_fail (filename != NULL, FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
superm1, line 926: why so hardpressed for error being set to something?

hughsie: This is shorthand for "either error is NULL, or error is not already set"

hughsie:
> You'll need to bump up the subproject and minimum libjcat requirements as part of this PR.

If we did `jcat_item_get_default` in the error path, that works without bumping the version too, right?

superm1:
> > You'll need to bump up the subproject and minimum libjcat requirements as part of this PR.
> 
> If we did `jcat_item_get_default` in the error path, that works without bumping the version too, right?

I guess that is technically true.  I had some thoughts on using `jcat_item_get_alias_ids` instead, but that's not worth bumping up.

hughsie:
Fixed, although ideally we'd merge this first to make it testable: https://github.com/hughsie/libjcat/pull/29

PR #1954, opened 2020-04-06
trivial: Allow using --force with install-blob

hughsie:


src/fu-engine.c
```diff
@@ -1819,6 +1819,7 @@ fu_engine_install_release (FuEngine *self,
 		} else {
 			fu_device_set_update_state (device, FWUPD_UPDATE_STATE_FAILED);
 		}
+		g_debug ("returned error was %s", error_local->message);
```
superm1, line 1822: I wasn't sure this was really necessary since update_error got set, but I suppose using fwupdtool you wouldn't see that error otherwise.  As an alternative what do you think about moving this debug message into the fu_device_set_update_error call?

hughsie: It is in the fu_device_set_update_error() call, no? Perhaps we should just show it in fwupdtool....

superm1: Yeah that sounds cleaner too.

Issue #1951, opened 2020-04-03
Logitech hidpp reciever is getting stuck in bootloader mode


superm1:
**Describe the bug**

Running 'reinstall' on this device is causing it to timeout and get stuck in bootloader mode.
```
  DeviceId:             50c93db0cd70243d33c15121e7d67d644f1718e3
  Guid:                 9d131a0c-a606-580f-8eda-80587250b8d6
  Guid:                 fcf55bf5-767b-51ce-9c17-f6f538c4ee9f <- HIDRAW\VEN_046D&DEV_C52B&REV_00
  Guid:                 279ed287-3607-549e-bacc-f873bb9838c4 <- HIDRAW\VEN_046D&DEV_C52B
  Summary:              A miniaturised USB wireless receiver
  Plugin:               logitech_hidpp
  Protocol:             com.logitech.unifying
  Flags:                supported|registered|wait-for-replug
  VendorId:             USB:0x046D
  Version:              RQR12.10_B0032
  VersionBootloader:    BOT01.02_B0014
  VersionFormat:        plain
  Icon:                 preferences-desktop-keyboard
  InstallDuration:      30
  Created:              2020-04-03
  Modified:             2020-04-03
  UpdateState:          failed
  PhysicalId:            DEVPATH=/devices/pci0000:00/0000:00:1d.0/0000:03:00.0/0000:04:04.0/0000:3c:00.0/0000:3d:02.0/0000:3e:00.0/usb5/5-2/5-2.3/5-2.3.3/5-2.3.3:1.2
  SignedFirmware:        false
```

**Steps to Reproduce**
Steps to reproduce the behavior.

`fwupdtool reinstall --plugin-whitelist=logitech_hidpp $DEVICEID`

I did happen to get a segfault once as well, but didn't get it the second time.
[segfault.txt](https://github.com/fwupd/fwupd/files/4429209/segfault.txt)

However there appears to be some memory cleanup bug problem with the connected devices (which is probably new due to `EnumerateAllDevices=true`.

```
18:22:05:0732 FuDeviceList         doing delayed removal
18:22:05:0732 FuDeviceList         ::removed 224b61aa4fea82347a28e1bb02c77ca689e38987
18:22:05:0732 FuMain               REMOVED:
FuLogitechHidPpPeripheral:
Logitech M510
  DeviceId:             224b61aa4fea82347a28e1bb02c77ca689e38987
[...]

PR #1957, opened 2020-04-06
logitech-hidpp: Do not add peripheral devices for receiver in BL mode

hughsie:
This should fix flashing firmware when there are >= 2 receivers plugged in.

Fixes https://github.com/fwupd/fwupd/issues/1951


superm1:
I don't have the two receiver setup handy, but I suppose you already tested it since you came up with the fix?

hughsie:
> but I suppose you already tested it since you came up with the fix

Well, it fixed the crash. There were more commits coming up on the branch, but I had to switch to daddy mode :) New PR here: https://github.com/fwupd/fwupd/pull/1958

PR #1962, opened 2020-04-07
ccgx: Show a better error message on failure

hughsie:


hughsie:
> there is no need to show the actual hpi_event

I included that as there are quite a few "gaps" in the enumerated space and in the future there might be new errors not in the _to_string table.

>  Also, why have the dashes in the error string

It's what I did everywhere else :)

superm1:
OK sounds good.

PR #1965, opened 2020-04-07
Only auto-add counterpart GUIDs when required

hughsie:
Doing this unconditionally means we accidentally 'bleed' one device mode into
another in a non-obvious way. For instance, a device might have two operating
modes with different GUIDs. If firmware is supplied for both modes in the same
cabinet archive then we might accidentally match the 'wrong' firmware when
the daemon has observed a mode switch and added the counterpart GUIDs.

We only really need the counterpart GUIDs when switching between Jabra and DFU
devices where the DFU bootloader VID:PID is not manually tagged with
`CounterpartGuid` in a quirk file. In the general case lets keep it simple to
avoid difficult to find bugs.

### DIFF
--- a/data/device-tests/devices/realtek-rts5855.json
+++ b/data/device-tests/devices/realtek-rts5855.json
@@ -1,0 +1,16 @@
+{
+  "name": "Realtek RTS5855 Webcam",
+  "guids": [
+    "9829f051-47f5-55e7-87dc-a49cf55602e2"
+  ],
+  "releases": [
+    {
+      "version": "0.4",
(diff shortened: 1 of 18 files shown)


hughsie:
@chlee75 this should fix CCGX mode change...

superm1:
> We only really need the counterpart GUIDs when switching between Jabra and DFU
devices where the DFU bootloader VID:PID is not manually tagged with
CounterpartGuid in a quirk file. In the general case lets keep it simple to
avoid difficult to find bugs.

This wouldn't also apply to 8bitdo and Logitech?  I didn't think those were put in quirk files either.

hughsie:
Will test with the regression test thing tomorrow.

PR #1970, opened 2020-04-08
Do not crash the daemon if a plugin does something dumb

hughsie:
In this case, a plugin was doing:

    g_autoptr(FuDevice) parent = fu_device_get_parent (device);

Which is not valid as it is `(transfer none)` but that shouldn't be enough to
get the daemon to crash. If the FuDevice refcount drops to zero just print
a critical warning with the pointer (no FuDevice details are available) and
remove the FuDeviceItem from the device list.

Using `G_DEBUG=fatal-criticals` it is very easy to backtrace at the right place
and find out which naughty plugin needs to go on the thinking spot.

### DIFF
--- a/src/fu-device-list.c
+++ b/src/fu-device-list.c
@@ -463,6 +463,37 @@ fu_device_list_add_missing_guids (FuDevice *device_new, FuDevice *device_old)
 }
 
 static void
+fu_device_list_item_finalized_cb (gpointer data, GObject *where_the_object_was)
+{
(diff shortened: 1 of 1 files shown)


superm1:
> naughty plugin

Which one did you catch doing the bad behavior to prompt this?

hughsie:
> Which one did you catch doing the bad behavior to prompt this?

wacom-usb :)

PR #1972, opened 2020-04-08
logitech-hidpp: Add two new GUIDs that are useful for the new device …

hughsie:
…tests

The new regresion device tests are awesome, but dumb. They look for a specific
GUID and run a test if found. The Logitech unifying devices in runtime mode
currently have the same set of GUIDs.

Add a new GUID which can be used by the regression tests without adding a ton
of complexity in the client code doing a requirements check.

FIXME: If a good idea also have to edit the GUIDs used in device-tests/devices/logitech-*

### DIFF
--- a/plugins/logitech-hidpp/README.md
+++ b/plugins/logitech-hidpp/README.md
@@ -38,6 +38,8 @@ These devices use the standard USB DeviceInstanceId values when in DFU mode:
 
 When in runtime mode, the HID raw DeviceInstanceId values are used:
 
+ * `HIDRAW\VEN_046D&DEV_C52B&TYPE_SIGNED`
+ * `HIDRAW\VEN_046D&DEV_C52B&TYPE_LEGACY`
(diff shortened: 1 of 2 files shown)


hughsie:
The alternate of course is to assign a specific error code for failed requirements...

superm1:
Well and then the other alternative is to introduce new protocols and fallback protocols and let the client look at that (as was discussed  in #1876)

FFY00:
> Well and then the other alternative is to introduce new protocols and fallback protocols and let the client look at that (as was discussed in #1876)

I think that's the best option but this is a resonable change for now.

superm1:
Well for now - there is another option though too.  How about using `fwupd_device_get_protocol ` in the client?  It is exported.

hughsie:
>  How about using fwupd_device_get_protocol

That's a really good idea! Will respin.

hughsie:
Closed in favour of https://github.com/fwupd/fwupd/pull/1974

PR #1974, opened 2020-04-08
trivial: Allow filtering the device-test by the protocol

hughsie:
This allows us to run the device tests with all 4 supported types of unifying
dongles plugged in at the same time.


hughsie:
Will test with the actual devices tomorrow when back in the office. Thanks.

hughsie:
> Will test with the actual devices tomorrow when back in the office

This worked rather well. Thanks for the idea!

PR #1976, opened 2020-04-08
Do not conditionalize attach() and detach() on IS_BOOTLOADER

hughsie:
This is nice in theory, until you need to look at the bootloader status of the
parent, or of a different device entirely. Handle this in plugins for the few
cases we care about and stop setting or clearing IS_BOOTLOADER manually just to
get the vfuncs to be run.

Note: I do not think we want to use cleanup() for attaching devices not in
bootloader states -- as cleanup is only run at the end of the composite update.


hughsie:
Not tested on the big table-o-hardware yet, but might fix the VLI usbhub issue.

hughsie:
The big table-o-hardware approves, and this also fixes VLI usbhub (tests added).

PR #1978, opened 2020-04-09
Wip/hughsie/dell dock regression

hughsie:
Only lightly tested as I only have the DVT2.2 dock here.

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -47,35 +47,6 @@ fu_plugin_dell_dock_create_node (FuPlugin *plugin,
 	if (locker == NULL)
 		return FALSE;
 
-	fu_plugin_device_add (plugin, device);
-
-	return TRUE;
-}
-
-static gboolean
-fu_plugin_dell_dock_probe (FuPlugin *plugin,
-			   FuDevice *symbiote,
-			   GError **error)
-{
-	g_autoptr(FuDellDockEc) ec_device = NULL;
-
-	/* create all static endpoints */
-	ec_device = fu_dell_dock_ec_new (symbiote);
-	if (!fu_plugin_dell_dock_create_node (plugin,
-					      FU_DEVICE (ec_device),
-					      error))
-		return FALSE;
-
-	/* create TBT endpoint if Thunderbolt SKU and Thunderbolt link inactive */
-	if (fu_dell_dock_ec_needs_tbt (FU_DEVICE (ec_device))) {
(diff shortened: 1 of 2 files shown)


superm1:
> Only lightly tested as I only have the DVT2.2 dock here.

Is the plugin rejecting this board ID from further stuff and from flashing?  I forget if that hardware build was supposed to be blocked.

hughsie:
https://fwupd.org/lvfs/firmware/3874 is still in embargo -- is that what you want me to try?

superm1:
> https://fwupd.org/lvfs/firmware/3874 is still in embargo -- is that what you want me to try?

Yes

hughsie:
Okay, that fails. Will need to do some more thinking here.

superm1:
I haven't gotten to mine yet today to try this PR, logs from failure for insight?

hughsie:
>  I forget if that hardware build was supposed to be blocked

FwupdError: Device Package level of Dell dock [b5a854df85ab30fff48e4365ec7ad2249c90806e] does not currently allow updates

But, I have another, simpler, idea.C=can I reference the firmware i embargo for the device-tests?

hughsie:
Simpler idea here: https://github.com/fwupd/fwupd/pull/1979

superm1:
> But, I have another, simpler, idea.C=can I reference the firmware i embargo for the device-tests?

Yeah, it's going to be stable eventually and if something newer makes it to stable adjusting device tests is a few line change.

superm1:
> FwupdError: Device Package level of Dell dock [b5a854df85ab30fff48e4365ec7ad2249c90806e] does not currently allow updates

Hmm, that sounds like it might be that the board level is too old indeed.  Is this the board I sent you or was it an internal one?

hughsie:
>  Is this the board I sent you or was it an internal one

This is the prototype DVT one. I'm still waiting for the "release" device from Karl.

superm1:
Hmm, I just looked through my records, I sent a `DVT2.3-C2`.  Do you have that handy?  That's really close to production and should take all the production payloads.

hughsie:
> I sent a DVT2.3-C2

I have that here.

superm1:
Oh, you said DVT2.2 above

hughsie:
> Oh, you said DVT2.2 above

Sorry, it says:

    Base: DVT2.2
    Module: DVT2.2

then there's a smaller label with:

    DP/N SLM-DVT2.3-C2 X02-03

Randomly I can install the updates now. Weird.

superm1:
> Randomly I can install the updates now. Weird.

Remember it's got the delayed activation stuff.  If it's not been unplugged after an update you either need to manually call activate or unplug it.

Issue #1975, opened 2020-04-08
dell_dock: several devices don't remove on unplug


superm1:
With current master I'm finding that the EC device, package level device and VMM5331 are not getting removed when unplugged now.

superm1:
And plugging back in it's not enumerating correctly too.

hughsie:
Do we call `fu_device_list_emit_device_removed()` on the package level child device? Some logs would be great if you have them.

superm1:
Hopefully can do one better.  With such an easy to repro condition bisect is easy.
```
d$  git bisect log
git bisect start
# bad: [6c0e3d4efa8d138540e8c8447e0e87fb3de8b0a8] Do not crash the daemon if a plugin does something dumb
git bisect bad 6c0e3d4efa8d138540e8c8447e0e87fb3de8b0a8
# good: [1e571730d0b39574c3d7f2d0cc38e7c859205dbc] Fix a daemon crash when removing children
git bisect good 1e571730d0b39574c3d7f2d0cc38e7c859205dbc
# good: [49797d7de4a45981cc271fb22474c2b78ffeb0c2] ccgx: Reset the i2c bridge on status failure
git bisect good 49797d7de4a45981cc271fb22474c2b78ffeb0c2
# bad: [2671aaceb26a9b6874468c6569465a0e16f5c6b7] synaptics-cxaudio: Only use the PATCH2 address when the patch level is 2
git bisect bad 2671aaceb26a9b6874468c6569465a0e16f5c6b7
# good: [674478d4b83f91168c8b55f0a29ee513c32d9d0a] bash: add completion for fwupdtool reinstall
git bisect good 674478d4b83f91168c8b55f0a29ee513c32d9d0a
# bad: [60103e22070fa7edffa3b84b4f5e0b8ba4b9ef0f] ccgx: Use Silicon ID and Application type to set flash parameters
git bisect bad 60103e22070fa7edffa3b84b4f5e0b8ba4b9ef0f
# good: [80d9b7450ce7f0172775513e81040a6a6dcab635] logitech_hidpp: set the bootloader protocol
git bisect good 80d9b7450ce7f0172775513e81040a6a6dcab635
# good: [3dbb92e5d5fac147eb5b6163ecd0d9e775024ac3] trivial: fu-config: Fix behavior for UpdateMotd key
git bisect good 3dbb92e5d5fac147eb5b6163ecd0d9e775024ac3
[...]

PR #1979, opened 2020-04-09
Wip/hughsie/dell dock regression2

hughsie:


superm1:
This was my idea too in https://github.com/fwupd/fwupd/issues/1975#issuecomment-611214756.
Can you check and make sure that the plug / unplug without a firmware update works here?

hughsie:
>  Can you check and make sure that the plug / unplug without a firmware update works here?

Yes, this fixes the unplug / replug.

data/device-tests/devices/dell-wd19tb-ec.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "Dell WD19TB Dock (EC)",
+  "guids": [
+    "cd357cf1-40b2-5d87-b8df-bb2dd82774aa"
+  ],
+  "releases": [
+    {
+      "version": "01.00.00.04",
+      "file": "0f7990c89255596d6258c9715b59a40878d37c0ac3ef368aa01c99bed0c59a33-WD19FirmwareUpdateLinux_01.00.15.cab"
```
superm1, line 9: Hmm this is actually leaking the full filename; IE someone could easily just prepend https://fwupd.org/downloads/.  Could you just make a local symlink for now to obfuscate it?

data/device-tests/devices/dell-wd19tb-main.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "Dell WD19TB Dock (Main)",
```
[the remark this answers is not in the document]
hughsie: so `s/main/pkgver` perhaps?

data/device-tests/devices/dell-wd19tb-ec.json
hughsie: Fixed. If you run that test you'll have to download it manually and rename it too.

data/device-tests/devices/dell-wd19tb-main.json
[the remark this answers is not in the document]
superm1: Yeah

data/device-tests/devices/dell-wd19tb-ec.json
superm1: OK thanks.  When it goes stable we can revert it back to this too.

superm1:
Just tested on my side, looks good now.

PR #1985, opened 2020-04-11
Add the ability to attach a different device to the one updated

hughsie:


hughsie:
I need some more info from the silicon vendor but I think we need to do something like this in more than one case (tiered and EC-style).

superm1:
Presumably this shouldn't be in 1.4.0. I would ask if counterpart guid is appropriate in this case though?

libfwupdplugin/fu-device.c
```diff
@@ -2124,6 +2172,8 @@ fu_device_add_string (FuDevice *self, guint idt, GString *str)
 		fu_common_string_append_kv (str, idt + 1, "PhysicalId", priv->physical_id);
 	if (priv->logical_id != NULL)
 		fu_common_string_append_kv (str, idt + 1, "LogicalId", priv->logical_id);
+	if (priv->attach_guid != NULL)
+		fu_common_string_append_kv (str, idt + 1, "LogicalId", priv->attach_guid);
```
superm1, line 2176: Wrong str

hughsie:
> I would ask if counterpart guid is appropriate

Counterpart is an alternate mode of operation, i.e. you're either in bootloader or runtime. I think this is a different thing, it's a bit like the symbiote thing you have in dell dock i.e. to do an update to Y you have to write to Y and then power cycle X.

The alternative is we have something like `USE_ROOT_FOR_ATTACH` and `USE_PARENT_FOR_ATTACH` but that seemed too restrictive and I don't think suitable when we're doing things like rebooting the USB hub to restart the PD controller.

I don't think attach_guid is a great name either tho. Maybe controller_guid, or ec_guid. Or maybe the plugin just sets up the "full" object rather than letting the engine do the lookup, although then you can't attach in a different plugin to the update, which seems like something we want to allow.

libfwupdplugin/fu-device.c
hughsie: Fixed, thanks.

superm1:
> it's a bit like the symbiote thing you have in dell dock

What do you think about instead of it just being an attach device making it a "proxy device"?  We could always change dell_dock to use that infrastructure too.  Generalizing it would probably help control the lifecycle of the device across plugins.

hughsie:
> What do you think about instead of it just being an attach device making it a "proxy device"?

Do we always assume the proxy device controls the attach, the detach and the write itself? Maybe this is the sort of thing we want *both* flags and the quirks for, e.g.

    ProxyDevice = USB\VID_17EF&PID_A391
    Flags = use-proxy-for-attach

superm1:
Well how about start with no flag (setting ProxyDevice = attach, detach and write), and if you find a situation that we don't want all 3, change the behavior that ProxyDevice = write and add a quirk for a flag to do attach/detach.

hughsie:
> What do you think about instead of it just being an attach device making it a "proxy device

Lets start with https://github.com/fwupd/fwupd/pull/1990

PR #1987, opened 2020-04-13
Wip/hughsie/inverted flags

hughsie:


libfwupdplugin/fu-device.c
```diff
@@ -1859,6 +1859,21 @@ fu_device_get_physical_id (FuDevice *self)
 void
 fu_device_add_flag (FuDevice *self, FwupdDeviceFlags flag)
 {
+	/* none is not used as an "exported" flag */
+	if (flag == FWUPD_DEVICE_FLAG_NONE)
+		return;
+
+	/* being both a bootloader and requiring a bootloader is invalid */
+	if (flag == FWUPD_DEVICE_FLAG_NONE ||
```
superm1, line 1867: aren't you not going to hit this clause since you checked for none and returned earlier?

libfwupdplugin/fu-device.c
```diff
@@ -1859,6 +1859,21 @@ fu_device_get_physical_id (FuDevice *self)
...
+	    flag == FWUPD_DEVICE_FLAG_NEEDS_BOOTLOADER) {
```
superm1, line 1868: I think that you should do a bitwise check here - in case multiple flags were sent at once.

libfwupdplugin/fu-device.c
```diff
@@ -1859,6 +1859,21 @@ fu_device_get_physical_id (FuDevice *self)
...
+		fu_device_remove_flag (self, FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+	}
+	if (flag == FWUPD_DEVICE_FLAG_NONE ||
+	    flag == FWUPD_DEVICE_FLAG_IS_BOOTLOADER) {
```
superm1, line 1872: same comments as above for these two lines

libfwupdplugin/fu-device.c
hughsie: Urgh, I missed the bit where I remove the ugly NONE hack.

libfwupdplugin/fu-device.c
hughsie: Agreed, thanks.

libfwupdplugin/fu-device.c
hughsie: Okay, respun, thanks.

PR #1988, opened 2020-04-13
Use black to format python source in a consistent manner

hughsie:
No code changes.

### DIFF
--- a/contrib/afl-fuzz.py
+++ b/contrib/afl-fuzz.py
@@ -6,6 +6,7 @@ import sys
 import subprocess
 import os
 
+
 def main():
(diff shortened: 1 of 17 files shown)


superm1:
Don't have concerns to this, but I would ask if there is a configuration file for "black", can you commit it into `contrib` (ie `contrib/python-black-style.cfg`)?  Perhaps also it's worth adding a CI task that runs this and fails if it changed?

hughsie:
Black is awesome in that you can hardly configure anything: https://black.readthedocs.io/en/stable/ -- i.e. no discussion about the optimal format as there's only one right answer :)

PR #1990, opened 2020-04-14
Wip/hughsie/proxy

hughsie:


plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -968,16 +954,13 @@ fu_dell_dock_ec_open (FuDevice *device, GError **error)
...
-
-	return fu_device_close (self->symbiote, error);
+	return fu_device_close (fu_device_get_proxy (device), error);
 }
 
 static void
 fu_dell_dock_ec_finalize (GObject *object)
 {
 	FuDellDockEc *self = FU_DELL_DOCK_EC (object);
-	g_object_unref (self->symbiote);
```
superm1, line 980: I guess this is actually a bug from previously, I didn't see it actually refcounted previously.

superm1: I do wonder if it's safer to actually ref it on open though, no?

hughsie: I don't think we can ref it, else we get a lifecycle dep loop and the device never finalizes.

superm1: Then maybe we shouldn't be doing that in the i2c TBT code too.

hughsie: Ohh, i missed that one. Will fix.

hughsie: Fixed, and added the GUID thing we needed to avoid the MST hack. The GUID engine matcher thing also allows me to fix the VLI PD update thing I think.

superm1:
This LGTM, but I haven't tested it locally yet. I guess you checked on dell-dock on your side already?

hughsie:
> I guess you checked on dell-dock on your side already?

Not yet, will when I get back to my little office.

hughsie:
> I guess you checked on dell-dock

The proxy device stuff works, the GUID matching needs more work. I've removed the last commit from this PR so I can think on it a bit more.

PR #1994, opened 2020-04-14
Expose device physical id to clients

sjoerdsimons:
To allow fwupd clients to better communicate which device is going to be
updated to the user it may be required to know which physical device a
fwupd device refers to. To that end expose the physical id that fwupd
already keeps.

Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>

Type of pull request:
- [ ] Code fix
- [ x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -34,7 +34,6 @@ static void fu_device_finalize			 (GObject *object);
 typedef struct {
 	gchar				*alternate_id;
 	gchar				*equivalent_id;
-	gchar				*physical_id;
 	gchar				*logical_id;
 	FuDevice			*alternate;
 	FuDevice			*parent;	/* noref */
@@ -90,7 +89,8 @@ fu_device_get_property (GObject *object, guint prop_id,
 		g_value_set_uint (value, priv->progress);
 		break;
 	case PROP_PHYSICAL_ID:
-		g_value_set_string (value, priv->physical_id);
+		g_value_set_string (value,
+				    fu_device_get_physical_id (self));
 		break;
 	case PROP_LOGICAL_ID:
 		g_value_set_string (value, priv->logical_id);
@@ -1706,8 +1706,8 @@ fu_device_set_version_bootloader (FuDevice *self, const gchar *version)
 gboolean
 fu_device_ensure_id (FuDevice *self, GError **error)
 {
-	FuDevicePrivate *priv = GET_PRIVATE (self);
 	g_autofree gchar *device_id = NULL;
+	gchar *physical_id = fu_device_get_physical_id (self);
 
 	g_return_val_if_fail (FU_IS_DEVICE (self), FALSE);
 	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
@@ -1717,7 +1717,7 @@ fu_device_ensure_id (FuDevice *self, GError **error)
 		return TRUE;
 
 	/* nothing we can do! */
-	if (priv->physical_id == NULL) {
+	if (physical_id == NULL) {
 		g_autofree gchar *tmp = fu_device_to_string (self);
 		g_set_error (error,
 			     G_IO_ERROR,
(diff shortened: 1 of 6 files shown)


superm1:
> To allow fwupd clients to better communicate which device is going to be
> updated to the user it may be required to know which physical device a
> fwupd device refers to. To that end expose the physical id that fwupd
> already keeps.
> 
> Signed-off-by: Sjoerd Simons [sjoerd@collabora.com](mailto:sjoerd@collabora.com)
> 
> Type of pull request:
> 
> * [ ]  New plugin (Please include [new plugin checklist](https://github.com/hughsie/fwupd/wiki/New-plugin-checklist))
> * [ ]  Code fix
> * [ x] Feature
> * [ ]  Documentation

Could you further elaborate what sort of thing the client would potentially reference?  In general I wouldn't think of physical IDs as user friendly strings to show to clients.  I think we would need another effort to make sure that all plugins using physical ID match to something in sysfs to be ready to do that.

libfwupdplugin/fu-device.c
```diff
@@ -1825,11 +1825,7 @@ fu_device_set_protocol (FuDevice *self, const gchar *protocol)
 void
 fu_device_set_physical_id (FuDevice *self, const gchar *physical_id)
 {
-	FuDevicePrivate *priv = GET_PRIVATE (self);
-	g_return_if_fail (FU_IS_DEVICE (self));
-	g_return_if_fail (physical_id != NULL);
-	g_free (priv->physical_id);
-	priv->physical_id = g_strdup (physical_id);
+	fwupd_device_set_physical_id (FWUPD_DEVICE (self), physical_id);
```
superm1, line 1828: I would recommend to instead change `fu-device.h` to map the function like other fu_* -> fwupd_* maps happen.

libfwupdplugin/fu-device.c
```diff
@@ -1848,9 +1844,7 @@ fu_device_set_physical_id (FuDevice *self, const gchar *physical_id)
 const gchar *
 fu_device_get_physical_id (FuDevice *self)
 {
-	FuDevicePrivate *priv = GET_PRIVATE (self);
-	g_return_val_if_fail (FU_IS_DEVICE (self), NULL);
-	return priv->physical_id;
+	return fwupd_device_get_physical_id (FWUPD_DEVICE (self));
```
superm1, line 1847: I would recommend to instead change `fu-device.h` to map the function like other fu_* -> fwupd_* maps happen.

src/fu-util-common.c
```diff
@@ -1252,6 +1252,12 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 		}
 	}
 
+	tmp = fwupd_device_get_physical_id (dev);
+	if (tmp != NULL) {
+		/* TRANSLATORS: Physical ID for hardware */
+		fu_common_string_append_kv (str, idt + 1, _("Physical ID"), tmp);
```
superm1, line 1258: Along the lines of my other comment if we are showing this to users, I want to make sure that we show something useful.  So depending on how this is looking for most devices, we might want to have a decoder function to walk sysfs and look for `name` attributes or to use udev to build some nice strings to present.

sjoerdsimons:
@superm1 I just opened issue #1995 to discuss the bigger issue.  Whether exposing physical id is the right solution here depends on that discussion (and indeed likely defining what it means).

Fwiw looking through the code i didn't spot any obvious place that relies on the exact content of the physical id

libfwupdplugin/fu-device.c
hughsie: I don't think we can do that, else we break plugin ABI as the symbol disappears. What Sjoerd did is right I think.

src/fu-util-common.c
hughsie: I think we need to nail down exactly what the physical ID **is** before we even think about displaying it.

libfwupdplugin/fu-device.c
superm1: Oh I see, good point.  I guess if/when plugin ABI breaks in the future that is's something we can remove then.

hughsie:
Maybe we close this now?

superm1:
I think so.  @sjoerdsimons if we determine this actually is the best API later can always re-open.

PR #1996, opened 2020-04-15
Wip/hughsie/cxaudio

hughsie:


dream-emily:
Issue:device has already updated and restarted, I still see following message looks like sth failed. While it can be enumerated by get-devices command and update again successfully later.

> failed to reload device: USB error on device 17ef:a396 : No such device (it may have been disconnected) [-4]

Verified. Fixed.

PR #2001, opened 2020-04-15
ccgx: Install both FW1 and FW2 firmwares

hughsie:
We have to do this client-side rather than using ANOTHER_WRITE_REQUIRED
as the device changes from `HID -> HPI-FW2 -> HPI->FW1 -> HID -> HPI-FW2` which
is somewhat complicated to handle in the daemon.

Using a 'client-side' retry loop also allows us to recover hub hardware when
one installing the correct device firmware on one device makes another device
appear, for instance making a shared SPI PD controller re-appear from a failed
hub upgrade.

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -45,6 +45,7 @@ typedef struct {
 	gchar				*daemon_version;
 	gchar				*host_product;
 	gchar				*host_machine_id;
+	gchar				*device_id_restart;
 	GDBusConnection			*conn;
 	GDBusProxy			*proxy;
 } FwupdClientPrivate;
@@ -210,6 +211,7 @@ fwupd_client_signal_cb (GDBusProxy *proxy,
 			GVariant *parameters,
 			FwupdClient *client)
 {
+	FwupdClientPrivate *priv = GET_PRIVATE (client);
 	g_autoptr(FwupdDevice) dev = NULL;
 	if (g_strcmp0 (signal_name, "Changed") == 0) {
 		g_debug ("Emitting ::changed()");
@@ -235,6 +237,13 @@ fwupd_client_signal_cb (GDBusProxy *proxy,
 		g_signal_emit (client, signals[SIGNAL_DEVICE_CHANGED], 0, dev);
 		g_debug ("Emitting ::device-changed(%s)",
 			 fwupd_device_get_id (dev));
+
+		/* save this in case we need to do ANOTHER_WRITE_REQUIRED */
+		if (fwupd_device_get_status (dev) == FWUPD_STATUS_DEVICE_RESTART &&
+		    g_strcmp0 (priv->device_id_restart, fwupd_device_get_id (dev)) != 0) {
+			g_free (priv->device_id_restart);
+			priv->device_id_restart = g_strdup (fwupd_device_get_id (dev));
+		}
 		return;
 	}
 	g_debug ("Unknown signal name '%s' from %s", signal_name, sender_name);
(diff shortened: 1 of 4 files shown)


hughsie:
I'm not sure about the `FWUPD_DEVICE_FLAG_ANOTHER_INSTALL_REQUIRED` name. Better ideas welcome!

superm1:
I'm a bit confused - typically don't you only want to write one bank so you can revert back to old FW in a failure scenario?

hughsie:
> typically don't you only want to write one bank

Yes, and you've uncovered a bug in my patch. In the case where the firmware is "symmetrical" we just flash the "other" bank and switch to that. In the case where the firmware is "asymmetric" the FW1 bank is a recovery partition that is only really useful for flashing FW2 (i.e. the hardware isn't functional to the user in FW1) -- but the catch is that FW1 needs to be updated to the same version as FW2 before it can be used to flash FW2. i.e. it's like a weird bootloader we update.

libfwupd/fwupd-client.c
```diff
@@ -1089,6 +1073,81 @@ fwupd_client_install (FwupdClient *client,
...
+		if (device == NULL) {
+			g_debug ("failed to find device: %s", error_local->message);
+			break;
+		}
+		if (!fwupd_device_has_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_INSTALL_REQUIRED))
+			break;
+	}
+
+	/* success */
+	return TRUE;
```
superm1, line 1150: I think you're missing a similar loop for fwupdtool too, no?

superm1:
> > typically don't you only want to write one bank
> 
> Yes, and you've uncovered a bug in my patch. In the case where the firmware is "symmetrical" we just flash the "other" bank and switch to that. In the case where the firmware is "asymmetric" the FW1 bank is a recovery partition that is only really useful for flashing FW2 (i.e. the hardware isn't functional to the user in FW1) -- but the catch is that FW1 needs to be updated to the same version as FW2 before it can be used to flash FW2. i.e. it's like a weird bootloader we update.

It seems preferable to keep that lifecycle in the plugin.  Like the "firmware version" of the device shouldn't be changed until all the components are changed.  Popping out and back in feels kinda hacky.

libfwupd/fwupd-client.c
```diff
@@ -235,6 +237,13 @@ fwupd_client_signal_cb (GDBusProxy *proxy,
 		g_signal_emit (client, signals[SIGNAL_DEVICE_CHANGED], 0, dev);
 		g_debug ("Emitting ::device-changed(%s)",
 			 fwupd_device_get_id (dev));
+
+		/* save this in case we need to do ANOTHER_WRITE_REQUIRED */
```
superm1, line 241: another install not write

libfwupd/fwupd-client.c
hughsie: Ya, I wanted some feedback on the general idea before sinking too much time into it :)

hughsie:
>  Popping out and back in feels kinda hacky.

Will have a rethink.

chlee75:
Richard,
I found something Strange

1) Remain time is strange on second FW write in fwupdmgr
   - Install duration is less than 1 min and remaining time is correct when a first firmware is writing such as "Less than one minute remaining..."
   - But Remaining time changes such as 67 min. -> 33 min. -> 5 min. -> etc. when a second firmware is writing. 
2) Additional Two Attach during two FW update .
   - [setup->write_fw->attach] --> [setup->attach] --> [setup->write_fw->attach] --> [setup->attach] 
   - There are two addtional [setup->attach]. Total four attach (reboot) happens during two FW update.

chlee75:
Richard.

Version mismatch issue happens in the case below.

FW2 (Primary) is running and fwupd update version (0.0.33), FW1 (0.0.33)/FW2 (0.0.33) 
1) Upgrade FW to 0.0.34 such as Run fwupdmgr install Lenovo-USBC-Gen2_Dock_v0.0.34.cab
2) Power off Gen2 Dock while FW1 is writing
3) Powe on Gen2 Dock 
4) FW2 (Primary) is running and fwupd update version (0.0.0), FW1 (0.0.0)/FW2 (0.0.33) 
   - FW1 version is 0.0.00 because FW1 is broken.
5) Upgrade FW to 0.0.34 such as Run fwupdmgr install Lenovo-USBC-Gen2_Dock_v0.0.34.cab
6) Only FW1 is updated and fwupmgr exit.
7) Run fwupdmgr get-devices : device version is 0.0.34 and GUID is FW1
   - In fwupd daemon,  FW1 (Backup) is running and fwupd update version (0.0.33), FW1 (0.0.34)/FW2 (0.0.33) 

At this time, fwupd daemon update version as 0.0.33 while version of fwupdmgr get-devices is 0.0.34.
Version mismatch happens and we can't upgrade FW2.

hughsie:
>  I found something Strange

Right, this branch is probably going to get thrown away without being merged. IT was proof of concept to find out what Mario thought (he hated it :)

I have a better idea I want to prototype this morning.

chlee75:
> Right, this branch is probably going to get thrown away without being merged. IT was proof of concept to find out what Mario thought (he hated it :)
> 
> I have a better idea I want to prototype this morning.

Well noted. Thanks.

superm1:
> (he hated it :)

I'd like to think my hate was justified by the above described experience that could happen by this happening outside of the plugin :)

hughsie:
> I'd like to think my hate was justified by the above

Dude, you were right to nak a hack. I'd have done the same to you :)

PR #2003, opened 2020-04-15
Allow devices to match the proxy device by GUID

hughsie:
This allows us to use a proxy even if the proxy device was created in a
different plugin or where the GTypes are not known to each other.

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -36,6 +36,7 @@ typedef struct {
 	gchar				*equivalent_id;
 	gchar				*physical_id;
 	gchar				*logical_id;
+	gchar				*proxy_guid;
 	FuDevice			*alternate;
 	FuDevice			*parent;	/* noref */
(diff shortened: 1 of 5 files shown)


hughsie:
I think I need this to do a couple of things in the future. I can expand if you like.

libfwupdplugin/fu-device.c
```diff
@@ -2200,6 +2242,8 @@ fu_device_add_string (FuDevice *self, guint idt, GString *str)
 		fu_common_string_append_kv (str, idt + 1, "LogicalId", priv->logical_id);
 	if (priv->proxy != NULL)
 		fu_common_string_append_kv (str, idt + 1, "Proxy", fu_device_get_id (priv->proxy));
```
superm1, line 2244: I think that you should change this to `ProxyID` or `ProxyDeviceID` to disambiguate now.

hughsie: Done.

PR #2004, opened 2020-04-15
build: Install systemd files relative to our prefix

jtojnar:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

-----

It is a good practice to install files relative to our own variables

bassi.io/articles/2018/03/15/pkg-config-and-paths

and it is required on systems like NixOS.

See also systemd/systemd@1c2c7c6

Note that on next systemd release, this will likely need to be changed:  https://github.com/systemd/systemd/commit/4908de44b0a0409f84a7cdc5641b114d6ce8ba03

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -22,6 +22,7 @@ option('plugin_modem_manager', type : 'boolean', value : false, description : 'e
 option('plugin_flashrom', type : 'boolean', value : false, description : 'enable libflashrom support')
 option('plugin_coreboot', type : 'boolean', value : true, description : 'enable coreboot support')
 option('systemd', type : 'boolean', value : true, description : 'enable systemd support')
+option('systemd_root_prefix', type: 'string', value: '', description: 'Directory to base systemd’s installation directories on (defaults to prefix)')
 option('systemdunitdir', type: 'string', value: '', description: 'Directory for systemd units')
 option('elogind', type : 'boolean', value : false, description : 'enable elogind support')
 option('tests', type : 'boolean', value : true, description : 'enable tests')
--- a/data/meson.build
+++ b/data/meson.build
@@ -92,7 +92,7 @@ if get_option('systemd') and build_daemon
     output : 'fwupd.shutdown',
     configuration : con2,
     install: true,
-    install_dir: systemd.get_pkgconfig_variable('systemdshutdowndir'),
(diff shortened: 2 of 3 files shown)


meson_options.txt
```diff
@@ -22,6 +22,7 @@ option('plugin_modem_manager', type : 'boolean', value : false, description : 'e
 option('plugin_flashrom', type : 'boolean', value : false, description : 'enable libflashrom support')
 option('plugin_coreboot', type : 'boolean', value : true, description : 'enable coreboot support')
 option('systemd', type : 'boolean', value : true, description : 'enable systemd support')
+option('systemd_root_prefix', type: 'string', value: '', description: 'Directory to base systemd’s installation directories on (defaults to prefix)')
```
hughsie, line 25: why not just prefix?

jtojnar: systemd allows to change this independently of `prefix` so we need to make it overridable too.

jtojnar:
Paths in `systemd.pc` on different systems:

[NixOS](https://github.com/NixOS/nixpkgs/blob/2da253a7ded5942283049b3a83388123adc27ab6/pkgs/os-specific/linux/systemd/default.nix#L74):

```
prefix=/nix/store/r59a9dwcwbbg10day7mwmfv94cff16i0-systemd-243.7
rootprefix=/nix/store/r59a9dwcwbbg10day7mwmfv94cff16i0-systemd-243.7
```

[Arch](https://www.archlinux.org/packages/core/x86_64/systemd/):

```
prefix=/usr
rootprefix=/usr
```

[Debian](https://packages.debian.org/sid/systemd)

```
prefix=/usr
rootprefix=
```

[Fedora](https://apps.fedoraproject.org/packages/systemd/)

```
prefix=/usr
rootprefix=/usr
```

meson_options.txt
jtojnar: Looks like Debian in fact sets to `/` so they would already need to override it here.

Alternately, we could install it to systemd root prefix by default and allow overriding it like we do with `systemdsystemunitdir`. In fact we could drop `systemdsystemunitdir` option since systemd hardcodes it to `${rootprefix}/lib/systemd/system`

jtojnar:
Closing in favour of #2006

PR #2006, opened 2020-04-15
build: Allow overriding prefix for systemd file installation

jtojnar:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

Replaces #2004

It is a good practice to install files relative to our own variables
bassi.io/articles/2018/03/15/pkg-config-and-paths
and it is required on systems like NixOS.

Unfortunately, systemd allows overriding the root prefix,
see also systemd/systemd@1c2c7c6,
so we cannot just do that.

Let's at least make the systemd installation prefix overridable.

This will also allow us to drop `systemdsystemunitdir` option since
systemd hardcodes it to `${rootprefix}/lib/systemd/system`.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -92,7 +92,7 @@ if get_option('systemd') and build_daemon
     output : 'fwupd.shutdown',
     configuration : con2,
     install: true,
-    install_dir: systemd.get_pkgconfig_variable('systemdshutdowndir'),
+    install_dir: systemd_shutdown_dir,
   )
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -349,11 +349,16 @@ if build_standalone and get_option('systemd')
   systemd = dependency('systemd', version : '>= 211')
   conf.set('HAVE_SYSTEMD' , '1')
   conf.set('HAVE_LOGIND' , '1')
-  systemdunitdir = get_option('systemdunitdir')
-  if systemdunitdir == '' and get_option('systemd')
+  systemd_root_prefix = get_option('systemd_root_prefix')
+  if systemd_root_prefix == ''
     systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
+    systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir')
+    systemd_shutdown_dir = systemd.get_pkgconfig_variable('systemdshutdowndir')
+  else
+    systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir', define_variable: ['rootprefix', systemd_root_prefix])
+    systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir', define_variable: ['rootprefix', systemd_root_prefix])
+    systemd_shutdown_dir = systemd.get_pkgconfig_variable('systemdshutdowndir', define_variable: ['rootprefix', systemd_root_prefix])
   endif
-  systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir')
 endif
 
 if build_standalone and get_option('elogind')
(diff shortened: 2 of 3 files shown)


jtojnar:
To overcome a [Meson issue](https://github.com/mesonbuild/meson/issues/6982), I have decided to choose a slightly more verbose solution:

```meson
systemd_root_prefix = get_option('systemd_root_prefix')
if systemd_root_prefix == ''
  systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
  systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir')
  systemd_shutdown_dir = systemd.get_pkgconfig_variable('systemdshutdowndir')
else
  systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir', define_variable: ['rootprefix', systemd_root_prefix])
  systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir', define_variable: ['rootprefix', systemd_root_prefix])
  systemd_shutdown_dir = systemd.get_pkgconfig_variable('systemdshutdowndir', define_variable: ['rootprefix', systemd_root_prefix])
endif
```

over a neater but rather hacky one:

```meson
systemd_root_prefix = get_option('systemd_root_prefix')
if systemd_root_prefix == ''
  systemd_root_prefix = systemd.get_pkgconfig_variable('rootprefix')
endif
if systemd_root_prefix == ''
  systemd_root_prefix = '""'
endif
systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir', define_variable: ['rootprefix', systemd_root_prefix])
systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir', define_variable: ['rootprefix', systemd_root_prefix])
systemd_shutdown_dir = systemd.get_pkgconfig_variable('systemdshutdowndir', define_variable: ['rootprefix', systemd_root_prefix])
```

superm1:
I think that `data/motd/meson.build` needs updating too no?

superm1:
> I think that `data/motd/meson.build` needs updating too no?

Oh maybe not, it's supposed to not be a shutdown unit.

jtojnar:
That seems to already use the variables.

hughsie:
> hughsie what do you think

I'm really confused why we need it. Surely the systemd pkgconfig file is saying "you need to use this directory" and then we're randomly doing something else?

jtojnar:
It is just a good practice to install files relative to our own variables, see
https://www.bassi.io/articles/2018/03/15/pkg-config-and-paths/

The assumption is that if user installs to a different prefix, they probably cannot install to the system one. For example:

* Developer installs the program to a local directory for development (might not work if fwupd is dependent on systemd to run it).
* On NixOS, every package is installed into its own immutable prefix so packages cannot write to the path listed in the systemd .pc file. We create a new directory tree linking all the services user enables and pass that to systemd from the top.

systemd explicitly supports this use case from 242: https://github.com/systemd/systemd/commit/1c2c7c6cb3d92315624f3711114e86b0acfbce63

hughsie:
> The assumption is that if user installs to a different prefix, they probably cannot install to the system one

So can we just use `prefix`?

jtojnar:
That would fail on Debian which has `prefix` ≠ `root_prefix`, see the CI failure in #2004.

jtojnar:
That would fail on Debian which has `prefix` ≠ `root_prefix`, see the CI failure in #2004.

hughsie:
> which has prefix ≠ root_prefix

Can you explain a little what the difference is? i.e. do we want a `root_prefix` option available in `meson` rather than changing all the upstream software?

jtojnar:
I think that is just systemd specific thing. Did not notice it in any other software I touched (freedesktop stack and almost all projects using meson).

jtojnar:
Looks like the `rootprefix` ≠ `prefix` is set by `split_usr` on Debian:

https://github.com/systemd/systemd/blob/9ba4029f8fdd88f1e774807cefa02491a535849b/meson.build#L78-L85

And from the linked PR, it looks like Exherbo sets it explicitly.

PR #2009, opened 2020-04-16
Do not skip attach() if the device is marked _WILL_DISAPPEAR

hughsie:
The idea here is that the device would not come back after it was restarted,
and skipping the attach in the engine was only working around the fact that the
ebitdo did not split out an ->attach() function.

We can't really blame it; we only decoupled the _IS_BOOTLOADER requirement
for ->attach() recently...

### DIFF
--- a/plugins/ebitdo/fu-ebitdo-device.c
+++ b/plugins/ebitdo/fu-ebitdo-device.c
@@ -511,6 +511,16 @@ fu_ebitdo_device_write_firmware (FuDevice *device,
 		return FALSE;
 	}
 
+	/* success! */
+	return TRUE;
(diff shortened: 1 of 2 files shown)


superm1:
I wonder if that also means that some places in the engine we were looking for `FWUPD_DEVICE_FLAG_WILL_DISAPPEAR` can go away too.  Did you look at those yet?

superm1:
> I wonder if that also means that some places in the engine we were looking for `FWUPD_DEVICE_FLAG_WILL_DISAPPEAR` can go away too. Did you look at those yet?

Actually disregard I think the others are still correct.

PR #2011, opened 2020-04-16
Allow a device to set the logical or physical ID during ->setup()

hughsie:
The device might not know what mode it is running in until after the device has
been opened.


hughsie:
Self tests FTW indeed.

PR #2012, opened 2020-04-16
Wip/hughsie/fixes

hughsie:


src/fu-engine.c
```diff
@@ -1455,6 +1455,8 @@ fu_engine_install_tasks (FuEngine *self,
 	devices = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
 	for (guint i = 0; i < install_tasks->len; i++) {
 		FuInstallTask *task = g_ptr_array_index (install_tasks, i);
+		g_debug ("composite update %u: %s", i + 1,
```
superm1, line 1458: I'd tihnk you only want to do this when `install_tasks->len` is > 1 though right?

hughsie: It was deliberate, to see how many tasks there were in the queue.

PR #2016, opened 2020-04-17
Wip/hughsie/ccgx fw1 fw2

hughsie:


plugins/ccgx/ccgx.quirk
```diff
@@ -9,9 +9,16 @@ Plugin = ccgx
...
+Name = ThinkPad USB-C Dock Gen2 PD Controller
 ImageKind = dual-asymmetric
 ParentGuid = USB\VID_17EF&PID_A391
 
+[DeviceInstanceId=USB\VID_04B4&PID_521A&SID_1F00&APP_6D64&MODE_FW1]
+Summary = PD-IC Backup Firmware
+
+[DeviceInstanceId=USB\VID_04B4&PID_521A&SID_1F00&APP_6D64&MODE_FW2]
+Summary = PD-IC Primary Firmware
+
```
chlee75, line 21: Could we change it such as CCGx PD Device (Primary Firmware) and CCGx PD Device (Backup Firmware)?

plugins/ccgx/ccgx.quirk
```diff
@@ -28,6 +35,7 @@ Plugin = ccgx
 GType = FuCcgxHpiDevice
 
 [DeviceInstanceId=USB\VID_04B4&PID_5218&SID_1F00&APP_6462]
+Name = ThinkPad USB-C Dock Hybrid PD Controller
+Summary = PD-IC Main Firmware
```
chlee75, line 39: Could we change summary such as CCGx PD Device (Main Firmware)?

chlee75:
I am testing this branch

plugins/ccgx/ccgx.quirk
hughsie: Sure! I guess we could expand out PD too -- maybe `CCGx Power Delivery Device (Primary Firmware)`?

chlee75:
Second FW writing remaining time do not match.
See below.

```
> Decompressing…           [***************************************]
> Authenticating…          [***************************************]
> Installing on ThinkPad USB-C Dock Gen2 PD Controller…            ]
> Restarting device…       [***************************************] Less than one minute remaining…
> Installing on ThinkPad USB-C Dock Gen2 PD Controller…*********** ]
> Restarting device…       [***                                    ] 37 minutes remaining…
```

hughsie:
> Second FW writing remaining time do not match

Ahh yes, I saw that too. That's going to be an unrelated fix, but it's one we should fix before merging this PR.

hughsie:
I think I have a fix for the time -- but I'm seeing this:

    Installing on ThinkPad USB-C Dock Gen2 PD Controller…            ]
    Restarting device…       [***************************************] Less than one minute remaining…
    Installing on ThinkPad USB-C Dock Gen2 PD Controller…************]
    Restarting device…       [***************************************] Less than one minute remaining…

The CLI uses the device name, and the FW1 and FW2 devices have the same name at the moment. Should we use the `(Main)` in the Name too? @chlee75

chlee75:
Two Devices are found on gnome-software.
Updating FW failed with error message when I select the device at the top.
Updating FW succeded when I select the device at the bottom.
I think only one device should be shown for updating.

Two devices on gnome-software.
image

Error message for the device at the top.
image

hughsie:
> Two Devices are found on gnome-software.

Grr. I'd not thought about that. Will have a think.

chlee75:
> The CLI uses the device name, and the FW1 and FW2 devices have the same name at the moment. Should we use the (Main) in the Name too? @chlee75 

Do you mean the name below?
  "ThinkPad USB-C Dock Gen2 PD Primary Controller" and "ThinkPad USB-C Dock Gen2 PD Backup Controller"

or 
  "ThinkPad USB-C Dock Gen2 PD Main Controller" and "ThinkPad USB-C Dock Gen2 PD Backup Controller"

Or 

  "ThinkPad USB-C Dock Gen2 PD Controller (Main)" and "ThinkPad USB-C Dock Gen2 PD Controller (Backup)"

Let me know your thought? @hughsie

plugins/ccgx/ccgx.quirk
chlee75: > CCGx Power Delivery Device (Primary Firmware)

It sounds good.

hughsie:
@chlee75 can you test with the branch again please? I've added a new flag that causes the proxy device to be updatable in the transaction, but not by itself.

Also, with the new names this is the much better output:


    Installing on ThinkPad USB-C Dock Gen2 PD Main Controller…       ]
    Restarting device…       [***************************************] Less than one minute remaining…
    Installing on ThinkPad USB-C Dock Gen2 PD Backup Controller…*****]
    Restarting device…       [***************************************] Less than one minute remaining…

chlee75:
> can you test with the branch again please? I've added a new flag that causes the proxy device to be updatable in the transaction, but not by itself.

Dinner time.
I will test it again after a meal. @hughsie

chlee75:
> Two Devices are found on gnome-software.
> Second FW writing remaining time do not match.
> CCGx Power Delivery Device (Primary Firmware)

Thress items have been fixed. @hughsie 

I will update stress test result (e.g power off/on during fw update)

src/fu-util-common.c
```diff
@@ -1090,6 +1090,10 @@ fu_util_device_flag_to_string (guint64 device_flag)
 		/* skip */
 		return NULL;
 	}
+	if (device_flag == FWUPD_DEVICE_FLAG_PROXY_UPDATABLE) {
+		/* TRANSLATORS: Device is updatable but via another deviced */
+		return _("Updatable (proxy)");
```
superm1, line 1095: I don't think end users need to know that it's a proxy update, to most users it should be the same either way.

libfwupd/fwupd-enums.h
```diff
@@ -104,6 +104,7 @@ typedef enum {
  * @FWUPD_DEVICE_FLAG_MD_SET_VERFMT:		Set the device version format from the metadata if available
  * @FWUPD_DEVICE_FLAG_ADD_COUNTERPART_GUIDS:	Add counterpart GUIDs from an alternate mode like bootloader
  * @FWUPD_DEVICE_FLAG_NO_GUID_MATCHING:		Force an explicit ID match when adding devices to the device list
+ * @FWUPD_DEVICE_FLAG_PROXY_UPDATABLE:		Device is updatable but only matched when a proxy device is used
```
superm1, line 107: I guess i'm not super sure about this.  Can't the plugin just only set the updateable flag when the `set_proxy` call is used?

superm1:
Everything else in the PR looks good, just those questions about the new flag.

src/fu-util-common.c
hughsie: Good point, will fix.

libfwupd/fwupd-enums.h
hughsie: Maybe it's not a very good name; this is supposed to mean "_updatable enough so that it passes the FuInstallTask checks, but not updatable enough that it shows up as a first class update in fwupdmgr or gnome-software_". The previous name I had for this was `UPDATABLE_HIDDEN`.

chlee75:
> I will update stress test result (e.g power off/on during fw update)

Case 1. Upgrade FW -> turn off/on docking power -> Upgrade FW : PASS.
Case 2. Downgrade FW -> turn off/on docking power -> Downgrade FW : PASS.

Case 3. Downgrade FW -> turn off/on docking power -> Upgrade FW : FAIL.
   - Primary (FW2) is running and FW1 ver(0.0.34)/FW2 ver(0.0.34) 
   - Downgrade FW to 0.0.33 
   - Turn off Dock Power when secoding FW is writing
   - Turn on Dock Power : FW1 is running,  FW1 ver(0.0.33)/ FW2 ve(0.0.0)
   - Upgrade FW to 0.0.34 
   -  Backup (FW1) FW is runing and FW1 ver(0.0.34)/FW2 ver(0.0.34) 

  There are no ways to go back to Primary(FW2) when Case3 happens even though run upgrading or downgrading or reinstalling FW.

 Could we make some workaround for case3?

hughsie:
> Could we make some workaround for case3?

Can you attach the verbose daemon logs for case 3 please?

chlee75:
> Can you attach the verbose daemon logs for case 3 please?

Is there any option to save fwupd log? 
I can't save log with  sudo fwupd | tee fwupd.log @hughsie

hughsie:
does this work? `sudo fwupd --verbose > log.txt` ?

chlee75:
> does this work? sudo fwupd --verbose > log.txt ?

do not work. try to find other way.

chlee75:
> do not work. try to find other way.

I attached case3 fwupd log. @hughsie 

[fwupd.log](https://github.com/fwupd/fwupd/files/4493911/fwupd.log)

libfwupd/fwupd-enums.h
superm1: Two more ideas:
1.  Use `FWUPD_DEVICE_FLAG_IGNORE_VALIDATION` for this purpose.
2. `FWUPD_DEVICE_FLAG_IGNORE_CHECKS`

hughsie:
Ohh, so I think what that log shows is:

 * Device in FW2
 * flash FW1
 * reboot into FW1
 * flash FW2 -- but remove before it's finished

then reinsert.

 * device starts in FW1
 * we flash FW2
 * we reboot into FW2
 * we flash FW1
 * **we reboot into FW1**

This would indicate 4adb7a11c20d987d63d71d845fe6994910322ddc isn't working correctly. I'll study the logs in more detail after dinner. I don't see the `CCGX device went away, so assuming no primary` anywhere in the logs which is suspicious.

chlee75:
> This would indicate 4adb7a1 isn't working correctly. I'll study the logs in more detail after dinner. I don't see the CCGX device went away, so assuming no primary anywhere in the logs which is suspicious.

I turn off docking power during seconding fw updating and turn on docking power.

hughsie:
> I turn off docking power during seconding fw updating and turn on docking power.

I see you unplugged at 15:46:29:

    15:46:29:0170 FuDeviceLocker       failed to close device: failed to clear events: read intr reg error: failed after 3 retries: write error: i2c get status error: failed to get i2c status: USB error on device 04b4:521a : No such device (it may have been disconnected) [-4]

This started a countdown:

> 15:46:29:0303 FuEngine             auto-removing GUsbDevice
> 15:46:29:0303 FuDeviceList         waiting 120000ms for ThinkPad USB-C Dock Gen2 PD Main Controller device removal

And then you re-inserted after a few seconds:

    15:46:33:0901 FuPlugin             emit added from vli: b0a78eb71f4eeea7df8fb114522556ba8ce22074

... and after a few more:

    15:46:37:0022 FuDevice             using 139e834982aea778526f37a982dd0bede328ccae for usb:01:00:01:03:03:01:FW2

If you'd have waited the 120 seconds (unlikely!) you would have got the device unregistered event which would have cause the `fu_plugin_device_removed` to be called which would have unset `priv->seen_primary`.

I'm not immediately sure how to fix this. Perhaps we want to watch a flag rather than wait for the `fu_plugin_device_removed` vfunc.

hughsie:
@chlee75  -- it seems FuDeviceList automatically unsets FWUPD_DEVICE_FLAG_UPDATABLE on instant unplug which is perfect for our needs. I'm in daddy mode, and am no longer in the office, but if you could test the unplug case 3 (with another verbose log if it fails) that would be great.

chlee75:
> it seems FuDeviceList automatically unsets FWUPD_DEVICE_FLAG_UPDATABLE on instant unplug which is perfect for our needs. I'm in daddy mode, and am no longer in the office, but if you could test the unplug case 3 (with another verbose log if it fails) that would be great.

I tested unplug case 3 and upgrade two times and finally FW1 is running.
I attached fwupd log. @hughsie 

[fwupd_2.log](https://github.com/fwupd/fwupd/files/4494732/fwupd_2.log)

For this case, 
 I think problem is that FW1 is running with FW1 version == FW2 version and it is very special case.
 So I set version to 0 in this case so that FW1 have to be updated forcely by FW upgrade.
 Then FW2 is running by FW update.
 below is my testing code.

```
		if (self->fw_image_type == FW_IMAGE_TYPE_DUAL_SYMMETRIC) {
			fu_ccgx_hpi_device_set_version_raw (self, versions[self->fw_mode]);
		} else if (self->fw_image_type == FW_IMAGE_TYPE_DUAL_ASYMMETRIC) {
			guint version_raw = versions[fu_ccgx_fw_mode_get_alternate (self->fw_mode)];
			if (self->fw_mode == FW_MODE_FW1 && versions[FW_MODE_FW1] == versions[FW_MODE_FW1])
				version_raw = 0x00 | self->fw_app_type;
			fu_ccgx_hpi_device_set_version_raw (self, version_raw);
			//fu_ccgx_hpi_device_set_version_raw (self, versions[fu_ccgx_fw_mode_get_alternate (self->fw_mode)]);
		}

```

hughsie:
> I tested unplug case 3

So I see `CCGX device went away, so assuming no primary` this time, so progress. I've merged a tiny fixup to fix the duplicated message.

I think the branch basically does what's expected. We could fix the FW mismatch but it's a pretty big change. Rather than swap around the versions like we do at the moment (e.g. `fu_ccgx_hpi_device_set_version_raw (self, versions[fu_ccgx_fw_mode_get_alternate (self->fw_mode)]);`) we could now set the version of the actual partition. If we that all the existing .cab files become invalid as the GUIDs would need switching around.

Is this something we want to do? Now seems like the last time we could make a change like this.

libfwupd/fwupd-enums.h
hughsie: I really want to describe the device, not describe what happens in the daemon. e.g. the device is updatable to the end user, but it doesn't need to manually update it as something else does. `UPDATABLE_HIDDEN` works in that idea, as does `UPDATABLE_PROXY` -- or `UPDATABLE_MASKED` or `UPDATABLE_PRIVATE` perhaps.

superm1: I think of those you proposed my preference lands on hidden.  And specifically the comments around the flag should talk about how clients shouldn't do anything with it.

Another thought though - what about adding splitting to two classes of flags?  One being private device flags that clients shouldn't do anything with and one that they should?

The transition might be a bit ugly since clients can access some of the private ones, but we could make new functions like `fwupd_device_get_public_flags` and deprecate the `fwupd_device_get_flags`

hughsie: UPDATABLE_HIDDEN seems most obvious to me too. Changed.

> Another thought though - what about adding splitting to two classes of flags

Yes, I'm aware we only have 64 bits to use too. I'll have a think for a future PR. 1.5.x stuff tho.

hughsie:
I've pushed new firmware in wip/hughsie/unswap and a test commit here. 100.0% untested.

chlee75:
Can't downgrade with message "device ID 139e834982aea778526f37a982dd0bede328ccae was not unique" @hughsie 

[fwupd_dev_id.txt](https://github.com/fwupd/fwupd/files/4502317/fwupd_dev_id.txt)

Reproduction.
1. Downgrade from 0.0.34 to 0.0.33
2. Turn off and on power during updating FW2
3. Upgrade to 0.0.34
4. Upgrade to 0.0.34 :  FWMode mismatch, expected FW1, got FW2
5. Downgrade from 0.0.31: can't downgrade with error message

chlee75:
Can't upgrade with error message "device ID e101551cc00c06b4bbdab33acafabfba622b0b4d was not found"

[fwupd_fw1_invalid_upgrade.log](https://github.com/fwupd/fwupd/files/4504058/fwupd_fw1_invalid_upgrade.log)

Pre-condition.
 -  FW2 is running with FW1 version (0.0.0) and FW2 version (0.0.34)
 - FW1 version is 0.0.0 because it is broken FW.

Upgrade FW to 0.0.34 and threre is Error message.

hughsie:
Ryan, I've had a think and I think we are doing this backwards. What we want to do is expose two devices for aysm devices, where the "other" device updates the current device and vice versa. This means we don't have to keep using the "alternate" `fw_mode` in random places.

I've pushed a branch called wip/hughsie/unswap to cypress-firmware.git and you can make the firmware files there using make like before. I've pushed a new experimental branch to fwupd.git called wip/hughsie/ccgx-fw1-fw2-unswap which shows more what I mean about unswapping the firmware.

Using the two branches the device behaves a lot more predictably. What I'm confused about is if you unplug while FW1 is flashing from FW2 then the device boots back up in FW1, when I suspect it should have booted in FW2 instead. I wonder if the address in `fu_ccgx_hpi_get_metadata_offset` are correct, but I didn't have the guts to change them.

Feedback welcome.

chlee75:
> I've pushed a branch called wip/hughsie/unswap to cypress-firmware.git and you can make the firmware files there using make like before. I've pushed a new experimental branch to fwupd.git called wip/hughsie/ccgx-fw1-fw2-unswap which shows more what I mean about unswapping the firmware.

Richard,
Actually we only need to update FW2 (Primary) and do not need to update FW1 (Backup) in Gen2 dock. Backup FW size is fixed and it is used only to update Primary FW.

If we update FW2 only, it will be more easy.

Normal case.
  1. FW2 is running, set FW2 version to device version.
  2. Jump to Alt mode if FW1 is valid. (4byte version is 0 if FW is not valid)
    (* Optionally Update FW1 if FW1 is invalid and Jump to Alt mode)
  3. Reboot 
  4. FW1 is running, update FW2 
  5. Reboot 

Abnormal case. (FW2 is broken)
  1. FW1 is running, FW2 is broken, set FW2 version to device version (0.0.0)
  2. update FW2 
  3. reboot

How about your though? @hughsie

hughsie:
> Actually we only need to update FW2 (Primary) and do not need to update FW1 (Backup) in Gen2 dock

This would have been good to know about 2 weeks ago...

hughsie:
> If we update FW2 only, it will be more easy.

Will this be true for all CCGX hardware? Are you now not planning to ship the FW1 and FW2 firmware in the cab file?

Did you try the branch I worked on most of today?

chlee75:
> This would have been good to know about 2 weeks ago...

Sorry for this inconvenience.
I recently checked Dock FW in detail for stress test.

chlee75:
> Will this be true for all CCGX hardware? 

It is true to Fixed Asymmetric FW which is Gen2 Dock.

We have 3 kinds of FW models.
 1) Symmetric FW
     - FW1 and FW2 size fixed. both FW1 and FW2 is Main FW.
 2) Fixed Asymmetric FW
     - FW1 (Backup) is fixed, FW2 is varied.
     - Update FW1 the FW2 or  FW2 only update cab be possible.
 3) Variable Asymmetric FW
     - FW1 and FW2 size are varied.
     - Always update FW1 then FW2.

chlee75:
> Are you now not planning to ship the FW1 and FW2 firmware in the cab file?

I am planning to ship FW1 and FW2 into cab file.
Sym and Asym FW always have FW1 and FW2.

chlee75:
> Did you try the branch I worked on most of today?

I really thank you for your hard work.
I have tested new branch and cab file.
While stress testing, FW1 only update case happen or FW2 only update case happen.
I can see FW1 (backup) is running after FW update.

I will share the logs for error case under stress test.

chlee75:
> Did you try the branch I worked on most of today?

> I will share the logs for error case under stress test.

I found some issue on wip/hughsie/ccgx-fw1-fw2-unswap. @hughsie 

Child version value is always 0.0.0 even though FW1 and FW2 version is 0.0.34.
In my stress test at night, I had strange results and I think it is because of 0.0.0 child version.
e.g. Updating FW is allowed always. 

[fwupd_unswap_keep_update.log](https://github.com/fwupd/fwupd/files/4507555/fwupd_unswap_keep_update.log)

"@setup() FW2 (0.0.34/0.0.34)" in log file is my own log to check FW versions.


```
├─ThinkPad USB-C Dock Gen2 PD Main Controller:
│ │ │   Device ID:         139e834982aea778526f37a982dd0bede328ccae
│ │ │   Summary:           CCGx Power Delivery Device (Main Firmware)
│ │ │   Current version:   0.0.34
│ │ │   Vendor:            Cypress Semiconductor (USB:0x04B4)
│ │ │   Install Duration:  1 minute
│ │ │   GUIDs:             2c4d586a-a2a6-5957-9427-768add598121 ← USB\VID_04B4&PID_521A&REV_0000
│ │ │                      ebd96940-fa49-5cb1-b1a1-aa2d4dea1416 ← USB\VID_04B4&PID_521A
│ │ │                      5108e726-44a6-5eca-bd1a-fbbdabbe7bc1 ← USB\VID_04B4&PID_521A&SID_1F00&APP_6D64
│ │ │                      31d41b9b-361a-50ad-8bea-8dad3339fa04 ← USB\VID_04B4&PID_521A&SID_1F00&APP_6D64&MODE_FW2
│ │ │   Device Flags:      • Requires AC power
│ │ │                      • Cryptographic hash verification is available
│ │ │                      • Device stages updates
│ │ │                      • Device can recover flash failures
│ │ │                      • Updatable
│ │ │ 
│ │ └─ThinkPad USB-C Dock Gen2 PD Backup Controller:
│ │       Device ID:       e101551cc00c06b4bbdab33acafabfba622b0b4d
│ │       Summary:         CCGx Power Delivery Device (Backup Firmware)
│ │       Current version: 0.0.0
│ │       Vendor:          Cypress Semiconductor (USB:0x04B4)
│ │       Install Duration:1 minute
│ │       GUIDs:           5108e726-44a6-5eca-bd1a-fbbdabbe7bc1 ← USB\VID_04B4&PID_521A&SID_1F00&APP_6D64
│ │                        f6b5141d-eeb0-5ff1-b8de-ee4f02b222b7 ← USB\VID_04B4&PID_521A&SID_1F00&APP_6D64&MODE_FW1
│ │       Device Flags:    • Updatable
│ │                        • Requires AC power
│ │                        • Supported on remote server
│ │                        • Cryptographic hash verification is available
│ │                        • Device stages updates
│ │                        • Device can recover flash failures

```

chlee75:
> Child version value is always 0.0.0

I found the root cause and fixed it.
- Before
```
fu_ccgx_hpi_device_set_version_raw (child, versions[child->fw_mode]);
fu_ccgx_hpi_device_set_silicon_id (child, self->silicon_id, NULL);
fu_ccgx_hpi_device_set_fw_mode (child, fu_ccgx_fw_mode_get_alternate (self->fw_mode));
```
- After
```
fu_ccgx_hpi_device_set_fw_mode (child, fu_ccgx_fw_mode_get_alternate (self->fw_mode));
fu_ccgx_hpi_device_set_version_raw (child, versions[child->fw_mode]);
fu_ccgx_hpi_device_set_silicon_id (child, self->silicon_id, NULL);
```
With fixed code, I did stress test and FW1 is running finally. @hughsie 
1. FW2 is running and FW1 and FW2 version is 0.0.34
2. Downgrade to 0.0.33 and turn off dock power during FW2 is wirtten.
3. Turn on dock power.
4. Upgrade  to 0.0.34 , FW2 is running after upgrade
5. Upgrade  to 0.0.34, FW1 is running after upgrade
6. Can't upgrade because FW1 and FW2 version is 0.0.34, But FW1 is running.

[fwupd_unswap_retest.log](https://github.com/fwupd/fwupd/files/4507619/fwupd_unswap_retest.log)

chlee75:
> If we update FW2 only, it will be more easy.

Most of Dock FW update in Cypress updates both FW1 and FW2.
Updating FW2 only is my idea to solve issues during stress test (power off/on).

As I know, Gen2 Dock Backup FW size is fixed. Primary FW is main FW.
I have been double checking my idea (FW2 update only) with Gen2 Dock FW owner.
I will let you know the result.

chlee75:
> I will let you know the result.

I got confirmation that FW2 update only is OK from Gen2 Dock FW owner. @hughsie 
Backup FW size is fixed and has only version changes.
I am really sorry for not catching this before.

I think it will take more time to figure out a way to solve issues happening while testing power off/on during FW1 and FW2 updating.
I think FW2 only updating could solve these issues (Not go to FW2 after updating FW).
It is because that we only update FW2 and we can update FW2 when FW2 is broken (fw2 version is 0.0.0 in this case in FW1) .
And Lenovo keeps pushing plugin schedule.

What do you think of applying FW2 update only for a Fixed Asymmetric FW type (Gen2 Dock Case)? 
And For a Variable Asymmetric FW type (FW1 and FW2 update), Could we debug it more later?

Technically I don't know how to make FW2 only plugin and cab file because we need to jump to FW1 to update FW2.

I really appreciate it if you make a new branch for updating FW2 only. 
I will do my best to validate it and give feedback to you to make it stable.
For cab file of FW2 update only, I think we need only FW2 firmware.

Please give me your suggestion.

hughsie:
> I think FW2 only updating could solve these issues

Well, it's certainly much simpler as it matches what all the other devices we support in fwupd do.

I've created a new PR on https://github.com/fwupd/fwupd/pull/2026 which should work with the updated wip/hughsie/unswap branch in the firmware repo that only includes the FW2 firmware in the archive. It works for me when installing from FW1 and FW2 even during unplug and recovery.

>  I will do my best to validate it and give feedback to you to make it stable.

I am busy for the next few days with another OEM but please let me know how you get on.

chlee75:
> I am busy for the next few days with another OEM but please let me know how you get on.

I appreciate it.
I am working on testing the fw2-only branch and cab file.

PR #2018, opened 2020-04-17
Wip/hughsie/inherit quirks

hughsie:


libfwupdplugin/fu-device.c
```diff
@@ -660,6 +660,13 @@ fu_device_set_parent (FuDevice *self, FuDevice *parent)
 		fu_device_set_order (parent, fu_device_get_order (self) + 1);
 	}
 
+	/* if the parent has quirks, make the child inherit it */
+	if (parent != NULL) {
+		if (fu_device_get_quirks (self) == NULL &&
+		    fu_device_get_quirks (parent) != NULL)
+			fu_device_set_quirks (self, fu_device_get_quirks (parent));
```
superm1, line 667: Is this really true that all parents quirks should go to children?  I can't think it would be. What if the quirks get loaded for the parent before the child but the child is supposed to be loading some from a quirk file?  This seems like the sort of thing that should be guarded by a flag indicating it's the desired behavior.

hughsie: It's not the quirks themselves, it's the FuQuirk object that allows the quirks to be applied on the child device.

superm1: oh.... I see OK.

PR #2021, opened 2020-04-18
vli: Use GPIOB to reset the VL817 found in two Lenovo products

hughsie:
If these devices really are shared SPI then we don't need to redirect the attach to another device and can directly handle the quirk in the usb hub device.  tl;dr: I over-thought the problem.

### DIFF
--- a/plugins/vli/fu-vli-usbhub-device.c
+++ b/plugins/vli/fu-vli-usbhub-device.c
@@ -325,10 +325,47 @@ fu_vli_usbhub_device_spi_write_data (FuVliDevice *self,
 	return TRUE;
 }
 
+#define VL817_ADDR_GPIO_OUTPUT_ENABLE		0xF6A0	/* 0=input, 1=output */
+#define VL817_ADDR_GPIO_SET_OUTPUT_DATA		0xF6A1	/* 0=low, 1=high */
+#define VL817_ADDR_GPIO_GET_INPUT_DATA		0xF6A2	/* 0=low, 1=high */
+
+static gboolean
+fu_vli_usbhub_device_attach_vl817_gpiob (FuDevice *device, GError **error)
+{
+	FuVliUsbhubDevice *self = FU_VLI_USBHUB_DEVICE (device);
+	guint8 tmp = 0x0;
+
+	/* set GPIOB output enable */
+	if (!fu_vli_usbhub_device_read_reg (self, VL817_ADDR_GPIO_OUTPUT_ENABLE,
+					    &tmp, error))
+		return FALSE;
+	if (!fu_vli_usbhub_device_write_reg (self, VL817_ADDR_GPIO_OUTPUT_ENABLE,
+					     tmp | (1 << 1), error))
+		return FALSE;
+
+	/* toggle GPIOB to trigger reset */
+	if (!fu_vli_usbhub_device_read_reg (self, VL817_ADDR_GPIO_SET_OUTPUT_DATA,
+					    &tmp, error))
+		return FALSE;
+	if (!fu_vli_usbhub_device_write_reg (self, VL817_ADDR_GPIO_SET_OUTPUT_DATA,
+					     tmp ^ (1 << 1), error))
+		return FALSE;
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
+	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+	return TRUE;
+}
+
 static gboolean
 fu_vli_usbhub_device_attach (FuDevice *device, GError **error)
 {
 	g_autoptr(GError) error_local = NULL;
+
+	/* the proxy might be using a GPIO instead */
+	if (fu_device_get_proxy (device) != NULL) {
+		FuDevice *proxy = fu_device_get_proxy (device);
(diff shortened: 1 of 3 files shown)


plugins/vli/fu-vli-usbhub-device.c
```diff
@@ -1006,12 +1036,25 @@ fu_vli_usbhub_device_write_firmware (FuDevice *device,
...
 
+static void
+fu_vli_usbhub_device_kind_changed_cb (FuVliDevice *device, GParamSpec *pspec, gpointer user_data)
+{
+	FuDeviceClass *klass_device = FU_DEVICE_GET_CLASS (device);
+	if (fu_vli_device_get_kind (device) == FU_VLI_DEVICE_KIND_VL817 &&
+	    fu_device_has_custom_flag (FU_DEVICE (device), "attach-with-gpiob"))
+		klass_device->attach = fu_vli_usbhub_device_attach_vl817_gpiob;
+}
+
```
memily, line 1047: We think the vl817 default attach should be fu_vli_device_attach and if it has the "attach-with-gpiob" flag then evoke "fu_vli_usbhub_device_attach_vl817_gpiob". Otherwise downport hubs without share spi will not attach properly under parent; see Proj 08 Mini Dock.

hughsie: That's what I did, no? The default attach handler is set up in `_init()`.

memily: During testing we're running into the issue of fu_vli_usbhub_device_attach_vl817_gpiob() trying to attach with VL211 and not VL817.

plugins/vli/fu-vli-usbhub-device.c
```diff
@@ -325,6 +325,36 @@ fu_vli_usbhub_device_spi_write_data (FuVliDevice *self,
 	return TRUE;
 }
 
+#define VL817_ADDR_GPIO_OUTPUT_ENABLE		0x60A0	/* 0=input, 1=output */
+#define VL817_ADDR_GPIO_SET_OUTPUT_DATA		0x60A1	/* 0=low, 1=high */
+#define VL817_ADDR_GPIO_GET_INPUT_DATA		0x60A2	/* 0=low, 1=high */
```
memily, line 330:
```c++
#define VL817_ADDR_GPIO_OUTPUT_ENABLE       0xF6A0  /*0=input, 1=output*/ 
#define VL817_ADDR_GPIO_SET_OUTPUT_DATA     0xF6A1  /*0=low, 1=high */ 
#define VL817_ADDR_GPIO_GET_INPUT_DATA      0xF6A2  /*0=low, 1=high*/ 
```

hughsie: Fixed, thanks.

plugins/vli/fu-vli-usbhub-device.c
hughsie: `write VL817 -> write VL103 -> vl817_gpiob` is what I'm trying to do here as a first step -- the `write VL211 -> attach VL211 -> vl817_gpiob` is still WIP.

plugins/vli/vli-usbhub-lenovo.quirk
```diff
@@ -167,7 +167,13 @@ CounterpartGuid = USB\VID_17EF&PID_3095
...
 Plugin = vli
 GType = FuVliUsbhubDevice
-Flags = usb2,has-shared-spi-pd,attach-with-gpiob
+Flags = usb2
+ParentGuid = USB\VID_17EF&PID_3094
+[DeviceInstanceId=USB\VID_17EF&PID_3097]
+Plugin = vli
+GType = FuVliUsbhubDevice
+Flags = usb2
+ParentGuid = USB\VID_17EF&PID_3095
```
hughsie, line 176: so this is a 2nd tier device? Why remove the other flags?

plugins/vli/fu-vli-usbhub-device.c
memily: Makes sense.

plugins/vli/vli-usbhub-lenovo.quirk
memily: VL211 is a 2nd tier device that only has a U2 port and does not share spi.

hughsie: But the VL817-Q7 still shares an SPI with the VL103-Q4, no?

memily: Oh, I understand your question now. Yes, VL817-Q7 usb2 can have the same flags "has-shared-spi-pd,attach-with-gpiob" as usb3. I didn't explain that our convention for VL817-Q7 usb2 is debugging only. As in, the end client will not update FW through VL817-Q7 usb2. The design choice is up to you, either way works!

plugins/vli/fu-vli-usbhub-device.c
hughsie: I've pushed a 100% untested patch to make the `write VL211 -> attach VL211 -> vl817_gpiob` sequence work.

hughsie: I've tested this, with the GPIOB toggle it seems to work on all firmware versions now.

PR #2022, opened 2020-04-20
Allow specifying the device on the command line by GUID

hughsie:
The GUID is the only stable identifier, and allowing using the GUID makes it
much easier to test specific devices.

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -470,6 +470,60 @@ fwupd_client_get_device_by_id (FwupdClient *client,
 }
 
 /**
+ * fwupd_client_get_devices_by_guid:
+ * @client: A #FwupdClient
+ * @guid: the GUID, e.g. `e22c4520-43dc-5bb3-8245-5787fead9b63`
+ * @cancellable: the #GCancellable, or %NULL
+ * @error: the #GError, or %NULL
+ *
+ * Gets any devices that provide a specific GUID. An error is returned if no
+ * devices contains this GUID.
+ *
+ * Returns: (element-type FwupdDevice) (transfer container): devices or %NULL
+ *
+ * Since: 1.4.1
+ **/
+GPtrArray *
+fwupd_client_get_devices_by_guid (FwupdClient *client,
+				  const gchar *guid,
+				  GCancellable *cancellable,
+				  GError **error)
+{
+	g_autoptr(GPtrArray) devices = NULL;
+	g_autoptr(GPtrArray) devices_tmp = NULL;
+
+	g_return_val_if_fail (FWUPD_IS_CLIENT (client), NULL);
+	g_return_val_if_fail (guid != NULL, NULL);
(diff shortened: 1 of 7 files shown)


sumitopensource:
Thanks Richard for considering this enhancement request. It would really be helpful.

libfwupd/fwupd-client.c
```diff
@@ -469,6 +469,65 @@ fwupd_client_get_device_by_id (FwupdClient *client,
...
+		if (fwupd_device_has_guid (dev_tmp, guid)) {
+			if (dev == NULL) {
+				dev = g_object_ref (dev_tmp);
+				continue;
+			}
+			g_set_error (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "multiple devices provided %s", guid);
+			return NULL;
```
superm1, line 517: I think specifically I would be happier with the client library returning an array of matched devices.  Then the command line frontends can use this for the prompting list, or adopt a behavior that is "update all" or similar.

src/fu-engine.c
```diff
@@ -3534,6 +3534,27 @@ fu_engine_get_device (FuEngine *self, const gchar *device_id, GError **error)
...
+ * Gets a specific device.
+ *
+ * Returns: (transfer full): a device, or %NULL if not found
+ **/
+FuDevice *
+fu_engine_get_device_by_guid (FuEngine *self, const gchar *guid, GError **error)
+{
+	FuDevice *device;
+
+	device = fu_device_list_get_by_guid (self->device_list, guid, error);
```
superm1, line 3552: This makes an assumption there is only one device (it will potentially cause the wrong behavior if multiple are plugged in)

libfwupd/fwupd-client.c
hughsie: Good idea.

src/fu-engine.c
hughsie: Fixed I think.

djcampello:
I'm getting the following while trying on master:

# fwupdtool update cd357cf1-40b2-5d87-b8df-bb2dd82774aa
device ID cd357cf1-40b2-5d87-b8df-bb2dd82774aa was not found

superm1:
@djcampello 

I just tried myself and caught two simple bugs in the `fwupdtool` implementation.  They were one line fixes, I just pushed to master.  Can you have another try?

djcampello:
Thanks, I had also hit the infinite loop. Now it seems to work!

djcampello:
On the other hand, I think other commands (than update) are still only using device ID...

# fwupdtool firmware-read file 8ceeeffd-51b6-580c-9b75-69143227aff
device ID 8ceeeffd-51b6-580c-9b75-69143227aff was not found

superm1:
I can't reproduce that.  I actually checked with a WD19TB too.

```
$ sudo ./build/src/fwupdtool firmware-read file 8ceeeffd-51b6-580c-9b75-69143227aff8
Selected device: Package level of Dell dock
not supported
```

djcampello:
You are right... I was missing a character

PR #2024, opened 2020-04-20
Fix the DeviceID set by GetDetails

hughsie:
The returned ID is the result of the SHA1 hash of the actual device ID. This
does not match anything found by the client, and so the install fails.

The symbol is exported as I think the device ID is an important identifier and
used in various fwupd tools.

When backported to the stable branch the verification should just be a static
function in src/fu-engine.c rather than a new symbol.

### DIFF
--- a/libfwupd/fwupd-common.c
+++ b/libfwupd/fwupd-common.c
@@ -818,6 +818,34 @@ fwupd_guid_hash_data (const guint8 *data, gsize datasz, FwupdGuidFlags flags)
 }
 
 /**
+ * fwupd_device_id_is_valid:
+ * @device_id: string to check, e.g. `d3fae86d95e5d56626129d00e332c4b8dac95442`
+ *
+ * Checks the string is a valid non-partial device ID. It is important to note
+ * that the wildcard ID of `*` is not considered a valid ID in this function and
+ * the client must check for this manually if this should be allowed.
+ *
+ * Returns: %TRUE if @guid was a fwupd device ID, %FALSE otherwise
+ *
+ * Since: 1.4.1
+ **/
+gboolean
+fwupd_device_id_is_valid (const gchar *device_id)
+{
+	if (device_id == NULL)
+		return FALSE;
+	if (strlen (device_id) != 40)
+		return FALSE;
+	for (guint i = 0; device_id[i] != '\0'; i++) {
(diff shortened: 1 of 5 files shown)


libfwupd/fwupd-common.c
```diff
@@ -817,6 +817,34 @@ fwupd_guid_hash_data (const guint8 *data, gsize datasz, FwupdGuidFlags flags)
...
 }
 
+/**
+ * fwupd_device_id_is_valid:
+ * @device_id: string to check, e.g. `d3fae86d95e5d56626129d00e332c4b8dac95442`
+ *
+ * Checks the string is a valid non-partial device ID. It is important to note
+ * that the wildcard ID of `*` is not considered a valid ID in this function and
+ * the client must check for this manually if this should be allowed.
+ *
```
superm1, line 827: Why not have a g_return_val_if_fail in this case to prevent coding errors?

libfwupd/fwupd-common.c
```diff
@@ -817,6 +817,34 @@ fwupd_guid_hash_data (const guint8 *data, gsize datasz, FwupdGuidFlags flags)
...
+{
+	if (device_id == NULL)
+		return FALSE;
+	if (strlen (device_id) != 40)
+		return FALSE;
+	for (guint i = 0; device_id[i] != '\0'; i++) {
+		gchar tmp = device_id[i];
+		/* isalnum isn't case specific */
+		if ((tmp < 'a' || tmp > 'f') && (tmp < '0' || tmp > '9'))
+			return FALSE;
```
superm1, line 843: Any particular reason not to use `g_ascii_strdown`?

hughsie: That dupes the string. I'm actually explicitly testing for not uppercase text, as that wouldn't be matched by the daemon -- which would be confusing.

superm1: OK...

libfwupd/fwupd-common.c
hughsie: Good question, but that doesn't seem right. I see this symbol being used for "is a fully specified ID" so I think it's perfectly fine to call it on `*` and get FALSE.

superm1: OK.

PR #2026, opened 2020-04-21
ccgx: Only update the FW2 partition of the ThinkPad USB-C Dock Gen2

hughsie:


plugins/ccgx/ccgx.quirk
```diff
@@ -10,15 +10,16 @@ GType = FuCcgxHpiDevice
 
 [DeviceInstanceId=USB\VID_04B4&PID_521A&SID_1F00&APP_6D64]
 ImageKind = dual-asymmetric
+Name = ThinkPad USB-C Dock Gen2 PD
```
chlee75, line 13: "ThinkPad USB-C Dock Gen2 PD Controller" is better to get align with Hybrid Dock Name: "ThinkPad USB-C Dock Hybrid PD Controller"

hughsie: Fixed.

hughsie:
> No need "enter_alt_mode" in fu-ccgx-hpi-device.c

Agreed, removed. Apart from those niggles it works?

chlee75:
I can't see any issue during upgrading/downgrading aging testing and unplug/plug test in offline mode with cab files.
I think it seems to be stable.
What do you think of it? @hughsie 

If you agree on it, I will ask sumit to upload new cab file for Testing State.
Also request customer test new plugin and cab file.

hughsie:
>  What do you think of it? @hughsie

I'm happy to merge this now. I'll demote the existing firmware as it really won't work with the new GUIDs. I can delete it too if that helps. Sumit will have to upload new firmware.

chlee75:
> I'll demote the existing firmware as it really won't work with the new GUIDs. I can delete it too if that helps. Sumit will have to upload new firmware.

Do you mean that you will delete existing cab files (Testing or Embargo) in LVFS for Gen2 Dock?
@hughsie

hughsie:
> Do you mean that you will delete existing cab files

We have to take them out of testing, otherwise anyone running fwupd git master gets the wrong FW installed on the device now. I've put them back in your private remote.

PR #2027, opened 2020-04-21
Make tss2-esys conditional

hughsie:
This is not provided in RHEL 8 in the default build root.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -219,6 +219,10 @@ libgcab = dependency('libgcab-1.0', version : '>= 1.0', fallback : ['gcab', 'gca
 gcab = find_program('gcab', required : true)
 bashcomp = dependency('bash-completion', required: false)
 python3 = find_program('python3')
+tpm2tss = dependency('tss2-esys', version : '>= 2.0', required: false)
+if tpm2tss.found()
+  conf.set('HAVE_TSS2', '1')
+endif
 
 platform_deps = []
 if get_option('default_library') != 'static'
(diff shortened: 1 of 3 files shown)


superm1:
What about `plugin_tpm`?  Doesn't that also use esys?

meson.build
```diff
@@ -280,7 +280,11 @@ if build_standalone and get_option('plugin_uefi')
   efiboot = dependency('efiboot')
   objcopy = find_program ('objcopy')
   readelf = find_program ('readelf')
-  tpm2tss = dependency('tss2-esys', version : '>= 2.0')
+  tpm2tss = dependency('tss2-esys', version : '>= 2.0', required: false)
```
superm1, line 283: I'm a little worried this could lead to an undefined situation.  Depending on what optimization meson does before evaluating `-Dplugin_tpm=true` and missing tsys could lead to a pass potentially.

So Perhaps lines 270-271 should instead show an error if it's missing?

hughsie:
> What about plugin_tpm

You can turn that off with `-Dplugin_tpm=false` and live quite happily.

meson.build
hughsie: Maybe it might be cleaner if we did the `required: false` above both checks, and then handled the unfound case as a failure if the TPM plugin is enabled.

hughsie: I've pushed a new version that does this. Until tomorrow dude.

hughsie:
>  Are there more things needed beyond this

There's no libjcat in RHEL 8 either. I've burned my karma for 8.3 getting libxmlb in (also needed for gnome-software) so libjcat should stabilize and then be in later minor releases I guess. In the meantime I've bundled it in fwupd.src.rpm. I'll prep some other PRs that should allow a centos 8 CI in the next few days.

Issue #2034, opened 2020-04-28
Changelog and releases


the-spyke:
**Is your feature request related to a problem? Please describe.**

No.

**Describe the solution you'd like**

Please add `CHANGELOG.md` file with history of changes for releases and/or post changes to GutHub releases. The first one is a common practice. The second one is too, but more importantly it is convenient: you can subscribe to release notifications on GitHub, so you want to see the changes in the release notification email.

**Describe alternatives you've considered**

Looking for changes in commit log or mail lists is great, but you can do better.

**Additional context**

An example of good changelog/releases: https://github.com/babel/babel

hughsie:
Would a NEWS file be sufficient? Generating it is already part of the release process, but it's just not checked into git...

the-spyke:
I think `CHANGELOG.md` is a more established name, but at least `NEWS` will be something. Nowadays some projects discard using a file and just post changes to GitHub Releases. I'm fine with this too.

the-spyke:
@hughsie `1.4.2` -- still no notes in GitHub releases and I haven't found the NEWS file.

superm1:
It looks like this automation failed somehow.  We'll have to dig into why.

superm1:
Maybe because we didn't backport a1fda2a179679a0377035321c03c62151c3b8b0d to 1_4_X

superm1:
backport to 1_4_X: 11b7d4dd384a97266c54fa4337014eb741456728.  1.4.3 and 1.5.0 should have it.

superm1:
Just FYI; it looks like automatic changelogs are working properly now.  They're showing in the recent 1.5.5 release:
https://github.com/fwupd/fwupd/releases/tag/1.5.5

PR #2035, opened 2020-04-28
trivial: Include the NEWS file in the repo

hughsie:
Fixes: https://github.com/fwupd/fwupd/issues/2034

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,0 +1,1353 @@
+Version 1.4.1
+~~~~~~~~~~~~~
+Released: 2020-04-27
+
+This release adds the following features:
+ * Allow adding a device 'proxy' device that can do actions on it
+ * Allow specifying the device on the command line by GUID
+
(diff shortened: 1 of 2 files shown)


hughsie:
It's possibly the largest _trivial_ change in the repo, but you get the idea :)

superm1:
I think it's a waste of space to distribute (essentially) the same changelog in two places in the tarball.  So I would counter two suggestions:

1. Distribute NEW as a symlink to `data/org.freedesktop.fwupd.metainfo.xml`.  I'm aware it's a different file format, but the information is all there.
2. Use `ghr` to publish the body to Github releases when the Windows EXE is published on tagging.  I checked and see there is `-b` argument for it: https://github.com/tcnksm/ghr.

superm1:
counter proposal in https://github.com/fwupd/fwupd/pull/2036

PR #2036, opened 2020-04-28
trivial: potentially publish changelog to github release

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,7 +165,8 @@ jobs:
           command: |
             go get github.com/tcnksm/ghr
             VERSION=$(cat dist/VERSION)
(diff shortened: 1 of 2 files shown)


hughsie:
I guess this works, but I don't think it's hugely useful. I think building the text we publish as the changelog would be much better being generated from the metainfo.xml file. If we commit something, then revert it, then commit another version of it, we'd get three weird lines in the release notes. The metainfo list is also a deliberate "interesting subset" which is probably more useful to random people.

superm1:
> I guess this works, but I don't think it's hugely useful. I think building the text we publish as the changelog would be much better being generated from the metainfo.xml file. If we commit something, then revert it, then commit another version of it, we'd get three weird lines in the release notes. The metainfo list is also a deliberate "interesting subset" which is probably more useful to random people.

OK, adjusted

PR #2041, opened 2020-04-29
trivial: dell_dock: stop using custom flag skip-restart

superm1:
Instead switch to usable-during-update, which implies the same
behavior.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/synaptics-mst/synaptics-mst.quirk
```diff
@@ -16,7 +16,7 @@ Plugin = synaptics_mst
 #  * These GUIDs will look like MST-${DEVICEKIND}-${CHIPID}-${BOARDID}
 #
 # By default the Synaptics MST device will restart after update
-# To override this behavior add the custom flag "skip-restart"
+# To override this behavior add FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE
```
hughsie, line 19: maybe `add the flag usable-during-update` might be clearer

hughsie:
> Instead switch to usable-during-update

Is this the same thing? "Usable during update" means you can carry on using the device as it's being updated (e.g. you're running partition A, and writing to partition B) but I don't think it's implies "don't rebooth to switch to the new mode".

superm1:
> > Instead switch to usable-during-update
> 
> Is this the same thing? "Usable during update" means you can carry on using the device as it's being updated (e.g. you're running partition A, and writing to partition B) but I don't think it's implies "don't rebooth to switch to the new mode".

Well, I originally was going to add a new `SKIPS_RESTART` flag.  I then thought about it more and can't think of a single device that has to go down (at least momentarily) to complete it's update.  So the thought process is this happens when the user isn't using the machine:
* Scheduling UEFI updates (uefi)
* Updates that happen automatically when power cycling (dell-dock, nvme)
* Updates that happen during activation (dell-dock, ata, synaptics mst)

plugins/synaptics-mst/synaptics-mst.quirk
superm1: fixed

superm1:
> (e.g. you're running partition A, and writing to partition B)

Isn't that `FWUPD_DEVICE_FLAG_DUAL_IMAGE`

hughsie:
> can't think of a single device that doesn't have to go down (at least momentarily) to complete it's update

Ha, I guess you're right.

> Usable during update

I think semantically this means you can carry on using the computer while the update is applying; for instance if a dock update took 5 minutes but you can carry on using the computer and then reboot the computer to complete that's fine. I don't think it means "will not reboot". I kinda see these flags as hints to the UI so we could say "it's okay to keep using your computer while the update is being applied" rather than saying something like "the device may be in a degraded state while the update is being applied".

> SKIPS_RESTART

I honestly think this is the least magic and most straightforward thing to use. I'm not sure it's terribly useful to show to the user in the fwupdmgr output, but becomes another "internal" flag we use.

superm1:
> I don't think it means "will not reboot". I kinda see these flags as hints to the UI 

The part about this that is stuck with me is that even if the device is usable for 99.9% of the time the update is installed from a fwupd client but there is .1% that it isn't, that flag shouldn't be on the device.  I've been equating it in my mind that the update has been installed/staged/prepared such that it will be finalized with no downtime.

I was thinking this flag when a client used it can be as one piece of the hints that if your security policy allows to install the update in the background automatically.

> so we could say "it's okay to keep using your computer while the update is being applied" rather than saying something like "the device may be in a degraded state while the update is being applied".

And that's what we're using in the command line clients.
`%s and all connected devices may not be usable while updating`

superm1:
OK i've come around to this, adjusted PR.

PR #2050, opened 2020-05-01
Wip/hughsie/hsi

hughsie:


hughsie:
This is the rough output from `fwupdmgr` and `fwupdtool`:

    $ ./src/fwupdmgr get-hsi
      AppstreamId:          org.fwupd.Hsi.Updates
      Flags:                success|runtime-updates
      Name:                 Firmware Supported
      Summary:              There is firmware less than 12 months old

      AppstreamId:          org.fwupd.Hsi.Attestation
      Flags:                success|runtime-attestation
      Name:                 Firmware Attestation
      Summary:              Firmware has attestation checksums

      AppstreamId:          org.trustedcomputinggroup.TpmEventLog
      HsiNumber:            2
      Flags:                success
      Name:                 TPM Reconstruction
      Summary:              TPM event log matched PCR0 reading

      AppstreamId:          com.uefi.SecureBoot
      HsiNumber:            1
      Flags:                success
      Name:                 UEFI Secure Boot
      Summary:              Enabled

    Calculated Host Security ID: HSI:2+UA

There's also the raw JSON data from `fwupdagent` too.

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	gchar				*appstream_id;
+	GPtrArray			*obsoletes;
+	gchar				*name;
+	gchar				*summary;
+	gchar				*uri;
+	guint32				 number;
+	FwupdHsiAttrFlags		 flags;
+} FwupdHsiAttrPrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (FwupdHsiAttr, fwupd_hsi_attr, G_TYPE_OBJECT)
```
hughsie, line 36: I'm not 100% about this object name if you have any better ideas...

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -112,6 +113,7 @@ fu_plugin_device_registered (FuPlugin *plugin, FuDevice *device)
 		for (guint j = 0; j < data->pcr0s->len; j++) {
 			const gchar *checksum_tmp = g_ptr_array_index (data->pcr0s, j);
 			if (g_strcmp0 (checksum, checksum_tmp) == 0) {
+				data->reconstructed = TRUE;
 				g_debug ("TPM reconstructed event log matched PCR0 reading");
```
superm1, line 117: I think this debug line can go away now

src/fu-engine.c
```diff
@@ -5017,6 +5020,188 @@ fu_engine_get_host_machine_id (FuEngine *self)
...
+fu_engine_add_hsi_attrs_supported (FuEngine *self, GPtrArray *attrs)
+{
+	FwupdHsiAttr *attr;
+	FwupdRelease *rel_newest = NULL;
+	guint64 now = (guint64) g_get_real_time () / G_USEC_PER_SEC;
+	g_autoptr(FuDevice) device = NULL;
+	g_autoptr(GPtrArray) releases = NULL;
+	g_autoptr(GError) error_local = NULL;
+
+	/* find out if there is firmware less than 6 months old */
```
superm1, line 5050: you changed code to 12, this should also be 12

src/fu-engine.c
```diff
@@ -5017,6 +5020,188 @@ fu_engine_get_host_machine_id (FuEngine *self)
...
+	fwupd_hsi_attr_set_summary (attr, "There is firmware less than 12 months old");
+	fwupd_hsi_attr_add_flag (attr, FWUPD_HSI_ATTR_FLAG_SUCCESS);
+	fwupd_hsi_attr_add_flag (attr, FWUPD_HSI_ATTR_FLAG_RUNTIME_UPDATES);
+	g_ptr_array_add (attrs, attr);
+
+	/* do we have attestation checksums */
+	if (fwupd_release_get_checksums(rel_newest)->len > 0) {
+		attr = fwupd_hsi_attr_new ("org.fwupd.Hsi.Attestation");
+		fwupd_hsi_attr_set_name (attr, "Firmware Attestation");
+		fwupd_hsi_attr_set_summary (attr, "Firmware has attestation checksums");
```
superm1, line 5092: I think this should be "Latest System Firmware"

src/fu-engine.c
```diff
@@ -5017,6 +5020,188 @@ fu_engine_get_host_machine_id (FuEngine *self)
...
+		g_ptr_array_add (attrs, attr);
+	}
+}
+
+GPtrArray *
+fu_engine_get_host_security_attrs (FuEngine *self, GError **error)
+{
+	GPtrArray *plugins = fu_plugin_list_get_all (self->plugin_list);
+	g_autoptr(GPtrArray) hsi_attrs = NULL;
+
```
superm1, line 5104: I think you want two extra checks here:
1. At least the way that HSI is done right now, should make sure the system is booted in UFEI mode and return an error if not.
2. Check that it's x86_64.  Most of the architecture for HSI is x86 centered.

src/fu-engine.c
```diff
@@ -5017,6 +5020,188 @@ fu_engine_get_host_machine_id (FuEngine *self)
...
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "no host security attributes");
+		return NULL;
+	}
+	return g_steal_pointer (&hsi_attrs);
+}
+
+static gchar *
+fu_engine_hsi_attrs_to_string (GPtrArray *hsi_attrs)
```
superm1, line 5136: I think this is actually `fu_engine_calculate_hsi`.  It's not just converting a bunch of attributes into a string as much as it's building all the components that go into that string.

src/fu-engine.c
```diff
@@ -5017,6 +5020,188 @@ fu_engine_get_host_machine_id (FuEngine *self)
...
+{
+	guint hsi_number = 0;
+	FwupdHsiAttrFlags flags = FWUPD_HSI_ATTR_FLAG_NONE;
+	GString *str = g_string_new ("HSI:");
+
+	/* find the highest HSI number where there are no failures and at least
+	 * one success */
+	for (guint j = 1; j < 9; j++) {
```
superm1, line 5144: Why did you offer 9 as the upper boundary?  Shouldn't this be 5?

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
hughsie: Agreed, fixed.

src/fu-engine.c
hughsie: Fixed.

src/fu-engine.c
hughsie: Hmm, should we be looking at the _latest_ firmware release, or actually the _current running_ firmware release? In retrospect, the latter makes more sense.

src/fu-engine.c
hughsie: > make sure the system is booted in UFEI mode

I don't think HSI needs to be x86 centered, I just think for the other architectures we return HSI:0 until we define some sane firmware platform tests.

src/fu-engine.c
superm1: That's true, I agree current is what you want.

src/fu-engine.c
hughsie: Agree, fixed.

src/fu-engine.c
hughsie: Maybe we want something like `FWUPD_HSI_ATTR_NUMBER_MAX` exported in fwupd-hsi-attr.h?

src/fu-tool.c
```diff
@@ -1937,6 +1937,27 @@ fu_util_get_remotes (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	if (!fu_util_start_engine (priv, FU_ENGINE_LOAD_FLAG_NONE, error))
+		return FALSE;
+	attrs = fu_engine_get_host_security_attrs (priv->engine, error);
+	if (attrs == NULL)
+		return FALSE;
+	for (guint i = 0; i < attrs->len; i++) {
+		FwupdHsiAttr *attr = g_ptr_array_index (attrs, i);
+		g_autofree gchar *str = fwupd_hsi_attr_to_string (attr);
+		g_print ("%s\n", str);
+	}
```
superm1, line 1953: What do you think about doing tree output here? We're using that everywhere else, so maybe should be considered here too.

src/fu-engine.c
hughsie: Fixed.

src/fu-engine.c
hughsie: This is what I did.

src/fu-engine.c
superm1: Well several things are part of the x86 PC architecture specifically and not really seen elsewhere: TPM, EC.  Things like the IOMMU are seen in different implementations on other architectures.

superm1: And those PCH bits in HSI:1 are x86 specific too.

src/fu-tool.c
hughsie: Right, the `fwupd_hsi_attr_to_string()` was just a hack to show something quickly. If nothing else we need to translate the attribute titles like we do releases and devices already. I guess a tree mode might be prettier. If you want to take this I'll try and implement some more of the tests from the document.

libfwupd/fwupd-hsi-attr.c
superm1: FwupdSecurityAttr?

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	if (flag == FWUPD_HSI_ATTR_FLAG_RUNTIME_UNTRUSTED)
+		return "runtime-untrusted";
+	return NULL;
+}
+
+/**
+ * fwupd_hsi_attr_get_obsoletes:
+ * @self: A #FwupdHsiAttr
+ *
+ * Gets the self obsoletes.
```
superm1, line 58: Needs better description

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	g_return_val_if_fail (FWUPD_IS_HSI_ATTR (self), NULL);
+	return priv->obsoletes;
+}
+
+/**
+ * fwupd_hsi_attr_add_obsolete:
+ * @self: A #FwupdHsiAttr
+ * @appstream_id: the attribute appstream_id
+ *
+ * Sets the attribute appstream_id.
```
superm1, line 77: Wrong description

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+fwupd_hsi_attr_add_obsolete (FwupdHsiAttr *self, const gchar *appstream_id)
+{
+	FwupdHsiAttrPrivate *priv = GET_PRIVATE (self);
+	g_return_if_fail (FWUPD_IS_HSI_ATTR (self));
+	g_return_if_fail (appstream_id != NULL);
+	for (guint i = 0; i < priv->obsoletes->len; i++) {
+		const gchar *obsolete_tmp = g_ptr_array_index (priv->obsoletes, i);
+		if (g_strcmp0 (obsolete_tmp, appstream_id) == 0)
+			return;
+	}
```
superm1, line 91: seems like this should just be `fwupd_hsi_attr_has_obsolete` rather than code duplication

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+ * Since: 1.4.2
+ **/
+void
+fwupd_hsi_attr_set_uri (FwupdHsiAttr *self, const gchar *uri)
+{
+	FwupdHsiAttrPrivate *priv = GET_PRIVATE (self);
+	g_return_if_fail (FWUPD_IS_HSI_ATTR (self));
+	g_free (priv->uri);
+	priv->uri = g_strdup (uri);
+}
```
superm1, line 154: URI?  What is URI stuff doing here?  I didn't see that in design doc before

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	return (priv->flags & flag) > 0;
+}
+
+/**
+ * fwupd_hsi_attr_get_number:
+ * @self: A #FwupdHsiAttr
+ *
+ * Gets the time estimate for firmware installation (in seconds)
+ *
+ * Returns: the estimated time to flash this self (or 0 if unset)
```
superm1, line 341: wrong description

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	g_return_val_if_fail (FWUPD_IS_HSI_ATTR (self), 0);
+	return priv->number;
+}
+
+/**
+ * fwupd_hsi_attr_set_number:
+ * @self: A #FwupdHsiAttr
+ * @number: The amount of time
+ *
+ * Sets the time estimate for firmware installation (in seconds)
```
superm1, line 358: wrong description

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	}
+	if (priv->summary != NULL) {
+		g_variant_builder_add (&builder, "{sv}",
+				       FWUPD_RESULT_KEY_SUMMARY,
+				       g_variant_new_string (priv->summary));
+	}
+	if (priv->uri != NULL) {
+		g_variant_builder_add (&builder, "{sv}",
+				       FWUPD_RESULT_KEY_URI,
+				       g_variant_new_string (priv->uri));
```
superm1, line 407: not sure why URI?

libfwupd/fwupd-hsi-attr.c
```diff
@@ -0,0 +1,730 @@
...
+	if (priv->obsoletes->len > 0) {
+		json_builder_set_member_name (builder, FWUPD_RESULT_KEY_CHECKSUM);
+		json_builder_begin_array (builder);
+		for (guint i = 0; i < priv->obsoletes->len; i++) {
+			const gchar *appstream_id = g_ptr_array_index (priv->obsoletes, i);
+			json_builder_add_string_value (builder, appstream_id);
+		}
+		json_builder_end_array (builder);
+	}
+	fwupd_hsi_attr_json_add_string (builder, FWUPD_RESULT_KEY_URI, priv->uri);
```
superm1, line 556: not sure need URI

hughsie: Fixed.

libfwupd/fwupd-hsi-attr.c
hughsie: Removed.

libfwupd/fwupd-hsi-attr.c
hughsie: Fixed.

libfwupd/fwupd-hsi-attr.c
hughsie: Fixed.

libfwupd/fwupd-hsi-attr.c
hughsie: Nuked.

libfwupd/fwupd-hsi-attr.c
hughsie: Agreed, fixed.

libfwupd/fwupd-hsi-attr.c
hughsie: Fixed.

libfwupd/fwupd-hsi-attr.c
hughsie: Fixed.

libfwupd/fwupd-hsi-attr.c
hughsie: I've used that, thanks.

libfwupd/fwupd-security-attr.h
```diff
@@ -0,0 +1,86 @@
...
+ * @FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ATTESTATION:	Suffix `A`
+ * @FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE:		Suffix `!`
+ * @FWUPD_SECURITY_ATTR_FLAG_RUNTIME_UNTRUSTED:		Suffix `?`
+ *
+ * The flags available for HSI attributes.
+ **/
+typedef enum {
+	FWUPD_SECURITY_ATTR_FLAG_NONE			= 0,
+	FWUPD_SECURITY_ATTR_FLAG_SUCCESS		= 1 << 0,
+	FWUPD_SECURITY_ATTR_FLAG_RUNTIME_UPDATES	= 1 << 8,
```
superm1, line 46: Why are you starting at 8 here?

hughsie: I left a bit of room for "standard" flags so we could keep the RUNTIME ones together. For instance I think I need a `FWUPD_SECURITY_ATTR_FLAG_OBSOLETED` so we can run though all the obsoletes after running all the vfuncs and then easily skip. Maybe not, I've not implemented this bit yet.

src/fu-engine.c
hughsie: > make sure the system is booted in UFEI mode

What's the best way to check if we're in EFI mode? Check that /sys/firmware/efi/runtime exists?

libfwupd/fwupd-security-attr.h
superm1: Ah OK.  The logic being there will probably be a lot more runtime ones later that can factor into stuff.

hughsie: Right.

src/fu-engine.c
superm1: That's what we do in the UEFI plugin: https://github.com/fwupd/fwupd/blob/0258c12af38eae27812695d3351a3797b65b8d6c/plugins/uefi/fu-plugin-uefi.c#L607

Maybe that should be generalized somewhere else now in libfwupdplugin?

hughsie:
@superm1 -- I was thinking about something like this, comments welcome:

Screenshot from 2020-05-02 20-18-21

superm1:
That flow looks good in screenshot, but on the last two you should probably use something different than OK because it reads as you are tainted and it's okay.

plugins/swap/fu-plugin-swap.c
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-plugin-vfuncs.h"
+#include "fu-hash.h"
+
+#define FU_PLUGIN_SWAP_SOURCE_FILE	"/proc/swaps"
```
superm1, line 12: I think it might be better to use the fwupd prefix stuff for this specifically.  It will then allow using this with the test suite and various mocked up files.

plugins/swap/fu-plugin-swap.c
```diff
@@ -0,0 +1,102 @@
...
+	if (g_strv_length (lines) <= 2) {
+		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+		return TRUE;
+	}
+
+	/* check each swap, ignoring header line */
+	for (guint i = 1; lines[i] != NULL && lines[i][0] != '\0'; i++) {
+		g_warning ("lines[i]=%s", lines[i]);
+		g_strdelimit (lines[i], " ", '\0');
+		if (!g_str_has_prefix (lines[i], "/dev/dm-")) {
```
superm1, line 88: Also need to look for `/dev/mapper` for people using swapfiles that are encrypted

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -91,6 +91,79 @@ fu_plugin_get_results (FuPlugin *plugin, FuDevice *device, GError **error)
...
+gboolean
+fu_plugin_add_security_attrs (FuPlugin *plugin, GPtrArray *attrs, GError **error)
+{
+	fu_plugin_uefi_add_hsi_secureboot (plugin, attrs);
+	fu_plugin_uefi_add_hsi_tainted (plugin, attrs);
+	fu_plugin_uefi_add_hsi_lockdown (plugin, attrs);
+	fu_plugin_uefi_add_hsi_s3sleep (plugin, attrs);
+	return TRUE;
+}
+
```
superm1, line 166: All of the above except secureboot I think should come from another plugin or the engine directly so this doesn't necessarily stay locked down into UEFI architecture.

src/fu-util.c
```diff
@@ -2678,7 +2700,12 @@ main (int argc, char *argv[])
 		     /* TRANSLATORS: command description */
 		     _("Reinstall current firmware on the device."),
 		     fu_util_reinstall);
-
+	fu_util_cmd_array_add (cmd_array,
+		     "get-hsi",
```
superm1, line 2704: What do you think about `get-security` instead of `get-hsi` for this?

hughsie: What about just `security` instead?

plugins/uefi/fu-plugin-uefi.c
hughsie: Agree. I think I'm going to need a `spiflash` plugin for the kernel querying so maybe it makes sense there.

plugins/swap/fu-plugin-swap.c
hughsie: Good catch, thanks.

plugins/swap/fu-plugin-swap.c
hughsie: Yup, the swap plugin is crying out for some unit tests too. Will fix.

hughsie:
> but on the last two you should probably use something different than OK

What about 

    {tick} Linux Kernel: Not tainted
    {tick} Linux Kernel: Locked down

I'm not sure about the double `Linux Kernel` tho.

src/fu-util.c
superm1: That seems fine to me. Don't forget to add it to bash completion too.

superm1:
> > but on the last two you should probably use something different than OK
> 
> What about
> 
> ```
> {tick} Linux Kernel: Not tainted
> {tick} Linux Kernel: Locked down
> ```
> 
> I'm not sure about the double `Linux Kernel` tho.

Those seem good to me.  These are definitely two aspects being measured from the kernel, and I think we want to keep them separate.

src/fu-util.c
hughsie: Fixed, thanks.

plugins/swap/fu-plugin-swap.c
hughsie: Unit tests added.

hughsie:
>  These are definitely two aspects being measured from the kernel

Agree, fixed. As a general note, I don't think merging this makes much sense until we have the BIOSWE kernel module; I don't want people using this getting HSI:3 and then when we add the kernel module getting downgraded to HSI:0.

superm1:
> > These are definitely two aspects being measured from the kernel
> 
> Agree, fixed. As a general note, I don't think merging this makes much sense until we have the BIOSWE kernel module; I don't want people using this getting HSI:3 and then when we add the kernel module getting downgraded to HSI:0.

A few other thoughts that would allow merging the code sooner:
* Ask developer what sysfs interface will be needed for BIOSWE.  Add a "dummy" module that will just check for presence of that sysfs interface and fail the security attribute check otherwise.  I think that is the behavior that we'll want anyway even when the code is ready.
* Merge the code but put a big disclaimer in `security` command output
* Merge the code, but guard running `security` command unless `--force`

Something else I was thinking is that someone can possibly get wildly different results depending upon which plugins are compiled with the daemon.  Could do something like a minimum threshold of plugins at different levels need to report to calculate values to help this.  It could work just as well for BIOSWE plugin not yet being ready.

hughsie:
> but guard running security command unless --force

I did this.

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -10,32 +10,64 @@
...
 	}
 
 	return TRUE;
 }
+
+gboolean
+fu_plugin_add_security_attrs (FuPlugin *plugin, GPtrArray *attrs, GError **error)
+{
+	FuPluginData *data = fu_plugin_get_data (plugin);
+	if (data->is_intel) {
```
superm1, line 60: So AMD will be having something as well (see https://community.amd.com/community/amd-business/blog/2019/10/21/amd-and-microsoft-secured-core-pc)

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -91,6 +91,81 @@ fu_plugin_get_results (FuPlugin *plugin, FuDevice *device, GError **error)
 	return TRUE;
 }
 
+static void
+fu_plugin_uefi_add_hsi_tainted (FuPlugin *plugin, GPtrArray *attrs)
+{
+	FwupdSecurityAttr *attr = fwupd_security_attr_new ("org.kernel.CheckTainted");
+	fwupd_security_attr_set_name (attr, "Linux Kernel");
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	if (fu_uefi_read_file_as_uint64 ("/proc/sys/kernel", "tainted") == 0) {
```
superm1, line 100: Didn't we grow a `/proc` path lookup as part of swap that should be used?

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -91,6 +91,81 @@ fu_plugin_get_results (FuPlugin *plugin, FuDevice *device, GError **error)
...
+	g_autofree gchar *buf = NULL;
+	fwupd_security_attr_set_name (attr, "Linux Kernel");
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	if (g_file_get_contents ("/sys/kernel/security/lockdown", &buf, &bufsz, NULL)) {
+		if (g_strstr_len (buf, bufsz, "[integrity]") != NULL ||
+		    g_strstr_len (buf, bufsz, "[confidentiality]") != NULL) {
+			fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+			fwupd_security_attr_set_summary (attr, "Locked down");
+		}
+	}
```
superm1, line 122: I don't think either taint or lockdown should be specific to UEFI architecture.

plugins/uefi/fu-plugin-uefi.c
hughsie: We did indeed, will fix.

plugins/uefi/fu-plugin-uefi.c
hughsie: A new "linux" plugin perhaps?

plugins/cpu/fu-plugin-cpu.c
hughsie: Is it going to be in cpuids as well?

plugins/uefi/fu-plugin-uefi.c
hughsie: Done.

plugins/cpu/fu-plugin-cpu.c
superm1: I would hope it shows up there, but again until the kernel is actually exporting flags details are slim.

plugins/uefi/fu-plugin-uefi.c
superm1: Ya that sounds good

superm1: That or merging swap functionality into a Linux plug-in

hughsie: > merging swap functionality into a Linux plug-in

It's certainly easier disabling plugins when the plugins just do one thing. I'd err on more simple plugins from a readability and maintenance point of view.

superm1: OK.

hughsie: I've split these out; I think I was right on lots of small plugins, there are lots of edge cases (e.g. like /sys/kernel/tainted) changing that we should handle.

plugins/linux-sleep/fu-plugin-linux-sleep.c
```diff
@@ -0,0 +1,43 @@
...
+		return FALSE;
+	}
+	attr = fwupd_security_attr_new ("org.kernel.CheckS3Sleep");
+	fwupd_security_attr_set_number (attr, 3);
+	fwupd_security_attr_set_name (attr, "Linux Kernel S3 Sleep");
+	if (g_strstr_len (buf, bufsz, "deep") == NULL) {
+		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+	} else {
+		fwupd_security_attr_set_summary (attr, "Deep sleep available");
+	}
```
superm1, line 40: So something I keep going back and forth here; is it necessarily bad that the firmware has s3 methods if the OS isn't using them?  There is a bit in the firmware that gets set setting a preference of whether to use modern standby/s0ix
On Windows this _disables_ S3.  It's not possible to use.
On Linux this sets the default to s2i, but s3 is still possible to invoke with root permissions.

So should we only be focusing on what that bit is set to rather than the complete lack of S3?  I know right now there are a lot of platforms that still offer the S3 ACPI methods but set the bit to S2I.  The S3 methods are there as a contingency in case of power management problems on S2I (which certainly exist).

superm1: So my idea would be maybe we should instead have this plugin make two assertions.
1. An assertion on the default the firmware is set, this should go into the numerical HSI equation.
2. The "current" value it's set to, this should go into the runtime "!" equation if it's set to S3.

superm1: If you agree, then the way to evaluate this is by looking at the FADT.  There is an offset that indicates whether to use S2I (At FADT.Flags[21]: https://uefi.org/sites/default/files/resources/Intel_ACPI_Low_Power_S0_Idle.pdf)

The kernel looks at it here to set the default:
https://github.com/torvalds/linux/commit/e870c6c87cf9484090d28f2a68aa29e008960c93

hughsie: > is it necessarily bad that the firmware has s3 methods if the OS isn't using them

I kinda felt the same. 

> On Linux this sets the default to s2i

As in what happens when systemd echos `mem` into `sleep`? Or is it more complicated now.

> by looking at the FADT

How can we do that from userspace?

src/fu-tool.c
hughsie: This is now pretty.

plugins/tpm/fu-plugin-tpm.c
```diff
@@ -11,10 +11,43 @@
...
+	FuPluginData *data = fu_plugin_get_data (plugin);
+	FwupdSecurityAttr *attr = fwupd_security_attr_new ("org.trustedcomputinggroup.Tpm");
+	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL);
+	fwupd_security_attr_set_name (attr, "TPM");
+	if (data->has_tpm_v20) {
+		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+		fwupd_security_attr_set_result (attr, "v2.0");
+	} else if (data->has_tpm) {
+		fwupd_security_attr_set_result (attr, "Not in v2.0 mode");
+	}
```
superm1, line 50: Doesn't there need to be a 3rd case of doesn't have TPM?

hughsie: We could have `fwupd_security_attr_set_result(attr, "Not detected")` I guess, but at the moment it should just show `Failed`.

plugins/acpi-dmar/fu-plugin-acpi-dmar.c
```diff
@@ -0,0 +1,55 @@
...
+		fwupd_security_attr_set_result (attr, error_local->message);
+		return TRUE;
+	}
+	dmar = fu_acpi_dmar_new (blob, error);
+	if (dmar == NULL) {
+		g_prefix_error (error, "could not parse %s: ", fn);
+		return FALSE;
+	}
+	if (fu_acpi_dmar_get_opt_in (dmar))
+		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
```
superm1, line 53: Don't you want to explain here the failure scenario and what it means in the attribute?

hughsie: I think i'd struggle to explain the failure in less than 5 words. Got any ideas?

superm1: Here's 4:
`Pre-boot DMA protection unavailable`

plugins/tpm/fu-plugin-tpm.c
superm1: Well that's probably enough, just wanted to make sure the test was going to show up, and since it's required for HSI I guess it will

plugins/linux-sleep/fu-plugin-linux-sleep.c
superm1: >As in what happens when systemd echos mem into sleep? Or is it more complicated now.

Yes when `mem` is echoed into `sleep` the current looks at the value of `mem_sleep` to determine what to do, s2i or s3.

> How can we do that from userspace?

Same way the DMAR stuff you just did, grab FADT from `/sys/firmware/acpi` and then parse it to figure out that bit in the spec I linked above.

plugins/acpi-dmar/fu-plugin-acpi-dmar.c
hughsie: Added, thanks.

plugins/linux-sleep/fu-plugin-linux-sleep.c
hughsie: Hmm, I don't have a `/sys/firmware/acpi/tables/LPIT` -- do you?

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
```diff
@@ -0,0 +1,122 @@
...
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+
+#include "fu-plugin-vfuncs.h"
+#include "fu-hash.h"
+
+#define FU_PLUGIN_LINUX_SPI_LPC_SYSFS_DIR	"/sys/kernel/security/spi"
```
superm1, line 13: I think this should go into fu-common stuff

plugins/linux-sleep/fu-plugin-linux-sleep.c
superm1: Yeah I do have that on my system, but that is not what _needs_ to be parsed.  Just the FADT at that one bit I said above which should be on any system with ACPI 5.0 and later.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
```diff
@@ -0,0 +1,122 @@
...
+		return FALSE;
+	attr = fwupd_security_attr_new ("org.kernel.SMM_BWP");
+	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL);
+	fwupd_security_attr_set_name (attr, "SPI");
+	if (g_strcmp0 (buf, "1\n") == 0) {
+		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+		fwupd_security_attr_set_result (attr, "SMM required");
+	} else {
+		fwupd_security_attr_set_result (attr, "SMM not required");
+	}
```
superm1, line 107: This stuff isn't mandating SMM, it's BIOS write protect

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
```diff
@@ -0,0 +1,122 @@
...
+fu_plugin_startup (FuPlugin *plugin, GError **error)
+{
+	if (!g_file_test (FU_PLUGIN_LINUX_SPI_LPC_SYSFS_DIR, G_FILE_TEST_IS_DIR)) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_NOT_FOUND,
+			     "Cannot find %s",
+			     FU_PLUGIN_LINUX_SPI_LPC_SYSFS_DIR);
+		return FALSE;
+	}
```
superm1, line 33: Checking this on a real system without the kernel patches this actually causes a runtime failure which I think is wrong, no?  Should you actually be able to get HSI1+UA! with a missing kernel module to really calculate or should this be HSI0+UA!?

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: Ya, lets get the kernel bits upstream first :)

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: Right, as I understand it it's write protect, unless in SMM mode -- right?

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: I think `HSI1+UA!` is expected when the kernel module is missing. The `!` kinda means "we don't trust this value" rather than "we failed one of the HSI:1 tests" -- logically we couldn't run the test, rather that it failing.

plugins/linux-sleep/fu-plugin-linux-sleep.c
hughsie: Ahh, the FADT is `/sys/firmware/acpi/tables/FACP` -- give me 30 mins

hughsie: I've pushed something -- is that what you had in mind? My P50 seems to only support s3 sleep -- if you have a system with LPIT handy I'd appreciate if you could upload to `plugins/acpi-facp/tests/FACP-with-LPIT` or something and add a successful test.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
superm1: OK that makes more sense.  I think that we should embellish this detail more clearly in the spec too then.

plugins/linux-sleep/fu-plugin-linux-sleep.c
```diff
@@ -1,43 +0,0 @@
...
-		return FALSE;
-	}
-	attr = fwupd_security_attr_new ("org.kernel.CheckS3Sleep");
-	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_THEORETICAL);
-	fwupd_security_attr_set_name (attr, "Linux Kernel S3 Sleep");
-	if (g_strstr_len (buf, bufsz, "deep") == NULL) {
-		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
-	} else {
-		fwupd_security_attr_set_result (attr, "Deep sleep available");
-	}
```
superm1: So I don't think this one should go "away". I think we should have it for runtime checks - specifically if the firmware was set to default S2I, but the user changed it to `deep` via `/sys/power/mem_sleep` then I think that should fail a runtime check and you no longer trust things.

plugins/linux-sleep/fu-plugin-linux-sleep.c
superm1: Added.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
superm1: Oh I guess the point being to add the path after it's upstream, yeah makes sense.

plugins/linux-sleep/fu-plugin-linux-sleep.c
superm1: Well two ideas.

1. Move the check into the facp plugin rather than a dedicated sleep plugin.  Create the s3 attribute only if facp set appropriately.
2. Keep sleep it's own plugin, and make anything that is set to s3 a runtime error on all systems set this way.

hughsie: Hang on, can't we just say to get HSI:3 we have to have `/sys/power/mem_sleep != [deep]` AND facp with LOW_POWER_S0_IDLE_CAPABLE set?

superm1: Certainly - but the first one of those is a runtime thing that you can potentially change.

hughsie: > that you can potentially change

Right, but as the user can with secure boot too... I think we want to limit the number of things that trigger `!` especially when the admin setting [deep] manually is probably deliberate.

superm1: I know of a few machines that have shipped out of the factory with a kernel command line option setting it to `deep`.  It's not because `s2idle` doesn't work, but specifically because of energy consumption requirements in a local market (IE something like component XYZ uses XXX mW on Windows and uses YYY mW on Linux, but that keeps SOC from hitting this state which makes whole platform use ZZZ W which means it doesn't meet local energy regulations).  This is firmly in the camp of driver is missing a feature or has a bug.

I think for most people that's a net positive tradeoff, but it should be viewed negatively by the `security` command.  However then again those things were set as shipped, not modified by the user.

TL;DR You're right.  Check it once at fwupd startup and that's enough.

hughsie: Okay, sleep plugin re-instated

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
superm1: It's locking down the region that the BIOS is stored to only SMM writes

plugins/linux-sleep/fu-plugin-linux-sleep.c
hughsie: Great, thanks. I've rebased this PR on top of FuSecurityAttrs and it's much nicer and easier to understand now. Can you have a look? I can break it up into smaller chunks if that's easier to digest.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: So for failure "Non-SMM writes permitted" and for success "SMM write only"?

plugins/acpi-dmar/fu-plugin-acpi-dmar.c
```diff
@@ -0,0 +1,61 @@
...
+	g_autoptr(GError) error_local = NULL;
+
+	/* only Intel */
+	if (!fu_common_is_cpu_intel ())
+		return;
+
+	/* create attr */
+	attr = fwupd_security_attr_new ("org.uefi.ACPI.Dmar");
+	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_THEORETICAL);
+	fwupd_security_attr_set_name (attr, "ACPI kDMAp");
```
superm1, line 36: Pre-boot kernel DMA protection

plugins/acpi-dmar/fu-self-test.c
```diff
@@ -0,0 +1,45 @@
...
+	g_autofree gchar *fn = NULL;
+
+	fn = g_build_filename (TESTDATADIR, "DMAR", NULL);
+	blob = fu_common_get_contents_bytes (fn, &error);
+	g_assert_no_error (error);
+	g_assert_nonnull (blob);
+	dmar = fu_acpi_dmar_new (blob, &error);
+	g_assert_no_error (error);
+	g_assert_nonnull (dmar);
+	g_assert_true (fu_acpi_dmar_get_opt_in (dmar));
```
superm1, line 30: Perhaps also worth having a case that this fails for testing with?

plugins/acpi-facp/fu-plugin-acpi-facp.c
```diff
@@ -0,0 +1,57 @@
...
+		return;
+	}
+	facp = fu_acpi_facp_new (blob, &error_local);
+	if (facp == NULL) {
+		g_warning ("failed to parse %s: %s", fn, error_local->message);
+		fwupd_security_attr_set_result (attr, "Could not parse FACP");
+		return;
+	}
+	if (!fu_acpi_facp_get_s2i (facp)) {
+		fwupd_security_attr_set_result (attr, "S3 suspend in use");
```
superm1, line 51: I think a better message:
`Firmware defaults to suspend-to-ram (S3)"

plugins/linux-sleep/fu-plugin-linux-sleep.c
```diff
@@ -0,0 +1,47 @@
...
+	fu_security_attrs_append (attrs, attr);
+
+	/* load file */
+	if (!g_file_load_contents (file, NULL, &buf, &bufsz, NULL, &error_local)) {
+		g_autofree gchar *fn = g_file_get_path (file);
+		g_warning ("could not open %s: %s", fn, error_local->message);
+		fwupd_security_attr_set_result (attr, "Deep sleep status unavailable");
+		return;
+	}
+	if (g_strstr_len (buf, bufsz, "deep") != NULL) {
```
superm1, line 40: Didn't we say that we would only check what it was set to at daemon startup, not actually whether it was offered?

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
```diff
@@ -0,0 +1,139 @@
...
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+}
+
+gboolean
+fu_plugin_startup (FuPlugin *plugin, GError **error)
+{
+	if (!g_file_test (FU_PLUGIN_LINUX_SPI_LPC_SYSFS_DIR, G_FILE_TEST_IS_DIR)) {
```
superm1, line 24: This check essentially does nothing now, since the `fu_plugin_add_security_attrs` vfunc runs either way.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
superm1: I think that's simplifying it too much.  How about this:
Title: `BIOS region of SPI`
Success: `Writes only available through BIOS code`
Fail: `Writes available to operating system`

plugins/linux-swap/fu-plugin-linux-swap.c
```diff
@@ -36,41 +39,72 @@ fu_plugin_linux_swap_changed_cb (GFileMonitor *monitor,
...
+{
+	FuPluginData *data = fu_plugin_get_data (plugin);
+	gsize bufsz = 0;
+	g_autofree gchar *buf = NULL;
+	g_autoptr(FuLinuxSwap) swap = NULL;
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+	g_autoptr(GError) error_local = NULL;
+
+	/* create attr */
+	attr = fwupd_security_attr_new ("org.kernel.Swap");
```
superm1, line 75: If this came in from the changed callback, (IE If the swap keeps changing) then don't we need to also evacuate the existing attribute

plugins/linux-tainted/fu-plugin-linux-tainted.c
```diff
@@ -0,0 +1,92 @@
...
+	gsize bufsz = 0;
+	g_autofree gchar *buf = NULL;
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+	g_autoptr(GError) error_local = NULL;
+
+	/* create attr */
+	attr = fwupd_security_attr_new ("org.kernel.CheckTainted");
+	fwupd_security_attr_set_name (attr, "Linux Kernel");
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	fu_security_attrs_append (attrs, attr);
```
superm1, line 76: similarly for this, does the attribute need to be removed when it was changed?

plugins/acpi-dmar/fu-self-test.c
hughsie: Do you have anything you know has a DMAR which isn't opt-in?

superm1: Unfortunately yes.

plugins/linux-sleep/fu-plugin-linux-sleep.c
hughsie: I wasn't sure. The daemon doesn't have to be started at early-boot, it might be started by gnome-software well after the admin has done his "fixup". Maybe in this case the admin would just blacklist that one plugin?

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: Yup, agreed, nuked.

plugins/linux-swap/fu-plugin-linux-swap.c
hughsie: No, the FuSecurityAttrs is created for each request.

plugins/linux-tainted/fu-plugin-linux-tainted.c
hughsie: Nope. :)

plugins/acpi-dmar/fu-self-test.c
superm1: pushed it

hughsie: Thanks.

plugins/linux-sleep/fu-plugin-linux-sleep.c
superm1: But so irrespective of when the daemon has actually been started this should be looking at "[deep]" instead.  And flagging that as "Suspend is set to S3"

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: I've done something like that, thanks.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
```diff
@@ -0,0 +1,125 @@
...
+	attr = fwupd_security_attr_new ("org.kernel.BIOSWE");
+	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL);
+	fwupd_security_attr_set_name (attr, "SPI");
+	fu_security_attrs_append (attrs, attr);
+
+	/* load file */
+	fn = g_build_filename (FU_PLUGIN_LINUX_SPI_LPC_SYSFS_DIR, "bioswe", NULL);
+	if (!g_file_get_contents (fn, &buf, &bufsz, &error_local)) {
+		g_warning ("could not open %s: %s", fn, error_local->message);
+		fwupd_security_attr_set_result (attr, "Could not open file");
```
superm1, line 40: This errors as:

> "BIOS SPI Region: Could not open file"

So I think to make this more user friendly in the case of missing kernel, perhaps look for the directory first (IE `/sys/kernel/security/spi`) and if that fails, then indicate:

> "Kernel support for reading SPI security settings not present"

(or something similar)

plugins/linux-sleep/fu-plugin-linux-sleep.c
superm1: I just pushed what I'm thinking to the branch, see what you think.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: Should this fail HSI:1 or mark it as tainted?

superm1: I think we should fail it.  This is probably the most important basic firmware test.
And I realize that means that `fwupdtool security` effectively won't work until at least kernel 5.9 at this point (or a distro backports the kernel support), but I think it's the right trade off.

plugins/linux-sleep/fu-plugin-linux-sleep.c
hughsie: Works for me, thanks.

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: Okay, agreed. I've pushed something.

superm1: LGTM now

superm1:
>SPI: Write disabled
    ✔  SPI: SMM required
    ✔  UEFI Secure Boot: Enabled

Did you actually sign your own kernel and enroll your own signing authority?  I didn't expect to see both of those as on simultaneously right now.

hughsie:
> Did you actually sign your own kernel

No, I hacked up the plugin for a demo. :/

superm1:
Ha, OK 😜

PR #2053, opened 2020-05-02
Switch the default of EnumerateAllDevices to false

hughsie:
This was causing dozens of regressions to be reported.


hughsie:
Lets flip the default and try to get a device added to the test matrix here that triggers the random events. Once that works maybe we can flip it back.

superm1:
Yup makes sense

Issue #2052, opened 2020-05-02
fwupd disables touchpad on Dell K12A kbd-dock until reboot


jwrdegoede:
As requested I'm filing a big for this issue which I initially reported by email.

After replacing a quite old Fedora 27 install on one of my test machines with a Fedora 32 install I hit an interesting issue. The machine in question is a Dell Venue 11 Pro 7130 tablet / 2-in-1, with a keyboard dock with touchpad, turning it into more or less a
normal laptop.

The touchpad works fine in gdm, but it would stop working when I logged into a user GNOME-session.  It took me a while to get to the bottom of this. The problem is that the user-session ends up dbus activating fwupd and fwupd does some probe of the touchpad which puts it in a mode where it no longer generates any events.

`sudo rpm -e fwupd gnome-software` works around the issue of the touchpad no longer sending any events after I login to the user session.

hughsie:
@superm1 is that going to be wacom? @jwrdegoede can you get me a `sudo /usr/libexec/fwupd/fwupd --verbose` log please.

jwrdegoede:
Here are answers to the questions asked in your email response:

> Can we work on a less brutal fix (then the `rpm -e`) ?

Certainly that is the whole reason I reached out to you :)

> Is it an RMI4 touchpad by any chance?

No it is an USB touchpad/kbd combo using the HID-multitouch protocol for the touchpad part.

> Setting /etc/fwupd/daemon.conf BlacklistPlugins to include wacom_usb perhaps?

I've tried blacklisting wacom_usb, that does not help.

superm1:
A verbose daemon log would be really useful.

Also Can you please try to blacklist "dell" plugin and see if it helps?

superm1:
Btw are you sure it's USB? I didn't think that was how it worked. If so I would expect undocking and redocking fixes it.

jwrdegoede:
`fwupd --verbose` [log](https://fedorapeople.org/~jwrdegoede/fwupd.log)

Yes I'm sure it is USB. Unplug + replug fixes it, if I kill fwupd first so that it does not mess it with after the replug.

jwrdegoede:
Here are some interesting bits from the log:

```
14:59:12:0640 FuEngine             2 devices with subsystem hidraw
14:59:12:0642 FuDevice             using 82f75bac8ca62bf795945db8b38673283dd38e6f for HID_PHYS=usb-0000:00:14.0-4/input1
[...]

PR #2054, opened 2020-05-02
synaptics-rmi: Whitelist the specific device to avoid causing regress…

hughsie:
…ions

Apparently some RMI4 devices get upset when setting the RMA page.

Fixes https://github.com/fwupd/fwupd/issues/2052

### DIFF
--- a/plugins/synaptics-rmi/fu-synaptics-rmi-device.c
+++ b/plugins/synaptics-rmi/fu-synaptics-rmi-device.c
@@ -1007,6 +1007,9 @@ fu_synaptics_rmi_device_init (FuSynapticsRmiDevice *self)
 	fu_device_set_name (FU_DEVICE (self), "Touchpad");
 	fu_device_set_remove_delay (FU_DEVICE (self), FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
 	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_TRIPLET);
+	/* there are a lot of Synaptics HID peripherals, but not all respond
+	 * well to probing -- so limit to ones with issued updates */
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_ONLY_SUPPORTED);
 	priv->functions = g_ptr_array_new_with_free_func (g_free);
 }
 
--- a/plugins/synaptics-rmi/synaptics-rmi.quirk
+++ b/plugins/synaptics-rmi/synaptics-rmi.quirk
@@ -1,3 +1,10 @@
 [DeviceInstanceId=HIDRAW\VEN_06CB]
 Plugin = synaptics_rmi
 Vendor = Synaptics
+
+[DeviceInstanceId=HIDRAW\VEN_06CB&DEV_4875]
+Flags = ~only-supported
+[DeviceInstanceId=HIDRAW\VEN_06CB&DEV_76AF]
+Flags = ~only-supported
+[DeviceInstanceId=HIDRAW\VEN_06CB&DEV_CD8B]
+Flags = ~only-supported


hughsie:
I've asked Synaptics to review the whitelist.

superm1:
Well so this stops the plugin from working on my XPS 13 7390 that I have on hand.  GUIDs for that:
```
  Guid:                 f4384034-9243-5334-8075-a534be913e46 <- HIDRAW\VEN_06CB&DEV_76AF&REV_00
  Guid:                 424bd00e-9789-5cdf-a12a-3c81bc4676d6 <- HIDRAW\VEN_06CB&DEV_76AF
  Guid:                 140f4458-951b-5bb9-85e2-879bd5b02615 <- SYNAPTICS_RMI\TM3038-003
  Guid:                 b29d3c85-cd0e-503e-9c7e-f6731c1eaf2d <- SYNAPTICS_RMI\TM3038
```

When you talk to synaptics can you check if they know why this older machine is failing?  I think it's preferable to have a blacklist if it's just an older class of devices or some known hardware problem.

hughsie:
>  GUIDs for that

I've added that one, thanks.

> I think it's preferable to have a blacklist

I'm really worried about RHEL regressions like we found with the logitech hidpp plugin. I'd rather at 10 things to the whitelist in `trivial: ` commits than deal with one RHEL customer bug about disabling the touchpad on thousands of identical in-field devices.

superm1:
> I've added that one, thanks.

There are a bunch of other laptops that I've seen these in working properly.  I think this is going to be a pretty big whitelist if it's complete.

>I'm really worried about RHEL regressions like we found with the logitech hidpp plugin. I'd rather at 10 things to the whitelist in trivial: commits than deal with one RHEL customer bug about disabling the touchpad on thousands of identical in-field devices.

Before we merge and go that way I'd really like to understand the _why_ for this happening.  If it's "oh you forgot to do this check for HW blah before setting the RMA page" or "we only support things on I2C or SMBUS, not USB" then those are pretty easy things to adjust to prevent having to go this directoin.

superm1:
One more thought; if we don't get a root cause for this, another more RHEL friendly way to approach it is to "blacklist" from quirks.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
```diff
@@ -1007,6 +1007,9 @@ fu_synaptics_rmi_device_init (FuSynapticsRmiDevice *self)
 	fu_device_set_name (FU_DEVICE (self), "Touchpad");
 	fu_device_set_remove_delay (FU_DEVICE (self), FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
 	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_TRIPLET);
+	/* there are a lot of Synaptics HID peripherals, but not all respond
+	 * well to probing -- so limit to ones with issued updates */
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_ONLY_SUPPORTED);
```
superm1, line 1012: I don't know the only supported flag makes sense.  We have a bunch of GUIDs exported by the devices that allow the same payload to span across multiple systems, don't we?  So if the "wrong" GUID is used then this wouldn't work on newer fwupd versions but it would on older.

hughsie: >  We have a bunch of GUIDs exported by the devices that allow the same payload to span across multiple systems, don't we

I don't think so; I think it's only Lenovo using this in anger.

superm1: But the point being why even export those GUIDs if they effectively can't be used.  Unless we find out this is widespread, I would rather a blacklist solution via a quirk.

Issue #2055, opened 2020-05-04
Regression with new thunderbolt plugin and X1 carbon


hughsie:
Trying to deploy the update I get:

    failed to read is-host: location is outside of the given image: attempted to read 0x01 bytes at offset 0x7f014b from buffer of 0x7e000

Will debug later a bit, but wanted to file first.

superm1:
On the X1 carbon host controller?  Any idea what is the generation of the controller, AR or TR?

I'd guess that @YehezkelShB got https://github.com/fwupd/fwupd/blob/master/plugins/thunderbolt/fu-thunderbolt-firmware.c#L530 backwards when correcting it on the PR, but I'll need to look at the original code.

superm1:
I looked through and I think what's happening is that the old code used the FARB pointer location in the "new" image as it's basis for offset to say where digital section starts.  It then used this plusrather than that of the NVM coming from the kernel.  So for logic of is-host is at digital + 0x10 doesn't make  sense if that farb pointer is completely off.

@YehezkelShB so is it a fair assumption that digital section starts at 0x00, or is that completely off base?  I guess we need this properly parsing the images the kernel spits out, not just brand new genuine images.

YehezkelShB:
When reading the "live" image from sysfs, there is no FARB whatsoever. You don't get the header there, it starts right away with the digital section. The original code used the FARB offset only when reading from a file.

Does it answer the question?

superm1:
Oh!  OK that makes more sense, I was trying to understand why this looked like garbage and not the same header.  OK I guess I need to re-do the PR some then.

superm1:
@YehezkelShB what about the other bytes in that header?  Are any of those a signature that we can be checking for to tell if it's a FARB header or start of a digital section?

superm1:
(or similarly are the first few bytes of the digital section always identical and can be triggered the same way)

YehezkelShB:
It's simple: if this is an image file, it MUST start with the headers (which include the FRAB pointer). If it's a "live" FW coming from sysfs, there are no headers at all. The code must know where it's reading from.

superm1:
I guess for the plugin flow it will know where it's reading from, but for the `firmware-parse` command this enables it won't know.  It will need to look at it and tell what it looks like.

Right now my PR is just checking those first 3 bytes and if they're bigger than the file it's no header.  I don't think that's guaranteed (who knows what digital starts with).

[...]

PR #2056, opened 2020-05-04
Adjust parser for Thunderbolt (Fixes: #2055)

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

This fixes the parser when dealing with the NVM the kernel returns.  The old code was only ever used to compare images, and the offsets used were those that came from the "new" image, not the old image.  So ignore the value of the initial farb pointer and assume that digital section starts at the beginning if it's outside the size of the file.  @YehezkelShB is this a a reasonable assumption?

Also adjust to not read the native value from the image on external devices as this is meaningless.


YehezkelShB:
IIUC, part of the idea in the redesign of this plugin was to stop parsing the image file. FRAB is relevant only when reading from the image file. Don't use it when parsing the "live" FW as comes from sysfs.

superm1:
> IIUC, part of the idea in the redesign of this plugin was to stop parsing the image file. FRAB is relevant only when reading from the image file. Don't use it when parsing the "live" FW as comes from sysfs.

Thanks, I've adjusted it.

superm1:
If 6990c97 looks OK i'll also backport something similar to 1_4_X due to that regression I alluded to in the commit message.

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -62,6 +63,7 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_udev_subsystem (plugin, "thunderbolt");
 	fu_plugin_set_device_gtype (plugin, FU_TYPE_THUNDERBOLT_DEVICE);
 	fu_plugin_add_firmware_gtype (plugin, "thunderbolt", FU_TYPE_THUNDERBOLT_FIRMWARE);
+	fu_plugin_add_firmware_gtype (plugin, "thunderbolt-update", FU_TYPE_THUNDERBOLT_FIRMWARE_UPDATE);
```
hughsie, line 66: Nice one. This is how I would have done it. The only thing that irks me is the name -- `update` isn't the best noun -- maybe `spi`?

superm1: Well "thunderbolt" is the one from the SPI, "thunderbolt-update" is the one you're getting from the web.  So maybe instead:

"thunderbolt-spi" and "thunderbolt-update"?

superm1: Or alternatively "thunderbolt-controller" and "thunderbolt-update"

hughsie: maybe `update` is fine then.

PR #2061, opened 2020-05-05
uefi-dbx: Add a plugin that analyses the UEFI dbx variable

hughsie:
This will be used for future functionality.


contrib/ci/dependencies.xml
```diff
@@ -5,6 +5,19 @@
       <package />
     </distro>
   </dependency>
+  <dependency type="build" id="secureboot-db">
```
superm1, line 8: I think these should be runtime dependencies too, no?

plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+gchar *
+fu_uefi_dbx_get_dbxupdate (GError **error)
+{
+	g_autofree gchar *dbxdir = NULL;
+	g_autofree gchar *glob = NULL;
+	g_autoptr(GPtrArray) files = NULL;
+
+	/* fall back out of prefix */
```
superm1, line 49: These  should go into fwupd runtime path detection stuff rather than hardcoding stuff like /usr into the plugin.

contrib/ci/dependencies.xml
```diff
@@ -5,6 +5,19 @@
...
+    </distro>
+    <distro id="debian">
+      <package />
+    </distro>
+  </dependency>
+  <dependency type="build" id="dbxtool">
+    <distro id="fedora">
+      <package />
+    </distro>
+  </dependency>
```
superm1, line 20: I think we should talk about what to do for the snap too here, should we rely upon the host distro to provide the database or ship it with the snap (my inclination is to ship it with the snap; we can keep control of it then which is nice)

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
```diff
@@ -0,0 +1,90 @@
...
+
+#include "fu-plugin-vfuncs.h"
+#include "fu-efivar.h"
+#include "fu-hash.h"
+#include "fu-uefi-dbx-common.h"
+#include "fu-uefi-dbx-file.h"
+
+struct FuPluginData {
+	gchar			*fn;
+};
```
superm1, line 17: This seems like overkill to keep a member just for something that you use in startup.

Is this because for HSI stuff later you want to have access to run at it outside of the startup function?

plugins/uefi-dbx/fu-self-test.c
```diff
@@ -0,0 +1,55 @@
...
+{
+	gboolean ret;
+	gsize bufsz = 0;
+	g_autofree gchar *fn = NULL;
+	g_autofree guint8 *buf = NULL;
+	g_autoptr(FuUefiDbxFile) uefi_dbx_file = NULL;
+	g_autoptr(GError) error = NULL;
+
+	/* load file */
+	fn = fu_uefi_dbx_get_dbxupdate (&error);
```
superm1, line 25: Rather than splitting out the function to it's own file, why not just load the plugin and run it's startup routine?

contrib/ci/dependencies.xml
hughsie: I did wonder if fwupd should keep a copy too, but I'm not particularly enamoured with the idea of shipping some static security database in RHEL and being on the hook to update the package at a moments notice for CVEs. Maybe we ship a fallback copy if the other sources are not found?

contrib/ci/dependencies.xml
hughsie: Heh, I forgot all about the `type`. I don't think we support declaring runtime deps in the dependencies.xml manifest do we?

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
hughsie: > because for HSI stuff later

Exactly that.

plugins/uefi-dbx/fu-self-test.c
hughsie: I was pondering fuzzing it.

contrib/ci/dependencies.xml
hughsie: > Maybe we ship a fallback copy

This is what I did.

plugins/uefi-dbx/fu-uefi-dbx-common.c
hughsie: Agreed, fixed.

contrib/ci/dependencies.xml
hughsie: I added some bits in the spec file and in `control.in` -- yell if this was wrong.

plugins/uefi-dbx/fu-self-test.c
superm1: Can't you still fuzz if you have it set to obey paths set via environment variable and loading plug-in?

hughsie: Loading the plugin dynamically is an order of magnitude slower.

libfwupdplugin/fu-common.c
```diff
@@ -1089,6 +1089,20 @@ fu_common_get_path (FuPathKind path_kind)
...
+		if (tmp != NULL)
+			return g_strdup (tmp);
+#ifdef FWUPD_EFI_DBXDIR
+		tmp = g_getenv ("SNAP");
+		if (tmp != NULL)
+			return g_build_filename (tmp, FWUPD_EFI_DBXDIR, NULL);
+		return g_strdup (FWUPD_EFI_DBXDIR);
+#else
+		basedir = fu_common_get_path (FU_PATH_KIND_EFIAPPDIR);
+		return g_build_filename (tmp, "dbx", NULL);
```
superm1, line 1105: shouldn't this be basedir?

hughsie: it should indeed! good catch.

plugins/uefi-dbx/fu-self-test.c
hughsie: I'm just adding the fuzzer bits now.

libfwupdplugin/fu-common.c
```diff
@@ -1089,6 +1089,20 @@ fu_common_get_path (FuPathKind path_kind)
 		return g_strdup (EFI_APP_LOCATION);
 #else
 		return NULL;
+#endif
+	/* /usr/libexec/fwupd/efi/dbx */
```
superm1, line 1093: `libexecdir` is supposed to be executables and this stuff shouldn't be executable.  I think this should be going into `/usr/share/fwupd/` instead

plugins/uefi-dbx/meson.build
```diff
@@ -0,0 +1,60 @@
...
+  ],
+  c_args : cargs,
+  dependencies : [
+    plugin_deps,
+  ],
+)
+
+# only install this copy if the user has not specified another location
+# downloaded from https://uefi.org/revocationlistfile on 2020-05-06
+if efi_dbxdir == ''
```
superm1, line 29: Since the build process doesn't normally use the web, I think it might be worth adding to one of the CI tasks to checks if this file has changed and makes noise for it to be updated

contrib/ci/dependencies.xml
superm1: You got it right

superm1:
> plugins/uefi-dbx/dbxupdate.bin

I think that the license in https://uefi.org/revocationlistfile probably needs to be distributed with fwupd to include this in the tarball.

I'll offer another two alternatives though;
1. We can just fail the build if that path isn't set and the user hasn't manually downloaded the database into the directory they downloaded the tarball.
2. We can also offer a simple python script in contrib to do the download for the user (and then use that in the snap and any distro in CI that doesn't distribute the database in it's own package).

plugins/uefi-dbx/fu-self-test.c
superm1: I guess that's true, a lot more code is loaded.  OK well no concerns then, fuzzing is a good thing.

hughsie:
> I think that the license in https://uefi.org/revocationlistfile

plugins/uefi-dbx/meson.build
hughsie: tbh, I don't really want to ship the file at all...

libfwupdplugin/fu-common.c
hughsie: Fixed.

plugins/uefi-dbx/meson.build
hughsie: Thinking about it more rationally, I think we have to ship a fallback, otherwise the HSI is going to be incomplete on every non-mainstream Linux. I really want this to "just work" without having to hunt things down. I've added a CI test to check if the file is up to date.

superm1:
> > I think that the license in https://uefi.org/revocationlistfile
> 
> I asked Peter about this yesterday, and he said the file is just raw data and thus has no licence. I could see it being an issue for Debian certainly.

@ximion would you be able to weigh in on how this would be for Debian?  Can the UEFI revocation list be distributed in the fwupd tarball?  As far as I can tell this revocation list isn't currently distributed in Debian.

contrib/debian/rules
```diff
@@ -42,7 +42,7 @@ override_dh_auto_configure:
 		export DELL="-Dplugin_dell=false"; \
 	fi; \
 	if pkg-config --exists efivar; then \
-		export UEFI="-Dplugin_uefi=true -Dplugin_redfish=true -Dplugin_nvme=true"; \
```
superm1, line 45: This is a case that we're going to have to deviate behaviors for Ubuntu and Debian.  `secureboot-db` is a Ubuntu package.  I'll push the changes needed for contrib/debian/rules to get this right between them.

superm1:
>> Thinking about it more rationally, I think we have to ship a fallback, otherwise the HSI is going to be incomplete on every non-mainstream Linux. I really want this to "just work" without having to hunt things down. I've added a CI test to check if the file is up to date.

Another thought to avoid the potential licensing problem.
1. What about adding a `fwupdtool` only argument that is `update-uefi-dbx`.  This can go out to the web and get a database and put it somewhere in `/var`.  
2. To figure out which file to use (that one or distro) pick the one that is changed most recently.
3. When trying to run `security` if the file was missing, tell the user how to run `fwupdtool update-uefi-dbx` to update it.
3. When `security` runs, display the timestamp the file last changed.  Tell people how manually update it with that command if they want.

Then we're all out of the loop from maintaining the file, avoid the licensing problem, and let distros either update it or users update it when they want to double check things against the most recent revocation list.

contrib/debian/rules
hughsie: Thanks.

hughsie:
> to avoid the potential licensing problem

ximion:
> > > I think that the license in https://uefi.org/revocationlistfile
> > 
> > 
> > I asked Peter about this yesterday, and he said the file is just raw data and thus has no licence. I could see it being an issue for Debian certainly.
> 
> @ximion would you be able to weigh in on how this would be for Debian? Can the UEFI revocation list be distributed in the fwupd tarball? As far as I can tell this revocation list isn't currently distributed in Debian.

Hmm, hard to say. Technically it's just data, but the license states that:
> [...] you may distribute the UEFI Revocation List to end users solely as part of the distribution of an operating system software product, or as part of the distribution of updates to an operating system product; and you may distribute the UEFI Revocation List to end users or through your distribution channels solely as embodied in a firmware product or hardware product that embodies nontrivial additional functionality. Without limiting the foregoing, copying or reproduction of the UEFI Revocation List to any other server or location for further reproduction or redistribution on a standalone basis is expressly prohibited. [...]

This is a distribution limitation that would without any doubt be unacceptable for Debian (and I am puzzled why that limitation even exists, it makes zero sense to me to limit distribution of that list at all...)
If you want to get a better and more definitive answer, it would be best to write an email to debian-legal: https://lists.debian.org/debian-legal/ - there are people there who know how to deal with these issues.
But from my initial feeling about this is that this will be at least problematic.
Is there any chance that the license could be changed to be more permissive? I've heard Microsoft is supposed to be nice now ;-) :-P

ximion:
> > > I think that the license in https://uefi.org/revocationlistfile
> > 
> > 
> > I asked Peter about this yesterday, and he said the file is just raw data and thus has no licence. I could see it being an issue for Debian certainly.
> 
> @ximion would you be able to weigh in on how this would be for Debian? Can the UEFI revocation list be distributed in the fwupd tarball? As far as I can tell this revocation list isn't currently distributed in Debian.

Hmm, hard to say. Technically it's just data and you could argue that this shouldn't be covered by a license, but there is one and it states that:
> [...] you may distribute the UEFI Revocation List to end users solely as part of the distribution of an operating system software product, or as part of the distribution of updates to an operating system product; and you may distribute the UEFI Revocation List to end users or through your distribution channels solely as embodied in a firmware product or hardware product that embodies nontrivial additional functionality. Without limiting the foregoing, copying or reproduction of the UEFI Revocation List to any other server or location for further reproduction or redistribution on a standalone basis is expressly prohibited. [...]

This is a distribution limitation that would without any doubt be unacceptable for Debian (and I am puzzled why that limitation even exists, it makes zero sense to me to limit distribution of that list at all...)
If you want to get a better and more definitive answer, it would be best to write an email to debian-legal: https://lists.debian.org/debian-legal/ - there are people there who know how to deal with these issues and have a legal background.
But my initial feeling about this is that this will be at least problematic.
Is there any chance that the license could be changed to be more permissive? I've heard Microsoft is supposed to be nice now ;-) :-P

superm1:
>This is a distribution limitation that would without any doubt be unacceptable for Debian

Thanks for checking, you nailed down my specific fear around it.

> Is there any chance that the license could be changed to be more permissive? I've heard Microsoft is supposed to be nice now ;-) :-P

Well I _think_ that revocation list is maintained by the USWG not Microsoft?  It just so happens that Microsoft has the root of trust here so they are the ones signing updates I think.

So this probably would need to be raised in USWG, and down the rabbit hole we go.

hughsie:
> What about adding a fwupdtool only argument that is update-uefi-dbx

This isn't going to work when the user is just looking at a tab in the gnome-control-center. I think if the file is missing we just fail the HSI calculation with an `?` and tell the user the data is unavailable.

> and down the rabbit hole we go

This isn't something I think that's useful to spend time on.

superm1:
> > What about adding a fwupdtool only argument that is update-uefi-dbx
> 
> This isn't going to work when the user is just looking at a tab in the gnome-control-center. I think if the file is missing we just fail the HSI calculation with an `?` and tell the user the data is unavailable.

Well I think I would say HSI-0+? specifically, but yeah.
And I think that's a fine response.

But what do you think about in the command line tool still offering that option to let them pull an updated dbx down?  At least in something like RHEL with glacial updates I feel it's a useful way to get the data updated.

> 
> > and down the rabbit hole we go
> 
> This isn't something I think that's useful to spend time on.

Me either.

hughsie:
>  At least in something like RHEL with glacial updates

Updating dbxtool would be a CVE, which means it's very hard to not update it on basically everything.

ximion:
Yikes... I wonder whether this is one of the cases where "the list" is copyrightable, but "the hashes" themselves are not, so if you would reassemble your own list, the license wouldn't be applicable. But that feels like a slippery slope and potentially dangerous.
We'll probably need a lawyer to take a look at this - in the best case there is no problem, even for Debian, and in the worst case the list isn't redistributable by Debian (Debian IS an operating system, but that particular restriction would still make the list non-free (as one could e.g. also distribute the individual package containing the list without the whole OS and immediately violate the license. Also, what's an "operating system"? Do containers without a kernel count?).

hughsie:
>Yikes... I wonder whether this is one of the cases where "the list" is copyrightable, but "the hashes" themselves are not

/me summons @vathpela  -- has this been discussed before on the UEFI forum? We ship it in RHEL so I can't see there's a problem like this.

vathpela:
I'm not a lawyer, and I'm definitely not your lawyer.

That said, I really don't think any of it is copyrightable.  The hashes are no different than a URL in this regard; they're descriptive and they're mechanically generated according to a standard. The list itself is also a purely mechanical transformation; there's a correct canonical form, and if you take that group of hashes and make the list out of it, you get the same thing every time.

And in fact the download page at uefi.org makes no assertion of a copyright either - the only agreement there is an agreement that if you download it *from that site*, you agree that you don't get any rights to anything other than the file, and you acknowledge the warranty disclaimer.  It also says you're allowed to distribute it in an OS or firmware, but doesn't place any limits on anyone who gets the file from you.  If you get the file from my github repo, which I think you have been for quite some time, you don't even have to agree to it.  Arguably I might be violating the last clause in the not-all-caps section (I don't think I am), but you're not bound by it at all.

hughsie:
> dh_missing: warning: usr/share/dbx/dbxupdate.bin exists in debian/tmp but is not installed to anywhere

@superm1 what do you want to do for Debian? I'm happy replacing the file with the one from @vathpela's repo (same contents, different filename) if you think that gets around the problem.

superm1:
> dh_missing: warning: usr/share/dbx/dbxupdate.bin exists in debian/tmp but is not installed to anywhere

This error doesn't make sense - it should be in `usr/share/fwupd/dbx`.  I think there is still a logic error.

>@superm1 what do you want to do for Debian? I'm happy replacing the file with the one from @vathpela's repo (same contents, different filename) if you think that gets around the problem.

I don't think that gets around the issue.  I don't feel comfortable keeping this in Debian until someone authoritative signs off.
I think we're pretty much at a crossroads of 3 options:
1. Raise this to debian legal as @ximion recommended.
2. Set up Debian to strip out this file in the tarball (that means HSI is busted on Debian and any other distro that has to strip it)
3. Decouple shipping it, make people fetch it themselves if not put in the distro already.

hughsie:
What if we shipped an preprocessed regenerated version? i.e. same SHA256 checksums, but no pkcs7 header.

plugins/uefi-dbx/meson.build
```diff
@@ -0,0 +1,90 @@
...
+  install_data(['dbxupdate.bin'],
+    install_dir : join_paths(datadir, 'dbx')
```
superm1: heres where the problem was for debian.

superm1: Yay CI for catching it.

superm1:
> What if we shipped an preprocessed regenerated version? i.e. same SHA256 checksums, but no pkcs7 header.

I feel like it's shifting the ownership sufficiently.  @ximion what do you think?

hughsie:
If so, maybe we hijack `plugins/uefi-dbx/create-fuzzing-targets.py` into plugins/uefi-dbx/dbxtool.py and then have commands like `--cmd=create-empty`, `--cmd-create-one` (for fuzzing) and then also `--cmd=repack` which takes the downloaded dbxupdate.bin and converts `--input` into what's specified by `--output` -- this way we can still do the CI check as we can call the python tool there. We check the `dbxupdate-fwupd.bin` into git, and remove the UEFI one.

ximion:
> > What if we shipped an preprocessed regenerated version? i.e. same SHA256 checksums, but no pkcs7 header.
> 
> I feel like it's shifting the ownership sufficiently. @ximion what do you think?

I really don't think I have any legal authority, and certainly not enough to make a general statement about Debian. That's what Debian and the DPL consult lawyers for, and what Red Hat and Canonical employ lawyers for, so they actually can give a definitive answer or at least on with some legal backing.
To be absolutely sure, ask debian-legal (there is also a private mail address for sensitive legal matters).

However, from an engineering perspective, I don't think a set of hashsums is copyrightable at all (or rather: shouldn't be). I do read statements like "[...]redistribution on a standalone basis is expressly prohibited[...]" though and wonder whether that makes the list *from that site* distributable by us at all.
If hughsie generates a new list, especially if it's different from the one one the UEFI site, which happens to have the same hashes, for all I can see that would be his creation then and not subjected to the license of that particular list file on the UEFI site.
So, I do think that would work, and if that works it's certainly no Problem for Debian, but I don't *know* it because I'm not a lawyer (and have especially no idea about U.S. law). There has been so much counter-intuitive and crazy-looking legal stuff in the software engineering world, that it's hard to rule out anything (the UEFI Forum or Microsoft suing the fwupd project or Debian over this list would be absolutely insane though and go against their mission, so I think that event is highly unlikely).

So, tl;dr I have no power here but believe a new list with the same hashes would be a new thing and wouldn't be bound by the website's terms.

superm1:
I raised some discussion to debian-legal about this.  Summary of the touched on points:

1. We should contact USWG and ask why it can't be BSD/MIT in the first place.
2. Not sure it meets requirements to copyrightability (license doesn't matter)
3. Distributing it with fwupd is certain that it will be out of date.  If we want up to date data and distributed in Debian it should come in another package that can be updated regularly (like how time zone data is done).
4. They don't think that fallback idea the files would be meaningfully different to the original files since they're derived.

FFY00:
Sorry, I only skimmed through the comments. It's late and I don't have enough strength to parse everything. But why aren't we downloading this, as needed, at runtime? No need for anyone to be redistributing anything. We also have the benefit that the list will always be up-to-date. I would say a fwupd user would be expected to have a network connection.

hughsie:
I've removed the binary file and have included a message in the tests and plugin failure message that the file is missing and where it can be obtained from. My "how much I care about Debian" has been trained to 0% by the complete lack of pragmatism.

@superm1 if this PR looks okay now I will squash down an merge it.

@FFY00 if you know where Arch stores the dbx on the filesystem that would be useful as well. Thanks.

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
```diff
@@ -0,0 +1,94 @@
...
+	g_autoptr(FuUefiDbxFile) dbx_system = NULL;
+	g_autoptr(FuUefiDbxFile) dbx_update = NULL;
+	g_autoptr(GError) error_local = NULL;
+
+	/* get binary blob */
+	data->fn = fu_uefi_dbx_get_dbxupdate (error);
+	if (data->fn == NULL) {
+		g_prefix_error (error,
+				"file can be downloaded from %s: ",
+				FU_UEFI_DBX_DATA_URL);
```
superm1, line 51: If we aren't offering a fwupdtool/fwupdmgr command that downloads it, we probably should put in the message where the user needs to store it.  Or distribute a helper script that does it with fwupd.

superm1:
> My "how much I care about Debian" has been drained to 0% by the complete lack of pragmatism.

You and me both.  Not sure if you looked at the debian-legal post but the people there seem to view dbx as a threat do Debian by revoking Debian hashes rather than a improve the security of the whole ecosystem.

Anyway, thanks for being pragmatic even when they're not.

> 
> @superm1 if this PR looks okay now I will squash down an merge it.
> 
> @FFY00 if you know where Arch stores the dbx on the filesystem that would be useful as well. Thanks.

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
hughsie: > where the user needs to store it

superm1: Yeah I think that makes sense.

hughsie: I did the /var/lib/fwupd/dbx thing. At the moment the error is just a console warning, but with the hsi branch it's visible in `fwupdmgr security`

superm1:
>I did the /var/lib/fwupd/dbx thing. At the moment the error is just a console warning, but with the hsi branch it's visible in fwupdmgr security

That's better. One more thing though; the revocation list is a zip file.  It's not going to be obvious that zip needs to be unpacked.

hughsie:
> the revocation list is a zip file

Urgh, I forgot about that. Maybe `file can be downloaded from %s and decompressed into %s`

superm1:
> > the revocation list is a zip file
> 
> Urgh, I forgot about that. Maybe `file can be downloaded from %s and decompressed into %s`

Yeah that works.

hughsie:
> Yeah that works.

Done. @superm1 ready to merge?

PR #2062, opened 2020-05-06
support MATEKF722SE quirck

amfern:
MATEKF722SE has unconvetional behavior for dfu protocol,
sector size isn't specified and sector type is shiffted left by 1
this happens only for one sector

sector parsing from MATEKF722SE
*016Kg
*64Kg
*128Kg
*048 e
*528e
*004 e

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/dfu-target.c
+++ b/plugins/dfu/dfu-target.c
@@ -198,6 +198,15 @@ dfu_target_parse_sector (DfuTarget *target,
 		return FALSE;
 	}
 
+	/* handle weirdness */
+	if (fu_device_has_custom_flag (FU_DEVICE (dfu_target_get_device (target)),
+				       "absent-sector-size")) {
+		if (tmp[1] == '\0') {
+			tmp[1] = tmp[0];
+			tmp[0] = 'B';
+		}
+	}
+
 	/* get multiplier */
 	switch (tmp[0]) {
 	case 'B':		/* byte */
--- a/plugins/dfu/dfu-device.c
+++ b/plugins/dfu/dfu-device.c
@@ -43,6 +43,7 @@
  * * `use-protocol-zero`:	Fix up the protocol number
  * * `legacy-protocol`:		Use a legacy protocol version
  * * `detach-for-attach`:	Requires a DFU_REQUEST_DETACH to attach
+ * * `absent-sector-size`:	In absense of sector size, assume byte
  *
  * Default value: `none`
  *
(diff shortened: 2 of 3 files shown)


plugins/dfu/dfu-target.c
```diff
@@ -198,6 +198,12 @@ dfu_target_parse_sector (DfuTarget *target,
 		return FALSE;
 	}
 
+  // MATEKF722SE board quirk, sector type is the first char, the second char is empty
+  if (tmp[1] == '\0') {
+      tmp[1] = tmp[0];
+      tmp[0] = 'B';
+  }
+
```
hughsie, line 206: Can you explain a bit on why this happens? Is it the kind of thing we should add a fwupd quirk for? It seems slightly dangerous to assume 'B' in all cases. Also, tabs please.

plugins/dfu/dfu-target.c
```diff
@@ -198,6 +198,12 @@ dfu_target_parse_sector (DfuTarget *target,
 		return FALSE;
 	}
 
+  // MATEKF722SE board quirk, sector type is the first char, the second char is empty
```
superm1, line 201: FYI the project uses C style comments, not C++ style

plugins/dfu/dfu-target.c
superm1: I agree; this should be done as a quirk on this particular board.

amfern: I am not sure why this happens, but this certainly isn't documented by dfu protocol. 
I copied this assumption from dfu-util https://github.com/Stefan-Schmidt/dfu-util/blob/fdc72a8a738994dc72cfd63ff08f690e5145277e/src/dfuse_mem.c#L170.

i agree this, should be a quirk, i can implement new quirk, but can you give me general guides or point me to example of similar quirk in the code?

superm1: At a high level add a new quirk type here: https://github.com/fwupd/fwupd/blob/02930f1f23e67c410c808d7ead6dd7da6982c461/plugins/dfu/dfu-device.c#L45

Add an entry into `dfu.quirk` for your device and the quirks it needs (via Flags key).

And then in the code look for that quirk and activate the changed behavior.

Just force push this branch with the fixes once you have them done.

amfern: ok, i added absent-sector-size

amfern:
i am not sure why the CI fails on dfu-self-test, how can i run this test locally?

hughsie:
> how can i run this test locally?

Like this: `ninja && ./plugins/dfu/dfu-self-test`

hughsie:
I've fixed this up and merged it to master. Thanks!

amfern:
Thanks you very, much and for this incredible project

PR #2063, opened 2020-05-06
swap: Add a plugin that parses /proc/swaps

hughsie:
This will be used for future functionality.


hughsie:
> if a drive is a SED

Do we trust that now?

superm1:
> Do we trust that now?

I don't know, but anyway that's for outside of this PR.

PR #2064, opened 2020-05-06
cpu: Parse the CPU flags to detect the CET status

hughsie:
New enough hardware to have this feature isn't going to be in the marketplace
for a while. To use that newer hardware requires a very recent kernel (5.6 at
least, although it will probably be at least 5.9 by the time the hardware is
released).

The CET status will be used in future functionality.


superm1:
(Just as an FYI I don't see those flags exported by the kernel _yet_, so if for some reason they change this might require a minor fixup)

PR #2065, opened 2020-05-06
thunderbolt/dell-dock: Delay activation for Thunderbolt updates

superm1:
Fixes a regression on 1_4_X related to replug behavior.

This is a backport of 0cd2f107113b23774187addb29611dd1daa019c6 to
1_4_X which hasn't yet gone through the thunderbolt plugin modernization.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
This is actually cleaner than the old approach, it should work on 1_3_X too, opposition to backporting there too?

Issue #2068, opened 2020-05-07
Build errors when building with -Werror=discarded-qualifiers


jwrdegoede:
Lets continue the offtopic thing about the build-errors from #2052 here in a new issue.

First of all let me say sorry, I just checked and I was wrong,  Fedora's default build-flags do not contain -Werror=discarded-qualifiers.

As mentioned I have a wrapper around meson to inject Fedora's buildflags into local builds when I configure them with meson. But it seems that at some point in time, for some reason I added "-Werror=discarded-qualifiers -Werror=implicit-function-declaration" as extra flags to these., probably while working on mutter.

So this should explain why I'm seeing these errors and you are not (oops, my bad).

Now I think these are still worth fixing. As mentioned in #2052 I already started fixing it, but I gave up when more errors popped up.

Besides the earlier mentioned errors in #2052 I'm also seeing this one:

```
../plugins/ccgx/fu-ccgx-hpi-device.c: In function ‘fu_ccgx_hpi_write_firmware’:
../plugins/ccgx/fu-ccgx-hpi-device.c:1176:12: error: passing argument 3 of ‘fu_ccgx_hpi_write_flash’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
 1176 |            g_bytes_get_data (rcd->data, NULL),
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../plugins/ccgx/fu-ccgx-hpi-device.c:815:13: note: expected ‘guint8 *’ {aka ‘unsigned char *’} but argument is of type ‘gconstpointer’ {aka ‘const void *’}
  815 |     guint8 *buf,
      |     ~~~~~~~~^~~
cc1: some warnings being treated as errors
```

This one seems tricky to fix, we end up storing the const pointer in  FuCcgxHpiDeviceRetryHelper and that one is used for both reading and writing, I guess we could put separate buf_in and buf_out pointers in that struct and make buf_in const
(and use the right one in fu_ccgx_hpi_device_reg_read_cb / fu_ccgx_hpi_device_reg_write_cb).

And with your wip branch I'm getting this error now:

```
../plugins/ep963x/fu-ep963x-device.c: In function ‘fu_ep963x_device_write_icp’:
../plugins/ep963x/fu-ep963x-device.c:73:26: error: passing argument 1 of ‘fu_memcpy_safe’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
[...]

PR #2069, opened 2020-05-07
Fix various build issues with -Wdiscarded-qualifiers

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2068

### DIFF
--- a/plugins/ep963x/fu-ep963x-device.c
+++ b/plugins/ep963x/fu-ep963x-device.c
@@ -22,7 +22,7 @@ G_DEFINE_TYPE (FuEp963xDevice, fu_ep963x_device, FU_TYPE_HID_DEVICE)
 static gboolean
 fu_ep963x_device_write (FuEp963xDevice *self,
 			guint8 ctrl_id, guint8 cmd,
-			guint8 *buf, gsize bufsz,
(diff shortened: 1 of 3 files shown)


plugins/ep963x/fu-ep963x-device.c
```diff
@@ -47,8 +47,9 @@ fu_ep963x_device_write (FuEp963xDevice *self,
 }
 
 static gboolean
-fu_ep963x_device_write_icp (FuEp963xDevice *self,
-			    guint8 cmd, guint8 *buf, gsize bufsz,
+fu_ep963x_device_write_icp (FuEp963xDevice *self, guint8 cmd,
+			    const guint8 *buf, gsize bufsz,
+			    guint8 *bufout, gsize bufoutsz,
```
superm1, line 52: I will admit this feels a bit hacky to pass the buf parameter twice, once in const and one not..  Would you consider instead excluding the qualifier check for the compiler in this function?

hughsie: The right thing to do is to not pass in the same buf. I'll fix.

hughsie: Done.

PR #2071, opened 2020-05-09
trivial: Add a helper method to check for an Intel CPU

hughsie:
Some plugins will be Intel specific.

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -19,6 +19,8 @@
 #include <shlwapi.h>
 #endif
 
+#include <cpuid.h>
+
 #include <archive_entry.h>
 #include <archive.h>
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -2026,3 +2028,31 @@ fu_common_kernel_locked_down (void)
...
+{
+	guint eax = 0;
+	guint ebx = 0;
+	guint ecx = 0;
+	guint edx = 0;
+	guint level = 0;
+
+	/* get vendor */
+	__get_cpuid(level, &eax, &ebx, &ecx, &edx);
+	if (ebx == signature_INTEL_ebx &&
```
superm1, line 2052: I'm not familiar with this header and lookup but I notice you didn't also check eax. Is that intentional?

hughsie: I think eax is used to modify what gets returned in the other registers, i.e. it's a `in` parameter, but weirdly passed by reference.

PR #2079, opened 2020-05-11
Wip/hughsie/hsi tiny steps

hughsie:


src/fu-agent.c
```diff
@@ -113,6 +114,37 @@ fu_util_add_updates_json (FuUtilPrivate *priv, JsonBuilder *builder, GError **er
...
+static gboolean
+fu_util_add_security_attributes_json (FuUtilPrivate *priv, JsonBuilder *builder, GError **error)
+{
+	g_autoptr(GPtrArray) attrs = NULL;
+
+	/* not ready yet */
+	if ((priv->flags & FWUPD_INSTALL_FLAG_FORCE) == 0) {
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_NOT_SUPPORTED,
```
superm1, line 126: I think for end user tools we're normally using `FWUPD_ERROR_NOT_SUPPORTED`

hughsie: Ahh, good catch, thanks.

PR #2082, opened 2020-05-12
hsi: Run the fu_plugin_add_security_attrs() vfunc even for disabled p…

hughsie:
…lugins

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -1216,16 +1216,25 @@ fu_plugin_runner_flagged_device_generic (FuPlugin *self, FwupdInstallFlags flags
 
 }
 
+typedef enum {
+	FU_PLUGIN_RUNNER_FLAG_NONE		= 0,
+	FU_PLUGIN_RUNNER_FLAG_EVEN_DISABLED	= 1 << 0,
+	FU_PLUGIN_RUNNER_FLAG_LAST,
+} FuPluginRunnerFlags;
+
 static gboolean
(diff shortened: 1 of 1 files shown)


superm1:
As a counter-discussion point to this, can't you argue that if a plugin fails startup routine it should fail all the HSI stuff?  So perhaps it's worth collecting the error from failed startup and using that for HSI instead?

hughsie:
> it should fail all the HSI stuff

So if the Intel CET plugin fails then we should drop down to HSI:0? Or of the dell plugin fails to load on a Lenovo? I think the plugin itself is best to decide whether it being disabled just "fails" the HSI level it's trying to get to, or if it should taint the entire calculation.

The other thing I was pondering as a counter-counter-point is to instead pass a FuSecurityAttrList rather than a GPtrArray -- which would allow us to move the HSI calculation there, and also ensure that plugins actually add an attr to the list (via `fu_security_attr_list_add()`) rather than a FuDevice or something else odd.

hughsie:
It was much more typing, but I think https://github.com/fwupd/fwupd/pull/2085 is a much better long term API. Comments welcome.

superm1:
> So if the Intel CET plugin fails then we should drop down to HSI:0? Or of the dell plugin fails to load on a Lenovo? I think the plugin itself is best to decide whether it being disabled just "fails" the HSI level it's trying to get to, or if it should taint the entire calculation.

Well I think those cases you wouldn't mark it as _required_ for HSI calculation.  So the startup check would fail but they wouldn't be required so you wouldn't collect the error.

> It was much more typing, but I think #2085 is a much better long term API. Comments welcome.

OK will look

PR #2085, opened 2020-05-12
hsi: Abstract out the list of FwupdSecurityAttr objects for plugins

hughsie:
This exports FuSecurityAttrs into libfwupdplugin so that we can pass the plugins
this object rather than a 'bare' GPtrArray. This greatly simplifies the object
ownership, and also allows us to check the object type before adding.

In the future we could also check for duplicate appstream IDs or missing
properties at insertion time.

This change also changes the fu_plugin_add_security_attrs() to not return an
error. This forces the plugin to handle the error, storing the failure in the
attribute itself.

Only the plugin know if a missing file it needs to read indicates a runtime
problem or a simple failure to obtain a specific HSI level.


plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -125,19 +125,22 @@ fu_plugin_device_registered (FuPlugin *plugin, FuDevice *device)
...
 {
 	FuPluginData *data = fu_plugin_get_data (plugin);
-	FwupdSecurityAttr *attr = fwupd_security_attr_new ("org.trustedcomputinggroup.TpmEventLog");
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	attr = fwupd_security_attr_new ("org.trustedcomputinggroup.TpmEventLog");
 	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_IMPORTANT);
 	fwupd_security_attr_set_name (attr, "TPM Reconstruction");
-	if (data->reconstructed) {
+	if (!fu_plugin_get_enabled (plugin)) {
```
superm1, line 137: Ah OK, so this lets the plugin vfunc run no matter enabled or not and it can make a decision then what to do with HSI .  I think this is a better approach indeed.

PR #2087, opened 2020-05-13
Some HSI improvements

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5045,6 +5045,8 @@ fu_engine_add_security_attrs_tainted (FuEngine *self, FuSecurityAttrs *attrs)
 	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
 	if (self->tainted) {
 		fwupd_security_attr_set_result (attr, "Tainted");
+	} else if (self->plugin_filter->len > 0) {
+		fwupd_security_attr_set_result (attr, "Disabled plugins");
 	} else {
 		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
 	}
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5040,7 +5040,7 @@ fu_engine_get_host_machine_id (FuEngine *self)
 static void
 fu_engine_add_security_attrs_tainted (FuEngine *self, FuSecurityAttrs *attrs)
 {
-	FwupdSecurityAttr *attr = fwupd_security_attr_new ("org.fwupd.Hsi.Plugins");
+	g_autoptr(FwupdSecurityAttr) attr = fwupd_security_attr_new ("org.fwupd.Hsi.Plugins");
 	fwupd_security_attr_set_name (attr, "fwupd plugins");
 	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
 	if (self->tainted) {
@@ -5056,8 +5056,8 @@ fu_engine_add_security_attrs_supported (FuEngine *self, FuSecurityAttrs *attrs)
 {
 	FwupdRelease *rel_current = NULL;
 	FwupdRelease *rel_newest = NULL;
-	FwupdSecurityAttr *attr_a;
-	FwupdSecurityAttr *attr_u;
+	g_autoptr(FwupdSecurityAttr) attr_a = NULL;
+	g_autoptr(FwupdSecurityAttr) attr_u = NULL;
 	guint64 now = (guint64) g_get_real_time () / G_USEC_PER_SEC;
 	g_autoptr(FuDevice) device = NULL;
 	g_autoptr(GPtrArray) releases = NULL;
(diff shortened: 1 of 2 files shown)


src/fu-engine.c
```diff
@@ -5040,11 +5040,23 @@ fu_engine_get_host_machine_id (FuEngine *self)
...
 {
+	gboolean disabled_plugins = FALSE;
+	GPtrArray *blacklist = fu_config_get_blacklist_plugins (self->config);
 	FwupdSecurityAttr *attr = fwupd_security_attr_new ("org.fwupd.Hsi.Plugins");
 	fwupd_security_attr_set_name (attr, "fwupd plugins");
 	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	for (guint i = 0; i < blacklist->len; i++) {
+		const gchar *name_tmp = g_ptr_array_index (blacklist, i);
+		if (g_strcmp0 (name_tmp, "test") != 0 &&
+		    g_strcmp0 (name_tmp, "invalid") != 0) {
```
hughsie, line 5051: Not sure about the hardcoding, but for these two it's probable fine.

src/fu-engine.c
```diff
@@ -5040,11 +5040,23 @@ fu_engine_get_host_machine_id (FuEngine *self)
 static void
 fu_engine_add_security_attrs_tainted (FuEngine *self, FuSecurityAttrs *attrs)
 {
+	gboolean disabled_plugins = FALSE;
+	GPtrArray *blacklist = fu_config_get_blacklist_plugins (self->config);
 	FwupdSecurityAttr *attr = fwupd_security_attr_new ("org.fwupd.Hsi.Plugins");
```
hughsie, line 5045: Ohh, this needs to be `g_autoptr()` -- my bad. Can you fix that up pls.

src/fu-util-common.c
```diff
@@ -1610,9 +1611,21 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
 				FwupdSecurityAttr *attr = g_ptr_array_index (attrs, i);
 				if (!fwupd_security_attr_has_flag (attr, hpi_suffixes[j]))
 					continue;
+				if (!help &&
```
hughsie, line 1614: maybe lose this, micro-optimization.

src/fu-util-common.c
```diff
@@ -1610,9 +1611,21 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
...
+				    fwupd_security_attr_has_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE) &&
+				    !fwupd_security_attr_has_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS))
```
hughsie, line 1616: woah, that's quite a high bar -- if you fail a HSI:5 check and you drop down to HSI:4 that's really still very high and I don't think we should show the notification. I'd much rather have two pages, one for runtime issues, and one if the HSI number is < 2 or something.

src/fu-util-common.c
```diff
@@ -1610,9 +1611,21 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
...
 				fu_security_attr_append_str (attr, str);
 			}
 		}
 	}
+
+	if (help) {
+		/* TRANSLATORS:  this is instructions on how to improve the HSI suffix */
+		g_string_append_printf (str, "\n%s: %s",
+					_("Your system has runtime issues. To improve your security visit"),
+					"https://github.com/fwupd/fwupd/wiki/Low-host-security-ID");
```
hughsie, line 1627: I don't think translators like having URLs - maybe build the string using `g_string_new()` and use %s for the URL.

src/fu-util-common.c
superm1: OK i'll rename that page and make it runtime focused (as that's what the PR is).

src/fu-engine.c
superm1: sure, same problem elsewhere it seems too.

src/fu-engine.c
superm1: The other option is to only include these when plugin_dummy is compiled, but I think for distro purposes that's all the time, so I think I would rather keep them hardcoded.

src/fu-util-common.c
superm1: I adjusted it to model off of what was done in `fu-util.c` for a URL.

src/fu-util-common.c
superm1: fixed

src/fu-util-common.c
```diff
@@ -1610,9 +1616,26 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
...
 	}
+
+	if (low_help) {
+		/* TRANSLATORS:  this is instructions on how to improve the HSI security level */
+		g_string_append_printf (str, "\n%s %s",
+					_("Your system has a low HSI security level. Visit this URL for more information: "),
+					"https://github.com/fwupd/fwupd/wiki/Low-host-security-level");
+	}
+	if (runtime_help) {
+		/* TRANSLATORS:  this is instructions on how to improve the HSI suffix */
```
hughsie, line 1634: frustratingly, the TRANSLATORS line has to be the line below, it only gets found if it's one logical line above the _()

superm1: wow, that's annoying.  OK fixed.

src/fu-engine.c
hughsie: Ya, I think two is fine.

PR #2088, opened 2020-05-13
amt: Add a security attestation for provisioning

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/amt/fu-plugin-amt.c
+++ b/plugins/amt/fu-plugin-amt.c
@@ -17,6 +17,11 @@
 #include "fu-plugin-vfuncs.h"
 #include "fu-hash.h"
 
+struct FuPluginData {
+	gboolean			 has_mei;
+	gboolean			 provisioned;
+};
+
 typedef struct {
 	uuid_le guid;
 	guint buf_size;
@@ -444,8 +449,9 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(mei_context, mei_context_free)
 #pragma clang diagnostic pop
 
 static FuDevice *
-fu_plugin_amt_create_device (GError **error)
+fu_plugin_amt_create_device (FuPlugin *plugin, GError **error)
 {
+	FuPluginData *data = fu_plugin_get_data (plugin);
 	guint8 state;
 	struct amt_code_versions ver;
 	fwupd_guid_t uu;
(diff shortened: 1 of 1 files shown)


plugins/amt/fu-plugin-amt.c
```diff
@@ -544,15 +553,35 @@ void
...
+
+	if (!fu_common_is_cpu_intel ())
+		return;
+
+	attr = fwupd_security_attr_new ("com.intel.AMT");
+	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_SYSTEM_PROTECTION);
+	fwupd_security_attr_set_name (attr, "Intel AMT");
+	fu_security_attrs_append (attrs, attr);
+	if (data->provisioned)
+		fwupd_security_attr_set_result (attr, "provisioned");
```
hughsie, line 584: i think just `return` here and match the style of the other plugins.

plugins/amt/fu-plugin-amt.c
```diff
@@ -544,15 +553,35 @@ void
...
+
+void
+fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+	FuPluginData *data = fu_plugin_get_data (plugin);
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	if (!fu_common_is_cpu_intel ())
+		return;
+
```
hughsie, line 578: Don't we need to keep track if there are any AMT devices detected too? remember this vfunc gets called even if fu_plugin_amt_create_device() never got called.

plugins/amt/fu-plugin-amt.c
hughsie, line 584: I think the result needs to be sentance case, e.g. `Provisioned`

superm1: fixed

plugins/amt/fu-plugin-amt.c
superm1: yes, fixed

plugins/amt/fu-plugin-amt.c
superm1: fixed

PR #2090, opened 2020-05-14
synaptics-rmi: Essentially blacklist Dell K12A

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2052


jwrdegoede:
Hmm, I was hoping we could do a less device specific fix. Anyways lets just wait and see if other devices affected by the same issue pop up.

superm1:
> Hmm, I was hoping we could do a less device specific fix. Anyways lets just wait and see if other devices affected by the same issue pop up.

I still do think it would be good to get a root cause from Synaptics, especially if it was along the lines of "Oh we don't support RMA page access over this USB interface or these types of devices" to make it more generic while not having to turn the plugin into a whitelist only approach.

PR #2093, opened 2020-05-14
Add a new plugin for discovering if IOMMU is on the system

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


plugins/iommu/iommu.quirk
```diff
@@ -0,0 +1,3 @@
+# match all devices with this udev subsystem
+[DeviceInstanceId=IOMMU]
```
hughsie, line 2: I have an empty `/sys/class/iommu/` is that expected?

superm1: Do you have your IOMMU enabled?  It's only enabled if DMAR opt in bit was set to 1 or you manually enabled with `iommu=on` through kernel command line.

hughsie: opt-in is set, which is why I'm confused.

superm1: And modern (5.0+) kernel?  Check dmesg for iommu and dmesg references?

YehezkelShB: And make sure it's enabled in your system BIOS/EFI setup, if such a setting is available there

superm1: Vtd needs to be turned on.  It should have been if you have DMAR opt in set to 1.  If not I don't know how DMAR opt in could possibly be set to 1.

hughsie: VT-d is certainly turned on in the BIOS. I'm using 5.6.11-300.fc32.x86_64

    $ dmesg  | grep -i dmar
    [    0.017590] ACPI: DMAR 0x000000004FFCB000 0000A8 (v01 LENOVO TP-N1E   00001590 PTEC 00000002)
    [    0.248920] DMAR: Host address width 39
    [    0.248921] DMAR: DRHD base: 0x000000fed90000 flags: 0x0
    [    0.248925] DMAR: dmar0: reg_base_addr fed90000 ver 1:0 cap 1c0000c40660462 ecap 7e3ff0505e
    [    0.248926] DMAR: DRHD base: 0x000000fed91000 flags: 0x1
    [    0.248929] DMAR: dmar1: reg_base_addr fed91000 ver 1:0 cap d2008c40660462 ecap f050da
    [    0.248929] DMAR: RMRR base: 0x0000004f49f000 end: 0x0000004f4befff
    [    0.248930] DMAR: RMRR base: 0x0000005a000000 end: 0x0000005e7fffff
    [    0.248932] DMAR-IR: IOAPIC id 2 under DRHD base  0xfed91000 IOMMU 1
    [    0.248933] DMAR-IR: HPET id 0 under DRHD base 0xfed91000
    [    0.248933] DMAR-IR: x2apic is disabled because BIOS sets x2apic opt out bit.
    [    0.248934] DMAR-IR: Use 'intremap=no_x2apic_optout' to override the BIOS setting.
    [    0.250431] DMAR-IR: Enabled IRQ remapping in xapic mode

    $ dmesg  | grep -i iommu
    [    0.248932] DMAR-IR: IOAPIC id 2 under DRHD base  0xfed91000 IOMMU 1
    [    0.502957] iommu: Default domain type: Translated

hughsie: So, the reason it's empty is because you have to include a kernel parameter like https://scottlinux.com/2017/05/10/how-to-enable-iommu-support-in-fedora-linux/

Of course, adding that, rebooting and actually using iommu means my external display only shows a few lines of desktop and then all white. Now I know why it's not on by default...

Given it's disabled by default, I don't think `CRITICAL` is the right HSI level to assign. Maybe important at a push, but I'm erring more on theoretical myself.

YehezkelShB: Ah, so the actual parameter name is `intel_iommu`. And yes, it's possible that many device drivers didn't took it into account, so they don't play well with it. Additionally, it may have performance implications.
Nevertheless, it becomes more and more important for users, with threats like [thunderclap](http://thunderclap.io/) become more and more common, so I hope more efforts will be invested in making it to just work.

superm1: I still don't understand.

Due to this kernel patch:
Force IOMMU on for platform opt in hint

Which is part of 5.0 and later iommu should be enabled.

This might warrant some upstream discussion. Is it perhaps because your firmware has x2apic opt out mentioned in that message?

I do think this is a critical item, especially with USB4.

superm1: > it's possible that many device drivers didn't took it into account, so they don't play well with it

What graphics driver is this causing the behavior?

hughsie: This machine is a dual-GPU NVIDIA and Intel machine,so called "NV117 / Mesa Intel® HD Graphics 530 (SKL GT2)". I don't have the nvidia nonfree binary drivers installed. I assume that means I'm using the Intel GPU, as I can manually launch apps on the "dedicated graphics card".

hughsie: > Is it perhaps because your firmware has x2apic opt out mentioned in that message

No idea, I'm well out of my expertise level here.

> I do think this is a critical item, especially with USB4.

Random results from twitter: Nobody has /sys/class/iommu/ yet.

YehezkelShB: > Force IOMMU on for platform opt in hint

>Intel VT-d spec added a new DMA_CTRL_PLATFORM_OPT_IN_FLAG flag

According to [this](https://software.intel.com/sites/default/files/managed/c5/15/vt-directed-io-spec.pdf) (PDF!), it was added in rev2.5, Nov '17. Maybe not all BIOS vendors implemented it already? Especially as the machine is SKL (I assume it means Skylake, which is from 2015).

superm1: At least for Dell it doesn't start showing up until machines from 2019 and on.

hughsie: Right, some success. Using `intel_iommu=on,igfx_off` I get the expected `/sys/class/iommu/dmar1/` and a working iGPU. The question is, are we expecting users to change the grub command line to get a higher than HSI:0 value? I think things like `intel_iommu=on,igfx_off` need to be turned on by default in the kernel before we can be so strict.

superm1: >The question is, are we expecting users to change the grub command line to get a higher than HSI:0 value?

No, I don't expect that users should do anything.

> think things like intel_iommu=on,igfx_off need to be turned on by default in the kernel before we can be so strict.

I really think something is wrong with your system's firmware setup, and so the hand off isn't working properly and we need to determine what it is and capture that.

On my XPS 7390, no command line options, running a signed distro kernel with secure boot.
```
BOOT_IMAGE=/boot/vmlinuz-5.4.0-29-generic root=UUID=1f5f8d86-98e7-4c93-ab73-891051ab9a59 ro quiet splash vt.handoff=7
```

With this branch and default compile options I'm seeing:
```
$ sudo fwupdtool security --force
Loading…                 [-                                      ]13:40:23:0313 FuPluginUefi         missing signed bootloader for secure boot: /usr/local/libexec/fwupd/efi/fwupdx64.efi.signed cannot be found
Host Security ID: HSI:0+!

HSI-1
✘  SPI: Kernel support not present
✔  TPM: v2.0
✘  UEFI dbx: DBX can be downloaded from https://uefi.org/revocationlistfile and decompressed into /var/local/lib/fwupd/dbx:
✔  Input/Output Memory Management Unit: OK
✔  UEFI Secure Boot: Enabled

HSI-2
✔  TPM Reconstruction: Matched PCR0 reading

HSI-3
✔  Linux Kernel S3 Sleep: OK
✔  Suspend To Idle: OK
✘  Pre-boot kernel DMA protection: Unavailable
✘  Intel CET: Unavailable

Runtime Suffix -U
✘  Firmware Updates: No releases

Runtime Suffix -A
✘  Firmware Attestation: No PCR0s

Runtime Suffix -!
✔  fwupd plugins: OK
✘  Linux Swap: Not encrypted
✔  Linux Kernel Taint: OK
✔  Linux Kernel Lockdown: Locked down

Your system has a low HSI security level. Visit this URL for more information:  https://github.com/fwupd/fwupd/wiki/Low-host-security-level
Your system has runtime issues. Visit this URL for more information:  https://github.com/fwupd/fwupd/wiki/Host-security-ID-runtime-issues
```

IE the IOMMU is working properly and all my external monitors are working with i915 as it my Thunderbolt dock (which is behind the IOMMU now).

superm1: Hmmm...

>✘  Pre-boot kernel DMA protection: Unavailable

That's wrong.  It's there on my system.  IE:

>$ cat /sys/bus/thunderbolt/devices/domain0/iommu_dma_protection
1

hughsie: > I really think something is wrong with your system's firmware setup

Certainly plausible. What bothers me is there's precisely one Intel user with an out-of-the-box working iommu on my informal twitter poll. Maybe we compromise and make it `IMPORTANT`?

hughsie: > That's wrong. It's there on my system

Perhaps we got the DMAR parsing wrong? I thought iommu_dma_protection == opt_in()

superm1: >Certainly plausible. What bothers me is there's precisely one Intel user with an out-of-the-box working iommu on my informal twitter poll. Maybe we compromise and make it IMPORTANT?


I think lets get to the root cause of this first before re-assigning where this should land.  The whole point of the HSI is to drive security. 

>Perhaps we got the DMAR parsing wrong? I thought iommu_dma_protection == opt_in()

It should be.  Quite possibly got the parsing wrong.  Since your system has problems with IOMMU it sounds like likely it's not really active pre-boot.

superm1: OK found the problem.  Pushed a fix.

plugins/cpu/fu-cpu-device.c
```diff
@@ -10,30 +10,27 @@
 
 struct _FuCpuDevice {
 	FuDevice		 parent_instance;
-	gboolean		 has_shstk;
-	gboolean		 has_ibt;
+	FuCpuDeviceFlag		 flags;
```
hughsie, line 13: I think the two commits got squashed together.

superm1: Not sure how that happened, just rebased again.

plugins/iommu/fu-plugin-iommu.c
```diff
@@ -0,0 +1,53 @@
...
+
+void
+fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+	FuPluginData *data = fu_plugin_get_data (plugin);
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	/* create attr */
+	attr = fwupd_security_attr_new ("org.kernel.IOMMU");
+	fwupd_security_attr_set_level (attr, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL);
```
hughsie, line 42: Still not convinced it's critical. Move to important, and then work out what's going on out-of-band?

plugins/iommu/fu-iommu-device.c
```diff
@@ -0,0 +1,49 @@
...
+	return fu_udev_device_set_physical_id (device, "iommu", error);
+}
+
+static gboolean
+fu_iommu_device_setup (FuDevice *device, GError **error)
+{
+	const gchar *path = fu_udev_device_get_sysfs_path (FU_UDEV_DEVICE (device));
+	g_autofree gchar *basename = g_path_get_basename (path);
+	g_autofree gchar *instance = g_strdup_printf ("system-iommu-%s", basename);
+	fu_device_add_instance_id (FU_DEVICE (device), instance);
```
hughsie, line 29: Do we need an actual device in the daemon? Can't we do the same trick we did in the pci-bcr plugin with `fu_plugin_udev_device_added()`?

superm1: I think we can do that.  I'll switch it.

plugins/iommu/fu-plugin-iommu.c
superm1: Well it requires an OEM to implement a bunch of firmware development for this to work.  Like I said it's only on Dell's 2019 and newer systems, so I would suspect that it's only on pretty new Lenovo stuff too.

The driving reason behind it is external port security (see Thunderclap and Thunderspy disclosures), and that's why I think it's really important to include higher in the HSI spec.  It's a requirement for Microsoft's secured core PC initiative too (https://docs.microsoft.com/en-us/windows-hardware/design/device-experiences/oem-highly-secure).

I think a potential consolation is setting it critical only if there is Thunderbolt or USB4 on the system (I think just checking for /`sys/bus/thunderbolt` can be a way to decide), and lowering it to important otherwise.

It's certainly still valid for internal PCIe too (someone can pop off your chassis cover and access a PCIe or NVME slot), but the barrier to entry for the evil maid type attack is so much lower when you have a Thunderbolt port.

hughsie: I think Thunderclap and Thunderspy are in the theoretical class at the moment, i.e. you can't just go any buy something on ebay to trivally use those vulnerabilities, unlike having something like BIOSWE set which is one "sudo flashrom -w" from being pwned. I think it's perfectly fine to move "down" the HSI levels over time, as exploits move from theoretical to practical.

plugins/iommu/fu-plugin-iommu.c
```diff
@@ -0,0 +1,53 @@
...
+	fwupd_security_attr_set_name (attr, "Input/Output Memory Management Unit");
```
hughsie, line 43: I think `IOMMU` is a much better name fwiw. We don't expand TPM or SPI so it just looks weird.

plugins/iommu/fu-plugin-iommu.c
superm1: >I think Thunderclap and Thunderspy are in the theoretical class at the moment,

Pre-boot I can get behind it staying theoretical (you would need access while the system was booting or be able to restart it).

I've adjusted it in the PR to theoretical but i'm still on the fence.

>.e. you can't just go any buy something on ebay to trivally use those vulnerabilitie

https://www.ebay.com/sch/i.html?_from=R40&_nkw=Intel%C2%AE+Arria+Development+Kit&_sacat=0&LH_TitleDesc=0&rt=nc&LH_Sold=1&LH_Complete=1

And a guide how to use it: https://thunderclap.io/gettingstarted.html including a SD card image that has it all set up.

Seems pretty easy to me.

plugins/iommu/fu-plugin-iommu.c
superm1: fixed

plugins/iommu/fu-iommu-device.c
superm1: fixed

plugins/iommu/fu-plugin-iommu.c
hughsie: > Seems pretty easy to me.

Eek, that's more than theoretical, that's basically a HOWTO. Move to important then I guess, i.e. HSI:2

PR #2094, opened 2020-05-14
Wip/hughsie/pci bcr

hughsie:


src/fu-util-common.c
```diff
@@ -1552,7 +1552,9 @@ fu_util_remote_to_string (FwupdRemote *remote, guint idt)
 static void
 fu_security_attr_append_str (FwupdSecurityAttr *attr, GString *str)
 {
-	if (fwupd_security_attr_has_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+	if (fwupd_security_attr_has_flag (attr, FWUPD_SECURITY_ATTR_FLAG_OBSOLETED)) {
```
superm1, line 1555: I'd argue we probably want to sort this list by flags now when printing it.  Success/Fail/Obsoleted.

superm1:
> The kernel approach is preferable, as the config registers can be read even
when the PCI device is hidden from userspace.

This comment is confusing.  Is it really preferable?  It seems to me that it's only preferable in the cases that PCI is missing.  Otherwise this is preferable because it works on older stuff too!

I did check and it works properly on my CML laptop, good find!

src/fu-util-common.c
superm1: I pushed this

hughsie:
> it's only preferable in the cases that PCI is missing

Good point, fixed up.

> but you may check you like my sorting change

Good idea! I moved it to the daemon, I hope that's okay. It's the sort of order I'd want in gnome-firmware too, so it made sense to just return the attrs in a sane order.

PR #2097, opened 2020-05-14
trivial: if not specified try to use some better dbx defaults

superm1:
On Ubuntu and Fedora try to look for existing dbx database
directories.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -289,11 +289,16 @@ if build_standalone and get_option('plugin_uefi')
   efi_app_location = join_paths(libexecdir, 'fwupd', 'efi')
   conf.set_quoted ('EFI_APP_LOCATION', efi_app_location)
 
-  # e.g. could be:
-  # * /usr/share/dbxtool for Red Hat
-  # * /usr/share/secureboot/updates/dbx for Ubuntu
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -289,11 +289,23 @@ if build_standalone and get_option('plugin_uefi')
   efi_app_location = join_paths(libexecdir, 'fwupd', 'efi')
   conf.set_quoted ('EFI_APP_LOCATION', efi_app_location)
 
-  # e.g. could be:
-  # * /usr/share/dbxtool for Red Hat
-  # * /usr/share/secureboot/updates/dbx for Ubuntu
   efi_dbxdir = get_option('efi_dbxdir')
+  if efi_dbxdir == ''
+    lsb = find_program('lsb_release', required : false)
```
hughsie, line 294: why parse LSB release? Can't we just try `/usr/share/secureboot/updates/dbx` and then `/usr/share/dbxtool`? I don't think either distro is going to provide both.

superm1: Oh good point. I'll redo.

superm1: fixed

PR #2098, opened 2020-05-15
cpu: Add support for a security attribute related to Intel TME

superm1:
This only checks that it was available from the CPU.
To be complete an additional check should be made to show that it
was actually enabled from the firmware.

This will require a kernel modification though because MSR access
will be forbidden from userland while in kernel lockdown.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
I rewrote this a bit I hope that's okay. Please check, and then LGTM.

PR #2100, opened 2020-05-15
Wip/hughsie/fwupd security attr set plugin

hughsie:


hughsie:
> so that it doesn't need to be exported directly to clients

Ohh, I guess. You just mean drop it from the fwupd_security_attr_to_variant() functions.

PR #2101, opened 2020-05-15
pci-mei: Check the ME device is not in manufacturing mode

hughsie:


superm1:
I added a CML ME to the list `8086:02e0`.  This particular one is is not production so it's possible that certain things aren't fused and this is correct, but it is showing up as manufacturing mode.

hughsie:
> I added a CML ME to the list 8086:02e0

Do you want to push that? Out of interest, what did lspci say?

superm1:
I gotta run, but it's:
```
00:16.0 Communication controller [0780]: Intel Corporation Comet Lake Management Engine Interface [8086:02e0]
```

superm1:
Oh forgot.  The list of stuff that is an ME is here to sync up:

https://github.com/torvalds/linux/blob/0ad5b053d438990fabaa324499abb6131b9d2202/drivers/misc/mei/hw-me-regs.h

hughsie:
>  The list of stuff that is an ME is here to sync up

Super useful, thanks.

superm1:
I did check and this is working properly on a canon point laptop I have on hand.

My comet lake one doesn't, but I suspect it's actually because it's not production it hasn't been fused.

PR #2103, opened 2020-05-16
trivial: add versioning into HSI measurement

superm1:
Show the fwupd version in the HSI string so items can be cross
referenced to the spec from when they were first added

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Should we build the version into the output of fu_engine_get_host_security_id() instead?

PR #2104, opened 2020-05-17
Wip/hughsie/hsi fixes

hughsie:


hughsie:
Untested; I'm on the wrong laptop to test.

superm1:
If translating ok and fail, I wonder if it's worth trying to make more plugins show that for success and fail rather than details shown now. IE make the description in the title rather than the result.

hughsie:
> If translating ok and fail

superm1:
> ✔  MEI detected

ME Manufacturing mode disabled

>> ✔  Something

Does making it instead like this come across better?

 Something : ✔ 

IE change the OK/Failed into the check / X but leave it in the same place.

hughsie:
> Does making it instead like this come across better?

I think all this is complicated enough for the user. I think we do need to say if something is "locked" or "unlocked" if nothing else so they can file a good bug report. PR 90% done with an idea to clean everything up a bit.

hughsie:
Obsoleted by https://github.com/fwupd/fwupd/pull/2106

PR #2106, opened 2020-05-18
Allow client tools to translate the HSI attributes and results

hughsie:
To do this, rely on the AppStream ID to map to a translated string (providing a
fallback for clients that do not care) and switch the free-form result string
into a set of enumerated values that can be translated.

This fixes some of the problems where some things have to be enabled to "pass"
and other attributes have to be some other state. For cases where we want the
user to "do" something, provide a URL to a wiki page that we update out-of-band
of fwupd releases.

### DIFF
--- a/libfwupd/fwupd-security-attr.c
+++ b/libfwupd/fwupd-security-attr.c
@@ -25,8 +25,9 @@ typedef struct {
 	GPtrArray			*obsoletes;
 	gchar				*name;
 	gchar				*plugin;
-	gchar				*result;
+	gchar				*url;
 	FwupdSecurityAttrLevel		 level;
+	FwupdSecurityAttrResult		 result;
 	FwupdSecurityAttrFlags		 flags;
 } FwupdSecurityAttrPrivate;
 
@@ -58,6 +59,50 @@ fwupd_security_attr_flag_to_string (FwupdSecurityAttrFlags flag)
 		return "runtime-attestation";
 	if (flag == FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE)
 		return "runtime-issue";
+	return NULL;
+}
+
+/**
+ * fwupd_security_attr_result_to_string:
+ * @result: A #FwupdSecurityAttrResult, e.g. %FWUPD_SECURITY_ATTR_RESULT_ENABLED
+ *
+ * Returns the printable string for the result enum.
+ *
+ * Returns: string, or %NULL
+ *
+ * Since: 1.5.0
+ **/
+const gchar *
+fwupd_security_attr_result_to_string (FwupdSecurityAttrResult result)
+{
+	if (result == FWUPD_SECURITY_ATTR_RESULT_VALID)
+		return "valid";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_VALID)
+		return "not-valid";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_ENABLED)
+		return "enabled";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED)
+		return "not-enabled";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_LOCKED)
+		return "locked";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_LOCKED)
+		return "not-locked";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_ENCRYPTED)
+		return "encrypted";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_ENCRYPTED)
+		return "not-encrypted";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_TAINTED)
+		return "tainted";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_TAINTED)
+		return "not-tainted";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_FOUND)
+		return "found";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND)
+		return "not-found";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_SUPPORTED)
+		return "supported";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_SUPPORTED)
+		return "not-supported";
 	return NULL;
 }
 
@@ -105,7 +150,7 @@ fwupd_security_attr_get_obsoletes (FwupdSecurityAttr *self)
 /**
  * fwupd_security_attr_add_obsolete:
  * @self: A #FwupdSecurityAttr
- * @appstream_id: the appstream_id
+ * @appstream_id: the appstream_id or plugin name
  *
  * Adds an attribute appstream_id to obsolete. The obsoleted attribute will not
  * contribute to the calculated HSI value or be visible in command line tools.
@@ -180,26 +225,31 @@ fwupd_security_attr_set_appstream_id (FwupdSecurityAttr *self, const gchar *apps
 {
 	FwupdSecurityAttrPrivate *priv = GET_PRIVATE (self);
 	g_return_if_fail (FWUPD_IS_SECURITY_ATTR (self));
+
+	/* sanity check */
+	if (!g_str_has_prefix (appstream_id, "org.fwupd.hsi."))
+		g_critical ("HSI attributes need to have a 'org.fwupd.hsi.' prefix");
+
 	g_free (priv->appstream_id);
 	priv->appstream_id = g_strdup (appstream_id);
 }
 
 /**
- * fwupd_security_attr_get_result:
+ * fwupd_security_attr_get_url:
  * @self: A #FwupdSecurityAttr
  *
- * Gets the attribute result.
+ * Gets the attribute URL.
  *
  * Returns: the attribute result, or %NULL if unset
  *
  * Since: 1.5.0
  **/
 const gchar *
-fwupd_security_attr_get_result (FwupdSecurityAttr *self)
+fwupd_security_attr_get_url (FwupdSecurityAttr *self)
 {
 	FwupdSecurityAttrPrivate *priv = GET_PRIVATE (self);
 	g_return_val_if_fail (FWUPD_IS_SECURITY_ATTR (self), NULL);
-	return priv->result;
+	return priv->url;
 }
 
 /**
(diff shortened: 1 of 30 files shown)


hughsie:
Screenshot from 2020-05-18 14-29-58

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
```diff
@@ -28,28 +28,27 @@ fu_plugin_add_security_attr_bioswe (FuPlugin *plugin, FuSecurityAttrs *attrs)
...
-	fwupd_security_attr_add_obsolete (attr, "org.fwupd.plugin.pci-bcr");
+	fwupd_security_attr_add_obsolete (attr, "org.fwupd.hsi.pci-bcr");
 	fu_security_attrs_append (attrs, attr);
 
 	/* load file */
 	fn = g_build_filename (FU_PLUGIN_LINUX_SPI_LPC_SYSFS_DIR, "bioswe", NULL);
 	if (!g_file_get_contents (fn, &buf, &bufsz, &error_local)) {
 		g_warning ("could not open %s: %s", fn, error_local->message);
-		fwupd_security_attr_set_result (attr, "Could not open file");
+		fwupd_security_attr_set_url (attr, "Could not open file");
```
superm1, line 41: This doesn't make sense as a URL

libfwupd/fwupd-security-attr.c
```diff
@@ -61,6 +62,50 @@ fwupd_security_attr_flag_to_string (FwupdSecurityAttrFlags flag)
...
+ *
+ * Since: 1.5.0
+ **/
+const gchar *
+fwupd_security_attr_result_to_string (FwupdSecurityAttrResult result)
+{
+	if (result == FWUPD_SECURITY_ATTR_RESULT_VALID)
+		return "valid";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_VALID)
+		return "not-valid";
```
superm1, line 81: why not-valid instead of invalid?

libfwupd/fwupd-security-attr.c
```diff
@@ -61,6 +62,50 @@ fwupd_security_attr_flag_to_string (FwupdSecurityAttrFlags flag)
...
+	if (result == FWUPD_SECURITY_ATTR_RESULT_ENABLED)
+		return "enabled";
+	if (result == FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED)
+		return "not-enabled";
```
superm1, line 85: also why not-enabled instead of disabled?

plugins/linux-spi-lpc/fu-plugin-linux-spi-lpc.c
hughsie: Fixed, thanks.

libfwupd/fwupd-security-attr.c
hughsie: I was trying to keep the boolean states symmetric and possibly invertable in the future. I've used the "invalid" form in the user-visble text.

superm1: ah OK.

hughsie:
@superm1 the other thing we could do is align the result:

    Host Security ID: HSI:1

    HSI-1
    ✔  BIOS region of SPI:            Enabled
    ✔  MEI manufacturing mode:        Locked
    ✔  SPI lock:                      Enabled
    ✔  SPI write:                     Disabled
    ✔  TPM v2.0:                      Enabled
    ✔  UEFI dbx:                      Found
    ✔  UEFI secure boot:              Enabled

    HSI-2
    ✔  TPM PCR0 reconstruction:       Valid
    ✘  IOMMU:                         Not found

    HSI-3
    ✘  Intel CET:                     Not supported
    ✘  Pre-boot kernel DMA protection:Disabled
    ✘  Suspend to idle:               Disabled
    ✘  Suspend-to-ram:                Enabled

    HSI-4
    ✔  Intel AMT:                     Disabled
    ✘  Intel TME:                     Not supported

    Runtime Suffix -U
    ✘  Firmware updates:              Not supported

    Runtime Suffix -A
    ✘  Firmware attestation:          Not supported

    Runtime Suffix -!
    ✔  Linux kernel:                  Untainted
    ✔  Linux kernel lockdown:         Enabled
    ✔  Linux swap:                    Encrypted
    ✔  fwupd plugins:                 Untainted

Or maybe make the result bold or something.

superm1:
I do think that aligning the result looks better.  I was thinking that same thing recently.

hughsie:
Or we make the label the success/failure color?

Screenshot from 2020-05-18 15-44-05

I can imagine the color blind people might get upset with that.

superm1:
> I can imagine the color blind people might get upset with that.

I like the coloring, but if it would upset colorblind people, perhaps add the check/X and colored wording both?

src/fu-security-attr.c
```diff
@@ -0,0 +1,174 @@
...
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI write");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI lock");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("BIOS region of SPI");
```
superm1, line 26: Does this read better as `SPI BIOS region` to match the others?

hughsie: It does, fixed.

hughsie:
tick at the result, or at the start?

Screenshot from 2020-05-18 15-51-14

src/fu-security-attr.c
```diff
@@ -0,0 +1,174 @@
...
+		/* TRANSLATORS: Title: AMT = Active Management Technology */
+		return _("Intel AMT");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET) == 0) {
+		/* TRANSLATORS: Title: CET = Control-flow Enforcement Technology */
+		return _("Intel CET");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_TME) == 0) {
+		/* TRANSLATORS: Title: TME = Total Memory Encryption */
+		return _("Intel TME");
```
superm1, line 42: Although it's intel specific right now the way we implemented, can we leave intel out of this?  AMD and others should have their own and I think it's the same way we should show it.

hughsie: `s\FWUPD_SECURITY_ATTR_ID_INTEL_TME\FWUPD_SECURITY_ATTR_ID_INTEL_ENCRYPTED_MEMORY` ? Then we have `Supported` and `Unsupported.

src/fu-security-attr.c
```diff
@@ -0,0 +1,174 @@
...
+		/* TRANSLATORS: Title: swap spaceor aswap partition */
+		return _("Linux swap");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_RAM) == 0) {
+		/* TRANSLATORS: Title: sleep state */
+		return _("Suspend-to-ram");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_IDLE) == 0) {
+		/* TRANSLATORS: Title: a better sleep state */
+		return _("Suspend to idle");
```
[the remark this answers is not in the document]
hughsie: Good catch, thanks.

superm1:
> tick at the result, or at the start?

At the start instead of the * I personally like better

hughsie:
>  At the start instead of the * I personally like better

Screenshot from 2020-05-18 15-57-59

src/fu-security-attr.c
superm1: Well i was hoping to make it so that we don't have intel at all in the attribute and name.   AMD has SME and SEV, and I think the same attribute should be shared for them too, no?

superm1:
> > At the start instead of the * I personally like better
> 
> Screenshot from 2020-05-18 15-57-59

Yep, looks good to me that way.

src/fu-security-attr.c
hughsie: Gahh, that was supposed to be `FWUPD_SECURITY_ATTR_ID_ENCRYPTED_MEMORY` sorry

hughsie: Maybe memory isn't specific enough. `_DRAM`?

superm1: Yeah that's good

hughsie:
I also have to fix the obsoletes stuff too, as that's now broken of course.

src/fu-security-attr.c
```diff
@@ -0,0 +1,174 @@
...
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_LOCKDOWN) == 0) {
+		/* TRANSLATORS: Title: lockdown is a security mode of te kernel */
+		return _("Linux kernel lockdown");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_TAINTED) == 0) {
+		/* TRANSLATORS: Title: if it's tainted or not */
+		return _("Linux kernel");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_SWAP) == 0) {
+		/* TRANSLATORS: Title: swap spaceor aswap partition */
```
superm1, line 57: missing spaces in this comment

src/fu-security-attr.c
```diff
@@ -0,0 +1,174 @@
...
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI write");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI lock");
+	}
+	if (g_strcmp0 (appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI BIOS region");
```
superm1, line 26: I think this needs to show something regarding writes to this region if it's using enabled/disabled.

hughsie: Maybe convert to Locked/Unlocked?

superm1: Yeah that works.

plugins/tpm/fu-plugin-tpm.c
```diff
@@ -41,23 +41,22 @@ fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
...
 	if (!data->has_tpm_v20) {
-		fwupd_security_attr_set_result (attr, "Not in v2.0 mode");
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED);
 		return;
 	}
 
 	/* success */
 	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
-	fwupd_security_attr_set_result (attr, "v2.0");
+	fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_ENABLED);
```
superm1, line 61: I think this should actually be found, looking again.  We have no guarantees that anything in the OS is using it, just that the platform has it.

hughsie: Fixed. Feel free to fix up any of these as required post merge.

PR #2109, opened 2020-05-18
pci-bcr: Read the ISA bridge BCR from the PCI device class

hughsie:
The SPI controllers are always identified with one of two device classes.


superm1:
That always works?  What sets this class up?

superm1:
Actually it fails on my comet lake U system.  The class is 0780 there.

superm1:
I have a better idea to propose though if you want to get out of the business of keeping up with quirks.  What about you read the symlink in `driver` for the PCI device?  It will always point to `mei_me` for the ME.

superm1:
> Actually it fails on my comet lake U system. The class is 0780 there.

And this also falls apart on a future unreleased system that has multiple PCI devices with the exact same class.  I did check though and only one of them does match to `mei_me` for it's driver though.

hughsie:
>  It will always point to mei_me

This is for the SPI BCR register.

superm1:
> > It will always point to mei_me
> 
> This is for the SPI BCR register.

Oh.... Sorry I was mixing those two up.

hughsie:
>  What sets this class up

This is the standard PCI class attribute defined by all devices: https://pci-ids.ucw.cz/read/PD/

superm1:
OK I see.  I think you still want to be matching the vendor though in this case.  Not all ISA bridges will be made the same.  If you encounter one on AMD it won't have this for example.

superm1:
>And this also falls apart on a future unreleased system that has multiple PCI devices with the exact same class. I did check though and only one of them does match to mei_me for it's driver though.

BTW my idea for the mei driver I think we should still consider outside this PR.  What about just setting it up to look at the driver in use rather than hardcoding the list of device IDs that were cloned from the kernel?

hughsie:
>  I think you still want to be matching the vendor though in this case.

The *device* vendor or the *bus* vendor?

superm1:
> > I think you still want to be matching the vendor though in this case.
> 
> The _device_ vendor or the _bus_ vendor?

Device vendor I would think.

hughsie:
> Device vendor I would think.

Okay, udpated to use the vendor too -- that fixes the SPI detection on my (your?) XPS13 and a really old X220 with coreboot.

superm1:
> > Device vendor I would think.
> 
> Okay, udpated to use the vendor too -- that fixes the SPI detection on my (your?) XPS13 and a really old X220 with coreboot.

Yeah works on my unreleased future thing as well as my CML XPS 13 too.

PR #2110, opened 2020-05-19
Simplify ME detection

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-udev-device.c
```diff
@@ -335,6 +335,18 @@ fu_udev_device_probe (FuDevice *device, GError **error)
 						FU_DEVICE_INSTANCE_FLAG_ONLY_QUIRKS);
 	}
 
+	/* add the driver */
+	tmp = g_udev_device_get_driver (priv->udev_device);
+	if (tmp != NULL) {
+		g_autofree gchar *devid = NULL;
+		devid = g_strdup_printf ("%s\\VEN_%04X&DRIVER_%s",
+					 subsystem,
+					 priv->vendor,
```
hughsie, line 344: do we need the vendor? I mean, a kernel module is supposed to provide an abstracted interface that's vendor neutral. I know in this case the MEI interface is Intel specific, but in the general case it's not.

superm1: My thought was later on we might have a case that multiple devices use the same driver and you want to differentiate.  However I suppose we can always change this later and add code when that happens.

PR #2111, opened 2020-05-19
synaptics-prometheus: Force the minor version from 0x02 to 0x01 to make

vhuag:
sure the devices can be updated back to 0x01.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


vhuag:
We are going to upload minor version 1 firmware to LVFS, so request to force the minor version from 2 to 1 to make sure new firmware can be updated.

vhuag:
We will skip 10.02 for all device IDs, but we can add major number 10 to the condition.

PR #2113, opened 2020-05-19
Export the host vendor and family

hughsie:
The 'product name' is not typically what the hardware is known as. We need the
vendor and family if the user is going to recognise the hardware.

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -45,6 +45,9 @@ typedef struct {
 	guint				 percentage;
 	gchar				*daemon_version;
 	gchar				*host_product;
+	gchar				*host_family;
+	gchar				*host_sku;
+	gchar				*host_vendor;
 	gchar				*host_machine_id;
 	gchar				*host_security_id;
 	GDBusConnection			*conn;
@@ -67,6 +70,9 @@ enum {
 	PROP_DAEMON_VERSION,
 	PROP_TAINTED,
 	PROP_HOST_PRODUCT,
+	PROP_HOST_FAMILY,
+	PROP_HOST_SKU,
+	PROP_HOST_VENDOR,
 	PROP_HOST_MACHINE_ID,
 	PROP_HOST_SECURITY_ID,
 	PROP_INTERACTIVE,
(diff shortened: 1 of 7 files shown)


hughsie:
(this will also be used for `upload-security` which shares the security attributes with the LVFS, and I need more data to match the systems up)

superm1:
> The 'product name' is not typically what the hardware is known as. 

Well for Dell it is, I guess other manufacturers might not fill out this field as user friendly?

I guess in a related PR https://github.com/systemd/systemd/issues/15493 is happening.  When this happens it might be better to rely upon rather than a potential heuristic in two places.

hughsie:
On Lenovo it's vendor=LENOVO family="ThinkPad P50" product=20EQS64N0C -- which kinda makes sense as there's lots of "different" P50s with different options.

superm1:
I suspect that if you're going to be uploading and showing all of these, you also want `product_sku`

hughsie:
Which for Lenovo is the uncomprehensible `LENOVO_MT_20EQ_BU_Think_FM_ThinkPad P50`. I don't think the user cares much about the SKU or product; most people don't know that there are multiple "versions" of a P50 for instance.

superm1:
Product SKU can actually be quite important when disambiguating results.  Even if it's not shown in the UX on LVFS side, I think it's better to have the data available.

You might have two systems that share the same marketing name but can be offered with a -U and a -H CPU.  They are most likely running different platform firmware and can have a different class of security, features and bugs.  I'm not saying that is necessarily how common that is, but if we don't have the data uploaded we'll have no way to find out when that's the case.

superm1:
Another point about not just showing the product family; for Dell this will show up as "XPS".  There are absolutely new security features in the XPS 13 7390 that are not present in the XPS 13 9380 and more that will be present in future XPS laptops.  I want people to be able to see the difference on LVFS when certain laptops measure more highly for security.

hughsie:
SKU added. I think we might quirk that we need the SKU for Dell and the Family for Lenovo. What's the exact "vendor" string for Dell hardware? just `Dell` or something like `Dell Inc."?

superm1:
Here's from my system on hand in dmidecode.

```
System Information
        Manufacturer: Dell Inc.
        Product Name: XPS 13 7390
        Version: Not Specified
        Serial Number: 3M466Y2
        UUID: 4c4c4544-004d-3410-8036-b3c04f365932
        Wake-up Type: Power Switch
        SKU Number: 0962
        Family: XPS
```

superm1:
> I think we might quirk on the LVFS that we need the SKU for Dell and the Family for Lenovo. What's the exact "vendor" string for Dell hardware? just `Dell` or something like `Dell Inc."?

Given the systemd pull request I referenced was discussing special casing Lenovo, I think that if you're going to put quirks on LVFS the default should probably be to use product name and if you see something Lenovo to quirk it to family.  I do think the quirks make most sense on the LVFS side so that if a manufacturer chooses to change how they use SMBIOS strings it doesn't break things needing a client change.

> SKU added. 

Thanks.

libfwupd/fwupd-client.c
```diff
@@ -2143,6 +2266,39 @@ fwupd_client_class_init (FwupdClientClass *klass)
...
+	 * Since: 1.5.0
+	 */
+	pspec = g_param_spec_string ("host-family", NULL, NULL,
+				     NULL, G_PARAM_READABLE | G_PARAM_STATIC_NAME);
+	g_object_class_install_property (object_class, PROP_HOST_FAMILY, pspec);
+
+	/**
+	 * FwupdClient:host-sku:
+	 *
+	 * The host family string, e.g. "ABC12345"
```
superm1, line 2283: SKU

hughsie:
> XPS 13 7390

src/fu-util.c
```diff
@@ -528,7 +528,10 @@ fu_util_build_device_tree (FuUtilPrivate *priv, GNode *root, GPtrArray *devs, Fw
 static gchar *
 fu_util_get_tree_title (FuUtilPrivate *priv)
 {
-	return g_strdup (fwupd_client_get_host_product (priv->client));
+	return g_strdup_printf ("%s - %s %s",
+				fwupd_client_get_host_product (priv->client),
+				fwupd_client_get_host_vendor (priv->client),
+				fwupd_client_get_host_family (priv->client));
```
superm1, line 534: This is now pretty verbose.

```
XPS 13 7390 - Dell Inc. XPS
│
```

Also the same thing needs to be set in `fu-tool.c` to stay consistent with `fwupdtool` and `fwupdmgr`

superm1:
> > XPS 13 7390
> 
> Out of interest, Is there only one type of XPS 13 7390 or do we always need to show the SKU too?

There is only one XPS 13 7390.  I don't think we should be showing the SKU in **most places**, I just want to have it on hand if/when we get to a situation that we need it.

src/fu-util.c
hughsie: Yup, I'll just drop this hunk for now.

PR #2115, opened 2020-05-19
Allow uploading security attributes to the LVFS

hughsie:
We sign the data with the client cert to allow users with LVFS accounts to
publish 'official' HSI ratings.


hughsie:
Of course, the alternative is just ask every time after just running `security` -- but we have no way of storing the "don't ever ask again" preference -- and it does seem a little pushy.

src/fu-util.c
```diff
@@ -2396,6 +2396,165 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	hash = fwupd_get_os_release (error);
+	if (hash == NULL)
+		return FALSE;
+	for (guint i = 0; distro_kv[i].key != NULL; i++) {
+		const gchar *tmp = g_hash_table_lookup (hash, distro_kv[i].key);
+		if (tmp != NULL) {
+			json_builder_set_member_name (builder, distro_kv[i].val);
+			json_builder_add_string_value (builder, tmp);
+		}
+	}
```
superm1, line 2423: So I think for analyzing how accurate things are you'll probably want `/proc/cmdline` too.

hughsie: Do we also want that for "normal" firmware reports?

superm1: I don't think it's needed there.  The reason for this is if someone has HSI:0 and you see they have `iommu=off`, potentially LVFS can either filter this or return back a message about something they can do.

superm1:
> Of course, the alternative is just ask every time after just running security -- but we have no way of storing the "don't ever ask again" preference -- and it does seem a little pushy.

Maybe store to the history database a new column on security, and then if it's changed you can be pushy?

The other thing that can be done is add automatic uploads like happens for reports and is opt-in the first time you run security.

src/fu-util.c
```diff
@@ -2396,6 +2396,165 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INVALID_FILE,
+			     "Failed to upload to %s: %s",
+			     report_uri, soup_status_get_phrase (status_code));
+		return FALSE;
+	}
+
+	/* TRANSLATORS: success, so say thank you to the user */
+	g_print ("%s\n", "Host Security ID attributes uploaded successfully, thanks!"));
```
superm1, line 2554: too many `)`

hughsie: Gotcha, fixed.

src/fu-util.c
hughsie: Hmm, there's quite a lot of personal-ish data there. 

For instance:

`BOOT_IMAGE=(hd2,gpt2)/vmlinuz-5.6.11-300.fc32.x86_64 root=UUID=01234dff-c7d0-4e29-b312-888e710ec39b ro resume=/dev/mapper/luks-012345-5c97-427c-87d8-e6c46c662562 rd.luks.uuid=luks-012345-5c97-427c-87d8-e6c46c662562 rhgb quiet`

I'm wondering if we should split it by spaces in the client tool and then whitelist/backlist some stuff. For instance, blacklisting `BOOT_IMAGE`, `root`, `resume`, `rd.luks.uuid` etc

superm1: Yes I think that makes sense to filter out some of those things we don't care about and could be sensitive.  I would prefer to blacklist items than whitelist, because we don't know yet what we don't know in terms of what people are doing that decreases the security on their system from `/proc/cmdline`.

hughsie: Okay, added.

hughsie:
> opt-in the first time you run security

I was thinking about this some more, and I think you're right. The number of users bothering to use `fwupdmgr security` is going to be small, and those knowing they need to run `fwupdmgr upload-security` is going to be even smaller. Pushing the upload nag into `security` gets the maximum data. 

Maybe we can use `--no-unreported-check` to prevent the nag for scripts that are using this functionality, although I suppose scripts could just use `fwupdtool` (which can't safely upload data, as it's running as root).

src/fu-util.c
```diff
@@ -2424,6 +2427,63 @@ fu_util_upload_security_build_metadata (JsonBuilder *builder, GError **error)
...
+	g_autofree gchar *buf = NULL;
+#ifdef HAVE_UTSNAME_H
+	struct utsname uname_tmp;
+#endif
+	const gchar *ignore[] = {
+		"",
+		"BOOT_IMAGE",
+		"console",
+		"quiet",
+		"quiet",
```
superm1, line 2412: double quiet

hughsie: Fixed, thanks. Twitter replies still coming in, so I might add a few more things here.

src/fu-util.c
```diff
@@ -2617,6 +2578,46 @@ fu_util_upload_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	str = fu_util_security_attrs_to_string (attrs);
+	g_print ("%s\n", str);
+
+	/* ask if we can upload them to the LVFS */
+	if (!priv->assume_yes) {
+		g_print ("\n%s [y|N]: ",
+			/* TRANSLATORS: ask the user to share */
+			 _("Can we upload these anonymous results to the LVFS?"));
+		if (!fu_util_prompt_for_boolean (FALSE))
+			return TRUE;
```
superm1, line 2616: What do you think about the idea for having a config file option to automatically do this and the client can change the daemon setting like we do for reports?  I think it only makes sense if security attributes are cached and only upload for changes though.

superm1: But part of my thought is that if that's done a GUI client could potentially be sourcing and using the same option like we have configured the update reports to be able to be done in a GUI tool.

hughsie: Aha, we have `ModifyConfig()` - sure, will give it a try.

superm1: And also `fu_util_maybe_enable_automatic` which can hopefully be re-used

hughsie: Actually, maybe we want to tie this to a remote? it means I don't have to hardcode the URL for the LVFS... and if there's no enabled remotes then we don't offer to upload it at all. Then we use te more conventional `ModifyRemote()`....

hughsie: And even further down the rabbithole; do we want

    @@ -5,6 +5,8 @@ Enabled=true
     Title=Linux Vendor Firmware Service
     MetadataURI=https://cdn.fwupd.org/downloads/firmware.xml.gz
     ReportURI=https://fwupd.org/lvfs/firmware/report
    +SecurityURI=https://fwupd.org/lvfs/hsireports/upload
     OrderBefore=fwupd
     AutomaticReports=false
    +AutomaticHsiReports=false
     ApprovalRequired=false

Or are security reports the "same" as firmware reports?

superm1: Yes I think that's perfect.  Then someone can opt into upload reports and security reports separately.  My vision for this is that something like gnome-software will send upload reports and something like gnome control center will send security reports.  Those "may" be separate opt in events sending separate data, so it's good to treat them as such.

hughsie: `fwupd_remote_get_security_uri` or `fwupd_remote_get_hsi_uri` or `fwupd_remote_get_security_report_uri`

superm1: `fwupd_remote_get_security_report_uri` or `fwupd_remote_get_security_uri` to me.  Remember the command is currently `security`, not `hsi` so the conf options should match that too.

hughsie: I've added this, although I opted to ask multiple questions rather than use `fu_util_maybe_enable_automatic()` to improve the flow a bit.

src/fu-util.c
```diff
@@ -2393,6 +2639,60 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	if (remote == NULL) {
+		g_debug ("failed to find suitable remote: %s", error_local->message);
+		return TRUE;
+	}
+
+	/* ask if we can upload them to the LVFS */
+	if (priv->no_unreported_check)
+		return TRUE;
+	if (!priv->assume_yes ||
+	    !fwupd_remote_get_automatic_security_reports (remote)) {
```
superm1, line 2654: But so don't you want to make sure it hasn't changed?  Otherwise users will upload every time they view security.

src/fu-util.c
```diff
@@ -2393,6 +2639,60 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+		/* TRANSLATORS: ask the user to share, %s is something like:
+		 * "Linux Vendor Firmware Service" */
+		tmp = g_strdup_printf ("Upload these anonymous results to the %s to help other users?",
+				       fwupd_remote_get_title (remote));
+
+		g_print ("\n%s [y|N]: ", tmp);
+		if (!fu_util_prompt_for_boolean (FALSE)) {
+			g_print ("%s [y|N]: ",
+				 /* TRANSLATORS: stop nagging the user */
+				 _("Disable the auto-upload functionality?"));
```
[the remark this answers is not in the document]
hughsie: Fixed.

src/fu-util.c
hughsie: I pondered this. I wondered if we should store this in the remote config file, or on the daemon, or somewhere else entirely. I'm somewhat regretting not returning metadata from the engine when getting the attr list. I guess we could just store the sha256 checksum in ~/cache somewhere, but it seems like the kind of thing the fwupd daemon should know.

superm1: > I'm somewhat regretting not returning metadata from the engine when getting the attr list

This is why it's good 1.5.0 isn't released yet while figuring all this stuff out.  I don't see a problem with changing the functions and symbols right now.

>I guess we could just store the sha256 checksum in ~/cache somewhere, but it seems like the kind of thing the fwupd daemon should know.

I agree it should be kept in the daemon not home directory.  I was thinking all the attributes and the end result would both make sense to go into the history database.  A logical extension then ends up being for the client to look at the history of security attribute runs.

The client command `security` could potentially then indicate when security has gone up or down too.

hughsie: Thinking more about this, I don't think re-uploading the security report is such a bad thing. Some rationalle:

 * If the user got a low score (e.g. SB off, or non-encrypted swap) they might "fix" the issue and then re-get the security level. We definitely want the new report.

* If the user updates the distro/kernel and something like lockdown changes then we want the new report too.

* If the user updates the firmware, we definately want the new report as the IOMMU might have just turned up.

Having a "cache" even if a SHA256 of the blob means you have to know when to invalidate it.

superm1: Re-uploading the security report in all of those scenarios makes perfect sense to me.  I don't disagree about re-uploading the report being bad, I just think it should only happen if it has changed in some way.

src/fu-util.c
```diff
@@ -2365,11 +2365,169 @@ fu_util_modify_config (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	if (priv->sign) {
+		sig = fwupd_client_self_sign (priv->client, data,
+					      FWUPD_SELF_SIGN_FLAG_ADD_TIMESTAMP,
+					      priv->cancellable, error);
+		if (sig == NULL)
+			return FALSE;
+	}
+
+	/* ask for permission */
+	if (!priv->assume_yes) {
```
superm1, line 2458: aren't you missing the stuff that lets this go automatic here?

src/fu-util.c
```diff
@@ -2393,6 +2551,60 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+
+	/* can we find a remote with a security attr */
+	remote = fu_util_get_remote_with_security_report_uri (priv, &error_local);
+	if (remote == NULL) {
+		g_debug ("failed to find suitable remote: %s", error_local->message);
+		return TRUE;
+	}
+
+	/* ask if we can upload them to the LVFS */
+	if (priv->no_unreported_check)
```
superm1, line 2563: I think this should go before we look for a remote that supports it

src/fu-util.c
```diff
@@ -2393,6 +2551,60 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
...
+			 /* TRANSLATORS: can we JFDI? */
+			 _("Automatically upload every time?"));
+		if (fu_util_prompt_for_boolean (FALSE)) {
+			if (!fwupd_client_modify_remote (priv->client,
+							 fwupd_remote_get_id (remote),
+							 "AutomaticSecurityReports", "true",
+							 NULL, error))
+				return FALSE;
+		}
+	}
```
superm1, line 2605: I think this whole block of code about uploading and the handling of it should go into the static `fu_util_upload_security` function.  Put all the logic around that upload there rather than disjointed looking up a remote here and then checking it it works and then looking it up again there.  Just do it all in one place.

src/fu-util.c
hughsie: Agree, fixed.

src/fu-util.c
hughsie: I am! Fixed.

src/fu-util.c
hughsie: Agreed, fixed.

src/fu-util.c
hughsie: Okay, I ruminated about this for over an hour, and the conclusion I came to was it's not worth adding the complexity. Sending the security report takes less than a second and is a few Kb of network, and is trivial to handle on the LVFS. It's literally the case of pressing enter to stop the action happening.

superm1: I guess the part that I'm thinking about is if a graphical client that runs on every boot checks security and reads the value of the config file it also could be checking in every boot.  It would mean that the graphical client needs to cache the information and decide not to run it (which I guess might be acceptable - command line always sends and graphical checks each time),

hughsie: In that case I'd simply put the successful upload timestamp and sha256 in GSettings. We don't have per-user settings in fwupd at the moment.

superm1: ok.

PR #2120, opened 2020-05-20
Wip/hughsie/get report metadata

hughsie:
_No matter how far you have gone on the wrong road, turn back._

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1342,14 +1342,43 @@ fu_engine_get_boot_time (void)
 	return NULL;
 }
 
-static GHashTable *
-fu_engine_get_report_metadata (FuEngine *self)
+static gboolean
+fu_engine_get_report_metadata_os_release (GHashTable *hash, GError **error)
 {
-	GHashTable *hash;
+	g_autoptr(GHashTable) os_release = NULL;
+	struct {
+		const gchar *key;
+		const gchar *val;
+	} distro_kv[] = {
+		{ "ID",			"DistroId" },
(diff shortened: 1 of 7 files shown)


src/fu-engine.c
```diff
@@ -1401,6 +1402,20 @@ fu_engine_get_report_metadata (FuEngine *self, GError **error)
...
+	tmp = fu_hwids_get_value (self->hwids, FU_HWIDS_KEY_FAMILY);
+	if (tmp != NULL)
+		g_hash_table_insert (hash, g_strdup ("HostFamily"), g_strdup (tmp));
+	tmp = fu_hwids_get_value (self->hwids, FU_HWIDS_KEY_PRODUCT_SKU);
+	if (tmp != NULL)
+		g_hash_table_insert (hash, g_strdup ("HostSku"), g_strdup (tmp));
+	tmp = fu_hwids_get_value (self->hwids, FU_HWIDS_KEY_MANUFACTURER);
+	if (tmp != NULL)
+		g_hash_table_insert (hash, g_strdup ("HostVendor"), g_strdup (tmp));
+
```
superm1, line 1474: Sending these up as part of update reports does open up some interesting things in terms of tying SMBIOS strings to what they are actually used for in UEFI capsule updates.  The concept of a "system" on LVFS and the types of updates it supports and encompasses is possible with this.

hughsie: Yes, it's a bit of _leakage_ for things like FMP, but it's probably going to be super useful for debugging.

PR #2123, opened 2020-05-21
Add an HSI attribute for Intel SMAP

hughsie:
See https://en.wikipedia.org/wiki/Supervisor_Mode_Access_Prevention for details.

### DIFF
--- a/plugins/cpu/fu-plugin-cpu.c
+++ b/plugins/cpu/fu-plugin-cpu.c
@@ -12,6 +12,7 @@
 
 struct FuPluginData {
 	gboolean		 has_cet;
+	gboolean		 has_smap;
 	gboolean		 has_tme;
 };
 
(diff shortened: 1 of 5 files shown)


plugins/cpu/fu-plugin-cpu.c
```diff
@@ -46,6 +47,9 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 			data->has_cet = TRUE;
 		if (fu_cpu_device_has_flag (dev, FU_CPU_DEVICE_FLAG_TME))
 			data->has_tme = TRUE;
+		if (fu_cpu_device_has_flag (dev, FU_CPU_DEVICE_FLAG_SMEP) &&
+		    fu_cpu_device_has_flag (dev, FU_CPU_DEVICE_FLAG_SMAP))
+			data->has_smap = TRUE;
```
superm1, line 52: At least the way the wikipedia page reads, SMAP is the extension to SMEP.  I think you just want to look for SMAP.

hughsie: So maybe just ignore SMEP completely?

superm1: Assuming wikipedia is right, yeah I think so.

hughsie: Gotcha, fixed.

PR #2124, opened 2020-05-21
Wip/hughsie/override strap

hughsie:


hughsie:
> the code looks okay to me

I've spun it out as it's own "thing" so we can reference it in the HSI doc.

I'm not sure how I feel about breaking API in libfwupdplugin. What do you think? I can add _pwrite_full() if you think that's more sane.

superm1:
> I'm not sure how I feel about breaking API in libfwupdplugin. What do you think? I can add _pwrite_full() if you think that's more sane.

I think probably adding a new `_full` makes more sense.  Bumping the ABI just for this seems like it's not a worthwhile tradeoff.

PR #2125, opened 2020-05-22
Check the device requirements when returning from GetDetails

hughsie:
One vendor is shipping a cab archive with two metadata files, both referencing
the same GUID. The 'correct' metainfo description is selected using a GUID
'other device' requirement. This works fine when installing, but breaks when
double clicking on the .cab file as both components are valid, and thus get
returned.

In this case, return the component that matches the requirement 'first' so that
it gets chosen by gnome-software as the default.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -3502,6 +3502,20 @@ fu_engine_get_result_from_component (FuEngine *self, XbNode *component, GError *
 	return g_steal_pointer (&dev);
 }
 
+static gint
+fu_engine_get_details_sort_cb (gconstpointer a, gconstpointer b)
+{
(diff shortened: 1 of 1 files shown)


superm1:
This isn't going to help gnome-software though.  This is only going to affect get-details on the CAB.

hughsie:
> This isn't going to help gnome-software though

This is what gnome-software does when double clicking to show the info page: https://gitlab.gnome.org/GNOME/gnome-software/-/blob/master/plugins/fwupd/gs-plugin-fwupd.c#L978

superm1:
Oh I see.... Yeah then this LGTM

Issue #2119, opened 2020-05-20
del-self-test FAIL with SIGTRAP (GLib-GObject-FATAL-WARNING: cannot register existing type 'FuUefiBgrt')


RiderALT:
**Describe the bug**
When building a new version for ALT, I got the following error:
```
# FuPlugin-DEBUG: performing init() on /usr/src/RPM/BUILD/fwupd-1.4.2/x86_64-alt-linux/plugins/dell/../uefi/libfu_plugin_uefi.so
Bail out! GLib-GObject-FATAL-WARNING: cannot register existing type 'FuUefiBgrt'
--- stderr ---

(/usr/src/RPM/BUILD/fwupd-1.4.2/x86_64-alt-linux/plugins/dell/dell-self-test:1539020): GLib-GObject-WARNING **: 12:33:34.220: cannot register existing type 'FuUefiBgrt'
```
bt from gdb:
```

#0  0x00007ffff7c4a2f5 in ?? () from /lib64/libglib-2.0.so.0
#1  0x00007ffff7c4b5c9 in g_logv () from /lib64/libglib-2.0.so.0
#2  0x00007ffff7c4b78f in g_log () from /lib64/libglib-2.0.so.0
#3  0x00007ffff7d53b98 in ?? () from /usr/lib64/libgobject-2.0.so.0
#4  0x00007ffff7d53e9b in g_type_register_static () from /usr/lib64/libgobject-2.0.so.0
#5  0x00007ffff7d540e5 in g_type_register_static_simple () from /usr/lib64/libgobject-2.0.so.0
#6  0x00007ffff63c110c in fu_uefi_bgrt_get_type_once () at ../plugins/uefi/fu-uefi-bgrt.c:21
#7  0x00007ffff63c1195 in fu_uefi_bgrt_get_type () at ../plugins/uefi/fu-uefi-bgrt.c:21
#8  0x00007ffff63c15d6 in fu_uefi_bgrt_new () at ../plugins/uefi/fu-uefi-bgrt.c:131
#9  0x00007ffff63bf48b in fu_plugin_init (plugin=0x43b320) at ../plugins/uefi/fu-plugin-uefi.c:38
#10 0x00007ffff7f848d4 in fu_plugin_open (self=self@entry=0x43b320, 
[...]

PR #2126, opened 2020-05-22
Skip module unloading only if we are actually running under valgrind

glebfm:
See also: https://github.com/fwupd/fwupd/issues/2119
Fixes: 576c012 ("Do not dlclose modules when running under valgrind")

Type of pull request:
- [x] Code fix

This fix makes issue #2119 more reproducible.

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -2792,10 +2792,11 @@ fu_plugin_finalize (GObject *object)
 	g_free (priv->data);
 	/* Must happen as the last step to avoid prematurely
 	 * freeing memory held by the plugin */
-#ifndef RUNNING_ON_VALGRIND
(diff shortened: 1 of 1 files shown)


hughsie:
I think this needs somethign like wip/hughsie/module-unloading to fix CI.

glebfm:
> I think this needs somethign like wip/hughsie/module-unloading to fix CI.

Thank you for quick response!
It looks as two separate changes to me, but it actually fixes both issues.

hughsie:
> but it actually fixes both issues

Nah, CI is still failing even with the dell plugin fix.

glebfm:
> Nah, CI is still failing even with the dell plugin fix.

Ok, it fixes it for isolated tpm-free environment I use to build software. :)

superm1:
superceded by #2140

PR #2129, opened 2020-05-22
Add a general plugin to detect memory encryption

hughsie:


hughsie:
This isn't what I imagined, I'd much rather read a sysfs attribute, but here we are.

hughsie:
This isn't what I imagined, I'd much rather read a sysfs attribute, but here we are.

plugins/journald/fu-plugin-journald.c
```diff
@@ -0,0 +1,91 @@
...
+	rc = sd_journal_open (&j, SD_JOURNAL_SYSTEM);
+	if (rc < 0) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "failed to open journal: %s [%i]",
+			     strerror (rc), rc);
+		return FALSE;
+	}
+	while (sd_journal_next (j) > 0) {
```
superm1, line 47: I'm a little worried this turns into a performance bottleneck.  Is there a way to filter it to only be running from the current boot?  Or only from time the kernel finished initializing?

hughsie: It takes 68ms on my Lenovo P50, which is a pretty powerful machine. I can imagine something like a raspberry pi might take a lot longer.

hughsie: Of course, we could just read the journal when getting the hsi report... will fix.

superm1: Well I think when paired with `sd_id128_get_boot`, `sd_journal_seek_monotonic_usec` it's possible to make it a super light operation for most people even with a large journal.  It would at least allow only looking at the current boot's journal.

hughsie: Hmm, I tried that and didn't get any speedup at all.

superm1:
So I checked this on a box that supports it, and I now think this is the wrong way to go.  I think we should go sysfs attribute showing current status or nothing.

```
# dmesg | grep tme
[    0.116284] x86/tme: enabled by BIOS
[    0.116285] x86/mktme: No known encryption algorithm is supported: 0x0
[    0.116288] x86/mktme: disabled by BIOS
```

fwupdtool security picks up as enabled because of the first one.

hughsie:
> and I now think this is the wrong way to go

I agree. We're currently arguing with LKML about the best way to export this: https://lkml.org/lkml/2020/6/18/1034

superm1:
> > and I now think this is the wrong way to go
> 
> I agree. We're currently arguing with LKML about the best way to export this: https://lkml.org/lkml/2020/6/18/1034

Looks like quite the rabbit hole :)

Issue #2044, opened 2020-04-30
Can I downgrade Logitech K375s keyboard firmware? And is there a way to backup the current firmware on the keyboard before I do so?


fondfire:
I have a Logitech K375s keyboard that I'm using on Ubuntu 18.04 systems. The following are screenshots from Solaar on my system:

![Screenshot_2020-04-30_01-53-12](https://user-images.githubusercontent.com/44927770/80681048-82534480-8a85-11ea-8725-07392e25dafb.png)
![Screenshot_2020-04-30_01-53-49](https://user-images.githubusercontent.com/44927770/80681059-867f6200-8a85-11ea-9736-291191fda35e.png)

The current MPK 03.04.B0014 firmware for my keyboard seems to have a bug.

The keyboard goes into a suspend or sleep mode after its untouched for a short while, which I am sure preserves the battery life. When it is touched again, it "wakes up".

On my Logitech K375s with the MPK 03.04.B0014 firmware, there's a string of gibberish that flies out of the keyboard every time it "wakes up" from this light, and otherwise perfectly functional, sleep. I've tried swapping the keyboard mode -- to iOS mode (by pressing Fn-I for about seven seconds), to MacOS mode (by pressing Fn-O for about seven seconds), and to Windows/Android mode (by pressing Fn-P for about seven seconds) -- and none of those modes is without this issue. In fact, every time a switch modes, a new set of garbage is randomly spewed by the keyboard (albeit, each mode seems to use a slightly different set of garbage characters). I suppose there is a small chance my keyboard was physically damaged in some way, but it's working perfectly in all other ways. And the keyboard has been doing this since I purchased it, IIRC.

I like this keyboard in every other way. I believe it's too late to return it. I really don't WANT to replace it.

Can I try downgrading to the MPK 03.02.B0009 firmware available on LVFS already? Is there a way I can back-up the current firmware in case that goes horribly wrong? I've tried with Ubuntu 18.04 systems, running this version of fwupd at the time of this writing:

```
user@xubuntu:~$ sudo fwupdmgr --version
client version:	1.2.10
compile-time dependency versions
	gusb:	0.2.11
	efivar:	34
daemon version:	1.2.10
user@xubuntu:~$ 
```

When I run `fwupdmgr get-devices`, the Unifying Receiver shows up, but not the keyboard device.

```
user@xubuntu:~$ sudo fwupdmgr --allow-older get-devices
Unifying Receiver
[...]

PR #2130, opened 2020-05-22
Initial Chrome OS EC plugin 

bleungatchromium:
Type of pull request:
- [ X] New plugin (Please include [new plugin checklist](https://github.com/hughsie/fwupd/wiki/New-plugin-checklist))
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

Initial checkin of the Chrome OS EC USB plugin. For now, this will check and provide the version number of a Servo Micro debug dongle. (https://chromium.googlesource.com/chromiumos/third_party/hdctools/+/refs/heads/master/docs/servo_micro.md).

Further changes will add support for the firmware file format and complete the mechanism of flashing the payload, as well as adding support for other USB targets.

### DIFF
--- a/contrib/ci/dependencies.xml
+++ b/contrib/ci/dependencies.xml
@@ -766,7 +766,7 @@
     </distro>
     <distro id="debian">
       <control>
-        <version>(>= 0.2.9)</version>
+        <version>(>= 0.3.3)</version>
       </control>
       <package variant="x86_64" />
       <package variant="s390x">libgusb-dev:s390x</package>
@@ -774,7 +774,7 @@
     </distro>
     <distro id="ubuntu">
       <control>
-        <version>(>= 0.2.9)</version>
+        <version>(>= 0.3.3)</version>
       </control>
       <package variant="x86_64" />
     </distro>
--- a/plugins/cros-ec/fu-cros-ec-common.c
+++ b/plugins/cros-ec/fu-cros-ec-common.c
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2020 Benson Leung <bleung@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <string.h>
+
+#include "fu-cros-ec-common.h"
+
+gboolean
+fu_cros_ec_parse_version (const gchar *version_raw,
+			  struct cros_ec_version *version, GError **error)
+{
+	g_auto(GStrv) v_split = NULL;
+	g_auto(GStrv) marker_split = NULL;
+	g_auto(GStrv) triplet_split = NULL;
+
+	if (NULL == version_raw || 0 == strlen (version_raw)) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "no version string to parse");
+		return FALSE;
+	}
+
+	/* sample version string: cheese_v1.1.1755-4da9520 */
+	v_split = g_strsplit (version_raw, "_v", 2);
+	if (g_strv_length (v_split) < 2) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "version marker not found");
+		return FALSE;
+	}
+	marker_split = g_strsplit_set (v_split[1], "-+", 2);
+	if (g_strv_length (marker_split) < 2) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "hash marker not found: %s", v_split[1]);
+		return FALSE;
+	}
+	triplet_split = g_strsplit_set (marker_split[0], ".", 3);
+	if (g_strv_length (triplet_split) < 3) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "improper version triplet: %s", marker_split[0]);
+		return FALSE;
+	}
+	g_strlcpy (version->triplet, marker_split[0], 32);
+	if (g_strlcpy (version->boardname, v_split[0], 32) == 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "empty board name");
+		return FALSE;
+	}
+	if (g_strlcpy (version->sha1, marker_split[1], 32) == 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "empty SHA");
+		return FALSE;
+	}
+	version->dirty = (g_strrstr(v_split[1], "+") != NULL);
+
+	return TRUE;
+}
--- a/plugins/cros-ec/fu-cros-ec-common.h
+++ b/plugins/cros-ec/fu-cros-ec-common.h
@@ -1,0 +1,125 @@
+/*
+ * Copyright (C) 2020 Benson Leung <bleung@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#define UPDATE_PROTOCOL_VERSION 6
+
+/*
+ * This is the format of the update PDU header.
+ *
+ * block digest: the first four bytes of the sha1 digest of the rest of the
+ *               structure (can be 0 on boards where digest is ignored).
+ * block_base:   offset of this PDU into the flash SPI.
+ */
+typedef struct __attribute__((packed)) {
+	guint32  block_digest;
+	guint32  block_base;
+	/* The actual payload goes here. */
+} update_command;
+
+/*
+ * This is the frame format the host uses when sending update PDUs over USB.
+ *
+ * The PDUs are up to 1K bytes in size, they are fragmented into USB chunks of
+ * 64 bytes each and reassembled on the receive side before being passed to
+ * the flash update function.
+ *
+ * The flash update function receives the unframed PDU body (starting at the
+ * cmd field below), and puts its reply into the same buffer the PDU was in.
+ */
+struct update_frame_header {
+	guint32 block_size; /* Total frame size, including this field. */
+	update_command cmd;
+};
+
+/*
+ * A convenience structure which allows to group together various revision
+ * fields of the header created by the signer (cr50-specific).
+ *
+ * These fields are compared when deciding if versions of two images are the
+ * same or when deciding which one of the available images to run.
+ */
+struct signed_header_version {
+	guint32 minor;
+	guint32 major;
+	guint32 epoch;
+};
+
+/*
+ * Response to the connection establishment request.
+ *
+ * When responding to the very first packet of the update sequence, the
+ * original USB update implementation was responding with a four byte value,
+ * just as to any other block of the transfer sequence.
+ *
+ * It became clear that there is a need to be able to enhance the update
+ * protocol, while staying backwards compatible.
+ *
+ * All newer protocol versions (starting with version 2) respond to the very
+ * first packet with an 8 byte or larger response, where the first 4 bytes are
+ * a version specific data, and the second 4 bytes - the protocol version
+ * number.
+ *
+ * This way the host receiving of a four byte value in response to the first
+ * packet is considered an indication of the target running the 'legacy'
+ * protocol, version 1. Receiving of an 8 byte or longer response would
+ * communicates the protocol version in the second 4 bytes.
+ */
+struct first_response_pdu {
+	guint32 return_value;
+
+	/* The below fields are present in versions 2 and up. */
+
+	/* Type of header following (one of first_response_pdu_header_type) */
+	guint16 header_type;
+
+	/* Must be UPDATE_PROTOCOL_VERSION */
+	guint16 protocol_version;
+
+	/* In version 6 and up, a board-specific header follows. */
+	union {
+		/* cr50 (header_type = UPDATE_HEADER_TYPE_CR50) */
+		struct {
+			/* The below fields are present in versions 3 and up. */
+			guint32  backup_ro_offset;
+			guint32  backup_rw_offset;
+
+			/* The below fields are present in versions 4 and up. */
+			/*
+			 * Versions of the currently active RO and RW sections.
+			 */
+			struct signed_header_version shv[2];
+
+			/* The below fields are present in versions 5 and up */
+			/* keyids of the currently active RO and RW sections. */
+			guint32 keyid[2];
+		} cr50;
+		/* Common code (header_type = UPDATE_HEADER_TYPE_COMMON) */
+		struct {
+			/* Maximum PDU size */
+			guint32 maximum_pdu_size;
+
+			/* Flash protection status */
+			guint32 flash_protection;
+
+			/* Offset of the other region */
+			guint32 offset;
+
+			/* Version string of the other region */
+			gchar version[32];
+
+			/* Minimum rollback version that RO will accept */
+			gint32 min_rollback;
+
+			/* RO public key version */
+			guint32 key_version;
+		} common;
+	};
+};
+
+enum first_response_pdu_header_type {
+	UPDATE_HEADER_TYPE_CR50 = 0, /* Must be 0 for backwards compatibility */
+	UPDATE_HEADER_TYPE_COMMON = 1,
+};
--- a/plugins/cros-ec/fu-cros-ec-common.h
+++ b/plugins/cros-ec/fu-cros-ec-common.h
@@ -4,6 +4,12 @@
  * SPDX-License-Identifier: LGPL-2.1+
  */
 
+#pragma once
+
+#include <glib.h>
+
+#include "fu-plugin.h"
+
 #define UPDATE_PROTOCOL_VERSION 6
 
 /*
@@ -123,3 +129,14 @@ enum first_response_pdu_header_type {
 	UPDATE_HEADER_TYPE_CR50 = 0, /* Must be 0 for backwards compatibility */
 	UPDATE_HEADER_TYPE_COMMON = 1,
 };
+
+struct cros_ec_version {
+	gchar	 boardname[32];
+	gchar	 triplet[32];
+	gchar	 sha1[32];
+	gboolean dirty;
+};
+
+gboolean	fu_cros_ec_parse_version	(const gchar 		*version_raw,
+						 struct cros_ec_version *version,
+						 GError **error);
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -322,7 +322,26 @@ static void
 fu_cros_ec_usb_device_to_string (FuDevice *device, guint idt, GString *str)
 {
 	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	g_autofree gchar *min_rollback = NULL;
+
 	fu_common_string_append_kv (str, idt, "GitHash", self->version.sha1);
+	fu_common_string_append_kb (str, idt, "Dirty",
+				    self->version.dirty);
+	fu_common_string_append_ku (str, idt, "ProtocolVersion",
+				    self->protocol_version);
+	fu_common_string_append_ku (str, idt, "HeaderType",
+				    self->header_type);
+	fu_common_string_append_ku (str, idt, "MaxPDUSize",
+				    self->targ.common.maximum_pdu_size);
+	fu_common_string_append_kx (str, idt, "FlashProtectionStatus",
+				    self->targ.common.flash_protection);
+	fu_common_string_append_kv (str, idt, "RawVersion",
+				    self->targ.common.version);
+	fu_common_string_append_ku (str, idt, "KeyVersion",
+				    self->targ.common.key_version);
+	min_rollback = g_strdup_printf ("%" G_GINT32_FORMAT,
+					self->targ.common.min_rollback);
+	fu_common_string_append_kv (str, idt, "MinRollback", min_rollback);
 }
 
 static void
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -9,9 +9,15 @@
 #include <string.h>
 
 #include "fu-cros-ec-usb-device.h"
+#include "fu-cros-ec-common.h"
 
 #define USB_SUBCLASS_GOOGLE_UPDATE	0x53
 #define USB_PROTOCOL_GOOGLE_UPDATE	0xff
+
+#define SETUP_RETRY_CNT			5
+#define FLUSH_TIMEOUT_MS		10
+#define BULK_SEND_TIMEOUT_MS		2000
+#define BULK_RECV_TIMEOUT_MS		5000
 
 struct _FuCrosEcUsbDevice {
 	FuUsbDevice		parent_instance;
@@ -19,9 +25,17 @@ struct _FuCrosEcUsbDevice {
 	guint8			ep_num;		/* bEndpointAddress */
 	guint16			chunk_len; 	/* wMaxPacketSize */
 
+	struct			first_response_pdu targ;
+	guint16			protocol_version;
+	guint16			header_type;
 };
 
 G_DEFINE_TYPE (FuCrosEcUsbDevice, fu_cros_ec_usb_device, FU_TYPE_USB_DEVICE)
+
+typedef union _START_RESP {
+	struct first_response_pdu rpdu;
+	guint32 legacy_resp;
+} START_RESP;
 
 static gboolean
 fu_cros_ec_usb_device_find_interface (FuUsbDevice *device,
@@ -105,8 +119,168 @@ fu_cros_ec_usb_device_probe (FuUsbDevice *device, GError **error)
 }
 
 static gboolean
+fu_cros_ec_usb_device_do_xfer (FuCrosEcUsbDevice * self, guint8 *outbuf,
+			       gsize outlen, guint8 *inbuf, gsize inlen,
+			       gboolean allow_less, gsize *rxed_count,
+			       GError **error)
+{
+	GUsbDevice *usb_device = fu_usb_device_get_dev (FU_USB_DEVICE (self));
+	gsize actual = 0;
+
+	/* send data out */
+	if (outbuf != NULL && outlen > 0) {
+		if (!g_usb_device_bulk_transfer (usb_device, self->ep_num,
+						 outbuf, outlen,
+						 &actual, BULK_SEND_TIMEOUT_MS,
+						 NULL, error)) {
+			return FALSE;
+		}
+		if (actual != outlen) {
+			g_set_error (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_PARTIAL_INPUT,
+				     "only sent %" G_GSIZE_FORMAT "/%"
+				     G_GSIZE_FORMAT " bytes",
+				     actual, outlen);
+			return FALSE;
+		}
+	}
+
+	/* read reply back */
+	if (inbuf != NULL && inlen > 0) {
+		actual = 0;
+		if (!g_usb_device_bulk_transfer (usb_device,
+						 self->ep_num | 0x80,
+						 inbuf, inlen,
+						 &actual, BULK_RECV_TIMEOUT_MS,
+						 NULL, error)) {
+			return FALSE;
+		}
+		if (actual != inlen && !allow_less) {
+			g_set_error (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_PARTIAL_INPUT,
+				     "only received %" G_GSIZE_FORMAT "/%"
+				     G_GSIZE_FORMAT " bytes",
+				     actual, outlen);
+			return FALSE;
+		}
+	}
+
+	if (rxed_count != NULL)
+		*rxed_count = actual;
+
+	return TRUE;
+}
+
+static gboolean
+fu_cros_ec_usb_device_flush (FuDevice *device, gpointer user_data,
+			     GError **error)
+{
+	GUsbDevice *usb_device = fu_usb_device_get_dev (FU_USB_DEVICE (device));
+	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	gsize actual = 0;
+	g_autofree guint8 *inbuf = g_malloc0 (self->chunk_len);
+
+	if (g_usb_device_bulk_transfer (usb_device, self->ep_num | 0x80, inbuf,
+					self->chunk_len, &actual,
+					FLUSH_TIMEOUT_MS, NULL, NULL)) {
+		g_debug ("flushing %" G_GSIZE_FORMAT " bytes", actual);
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "flushing %" G_GSIZE_FORMAT " bytes", actual);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_cros_ec_usb_device_start_request (FuDevice *device, gpointer user_data,
+				     GError **error)
+{
+	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	guint8 *start_resp = (guint8 *) user_data;
+	struct update_frame_header ufh;
+	gsize rxed_size = 0;
+
+	memset(&ufh, 0, sizeof (ufh));
+	ufh.block_size = GUINT32_TO_BE (sizeof(ufh));
+	if (!fu_cros_ec_usb_device_do_xfer (self, (guint8 *)&ufh, sizeof(ufh),
+					    start_resp,
+					    sizeof(START_RESP), TRUE,
+					    &rxed_size, error))
+		return FALSE;
+
+	/* we got something, so check for errors in response */
+	if (rxed_size < 8) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_PARTIAL_INPUT,
+			     "unexpected response size %" G_GSIZE_FORMAT,
+			     rxed_size);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
 fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
 {
+	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	guint32 error_code;
+	START_RESP start_resp;
+
+	/* flush all data from endpoint to recover in case of error */
+	if (!fu_device_retry (device, fu_cros_ec_usb_device_flush,
+			      SETUP_RETRY_CNT, NULL, error)) {
+		g_prefix_error (error, "failed to flush device to idle state: ");
+		return FALSE;
+	}
+
+	/* send start request */
+	if (!fu_device_retry (device, fu_cros_ec_usb_device_start_request,
+			      SETUP_RETRY_CNT, &start_resp, error)) {
+		g_prefix_error (error, "failed to send start request: ");
+		return FALSE;
+	}
+
+	self->protocol_version = GUINT16_FROM_BE (start_resp.rpdu.protocol_version);
+
+	if (self->protocol_version < 5 || self->protocol_version > 6) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_NOT_SUPPORTED,
+			     "unsupported protocol version %d",
+			     self->protocol_version);
+		return FALSE;
+	}
+	self->header_type = GUINT16_FROM_BE (start_resp.rpdu.header_type);
+
+	error_code = GUINT32_FROM_BE (start_resp.rpdu.return_value);
+	if (error_code != 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "target reporting error %u", error_code);
+		return FALSE;
+	}
+
+	memcpy (self->targ.common.version, start_resp.rpdu.common.version,
+		sizeof(start_resp.rpdu.common.version));
+	self->targ.common.maximum_pdu_size =
+		GUINT32_FROM_BE (start_resp.rpdu.common.maximum_pdu_size);
+	self->targ.common.flash_protection =
+		GUINT32_FROM_BE (start_resp.rpdu.common.flash_protection);
+	self->targ.common.min_rollback = GINT32_FROM_BE (start_resp.rpdu.common.min_rollback);
+	self->targ.common.key_version = GUINT32_FROM_BE (start_resp.rpdu.common.key_version);
+
+	fu_device_set_version (FU_DEVICE (device), self->targ.common.version);
+
 	/* success */
 	return TRUE;
 }
@@ -131,7 +305,7 @@ fu_cros_ec_usb_device_close (FuUsbDevice *device, GError **error)
 static void
 fu_cros_ec_usb_device_init (FuCrosEcUsbDevice *device)
 {
-	fu_device_set_version_format (FU_DEVICE (device), FWUPD_VERSION_FORMAT_TRIPLET);
+	fu_device_set_version_format (FU_DEVICE (device), FWUPD_VERSION_FORMAT_PLAIN);
 }
 
 static void
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -289,6 +289,7 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
 	}
 
 	fu_device_set_version (FU_DEVICE (device), self->version.triplet);
+	fu_device_add_instance_id (FU_DEVICE (device), self->version.boardname);
 
 	/* success */
 	return TRUE;
@@ -318,11 +319,19 @@ fu_cros_ec_usb_device_init (FuCrosEcUsbDevice *device)
 }
 
 static void
+fu_cros_ec_usb_device_to_string (FuDevice *device, guint idt, GString *str)
+{
+	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	fu_common_string_append_kv (str, idt, "GitHash", self->version.sha1);
+}
+
+static void
 fu_cros_ec_usb_device_class_init (FuCrosEcUsbDeviceClass *klass)
 {
 	FuDeviceClass *klass_device = FU_DEVICE_CLASS (klass);
 	FuUsbDeviceClass *klass_usb_device = FU_USB_DEVICE_CLASS (klass);
 	klass_device->setup = fu_cros_ec_usb_device_setup;
+	klass_device->to_string = fu_cros_ec_usb_device_to_string;
 	klass_usb_device->open = fu_cros_ec_usb_device_open;
 	klass_usb_device->probe = fu_cros_ec_usb_device_probe;
 	klass_usb_device->close = fu_cros_ec_usb_device_close;
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2020 Benson Leung <bleung@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <string.h>
+
+#include "fu-cros-ec-usb-device.h"
+
+G_DEFINE_TYPE (FuCrosEcUsbDevice, fu_cros_ec_usb_device, FU_TYPE_USB_DEVICE)
+
+static gboolean
+fu_cros_ec_usb_device_open (FuUsbDevice *device, GError **error)
+{
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
+{
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_cros_ec_usb_device_close (FuUsbDevice *device, GError **error)
+{
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_cros_ec_usb_device_init (FuCrosEcUsbDevice *device)
+{
+	fu_device_set_version_format (FU_DEVICE (device), FWUPD_VERSION_FORMAT_TRIPLET);
+}
+
+static void
+fu_cros_ec_usb_device_class_init (FuCrosEcUsbDeviceClass *klass)
+{
+	FuDeviceClass *klass_device = FU_DEVICE_CLASS (klass);
+	FuUsbDeviceClass *klass_usb_device = FU_USB_DEVICE_CLASS (klass);
+	klass_device->setup = fu_cros_ec_usb_device_setup;
+	klass_usb_device->open = fu_cros_ec_usb_device_open;
+	klass_usb_device->close = fu_cros_ec_usb_device_close;
+}
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -20,14 +20,15 @@
 #define BULK_RECV_TIMEOUT_MS		5000
 
 struct _FuCrosEcUsbDevice {
-	FuUsbDevice		parent_instance;
-	guint8			iface_idx; 	/* bInterfaceNumber */
-	guint8			ep_num;		/* bEndpointAddress */
-	guint16			chunk_len; 	/* wMaxPacketSize */
+	FuUsbDevice			parent_instance;
+	guint8				iface_idx; 	/* bInterfaceNumber */
+	guint8				ep_num;		/* bEndpointAddress */
+	guint16				chunk_len; 	/* wMaxPacketSize */
 
-	struct			first_response_pdu targ;
-	guint16			protocol_version;
-	guint16			header_type;
+	struct first_response_pdu	targ;
+	guint16				protocol_version;
+	guint16				header_type;
+	struct cros_ec_version		version;
 };
 
 G_DEFINE_TYPE (FuCrosEcUsbDevice, fu_cros_ec_usb_device, FU_TYPE_USB_DEVICE)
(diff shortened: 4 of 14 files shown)


contrib/ci/dependencies.xml
```diff
@@ -784,15 +784,15 @@
     </distro>
     <distro id="debian">
       <control>
-        <version>(>= 0.2.9)</version>
+        <version>(>= 0.3.3)</version>
```
superm1, line 787: I think you also need to bump up the top level meson.build for this

plugins/cros_ec-usb/fu-plugin-cros_ec-usb.c
```diff
@@ -0,0 +1,20 @@
...
+#include "fu-hash.h"
+
+#include "fu-cros_ec-usb-device.h"
+
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+	fu_plugin_set_device_gtype (plugin, FU_TYPE_CROS_EC_USB_DEVICE);
+	g_print ("fu-plugin-cros_ec_usb fu_plugin_init!\n");
```
superm1, line 19: I suspect this was a debugging line originally, but it should be unnecessary now

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,309 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_cros_ec_usb_device_init (FuCrOSECUsbDevice *device)
+{
+	fu_device_set_version_format (FU_DEVICE (device), FWUPD_VERSION_FORMAT_TRIPLET);
+	fu_device_set_version (FU_DEVICE (device), "0.0.0");
```
superm1, line 288: Is it necessary to set the default version to 0.0.0?  I would think no; the daemon handles not having a version defined just fine.  Also since you always set it in `_setup` it should work out OK.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,309 @@
...
+	self->targ.common.flash_protection =
+		be32toh (start_resp.rpdu.common.flash_protection);
+	self->targ.common.min_rollback = be32toh (start_resp.rpdu.common.min_rollback);
+	self->targ.common.key_version = be32toh (start_resp.rpdu.common.key_version);
+
+	g_debug ("maximum PDU size: %u\n", self->targ.common.maximum_pdu_size);
+	g_debug ("Flash protection status: %04x\n", self->targ.common.flash_protection);
+	g_debug ("version: %32s\n", self->targ.common.version);
+	g_debug ("key_version: %u\n", self->targ.common.key_version);
+	g_debug ("min_rollback: %d\n", self->targ.common.min_rollback);
```
superm1, line 244: I think in general all of this debug output code is better put into the `to_string` .  That's what other plugins typically do, and it's more easily accessible then.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,309 @@
...
+		return FALSE;
+	}
+
+	g_debug ("boardname = %s\n", self->version.boardname);
+	g_debug ("triplet = %s\n", self->version.triplet);
+	g_debug ("sha1 = %s\n", self->version.sha1);
+	g_debug ("%s\n", self->version.dirty?"dirty":"clean");
+
+	fu_device_set_version (FU_DEVICE(device), self->version.triplet);
+	fu_device_add_guid (FU_DEVICE(device), self->version.boardname);
```
superm1, line 261: I think that you actually want to use `fu_device_add_instance_id` usually here instead of `fu_device_add_guid`.  IIRC this allows quirks to be applied (if they end up being necessary).

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
```diff
@@ -0,0 +1,39 @@
...
+
+#include "config.h"
+
+#include <string.h>
+
+#include "fu-cros_ec-usb-common.h"
+
+gboolean
+fu_cros_ec_usb_parse_version (const gchar *version_raw,
+			      struct cros_ec_version *version)
```
superm1, line 15: I think when you add self tests there should probably be one that does parsing of this to match all your boundary conditions properly.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+#include "config.h"
+
+#include <string.h>
+
+#include "fu-cros_ec-usb-common.h"
+
+gboolean
+fu_cros_ec_usb_parse_version (const gchar *version_raw,
```
superm1, line 14: I think it would be reasonable to pass a `GError` as a 3rd argument for this function.  Then in all the `FALSE` conditions you can add more detail about what in the parsing actually failed.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+				continue;
+			ep = g_ptr_array_index (endpoints, 0);
+			self->iface_idx = g_usb_interface_get_number (intf);
+			self->ep_num = g_usb_endpoint_get_address (ep) & 0x7f;
+			self->chunk_len = g_usb_endpoint_get_maximum_packet_size (ep);
+
+			return TRUE;
+		}
+	}
+	return FALSE;
```
superm1, line 65: You're missing an error being set in this `FALSE` condition for this function.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+	intfs = g_usb_device_get_interfaces (usb_device, error);
+	if (intfs == NULL)
+		return FALSE;
+	for (guint i = 0; i < intfs->len; i++) {
+		GUsbInterface *intf = g_ptr_array_index (intfs, i);
+		if (g_usb_interface_get_class (intf) == 255 &&
+		    g_usb_interface_get_subclass (intf) == USB_SUBCLASS_GOOGLE_UPDATE &&
+		    g_usb_interface_get_protocol (intf) == USB_PROTOCOL_GOOGLE_UPDATE) {
+			g_autoptr(GPtrArray) endpoints = NULL;
+			GUsbEndpoint *ep;
```
superm1, line 52: (nit) project style has `g_autoptr` and `g_autofree` declared as the last variables in a block, so re-order these two.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+
+static gboolean
+fu_cros_ec_usb_device_open (FuUsbDevice *device, GError **error)
+{
+	GUsbDevice *usb_device = fu_usb_device_get_dev (device);
+	FuCrOSECUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+
+	/* very much like usb_updater2's usb_findit() */
+
+	if (!fu_cros_ec_usb_device_find_interface (device, error)) {
```
superm1, line 76: I feel like this particular code to find the interface (and the failure check for invalid stuff) should probably live in a `_probe` function.  It's the type of stuff that is normally done by the parent class in it's probe function. 
 (`fu_usb_device_probe`)

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+}
+
+static gboolean
+do_xfer (FuCrOSECUsbDevice * cros_ec_device, guint8 *outbuf, gsize outlen,
+	 guint8 *inbuf, gsize inlen, gboolean allow_less,
+	 gsize *rxed_count, GError **error)
+{
+	GUsbDevice *usb_device = fu_usb_device_get_dev (FU_USB_DEVICE (cros_ec_device));
+	FuCrOSECUsbDevice *self = cros_ec_device;
+	gsize actual;
```
superm1, line 110: I think you probably should initialize this, as there is a set of circumstances that garbage gets set back if either outbuf, outlen, or inbuf or inlen evaluate to false but rxed_count is > 0.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+
+	/* Send data out */
+	if (outbuf && outlen) {
```
superm1, line 113: I think it should be clearer to make it explicit what you're checking for as that's what the rest of the project usually does.
IE:
`if (outbug != NULL && outlen > 0)`

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+				     G_IO_ERROR_PARTIAL_INPUT,
+				     "only sent %" G_GSIZE_FORMAT "/%"
+				     G_GSIZE_FORMAT " bytes\n",
+				     actual, outlen);
+			return FALSE;
+		}
+	}
+
+	/* Read reply back */
+	if (inbuf && inlen) {
```
superm1, line 132: same comment at outbuf/outlen

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,308 @@
...
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "Target reporting error %u: \n", error_code);
+		return FALSE;
+	}
+
+	memcpy(self->targ.common.version, start_resp.rpdu.common.version,
+		sizeof(start_resp.rpdu.common.version));
+	self->targ.common.maximum_pdu_size =
+		be32toh (start_resp.rpdu.common.maximum_pdu_size);
```
superm1, line 234: `GUINT32_FROM_BE` perhaps?

contrib/ci/dependencies.xml
bleungatchromium: Hi @superm1,

I made this change, but this made the lgtm.com builder upset. Do you have some advice how to make that builder happy?

https://lgtm.com/projects/g/fwupd/fwupd/logs/rev/pr-567e623e129364154f1ea9ca31a3ac99d275c7ee/lang:cpp/stage:Build%20merge_a020dfb124810e01efe81b8a3e16c7aa4e684b00

superm1: Probably modifications to https://github.com/fwupd/fwupd/blob/master/.lgtm.yml, but I don't know for sure.  Don't worry about this part for now, we can figure that out later when we're closer to merging.  I think it's more correct to have the top level `meson.build`.

bleungatchromium:
Thanks for the review, @superm1.

Could you give advice on what this error means on the fedora ci builder?

Checking for unpackaged file(s): /usr/lib/rpm/check-files /root/rpmbuild/BUILDROOT/fwupd-1.5.0-0.1alpha.fc31.x86_64
error: Installed (but unpackaged) file(s) found:
   /usr/lib/debug/usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so-1.5.0-0.1alpha.fc31.x86_64.debug
   /usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so
RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/lib/debug/usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so-1.5.0-0.1alpha.fc31.x86_64.debug
   /usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so

https://travis-ci.org/github/fwupd/fwupd/jobs/690231181

Thanks!

superm1:
The other CI failure you're seeing is because it needs to be added to Fedora 

> Thanks for the review, @superm1.
> 
> Could you give advice on what this error means on the fedora ci builder?
> 
> Checking for unpackaged file(s): /usr/lib/rpm/check-files /root/rpmbuild/BUILDROOT/fwupd-1.5.0-0.1alpha.fc31.x86_64
> error: Installed (but unpackaged) file(s) found:
> /usr/lib/debug/usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so-1.5.0-0.1alpha.fc31.x86_64.debug
> /usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so
> RPM build errors:
> Installed (but unpackaged) file(s) found:
> /usr/lib/debug/usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so-1.5.0-0.1alpha.fc31.x86_64.debug
> /usr/lib64/fwupd-plugins-3/libfu_plugin_cros_ec_usb.so
> 
> https://travis-ci.org/github/fwupd/fwupd/jobs/690231181
> 
> Thanks!

Sure!
Basically needs to be added into the Fedora spec file.  There is a list of all the plugins packaged in that `spec.in`, just add it in the correct alphabetical part of that file.

bleungatchromium:
> Sure!
> Basically needs to be added into the Fedora spec file. There is a list of all the plugins packaged in that `spec.in`, just add it in the correct alphabetical part of that file.

Great, thanks!

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Alright. I'll work on adding the _probe for this. Thanks!

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/meson.build
```diff
@@ -0,0 +1,29 @@
...
+
+install_data(['cros_ec-usb.quirk'],
+  install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
+)
+
+shared_module('fu_plugin_cros_ec_usb',
+  fu_hash,
+  sources : [
+    'fu-plugin-cros_ec-usb.c',
+    'fu-cros_ec-usb-common.c',
```
hughsie, line 11: Can I be a right pain? I think it should be `fu-cros-ec*' (i.e. with dashes, not underscores) to match the other plugins.

plugins/meson.build
```diff
@@ -1,6 +1,7 @@
 subdir('acpi-dmar')
 subdir('acpi-facp')
 subdir('ccgx')
+subdir('cros_ec-usb')
```
hughsie, line 4: same here. Git rewrite to the rescue?

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+	if (rxed_count)
+		*rxed_count = actual;
+
+	return TRUE;
+}
+
+static gboolean
+fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
+{
+	FuCrOSECUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
```
hughsie, line 171: Maybe also `FuCrOSECUsbDevice` -> `FuCrosEcUsbDevice` to match convention.

plugins/cros_ec-usb/meson.build
bleungatchromium: Hi @hughsie. Thanks for the review.

Do you recommend that I rename the whole directory to be "cros-ec" instead of "cros_ec-usb"?

How about the following renames:
fu-plugin-cros_ec-usb.c -> fu-plugin-cros-ec.c
fu-cros_ec-usb-device.c/h -> fu-cros-ec-usb-device.c/h
fu-cros_ec-usb-common.c/h -> fu-cros-ec-common.h
cros_ec-usb.quirk -> cros-ec-usb.quirk


This will leave open the possibility of non-USB based updaters being written (which we have, but are not in scope right at the moment).

plugins/meson.build
bleungatchromium: So just to be clear, cros_ec-usb -> cros-ec?

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Ok, will do. Sorry for the weird CamelCase.

plugins/meson.build
hughsie: yup

plugins/cros_ec-usb/meson.build
hughsie: I think  `fu-cros_ec-usb-device.c/h` -> `fu-cros-ec-device.c/h` is better

hughsie: And `cros_ec-usb.quirk` -> `cros-ec.quirk`

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
hughsie: No worries :) Maybe FuCrosEcDevice if we drop the USB everywhere too.

contrib/ci/dependencies.xml
hughsie: I think https://github.com/hughsie/libjcat/blob/master/.lgtm.yml is what you want to copy from.

plugins/cros_ec-usb/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+Information about the update protocol is available at [2].
+
+
+GUID Generation
+---------------
+
+These devices use the standard USB DeviceInstanceId values, e.g.
+
+ * `USB\VID_18d1&PID_501a`
```
hughsie, line 20: upper case last A

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
```diff
@@ -0,0 +1,75 @@
...
+{
+	g_auto(GStrv) v_split = NULL;
+	g_auto(GStrv) marker_split = NULL;
+	g_auto(GStrv) triplet_split = NULL;
+
+	if (NULL == version_raw || 0 == strlen (version_raw)) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "No version string to parse.");
```
hughsie, line 26: no fullstops at the end of the error message please.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
```diff
@@ -0,0 +1,75 @@
...
+		return FALSE;
+	}
+
+	/* sample version string: cheese_v1.1.1755-4da9520 */
+	v_split = g_strsplit (version_raw, "_v", 2);
+	if (2 > g_strv_length (v_split)) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "Version marker not found.");
```
hughsie, line 36: same

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
```diff
@@ -0,0 +1,75 @@
...
+	triplet_split = g_strsplit_set (marker_split[0], ".", 3);
+	if (3 > g_strv_length (triplet_split)) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Improper version triplet: %s", marker_split[0]);
+		return FALSE;
+	}
+	g_strlcpy (version->triplet, marker_split[0], 32);
+	if (0 == g_strlcpy (version->boardname, v_split[0], 32)) {
```
hughsie, line 56: Would g_strndup() be more suitable here? It's hardly fast path.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
```diff
@@ -0,0 +1,75 @@
...
+				     FWUPD_ERROR_INTERNAL,
+				     "Empty sha1 hash.");
+		return FALSE;
+	}
+	version->dirty = (NULL != g_strrstr(v_split[1], "+"));
+
+	return TRUE;
+}
+
+
```
hughsie, line 75: extra newlines

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+			self->ep_num = g_usb_endpoint_get_address (ep) & 0x7f;
+			self->chunk_len = g_usb_endpoint_get_maximum_packet_size (ep);
+
+			return TRUE;
+		}
+	}
+	g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_FOUND,
+			     "No Google Update interface found.");
```
hughsie, line 68: no ., and maybe drop the Google?

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+	FuCrOSECUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+
+	/* very much like usb_updater2's usb_findit() */
+
+	if (!fu_cros_ec_usb_device_find_interface (device, error)) {
+		g_prefix_error (error, "failed to find update interface: ");
+		return FALSE;
+	}
+
+	if (0 == self->chunk_len) {
```
hughsie, line 102: by convention, and for no other reason, we tend to do `self->chunk_len == 0`. Not a hill I'm going to die on tho.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+		}
+	}
+
+	/* read reply back */
+	if (inbuf != NULL && inlen > 0) {
+		actual = 0;
+		if (!g_usb_device_bulk_transfer (usb_device,
+						 self->ep_num | 0x80,
+						 inbuf, inlen,
+						 &actual, 5000, NULL, error)) {
```
hughsie, line 148: the 5000 and 2000 seem a bit magic. Maybe breeak those out as #2044

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+			return FALSE;
+		}
+		if ((actual != inlen) && !allow_less) {
```
hughsie, line 151: no need for the inner brackets.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+				     G_IO_ERROR,
+				     G_IO_ERROR_PARTIAL_INPUT,
+				     "only received %" G_GSIZE_FORMAT "/%"
+				     G_GSIZE_FORMAT " bytes\n",
+				     actual, outlen);
+			return FALSE;
+		}
+	}
+
+	if (rxed_count)
```
hughsie, line 162: != NULL please

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+	fu_common_string_append_ku (str, idt, "ProtocolVersion",
+				    self->protocol_version);
+	fu_common_string_append_ku (str, idt, "HeaderType",
+				    self->header_type);
+	fu_common_string_append_ku (str, idt, "MaxPDUSize",
+				    self->targ.common.maximum_pdu_size);
+	flash_protection = g_strdup_printf ("%04x",
+					    self->targ.common.flash_protection);
+	fu_common_string_append_kv (str, idt, "FlashProtectionStatus",
+				    flash_protection);
```
hughsie, line 306: I think you can use fu_common_string_append_kx for this.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+	fu_common_string_append_kv (str, idt, "RawVersion",
+				    self->targ.common.version);
+	fu_common_string_append_ku (str, idt, "KeyVersion",
+				    self->targ.common.key_version);
+	min_rollback = g_strdup_printf ("%" G_GINT32_FORMAT,
+					self->targ.common.min_rollback);
+	fu_common_string_append_kv (str, idt, "Minimum Rollback", min_rollback);
```
hughsie, line 313: and fu_common_string_append_ku? with a cast?

plugins/cros_ec-usb/meson.build
bleungatchromium: > I think `fu-cros_ec-usb-device.c/h` -> `fu-cros-ec-device.c/h` is better

Hi @hughsie. In the future, if this plugin were to support non-USB updates would it make sense for this file to stay as"fu-cros-ec-usb-device.c"? 

Off the top of my head, there are two non-USB methods to update other cros-ec devices:
1. Google USB-C Alt Mode using a custom kernel driver
2. flashrom

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: The minimum rollback is a signed int, and the output of "-1" is meaningful here, and this was intended to display that number.

plugins/cros_ec-usb/meson.build
hughsie: Yup, please leave this as usb if there are other methods planned.

bleungatchromium: Done.

plugins/meson.build
bleungatchromium: Done.

plugins/cros_ec-usb/README.md
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

plugins/cros_ec-usb/fu-cros_ec-usb-device.c
bleungatchromium: Done.

contrib/ci/dependencies.xml
bleungatchromium: Hi @hughsie,

I wasn't able to figure out how to install this from pip, as libgusb isn't listed on https://pypi.org/. Any other suggestion?

`[2020-05-23 00:47:10] [build] |Subproject gusb finished.
[2020-05-23 00:47:10] [build] meson.build:198:0: ERROR: Version 0.3.2 of subproject dependency gusb already cached, requested incompatible version ['>= 0.3.3'] for dep gusb`

It's not clear to me why it goes to 0.3.2 of gusb, but not any further.

superm1: Ah... I see what's happening!  it's because of `subprojects/libgusb.wrap`.  It's anchored at 0.3.2 (IE https://github.com/hughsie/libgusb/commit/43b327a1e213aff00833842c455a796a068077cd).  Update that to the commit matching 0.3.3 and the subproject fallback magic should start working and fix CI.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+	 */
+	union {
+		struct first_response_pdu rpdu;
+		guint32 legacy_resp;
+	} start_resp;
+
+	/* send start request */
+
+	/* Flush all data from endpoint to recover in case of error. */
+	/* FIXME(bleung): b/157109635. This flush causes problems with the next
```
superm1, line 191: outside of chrome OS development I don't think b/#### is very well understood.  Assuming this is a public bug number, could you expand this to something resolvable?

That being said having no idea on the underlying architecture, is this just a case that the flush command returns before the content has _actually_ propagated to the other parts in the black box?  If so, a reasonable thing to do would be to throw a loop here that will try some reasonable number of times until the transfers stop timing out.

I do something similar in the Dell dock setup routine where there is a corner race condition that you could potentially try to probe some composite devices before they've finished booting if you plug everything in at once.
https://github.com/fwupd/fwupd/blob/master/plugins/dell-dock/fu-dell-dock-i2c-ec.c#L916

This is also the reason we have the helper function `fu_device_retry` for sending some commands.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+			     self->chunk_len);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+do_xfer (FuCrosEcUsbDevice * cros_ec_device, guint8 *outbuf, gsize outlen,
```
superm1, line 117: We typically try prefix all function symbols with the plugin name if possible to prevent an unlikely collision when multiple plugins or libraries get loaded.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+	 guint8 *inbuf, gsize inlen, gboolean allow_less,
+	 gsize *rxed_count, GError **error)
+{
+	GUsbDevice *usb_device = fu_usb_device_get_dev (FU_USB_DEVICE (cros_ec_device));
+	FuCrosEcUsbDevice *self = cros_ec_device;
```
superm1, line 122: This line seems superfluous as no casting gets done at all.  Perhaps just rename the argument parameter to `self`?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,327 @@
...
+					   self->ep_num | 0x80,
+					   inbuf, self->chunk_len,
+					   &actual, 10, NULL, error)) {
+
+		g_debug ("flush!\n");
+	}
+	*/
+
+	memset(&ufh, 0, sizeof(ufh));
+	ufh.block_size = htobe32 (sizeof(ufh));
```
superm1, line 207: `GINT32_TO_BE` perhaps?

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Hi @superm1,

I can change this to an issue in this public project. I got some comments from the initial developer of usb_updater2.c that I can reflect out to the public issue.

I suspect that there is some underlying difference between my fwupd plugin implementation's use of bulk transfer (which is async) versus, the one used in usb_updater2, which calls the synchronous libusb bulk transfer. We can discuss it in the issue further.

contrib/ci/dependencies.xml
bleungatchromium: @superm1, thanks! That did it! New builds will pass green now.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,321 @@
...
+	} start_resp;
+
+	/* send start request */
+
+	/* flush all data from endpoint to recover in case of error */
+	while (g_usb_device_bulk_transfer (usb_device,
+					   self->ep_num | 0x80,
+					   inbuf, self->chunk_len,
+					   &actual, 10, NULL, NULL))
+		g_debug ("flush!\n");
```
superm1, line 198: Some more conversation around this flushing routine.  Since it's a while loop matching on the success return case, waiting for the fail case to break seems counterintutive.  Can it be a fixed number of calls in a for loop (or a `fu_device_retry`)?  Or does it really need to be bulk transfer until it fails?

If done with `fu_device_retry` you can probably set the exact circumstance that it should be a success and fail, how many retries and it will be much more obvious what's going on.

bleungatchromium: Hi @superm1 

I will look into implementing it as a fu_device_retry.

From what I understand about what this is seeking to accomplish, the normal case would result in a failed bulk read on the 1st attempt, since when the program starts, or when the usb device is attached, there was no command sent to the device, and no data is ready to be sent back to the host (since nothing was requested).

In the case where a flush is necessary, it would actually read bytes out and the bulk transfer would succeed, indicating that something was requested by a previous iteration of this updater (but maybe it crashed or was interrupted leaving the data buffered by the EC).

I think we can set some upper limit, but I'm still mulling over what that number should be.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: https://github.com/fwupd/fwupd/issues/2138 - This was fixed. Thanks!

plugins/cros-ec/fu-cros-ec-usb-device.c
superm1: Thanks for explaining it.  I think an upper limit makes sense so you don't have the possibility to hang the daemon in a loop here because of a possible firmware bug.

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done

plugins/cros-ec/fu-cros-ec-common.c
```diff
@@ -0,0 +1,72 @@
...
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "No version string to parse");
+		return FALSE;
+	}
+
+	/* sample version string: cheese_v1.1.1755-4da9520 */
+	v_split = g_strsplit (version_raw, "_v", 2);
+	if (2 > g_strv_length (v_split)) {
```
hughsie, line 31: other way around please

plugins/cros-ec/fu-cros-ec-common.c
```diff
@@ -0,0 +1,72 @@
...
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "Version marker not found");
+		return FALSE;
+	}
+	marker_split = g_strsplit_set (v_split[1], "-+", 2);
+	if (2 > g_strv_length (marker_split)) {
```
hughsie, line 39: same

plugins/cros-ec/fu-cros-ec-common.c
```diff
@@ -0,0 +1,72 @@
...
+	triplet_split = g_strsplit_set (marker_split[0], ".", 3);
+	if (3 > g_strv_length (triplet_split)) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Improper version triplet: %s", marker_split[0]);
+		return FALSE;
+	}
+	g_strlcpy (version->triplet, marker_split[0], 32);
+	if (0 == g_strlcpy (version->boardname, v_split[0], 32)) {
```
hughsie, line 55: same

plugins/cros-ec/fu-cros-ec-common.c
```diff
@@ -0,0 +1,72 @@
...
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "Empty board name");
+		return FALSE;
+	}
+	if (0 == g_strlcpy (version->sha1, marker_split[1], 32)) {
```
hughsie, line 62: same

plugins/cros-ec/fu-cros-ec-common.c
```diff
@@ -0,0 +1,72 @@
...
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "Empty sha1 hash");
```
hughsie, line 66: SHA

plugins/cros-ec/fu-cros-ec-common.h
```diff
@@ -0,0 +1,142 @@
...
+	UPDATE_HEADER_TYPE_CR50 = 0, /* Must be 0 for backwards compatibility */
+	UPDATE_HEADER_TYPE_COMMON = 1,
+};
+
+struct cros_ec_version {
+	gchar boardname[32];
+	gchar triplet[32];
+	gchar sha1[32];
+	gboolean dirty;
+};
```
hughsie, line 138: can you align the struct members with tab then a single space please.

plugins/cros-ec/fu-cros-ec-common.h
```diff
@@ -0,0 +1,142 @@
...
+
+gboolean fu_cros_ec_parse_version (const gchar *version_raw,
+				   struct cros_ec_version *version,
+				   GError **error);
```
hughsie, line 142: can you copy the indenting from other source files please.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,323 @@
...
+#include <string.h>
+
+#include "fu-cros-ec-usb-device.h"
+#include "fu-cros-ec-common.h"
+
+#define USB_SUBCLASS_GOOGLE_UPDATE 0x53
+#define USB_PROTOCOL_GOOGLE_UPDATE 0xff
+
+#define BULK_SEND_TIMEOUT_MS 2000
+#define BULK_RECV_TIMEOUT_MS 5000
```
hughsie, line 18: align the values with tabs pls

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,323 @@
...
+			     G_IO_ERROR,
+			     G_IO_ERROR_NOT_SUPPORTED,
+			     "Unsupported protocol version %d: \n",
+			     self->protocol_version);
+		return FALSE;
+	}
+	self->header_type = GUINT16_FROM_BE (start_resp.rpdu.header_type);
+
+	error_code = GUINT32_FROM_BE (start_resp.rpdu.return_value);
+	if (error_code) {
```
hughsie, line 230: != 0?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,323 @@
...
+		GUINT32_FROM_BE (start_resp.rpdu.common.maximum_pdu_size);
+	self->targ.common.flash_protection =
+		GUINT32_FROM_BE (start_resp.rpdu.common.flash_protection);
+	self->targ.common.min_rollback = GINT32_FROM_BE (start_resp.rpdu.common.min_rollback);
+	self->targ.common.key_version = GUINT32_FROM_BE (start_resp.rpdu.common.key_version);
+
+	if (!fu_cros_ec_parse_version (self->targ.common.version,
+				       &self->version, error)) {
+		g_prefix_error (error,
+				"Failed parsing device's version: %32s\n",
```
hughsie, line 250: no newline for GErrors please

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,323 @@
...
+				self->targ.common.version);
+		return FALSE;
+	}
+
+	fu_device_set_version (FU_DEVICE(device), self->version.triplet);
+	fu_device_add_instance_id (FU_DEVICE(device), self->version.boardname);
```
hughsie, line 256: space before bracket

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,323 @@
...
+				    self->targ.common.maximum_pdu_size);
+	fu_common_string_append_kx (str, idt, "FlashProtectionStatus",
+				    self->targ.common.flash_protection);
+	fu_common_string_append_kv (str, idt, "RawVersion",
+				    self->targ.common.version);
+	fu_common_string_append_ku (str, idt, "KeyVersion",
+				    self->targ.common.key_version);
+	min_rollback = g_strdup_printf ("%" G_GINT32_FORMAT,
+					self->targ.common.min_rollback);
+	fu_common_string_append_kv (str, idt, "Minimum Rollback", min_rollback);
```
hughsie, line 309: No space in the title pls

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,323 @@
...
+}
+
+
+static void
```
hughsie, line 313: double newline

subprojects/gusb.wrap
```diff
@@ -1,4 +1,4 @@
 [wrap-git]
 directory = gusb
 url = https://github.com/hughsie/libgusb.git
-revision = 43b327a1e213aff00833842c455a796a068077cd
+revision = 1281d6ff0789ea8e4a825d0c379d175254e269ee
```
hughsie, line 4: can we use a tag name here? Like `0.3.4`

hughsie:
Tons more nits, but I think we're nearly there now.

superm1:
> Tons more nits, but I think we're nearly there now.

Likewise outside of my comment around the flushing behavior and trying to use `fu_device_retry` instead with an upper limit, I have no other concerns.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,322 @@
...
+	return TRUE;
+}
+
+static void
+fu_cros_ec_usb_device_init (FuCrosEcUsbDevice *device)
+{
+	fu_device_set_version_format (FU_DEVICE (device), FWUPD_VERSION_FORMAT_TRIPLET);
+}
+
+
```
superm1, line 285: double newline here

subprojects/gusb.wrap
bleungatchromium: Done. Set to 0.3.3.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Fixed.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Fixed.

plugins/cros-ec/fu-cros-ec-common.h
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Hi @superm1.

I've gone ahead and implemented the fu_device_retry. After experimenting with purposefully wedging the device, I got the handle on how fu_device_retry works.

I think I might actually refactor another part of the _setup() in order to use a separate fu_device_retry as well. Thanks for the tip.

bleungatchromium:
Hi @superm1,

I've pushed this version which implements `fu_device_retry` for flush, but I've decided to try to refactor another part of `fu_cros_ec_usb_device_setup` which has been unreliable with a retry as well. So please standby, there will be at least one more revision.

plugins/cros_ec-usb/fu-cros_ec-usb-common.c
bleungatchromium: Hmm. I've allocated enough space in cros_ec_version for these strings, so a straight up g_strlcpy saves me from having to free it later separately.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,343 @@
...
+		struct first_response_pdu rpdu;
+		guint32 legacy_resp;
+	} start_resp;
+
+	/* send start request */
+
+	/* flush all data from endpoint to recover in case of error */
+	if (!fu_device_retry (device, fu_cros_ec_usb_device_flush,
+			      FLUSH_RETRY_CNT, NULL, error)) {
+		g_prefix_error (error, "Failed to flush device to idle state");
```
superm1, line 218: since this is a prefix error you should put a ": " at the end so it reads out properly.

superm1: also, no need to set capital F, it might not be the very first error prefixed so it might look funny if you get

`error calibrating flux capacitor: Failed to flush device to idle stateflushing 5 bytes`

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,343 @@
...
+	g_autofree guint8 *inbuf = g_malloc0 (self->chunk_len);
+
+	if (g_usb_device_bulk_transfer (usb_device, self->ep_num | 0x80, inbuf,
+					self->chunk_len, &actual,
+					FLUSH_TIMEOUT_MS, NULL, NULL)) {
+		g_debug ("flushing %" G_GSIZE_FORMAT " bytes", actual);
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "flushing %" G_GSIZE_FORMAT " bytes\n", actual);
```
superm1, line 188: don't think you need the end `\n` here

superm1:
> Hi @superm1,
> 
> I've pushed this version which implements `fu_device_retry` for flush, but I've decided to try to refactor another part of `fu_cros_ec_usb_device_setup` which has been unreliable with a retry as well. So please standby, there will be at least one more revision.

looks like an effective use of it and it makes this logic much more obvious to follow, good job 👍

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Thanks. Did this and cleaned this up elsewhere too.

bleungatchromium:
- [X] Fill out `README.md` with update protocol
- [X] Fill out `README.md` with any custom quirks and flags
- [X] Fill out `README.md` with the vendor ID security value
- [X] CI run for at least one target
- [X] Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc).

Thanks!

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,357 @@
...
+fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
+{
+	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	guint32 error_code;
+	START_RESP start_resp;
+
+	/* flush all data from endpoint to recover in case of error */
+	if (!fu_device_retry (device, fu_cros_ec_usb_device_flush,
+			      SETUP_RETRY_CNT, NULL, error)) {
+		g_prefix_error (error, "Failed to flush device to idle state: ");
```
superm1, line 242: nit: lower case F for this

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,357 @@
...
+		return FALSE;
+	}
+
+	/* send start request */
+	if (!fu_device_retry (device, fu_cros_ec_usb_device_start_request,
+			      SETUP_RETRY_CNT, &start_resp, error)) {
+		g_prefix_error (error, "Failed to send start request: ");
```
superm1, line 249: nit: lower case F for this

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,357 @@
...
+		return FALSE;
+	}
+	self->header_type = GUINT16_FROM_BE (start_resp.rpdu.header_type);
+
+	error_code = GUINT32_FROM_BE (start_resp.rpdu.return_value);
+	if (error_code != 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "Target reporting error %u", error_code);
```
superm1, line 270: lower case T here

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -0,0 +1,357 @@
...
+		GUINT32_FROM_BE (start_resp.rpdu.common.maximum_pdu_size);
+	self->targ.common.flash_protection =
+		GUINT32_FROM_BE (start_resp.rpdu.common.flash_protection);
+	self->targ.common.min_rollback = GINT32_FROM_BE (start_resp.rpdu.common.min_rollback);
+	self->targ.common.key_version = GUINT32_FROM_BE (start_resp.rpdu.common.key_version);
+
+	if (!fu_cros_ec_parse_version (self->targ.common.version,
+				       &self->version, error)) {
+		g_prefix_error (error,
+				"Failed parsing device's version: %32s: ",
```
superm1, line 286: lower case F here

bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done

plugins/cros-ec/fu-cros-ec-usb-device.c

PR #2134, opened 2020-05-26
dell-dock: Capture the dock SKU in metadata

superm1:
Should be helpful in reproducing failure reports.
If no opposition, I do intend to backport this back as far as possible too in hopes it increases the content in failure reports.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -165,6 +165,25 @@ fu_plugin_dell_dock_get_ec (GPtrArray *devices)
 }
 
 gboolean
+fu_plugin_composite_prepare (FuPlugin *plugin, GPtrArray *devices,
+			     GError **error)
+{
+	FuDevice *parent = fu_plugin_dell_dock_get_ec (devices);
+	const gchar *sku;
+	if (parent == NULL)
+		return TRUE;
+	sku = fu_dell_dock_ec_get_module_type (parent);
+	if (sku == NULL) {
+		g_set_error_literal (error, FWUPD_ERROR, FWUPD_ERROR_INTERNAL,
+				     "unable to detect SKU");
+		return FALSE;
+	}
+	fu_plugin_add_report_metadata (plugin, "DellDockSKU", sku);
+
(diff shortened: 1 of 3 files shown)


plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -164,6 +164,27 @@ fu_plugin_dell_dock_get_ec (GPtrArray *devices)
 	return ec_parent;
 }
 
+gboolean
+fu_plugin_update (FuPlugin *plugin,
```
hughsie, line 168: maybe fu_plugin_runner_composite_prepare instead?

hughsie: Actually, maybe this is a hack, as it seems the "last set" thing wins. Shouldn't we get the metadata from the FuDevice directly? Maybe an extra vfunc or property? I'm happy to do the plumbing if you think this is a good idea.

superm1: Sure a vfunc on the device that adds metadata sounds good. Go ahead :)

hughsie: Maybe we want everything in `fu_device_set_metadata()`? I can't see anything that we wouldn't want to include.

hughsie: Ya, I *think* doing `fu_device_set_metadata(dev, "DellDockSKU", sku)` gives us the metadata in in the report right now.

superm1: We might need to be cognizant that there is no PII in the metadata, but otherwise yes I think that makes sense.

superm1: I tried `fu_device_set_metadata*` and it doesn't flow into the report.  However I do like doing it in `composite_prepare` as this metadata can then be set even in the context that another non-dell-dock plugin performs the update.

Think of it as "If this is a Dell dock add the metadata in the upload report of which model it was"

PR #2140, opened 2020-05-26
Skip module unloading only if we are actually running under valgrind

superm1:
See also: https://github.com/fwupd/fwupd/issues/2119
Reported-by: Anton Farygin <rider@altlinux.org>

Based on a patch by Gleb Fotengauer-Malinovskiy <glebfm@altlinux.org>

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Fails CI; I ran out of time before PTO last week. :(

superm1:
I'll see if I can figure out the CI problem.  Stop being so responsive on PTO though 🍵

PR #2142, opened 2020-05-27
dell-dock: prevent updates to occur via synaptics-mst plugin

superm1:
Although they normally work, some failures have been reported in the field
related to the MST hub not responding in the MST plugin.

When these failures have occurred the dell_dock plugin also fails to
enumerate.

So rather than allow some people who don't have dell_dock compiled to
update their MST hub using synaptics_mst, perform ALL updates for
mst hub via dell_dock.

```
18:06:24:0324 FuPluginSynapticsMST no device found on drm_dp_aux1: VMM5331 inside Dell dock is only supported by dell_dock
```

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/synaptics-mst/synaptics-mst.quirk
```diff
@@ -40,7 +40,8 @@ Name = Dell Rugged Platform
 DeviceKind = system
 
 [SynapticsMSTBoardID=259]
-Name = Dell dock
+Name = Dell WD19 dock
+Plugin = dell_dock
```
hughsie, line 44: I feel like this chunk should probably live in the dell-dock quirk file now.

plugins/synaptics-mst/fu-synaptics-mst-device.c
```diff
@@ -1092,6 +1093,16 @@ fu_synaptics_mst_device_rescan (FuDevice *device, GError **error)
 	}
 	fu_device_set_name (FU_DEVICE (self), name);
 
+	plugin = fu_quirks_lookup_by_id (quirks, group, FU_QUIRKS_PLUGIN);
+	if (plugin != NULL && g_strcmp0 (plugin, "synaptics_mst") != 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_DATA,
```
hughsie, line 1100: NOT_SUPPORTED?

plugins/synaptics-mst/fu-synaptics-mst-device.c
```diff
@@ -1000,6 +1000,7 @@ fu_synaptics_mst_device_rescan (FuDevice *device, GError **error)
 	const gchar *guid_template;
 	const gchar *name_parent;
 	const gchar *name_family;
+	const gchar *plugin = NULL;
```
hughsie, line 1003: no need for the NULL

plugins/synaptics-mst/synaptics-mst.quirk
hughsie: Actually, maybe we just remove the chunk altogether? Or do something like Flags=ignore?

PR #2148, opened 2020-05-28
trivial: uefi: fix dell TPM updates

superm1:
Adding an extra header makes the firmware reject the GUID in the real
header.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-uefi-device.c
+++ b/plugins/uefi/fu-uefi-device.c
@@ -35,6 +35,7 @@ struct _FuUefiDevice {
 	guint64			 fmp_hardware_instance;
 	gboolean		 missing_header;
 	gboolean		 automounted_esp;
+	gboolean		 requires_header;
 };
 
 G_DEFINE_TYPE (FuUefiDevice, fu_uefi_device, FU_TYPE_DEVICE)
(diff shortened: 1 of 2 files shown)


plugins/uefi/fu-uefi-device.c
```diff
@@ -324,6 +324,9 @@ fu_uefi_device_fixup_firmware (FuDevice *device, GBytes *fw, GError **error)
 	} else if (fu_uefi_device_get_kind (self) == FU_UEFI_DEVICE_KIND_FMP) {
 		g_debug ("performing FMP update");
 		return g_bytes_new_from_bytes (fw, 0, fw_length);
+	} else if (fu_uefi_device_get_kind (self) == FU_UEFI_DEVICE_KIND_DELL_TPM_FIRMWARE) {
```
hughsie, line 327: I'd rather have an explicit property, e.g. `fu_uefi_device_get_requires_header()` or something like that which could explicitly be unset by FMP and the TPM hardware at device creation time.

superm1: adjusted

PR #2150, opened 2020-05-29
Always enforce the metadata signature has a valid timestamp

hughsie:
Although this is something that we have always done on the LVFS, corporate
deployments that resign the firmware or metadata might not be signing the files
in the same way.

Always require a timestamp to prevent allowing an inadvertent rollback attack.


hughsie:
Snap and win32 seems to be failing

    subprojects/gusb/gusb/meson.build:98:0: ERROR: Program(s) ['g-ir-scanner'] not found or not executable

@superm1 any chance you can have a look at this please?

PR #2155, opened 2020-06-01
Detect if the MEI device has known security issues

hughsie:
If it has, fail HSI-1.

### DIFF
--- a/libfwupd/fwupd-security-attr.h
+++ b/libfwupd/fwupd-security-attr.h
@@ -124,6 +124,7 @@ typedef enum {
 #define FWUPD_SECURITY_ATTR_ID_KERNEL_TAINTED		"org.fwupd.hsi.KernelTainted"		/* Since: 1.5.0 */
 #define FWUPD_SECURITY_ATTR_ID_MEI_MANUFACTURING_MODE	"org.fwupd.hsi.MeiManufacturingMode"	/* Since: 1.5.0 */
 #define FWUPD_SECURITY_ATTR_ID_MEI_OVERRIDE_STRAP	"org.fwupd.hsi.MeiOverrideStrap"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_MEI_VERSION		"org.fwupd.hsi.MeiVersion"		/* Since: 1.5.0 */
 #define FWUPD_SECURITY_ATTR_ID_SPI_BIOSWE		"org.fwupd.hsi.SpiBioswe"		/* Since: 1.5.0 */
 #define FWUPD_SECURITY_ATTR_ID_SPI_BLE			"org.fwupd.hsi.SpiBle"			/* Since: 1.5.0 */
 #define FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP		"org.fwupd.hsi.SpiSmmBwp"		/* Since: 1.5.0 */
--- a/plugins/pci-mei/fu-mei-common.c
+++ b/plugins/pci-mei/fu-mei-common.c
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-mei-common.h"
+
+const gchar *
+fu_mei_common_family_to_string (FuMeiFamily family)
+{
+	if (family == FU_MEI_FAMILY_SPS)
+		return "SPS";
+	if (family == FU_MEI_FAMILY_TXE)
+		return "TXE";
+	if (family == FU_MEI_FAMILY_ME)
+		return "ME";
+	if (family == FU_MEI_FAMILY_CSME)
+		return "CSME";
+	return "AMT";
+}
+
+static gint
+fu_mei_common_cmp_version (FuMeiVersion *vers1, FuMeiVersion *vers2)
+{
+	guint16 vers1buf[] = {
+		vers1->major,
+		vers1->minor,
+		vers1->hotfix,
+		vers1->buildno,
+	};
+	guint16 vers2buf[] = {
+		vers2->major,
+		vers2->minor,
+		vers2->hotfix,
+		vers2->buildno,
+	};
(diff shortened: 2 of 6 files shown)


libfwupd/fwupd-security-attr.h
```diff
@@ -88,6 +88,8 @@ typedef enum {
  * @FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND:		NOt found
  * @FWUPD_SECURITY_ATTR_RESULT_SUPPORTED:		Supported
  * @FWUPD_SECURITY_ATTR_RESULT_NOT_SUPPORTED:		Not supported
+ * @FWUPD_SECURITY_ATTR_RESULT_SECURE:			Secure
```
superm1, line 91: I think this is a risky word to use.  It's only going to be secure in the sense that the current known vulnerabilities from when the daemon was last compiled/updated.
I would err on something like "No known vulnerabilities", or "Vulnerabilities present".

hughsie: Ya, I wasn't sure about that to. Got any good alternatives to `SECURE`? Perhaps we just VALID for no issues, and NOT_VALID when known issues are present?

superm1: Well I was thinking vulnerabilities/no vulnerabilities is better since it can be overloaded into other things that might be measured for vulnerabilities later.

plugins/pci-mei/fu-mei-common.c
```diff
@@ -0,0 +1,148 @@
...
+		if (vers1buf[i] < vers2buf[i])
+			return -1;
+		if (vers1buf[i] > vers2buf[i])
+			return 1;
+	}
+	return 0;
+}
+
+FuMeiIssue
+fu_mei_common_is_csme_vulnerable (FuMeiVersion *vers)
```
superm1, line 50: To any layman looking at this, it's missing the CVE or INTEL-SA identifiers matching the particular firmware versions.  I think that those are important to include, especially as this is going to rev up with the daemon version as more vulnerabilities are disclosed.

hughsie: Intel don't give us those :( If you know of a list I can copy from, I'm happy to do that.

superm1: Well I think at least some of them you can probably tell by looking at CVE's listed in https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00086.html

superm1: If that's not workable enough, at least need to mention where to look for updated version numbers with the right hotfixes.

libfwupd/fwupd-security-attr.h
hughsie: The other thing I pondered was making the HSI from a two dimensional thing (e.g. firmware platform security * runtime issues) into a three dimensional thing (firmware platform security * runtime issues * known security issues) but I think that's making it too complicated. The user can't fix the firmware if there is a known CSME vulnerability so it's not runtime, and we've said that the system can be downgraded HSI version after launch as we add new tests, so I guess it makes sense the platform HSI can be downgraded if the vendor is not "quick enough" to build a new ME during the CVE embargo period.

superm1: >but I think that's making it too complicated

Yeah 3d is too complex I agree.

>The user can't fix the firmware if there is a known CSME vulnerability so it's not runtime, and we've said that the system can be downgraded HSI version after launch as we add new tests, 

Which is part of why I think it will be important to document exactly which CVEs or vulnerabilities are being checked for/measured in the spec.  If a new one is added, it goes in the spec and it's obvious in the code the changes needed for it.

>so I guess it makes sense the platform HSI can be downgraded if the vendor is not "quick enough" to build a new ME during the CVE embargo period.

Sometimes it's not just this, but if it causes a genuine regression that needs to be fixed.

hughsie: Thinking about this a bit more, I think we could just print "OK" for the "no vulnerability found" message, and "Not valid" for the case where we think there might be a CSME issue.

plugins/pci-mei/fu-mei-common.c
hughsie: I did imagine such a thing would be explained in the _Host Security ID Specification_ on the logic that we can update it async with fwupd releases. I think every `FWUPD_SECURITY_ATTR_ID_*` would have a href link in the spec with a lot more detail.

libfwupd/fwupd-security-attr.h
superm1: That seems fine

plugins/pci-mei/fu-mei-common.c
superm1: OK..

plugins/pci-mei/fu-plugin-pci-mei.c
```diff
@@ -99,6 +193,32 @@ fu_plugin_add_security_attrs_override_strap (FuPlugin *plugin, FuSecurityAttrs *
...
 }
 
+static void
+fu_plugin_add_security_attrs_csme_version (FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+	FuPluginData *priv = fu_plugin_get_data (plugin);
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	if (priv->issue == FU_MEI_ISSUE_UNKNOWN)
+		return;
```
superm1, line 203: I think a warning into syslog makes sense here on the logic that an unknown ME architecture needs to start getting tracked too and this will cause some noise to be stirred.

hughsie: syslog, or just fail the HSI test?

superm1: Hmm, I think I'm swayed more towards fail the syslog approach.  A "new" ME architecture is very likely to include all the previous hotfixes.  By the time a vulnerability came out it should have more details.

hughsie: I've added a g_warning which should make it into the logs. I also saved the FuMeiVersion so the log message would be half-useful.

Issue #2156, opened 2020-06-02
This is perhaps a feature request: document somewhere prominent how to install the Flatpak version


LinuxOnTheDesktop:
**Describe the question**
I think it is insufficiently clear how to install (and run) the Flatpak version. Admittedly many distributions have Software Managers that make it somewhat evident how to do that, but why not, on github.com/fwupd/fwupd, provide the terminal commands?

**fwupd version information**
Not applicable.

EDIT: This is now a support request too. For, I see this - which I did not before, when I used the Snap version of fwupd. (Note: Linux Mint 20 is going to drop Snap.)
```

$ sudo flatpak run org.freedesktop.fwupd refresh
Loading…                 [-                                      ]03:04:37:0751 FuEngine             device fb1a5fd6e7d8cbf5d5a2ba7df68aee106ce41027 [Unifying Receiver] does not define an update protocol
03:04:37:0963 FuEngine             device fb1a5fd6e7d8cbf5d5a2ba7df68aee106ce41027 [Unifying Receiver] does not define an update protocol
Loading…                 [***************************************]
$ 
```
And my initial attempt, that did not used sudo, produced a confusing welter of error messages.

Also, `sudo flatpak run org.freedesktop.fwupd get-updates` did nothing for so long (on a fast computer) that I thought the program had hung, though eventually the command worked as it should.

EDIT: even `flatpak run org.freedesktop.fwupd help` took over twenty seconds (and quite a few seconds to type!).

superm1:
I think this is perfect content to be contributed to the wiki: https://github.com/fwupd/fwupd/wiki

Please feel free to create a new wiki page there and link back here.

> EDIT: This is now a support request too. For, I see this - which I did not before, when I used the Snap version of fwupd. (Note: Linux Mint 20 is going to drop Snap.)
>03:04:37:0963 FuEngine             device fb1a5fd6e7d8cbf5d5a2ba7df68aee106ce41027 [Unifying Receiver] does not define an update protocol

I suspect this was happening in snap too, but you didn't notice it because it was in the daemon log.
When you run the snap you're running `fwupdtool` not `fwupdmgr` . `fwupdtool` includes both the daemon (engine) output.

Could you run:
```
sudo flatpak run org.freedesktop.fwupd get-devices --verbose
```

And share the resultant output?

>EDIT: even flatpak run org.freedesktop.fwupd help took over twenty seconds (and quite a few seconds to type!).

That's my experience with the flatpak too.

[...]

PR #2157, opened 2020-06-02
trivial: logitech_hidpp: set the protocol properly when bootloader un…

superm1:
…known

This appears to be a regression from c6ae0d998b7e5577b3de4763c25d9d90056bebf1
where the case of bootloader version 0 didn't get protocol set.

Fixes: #2156

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-runtime.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-runtime.c
@@ -230,10 +230,10 @@ fu_logitech_hidpp_runtime_setup_internal (FuDevice *device, GError **error)
 		    (self->version_bl_major == 0x03 && config[8] >= 0x02)) {
 			self->signed_firmware = TRUE;
 			fu_device_set_protocol (device, "com.logitech.unifyingsigned");
-		} else {
-			fu_device_set_protocol (device, "com.logitech.unifying");
 		}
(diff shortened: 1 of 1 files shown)


plugins/logitech-hidpp/fu-logitech-hidpp-runtime.c
```diff
@@ -230,10 +230,10 @@ fu_logitech_hidpp_runtime_setup_internal (FuDevice *device, GError **error)
 		    (self->version_bl_major == 0x03 && config[8] >= 0x02)) {
 			self->signed_firmware = TRUE;
 			fu_device_set_protocol (device, "com.logitech.unifyingsigned");
-		} else {
-			fu_device_set_protocol (device, "com.logitech.unifying");
 		}
 	}
+	if (!self->signed_firmware)
+		fu_device_set_protocol (device, "com.logitech.unifying");
```
hughsie, line 236: I'm wondering if we should set com.logitech.unifyingsigned here, but not a hill I want to die on.

superm1: I think if the code refactors that it makes sense such a second set of circumstances for signed to be set, otherwise not necessary.

PR #2158, opened 2020-06-02
trivial: add a bug report target for Dell WD19

superm1:
This is to direct people from https://github.com/fwupd/fwupd/wiki/LVFS-Triaged-Issue%3A-WD19%2A%3A-Device-ID-was-not-found/_edit

At some point I think we may remove it and collapse back to a generic report.  For now I would rather gather explicit details about the dock that aren't part of the generic report.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report-wd19.md
+++ b/.github/ISSUE_TEMPLATE/bug-report-wd19.md
@@ -1,0 +1,64 @@
+---
+name: Bug report (Dell WD19)
+about: Create a report to help us improve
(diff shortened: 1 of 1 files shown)


superm1:
I'm getting some pressure to get this enabled, so I'll merge this.  If you have any heartache, please scream after the fact and I'll adjust the report as needed.

hughsie:
No concerns dude.

PR #2159, opened 2020-06-02
Add a plugin to fail the HSI runtime for WPBT

hughsie:
The logic of the runtime failure is that it is something the user can fix
without requiring help from the OEM.

### DIFF
--- a/plugins/acpi-wpbt/README.md
+++ b/plugins/acpi-wpbt/README.md
@@ -1,0 +1,23 @@
+ACPI WPBT
+=========
+
+Introduction
+------------
+
+This plugin checks if the WPBT ACPI table has been exported. Although unused by
+Linux, Microsoft Windows will execute the EXE file specified by the WPBT table.
+The launching of the executable may be conditionalized on it being signed by a
+specific cryptographic key or can depend on site or system policy.
+
+WPBT should be disabled in the machine firmware setup, and it certainly should
+not be enabled by default at the factory. Most commonly WPBT is enabled when
+enabling the Computrace feature.
+
+Computrace uses WPBT for legitimate hardware recovery purposes. Less commonly
+an OEM may use it to install OEM-specific drivers and utilities.
+
+The result will be stored in a security attribute for HSI.
+
+If you wish to ignore the HSI failure (for example, because Computrace is
+actually desired) then you can remove this runtime issue by setting
+`BlacklistPlugins=acpi_wpbt` in `/etc/fwupd/daemon.conf`
--- a/plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
+++ b/plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-plugin-vfuncs.h"
+#include "fu-hash.h"
+
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+}
+
+void
+fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+	g_autofree gchar *fn = NULL;
+	g_autofree gchar *path = NULL;
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	/* create attr */
+	attr = fwupd_security_attr_new (FWUPD_SECURITY_ATTR_ID_ACPI_WPBT);
+	fwupd_security_attr_set_plugin (attr, fu_plugin_get_name (plugin));
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	fu_security_attrs_append (attrs, attr);
+
+	/* load WPBT table */
+	path = fu_common_get_path (FU_PATH_KIND_ACPI_TABLES);
+	fn = g_build_filename (path, "WPBT", NULL);
+	if (g_file_test (fn, G_FILE_TEST_EXISTS)) {
+		gsize bufsz = 0;
+		guint8 csum = 0;
+		g_autofree gchar *buf = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		if (!g_file_get_contents (fn, &buf, &bufsz, &error_local)) {
+			g_warning ("failed to read %s: %s", fn, error_local->message);
+			fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+			return;
+		}
+		/* test if the checksum is valid */
+		for (gsize i = 0; i < bufsz; i++)
+			csum += buf[i];
+		if (csum != 0) {
+			fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_VALID);
+			return;
+		}
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_ENABLED);
+		return;
+	}
+
+	/* success */
+	fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED);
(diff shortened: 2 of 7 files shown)


plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
```diff
@@ -0,0 +1,42 @@
...
+	/* create attr */
+	attr = fwupd_security_attr_new (FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_IDLE);
+	fwupd_security_attr_set_plugin (attr, fu_plugin_get_name (plugin));
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	fu_security_attrs_append (attrs, attr);
+
+	/* load WPBT table */
+	path = fu_common_get_path (FU_PATH_KIND_ACPI_TABLES);
+	fn = g_build_filename (path, "WPBT", NULL);
+	if (g_file_test (fn, G_FILE_TEST_EXISTS)) {
```
superm1, line 34: Is testing it exists enough for this? I would think you want to make sure it's actually a windows binary?

plugins/acpi-wpbt/README.md
```diff
@@ -0,0 +1,15 @@
...
+Introduction
+------------
+
+This plugin checks if the WPBT table has been exported. Although unused by
+Linux, Microsoft Windows will unconditionally execute the EXE file specified by
+the WPBT table. WPBT should be disabled in the firmware setup menu, and it
+certainly should not be enabled by default at the factory.
+
+This is used by CompuTrace for legitimate hardware recovery and less commonly
+to install OEM-specific drivers and utilities on a "clean" windows tool.
```
superm1, line 13: Since this has a legitimate use, I feel that we need a user who has chosen to enable something like computrace for their threat model to be able opt out of this being a runtime error.

plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
```diff
@@ -0,0 +1,42 @@
...
+
+void
+fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+	g_autofree gchar *fn = NULL;
+	g_autofree gchar *path = NULL;
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	/* create attr */
+	attr = fwupd_security_attr_new (FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_IDLE);
```
superm1, line 26: Suspend to idle?

plugins/acpi-wpbt/README.md
```diff
@@ -0,0 +1,15 @@
+ACPI WPBT
+=========
+
+Introduction
+------------
+
+This plugin checks if the WPBT table has been exported. Although unused by
+Linux, Microsoft Windows will unconditionally execute the EXE file specified by
+the WPBT table. WPBT should be disabled in the firmware setup menu, and it
+certainly should not be enabled by default at the factory.
```
superm1, line 10: Does any manufacturer call it this in BIOS setup?

plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
hughsie: Have you got a link to the WPBT spec? I can sure enough check for the PE header.

plugins/acpi-wpbt/README.md
hughsie: `BlacklistPlugins=`, no?

plugins/acpi-wpbt/README.md
hughsie: IIRC, it's exported when you enable Computrace on Lenovo.

plugins/acpi-wpbt/README.md
superm1: Yes, I think that is a sufficient lever. We should probably have it mentioned in the readme though to discuss your threat model how to actually do it.

plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
superm1: https://view.officeapps.live.com/op/view.aspx?src=http%3A%2F%2Fdownload.microsoft.com%2Fdownload%2F8%2FA%2F2%2F8A2FB72D-9B96-4E2D-A559-4A27CF905A80%2Fwindows-platform-binary-table.docx

superm1: Looking at that spec the WPBT itself doesn't have the binary, but rather points to memory that will have the binary.  I don't anticipate that the memory region is actually accessible from Linux userland.  So I think that leaves two options of varying degrees of difficulty:
1. Parse the table from fwupd, make sure if it measures out to that spec.
2. Add a new kernel module for parsing the table and _exporting_ out the binary to userland in a sysfs interface.  I think this is most interesting for a researchers interface of being able to do some analysis on actual things with what is in WPBT and probably overkill.

If it looks strikingly like a duck and it's quacks at least checksum to that of a duck;'s it's probably a duck.

plugins/acpi-wpbt/README.md
superm1: Right, my point was that I don't believe any manufacturer exports this to a user with the phrase `WPBT`, that's just the means to the end to enable a technology.

Most manufacturers export it as `Computrace` and that is probably what should be referenced here but just should be worded carefully not to step on the hypothetical toes by calling out their software as bad.  
It's just the way that the software and infrastructure works and some admins may choose to have it as an element of their security threat model.

plugins/acpi-wpbt/README.md
```diff
@@ -0,0 +1,15 @@
+ACPI WPBT
+=========
+
+Introduction
+------------
+
+This plugin checks if the WPBT table has been exported. Although unused by
+Linux, Microsoft Windows will unconditionally execute the EXE file specified by
```
superm1, line 8: According to the spec there are conditions to the launch (including failures for lack of signing or system policy).

plugins/acpi-wpbt/README.md
hughsie: I've added a chunk to the README.

plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
hughsie: Fixed, i didn't want to do all the work if you NAKd the idea :)

plugins/acpi-wpbt/fu-plugin-acpi-wpbt.c
hughsie: I also agree it's not terribly useful to see the exe itself, unless of course we can verify the checksum isn't one of the known-bad ones -- but that's getting dangerously close to an _AV scanner_, and I'm not sure that's where I want this to head.

I've added a test to check the file is readable and to verify the checksum isn't invalid.

plugins/acpi-wpbt/README.md
hughsie: Great, thanks. Added.

plugins/acpi-wpbt/README.md
hughsie: I've reworded things a bit, now about that?

superm1:
> The logic of the runtime failure is that it is something the user can fix
without requiring help from the OEM.

So there are a few more things I wanted to mention here about this being a runtime check.  

Although in the Computrace case a user can adjust it, the non-Computrace population of this (IE superfish) a user can't fix.  So at that point you almost need to _actually_ check the content to really decide if it's a bad thing.

Also these days WPBT isn't the only lever that exists to get binaries installed on a machine automatically in Windows.  WPBT is just notable because of it acting automatically without a network.

There are certainly other things that will cause the the Microsoft store to automatically download things too.

superm1:
@hughsie while we're cleaning up old PR's and issues, what's your intent on this one?

hughsie:
> what's your intent on this one

I'm not sure. I keep flip-flopping both ways.

superm1:
I guess to me it comes down to what you want the outcome to be from populating it.
* Is it something that a user should see and be going into firmware setup to try to turn off
* Something telling users to sharpen the pitchforks if they can't turn it off?
* Draw awareness to `WPBT`?

As it's not harmful to Linux at all, I don't really see any point in *any* of that.

hughsie:
> I don't really see any point in any of that.

Fair point.

PR #2162, opened 2020-06-08
Wip/hughsie/vli g object class

hughsie:


superm1:
regarding the CI failure, I think simply removing this line from debian.sh should fix it:
https://github.com/fwupd/fwupd/blob/master/contrib/ci/debian.sh#L42

hughsie:
> vli: Swapped proxy for device twice, removed else in attach_full() 

Does that work?! That patch doesn't make sense to me.

> We're a little confused as to the purpose of FuDevice *proxy

The `proxy` device performs actions on behalf of the `device` -- so in this case the proxy is the VL817 performing the GPIOB reset and the device is the VL211. We need to set the flags on the @device not the @proxy as we wait for the VL211 to "come back" after re-enumeration, and because it's downstream of the VL817 it takes a few hundred ms more to come back.

memily:
> Does that work?! That patch doesn't make sense to me.

It do.

> the proxy is the VL817 performing the GPIOB reset and the device is the VL211.

VL211 is standalone, independent of VL817. VL211 resets with standard control transfer. GPIOB reset should be triggered from VL817 for VL103 after write VL103 FW to Share SPI. (GPIOB should also be able to reset VL817 but I can't get VL817 to reset with GPIOB on my end...)

hughsie:
> because VL103 is not updating

I think the PID mapping is wrong; I've sent email.

PR #2165, opened 2020-06-09
fu-main: Make it clearer what lost name is and bump it to warning

superm1:
This happens often enough when people switch from snap to distro
package that we should mention it in non-verbose logs.

Fixes: #2112

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -1493,7 +1493,7 @@ fu_main_on_name_lost_cb (GDBusConnection *connection,
 			 gpointer user_data)
 {
 	FuMainPrivate *priv = (FuMainPrivate *) user_data;
-	g_debug ("FuMain: lost name: %s", name);
+	g_warning ("another service has claimed the dbus name %s", name);
 	g_main_loop_quit (priv->loop);
 }
 


src/fu-main.c
```diff
@@ -1493,7 +1493,7 @@ fu_main_on_name_lost_cb (GDBusConnection *connection,
 			 gpointer user_data)
 {
 	FuMainPrivate *priv = (FuMainPrivate *) user_data;
-	g_debug ("FuMain: lost name: %s", name);
+	g_warning ("another service has claimed the dbus name %s", name);
```
hughsie, line 1496: isn't this is a legitimate code path when we replace the system version of fwupd with one compiled locally?

superm1: Can one compiled locally actually take over?  Locally compiled one needs messing with default place to find dbus services (IE look at compilation page what we have down https://github.com/fwupd/fwupd/wiki/Compilation)

superm1: And furthermore - if that was the case, a system one was running and a locally compiled one takes over, I would certainly rather see that in logs that don't have verbose turned on..

hughsie: True.

Issue #2118, opened 2020-05-20
My lenovo X1 reboots and updates are not applied.


lhirlimann:
**Describe the bug**
I have firmware updates offered when I run fwupmgr :
```
[ludovic@saraan]~% sudo fwupdmgr update 
• Thunderbolt Controller has the latest available firmware version
• Embedded Controller has the latest available firmware version
Upgrade available for Intel Management Engine from 184.70.3626 to 184.77.3664
20KHCTO1WW must remain plugged into a power source for the duration of the update to avoid damage. Continue with update? [Y|n]: y
Downloading 184.77.3664 for Intel Management Engine...
Decompressing…           [***************************************]
Authenticating…          [***************************************]
Updating Intel Management Engine…                                ]
Scheduling…              [***************************************]
Successfully installed firmware
• SAMSUNG MZVLB512HAJQ-000L7 has the latest available firmware version
Upgrade available for System Firmware from 0.1.46 to 0.1.48
20KHCTO1WW must remain plugged into a power source for the duration of the update to avoid damage. Continue with update? [Y|n]: y
Downloading 0.1.48 for System Firmware...
Decompressing…           [***************************************]
Authenticating…          [***************************************]
Updating System Firmware…[            /                          ]
Scheduling…              [***************************************]
Successfully installed firmware

An update requires a reboot to complete. Restart now? [y|N]: n
```
When I reboot, I see messages about new capsule being found - it takes a while and then boot continues. firmware update don't seem to have been applied.

**Steps to Reproduce**
sudo fwupmgr update
reboot

**Expected behavior**

After the reboot running the update command returns that I'm up to date.

**fwupd version information**
Please provide the version of the daemon and client.
```shell
[ludovic@saraan]~% fwupdmgr --version
client version:	1.4.2
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.4.2
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):

dnf with enable-testing

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
[ludovic@saraan]~% fwupdmgr get-devices --show-all-devices
20KHCTO1WW
│
├─Thunderbolt Controller:
│     Device ID:           199c280418de0aa1347fc161a9df3c68dd5b0848
│     Summary:             Unmatched performance for high-speed I/O
│     Current version:     43.00
│     Vendor:              Lenovo (TBT:0x0109)
[...]

PR #2170, opened 2020-06-11
Record the UEFI failure in more cases

hughsie:
Ensure the historical error is set for failed NEEDS_REBOOT UEFI devices that do
not set LastAttemptStatus.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5707,6 +5707,7 @@ fu_engine_update_history_device (FuEngine *self, FuDevice *dev_history, GError *
 		fu_device_set_version (dev_history, fu_device_get_version (dev));
 		fu_device_remove_flag (dev_history, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION);
 		fu_device_set_update_state (dev_history, FWUPD_UPDATE_STATE_SUCCESS);
+		fu_device_set_update_error (dev_history, NULL);
 		return fu_history_modify_device (self->history, dev_history, error);
 	}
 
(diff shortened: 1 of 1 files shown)


hughsie:
@superm1 some sanity checking please

src/fu-engine.c
```diff
@@ -5707,6 +5707,7 @@ fu_engine_update_history_device (FuEngine *self, FuDevice *dev_history, GError *
 		fu_device_set_version (dev_history, fu_device_get_version (dev));
 		fu_device_remove_flag (dev_history, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION);
 		fu_device_set_update_state (dev_history, FWUPD_UPDATE_STATE_SUCCESS);
+		fu_device_set_update_error (dev_history, NULL);
```
superm1, line 5710: Has this been a problem in practice?  I think it does seem like a _logical_ thing to do, but when you're talking about UEFI errors I feel like it's not going to matter.  You'll get success or you won't, not a weird transient during your current boot.

src/fu-engine.c
```diff
@@ -5723,11 +5724,14 @@ fu_engine_update_history_device (FuEngine *self, FuDevice *dev_history, GError *
 	if (fu_device_get_update_state (dev) != FWUPD_UPDATE_STATE_FAILED &&
 	    fu_device_get_update_state (dev) != FWUPD_UPDATE_STATE_FAILED_TRANSIENT) {
 		g_debug ("falling back to generic failure");
+		fu_device_set_update_state (dev_history, FWUPD_UPDATE_STATE_FAILED);
```
superm1, line 5727: do you know what state is actually coming into this method?  If it's not failed and not failed-transient, doesn't that mean it's a plugin bug in not setting it?

src/fu-engine.c
hughsie: One Lenovo update: https://github.com/fwupd/fwupd/issues/2171

src/fu-engine.c
hughsie: If we're calling into GetResults vfunc (like the UEFI plugin) then the plugin doesn't know if a firmware was deployed, scheduled or whatever and just returns the ESRT value. In the case where we've just launched fwupd and the pending update firmware does not match the running firmware version then returning success is "correct" but not true.

My worry here is that we schedule an update that doesn't get run (e.g. because the user has set up custom secure boot keys, or deliberately aborts the flashing firmware and we'd record a hard failure, not `FAILED_TRANSIENT`. Of course In some cases we do want a hard `FAILED` status where the capsule was run, but didn't update the `Status` value.

src/fu-engine.c
hughsie: Nope, I meant https://github.com/fwupd/fwupd/issues/2118 -- sorry.

src/fu-engine.c
superm1: I guess relying upon strictly the ESRT status is not enough because users can put other items into the flow that break things like you say.

superm1:
While doing this, I do think you should also add some new categories to the report metadata though:
*  Explicit ESRT status
* Custom secure boot conf set

hughsie:
> Explicit ESRT status

One of those requests that leads me down a rabbithole... Big PR coming up next week.

PR #2177, opened 2020-06-12
Add support for Google flashmap layout

### LINKED ISSUE #2149: Add a parser for Google's FMAP firmware format.
Google firmware created as a part of the ChromiumOS project (Chromebook EC ifrmware, Chromebook system firmware, peripheral firmware) use the "fmap" layout.

As a part of building the cros-ec plugin (https://github.com/fwupd/fwupd/pull/2130), the next step is to process cros-ec's payloads, which are in fmap layout.

 See: 
https://www.chromium.org/chromium-os/firmware-porting-guide/fmap

A reference implementation that is a part of chromiumos is here:
https://chromium.googlesource.com/chromiumos/third_party/flashmap

Port the core of fmap into fwupd as libfwupdplugin/fu-fmap-firmware.[c|h].

I considered including chromiumos's libfmap as a library, but the library is not widely adopted by other distributions.

**Additional context**

bleungatchromium:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

This is required for the ongoing cros-ec work. This is a simple port of the core flashmap code here:
https://chromium.googlesource.com/chromiumos/third_party/flashmap/+/refs/heads/master/lib/

Resolves #2149


libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+}
+
+static void
+fu_fmap_firmware_class_init (FuFmapFirmwareClass *klass)
+{
+	FuFirmwareClass *klass_firmware = FU_FIRMWARE_CLASS (klass);
+	klass_firmware->parse = fu_fmap_firmware_parse;
+}
+
+FuFirmware *
```
superm1, line 274: This will need gtk doc documentation to indicate which version of fwupd it's landing in (1.5.0)

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+	guint32	offset;                /* offset relative to base */
+	guint32	size;                  /* size in bytes */
+	guint8	name[FMAP_STRLEN];     /* descriptive name */
+	guint16	flags;                 /* flags for this area */
+}  __attribute__((packed));
+
+struct fmap {
+	guint8	signature[8];		/* "__FMAP__" (0x5F5F464D41505F5F) */
+	guint8	ver_major;		/* major version */
+	guint8	ver_minor;		/* minor version */
```
superm1, line 34: I would think you also want to parse the major/minor too, no?  I didn't see it used at all.

bleungatchromium:
> A few generalized statements:
> 
> 1. Do you envision this to be used by multiple plugins (IE more than `cros-ec`)?  I would guess yes since you mentioned a few things in ChromeOS use it.  If so, then yes I agree it should go to libfwupdplugin like you have done.

Yes. I tested this with a coreboot (chrome os system firmware), so i envision this being used by a flashrom plugin capable of flashing a Chromebook's system firmware.

> 2. Presumably you're going to actually connect it to the cros-ec plugin for firmware verification too, right?  I would think that should go in the same PR (as another commit)

Yes, that one is still in progress. I'll undraft this once I add those commits.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Thanks! I was wondering why the CIs were upset.

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+
+/* mapping of volatile and static regions in firmware binary */
+struct fmap_area {
+	guint32	offset;                /* offset relative to base */
+	guint32	size;                  /* size in bytes */
+	guint8	name[FMAP_STRLEN];     /* descriptive name */
+	guint16	flags;                 /* flags for this area */
+}  __attribute__((packed));
+
+struct fmap {
```
hughsie, line 31: in fwupd a `typedef struct` is more usual, also typically called namespaced something like FuFmap or something.

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+		return FALSE;
+	}
+
+	for (i = 0; i < len - strlen(FMAP_SIGNATURE); i++)
+		if (!memcmp(&image[i],
+		            FMAP_SIGNATURE,
+		            strlen(FMAP_SIGNATURE))) {
+			fmap_found = TRUE;
+			break;
+		}
```
hughsie, line 75: braces here

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+
+	return count;
+}
+
+static gboolean
+fmap_find (const guint8 *image, gsize image_len, gsize *offset, GError **error)
+{
+	gboolean ret;
+
+	if ((image == NULL) || (image_len == 0)) {
```
hughsie, line 168: no need for the inner brackets

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+	if (fmap->nareas < 1) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_DATA,
+			     "number of areas too small, got %" G_GUINT16_FORMAT,
+			     fmap->nareas);
+		return FALSE;
+	}
+
+	for (i = 0; i < fmap->nareas; i++) {
```
hughsie, line 242: maybe more the i into the for loop?

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,278 @@
...
+
+struct _FuFmapFirmware {
+	FuFirmwareClass		 parent_instance;
+};
+
+G_DEFINE_TYPE (FuFmapFirmware, fu_fmap_firmware, FU_TYPE_FIRMWARE)
+
+#define FMAP_SIGNATURE		"__FMAP__"
+#define FMAP_VER_MAJOR		1	/* this header's FMAP minor version */
+#define FMAP_VER_MINOR		1	/* this header's FMAP minor version */
```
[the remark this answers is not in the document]
bleungatchromium: Fixed. Thanks.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done. This major/minor is the FMAP's version number, not the firmware payloads, so I attached this to the fmap's own image.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,139 @@
...
+{
+	FuCrosEcFirmware *self = FU_CROS_EC_FIRMWARE (firmware);
+	g_autoptr(FuFirmware) fmap_firmware = fu_fmap_firmware_new ();
+
+	self->sections[0].name = "RO";
+	self->sections[1].name = "RW";
+
+	if (!fu_firmware_parse (fmap_firmware, fw, flags, error)) {
+		return FALSE;
+	}
```
superm1, line 39: no need for the braces for one line statement

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,139 @@
...
+						       &self->version,
+						       error)) {
+				g_prefix_error (error,
+						"failed parsing firmware's version: %32s: ",
+						section.version);
+				return FALSE;
+			}
+			fu_firmware_set_version (firmware,
+						 self->version.triplet);
+
```
superm1, line 115: extra whitespace

plugins/cros-ec/fu-cros-ec-firmware.h
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#pragma once
+
+#include "fu-firmware.h"
+#include "fu-fmap-firmware.h"
+
+#define FU_TYPE_CROS_EC_FIRMWARE (fu_cros_ec_firmware_get_type ())
+G_DECLARE_FINAL_TYPE (FuCrosEcFirmware, fu_cros_ec_firmware, FU, CROS_EC_FIRMWARE, FuFirmware)
```
superm1, line 13: I sorta expected that this would actually derive from fmap firmware, shouldn't it?

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-firmware.h
bleungatchromium: Good catch. Done.

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,265 @@
...
+		return FALSE;
+	}
+
+	/*
+	 * For efficient operation, we start with the largest stride possible
+	 * and then decrease the stride on each iteration. Also, check for a
+	 * remainder when modding the offset with the previous stride. This
+	 * makes it so that each offset is only checked once.
+	 */
+	for (stride = len / 2; stride >= 1; stride /= 2) {
```
hughsie, line 94: maybe move `gint stride` into the for loop.

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,265 @@
...
+	if (fmap->size != image_len) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_DATA,
+			     "file size incorrect, expected 0x%04x got 0x%04x",
+			     (guint) fmap->size,
+			     (guint) image_len);
+		return FALSE;
+
+	}
```
hughsie, line 204: extra newline

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,265 @@
...
+		return FALSE;
+	}
+
+	if (i + fmap_size ((FuFmap *)&image[i]) > len) {
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_INVALID_DATA,
+				     "malformed fmap too close to end of image");
+
+		return FALSE;
```
hughsie, line 65: no extra newline

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,265 @@
...
+		ret = fmap_bsearch (image, image_len, offset, error);
+	else
+		ret = fmap_lsearch (image, image_len, offset, error);
+
+	if (!ret) {
+		g_prefix_error (error, "failed fmap_find: ");
+		return FALSE;
+	}
+
+	return TRUE;
```
hughsie, line 160: could we redo this in a nicer way? e.g. 

    if (popcnt (image_len) == 1) {
        if (!fmap_bsearch (image, image_len, offset, error)) {
            g_prefix_error (error, "failed fmap_find for cnt 1: ");
            return FALSE;
        }
    }

    if (!fmap_lsearch (image, image_len, offset, error)) {
        g_prefix_error (error, "failed fmap_find: ");
        return FALSE;
    }

    return TRUE;

Then the error message tells us which path was taken.

libfwupdplugin/fu-fmap-firmware.h
```diff
@@ -0,0 +1,43 @@
...
+	guint32		size;			/* size of firmware binary in bytes */
+	guint8		name[FMAP_STRLEN];	/* name of this firmware binary */
+	guint16		nareas;			/* number of areas described by
+						   areas[] below */
+	FuFmapArea	areas[];
+} FuFmap;
+
+
+
+FuFirmware			*fu_fmap_firmware_new			(void);
```
hughsie, line 43: too many newlines

libfwupdplugin/fu-fmap-firmware.h
```diff
@@ -0,0 +1,43 @@
...
+
+#define FMAP_SIGNATURE		"__FMAP__"
+#define FMAP_VER_MAJOR		1	/* this header's FMAP major version */
+#define FMAP_VER_MINOR		1	/* this header's FMAP minor version */
+#define FMAP_STRLEN		32	/* maximum length for strings, */
+					/* including null-terminator */
+#define FMAP_AREANAME		"FMAP"
+
+#define FU_TYPE_FMAP_FIRMWARE (fu_fmap_firmware_get_type ())
+G_DECLARE_FINAL_TYPE (FuFmapFirmware, fu_fmap_firmware, FU, FMAP_FIRMWARE, FuFirmware)
```
hughsie, line 19: FINAL is _okay_, but this means you can never create a subclass of FuFmapFirmware. If you think they'll ever be different types of FuFmapFirmware e.g. FuLegacyFmapFirmware or FuSuperFmapFirmware this should probably be DERIVED. If the latter is true, I can help you convert the various `self`'s to `priv`s if you want.

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,137 @@
...
+		fwid_bytes = fu_firmware_image_write(fwid_img, error);
+		if (fwid_bytes == NULL) {
+			g_prefix_error (error,
+					"unable to get bytes from %s: ",
+					fmap_fwid_name);
+			return FALSE;
+		}
+		memcpy (section.version,
+			g_bytes_get_data (fwid_bytes, NULL),
+			g_bytes_get_size (fwid_bytes));
```
hughsie, line 89: this makes me a bit twitchy. Maybe use `fu_memcpy_safe`?

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		fwid_img = fu_firmware_get_image_by_id (fmap_firmware,
+							fmap_fwid_name, error);
+		if (fwid_img == NULL) {
+			g_prefix_error (error, "%s image not found: ",
+					fmap_fwid_name);
+			return FALSE;
+		}
+		fwid_bytes = fu_firmware_image_write(fwid_img, error);
```
hughsie, line 80: space before bracket pls

plugins/cros-ec/fu-cros-ec-firmware.h
```diff
@@ -0,0 +1,41 @@
...
+ * Each RO or RW section of the new image can be in one of the following
+ * states.
+ */
+typedef enum {
+	not_needed = 0,  /* Version below or equal that on the target. */
+	not_possible,    /*
+			  * RO is newer, but can't be transferred due to
+			  * target RW shortcomings.
+			  */
+	needed           /*
```
hughsie, line 25: upper case enum names please

libfwupdplugin/fu-fmap-firmware.h
superm1: Well specifically CrosEC is a type of fmap firmware per my understanding.  So it shouldn't be final then.

bleungatchromium: Sounds good. Will fix.

bleungatchromium: The good news is that the fu-fmap-firmware was so simple at this point that it didn't use any `self`s. New version will use `G_DECLARE_DERIVABLE_TYPE ` instead.

plugins/cros-ec/fu-cros-ec-firmware.h
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done. Also added another change to replace the memcpy in fu-cros-ec-usb-device.c as well.

libfwupdplugin/fu-fmap-firmware.h
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.h
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,258 @@
...
+			g_prefix_error (error, "failed fmap_find using lsearch: ");
+			return FALSE;
+		}
+	}
+
+	return TRUE;
+}
+
+static gboolean
+fu_fmap_firmware_parse (FuFirmware *firmware,
```
superm1, line 158: Since this firmware can subclass, you need to call into potentially subclassed parse methods for it.  Example of how another plugin does this: https://github.com/fwupd/fwupd/blob/master/plugins/thunderbolt/fu-thunderbolt-firmware.c#L393

And yes that means in the subclassed one you don't call the superclassed parse (it's the other way around).
And from the cros-ec plugin you would just call `fu_firmware_parse` .  Another example here: https://github.com/fwupd/fwupd/blob/master/plugins/thunderbolt/fu-thunderbolt-device.c#L484

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -0,0 +1,258 @@
...
+	if (!fmap_found) {
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_INVALID_DATA,
+				     "fmap not found using binary search");
+		return FALSE;
+	}
+
+	if (i + fmap_size ((FuFmap *)&image[i]) > len)
+		return FALSE;
```
superm1, line 113: you're missing an error to be set here

libfwupdplugin/fu-fmap-firmware.h
```diff
@@ -0,0 +1,46 @@
...
+#pragma once
+
+#include "fu-firmware.h"
+
+#define FMAP_SIGNATURE		"__FMAP__"
+#define FMAP_VER_MAJOR		1	/* this header's FMAP major version */
+#define FMAP_VER_MINOR		1	/* this header's FMAP minor version */
+#define FMAP_STRLEN		32	/* maximum length for strings, */
+					/* including null-terminator */
+#define FMAP_AREANAME		"FMAP"
```
superm1, line 16: This define only appears to be used in one source file, I think you should just move the define there.

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,140 @@
...
+			   GError **error)
+{
+	FuCrosEcFirmware *self = FU_CROS_EC_FIRMWARE (firmware);
+	g_autoptr(FuFirmware) fmap_firmware = fu_fmap_firmware_new ();
+
+	self->sections[0].name = "RO";
+	self->sections[1].name = "RW";
+
+	if (!fu_firmware_parse (fmap_firmware, fw, flags, error))
+		return FALSE;
```
superm1, line 38: this specifically should go away when you do the subclass call

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,140 @@
...
+static gboolean
+fu_cros_ec_firmware_parse (FuFirmware *firmware,
+			   GBytes *fw,
+			   guint64 addr_start,
+			   guint64 addr_end,
+			   FwupdInstallFlags flags,
+			   GError **error)
+{
+	FuCrosEcFirmware *self = FU_CROS_EC_FIRMWARE (firmware);
+	g_autoptr(FuFirmware) fmap_firmware = fu_fmap_firmware_new ();
```
superm1, line 32: I expect this also goes away in the new call too.  If you need to refer the fmap firmware it should be something like `FU_FMAP_FIRMWARE ()`

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,140 @@
...
+fu_cros_ec_firmware_class_init (FuCrosEcFirmwareClass *klass)
+{
+	FuFirmwareClass *klass_firmware = FU_FIRMWARE_CLASS (klass);
+	klass_firmware->parse = fu_cros_ec_firmware_parse;
+}
+
+FuFirmware *
+fu_cros_ec_firmware_new (void)
+{
+	return FU_FIRMWARE (g_object_new (FU_TYPE_CROS_EC_FIRMWARE, NULL));
```
superm1, line 139: do you actually need the `FU_FIRMWARE (` macro?  I think just `g_object_new` is sufficient

libfwupdplugin/fu-fmap-firmware.c
hughsie: I think the way it's done here is fine for the moment IMHO. If/when we get a subclassed version we can rework how the parse is handled.

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -0,0 +1,140 @@
...
+#include "config.h"
+
+#include "fu-fmap-firmware.h"
+#include "fu-cros-ec-common.h"
+#include "fu-cros-ec-firmware.h"
+
+#define MAXSECTIONS	2
+
+struct _FuCrosEcFirmware {
+	FuFirmware			parent_instance;
```
hughsie, line 16: I think this needs to be `FuFmapFirmware parent_instance`, no?

plugins/cros-ec/fu-cros-ec-firmware.c
hughsie: Right. If `FuCrosEcFirmware` is a subclass of `FuFmapFirmware` then you can just cast it to `FU_FIRMWARE()` without creating a new object.

plugins/cros-ec/fu-cros-ec-firmware.c
hughsie: Agree.

plugins/cros-ec/fu-cros-ec-firmware.c
hughsie: I don't think it's required, but I also don't have a problem with it being there. If `FU_TYPE_CROS_EC_FIRMWARE` isn't squeezable down into a FuFirmware we have bigger problems. :)

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.h
bleungatchromium: Dnoe.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Done.

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Thanks for the guidance. I've gone ahead and reworked this so that the subclass's parse just does the extra step of figuring out which EC specific image contains the section version strings and attaches them to the corresponding images. Much cleaner. Thanks!

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done. Removed.

libfwupdplugin/fu-fmap-firmware.h
```diff
@@ -0,0 +1,48 @@
...
+struct _FuFmapFirmwareClass
+{
+	FuFirmwareClass		 parent_class;
+	gboolean		 (*parse)		(FuFirmware	*self,
+							 GBytes		*fw,
+							 guint64	 addr_start,
+							 guint64	 addr_end,
+							 FwupdInstallFlags flags,
+							 GError		**error);
+};
```
hughsie, line 26: could you add:

    /*< private >*/
    gpointer	padding[14];

...here please -- then we can add more vfuncs in the future without increasing the size of the struct and breaking ABI.

bleungatchromium: Sounds good. Will do.

Issue #2179, opened 2020-06-15
fwupd does not cleanup files in /boot/efi


xclaesse:
My /boot/efi partition is only 50M because before fwupd existed that partition only contained a few text files. Fortunately a firmware update seems to take only 17M in `/boot/efi/EFI/ubuntu/fw` so it can fit and the update worked. But when I got a 2nd update it seems previous files where not deleted and downloading a 2nd firmware exceeds the space available on that partition. gnome-software gives an error but with no way or clue how to fix it. Manually deleting `/boot/efi/EFI/ubuntu/fw` folder fixed the error and the firware update succeeded.

I'm not sure who should be responsible for deleting old files, fwupd or gnome-software?

superm1:
What version of fwupd do you have installed?

The old files are supposed to be cleaned up when a new update is run:
https://github.com/fwupd/fwupd/blob/master/plugins/uefi/fu-uefi-device.c#L449

xclaesse:
1.3.9-4ubuntu0.1 on Ubuntu 20.04.

xclaesse:
When is that function supposed to be called? Before downloading an update, or after having installed the update? Because those files are still there on my system:
```
ls /boot/efi/EFI/ubuntu/fw
fwupd-39f17d37-5d18-4e72-934c-226ba948743e.cap
fwupd-3b8c8162-188c-46a4-aec9-be43f1d65697.cap
```

superm1:
It's _supposed_ to get called next time you try to run an update.  The general flow is like this:

1. Download CAB update to home directory
2. Extract update to temporary directory.
3. Clean up old updates on ESP
4. Copy CAP file to the ESP.
5. Schedule for next reboot
6. Reboot machine
7. Install update
8. Boot back into OS

Are you saying that you're having a problem with that flow because of the small ESP? If so we might have a bug still.
If you can please turn on verbose logging in the daemon and show the output when trying to install the second update that would be appreciated.

xclaesse:
Those steps make sense, as long as steps 3 and 4 does not get inverted by mistake? 

Unfortunately the update is done now because I deleted files manually. I'll only be able to give logs when/if lenovo release a new firmware for my x395.

xclaesse:
@superm1 is there a debug tool or something I can use to simulare running step 3?

xclaesse:
ooohhhh, reading fu_uefi_device_prepare() it checks free space before cleanup.

superm1:
[...]

PR #2180, opened 2020-06-15
uefi: check for free space after cleaning up ESP

superm1:
In a very small ESP situation it's possible that the amount of free
space is insufficient until it's actually been cleaned.

Fixes: #2179

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


xclaesse:
Thanks for the quick fix, I'm not a fwupd dev but that patch looks good to me.

PR #2183, opened 2020-06-15
tpm-eventlog: fix PCR0 calculation

superm1:
One of the core problems is that systems with both sha1 and sha256
were miscalculating.

Fixes: #2181

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Well I missed it too during code review 😜 
So maybe our number of "real" PCR0 calculation problems is a lot smaller than we thought.

Side note: I wonder if we really even should still have a standalone tool anymore.  `tpm2-tools` provides `tpm2_eventlog` which does basically the same thing..

PR #2186, opened 2020-06-16
Wip/hughsie/review metadata

hughsie:


libfwupdplugin/fu-device.c
```diff
@@ -2972,6 +2972,63 @@ fu_device_probe_invalidate (FuDevice *self)
...
+	FuDeviceClass *klass = FU_DEVICE_GET_CLASS (self);
+	g_autoptr(GHashTable) metadata = NULL;
+
+	g_return_val_if_fail (FU_IS_DEVICE (self), NULL);
+
+	/* metadata for all devices */
+	metadata = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+	g_hash_table_insert (metadata,
+			     g_strdup ("GType"),
+			     g_strdup (G_OBJECT_TYPE_NAME (self)));
```
superm1, line 2997: Is this actually valuable for all the devices?

hughsie: For things like VLI or Wacom that are subclassed quite deeply, yes. For others perhaps less so.

libfwupdplugin/fu-device.c
```diff
@@ -2972,6 +2972,63 @@ fu_device_probe_invalidate (FuDevice *self)
...
+
+	g_return_val_if_fail (FU_IS_DEVICE (self), NULL);
+
+	/* metadata for all devices */
+	metadata = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+	/* optionally subclassed */
+	if (klass->report_metadata_post != NULL)
+		klass->report_metadata_post (self, metadata);
+	return g_steal_pointer (&metadata);
```
superm1, line 3029: Most device types won't have this new vfunc, I wonder if it's really worth the extra memory to create the hash table and return an empty hash table in that common case.  Instead why not

```
if (klass->report_metadata_post == NULL)
        return NULL;
metadata = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
klass->report_metadata_post (self, metadata);
return g_steal_pointer (&metadata);
```

libfwupdplugin/fu-device.c
superm1: But so in those it actually matters, shouldn't they be device metadata made by the plugin?

libfwupdplugin/fu-device.c
hughsie: Sure, but this isn't fast path at all, so I went for clarity. It's called just once at startup if it's the first boot after a UEFI update.

libfwupdplugin/fu-device.c
hughsie: Sure, I can do that if you think it's clearer.

superm1: Yes I do think it's clearer.

libfwupdplugin/fu-device.c
hughsie: I guess if `GType` isn't always added then my arguement is null and void. Will fix.

libfwupdplugin/fu-device.c
hughsie: Okay, fixed.

hughsie:
> to adjust too to be more device centric (tpm-eventlog

That one is a bit special as we add it as a dep to the UEFI devices.

> dell-dock

Would this need a vfunc implemented for all subclassed FuDevices? I don't know if it's just needed for EC etc.

superm1:
> Would this need a vfunc implemented for all subclassed FuDevices? I don't know if it's just needed for EC etc.

It's relevant information for all of them yeah.  So maybe the way it's done now is fine.

PR #2187, opened 2020-06-16
Wip/hughsie/msp

hughsie:


libfwupdplugin/fu-device.c
```diff
@@ -770,6 +770,17 @@ fu_device_add_child (FuDevice *self, FuDevice *child)
 	}
 	g_ptr_array_add (priv->children, g_object_ref (child));
 
+	/* ensure the parent has the MAX() of the childrens removal delay  */
+	for (guint i = 0; i < priv->children->len; i++) {
+		FuDevice *child_tmp = g_ptr_array_index (priv->children, i);
+		guint remove_delay = fu_device_get_remove_delay (child_tmp);
+		if (remove_delay > priv->remove_delay) {
```
superm1, line 777: is it possible that a parent would have a remove delay of 0 and this inadvertently sets one?

hughsie: I think it would deliberately set one. If the physical child needs a remove delay then the parent has to inherit it. I don't think a remove delay of 0 makes a lot of sense, I think we even have a daemon warning if a FuDevice subclass forgets to set it...

superm1: Ack OK.

PR #2189, opened 2020-06-17
TPM event log fixups

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
```diff
@@ -115,19 +115,25 @@ fu_plugin_device_registered_uefi (FuPlugin *plugin, FuDevice *device)
...
 			if (g_strcmp0 (checksum, checksum_tmp) == 0) {
 				data->reconstructed = TRUE;
-				return;
+				break;
 			}
 		}
+		/* check at least one reconstruction for this algorithm */
+		if (!data->reconstructed) {
+			fu_device_set_update_message (device,
+						     "TPM PCR0 %s differs from reconstruction, "
```
hughsie, line 134: I think you need to `g_strdup_printf` if you're using %s here.

plugins/tpm-eventlog/fu-tpm-eventlog.c
```diff
@@ -9,6 +9,7 @@
 #include "config.h"
 
 #include <fwupd.h>
+#include "fwupd-common-private.h"
```
hughsie, line 12: maybe move this down a bit after the <> defines?

plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
superm1: I think that %s was a typo when I moved it.  will fix

hughsie:
LGTM, modulo a few niggles.

PR #2191, opened 2020-06-17
dell-dock: Add more module types to the enum

superm1:
Unfortunately module type has more than I previously realized.  This wasn't caught because of the large number of options that can be plugged into the dock.

The meanings that previously were applied fortunately worked for
the most important case (130-180W TBT) but didn't for single C, dual
C or small power (45W) cases.

Since composite_prepare was trying to read and interpret these, it
causes failures when these other ones are encountered.

I reproduced this on a 130W adapter plugged into a single C (type 0x4).
This meant the update wouldn't install since NULL was returned for the
type.

In case a new module ID is added later, also return an "unknown" for
the metadata.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-dell-dock-i2c-ec.c
+++ b/plugins/dell-dock/fu-dell-dock-i2c-ec.c
@@ -100,9 +100,13 @@ typedef struct __attribute__ ((packed)) {
 } FuDellDockEcQueryEntry;
 
 typedef enum {
-	MODULE_TYPE_SINGLE = 1,
-	MODULE_TYPE_DUAL,
-	MODULE_TYPE_TBT,
+	MODULE_TYPE_45_TBT = 1,
+	MODULE_TYPE_45,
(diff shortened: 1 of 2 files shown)


superm1:
@hughsie I know it's really soon since you just tagged 1.3.10, but could you tag a 1.3.11 now?  I cherry-picked this and the PCR0 thing.
I have the attention in Ubuntu to get 1.3.x updated to a new tagged release, and so it makes more sense to include this and the PCR0 fix to prevent large swaths of people to hit these two issues.

hughsie:
@superm1 sure, I can do this today if that helps. Before I do, is there anything else we should backport?

superm1:
Thanks.  I had a look and found 2 other unrelated things that made sense and backported.  Nothing else from my side!

hughsie:
Release done!

PR #2202, opened 2020-06-22
ccgx: add new protocol for dmc dock

chlee75:
Type of pull request:
- [ ] Code fix: add new protocol for dmc dock to support quad firmware version format


chlee75:
Add new protocol "com.cypress.ccgx.dmc" for quad firmware version format.

PR #2204, opened 2020-06-22
Add support for a delayed activation flow for Thunderbolt

superm1:
This allows delaying the activation of Thunderbolt firmware until
shutdown/reboot or when the dock is unplugged.

This functionality requires features in the kernel:
https://lore.kernel.org/linux-usb/20200622143035.25327-1-mario.limonciello@dell.com/T/#t

Matrix of cases to support:

* Distro Old Linux kernel (doesn't support authenticate on disconnect)

  - WD19TB: Should have `skips-restart` flag set
    No flush or activate features called in `thunderbolt` plugin.
    `dell_dock` plugin will activate at end of composite update

  - All other devices: Shouldn't have flags set
    Should authenticate in Thunderbolt plugin.
    `1 > nvm_authenticate`

* Distro New Linux kernel (supports authenticate on disconnect)

  - WD19TB: Should have `usable-during-update` flag set but not `skips-restart`
    Should flush image to SPI in `thunderbolt` plugin
    `2 > nvm_authenticate_on_disconnect`
    Should configure TBT device for authenticate on disconnect
    `1 > nvm_authenticate_on_disconnect`
    `dell_dock` plugin will configure dock for authenticate on disconnect

  - All other devices: Shouldn't have flags set
    Should authenticate in `thunderbolt` plugin.
    `1 > nvm_authenticate`

* ChromeOS (supports authenticate on disconnect)

  - `thunerbolt.conf` will have `DelayedActivation=true`.

  - WD19TB: Should have `usable-during-update` flag set but not `skips-restart`
    Should flush image to SPI in `thunderbolt` plugin
    `2 > nvm_authenticate_on_disconnect`
    Should configure device for authenticate on disconnect
    `1 > nvm_authenticate_on_disconnect`
    `dell_dock` plugin will configure dock for authenticate on disconnect

  - All other devices: Should have both `usable-during-update` and `skips-restart` set
    Should flush image to SPI in `thunderbolt` plugin
    `2 > nvm_authenticate`
    Will activate upon logout/shutdown/reboot
    `1 > nvm_authenticate`

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -185,10 +185,28 @@ fu_plugin_composite_cleanup (FuPlugin *plugin,
 			     GError **error)
 {
 	FuDevice *parent = fu_plugin_dell_dock_get_ec (devices);
+	FuDevice *dev = NULL;
 	g_autoptr(FuDeviceLocker) locker = NULL;
+	gboolean needs_activation = FALSE;
 
 	if (parent == NULL)
 		return TRUE;
+
+	/* if thunderbolt is in the transaction it needs to be activated separately */
+	for (guint i = 0; i < devices->len; i++) {
+		dev = g_ptr_array_index (devices, i);
+		if (g_strcmp0 (fu_device_get_plugin (dev), "thunderbolt") == 0 &&
+		    fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
+			/* the kernel and/or thunderbolt plugin have been configured to let HW finish the update */
+			if (fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
+				fu_dell_dock_ec_tbt_passive (parent);
+			/* run the update immediately - no kernel support */
+			} else {
+				needs_activation = TRUE;
+				break;
+			}
+		}
+	}
 
 	locker = fu_device_locker_new (parent, error);
 	if (locker == NULL)
@@ -201,15 +219,9 @@ fu_plugin_composite_cleanup (FuPlugin *plugin,
 	if (!fu_device_locker_close (locker, error))
 		return FALSE;
 
-	/* if thunderbolt is in the transaction it needs to be activated separately */
-	for (guint i = 0; i < devices->len; i++) {
-		FuDevice *dev = g_ptr_array_index (devices, i);
-		if (g_strcmp0 (fu_device_get_plugin (dev), "thunderbolt") == 0 &&
-		    fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
-			if (!fu_device_activate (dev, error))
-				return FALSE;
-			break;
-		}
+	if (needs_activation && dev != NULL) {
+		if (!fu_device_activate (dev, error))
+			return FALSE;
 	}
 
 	return TRUE;
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -28,6 +28,7 @@ struct _FuThunderboltDevice {
 	gboolean		 is_native;
 	guint16			 gen;
 	gchar			*devpath;
+	const gchar		*auth_method;
 };
 
 #define TBT_NVM_RETRY_TIMEOUT				200	/* ms */
@@ -178,6 +179,7 @@ fu_thunderbolt_device_to_string (FuDevice *device, guint idt, GString *str)
 	fu_common_string_append_kb (str, idt, "Safe Mode", self->safe_mode);
 	fu_common_string_append_kb (str, idt, "Native mode", self->is_native);
 	fu_common_string_append_ku (str, idt, "Generation", self->gen);
+	fu_common_string_append_kv (str, idt, "AuthAttribute", self->auth_method);
 }
 
 static gboolean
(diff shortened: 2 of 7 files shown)


superm1:
CC @djcampello

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -178,6 +179,7 @@ fu_thunderbolt_device_to_string (FuDevice *device, guint idt, GString *str)
 	fu_common_string_append_kb (str, idt, "Safe Mode", self->safe_mode);
 	fu_common_string_append_kb (str, idt, "Native mode", self->is_native);
 	fu_common_string_append_ku (str, idt, "Generation", self->gen);
+	fu_common_string_append_kv (str, idt, "AuthAttribute", self->authenticate_attribute);
```
gicmo, line 182: I stumbled over this when I first read it but I cannot think how to better name that is now horrendously long. I was thinking maybe `AuthMethod`/`AuthMechanism`? Maybe just good as is.

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -201,15 +217,9 @@ fu_plugin_composite_cleanup (FuPlugin *plugin,
...
-	/* if thunderbolt is in the transaction it needs to be activated separately */
-	for (guint i = 0; i < devices->len; i++) {
-		FuDevice *dev = g_ptr_array_index (devices, i);
-		if (g_strcmp0 (fu_device_get_plugin (dev), "thunderbolt") == 0 &&
-		    fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
-			if (!fu_device_activate (dev, error))
-				return FALSE;
-			break;
-		}
+	if (needs_activation && dev != NULL) {
```
gicmo, line 220: I dont think `needs_activation` can be non-`NULL` but `dev` be `NULL` but doesn't hurt.

superm1: `needs_activation` is boolean

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: thanks fixed

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -185,11 +185,27 @@ fu_plugin_composite_cleanup (FuPlugin *plugin,
...
 	if (parent == NULL)
 		return TRUE;
 
+	/* if thunderbolt is in the transaction it needs to be activated separately */
+	for (guint i = 0; i < devices->len; i++) {
+		dev = g_ptr_array_index (devices, i);
+		if (g_strcmp0 (fu_device_get_plugin (dev), "thunderbolt") == 0 &&
+		    fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
+			if (fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
+				fu_dell_dock_ec_tbt_passive (parent);
```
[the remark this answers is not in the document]
superm1: added a comment

plugins/dell-dock/fu-plugin-dell-dock.c
gicmo: Ah yes, sorry, I mean, `needs_activation` can not be `TRUE` and  `dev` be `NULL` at the same time. But being defensive is probably a better idea anyway!

superm1: Yeah it's mostly to help prevent if some other code changes this logic in the future and there is a situation that could occur. Thanks!

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -321,36 +323,45 @@ fu_thunderbolt_device_setup (FuDevice *device, GError **error)
...
+		self->auth_method = "nvm_authenticate";
+	}
+
 	/* success */
 	return TRUE;
 }
 
 static gboolean
-fu_thunderbolt_device_trigger_update (FuDevice *device, GError **error)
+_write_simple_sysfs (gchar *path, const gchar* val, GError **error)
```
hughsie, line 344: could we split this out in FuUdevDevice?

plugins/thunderbolt/thunderbolt.conf
```diff
@@ -4,3 +4,6 @@
 # It's important that all backports from this kernel have been
 # made if using an older kernel
 MinimumKernelVersion=4.13.0
+
+# Forces delaying activation until shutdown/logout/reboot
+DelayedActivation=false
```
hughsie, line 9: why a config option?

superm1: ChromeOS will want to turn this on for everything with a guarantee that they always run activation on logout/shutdown/reboot.  I didn't think this made sense for generically doing in all distros.  

If we want to shift fwupd to behave this way, we can drop the config option (and some logic).  See the matrix of situations to support in the commit message for background.

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: You mean make a "write simple sysfs" method for FuUdevDevice that we call?

superm1: I just pushed what I think you mean.

superm1:
V2 of kernel patches needed for this: https://lore.kernel.org/linux-usb/20200622185758.28145-1-mario.limonciello@dell.com/T/#t

PR #2206, opened 2020-06-22
trivial: add Vendor ID into Modem manager devices

superm1:
I don't actually know for sure this will fix it sufficiently or it's the correct vendor ID in this instance, so hopefully @wicadmin or @aleksander0m  you can double check it.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/modem-manager.quirk
+++ b/plugins/modem-manager/modem-manager.quirk
@@ -3,38 +3,46 @@
 [DeviceInstanceId=USB\VID_413C&PID_81D7]
 Summary = Dell DW5821e LTE modem
 CounterpartGuid = USB\VID_413C&PID_81D6
+VendorId = USB:0x413C
 
 # DW5821e in fastboot mode
 [DeviceInstanceId=USB\VID_413C&PID_81D6]
 Summary = Dell DW5821e LTE modem (fastboot)
 CounterpartGuid = USB\VID_413C&PID_81D7
+VendorId = USB:0x413C
 
 # DW5821e/eSIM
 [DeviceInstanceId=USB\VID_413C&PID_81E0]
 Summary = Dell DW5821e/eSIM LTE modem
 CounterpartGuid = USB\VID_413C&PID_81E1
+VendorId = USB:0x413C
 
 # DW5821e/eSIM in fastboot mode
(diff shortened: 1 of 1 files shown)


plugins/modem-manager/modem-manager.quirk
```diff
@@ -3,38 +3,46 @@
...
 [DeviceInstanceId=USB\VID_0489&PID_E0B7]
 Summary = Foxconn T77w968 LTE modem (fastboot)
 CounterpartGuid = USB\VID_0489&PID_E0B4
+VendorId = USB:0x0489
 
 # T77W968/eSIM
 [DeviceInstanceId=USB\VID_0489&PID_E0B5]
 Summary = Foxconn T77w968/eSIM LTE modem
 CounterpartGuid = USB\VID_0489&PID_E0B8
+VendorId = USB:0x0489
```
hughsie, line 42: don't we get it from the USB vendor automatically?

superm1: No, see the reported issue linked to this.  `FuMmDevice` derives from `FuDevice`, not `FuUdevDevice`

hughsie: I think in this case I think I'd rather set the instance ID from the constructed VID and then only override them in the quirk files if there are different to the `VID_` entry. i.e. we can get the IDs from `mm_firmware_update_settings_get_device_ids` -- so maybe we can get the vendor in the same way.

superm1: I don't see another method that gets just the device IDs.  So it's either:
1. do surgery on the string
2. try to read it another way.
3. Set up `fu_udev_set_vendor` do use some heuristics and set id if unset (which would mean audit everything and make sure that id gets set before vendor string).

hughsie: > do surgery on the string

Eww. MM returns multiple IDs for each device, so I'd rather not map the instance-id to the vendor_id in any way. @aleksander0m can we get a method to get the USB vendor ID please? For the meantime adding the quirks is fine, but long term doesn't really scale IMHO.

superm1:
OK will merge for now and leave an issue open for us to track that change to add a new interface. https://github.com/fwupd/fwupd/issues/2209

PR #2207, opened 2020-06-22
Wip/hughsie/leak

hughsie:


libfwupdplugin/fu-plugin.c
```diff
@@ -1650,6 +1650,8 @@ void
 fu_plugin_add_udev_subsystem (FuPlugin *self, const gchar *subsystem)
 {
 	FuPluginPrivate *priv = GET_PRIVATE (self);
+	if (priv->udev_subsystems == NULL)
+		priv->udev_subsystems = g_ptr_array_new_with_free_func (g_free);
```
superm1, line 1654: I'm not convinced this will actually help anything.  the engine calls `fu_plugin_set_udev_subsystems` when plugins get loaded which will still initialize it.

libfwupdplugin/fu-device.c
```diff
@@ -3228,9 +3240,10 @@ fu_device_finalize (GObject *object)
 		g_object_unref (priv->quirks);
 	if (priv->poll_id != 0)
 		g_source_remove (priv->poll_id);
+	if (priv->metadata != NULL)
```
superm1, line 3243: this looks like you actually caught a bug with this change too!  It shouldn't have cleared it when the mutex was released.

libfwupdplugin/fu-device.c
```diff
@@ -3228,9 +3240,10 @@ fu_device_finalize (GObject *object)
...
+		g_hash_table_unref (priv->metadata);
 	g_rw_lock_clear (&priv->metadata_mutex);
 	g_rw_lock_clear (&priv->parent_guids_mutex);
-	g_hash_table_unref (priv->metadata);
 	g_ptr_array_unref (priv->children);
 	g_ptr_array_unref (priv->parent_guids);
```
superm1, line 3248: Perhaps this should also move up above the mutex release too?

libfwupdplugin/fu-plugin.c
hughsie: I think it's for the tests that just use FuPlugin and do not create a FuEngine.

libfwupdplugin/fu-device.c
hughsie: 100% unintentional :)

libfwupdplugin/fu-plugin.c
superm1: I don't think so?  I see the engine calling it when you do `fu_engine_load_plugins` which is part of `fu_engine_load`, should run in the main case

libfwupdplugin/fu-device.c
superm1: Since it's the case for both of the mutex use, maybe double check that it wasn't already right. `g_rw_lock_clear` releases the mutex, but if it wasn't released will trying to clear cause a deadlock?  or did this function take the mutex at the start?

libfwupdplugin/fu-plugin.c
hughsie: Right, but in the self tests case we don't have an engine instance:

    #0  0x00007ffff7f74e5f in fu_plugin_add_udev_subsystem (self=0x43c160, subsystem=0x7ffff7f3c08f "drm") at ../libfwupdplugin/fu-plugin.c:1655
    #1  0x00007ffff7f370e5 in fu_plugin_init (plugin=0x43c160) at ../plugins/synaptics-mst/fu-plugin-synaptics-mst.c:172
    #2  0x00007ffff7f72b6e in fu_plugin_open (self=0x43c160, filename=0x43d050 "/home/hughsie/Code/fwupd/build/plugins/synaptics-mst/libfu_plugin_synaptics_mst.so", error=0x7fffffffd8a0)
        at ../libfwupdplugin/fu-plugin.c:448
    #3  0x0000000000403d22 in fu_plugin_synaptics_mst_none_func () at ../plugins/synaptics-mst/fu-self-test.c:67
    #4  0x00007ffff7c372be in g_test_run_suite_internal () at /lib64/libglib-2.0.so.0
    #5  0x00007ffff7c370bb in g_test_run_suite_internal () at /lib64/libglib-2.0.so.0
    #6  0x00007ffff7c370bb in g_test_run_suite_internal () at /lib64/libglib-2.0.so.0
    #7  0x00007ffff7c377aa in g_test_run_suite () at /lib64/libglib-2.0.so.0
    #8  0x00007ffff7c377c5 in g_test_run () at /lib64/libglib-2.0.so.0
    #9  0x0000000000404301 in main (argc=1, argv=0x7fffffffdd68) at ../plugins/synaptics-mst/fu-self-test.c:131

libfwupdplugin/fu-device.c
hughsie: > but if it wasn't released will trying to clear cause a deadlock

It's apparently _undefined_. I'll do some review.

libfwupdplugin/fu-device.c
hughsie: I think we should clear the mutex before we finalize the thing it protects.

superm1: OK.

superm1: I guess the other one needs to be inverted then again too.

hughsie: Fixed up.

PR #2211, opened 2020-06-23
trivial: modem-manager: add a protocol for updating the devices

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -243,6 +243,18 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 	fu_device_set_version (device, version);
 	for (guint i = 0; device_ids[i] != NULL; i++)
 		fu_device_add_instance_id (device, device_ids[i]);
+	if (fu_device_get_vendor_id (device) == NULL) {
+		g_autofree gchar *path = g_build_filename (device_sysfs_path, "idVendor", NULL);
+		g_autofree gchar *value = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		if (!g_file_get_contents (path, &value, NULL, &error_local)) {
+			g_warning ("failed to set vendor ID: %s", error_local->message);
+		} else {
+			g_autofree gchar *vendor_id = g_strdup_printf ("USB:0x%s", value);
+			fu_device_set_vendor_id (device, vendor_id);
+		}
+	}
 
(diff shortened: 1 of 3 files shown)


hughsie:
Thanks for working this one out dude.

aleksander0m:
The vendor ID retrieval from sysfs attributes should be fine, so far we don't have non-USB modems with firmware upgrade supported, so should be enough. Whenever we add support for non-USB modems, it may make sense to have the vendor ID provided by ModemManager itself, if there is some other way to retrieve that.

And regarding the protocol reported by the plugin, right now the upgrade support works for a device that requires fastboot+QMI.PDC, both together, but the "main" one is fastboot, so it's fine also.

All good! Thanks for looking into this

hughsie:
> it may make sense to have the vendor ID provided by ModemManager itself

Agree. Yell if this API gets added and we'll add some `#ifdef`s.

aleksander0m:
@hughsie  I will, here's the MM issue report for context: https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/issues/229

plugins/modem-manager/fu-mm-device.c
```diff
@@ -189,6 +190,9 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 				break;
 			}
 		}
+		/* only set if fastboot wasn't already set */
+		if (fu_device_get_protocol (device) != NULL)
+			fu_device_set_protocol (device, "com.qualcomm.qmi_pdc");
```
qhill, line 195: Is this correct? On a naive reading it seems the code doesn't seem to match the comment.

superm1: It's due to the order that the function is populated. If fastboot was found higher up it will set it to fastboot. If that wasn't set the protocol should be NULL and if qmi is found it gets set.

qhill: I assumed fu_device_get_protocol (device) returning NULL when it is not yet set to fastboot, and non-NULL if it was set to fastboot.  The code sets qmi in the latter (non-NULL) case. (i.e. I expected the condition to read fu_device_get_protocol (device) == NULL.)

superm1: Oh, you're right, great catch! I'll push a follow up fix.

PR #2215, opened 2020-06-23
Use libxmlb bound parameters to speed up te device verification

hughsie:
This means we only parse the complicated xpath query once, rather than for
every GUID the device has.

The code flow is also simplified and split out into two functions.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -785,6 +785,115 @@ fu_engine_get_component_by_guids (FuEngine *self, FuDevice *device)
 	return NULL;
 }
 
+static XbNode *
+fu_engine_verify_from_local_metadata (FuEngine *self,
+				      FuDevice *device,
+				      GError **error)
+{
+	g_autofree gchar *fn = NULL;
+	g_autofree gchar *localstatedir = NULL;
+	g_autofree gchar *xpath = NULL;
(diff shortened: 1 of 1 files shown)


src/fu-engine.c
```diff
@@ -785,6 +785,106 @@ fu_engine_get_component_by_guids (FuEngine *self, FuDevice *device)
...
+			     "failed to find %s", fn);
+		return NULL;
+	}
+
+	if (!xb_builder_source_load_file (source, file,
+					  XB_BUILDER_SOURCE_FLAG_NONE,
+					  NULL, error))
+		return NULL;
+	xb_builder_import_source (builder, source);
+	silo = xb_builder_compile (builder,
```
superm1, line 819: Isn't this a memory leak?

superm1: (specifically because of the call to `xb_silo_new` earlier on with this same pointer)

hughsie: Yes, and it was before! Will fix in a smarter way, good catch.

hughsie: `g_object_set_data_full` to the rescue. We can't unconditionally ref the silo in the node else we get a dep loop and just leak tons of memory.

Issue #2212, opened 2020-06-23
Dell WD19TB failing to update


bulletmark:
**Describe the bug**
Trying to update Dell dock WD19TB but the update always stays "Updates will complete ..".

**Steps to Reproduce**
1. sudo fwupdmgr enable-remote lvfs-testing
2. sudo fwupdmgr refresh
3. sudo fwupdmgr get-devices
4. sudo fwupdmgr get-updates
5. sudo fwupdmgr update

Unplugged laptop from dock and waited 15 mins. Also, power recycled the dock later.

**Expected behavior**
Expected firmware to update.

**fwupd version information**
```shell
lt:~ sudo fwupdmgr --version
client version:	1.4.4
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.4.4
```
Installed by pacman on Arch Linux.

**fwupd device information**
```shell
$ fwupdmgr get-devices --filter=~internal
lt:~ sudo fwupdmgr get-devices --filter=~internal
XPS 13 9300
│
├─WD19TB:
│ │   Device ID:           570867f4ddda9915445da59bd85cd0e0e507270e
│ │   Summary:             High performance dock
│ │   Current version:     01.00.00.04
│ │   Minimum Version:     01.00.00.00
│ │   Vendor:              Dell Inc. (USB:0x413C)
│ │   Install Duration:    1 minute
│ │   Serial Number:       4DN5523/3159128927644594
│ │   Update Message:      Updates will complete after the dock is unplugged. The dock will be unavailable for approximately 1 minute.
│ │   GUID:                cd357cf1-40b2-5d87-b8df-bb2dd82774aa ← USB\VID_413C&PID_B06E&hub&embedded
│ │   Device Flags:        • Updatable
│ │                        • Requires AC power
│ │                        • Supported on remote server
│ │                        • Device stages updates
│ │                        • Device can recover flash failures
│ │                        • Device is usable for the duration of the update
│ │ 
│ ├─Thunderbolt controller in Dell dock:
│ │     Device ID:         5e8cd9f97ee07a0af3a5e0c70fffb713aa32362e
│ │     Summary:           Thunderbolt controller
│ │     Current version:   43.00
│ │     Vendor:            Dell Inc. (TBT:0x00D4)
│ │     Install Duration:  22 seconds
│ │     GUID:              c94770ca-1773-592c-b20a-e87243bc7cd0 ← TBT-00d4b070
│ │     Device Flags:      • Updatable
[...]

PR #2218, opened 2020-06-23
fu-engine: only set update message from metadata on upgrades

superm1:
After a release finishes installing, this isn't useful metadata.

Fixes: #2212

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1592,6 +1592,7 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
 	GString *str = g_string_new (NULL);
 	gboolean low_help = FALSE;
 	gboolean runtime_help = FALSE;
+	gboolean pcr0_help = FALSE;
 
 	for (guint j = 1; j <= FWUPD_SECURITY_ATTR_LEVEL_LAST; j++) {
 		gboolean has_header = FALSE;
@@ -1608,6 +1609,12 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
 			if (j < FWUPD_SECURITY_ATTR_LEVEL_IMPORTANT &&
 			    !fwupd_security_attr_has_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS))
 				low_help = TRUE;
+
+			/* check for PCR0 not matching */
+			if (g_strcmp0 (fwupd_security_attr_get_appstream_id (attr),
+					FWUPD_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0) == 0 &&
+			    fwupd_security_attr_get_result (attr) == FWUPD_SECURITY_ATTR_RESULT_NOT_VALID)
+				    pcr0_help = TRUE;
 		}
 	}
 	for (guint i = 0; i < attrs->len; i++) {
@@ -1643,6 +1650,14 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
 					/* TRANSLATORS: this is instructions on how to improve the HSI suffix */
 					_("This system has HSI runtime issues."),
 					"https://github.com/fwupd/fwupd/wiki/Host-security-ID-runtime-issues");
+	}
+
+	if (pcr0_help) {
+		g_string_append_printf (str, "\n%s\n » %s\n",
+					/* TRANSLATORS: this is more background on a security measurement problem */
+					_("The TPM PCR0 differes from reconstruction."),
+					"https://github.com/fwupd/fwupd/wiki/TPM-PCR0-differs-from-reconstruction");
+
 	}
 
 	return g_string_free (str, FALSE);
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1210,16 +1210,19 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 		/* TRANSLATORS: hardware state, e.g. "pending" */
 		fu_common_string_append_kv (str, idt + 1, _("Update State"),
 					    fwupd_update_state_to_string (state));
+
+		if (state == FWUPD_UPDATE_STATE_SUCCESS) {
+			tmp = fwupd_device_get_update_message (dev);
+			if (tmp != NULL) {
+				/* TRANSLATORS: helpful messages from last update */
+				fu_common_string_append_kv (str, idt + 1, _("Update Message"), tmp);
+			}
+		}
 	}
 	tmp = fwupd_device_get_update_error (dev);
 	if (tmp != NULL) {
 		/* TRANSLATORS: error message from last update attempt */
 		fu_common_string_append_kv (str, idt + 1, _("Update Error"), tmp);
-	}
-	tmp = fwupd_device_get_update_message (dev);
-	if (tmp != NULL) {
-		/* TRANSLATORS: helpful messages from last update */
-		fu_common_string_append_kv (str, idt + 1, _("Update Message"), tmp);
 	}
 
 	/* modified date: for history devices */
--- a/plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
+++ b/plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
@@ -126,12 +126,8 @@ fu_plugin_device_registered_uefi (FuPlugin *plugin, FuDevice *device)
 			}
 		}
 		/* check at least one reconstruction for this algorithm */
-		if (!data->reconstructed) {
-			fu_device_set_update_message (device,
-						     "TPM PCR0 differs from reconstruction, "
-						     "please see https://github.com/fwupd/fwupd/wiki/TPM-PCR0-differs-from-reconstruction");
+		if (!data->reconstructed)
 			return;
-		}
 	}
 }
 
--- a/plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
+++ b/plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
@@ -10,11 +10,9 @@
 #include "fu-plugin-vfuncs.h"
 
 #include "fu-tpm-eventlog-device.h"
-#include "fu-efivar.h"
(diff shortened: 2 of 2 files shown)


hughsie:
Using FWUPD_RELEASE_FLAG_IS_UPGRADE seems a bit fishy... don't we also want to show it for downgrades too?

superm1:
> Using FWUPD_RELEASE_FLAG_IS_UPGRADE seems a bit fishy... don't we also want to show it for downgrades and reinstalls too?

This is only when it's copied for the device.  IE what you would see in something from get-devices after a metadata refresh.  If you actually upgrade, downgrade or reinstall it should be displayed from `fu-util.c'` `display_current_message`.

My logic is you don't want to see it unless there is an upgrade available, or you actually upgrade/downgrade/reinstall.

superm1:
With this branch here is what you experience in `reinstall`
```shell
$ sudo fwupdmgr reinstall
Choose a device:
0.      Cancel
1.      c29fbc06f3cf2d8ff0ba78a66c09ccdea3fc9850 (Thunderbolt controller in Dell dock)
2.      073624c16dd99abe01ba1da223a70321e4f29beb (Package level of Dell dock)
3.      0acceea54e71c4d8002593822afe0f4705616613 (RTS5413 in Dell dock)
4.      96969b82d977fdcbf607df084ef0dcf10119409b (RTS5487 in Dell dock)
5.      228ac6846b64d532dd1ea1b7e651e8f5f55fbc34 (VMM5331 in Dell dock)
6.      570867f4ddda9915445da59bd85cd0e0e507270e (WD19TB)
7.      b6c08fb9e5384d9d101853cc1ca20cf0ce2df2e2 (System Firmware)
6
Downloading 01.00.00.04 for WD19TB...
Installing on Package level of Dell dock…                        ]
Restarting device…       [***************************************]
Installing on VMM5331 in Dell dock…                              ]
Restarting device…       [***************************************] Less than one minute remaining…
Installing on WD19TB…    [************************************** ]
Writing…                 [***************************************] Less than one minute remaining…
Successfully installed firmware: Updates will complete after the dock is unplugged. The dock will be unavailable for approximately 1 minute.
```

superm1:
It works because the device doesn't have `FuDevice` message set unless there is an upgrade available from metadata.  So running get-devices won't show the key unless you've refreshed metadata and there is that upgrade.

hughsie:
> I'm just really confused why it works...

hughsie:
> I'm just really confused why it works...

I think I’m having a really hard time understanding what this patch does because we’ve never actually defined what `LVFS::UpdateMessage` really is. At the moment it’s:

* A string on the FwupdRelease
* A string on the FwupdDevice
* Settable using a quirk
* Settable using the metainfo.xml file by the uploader, which makes its way into the AppStream metadata
* Used to show warnings like the PCR0 reconstruction failure and secure boot disabled message
* Saved in the command line tools whenever device-changed is fired, and then the last set message is shown after install/update/reinstall

So my confusion is whether the update message is a property on the release (i.e. set by the OEM) or a property on the device (i.e. set by the quirk) – or both where the implication is the device should inherit the message from the/any release. Or, if it’s a transient message generated by the daemon when an operation is in progress. I think we need to decide when we’re supposed to firstly set the metadata property, and the secondly when we’re supposed to show the property – and if we need to split the property into two different things.

What I think we want to define the  LVFS::UpdateMessage as is “this is the text that is shown at the end of a successful install/update/reinstall/downgrade. What I’m not sure about is whether

* the front end should be dumb and just show the update message and rely on the daemon to get and clear the status depending on the last performed action, and so show the message unconditionally.
* the front end should only show the update message in response to what it’s done itself, i.e. so we unconditionally set the UpdateMessage in the daemon at all times even when an update has not been performed.
* whether we should show it for failed updates too

Another thing is that I think we should think about is showing an image for GUI clients like GNOME software. For instance showing a line art image of the dock showing removing the TBT cable and not the power – which is something VLI also wants to use.

Although your patch works, I think it’s very much a workaround that will come back to bite us. I’m happy to do the work and add tests for this kind of thing, but I think it’s something that’s worth doing properly to make sure everything always works like we’d expect.

superm1:
Thanks for explaining the confusion.

>What I think we want to define the LVFS::UpdateMessage as is “this is the text that is shown at the end of a successful install/update/reinstall/downgrade.

Yes I agree.  We probably shouldn't be showing this key at any other time that means.  Every other use we've overloaded it with should be changed.

>the front end should be dumb and just show the update message and rely on the daemon to get and clear the status depending on the last performed action, and so show the message unconditionally.
the front end should only show the update message in response to what it’s done itself, i.e. so we unconditionally set the UpdateMessage in the daemon at all times even when an update has not been performed.
whether we should show it for failed updates too

My preference is frontend to be dumb and show it every time you do something like upgrade/downgrade.

>Another thing is that I think we should think about is showing an image for GUI clients like GNOME software. For instance showing a line art image of the dock showing removing the TBT cable and not the power – which is something VLI also wants to use.

Well originally it was going to be showing the text key, but I guess this never landed:
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/806

>Although your patch works, I think it’s very much a workaround that will come back to bite us. I’m happy to do the work and add tests for this kind of thing, but I think it’s something that’s worth doing properly to make sure everything always works like we’d expect.

Yeah it sounds like the biggest problem here is that we overloaded it as a pipe of messages to a user..

hughsie:
> My preference is frontend to be dumb and show it every time you do something like upgrade/downgrade.

I think that's the best approach; we then need to decide when to *unset* it -- e.g. if the device unplugs we have to manually clear it for instance. Should we only add the AppStream (per-release) LVFS:UpdateMessage at the end of the `fu_engine_install_release()` call for instance? What about the case where the UpdateMessage is added as a quirk on the device?

It all might be a lot cleaner to just show the UpdateMessage in the client only if the update_state is FWUPD_UPDATE_STATE_SUCCESS -- or something like FWUPD_DEVICE_FLAG_SHOW_UPDATE_MESSAGE which can toggle in the right place without requerying quirks and going through the list of releases.

> Yeah it sounds like the biggest problem here is that we overloaded it as a pipe of messages to a user..

Something like the PCR) not reconstructing properly is really just a message; and there's no reason why we can't have more than one message. An array of strings seems most sensible for this.

superm1:
> It all might be a lot cleaner to just show the UpdateMessage in the client only if the update_state is FWUPD_UPDATE_STATE_SUCCESS -- or something like FWUPD_DEVICE_FLAG_SHOW_UPDATE_MESSAGE which can toggle in the right place without requerying quirks and going through the list of releases.

That seems pretty sensible.  So it always clones from release -> device if it's not set from quirks.  And client decides when to show it.

> Something like the PCR) not reconstructing properly is really just a message; and there's no reason why we can't have more than one message. An array of strings seems most sensible for this.

The PCR reconstructing message I think was useful to get an idea of where PCR reconstruction errors exist.  I think it uncovered a lot of firmware issues, at least one tool issue.  In general I don't think it should stick around and be shown to people as it is today.  I would prefer to move it to the output of `fwupdmgr security`.  If you're aligned on this, i'll send up a PR to remove it.

The secure boot message didn't that just get dropped?  (IE https://github.com/fwupd/fwupd/commit/4669dd590a9a0cc4cc962b6809c8dcd4d4cfe51c#diff-2f5c39c775e4cac7ebbeabed4cefca6a)

hughsie:
> That seems pretty sensible

The SUCCESS checking or new flag?

>  I would prefer to move it to the output of fwupdmgr security

Yes, good idea.

superm1:
> > That seems pretty sensible
> 
> The SUCCESS checking or new flag?

Success checking.  Reverted old commit idea and added to this PR.

> 
> > I would prefer to move it to the output of fwupdmgr security
> 
> Yes, good idea.

I've added it to this PR.

src/fu-util-common.c
```diff
@@ -1608,6 +1609,12 @@ fu_util_security_attrs_to_string (GPtrArray *attrs)
 			if (j < FWUPD_SECURITY_ATTR_LEVEL_IMPORTANT &&
 			    !fwupd_security_attr_has_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS))
 				low_help = TRUE;
+
+			/* check for PCR0 not matching */
+			if (g_strcmp0 (fwupd_security_attr_get_appstream_id (attr),
+					FWUPD_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0) == 0 &&
+			    fwupd_security_attr_get_result (attr) == FWUPD_SECURITY_ATTR_RESULT_NOT_VALID)
+				    pcr0_help = TRUE;
```
hughsie, line 1620: I think this is a much better way to handle this kind of thing.

src/fu-util-common.c
```diff
@@ -1210,17 +1210,20 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
...
 		fu_common_string_append_kv (str, idt + 1, _("Update State"),
 					    fwupd_update_state_to_string (state));
+
+		if (state == FWUPD_UPDATE_STATE_SUCCESS) {
+			tmp = fwupd_device_get_update_message (dev);
+			if (tmp != NULL) {
+				/* TRANSLATORS: helpful messages from last update */
+				fu_common_string_append_kv (str, idt + 1, _("Update Message"), tmp);
+			}
+		}
```
hughsie, line 1220: do we want it here, or in fu_util_display_current_message?

superm1: Here I think, this is what will translate into `get-devices` output.

hughsie: R5ght, i think we probably want it in both places.

superm1: No, I don't think so.  If the firmware fails we don't call `fu_util_display_current_message`.  So why check for success?

hughsie: Indeed, you're correct, my apologies.

Issue #2219, opened 2020-06-23
Dell XPS 9570 System Firmware - Update errors


ericonr:
**Describe the bug**

https://fwupd.org/lvfs/devices/com.dell.uefie4d7bdc4.firmware (as well as Dell's own page) shows that the 1.16.2 version of the system firmware for this laptop is available. My system is running the 1.15.0 one, but I get update errors when running `get-devices`, and the update isn't even shown. I have already refreshed and force refreshed the database.

```
$ fwupdmgr get-devices
[...]
├─System Firmware:
│     Device ID:           xxxxxx
│     Current version:     1.15.0
│     Minimum Version:     1.15.0
│     Vendor:              Dell Inc. (DMI:Dell Inc.)
│     Update Error:        missing signed bootloader for secure boot: /usr/libexec/fwupd/efi/fwupdx64.efi.signed cannot be found
│     GUID:                xxxxxx
│     Device Flags:        • Internal device
│                          • Requires AC power
│                          • Needs a reboot after installation
│                          • Cryptographic hash verification is available
[...]

$ fwupmgr update
• Thunderbolt Controller has no available firmware updates
• PC401 NVMe SK hynix 512GB has no available firmware updates
• Touchpad has no available firmware updates
No updatable devices
```

I have signed the bootloader with `sbsign` (the same error appeared before and after that):

```
$ cd cd /usr/libexec/fwupd/efi/
$ sudo sbsign --cert /etc/refind.d/keys/refind_local.crt --key /etc/refind.d/keys/refind_local.key fwupdx64.efi
$ ls
fwupdx64.efi
fwupdx64.efi.signed
```

I don't know if it's related, but I get this info about the TPM module:

```
├─TPM 2.0:
│     Device ID:           xxxxxxx
│     Summary:             Platform TPM device
│     Current version:     7.2.0.1
│     Vendor:              Dell Inc. (PCI:0x1028)
│     Update Error:        Updating disabled due to TPM ownership
│     GUIDs:               xxxxx 
[...]
│     Device Flags:        • Internal device
│                          • Requires AC power
```

I have erased all the codes because I don't know how unique they are, but if you need any of them, let me know.

**Steps to Reproduce**

Explained above, I think.

**Expected behavior**

The update would show up regardless of the presence of a signed bootloader. Ideally, some mechanism to create the signed bootloader should be available as well.

**fwupd version information**

Please provide the version of the daemon and client.

```
$ fwupdmgr --version
client version:	1.4.2
[...]

PR #2220, opened 2020-06-23
trivial: uefi: drop secure boot check at coldplug

superm1:
We'll instead check this when the user tries to run an update.  This
allows them to sign a bootloader after the daemon starts (or remove
a signed bootloader after starting)

Fixes: #2219

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -730,10 +730,8 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 {
 	FuPluginData *data = fu_plugin_get_data (plugin);
 	const gchar *str;
-	g_autofree gchar *bootloader = NULL;
 	g_autofree gchar *esrt_path = NULL;
 	g_autofree gchar *sysfsfwdir = NULL;
-	g_autoptr(GError) error_bootloader = NULL;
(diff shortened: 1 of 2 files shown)


plugins/uefi/fu-uefi-bootmgr.c
```diff
@@ -302,8 +302,11 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
 
 	/* if secure boot was turned on this might need to be installed separately */
 	source_app = fu_uefi_get_built_app_path (error);
-	if (source_app == NULL)
+	if (source_app == NULL) {
+		if (fu_efivar_secure_boot_enabled ())
+			g_prefix_error (error, "missing signed bootloader for secure boot: ");
 		return FALSE;
+	}
```
hughsie, line 309: ...this is way too late. This is after the update has been deployed to the ESP -- checks that the update is going to be deployable need to be done in `->prepare_firmware()` not `->write_firmware()`. From a UX point of view the user is going to see it be "installed" and then for the process to fail. Will the update work? Should the user reboot? etc.

superm1: Good point, thanks.  I just pushed a fixup.

PR #2221, opened 2020-06-24
trivial: fu-util: show a better error for non-responsive daemon

superm1:
See #2212 for more context.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -2544,6 +2544,15 @@ fu_util_check_daemon_version (FuUtilPrivate *priv, GError **error)
 {
 	const gchar *daemon = fwupd_client_get_daemon_version (priv->client);
 
+	if (daemon == NULL) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     /* TRANSLATORS: error message */
+				     _("Unable to connect to service"));
+		return FALSE;
+	}
+
 	if (g_strcmp0 (daemon, SOURCE_VERSION) != 0) {
 		g_set_error (error,
 			     FWUPD_ERROR,


src/fu-util.c
```diff
@@ -2545,12 +2545,20 @@ fu_util_check_daemon_version (FuUtilPrivate *priv, GError **error)
 	const gchar *daemon = fwupd_client_get_daemon_version (priv->client);
 
 	if (g_strcmp0 (daemon, SOURCE_VERSION) != 0) {
-		g_set_error (error,
-			     FWUPD_ERROR,
-			     FWUPD_ERROR_NOT_SUPPORTED,
-			     /* TRANSLATORS: error message */
-			     _("Unsupported daemon version %s, client version is %s"),
-			     daemon, SOURCE_VERSION);
+		if (daemon == NULL) {
```
hughsie, line 2548: i think this should go before the `if (g_strcmp0 (daemon, SOURCE_VERSION`

src/fu-util.c
```diff
@@ -2545,12 +2545,20 @@ fu_util_check_daemon_version (FuUtilPrivate *priv, GError **error)
...
+			g_set_error_literal (error,
+					     FWUPD_ERROR,
+					     FWUPD_ERROR_NOT_SUPPORTED,
+					     /* TRANSLATORS: error message */
+					     _("Unable to connect to daemon"));
```
hughsie, line 2553: maybe `daemon` is too technical and not specific enough? maybe `fwupd service` or something like that?

src/fu-util.c
superm1: fixed

src/fu-util.c
superm1: I made it `service` on the logic that snap uses `snap.fwupd.fwupd.service` and regular uses `fwupd.service` and I don't want to have to show either in an error string.

PR #2226, opened 2020-06-25
Drop the automatic /usr/bin/python3 rpmbuild dependency

hughsie:
We ship 4 *tiny* python scripts that are useful for ODMs and other people
working with low level firmware blobs.

These helper utilities do not warrant dragging Python onto the CoreOS image.

### DIFF
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -6,6 +6,10 @@
 %global systemd_version 231
 %global json_glib_version 1.1.1
 
+# although we ship a few tiny python files these are utilities that 99.99%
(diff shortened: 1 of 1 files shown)


hughsie:
> Do you need to also drop/change to suggests Recommends: python3 too

Good question, but I think it can stay at Recommends -- which means a "dnf install fwupd" would drag it in if not already on the system, but python wouldn't be included when using imgbuilder or equiv.

superm1:
LGTM then

LorbusChris:
@hughsie can we get this ported to the 1.4.x and 1.3.x releases?

superm1:
It's already backported to 1.4.x, just backported to 1.3.x

Issue #2223, opened 2020-06-25
lvfs-testing: no signature method in results


joakim-tjernlund:
**Describe the bug**
 fwupdmgr refresh
Fetching signature https://cdn.fwupd.org/downloads/firmware-testing.xml.gz.jcat
Downloading…             [***************************************]
Fetching metadata https://cdn.fwupd.org/downloads/firmware-00432-testing.xml.gz
Downloading…             [***************************************]
Failed to update metadata for lvfs-testing: no signature method in results

**fwupd version information**
fwupdmgr --version
client version:	1.4.4
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.4.4

emerge sys-apps/fwupd

**Additional questions**
- Operating system and version: Gentoo
- Have you tried rebooting? Yes
- Is this a regression? Yes

hughsie:
> Operating system and version: Gentoo

Did you compile libjcat without GPG **and** GNUTLS? What libjcat version are you using?

joakim-tjernlund:
libjcat 0.1.3

It is built with both gpg an gnutls

joakim-tjernlund:
```
Program diff found: YES (/usr/bin/diff)
Found pkg-config: /usr/bin/x86_64-pc-linux-gnu-pkg-config (0.29.2)
Run-time dependency gio-2.0 found: YES 2.62.6
Run-time dependency gio-unix-2.0 found: YES 2.62.6
Run-time dependency json-glib-1.0 found: YES 1.4.4
Run-time dependency gnutls found: YES 3.6.14
Library gpgme found: YES
Library gpg-error found: YES
Configuring config.h using configuration                                                            
Configuring jcat-version.h using configuration

**Comment by joakim-tjernlund** [comment]
This might be a clue:
```
fwupdtool refresh
Loading…                 [-                                      ]ERROR:esys:src/tss2-esys/esys_context.c:69:Esys_Initialize() Initialize default tcti. ErrorCode (0x000a000a) 
Loading…                 [***************************************]
11:48:02:0244 GLib-GObject         g_object_get_is_valid_property: object class 'JcatGpgEngine' has no property named 'verify-kind'
11:48:02:0244 GLib-GObject         g_object_get_is_valid_property: object class 'JcatPkcs7Engine' has no property named 'verify-kind'
11:48:02:0244 GLib-GObject         g_object_get_is_valid_property: object class 'JcatSha1Engine' has no property named 'verify-kind'
[...]

PR #2227, opened 2020-06-25
Do not show a critical warning if the user-cache is not set

hughsie:
Fixes half of https://github.com/fwupd/fwupd/issues/2223

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -248,7 +248,7 @@ fu_util_is_interesting_device (FwupdDevice *dev)
 }
 
 gchar *
-fu_util_get_user_cache_path (const gchar *fn)
+fu_util_get_user_cache_path (const gchar *fn, GError **error)
 {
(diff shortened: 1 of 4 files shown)


src/fu-util-common.c
```diff
@@ -248,7 +248,7 @@ fu_util_is_interesting_device (FwupdDevice *dev)
 }
 
 gchar *
-fu_util_get_user_cache_path (const gchar *fn)
+fu_util_get_user_cache_path (const gchar *fn, GError **error)
 {
 	const gchar *root = g_get_user_cache_dir ();
 	g_autofree gchar *basename = g_path_get_basename (fn);
```
joakim-tjernlund, line 254: I think my error comes from here because fn is NULL

superm1: In that case we probably shouldn't merge this PR, but fix the case that was causing `fn` to be `NULL`

PR #2228, opened 2020-06-25
Support LVFS::UpdateImage in GUI clients

hughsie:
The idea here is that we can show the user both a string and an optional
line-art image when the update has completed. The line art is often more well
understood for non-English speakers.

### DIFF
--- a/libfwupd/fwupd.map
+++ b/libfwupd/fwupd.map
@@ -447,6 +447,15 @@ LIBFWUPD_1.4.1 {
   local: *;
 } LIBFWUPD_1.4.0;
 
+LIBFWUPD_1.4.5 {
+  global:
+    fwupd_device_get_update_image;
+    fwupd_device_set_update_image;
+    fwupd_release_get_update_image;
+    fwupd_release_set_update_image;
+  local: *;
+} LIBFWUPD_1.4.1;
+
 LIBFWUPD_1.5.0 {
   global:
     fwupd_client_get_host_security_attrs;
@@ -484,4 +493,4 @@ LIBFWUPD_1.5.0 {
     fwupd_security_attr_to_string;
     fwupd_security_attr_to_variant;
   local: *;
-} LIBFWUPD_1.4.1;
+} LIBFWUPD_1.4.5;
--- a/libfwupd/fwupd-device.c
+++ b/libfwupd/fwupd-device.c
@@ -59,6 +59,7 @@ typedef struct {
 	FwupdUpdateState		 update_state;
 	gchar				*update_error;
 	gchar				*update_message;
+	gchar				*update_image;
 	FwupdStatus			 status;
 	GPtrArray			*releases;
 	FwupdDevice			*parent;
@@ -1201,6 +1202,8 @@ fwupd_device_incorporate (FwupdDevice *self, FwupdDevice *donor)
 		fwupd_device_set_update_error (self, priv_donor->update_error);
 	if (priv->update_message == NULL)
 		fwupd_device_set_update_message (self, priv_donor->update_message);
+	if (priv->update_image == NULL)
+		fwupd_device_set_update_image (self, priv_donor->update_image);
 	if (priv->version == NULL)
 		fwupd_device_set_version (self, priv_donor->version);
 	if (priv->version_lowest == NULL)
@@ -1394,6 +1397,11 @@ fwupd_device_to_variant_full (FwupdDevice *device, FwupdDeviceFlags flags)
 				       FWUPD_RESULT_KEY_UPDATE_MESSAGE,
 				       g_variant_new_string (priv->update_message));
 	}
+	if (priv->update_image != NULL) {
+		g_variant_builder_add (&builder, "{sv}",
+				       FWUPD_RESULT_KEY_UPDATE_IMAGE,
+				       g_variant_new_string (priv->update_image));
+	}
 	if (priv->update_state != FWUPD_UPDATE_STATE_UNKNOWN) {
 		g_variant_builder_add (&builder, "{sv}",
 				       FWUPD_RESULT_KEY_UPDATE_STATE,
@@ -1576,6 +1584,10 @@ fwupd_device_from_key_value (FwupdDevice *device, const gchar *key, GVariant *va
 	}
 	if (g_strcmp0 (key, FWUPD_RESULT_KEY_UPDATE_MESSAGE) == 0) {
 		fwupd_device_set_update_message (device, g_variant_get_string (value, NULL));
+		return;
+	}
+	if (g_strcmp0 (key, FWUPD_RESULT_KEY_UPDATE_IMAGE) == 0) {
+		fwupd_device_set_update_image (device, g_variant_get_string (value, NULL));
 		return;
 	}
 	if (g_strcmp0 (key, FWUPD_RESULT_KEY_UPDATE_STATE) == 0) {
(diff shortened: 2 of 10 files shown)


libfwupd/fwupd.map
```diff
@@ -447,6 +447,15 @@ LIBFWUPD_1.4.1 {
   local: *;
 } LIBFWUPD_1.4.0;
 
+LIBFWUPD_1.4.5 {
```
superm1, line 450: 1.4.5?

superm1:
> The idea here is that we can show the user both a string and an optional
line-art image when the update has completed. The line art is often more well
understood for non-English speakers.

What does line art mean?  Is that different than ASCII art?
is it supposed to be a GIF?  
Is it supposed to be rendered on a command line client too with some jpg to ascii translation type thing?

libfwupd/fwupd.map
hughsie: I wondered if you would notice that; my intention for this would be to backport the new API into 1_4_X so that we could include it in Fedora 32 etc. I think 1.5.0 is some way off being "ready" as I think the HSI spec kinda determines when we need to branch again.

superm1:
I would almost question if it's an image if we're better off putting the image _inside_ of the CAB file itself?

hughsie:
> What does line art mean

Maybe that's the wrong word to use. I mean the blocky style we're using for the other vendors already: https://github.com/fwupd/8bitdo-firmware/blob/master/screenshots/FC30PRO.png

libfwupd/fwupd.map
superm1: Hmm, new API in the stable branch..  I guess it's not a _super_ risky API.

>I think 1.5.0 is some way off being "ready" as I think the HSI spec kinda determines when we need to branch again.

I'm of the perspective we should probably release 1.5.0 soon.  Lots of changes that should get flushed out and HSI will stay _pre-release_ in 1.5.0.  When we're ready to have it live, drop the need for `--force`

hughsie:
> I would almost question if it's an image if we're better off putting the image inside of the CAB file itself?

Hmm, that's not awesome from a dedupe point of view; if there are 50 firmwares with 5 different 50kB images in then the storage size just keeps increasing. It's also very hard to change the style of an existing firmware if we embed it, rather than just updating where the URL points to. It's also nontrivial to safely export a blob of memory from the parsed file to the client process.

superm1:
> > What does line art mean
> 
> Maybe that's the wrong word to use. I mean the blocky style we're using for the other vendors already: https://github.com/fwupd/8bitdo-firmware/blob/master/screenshots/FC30PRO.png

With push a button to do this a fixed png makes perfect sense.  But an "unplug" your dock, how do you do it without an animation?  How do you indicate which cable to unplug?

image
That is the best (type of) design I could think.

libfwupd/fwupd.map
hughsie: I'd like to get the ARM and AMD opinions before we commit to API.

superm1:
Another thought - if this is mostly going to be used for unplug / power cycle your device type scenarios, what about distributing a set of generic graphics _with_ fwupd and some translatable strings?  They can be generic and then when it's time to unplug your device a GUI client can flash up the graphic.

hughsie:
I pondered something like this:

ss

libfwupd/fwupd.map
superm1: Oh you mean because of library implications of potentially breaking API.   Is there really that much that can change from another set of vendors coming on board?  I'm curious what sort of things you envision could have to happen.

superm1:
To be honest, I don't understand what that picture means.  Is it the port powering the dock?  The port you have something plugged into upstream on the dock?

hughsie:
> To be honest, I don't understand what that picture means

Ya, I'm no graphic designer. I think it does need some arrows. This is for a Lenovo dock that has both ends removable, and maybe something like the TB lightning bolt and an "out" arrow might be clearer.

libfwupd/fwupd.map
hughsie: > Oh you mean because of library implications of potentially breaking API

Right.

>  Is there really that much that can change from another set of vendors coming on board

I'm not really sure. I wouldn't be at all surprised if there were some kind of random requirement on displaying the results for instance.

superm1:
What I was thinking for a generic thing is something host focused and can be drawn/animated.

Here's an example shamelessly borrowed from https://hackaday.com/2019/07/29/usb-c-one-plug-to-connect-them-all-and-in-confusion-bind-them/
image

Make something similar with USB-A and another that is USB-C that has a generic computer and a generic cable hooked up to it.

superm1:
(Looks like Github isn't animating it here is the original: https://hackaday.com/wp-content/uploads/2019/07/usb-orientation.gif?w=250)

hughsie:
> What I was thinking for a generic thing is something host focused and can be drawn/animated.

Yes, that's really good. I can ask jimmac for some more examples in the GNOME style. I do think these the stock artwork should be distributed still by the LVFS; putting the artwork in the fwupd package means we can't deploy newer firmware on anything but the latest fwupd if we need anything out of the ordinary from a user-interaction point of view.

superm1:
>putting the artwork in the fwupd package 

I was looking at it from the perspective that I want this stuff to be able to work even if someone has a SUT without web access and double clicks a CAB file transferred via sneakernet or locked down intranet.  That was why I originally suggested embedding a graphic in the CAB.  If it's a generic graphic then distribute it with fwupd and all frontends can just show it.

Another idea here could be to do both: ship generic ones with fwupd and on LVFS, and then if you add a new one it can be in a newer fwupd version, and older versions could reference the LVFS location?

> we can't deploy newer firmware on anything but the latest fwupd if we need anything out of the ordinary from a user-interaction point of view.

Well we don't usually backport plugins supporting new devices to older versions though.  Older "stable" release versions are just bug fixes usually.

So if there is something that needs more than basic user interaction like lets say flipping a switch or pressing a button, the code to support that would probably be a new plugin wouldn't it be?

hughsie:
> that I want this stuff to be able to work even if someone has a SUT without web access

The detach image is in the same boat then -- I don't know if we always have the .cab file when we detach (e.g. verify). I guess I can see the argument for including them both in the cab, but we'd have to rewrite the cab file on the LVFS and export the blob from the daemon. I'm not so sold installing them on the filesystem, although I do see your logic on not backporting new plugins.

hughsie:
> I'm not so sold installing them on the filesystem

Although, if we do decide this is the way to go, exporting the URL is as easy as `file:///usr/share/fwupd/`etc...

superm1:
Well /usr shouldn't go in metadata, that locks it to a certain prefix. I was going to suggest that just list some predefined filenames in the XML.

Fwupd can then try to go local first to find it and build the URI it exports with the install prefix and if it's not available then build an lvfs URL to get it from.

hughsie:
> some predefined filenames in the XML

I don't think that's forwards and backwards compat; I think what we'd need to do is have a URL like http://fwupd.org/downloads/usb-unplug.png and then if usb-unplug.png exists on the filesystem then we use that instead. I don't think we can know the URL base when just processing a random cab without a remote_id set.

With your blessing, I'd like to commit this simple version and then work with the GNOME designers on getting some "stock" artwork we can use for the most common cases which can be shipped in fwupd.

superm1:
Okay

PR #2232, opened 2020-06-26
Allow firmware to require specific functionality from front-end clients

hughsie:
At the moment we just blindly assume the capabilities of the front-end client
when installing firmware. We can somewhat work around by requiring a new enough
fwupd daemon version, but the client software may be older or just incomplete.

This would allow, for instance, the firmware to specify that it requries the
client to be able to show a detach image. This would not be set by a command
line tool using FwupdClient, but would be set by a GUI client that is capable
of downloading a URL and showing a PNG image.

Clients that do not register features are assumed to be dumb.


hughsie:
If this looks acceptable, I'll of course document this new requirement type in lvfs-website.

hughsie:
Screenshot from 2020-06-26 14-50-55

libfwupd/fwupd-enums.h
```diff
@@ -64,6 +64,28 @@ typedef enum {
 	FWUPD_TRUST_FLAG_LAST
 } FwupdTrustFlags;
 
+/**
+ * FwupdFeatureFlags:
+ * @FWUPD_FEATURE_FLAG_NONE:			No trust
+ * @FWUPD_FEATURE_FLAG_CAN_REPORT:		Can report successful update
+ * @FWUPD_FEATURE_FLAG_DETACH_TEXT:		Can show detach text
```
superm1, line 71: How are detach text and post-update text different?

src/fu-engine.c
```diff
@@ -1355,8 +1356,75 @@ fu_engine_check_requirement_hardware (FuEngine *self, XbNode *req, GError **erro
...
+fu_engine_check_requirement_client (FuEngine *self,
+				    XbNode *req,
+				    const gchar *sender,
+				    GError **error)
+{
+	FwupdFeatureFlags flags;
+	g_auto(GStrv) feature_split = NULL;
+
+	/* should not be set in normal operation */
+	if (g_strcmp0 (sender, FU_ENGINE_DBUS_SENDER_INVALID) == 0) {
```
superm1, line 1383: Isn't this going to lock out a `fwupdtool` client that can show text messages?

libfwupd/fwupd-enums.h
hughsie: the detach image is stored in a `<screenshot>` and is shown before the update is deployed, and the update image is stored in a `<custom>` section and shown after the update has successfully completed.

src/fu-engine.c
hughsie: Yes, good catch. Either this `if` can just be deleted, or we add something like `FU_TOOL_SENDER_FAKE = "fake"`

libfwupd/fwupd-enums.h
```diff
@@ -64,6 +64,28 @@ typedef enum {
...
+ * @FWUPD_FEATURE_FLAG_DETACH_IMAGE:		Can show detach image
+ * @FWUPD_FEATURE_FLAG_UPDATE_TEXT:		Can show post-update text
+ * @FWUPD_FEATURE_FLAG_UPDATE_IMAGE:		Can show post-update image
```
superm1, line 74: This has the potential to set up a weird dependency where you can only install certain updates using a GUI client and certain updates using a text client.  Maybe it can be avoided by the text client calls something like `jp2a`, but that's definitely not the ideal solution.

I think really you want the message to be that the client can show the user post-update steps and it's up to the firmware metainfo to specify both the text and graphical steps.

hughsie:
> before it performs the request

isn't this a massive race?

libfwupd/fwupd-enums.h
superm1: Oh... like the 8bitdo thing.  It either needs more description or a better name then.  Maybe PRE-INSTALL text?

src/fu-engine.c
superm1: Well see my comment about a client setting features to the engine, I think the other option is to have fu-tool set features instead of this sender checking stuff.  You just check if the engine's current client supports the features, fu-tool included.

libfwupd/fwupd-enums.h
hughsie: > set up a weird dependency where you can only install certain updates using a GUI client

yes, that's kinda the idea :)

> I think really you want the message to be that the client can show the user post-update steps

Ahh, so we'd have something like FWUPD_FEATURE_FLAG_DETACH_ACTION and FWUPD_FEATURE_FLAG_UPDATE_ACTION and just silently fall back to the text if text+image was not available?

superm1:
>isn't this a massive race?

Encapsulate it in a libfwupd call, protect operations to the daemon with a mutex (I know being single threaded not a problem now, but for $futureproof)

libfwupd/fwupd-enums.h
superm1: >yes, that's kinda the idea :)

I find it impossible that there is a message that can ONLY be conveyed via GUI.  GUI is probably preferable, but making a hard requirement on a GUI is going to ruin a lot of use cases I feel like.

>Ahh, so we'd have something like FWUPD_FEATURE_FLAG_DETACH_ACTION and FWUPD_FEATURE_FLAG_UPDATE_ACTION and just silently fall back to the text if text+image was not available?

Yes!

libfwupd/fwupd-enums.h
hughsie: >  Maybe PRE-INSTALL text

Well, at the moment gnome-software explicitly tests for:

    if (gs_app_get_kind (app) == AS_APP_KIND_FIRMWARE &&
        fwupd_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_BOOTLOADER)) {

...so it really is detach specific. That said, I think *another* feature of pre-install could be added in the future if needed.

src/fu-engine.c
hughsie: > the engine's current client

hughsie: > the engine's current client

That seems a huge assumption for a shared daemon.

libfwupd/fwupd-enums.h
superm1: OK

src/fu-engine.c
superm1: But if you bundle `fwupd_client_install` with something that sets up all the current features?

libfwupd/fwupd-enums.h
hughsie: > Yes!

Will fix.

src/fu-engine.c
hughsie: you need it way before install time; to check the requirements you need the featureset at `GetDetails` and `GetUpgrades` time for instance. There's no point showing the user they have an update with no indication it might not work but then just failing on the install step.

superm1: But so bundle it with each call to the daemon.  When you do fwupd_client_get_details, update the features, you get the relevant results.  When you do fwupd_client_install, you update the features again.  I just think sending the sender id all over the daemon seems like a lot of unnecessary plumbing here.

hughsie: > But so bundle it with each call to the daemon

One thing I considered is breaking all the D-Bus method API and adding a `a{sv}` called `options` or something where each request could set the feature flags, the locale etc. That's probably architecturally cleanest, but it's a **huge** API break I'm super keen on right now -- a few projects are using the D-Bus APIs rather than the GObject client. Although translated device descriptions and release notes sounds like a good idea, it's impossible to get the content (both from the hardware, and from OEMs) and so I kinda gave up on that a while ago.

> I just think sending the sender id all over the daemon seems like a lot of unnecessary plumbing here.

What about rather than sending the `sender` all over the place we instead looked up the feature flags in the `fu-main.c` file and passed the feature_flags to each engine method. That means FU_ENGINE_DBUS_SENDER_INVALID kinda becomes unnecessary for fwupdtool.

> why not just add a new private detail for client_features

superm1: >One thing I considered is breaking all the D-Bus method API 

But it's all abstracted by the library anyway.  And it re-inforces the point that people shouldn't be (ab)using the dbus API directly, they **should** be using the library.

> I'm super keen on right now

You mean _not_ super keen on presumably?  You seems to have conflicting thoughts.

>translated device descriptions and release notes sounds like a good idea, it's impossible to get the content (both from the hardware, and from OEMs) and so I kinda gave up on that a while ago.

You really need an army of translator monkeys for this.  I agree it's not worth the effort.

>fu_engine_set_feature_flags

Yes that's exactly what I was trying to convey.

libfwupd/fwupd-enums.h
```diff
@@ -64,6 +64,24 @@ typedef enum {
 	FWUPD_TRUST_FLAG_LAST
 } FwupdTrustFlags;
 
+/**
+ * FwupdFeatureFlags:
+ * @FWUPD_FEATURE_FLAG_NONE:			No trust
+ * @FWUPD_FEATURE_FLAG_CAN_REPORT:		Can report successful update
```
superm1, line 70: presumably not just successful updates?

libfwupd/fwupd-enums.h
```diff
@@ -64,6 +64,24 @@ typedef enum {
...
+ * @FWUPD_FEATURE_FLAG_DETACH_ACTION:		Can perform detach action, typically showing text
+ * @FWUPD_FEATURE_FLAG_UPDATE_ACTION:		Can perform detach action, typically showing text
```
superm1, line 72: s/detach/update/

src/fu-self-test.c
```diff
@@ -214,12 +214,89 @@ fu_engine_requirements_missing_func (gconstpointer user_data)
...
+	gboolean ret;
+	g_autoptr(XbNode) component = NULL;
+	g_autoptr(XbSilo) silo = NULL;
+	g_autoptr(FuEngine) engine = fu_engine_new (FU_APP_FLAGS_NONE);
+	g_autoptr(FuInstallTask) task = NULL;
+	g_autoptr(GError) error = NULL;
+	const gchar *xml =
+		"<component>"
+		"  <requires>"
+		"    <client>detach-action</client>"
```
superm1, line 270: do you want another case that has invalid client requirement perhaps?

libfwupd/fwupd-enums.h
hughsie: `Can upload a report of the update back to the server`?

libfwupd/fwupd-enums.h
hughsie: Ta.

libfwupd/fwupd-enums.h
superm1: ya

src/fu-self-test.c
hughsie: Added, good idea. I used FWUPD_ERROR_NOT_FOUND for not known, and FWUPD_ERROR_NOT_SUPPORTED for not valid, as INVALID_FILE didn't seem right.

src/fu-engine.c
hughsie: > they should be using the library.

It's KDE that wasn't keen on GObject from memory.

> You mean not super keen on presumably

Yes, sorry. As in I really don't want to break the world with a RHEL head on.

> fu_engine_set_feature_flags

I'll prototype something and see what happens.

hughsie: > I'll prototype something and see what happens.

So I started doing that, and it doesn't feel right. For instance, if you call into the daemon with fwupdmgr and then the daemon calls fu_engine_update_motd() then we use the features of the client to calculate the MOTD. If we don't call into the daemon first, then we have no features set and the MOTD is different. I think it's much better to be explicit for things like this otherwise it all gets a bit Rube Goldberg.

That said, I don't like passing the sender around, and the fact that FU_ENGINE_DBUS_SENDER_INVALID exists kinda indicates it's the wrong approach. The other thing I'd like to try is to pass around the FwupdFeatureFlags with the request -- or maybe wrap that up in a GObject something like FuRequest. The latter would let us add other things in the future without breaking internal ABI yet again.

hughsie: > maybe wrap that up in a GObject something like FuRequest

Also, this allows us to nicely tidy up the GetDevices() device flags which depends on the sender; the "if you're root you get the instance ID" thing.

superm1: > MOTD is different

Hmm, I hadn't even started to consider the implications for the motd.  I almost think that should always be calculated as though all the features are supported.  The logic being that the motd message **tells you** to use `fwupdmgr`.

hughsie: > Hmm, I hadn't even started to consider the implications for the motd

Right, when making all the changes I kept having to think "what is set before this API call" which was a mental load that's ripe for getting wrong.

> motd message tells you to use fwupdmgr

Yes, we'd juts hardcode a FuEngineRequest with the same FwupdFeatureFlags as fwupdmgr.

src/fu-util.c
```diff
@@ -2973,6 +2973,17 @@ main (int argc, char *argv[])
...
+	/* send our implemented feature set */
+	if (!fwupd_client_set_feature_flags (priv->client,
+					     FWUPD_FEATURE_FLAG_CAN_REPORT |
+					     FWUPD_FEATURE_FLAG_UPDATE_ACTION |
+					     FWUPD_FEATURE_FLAG_DETACH_ACTION,
+					     priv->cancellable, &error)) {
+		g_printerr ("Failed to set front-end features: %s\n",
+			    error->message);
+		return EXIT_FAILURE;
+	}
```
superm1, line 2985: To add a few other dimensions of of complexity to your thought process that found motd:
* the other non-obvious implication is how the command line client behaves if it has a tty or not.  In non-interactive mode you wouldn't be able to see those messages for detach or unplug anymore.

* The message to unplug, it might not actually be _needed_.  IE it's just to start the boot into the new firmware, or to finalize a firmware upgrade.  If you want to be able to support fully automated updates for stuff that has the flag `usable-during-update` you would probably put in the XML `LVFS::UpdateMessage`, and `LVFS::UpdateImage` but not put `update-action` in client requirements.

hughsie: >  In non-interactive mode you wouldn't be able to see those messages for detach or unplug

I didn't think about that. I guess we can just set the flags correctly if `isatty==0`

> you would probably put in the XML LVFS::UpdateMessage, and LVFS::UpdateImage but not put update-action in client requirements

Right, this is why I didn't "auto require" the features when `LVFS::UpdateMessage` was set.

hughsie: > I didn't think about that. I guess we can just set the flags correctly if isatty==0

That's what I did. I think it's a lot nicer with FuEngineRequest. Comments welcome!

src/fu-engine-request.c
```diff
@@ -0,0 +1,74 @@
...
+}
+
+static void
+fu_engine_request_init (FuEngineRequest *self)
+{
+	/* untrusted */
+	self->device_flags = FWUPD_DEVICE_FLAG_NONE;
+
+	/* a subset of what fwupdmgr can do */
+	self->feature_flags = FWUPD_FEATURE_FLAG_DETACH_ACTION |
```
superm1, line 59: I guess this is how it is specifically so that motd can show everything that `fwupdmgr` can do right?  I am a bit concerned that detail is going to get lost when a new feature comes up.  How about instead defaulting to `FWUPD_FEATURE_FLAG_NONE` and then setting these two in `fu-engine-helper.c`?

src/fu-main.c
```diff
@@ -206,16 +207,26 @@ fu_main_engine_percentage_changed_cb (FuEngine *engine,
...
 	g_autoptr(GVariant) value = NULL;
 
-	g_return_val_if_fail (sender != NULL, FALSE);
-	g_return_val_if_fail (flags != NULL, FALSE);
+	g_return_val_if_fail (sender != NULL, NULL);
+
+	/* did the client set the list of supported feature */
+	feature_flags = g_hash_table_lookup (priv->sender_features, sender);
+	if (feature_flags == NULL) {
+		fu_engine_request_set_feature_flags (request, FWUPD_FEATURE_FLAG_NONE);
```
superm1, line 224: If you do take my other suggestion to set the defaults to none, this can probably be cleaned up that feature_flags set is only called when you have new features to add.

src/fu-util.c
superm1: I think you only got it done for `fu-tool.c`, forgot for `fu-util.c`

src/fu-main.c
hughsie: Done.

src/fu-engine-request.c
hughsie: Ya, that's probably better. Done.

superm1:
I think just that one last nit about fu-tool/fu-util interactive an then LGTM

PR #2235, opened 2020-06-29
Backport the thunderbolt overhaul to 1.4.x

superm1:
The plugins should work the same now as 1.5.x does.  This will make enabling retimers and the authenticate on disconnect stuff more widely.

These should both be empty:
```
diff -u $master/plugins/thunderbolt $1_4_x/plugins/thunderbolt
diff -u $master/plugins/dell-dock $1_4_x/plugins/dell-dock
```

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
Dropping the force power stuff does worry me slightly for the stable branch...

superm1:
> Dropping the force power stuff does worry me slightly for the stable branch...

With how many reports there are of failures to come back on older designs, I was thinking it's probably the right thing to do..

hughsie:
> many reports there are of failures to come back on older designs

Remind me, how does this new patch fix that problem? I thought it was a hardware issue...

superm1:
It avoids the issue. 

The issue happens specifically because the device is force powered on. If you keep something connected instead over the update cycle you don't encounter it.

I do agree it's actually either a thunderbolt controller hardware issue or a thunderbolt controller firmware issue though.

hughsie:
> If you keep something connected instead over the update cycle you don't encounter it

Right, I guess on one hand it's ceding defeat to the hardware, but on the other hand it's being pragmatic. I do think we should do this, but we should squash the commits down into something more manageable, and also merge the other PR into the same commit too. We probably ought to make the behaviour change explicit in the commit message too.

superm1:
> > If you keep something connected instead over the update cycle you don't encounter it
> 
> Right, I guess on one hand it's ceding defeat to the hardware, but on the other hand it's being pragmatic. I do think we should do this, but we should squash the commits down into something more manageable, and also merge the other PR into the same commit too. We probably ought to make the behaviour change explicit in the commit message too.

Can you recommend what combination of commits you want squashed?  The other approach is the "merge commit"

hughsie:
> Can you recommend what combination of commits you want squashed

Not easily, so maybe this is a bad idea.

PR #2238, opened 2020-06-30
Remove potentially problematic language

hughsie:
Red Hat wants to drive an initiative in correcting problematic and potentially
divisive language in open source projects. These naming conventions and
descriptive phrases are hurtful and offensive to many of our colleagues across
the open source universe.

### DIFF
--- a/plugins/csr/fu-csr-device.c
+++ b/plugins/csr/fu-csr-device.c
@@ -397,7 +397,7 @@ fu_csr_device_probe (FuUsbDevice *device, GError **error)
 {
 	FuCsrDevice *self = FU_CSR_DEVICE (device);
 
-	/* devices have to be whitelisted */
+	/* proxy the quirk delay */
 	if (fu_device_has_custom_flag (FU_DEVICE (device),
 				       FU_CSR_DEVICE_FLAG_REQUIRE_DELAY))
 		self->quirks = FU_CSR_DEVICE_QUIRK_REQUIRE_DELAY;
--- a/contrib/ci/fedora.sh
+++ b/contrib/ci/fedora.sh
@@ -53,7 +53,7 @@ mkdir -p dist
 cp $HOME/rpmbuild/RPMS/*/*.rpm dist
 
 if [ "$CI" = "true" ]; then
-	sed "s,^BlacklistPlugins=test;invalid,BlacklistPlugins=," -i /etc/fwupd/daemon.conf
+	sed "s,^DisabledPlugins=test;invalid,DisabledPlugins=," -i /etc/fwupd/daemon.conf
 
 	# set up enough PolicyKit and D-Bus to run the daemon
 	mkdir -p /run/dbus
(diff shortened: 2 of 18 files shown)


plugins/csr/fu-csr-device.c
```diff
@@ -397,7 +397,7 @@ fu_csr_device_probe (FuUsbDevice *device, GError **error)
 {
 	FuCsrDevice *self = FU_CSR_DEVICE (device);
 
-	/* devices have to be whitelisted */
+	/* devices have to be quirked */
```
superm1, line 400: I think instead of quirked - `allowed`

superm1:
Are you also thinking to rename the `master` branch?

plugins/csr/fu-csr-device.c
hughsie: the comment is kinda obscure anyway, I've rewritten it to make sense.

hughsie:
> Are you also thinking to rename the master branch?

Good question; I was tempted to wait until github/gitlab change the default from master to main and then change at the same time to avoid confusion. It'll also let other people fix all the tiny issues for the tools that are hardcoded to expect `master`....

hughsie:
The one I'm not sure about is i2c master/slave. There's not really any prior art on i2c specifically, and all the suggestions make sense in a SPI-type bus but not for low-level i2c. Ideas very welcome....

superm1:
> The one I'm not sure about is i2c master/slave. There's not really any prior art on i2c specifically, and all the suggestions make sense in a SPI-type bus but not for low-level i2c. Ideas very welcome....

I've had the same discussions with others.  There are a few things that have been suggested you can see here:
https://en.wikipedia.org/wiki/Master/slave_(technology)

I personally like primary and secondary.

superm1:
> > Are you also thinking to rename the master branch?
> 
> Good question; I was tempted to wait until github/gitlab change the default from master to main and then change at the same time to avoid confusion. It'll also let other people fix all the tiny issues for the tools that are hardcoded to expect `master`....

I was suspecting that Github wide change will have to be opt-in in some sense because of the fact that you raised.  I see no problem changing it now though.  Realistically those types of problems don't just go away later, it just puts off fixing them to later.

hughsie:
> I personally like primary and secondary.

That's good, but doesn't really show the `do this` relationship of master/slave. Primary/secondary sound like the devices are doing the same thing, but the primary one comes first. Same with main/secondary. tbh, it's not really the _master_ that I have problems with justifying, so maybe master/subordinate or master/serf or master/client or master/worker or master/peon or master/servant works instead.

YehezkelShB:
If the branch name will be change from master, probably is better to remove the usage for I2C too, isn't it?
What about main/client or controller/client?

hughsie:
> What about main/client or controller/client?

For i2c the winner seems to be initiator/responder so far.

hughsie:
Or initiator/target -- naming is hard.

hughsie:
See https://github.com/fwupd/fwupd/pull/2244

Issue #2234, opened 2020-06-29
log messages have terminal control codes, even when destination is not a terminal


chadmiller:
**Describe the bug**
Syslogged messages have terminal color codes

**Steps to Reproduce**
Syslog file contains 

```
$ cat -v /var/log/syslog
fwupd[16964]: 21:30:39:0966 FuEngine Failed to load SMBIOS: Failed to open file M-bM-^@M-^\/sys/firmware/dmi/tables/smbios_entry_pointM-bM-^@M-^]: No such file or directory

$ grep "Failed to open file ./sys/firmware" ...

$ grep -Po "(?<=file ).*smbios.*(?=:)" /var/log/syslog |uniq |od -a
0000000 b nul fs / s y s / f i r m w a r e
0000020 / d m i / t a b l e s / s m b i
0000040 o s _ e n t r y _ p o i n t b nul
0000060 gs nl
0000062
```

**Expected behavior**

I expect clean lines in syslog. Things that are grepable and "locheck"-able and can go to a line printer. It shouldn't make assumptions about how syslog message will be displayed.

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version
```

```
Package: fwupd 1.3.9-4ubuntu0.1
ProcVersionSignature: User Name 5.4.0-1012.12-raspi 5.4.41
Uname: Linux 5.4.0-1012-raspi aarch64
ApportVersion: 2.20.11-0ubuntu27.2
Architecture: arm64
CasperMD5CheckResult: skip
Date: Thu Jun 25 21:38:21 2020
ImageMediaBuild: 20200423.1
ProcEnviron:
 TERM=xterm-256color
 PATH=(custom, no user)
 XDG_RUNTIME_DIR=<set>
 LANG=C.UTF-8
 SHELL=/bin/bash
SourcePackage: fwupd
UpgradeStatus: No upgrade log present (probably fresh install)
```

superm1:
I think this might actually be a glib bug.   Let me explain.  Fwupd tries to open the path, but doesn't do any processing on the error string for it.
https://github.com/fwupd/fwupd/blob/master/libfwupdplugin/fu-smbios.c#L257

If you go and look at `g_file_get_contents` you'll see there is a posix handler.
https://github.com/GNOME/glib/blob/3dec72b946a527f4b1f35262bddd4afb060409b7/glib/gfileutils.c#L1005

In the posix handler the quotations that are used look like they're UTF to me not ASCII:
[...]

PR #2239, opened 2020-06-30
Convert warning messages to ASCII when logged

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-debug.c
+++ b/src/fu-debug.c
@@ -95,11 +95,12 @@ fu_debug_handler_cb (const gchar *log_domain,
 
 	/* to file */
 	if (!self->console) {
+		g_autofree gchar *ascii_message = g_str_to_ascii (message, NULL);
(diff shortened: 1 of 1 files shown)


superm1:
CC @djcampello this might actually also clean up some of the fwupdtool output on ChromeOS, at least when run by the udev runner.  I know it had output some `???` characters previously.

hughsie:
Just so I understand this completely, why are we including UTF-8 in error messages anyway?

superm1:
> Just so I understand this completely, why are we including UTF-8 in error messages anyway?

It comes from glib.  Take at the linked issue to see my sleuthing (and prove it wrong if i'm wrong!): #2234

hughsie:
> It comes from glib

Ahh, that! I wondered if fwupd was injecting color codes or something!

PR #2241, opened 2020-06-30
Wip/hughsie/detect cet

hughsie:


hughsie:
Screenshot from 2020-06-30 20-34-06

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -73,6 +75,18 @@ fu_plugin_add_security_attrs_intel_cet (FuPlugin *plugin, FuSecurityAttrs *attrs
 		return;
 	}
 
+	/* check that userspace has been compiled for CET support */
+	if (!g_spawn_command_line_sync (FWUPD_LIBEXECDIR "/fwupd/fwupd-detect-cet",
```
superm1, line 79: I feel like this is asking for a fu_common_get_path type thing.

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -73,6 +75,18 @@ fu_plugin_add_security_attrs_intel_cet (FuPlugin *plugin, FuSecurityAttrs *attrs
...
+					NULL, NULL, &exit_status, &error_local)) {
+		g_warning ("failed to test CET: %s", error_local->message);
```
superm1, line 81: Shouldn't this be a failure or unused too?

plugins/cpu/fu-cpu-helper-cet.c
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
```
superm1, line 2: Presumably `H.J. Lu <hjl.tools@gmail.com>` should be in this?

meson.build
```diff
@@ -269,6 +270,9 @@ endif
 if cc.has_function('realpath')
   conf.set('HAVE_REALPATH', '1')
 endif
+if cc.has_function('sigaction')
```
superm1, line 273: Do we have any idea if current versions of clang and gcc both support this going in?

meson.build
```diff
@@ -125,6 +125,7 @@ test_link_args = [
   '-Wl,-z,relro',
   '-Wl,-z,defs',
   '-Wl,-z,now',
+  '-Wl,-z,ibt,-z,shstk',
```
superm1, line 128: Is this an implicit dependency on newer compilers?  Shouldn't this only go in if supported?

superm1:
You results - those are faked?  Or you have something with a working CET stack?  On my side, I have some hardware that should support it, but even with 5.8rc1 the flags are not exporting in `/proc/cpuinfo`.  Not sure if I was missing something on the kernel side..

plugins/cpu/fu-cpu-helper-cet.c
hjl-tools: Yes.  LGTM.

plugins/cpu/fu-plugin-cpu.c
hughsie: I wondered that too, but do we ever want to override it? Could it just be `g_build_filename (FWUPD_LIBEXECDIR, "fwupd", "fwupd-detect-cet", NULL)` instead?

plugins/cpu/fu-plugin-cpu.c
hughsie: I wasn't sure. This is more of a "distributor of fwupd f*cked up" situation rather than anything the hardware vendor did.

plugins/cpu/fu-cpu-helper-cet.c
hughsie: Agree, thanks.

meson.build
hughsie: Well, sigaction is a pretty old interface -- I'd be amazed if all libc's didn't support it.

meson.build
hughsie: Ya, it's only used if `cc.has_link_argument` passes. For me it passes already, which was slightly unexpected:

    Compiler for C supports link arguments -Wl,-z,ibt,-z,shstk: YES

hughsie:
> You results - those are faked

Completely faked, sorry I should have said that:

    -               return;
    +//             return;

plugins/cpu/fu-plugin-cpu.c
superm1: That seems fine to me.

meson.build
superm1: Well the compiler stuff for CET started getting put in a while back, so not too surprising.  OK.

superm1:
> > You results - those are faked
> 
> Completely faked, sorry I should have said that:
> 
> ```
> -               return;
> +//             return;
> ```

Oh OK.  So then can @hjl-tools  double check this?  Sounds like they have working hardware.

plugins/cpu/fu-plugin-cpu.c
superm1: Well I think it's missing a `return` then.  Right now if that fails to run, it will end up with a success attribute and enabled which is incorrect.

hjl-tools:
> > > You results - those are faked
> > 
> > 
> > Completely faked, sorry I should have said that:
> > ```
> > -               return;
> > +//             return;
> > ```
> 
> Oh OK. So then can @hjl-tools double check this? Sounds like they have working hardware.

Test needs a small change:

```
[hjl@gnu-tgl-1 tmp]$ cat x.c
#include <signal.h>
#include <stdlib.h>

extern void testfn1 (void);

static void
segfault_sigaction (int signal, siginfo_t *si, void *arg)
{
	/* CET did exactly as it should to protect the system */
	exit (0);
}

int
main (int argc, char *argv[])
{
	struct sigaction sa = { 0 };

	sigemptyset (&sa.sa_mask);
	sa.sa_sigaction = segfault_sigaction;
	sa.sa_flags = SA_SIGINFO;
	sigaction (SIGSEGV, &sa, NULL);

	testfn1 ();

	/* this means CET did not work */
	return 1;
}
[hjl@gnu-tgl-1 tmp]$ cat y.c
static void
testfn_fptr (void)
{
}

static void
__attribute__ ((noinline, noclone))
testfn_call_fptr (void (*func) (void))
{
	func ();
}

void
testfn1 (void)
{
	testfn_call_fptr (testfn_fptr);
}
[hjl@gnu-tgl-1 tmp]$ gcc -c -fcf-protection x.c -g
[hjl@gnu-tgl-1 tmp]$ gcc -Wl,-z,shstk,-z,ibt x.o y.c -g
[hjl@gnu-tgl-1 tmp]$ ./a.out 
[hjl@gnu-tgl-1 tmp]$ echo $?
0
```
main must be compiled with -fcf-protection.  Otherwise main () will get a segfault due
to missing ENDBR.

superm1:
> > > > You results - those are faked
> > > 
> > > 
> > > Completely faked, sorry I should have said that:
> > > ```
> > > -               return;
> > > +//             return;
> > > ```
> > 
> > 
> > Oh OK. So then can @hjl-tools double check this? Sounds like they have working hardware.
> 
> Test needs a small change:
> 
> ```
> [hjl@gnu-tgl-1 tmp]$ cat x.c
> #include <signal.h>
> #include <stdlib.h>
> 
> extern void testfn1 (void);
> 
> static void
> segfault_sigaction (int signal, siginfo_t *si, void *arg)
> {
> 	/* CET did exactly as it should to protect the system */
> 	exit (0);
> }
> 
> int
> main (int argc, char *argv[])
> {
> 	struct sigaction sa = { 0 };
> 
> 	sigemptyset (&sa.sa_mask);
> 	sa.sa_sigaction = segfault_sigaction;
> 	sa.sa_flags = SA_SIGINFO;
> 	sigaction (SIGSEGV, &sa, NULL);
> 
> 	testfn1 ();
> 
> 	/* this means CET did not work */
> 	return 1;
> }
> [hjl@gnu-tgl-1 tmp]$ cat y.c
> static void
> testfn_fptr (void)
> {
> }
> 
> static void
> __attribute__ ((noinline, noclone))
> testfn_call_fptr (void (*func) (void))
> {
> 	func ();
> }
> 
> void
> testfn1 (void)
> {
> 	testfn_call_fptr (testfn_fptr);
> }
> [hjl@gnu-tgl-1 tmp]$ gcc -c -fcf-protection x.c -g
> [hjl@gnu-tgl-1 tmp]$ gcc -Wl,-z,shstk,-z,ibt x.o y.c -g
> [hjl@gnu-tgl-1 tmp]$ ./a.out 
> [hjl@gnu-tgl-1 tmp]$ echo $?
> 0
> ```
> 
> main must be compiled with -fcf-protection. Otherwise main () will get a segfault due
> to missing ENDBR.

Thanks!  By chance can you point out what exactly is needed for kernel support of this?  Are there pieces not in 5.8 that need to be patched in?  And can the rest of userland operate fine without shadow stack support (just no CET protection in those parts of the stack)?

hjl-tools:
> Thanks! By chance can you point out what exactly is needed for kernel support of this? Are there pieces not in 5.8 that need to be patched in? And can the rest of userland operate fine without shadow stack support (just no CET protection in those parts of the stack)?

Here is the CET enabled 5.7.6 kernel tree:

https://github.com/hjl-tools/linux/commits/hjl/cet/5.7.6

I have been running the same CET kernel binary on all my machines.

plugins/cpu/fu-plugin-cpu.c
hughsie: Right, agreed, fixed. Thanks.

plugins/cpu/fu-plugin-cpu.c
hughsie: Fixed.

hughsie:
> main must be compiled with -fcf-protection

@hjl-tools can we just compile **all** of fwupd with `-fcf-protection=full`?

hjl-tools:
> > main must be compiled with -fcf-protection
> 
> @hjl-tools can we just compile **all** of fwupd with `-fcf-protection=full`?

Yes.

hughsie:
> main () must compiled separately with -fcf-protection and testfn1 () must be compiled in a different file with
-fcf-protection=none

I'm a bit confused; I thought fcf-protection would essentially "protect" the binary when CET is available? If CET is working the test binary should segfault, no?

hjl-tools:
> > main () must compiled separately with -fcf-protection and testfn1 () must be compiled in a different file with
> > -fcf-protection=none
> 
> I'm a bit confused; I thought fcf-protection would essentially "protect" the binary when CET is available? If CET is working the test binary should segfault, no?

On Linux, main () is called via indirect branch.  When CET is enabled, if main () doesn't start with ENDBR, you get a
segfault on CET machine BEFORE you have a chance to install a signal handler for segfault.

hughsie:
> you get a segfault on CET machine BEFORE you have a chance to install a signal handler for segfault

But I'm confused; the test binary sets up the `sigaction(SIGSEGV)` before calling `testfn1()`

hjl-tools:
> > you get a segfault on CET machine BEFORE you have a chance to install a signal handler for segfault
> 
> But I'm confused; the test binary sets up the `sigaction(SIGSEGV)` before calling `testfn1()`

You get a segfault at entrance of main () before `sigaction(SIGSEGV)` is called.

hughsie:
> You get a segfault at entrance of main () before sigaction(SIGSEGV) is called

Ahh, thanks. I'll see what I can do.

hughsie:
@hjl-tools something like that ^^^?

hjl-tools:
> @hjl-tools something like that ^^^?

Looks right.

hughsie:
> Looks right.

superm1:
> > Looks right.
> 
> If I build a Fedora package with this change would it be possible to test it for me on a machine with functioning CET? @hjl-tools Thanks!

I tried, but after booting that CET enabled kernel from @hjl-tools tree and these config options added:
```
CONFIG_AS_HAS_SHADOW_STACK=y
CONFIG_X86_INTEL_CET=y
```

I'm getting into a no boot scenario.  Pretty much all of the services that try to start up fail.  Even in the kernel's single user mode I end up with this:
```
$ fwupdtool security --force
: legacy bitmap isn't available
```

hjl-tools:
> > > Looks right.
> > 
> > 
> > If I build a Fedora package with this change would it be possible to test it for me on a machine with functioning CET? @hjl-tools Thanks!
> 
> I tried, but after booting that CET enabled kernel from @hjl-tools tree and these config options added:
> 
> ```
> CONFIG_AS_HAS_SHADOW_STACK=y
> CONFIG_X86_INTEL_CET=y
> ```
> 
> I'm getting into a no boot scenario. Pretty much all of the services that try to start up fail. Even in the kernel's single user mode I end up with this:
> 
> ```
> $ fwupdtool security --force
> : legacy bitmap isn't available
> ```

Glibc 2.32 is needed for the CET kernel.  Which Linux distro are you using? I have updated glibc rpms for Fedora 32.

superm1:
> > > > Looks right.
> > > 
> > > 
> > > If I build a Fedora package with this change would it be possible to test it for me on a machine with functioning CET? @hjl-tools Thanks!
> > 
> > 
> > I tried, but after booting that CET enabled kernel from @hjl-tools tree and these config options added:
> > ```
> > CONFIG_AS_HAS_SHADOW_STACK=y
> > CONFIG_X86_INTEL_CET=y
> > ```
> > 
> > 
> > I'm getting into a no boot scenario. Pretty much all of the services that try to start up fail. Even in the kernel's single user mode I end up with this:
> > ```
> > $ fwupdtool security --force
> > : legacy bitmap isn't available
> > ```
> 
> Glibc 2.32 is needed for the CET kernel. Which Linux distro are you using? I have updated glibc rpms for Fedora 32.

Ah... I have 2.31.  I was on Ubuntu 20.04, but I just upgraded to a daily snapshot of Ubuntu 20.10 and they still don't have 2.32.  Let me see if there is a PPA or something similar I can use.

superm1:
> > > > > Looks right.
> > > > 
> > > > 
> > > > If I build a Fedora package with this change would it be possible to test it for me on a machine with functioning CET? @hjl-tools Thanks!
> > > 
> > > 
> > > I tried, but after booting that CET enabled kernel from @hjl-tools tree and these config options added:
> > > ```
> > > CONFIG_AS_HAS_SHADOW_STACK=y
> > > CONFIG_X86_INTEL_CET=y
> > > ```
> > > 
> > > 
> > > I'm getting into a no boot scenario. Pretty much all of the services that try to start up fail. Even in the kernel's single user mode I end up with this:
> > > ```
> > > $ fwupdtool security --force
> > > : legacy bitmap isn't available
> > > ```
> > 
> > 
> > Glibc 2.32 is needed for the CET kernel. Which Linux distro are you using? I have updated glibc rpms for Fedora 32.
> 
> Ah... I have 2.31. I was on Ubuntu 20.04, but I just upgraded to a daily snapshot of Ubuntu 20.10 and they still don't have 2.32. Let me see if there is a PPA or something similar I can use.

Wait a minute..  2.32 isn't released yet.  So I guess it's a series of patches that were post 2.31 but not yet in a released version?
https://www.gnu.org/software/libc/

superm1:
Ah here we go.  It looks like the patches must have been backported into glibc in 20.10's package.  Don't mind the IOMMU errors, there is a problem here with this CPU stepping.

```
$ sudo fwupdtool security --force
17:34:39:0205 FuPluginAcpiDmar     failed to load /sys/firmware/acpi/tables/DMAR: Failed to open file “/sys/firmware/acpi/tables/DMAR”: No such file or directory
Host Security ID: HSI:0+! (v1.5.0)
17:34:39:0209 FuPluginAcpiDmar     failed to load /sys/firmware/acpi/tables/DMAR: Failed to open file “/sys/firmware/acpi/tables/DMAR”: No such file or directory

HSI-1
✔ MEI override strap:            Locked
✔ MEI version:                   Valid
✔ SPI BIOS region:               Locked
✔ SPI lock:                      Enabled
✔ SPI write:                     Disabled
✔ TPM v2.0:                      Found
✔ UEFI dbx:                      Found
✘ MEI manufacturing mode:        Unlocked

HSI-2
✘ IOMMU:                         Not found
✘ TPM PCR0 reconstruction:       Invalid

HSI-3
✔ Intel CET:                     Enabled
✔ Suspend-to-idle:               Enabled
✔ Suspend-to-ram:                Disabled
✘ Pre-boot DMA protection:       Invalid

HSI-4
✔ Encrypted RAM:                 Enabled
✔ Intel SMAP:                    Enabled

Runtime Suffix -U
✘ Firmware updates:              Not supported

Runtime Suffix -A
✘ Firmware attestation:          Not supported

Runtime Suffix -!
✔ Linux kernel:                  Untainted
✔ fwupd plugins:                 Untainted
✘ Linux kernel lockdown:         Disabled
✘ Linux swap:                    Unencrypted
✘ UEFI secure boot:              Disabled
```

hughsie:
> ✘ MEI manufacturing mode:        Unlocked

offtopic, but is that actually plausible? e.g. prerelease silicon?

> ✔ Intel CET:                     Enabled

With this branch?! the helper actually worked?

plugins/cpu/fu-plugin-cpu.c
```diff
@@ -73,6 +76,17 @@ fu_plugin_add_security_attrs_intel_cet (FuPlugin *plugin, FuSecurityAttrs *attrs
...
 
+	/* check that userspace has been compiled for CET support */
+	toolfn = g_build_filename (FWUPD_LIBEXECDIR, "fwupd", "fwupd-detect-cet", NULL);
+	if (!g_spawn_command_line_sync (toolfn, NULL, NULL, &exit_status, &error_local)) {
+		g_warning ("failed to test CET: %s", error_local->message);
+		return;
+	}
+	if (!g_spawn_check_exit_status (exit_status, &error_local)) {
+		g_debug ("CET does not function, not supported: %s", error_local->message);
+		fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_UNUSED);
```
superm1, line 87: Given how much it actually takes to make a CET measurement work (correct kernel patches, correct glibc, and fwupd supporting it) i'm not sure the unused flag makes any sense.

If I had the correct kernel support that the flags showed up (`ibt` and `shstk`) but hte missing glibc it wouldn't have booted.  If I had the correct glibc but fwupd was not compiled with support, then I wouldn't have this check in the first place.

hjl-tools:
> Ah here we go. It looks like the patches must have been backported into glibc in 20.10's package. Don't mind the IOMMU errors, there is a problem here with this CPU stepping.
> 

We have requested Canonical to backport CET fix to Ubuntu 20.10.  Does CET kernel
work with the updated Ubuntu now?

superm1:
> offtopic, but is that actually plausible? e.g. prerelease silicon?

It is pre-release silicon yes.

>With this branch?! the helper actually worked?

Yes!

```
$ fwupdtool --version
This program may only work correctly as root
client version: 1.4.1-229-gfc6e29dc
compile-time dependency versions
        gusb:   0.3.4
        efivar: 37
```

>We have requested Canonical to backport CET fix to Ubuntu 20.10. Does CET kernel
work with the updated Ubuntu now?

Yes it does.

plugins/cpu/fu-plugin-cpu.c
hughsie: > i'm not sure the unused flag makes any sense

My logic for the extra data was to signify if the support was enabled in the OS; as I understand it you would still have cet in cpuinfo, but if glibc wasn't compiled with cet then it would essentially be unused. Is that wrong?

superm1: I could only boot into single user mode and a ton of my binaries wouldn't execute without it being in glibc.

hughsie: So, forgive the stupid question, does that mean a CPU with CET enabled is unable to boot older distros like Ubuntu LTS?

superm1: Well older distros that don't have CET enabled in the kernel won't export `ibt` or `shstk` in `/proc/cpuinfo`.  It was only after I booted the CET enabled kernel that those start to export.

hughsie: So maybe we don't need this PR at all? i.e. stick to what we were doing before?

superm1: Interesting thought.  I really like a test that the shadow stack really is working though.

hughsie: Maybe we just fail the check with a runtime `!`?

hjl-tools: Majority of applications are compatible with shadow stack.   Here is an example of shadow stack:

https://gitlab.com/cet-software/cet-smoke-test/-/tree/master/violations

hughsie: >  I really like a test that the shadow stack really is working though

Yes, agreed. I think we either need the "unused" thing or a different runtime attr for CET. I'm happy to do either. The different attributes are already being suggested for DCI lock/enabled so maybe this isn't so bad.

superm1: You don't think setting CET to invalid would be appropriate in this case?

hughsie: No, invalid would fail the HSI test which makes it look like the OEM is at fault, when really it's the users system that is screwed up or a mixture of old and new binaries.

hughsie: I've updated the branch with a much better check, splitting these out like we did the others. It's also better as now CET being available but not being used by the OS is a *runtime* failure, not something that affects the HSI value.

plugins/cpu/fu-cpu-helper-cet.c
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <signal.h>
+#include <stdlib.h>
+
+#include "fu-cpu-helper-cet-helper.h"
```
superm1, line 13: cpu-helper-cet-helper?  I think this should have a better name

hughsie: fu-cpu-helper-cet-common?

superm1: Yeah, just avoid helper-*-helper

PR #2243, opened 2020-07-01
ccgx: remove verify flag in plugin

chlee75:
Type of pull request:
- [ ] Code fix
   Remove Verify support in Lenvo and HP Dock. Those docks do not support read-firmware.


chlee75:
@hughsie 
I disable "FWUPD_DEVICE_FLAG_CAN_VERIFY_IMAGE" in Lenovo and Gen2 Dock.
"Verify" button disabled in gnome-firmware when I connected Lenovo Gen2 Dock or HP Dock at first.
But "Verify" button enabled again after I update firmware.
Is it bug in gnome-firmware?

PR #2244, opened 2020-07-01
Change all instances of master/slave to initiator/target

hughsie:
This makes perfect sense, because the 'initiator' starts the transaction and
the 'target' is the addressee of the transaction. Even the I²C spec defines the
'master' as 'initiating' the transaction.

This is the same nomenclature now used by the Glasgow project too.


bonzini:
SPI is considering usage of Controller and Peripheral, probably should use the same for I2C?

whitequark:
> SPI is considering usage of Controller and Peripheral, probably should use the same for I2C?

These are different cases--with SPI the direction is fixed, with I2C not so much as it has the "multi-master" capability.

PR #2246, opened 2020-07-01
Wip/hughsie/fwupd security attr metadata

hughsie:
Screenshot from 2020-07-01 19-13-03


hughsie:
> superm1 approved these changes

Did. Not. Expect. :)

PR #2248, opened 2020-07-01
Group devices in get-updates output

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1776,3 +1776,23 @@ fu_util_send_report (SoupSession *soup_session,
 	/* success */
 	return TRUE;
 }
+
+gint
+fu_util_sort_devices_by_flags_cb (gconstpointer a, gconstpointer b)
+{
+	FuDevice *dev_a = *((FuDevice **) a);
+	FuDevice *dev_b = *((FuDevice **) b);
+
+	if ((!fu_device_has_flag (dev_a, FWUPD_DEVICE_FLAG_UPDATABLE) &&
+	     fu_device_has_flag (dev_b, FWUPD_DEVICE_FLAG_UPDATABLE)) ||
+	    (!fu_device_has_flag (dev_a, FWUPD_DEVICE_FLAG_SUPPORTED) &&
+	     fu_device_has_flag (dev_b, FWUPD_DEVICE_FLAG_SUPPORTED)))
+		return -1;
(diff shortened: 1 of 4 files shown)


superm1:
Oh and sample output on my local system.
```
$ fwupdmgr update
Devices with no available firmware updates:
 • Thunderbolt Controller
 • PC601 NVMe SK hynix 512GB
 • Touchpad
Devices with the latest available firmware version:
 • Package level of Dell dock
 • RTS5413 in Dell dock
 • RTS5487 in Dell dock
 • VMM5331 in Dell dock
 • WD19
 • System Firmware
```

src/fu-engine.c
```diff
@@ -3889,6 +3889,18 @@ fu_engine_sort_devices_by_priority_name (gconstpointer a, gconstpointer b)
...
+	    fu_device_has_flag (dev_b, FWUPD_DEVICE_FLAG_UPDATABLE)) ||
+	    (!fu_device_has_flag (dev_a, FWUPD_DEVICE_FLAG_SUPPORTED) &&
+	    fu_device_has_flag (dev_b, FWUPD_DEVICE_FLAG_SUPPORTED)))
+		return -1;
+	if ((fu_device_has_flag (dev_a, FWUPD_DEVICE_FLAG_UPDATABLE) &&
+	    !fu_device_has_flag (dev_b, FWUPD_DEVICE_FLAG_UPDATABLE)) ||
+	    (fu_device_has_flag (dev_a, FWUPD_DEVICE_FLAG_SUPPORTED) &&
+	    !fu_device_has_flag (dev_b, FWUPD_DEVICE_FLAG_SUPPORTED)))
+		return 1;
+	/* then support by priority, so clients can show parents */
```
hughsie, line 3903: i'm not super keen on this being in the daemon; at the moment we can say the devices are ordered "by priority then name" but with this change it all gets a bit more fuzzy.

superm1: Adjusted

src/fu-util-common.c
```diff
@@ -1776,3 +1776,24 @@ fu_util_send_report (SoupSession *soup_session,
 	/* success */
 	return TRUE;
 }
+
+gint
+fu_util_sort_devices_by_flags (gconstpointer a, gconstpointer b)
```
hughsie, line 1781: I think the convention for this kind of sort function is for a _cb suffix, so `fu_util_sort_devices_by_flags_cb`

PR #2250, opened 2020-07-02
Wip/hughsie/mei more more more

hughsie:


superm1:
The one thing I would caution is that the definition of data in these registers might end up being silicon dependent, but that might be a bridge to be crossed when arrived.

hughsie:
> these registers might end up being silicon dependent

I think they very much are; on some platforms the flags were `reserved` and in other platforms they were defined. Luckily there wasn't any re-use or overlap between platforms.

Out of interest, did you try this on any modernish platforms? Verified boot has been around for a while now and so I guess most/all supported Dell stuff should be valid.

Also, do you have any insight on what `ProtectBiosEnv` or `FpfSocLock` is, or what they do?

superm1:
> > these registers might end up being silicon dependent
> 
> I think they very much are; on some platforms the flags were `reserved` and in other platforms they were defined. Luckily there wasn't any re-use or overlap between platforms.
> 
> Out of interest, did you try this on any modernish platforms? Verified boot has been around for a while now and so I guess most/all supported Dell stuff should be valid.

Yeah it works properly on the CML system I have on hand.  I didn't check anything else though.

> 
> Also, do you have any insight on what `ProtectBiosEnv` or `FpfSocLock` is, or what they do?

No, sorry I don't.

PR #2251, opened 2020-07-03
cros-ec: cros-ec-usb-device write firmware

bleungatchromium:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

Implement the pieces necessary to write firmware to cros-ec usb devices utilizing Protocol 6 or later.

### DIFF
--- a/plugins/cros-ec/fu-cros-ec-firmware.c
+++ b/plugins/cros-ec/fu-cros-ec-firmware.c
@@ -13,9 +13,9 @@
 #define MAXSECTIONS	2
 
 struct _FuCrosEcFirmware {
-	FuFmapFirmware			parent_instance;
-	struct cros_ec_version		version;
-	FuCrosEcFirmwareSection		sections[MAXSECTIONS];
+	FuFmapFirmware			 parent_instance;
+	struct cros_ec_version		 version;
+	GPtrArray			*sections;
 };
 
 G_DEFINE_TYPE (FuCrosEcFirmware, fu_cros_ec_firmware, FU_TYPE_FMAP_FIRMWARE)
@@ -31,12 +31,9 @@ fu_cros_ec_firmware_parse (FuFirmware *firmware,
 	FuCrosEcFirmware *self = FU_CROS_EC_FIRMWARE (firmware);
 	FuFirmware *fmap_firmware = FU_FIRMWARE (firmware);
 
-	self->sections[0].name = "RO";
-	self->sections[1].name = "RW";
-
-	for (gsize i = 0; i < G_N_ELEMENTS (self->sections); i++) {
+	for (gsize i = 0; i < self->sections->len; i++) {
 		gboolean rw = FALSE;
-		FuCrosEcFirmwareSection *section = &self->sections[i];
+		FuCrosEcFirmwareSection *section = g_ptr_array_index (self->sections, i);
 		const gchar *fmap_name;
 		const gchar *fmap_fwid_name;
 		g_autoptr(FuFirmwareImage) img = NULL;
@@ -120,13 +117,32 @@ fu_cros_ec_firmware_parse (FuFirmware *firmware,
 static void
 fu_cros_ec_firmware_init (FuCrosEcFirmware *self)
 {
+	FuCrosEcFirmwareSection *section;
+
+	self->sections = g_ptr_array_new_with_free_func (g_free);
+	section = g_new0 (FuCrosEcFirmwareSection, 1);
+	section->name = "RO";
+	g_ptr_array_add (self->sections, section);
+	section = g_new0 (FuCrosEcFirmwareSection, 1);
+	section->name = "RW";
+	g_ptr_array_add (self->sections, section);
+}
+
+static void
+fu_cros_ec_firmware_finalize (GObject *object)
+{
+	FuCrosEcFirmware *self = FU_CROS_EC_FIRMWARE (object);
+	g_ptr_array_free (self->sections, TRUE);
+	G_OBJECT_CLASS (fu_cros_ec_firmware_parent_class)->finalize (object);
 }
 
 static void
 fu_cros_ec_firmware_class_init (FuCrosEcFirmwareClass *klass)
 {
+	GObjectClass *object_class = G_OBJECT_CLASS (klass);
 	FuFmapFirmwareClass *klass_firmware = FU_FMAP_FIRMWARE_CLASS (klass);
 	klass_firmware->parse = fu_cros_ec_firmware_parse;
+	object_class->finalize = fu_cros_ec_firmware_finalize;
 }
 
 FuFirmware *
--- a/plugins/cros-ec/fu-cros-ec-firmware.c
+++ b/plugins/cros-ec/fu-cros-ec-firmware.c
@@ -51,6 +51,12 @@ fu_cros_ec_firmware_pick_sections (FuCrosEcFirmware *self,
 	return TRUE;
 }
 
+GPtrArray *
+fu_cros_ec_firmware_get_sections (FuCrosEcFirmware *self)
+{
+	return self->sections;
+}
+
 static gboolean
 fu_cros_ec_firmware_parse (FuFirmware *firmware,
 			   GBytes *fw,
--- a/plugins/cros-ec/fu-cros-ec-firmware.c
+++ b/plugins/cros-ec/fu-cros-ec-firmware.c
@@ -20,6 +20,37 @@ struct _FuCrosEcFirmware {
 
 G_DEFINE_TYPE (FuCrosEcFirmware, fu_cros_ec_firmware, FU_TYPE_FMAP_FIRMWARE)
 
+gboolean
+fu_cros_ec_firmware_pick_sections (FuCrosEcFirmware *self,
+				   guint32 writeable_offset,
+				   GError **error)
+{
+	gboolean found = FALSE;
+
+	for (gsize i = 0; i < self->sections->len; i++) {
+		FuCrosEcFirmwareSection *section = g_ptr_array_index (self->sections, i);
+		guint32 offset = section->offset;
+
+		if (offset != writeable_offset)
+			continue;
+
+		section->ustatus = FU_CROS_EC_FW_NEEDED;
+		found = TRUE;
+	}
+
+	if (!found) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_DATA,
+			     "no writeable section found with offset: 0x%x",
+			     writeable_offset);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
 static gboolean
 fu_cros_ec_firmware_parse (FuFirmware *firmware,
 			   GBytes *fw,
--- a/plugins/cros-ec/fu-cros-ec-firmware.c
+++ b/plugins/cros-ec/fu-cros-ec-firmware.c
@@ -132,6 +132,7 @@ fu_cros_ec_firmware_parse (FuFirmware *firmware,
 		section->offset = fu_firmware_image_get_addr (img);
 		section->size = g_bytes_get_size (payload_bytes);
 		fu_firmware_image_set_version (img, section->version);
+		section->image_idx = fu_firmware_image_get_idx (img);
 
 		if (rw) {
 			if (!fu_cros_ec_parse_version (section->version,
--- a/plugins/cros-ec/fu-cros-ec-firmware.h
+++ b/plugins/cros-ec/fu-cros-ec-firmware.h
@@ -38,4 +38,7 @@ typedef struct {
 	guint32				key_version;
 } FuCrosEcFirmwareSection;
 
+gboolean			 fu_cros_ec_firmware_pick_sections	(FuCrosEcFirmware *self,
+									 guint32 writeable_offset,
+									 GError **error);
 FuFirmware			*fu_cros_ec_firmware_new		(void);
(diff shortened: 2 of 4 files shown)


plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+fu_cros_ec_usb_device_send_done (FuDevice *device, GError **error)
+{
+	guint32 out = GUINT32_TO_BE (UPDATE_DONE);
+
+	/* send stop request, ignoring reply */
+	if (!fu_cros_ec_usb_device_do_xfer (FU_CROS_EC_USB_DEVICE (device),
+					    (guint8 *)&out, sizeof (out),
+					    (guint8 *)&out, 1,
+					    FALSE, NULL, error)) {
+		g_debug ("error on transfer of done. clearing error.\n");
```
superm1, line 500: you shouldn't need to add a newline to the end here

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+		g_clear_error (error);
```
superm1, line 501: If you're just going to be throwing away the error and always returning `TRUE` for this function, I would argue there is no point in passing the `GError` parameter in the first place.  Instead use an `error_local` for your call to `fu_cros_ec_usb_device_do_xfer` and output the error that passed up into `g_debug` message.

If you truly don't care what the error was, pass `NULL` into `fu_cros_ec_usb_device_do_xfer` though.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -316,9 +573,35 @@ fu_cros_ec_usb_device_close (FuUsbDevice *device, GError **error)
...
+		g_prefix_error (error, "failed to pick sections: ");
+		return NULL;
+	}
+	return g_steal_pointer (&firmware);
+}
+
 static void
 fu_cros_ec_usb_device_init (FuCrosEcUsbDevice *device)
 {
+	fu_device_set_protocol (FU_DEVICE (device), "com.google.usb.crosec");
```
superm1, line 604: You should add this to `README.md` too in this PR.

plugins/cros-ec/fu-cros-ec-firmware.h
```diff
@@ -36,6 +36,11 @@ typedef struct {
 	gchar				version[FU_FMAP_FIRMWARE_STRLEN];
 	gint32				rollback;
 	guint32				key_version;
+	guint64				image_idx;
 } FuCrosEcFirmwareSection;
 
+gboolean			 fu_cros_ec_firmware_pick_sections	(FuCrosEcFirmware *self,
+									 guint32 writeable_offset,
+									 GError **error);
+GArray				*fu_cros_ec_firmware_get_sections	(FuCrosEcFirmware *self);
```
superm1, line 45: Is the whitespace here lined up?  (It looks off; but it might just be Github rendering it poorly - I didn't look in a local checkout).

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_cros_ec_usb_device_transfer_section (FuDevice *device,
+					FuFirmware *firmware,
+					FuCrosEcFirmwareSection *section,
+					gboolean smart_update,
```
superm1, line 408: It seems to me that you only use this function in one place and always send `smart_update` as `TRUE`.  Worth dropping?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+					GError **error)
+{
+	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
+	const guint8 * data_ptr = NULL;
+	guint32 section_addr = 0;
+	gsize data_len = 0;
+	gsize offset = 0;
+	g_autoptr(GBytes) img_bytes = NULL;
+
+	if (section == NULL) {
```
superm1, line 418: This smells like it should be a `g_return_val_if_fail` type of assertion.  If this was `NULL` it's probably a coding error.

plugins/cros-ec/fu-cros-ec-firmware.c
```diff
@@ -13,13 +13,52 @@
 #define MAXSECTIONS	2
 
 struct _FuCrosEcFirmware {
-	FuFmapFirmware			parent_instance;
-	struct cros_ec_version		version;
-	FuCrosEcFirmwareSection		sections[MAXSECTIONS];
+	FuFmapFirmware			 parent_instance;
+	struct cros_ec_version		 version;
+	GArray				*sections;
```
hughsie, line 18: I think a GPtrArray might be more useful here, no?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -38,6 +44,13 @@ typedef union _START_RESP {
 	guint32 legacy_resp;
 } START_RESP;
 
+typedef struct _fu_cros_ec_usb_block_info {
```
hughsie, line 47: The `_fu_cros_ec_usb_block_info` is redundant, no?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -38,6 +44,13 @@ typedef union _START_RESP {
...
+	struct update_frame_header ufh;
+	GBytes *image_bytes;
+	gsize offset;
+	gsize payload_size;
+} fu_cros_ec_usb_block_info;
```
hughsie, line 52: CamelCase please

plugins/cros-ec/fu-cros-ec-usb-device.c
hughsie: or the `.`

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -271,6 +284,8 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
 		return FALSE;
 	}
 
+	self->writeable_offset =
+		GUINT32_FROM_BE (start_resp.rpdu.common.offset);
```
hughsie, line 288: this can be allone line if you want, I'm not a nazi with line width.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+	g_autoptr(GBytes) block_bytes = NULL;
+	g_autoptr(GPtrArray) chunks = NULL;
+
+	if (block_info == NULL) {
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_INVALID_DATA,
+				     "no block_info provided");
+		return FALSE;
+	}
```
hughsie, line 335: `g_return_val_if_fail`?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+			     G_IO_ERROR_INVALID_DATA,
+			     "image and section sizes do not match: image = %"
+			     G_GSIZE_FORMAT " bytes vs section size = %"
+			     G_GSIZE_FORMAT " bytes",
+			     data_len, section->size);
+		return FALSE;
+	}
+
+	if (smart_update) {
+		while (data_len != 0 && (data_ptr[data_len -1] == 0xff))
```
hughsie, line 448: `data_len - 1` pls.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -299,6 +314,248 @@ fu_cros_ec_usb_device_setup (FuDevice *device, GError **error)
...
+								     section,
+								     TRUE,
+								     error)) {
+				return FALSE;
+			}
+			num_txed_sections++;
+		}
+	}
+	/* send done */
+	fu_cros_ec_usb_device_send_done (device, error);
```
hughsie, line 545: if you're setting error you need to return if this fails.

plugins/cros-ec/fu-cros-ec-usb-device.c
hughsie: And we'll need to add to the LVFS at some point. On that, is `com.google.usb.crosec` *signed* i.e. with some strong public/private method?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -353,9 +636,11 @@ fu_cros_ec_usb_device_class_init (FuCrosEcUsbDeviceClass *klass)
 {
 	FuDeviceClass *klass_device = FU_DEVICE_CLASS (klass);
 	FuUsbDeviceClass *klass_usb_device = FU_USB_DEVICE_CLASS (klass);
+	klass_device->write_firmware = fu_cros_ec_usb_device_write_firmware;
 	klass_device->setup = fu_cros_ec_usb_device_setup;
 	klass_device->to_string = fu_cros_ec_usb_device_to_string;
 	klass_usb_device->open = fu_cros_ec_usb_device_open;
 	klass_usb_device->probe = fu_cros_ec_usb_device_probe;
 	klass_usb_device->close = fu_cros_ec_usb_device_close;
+	klass_device->prepare_firmware = fu_cros_ec_usb_device_prepare_firmware;
```
hughsie, line 645: a total nit, but please order the `klass_device` stuff together

plugins/cros-ec/fu-cros-ec-firmware.h
bleungatchromium: Hi @superm1 

It looks lined up to me in my editor, and when you look at it in non diff mode: 
https://github.com/fwupd/fwupd/blob/7527538bd74e0aa2ca1c6a51c3787bb8991e92bd/plugins/cros-ec/fu-cros-ec-firmware.h

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Cleaned this up. Error no longer required for this.

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-firmware.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Added to the README.md.

Hi @hughsie:
I'll tackle getting two versions of servo_micro firmware added to LVFS next.

I have to check with someone who knows more about the signing question. I suspect that some of our end products are signed, but possibly not all.

hughsie: > I suspect that some of our end products are signed, but possibly not all.

Can you tell from asking the hardware? For logitech we actually have two defined protocols, `com.logitech.unifying` and `com.logitech.unifyingsigned` -- that allows us to show the nice green tick for firmware updates that are signed and the red cross for ones that are not.

superm1: And if not from the hardware, I would think from looking at that payload structure you should be able to tell?

bleungatchromium: Hi @hughsie and @superm1 

So the servo micro board is NOT signed, but some commercially available products that use the cros-ec usb update method are signed.

I'm still investigating what steps I have to take to support signed updates.

What do you recommend for this patch series?

superm1: My thought is to merge this as is right now. On lvfs mark that protocol as unsigned. When it comes time to support a device with signed payloads modify this plugin to resolve the differences and use a new protocol that we mark as signed.

From my perspective this PR looks great now 🙌 (and I suppose is the last one in this series for this device).

hughsie: Okay, I've added `com.google.usb.crosec` as unsigned on the LVFS for now.

PR #2252, opened 2020-07-03
Add support for Thunderbolt retimers

superm1:
(I don't actually have hardware that can test this, so it's best effort)

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -1241,6 +1241,28 @@ fu_udev_device_write_sysfs (FuUdevDevice *self, const gchar *attribute,
 #endif
 }
 
+/**
+ * fu_udev_device_get_devtype
+ * @self: A #FuUdevDevice
+ *
+ * Returns the Udev device type
+ *
+ * Returns: device type specified in the uevent
+ *
+ * Since: 1.4.5
+ **/
+const gchar *
+fu_udev_device_get_devtype (FuUdevDevice *self)
+{
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+
+	return g_udev_device_get_devtype (priv->udev_device);
+#else
+	return NULL;
+#endif
+}
+
 static void
 fu_udev_device_get_property (GObject *object, guint prop_id,
 			    GValue *value, GParamSpec *pspec)
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -21,9 +21,15 @@
 #include "fu-thunderbolt-firmware.h"
 #include "fu-thunderbolt-firmware-update.h"
 
+typedef enum {
+	FU_THUNDERBOLT_DEVICE_TYPE_DEVICE_CONTROLLER,
+	FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER,
+	FU_THUNDERBOLT_DEVICE_TYPE_RETIMER
+} FuThunderboltDeviceType;
+
 struct _FuThunderboltDevice {
 	FuUdevDevice		 parent_instance;
-	gboolean		 host;
+	FuThunderboltDeviceType	 device_type;
 	gboolean		 safe_mode;
 	gboolean		 is_native;
 	guint16			 gen;
@@ -160,7 +166,7 @@ static void
 fu_thunderbolt_device_check_safe_mode (FuThunderboltDevice *self)
 {
 	/* failed to read, for host check for safe mode */
-	if (!self->host || self->gen >= 4)
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_DEVICE_CONTROLLER)
 		return;
 	g_warning ("%s is in safe mode --  VID/DID will "
 		   "need to be set by another plugin",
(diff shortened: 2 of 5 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -1241,6 +1241,28 @@ fu_udev_device_write_sysfs (FuUdevDevice *self, const gchar *attribute,
...
 
+/**
+ * fu_udev_device_get_devtype
+ * @self: A #FuUdevDevice
+ *
+ * Returns the Udev device type
+ *
+ * Returns: device type specified in the uevent
+ *
+ * Since: 1.5.0
```
hughsie, line 1252: Do we want this to go into 1.4.5 eventually too?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -331,14 +338,76 @@ fu_thunderbolt_device_setup (FuDevice *device, GError **error)
...
+	}
+
+	instance = g_strdup_printf ("TBT-%04x%04x-retimer%s",
+				    (guint) vid,
+				    (guint) did,
+				    idx);
+	fu_device_add_instance_id (device, instance);
+
+	/* hardcoded for now - we don't recognize anyone else yet */
+	fu_device_set_vendor (device, "Intel");
```
hughsie, line 386: I think this should already be set from the `ID_VENDOR_FROM_DATABASE` key in `FuUdevDevice->probe`, no?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -331,14 +338,76 @@ fu_thunderbolt_device_setup (FuDevice *device, GError **error)
...
+fu_thunderbolt_device_setup_retimer (FuDevice *device, GError **error)
+{
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE (device);
+	guint16 did;
+	guint16 vid;
+	g_autofree gchar *idx = g_path_get_basename (self->devpath);
+	g_autofree gchar *instance = NULL;
+
+	fu_device_set_physical_id (device, idx);
+	fu_device_set_name (device, "Retimer");
```
hughsie, line 356: I'm not sure the user is going to know what that actually is... maybe also `summary` here?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -331,14 +338,76 @@ fu_thunderbolt_device_setup (FuDevice *device, GError **error)
...
+
+	/* try to read the version */
+	if (!fu_thunderbolt_device_get_version (self))
+		g_debug ("failed to read version");
+
+	/* default behavior */
+	self->auth_method = "nvm_authenticate";
+
+	/* configure differences between retimer and controller */
+	return self->is_retimer ? fu_thunderbolt_device_setup_retimer (device, error) :
```
hughsie, line 407: rather than is_retimer i think it might be better to have something like `device_mode` where `FU_THUNDERBOLT_DEVICE_MODE_RETIMER` and `FU_THUNDERBOLT_DEVICE_MODE_CONTROLLER` are the enumerated types.

libfwupdplugin/fu-udev-device.c
superm1: Maybe. I wasn't sure since I couldn't test it.

hughsie: i think this kind of safe/obvious stuff can just go into 1.4.5 even without the retimer bits.

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: adjusted

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: I added a summary, quoting the PCIe spec

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: I don't know, I've never seen that key set in the `thunderbolt` subsystem.  I've adjusted to only set if it's not set.

libfwupdplugin/fu-udev-device.c
superm1: OK, adjusted it.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -171,11 +177,31 @@ fu_thunderbolt_device_check_safe_mode (FuThunderboltDevice *self)
 	fu_device_set_metadata_boolean (FU_DEVICE (self), FU_DEVICE_METADATA_TBT_IS_SAFE_MODE, TRUE);
 }
 
+static const gchar*
+fu_thunderbolt_device_type_to_string (FuThunderboltDevice *self)
+{
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER) {
+		if (self->gen >= 4)
+			return "USB4 host controller";
```
hughsie, line 185: is it USB4 or `USB 4`? I've not been following all the branding stuff...

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -171,11 +177,31 @@ fu_thunderbolt_device_check_safe_mode (FuThunderboltDevice *self)
...
+			return "Thunderbolt host controller";
+	}
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_DEVICE_CONTROLLER) {
+		if (self->gen >= 4)
+			return "USB4 device controller";
+		else
+			return "Thunderbolt device controller";
+	}
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_RETIMER)
+		return "Retimer";
```
hughsie, line 196: I don't think the user is going to be familiar with that when it appears in `get-devices`...

superm1: Do you have a suggestion?

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: `USB4™` https://www.usb.org/document-library/usb4tm-specification

hughsie: I don't think we want the TM else we'll have to scatter them all over the place :)

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: `USB4 extension device` perhaps?

superm1: What about if it's renamed to USB4 retimer and put something along those lines in the summary?

hughsie: Yes, i think that's fine. Thanks.

PR #2253, opened 2020-07-03
Cache the FuSecurityAttrs in the daemon

hughsie:
At the moment at startup we're calculating the attrs so we can export the HSI
string property on the D-Bus interface. Running `fwupdtool security` actually
gets all the security attributes at least twice!

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -69,6 +69,7 @@
 #endif
 
 static void fu_engine_finalize	 (GObject *obj);
+static void fu_engine_ensure_security_attrs	(FuEngine *self);
 
 struct _FuEngine
 {
@@ -107,7 +108,7 @@ struct _FuEngine
 	JcatContext		*jcat_context;
 	gboolean		 loaded;
 	gchar			*host_security_id;
-	gboolean		 host_security_id_valid;
+	FuSecurityAttrs		*host_security_attrs;
 };
 
 enum {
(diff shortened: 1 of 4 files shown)


hughsie:
>  For example runtime things need to be checked on every call.

I thought we were pretty good at invalidating? e.g. `fu_plugin_security_changed()` -- if we're not doing that then the HSI string is going to be invalid, and the GUIs that are displaying it are not going to be valid either.

superm1:
But like the built in stuff - taints for example

hughsie:
> taints for example

I think if we taint the computer and have gnome-control-center open then the display should immediately change to add `!`. I can look at making that happen if you want.

superm1:
Yeah if that's possible I think it makes sense to do.

hughsie:
Okay, I've been more aggressive invalidating the cache now. All the existing plugins seem to do the right thing and call `fu_plugin_security_changed()` when something changes.

src/fu-engine.c
```diff
@@ -106,8 +106,9 @@ struct _FuEngine
 	gchar			*host_machine_id;
 	JcatContext		*jcat_context;
 	gboolean		 loaded;
+	gboolean		 host_security_valid;
```
superm1, line 109: Rather than a boolean attribute, what do you think about just setting `host_security_id` to `NULL` when no longer valid and check for that everywhere?

hughsie: Something like that?

PR #2256, opened 2020-07-07
Wip/hughsie/libfwupd libsoup

hughsie:


libfwupd/fwupd-client.c
```diff
@@ -2039,6 +2112,129 @@ fwupd_client_get_remote_by_id (FwupdClient *client,
...
+ * Since: 1.5.0
+ **/
+void
+fwupd_client_set_user_agent (FwupdClient *client, const gchar *user_agent)
+{
+	FwupdClientPrivate *priv = GET_PRIVATE (client);
+	g_return_if_fail (FWUPD_IS_CLIENT (client));
+	g_return_if_fail (user_agent != NULL);
+	g_free (priv->user_agent);
+	priv->user_agent = g_strdup (user_agent);
```
superm1, line 2131: Is it worth some extra checks here that the agent looks like the expected format?

libfwupd/fwupd-client.c
```diff
@@ -2039,6 +2112,129 @@ fwupd_client_get_remote_by_id (FwupdClient *client,
...
+	fwupd_client_set_percentage (client, percentage);
+}
+
+/**
+ * fwupd_client_download_bytes:
+ * @client: A #FwupdClient
+ * @url: the remote URL
+ *
+ * Downloads data from a remote server. The fwupd_client_set_user_agent() function
+ * should typically be called before this method is used.
```
superm1, line 2168: I think it's worth mandating this. Fail the call if agent not set

libfwupd/fwupd-client.c
hughsie: is there an expected format? I mean, any non-zero string is okay, e.g. `dave` or `a b c d=e f="g"`

libfwupd/fwupd-client.c
hughsie: Right, can do.

libfwupd/fwupd-client.c
superm1: But since this will go up to lvfs you can actually mandate a particular format to make it easier to distinguish with client is in use slot.

superm1: To your point in the comment if you want clients to always have the "name/version" syntax you can mandate that.  But if you don't think this is a useful effort, understand.

libfwupd/fwupd-client.c
hughsie: Done.

libfwupd/fwupd-client.c
hughsie: Aha, of course. I've added a check to see if at least the fwupd version is set, which is the one we really care about from an LVFS point of view. In the process, I solved a mystery, which is fixed in the final commit in the branch.

hughsie:
> superm1 approved these changes

We only want this for 1.5.0? or 1.4.x too?

superm1:
>We only want this for 1.5.0? or 1.4.x too?

I don't think it's too risky for 1.4.x if you want it backported there too (and `s,1.5.0,1.4.5`)

hughsie:
> I don't think it's too risky for 1.4.x if you want it backported there too

I think that's a good idea, we really **need** to be using the installed fwupd version when asking firmware from the LVFS; some people are missing out on important firmware. I'll respin.

superm1:
btw after you backport that to 1.4.5 what else are you planning to bring back?  I guess the rest of that dock stuff?

hughsie:
> what else are you planning to bring back

Currently it's anything that's not HSI specific....

superm1:
Oh so a lot then, maybe the retimer stuff should come back too afterall.

PR #2257, opened 2020-07-07
fix syntax in German translation file (de.po)

mendel5:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [x] Documentation

This is just a small fix in a comment of the German translation file de.po.


mendel5:
@hughsie Thank you!

PR #2258, opened 2020-07-07
also fix syntax for 20 other translation files

mendel5:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [x] Documentation

Related to https://github.com/fwupd/fwupd/pull/2257 .

This is the same type of syntax fix as in the pull request for the German translation file (see https://github.com/fwupd/fwupd/pull/2257 ). I have now used search and replace to fix this issue in all .po translation files.

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -15,7 +15,7 @@ msgstr ""
 "Language: af\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
-#. more than a minute
+#. TRANSLATORS: more than a minute remaining
 #, c-format
 msgid "%.0f minute remaining"
 msgid_plural "%.0f minutes remaining"
--- a/po/ca.po
+++ b/po/ca.po
@@ -16,7 +16,7 @@ msgstr ""
 "Language: ca\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
-#. more than a minute
+#. TRANSLATORS: more than a minute remaining
 #, c-format
 msgid "%.0f minute remaining"
 msgid_plural "%.0f minutes remaining"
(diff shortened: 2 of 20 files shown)


mendel5:
> The translation comment should automatically update.

Where do the updated translation comments come from?

hughsie:
> Where do the updated translation comments come from?

`ninja fwupd-pot` regenerates the .pot file from the source files

mendel5:
Do I understand it correctly that if this pull request was merged and you generated a new translation file, all the changes in this pull request would be overwritten?

Therefore changing `/* more than a minute */` to `/* more than a minute remaining */` will partially fix that?
https://github.com/fwupd/fwupd/blob/72f78dc060f82a537399ded5536cb6c0cd481cff/src/fu-progressbar.c#L142

Why does it say `TRANSLATORS:` in front of all other comments but not in front of the first comment?
https://github.com/fwupd/fwupd/blob/72f78dc060f82a537399ded5536cb6c0cd481cff/po/ca.po#L19

hughsie:
> all the changes in this pull request would be overwritten

Correct. I've just synced transifex if you'd like to check.

mendel5:
> I've just synced transifex if you'd like to check.

I would like to check it, but I don't know how... Can you tell me?

If you have synced transifex I would expect the files in https://github.com/fwupd/fwupd/tree/master/po to have changed recently due to the regeneration, but that does not seem to be the case. Why is that?

hughsie:
> Can you tell me?

https://www.transifex.com/freedesktop/fwupd/dashboard/

> to have changed recently due to the regeneration

This is only done at release time to prevent git churn.

PR #2259, opened 2020-07-09
Wip/hughsie/remote refresh client

hughsie:


hughsie:
As an aside, fwupd_client_install_release() is something I also want to use from fwupdagent...

meson.build
```diff
@@ -146,6 +146,9 @@ add_project_arguments('-DFWUPD_DISABLE_DEPRECATED', language : 'c')
 add_project_arguments('-D_BSD_SOURCE', language : 'c')
 add_project_arguments('-D_XOPEN_SOURCE=700', language : 'c')
 
+# needed for memfd_create()
+add_project_arguments('-D_GNU_SOURCE', language : 'c')
```
superm1, line 150: Rather than `_BSD_SOURCE` and `_GNU_SOURCE`, can you just set `_DEFAULT_SOURCE`?

superm1:
>As an aside, fwupd_client_install_release() is something I also want to use from fwupdagent...

An agent with super powers!

meson.build
hughsie: No, it really needs to be `_GNU_SOURCE`

PR #2260, opened 2020-07-09
msr: Add a new plugin to detect the Intel DCI state

hughsie:


hughsie:
With this I get:

    HSI-1
    ✔ Intel DCI debugger:            Disabled

    HSI-2
    ✔ Intel DCI debugger:            Locked

superm1:
For me:
```
14:07:32:0123 FuEngine             0 devices with subsystem msr
```

superm1:
`msr.ko` doesn't seem to load for me by default.  Loading it I then see this on a CML system.
```
HSI-1
✔ Intel DCI debugger:            Disabled

HSI-2
✔ Intel DCI debugger:            Locked
```

plugins/msr/README.md
```diff
@@ -0,0 +1,14 @@
+MSR
+===
+
+Introduction
+------------
+
+This plugin checks if the MSRs indicate the Direct Connect Interface (DCI) is
```
superm1, line 7: Probably need to describe what MSR actually is in this readme or link to something explaining it.

hughsie:
> but we need a way to get msr.ko loaded from fwupd

Ahh, it's builtin to the Fedora kernel. Is there a clever systemd thing we can do? Maybe just shelling out to `modprobe msr` is acceptable? It's a NOP if it's builtin.

hughsie:
@superm1 does that ^^ work for you?

superm1:
> @superm1 does that ^^ work for you?

Yes it is functional for me on Ubuntu.

A few boundary conditions we should consider though:
* What happens if `msr` isn't compiled into kernel and also isn't available as a module?  I think fwupd would just not show these as part of the HSI equation, which I don't think is our expected behavior.  It should be something about being unable to check or similar.
* Should that conf file to load `msr` be loaded by a snap as well?  If so, this needs to get installed in `snap/hooks/install` and removed in `snap/hooks/remove` too.
* Is the MSR supported on **all** Intel systems that support MSR?  I would guess that the answer is no, specifically older stuff not being supported, and what happens on those?  Garbage data?  Or a failure to run the MSR lookup?

hughsie:
> I think fwupd would just not show these as part of the HSI equation, which I don't think is our expected behavior

Agree.

> Should that conf file to load msr be loaded by a snap as well

If we did something like `Want=modprobe@msr.service` in the fwupd.service (as well as the autoload) does that help snap too?

> Is the MSR supported on all Intel systems that support MSR

I think everything we care about, but I do believe there's an MSR you can check to see which MSRs are supported on a given platform. I'll investigate.

superm1:
> > I think fwupd would just not show these as part of the HSI equation, which I don't think is our expected behavior
> 
> Agree.
> 
> > Should that conf file to load msr be loaded by a snap as well
> 
> If we did something like `Want=modprobe@msr.service` in the fwupd.service (as well as the autoload) does that help snap too?
> 

So that would break if you don't have msr compiled into your kernel or as a module, right?  It seems like that's unfair to a user to block fwupd because a specific security thing that they might not care about isn't able to be measured if they really don't need msr in their kernel.

But as for snap - Unfortunately not for - snapd creates it's own systemd unit that we don't get to touch.  I think the `modload.d` approach is fine for snap.  It just needs to be copied over in those hooks too.

> > Is the MSR supported on all Intel systems that support MSR
> 
> I think everything we care about, but I do believe there's an MSR you can check to see which MSRs are supported on a given platform. I'll investigate.

OK.  That would also help if say some future platform blocks that debugging MSR in the future.

superm1:
> modprobe@msr.service

Looks like this is specific to systemd 245 and later, so probably needs to be conditionally added too.

hughsie:
> So that would break if you don't have msr compiled into your kernel or as a module, right?

plugins/msr/fu-plugin-msr.c
```diff
@@ -34,6 +37,21 @@ fu_plugin_init (FuPlugin *plugin)
...
+	FuPluginData *priv = fu_plugin_get_data (plugin);
+	guint eax = 0;
+	guint ebx = 0;
+	guint ecx = 0;
+	guint edx = 0;
+
+	/* sdbg is supported: https://en.wikipedia.org/wiki/CPUID */
+	__get_cpuid (0x01, &eax, &ebx, &ecx, &edx);
+	priv->is_supported = ((ecx >> 11) & 0x1) > 0;
+	return TRUE;
```
superm1, line 52: why not just abort plugin startup here?

hughsie: good question, i did ponder outputting the status as "not supported" but changed by mind. Do you think not-suported makes sense, or just not showing anything about DCI?

superm1: I think showing `not supported` just begs the questions:
* what do I do to support it?  
* Is that a good thing it's not supported?
* Is something wrong with my system?

Since the system doesn't support it I think just return FALSE in that circumstance and leave it out of the equation.  Here's the perfect example:
* If you're running on AMD and AMD doesn't support something like DCI, would you show not supported?  I think the answer is no.

hughsie: The other thing I wondered was when we add more things to the MSR plugins, only some of the HSI checks need to be shipped. I guess we could rename `is_supported` to `ia32_debug_supported` and then skip the tests in the two `ia32_debug`-specific tests.

superm1: that makes sense to me

hughsie:
Something like that^?

Issue #2265, opened 2020-07-16
Intel BootGuard reported as Invalid on platform with BootGuard enabled


dnoliver:
**Describe the bug**

The `fwupdmgr security --force` report says that Intel BootGuard is invalid, in a CompuLab Fitlet2
The OEM says that Intel BootGuard is correctly enabled in the device. 

After inspecting the test code in here: https://github.com/fwupd/fwupd/commit/6269a839ebf770ecbef8700656426c0fe338f7bb, I can see that the reason why the report is showing invalid can be:
1.	Verified Boot is not enabled or,
2.	ACM protection is not required or,
3.	Policy is not “immediately shutdown” or,
4.	The vendor did not set the FPF OTP fuse

Is there any way to have a more granular report on why `fwupdmgr` is saying that Boot Guard is disabled?

```shell
[test@compulab-fitlet2-1181004-04904 ~]$ rpm -qa fwupd
fwupd-1.5.0-0.603.20200706git.fc32.x86_64

[test@compulab-fitlet2-1181004-04904 ~]$ fwupdmgr --version
client version: 1.4.1-247-gfb0a938f
compile-time dependency versions
        gusb:   0.3.4
        efivar: 37
daemon version: 1.4.1-247-gfb0a938f

[test@compulab-fitlet2-1181004-04904 ~]$ sudo fwupdmgr security --force
Host Security ID: HSI:1+U (v1.5.0)

HSI-1
✔ SPI BIOS region:               Locked
✔ SPI lock:                      Enabled
✔ SPI write:                     Disabled
✔ TPM v2.0:                      Found
✔ TXE manufacturing mode:        Locked
✔ TXE override strap:            Locked
✔ TXE v0:3.1.70.2334:            Valid
✔ UEFI dbx:                      Found

HSI-2
✔ IOMMU:                         Enabled
✔ TPM PCR0 reconstruction:       Valid
✘ Intel BootGuard:               Invalid

HSI-3
✔ Suspend-to-ram:                Disabled
✘ Intel CET:                     Not supported
✘ Pre-boot DMA protection:       Disabled
✘ Suspend-to-idle:               Disabled

HSI-4
✔ Intel SMAP:                    Enabled
✘ Encrypted RAM:                 Not supported

Runtime Suffix -U
✔ Firmware updates:              Supported

Runtime Suffix -A
✘ Firmware attestation:          Not supported

Runtime Suffix -!
✔ Linux kernel:                  Untainted
[...]

PR #2266, opened 2020-07-16
trivial: pci-mei: show why bootguard is invalid more verbosely

superm1:
See #2265

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
FYI this does fail on my pre-production XPS 7390 on hand because:
```
19:21:57:0350 FuPluginPciMei       FPF OTP fuse not set
```

PR #2275, opened 2020-07-24
Wip/hughsie/runtime update protocol

hughsie:
This isn't 100% true, as these are the runtime devices only used for detaching.


superm1:
Maybe to make if clearer - have a predefined "dummy" protocol that the daemon knows can't be used for updating.

hughsie:
> Maybe to make if clearer - have a predefined "dummy" protocol that the daemon knows can't be used for updating.

I did think of having something like FWUPD_DEVICE_FLAG_RUNTIME that would just neuter the warning -- but when I did that the firmware failed to deploy as it checks the protocol matches the metadata *before* with switches the device the bootloader (which is probably the right thing to do). I think this PR is probably the most pragmatic thing, where the protocol is "what would we use to flash" rather than "what are we using right now". Maybe the docs for `_set_protocol()` just need to reflect that.

PR #2281, opened 2020-07-27
Wip/hughsie/hsi docbook

hughsie:


hughsie:
Screenshot from 2020-07-27 19-42-10
Screenshot from 2020-07-27 19-42-27

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+          <firstname>Alex</firstname>
+          <surname>Matrosov</surname>
+        </personname>
+        <affiliation>
+          <orgname>Nvidia</orgname>
+        </affiliation>
+      </author>
+    </authorgroup>
+  </info>
+-->
```
superm1, line 55: If these weren't commented out, how do they actually render?  Or do they?

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+        from signed reports uploaded to the LVFS and from analyzing firmware binaries.
+        However this information alone does not explain firmware security to the
+        user in a way they can actually interpret.
+      </para>
+    </refsect1>
+
+    <refsect2 id="existing-tooling">
+      <title>Existing Tooling</title>
+      <para>
+        Today figuring out the true security of your hardware and firmware
```
superm1, line 85: The word "Today" doesn't generally age well in something like this.

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+        people buying hardware.
+        The service already knows a huge amount of information about machines
+        from signed reports uploaded to the LVFS and from analyzing firmware binaries.
+        However this information alone does not explain firmware security to the
+        user in a way they can actually interpret.
+      </para>
+    </refsect1>
+
+    <refsect2 id="existing-tooling">
+      <title>Existing Tooling</title>
```
superm1, line 83: Right now existing means "pre-fwupd 1.5.x tooling".

Maybe this section needs to be renamed to "Other tools"?

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+    </refsect1>
+
+    <refsect2 id="runtime-behaviour">
+      <title>Runtime Behavior</title>
+      <para>
+        Orthogonal to the security features provided by the firmware there are
+        other security considerations related to the firmware which may require
+        internet access to discover or that runtime OS changes directly affect
+        the security of the firmware.
+        It would not make sense to have <emphasis>“have updates on the LVFS”</emphasis>
```
superm1, line 140: with the docbook markup, does it still need `"`

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+        as a requirement for a specific security level as this would mean
+        offline the platform might be a higher level initially but as soon as
+        it is brought online it is downgraded which would be really confusing to
+        users.
+        The “core” security level will not change at runtime, but the suffix may.
```
superm1, line 145: Operating System runtime

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+      </para>
+    </refsect2>
+
+    <refsect2 id="hsi-level0">
+      <title>HSI:0 (Insecure)</title>
+      <para>
+        The lowest security level with very few detected firmware protections.
```
superm1, line 152: I would suggest `s/very few/little or no/`

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+      <para>
+        Updates available on the <ulink url="https://fwupd.org/">
+        Linux Vendor Firmware Service </ulink> which are less than 12 months old.
+      </para>
+    </refsect3>
+
+    <refsect3 id="runtime-a">
+      <title>HSI Runtime Suffix <code>A</code></title>
+      <para>
+        Attestation data is available to verify the current PCR0 TPM hash.
```
superm1, line 207: Something about this being UFEI specific and linking to an explanation of PCR0 would be useful here.

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+        <listitem>
+          <para>
+            The kernel is not not <ulink url="https://mjg59.dreamwidth.org/50577.html">
+            locked down</ulink>. <emphasis>v1.5.0</emphasis>
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            Unencrypted <ulink url="https://wiki.archlinux.org/index.php/Dm-crypt/Swap_encryption">
+            swap partition</ulink>. <emphasis>v1.5.0</emphasis>
```
superm1, line 238: Maybe worth including a link to your blog for this too?

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+            </listitem>
+          </itemizedlist>
+        </para>
+      </note>
+    </refsect3>
+
+    <refsect3 id="org.fwupd.hsi.Spi.Ble">
+      <title>BIOS Lock Enable (BLE)</title>
+      <para>
+        If the lock bit is set then SMIs are raised when setting BIOS Write Enable.
```
superm1, line 289: Probably need to expand SMI as this is the first use of it.

docs/hsi.xml
superm1: Also this is the first time a TPM is mentioned and it might need to be explained too.

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+          </itemizedlist>
+        </para>
+      </note>
+    </refsect3>
+
+    <refsect3 id="org.fwupd.hsi.Mei.Version">
+      <title>CSME Version</title>
+      <para>
+        Converged Security and Manageability Engine is a standalone management
+        module that can manage and control all the local devices without the host
```
superm1, line 468: That's a bit of an overreach.  It only controls very specific devices.

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+              <ulink url="https://www.kernel.org/doc/html/latest/security/tpm/tpm_event_log.html">
+                Linux Kernel TPM Documentation
+              </ulink>
+            </listitem>
+          </itemizedlist>
+        </para>
+      </note>
+    </refsect3>
+    <refsect3 id="org.fwupd.hsi.AcpiDmar">
+      <title>Kernel DMA protection</title>
```
superm1, line 570: This is pre-boot specifically, no?

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+      </para>
+    </refsect1>
+
+    <refsect2 id="ommissions">
+      <title>Intentional Omissions</title>
+    </refsect2>
+    <refsect3 id="intel-txt">
+      <title>Intel TXT</title>
+      <para>
+        This security functionality requires a binary blob to operate.
```
superm1, line 839: eh?  https://github.com/TrenchBoot

docs/hsi.xml
```diff
@@ -0,0 +1,875 @@
...
+      <title>Intel SGX</title>
+      <para>
+        This is not widely used as it has several high severity security issues.
+      </para>
+    </refsect3>
+    <refsect3 id="intel-mpx">
+      <title>Intel MPX</title>
+      <para>
+        MPX support has now been removed from popular compilers and the kernel
+        several years ago and it is now considered obsolete.
```
superm1, line 853: this statement won't age well - probably need a specific time frame to include

docs/hsi.xml
hughsie: I commented them out for two reasons:

 * I'd like people to sign-off the final version, possibly when the ARM/AMD stuff has landed too
 * The rendering of the names+vendors is simply rubbish in the gtk-doc style sheet -- if we **do** want the author block then I might resort to `<para>` and `<itemizedlist>` as a hack.

Screenshot_2020-07-28 Host Security ID Specification fwupd Reference Manual

docs/hsi.xml
hughsie: how about,

        Traditionally, figuring out the true security of your hardware and firmware
        requires sifting through the marketing documentation provided by the
        OEM and in many cases just “trusting” they did it right.

docs/hsi.xml
hughsie: Agree, fixed.

docs/hsi.xml
hughsie: Good point, fixed.

docs/hsi.xml
hughsie: I considered it, but I wanted an independent source really. Like a wikipedia page, if it existed.

docs/hsi.xml
hughsie: How about ... `manage and control some local devices`

docs/hsi.xml
hughsie: How about

        MPX support was removed from GCC and the Linux kernel in 2019 and it is
        now considered obsolete.

docs/hsi.xml
hughsie: like?

        If the lock bit is set then System Management Interrupts (SMIs) are
        raised when setting BIOS Write Enable.

docs/hsi.xml
hughsie: How about

        Attestation data is available to verify the current system firmware.
        For most UEFI platforms, this is usually the Trusted Platform Module (TPM)
        PCR0 hash, but other attestation checksums could be used.

docs/hsi.xml
hughsie: I think this is a good example of where the spec and fwupd were disagreeing on what stuff actually meant; and one of the big reasons I wanted to bring it all into one place (the other, being able to add URLs to the various HSI attrs to link back to the spec...).

This is of course pre-boot IOMMU. How about this:

      <title>Pre-boot DMA protection</title>
      <para>
        The IOMMU on modern systems is used to mitigate against DMA attacks.
        All I/O for devices capable of DMA is mapped into a private virtual
        memory region.
        The ACPI DMAR table is used to set up pre-boot DMA protection which
        eliminates some firmware attacks.
      </para>

docs/hsi.xml
hughsie: I'll admit I didn't know how TrenchBoot and TXT operate together. Shall I just remove this TXT section completely for now?

docs/hsi.xml
superm1: Ah OK, thanks.

docs/hsi.xml
superm1: Yeah, that's better

docs/hsi.xml
superm1: Sounds good, and it's good to disambiguate that something non-UEFI may exist some day.

docs/hsi.xml
superm1: OK.

docs/hsi.xml
superm1: 👍

docs/hsi.xml
superm1: Yes

docs/hsi.xml
superm1: 👍

docs/hsi.xml
superm1: I think that's better for now, yes.  But also it's worth looking into, it's how TXT can come to the masses.

docs/hsi.xml
superm1: 👍

hughsie:
@superm1 should I push this to master now? I'd like to propose some new tests in a few days and I'd like to use the URL thing to point to the spec so I can add the spec entry at the same time.

PR #2282, opened 2020-07-28
ccgx: modify readmd to remove customer info

chlee75:
1. Reorder section in readme.md due to Lenovo request.
2. Modify readme.md to remove customer specific information.

### DIFF
--- a/plugins/ccgx/README.md
+++ b/plugins/ccgx/README.md
@@ -3,15 +3,17 @@ Cypress support
 
 Introduction
 ------------
+
 This plugin can flash firmware on Cypress CCGx USB-C controller family of
-devices used in dock solutions.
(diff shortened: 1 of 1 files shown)


chlee75:
@superm1 
I fixed rendering issue of README.md. 
I've just got file from customer (Lenovo) and checked text only.
Sorry for not to check rendering.

superm1:
Did the line endings change or something from `\n` to `\n\r`?  Things that shouldn't be a diff are showing up as diffs still.

chlee75:
@superm1 
`Did the line endings change or something from \n to \n\r? Things that shouldn't be a diff are showing up as diffs still.`
I convert DOS format to UNIX format using dos2unit command

PR #2283, opened 2020-07-28
Created new plugin for PS175 device

shiyu-s:
Type of pull request:
- [x] New plugin (Please include [new plugin checklist](https://github.com/hughsie/fwupd/wiki/New-plugin-checklist))

- [] CI run for at least one target
- [] Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc).

### DIFF
--- a/libfwupdplugin/fu-common.h
+++ b/libfwupdplugin/fu-common.h
@@ -52,6 +52,7 @@ typedef guint FuEndianType;
  * @FU_PATH_KIND_SYSCONFDIR_PKG:	The package configuration location (IE /etc/fwupd)
  * @FU_PATH_KIND_SYSFSDIR_FW:		The sysfs firmware location (IE /sys/firmware)
  * @FU_PATH_KIND_SYSFSDIR_DRIVERS:	The platform sysfs directory (IE /sys/bus/platform/drivers)
+ * @FU_PATH_KIND_I2C_DEVICES:		The i2c devices directory (IE /sys/bus/i2c/devices)
  * @FU_PATH_KIND_SYSFSDIR_TPM:		The TPM sysfs directory (IE /sys/class/tpm)
  * @FU_PATH_KIND_PROCFS:		The procfs location (IE /proc)
  * @FU_PATH_KIND_POLKIT_ACTIONS:	The directory for policy kit actions (IE /usr/share/polkit-1/actions/)
@@ -71,6 +72,7 @@ typedef enum {
 	FU_PATH_KIND_PLUGINDIR_PKG,
 	FU_PATH_KIND_SYSCONFDIR,
 	FU_PATH_KIND_SYSCONFDIR_PKG,
+	FU_PATH_KIND_I2C_DEVICES,
 	FU_PATH_KIND_SYSFSDIR_FW,
 	FU_PATH_KIND_SYSFSDIR_DRIVERS,
 	FU_PATH_KIND_SYSFSDIR_TPM,
--- a/plugins/flashrom-i2c/flashrom-i2c.quirk
+++ b/plugins/flashrom-i2c/flashrom-i2c.quirk
@@ -1,0 +1,5 @@
+[HwId=1AF80175]
+Programmer=lspcon_i2c_spi
+Device=PS175
+Protocol=com.parade.ps175
+VendorName=Parade
--- a/plugins/flashrom-i2c/fu-i2c-device.c
+++ b/plugins/flashrom-i2c/fu-i2c-device.c
@@ -6,24 +6,272 @@
 #include "config.h"
 #include "fu-i2c-device.h"
 
+#include <glib.h>
+#include <glib/gstdio.h>
+
+#define LAYOUT_FLAG_NAME "FLAG"
+#define LAYOUT_PARTITION_NAME "BLK"
+#define IMG_LAYOUT_NAME "layout"
+#define IMG_FLAG1_NAME "flag1.bin"
+#define IMG_FLAG2_NAME "flag2.bin"
+#define IMG_FIRMWARE_NAME "fw.bin"
+
+struct FlashromArgs
+{
+	const gchar *spi_master;
+	const gchar *layout;
+	const gchar *image;
+	const gchar *operation;
+};
+
 struct _FuI2cDevice
 {
-	FuDevice		 parent_instance;
+	FuDevice parent_instance;
 };
 
 G_DEFINE_TYPE (FuI2cDevice, fu_i2c_device, FU_TYPE_DEVICE)
+
+/* This calls fu_common_get_contents_fd which will close the fd when done. */
+static gint fu_i2c_device_get_boot_block_from_fd (int fd, GError **error)
+{
+	g_autoptr (GBytes) result = NULL;
+	gsize result_len = 0;
+	result = fu_common_get_contents_fd (fd, 1, error);
+	if (result == NULL)
+		return -1;
+
+	const guint8 *result_bytes = g_bytes_get_data (result, &result_len);
+	if (result_len != 1) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "%s: bootblock info size is expected to be 1 byte, \
+			     got %d",
+			     __func__,
+			     result_len);
+		return -1;
+	}
+
+	return result_bytes[0];
+}
+
+static gboolean fu_i2c_device_validate_flashrom_args (
+	const struct FlashromArgs *args,
+	GError **error)
+{
+	if (args->spi_master == NULL ||
+	    args->layout == NULL ||
+	    args->image == NULL ||
+	    args->operation == NULL) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "%s: all arguments under FlashromArgs has \
+			     to be set",
+			     __func__);
+		return FALSE;
+	}
+
+	return TRUE;
+
+}
+
+static gboolean fu_i2c_device_run_command (const struct FlashromArgs *args,
+					   GError **error)
+{
+	if (fu_i2c_device_validate_flashrom_args (args, error) == FALSE)
+		return FALSE;
+
+	const gchar *argv[9] = { "flashrom",
+				 "-p", args->spi_master,
+				 "--layout", args->layout,
+				 "--image", args->image,
+				 args->operation,
+				 NULL };
+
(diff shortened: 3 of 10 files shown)


libfwupdplugin/fu-common.h
```diff
@@ -52,6 +52,7 @@ typedef guint FuEndianType;
  * @FU_PATH_KIND_SYSCONFDIR_PKG:	The package configuration location (IE /etc/fwupd)
  * @FU_PATH_KIND_SYSFSDIR_FW:		The sysfs firmware location (IE /sys/firmware)
  * @FU_PATH_KIND_SYSFSDIR_DRIVERS:	The platform sysfs directory (IE /sys/bus/platform/drivers)
+ * @FU_PATH_KIND_I2C_DEVICES:		The i2c devices directory (IE /sys/bus/i2c/devices)
```
hughsie, line 55: I don't think we need this define, see below...

plugins/flashrom-i2c/flashrom-i2c.quirk
```diff
@@ -0,0 +1,4 @@
+[HwId=1AF80175]
+Programmer=lspcon_i2c_spi
+Device=PS175
+Protocol=com.parade.ps175
```
hughsie, line 4: Is the protocol specific to the PS175? It looks like a generic flashrom blob to me.

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+	const gchar *image;
+	const gchar *operation;
+};
+
+struct _FuI2cDevice
+{
+	FuDevice parent_instance;
+};
+
+G_DEFINE_TYPE (FuI2cDevice, fu_i2c_device, FU_TYPE_DEVICE)
```
hughsie, line 32: This namespacing is wrong, you want something like FuFlashromI2cDevice here

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+
+/* This calls fu_common_get_contents_fd which will close the fd when done. */
+static gint fu_i2c_device_get_boot_block_from_fd (int fd, GError **error)
```
hughsie, line 35: style is to put the `static gint` on a new line

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+{
+	g_autoptr (GBytes) result = NULL;
+	gsize result_len = 0;
+	result = fu_common_get_contents_fd (fd, 1, error);
+	if (result == NULL)
+		return -1;
+
+	const guint8 *result_bytes = g_bytes_get_data (result, &result_len);
```
hughsie, line 43: declarations at the top of the module pls

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+	if (result_len != 1) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "%s: bootblock info size is expected to be 1 byte, \
+			     got %d",
+			     __func__,
```
hughsie, line 50: we don't really do the __func__ thing

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+{
+	if (fu_i2c_device_validate_flashrom_args (args, error) == FALSE)
+		return FALSE;
+
+	const gchar *argv[9] = { "flashrom",
+				 "-p", args->spi_master,
+				 "--layout", args->layout,
+				 "--image", args->image,
+				 args->operation,
+				 NULL };
```
hughsie, line 90: nah, we shouldn't do this. we have a perfectly good flashrom plugin that uses libflashrom. spawning a random binary is a quick hack, sorry. You don't get errors, progress, etc.

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+	const gchar *spi_master,
+	const gchar *layout,
+	const gchar *flag_name,
+	GError **error)
+{
+	struct FlashromArgs flash_args;
+	g_autofree gchar *img_arg = NULL;
+	g_autofree gchar *tmp_file_name = NULL;
+	tmp_file_name = g_build_filename (dir_name, "XXXXXX", NULL);
+	gint fd = g_mkstemp (tmp_file_name);
```
hughsie, line 109: I don't think we want to be creating temp files.

plugins/flashrom-i2c/fu-plugin-flashrom-i2c.c
```diff
@@ -0,0 +1,198 @@
...
+	while ((ent_name = g_dir_read_name (dir)) != NULL) {
+		if (strlen (ent_name) <
+		    strlen (DEVICE_NAME_PREFIX) + HID_LENGTH)
+			continue;
+
+		if (add_i2c_device (plugin, i2c_device_dir,
+				    ent_name, error) == FALSE) {
+			return FALSE;
+		}
+	}
```
hughsie, line 180: This is really not the way to deal with i2c devices; you want to use FuUdevDevice and just ask fwupd to enumerate the correct sysfs class. e.g. see the WIP https://github.com/fwupd/fwupd/blob/wip/hughsie/elantp/plugins/elantp/fu-plugin-elantp.c and the other files in https://github.com/fwupd/fwupd/tree/wip/hughsie/elantp/plugins/elantp

plugins/flashrom-i2c/fu-i2c-device.h
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2020 The Chromium OS Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
```
hughsie, line 3: all plugin code has to be LGPLV2+

plugins/flashrom-i2c/fu-i2c-device.c
shiyu-s: Yes we are aware of the flashrom plugin and ultimately we do like to use the libflashrom, however the flashrom on chromium side: https://chromium.googlesource.com/chromiumos/third_party/flashrom/ does not have the libflashrom support yet, as such feature is expected to take a long time to settle. We would like to get firmware update support now with command line flsahrom. Does it sounds reasonable?

plugins/flashrom-i2c/fu-i2c-device.c
shiyu-s: Is there any particular reason for that? If it's just no temp file, we can just create a normal file inside the temp dir.

plugins/flashrom-i2c/fu-i2c-device.c
hughsie: > however the flashrom on chromium side flashrom does not have the libflashrom support yet

Why not?

> Does it sounds reasonable?

Not really. The "spawn an executable" just hopes the flash process completes. How to you report the progress back to the user? How do we convey the various erase/write/verify stages to the front end? Do we just display a bouncing cursor back and forth for 5 minutes? 30 seconds? What if it hangs during initialisation?

superm1: I think this turns into a bit of a larger question - what is actually different with that flashrom and https://review.coreboot.org/plugins/gitiles/flashrom?  Is the chromium one a fork or just moving at a faster pace?  Or  are things actually flowing back up to the the one at coreboot?

If you take this Parade case, I would wonder specifically if there needs to be specific code on the flashrom side only present in your fork to handle it, or if the (recently tagged) flashrom 1.2 has it?  What kind of check could be done at compile or even runtime time to know there is support?

>as such feature is expected to take a long time to settle

Have they really deviated so much that you can't cherry pick the 25 commits into your fork for it?

plugins/flashrom-i2c/fu-i2c-device.c
```diff
@@ -0,0 +1,288 @@
...
+				   error_local->message);
+		}
+
+		fd = -1;
+	}
+
+	g_unlink (tmp_file_name);
+
+	if (current_block == -1)
+		return -1;
```
superm1, line 146: I think you need to set an error in this context

plugins/flashrom-i2c/fu-plugin-flashrom-i2c.c
```diff
@@ -0,0 +1,198 @@
...
+	quirk_vendor_name = fu_plugin_lookup_quirk_by_id (
+                plugin, quirk_key, DEVICE_VENDOR_NAME);
+	/* Add devices with quirk configuration only. */
+	if (quirk_programmer_name == NULL)
+		return TRUE;
+
+	device_path = g_build_filename (
+		i2c_device_dir, i2c_name, NULL);
+	bus_no = get_i2c_bus_number_from_path (device_path, error);
+	if (bus_no == -1) {
```
superm1, line 103: If this happens I think you end up overwriting the error that came from the function

plugins/flashrom-i2c/fu-i2c-device.h
superm1: We have done dual license on one plugin (dell-dock), so if it's needed to have BSD style, it should be BSD and LGPLv2+

plugins/flashrom-i2c/fu-i2c-device.c
superm1: So specifically I think the temp file is an artifact of interacting with another executable binary, which as mentioned in other comments is not directionally where we want to be going with fwupd.  Where possible we work directly with sysfs interfaces or via ioctl on devfs interfaces.  If there is an abstraction library, then we use that.  So `libflashrom` is the intended way to work with devices that otherwise use flashrom for flashing.

superm1:
So fundamentally in fwupd's architecture we don't want to be having plugins shelling out to other applications in the upstream fwupd codebase.  

For flashrom specifically we have a view of using the upstream flashrom codebase and libflashrom.  So I think the way forward for this is that the chrome flashrom fork needs to align to upstream flashrom - specifically to libflashrom.  At that point I think this will most likely turn into some very small modifications to the existing fwupd flashrom plugin, or possibly just a new entry to quirks to describe the parade device.

So as such I'm going to close this, and we can revisit when it doesn't have a dependency on calling out to `$bindir/flashrom`

hughsie:
> we don't want to be having plugins shelling out to other applications

For the record, I completely agree with Mario. It's not a precedent I want to set.

PR #2285, opened 2020-07-28
elantp: Add a plugin to update Elan Touchpads

hughsie:


hughsie:
This is 100.0% untested, but should be good enough to start a review for sanity.

plugins/elantp/README.md
```diff
@@ -0,0 +1,47 @@
+Elan TouchPad
+=============
+
+Introduction
+------------
+
+This plugin allows updating Touchpad devices from Elan. Devices are enumerated
```
superm1, line 7: Does this also happen to work with touchscreens?  It would be good to confirm if it's the same protocol as the protocol might be renamed and plugin renamed.

hughsie: > Does this also happen to work with touchscreens

I'm lead to believe these are only touchpads, i.e. palm-detection in the keyboard kinda things.

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -0,0 +1,241 @@
...
+	if (g_getenv ("FWUPD_ELANTP_VERBOSE") != NULL)
+		fu_common_dump_raw (G_LOG_DOMAIN, "ReadCmd", buf, sizeof(buf));
+	return fu_elantp_i2c_device_send_cmd (self, buf, sizeof(buf),
+					      rx, rxsz, error);
+}
+
+static gboolean
+fu_elantp_i2c_device_ensure_iap_ctrl (FuElantpI2cDevice *self, GError **error)
+{
+	return TRUE;
```
superm1, line 64: missing code?  Or should this function not be here?

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -0,0 +1,241 @@
...
+	buf[1] = ETP_I2C_IAP_REG_H;
+	memcpy (buf + 2, raw_data, FW_PAGE_SIZE);
+	fu_common_write_uint16 (buf + FW_PAGE_SIZE + 2, checksum, G_LITTLE_ENDIAN);
+
+	if (!fu_elantp_i2c_device_send_cmd (self, buf, sizeof(buf), 0, 0, error))
+		return FALSE;
+	g_usleep (35 * 1000);
+	if (!fu_elantp_i2c_device_ensure_iap_ctrl (self, error))
+		return FALSE;
+	if (self->iap_ctrl & (ETP_FW_IAP_PAGE_ERR | ETP_FW_IAP_INTF_ERR)) {
```
superm1, line 85: I can't see how this is actually set

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -0,0 +1,241 @@
...
+	guint8 buf[FW_PAGE_SIZE + 4];
+
+	buf[0] = ETP_I2C_IAP_REG_L;
+	buf[1] = ETP_I2C_IAP_REG_H;
+	memcpy (buf + 2, raw_data, FW_PAGE_SIZE);
+	fu_common_write_uint16 (buf + FW_PAGE_SIZE + 2, checksum, G_LITTLE_ENDIAN);
+
+	if (!fu_elantp_i2c_device_send_cmd (self, buf, sizeof(buf), 0, 0, error))
+		return FALSE;
+	g_usleep (35 * 1000);
```
superm1, line 82: I think this is better placed in a `#define` at the top of file

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -0,0 +1,241 @@
...
+		return FALSE;
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_elantp_ic2_device_open (FuUdevDevice *device, GError **error)
+{
+//	FuElantpI2cDevice *self = FU_ELANTP_I2C_DEVICE (device);
+	gint addr = 0x15;
```
superm1, line 100: wait, does this mean that all elan touchpads are always hardcoded to I2C address `0x15`?

hughsie: I believe so. Is that the sort of thing that belongs in a quirk?

plugins/elantp/fu-elantp-i2c-device.c
hughsie: Ya, I need to copy the logic from the HID version. It's more of a placeholder for now.

hughsie:
> Am I to understand that IAP == bootloader in the context of this plugin?

Yes, I believe so.

superm1:
> > Am I to understand that IAP == bootloader in the context of this plugin?
> 
> Yes, I believe so.

Unless there is a strong need to keep the IAP term, would you consider to update all the references to bootloader?

plugins/elantp/fu-elantp-i2c-device.c
superm1: If it's a copy of the logic that's also asking to just split out to a common function too.

plugins/elantp/fu-elantp-i2c-device.c
superm1: At least something that _supports_ a quirk, even if the default is `0x15` it would allow for designs to use Elan on a different I2C address.

jinglewu: Yes, elan touchpad has some other i2c addresses.

jinglewu:
> Am I to understand that `IAP` == `bootloader` in the context of this plugin?

IAP is defined as "in-application programming"

hughsie:
> would you consider to update all the references to bootloader

Done. I'm awaiting hardware to validate the HID side of this, and am tempted to drop the i2c side for the initial review. I'm still a bit confused why we can't just use HID everywhere.

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -0,0 +1,474 @@
...
+static gint
+fu_elantp_hid_device_write_cmd (FuElantpHidDevice *self,
+				guint16 reg, guint16 cmd,
+				GError **error)
+{
+	guint8 buf[5];
+
+	buf[0] = 0x0d; /* report number */
+	fu_common_write_uint16 (buf + 0x1, reg, G_LITTLE_ENDIAN);
+	fu_common_write_uint16 (buf + 0x3, cmd, G_LITTLE_ENDIAN);
```
superm1, line 106: same here, don't you have one byte of garbage since you're only writing 2 bytes?

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -0,0 +1,474 @@
...
+					    buf, sizeof(buf), error)) {
+		g_prefix_error (error, "failed to read fw version: ");
+		return FALSE;
+	}
+	fwver = fu_common_read_uint16 (buf, G_LITTLE_ENDIAN);
+	version = fu_common_version_from_uint16 (fwver, FWUPD_VERSION_FORMAT_HEX);
+	fu_device_set_version (device, version);
+
+	/* get EAP firmware version */
+	is_new_pattern = fu_device_has_custom_flag (FU_DEVICE (self), "new-pattern");
```
superm1, line 159: the word new doesn't age well.  is there a better name than new?

jinglewu:
> > would you consider to update all the references to bootloader
> 
> Done. I'm awaiting hardware to validate the HID side of this, and am tempted to drop the i2c side for the initial review. I'm still a bit confused why we can't just use HID everywhere.

In the case of IAP failure, if the linux OSis restarted, the OS cannot detect the elan hid device, so the i2c interface can be used to update the firmware.

hughsie:
> if the linux OSis restarted, the OS cannot detect the elan hid device

Okay, that's reason enough to finish the i2c part too I guess. Thanks for the info.

plugins/elantp/fu-elantp-i2c-device.c
hughsie: Punted for now, lets get the HID stuff upstream first.

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -0,0 +1,474 @@
...
+static gboolean
+fu_elantp_hid_device_read_cmd (FuElantpHidDevice *self, guint16 reg,
+			       guint8 *rx, gsize rxsz, GError **error)
+{
+	guint8 buf[5];
+
+	buf[0] = 0x0d; /* report number */
+	buf[1] = 0x05;
+	buf[2] = 0x03;
+	fu_common_write_uint16 (buf + 0x3, reg, G_LITTLE_ENDIAN);
```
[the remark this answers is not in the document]
hughsie: `sizeof(buf)` is 5, so three bytes offset is buf[3] and buf[4] which is the last addressable location. I've added the explicit setting of the array to make this clear.

plugins/elantp/fu-elantp-hid-device.c
hughsie: I used `iap-version-2`

plugins/elantp/fu-elantp-i2c-device.c
hughsie: Will do, but not by me :)

plugins/elantp/fu-elantp-i2c-device.c
hughsie: Done, agreed.

hughsie:
> This is 100.0% untested

This now is tested with firmware provided by the vendor.

PR #2286, opened 2020-07-28
vli: Add support for the Realtek USB hub I²C protocol

hughsie:


hughsie:
100.0% untested.

plugins/vli/fu-vli-usbhub-realtek-device.c
```diff
@@ -0,0 +1,472 @@
...
+					gpointer user_data,
+					GError **error)
+{
+	FuVliUsbhubRealtekDevice *self = FU_VLI_USBHUB_REALTEK_DEVICE (device);
+	guint8 status = 0xff;
+	if (!fu_vli_usbhub_device_realtek_detach_raw (self, error))
+		return FALSE;
+	if (!fu_vli_usbhub_device_realtek_read_status_raw (self, &status, error))
+		return FALSE;
+	if (status != 0x11) {
```
superm1, line 216: this and other status codes I think should be in definitions

hughsie: I've got some more data back from Realtek, so I've added a ton of them to the code.

Issue #2297, opened 2020-08-03
fu-engine: ANOTHER_WRITE_REQUIRED doesn't work if a device has been replaced during fu_engine_install_blob


bleungatchromium:
**Describe the bug**
This is the update loop in fw_engine_install_blob:

`        /* plugins can set FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED to run again, but they
         * must return TRUE rather than an error */
        device_id = g_strdup (fu_device_get_id (device));
        do {
                /* check for a loop */
                if (++retries > 5) {
                        g_set_error_literal (error,
                                             FWUPD_ERROR,
                                             FWUPD_ERROR_INTERNAL,
                                             "aborting device write loop, limit 5");
                        return FALSE;
                }

                /* don't rely on a plugin clearing this */
                fu_device_remove_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED);

                /* signal to all the plugins the update is about to happen */
                if (!fu_engine_update_prepare (self, flags, device_id, error))
                        return FALSE;

                /* detach to bootloader mode */
                if (!fu_engine_update_detach (self, device_id, error))
                        return FALSE;

                /* install */
                if (!fu_engine_update (self, device_id, blob_fw, flags, error))
                        return FALSE;

                /* attach into runtime mode */
                if (!fu_engine_update_attach (self, device_id, error))
                        return FALSE;

        } while (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED));`

Inside each of _prepare, _detach, _update, and _attach, these are careful not to use the `device` handle, instead they get a device_id, and have to look it up in case the device has rebooted, and needed to be replugged, replaced, or whatehaveyou.

The loop itself, however still has a handle to the initial `device` which it uses for the `fu_device_has_flag` test. I found that when I set `FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED`, and my device had to reboot, drop off the USB bus, and come back on `_attach` that the loop would just exit even though I had set it in my plugin's `_write_firmware` function..

**Steps to Reproduce**
Steps to reproduce the behavior.
Reproduced it consistently using my WIP series here https://github.com/fwupd/fwupd/pull/2289

**Expected behavior**
[...]

PR #2300, opened 2020-08-03
Refresh the handle to device before testing ANOTHER_WRITE_REQUIRED

hughsie:
fu_engine_install_blob may result in the device being reset during attach or
detach and needing to be replugged.
The device handle we're holding may be stale, but it is still used by the do
while loop itself for the ANOTHER_WRITE_REQUIRED test.

Similar to the other functions in the loop, let's get the device handle by id
in case that happened.

Original patch by Benson Leung <bleung@chromium.org>, many thanks.

Fixes https://github.com/fwupd/fwupd/issues/2297


hughsie:
> There's another use of the stale handle on line 2807.

Good catch. How does this new version work?

Issue #2280, opened 2020-07-27
Add the ability to reject specific firmware updates


Yawning:
**Is your feature request related to a problem? Please describe.**

Certain firmware updates cause problems for end users (eg: https://github.com/fwupd/firmware-lenovo-thinkpad/issues/54 https://github.com/fwupd/firmware-lenovo-thinkpad/issues/63), and are repeatedly offered as update candidates when they should not be (or are already) installed.

**Describe the solution you'd like**

It should be possible for the end user to reject specific firmware updates, without having to reject all updates by device GUID.

eg: Explicitly reject `Lenovo ThinkPad T14s Gen 1 / ThinkPad X13 Gen 1 Embedded Controller Firmware 1.0.04`, but accept `Lenovo ThinkPad T14s Gen 1 / ThinkPad X13 Gen 1 Embedded Controller Firmware 1.0.06`.

**Describe alternatives you've considered**

The current solution of rejecting all updates for the troublesome device is poor as "having to remember to check the vendor's website for newer, hopefully fixed updates" is not great UX.

hughsie:
@Yawning do you want to restrict to a specific firmware, or a specific download? For instance, Lenovo could issue an update 1.2.3 with checksum `deadbeef` that doesn't work. They might pull that and reissue 1.2.3 with checksum `beefbeef`.

Would you want to block the install of 1.2.3 or just the one with checksum `deadbeef`?

Yawning:
> Would you want to block the install of 1.2.3 or just the one with checksum `deadbeef`?

Probably the latter (eg: reject downloads by SHA256 digest), since a fixed firmware release would resolve the sort of issues that make me want this capability in the first place.

I think most of the cases I can think of for the former (reject all downloads with a specific version) would be covered by the existing device GUID based option, but maybe I'm not being imaginative enough.

hughsie:
> eg: reject downloads by SHA256 digest

So the problem here is that we don't actually expose the SHA256 hashes of the payload or archive to the user in the `fwupdmgr get-releases` output. @superm1 got any ideas?

superm1:
I don't think I understand the problem.  There is a firmware issue with applying the update and you wan the ability to blacklist that individual update?  Basically https://github.com/fwupd/fwupd/issues/900 ?

Yawning:
> I don't think I understand the problem. There is a firmware issue with applying the update and you wan the ability to blacklist that individual update? Basically #900 ?

Essentially yes.

[...]

PR #2305, opened 2020-08-07
Allow blocking specific firmware releases by checksum

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2280

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -2129,6 +2129,98 @@ fwupd_client_set_approved_firmware (FwupdClient *client,
 }
 
 /**
+ * fwupd_client_get_blocked_firmware:
+ * @client: A #FwupdClient
+ * @cancellable: the #GCancellable, or %NULL
+ * @error: the #GError, or %NULL
+ *
+ * Gets the list of blocked firmware.
+ *
+ * Returns: (transfer full): list of checksums, or %NULL
+ *
+ * Since: 1.4.6
+ **/
+gchar **
+fwupd_client_get_blocked_firmware (FwupdClient *client,
+				    GCancellable *cancellable,
+				    GError **error)
+{
+	FwupdClientPrivate *priv = GET_PRIVATE (client);
+	g_autoptr(GVariant) val = NULL;
+	gchar **retval = NULL;
+
+	g_return_val_if_fail (FWUPD_IS_CLIENT (client), NULL);
+	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), NULL);
+	g_return_val_if_fail (error == NULL || *error == NULL, NULL);
+
+	/* connect */
+	if (!fwupd_client_connect (client, cancellable, error))
+		return NULL;
+
+	/* call into daemon */
+	val = g_dbus_proxy_call_sync (priv->proxy,
+				      "GetBlockedFirmware",
+				      NULL,
+				      G_DBUS_CALL_FLAGS_NONE,
+				      -1,
+				      cancellable,
+				      error);
+	if (val == NULL) {
+		if (error != NULL)
+			fwupd_client_fixup_dbus_error (*error);
+		return NULL;
+	}
+	g_variant_get (val, "(^as)", &retval);
+	return retval;
+}
+
+/**
+ * fwupd_client_set_blocked_firmware:
+ * @client: A #FwupdClient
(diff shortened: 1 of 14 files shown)


libfwupd/fwupd-client.c
```diff
@@ -2128,6 +2128,98 @@ fwupd_client_set_approved_firmware (FwupdClient *client,
...
 
+/**
+ * fwupd_client_get_blocked_firmware:
+ * @client: A #FwupdClient
+ * @cancellable: the #GCancellable, or %NULL
+ * @error: the #GError, or %NULL
+ *
+ * Gets the list of blocked firmware.
+ *
+ * Returns: (transfer full): list of remotes, or %NULL
```
superm1, line 2139: not remotes

src/fu-engine.c
```diff
@@ -527,6 +528,7 @@ fu_engine_modify_config (FuEngine *self, const gchar *key, const gchar *value, G
 	const gchar *keys[] = {
 		"ArchiveSizeMax",
 		"DisabledDevices",
+		"DisabledFirmware",
```
superm1, line 531: BlockedFirmware no?

src/fu-history.c
```diff
@@ -325,6 +343,10 @@ fu_history_create_or_migrate (FuHistory *self, guint schema_ver, GError **error)
 		g_debug ("migrating v%u database by altering", schema_ver);
 		if (!fu_history_migrate_database_v4 (self, error))
 			return FALSE;
+	} else if (schema_ver == 5) {
+		g_debug ("migrating v%u database by altering", schema_ver);
+		if (!fu_history_migrate_database_v5 (self, error))
+			return FALSE;
```
superm1, line 349: don't you also need to do this if someone goes from 1->6 and 2->6 and 3->6 etc?

src/fu-util.c
```diff
@@ -2311,6 +2352,153 @@ fu_util_check_polkit_actions (GError **error)
...
+	devices = fwupd_client_get_history (priv->client, NULL, error);
+	if (devices == NULL)
+		return NULL;
+	device = fu_util_prompt_for_device (priv, devices, error);
+	if (device == NULL)
+		return NULL;
+	release = fu_util_prompt_for_release (priv, fwupd_device_get_releases (device), error);
+	if (release == NULL)
+		return NULL;
+	return fu_util_prompt_for_checksum (priv, fwupd_release_get_checksums (release), error);
```
superm1, line 2371: I think it's going to be confusing that multiple checksums are going to be available to block.  Why not just block the largest (and presumably most secure) checksum?

src/fu-util.c
```diff
@@ -2557,6 +2745,24 @@ main (int argc, char *argv[])
...
+		     "[CHECKSUM]",
+		     /* TRANSLATORS: command description */
+		     _("Blocks a specific firmware from being installed."),
+		     fu_util_unblock_firmware);
+	fu_util_cmd_array_add (cmd_array,
+		     "get-blocked-firmware",
+		     NULL,
+		     /* TRANSLATORS: command description */
+		     _("Gets the list of blocked firmware."),
+		     fu_util_get_blocked_firmware);
```
superm1, line 2765: This same set of commands should be added to fu-tool.c too

libfwupd/fwupd-client.c
hughsie: Fixed.

src/fu-engine.c
hughsie: Eeek, indeed, good catch. Fixed.

src/fu-history.c
hughsie: Yes, good catch, fixed.

src/fu-util.c
hughsie: Valid point. We can use the already-exported `fwupd_checksum_get_best()` -- Fixed.

hughsie:
> Missing bash completion for the new commands

Fixed.

src/fu-util.c
hughsie: I'm not so sure about this; the `approved` stuff isn't in `fwupdtool` either, and I don't think it's particularly useful to allow the admin running as root to say "don't install foo" which then prevents the (same?) admin from installing that firmware update. I don't think fwupdtool has to have the same featureset as fwupdmgras longer term I want to push the power users using fwupdtool to use the daemon even if the daemon auto-quits straight after it's completed.

PR #2306, opened 2020-08-10
Switch the Ubuntu CI to Focal as Groovy is too broken...

hughsie:


superm1:
I'm surprised it's getting broken since there is the whole proposed promotion stuff (it should work like debian:testing in that regard).  But if it's breaking enough to warrant this PR LGTM.

superm1:
Actually hold off a minute - is the breakage because of missing clang-6?  Maybe we should instead move to clang-9 which is in Groovy.

https://launchpad.net/ubuntu/+source/llvm-toolchain-9

hughsie:
> is the breakage because of missing clang-

Yes, I think that's what I was seeing. Is there a "clang-latest" metapackage or something we could use instead?

superm1:
Yeah.  `clang` and `clang-tools`.  I'll see if those work: https://github.com/fwupd/fwupd/pull/2307

PR #2312, opened 2020-08-12
libfwupdplugin: Allow recursion in fu_common_filename_glob()

hughsie:
We want to return filenames that match the glob in any subdirectory.

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1711,13 +1711,38 @@ fu_common_filename_glob_sort_cb (gconstpointer a, gconstpointer b)
 	return g_strcmp0 (*(const gchar **)a, *(const gchar **)b);
 }
 
+static gboolean
+fu_common_filename_glob_recurse (GPtrArray *files,
+				 const gchar *directory,
+				 const gchar *pattern,
+				 GError **error)
+{
+	const gchar *basename;
+	g_autoptr(GDir) dir = g_dir_open (directory, 0, error);
+	if (dir == NULL)
+		return FALSE;
+	while ((basename = g_dir_read_name (dir)) != NULL) {
+		g_autofree gchar *fn = NULL;
+		if (pattern != NULL && !fu_common_fnmatch (pattern, basename))
+			continue;
+		fn = g_build_filename (directory, basename, NULL);
+		if (g_file_test (fn, G_FILE_TEST_IS_DIR)) {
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -1711,13 +1711,38 @@ fu_common_filename_glob_sort_cb (gconstpointer a, gconstpointer b)
...
+	const gchar *basename;
+	g_autoptr(GDir) dir = g_dir_open (directory, 0, error);
+	if (dir == NULL)
+		return FALSE;
+	while ((basename = g_dir_read_name (dir)) != NULL) {
+		g_autofree gchar *fn = NULL;
+		if (pattern != NULL && !fu_common_fnmatch (pattern, basename))
+			continue;
+		fn = g_build_filename (directory, basename, NULL);
+		if (g_file_test (fn, G_FILE_TEST_IS_DIR)) {
```
superm1, line 1729: So I assume the intention is to not follow symlinks?

hughsie: I thought that would follow the symlink... no?

superm1: Documentation is unclear to me. It has explicit mention that G_FILE_TEST_IS_REGULAR follows symlinks but I don't know if this does or not, so I figured this is something we should confirm.

hughsie: The code specifies them as a bitfield, so it's possible that `G_FILE_TEST_IS_SYMLINK` and `G_FILE_TEST_IS_DIR` are both true. I think we always want to follow symlinks, no?

superm1: Well I _think so_, but I don't know the situation that this came up for.  Following symlinks could be _potentially_ dangerous, for example if it's a symlink to a non protected directory.

hughsie: So I guess what you're trying to say is *add a flags argument* which is a perfectly good idea. For the issue at hand, do we ever want to follow symlinks on a FAT32 ESP? Is a symlink even possible there?

superm1: Yeah a flags argument seems sensible.

On FAT32 ESP?  Symlinks don't exist in FAT32 - so this wouldn't matter at all anyway.

hughsie:
Actually, I can just use `fu_common_get_files_recursive()` in the new code -- sorry for the noise.

PR #2314, opened 2020-08-12
uefi-dbx: Move the HSI failure from HSI:1 to a runtime issue

hughsie:
The dbx is controlled by the local admin now, and is something the user can fix
themselves.


superm1:
Yes and no. All the distros install by default and have a user facing tool. However if if fails to install, this is a firmware problem that a user can't solve.

hughsie:
> However if if fails to install, this is a firmware problem that a user can't solve.

Won't *not installing the dbx update because shim is not updated* be much more common than *dbx update failed to be deployed*?

superm1:
>Won't not installing the dbx update because shim is not updated be much more common than dbx update failed to be deployed?

I don't know - we haven't seen the recent 14.7**kb** dbx update try to install in the wild at scale yet. 😛

PR #2315, opened 2020-08-12
Wip/hughsie/fu volume

hughsie:


libfwupdplugin/fu-common.c
```diff
@@ -2069,3 +2070,196 @@ fu_common_is_cpu_intel (void)
...
+ * @kind: A volume kind, typically a GUID
+ * @error: A #GError or NULL
+ *
+ * Call into the plugin's get results routine
+ *
+ * Finds all volumes of a specific type
+ *
+ * Returns: (transfer container) (element-type FuVolume): a #GPtrArray, or %NULL if the kind was not found
+ *
+ * Since: 1.5.0
```
superm1, line 2128: Not going to be a backport to 1.4.6 feature?

hughsie: I was tempted to push 1.5.0 soon. I can do these for 1.4.6 if you'd prefer.

libfwupdplugin/fu-common.c
```diff
@@ -2069,3 +2070,196 @@ fu_common_is_cpu_intel (void)
 #endif
 	return FALSE;
 }
+
+#define UDISKS_DBUS_SERVICE		"org.freedesktop.UDisks2"
+#define UDISKS_DBUS_PATH		"/org/freedesktop/UDisks2/Manager"
+#define UDISKS_DBUS_MANAGER_INTERFACE	"org.freedesktop.UDisks2.Manager"
+#define UDISKS_DBUS_PART_INTERFACE 	"org.freedesktop.UDisks2.Partition"
+#define UDISKS_DBUS_FILE_INTERFACE	"org.freedesktop.UDisks2.Filesystem"
```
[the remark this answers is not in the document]
hughsie: sure can; i wanted your sanity check before I spent too much time on the branch

libfwupdplugin/fu-common.c
superm1: Well I'm just hesitant on 1.5.0 until there is more !intel in HSI..

libfwupdplugin/fu-common.c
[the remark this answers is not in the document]
superm1: rest of it _looks_ good (didn't actually test locally though).

superm1:
Branch looks good, the only thing I would think is possibly missing is self tests for the volume stuff (although I guess you can argue that uefi is exercising a lot of it)

hughsie:
> self tests for the volume stuff

Do you know how to mock that up? It's all using UDisks rather than poking at mtab now.

superm1:
> Do you know how to mock that up? It's all using UDisks rather than poking at mtab now.

Oh True.  I was going to suggest umockdev, but that might not actually work with udisks.  Considering this, maybe skipping tests is the right answer for this stuff.

libfwupdplugin/fu-common.c
hughsie: Fair point. Moved to 1.4.6

libfwupdplugin/fu-common.c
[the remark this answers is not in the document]
hughsie: Fixed.

hughsie:
I'm particularly proud of `fu_volume_locker()` as we didn't do a terribly good job of unmounting the ESP on the failure path before.

PR #2316, opened 2020-08-12
fwupd-remote: Download remote firmware on local remote

djcampello:
This adds a new kind of remote for local metadata (similar to LOCAL) and
remote firmware files (similar to DOWNLOAD). In this way the device can
control the available firmware without having to keep a local copy of
the firmware itself. This option is also useful to setup a remote mirror
with just firmware files without the complete LVFS website support.

Type of pull request: Feature

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -1469,9 +1469,11 @@ fwupd_client_install_release (FwupdClient *client,
 	g_autofree gchar *checksum_actual = NULL;
 	g_autofree gchar *uri_str = NULL;
 	g_autoptr(GBytes) blob = NULL;
+	g_autoptr(SoupURI) uri = NULL;
 
 	/* work out what remote-specific URI fields this should use */
 	uri_tmp = fwupd_release_get_uri (release);
+	uri = soup_uri_new (uri_tmp);
 	remote_id = fwupd_release_get_remote_id (release);
 	if (remote_id != NULL) {
 		g_autoptr(FwupdRemote) remote = NULL;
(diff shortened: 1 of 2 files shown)


superm1:
I'm confused - how is this different than a download remote?  You can just make a download remote.  That remote doesn't have to be LVFS (just compile with LVFS remote turned off and drop in the right remote for what you want).

djcampello:
This allows for metadata.xml.gz to be local with the listing of available firmware in a remote server. In this way you can construct your "vendor.xml.gz" and ship it with the device without having to ship the firmware itself too that can be hosted on any remote server without having to run a full LVFS-website mirror.

superm1:
But isn't it the same thing as having a download remote and then shipping file.xml.gz in the filesystem and never refreshing it?

djcampello:
Not according the current implementation. First of all, there is no way to explicitly set a Type=download without my patch. Type is inferred by the MetadataUri= file:// -> LOCAL http(s):// -> DOWNLOAD. Also without distinguishing between LOCAL_METADATA and DOWNLOAD, fwupdtool refresh will attempt to wget the MetadataUri also failing if it points to a file://

superm1:
Ah okay thanks for clarifying. I think I would rather have an explicit type of download be functional, and remove that automatic inference - make everything explicit with a Type in the file.

@hughsie thoughts?

libfwupd/fwupd-remote.c
```diff
@@ -416,6 +421,12 @@ fwupd_remote_load_from_filename (FwupdRemote *self,
 		return FALSE;
 	}
 
+	/* override remote kind if explict */
+	remote_kind = g_key_file_get_string (kf, group, "Type", NULL);
+	if (remote_kind != NULL) {
+		priv->kind = fwupd_remote_kind_from_string (remote_kind);
+	}
+
```
superm1, line 429: I think specifically keeping this block in, and putting the type into remotes in tree that are missing it makes sense (drop the override remote kind comment though)

hughsie:
> and remove that automatic inference

Maybe; I know we used to require it.

superm1:
And probably here https://github.com/fwupd/fwupd/blob/master/src/fu-tool.c#L1898 you should have an explicit check for the prefix of `file://` and skip saying file remotes can't be refreshed

djcampello:
I just changed the patch for something different: lets accomplish the same by allowing the existing local remote to point to remote locations for firmware

PR #2317, opened 2020-08-12
uefi-dbx: Check the binaries in the ESP are not in the dbx

hughsie:
If found, fail the HSI runtime tests.

### DIFF
--- a/plugins/uefi-dbx/fu-plugin-uefi-dbx.c
+++ b/plugins/uefi-dbx/fu-plugin-uefi-dbx.c
@@ -32,6 +32,37 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 	return TRUE;
 }
 
+static void
+fu_plugin_add_security_attrs_runtime (FuPlugin *plugin, GPtrArray *dbx, FuSecurityAttrs *attrs)
+{
+	g_autoptr(GError) error_local = NULL;
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	/* create attr */
+	attr = fwupd_security_attr_new (FWUPD_SECURITY_ATTR_ID_UEFI_DBX_ESP);
+	fwupd_security_attr_set_plugin (attr, fu_plugin_get_name (plugin));
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	fu_security_attrs_append (attrs, attr);
+
+	/* no binary blob */
+	if (!fu_plugin_get_enabled (plugin)) {
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+		fwupd_security_attr_set_url (attr, "https://github.com/fwupd/fwupd/wiki/Missingdbx");
+		return;
+	}
+
+	/* verify all files */
+	if (!fu_uefi_dbx_signature_list_validate (dbx, &error_local)) {
+		g_warning ("failed to check ESP: %s", error_local->message);
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_FOUND);
+		return;
+	}
+
+	/* success */
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+	fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+}
+
 void
 fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
 {
@@ -94,6 +125,9 @@ fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
 		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_VALID);
 		return;
 	}
+
+	/* verify there are no files in the ESP that match this dbx */
+	fu_plugin_add_security_attrs_runtime (plugin, dbx_update, attrs);
 
 	/* look for each checksum in the update in the system version */
 	if (!fu_efi_signature_list_array_inclusive (dbx_system, dbx_update)) {
--- a/plugins/uefi-dbx/fu-uefi-dbx-common.c
+++ b/plugins/uefi-dbx/fu-uefi-dbx-common.c
@@ -7,6 +7,9 @@
 #include "config.h"
 
 #include "fu-common.h"
+#include "fu-efi-signature-common.h"
+#include "fu-volume.h"
+
 #include "fu-uefi-dbx-common.h"
 
 gchar *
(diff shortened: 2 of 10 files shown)


plugins/uefi-dbx/fu-plugin-uefi-dbx.c
```diff
@@ -41,6 +41,37 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
...
+	if (!fu_plugin_get_enabled (plugin)) {
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+		fwupd_security_attr_set_url (attr, "https://github.com/fwupd/fwupd/wiki/Missingdbx");
+		return;
+	}
+
+	/* verify all files */
+	if (!fu_uefi_dbx_file_validate_esp (dbx, &error_local)) {
+		g_warning ("failed to check ESP: %s", error_local->message);
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_FOUND);
```
superm1, line 66: I have a sense that this should not be a found / not found, but rather vulnerable / not vulnerable

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
```diff
@@ -41,6 +41,37 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
...
+	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
+	fu_security_attrs_append (attrs, attr);
+
+	/* no binary blob */
+	if (!fu_plugin_get_enabled (plugin)) {
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+		fwupd_security_attr_set_url (attr, "https://github.com/fwupd/fwupd/wiki/Missingdbx");
+		return;
+	}
+
```
superm1, line 62: before even digging into the ESP, since this check has a dependency on `pesign` being present, I think you should have a check right now whether the check can even be run, and set a result accordingly.

plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -26,3 +26,63 @@ fu_uefi_dbx_get_dbxupdate (GError **error)
 		return NULL;
 	return g_strdup (g_ptr_array_index (files, 0));
 }
+
+gchar *
+fu_uefi_dbx_get_authenticode_hash (const gchar *fn, GError **error)
+{
+	const gchar *argv[] = { "pesign", "-P", "-i", fn, "-h", NULL };
```
superm1, line 33: Does this by chance also work with `sbsigntool`?  If not, it's going to be an upward battle to get `pesign` pulled in by default in distros that don't use it.

plugins/uefi-dbx/fu-uefi-dbx-file.c
```diff
@@ -180,6 +180,68 @@ fu_uefi_dbx_file_get_checksums (FuUefiDbxFile *self)
...
+	for (guint i = 0; i < files->len; i++) {
+		const gchar *fn = g_ptr_array_index (files, i);
+		g_autofree gchar *checksum = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		/* get checksum of file */
+		checksum = fu_uefi_dbx_get_authenticode_hash (fn, &error_local);
+		if (checksum == NULL) {
+			g_debug ("failed to get checksum for %s: %s", fn, error_local->message);
+			continue;
```
superm1, line 207: currently it looks like if pesign wasn't around valid files will get skipped

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
hughsie: Well, it might be...

    commit 676c6fe263afd159ad42ef864d23dc50f4359dfb (wip/hughsie/esp)
    Author: Richard Hughes <richard@hughsie.com>
    Date:   Wed Aug 12 15:04:24 2020 +0100
    
        Add a compatible re-implementation of the rhboot dbxtool
    
    :000000 100644 00000000 a63b9ecb A      plugins/uefi-dbx/fu-dbxtool.c
    :100644 100644 32f6b9dc 93690720 M      plugins/uefi-dbx/meson.build

hughsie: > but rather vulnerable / not vulnerable

What I'm trying to say is *we can't deploy the new dbx because of stuff we've found in your ESP* -- hence `FOUND`... I'm not sure that's super clear or not...

plugins/uefi-dbx/fu-uefi-dbx-common.c
hughsie: I didn't find any way of making the various `sbsigntool` command lines give me the authenticode hash of the EFI binary. If you know a way offhand that would be great, otherwise I can tap the wisdom of twitter.

plugins/uefi-dbx/fu-uefi-dbx-file.c
hughsie: This was intentional; if we can do the fallback to another tool as well maybe we can make this an error.

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
superm1: I guess with the lens in mind that this is intended to become a prerequisite to a fu-dbxtool loading the new dbx rather than just as a security attribute I look at it differently.  I think the proper descriptive phrasing depends upon how fu-dbxtool is going to be used.  Maybe you can walk through what you have in mind right now?  I'll mind dump how I think it can be useful now that I know it's a thing.

1. fwupd daemon starts up, checks for vulnerable binaries, checks for out of date dbx, checks for running under virtualization or live environment.
3. if vulnerable binaries fail the runtime thing (like this PR)
4. if out of date dbx, no vulnerable binaries, not live environment try to install new dbx
5. set up a file monitor on a fixed location to store dbx binaries
5. publish dbx hashes with lvfs metadata
6. pull down new dbx as necessary as part of `fwupdmgr refresh`

plugins/uefi-dbx/fu-uefi-dbx-common.c
superm1: I don't know off hand.  Entirely possible it doesn't support it too!

One more thought - is it a lot of code to lift from pesign?  Maybe just borrow it directly into here?

hughsie: > is it a lot of code to lift from pesign

It looked a lot of code, but I can find out exactly how much tomorrow.

plugins/uefi-dbx/fu-plugin-uefi-dbx.c
hughsie: >  Maybe you can walk through what you have in mind right now

I'm not completely sure myself yet. I’m also wondering if we _should_ ship dbx updates via the LVFS longer term; shipping binary blobs in packages for security updates doesn’t really scale and gives us almost insight to what's going on. That said, I’m not sure I want the LVFS to provide a few hundred kb file to every Linux machine on the planet... and it's not going to work for RHEL where the LVFS is still disabled by default. Maybe we just put it on the CDN or something. Much more pondering required.

One thing I also considered was making the dbx entry a FuDevice, as in expose a version and match a GUID like the other devices. The "version" could be the number of hashes in the entry, or something like that with some heuristic hacks.

superm1: >One thing I also considered was making the dbx entry a FuDevice, as in expose a version and match a GUID like the other devices. The "version" could be the number of hashes in the entry, or something like that with some heuristic hacks.

I guess if you run with this idea, the DBX can be packaged into a CAB file that's on the LVFS.  And the CAB file only gets pulled down when you're missing hashes, so it's not a few hundred kb to every box.

That being said, all of this work and assumptions made are predicated upon the MS 3rd party UEFI CA remaining the root of trust for secure boot in Linux, which after what a mess BootHole was, I don't know it's the right answer for Linux in the future.

hughsie: > the DBX can be packaged into a CAB file that's on the LVFS

Yes, I think so.

>  And the CAB file only gets pulled down when you're missing hashes

You mean in the case where the firmware vendor doesn't update the dbx?

> predicated upon the MS 3rd party UEFI CA remaining the root of trust

I think that's the future we're in now, and I did think of constructing the FuDevice GUID based on the first hash of the dbx, or the PK or something. Ideas very welcome.

> I don't know it's the right answer for Linux in the future.

I think we need a generic kernel interface to configure the SB root of trust; unless this is super easy to configure nobody is going to change the CA.

superm1: >You mean in the case where the firmware vendor doesn't update the dbx?

Well I guess make an assumption that everyone is up to date from before Boothole.  Most people will get the 2020 update from dbxtool/secureboot-db in their distros.  By the time this thing could be live and could replace those, there shouldn't be a _lot_ to update.  Now once the next vulnerability gets queued up, then yes LVFS would be the source to push all the updates.

>I think that's the future we're in now, and I did think of constructing the FuDevice GUID based on the first hash of the dbx, or the PK or something. Ideas very welcome.

Based off what's in KEK I think.

>I think we need a generic kernel interface to configure the SB root of trust; unless this is super easy to configure nobody is going to change the CA.

Completely agree it needs to be a standard way to enroll a new CA.  Maybe the right answer isn't a kernel interface though, rather something with proved physical presence like creating an industry standard BIOS setup option/enrollment flow.

Dell offers a "custom" mode, but I think it's not for your average user

I would propose something like this:
* Define a location on install media that can store a PK and CA for enrollment.
* If a system has SB turned on but a user tries to boot a binary signed with different PK and CA show an error message along the lines of "`This system has secure boot enabled and this binary isn't currently trusted.  To enroll it turn off secure boot"`
* User goes into menus and turns off secure boot.  This puts the system into "UEFI Setup Mode"
* When user tries to boot the new binary while in setup mode, system sees it's signed and looks for the CA to check it against in known location on same install media.
* If CA matches, offer to enroll PK and CA and turn on secure boot (put system in Deployed Mode)
* Continue booting binary.

But then the question is what should that CA be?  The ideas that come to mind are:
* LF?  Seems like it just shifts the problem to not affect Windows boxes.  Still a vulnerability local Debian would affect a Fedora box.
* Each distro provides the CA?  Puts a lot weight upon the infrastructure of each distro and makes switching distros more difficult.

plugins/uefi-dbx/fu-uefi-dbx-common.c
Foxboron: `sbsigntools` doesn't expose the authenticode hash in any way. I wonder if it would be easier to contribute some tools to fetch it, they do implement their own PE/COFF parser library for it after all.

superm1:
Presumably this should be touching dbxtool too now right?

plugins/uefi-dbx/fu-uefi-dbx-common.c
Foxboron: Trammell Hudson implemented `--hash-only` for `sbsign`. I reckon this is what you wanted?

https://groups.io/g/sbsigntools/message/42

hughsie: That's really helpful, I've updated the patch to support `sbsign` as well now.

superm1:
closing in favor of https://github.com/fwupd/fwupd/pull/2325

PR #2318, opened 2020-08-13
Add a compatible re-implementation of the rhboot dbxtool

hughsie:
This isn't installed anywhere and is just useful for local debugging.


plugins/uefi-dbx/fu-dbxtool.c
```diff
@@ -0,0 +1,187 @@
...
+
+	bindtextdomain (GETTEXT_PACKAGE, FWUPD_LOCALEDIR);
+	bind_textdomain_codeset (GETTEXT_PACKAGE, "UTF-8");
+	textdomain (GETTEXT_PACKAGE);
+
+	/* ensure root user */
+#ifdef HAVE_GETUID
+	if (getuid () != 0 || geteuid () != 0)
+		/* TRANSLATORS: we're poking around as a power user */
+		g_printerr ("%s\n", _("This program may only work correctly as root"));
```
superm1, line 75: Since this app only support apply and list, I think that this check should only be shown with the apply method, right?

plugins/uefi-dbx/fu-dbxtool.c
```diff
@@ -0,0 +1,187 @@
...
+#endif
+
+	/* get a action_list of the commands */
+	context = g_option_context_new (NULL);
+	g_option_context_set_description (context,
+		/* TRANSLATORS: description of dbxtool */
+		_("This tool allows an administrator to apply UEFI dbx updates."));
+
+	/* TRANSLATORS: program name */
+	g_set_application_name (_("UEFI Firmware Utility"));
```
superm1, line 85: should change no?

plugins/uefi-dbx/fu-dbxtool.c
```diff
@@ -0,0 +1,187 @@
...
+	g_autoptr(FuUefiDbxFile) dbx_update = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GOptionContext) context = NULL;
+	const GOptionEntry options[] = {
+		{ "verbose", 'v', 0, G_OPTION_ARG_NONE, &verbose,
+			/* TRANSLATORS: command line option */
+			_("Show extra debugging information"), NULL },
+		{ "list", 'l', 0, G_OPTION_ARG_NONE, &action_list,
+			/* TRANSLATORS: command line option */
+			_("List entries in dbx"), NULL },
```
superm1, line 52: I think you should have two list commands - one for parsing the current dbx and one for parsing the new dbx

plugins/uefi-dbx/fu-dbxtool.c
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	/* check this is a newer dbx update */
+	checksums_update = fu_uefi_dbx_file_get_checksums (dbx_update);
+	if (!force && fu_uefi_dbx_file_has_checksums (dbx_system, checksums_update)) {
+		/* TRANSLATORS: same or newer update already applied */
+		g_printerr ("%s\n", _("Cannot apply update as this dbx update has already been applied."));
+		return EXIT_FAILURE;
+	}
+
```
superm1, line 168: I guess assuming this is merged first, the checks added as part of https://github.com/fwupd/fwupd/pull/2317 should slip right in here.

hughsie: Exactly. I think #2317 needs a lot more work, and I think a standalone dbxtool is useful without the extra tests.

plugins/uefi-dbx/fu-dbxtool.c
hughsie: Fixed.

plugins/uefi-dbx/fu-dbxtool.c
hughsie: Fixed.

plugins/uefi-dbx/fu-dbxtool.c
hughsie: Fixed, but keeping one command. If you specify `--dbx=` you list the contents of the file, otherwise you get the system values.

hughsie:
Okay, could you please re-review -- I've rewritten the `FuEfiSignature` stuff as well for the future KEK work.

plugins/uefi-dbx/fu-dbxtool.c
```diff
@@ -59,6 +60,31 @@ fu_dbxtool_get_siglist_local (const gchar *filename, GError **error)
...
+	if (bufsz == 0)
+		return FALSE;
+	tokens = fu_common_strnsplit (buf, bufsz - 1, " ", -1);
+	for (guint i = 0; args[i] != NULL; i++) {
+		if (g_strv_contains ((const gchar * const *) tokens, args[i]))
+			return TRUE;
+	}
+	return FALSE;
+}
+
```
superm1, line 87: I think this function is actually useful to all of fwupd and should go into the daemon.  For example we can use it as a hint that we shouldn't run UEFI capsule updates.

hughsie: Something like `fu_common_is_live_media()` in `libfwupdplugin/fu-common.c`?

superm1:
> Okay, could you please re-review -- I've rewritten the FuEfiSignature stuff as well for the future KEK work.

That stuff LGTM.

plugins/uefi-dbx/fu-dbxtool.c
superm1: Yeah that's what I was thinking.  Then plugins can mark devices not updatable based upon that.
I'm also not sure if that argument list is sufficient for Ubuntu 20.04+ with their casper stuff.  I know they used to use `boot=casper` in 19.10 and earlier, but I don't know for sure what happens in 20.04 and htat will need double checking.
@xnox can you comment?

xnox:
Why is this being added?

On Ubuntu we already ship sbkeysync to apply dbxupdates, and fwupd can depend on that tool, and execute it with a different keystore to apply updates that it fetches and validated.

Do we really need one more implementation?

Or why is this one better and how will it replace sbkeysync?

plugins/uefi-dbx/fu-dbxtool.c
xnox: It's best to check for presence of `/cdrom/.disk/info` to detect "installer media"

hughsie:
> Why is this being added?

Fair question. It's to provide compatibility with the older dbxtool that admins are using on RHEL and Fedora.

> On Ubuntu we already ship sbkeysync to apply dbxupdates, and fwupd can depend on that tool

No, we can't depend on that. It's not available in anything non-Ubuntu.

> Do we really need one more implementation?

If we're going to obsolete dbxtool one day then yes we need to provide a compatible CLI. The main reason we need this is that it's a stepping stone to doing dbx updates using the LVFS.

plugins/uefi-dbx/fu-dbxtool.c
hughsie: Useful, thanks. Will update the patch.

xnox:
> > Why is this being added?
> 
> Fair question. It's to provide compatibility with the older dbxtool that admins are using on RHEL and Fedora.
> 
> > On Ubuntu we already ship sbkeysync to apply dbxupdates, and fwupd can depend on that tool
> 
> No, we can't depend on that. It's not available in anything non-Ubuntu.
> 
> > Do we really need one more implementation?
> 
> If we're going to obsolete dbxtool one day then yes we need to provide a compatible CLI. The main reason we need this is that it's a stepping stone to doing dbx updates using the LVFS.

Having better tooling, that is universally available would be great.
Please ensure if and when it arrives in LVFS there is space and ability for distributions to ship their own dbxupdates, and for users to drop in their own, in addition to the ones that are fetched by lvfs.

Also ensure that there are hooks and ahead of time TPM measurements available for various tools to reseal against. I.e. snapd, Google Cloud, Azure Cloud all need to know new TPM measurements to prevent alerts and to reseal TPM secrets.

hughsie:
> there is space and ability for distributions to ship their own dbxupdates

The previous tools will continue to work, although you're right we probably need more control than `BlacklistPlugins=uefi_dbx`. 

> Also ensure that there are hooks and ahead of time TPM measurements available for various tools to reseal against

This is something we also need for firmware updates when we alter the PCR**0**. What do you do now out of interest?

plugins/uefi-dbx/fu-dbxtool.c
hughsie: Fixed.

PR #2321, opened 2020-08-13
uefi-dbx: Allow updating the dbx using the LVFS

hughsie:
The GUID is built using the SHA256 of the certificates in the KEK.

### DIFF
--- a/plugins/uefi-dbx/README.md
+++ b/plugins/uefi-dbx/README.md
@@ -4,5 +4,38 @@ UEFI dbx Support
 Introduction
 ------------
 
-This plugin checks if the UEFI dbx contains all the most recent revoked
+Updating the UEFI revocation database prevents starting EFI binaries with known
+security issues, and is typically no longer done from a firmware update due to
+the risk of the machine being "bricked" if the bootloader is not updated first.
+
+This plugin also checks if the UEFI dbx contains all the most recent revoked
 checksums. The result will be stored in an security attribute for HSI.
+
+Firmware Format
+---------------
+
+The daemon will decompress the cabinet archive and extract a firmware blob in
+EFI_SIGNATURE_LIST format.
+
+See https://www.uefi.org/sites/default/files/resources/UEFI%20Spec%202_6.pdf
+for details.
+
+This plugin supports the following protocol ID:
+
+ * org.uefi.dbx
+
+GUID Generation
+---------------
+
+These devices use the GUID constructed of the uppercase SHA256 of the X509
+certificates found in the system KEK and optionally the EFI architecture. e.g.
+
+ * `UEFI\CRT_{sha256}`
+ * `UEFI\CRT_{sha256}&ARCH_{arch}`
+
+...where `arch` is typically one of `IA32`, `X64`, `ARM` or `AA64`
+
+Vendor ID Security
+------------------
+
+The vendor ID is hardcoded to `UEFI:Microsoft` for all devices.
--- a/plugins/uefi-dbx/fu-uefi-dbx-device.c
+++ b/plugins/uefi-dbx/fu-uefi-dbx-device.c
@@ -1,0 +1,148 @@
+/*
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-efivar.h"
+
+#include "fu-efi-signature-common.h"
+#include "fu-efi-signature-parser.h"
+#include "fu-uefi-dbx-device.h"
+
+struct _FuUefiDbxDevice {
+	FuDevice		 parent_instance;
+};
+
+G_DEFINE_TYPE (FuUefiDbxDevice, fu_uefi_dbx_device, FU_TYPE_DEVICE)
+
+static gboolean
+fu_uefi_dbx_device_write_firmware (FuDevice *device,
+				   FuFirmware *firmware,
+				   FwupdInstallFlags install_flags,
+				   GError **error)
+{
+	const guint8 *buf;
+	gsize bufsz = 0;
+	g_autoptr(GBytes) fw = NULL;
+
+	/* get default image */
+	fw = fu_firmware_get_image_default_bytes (firmware, error);
+	if (fw == NULL)
+		return FALSE;
+
+	/* write entire chunk to efivarfs */
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_WRITE);
+	buf = g_bytes_get_data (fw, &bufsz);
+	if (!fu_efivar_set_data (FU_EFIVAR_GUID_SECURITY_DATABASE,
+				 "dbx", buf, bufsz,
+				 FU_EFIVAR_ATTR_APPEND_WRITE |
+				 FU_EFIVAR_ATTR_TIME_BASED_AUTHENTICATED_WRITE_ACCESS |
+				 FU_EFIVAR_ATTR_RUNTIME_ACCESS |
+				 FU_EFIVAR_ATTR_BOOTSERVICE_ACCESS |
+				 FU_EFIVAR_ATTR_NON_VOLATILE,
+				 error)) {
+		return FALSE;
+	}
+
+	/* success! */
+	return TRUE;
+}
+
+static gboolean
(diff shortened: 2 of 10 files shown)


hughsie:
The kind of metainfo file I would imagine would look like this:

    <?xml version="1.0" encoding="UTF-8"?>
    
    <component type="firmware">
      <id>org.uefi.dbx.x64.firmware</id>
      <name>UEFI dbx</name>
      <summary>UEFI Security Blocklist</summary>
      <description>
        <p>
          Updating the UEFI dbx prevents starting EFI binaries with known security issues.
        </p>
      </description>
      <provides>
        
        <firmware type="flashed">47515c95-388c-5114-adcc-07b7501d1188</firmware>
      </provides>
      <url type="homepage">https://uefi.org/revocationlistfile</url>
      <metadata_license>CC0-1.0</metadata_license>
      <project_license>proprietary</project_license>
      <developer_name>Richard Hughes</developer_name>
      <releases>
        
        <release urgency="high" version="190" date="2020-07-29">
          <checksum filename="DBXUpdate-20200729.x64.bin" target="content"/>
          <description>
            <p>
              This updates the dbx to the latest release from Microsoft.
            </p>
          </description>
          <issues>
            <issue type="cve">CVE-2020-10713</issue>
          </issues>
        </release>
      </releases>
      
      <requires>
        <id compare="ge" version="1.4.6">org.freedesktop.fwupd</id>
      </requires>
      
      <keywords>
        <keyword>boothole</keyword>
      </keywords>
    </component>

plugins/uefi-dbx/README.md
```diff
@@ -4,5 +4,33 @@ UEFI dbx Support
 Introduction
 ------------
 
-This plugin checks if the UEFI dbx contains all the most recent revoked
+Updating the UEFI dbx prevents starting EFI binaries with known security issues,
```
superm1, line 7: probably should mention what dbx actually is 🗡️

hughsie:
One thing I'm not sure about. Would the `aa64` and `ia32` Microsoft X509 certificate found in the KEK be the same as the `x64` one?

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -0,0 +1,153 @@
...
+		for (guint j = 0; j < sigs->len; j++) {
+			FuEfiSignature *sig = g_ptr_array_index (sigs, j);
+			fu_device_add_instance_id (device, fu_efi_signature_get_checksum (sig));
+		}
+	}
+	return fu_uefi_dbx_device_set_version_number (device, error);
+}
+
+static void
+fu_uefi_dbx_device_init (FuUefiDbxDevice *self)
```
superm1, line 122: I think it makes sense to set the parent of this device as system firmware

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -0,0 +1,153 @@
...
+{
+	fu_device_set_physical_id (FU_DEVICE (self), "dbx");
+	fu_device_set_name (FU_DEVICE (self), "UEFI dbx");
```
superm1, line 125: To make this more user friendly how about "UEFI Revocation database"

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -0,0 +1,153 @@
...
+	fu_device_set_summary (FU_DEVICE (self), "UEFI dbx Virtual Device");
+	fu_device_set_vendor (FU_DEVICE (self), "Microsoft");
```
superm1, line 127: I don't think this is reasonable to set here - should be looking at what's in KEK and export it based on that (it would actually be Microsoft Corporation for most systems)

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: That's a better summary for sure.

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: Good idea, fixed.

plugins/uefi-dbx/README.md
hughsie: I used `revocation database` again here, thanks.

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: > should be looking at what's in KEK

I'm not sure I actually want to be *parsing* the X509 certificate, although I suppose we might have to start doing that in the future for other things. If we did the vendor would be `Microsoft Corporation KEK CA 2011` which probably isn't what the user would expect. Maybe it's best we don't set a vendor at all. Keep reading...

The more important one to get right is the `vendor_id` and with a brain turned on hardcoding `UEFI:Microsoft` isn't smart either. I briefly thought we want to use the KEK checksum again, although then the question becomes _which_ certificate in the KEK should be used. I think hardcoding a heuristic would be pretty nasty given the OEM usually puts a cert there too.

Given I think the way this should work is by having a super-locked down LVFS vendor group (with just the DBX updates there, only pushable by LVFS admins...) so maybe having a vendor_id of `"LVFS"` or `"unspecified"` isn't so bad.

superm1:
>    

This seems like it's asking for an argument to `dbxtool` to do exactly that!

>        <issue type="cve">CVE-2020-10713</issue>

There are a bunch more than that, no?

hughsie:
> This seems like it's asking for an argument to dbxtool to do exactly that!

Not sure what you mean, sorry.

>       <issue type="cve">CVE-2020-10713</issue>
> There are a bunch more than that, no?

Sure, it was just an example. It actually worked when built into a .cab file too :hot_face:

plugins/uefi-dbx/fu-uefi-dbx-device.c
superm1: >I'm not sure I actually want to be parsing the X509 certificate, although I suppose we might have to start doing that in the future for other things. If we did the vendor would be Microsoft Corporation KEK CA 2011 which probably isn't what the user would expect. Maybe it's best we don't set a vendor at all. Keep reading...

I think it would be the "O" field not the "CN" field.
`Subject: C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Corporation KEK CA 2011`

>Given I think the way this should work is by having a super-locked down LVFS vendor group (with just the DBX updates there, only pushable by LVFS admins...) so maybe having a vendor_id of "LVFS" or "unspecified" isn't so bad.

I think until there is a third party we know of actually _uploading_ a dbx to LVFS, how about "Linux Foundation"?

superm1:
>Not sure what you mean, sorry.

I mean adding an argument to dbxtool that parses a file, counts up the microsoft GUIDs and spits out that number.  Like `--version`

hughsie:
> I mean adding an argument to dbxtool that parses a file, counts up the microsoft GUIDs and spits out that number.

Yes, that makes sense indeed. Will ponder after some sleep.

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: > I think it would be the "O" field

Of which certificate tho?

> how about "Linux Foundation"?

So `UEFI:Linux Foundation` with a _space_? I guess that's okay, it just looked weird.

superm1: >Of which certificate tho?

I pulled that from the one in key 2 in my KEK

>So UEFI:Linux Foundation with a space? I guess that's okay, it just looked weird.

Dell's is `(DMI: Dell Inc.)`

hughsie: > I pulled that from the one in key 2 in my KEK

But how did you decide to use the 2nd key and not the 1st?

hughsie:
> Not actually sure. Maybe peter knows.

He seems to think it should be the same key. I've added the EFI arch in the GUID so we can disambiguate them.

hughsie:
> I mean adding an argument to dbxtool that parses a file, counts up the microsoft GUIDs

Done.

PR #2325, opened 2020-08-17
uefi-dbx: Validate the dbx update is safe to apply

hughsie:
To do this mount all ESP partitions and check all the binaries there to see if
they match any entries in the new dbx. If we applied the update when a hash
matched, we would unintentially 'brick' the users machine, as the grub and shim
binaries *have* to be updated first.

This functionality does reimplement the PE hashing functionality found in
sbsigntools and pesign. This was done for 4 main reasons:

 * There were some memory safety issues found when fuzzing random binaries
 * Executing the tools hundreds of times was a lot of overhead
 * Operating from a blob of immutable mmap'd memory is much faster
 * We only need a very small amount of functionality from both tools


hughsie:
Screenshot from 2020-08-17 15-10-22

hughsie:
For my future self, the metainfo.xml files are here: https://github.com/hughsie/dbx-firmware

hughsie:
Yes, you're right, they belong in the fwupd vendor group. I've moved them to here https://github.com/fwupd/dbx-firmware

PR #2326, opened 2020-08-17
Install the new dbx utility as fwupd-dbxtool

hughsie:


hughsie:
Of course, the other name choice is `dbxtool2`

hughsie:
Or, even install to `$libexec/fwupd/` and then obsolete the upstream project if the higher-ups agree. Certainly from RHEL's point of view adding obsoletes in a stable series in a lot of "paperwork".

hughsie:
@vathpela agreed that it was right to obsolete dbxtool -- so I've adapted the name with that in mind.

cgwalters:
AIUI previously in Fedora 32, `dbxtool.service` ran by default and updated the DBX database.  With this move, because (again AIUI) no fwupd operations happen until an agent initiates them, it means that the DBX will *not* be updated by default when the operating system boots unless an agent like gnome-software initiates it.  Is that correct?

superm1:
Matches my expectations yes.

PR #2327, opened 2020-08-17
Add X-Configuration to use for dbx updates

hughsie:
This would fall back to X-System.


hughsie:
@superm1 I didn't actually mean to merge this; any concerns? Happy to revert if required.

superm1:
It's fine

Issue #1956, opened 2020-04-06
Please also install the unit tests


rossburton:
fwupd has two test suites: a set of unit tests (as ran by `meson test`) and an integration test (which is installed).

In some situations, such as in cross-compilation, it's useful to be able to also install the unit tests so that they can be executed on the target.  Please consider also installing the unit tests.

superm1:
Several projects have started to do this, I don't think there is any problem with it.  If you've got some cycles, definitely happy to review a PR.

FFY00:
I think this should be optional, adding a meson option should be pretty easy.

hughsie:
Where would we install the local self tests? Would they be called by a new `installed-test` as well?

superm1:
For libjcat we did:
`/usr/libexec/installed-tests/libjcat/`

superm1:
should be fixed in master now

superm1:
Something to note, as part of this change many binaries were removed from tree and won't be installed.  You can fetch them at runtime using `contrib/ci/get_test_firmware.sh`

PR #2330, opened 2020-08-19
Install more installed-tests

hughsie:
Fixes some of https://github.com/fwupd/fwupd/issues/1956


hughsie:
The problem with this, of course, is that we're now including random files with unclear licences into the debian subpackage -- e.g. `DMAR-OPTOUT`, `Samsung SSD 860 EVO 500GB.bin`, `dev_VRBRAIN.dfu` and `TOSHIBA_THNSN5512GPU7.bin` -- is that a blocker or are these kinds of proprietary blobs obtained legitimately actually okay? I don't think its much use generating our own test files for this kind of thing, else we're not really testing on "true" data.

hughsie:
@superm1 ^^

superm1:
To be honest I think blobs like that don't have redistribution licenses shouldn't be in tree or in the tarball releases in the first place, but downloaded on demand as part of CI from another place. And let tests get by without them (skip tests if they're missing).

superm1:
Maybe @ximion can weigh in here though?

ximion:
It's actually a matter of debate in the Debian community whether "random binary files used only for testing, but without source" is actually okay. To me, consensus seems to be that those are okay if it's obvious how they were made and their license is clear. (A much more complicated subsection of this issue is the "source" of pretrained neural networks, not sure if that has a resolution yet).
But @hughsie saying "random files with unclear licences" is a pretty big red flag, and I am pretty sure that Debian can't (and won't) ship those in `main`. I am sure similar reasons apply to those files as they do to non-free firmware, which Debian also ships in non-free: https://packages.debian.org/bullseye/firmware-linux-nonfree

Downloading these files on-demand for testing would likely circumvent that problem, but you couldn't do that during package build time but it would have to be done in an autopkgtest.

superm1:
@ximion so lets take for example a sample "DMAR" file.  This is captured by `cp /sys/firmware/acpi/tables/DMAR $tree_place` on a sample machine.  In that case the license is unclear, but how it was obtained is very clear.  We've got a few cases like that which are in the source package today to run tests, but not in any binary package.  If those gotta go, I'm entirely fine with it moving to autopkgtest.  Presumably we could actually use the same script for Travis/Circle CI/autopkgtest/installed tests to use at runtime.

ximion:
Where does that ACPI data come from? Likely the device vendor, so maybe some licensing applies. That Debian example I mentioned was more for random-ish data in specific formats (e.g. generated by a fuzzing tool) or for things like weather data or temperature measurements.
The way to 100% know for sure whether this is okay within Debian is to ask on debian-devel or, for the less public option, on debian-legal (debian-legal IMHO is usually the best source of advice). Personally I would probably just download that data from the web for testing on-demand, just in case there are license claims (files can easily be modified or removed then) and simply because it is more straightforward to do that a possibly bigger discussion on debian-devel (I am also quite sure that debian-legal would say "this doesn't go into main!" once you mention that you are not sure about the license - because that would make the binary *potentially* unredistributable, and Debian really doesn't like to take that kind of risk).

superm1:
>Where does that ACPI data come from?

That depends on how many steps back you want to go.  The kernel?  The OEM?  An ODM?  IHV reference code?  Lots of things touch it.

>The way to 100% know for sure whether this is okay within Debian is to ask on debian-devel or, for the less public option, on debian-legal (debian-legal IMHO is usually the best source of advice). Personally I would probably just download that data from the web for testing on-demand, just in case there are license claims (files can easily be modified or removed then) and simply because it is more straightforward to do that a possibly bigger discussion on debian-devel (I am also quite sure that debian-legal would say "this doesn't go into main!" once you mention that you are not sure about the license - because that would make the binary potentially unredistributable, and Debian really doesn't like to take that kind of risk).

I agree; I think I'd rather avoid those conversations and move all the binaries out of fwupd-X.Y.Z.tar.gz and pull on demand.  @hughsie that sound OK to you?  Smaller tarball for releases!  I'll help sort out the autopkgtest changes.

superm1:
So the above commit I just pushed is an example of what I think we should do (at scale with the rest of tree).
I'm thinking land all the binaries into https://github.com/fwupd/fwupd-test-firmware/tree/wip/superm1/move-binaries and then that gets pulled down as part of CI and overlaid on the tree here back in the old locations, or usable as part of installed-tests if cloned locally.

ximion:
Could the test auto-download the file if it is missing and only skip the test if a `FWUPD_TESTS_NO_NET` environment variable is also set? Or the other way round, skip the test only unless a `FWUPD_TESTS_ALLOW_NET` var is set. That would prevent tests being semi-silently skipped when the user expects them to happen (I had this a few times and it took a while to notice that a test was skipped because of an error and not because it wasn't supposed to happen intentionally).

hughsie:
> and overlaid on the tree here back in the old locations

Seems sane to me.

superm1:
>Could the test auto-download the file if it is missing and only skip the test if a FWUPD_TESTS_NO_NET environment variable is also set? Or the other way round, skip the test only unless a FWUPD_TESTS_ALLOW_NET var is set. That would prevent tests being semi-silently skipped when the user expects them to happen (I had this a few times and it took a while to notice that a test was skipped because of an error and not because it wasn't supposed to happen intentionally).

I'm not sure it buys a lot other than more code TBH.  The major cases this will matter are CI, autopkgtest and tests run on buildds.
* On buildds: They'll run without the binaries every time
* on CI: they'll run with binaries every time
* On autopkgtest: They'll run with binaries every time

Only time it's up in the air is installed-tests, which I don't think a lot of people will use anyway..

ximion:
Adding the explicit method would protect against scenarios like renaming files but not adjusting their name in the code, accidentally removed files or download failures, or people not using the tests right. It would be an additional protection against accidental coding errors leading to unexpected test-skips.
It's not a must-have for me though, and also definitely hughsie's call anyway. I just found a mechanism like this simple to add and pretty useful in the past (granted, the project where it mattered was a way less straightforward case).

hughsie:
FYI, I think one mega commit is fine for this.

superm1:
> FYI, I think one mega commit is fine for this.

OK.  Will squash once it's all actually working :)

superm1:
@hughsie I got it working everywhere but on Fedora.  Seems like it's related to binaries not being put into the tarball maybe.  Help?

hughsie:
>  Seems like it's related to binaries not being put into the tarball maybe

Are you trying to put them into the .rpm file itself? If so, why? I don't think we can ship them in Fedora using the same logic anyway.

superm1:
> Are you trying to put them into the .rpm file itself? If so, why? I don't think we can ship them in Fedora using the same logic anyway.

Well they have to be to run in the self tests if it's supposed to be in tree no?  Yeah, so if you can advise the right thing to do here..

hughsie:
> Well they have to be to run in the self tests if it's supposed to be in tree no?

I think the installed tests are nice to have, but don't blow an artery over them. The self tests are way more valuable IMHO.

superm1:
> I think the installed tests are nice to have, but don't blow an artery over them. The self tests are way more valuable IMHO.

Well this breaks the self tests in Fedora too.. It's not the installed tests part that fails, it's the self test during build (because of the CI flag)

plugins/nvme/fu-self-test.c
```diff
@@ -16,14 +16,15 @@ fu_nvme_cns_func (void)
 {
 	gboolean ret;
 	gsize sz;
+	const gchar *ci = g_getenv ("CI");
```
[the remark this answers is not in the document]
superm1: I personally think it's confusing to read.

superm1:
@hughsie OK so I've got it all working now.  If this looks good i'll merge fwupd-test-firmware, and adjust the branch it pulls from to master.

superm1:
https://github.com/fwupd/fwupd-test-firmware/pull/1
Will adjust this branch once that's merged.

data/installed-tests/fwupd.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+
+exec 2>&1
+dirname=`dirname $0`
+
+# ACPI DMAR
+$dirname/acpi-dmar-self-test
+rc=$?; if [[ $rc != 0 ]]; then exit $rc; fi
+
+# xxxxxxxxxxxxxxxxx
```
hughsie, line 10: maybe we just nuke the `xxxxx`'s?

superm1: fixed

contrib/debian/tests/ci
```diff
@@ -2,4 +2,5 @@
 set -e
 sed "s,^DisabledPlugins=.*,DisabledPlugins=," -i /etc/fwupd/daemon.conf
 sed "s,^VerboseDomains=.*,VerboseDomains=*,"  -i /etc/fwupd/daemon.conf
-gnome-desktop-testing-runner fwupd
+git clone https://github.com/fwupd/fwupd-test-firmware
+CI_NETWORK=true gnome-desktop-testing-runner fwupd
```
superm1, line 6: I guess this is missing a line to export the path to the test firmware still

superm1: `export G_TEST_SRCDIR=`pwd`/fwupd-test-firmware/installed-tests`

plugins/synaptics-prometheus/meson.build
```diff
@@ -31,8 +31,9 @@ shared_module('fu_plugin_synaptics_prometheus',
 )
 
 if get_option('tests')
-  testdatadir = join_paths(meson.current_source_dir(), 'data')
-  cargs += '-DTESTDATADIR="' + testdatadir + '"'
+  promtest = environment()
```
hughsie, line 34: not testdatadirs?

plugins/tpm-eventlog/meson.build
```diff
@@ -27,8 +27,9 @@ shared_module('fu_plugin_tpm_eventlog',
 )
 
 if get_option('tests')
-  testdatadir = join_paths(meson.current_source_dir(), 'tests')
-  cargs += '-DTESTDATADIR="' + testdatadir + '"'
+  tpmeventtest = environment()
```
hughsie, line 30: testdatadirs again?

hughsie:
Two little niggles, then LGTM, thanks!

plugins/synaptics-prometheus/meson.build
superm1: I was running into problems re-using the same variable across files..

hughsie: unexpected. I know there's only a "global scope" in meson, but i don't think re-defining should affect things...

superm1: Well I think it was an order of events thing.  Meson builds up all the variables, compiles things, and then runs things.  Since these variables are very specific to a specific directory failures ensued.

hughsie: I think then we should do something consistent for all the plugins. Perhaps `prometheus_env` or something?

Issue #2328, opened 2020-08-18
PCI BCR Security checks in Intel Bay Trail have incorrect offset


dnoliver:
**Describe the bug**

The current BCR offset defined in https://github.com/fwupd/fwupd/blob/master/plugins/pci-bcr/fu-plugin-pci-bcr.c#L17 is not correct for Bay Trail. 
The correct offset for Bay Trail is `0x54h`
Having and incorrect offset makes the 3 checks for the BIOS Lock functionality to be not correct.

The current output of `lspci` in the test device is:

```shell
lspci -xxx -nn -s0:1f.0
00:1f.0 ISA bridge [0601]: Intel Corporation Atom Processor Z36xxx/Z37xxx Series Power Control Unit [8086:0f1c] (rev 11)
00: 86 80 1c 0f 07 00 10 02 11 00 01 06 00 00 80 00
10: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
20: 00 00 00 00 00 00 00 00 00 00 00 00 86 80 1c 0f
30: 00 00 00 00 e0 00 00 00 00 00 00 00 00 00 00 00
40: 03 04 00 00 02 30 d0 fe 03 05 00 00 02 c0 d0 fe
50: 02 80 d0 fe 02 10 d0 fe 02 00 f0 fe 02 50 d0 fe
60: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
70: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
90: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
a0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
c0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
d0: 00 00 00 00 00 00 00 00 cf ff 00 00 00 00 00 00
e0: 09 00 0c 10 00 00 00 00 00 00 00 00 00 00 00 00
f0: 01 c0 d1 fe 00 00 00 00 1a 0f 13 01 03 03 00 00
```

Running the `fwupdtool security --force` (with the 0xdc) on this platform yields:

```shell
HSI-1
[...]

PR #2332, opened 2020-08-20
pcb-bcr: Use the correct BCR register for Bay Trail CPUs

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2328

### DIFF
--- a/plugins/pci-bcr/fu-plugin-pci-bcr.c
+++ b/plugins/pci-bcr/fu-plugin-pci-bcr.c
@@ -11,10 +11,10 @@
 
 struct FuPluginData {
 	gboolean		 has_device;
+	guint8			 bcr_addr;
 	guint8			 bcr;
 };
 
-#define BCR			0xdc
 #define BCR_WPD			(1 << 0)
 #define BCR_BLE			(1 << 1)
 #define BCR_SMM_BWP		(1 << 5)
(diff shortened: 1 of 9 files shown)


plugins/pci-bcr/fu-plugin-pci-bcr.c
```diff
@@ -11,20 +11,45 @@
...
+{
+	FuPluginData *priv = fu_plugin_get_data (plugin);
+	guint32 eax = 0;
+	guint64 tmp;
+	g_autofree gchar *group = NULL;
+
+	/* has a quirk */
+	if (!fu_common_cpuid (0x1, &eax, NULL, NULL, NULL, error))
+		return FALSE;
+	group = g_strdup_printf ("DeviceInstanceId=CPUID\\PRO_%01X&FAM_%01X&MOD_%01X",
```
superm1, line 44: As a another thought - shouldn't this populate from the cpu plugin?  Like CPU plugin can carry the quirk and some metadata w/ `BcrAddr`  And this plugin can lookup the device from the CPU plugin?

hughsie: So we'd add the GUIDs to the `FuCpuDevice` and then set up a `fu_plugin_device_registered()` in the `pci-bcr` plugin, then check if the device has some kind of metadata using `fu_device_get_metadata_integer()` and if so set the `bcr_addr`?

superm1: Exactly!

hughsie: I've updated the patch with this suggestion; I agree it's a better design.

PR #2334, opened 2020-08-23
Wip/hughsie/cpuid

hughsie:


plugins/cpu/fu-cpu-device.c
```diff
@@ -153,34 +153,61 @@ fu_cpu_device_init (FuCpuDevice *self)
...
 	fu_device_add_instance_id (device, devid3);
 	return TRUE;
 }
 
+static gboolean
+fu_cpu_device_probe (FuDevice *device, GError **error)
+{
+	if (!fu_cpu_device_add_instance_ids (device, error))
+		return FALSE;
+	return TRUE;
```
superm1, line 243: maybe just `return fu_cpu_device_add_instance_ids();`?

plugins/cpu/fu-cpu-device.c
```diff
@@ -35,6 +35,64 @@ fu_cpu_device_to_string (FuDevice *device, guint idt, GString *str)
 				    fu_cpu_device_has_flag (self, FU_CPU_DEVICE_FLAG_SMAP));
 }
 
+static const gchar *
+fu_cpu_device_convert_vendor (const gchar *vendor)
```
superm1, line 39: Were these sourced from kernel?  Might be good to reference in a comment where to get updates from

plugins/cpu/fu-cpu-device.c
```diff
@@ -92,37 +100,149 @@ fu_cpu_device_init (FuCpuDevice *self)
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_INTERNAL);
 	fu_device_add_icon (FU_DEVICE (self), "computer");
 	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_HEX);
+	fu_device_set_physical_id (FU_DEVICE (self), "cpu:0");
```
superm1, line 103: This shouldn't be hardcoded in case of a multi-socket system no?

plugins/cpu/fu-cpu-device.c
hughsie: I know it looks janky at the moment, but the later patches insert other things in here too.

plugins/cpu/fu-cpu-device.c
hughsie: Wikipedia! Is that good enough?

plugins/cpu/fu-cpu-device.c
hughsie: I'm explicitly just querying the first physical processor;I don't think it's super interesting to enumerate all of them like before.

plugins/cpu/fu-cpu-device.c
superm1: Probably..

PR #2336, opened 2020-08-25
Wip/hughsie/async

hughsie:
More to do, but I wanted to check the concept. These async functions allow us to unblock the UI thread in gnome-firmware without using threads.


superm1:
Sure.  Seems like a good idea to me

hughsie:
@superm1 do we want this in 1.4.6 or 1.5.0? It's a metric boat-load of new code.

superm1:
> @superm1 do we want this in 1.4.6 or 1.5.0? It's a metric boat-load of new code.

I mean basically the approach has been everything that's not related to security attributes is going on 1_4_X right?

hughsie:
> I mean basically the approach has been everything that's not related to security attributes is going on 1_4_X right?

Yes, but I'm getting a bit sick of backporting things and not exactly having a stable branch.This backport would be a huge mess as it would have to be merged,cherry picked, manually reverted a few hunks and then fixed'up manually.

I'm tempted to just roll a 1.5.0 but not "announce" the HSI work yet. The `--force` kinda shows this isn't ready yet.

hughsie:
> This backport would be a huge mess

Yes, thinking about it a bit more, this is not just a few tens of lines of similar code, it's basically a rework of the FwupdClient internals. I think the plan should be include this on master only, and just use master in gnome-firmware -- putting off the 1.5.0 release a bit further.

hughsie:
@superm1 do you want to review the new code? No is a valid answer, although a "yes" might need a few beers and a quiet night.

superm1:
>@superm1 do you want to review the new code? No is a valid answer, although a "yes" might need a few beers and a quiet night.

*+5,339 −1,450 * 
Yikes!  I'll give it some high level glances, but probably not as in depth as I usually look.

superm1:
> > This backport would be a huge mess
> 
> Yes, thinking about it a bit more, this is not just a few tens of lines of similar code, it's basically a rework of the FwupdClient internals. I think the plan should be include this on master only, and just use master in gnome-firmware -- putting off the 1.5.0 release a bit further.

OK

libfwupd/fwupd-client.c
```diff
@@ -428,7 +366,90 @@ fwupd_client_connect (FwupdClient *self, GCancellable *cancellable, GError **err
...
+
+/**
+ * fwupd_client_connect_async:
+ * @self: A #FwupdClient
+ * @cancellable: the #GCancellable, or %NULL
+ * @callback: the function to run on completion
+ * @callback_data: the data to pass to @callback
+ *
+ * Sets up the client ready for use. This is probably the first method you call
+ * when wanting to use libfwupd.
```
superm1, line 408: "wanting to use libfwupd in an asynchronous manner"

libfwupd/fwupd-client.c
```diff
@@ -428,7 +366,90 @@ fwupd_client_connect (FwupdClient *self, GCancellable *cancellable, GError **err
...
+ *
+ * Since: 1.5.0
+ **/
+void
+fwupd_client_connect_async (FwupdClient *self, GCancellable *cancellable,
```
superm1, line 413: I'm not sure an async design you want a separate connect method.  By definition connecting is supposed to be synchronous.  If this doesn't finish but you've already called something that expectes you to have connected isn't that a problem?

hughsie: It forces initialization into a three step process (which is actually a good thing). Let me explain a bit:

* phase 1 is app startup, loading GUIs etc
* phase 2 is contacting fwupd and connecting the proxy -- showing an error dialog if not available
* phase 3 is using the proxy and firing off all the requests async to be completed whenever

You can see all this in action in https://gitlab.gnome.org/hughsie/gnome-firmware-updater/-/blob/wip/hughsie/async/src/gfu-main.c#L1455

superm1: Ahhh.. seeing `fwupd_client_connect_finish` in action that flow makes a lot more sense.

Why not also bring fu-util into the same approach?   I guess maybe not as valuable since there isn't a GUI to click and interact that gets blocked while waiting for stuff to get done.

libfwupd/fwupd-client.c
```diff
@@ -1309,323 +1887,533 @@ fwupd_client_install_fd (FwupdClient *self,
...
+
+	/* if a remote-id was specified, the remote has to exist */
+	remote = fwupd_client_get_remote_by_id_finish (FWUPD_CLIENT (source), res, &error);
+	if (remote == NULL) {
+		g_task_return_error (task, g_steal_pointer (&error));
+		return;
+	}
+
+	/* local and directory remotes may have the firmware already */
+	if (fwupd_remote_get_kind (remote) == FWUPD_REMOTE_KIND_LOCAL /*&& uri == NULL*/) {
```
superm1, line 2167: didn't you lose the `uri == NULL` from the original function?

libfwupd/fwupd-client.c
hughsie: Fixed, thanks.

libfwupd/fwupd-client.c
hughsie: > that flow makes a lot more sense

I've added some more doctext to the fwupd_client_connect_async() method making this a bit more clear.

> Why not also bring fu-util into the same approach

libfwupd/fwupd-client.c
hughsie: Agh yes, thanks. Good catch! Late night coding. I've restored it back so it worked the same was a before.

superm1 (approved):

PR #2338, opened 2020-08-27
Fix building GUsb as a subproject with introspection enabled

hughsie:


superm1:
I guess for Windows this needs some changing though, only add to introspection deps if introspection is turned on.

PR #2339, opened 2020-08-28
trivial: dell-dock: check for valid dock type on open

superm1:
Only add instance ID if it actually probes properly.

Otherwise this makes an invalid assumption that the device is a WD19
EC just because it had the correct hub in front.

Instead check the first time it's opened that the correct device
is identified (`EXPECTED_DOCK_TYPE`)

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-dell-dock-i2c-ec.c
+++ b/plugins/dell-dock/fu-dell-dock-i2c-ec.c
@@ -301,8 +301,9 @@ fu_dell_dock_ec_write (FuDevice *device, gsize length, guint8 *data, GError **er
 static gboolean
 fu_dell_dock_is_valid_dock (FuDevice *device, GError **error)
 {
-	g_autoptr(GBytes) data = NULL;
 	const guint8 *result = NULL;
+	gsize sz = 0;
+	g_autoptr(GBytes) data = NULL;
 
 	g_return_val_if_fail (device != NULL, FALSE);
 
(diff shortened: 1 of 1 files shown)


plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -312,12 +312,24 @@ fu_dell_dock_is_valid_dock (FuDevice *device, GError **error)
...
-		g_set_error_literal (error, FWUPD_ERROR, FWUPD_ERROR_NOT_FOUND,
+	if (result == NULL) {
+		g_set_error_literal (error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED,
 				     "No valid dock was found");
 		return FALSE;
 	}
-	return TRUE;
+
+	/* this will trigger setting up all the quirks */
+	if (*result == EXPECTED_DOCK_TYPE) {
```
hughsie, line 322: are we sure *result is valid? I'd feel a lot more content getting the size from `g_bytes_get_data` and also checking the `sz >= 1`

hughsie:
One niggle then LGTM, thanks.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
superm1: Right before checked that result wasn't NULL

superm1: But I guess that's a cleaner style way to check via size. I'll adjust.

superm1: fixed

hughsie (approved):

PR #2340, opened 2020-08-28
Don't run debian packaging when subproject is active

superm1:
Cleaner fix for https://github.com/fwupd/fwupd/pull/2289
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
"I have no power here" :)

PR #2341, opened 2020-08-28
Do not return HSI attributes when running in a VM

hughsie:


src/fu-main.c
```diff
@@ -1662,6 +1670,16 @@ fu_main_load_introspection (const gchar *filename, GError **error)
...
 }
 
+static gboolean
+fu_main_is_hypervisor (void)
+{
+	g_autofree gchar *buf = NULL;
+	gsize bufsz = 0;
+	if (!g_file_get_contents ("/proc/cpuinfo", &buf, &bufsz, NULL))
+		return FALSE;
+	return g_strstr_len (buf, (gssize) bufsz, "hypervisor") != NULL;
```
superm1, line 1680: Is this valid for all VMs?  Or does there need to be more ways to detect?

src/fu-main.c
```diff
@@ -1105,6 +1106,13 @@ fu_main_daemon_method_call (GDBusConnection *connection, const gchar *sender,
 	if (g_strcmp0 (method_name, "GetHostSecurityAttrs") == 0) {
 		g_autoptr(FuSecurityAttrs) attrs = NULL;
 		g_debug ("Called %s()", method_name);
+		if (priv->is_vm) {
+			g_dbus_method_invocation_return_error_literal (invocation,
+								       FWUPD_ERROR,
+								       FWUPD_ERROR_NOT_SUPPORTED,
+								       "HSI unavailable for hypervisor");
```
superm1, line 1113: I think the same thing should probably apply for containers too.

src/fu-main.c
hughsie: I think the kernel has a pretty long list of ways to detect a VM.

src/fu-main.c
hughsie: Oof, that's harder. Do you know how to detect that?

superm1: https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker seems to have a few ways to do it

exploide: `systemd-detect-virt` has also implemented such checks: https://github.com/systemd/systemd/blob/5b14956385ee1076b98c3ca730bb9c1c9e95fc22/src/basic/virt.c#L450

hughsie: I've added something that should catch 99% of people trying this.

PR #2344, opened 2020-08-30
build: fix systemd_root_prefix for systemd 246

jtojnar:
systemd 246 replaced the variable names (https://github.com/systemd/systemd/commit/4908de44b0a0409f84a7cdc5641b114d6ce8ba03).
The old ones are kept for backwards compatibility but some of them interpolate new variable names so redefining them breaks.
Curiously, only systemdshutdowndir is affected by that here.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -378,7 +378,7 @@ if build_standalone and get_option('systemd')
   else
     systemdunitdir = systemd.get_pkgconfig_variable('systemdsystemunitdir', define_variable: ['rootprefix', systemd_root_prefix])
     systemdsystempresetdir = systemd.get_pkgconfig_variable('systemdsystempresetdir', define_variable: ['rootprefix', systemd_root_prefix])
-    systemd_shutdown_dir = systemd.get_pkgconfig_variable('systemdshutdowndir', define_variable: ['rootprefix', systemd_root_prefix])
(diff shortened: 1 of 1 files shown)


dasj19 (approved):

FFY00 (commented):
Shouldn't you do this in all the variables?

jtojnar:
As mentioned above, only shutdowndir uses `root_prefix`, others still use `rootprefix`.

FFY00:
That is irrelevant, because
```
rootprefix=${root_prefix} 
```

The commit says that all names without prefixes (the ones that are not using underscores) are deprecated:
```
 {systemd,udev}.pc: add names with underscores, deprecate the old ones

The names with multiple lowercase words run together are hard to read.  We
started that way with very short names like rootprefix, but then same pattern
was applied to longer and longer names. Looking at the body of .pc files
available on my machine, many packages use underscores; let's do the same. Old
names are kept for compatiblity, so this is backwards compatible.
```

`systemd_system_preset_dir` using the `rootprefix` variable instead of `root_prefix` internally is just a implementation detail. As the commit says, you should use `root_prefix`. You should also use `systemd_system_preset_dir`, `systemd_system_unit_dir` instead of `systemdsystempresetdir`, `systemdsystemunitdir`, respectively :blush:

jtojnar:
You are right, using `root_prefix` everywhere would be nicer. I misremembered how the `rootprefix` itself was changed.

PR #2348, opened 2020-08-31
dfu: Support polling the status from device in dfuManifest state

zhj8407:
Some devices may accumulate the firmware image and perform the
entire reprogramming operation at one time. In this case, the
device enters dfuMANIFEST-SYNC or dfuMANIFEST state after
dfuDNLOAD-IDLE.

The fwupd shall be able to poll the status from the device via
DFU_GETSTATUS until the device completes the reprogramming or
reports an error.

For details, please refer to Section 7. Manifestation Phase and
A.1 Interface State Transition Diagram in the USB DFU protocol.
https://www.usb.org/sites/default/files/DFU_1.1.pdf

For not affecting the other DFU capable devices, introduce a quirk
"poll-in-manifest" to limit the logic.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/dfu-device.c
+++ b/plugins/dfu/dfu-device.c
@@ -44,6 +44,7 @@
  * * `legacy-protocol`:		Use a legacy protocol version
  * * `detach-for-attach`:	Requires a DFU_REQUEST_DETACH to attach
  * * `absent-sector-size`:	In absence of sector size, assume byte
+ * * `manifest-poll`:		Requires polling via GetStatus in dfuManifest state
  *
  * Default value: `none`
  *
--- a/plugins/dfu/dfu-target.c
+++ b/plugins/dfu/dfu-target.c
@@ -30,6 +30,8 @@
 
 #include "fwupd-error.h"
 
+#define DFU_TARGET_MANIFEST_MAX_POLLING_TRIES	200
+
 static void dfu_target_finalize			 (GObject *object);
 
 typedef struct {
@@ -470,6 +472,50 @@ dfu_target_status_to_error_msg (DfuStatus status)
 	if (status == DFU_STATUS_ERR_STALLDPKT)
 		return "Device stalled an unexpected request";
 	return NULL;
+}
+
+static gboolean
+dfu_target_manifest_wait (DfuTarget *target, GError **error)
+{
+	DfuTargetPrivate *priv = GET_PRIVATE (target);
+	DfuStatus status;
+	guint polling_count = 0;
+
+	/* get the status */
+	if (!dfu_device_refresh (priv->device, error))
+		return FALSE;
+
+	/* wait for DFU_STATE_DFU_MANIFEST to not be set */
+	while (dfu_device_get_state (priv->device) == DFU_STATE_DFU_MANIFEST_SYNC ||
+	       dfu_device_get_state (priv->device) == DFU_STATE_DFU_MANIFEST) {
+		g_debug ("waiting for DFU_STATE_DFU_MANIFEST to clear");
+
+		if (polling_count++ > DFU_TARGET_MANIFEST_MAX_POLLING_TRIES) {
+			g_set_error_literal (error,
+					     FWUPD_ERROR,
+					     FWUPD_ERROR_INTERNAL,
+					     "reach to max polling tries");
+			return FALSE;
+		}
+
+		g_usleep ((dfu_device_get_download_timeout (priv->device) + 1000) * 1000);
+		if (!dfu_device_refresh (priv->device, error))
+			return FALSE;
+	}
+
+	/* in an error state */
+	if (dfu_device_get_state (priv->device) == DFU_STATE_DFU_ERROR) {
+		status = dfu_device_get_status (priv->device);
+		/* use a proper error description */
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     dfu_target_status_to_error_msg (status));
+
+		return FALSE;
+	}
+
+	return TRUE;
 }
 
 gboolean
@@ -1284,6 +1330,12 @@ dfu_target_download (DfuTarget *target, DfuImage *image,
 		if (!ret)
 			return FALSE;
 	}
+
+	if (fu_device_has_custom_flag (FU_DEVICE (dfu_target_get_device (target)),
+				       "manifest-poll") &&
+	    dfu_device_has_attribute (priv->device, DFU_DEVICE_ATTRIBUTE_MANIFEST_TOL))
+		if (!dfu_target_manifest_wait (target, error))
+			return FALSE;
 
 	/* success */
 	return TRUE;
--- a/plugins/dfu/dfu.quirk
+++ b/plugins/dfu/dfu.quirk
@@ -329,3 +329,43 @@ Flags = absent-sector-size
 Plugin = dfu
 DfuForceVersion = 011a
 DfuForceTimeout = 5000
+
+# Poly Studio
+[DeviceInstanceId=USB\VID_095D&PID_9217]
+Plugin = dfu
+Flags = manifest-poll
+[DeviceInstanceId=USB\VID_095D&PID_9218]
+Plugin = dfu
+Flags = manifest-poll
+
+# Poly Eagle Eye Cube
+[DeviceInstanceId=USB\VID_095D&PID_9212]
+Plugin = dfu
+Flags = manifest-poll
+[DeviceInstanceId=USB\VID_095D&PID_9213]
+Plugin = dfu
+Flags = manifest-poll
+
+# Poly P30
+[DeviceInstanceId=USB\VID_095D&PID_9290]
+Plugin = dfu
+Flags = manifest-poll
+[DeviceInstanceId=USB\VID_095D&PID_9291]
+Plugin = dfu
+Flags = manifest-poll
+
+# Poly ULCC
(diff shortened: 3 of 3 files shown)


plugins/dfu/dfu.quirk
```diff
@@ -329,3 +329,43 @@ Flags = absent-sector-size
...
+Plugin = dfu
+Flags = poll-in-manifest
+
+# Poly Eagle Eye Mini
+[DeviceInstanceId=USB\VID_095D&PID_3001]
+Plugin = dfu
+Flags = poll-in-manifest
+[DeviceInstanceId=USB\VID_095D&PID_3002]
+Plugin = dfu
+Flags = poll-in-manifest
```
superm1, line 371: missing a new line at end of file

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +472,42 @@ dfu_target_status_to_error_msg (DfuStatus status)
...
+
+	/* not in an error state */
+	if (dfu_device_get_state (priv->device) != DFU_STATE_DFU_ERROR)
+		return TRUE;
+
+	status = dfu_device_get_status (priv->device);
+	/* use a proper error description */
+	g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
```
superm1, line 505: is this really the right description for the error?

plugins/dfu/dfu.quirk
zhj8407: Okay. Got it. I will add the new line.

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +472,42 @@ dfu_target_status_to_error_msg (DfuStatus status)
...
+
+	/* get the status */
+	if (!dfu_device_refresh (priv->device, error))
+		return FALSE;
+
+	/* wait for DFU_STATE_DFU_MANIFEST to not be set */
+	while (dfu_device_get_state (priv->device)
+					== DFU_STATE_DFU_MANIFEST_SYNC ||
+			dfu_device_get_state (priv->device)
+					== DFU_STATE_DFU_MANIFEST) {
```
superm1, line 489: so if the device is permanently stuck in this state the tool what happens to tool?  It seems it gets hung.  I don't know how common that is, but I wonder if we want an upper bound to prevent it?  It seems to not be in the spec.

plugins/dfu/dfu-target.c
zhj8407: Checked the error code of FWUPD, maybe the "FWUPD_ERROR_INTERNAL" is more proper? The real error message returned by device is dfu_target_status_to_error_msg (status).

plugins/dfu/dfu-target.c
zhj8407: Yeah, you are right. If something is wrong with the reprogramming, the device may be stuck in this state. We can limit the max tries here, but not sure what value should be assigned. It is device-specific.

plugins/dfu/dfu-target.c
```diff
@@ -1285,6 +1321,15 @@ dfu_target_download (DfuTarget *target, DfuImage *image,
 			return FALSE;
 	}
 
+	if (fu_device_has_custom_flag (FU_DEVICE (dfu_target_get_device (target)),
+				       "poll-in-manifest") &&
+		dfu_device_has_attribute (priv->device,
+						DFU_DEVICE_ATTRIBUTE_MANIFEST_TOL)) {
+		ret = dfu_target_poll_in_manifest (target, error);
+		if (!ret)
```
hughsie, line 1329: I think lose the `ret` temp var and just put on one line

plugins/dfu/dfu-device.c
```diff
@@ -44,6 +44,7 @@
  * * `legacy-protocol`:		Use a legacy protocol version
  * * `detach-for-attach`:	Requires a DFU_REQUEST_DETACH to attach
  * * `absent-sector-size`:	In absence of sector size, assume byte
+ * * `poll-in-manifest`:	Requires polling via GetStatus in dfuManifest state
```
hughsie, line 47: would `manifest-poll` be better here?

plugins/dfu/dfu-target.c
zhj8407: Yeah. I will change it.

plugins/dfu/dfu-device.c
zhj8407: I have changed the quirk flag. Thanks.

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +474,52 @@ dfu_target_status_to_error_msg (DfuStatus status)
 	return NULL;
 }
 
+static gboolean
+dfu_target_poll_in_manifest (DfuTarget *target, GError **error)
```
hughsie, line 478: I think you want to call it something like `dfu_target_manifest_poll` or  `dfu_target_manifest_wait`

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +474,52 @@ dfu_target_status_to_error_msg (DfuStatus status)
...
+	DfuStatus status;
+	guint polling_count = 0;
+
+	/* get the status */
+	if (!dfu_device_refresh (priv->device, error))
+		return FALSE;
+
+	/* wait for DFU_STATE_DFU_MANIFEST to not be set */
+	while (dfu_device_get_state (priv->device)
+					== DFU_STATE_DFU_MANIFEST_SYNC ||
```
hughsie, line 490: just make a long line, 80 chars is a guideline, not a limit :)

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +474,52 @@ dfu_target_status_to_error_msg (DfuStatus status)
...
+			dfu_device_get_state (priv->device)
+					== DFU_STATE_DFU_MANIFEST) {
+		g_debug ("waiting for DFU_STATE_DFU_MANIFEST to clear");
+
+		if (polling_count++ > DFU_TARGET_MANIFEST_MAX_POLLING_TRIES) {
+			g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "reach to max polling tries");
```
hughsie, line 499: indenting is weird, 8 space tabs please

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +474,52 @@ dfu_target_status_to_error_msg (DfuStatus status)
...
+			return FALSE;
+		}
+
+		g_usleep ((dfu_device_get_download_timeout (
+					priv->device) + 1000) * 1000);
```
hughsie, line 504: indenting again

hughsie: plus, just use a long line pls

plugins/dfu/dfu-target.c
```diff
@@ -472,6 +474,52 @@ dfu_target_status_to_error_msg (DfuStatus status)
...
+		return TRUE;
+
+	status = dfu_device_get_status (priv->device);
+	/* use a proper error description */
+	g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     dfu_target_status_to_error_msg (status));
+
+	return FALSE;
```
hughsie, line 520: it might be nicer to flip this around, so that `dfu_device_get_state (priv->device) == DFU_STATE_DFU_ERROR` is the `if` block which returns FALSE and then we return TRUE at the end. Then status can move up into that scope too.

plugins/dfu/dfu-target.c
```diff
@@ -1285,6 +1333,13 @@ dfu_target_download (DfuTarget *target, DfuImage *image,
 			return FALSE;
 	}
 
+	if (fu_device_has_custom_flag (FU_DEVICE (dfu_target_get_device (target)),
+				       "manifest-poll") &&
+		dfu_device_has_attribute (priv->device,
+						DFU_DEVICE_ATTRIBUTE_MANIFEST_TOL))
```
hughsie, line 1339: indenting and weird line breaking

plugins/dfu/dfu-target.c
zhj8407: Okay. I will change it to dfu_target_manifest_wait.

plugins/dfu/dfu-target.c
zhj8407: I will make the change. I thought the 80-chars was mandatory.

plugins/dfu/dfu-target.c
zhj8407: I will check the indenting.

plugins/dfu/dfu-target.c
zhj8407: I will remove the indenting.

plugins/dfu/dfu-target.c
zhj8407: I was just following the same coding style in dfu_target_check_status(). But I will change it as you suggested.

plugins/dfu/dfu-target.c
zhj8407: I will make the change

hughsie:
I've made a tiny fixup and pushed this to master. Thanks!

zhj8407:
Thank you. Sorry for the bad code format. I should have set my VS Code properly.

PR #2354, opened 2020-09-03
Wip/hughsie/nvme

hughsie:


superm1 (approved):

hughsie:
> meaning that all the OEMs need to have more things in the allowed to upload Vendors in LVFS

Nah, this only changes the *vendor*, not the *vendor_id*.

superm1:
> Nah, this only changes the vendor, not the vendor_id.

Oh phew, OK :)

PR #2355, opened 2020-09-04
dfu: Allow device to attach to normal without a bus reset

zhj8407:
As described in DFU protocol (7. Manifestation Phase), after the
firmware reprogramming is done, if bitWillDetach = 1, it doesn't
require the host to issue a USB bus reset, but the device can
generate a detach-attach sequence itself to go back to normal.

Add a quirk flag "no-bus-reset-attach" to skip the bus reset in
dfu_device_attach(), and increase the "RemoveDelay" as well.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/dfu-device.c
+++ b/plugins/dfu/dfu-device.c
@@ -45,6 +45,7 @@
  * * `detach-for-attach`:	Requires a DFU_REQUEST_DETACH to attach
  * * `absent-sector-size`:	In absence of sector size, assume byte
  * * `manifest-poll`:		Requires polling via GetStatus in dfuManifest state
+ * * `no-bus-reset-attach`:	Do not require a bus reset to attach to normal
  *
  * Default value: `none`
  *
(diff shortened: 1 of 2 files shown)


plugins/dfu/dfu-device.c
```diff
@@ -1365,7 +1366,10 @@ dfu_device_attach (FuDevice *device, GError **error)
 		return FALSE;
 
 	/* normal DFU mode just needs a bus reset */
-	if (!dfu_target_attach (target, error))
+	if (fu_device_has_custom_flag (device, "no-bus-reset-attach") &&
+	    dfu_device_has_attribute (self, DFU_DEVICE_ATTRIBUTE_WILL_DETACH))
```
hughsie, line 1370: do we need the quirk? It seems that we should just look at `DFU_DEVICE_ATTRIBUTE_WILL_DETACH` for this behaviour for a DFU compliant device.

zhj8407: I think it would be better to keep this quirk. Maybe some other existing DFU capable devices still rely on this bus reset to attach. Removing this quirk may have impact on them.

PR #2357, opened 2020-09-07
Wip/hughsie/mei txt

hughsie:


superm1 (approved):
I have no idea if those changes are actually correct; but the effective outcome of blocking running checks on TXE seems reasonable considering the limitations of available information.

PR #2358, opened 2020-09-08
Wip/hughsie/elaptp recovery

hughsie:


plugins/elantp/elantp.quirk
```diff
@@ -6,6 +6,12 @@ GType = FuElantpHidDevice
 Plugin = elantp
 GType = FuElantpHidDevice
 
+# recovery device
+[DeviceInstanceId=I2C\NAME_Synopsys-DesignWare-I2C-adapter]
```
superm1, line 10: Hmmm.... How is this not going to probe every single machine with an i2c adapter at address 0x15?

superm1: I2C doesn't have any specification for what is a probe and what is a more destructive action.  So I'm worried specifically that sending `tx_buf` out on the i2c bus to that address might not be safe.  What if you happen to have a retimer on the i2c bus and that retimer happens to have a power cycle command corresponding to what's in `tx_buf` to probe an Elan touchpad?  You might sever your USB4 link whenever fwupd starts up and cause some data loss on a downstream device.

hughsie: > for what is a probe and what is a more destructive action

Valid. We get the "name" from the `cat /sys/class/i2c-adapter/i2c-*/name` e.g `nvkm-0000:01:00.0-bus-0002` but I agree in the `Synopsys-DesignWare-I2C-adapter` case this is super generic and likely to match on more hardware than just the Acer laptop I have here. I think restricting it to a DMI match might be enough in this case, right?

superm1: >Synopsys-DesignWare-I2C-adapter case this is super generic and likely to match on more hardware than just the Acer laptop I have here

I get a `Synposys DesignWare I2C adapter` on my XPS 13.  
How about using `HIDRAW\*` like we did on Wacom?  Does that work properly?

> I think restricting it to a DMI match might be enough in this case, right?

I think lets fall back to that as last resort if there's no better way.

superm1: > How about using HIDRAW\* like we did on Wacom? Does that work properly?

Oh I guess this wouldn't work.  The whole point is that this doesn't export as HID device, it's raw I2C device.

So yes, DMI match is what we need to move this to for anything with an Elan touchpad that can recover this way.

hughsie: >  The whole point is that this doesn't export as HID device, it's raw I2C device

Right. We do get the HID descriptor using command 0x0001 but I don't think that's at all standardized on all i2c devices.

jinglewu: Yes, Can HID descriptor commands be read from the acpi table? > 
-> Device id : ELANXXXX:XX
-> i2c-address: 0x15
-> HID Descriptor Register Address: 0x0001

plugins/elantp/fu-elantp-common.h
```diff
@@ -8,6 +8,7 @@
 
 #include <glib.h>
 
+#define ETP_CMD_GET_HID_DESCRIPTOR	0x0001
```
jinglewu, line 11: This ETP_CMD_GET_HID_DESCRIPTOR can be read from the acpi table or get from the metainfo.xml?
-> Device id : ELANXXXX:XX
-> i2c-address: 0x15
-> HID Descriptor Register Address: 0x0001

plugins/elantp/elantp.quirk
hughsie: I've added some DMI allow-listing. @superm1 can you verify it **does not** enumerate a i2c device on you XPS 13 please.

plugins/elantp/fu-elantp-common.h
hughsie: I've added the DMI check; you can get more hwids for other machines using `sudo fwupdtool hwids`

jinglewu: I mean, not all elan tp's "ETP_CMD_GET_HID_DESCRIPTOR" are 0x0001 . 
And this value is defined in ACPI TABLE.

hughsie: > I mean, not all elan tp's "ETP_CMD_GET_HID_DESCRIPTOR" are 0x0001

More information required! :) Is this the kind of thing we should include in the .quirk file like we do the i2c address? Or can you share more details about which ACPI table you mean, and the format of it?

superm1: I think the point is the ACPI tables should still be populated even if the device is in recovery state, and that information should come from ACPI tables.  It would scale better than duplication in a quirk file.
Can you put this into a recovery state and check what is still exporting from them into `/sys`?

hughsie: > that information should come from ACPI tables

But which table?

superm1: Generally I'd think DSDT should have it.  Perhaps if you acpidump and decompile and put in a gist we can look it over?

plugins/elantp/elantp.quirk
superm1: As far as I can tell it's not doing anything.

```
$ sudo FWUPD_PROBE_VERBOSE=1 fwupdtool get-devices --plugins=elantp --daemon-verbose=FuEngine
18:43:47:0538 FuEngine             client certificate exists and working
18:43:47:0545 FuEngine             remote lvfs-testing not enabled, so skipping
18:43:47:0545 FuEngine             remote fwupd-tests not enabled, so skipping
18:43:47:0545 FuEngine             remote vendor not enabled, so skipping
18:43:47:0545 FuEngine             remote vendor-directory not enabled, so skipping
18:43:47:0545 FuEngine             468 components now in silo
18:43:47:0555 FuEngine             plugin logind is not enabled
18:43:47:0555 FuEngine             plugin ccgx is not enabled
18:43:47:0555 FuEngine             plugin csr is not enabled
18:43:47:0555 FuEngine             plugin vli is not enabled
18:43:47:0555 FuEngine             adding plugin /usr/local/lib/x86_64-linux-gnu/fwupd-plugins-3/libfu_plugin_elantp.so
18:43:47:0556 FuEngine             plugin linux_sleep is not enabled
18:43:47:0556 FuEngine             plugin jabra is not enabled
18:43:47:0556 FuEngine             plugin rts54hid is not enabled
18:43:47:0556 FuEngine             plugin altos is not enabled
18:43:47:0556 FuEngine             plugin bios is not enabled
18:43:47:0556 FuEngine             plugin ata is not enabled
18:43:47:0556 FuEngine             plugin nitrokey is not enabled
18:43:47:0556 FuEngine             plugin uefi is not enabled
18:43:47:0556 FuEngine             plugin steelseries is not enabled
18:43:47:0556 FuEngine             plugin coreboot is not enabled
18:43:47:0556 FuEngine             plugin synaptics_rmi is not enabled
18:43:47:0556 FuEngine             plugin superio is not enabled
18:43:47:0556 FuEngine             plugin thunderbolt is not enabled
18:43:47:0556 FuEngine             plugin synaptics_prometheus is not enabled
18:43:47:0556 FuEngine             plugin tpm is not enabled
18:43:47:0556 FuEngine             plugin cros_ec is not enabled
18:43:47:0556 FuEngine             plugin uefi_dbx is not enabled
18:43:47:0556 FuEngine             plugin dfu is not enabled
18:43:47:0556 FuEngine             plugin linux_swap is not enabled
18:43:47:0556 FuEngine             plugin synaptics_mst is not enabled
18:43:47:0556 FuEngine             plugin fresco_pd is not enabled
18:43:47:0556 FuEngine             plugin wacom_raw is not enabled
18:43:47:0556 FuEngine             plugin ep963x is not enabled
18:43:47:0556 FuEngine             plugin emmc is not enabled
18:43:47:0556 FuEngine             plugin msr is not enabled
18:43:47:0556 FuEngine             plugin fastboot is not enabled
18:43:47:0556 FuEngine             plugin solokey is not enabled
18:43:47:0556 FuEngine             plugin ebitdo is not enabled
18:43:47:0556 FuEngine             plugin rts54hub is not enabled
18:43:47:0556 FuEngine             plugin iommu is not enabled
18:43:47:0556 FuEngine             plugin nvme is not enabled
18:43:47:0556 FuEngine             plugin wacom_usb is not enabled
18:43:47:0556 FuEngine             plugin logitech_hidpp is not enabled
18:43:47:0556 FuEngine             plugin thelio_io is not enabled
18:43:47:0556 FuEngine             plugin pci_bcr is not enabled
18:43:47:0556 FuEngine             plugin linux_tainted is not enabled
18:43:47:0556 FuEngine             plugin redfish is not enabled
18:43:47:0556 FuEngine             plugin dell is not enabled
18:43:47:0556 FuEngine             plugin optionrom is not enabled
18:43:47:0556 FuEngine             plugin colorhug is not enabled
18:43:47:0556 FuEngine             plugin dell_esrt is not enabled
18:43:47:0556 FuEngine             plugin acpi_facp is not enabled
18:43:47:0556 FuEngine             plugin upower is not enabled
18:43:47:0556 FuEngine             plugin tpm_eventlog is not enabled
18:43:47:0556 FuEngine             plugin amt is not enabled
18:43:47:0556 FuEngine             plugin uefi_recovery is not enabled
18:43:47:0556 FuEngine             plugin pci_mei is not enabled
18:43:47:0556 FuEngine             plugin linux_lockdown is not enabled
18:43:47:0556 FuEngine             plugin synaptics_cxaudio is not enabled
18:43:47:0556 FuEngine             plugin acpi_dmar is not enabled
18:43:47:0556 FuEngine             plugin dell_dock is not enabled
18:43:47:0556 FuEngine             plugin cpu is not enabled
18:43:47:0556 FuEngine             Emitting PropertyChanged('Status'='loading')
Loading…                 [-                                      ]18:43:47:0556 FuEngine             using plugins: elantp
18:43:47:0556 FuEngine             USB 0bda:8153 added
18:43:47:0557 FuEngine             failed to find specified plugin dell: no plugin dell found
18:43:47:0558 FuEngine             USB 0bda:0413 added
18:43:47:0558 FuEngine             USB 0bda:0487 added
18:43:47:0559 FuEngine             USB 1d6b:0003 added
18:43:47:0560 FuEngine             USB 1d6b:0002 added
18:43:47:0561 FuEngine             USB 1d6b:0003 added
18:43:47:0562 FuEngine             USB 8087:0029 added
18:43:47:0566 FuEngine             USB 413c:b06e added
18:43:47:0566 FuEngine             failed to find specified plugin dell_dock: no plugin dell_dock found
18:43:47:0566 FuEngine             USB 413c:b06f added
18:43:47:0567 FuEngine             failed to find specified plugin dell_dock: no plugin dell_dock found
18:43:47:0567 FuEngine             USB 0bda:402e added
18:43:47:0570 FuEngine             USB 0bda:5413 added
18:43:47:0572 FuEngine             USB 0bda:5487 added
18:43:47:0573 FuEngine             USB 0c45:6723 added
18:43:47:0576 FuEngine             USB 1050:0407 added
18:43:47:0578 FuEngine             USB 1d6b:0002 added
18:43:47:0580 FuEngine             10 devices with subsystem i2c-adapter
18:43:47:0580 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/drm/card0/card0-DP-1/i2c-8 added
18:43:47:0581 FuEngine             elantp ignoring: no device file
18:43:47:0581 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/drm/card0/card0-DP-2/i2c-9 added
18:43:47:0582 FuEngine             elantp ignoring: no device file
18:43:47:0582 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/drm/card0/card0-eDP-1/i2c-7 added
18:43:47:0582 FuEngine             elantp ignoring: no device file
18:43:47:0583 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/i2c-3 added
18:43:47:0583 FuEngine             elantp ignoring: no device file
18:43:47:0583 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/i2c-4 added
18:43:47:0584 FuEngine             elantp ignoring: no device file
18:43:47:0584 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/i2c-5 added
18:43:47:0585 FuEngine             elantp ignoring: no device file
18:43:47:0585 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:02.0/i2c-6 added
18:43:47:0585 FuEngine             elantp ignoring: no device file
18:43:47:0585 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:15.0/i2c_designware.0/i2c-1 added
18:43:47:0586 FuEngine             elantp ignoring: no device file
18:43:47:0586 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-2 added
18:43:47:0586 FuEngine             elantp ignoring: no device file
18:43:47:0587 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:1f.4/i2c-0 added
18:43:47:0587 FuEngine             elantp ignoring: no device file
18:43:47:0588 FuEngine             5 devices with subsystem hidraw
18:43:47:0588 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.0/0003:1050:0407.000A/hidraw/hidraw3 added
18:43:47:0589 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:14.0/usb1/1-1/1-1:1.1/0003:1050:0407.000B/hidraw/hidraw4 added
18:43:47:0590 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6.3/1-6.3.5/1-6.3.5:1.0/0003:413C:B06F.004F/hidraw/hidraw2 added
18:43:47:0592 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6.5/1-6.5:1.0/0003:413C:B06E.004E/hidraw/hidraw1 added
18:43:47:0593 FuEngine             UDEV /sys/devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-2/i2c-CUST0001:00/0018:06CB:76AF.0001/hidraw/hidraw0 added
18:43:47:0593 FuEngine             failed to find specified plugin synaptics_rmi: no plugin synaptics_rmi found
18:43:47:0594 FuEngine             Emitting PropertyChanged('Status'='idle')
No detected devices
```

plugins/elantp/fu-elantp-common.h
superm1: So on my system with another vendor's I2C touchpad, it shows up like `/sys/bus/acpi/devices/CUST0001:00`.  I can see in that heirarchy that `physical_node` points back over to the I2C controller:
`../../../../../pci0000:00/0000:00:15.1/i2c_designware.1/i2c-2/i2c-CUST0001:00`

Looking at my DSDT I can see that it's I2C address 0x17, but I don't see the kernel actually exporting that bit into userspace.  @bentiss is there a way that the information requested (i2c address and HID descriptor register address) from an i2c HID export from ACPI entry the kernel parsed to userspace that we're not aware of?

jinglewu: cd /sys/bus/acpi/devices/ELANXXXX:XX
cat status
--------------------
sys node also only found i2c address.

jinglewu: https://elixir.bootlin.com/linux/v4.5/source/drivers/hid/i2c-hid/i2c-hid.c
wHIDDescRegister -> read from ACPI table.

static int i2c_hid_of_probe(struct i2c_client *client,
		struct i2c_hid_platform_data *pdata)
{
	struct device *dev = &client->dev;
	u32 val;
	int ret;

	ret = of_property_read_u32(dev->of_node, "hid-descr-addr", &val);
	if (ret) {
		dev_err(&client->dev, "HID register address not provided\n");
		return -ENODEV;
	}
	if (val >> 16) {
		dev_err(&client->dev, "Bad HID register address: 0x%08x\n",
			val);
		return -EINVAL;
	}
	pdata->hid_descriptor_address = val;

	return 0;
}

jinglewu: static int i2c_hid_acpi_pdata(struct i2c_client *client,
		struct i2c_hid_platform_data *pdata)
{
	static u8 i2c_hid_guid[] = {
		0xF7, 0xF6, 0xDF, 0x3C, 0x67, 0x42, 0x55, 0x45,
		0xAD, 0x05, 0xB3, 0x0A, 0x3D, 0x89, 0x38, 0xDE,
	};
	union acpi_object *obj;
	struct acpi_device *adev;
	acpi_handle handle;
	int ret;

	handle = ACPI_HANDLE(&client->dev);
	if (!handle || acpi_bus_get_device(handle, &adev))
		return -ENODEV;

	obj = acpi_evaluate_dsm_typed(handle, i2c_hid_guid, 1, 1, NULL,
				      ACPI_TYPE_INTEGER);
	if (!obj) {
		dev_err(&client->dev, "device _DSM execution failed\n");
		return -ENODEV;
	}

	pdata->hid_descriptor_address = obj->integer.value;
	ACPI_FREE(obj);

	/* GPIOs are optional */
	ret = acpi_dev_add_driver_gpios(adev, i2c_hid_acpi_gpios);
	return ret < 0 && ret != -ENXIO ? ret : 0;
}

superm1: That's great, thanks!  I didn't realize that's what the number in status actually exported.  But so in this recovery situation, does that mean that basically the `physical_node` symlink wouldn't exist?

jinglewu: It seems that it is difficult to read this information? 
For x86 OS is in the ACPI table, ARM OS is in the DTS.

bentiss: It seems that the I2C device address is covered. Regarding the HID descriptor address, I don't think we export it anywhere in the i2c-hid kernel module.

hughsie: > Regarding the HID descriptor address, I don't think we export it anywhere

But _could_ we? It would save fwupd parsing the ACPI blob itself.

bentiss: > But could we? It would save fwupd parsing the ACPI blob itself.

Depends on what you want/need. My first reaction was to say that we should be able to easily export that from i2c-hid, but if this is a recovery firmware, there is a chance this needs to go lower at the ACPI level. And this gets a little bit more complex, because why would we expose a protocol specific on every ACPI devices?

As Jingle mentioned, the information is not on the ACPI table, but in an ACPI function. So for ACPI devices, this is a matter of just running the blob above, and you'll get the information.

For Arm, and the DTS, well, we are kind of screwed.

hughsie: > but if this is a recovery firmware, there is a chance this needs to go lower at the ACPI level

I guess @jinglewu might know more about this.

> this is a matter of just running the blob above

bentiss: > But can we do this from userspace?

right. Damn. Forgot about that. The answer is "I don't know for sure, but I highly doubt it".

>> For Arm, and the DTS, well, we are kind of screwed.
>
> So perhaps putting it in the fwupd .quirk file isn't so crazy afterall?

Well, there is one thing I am doing more and more, is to use eBPF to monitor unmodified kernel functions. I wonder if we could not instrument the kernel, force a i2c-hid discovery and then fetch all of those information :)

hughsie: > is to use eBPF to monitor unmodified kernel functions

The *whoooosh* as all that goes right over my head :)

> I wonder if we could not instrument the kernel

I think the `we` in that is highly optimistic :) For fwupd, especially for device /recovery/ I'd prefer we KISS if we can help it.

superm1: Since we're talking about a recovery scenario, what about caching some stuff around this into the history database?

hughsie: Something like `FWUPD_DEVICE_FLAG_IS_RECOVERY`? I could use that for the USB2 VLI hub updates too.

superm1: I haven't thought through all the details for how it would work yet, but my initial thought is something like if you see an elan touchpad device, create some metadata on that device for the i2c address and the hid descriptor.  Save that in the history database.  Next time elan tries to enumerate if that device isn't showing up, but you see in the history database there was a previous device with that information you use that information to try to probe for recovery scenario.

bentiss: Just to give an idea of what I am talking about:
On fedora, `sudo dnf install bcc`, then run the following snippet as root https://gist.github.com/bentiss/b694253a358c6a2e11da3bc54786a9d2

When `Tracing... Hit Ctrl-C to end.` is shown, use an other terminal to unbind/rebind the wanted i2c-hid device:
```bash
cd /sys/module/i2c_hid/drivers/i2c:i2c_hid
DEVICE=i2c-00304238304C4C44:00 ; echo $DEVICE > unbind ; echo $DEVICE> bind
```

(replace `i2c-00304238304C4C44:00` with the correct I2C device you are interested in)

Note: currently bringing in `bcc` installs llvmlite and do a on the fly compilation. AFAIU, libbpf would be capable of providing kernel agnostic files you can distribute and not need a compiler anymore for the clients. But that's not something I have explored, yet

hughsie: Although super-cool, this isn't really in-scope for fwupd :)

hughsie: > there was a previous device with that information

I think that sounds super-fragile to be honest; I think the simplest thing is just to encode the i2c address in the quirk file. I'm not super keen on making fwupd much more complicated; it's tricky enough to debug as it is.

Issue #2364, opened 2020-09-09
vli: Hyper Hub+PD Update 


memily:
**Describe the bug**
After adding a hyper quirk file we can update the parent VL817 hub; but are unable to update the BB/VL103 PD child. 

**Steps to Reproduce**
Tried fwupdtool install command on corresponding hyper *.cab file. 

**Expected behavior**
Same setup and update protocol as current Lenovo TR Gen2 Hub.

**fwupd version information**
1.4.1, (dnf)

**fwupd device information**
Attached install verbose log, red error twiced on:

"ignoring device eed3f7906b7b04db3244b7233989fd12f141d2fc [vli] existing device c843d6b450bf0a70c9b3e794904887ee165c52da [vli] already exists"

[verbose.txt](https://github.com/fwupd/fwupd/files/5194173/verbose.txt)

hughsie:
> ignoring device eed3f7906b7b04db3244b7233989fd12f141d2fc [vli] existing device c843d6b450bf0a70c9b3e794904887ee165c52da [vli] already exists

So we have a

* USB HUB `USB\VID_2D01&PID_0385` with that gets the device_id of `b0c12d93b90105a3a7d0860fbecfb9fca76e0fc8` which has a VL103 child of `USB\VID_2D01&PID_0380` which has device_id `c843d6b450bf0a70c9b3e794904887ee165c52da`
* USB HUB `USB\VID_2D01&PID_0382` with that gets the device_id of `4295296d98b3ba38c72f6baa33d24f03a1d428f6` which has a VL103 child of `USB\VID_2D01&PID_0380` which has device_id `eed3f7906b7b04db3244b7233989fd12f141d2fc`

Are there really two USB hubs on this one physical device?

hughsie:
Either way I think something like this is a good idea:
[usbhub-pd.txt](https://github.com/fwupd/fwupd/files/5194749/usbhub-pd.txt)

PR #2365, opened 2020-09-09
vli: Do not dedupe USB hub PD devices

hughsie:
This object derives from FuDevice rather than FuVliDevice and so does not
inherit the NO_GUID_MATCHING flag like other devices.

Fixes https://github.com/fwupd/fwupd/issues/2364

### DIFF
--- a/plugins/nvme/fu-nvme-device.c
+++ b/plugins/nvme/fu-nvme-device.c
@@ -390,6 +390,7 @@ fu_nvme_device_init (FuNvmeDevice *self)
 {
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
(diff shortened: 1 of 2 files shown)


plugins/nvme/fu-nvme-device.c
```diff
@@ -390,6 +390,7 @@ fu_nvme_device_init (FuNvmeDevice *self)
 {
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
```
superm1, line 393: Is that right?  I was thinking this might be a failure that logical ID is wrong

superm1: #2367

hughsie: I think FWUPD_DEVICE_FLAG_NO_GUID_MATCHING should have been the default, but alas it's too risky to change now.

hughsie: >  I was thinking this might be a failure that logical ID is wrong

Leaving NO_GUID_MATCHING turned off means we don't compare the physical or logical IDs at all. It's just a bad default IMHO.

Issue #2366, opened 2020-09-09
Duplicate Hardware Filtered-out Improperly


zx2c4:
My system has 2 of the same NVMe in it. When using `get-devices`, I see only one:

```
├─Samsung SSD 970 EVO Plus 2TB:
│     Device ID:          3743975ad7f64f8d6575a9ae49fb3a8856fe186f
│     Summary:            NVM Express Solid State Drive
│     Current version:    2B2QEXM7
│     Vendor:             Samsung Electronics Co Ltd (NVME:0x144D)
│     Serial Number:      {redacted}
│     GUIDs:              6e54c992-d302-59ab-b454-2d26ddd63e6d ← NVME\VEN_144D&DEV_A808&REV_00
│                         47335265-a509-51f7-841e-1c94911af66b ← NVME\VEN_144D&DEV_A808
│                         66cc31b5-6dc4-55da-9008-bf73c91a0826 ← Samsung SSD 970 EVO Plus 2TB
│     Device Flags:       • Internal device
│                         • Updatable
│                         • Requires AC power
│                         • Needs a reboot after installation
│                         • Device is usable for the duration of the update
```

Looking in the journal, I see this warning:

```
Sep 09 15:50:34 thinkpad fwupd[93842]: 13:50:34:0747 FuDeviceList         ignoring device 03281da317dccd2b18de2bd1cc70a782df40ed7e [nvme] existing device 3743975ad7f64f8d6575a9ae49fb3a8856fe186f [nvme] already exists
```

Ignoring the second NVMe like that seems improper; I really do have two NVMes, which different serial numbers, so it's not filtering some kind of mistake.

System info:
```
zx2c4@thinkpad ~ $ fwupdmgr --version
client version: 1.4.6
compile-time dependency versions
        gusb:   0.3.5
        efivar: 37
daemon version: 1.4.6
zx2c4@thinkpad ~ $ cat /proc/version
Linux version 5.9.0-rc4+ (root@thinkpad) (gcc (Gentoo 10.2.0-r1 p2) 10.2.0, GNU ld (Gentoo 2.34 p6) 2.34.0) #136 SMP Mon Sep 7 11:46:41 CEST 2020
```

zx2c4:
That was incredibly fast. Thank you!

PR #2367, opened 2020-09-09
nvme: use serial number to build device ID (Fixes: #2366)

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I don't get it; surely the physical ID (the PCI bus address) is going to be different?

superm1:
> I don't get it; surely the physical ID (the PCI bus address) is going to be different?

I would think so too, perhaps we need to discuss that more on the bug to find out why they are the same.

superm1:
Lookin at their bug report more closely, I think you're right.

PR #2368, opened 2020-09-09
Wip/hughsie/fwupdtool unsafe

hughsie:


superm1 (approved):

libfwupdplugin/fu-volume.c
```diff
@@ -81,7 +81,7 @@ fu_volume_get_id (FuVolume *self)
 gchar *
 fu_volume_get_mount_point (FuVolume *self)
 {
-	const gchar **mountpoints = NULL;
+	g_autofree const gchar **mountpoints = NULL;
```
superm1, line 84: With this change, does it make sense to perhaps `g_steal_pointer()` on the function return?

hughsie: No, as it's only returning `container` and we're returning one of the array _elements_ that is returned const.

superm1:
For my own learning, how did you manage to narrow this down to the inotify firing?  I couldn't step into a `g_main_loop` run with a vscode frontend for gdb.

hughsie:
> how did you manage to narrow this down to the inotify firing

Good question. Valgrind was really helpful in telling me the memory was being used after free, and I then I worked back to places that were allocating the object. And a good dollop of luck too.

PR #2369, opened 2020-09-09
cpu: Correct the BCR address for Atom Bay Trail

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2328

### DIFF
--- a/plugins/cpu/cpu.quirk
+++ b/plugins/cpu/cpu.quirk
@@ -1,3 +1,3 @@
 # Intel Atom Bay Trail [Silvermont]
-[DeviceInstanceId=CPUID\PRO_0&FAM_06&MOD_07]
-BcrAddr = 0x54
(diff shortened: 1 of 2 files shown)


plugins/cpu/cpu.quirk
```diff
@@ -1,3 +1,3 @@
 # Intel Atom Bay Trail [Silvermont]
-[DeviceInstanceId=CPUID\PRO_0&FAM_06&MOD_07]
-BcrAddr = 0x54
+[DeviceInstanceId=CPUID\PRO_0&FAM_06&MOD_37]
```
superm1, line 2: I would think you may want to just have all of family 06 rather than this specific model.  GUIDs should be made for both

superm1: Oh but wait, fam_06 actually hits big core too.  Never mind.

superm1 (approved):

Issue #2371, opened 2020-09-09
Fail to ensure Id on thunderbolt retimer devices


djcampello:
21:28:31:0034 FuPlugin             performing fu_plugin_device_created() on thunderbolt
21:28:31:0034 FuIdle               inhibiting: thunderbolt requires device wakeup
21:28:31:0034 FuIdle               inhibiting: thunderbolt requires device wakeup
21:28:31:0034 FuIdle               inhibiting: thunderbolt requires device wakeup
21:28:31:0035 FuEngine             failed to add udev device /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/0-0:3.1: cannot ensure ID: FuThunderboltDevice:
Unknown Device
  InstanceIds:          THUNDERBOLT\VEN_8087&DEV_15EE&REV_00
  InstanceIds:          THUNDERBOLT\VEN_8087&DEV_15EE
  Protocol:             com.intel.thunderbolt
  Flags:                require-ac
  VendorId:             THUNDERBOLT:0x8087
  VersionFormat:        pair
  Icon:                 thunderbolt
  Device Type:           USB4 Retimer
  Safe Mode:             false
  Native mode:           false
  Generation:            0
  AuthAttribute:

Looking at the code:

Error will show up if (priv->physical_id == NULL) on fu_device_ensure_id(). This method is called after probe but before setup on fu_device_open. Looking at the thunderbolt plugin fu_thunderbolt_device_probe() fails to call fu_device_set_physical_id() for the retimer case.

PR #2372, opened 2020-09-09
thunderbolt: Call fu_device_set_physical_id() during probe for retimer

djcampello:
Error will show up if (priv->physical_id == NULL) on
fu_device_ensure_id(). This method is called after probe but before
setup on fu_device_open.

Call fu_device_set_physical_id() on fu_thunderbolt_device_probe()
for the retimer case.

Fixes https://github.com/fwupd/fwupd/issues/2371

Change-Id: I0e462fff5e8abf6073318f6424b6736afc8259b8

Type of pull request:
- [ X ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Thanks for the fix!

superm1 (approved):
Will merge once CI passes.

Issue #2373, opened 2020-09-09
Thunderbolt plugin fails to read NVM version


djcampello:
During fwuptool get-updates I see this output on verbose invocation. Each attempt takes .2 seconds delaying the command.

```
23:30:19:0882 FuPluginThunderbolt  Attempt 0: Failed to read NVM version
23:30:20:0082 FuPluginThunderbolt  Attempt 1: Failed to read NVM version
23:30:20:0282 FuPluginThunderbolt  Attempt 2: Failed to read NVM version
23:30:20:0483 FuPluginThunderbolt  Attempt 3: Failed to read NVM version
23:30:20:0683 FuPluginThunderbolt  Attempt 4: Failed to read NVM version
23:30:20:0883 FuPluginThunderbolt  Attempt 5: Failed to read NVM version
23:30:21:0083 FuPluginThunderbolt  Attempt 6: Failed to read NVM version
23:30:21:0283 FuPluginThunderbolt  Attempt 7: Failed to read NVM version
23:30:21:0483 FuPluginThunderbolt  Attempt 8: Failed to read NVM version
23:30:21:0683 FuPluginThunderbolt  Attempt 9: Failed to read NVM version
23:30:21:0884 FuPluginThunderbolt  Attempt 10: Failed to read NVM version
23:30:22:0084 FuPluginThunderbolt  Attempt 11: Failed to read NVM version
23:30:22:0284 FuPluginThunderbolt  Attempt 12: Failed to read NVM version
23:30:22:0484 FuPluginThunderbolt  Attempt 13: Failed to read NVM version
23:30:22:0684 FuPluginThunderbolt  Attempt 14: Failed to read NVM version
23:30:22:0884 FuPluginThunderbolt  Attempt 15: Failed to read NVM version
23:30:23:0084 FuPluginThunderbolt  Attempt 16: Failed to read NVM version
23:30:23:0284 FuPluginThunderbolt  Attempt 17: Failed to read NVM version
23:30:23:0485 FuPluginThunderbolt  Attempt 18: Failed to read NVM version
23:30:23:0685 FuPluginThunderbolt  Attempt 19: Failed to read NVM version
23:30:23:0885 FuPluginThunderbolt  Attempt 20: Failed to read NVM version
23:30:24:0085 FuPluginThunderbolt  Attempt 21: Failed to read NVM version
23:30:24:0285 FuPluginThunderbolt  Attempt 22: Failed to read NVM version
[...]

PR #2375, opened 2020-09-10
thunderbolt: ignore host controller in SW CM mode (Fixes: #2373)

superm1:
nvm_version will not export in SW CM, there is no point in fwupd
displaying a device for it.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -444,7 +444,6 @@ fu_thunderbolt_device_setup (FuDevice *device, GError **error)
 	g_autoptr(GError) error_version = NULL;
 
 	self->devpath = g_strdup (fu_udev_device_get_sysfs_path (FU_UDEV_DEVICE (device)));
-	fu_device_set_metadata (device, "sysfs-path", self->devpath);
 
 	/* try to read the version */
 	if (!fu_thunderbolt_device_get_version (self, &error_version)) {
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -126,13 +126,21 @@ fu_thunderbolt_device_can_update (FuThunderboltDevice *self)
 }
 
 static gboolean
-fu_thunderbolt_device_get_version (FuThunderboltDevice *self)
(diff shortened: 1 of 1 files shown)


superm1:
@djcampello can you see if that helps?

djcampello:
Yes it does! There is no delay anymore attempting the read the version. It does still print this on verbose, not sure if we care to clean it up:

```
02:07:14:0995 FuPlugin             performing fu_plugin_device_created() on thunderbolt
02:07:14:0995 FuIdle               inhibiting: thunderbolt requires device wakeup
02:07:14:0995 FuIdle               inhibiting: thunderbolt requires device wakeup
02:07:14:0995 FuDevice             using 2059395a350c43d978084812d57853ff399e3d73 for a26e8780-9013-cd59-ffff-ffffffffffff
02:07:14:0995 FuPluginThunderbolt  failed to read version
02:07:14:0995 FuPluginThunderbolt  failed to get Vendor ID
02:07:14:0995 FuPluginThunderbolt  failed to get Device ID
```

superm1:
I think those debug messages are all expected since it doesn't bail until a lot later in init (when we can confirm it was a USB4 controller and a host controller).  I'd rather leave them in place in case we need to do any other debugging in these codepaths for the future, but @hughsie and @bleungatchromium any thoughts on those?

hughsie:
> any thoughts on those

they seem pragmatic to me.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -134,6 +134,9 @@ fu_thunderbolt_device_get_version (FuThunderboltDevice *self)
 	/* read directly from file to prevent udev caching */
 	g_autofree gchar *safe_path = g_build_path ("/", self->devpath, "nvm_version", NULL);
 
+	if (!g_file_test (safe_path, G_FILE_TEST_EXISTS))
+		return FALSE;
```
hughsie, line 138: I think if there are multiple points of failure then we should perhaps add a `GError` to `fu_thunderbolt_device_get_version` and print the failure in the caller rather than just the generic `failed to read version` string.

superm1: Sounds like a good idea, I'll rework it.

hughsie (approved):

PR #2379, opened 2020-09-10
Allow simulating a composite update without applying firmware

hughsie:
This allows us to run all the composite firmware vfuncs, as well as the actual
FuDevice ->detach() and ->attach() functions so that we can test the ordering
and enumeration of composite devices without actually writing to SPI.

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -2291,6 +2291,16 @@ fu_plugin_runner_update (FuPlugin *self,
 		return TRUE;
 	}
 
+	/* simulation */
+	if (flags & FWUPD_INSTALL_FLAG_SIMULATE) {
+		g_debug ("simulating, skipping write");
+		for (guint i = 0; i < 100; i++) {
+			fu_device_set_progress (device, i);
+			g_usleep (10 * 1000);
+		}
+		return TRUE;
+	}
+
 	/* optional */
 	g_module_symbol (priv->module, "fu_plugin_update", (gpointer *) &update_func);
 	if (update_func == NULL) {
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2115,6 +2115,7 @@ main (int argc, char *argv[])
 	gboolean allow_reinstall = FALSE;
 	gboolean force = FALSE;
 	gboolean ret;
+	gboolean simulate = FALSE;
 	gboolean version = FALSE;
 	gboolean interactive = isatty (fileno (stdout)) != 0;
 	g_auto(GStrv) plugin_glob = NULL;
@@ -2133,6 +2134,9 @@ main (int argc, char *argv[])
 		{ "allow-older", '\0', 0, G_OPTION_ARG_NONE, &allow_older,
 			/* TRANSLATORS: command line option */
 			_("Allow downgrading firmware versions"), NULL },
+		{ "simulate", '\0', 0, G_OPTION_ARG_NONE, &simulate,
+			/* TRANSLATORS: command line option */
+			_("Simulate installation of the new firmware"), NULL },
 		{ "force", '\0', 0, G_OPTION_ARG_NONE, &force,
 			/* TRANSLATORS: command line option */
 			_("Override plugin warning"), NULL },
@@ -2448,6 +2452,10 @@ main (int argc, char *argv[])
 		priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_OLDER;
 	if (force)
 		priv->flags |= FWUPD_INSTALL_FLAG_FORCE;
+	if (simulate) {
+		priv->flags |= FWUPD_INSTALL_FLAG_SIMULATE;
+		priv->flags |= FWUPD_INSTALL_FLAG_NO_HISTORY;
+	}
 
 	/* load engine */
 	priv->engine = fu_engine_new (FU_APP_FLAGS_NO_IDLE_SOURCES);
--- a/data/bash-completion/fwupdtool.in
+++ b/data/bash-completion/fwupdtool.in
@@ -40,6 +40,7 @@ _fwupdtool_opts=(
 	'--allow-reinstall'
 	'--allow-older'
 	'--force'
+	'--simulate'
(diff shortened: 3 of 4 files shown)


src/fu-util.c
```diff
@@ -2815,6 +2819,8 @@ main (int argc, char *argv[])
 		priv->flags |= FWUPD_INSTALL_FLAG_FORCE;
 	if (no_history)
 		priv->flags |= FWUPD_INSTALL_FLAG_NO_HISTORY;
+	if (simulate)
+		priv->flags |= FWUPD_INSTALL_FLAG_SIMULATE;
```
superm1, line 2823: Is there actually value in this going into fwupdmgr?  It feels like a developer feature, it should probably only be in fwupdtool.

src/fu-tool.c
```diff
@@ -2133,6 +2134,9 @@ main (int argc, char *argv[])
 		{ "allow-older", '\0', 0, G_OPTION_ARG_NONE, &allow_older,
 			/* TRANSLATORS: command line option */
 			_("Allow downgrading firmware versions"), NULL },
+		{ "simulate", '\0', 0, G_OPTION_ARG_NONE, &simulate,
+			/* TRANSLATORS: command line option */
+			_("Simulate installation of the new firmware"), NULL },
```
superm1, line 2139: need to update bash completion (also for fwupdmgr if that one stays)

superm1, line 2139: need to update bash completion (also for fwupdmgr if that one stays)

hughsie:
> and if you won't get that just by skipping the write vfunc necessarily

I do concede that in some cases the write() is the thing that triggers the reboot back into attach() but as we've added these recently I've always tried to "emulate" the same behaviour in ->attach() where possible.

src/fu-util.c
hughsie: I guess it depends if the behaviour is expected to be identical between fwupdtool and fwupdmgr -- in the single device case that should be true, but I'm not sure about the composite case.

libfwupdplugin/fu-plugin.c
```diff
@@ -2291,6 +2291,16 @@ fu_plugin_runner_update (FuPlugin *self,
 		return TRUE;
 	}
 
+	/* simulation */
+	if (flags & FWUPD_INSTALL_FLAG_SIMULATE) {
+		g_debug ("simulating, skipping write");
+		for (guint i = 0; i < 100; i++) {
+			fu_device_set_progress (device, i);
+			g_usleep (10 * 1000);
```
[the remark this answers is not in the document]
hughsie: I was trying to make it more representative, i.e. block the mainloop for more than a few ms.

superm1:
But so maybe what you need is something that feeds some fake data - maybe destroy the USB devices and re-enumerate the USB context?

src/fu-util.c
superm1: A while back yes - but the composite case I was expecting to be true as well.  We explicitly moved a lot of code into the daemon to make them behave identically.  If there actually is a case it's not that this will help root out, then by all means lets leave it in both.

hughsie:
> But so maybe what you need is something that feeds some fake data

Maybe we just need to skip the `->detach()` on devices that don't attach in `->attach()`?

src/fu-util.c
hughsie: I also wondered if we should expose the simulation as part of a "test transaction" in gnome-firmware for instance.

libfwupdplugin/fu-plugin.c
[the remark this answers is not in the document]
superm1: Ah ok.

superm1:
I tested this branch with WD19, and noticed that it causes `composite_cleanup` to be called (turning on the dock's passive update flow).  I think that needs to be skipped for the purposes of a simulation.
```
16:57:32:0711 FuPlugin             performing composite_cleanup() on dell_dock
16:57:32:0711 FuPluginDellDock     activating passive flow (2) for WD19TB
```

superm1:
> > But so maybe what you need is something that feeds some fake data
> 
> Maybe we just need to skip the `->detach()` on devices that don't attach in `->attach()`?

Well I think part of the problem is not everything uses attach/detach.  Also when I tested with WD19 the only events that happened were on the Thunderbolt device.  Nothing else uses those vfuncs

hughsie:
> Well I think part of the problem is not everything uses attach/detach.

Right. I think this is useful for debugging, but solely in fwupdtool territory.  I think given it's not 100% reliable it's not suitable for use in gnome-firmware anyway. I'll respin.

src/fu-tool.c
hughsie: Fixed.

superm1:
I think you still want to explicitly skip `fu_plugin_composite_cleanup`, or at least pass that down into dell-dock's plugin to not run that code path on simulate

hughsie:
I guess `fu_plugin_composite_cleanup()` and `_prepare()` need the `FwupdInstallFlags` added to them; but that's not insane, right?

superm1:
> I guess fu_plugin_composite_cleanup() and _prepare() need the FwupdInstallFlags added to them; but that's not insane, right?

The only problem I see, is that breaks plugin ABI..

hughsie:
> The only problem I see, is that breaks plugin ABI..

Hmm, indeed. Do we care? I think possibly we should; but we can do the same trick as an external library would have have more symbols, e.g. `fu_plugin_composite_cleanup_full()`. It adds complexity, and I'm not sure we want to make things more complicated for a little used simulate feature. Maybe writing to SPI lots of times isn't so bad after-all.

superm1:
I don't see a problem with bumping up ABI to keep it cleaner - there are realistically no other external consumers right now, and 1.5.0 is the release that is perpetually put off, I think accepting the ABI change should be fine.

hughsie:
> Maybe writing to SPI lots of times isn't so bad after-all

I think leave the branch, and close the PR. Either of us can re-open this if we need this kind of functionality in the future.

Issue #2377, opened 2020-09-10
Composite out-of-order updates fail when 


hughsie:
Imagine this scenario:

    Thunderbolt dock
      - TBT controller
        - USB 3.1 HUB
          - Audio device

We then try to install firmware on the TBT controller and Audio device, one after another in a **composite transaction** with multiple firmware types in one .cab file. In all, three devices get removed, some multiple times. Lets say the remove delay on the Audio device is 10 seconds. The TBT takes 30 seconds to detach, write, attach the hub:

 * prepare TBT firmware
 * detach TBT controller
 * write firmware 100%
 * reboot controller (takes ~6 seconds) which causes the USB hub and audio device to "fall off the bus" and the 10s disconnection timeout starts for Audio and HUB
 * wait for the tbt controller to come back
 * move onto writing the audio device
 * fail, as it's not yet been reenumerated
 * a few seconds later, the audio device enumerates -- TOO LATE

The same thing happens if you generate a composite update for the HUB and Audio device; after installing to the hub, we wait for the hub to come back then try to install onto the audio device which is still yet to enumerate.

I'm not sure the best way to fix this with minimum extra code and magic.  I think the only thing we can do is enforce the remove delay of the children to be always bigger that the write time of the parent, e.g.:

 * Make the *effective* remove_delay for a device is a sum of the priv->remove_delay and the MAX(estimated write time of all the parents), although sometimes we don't know the detach+write+attach time, so to speak
 * Set the remove_delay to infinite for all the children when the parent is being updated -- which has the effect that they're never going to be removed if the parent doesn't come back
 * Just set the RemoveDelay of the Audio and HUB devices to be 1.5*(detach+write+attach time of TBT controller) in a quirk for the dock

What we also need to do (which I think is the bug we're seeing) is wait for the child to re-attach if it's in the removed state. Either:

 * Enforce that we do the grandchild -> child -> parent device to workaround the bug -- which is already done using `fu_main_install_task_sort_cb()` and the device order. Maybe this is just a case where we don't set the device order correctly? We do the order promotion dance (using FWUPD_DEVICE_FLAG_INSTALL_PARENT_FIRST) when using `fu_device_add_child()` but seemingly not when using `fu_device_set_parent()` from something like `ParentGuid`

* Manually wait for the removed device as the very first thing in `fu_engine_install()`

Other ideas (and sanity check) most welcome. Thanks!

superm1:
[...]

PR #2380, opened 2020-09-10
synaptics-cxaudio: Fix the topology of the audio device on the TR dock

hughsie:
This ensures we perform the updates in this order:

* cxaudio
* vli
* thunderbolt

As any other order causes enumeration failures.

Fixes some of https://github.com/fwupd/fwupd/issues/2377


superm1 (approved):
makes sense

Issue #2374, opened 2020-09-10
thunderbolt update for W19TB does not complete


djcampello:
```
localhost /sys/bus/thunderbolt/devices/0-3 # fwupdtool update     
Decompressing?           [***************************************]
Decompressing?           [***************************************]
Loading?                 [***************************************]
Thunderbolt controller in Dell dock must remain connected for the duration of the update to avoid damage. Continue with update? [Y|n]: 
Decompressing?           [***************************************]
Installing on Package level of Dell dock?                        ]
Restarting device?       [***************************************]
Installing on VMM5331 in Dell dock?                              ]
Restarting device?       [***************************************] Less than one minute remaining?
Installing on Thunderbolt controller in Dell dock?               ]
Writing?                 [***************************************]
Installing on WD19TB?    [ -                                     ]
Writing?                 [***************************************]
Updates will complete after the dock is unplugged.  The dock will be unavailable for approximately 1 minute.
? Package level of Dell dock has the latest available firmware version
? RTS5413 in Dell dock has the latest available firmware version
? RTS5487 in Dell dock has the latest available firmware version
? VMM5331 in Dell dock has the latest available firmware version
? WD19TB has the latest available firmware version
localhost /sys/bus/thunderbolt/devices/0-3 # fwupdtool get-updates
Decompressing?           [***************************************]
Decompressing?           [***************************************]
Loading?                 [***************************************]
??Thunderbolt controller in Dell dock:
  ?   Device ID:           201af6dc56a22742de828b6bab212f7d3e29254c
  ?   Summary:             Thunderbolt controller
  ?   Current version:     40.00
  ?   Vendor:              Dell Inc. (TBT:0x00D4)
  ?   Install Duration:    22 seconds
  ?   GUIDs:               48711a14-0aaa-55b8-a17f-f7c494b318c6 ? THUNDERBOLT\VEN_00D4&DEV_B070&REV_00
  ?                        08a8c886-2818-544c-b1a3-588eb07ae487 ? THUNDERBOLT\VEN_00D4&DEV_B070
  ?                        c94770ca-1773-592c-b20a-e87243bc7cd0 ? TBT-00d4b070
  ?   Device Flags:        ? Updatable
  ?                        ? Requires AC power
  ?                        ? Supported on remote server
  ?                        ? Device update needs activation
  ?                        ? Device stages updates
  ?                        ? Device is usable for the duration of the update
  ? 
  ??Thunderbolt controller Update:
        New version:       43.00
[...]

PR #2383, opened 2020-09-10
thunderbolt: make sure that authorized has been set before flashing f…

superm1:
…irmware

If the device is not authorized, it may cause a composite update that it's part
of to not behave properly.

See #2374 for more details

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -1754,18 +1754,6 @@ fu_plugin_usb_device_added (FuPlugin *self, FuUsbDevice *device, GError **error)
 }
 
 static gboolean
-fu_plugin_udev_device_changed (FuPlugin *self, FuUdevDevice *device, GError **error)
-{
-	g_autoptr(FuDeviceLocker) locker = NULL;
-
-	/* open */
-	locker = fu_device_locker_new (FU_DEVICE (device), error);
-	if (locker == NULL)
-		return FALSE;
-	return fu_device_rescan (FU_DEVICE (device), error);
-}
-
-static gboolean
 fu_plugin_udev_device_added (FuPlugin *self, FuUdevDevice *device, GError **error)
 {
 	FuPluginPrivate *priv = GET_PRIVATE (self);
@@ -1944,14 +1932,8 @@ fu_plugin_runner_udev_device_changed (FuPlugin *self, FuUdevDevice *device, GErr
 
 	/* optional */
 	g_module_symbol (priv->module, "fu_plugin_udev_device_changed", (gpointer *) &func);
-	if (func == NULL) {
-		if (priv->device_gtype != G_TYPE_INVALID ||
-		    fu_device_get_specialized_gtype (FU_DEVICE (device)) != G_TYPE_INVALID) {
-			if (!fu_plugin_udev_device_changed (self, device, error))
-				return FALSE;
-		}
+	if (func == NULL)
 		return TRUE;
-	}
 	g_debug ("performing udev_device_changed() on %s", priv->name);
 	if (!func (self, device, &error_local)) {
 		if (error_local == NULL) {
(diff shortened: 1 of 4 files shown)


superm1:
@djcampello ^ FYI

djcampello:
Update still shows up. Not sure if this is right. If the device is not updatable until authorized I think that it should not be advertised.

```
localhost ~ # fwupdtool get-updates
??Thunderbolt controller in Dell dock:
  ?   Device ID:           201af6dc56a22742de828b6bab212f7d3e29254c
  ?   Summary:             Thunderbolt controller
  ?   Current version:     40.00
  ?   Vendor:              Dell Inc. (TBT:0x00D4)
  ?   Install Duration:    22 seconds
  ?   GUIDs:               48711a14-0aaa-55b8-a17f-f7c494b318c6 ? THUNDERBOLT\VEN_00D4&DEV_B070&REV_00
  ?                        08a8c886-2818-544c-b1a3-588eb07ae487 ? THUNDERBOLT\VEN_00D4&DEV_B070
  ?                        c94770ca-1773-592c-b20a-e87243bc7cd0 ? TBT-00d4b070
  ?   Device Flags:        ? Updatable
  ?                        ? Requires AC power
  ?                        ? Supported on remote server
  ?                        ? Device update needs activation
  ?                        ? Device stages updates
  ?                        ? Device is usable for the duration of the update
  ? 
  ??Thunderbolt controller Update:
        New version:       43.00
        Remote ID:         vendor-directory
        Summary:           Update for the Thunderbolt controller within a Dell dock
        License:           Proprietary
        Created:           2019-10-23
        Urgency:           High
        Vendor:            Dell Inc.
        Duration:          22 seconds
        Flags:             is-upgrade
      
localhost ~ # fwupdtool update     
Thunderbolt controller in Dell dock must remain connected for the duration of the update to avoid damage. Continue with update? [Y|n]: 
Thunderbolt controller in Dell dock not authorized (status 0)
```

superm1:
Hmmm, the problem is that there is a possible race condition of someone sets authorized after you've checked it during device enumeration unless you check when installing the update.

Does the kernel emit a change event when authorized goes to 1?  If so, that might be another way to handle it.

superm1:
Pushed another way to do it that I think is more in line with what you're expecting @djcampello.
@hughsie the revert is because I noticed that rescan wasn't behaving properly.  I don't think the way that we were doing it is correct.  See stuff like this on the change event:

```
02:10:53:0712 FuUdevDevice         FuUdevDevice emit changed
02:10:53:0712 FuEngine             adding rate-limited timeout for /sys/devices/pci0000:00/0000:00:1d.0/0000:03:00.0/0000:04:00.0/0000:05:00.0/domain0/0-0/0-3
02:10:54:0213 FuEngine             ccgx failed to change udev device /sys/devices/pci0000:00/0000:00:1d.0/0000:03:00.0/0000:04:00.0/0000:05:00.0/domain0/0-0/0-3: cannot ensure ID: FuUdevDevice:
Unknown Device
  InstanceIds:          THUNDERBOLT\VEN_00D4&DEV_B070&REV_00
  InstanceIds:          THUNDERBOLT\VEN_00D4&DEV_B070
  Flags:                none
  VendorId:             THUNDERBOLT:0x00D4
  SysfsPath:            /sys/devices/pci0000:00/0000:00:1d.0/0000:03:00.0/0000:04:00.0/0000:05:00.0/domain0/0-0/0-3

```

hughsie:
> ccgx failed to change udev device

Yes, I saw that too, thanks for tracking it down. If it works for @djcampello then LGTM. Thanks.

djcampello:
I think this is much better... Now get-updates does not show a pending update... Verbose output shows this:
  UpdateError:          Not authorized
which is useful to debug.
Thank you!

superm1:
Awesome thanks.

You'll also get that in `# fwupdtool get-devices` output.

PR #2385, opened 2020-09-11
dell-dock: mirror updatable flag into thunderbolt

superm1:
This makes sure that if an update is pending (as stored in the EC), Thunderbolt
won't show in a needing update state again.

Fixes: #2374

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
^ @djcampello

hughsie (approved):

PR #2386, opened 2020-09-11
Sort out TBT activation for WD19TB on non-disconnect

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -1330,20 +1330,57 @@ fu_util_attach (FuUtilPrivate *priv, gchar **values, GError **error)
 }
 
 static gboolean
-fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
+fu_util_check_activation_needed (FuUtilPrivate *priv, GError **error)
 {
 	gboolean has_pending = FALSE;
 	g_autoptr(FuHistory) history = fu_history_new ();
+	g_autoptr(GPtrArray) devices = fu_history_get_devices (history, error);
+	if (devices == NULL)
+		return FALSE;
+
+	/* only start up the plugins needed */
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *dev = g_ptr_array_index (devices, i);
+		if (fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
+			fu_engine_add_plugin_filter (priv->engine,
+						     fu_device_get_plugin (dev));
+			has_pending = TRUE;
+		}
+	}
+
+	if (!has_pending) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOTHING_TO_DO,
+				     "No devices to activate");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean
+fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
+{
+	gboolean has_pending = FALSE;
 	g_autoptr(GPtrArray) devices = NULL;
 
 	/* check the history database before starting the daemon */
+	if (!fu_util_check_activation_needed (priv, error))
+		return FALSE;
+
+	/* load engine */
+	if (!fu_util_start_engine (priv, FU_ENGINE_LOAD_FLAG_READONLY_FS, error))
+		return FALSE;
+
+	/* parse arguments */
 	if (g_strv_length (values) == 0) {
-		devices = fu_history_get_devices (history, error);
+		devices = fu_engine_get_devices (priv->engine, error);
 		if (devices == NULL)
 			return FALSE;
 	} else if (g_strv_length (values) == 1) {
 		FuDevice *device;
-		device = fu_history_get_device_by_id (history, values[0], error);
+		device = fu_engine_get_device_by_id (priv->engine, values[0], error);
 		if (device == NULL)
 			return FALSE;
 		devices = g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);
@@ -1356,36 +1393,28 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
 		return FALSE;
 	}
 
-	/* nothing to do */
-	for (guint i = 0; i < devices->len; i++) {
-		FuDevice *dev = g_ptr_array_index (devices, i);
-		if (fu_device_has_flag (dev, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
-			fu_engine_add_plugin_filter (priv->engine,
-						     fu_device_get_plugin (dev));
-			has_pending = TRUE;
-		}
-	}
-
-	if (!has_pending) {
-    		g_printerr ("No firmware to activate\n");
-    		return TRUE;
-	}
-
-	/* load engine */
-	if (!fu_util_start_engine (priv, FU_ENGINE_LOAD_FLAG_READONLY_FS, error))
-		return FALSE;
-
 	/* activate anything with _NEEDS_ACTIVATION */
+	/* order by device priority */
+	g_ptr_array_sort (devices, fu_util_device_order_sort_cb);
 	for (guint i = 0; i < devices->len; i++) {
 		FuDevice *device = g_ptr_array_index (devices, i);
 		if (!fu_util_filter_device (priv, FWUPD_DEVICE (device)))
 			continue;
 		if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION))
 			continue;
+		has_pending = TRUE;
 		/* TRANSLATORS: shown when shutting down to switch to the new version */
 		g_print ("%s %s…\n", _("Activating firmware update"), fu_device_get_name (device));
 		if (!fu_engine_activate (priv->engine, fu_device_get_id (device), error))
 			return FALSE;
+	}
+
+	if (!has_pending) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOTHING_TO_DO,
+				     "No devices to activate");
+		return FALSE;
 	}
 
 	return TRUE;
--- a/src/fu-util-common.h
+++ b/src/fu-util-common.h
@@ -94,3 +94,5 @@ gboolean	 fu_util_send_report		(FwupdClient	*client,
 						 GError		**error);
 gint		 fu_util_sort_devices_by_flags_cb	(gconstpointer	 a,
 						 gconstpointer	 b);
+gint		 fu_util_device_order_sort_cb	(gconstpointer a,
+						 gconstpointer b);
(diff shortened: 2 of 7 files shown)


superm1:
^ @djcampello

src/fu-tool.c
```diff
@@ -1375,13 +1376,32 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	for (guint i = 0; i < history_devices->len; i++) {
+		FuDevice *device = g_ptr_array_index (history_devices, i);
 		if (!fu_util_filter_device (priv, FWUPD_DEVICE (device)))
 			continue;
 		if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION))
 			continue;
+		/* make sure that it's plugged in */
+		has_pending = FALSE;
+		for (guint j = 0; j < engine_devices->len; j++) {
+			FuDevice *engine_device = g_ptr_array_index (engine_devices, j);
```
hughsie, line 1397: `device_tmp`?

src/fu-tool.c
```diff
@@ -1375,13 +1376,32 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
...
+			if (g_strcmp0 (fu_device_get_id (engine_device), fu_device_get_id (device)) == 0)
+				has_pending = TRUE;
+		}
+		if (!has_pending) {
+			g_debug ("unable to find %s (%s)", fu_device_get_name (device), fu_device_get_id (device));
```
hughsie, line 1402: don't we normally use square brackets for the ID?

src/fu-tool.c
```diff
@@ -1334,20 +1334,21 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
 {
 	gboolean has_pending = FALSE;
 	g_autoptr(FuHistory) history = fu_history_new ();
-	g_autoptr(GPtrArray) devices = NULL;
+	g_autoptr(GPtrArray) history_devices = NULL;
+	g_autoptr(GPtrArray) engine_devices = NULL;
```
hughsie, line 1338: maybe `s/history_devices/devices` and `s/engine_devices/devices_engine`?

djcampello:
This fixed the issue with the thunderbolt controller activation but after a successful activation. The thunderbolt device came back unauthorized:
```
?     Update Error:        Not authorized
```
so I authorized it and then it became pending activation again (as all other devices on the composite)
```
? ?   Update Error:        A pending update will be completed next time the dock is unplugged from your computer
```

Issuing a second activate did not help:

```
localhost ~ # fwupdtool activate   
Activating firmware update Thunderbolt controller in Dell dock?
Device 201af6dc56a22742de828b6bab212f7d3e29254c does not need activation
```

superm1:
>This fixed the issue with the thunderbolt controller activation but after a successful activation. The thunderbolt device came back unauthorized:

This I would expect if you don't have a udev rule in place to automatically authorize.

>so I authorized it and then it became pending activation again (as all other devices on the composite)

Hmm... That was with calling activate on just the Thunderbolt controller, or a full `fwupdtool activate`?

src/fu-tool.c

src/fu-tool.c

src/fu-tool.c

djcampello:
This was with a global fwupdtool activate (no device is)

superm1:
Did you observe that it called activate for the Thunderbolt first? If so, that may explain the outcome. Thunderbolt and whole chain was gone by the time activate was called for EC and stored pending activation in EC remained.

djcampello:
Yes, thunderbolt happened first. Not sure what would be the right approach: if thunderbolt does not activate first then we are back to the same issue as before, right?

superm1:
If TBT activate is called _right after_  EC's activate the timing works out properly for me.  See if this works for you.

djcampello:
I don't think the patch worked... I still get thunderbolt before:

```
localhost ~ # fwupdtool activate   
Activating firmware update Thunderbolt controller in Dell dock?
Activating firmware update WD19TB?
USB error on device 413c:b06e : No such device (it may have been disconnected) [-4]
```

Also on subsequent calls I get

```
localhost ~ # fwupdtool activate
Activating firmware update Thunderbolt controller in Dell dock?
Device 201af6dc56a22742de828b6bab212f7d3e29254c does not need activation
```

It does not try to activate the other devices (this seems like a bug). Issuing an activate directly to the composite solves the issue:

```
localhost ~ # fwupdtool activate -v 570867f4ddda9915445da59bd85cd0e0e507270e
Activating firmware update WD19TB?
```

superm1:
Oh I think both bugs might be because order isn't stored to the history database. I will rework the loop to use devices coming from the engine not history database.

djcampello:
Global activate now works! But order is still wrong:

```
localhost /sys/bus/thunderbolt/devices/0-3 # fwupdtool activate   
Activating firmware update Thunderbolt controller in Dell dock?
Activating firmware update WD19TB?
USB error on device 413c:b06e : Input/Output Error [-1]
```

superm1:
Oh I guess the default policy is actually to install child first - which is what was desirable until we added this specific activation flow.
See if this other commit I pushed helps.

This will probably need a bunch of testing to make sure it hasn't introduced a corner case scenario though.

djcampello:
Install and activate are still in the same order

```
localhost /sys/bus/thunderbolt/devices/0-3 # fwupdtool install /usr/share/fwupd/remotes.d/vendor/firmware/e66d7258d272c78ee46a8b1770474d046a46df8e-WD19FirmwareUpdateLinux_01.00.04.cab --allow-reinstal
l --allow-older
Installing on Package level of Dell dock?                        ]
Restarting device?       [***************************************]
Installing on VMM5331 in Dell dock?                              ]
Restarting device?       [***************************************] Less than one minute remaining?
Installing on RTS5413 in Dell dock?                              ]
Restarting device?       [***************************************] Less than one minute remaining?
Installing on RTS5487 in Dell dock?                    /         ]
Restarting device?       [***************************************]
Installing on Thunderbolt controller in Dell dock?          -    ]
Writing?                 [***************************************]
Installing on WD19TB?    [ -                                     ]
Writing?                 [***************************************]
Updates will complete after the dock is unplugged. The dock will be unavailable for approximately 1 minute.
localhost /sys/bus/thunderbolt/devices/0-3 # fwupdtool activate
Activating firmware update Thunderbolt controller in Dell dock?
Activating firmware update WD19TB?
USB error on device 413c:b06e : Input/Output Error [-1]
```

superm1:
Hmm @hughsie any thoughts? I figured that should have forced the order with the sort.

djcampello:
I think that flag is only relevant when calling fu_device_add_child(). I run gdb with a break-point there and it is only called twice and it seems to be called based on this entry on the quirks file: https://github.com/fwupd/fwupd/blob/0f6d754d5a57c9ad9c74166841c8f615007c60a2/plugins/dell-dock/dell-dock.quirk#L66

That entry does not list the TBT controller as a child.

superm1:
Thanks for digging into that.  I guess since nothing is actually using the install-parent-first flag this is the first thing catching it.  I think that it needs to be applied in more cases.  I just pushed another commit that should move where it's applied.  (I haven't gotten a chance to check it on my end though)

libfwupdplugin/fu-device.c
```diff
@@ -657,7 +657,10 @@ fu_device_set_parent (FuDevice *self, FuDevice *parent)
 	if (parent != NULL &&
 	    fu_device_get_order (parent) == fu_device_get_order (self)) {
 		g_debug ("auto-setting %s order", fu_device_get_id (parent));
-		fu_device_set_order (parent, fu_device_get_order (self) + 1);
+		if (fu_device_has_flag (self, FWUPD_DEVICE_FLAG_INSTALL_PARENT_FIRST))
+			fu_device_set_order (self, fu_device_get_order (parent) + 1);
+		else
+			fu_device_set_order (parent, fu_device_get_order (self) + 1);
```
hughsie, line 663: The reason I didn't do this originally is that it's not going to work in the following scenario:

 * A is parent to B
 * B is parent to C

So if we have:

 * A added as parent to B (`A=>1, B=0`)
 * B added as parent to C (`B=>1, C=0`)

So, `A=1, B=1, C=0` which isn't what we want (we want `A=2, B=1, C=0`). I think you're on the right track, but I think we need to do a mini-depsolve for the entire composite device.

I can have a go at doing this on Monday if you want.

hughsie: https://github.com/fwupd/fwupd/pull/2387

djcampello: #2387 (plus the 3 first patches here) fixes the issue with install/activation order:

```
localhost ~ # fwupdtool update     
Thunderbolt controller in Dell dock must remain connected for the duration of the update to avoid damage. Continue with update? [Y|n]: 
Installing on Package level of Dell dock?                        ]
Restarting device?       [***************************************]
Installing on VMM5331 in Dell dock?                              ]
Restarting device?       [***************************************] Less than one minute remaining?
Installing on WD19TB?    [    \                                  ]
Writing?                 [***************************************]
Installing on Thunderbolt controller in Dell dock?               ]
Writing?                 [***************************************]
Updates will complete after the dock is unplugged.  The dock will be unavailable for approximately 1 minute.
? Package level of Dell dock has the latest available firmware version
? RTS5413 in Dell dock has the latest available firmware version
? RTS5487 in Dell dock has the latest available firmware version
? VMM5331 in Dell dock has the latest available firmware version
? WD19TB has the latest available firmware version

localhost ~ # fwupdtool activate   
Activating firmware update WD19TB?
Activating firmware update Thunderbolt controller in Dell dock?
```

superm1: Thanks, that's great to hear.  I've rebased this branch on top of that and dropped last commit.

src/fu-util-common.h
```diff
@@ -94,3 +94,5 @@ gboolean	 fu_util_send_report		(FwupdClient	*client,
 						 GError		**error);
 gint		 fu_util_sort_devices_by_flags_cb	(gconstpointer	 a,
 						 gconstpointer	 b);
+gint		 fu_util_activate_task_sort_cb	(gconstpointer a,
```
hughsie, line 97: is there anything activation or task-specific about this? `fu_util_activate_task_sort_cb` should probably be something like `fu_util_device_order_sort_cb` and `fu_util_activate_task_compare` maybe something like `fu_util_device_order_compare`

superm1: no, nothing activation specific.  will rename them both.

src/fu-tool.c
```diff
@@ -1356,39 +1392,30 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
...
 
-	return TRUE;
+	if (!has_pending) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOTHING_TO_DO,
+				     "No devices to activate");
+	}
+
+	return has_pending;
```
hughsie, line 1418: I think `return TRUE` is much easier on your mental model here.

src/fu-tool.c
```diff
@@ -1330,20 +1330,56 @@ fu_util_attach (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	}
+
+	if (!has_pending) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOTHING_TO_DO,
+				     "No devices to activate");
+	}
+
+	return has_pending;
```
hughsie, line 1358: again, return TRUE pls

src/fu-util-common.h
superm1: fixed

src/fu-tool.c

src/fu-tool.c

hughsie (approved):

PR #2387, opened 2020-09-13
Wip/hughsie/device order

hughsie:


libfwupdplugin/fu-device.c
```diff
@@ -2299,7 +2283,7 @@ fu_device_add_string (FuDevice *self, guint idt, GString *str)
 		g_autofree gchar *sz = g_strdup_printf ("%" G_GUINT64_FORMAT, priv->size_max);
 		fu_common_string_append_kv (str, idt + 1, "FirmwareSizeMax", sz);
 	}
-	if (priv->order > 0)
+	if (priv->order != 0)
```
superm1, line 2286: I think `priv->order` is now relevant at all times, given things will be both > and < 0, we should show order every time for this string.

superm1 (approved):
one small comment, otherwise LGTM, thanks so much for this.

libfwupdplugin/fu-device.c
hughsie: Okay, fixed, thanks.

PR #2388, opened 2020-09-14
platform-integrity: Use the BCR values provided by the kernel class

hughsie:
This requires the new kernel driver by Daniel Gutson.


hughsie:
The kernel class will provide BCR data (and more in the future) which allows the HSI calculation to work on more unusual platforms such as Atom much more reliably.

plugins/platform-integrity/fu-plugin-platform-integrity.c
```diff
@@ -10,17 +10,47 @@
...
+{
+	FuPluginData *priv = fu_plugin_get_data (plugin);
+
+	/* interesting device? */
+	if (g_strcmp0 (fu_udev_device_get_subsystem (device), "platform-integrity") != 0)
+		return TRUE;
+
+	/* we only care about the first instance */
+	if (priv->sysfs_path != NULL)
+		return TRUE;
```
superm1, line 43: Aren't these supposed to return FALSE and have an error?

superm1:
I think in commit message you should link to the most recent kernel submission (or if landing for 5.10 maybe link to the maintainer tree it's coming in).

plugins/platform-integrity/fu-plugin-platform-integrity.c
hughsie: I'm not sure. In this case this is a sysfs class that will only realistically have one thing implementing it. Maybe an error is a good idea as then we can see if this assumption is ever voided.

superm1: There are certainly multiple things that can provide a view of platform integrity, but how much is expressed in open source, I don't know.  I think it's better to switch to FALSE and show an error in case that ever happens.

hughsie: > but how much is expressed in open source, I don't know

superm1: `NOT_SUPPORTED` seems fine to me, this isn't useful by default to anyone other than a debugging capacity.

hughsie: Okay, fixed.

superm1 (approved):

PR #2389, opened 2020-09-14
Do not show HSI obsoleted attributes by default

hughsie:
When one result is obsoleted by another, then do not show the old result by
default.

Additionally hide the HSI URLs as this was designed more for GUI clients like
gnome-firmware than CLI tools such as fwupdmgr.

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2160,6 +2160,9 @@ main (int argc, char *argv[])
 		{ "plugins", '\0', 0, G_OPTION_ARG_STRING_ARRAY, &plugin_glob,
 			/* TRANSLATORS: command line option */
 			_("Manually enable specific plugins"), NULL },
+		{ "plugin-whitelist", '\0', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_STRING_ARRAY, &plugin_glob,
+			/* TRANSLATORS: command line option */
+			_("Manually enable specific plugins"), NULL },
 		{ "prepare", '\0', 0, G_OPTION_ARG_NONE, &priv->prepare_blob,
 			/* TRANSLATORS: command line option */
 			_("Run the plugin composite prepare routine when using install-blob"), NULL },
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -61,7 +61,7 @@ struct FuUtilPrivate {
 	gboolean		 cleanup_blob;
 	gboolean		 enable_json_state;
 	FwupdInstallFlags	 flags;
-	gboolean		 show_all_devices;
+	gboolean		 show_all;
 	gboolean		 disable_ssl_strict;
 	/* only valid in update and downgrade */
 	FuUtilOperation		 current_operation;
@@ -411,7 +411,7 @@ fu_util_get_details (FuUtilPrivate *priv, gchar **values, GError **error)
 	}
 
 	/* implied, important for get-details on a device not in your system */
-	priv->show_all_devices = TRUE;
+	priv->show_all = TRUE;
 
 	/* open file */
 	fd = open (values[0], O_RDONLY);
@@ -473,7 +473,7 @@ fu_util_build_device_tree (FuUtilPrivate *priv, GNode *root, GPtrArray *devs, Fu
 		FuDevice *dev_tmp = g_ptr_array_index (devs, i);
 		if (!fu_util_filter_device (priv, FWUPD_DEVICE (dev_tmp)))
 			continue;
-		if (!priv->show_all_devices &&
+		if (!priv->show_all &&
 		    !fu_util_is_interesting_device (FWUPD_DEVICE (dev_tmp)))
 			continue;
 		if (fu_device_get_parent (dev_tmp) == dev) {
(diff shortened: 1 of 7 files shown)


src/fu-tool.c
```diff
@@ -2018,10 +2019,14 @@ fu_util_security (FuUtilPrivate *priv, gchar **values, GError **error)
 	g_print ("%s \033[1m%s\033[0m\n", _("Host Security ID:"),
 		 fu_engine_get_host_security_id (priv->engine));
 
+	/* show or hide different elements */
+	if (priv->show_all_devices)
```
superm1, line 2023: this feels like it's overloading `--show-all-devices`.  Either it should be a new option `--show-all-security`, or I think we should rename to `--show-all`.

hughsie: > or I think we should rename to --show-all

I think you're right. New patch in 5..4...

src/fu-util-common.h
```diff
@@ -24,6 +24,14 @@ typedef struct {
...
 } FuUtilCmd;
 
+typedef enum {
+	FU_SECURITY_ATTR_TO_STRING_FLAG_NONE			= 0,
+	FU_SECURITY_ATTR_TO_STRING_FLAG_SHOW_OBSOLETES		= 1 << 0,
+	FU_SECURITY_ATTR_TO_STRING_FLAG_SHOW_URLS		= 1 << 1,
+	/*< private >*/
+	FU_SECURITY_ATTR_TO_STRING_FLAG_LAST
+} FuSecurityAttrToStringFlags;
+
```
superm1, line 34: If adding all this information specific to the command line client, I'd argue that `--verbose` or something similar should add the URL too

hughsie: Maybe just tie it to `--show-all` as well?

src/fu-util.c
```diff
@@ -2524,7 +2529,10 @@ main (int argc, char *argv[])
 		{ "no-history", '\0', 0, G_OPTION_ARG_NONE, &no_history,
 			/* TRANSLATORS: command line option */
 			_("Do not write to the history database"), NULL },
-		{ "show-all-devices", '\0', 0, G_OPTION_ARG_NONE, &priv->show_all_devices,
+		{ "show-all", '\0', 0, G_OPTION_ARG_NONE, &priv->show_all,
+			/* TRANSLATORS: command line option */
+			_("Show all results"), NULL },
+		{ "show-all-devices", '\0', G_OPTION_FLAG_HIDDEN, G_OPTION_ARG_NONE, &priv->show_all,
```
superm1, line 2535: this line should be dropped with a rename of the option

src/fu-util.c
```diff
@@ -2524,7 +2529,10 @@ main (int argc, char *argv[])
 		{ "no-history", '\0', 0, G_OPTION_ARG_NONE, &no_history,
 			/* TRANSLATORS: command line option */
 			_("Do not write to the history database"), NULL },
-		{ "show-all-devices", '\0', 0, G_OPTION_ARG_NONE, &priv->show_all_devices,
+		{ "show-all", '\0', 0, G_OPTION_ARG_NONE, &priv->show_all,
```
superm1, line 2532: make sure to update the bash completion for both `fwupdtool` and `fwupdmgr`

src/fu-util-common.h
superm1: Sure.

src/fu-util.c
hughsie: Ohh, I don't think there's any way we can drop it; `G_OPTION_FLAG_HIDDEN` means it's not in --help any more, but there are hundreds of places all over the internet that specify running `fwupdmgr --show-all-devices` -- it's essentially part of our CLI "ABI" now.

superm1: Ohhh I see.  In that case, I wonder if we actually should have kept `--plugin-whitelist` to alias to `--plugins` like that too.

src/fu-util.c
superm1: And fish completion I guess uses this too

hughsie: Eeek, done.

src/fu-util.c
hughsie: I think so too; I've added it as a follow up to this branch.

.github/ISSUE_TEMPLATE/bug-report-general.md
```diff
@@ -30,7 +30,7 @@ Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
 Please provide the output of the fwupd devices recognized in your system.
 
 ```shell
-$ fwupdmgr get-devices --show-all-devices
+$ fwupdmgr get-devices --show-all
```
superm1, line 33: ugh, so this is going to be problematic being in the template - most people have 1.4.x or less and won't have `--show-all` to your point.

hughsie: I wasn't going to update the template; did I?

hughsie: I did! not what i wanted!

hughsie: Okay, fixed.

superm1 (approved):

PR #2393, opened 2020-09-15
Wip/hughsie/hsi arm

hughsie:


docs/hsi.xml
```diff
@@ -128,6 +128,13 @@
         connected, additional software to be installed, or disabling any existing
         security layers to measure.
       </para>
+      <para>
+        The HSI specification is primarily designed for consumer laptop, desktop
+        and server hardware, although some tests <emphasis>may</emphasis> still
```
superm1, line 133: I don't necessarily agree on the server hardware aspect.  I haven't seen a single thing that talks about a BMC, or looking for surprise removal events or anything else server-y.

docs/hsi.xml
```diff
@@ -128,6 +128,13 @@
...
+        make sense on embedded hardware such as mobile phones and other consumer
+        appliances.
+        It is not expected that embedded hardware will publish an HSI number.
```
superm1, line 136: I think it's worth taking this a step further.  Have an allowlist of chassis types, and only calculate HSI for those.

docs/hsi.xml
```diff
@@ -161,7 +168,8 @@
       <para>
         This security level corresponds to the most basic of security protections
         considered essential by security professionals.
-        Any failures at this level would have critical security impact.
+        Any failures at this level would have critical security impact and could
+        likely be used to compromise the system firmware without opening the case.
```
superm1, line 172: Rather than focusing on the case, how about "physical access"

docs/hsi.xml
```diff
@@ -259,6 +267,38 @@
...
     </refsect2>
 
+    <refsect3 id="org.fwupd.hsi.Uefi.SecureBoot">
+      <title>UEFI SecureBoot</title>
+      <para>
+        UEFI Secure boot is a verification mechanism for ensuring that code
+        launched by firmware is trusted.
+      </para>
+      <para>
+        Secure use of UEFI Secure Boot requires that each binary loaded at
```
superm1, line 277: I don't think you need to use "Secure use of UEFI Secure Boot".  Just "UEFI secure boot"

docs/hsi.xml
```diff
@@ -259,6 +267,38 @@
...
+        boot is validated against trusted certifictes.
+      </para>
+      <itemizedlist>
+        <listitem>
+          <para>
+            For HSI-1 SecureBoot must be available for use on the platform.
```
superm1, line 283: Although it's probably implied, you should say for HSI-1 on UEFI systems

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -102,11 +102,17 @@ fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
...
 	attr = fwupd_security_attr_new (FWUPD_SECURITY_ATTR_ID_UEFI_SECUREBOOT);
 	fwupd_security_attr_set_plugin (attr, fu_plugin_get_name (plugin));
-	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
 	fu_security_attrs_append (attrs, attr);
 
+	/* SB not available */
+	if (!fu_efivar_exists (FU_EFIVAR_GUID_EFI_GLOBAL, "SecureBoot")) {
+		fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+		return;
+	}
```
superm1, line 111: Use `fu_efivar_secure_boot_enabled` instead

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -102,11 +102,17 @@ fu_plugin_add_security_attrs (FuPlugin *plugin, FuSecurityAttrs *attrs)
 	/* create attr */
 	attr = fwupd_security_attr_new (FWUPD_SECURITY_ATTR_ID_UEFI_SECUREBOOT);
 	fwupd_security_attr_set_plugin (attr, fu_plugin_get_name (plugin));
-	fwupd_security_attr_add_flag (attr, FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE);
 	fu_security_attrs_append (attrs, attr);
 
+	/* SB not available */
+	if (!fu_efivar_exists (FU_EFIVAR_GUID_EFI_GLOBAL, "SecureBoot")) {
```
superm1, line 108: What do you think about instead making `fu_efivar_secure_boot_enabled` to take an argument of `GError` and then on the FALSE scenarios (missing variable, or disabled) explicitly put the reason in the error?

docs/hsi.xml
hughsie: Something like

        The HSI specification is primarily designed for consumer laptop and desktop
        hardware, although some tests <emphasis>may</emphasis> still make sense on
        server or embedded hardware.
        It is not expected that non-consumer hardware will publish an HSI number.

docs/hsi.xml
hughsie: I'm not so sure the DMI data for chassis types is always correct; there's a lot of `To Be Filled By Te O.E.M.` out there.

docs/hsi.xml
hughsie: That's much better, thanks.

docs/hsi.xml
superm1: that's better to me

docs/hsi.xml
hughsie: Fixed.

docs/hsi.xml
hughsie: Do we care about non-UEFI systems? CSM boot should be an HSI failure, surely?

docs/hsi.xml
superm1: The manufacturer field sure, but the chasssis type?  Isn't that a number?
To me that's not a reason to ignore a spec though.  If the chassis type is bad information (isn't in the allow list), HSI can't be computed and show an error.

plugins/uefi/fu-plugin-uefi.c
hughsie: It's a good idea; do we care about libfwupdplugin API now? I can't remember what policy is on breaking that.

docs/hsi.xml
superm1: Well you can ask the linuxboot, coreboot and oreboot guys that and you'll get a different answer than me :)

plugins/uefi/fu-plugin-uefi.c
superm1: ABI yes, API, not so much.  I guess this would break ABI, so maybe it needs to be a new `_full` function then

hughsie: I wonder if we should expand DB things out into it's own class, so that we can easily get the PK, status, etc.

docs/hsi.xml
hughsie: > linuxboot, coreboot and oreboot

These all imply SB=off, right? None of them provide a chain of trust with a PK, right?

superm1: Secure boot in the context described is a UEFI construct.  All these other things would have a _different_ trust chain approach.

plugins/uefi/fu-plugin-uefi.c
superm1: Makes sense to me

docs/hsi.xml
hughsie: > To me that's not a reason to ignore a spec though.

That's a fair enough point, but the devil is in the detail. Could you say which of these are "in" or "out" from an HSI point of view?

    02h Unknown
    03h Desktop
    04h Low Profile Desktop
    05h Pizza Box
    06h Mini Tower
    07h Tower
    08h Portable
    09h LapTop
    0Ah Notebook
    0Bh Hand Held
    0Ch Docking Station
    0Dh All in One
    0Eh Sub Notebook
    0Fh Space-saving
    10h Lunch Box
    11h Main Server Chassis
    12h Expansion Chassis
    13h SubChassis
    14h Bus Expansion Chassis
    15h Peripheral Chassis
    16h RAID Chassis
    17h Rack Mount Chassis
    18h Sealed-case PC
    19h Multi-system chassis
    1Ah Compact PCI
    1Bh Advanced TCA
    1Ch Blade
    1Dh Blade Enclosure

superm1: I would say in is:
3, 4, 6, 7, 8, 9, A, D

docs/hsi.xml
hughsie: Fixed.

plugins/uefi/fu-plugin-uefi.c
hughsie: > needs to be a new _full function

I did this, a whole new object just seemed like a waste for two bits of data.

docs/hsi.xml
```diff
@@ -128,6 +128,12 @@
         connected, additional software to be installed, or disabling any existing
         security layers to measure.
       </para>
+      <para>
+        The HSI specification is primarily designed for consumer laptop and
```
superm1, line 132: Consumer?  I would think it's actually more valuable in business client.

superm1 (approved):

docs/hsi.xml
hughsie: `designed for business and consumer laptop` perhaps?

superm1: Maybe just "laptops"

docs/hsi.xml
hughsie: Filed as https://github.com/fwupd/fwupd/issues/2440

docs/hsi.xml
hughsie: Agree. fixed.

PR #2399, opened 2020-09-17
bios: Only compile when UEFI is available

hughsie:
We're telling people on PPC64 to switch to UEFI mode...

### DIFF
--- a/plugins/bios/fu-plugin-bios.c
+++ b/plugins/bios/fu-plugin-bios.c
@@ -73,15 +73,17 @@ gboolean
 fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 {
 	g_autofree gchar *sysfsfwdir = NULL;
-	g_autoptr(GError) error_local = NULL;
 	g_autofree gchar *esrt_path = NULL;
 
 	/* are the EFI dirs set up so we can update each device */
+#if defined(__x86_64__) || defined(__i386__)
+	g_autoptr(GError) error_local = NULL;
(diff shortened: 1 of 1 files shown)


superm1:
Is it useful to still have the plugin but instead to use something like `HAVE_UEFI` to control whether or not `fu_plugin_bios_create_dummy` uses `NULL` for `reason` perhaps?

hughsie:
> Is it useful to still have the plugin

What would it do? compiling with `-Dplugin_uefi` is basically saying "I don't have UEFI" and thus are unable to get updates using capsule for instance.

superm1:
>What would it do? compiling with -Dplugin_uefi is basically saying "I don't have UEFI" and thus are unable to get updates using capsule for instance.

"Information only"
No different then the `tpm` or `cpu` plugin.  It just shows tpm version or cpu microcode version.

hughsie:
> Information only

Ahh, in this case we don't even have FU_HWIDS_KEY_BIOS_VERSION as we only support SMBIOS rather than DT. I've asked the company driving this forward to provide the mapping between DT attributes and the HWID values we use.

Perhaps the right thing to do here is:

 * Set up the DT mapping for the HWIDs (which we might need in the future for other reasons)
 * Somehow conventionalize the `Firmware can not be updated in legacy BIOS mode, switch to UEFI mode` on being x64 and ix86 specific.

Note, it also doesn't make sense to talk about BIOS mode for ARM too, right?

superm1:
Right

superm1 (approved):

PR #2400, opened 2020-09-18
Wip/hughsie/nvme with optionrom

hughsie:


superm1 (approved):

hughsie:
Thanks for the ever-speedy review. I'll get the reporter to test this, then merge this if if works on their system.

superm1:
Sure. I am surprised by this need, but commit message explained it pretty well.

hughsie:
So, it turns out we need this too, as we've now got a network card with option rom (which is far more likely) and also a new plugin to update it. I'll hold off applying it for a few days, as the option rom verification might just be available in the main firmware image and so we'd just have a device priority greater than the option rom device to obsolete it in the daemon.

superm1:
>network card with option rom

Well that's certainly interesting.  If it's an option ROM does it have DXE code that runs in BIOS?  I'd suspect you'll need signed binaries on it to run if so.

hughsie:
>  If it's an option ROM does it have DXE code that runs in BIOS

I'll dump all the code sections and work it out. I guess I'm looking for an EFI binary, right?

superm1:
> I'll dump all the code sections and work it out. I guess I'm looking for an EFI binary, right?

I'd guess so?  This is getting outside my wheelhouse of knowledge of how code is loaded from an option rom.

hughsie:
> I'd guess so

Yes, it's a valid PE file. For my future self, it's `CPU 0, Type 0` image.

PR #2401, opened 2020-09-18
Wip/hughsie/toc

hughsie:


hughsie:
The other name for this I had was `FuXmlFirmware` but I felt TableOfContents reflected this better. Yell if you have any better ideas, or if you prefer the Xml prefix.

superm1:
So is it essentially a firmware image that happens to use Xml for it's header rather than a data structure? That's pretty unusual, usually headers are a lot smaller. It might be just worth naming it as whatever device it's like until we know something other than this device actually uses it.

hughsie:
> So is it essentially a firmware image that happens to use Xml for it's header rather than a data structure

Not really, maybe my commit message wasn't good. It's a way of a human writing a "_how do I build a FuFirmware_" list of instructions. We could have something like this instead:

    create image
    assign image idx 345
    assign image addr 0x8000
    copy image bytes from firmware.ROM
    add image to firmware

the idea being it's just a set of instructions that let you build a composite image without having to write some C code. Like a firmware _manifest_ if you will. It's entirely built to generate a FuFirmware by me, so it's not like we have a spec anywhere to copy.

hughsie:
Maybe `FuFirmwareManifest` or `FuFirmwareFactory` is the right name for this; it's not really a "type" of FuFirmware, just an object that can create one given an XML manifest.

superm1:
> Maybe `FuFirmwareManifest` or `FuFirmwareFactory` is the right name for this; it's not really a "type" of FuFirmware, just an object that can create one given an XML manifest.

Yeah I think so

hughsie:
I think this is a much better idea: https://github.com/fwupd/fwupd/pull/2408

PR #2405, opened 2020-09-20
Allow binding and unbinding kernel modules

hughsie:


libfwupdplugin/fu-udev-device.c
```diff
@@ -518,6 +521,124 @@ fu_udev_device_get_slot_depth (FuUdevDevice *self, const gchar *subsystem)
...
+			    GError **error)
+{
+#ifndef _WIN32
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	g_autofree gchar *fn = NULL;
+	g_autoptr(GFile) file = NULL;
+	g_autoptr(GOutputStream) stream = NULL;
+
+	/* driver exists */
+	fn = g_strdup_printf ("/sys/module/%s/drivers/%s:%s/bind",
```
superm1, line 606: Is this path correct when complied into the kernel?

hughsie: Yes, that works too. Is that a hint that you think it should be `s/fu_udev_device_bind_module/fu_udev_device_bind_driver`?

superm1: Yes, you're binding a device to a driver not to a module.

hughsie: You're completely right. I've fixed up the naming and the comments.

libfwupdplugin/fu-udev-device.c
```diff
@@ -793,6 +914,7 @@ fu_udev_device_set_physical_id (FuUdevDevice *self, const gchar *subsystems, GEr
 			return FALSE;
 		}
 		physical_id = g_strdup_printf ("PCI_SLOT_NAME=%s", tmp);
+		priv->bus_id = g_strdup (tmp);
```
superm1, line 917: Why does bus-id need to be subsystem specific? Isn't the kernel interface to bind and unbind identical? Shouldn't all devices have this?

hughsie: For the PCI bus you use `PCI_SLOT_NAME`, for USB you use the USB bus ID, e.g. `1-1:1.0`

superm1: But FuUsbDevice doesn't inherit from FuUdevDevice.  For USB the udev device isn't used at all I thought

hughsie: Agree; did you expect to see bus_id set for USB devices too?

superm1: OK I think i see what you're meaning now looking at more context of this code.  Looking on my system at another device, i2c-hid:
```
/sys/module/i2c_hid/drivers/i2c:i2c_hid$ ls
bind  i2c-CUST0001:00  module  uevent  unbind
```
I can see that is actually the ACPI ID:
```
/sys/bus/acpi/devices$ ls -l | grep CUST
lrwxrwxrwx 1 root root 0 Sep 21 09:42 CUST0000:00 -> ../../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:7d/CUST0000:00
lrwxrwxrwx 1 root root 0 Sep 21 09:42 CUST0001:00 -> ../../../devices/LNXSYSTM:00/LNXSYBUS:00/PNP0A08:00/device:7e/CUST0001:00
```

So it makes me ask is there a more generic value that fits in here from resolving symlinks?

If not, I guess if/when it's needed outside of PCI the ID's used for other types can be added too.

superm1: > Agree; did you expect to see bus_id set for USB devices too?

I did for i2c and USB too, yeah.

hughsie: > But FuUsbDevice doesn't inherit from FuUdevDevice. For USB the udev device isn't used at all I thought

Ahh, yes, now I see what you mean. What we probably need is a `fu_device_unbind_driver` on the base class, and then vfuncs implemented for FuUsbDevice and FuUdevDevice.

superm1: Yep 👍

hughsie: I think I've done this; I don't have any i2c-hid devices handy so if you've got any more input for `fu_udev_device_get_bind_id()` that would be great. Unfortunately it's all very subsystem specific as it's all kinda evolved rather than with a nice abstracted sysfs key we can use.

src/fu-tool.c
```diff
@@ -1301,6 +1301,68 @@ fu_util_detach (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	/* get device */
+	if (g_strv_length (values) == 1) {
+		device = fu_util_get_device (priv, values[0], error);
+		if (device == NULL)
+			return FALSE;
+	} else {
+		device = fu_util_prompt_for_device (priv, NULL, error);
+		if (device == NULL)
+			return FALSE;
+	}
```
superm1, line 1323: I think you can simplify this to have the `(device == NULL)` check pop out of the if loop since both cases end with it.

src/fu-tool.c
```diff
@@ -1301,6 +1301,68 @@ fu_util_detach (FuUtilPrivate *priv, gchar **values, GError **error)
...
+static gboolean
+fu_util_bind_driver (FuUtilPrivate *priv, gchar **values, GError **error)
+{
+	g_autoptr(FuDevice) device = NULL;
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	/* load engine */
+	if (!fu_util_start_engine (priv, FU_ENGINE_LOAD_FLAG_NONE, error))
+		return FALSE;
+
```
superm1, line 1341: Since this is user facing I think it's worth adding some magic here to check if the driver provided exists in the kernel, and/or if `-` or `_` are done right.

For example for me I can interact with the kernel module `i2c-hid`, but `/sys/module/i2c_hid` is what you interact with.
So you probably need to check first if `/sys/module/%s` exists and if it doesn't, when you `s/-/_` if it does.

And I can totally see this magic test happening in `fu_device_bind_driver` too, but I think it will make this more user friendly.

libfwupdplugin/fu-udev-device.c
```diff
@@ -518,6 +518,114 @@ fu_udev_device_get_slot_depth (FuUdevDevice *self, const gchar *subsystem)
...
 
+#ifndef _WIN32
+static gchar *
+fu_udev_device_get_bind_id (FuUdevDevice *self)
+{
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	if (g_strcmp0 (fu_udev_device_get_subsystem (self), "pci") == 0)
+		return g_strdup (g_udev_device_get_property (priv->udev_device, "PCI_SLOT_NAME"));
+	if (g_strcmp0 (fu_udev_device_get_subsystem (self), "usb") == 0)
+		return g_path_get_basename (g_udev_device_get_sysfs_path (priv->udev_device));
```
superm1, line 529: for I2C it seems to be the stuff in `PhysicalId`

`HID_PHYS=i2c-CUST0001:00`

hughsie: What's the subsystem tho, `hid`, or `i2c`?

src/fu-tool.c
hughsie: > s/-/_ if it does

I think modprobe always unconditionally does this, maybe we just do the same?

src/fu-tool.c
hughsie: Fixed.

libfwupdplugin/fu-udev-device.c
superm1: This was from the synaptics-rmi device.

```
Touchpad
  DeviceId:             b26933c085b020ecf84c490812458523aee710ac
  Guid:                 f4384034-9243-5334-8075-a534be913e46 <- HIDRAW\VEN_06CB&DEV_76AF&REV_00
  Guid:                 424bd00e-9789-5cdf-a12a-3c81bc4676d6 <- HIDRAW\VEN_06CB&DEV_76AF
  Guid:                 140f4458-951b-5bb9-85e2-879bd5b02615 <- SYNAPTICS_RMI\TM3038-003
  Guid:                 b29d3c85-cd0e-503e-9c7e-f6731c1eaf2d <- SYNAPTICS_RMI\TM3038
  Plugin:               synaptics_rmi
  Protocol:             com.synaptics.rmi
  Flags:                internal|updatable|registered
  Vendor:               Synaptics
  VendorId:             HIDRAW:0x06CB
  Version:              1.5.2767034
  VersionBootloader:    54.0.0
  VersionFormat:        triplet
  Created:              2020-09-21
  PhysicalId:           HID_PHYS=i2c-CUST0001:00
  SysfsPath:            /sys/devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-2/i2c-CUST0001:00/0018:06CB:76AF.0001/hidraw/hidraw0
  BlVer:                0x6
  BootloaderId:         56.36
  BlockSize:            0x10
  BlockCountFw:         0xe60
  BlockCountCfg:        0x40
  FlashConfigLength:    0x0
  PayloadLength:        0x0
  BuildID:              0x2a38ba
```

src/fu-tool.c
superm1: Sure, makes sense to me.

libfwupdplugin/fu-udev-device.c
hughsie: You need to use `FU_UDEV_DEVICE_DEBUG=1` to see the subsystem iirc.

src/fu-tool.c
hughsie: Done.

libfwupdplugin/fu-udev-device.c
superm1: Oh it's hid - you're right.

hughsie: Okay, added, thanks.

src/fu-tool.c
```diff
@@ -1301,6 +1301,66 @@ fu_util_detach (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	/* get device */
+	if (g_strv_length (values) == 1) {
+		device = fu_util_get_device (priv, values[0], error);
+	} else {
```
superm1, line 1317: can lose the curly braces

superm1 (approved):

Issue #2406, opened 2020-09-21
fwupdate crashes on archlinux on Lenovo Ideapad 14are05


anarsoul:
**Describe the bug**
fwupdate crashes on archlinux

**Steps to Reproduce**
```
G_MESSAGES_DEBUG=all fwupdate -a {e20bafd3-9914-4f4f-9537-3129e090eb3c} ./EECN26WW.cap(fwupdate:3591): FuCommon-DEBUG: 23:28:07.340: device /org/freedesktop/UDisks2/block_devices/nvme0n1p2, type: 0fc63daf-8483-4772-8e79-3d69d8477de4
(fwupdate:3591): FuCommon-DEBUG: 23:28:07.344: device /org/freedesktop/UDisks2/block_devices/nvme0n1p3, type: 0fc63daf-8483-4772-8e79-3d69d8477de4
(fwupdate:3591): FuCommon-DEBUG: 23:28:07.352: device /org/freedesktop/UDisks2/block_devices/nvme0n1p1, type: c12a7328-f81f-11d2-ba4b-00a0c93ec93b
(fwupdate:3591): FuCommon-DEBUG: 23:28:07.368: reading ./EECN26WW.cap with 17350240 bytes
 
(fwupdate:3591): FuVolume-CRITICAL **: 23:28:07.368: fu_volume_is_mounted: assertion 'FU_IS_VOLUME (self)' failed
 
(fwupdate:3591): FuDeviceLocker-CRITICAL **: 23:28:07.368: fu_device_locker_new_full: assertion 'device != NULL' failed
Segmentation fault (core dumped)
```

ESP is at /boot/efi and is mounted:

```
mount | grep boot/efi
/dev/nvme0n1p1 on /boot/efi type vfat (rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,utf8,errors=remount-ro,discard)
```

**Expected behavior**
Update succeeds

**fwupd version information**
```
fwupdmgr --version
client version: 1.4.6
compile-time dependency versions
        gusb:   0.3.5
        efivar: 37
daemon version: 1.4.6
```

Also tried fwupd from git master - behavior is the same.

1.4.6 was installed via pacman

**fwupd device information**

```81X2
│
├─AMD Ryzen 5 4500U with Radeon Graphics:
│     Device ID:          4bde70ba4e39b28f9eab1628f9dd6e6244c03027
│     Current version:    0x8600102
[...]

PR #2407, opened 2020-09-21
uefi: Fix critical warning regression with 'fwupdate -a'

hughsie:
Set the ESP when creating the device with a known GUID.

Fixes https://github.com/fwupd/fwupd/issues/2406


superm1 (approved):
This looks correct to me, especially considering that the esp is an optional argument that can be passed.

anarsoul:
@hughsie thanks, it fixed the issue!

PR #2408, opened 2020-09-21
Wip/hughsie/firmware builder2

hughsie:


superm1:
I guess this is in many ways a lower level, more specific  implementation of `fu_common_firmware_builder`.

hughsie:
> more specific implementation of fu_common_firmware_builder

Yes, except with explicit values rather than running random binaries.

superm1 (approved):

PR #2411, opened 2020-09-22
Wip/hughsie/fu firmware image

hughsie:
Type of pull request:
- [X] Feature

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -1834,6 +1834,11 @@ fu_util_firmware_extract (FuUtilPrivate *priv, gchar **values, GError **error)
 		g_autofree gchar *fn = NULL;
 		g_autoptr(GBytes) blob_img = NULL;
 
+		/* get raw image without generated header, footer or crc */
+		blob_img = fu_firmware_image_get_bytes (img);
+		if (blob_img == NULL || g_bytes_get_size (blob_img) == 0)
+			continue;
+
 		/* use suitable filename */
 		if (fu_firmware_image_get_filename (img) != NULL) {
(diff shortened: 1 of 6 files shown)


src/fu-tool.c
```diff
@@ -2576,6 +2606,12 @@ main (int argc, char *argv[])
 		     /* TRANSLATORS: command description */
 		     _("Convert a firmware file"),
 		     fu_util_firmware_convert);
+	fu_util_cmd_array_add (cmd_array,
+		     "firmware-build",
+		     "BUILDER-XML FILENAME-DST",
```
superm1, line 2611: I guess actually having the XML as an input means you should document the expected schema somewhere.

hughsie: It's tricky, as the schema will depend on the subclass of FuFirmware. For instance, FuBcm57xxFirmware has attributes `model` and `vendor`.

superm1: Maybe a generic schema that would apply to everyone and leave optional fields as just that?

hughsie: Yup, that works. Any preference for XSD over RELAX NG, or do you just think an HTML document explaining the format is suitable? I could also do a _mega_-gtk-doc comment explaining it all in the existing generated documentation.

hughsie: I went with a mega-gtk-doc comment in the end, but I don't mind adding a schema too.

superm1: I think the gtk documentation is fine, it will end up in fwupd.github.io (although I acknowledge you'll need to know how to find it).

superm1 (approved):

PR #2415, opened 2020-09-24
Add fu_device_dump_firmware()

hughsie:
Conceptually we were trying to stuff subtly different actions into one vfunc:

 * Read firmware from the device to update the verification checksums

 * Read a firmware blob from the device for debugging

For the first action we might want to mask out the sections of the flash with
serial numbers (so the verification hashes match the ones published on the LVFS)
and for the second we want just a raw ROM file from the hardware with no
pre-processing that we can compare against an external SPI dumper.

Split out ->dump_firmware to get the raw blob, and allow plugins to also
implement ->read_firmware() if they have to mask out specific offsets or remove
specific images from the FuFirmware container.

In the common case when masking is not required, fall back to using a 'binary'
FuFirmware automatically to make most plugins simpler.

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -2515,6 +2515,11 @@ fu_device_prepare_firmware (FuDevice *self,
  * @error: A #GError
  *
  * Reads firmware from the device by calling a plugin-specific vfunc.
+ * The device subclass should try to ensure the firmware does not contain any
+ * serial numbers or user-configuration values and can be used to calculate the
+ * device checksum.
+ *
+ * The return value can be converted to a blob of memory using fu_firmware_write().
  *
  * Returns: (transfer full): A #FuFirmware, or %NULL for error
  *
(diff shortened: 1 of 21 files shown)


src/fu-tool.c
```diff
@@ -800,7 +800,7 @@ fu_util_firmware_read (FuUtilPrivate *priv, gchar **values, GError **error)
 			  G_CALLBACK (fu_util_update_device_changed_cb), priv);
 
 	/* dump firmware */
-	blob_fw = fu_engine_firmware_read (priv->engine, device, priv->flags, error);
+	blob_fw = fu_engine_firmware_dump (priv->engine, device, priv->flags, error);
```
superm1, line 803: does the fu-tool command want to be re-named now?

libfwupdplugin/fu-device.c
```diff
@@ -2515,6 +2515,10 @@ fu_device_prepare_firmware (FuDevice *self,
  * @error: A #GError
  *
  * Reads firmware from the device by calling a plugin-specific vfunc.
+ * The firmware will not contain any serial numbers or user-configuration values
+ * and can be used to calculate the device checksum.
```
superm1, line 2519: Isn't this assuming that the device SPI is partitioned in a such that the configuration specific stuff is in a certain region and can be ignored for a checksum purpose?  It seems like a general statement that can't be guaranteed.

src/fu-tool.c
hughsie: I'm not sure. I think to the end user it's still "Logically correct" in that it's reading the firmware from the device, it's just fwupd that cares about the distinction between `dump` and `read`. The user isn't going to want to read the firmware and do `sha1sum` on it manually to get the LVFS checksum, I don't think.

libfwupdplugin/fu-device.c
hughsie: Yes, it's assuming that the specific FuDevice subclass knows how to parse the firmware and blank something out at a known offset. Maybe the wording is just funky; how about `The device subclass should try to ensure the firmware does not contain any serial numbers or user-configuration values...`

superm1: That's better

src/fu-tool.c
superm1: I guess since there is also a parse command, I would think renaming to dump makes sense.  You can conceivable "dump" and then bring it into "parse".

hughsie: >  You can conceivable "dump" and then bring it into "parse".

Yes, that makes much more sense. Fixed, thanks.

PR #2417, opened 2020-09-25
bcm57xx: Add a plugin that can update the BCM5719 network adapter

hughsie:
Type of pull request:

### DIFF
--- a/plugins/bcm57xx/README.md
+++ b/plugins/bcm57xx/README.md
@@ -1,0 +1,28 @@
+BCM57xx Support
+===============
+
+Introduction
+------------
+
+This plugin updates BCM57xx wired network adaptors from Broadcom using a
+reverse-engineered flashing protocol. It is designed to be used with the
+clean-room reimplementation of the BCM5719 firmware found here:
+https://github.com/meklort/bcm5719-fw
+
+Protocol
+--------
+BCM57xx devices support a custom `com.broadcom.bcm57xx` protocol which is
+implemented as ioctls like ethtool does.
+
+GUID Generation
+---------------
+
+These devices use the standard PCI instance IDs, for example:
+
+ * `PCI\VEN_14E4&DEV_1657`
+ * `PCI\VEN_14E4&DEV_1657&SUBSYS_17AA222E`
+
+Vendor ID Security
+------------------
+
+The vendor ID is set from the PCI vendor, in this instance set to `PCI:0x14E4`
--- a/plugins/bcm57xx/bcm57xx.quirk
+++ b/plugins/bcm57xx/bcm57xx.quirk
@@ -1,0 +1,9 @@
+# Broadcom BCM5719
+[DeviceInstanceId=PCI\VEN_14E4&DEV_1657]
+Plugin = bcm57xx
+
+# Dell PCI card
+[DeviceInstanceId=PCI\VEN_14E4&DEV_1657&SUBSYS_14E41904]
+FirmwareSize = 0x80000
+[DeviceInstanceId=PCI\VEN_14E4&DEV_1657&SUBSYS_94E41904]
+FirmwareSize = 0x80000
--- a/plugins/bcm57xx/fu-bcm57xx-common.c
+++ b/plugins/bcm57xx/fu-bcm57xx-common.c
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2018 Evan Lojewski
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: GPL-2+
+ */
+
+#include "config.h"
+
+#include "fu-common.h"
+
+#include "fu-bcm57xx-common.h"
+
+guint32
+fu_bcm57xx_nvram_crc (const guint8 *buf, gsize bufsz)
+{
+	return GUINT32_FROM_BE(fu_common_crc32 (buf, bufsz));
+}
+
+gboolean
+fu_bcm57xx_verify_crc (GBytes *fw, GError **error)
+{
+	guint32 crc_actual;
+	guint32 crc_file = 0;
+	gsize bufsz = 0x0;
+	const guint8 *buf = g_bytes_get_data (fw, &bufsz);
+
+	/* expected */
+	if (!fu_common_read_uint32_safe (buf, bufsz, bufsz - sizeof(guint32),
+					 &crc_file, G_BIG_ENDIAN, error))
+		return FALSE;
+
+	/* reality */
+	crc_actual = fu_bcm57xx_nvram_crc (buf, bufsz - sizeof(guint32));
+	if (crc_actual != crc_file) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "invalid CRC, expected 0x%08x got: 0x%08x",
+			     (guint) crc_file, (guint) crc_actual);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+gboolean
+fu_bcm57xx_verify_magic (GBytes *fw, gsize offset, GError **error)
+{
+	guint32 magic = 0;
+	gsize bufsz = 0x0;
+	const guint8 *buf = g_bytes_get_data (fw, &bufsz);
+
+	/* hardcoded value */
+	if (!fu_common_read_uint32_safe (buf, bufsz, offset, &magic, G_BIG_ENDIAN, error))
+		return FALSE;
+	if (magic != BCM_NVRAM_MAGIC) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "invalid magic, got: 0x%x",
+			     (guint) magic);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+void
+fu_bcm57xx_veritem_free (Bcm57xxVeritem *veritem)
+{
+	g_free (veritem->branch);
+	g_free (veritem->version);
+	g_free (veritem);
+}
+
+Bcm57xxVeritem *
+fu_bcm57xx_veritem_new (const guint8 *buf, gsize bufsz)
+{
+	g_autofree gchar *tmp = NULL;
+	g_autoptr(Bcm57xxVeritem) veritem = g_new0 (Bcm57xxVeritem, 1);
+	struct {
+		const gchar *prefix;
+		const gchar *branch;
+		FwupdVersionFormat verfmt;
+	} data[] = {
+		{ "5719-v",	BCM_FW_BRANCH_UNKNOWN,		FWUPD_VERSION_FORMAT_PAIR },
+		{ "stage1-",	BCM_FW_BRANCH_OSS_FIRMWARE,	FWUPD_VERSION_FORMAT_TRIPLET },
+		{ NULL, NULL, 0 }
+	};
+
+	/* dont assume this is NUL terminated */
+	tmp = g_strndup ((const gchar *) buf, bufsz);
+	if (tmp == NULL || tmp[0] == '\0')
+		return NULL;
+
+	/* use prefix to define object */
+	for (guint i = 0; data[i].prefix != NULL; i++) {
+		if (g_str_has_prefix (tmp, data[i].prefix)) {
+			veritem->version = g_strdup (tmp + strlen (data[i].prefix));
+			veritem->branch = g_strdup (data[i].branch);
+			veritem->verfmt = data[i].verfmt;
+			return g_steal_pointer (&veritem);
+		}
+	}
+	veritem->verfmt = FWUPD_VERSION_FORMAT_UNKNOWN;
+	veritem->version = g_strdup (tmp);
+	return g_steal_pointer (&veritem);
+}
--- a/plugins/bcm57xx/fu-bcm57xx-common.h
+++ b/plugins/bcm57xx/fu-bcm57xx-common.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#pragma once
+
+#include <glib.h>
+#include <fwupd.h>
+
+#define BCM_VENDOR_BROADCOM			0x14E4
+
+#define BCM_FW_BRANCH_UNKNOWN			NULL
+#define BCM_FW_BRANCH_OSS_FIRMWARE		"oss-firmware"
+
+#define BCM_FIRMWARE_SIZE			0x40000 /* x2 for Dell */
+#define BCM_PHYS_ADDR_DEFAULT			0x08003800
+
+#define BCM_NVRAM_MAGIC				0x669955AA
+
+/* offsets into NVMRAM */
+#define BCM_NVRAM_HEADER_BASE			0x00
+#define BCM_NVRAM_DIRECTORY_BASE		0x14
+#define BCM_NVRAM_INFO_BASE			0x74
+#define BCM_NVRAM_VPD_BASE			0x100
+#define BCM_NVRAM_INFO2_BASE			0x200
+#define BCM_NVRAM_STAGE1_BASE			0x28c
+
+#define BCM_NVRAM_HEADER_MAGIC			0x00
+#define BCM_NVRAM_HEADER_PHYS_ADDR		0x04
+#define BCM_NVRAM_HEADER_SIZE_WRDS		0x08
+#define BCM_NVRAM_HEADER_OFFSET			0x0C
+#define BCM_NVRAM_HEADER_CRC			0x10
+#define BCM_NVRAM_HEADER_SZ			0x14
+
+#define BCM_NVRAM_INFO_MAC_ADDR0		0x00
+#define BCM_NVRAM_INFO_VENDOR			0x2C
+#define BCM_NVRAM_INFO_DEVICE			0x2E
+#define BCM_NVRAM_INFO_SZ			0x8C
+
+#define BCM_NVRAM_DIRECTORY_ADDR		0x00
+#define BCM_NVRAM_DIRECTORY_SIZE_WRDS		0x04
+#define BCM_NVRAM_DIRECTORY_OFFSET		0x08
+#define BCM_NVRAM_DIRECTORY_SZ			0x0c
+
+#define BCM_NVRAM_VPD_SZ			0x100
+
+#define BCM_NVRAM_INFO2_SZ			0x8c
+
+#define BCM_NVRAM_STAGE1_VERADDR		0x08
+#define BCM_NVRAM_STAGE1_VERSION		0x0C
+
+typedef struct {
+	gchar			*branch;
+	gchar			*version;
+	FwupdVersionFormat	 verfmt;
+} Bcm57xxVeritem;
+
+guint32		 fu_bcm57xx_nvram_crc		(const guint8	*buf,
+						 gsize		 bufsz);
+gboolean	 fu_bcm57xx_verify_crc		(GBytes		*fw,
+						 GError		**error);
+gboolean	 fu_bcm57xx_verify_magic	(GBytes		*fw,
+						 gsize		 offset,
+						 GError		**error);
+
+/* parses stage1 version */
+void		 fu_bcm57xx_veritem_free	(Bcm57xxVeritem	*veritem);
+Bcm57xxVeritem	*fu_bcm57xx_veritem_new		(const guint8	*buf,
+						 gsize		 bufsz);
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(Bcm57xxVeritem, fu_bcm57xx_veritem_free)
--- a/plugins/bcm57xx/fu-bcm57xx-device.c
+++ b/plugins/bcm57xx/fu-bcm57xx-device.c
@@ -1,0 +1,626 @@
+/*
+ * Copyright (C) 2018-2020 Evan Lojewski
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: GPL-2+
+ */
+
+#include "config.h"
+
+#ifdef HAVE_ETHTOOL_H
+#include <linux/ethtool.h>
+#include <linux/sockios.h>
+#include <net/if.h>
+#endif
+#ifdef HAVE_IOCTL_H
+#include <sys/ioctl.h>
+#endif
+#ifdef HAVE_SOCKET_H
+#include <sys/socket.h>
+#endif
+
+#include "fu-chunk.h"
+#include "fu-common.h"
+
+#include "fu-bcm57xx-common.h"
+#include "fu-bcm57xx-device.h"
+#include "fu-bcm57xx-recovery-device.h"
+#include "fu-bcm57xx-firmware.h"
+#include "fu-bcm57xx-dict-image.h"
+
+#define FU_BCM57XX_BLOCK_SZ		0x4000 /* 16kb */
+
+struct _FuBcm57xxDevice {
+	FuUdevDevice		 parent_instance;
+	FuBcm57xxRecoveryDevice	*recovery;
+	gchar			*ethtool_iface;
+	int			 ethtool_fd;
+};
+
+G_DEFINE_TYPE (FuBcm57xxDevice, fu_bcm57xx_device, FU_TYPE_UDEV_DEVICE)
+
+static void
+fu_bcm57xx_device_to_string (FuUdevDevice *device, guint idt, GString *str)
+{
+	FuBcm57xxDevice *self = FU_BCM57XX_DEVICE (device);
+	fu_common_string_append_kv (str, idt, "EthtoolIface", self->ethtool_iface);
+}
+
+static gboolean
+fu_bcm57xx_device_probe (FuUdevDevice *device, GError **error)
+{
+	FuBcm57xxDevice *self = FU_BCM57XX_DEVICE (device);
+	g_autofree gchar *fn = NULL;
+	g_autoptr(GPtrArray) ifaces = NULL;
+
+	/* only enumerate number 1 */
+	if (fu_udev_device_get_number (device) != 1) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "only device 1 supported on multi-device card");
+		return FALSE;
+	}
+
+	/* we need this even for non-recovery to reset APE */
+	fu_device_set_quirks (FU_DEVICE (self->recovery),
+			      fu_device_get_quirks (FU_DEVICE (self)));
+	fu_device_incorporate (FU_DEVICE (self->recovery), FU_DEVICE (self));
+	if (!fu_device_probe (FU_DEVICE (self->recovery), error))
+		return FALSE;
+
+	/* only if has an interface */
+	fn = g_build_filename (fu_udev_device_get_sysfs_path (device), "net", NULL);
+	ifaces = fu_common_filename_glob (fn, "en*", NULL);
+	if (ifaces == NULL || ifaces->len == 0) {
+		fu_device_add_child (FU_DEVICE (self), FU_DEVICE (self->recovery));
+	} else {
+		self->ethtool_iface = g_path_get_basename (g_ptr_array_index (ifaces, 0));
+	}
+
+	/* success */
+	return fu_udev_device_set_physical_id (device, "pci", error);
+}
+
+static gboolean
+fu_bcm57xx_device_nvram_write (FuBcm57xxDevice *self,
+			       guint32 address,
+			       const guint8 *buf,
+			       gsize bufsz,
+			       GError **error)
+{
+#ifdef HAVE_ETHTOOL_H
+	gsize eepromsz;
+	gint rc = -1;
+	struct ifreq ifr = { 0 };
+	g_autofree struct ethtool_eeprom *eeprom = NULL;
+
+	/* failed to load tg3 */
+	if (self->ethtool_iface == NULL) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "Not supported as ethtool interface disabled");
+		return FALSE;
+	}
+
+	/* sanity check */
+	if (address + bufsz > fu_device_get_firmware_size_max (FU_DEVICE (self))) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "tried to read outside of EEPROM size [0x%x]",
+			     (guint) fu_device_get_firmware_size_max (FU_DEVICE (self)));
+		return FALSE;
+	}
+
+	/* write EEPROM (NVRAM) data */
+	eepromsz = sizeof(struct ethtool_eeprom) + bufsz;
+	eeprom = (struct ethtool_eeprom *) g_malloc0 (eepromsz);
+	eeprom->cmd = ETHTOOL_SEEPROM;
+	eeprom->magic = BCM_NVRAM_MAGIC;
+	eeprom->len = bufsz;
+	eeprom->offset = address;
+	memcpy (eeprom->data, buf, eeprom->len);
+	strncpy (ifr.ifr_name, self->ethtool_iface, IFNAMSIZ - 1);
+	ifr.ifr_data = (char *) eeprom;
+#ifdef HAVE_IOCTL_H
+	rc = ioctl (self->ethtool_fd, SIOCETHTOOL, &ifr);
+#else
+	g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "Not supported as <sys/ioctl.h> not found");
+	return FALSE;
+#endif
+	if (rc < 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "cannot write eeprom [%i]", rc);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+#else
+	g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "Not supported as <linux/ethtool.h> not found");
+	return FALSE;
+#endif
+}
+
+static gboolean
+fu_bcm57xx_device_nvram_read (FuBcm57xxDevice *self,
+			      guint32 address,
+			      guint8 *buf,
+			      gsize bufsz,
+			      GError **error)
+{
+#ifdef HAVE_ETHTOOL_H
+	gsize eepromsz;
+	gint rc = -1;
+	struct ifreq ifr = { 0 };
+	g_autofree struct ethtool_eeprom *eeprom = NULL;
+
+	/* failed to load tg3 */
+	if (self->ethtool_iface == NULL) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "Not supported as ethtool interface disabled");
+		return FALSE;
+	}
+
+	/* sanity check */
+	if (address + bufsz > fu_device_get_firmware_size_max (FU_DEVICE (self))) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "tried to read outside of EEPROM size [0x%x]",
+			     (guint) fu_device_get_firmware_size_max (FU_DEVICE (self)));
+		return FALSE;
+	}
+
+	/* read EEPROM (NVRAM) data */
+	eepromsz = sizeof(struct ethtool_eeprom) + bufsz;
+	eeprom = (struct ethtool_eeprom *) g_malloc0 (eepromsz);
+	eeprom->cmd = ETHTOOL_GEEPROM;
+	eeprom->len = bufsz;
+	eeprom->offset = address;
+	strncpy (ifr.ifr_name, self->ethtool_iface, IFNAMSIZ - 1);
+	ifr.ifr_data = (char *) eeprom;
+#ifdef HAVE_IOCTL_H
+	rc = ioctl (self->ethtool_fd, SIOCETHTOOL, &ifr);
+#else
+	g_set_error_literal (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "Not supported as <sys/ioctl.h> not found");
+	return FALSE;
+#endif
+	if (rc < 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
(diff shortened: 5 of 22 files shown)


hughsie:
Most of the logic is courtesy of @meklort -- I've not actually had the guts to actually flash the hardware with the open source firmware yet, but it might work.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+	tmp.bits.WriteEnable = mode == FU_BCM57XX_NVRAM_MODE_ENABLED_WRITE;
+	fu_bcm57xx_device_bar_write (self, FU_BCM57XX_BAR_DEVICE, REG_NVM_ACCESS, tmp.r32);
+	return TRUE;
+}
+
+static gboolean
+fu_bcm57xx_device_nvram_acquire_lock (FuBcm57xxDevice *self, GError **error)
+{
+	RegNVMSoftwareArbitration_t tmp = { 0 };
+	GTimer *timer = g_timer_new ();
```
superm1, line 149: missing g_timer_destroy() no?

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+	fu_bcm57xx_device_bar_write (self, FU_BCM57XX_BAR_DEVICE,
+				     REG_NVM_SOFTWARE_ARBITRATION, tmp.r32);
+	return TRUE;
+}
+
+static gboolean
+fu_bcm57xx_device_nvram_wait_done (FuBcm57xxDevice *self, GError **error)
+{
+	RegNVMCommand_t tmp = { 0 };
+	GTimer *timer = g_timer_new ();
```
superm1, line 187: missing g_timer_destroy

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+			     bufsz_wrds * sizeof(guint32), error))
+		return FALSE;
+
+	/* success */
+	return TRUE;
+#else
+	g_set_error (error,
+		     FWUPD_ERROR,
+		     FWUPD_ERROR_NOT_SUPPORTED,
+		     "Not supported as <sys/ioctl.h> not found");
```
superm1, line 552: isn't this supposed to be ethtool.h?

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+	strncpy (ifr.ifr_name, self->ethtool_iface, IFNAMSIZ - 1);
+	ifr.ifr_data = (char *) &drvinfo;
+#ifdef HAVE_IOCTL_H
+	rc = ioctl (self->ethtool_fd, SIOCETHTOOL, &ifr);
+#endif
+	if (rc < 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "cannot get driver information [%i]", rc);
```
superm1, line 498: if you have ethtool and don't have ioctl.h, this error code will be wrong.  maybe the `#ifdef` needs `#else` clause for that.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+	return TRUE;
+}
+
+static void
+fu_bcm57xx_device_init (FuBcm57xxDevice *self)
+{
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_CAN_VERIFY_IMAGE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_INTERNAL);
```
superm1, line 728: you can't guarantee this, could be in a PCIe enclosure

superm1 (commented):
So I think you need to look at the firmware installed to confirm if it has that magic header for the custom firmware and add a GUID specific to the custom firmware when it's present.  This should make sure that people on the regular firmware aren't forced onto the custom firmware.

And can people "switch" using this plugin?  I would guess so, but it should probably require `--force` or some sort of incantation.

libfwupdplugin/fu-udev-device.c
```diff
@@ -743,6 +743,28 @@ fu_udev_device_get_sysfs_path (FuUdevDevice *self)
...
+guint64
+fu_udev_device_get_number (FuUdevDevice *self)
+{
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (FU_IS_UDEV_DEVICE (self), 0);
+	if (priv->udev_device != NULL)
+		return fu_common_strtoull (g_udev_device_get_number (priv->udev_device));
+#endif
+	return 0;
```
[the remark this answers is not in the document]
hughsie: Yes, good point, fixed.

plugins/bcm57xx/README.md
```diff
@@ -0,0 +1,28 @@
...
+BCM57xx devices support a custom `com.broadcom.bcm57xx` protocol which is
+implemented as ioctls like ethtool does.
+
+GUID Generation
+---------------
+
+These devices use the standard PCI instance IDs, for example:
+
+ * `PCI\VEN_14E4&DEV_1657`
+ * `PCI\VEN_14E4&DEV_1657&SUBSYS_17AA222E`
```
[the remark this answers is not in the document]
hughsie: I *think* the plan here is to use a different subsystem PID for the open source firmware.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: Fixed.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: Indeed, thanks.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: Ya, I did think it would just call back to rc==-1, but I agree it's more explicit with a #else

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: Unlikely, but true I guess.

plugins/bcm57xx/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Vendor ID Security
+------------------
+
+The vendor ID is set from the PCI vendor, for example set to `PCI:{$manfid}`
```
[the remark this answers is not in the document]
hughsie: I don't think the oss firmware author wants to change the VID/PID so that the tg3 driver still works unmodified. I'll hardcode it in the README.

plugins/bcm57xx/fu-bcm57xx-common.c
```diff
@@ -0,0 +1,79 @@
...
+guint32
+fu_bcm57xx_nvram_crc (const guint8 *buf, guint32 bufsz, guint32 crc)
+{
+	for (guint32 idx = 0; idx < bufsz; idx++) {
+		guint8 data = *buf++;
+		for (guint32 bit = 0; bit < 8; bit++, data >>= 1) {
+			crc = (crc >> 1) ^ (((crc ^ data) & 1) ? CRC32_POLYNOMIAL : 0);
+		}
+	}
+	return GUINT32_FROM_BE(~crc);
```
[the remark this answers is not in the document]
hughsie: it's now one line, indeed :)

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+
+	/* do we have a driver (e.g. tg3) providing an ethtool interface */
+	fn = g_build_filename (fu_udev_device_get_sysfs_path (device), "net", NULL);
+	ifaces = fu_common_filename_glob (fn, "en*", NULL);
+	if (ifaces != NULL && ifaces->len > 0) {
+		self->ethtool_iface = g_path_get_basename (g_ptr_array_index (ifaces, 0));
+	} else {
+		g_clear_pointer (&self->ethtool_iface, g_free);
+	}
+}
```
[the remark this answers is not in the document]
hughsie: Good catch! You're correct of course. I've switched everthing to call `fu_bcm57xx_device_ensure_ethtool_iface` before each check.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,767 @@
...
+	fu_bcm57xx_device_ensure_ethtool_iface (self);
+	return fu_udev_device_set_physical_id (device, "pci", error);
+}
+
+#ifdef __ppc64__
+#define BARRIER()	__asm__ volatile ("sync 0\neieio\n" : : : "memory")
+#else
+#define BARRIER()	__asm__ volatile ("" : : : "memory");
+#endif
+
```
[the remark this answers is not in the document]
hughsie: I believe it should work with lockdown; I don't _think_ I'm writing to writing random system memory, I think the driver is writing to a `mmap`'d PCI resource.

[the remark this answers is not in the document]
hughsie: Ohh, the barrier is just because we're reading from a mmio resource; I don't think it's hugely important for x64 but for ppc64le it seems like it's more so.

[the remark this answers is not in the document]
superm1: Why can't it use `fu_udev_device_pread` or `fu_udev_device_pwrite` then?

hughsie:
> And can people "switch" using this plugin? I would guess so, but it should probably require --force or some sort of incantation.

I've not worked this out yet; we need to do this same kind of balance act for hardware with nonfree BIOS / free software coreboot firmware.

superm1:
> I've not worked this out yet; we need to do this same kind of balance act for hardware with nonfree BIOS / free software coreboot firmware.

Well and not to mention if flashing this causes your system to no longer be able to secure boot w/ PXE etc.

hughsie:
> Well and not to mention if flashing this causes your system to no longer be able to secure boot w/ PXE etc.

Yes, agree. PXE is completely nonfunctional with the open source firmware -- at the moment. I'm not sure of the grand plan. We'd certainly need to show a pretty big warning.

superm1:
>  -- at the moment. I'm not sure of the grand plan

Even if it's part of grand plan, that is going to require signing the DXE with something and that something being trusted by the system firmware.  So it's not going to be an easy path at all.

hughsie:
> signing the DXE with something and that something being trusted by the system firmware

Maybe an odd question; but do we need the PXE blob? I know we'll lose things like http boot, but the kernel should still load tg3 even with sb turned on, right?

libfwupdplugin/fu-udev-device.c
```diff
@@ -743,6 +743,28 @@ fu_udev_device_get_sysfs_path (FuUdevDevice *self)
...
+ * @self: A #FuUdevDevice
+ *
+ * Gets the device number, if any.
+ *
+ * Returns: integer, or 0 if unavailable
+ *
+ * Since: 1.5.0
+ **/
+guint64
+fu_udev_device_get_number (FuUdevDevice *self)
```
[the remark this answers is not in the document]
hughsie: https://github.com/fwupd/fwupd/pull/2419

plugins/bcm57xx/fu-bcm57xx-device.c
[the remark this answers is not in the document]
hughsie: Well, I'm not reading from a fd, I'm reading an address. Or do you think I can `open()` the `resource4` file and then do `pread`?

superm1:
> Maybe an odd question; but do we need the PXE blob? I know we'll lose things like http boot, but the kernel should still load tg3 even with sb turned on, right?

Beyond networking stuff in BIOS you might also "lose configurability".  That is if the option rom can launch into a configuration applet that might store some settings in VPD.  Maybe this is programming a MAC, or enabling/ disabling a port.  It might not even be something the Broadcom option rom offers, but just types of things that would be there.

plugins/bcm57xx/fu-bcm57xx-device.c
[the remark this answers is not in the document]
superm1: Isn't the resource4 just pointing to pci address space?  and you're interacting with that pci address space?  I would think an ioctl out to that address space should work just fine.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,779 @@
...
+	return TRUE;
+}
+
+static void
+fu_bcm57xx_device_init (FuBcm57xxDevice *self)
+{
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_CAN_VERIFY_IMAGE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
```
superm1, line 741: Does it need a reboot?  or can you just issue a PCI reset from Linux?

hughsie: For the APE update (application program) yes, you can reboot the firmware and it comes back up. But stage1 actually needs a full system reset :(

meklort:
> > signing the DXE with something and that something being trusted by the system firmware
> 
> Maybe an odd question; but do we need the PXE blob? I know we'll lose things like http boot, but the kernel should still load tg3 even with sb turned on, right?

Just a note:
The firmware's primary purpose is for POWER9 based systems form Raptor Computing Systems. These machines use petitboot and do not support DXE/UEFI type binaries. With petitboot they are able to boot off the network.

Note that the hardware will come up with default settings and tg3 will still load even if the NVM isn't present / is completely empty. I'll take a look at implementing support for stage1 to provide the optionrom binary to an x86 machine, but it's not a priority for me. My goal here would be to use iPXE's payload, as this already exists and support the hardware.

Somewhat related, iPXE could potentially be supported here as a replacement when using the proprietary firmware for stage1/APE. So, I could potentially see a standalone iPXE binary being in LVFS.

plugins/bcm57xx/fu-bcm57xx-device.c
meklort: Some clarification:
the APE can and must be reset by writing to a specific register. A PCI reset does not cause the APE to reset. The alternative is to power cycle the machine / card (rebooting the OS is not sufficient).

For stage1, This actually is reloaded by the hardware on a PCI reset. the stage1 code can also be reset manually by setting registers. That said, the stage1 code modifies PCI configuration registers, and so resetting the firmware here without doing a PCI rescan can be problematic.
A full system reset is of course safest here.

hughsie: > the APE can and must be reset by writing to a specific register

Do you know if you can also do this with the ethtool api too?

plugins/bcm57xx/fu-bcm57xx-common.h
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#pragma once
+
+#include <glib.h>
+
+#define BCM_VENDOR_BROADCOM			0x14E4
+#define BCM_FIRMWARE_SIZE			0x80000
```
meklort, line 12: This is true for the Dell NIC, however the Talos II and Balckbird have a size of 0x40000

plugins/bcm57xx/fu-bcm57xx-firmware.c
```diff
@@ -0,0 +1,571 @@
...
+	gsize			 source_size;
+	guint8			 source_padchar;
+};
+
+G_DEFINE_TYPE (FuBcm57xxFirmware, fu_bcm57xx_firmware, FU_TYPE_FIRMWARE)
+
+#define BCM_STAGE1_HEADER_MAGIC_BROADCOM	0x0E000E03
+#define BCM_STAGE1_HEADER_MAGIC_MEKLORT		0x3C1D0800
+
+#define BCM_APE_HEADER_MAGIC			0x42434D1A
```
meklort, line 34: With the APE format now matching the NVM format, I'm seeing 0x1A4D4342 here.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,847 @@
...
+					    (FuDeviceLockerFunc) fu_bcm57xx_device_nvram_disable,
+					    error);
+	if (locker == NULL)
+		return FALSE;
+
+	/* get NVRAM version */
+	if (!fu_bcm57xx_device_nvram_read (self, BCM_NVRAM_STAGE1_BASE + BCM_NVRAM_STAGE1_VERSION,
+					   &fwversion, 1, error))
+		return FALSE;
+	if (fwversion == 0x0) {
```
meklort, line 638: if (fwversion != 0x0)

plugins/bcm57xx/fu-bcm57xx-device.c
meklort: It looks like this cannot be done using the ethtool api.

meklort (commented):
I needed to make a few changes to get things closer to working on my side.
Also note that I've started adding support for generating the .cab file at https://github.com/meklort/bcm5719-fw/tree/fwupd

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,847 @@
...
+#include <valgrind.h>
+#endif /* HAVE_VALGRIND */
+
+typedef struct {
+	guint8	*buf;
+	gsize	 bufsz;
+} FuBcm57xxMmap;
+
+#define FU_BCM57XX_BAR_DEVICE	0
+#define FU_BCM57XX_BAR_APE	2
```
meklort, line 45: This should be #define FU_BCM57XX_BAR_APE	1
APE is at resource2, which is in bar[1] since you are multiplying *2 inside of fu_bcm57xx_device_open

hughsie: Indeed, thanks!

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: Fixed, thanks.

plugins/bcm57xx/fu-bcm57xx-firmware.c
hughsie: Ack, thanks.

plugins/bcm57xx/fu-bcm57xx-common.h
hughsie: Ohh! I assumed they'd all use the same SPI chip, apologies. I've set the default to 0x40000 and quirked the Dell card to 0x80000 now.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: What about if we do a full system *power down* rather than just a warm reset?

plugins/bcm57xx/bcm57xx.quirk
```diff
@@ -0,0 +1,7 @@
+# Broadcom BCM5719
+[DeviceInstanceId=PCI\VEN_14E4&DEV_1657]
+Plugin = bcm57xx
+
+# Dell PCI card
+[DeviceInstanceId=PCI\VEN_14E4&DEV_1657&SUBSYS_94E41904]
```
superm1, line 6: Are you sure this subsystem vendor ID isn't from the host system?  Most of the devices in my system pick up the host's subssystem vendor ID.

Example:
```
00:1f.5 Serial bus controller [0c80]: Intel Corporation Comet Lake SPI (flash) Controller [8086:02a4]
        Subsystem: Dell Comet Lake SPI (flash) Controller [1028:0962]
        Flags: fast devsel
        Memory at fe010000 (32-bit, non-prefetchable) [size=4K]
```

hughsie: > Are you sure this subsystem vendor ID isn't from the host system

I'm pretty sure the subsystem ID is part of the PCI spec for the device; I've seen other cards just change the subsys IDs when they are rebranded for OEMs, but the VID/PID remains the same so the kernel driver still binds correctly.

meklort: There are populated and match the values in NVM

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,860 @@
...
+	/* get NVRAM version */
+	if (!fu_bcm57xx_device_nvram_read (self, BCM_NVRAM_STAGE1_BASE + BCM_NVRAM_STAGE1_VERSION,
+					   &fwversion, 1, error))
+		return FALSE;
+	if (fwversion != 0x0) {
+		g_autofree gchar *fwversion_str = NULL;
+
+		/* this is always 0x0 on the non-free firmware */
+		fu_device_set_version_format (device, FWUPD_VERSION_FORMAT_TRIPLET);
+		fu_device_set_version (device, fwversion_str);
```
meklort, line 643: This is NULL here and causes a failure when trying to flash.

 `Device NetXtreme BCM5719 Gigabit Ethernet PCIe [48e3a2f898a9e9d83dd82699ca2c453b5f52fc36] has no firmware version`

plugins/bcm57xx/fu-bcm57xx-device.c
meklort: So, the only way a full power down will work is if power is completely cut standby power cannot be present. That is, the machine must be disconnected from the power outlet, or the PSU power switch must be flipped.

To clarify:
- The APE can be reset by writing to the RegAPEMode_t with Reset = 1 and FastBoot = 0 to cause it to boot from flash
- The APE can also be reset bu pulling power for about 30 seconds. The system needs to be *completely* powered down - the PSU must be switched off or the machine needs to be unplugged from the wall. 

- The AST2500 BMC on the board maintains state information about the NIC that was gathered via NCSI when it first booted. Most of this can be reset by sshing in to the BMC itself and running ifconfig down/up on the BMC itself. The other option here is to reset the BMC or power cycle it. (a host reset is not sufficient).

So.. the best option to make sure everything is in a known state is to power down the machine and unplug it for about 30 seconds.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,860 @@
...
+	}
+
+	for (guint i = 0; i < bufsz; i++) {
+		RegNVMCommand_t tmp = { 0 };
+		fu_bcm57xx_device_nvram_clear_done (self);
+		fu_bcm57xx_device_bar_write (self, FU_BCM57XX_BAR_DEVICE,
+					     REG_NVM_ADDR, address);
+		tmp.bits.Doit = 1;
+		tmp.bits.First = (i == 0);
+		tmp.bits.Last = (i == bufsz - 1);
```
meklort, line 369: I had problems when I originally wrote the code in the fw repo with this. Changing this to write one word at a time fixes the data corruption issue I was seeing before, although it's not the fastest.

I'll see if I can get burst writes working, but for now changing this to the following gets writes working:
tmp.bits.First = 1;
tmp.bits.Last = 1;

meklort: Note that I'll also work on updating my repo to use ethtool for writing, maybe switching to that here makes sense too.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,860 @@
...
+	if (!fu_firmware_parse (firmware, fw, FWUPD_INSTALL_FLAG_NONE, error))
+		return NULL;
+
+	/* remove images that will contain user-data */
+	if (!fu_firmware_remove_image_by_id (firmware, "info", error))
+		return NULL;
+	if (!fu_firmware_remove_image_by_id (firmware, "info2", error))
+		return NULL;
+	if (!fu_firmware_remove_image_by_id (firmware, "vpd", error))
+		return NULL;
```
meklort, line 506: This causes the firmware image that is written back to hardware to have these sections erased.

meklort:
With the issues mentioned previously worked-around, I was able to flash a functional ape image to the Talos II.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,860 @@
...
+static gboolean
+fu_bcm57xx_device_write_firmware (FuDevice *device,
+				  FuFirmware *firmware,
+				  FwupdInstallFlags flags,
+				  GError **error)
+{
+	FuBcm57xxDevice *self= FU_BCM57XX_DEVICE (device);
+	const guint8 *buf;
+	gsize bufsz = 0;
+	gsize bufsz_dwrds = fu_device_get_firmware_size_max (FU_DEVICE (self)) / sizeof(guint32);
```
meklort, line 571: If fu_device_get_firmware_size_max is ever something larger than the actual nvm/eeprom size in hardware, this could cause a wrap-around in hardware and cause garbage to be written to eeprom. Maybe it'd be better to only write the size of the firmware image provided in the update.cab file? This would mean that a small firmware update image would result in a portion of the eeprom being unmodified.

(note: I've update the bcm5719-fw repo to  generate an image that's as small as possible).

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: Many thanks, changed.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: My mistake, thanks. Fixed.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: I had the commands the wrong way around, fixed, thanks.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: > only write the size of the firmware image provided in the update.cab file

Indeed, good catch. Fixed.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: > although it's not the fastest

It took over 16 minutes to write the image using single word writes; I'd hate to think what's this is doing to the flash write limit if it's doing something like a 1024 block read, erase, write for every dword..

hughsie: I just switched the default to use the ethtool API for writing. I think that's a lot safer if there's a data corruption issue using raw writes.

plugins/bcm57xx/fu-bcm57xx-ethtool-device.c
```diff
@@ -0,0 +1,534 @@
...
+			     (guint) fu_device_get_firmware_size_max (FU_DEVICE (self)));
+		return FALSE;
+	}
+
+	/* write EEPROM (NVRAM) data */
+	eepromsz = sizeof(struct ethtool_eeprom) + bufsz_wrds * sizeof(guint32);
+	eeprom = (struct ethtool_eeprom *) g_malloc0 (eepromsz);
+	eeprom->cmd = ETHTOOL_SEEPROM;
+	eeprom->len = bufsz_wrds * sizeof(guint32);
+	eeprom->offset = address;
```
meklort, line 104: This returns -1 if the following isn't added:
    eeprom->magic = BCM_NVRAM_MAGIC;

hughsie: Indeed, thanks!

hughsie:
@superm1 did you want to (re-?) review this? Quite a bit has changed since the original commit -- but "no" is a perfectly valid answer too.

plugins/bcm57xx/fu-bcm57xx-common.c
```diff
@@ -0,0 +1,111 @@
...
+	} data[] = {
+		{ "5719-v",	BCM_FW_BRANCH_UNKNOWN,		FWUPD_VERSION_FORMAT_PAIR },
+		{ "stage1-",	BCM_FW_BRANCH_OSS_FIRMWARE,	FWUPD_VERSION_FORMAT_TRIPLET },
+		{ NULL, NULL, 0 }
+	};
+
+	/* dont assume this is NUL terminated */
+	tmp = g_strndup ((const gchar *) buf, bufsz);
+	if (tmp == NULL || tmp[0] == '\0')
+		return NULL;
```
superm1, line 97: shouldn't the check for this have happened on `buf` before copying the string?  IE if `buf[0]` was `\0`, but only 1 byte - you just overflowed if bufsz was more than 1 byte.
Similarly shouldn't there be a check on how big bufsz is before you copy?

plugins/bcm57xx/fu-bcm57xx-dict-image.c
```diff
@@ -0,0 +1,171 @@
...
+	const guint8 *buf;
+	gsize bufsz = 0;
+	guint32 crc;
+	g_autoptr(GByteArray) blob = NULL;
+	g_autoptr(GBytes) fw_nocrc = NULL;
+
+	/* get the CRC-less data */
+	fw_nocrc = fu_firmware_image_get_bytes (image);
+	if (fw_nocrc == NULL || g_bytes_get_size (fw_nocrc) == 0)
+		return g_bytes_new (NULL, 0);
```
superm1, line 60: don't this need to set error?  Or it's expected scenario?

plugins/bcm57xx/fu-bcm57xx-dict-image.c
```diff
@@ -0,0 +1,171 @@
...
+		{ 0x05,	0x00,	"iscsi2" },
+		{ 0x0b,	0x00,	"iscsi3" },
+		{ 0x00,	0x01,	"cfg1000" },
+		{ 0x04,	0x01,	"vpd2" },
+		{ 0xff,	0xff,	NULL }
+	};
+	if (self->target == 0xff || self->kind == 0xff)
+		return;
+	for (guint i = 0; ids[i].id != NULL; i++) {
+		if (self->target == ids[i].target && self->kind == ids[i].kind) {
```
superm1, line 112: If new ID's are found that aren't recognized is it useful to emit some messaging at this time?

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
```diff
@@ -0,0 +1,753 @@
...
+			g_set_error (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_NOT_SUPPORTED,
+				     "could not stat %s", fn);
+			close (memfd);
+			return FALSE;
+		}
+
+		/* mmap */
+		g_debug ("mapping BAR[%u] %s for 0x%x bytes", i, fn, (guint) st.st_size);
```
superm1, line 646: this should probably be hidden behind a plugin verbose flag

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
```diff
@@ -0,0 +1,753 @@
...
+{
+	/* unbind tg3 */
+	return fu_device_unbind_driver (device, error);
+}
+
+static gboolean
+fu_bcm57xx_recovery_device_attach (FuDevice *device, GError **error)
+{
+	/* bind tg3 */
+	return fu_device_bind_driver (device, "pci", "tg3", error);
```
superm1, line 384: Thinking through the scenarios here - what if someone doesn't have tg3 compiled as a module or into their kernel?  This presumably will fail, but I don't think that should actually fail the end of the upgrade / write process.  It probably should be a warning.

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -0,0 +1,598 @@
...
+		return FALSE;
+
+	/* get NVRAM version */
+	if (!fu_bcm57xx_device_nvram_read (self, BCM_NVRAM_STAGE1_BASE + BCM_NVRAM_STAGE1_VERSION,
+					   (guint8 *) &fwversion, sizeof(guint32), error))
+		return FALSE;
+	if (fwversion != 0x0) {
+		g_autofree gchar *fwversion_str = NULL;
+
+		/* this is only set on the non-free firmware */
```
superm1, line 494: isn't this comment backwards?

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
```diff
@@ -0,0 +1,753 @@
...
+			return FALSE;
+		tmp.bits.Wr = 1;
+		tmp.bits.Doit = 1;
+		tmp.bits.First = 1; /* this is really slow but prevents corruption */
+		tmp.bits.Last = 1;
+		if (!fu_bcm57xx_recovery_device_bar_write (self, FU_BCM57XX_BAR_DEVICE,
+							   REG_NVM_COMMAND, tmp.r32, error))
+			return FALSE;
+		if (!fu_bcm57xx_recovery_device_nvram_wait_done (self, error)) {
+			g_prefix_error (error, "failed to read @0x%x: ", address);
```
meklort, line 362: write

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
```diff
@@ -0,0 +1,753 @@
...
+	for (guint i = 0; i < bufsz; i++) {
+		BcmRegNVMCommand tmp = { 0 };
+		guint32 val32 = 0;
+		if (!fu_bcm57xx_recovery_device_nvram_clear_done (self, error))
+			return FALSE;
+		if (!fu_bcm57xx_recovery_device_bar_write (self, FU_BCM57XX_BAR_DEVICE,
+						       REG_NVM_ADDR, address, error))
+			return FALSE;
+		tmp.bits.Doit = 1;
+		/* write one word at a time to avoid data corruption issue */
```
meklort, line 317: read

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
```diff
@@ -0,0 +1,753 @@
...
+		tmp.bits.First = TRUE;
+		tmp.bits.Last = TRUE;
```
meklort, line 319: So, I've been running into an issue when using a function other than 0 often times out when reading individual words. I don't see this with the utility I have, or when I change this code to do a single burst read.

I do not see this issue when using function 0, only function 1 or later (fwupd is always using the last function, not the first)

Also note that the ethtool apis show the same issue when using a function other than 0. (the tg3 driver does reads similar to this, which results in the timeout).

meklort: Also note that if I update bcmflash to use this sequence, it also ends up hanging in certain cases.

So, the most reliable way to do this here may be to use something like:
		if (i ==0)
			tmp.bits.First = TRUE;
		else
			tmp.bits.First = FALSE;
		if (i + sizeof(guint32) < bufsz)
			tmp.bits.Last = FALSE;
		else
			tmp.bits.Last = TRUE;

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
```diff
@@ -0,0 +1,753 @@
...
+			return FALSE;
+		if (!fu_bcm57xx_recovery_device_bar_write (self, FU_BCM57XX_BAR_DEVICE,
+							   REG_NVM_WRITE, GUINT32_TO_BE(buf[i]), error))
+			return FALSE;
+		if (!fu_bcm57xx_recovery_device_bar_write (self, FU_BCM57XX_BAR_DEVICE,
+							   REG_NVM_ADDR, address, error))
+			return FALSE;
+		tmp.bits.Wr = 1;
+		tmp.bits.Doit = 1;
+		tmp.bits.First = 1; /* this is really slow but prevents corruption */
```
meklort, line 356: I've updated the code in my repo to properly bust write: https://github.com/meklort/bcm5719-fw/commit/85e0e3fe2ba84b54fb0282fdfae1775e6e25e6cf

Effectively, write must not cross a page boundary. Once .First and .Last are set based on the page, everything is working correctly with the burst writes.

plugins/bcm57xx/fu-bcm57xx-common.c
hughsie: g_strndup() mallocs memory under the covers, so it it's the equivalent of:

len = strlen(str)
newbuf = malloc(len+1)
memcpy (newbuf, buf, len)
newbuf[len] = '\0'

i.e. you can be sure that whatever happens `tmp` in this case is a newly allocated null terminated string.

plugins/bcm57xx/fu-bcm57xx-dict-image.c
hughsie: That's a good question. `fw_nocrc == NULL` should be the error, and even for an empty section we should write a CRC.

Fixed.

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
hughsie: Agree, fixed.

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
hughsie: > but I don't think that should actually fail the end of the upgrade

Agreed, fixed.

plugins/bcm57xx/fu-bcm57xx-device.c
hughsie: It is! Fixed.

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
hughsie: Fixed, thanks.

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
hughsie: Fixed, thanks.

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
hughsie: So maybe just:

		tmp.bits.First = i == 0;
		tmp.bits.Last = i == bufsz - 1;

`i` is incrementing the address in guint32's, so I think it's just `- 1`

plugins/bcm57xx/fu-bcm57xx-recovery-device.c
hughsie: I think for fwupd, where we're only using the recovery device to flash a **complete** backup image we can simplify things quite a bit and just check the address is aligned with 256, and the bufsz_dwrds is modulo 64.

Then I think this might just work:

		tmp.bits.First = i % page_size_dwrds == 0;
		tmp.bits.Last = (i + 1) % page_size_dwrds == 0;

I'll test in a minute on actual hardware. Thanks for figuring it out and explaining it to me!

superm1 (approved):

PR #2418, opened 2020-09-25
Add common CRC32 routines

hughsie:
We have quite a few versions of CRC32 in-tree, and are about to get two more...


superm1 (approved):
commit message should be updated since it's more than just CRC32 routines

PR #2419, opened 2020-09-25
trivial: Add fu_udev_device_get_number()

hughsie:
Type of pull request:
- [X] Feature

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -744,6 +744,29 @@ fu_udev_device_get_sysfs_path (FuUdevDevice *self)
 }
 
 /**
+ * fu_udev_device_get_number:
+ * @self: A #FuUdevDevice
+ *
+ * Gets the device number, if any.
+ *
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -743,6 +743,28 @@ fu_udev_device_get_sysfs_path (FuUdevDevice *self)
 	return NULL;
 }
 
+/**
+ * fu_udev_device_get_number:
+ * @self: A #FuUdevDevice
+ *
+ * Gets the device number, if any.
+ *
+ * Returns: integer, or 0 if unavailable
```
superm1, line 752: comment needs fixing, should be G_MAXUINT64

hughsie: Well, it's a a bit funcky but it's not untrue. In the case of a random device where numbering isn't available, the return value will be zero. When udev isn't available it'll be G_MAXUINT64. How about `Returns: integer, 0 if the number is unavailable, or %G_MAXUINT64 if udev is not available`?

superm1: That's good

PR #2420, opened 2020-09-25
Add the concept of firmware 'branches'

hughsie:
This allows a device to identify with different streams, for instance a Lenovo
laptop could have a coreboot firmware or a AMI firmware. The GUIDs would be the
same, but switching firmware would only be done rarely and very carefully.

Another example would be switching the Broadcom BCM57xx nework adaptors from the
vendor nonfree firmware with a signed PXE image, to the free software reverse
engineered driver with no PXE support (and thus no signed DXE) at all.

It is expected firmware would have additional metadata something like this:

    ...
    <branch>sdcc</branch>
    <description>
      <p>
        This is an alternate firmware built by the community using only free
        software tools.
      </p>
      <p>
        Your hardware may be damaged using this hardware, and installing this
        alternate firmware may void any warranty with the vendor.
      </p>
    </description>
    <requires>
      <id compare="ge" version="1.5.0">org.freedesktop.fwupd</id>
      <client>switch-branch</client>
    </requires>
    ...

Additionally, alternate branch firmware will not be returned for clients not
setting the FWUPD_FEATURE_FLAG_SWITCH_BRANCH before the GetReleases request.


hughsie:
Totally untested, this also needs the LVFS side here: https://gitlab.com/fwupd/lvfs-website/-/merge_requests/805

superm1:
I don't know branch is the right nomenclature for this concept.  Branch makes me think it's like a fork of the upstream firmware and will merge back in.  That might be true for some firmwares, but the cases you describe not at all the case.

How about "channel" instead?

hughsie:
Although I agree branch isn't perfect, it's the existing name of the tag in the AppStream specification. e.g. https://github.com/hughsie/appstream-glib/blob/master/libappstream-glib/as-app.c#L2596 -- it's already used by flatpak for this kind of purpose. It's perhaps not 100% the same as multiple branches can be installed at the same time. That said I'd rather stick to the AppStream specification where we can do, rather than having another sucky `<value key="LVFS::Channel">oss</value> or `<X-channel>oss</X-channel>`.

Source is also out of the equation as this describes where the firmware is coming from, rather than the variant. Note: `variant` is also claimed for the `<name_variant_suffix>` thing we use for different logical firmwares for the same physical device (e.g. usb hub).

src/fu-engine.c
```diff
@@ -4413,6 +4431,24 @@ fu_engine_get_releases_for_device (FuEngine *self,
...
+			break;
+		}
+	}
+	if (has_multiple_branches) {
+		for (guint i = 0; i < releases->len; i++) {
+			FwupdRelease *rel = FWUPD_RELEASE (g_ptr_array_index (releases, i));
+			if (g_strcmp0 (fwupd_release_get_branch (rel),
+				       fu_device_get_branch (device)) == 0) {
+				fwupd_release_add_flag (rel, FWUPD_RELEASE_FLAG_IS_CURRENT_BRANCH);
+			}
```
superm1, line 4448: Don't you also need to set alternate branch here?

superm1 (commented):
i don't know switch branch should be in the main tool, are you sure about that?  I would think it's better to live in fwupdtool.  In either case bash completion is needed for updates

hughsie:
> i don't know switch branch should be in the main tool

I think we should leave fwupdtool as a debugging tool, not an advanced tool. I do wonder if changing the branch should require a higher policykit authorisation, although from a fwupd-daemon point of view firmware is either signed, or unsigned -- and version 1.23 is no different to 99.99 on a different branch. I suppose if it's signed we at least know the metadata hasn't been altered since it came from the LVFS so maybe we can do something in the non-unsigned case,

I'm pretty sure we need to have something better on the CLI -- I only really turned `reinstall` into _install any specified version on any specified branch, as well as reinstall_ as a quick hack to see if the idea worked. We probably need a way to show what branches are available, or to show devices that can switch branch using a filter or something.

It does get a bit more confusing when talking about the UX of changing branches. Would I do `fwupdmgr update DEVICE_ID BRANCH_ID` to go from official vendor 1.2.3 to random-vendor 0.0.1? An update downgrades? It would be weird.

The core design issue here is that I've made it the device itself to say what branch it's on, rather than have a quirk or config file entry that selects the filtered branch; the latter of which doesn't work super well with our existing API. I think it has to be a atomic operation to _switch branch to foo and install one of the releases on foo_ rather than having the weird state where we are using an old firmware but showing the updates for a different stream.

src/fu-engine.c
hughsie: I don't _think_ so, why?

superm1: when you return the releases don't you want to show releases for all branches and disambiguate?

hughsie: I wanted something backwards compat, and where FwupdRelease flags are unset in the common case where there are no alternate branches. In this way we wouldn't show `Branch:  default` in thousands of places where it simply wasn't relevant. So, in the common case of no branches:

* FwupdRelease (branch=NULL) 1.2.3, no flags
* FwupdRelease (branch=NULL) 1.2.4, no flags

...and in the case where there are branches, and the FuDevice is on branch NULL (default):

* FwupdRelease (branch=NULL) 1.2.3, is-current-branch
* FwupdRelease (branch=NULL) 1.2.4, is-current-branch
* FwupdRelease (branch=crazy) 1.2.4, is-alternate-branch

and then we'd only show the branch in the UI if `is-current-branch` is set, **or** `is-alternate-branch` is set.

superm1: But the client can tell you what it supports.  So if the client supports all the right stuff then send all the details, if it doesn't then strip the alternate branch releases and throw them out.

superm1:
>I think we should leave fwupdtool as a debugging tool, not an advanced tool.

We bounce back and forth, especially since a notable big deployment of fwupd doesn't use the daemon and still needs a lot of the advanced features.

>We probably need a way to show what branches are available, or to show devices that can switch branch using a filter or something.

Agree

>It does get a bit more confusing when talking about the UX of changing branches. Would I do fwupdmgr update DEVICE_ID BRANCH_ID to go from official vendor 1.2.3 to random-vendor 0.0.1? An update downgrades? It would be weird.

As part of changing branches you can set expectation this about to happen thought at least with some prompts / messages.

>The core design issue here is that I've made it the device itself to say what branch it's on, rather than have a quirk or config file entry that selects the filtered branch; the latter of which doesn't work super well with our existing API. I think it has to be a atomic operation to switch branch to foo and install one of the releases on foo rather than having the weird state where we are using an old firmware but showing the updates for a different stream.

This was the part I wanted to mention but didn't get a chance.  If the device is telling you what branch it's on then the device has to have a way to advertise this information, which as I gather - it doesn't.

Here's my feeling of what makes sense command wise and what it should do:
* `get-releases`: Get all releases across all branches, and show what release is on what branch
* `get-branches`: Get all the branches supported, and show the one you're currently on
* `sidegrade`: Switch to a new release branch and version from that branch.

I do think that it's possible for devices to advertise what branch they're on though.  It's always a custom firmware relative to the stable, and that custom firmware could be detected and a device metadata flag set.  If you start that idea now with the first devices that support this, it should be able to scale for future designed firmware.

src/fu-engine.c
hughsie: I think I'm getting my device and release flags muddled up. I think I want a `is-alternate-branch` **or** `is-current-branch` as a single flag on the FwupdRelease and also a `has-multiple-branches` flag on the Fwupd**Device**.

hughsie:
> I do think that it's possible for devices to advertise what branch they're on though

I agree

>  If you start that idea now with the first devices that support this, it should be able to scale for future designed firmware

Concur, even if we have to fall back to a heuristic like we do with bcm57xx.

hughsie:
> sidegrade: Switch to a new release branch and version from that branch

Do we ever want to install anything other than the *newest* firmware from the new branch? It would be a lot simpler if we could make that assumption.

superm1:
> Concur, even if we have to fall back to a heuristic like we do with bcm57xx.

Even there, the firmware can be changed!  Add a new field to user data area to advertise a git branch or build tag or something.
By the time the plugin is merged it can be sorted.

>Do we ever want to install anything other than the newest firmware from the new branch? It would be a lot simpler if we could make that assumption.

idk if it actually matters.  you can always downgrade later after you switch branches I guess

superm1:
Something else comes to mind with branches and firmware versions. If hypothetically you're on 1.0.0 for standard firmware and 2.0.0 for the alternate branch but using an old daemon that doesn't understand branches you could get into a situation that it's migrating you on branches. So I guess that means Xml for the alternate branch firmware needs to advertise a minimum fwupd version requirement.

hughsie:
> So I guess that means Xml for the alternate branch firmware needs to advertise a minimum fwupd version requirement.

I've added a check for that on the LVFS website version of this: https://gitlab.com/fwupd/lvfs-website/-/merge_requests/805

>  Add a new field to user data area to advertise a git branch or build tag or something

@meklort -- do you think you could add another string-from-offset (like the version number you recently implemented) which points to a branch name like `oss-firmware` or `meklort` or whatever you want it to be named?

libfwupdplugin/fu-plugin.c
```diff
@@ -1052,6 +1052,32 @@ fu_plugin_device_write_firmware (FuPlugin *self, FuDevice *device,
...
+		fn = g_strdup_printf ("%s.bin", fu_device_get_version (device));
+		path = g_build_filename (localstatedir,
+					 "backup",
+					 fu_device_get_id (device),
+					 fu_device_get_serial (device) != NULL ?
+						fu_device_get_serial (device) :
+						"default",
+					 fn, NULL);
+		if (!fu_common_set_contents_bytes (path, fw_old, error))
+			return FALSE;
```
superm1, line 1078: One thought - what do you think about making a `latest.bin` symlink?  I guess it depends on how this backup would actually be used.

libfwupd/fwupd-enums.h
```diff
@@ -124,6 +126,8 @@ typedef enum {
  * @FWUPD_DEVICE_FLAG_NO_GUID_MATCHING:		Force an explicit ID match when adding devices to the device list
  * @FWUPD_DEVICE_FLAG_UPDATABLE_HIDDEN:		Device is updatable but should not be called by the client
  * @FWUPD_DEVICE_FLAG_SKIPS_RESTART:		Device relies upon activation or power cycle to load firmware
+ * @FWUPD_DEVICE_FLAG_HAS_MULTIPLE_BRANCHES:	Device has multiple branches of firmware
```
superm1, line 129: this description should explain what a branch actually is.  IE "Device supports switching to a different stream of firmware"

src/fu-util.c
```diff
@@ -1870,6 +1870,146 @@ fu_util_reinstall (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	g_autofree gchar *desc_plain = NULL;
+
+	/* get formatted text */
+	desc_markup = fwupd_release_get_description (rel);
+	if (desc_markup == NULL)
+		return TRUE;
+	desc_plain = fu_util_convert_description (desc_markup, error);
+	if (desc_plain == NULL)
+		return FALSE;
+
```
superm1, line 1886: I think it would be good to have a boilerplate warning here like we do for the lvfs-testing remote

src/fu-util.c
```diff
@@ -2736,6 +2876,12 @@ main (int argc, char *argv[])
 		     /* TRANSLATORS: command description */
 		     _("Reinstall current firmware on the device."),
 		     fu_util_reinstall);
+	fu_util_cmd_array_add (cmd_array,
+		     "switch-branch",
```
superm1, line 2880: aww you didn't like my command name `sidegrade`?  I thought it captured the possibility for firmware version schema being entirely different.

meklort:
> @meklort -- do you think you could add another string-from-offset (like the version number you recently implemented) which points to a branch name like `oss-firmware` or `meklort` or whatever you want it to be named?

Yes, I will add this. (I'll put oss-firmware there, similar to the current method as you mentioned).

It looks the following needs to happen:
1. I'll add <branch>oss-firmware</branch> to the xml file
1. I'll add oss-firmware to the firmware itself. This should be in a location where the string will never be found in the proprietary firmware (this should be anywhere, since that string does not exist there.
1. The fwupd plugin will need to be updated to read the fixed location and (preferably) to verify that it's set before writing the updated firmware to nvm.

libfwupdplugin/fu-plugin.c
hughsie: I think it's a bit confusing tbh. Latest compared to what? Latest downgrade? Latest recovery?

libfwupd/fwupd-enums.h
hughsie: fixed.

src/fu-util.c
hughsie: I didn't like `sidegrade` for three reasons:

 * It didn't include the word branch, and that's what we've used everywhere else
 * It's not sideloading a firmware from another repo like you would on a phone
 * It's not really understandable if your primary language isnt' `en_*`

libfwupdplugin/fu-plugin.c
superm1: Most recent?  I want to be cognizant that version schema might be different between branches so latest is pretty confusing.  Maybe also need to include the branch in the path instead ?

src/fu-util.c
hughsie: What would it say? I don't think an alternate branch is always bad, for instance, what if System76 supported a coreboot firmware for their laptop and also a nonfree bios from AMI. They're both supported by the vendor, and both as valid as each other -- which is why I left the warning to be the `<description>` in the metainfo file; it can be changed as appropriate.

superm1: Hmm, it puts an implied requirement on the uploader of the alternate firmware to explain the implications for warranty or functionality in the description.  That seems a bit unfair to the original firmware creator.

libfwupdplugin/fu-plugin.c
hughsie: > version schema might be different between branches

Yes, you're right. We need both. Fixed.

src/fu-util.c
hughsie: > it puts an implied requirement on the uploader of the alternate firmware to explain the implications for warranty or functionality in the description

superm1: How about if you compare the vendor of the two different release branches and show such a boilerplate warning only if it's different?

hughsie: Plus, as we're only allowing specific OEMs on the LVFS access to the branch setting feature it's something we can pre-agree before allowing the user to upload the firmware.

hughsie: > compare the vendor of the two different release branches

That's not a bad idea at all. I suck at the prose stuff, so what about:

    This firmware was not supplied by the original hardware vendor
    Your hardware may be damaged using this firmware, and installing this
    release may void any warranty with the vendor.

superm1: Sounds good to me, but you can probably switch out `vendor` for `%s` and use the vendor's name since we have it.

hughsie: So how about:

    The firmware supplied by {Hughski Limited} is not from {Broadcom Inc}, the hardware vendor. Your hardware may be damaged using this firmware, and installing this release may void any warranty with {Broadcom Inc}.

hughsie: maybe `s/not from/not supplied by`

superm1: LGTM

hughsie: Fixed, thanks.

superm1 (approved):

PR #2421, opened 2020-09-27
Add plugin for goodix fingerprint sensors

hughsie:
Support updating firmware for Goodix MOC fingerprint sensors.


boger047:
It is strange that there may be some problem with my editer setting. I am sure that I submitted it after alignment, but it is not shown here ,I should modify an uncrustify cfg file to automatically format.

superm1:
> It is strange that there may be some problem with my editer setting. I am sure that I submitted it after alignment, but it is not shown here ,I should modify an uncrustify cfg file to automatically format. 

If you come up with a good config that matches style for the project please share it to contrib/ directory in another pull request. I made a few attempts to get it right, but in the end some things still needed manually fix-up.

hughsie:
I'm also wondering if I should just adopt a more standardized whitespace style on the codebase, eg plain lkml or something.

superm1:
> I'm also wondering if I should just adopt a more standardized whitespace style on the codebase, eg plain lkml or something.

No complaints here.  It would be much easier to set up an IDE to auto-correct, CI to check it etc.

Issue #2429, opened 2020-09-29
fwupdmr does not work with pkttyagent


RPigott:
fwupdmgr does not appear to work with pkttyagent.

What I mean is for some reason when I don't have graphical polkit helper running, fwupdmgr will exit with "Failed to obtain auth".

This will demonstrate what I mean:

```
$ systemctl enable fwupd-refresh.timer 
==== AUTHENTICATING FOR org.freedesktop.systemd1.manage-unit-files ====
Authentication is required to manage system service or unit files.
Authenticating as: ronan
Password: 
==== AUTHENTICATION COMPLETE ====
Created symlink /etc/systemd/system/timers.target.wants/fwupd-refresh.timer → /usr/lib/systemd/system/fwupd-refresh.timer.

$ fwupdmgr get-updates 
[...]
Devices that have been updated successfully:
 • System Firmware (1.19.2 → 1.20.0)
[...]
Upload report now? (Requires internet connection):
[...]
Failed to obtain auth
```

In the above I enable the fwupd-refresh timer which demonstrates pkttyagent in action. Later I try to `fwupdmgr get-updates` which prompts me to consider uploading a firmware report, but when I try to upload one, the password prompt never appears and fwupdmgr fails to authorize. I am using fwupdmgr and systemctl from the same (local) session.

I would expect fwupdmgr to prompt for a password just like systemctl in this case. When I manually run a helper, like `polkit-gnome-authentication-agent-1` it works as expected. I haven't observed other software failing with only with pkttyagent, and based on my understanding of polkit I also don't really know how this is possible :/.

```
$ (. /etc/os-release; echo $NAME)
Arch Linux
$ pacman -Q fwupd polkit systemd polkit-gnome
fwupd 1.4.6-1
polkit 0.118-1
systemd 246.6-1
polkit-gnome 0.105-5
```

I have a little shell script to parse polkit policies for me, and it seems the allow_active policy directives are the same:

```
$ pklist org.freedesktop.fwupd.policy | grep -A4 modify-remote
org.freedesktop.fwupd.modify-remote
allow_any:      auth_admin
allow_inactive:	no
allow_active:   auth_admin_keep
$ pklist org.freedesktop.systemd1.policy | grep -A4 manage-unit-files
org.freedesktop.systemd1.manage-unit-files
allow_any:      auth_admin
allow_inactive: auth_admin
allow_active:   auth_admin_keep
```

superm1:
Hmm, I can reproduce the same thing.

[...]

PR #2430, opened 2020-09-29
Use pkttyagent to request user passwords if running without GUI

hughsie:
This change will allow it to use pkcon over remote-shells (ssh) or
to use it witout a running GUI desktop environment in the background.

Should fix https://github.com/fwupd/fwupd/issues/2429

### DIFF
--- a/src/fu-polkit-agent.c
+++ b/src/fu-polkit-agent.c
@@ -1,0 +1,237 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2011 Lennart Poettering <lennart@poettering.net>
+ * Copyright (C) 2012 Matthias Klumpp <matthias@tenstral.net>
+ * Copyright (C) 2015-2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include <stdio.h>
+#include <sys/types.h>
+#ifdef __linux__
+#include <sys/prctl.h>
+#endif
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <dirent.h>
+#include <signal.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <inttypes.h>
+#include <sys/poll.h>
+#include <sys/wait.h>
+#include <glib.h>
+
+#include "fu-common.h"
+#include "fu-polkit-agent.h"
+
+static pid_t agent_pid = 0;
+
+static int
+fork_agent (pid_t *pid, const char *path, ...)
+{
+	char **l;
+	gboolean stderr_is_tty;
+	gboolean stdout_is_tty;
+	int fd;
+	pid_t n_agent_pid;
+	pid_t parent_pid;
+	unsigned n, i;
+	va_list ap;
+
+	g_return_val_if_fail (pid != 0, 0);
+	g_assert (path);
+
+	parent_pid = getpid ();
+
+	/* spawns a temporary TTY agent, making sure it goes away when
+	 * we go away */
+	n_agent_pid = fork ();
+	if (n_agent_pid < 0)
+		return -errno;
+
+	if (n_agent_pid != 0) {
+		*pid = n_agent_pid;
+		return 0;
+	}
+
+#ifdef __linux__
+	/* make sure the agent goes away when the parent dies */
+	if (prctl (PR_SET_PDEATHSIG, SIGTERM) < 0)
+		_exit (EXIT_FAILURE);
+#endif
+	/* check whether our parent died before we were able
+	 * to set the death signal */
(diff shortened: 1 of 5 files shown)


src/fu-util.c
```diff
@@ -2845,6 +2846,9 @@ main (int argc, char *argv[])
 	if (no_history)
 		priv->flags |= FWUPD_INSTALL_FLAG_NO_HISTORY;
 
+	/* start polkit tty agent to listen for password requests */
+	fu_polkit_agent_open ();
```
superm1, line 2850: does this need to run on every run of the app?

superm1: My thought being for example it wouldn't run in non-interactive mode

superm1, line 2850: the return code seems to be int, shouldn't GError be passed in and the return code be boolean instead (maybe with a local error)?

hughsie: Yes, it gets killed after the command quits. It's exactly the same logic as pkcon has been using for ~ 5 years.

hughsie: Yes, good idea.

src/fu-polkit-agent.c
```diff
@@ -0,0 +1,219 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2011 Lennart Poettering <lennart@poettering.net>
+ * Copyright (C) 2012 Matthias Klumpp <matthias@tenstral.net>
```
superm1, line 4: I guess this was lifted from systemd?  It feels like systemd style code.  If it was new code I guess a few things would have been differently like not using variable arguments.

src/fu-polkit-agent.c
```diff
@@ -0,0 +1,219 @@
...
+#include <fcntl.h>
+#include <errno.h>
+#include <inttypes.h>
+#include <sys/poll.h>
+#include <sys/wait.h>
+#include <glib.h>
+
+#include "fu-polkit-agent.h"
+
+#define POLKIT_TTY_AGENT_BINARY_PATH "/usr/bin/pkttyagent"
```
superm1, line 29: Probably shouldn't hardcode this path, we have infrastructure to lookup paths as configured by distros.

src/fu-polkit-agent.c
hughsie: > I guess this was lifted from systemd

Yes, a long time ago (LGPLv2+ also) -- we could clean up some stuff, but I'm hesitant to change the low level UNIX bits as in all honesty I don't understand all the nuances mentioned.

src/fu-polkit-agent.c
hughsie: Fixed.

src/fu-util.c
hughsie: Done,a few things cleaned up.

superm1:
So it's a bit awkward using this since there is a fighting for stdout.  

```
$ fwupdmgr enable-remote lvfs-testing
╔══════════════════════════════════════════════════════════════════════════════╗
║ The LVFS is a free service that operates as an independent legal entity and  ║
║ has no connection with Ubuntu. Your distributor may not have verified any    ║
║ of the firmware updates for compatibility with your system or connected      ║
║ devices. All firmware is provided only by the original equipment             ║
║ manufacturer.                                                                ║
║                                                                              ║
║ This remote contains firmware which is not embargoed, but is still being     ║
║ tested by the hardware vendor. You should ensure you have a way to manually  ║
║ downgrade the firmware if the firmware update fails.                         ║
║                                                                              ║
║ Enabling this functionality is done at your own risk, which means you have   ║
║ to contact your original equipment manufacturer regarding any problems       ║
║ caused by these updates. Only problems with the update process itself        ║
║ should be filed at https://bugs.launchpad.net/ubuntu/.                       ║
║                                                                              ║
╚══════════════════════════════════════════════════════════════════════════════╝

Agree and enable the remote? [Y|n]: y
Authenticating…          [-                                      ]==== AUTHENTICATING FOR org.freedesktop.fwupd.modify-remote ===
Authentication is required to modify a configured remote used for firmware updates
Authenticating as: supermario,,, (supermario)
Authenticating…          [                                     \ ]
Authenticating…          [                                      \]==== AUTHENTICATION COMPLETE ===
Successfully enabled remote
```

src/fu-util.c
```diff
@@ -2845,6 +2847,12 @@ main (int argc, char *argv[])
 	if (no_history)
 		priv->flags |= FWUPD_INSTALL_FLAG_NO_HISTORY;
 
+	/* start polkit tty agent to listen for password requests */
```
superm1, line 2850: don't you also want to check `is_interactive`?

hughsie: Yes! Fixed.

hughsie:
> fighting for stdout

How's that new version working? ^^

superm1 (approved):
Much better.  branch looks good to me now.

PR #2434, opened 2020-09-29
Fix get-updates and activate to use GUID in addition to device-id

djcampello:
Also merge fu_engine_get_device and fu_engine_get_device_by_id

Type of pull request:
- [ X ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2399,19 +2399,17 @@ fu_engine_get_plugins (FuEngine *self)
 }
 
 /**
- * fu_engine_get_device_by_id:
+ * fu_engine_get_device:
  * @self: A #FuEngine
- * @device_id: A string
- * @error: A #GError or NULL
+ * @device_id: A device ID
+ * @error: A #GError, or %NULL
  *
- * Returns the device from the engine or NULL if not found
+ * Gets a specific device.
  *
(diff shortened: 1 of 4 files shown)


src/fu-engine.c
```diff
@@ -2399,19 +2399,17 @@ fu_engine_get_plugins (FuEngine *self)
...
- *
- * Returns the device from the engine or NULL if not found
+ * @device_id: A device ID
+ * @error: A #GError, or %NULL
  *
- * Returns: (transfer full): a #FuDevice
+ * Gets a specific device.
  *
- * Since: 1.5.0
+ * Returns: (transfer full): a device, or %NULL if not found
```
superm1, line 2409: This should still have the clause mentioning when it was introduced.  IE:
` * Since: 1.5.0	`

superm1 (approved):
just a small nit, otherwise LGTM

src/fu-engine.c
djcampello: This function has been there from before.... I just eliminated the _by_id that was going to be introduced on 1.5.0

superm1: Oh I see... OK.
LGTM then

PR #2435, opened 2020-09-30
Support loading DMI data from DT systems

hughsie:
To do this with the existing codebase just fake the required SMBIOS structures.


libfwupdplugin/fu-smbios.c
```diff
@@ -64,12 +64,71 @@ typedef struct __attribute__((packed)) {
...
+	fu_smbios_convert_dt_string (self, FU_SMBIOS_STRUCTURE_TYPE_SYSTEM, 0x04,
+				     path, "vendor");
+
+	/* DMI:Family */
+	fu_smbios_convert_dt_string (self, FU_SMBIOS_STRUCTURE_TYPE_SYSTEM, 0x1a,
+				     path, "model-name");
+
+	/* DMI:ProductName */
+	fu_smbios_convert_dt_string (self, FU_SMBIOS_STRUCTURE_TYPE_SYSTEM, 0x05,
+				     path, "model");
```
superm1, line 115: on a DT system do these also populate `/sys/class/dmi/id`?  If so, maybe it's better to just fallback to that path in fu-smbios.

hughsie: On my qemu system there's no `/sys/class/dmi` whatsoever. @meklort is this populated on actual hardware please?

superm1: if nothing brings devicetree stuff like this to /sys/class/dmi on real hardware then yes agree this PR looks fine

meklort: > On my qemu system there's no `/sys/class/dmi` whatsoever. @meklort is this populated on actual hardware please?

Correct, this does not exist on the Talos II or Blackbird systems:

    meklort@talos2 ~/g/b/build ((e8e9e8df…))> ls /sys/class/dmi
    ls: cannot access '/sys/class/dmi': No such file or directory

PR #2436, opened 2020-09-30
fwupdtool/fwupdmgr: make return code different for get-updates with n…

superm1:
…o updates

This allows other tools to depend on it when checking for updates rather than
reading the output.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I like the concept, but needs CI fixes.

superm1:
> I like the concept, but needs CI fixes.

Hmmm, actually it looks like it's a bug that was there already.  It hasn't been running the CI tests after get-updates for a while now!  The fwupdmgr.sh was _expecting_ return code 2, and it's been "erroring" with return code 0....

So there is some sort of issue with verification.  I can repo it locally.  Does it stand out to you?
```
failed to verify Integrated Webcam™: No device checksums for 1.2.4
```

superm1:
I think it might be broken as far back as https://github.com/fwupd/fwupd/commit/af140735220cb769f398cdf346af23a96c4b7a51

The content read in is
```
$ cat fakedevice124.bin
fakedevice124 -n
```

And that probably won't parse properly by `fu_plugin_test_get_version`

superm1:
That explains the version error in the log (`fu_plugin_test_get_version` returns NULL).  I don't think I can fix it, it requires re-creating and detached re-signing those files, so @hughsie think you will have to.  Furthermore, it seems nothing ever enrolls the device checksums again after the update is installed in the first place.  I'd be tempted to just call fu_plugin_verify from the update vfunc, but it needs the right version to be parsed..

hughsie:
> it requires re-creating and detached re-signing those files

Done, feel free to squash it into your commit if it actually works.

superm1:
Whew, got it working.  Let me know if you want that squashed or not.

data/installed-tests/fwupdmgr.sh
```diff
@@ -43,7 +43,7 @@ rc=$?; if [[ $rc != 0 ]]; then error $rc; fi
 
 # ---
 echo "Installing test firmware..."
-fwupdmgr install ${dirname}/fakedevice124.cab
+fwupdmgr update $device -y
```
hughsie, line 46: is this testing the same thing? where does the metadata come from?

superm1: It's pulling from local `fwupd-tests` metadata.  It should be same file.  It let me run it in tree without having to manually fixup paths.

hughsie: Okay, sounds good then. I guess we're testing the checksum verification rather than installing a naked file, although it does seems to shame to loose the local-file-testing too.

hughsie (approved):

PR #2441, opened 2020-10-02
Add a device flag for when a detach-less firmware dump is possible

hughsie:
This would indicate to the use that the device is not going to 'go away' for a
few seconds and is better UX than just unconditionally detaching when it's not
actually required.

### DIFF
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -201,6 +201,8 @@ fwupd_device_flag_to_string (FwupdDeviceFlags device_flag)
 		return "has-multiple-branches";
 	if (device_flag == FWUPD_DEVICE_FLAG_BACKUP_BEFORE_INSTALL)
 		return "backup-before-install";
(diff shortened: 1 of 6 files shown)


superm1:
Why can't you use `FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE`?

hughsie:
> Why can't you use FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE

I guess I could, but it seems a bit weird; USABLE_DURING_UPDATE tells the user that "you can install this update without interruption" and FWUPD_DEVICE_FLAG_CAN_DUMP_RUNTIME tells the user "you can dump an image without interruption". Maybe the name is bad; what about `FWUPD_DEVICE_FLAG_USABLE_DURING_DUMP`?

hughsie:
Lets just implement BCM57xx in a simpler way, i.e. split up the device objects and not implement ->detach() in the runtime case.

PR #2442, opened 2020-10-02
trivial: Allow fwuptool to handle firmware that switches branch

hughsie:
This allows testing the bcm57xx firmware with fwupdtool.

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2711,7 +2711,8 @@ main (int argc, char *argv[])
 		/* set our implemented feature set */
 		fu_engine_request_set_feature_flags (priv->request,
 						     FWUPD_FEATURE_FLAG_DETACH_ACTION |
-						     FWUPD_FEATURE_FLAG_UPDATE_ACTION);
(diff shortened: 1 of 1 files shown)


superm1 (commented):
If it's offering that flag doesn't it also need switch-branch command added?

hughsie:
> it also need switch-branch command added

It can't interact with the metadata, and so the only thing it can do is install a 'naked' downloaded cab. We don't have the `FwupdRelease` that we can inspect and do the same kind of checks.

hughsie:
> and do the same kind of checks

Maybe we do want a --allow-switch-branch and a FWUPD_INSTALL_FLAG_ALLOW_SWITCH_BRANCH for exactly this reason...

superm1:
> It can't interact with the metadata, and so the only thing it can do is install a 'naked' downloaded cab. We don't have the FwupdRelease that we can inspect and do the same kind of checks.

What do you mean can't interact?  We have `fwupdtool refresh` and `fwupdtool update`

hughsie:
Ya, this doesn't work. Thanks for the sanity check.

PR #2447, opened 2020-10-05
Add a --allow-branch-switch to fwupdmgr

hughsie:
This allows us to turn off the tests like version format differences and
checking for downgrades.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2397,11 +2397,15 @@ fu_util_esp_list (FuUtilPrivate *priv, gchar **values, GError **error)
 int
 main (int argc, char *argv[])
 {
+	gboolean allow_branch_switch = FALSE;
 	gboolean allow_older = FALSE;
 	gboolean allow_reinstall = FALSE;
 	gboolean force = FALSE;
 	gboolean ret;
 	gboolean version = FALSE;
+	gboolean ignore_checksum = FALSE;
+	gboolean ignore_power = FALSE;
+	gboolean ignore_vid_pid = FALSE;
 	gboolean interactive = isatty (fileno (stdout)) != 0;
 	g_auto(GStrv) plugin_glob = NULL;
 	g_autoptr(FuUtilPrivate) priv = g_new0 (FuUtilPrivate, 1);
@@ -2419,9 +2423,21 @@ main (int argc, char *argv[])
 		{ "allow-older", '\0', 0, G_OPTION_ARG_NONE, &allow_older,
 			/* TRANSLATORS: command line option */
 			_("Allow downgrading firmware versions"), NULL },
+		{ "allow-branch-switch", '\0', 0, G_OPTION_ARG_NONE, &allow_branch_switch,
+			/* TRANSLATORS: command line option */
+			_("Allow switching firmware branch"), NULL },
 		{ "force", '\0', 0, G_OPTION_ARG_NONE, &force,
 			/* TRANSLATORS: command line option */
-			_("Override plugin warning"), NULL },
+			_("Force the action by relaxing some runtime checks"), NULL },
+		{ "ignore-checksum", '\0', 0, G_OPTION_ARG_NONE, &ignore_checksum,
+			/* TRANSLATORS: command line option */
+			_("Ignore firmware checksum failures"), NULL },
+		{ "ignore-vid-pid", '\0', 0, G_OPTION_ARG_NONE, &ignore_vid_pid,
+			/* TRANSLATORS: command line option */
+			_("Ignore firmware hardware mismatch failures"), NULL },
+		{ "ignore-power", '\0', 0, G_OPTION_ARG_NONE, &ignore_power,
+			/* TRANSLATORS: command line option */
+			_("Ignore requirement of external power source"), NULL },
 		{ "no-reboot-check", '\0', 0, G_OPTION_ARG_NONE, &priv->no_reboot_check,
 			/* TRANSLATORS: command line option */
 			_("Do not check for reboot after update"), NULL },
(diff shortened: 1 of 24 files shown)


superm1:
So we have both a `switch-branch` command and a `--allow-switch-branch` option?  It feels like it should be an either/or.

hughsie:
> and a --allow-switch-branch option

From a user point of view, the only time they'd need to use `--allow-switch-branch option` is when installing a local .cab file. It also lets us add the functionality into fwuptool for testing in another commit.

superm1:
> From a user point of view, the only time they'd need to use --allow-switch-branch option is when installing a local .cab file. It also lets us add the functionality into fwuptool for testing in another commit.

But so then maybe `switch-branch` should not be a command is my point.  It should be `install-release` and then --allow-switch-branch would show all the different releases including multiple branches.

hughsie:
> But so then maybe switch-branch should not be a command is my point

Not so convinced; from a 40,000ft view the CLI and API between downgrading and switching a release should be the same; they're both actions that have security implications and don't fit in the "JFDI update" flow. You could argue we don't need downgrade as a CLI command, as you could just have `install-release` and get the user to select the one that has the same branch as existing and that doesn’t have is_upgrade or doesn't have is_downgrade. The point is the user explicitly wants to "switch the branch" to a different stream rather than specifically install a release on that branch.

That said, maybe `install-release [DEVICE_ID] [VERSION] [BRANCH] [REMOTE]` is a useful thing to have, in addition. Just not on this PR :)

superm1:
But so then how is --allow-switch-branch different than force?

hughsie:
> how is --allow-switch-branch different than force

`--force is so much more powerful` -- insomuch that it will ignore the verfmt (like --allow-branch-switch) but **also** checks like "is the binary blob the right size" and checking CRCs and checksums are correct. I'm actually wondering if we want to allow `--force` from the daemon at all, and leave it as something only supported in fwpdtool.

superm1:
> I'm actually wondering if we want to allow --force from the daemon at all, and leave it as something only supported in fwpdtool.

We've always left `--force` as the big hammer to overrride any specific checks.  It's useful even in `fwupdmgr`, for example if you want to disable upower plugin checks.

Maybe what would make sense is to instead split up force into a variety of individual types of force?

hughsie:
> split up force into a variety of individual types of force

Yes, I think so too. Will ponder.

hughsie:
@superm1 what do you think about ^? that means that you can only ignore failed CRC and VID/PID mismatch from fwupdtool, but I think that's actually what we want. `--force` then becomes a _gentle nudge_ rather than a _potentially-hardware-destroying_ action.

src/fu-tool.c
```diff
@@ -2422,6 +2424,12 @@ main (int argc, char *argv[])
 		{ "force", '\0', 0, G_OPTION_ARG_NONE, &force,
 			/* TRANSLATORS: command line option */
 			_("Override plugin warning"), NULL },
+		{ "ignore-checksum", '\0', 0, G_OPTION_ARG_NONE, &ignore_checksum,
+			/* TRANSLATORS: command line option */
+			_("Ignore firmware checksum failures"), NULL },
+		{ "ignore-vid-pid", '\0', 0, G_OPTION_ARG_NONE, &ignore_vid_pid,
```
superm1, line 2430: these should be added to bash completion

superm1:
> @superm1 what do you think about ^? that means that you can only ignore failed CRC and VID/PID mismatch from fwupdtool, but I think that's actually what we want. `--force` then becomes a _gentle nudge_ rather than a _potentially-hardware-destroying_ action.

I think yeah that's the right direction.  We probably should completely remove `--force` though, and everything it did should be it's own thing, and some things only doable from `fwupdtool`.  IE also have `--ignore-power`

src/fu-tool.c
hughsie: I think there should be some kind of swear-jar for every time I forget to add these. My apologies.

hughsie:
>  We probably should completely remove --force though

Isn't that part of our CLI, API and ABI? I'd be fine slightly redefining `--force` to be a "soft nudge" that's not going to cause damage. I'm not super-keen on removing the API or emulating it.

superm1:
> I'd be fine slightly redefining --force to be a "soft nudge" that's not going to cause damage

Well I was thinking marking it as adding `--force` shows a message `--force` is removed, please see these other commands for the same things.

superm1:
> to be a "soft nudge

Or maybe rename it `-ignore-version-format` but let `--force` be a synonym for it?

hughsie:
I added `FWUPD_INSTALL_FLAG_IGNORE_POWER` but I'm not sure we want/need a `_IGNORE_FLASHES_LEFT` and `IGNORE_UEFI_DBX_ESP` and `_IGNORE_VERSION_FORMAT`. The latter is really just a _the uploader screwed something up_ and now we can rewrite the metainfo.xml file on the LVFS it's easiest just to fix the problem IMHO.

src/fu-util.c
```diff
@@ -3003,10 +3012,16 @@ main (int argc, char *argv[])
 		priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_REINSTALL;
 	if (allow_older)
 		priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_OLDER;
-	if (force)
+	if (allow_branch_switch)
+		priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_BRANCH_SWITCH;
+	if (force) {
 		priv->flags |= FWUPD_INSTALL_FLAG_FORCE;
+		priv->flags |= FWUPD_INSTALL_FLAG_IGNORE_POWER;
```
superm1, line 3019: this same behavior adjustment seems missing on `fu-tool.c`

hughsie: Indeed, good catch.

hughsie: Done.

src/fu-util.c
```diff
@@ -2011,6 +2011,7 @@ fu_util_switch_branch (FuUtilPrivate *priv, gchar **values, GError **error)
 	g_signal_connect (priv->client, "device-changed",
 			  G_CALLBACK (fu_util_update_device_changed_cb), priv);
 	priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_REINSTALL;
+	priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_BRANCH_SWITCH;
```
superm1, line 2014: I guess it's worth asking - why not also have this same argument on `fwupdtool`?  Installing a CAB could encounter the same situation.

src/fu-util.c
```diff
@@ -2683,6 +2686,9 @@ main (int argc, char *argv[])
 		{ "allow-older", '\0', 0, G_OPTION_ARG_NONE, &allow_older,
 			/* TRANSLATORS: command line option */
 			_("Allow downgrading firmware versions"), NULL },
+		{ "allow-branch-switch", '\0', 0, G_OPTION_ARG_NONE, &allow_branch_switch,
+			/* TRANSLATORS: command line option */
+			_("Allow switching firmware branch"), NULL },
 		{ "force", '\0', 0, G_OPTION_ARG_NONE, &force,
 			/* TRANSLATORS: command line option */
 			_("Override warnings and force the action"), NULL },
```
superm1, line 2694: This should probably change to reflect it's been gimped.

hughsie: I changed this to `Force the action by relaxing some runtime checks`

src/fu-util.c
hughsie: You're completely correct. Fixed.

superm1 (approved):

PR #2449, opened 2020-10-06
trivial: uefi: when registering devices take their version format

superm1:
Setting the default to number causes problems when the release
is set to 'quad'.

This regression was introduced by b097603800c03f6f99e2d559f60b4891d5fa0a88 in 1_4_X and master :sob: 
I'm not sure what we should set the version requirement on the CAB as a result since 1_3_X worked right.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-uefi-device.c
+++ b/plugins/uefi/fu-uefi-device.c
@@ -721,7 +721,6 @@ static void
 fu_uefi_device_init (FuUefiDevice *self)
 {
 	fu_device_set_protocol (FU_DEVICE (self), "org.uefi.capsule");
-	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_NUMBER);
 }
 
 static void
@@ -764,6 +763,9 @@ fu_uefi_device_new_from_entry (const gchar *entry_path, GError **error)
 
 	/* create object */
 	self = g_object_new (FU_TYPE_UEFI_DEVICE, NULL);
+
+	/* assume a uint64_t unless told otherwise by a quirk entry or metadata */
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_NUMBER);
 
 	/* read values from sysfs */
 	fw_class_fn = g_build_filename (entry_path, "fw_class", NULL);
@@ -824,5 +826,6 @@ fu_uefi_device_new_from_guid (const gchar *guid)
 	FuUefiDevice *self;
 	self = g_object_new (FU_TYPE_UEFI_DEVICE, NULL);
 	self->fw_class = g_strdup (guid);
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_NUMBER);
 	return self;
 }


superm1:
I guess can set the firmware as 'quad' on LVFS and fallback to 'number'.  This seems to work with older versions too then.

hughsie:
> I guess can set the firmware as 'quad' on LVFS and fallback to 'number

I think being explicit in the firmware is the safest thing.

superm1:
Hmm it seems LVFS is discarding the second defined version format now though.

hughsie:
> Hmm it seems LVFS is discarding the second defined version format now though.

Ohh? What firmware you setting with what verfmt? I can look now.

superm1:
"TPM firmware 7.2.0.2", just uploaded it yesterday to experiment. It had two version formats set and when LVFS rewrote the version format it was just one.

hughsie:
> It had two version formats set

superm1:
It had this:

```
 <custom>
  <value key="LVFS::UpdateProtocol">org.uefi.capsule</value>
  <value key="LVFS::VersionFormat">quad</value>
  <value key="LVFS::VersionFormat">number</value>
 </custom>
```

With the expectation to work around this bug.

hughsie:
> It had this

So the expectation is that this would be rewritten to `quad` -- as that was added much before `number`. Can we go back a bit, can you describe what's happening inside fwupd? e.g. device created in plugin foo, registered in plugin bar, verfmt set from metadata etc? Thanks.

superm1:
> Can we go back a bit, can you describe what's happening inside fwupd

A `FuDevice` is created from plugin `dell`.  This has version format of quad (always has) up until https://github.com/fwupd/fwupd/commit/b097603800c03f6f99e2d559f60b4891d5fa0a88.
When registered, `fu_plugin_device_registered` is called which calls `fu_plugin_uefi_register_proxy_device`, which calls `fu_uefi_device_new_from_dev`.

`fu_uefi_device_new_from_dev` creates a new device which calls `fu_uefi_device_init` which sets the version format to `number` by default.  `fu_device_incorporate` is called to copy everything from the donor device from the registration, but since the version format was already set to `number` by default it never gets updated.

Now in terms of metadata, I didn't check specifically with a remote "enabled" yet, I was checking specifically that the CAB would work if "hand" downloaded.

hughsie:
Couldn't the TPM device set `FWUPD_DEVICE_FLAG_MD_SET_VERFMT` and then we get the "correct" version format from the metadata?

superm1:
>Couldn't the TPM device set FWUPD_DEVICE_FLAG_MD_SET_VERFMT and then we get the "correct" version format from the metadata?

So the metadata (and CAB) would both set it to quad, right?  I guess that would work, but I'd like to make sure the hand install case still "works" too.  So I think LVFS still needs to have the fallback stuff working correctly.

hughsie:
> but I'd like to make sure the hand install

Assuming you're installing the LVFS-signed .cab file with the `quad` (only) version format? I think I see what you mean, the existing device is going to be `number` and the cab file is going to be `quad` and you're going to get a verfmt mismatch.

I think your patch is on the right lines, but I'd rather only set the verfmt when we know it's correct rather than change it at runtime and cause other side effects. What about something like 58108a196c0846d507eb5d218eb81b54bf6e6d7d -- if that works I'd be fine for you to use that on your branch, check it works in CI (in this PR) and then commit.

superm1:
>Assuming you're installing the LVFS-signed .cab file with the quad (only) version format? I think I see what you mean, the existing device is going to be number and the cab file is going to be quad and you're going to get a verfmt mismatch.

So in that regard, what do you think about letting LVFS keep multiple version formats if they're applied in the uploaded XML?  I guess that or some heuristics of things that need multiple formats.

>I think your patch is on the right lines, but I'd rather only set the verfmt when we know it's correct rather than change it at runtime and cause other side effects. What about something like 58108a1 -- if that works I'd be fine for you to use that on your branch, check it works in CI (in this PR) and then commit.

Yeah, that seems like the better solution than the one I proposed.

hughsie:
> what do you think about letting LVFS keep multiple version formats if they're applied in the uploaded XML

Not so much of a fan; I think we should do the backwards compat trick where we add the fallback last so it gets used by legacy clients, but I think layering on any kind of _other_ logic to the ordering of the verfmts on upload is probably trying to be too clever.

Issue #2360, opened 2020-09-08
fwupdtool refresh : failed to verify data, expected abe4258d1dbbc02ad732dc1feefeb266ae872e6c


joakim-tjernlund:
**Describe the bug**
fwupdtool refresh
Loading…                 [-                                      ]ERROR:esys:src/tss2-esys/esys_context.c:69:Esys_Initialize() Initialize default tcti. ErrorCode (0x000a000a) 
Loading…                 [***************************************]
checksum failure: failed to verify data, expected abe4258d1dbbc02ad732dc1feefeb266ae872e6c

**Steps to Reproduce**
See above 

**fwupd version information**
master from today

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
emerge sys-apps/fwupd

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

20L7004MMX
│
├─Thunderbolt host controller:
│     Device ID:          ab7d5105abea00e669b468df7882afc0b34e133e
│     Summary:            Unmatched performance for high-speed I/O
│     Current version:    20.00
│     Vendor:             Lenovo (TBT:0x0109)
│     GUIDs:              3849839e-d8b2-5941-9e69-5acc397d3ed3 ← THUNDERBOLT\VEN_0109&DEV_1708&REV_00
│                         7e3190a2-d5a5-5a9e-b582-494b47f7288c ← THUNDERBOLT\VEN_0109&DEV_1708
│                         061f2847-061b-5cfb-9c39-3dbfe2b49d69 ← TBT-01091708-native
│                         65a3d897-4a18-5d27-83d8-eb8a231b7aa4 ← TBT-01091708-native-controller0-0
│     Device Flags:       • Internal device
│                         • Updatable
│                         • Requires AC power
│                         • Supported on remote server
│                         • Device stages updates
│   
├─USB3.1 Hub:
│     Device ID:          d68b5d74453efa336214e42d39ec96008a3fed64
│     Summary:            USB 3.x Hub
│     Current version:    10.74
│     Vendor:             VIA Labs, Inc. (USB:0x17EF)
│     Install Duration:   15 seconds
│     GUIDs:              e2c96871-8be3-530e-a796-5652cee64eeb
│                         351195ca-bd3e-5a03-911c-0be6686d91dd ← USB\VID_17EF&PID_3070&REV_0A74
│                         2c08e9d9-1f25-5564-b4db-49867dd2c593 ← USB\VID_17EF&PID_3070
[...]

PR #2452, opened 2020-10-07
libfwupd: Actually use the JCat file to select the metadata file

hughsie:
This recent regression was caused when porting libfwupd to the async model.

Fixes https://github.com/fwupd/fwupd/issues/2360

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -2802,6 +2802,10 @@ fwupd_client_refresh_remote_signature_cb (GObject *source,
 		return;
 	}
 	data->signature = g_steal_pointer (&bytes);
+	if (!fwupd_remote_load_signature_bytes (data->remote, data->signature, &error)) {
+		g_task_return_error (task, g_steal_pointer (&error));
+		return;
+	}
 
 	/* download metadata */
 	fwupd_client_download_bytes_async (self,
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2168,25 +2168,6 @@ fu_util_refresh_remote (FuUtilPrivate *priv, FwupdRemote *remote, GError **error
 	g_autoptr(GBytes) bytes_raw = NULL;
 	g_autoptr(GBytes) bytes_sig = NULL;
 
-	/* payload */
-	metadata_uri = fwupd_remote_get_metadata_uri (remote);
-	if (metadata_uri == NULL) {
-		g_set_error (error,
-			     FWUPD_ERROR,
-			     FWUPD_ERROR_NOTHING_TO_DO,
-			     "no metadata URI available for %s",
-			     fwupd_remote_get_id (remote));
-		return FALSE;
-	}
-	fn_raw = fu_util_get_user_cache_path (metadata_uri);
-	if (!fu_common_mkdir_parent (fn_raw, error))
-		return FALSE;
-	if (!fu_util_download_out_of_process (metadata_uri, fn_raw, error))
-		return FALSE;
-	bytes_raw = fu_common_get_contents_bytes (fn_raw, error);
-	if (bytes_raw == NULL)
-		return FALSE;
-
 	/* signature */
 	metadata_uri = fwupd_remote_get_metadata_uri_sig (remote);
 	if (metadata_uri == NULL) {
@@ -2198,10 +2179,31 @@ fu_util_refresh_remote (FuUtilPrivate *priv, FwupdRemote *remote, GError **error
 		return FALSE;
 	}
 	fn_sig = fu_util_get_user_cache_path (metadata_uri);
+	if (!fu_common_mkdir_parent (fn_sig, error))
+		return FALSE;
 	if (!fu_util_download_out_of_process (metadata_uri, fn_sig, error))
 		return FALSE;
 	bytes_sig = fu_common_get_contents_bytes (fn_sig, error);
 	if (bytes_sig == NULL)
+		return FALSE;
+	if (!fwupd_remote_load_signature_bytes (remote, bytes_sig, error))
+		return FALSE;
+
+	/* payload */
+	metadata_uri = fwupd_remote_get_metadata_uri (remote);
+	if (metadata_uri == NULL) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOTHING_TO_DO,
+			     "no metadata URI available for %s",
+			     fwupd_remote_get_id (remote));
+		return FALSE;
+	}
+	fn_raw = fu_util_get_user_cache_path (metadata_uri);
+	if (!fu_util_download_out_of_process (metadata_uri, fn_raw, error))
+		return FALSE;
+	bytes_raw = fu_common_get_contents_bytes (fn_raw, error);
+	if (bytes_raw == NULL)
 		return FALSE;
 
 	/* send to daemon */


hughsie:
@joakim-tjernlund can you verify this fixes the issue for you please.

joakim-tjernlund:
Above patch does not fix the problem fully for me, sorry.
fwupdmgr refresh works but fwupdtool -v refresh does not:
```
1:01:12:0275 FuCommon             running 'wget https://cdn.fwupd.org/downloads/firmware-testing.xml.gz -O /root/.cache/fwupdmgr/firmware-testing.xml.gz'
11:01:12:0275 GLib                 posix_spawn avoided (fd close requested) (child_setup specified) 
11:01:12:0314 FuCommon             reading /root/.cache/fwupdmgr/firmware-testing.xml.gz with 44773 bytes
11:01:12:0314 FuCommon             running 'wget https://cdn.fwupd.org/downloads/firmware-testing.xml.gz.jcat -O /root/.cache/fwupdmgr/firmware-testing.xml.gz.jcat'
11:01:12:0314 GLib                 posix_spawn avoided (fd close requested) (child_setup specified) 
11:01:12:0351 FuCommon             reading /root/.cache/fwupdmgr/firmware-testing.xml.gz.jcat with 2236 bytes
11:01:12:0351 FuMain               updating lvfs-testing
checksum failure: failed to verify data, expected 0cbcf8ba6cf9004a8bcafb7e95406dc8ab95501c
11:01:12:0356 FuPlugin             destroy(linux_swap)
11:01:12:0357 FuPlugin             destroy(upower)

**Comment by joakim-tjernlund** [comment]
There is this unknown printout, not sure if it is relevant:
```
fwupdmgr refresh --force
Updating lvfs-testing
Unknown                  [***************************************]
Updating lvfs
Successfully downloaded new metadata: 5 local devices supported

hughsie:
> Above patch does not fix the problem fully for me, sorry.

Are you sure you applied the patch and installed/linked the new libfwupd? You should see something like:

    (fwupdmgr:99245): Fwupd-DEBUG: 10:51:36.656: changing metadata URI from https://cdn.fwupd.org/downloads/firmware.xml.gz to https://cdn.fwupd.org/downloads/firmware-00954-stable.xml.gz

...in the --verbose logs.

joakim-tjernlund:
> > Above patch does not fix the problem fully for me, sorry.
> 
> Are you sure you applied the patch and installed/linked the new libfwupd? You should see something like:
> 
> ```
> (fwupdmgr:99245): Fwupd-DEBUG: 10:51:36.656: changing metadata URI from https://cdn.fwupd.org/downloads/firmware.xml.gz to https://cdn.fwupd.org/downloads/firmware-00954-stable.xml.gz
> ```
> 
> ...in the --verbose logs.

Yes and fwupdmgr did not work until I added the patch.
I do not see "changing metadata URI..." in the log:
```
fwupdtool -v refresh |& grep metada
12:23:49:0041 Jcat                 trying to load certificate from /etc/pki/fwupd-metadata/LVFS-CA.pem
12:23:49:0041 Jcat                 reading /etc/pki/fwupd-metadata/LVFS-CA.pem with 1679 bytes
12:23:49:0053 XbSilo               attempting to load /var/cache/fwupd/metadata.xmlb

**Comment by joakim-tjernlund** [comment]
fwupdmgr does show:
```
fwupdmgr refresh -v --force | grep meta
(fwupdmgr:27737): Fwupd-DEBUG: 14:24:55.996: changing metadata URI from https://cdn.fwupd.org/downloads/firmware-testing.xml.gz to https://cdn.fwupd.org/downloads/firmware-00601-testing.xml.gz
(fwupdmgr:27737): Fwupd-DEBUG: 14:24:56.186: changing metadata URI from https://cdn.fwupd.org/downloads/firmware.xml.gz to https://cdn.fwupd.org/downloads/firmware-00954-stable.xml.gz

joakim-tjernlund:
BTW, not sure if this is expected but each time I reinstall fwupd, the daemon crashes.
Just have to restart it again though.

hughsie:
> fwupdtool

Ahh! sorry, I missed the fact you tested with `fwupdtool`! Same fix, just needs a bit of a re-arrange.

> but each time I reinstall fwupd, the daemon crashes

Hopefully it's not crashing, but it certainly is designed to exit gratefully when the binary is modified.

Can you try with the new patch on this branch too please. Thanks!

Richard.

joakim-tjernlund:
> > fwupdtool
> 
> Ahh! sorry, I missed the fact you tested with `fwupdtool`! Same fix, just needs a bit of a re-arrange.
> 
> > but each time I reinstall fwupd, the daemon crashes
> 
> Hopefully it's not crashing, but it certainly is designed to exit gratefully when the binary is modified.

I see, would prefer that the old daemon just continued. There may be some time in between install and activation. 
> 
> Can you try with the new patch on this branch too please. Thanks!

Just did, now it works for both tools :) Thanks

> 
> Richard.

joakim-tjernlund:
One final question, is 1.5.0 coming soon?

hughsie:
> I see, would prefer that the old daemon just continued

So, we had that and we got reports because the server-side bit was a different version to the client-side bits, and everyone got very confused. If it helps, fwupd also auto-quits on idle now, unless you have slow-to-query hardware installed.

> is 1.5.0 coming soon

That's a good question. @superm1 wanted to veto the release until we had the HSI spec more ready, but now we have Intel and ARM (although the latter hasn't actually contributed any code yet...) onboard I think that's more palatable. I think even for 1.5.0 the HSI work would be still labelled as proof-of-concept.

In a related sense, I need the async libfwupd functions for the async branch of gnome-firmware, which makes it actually predictable and responsive.

superm1:
> > is 1.5.0 coming soon
> 
> That's a good question. @superm1 wanted to veto the release until we had the HSI spec more ready, but now we have Intel and ARM (although the latter hasn't actually contributed any code yet...) onboard I think that's more palatable. I think even for 1.5.0 the HSI work would be still labelled as proof-of-concept.
> 
> In a related sense, I need the async libfwupd functions for the async branch of gnome-firmware, which makes it actually predictable and responsive.

Other than vendor support for HSI (and backing into an architectural or vendor specific corner) I have some concerns about how and what TPM stuff is measured in HSI that I think need to be reconciled. I've got a write-up I will share after I can align with a few more people.

hughsie:
> I have some concerns

Would those concerns be abridged somewhat if we disabled the `hsi.xml` in the official docs? Of if we didn't install the `fwupd-security-attrs.h` header? I'm keen to get a release out for all the new plugins and async libfwupd and we can always revert the disable if required.

superm1:
> Would those concerns be abridged somewhat if we disabled the hsi.xml in the official docs? Of if we didn't install the fwupd-security-attrs.h header? I'm keen to get a release out for all the new plugins and async libfwupd and we can always revert the disable if required.

You mean `fwupd-security-attrs.h` header right?  The one in `libfwupd`, not `libfwupdplugin`.
Specifically this section:
https://github.com/fwupd/fwupd/blob/82c3e3471d5f106ef6d75defac8030e2fde0e6af/libfwupd/fwupd-security-attr.h#L114 

Those defines essentially become a _loose_ plugin ABI as well as an ABI for any client to use and show strings.  If any of them need to be moved/removed/changed they can't (or shouldn't) be.

superm1:
> Those defines essentially become a loose plugin ABI as well as an ABI for any client to use and show strings. If any of them need to be moved/removed/changed they can't (or shouldn't) be.

It's kinda a big hammer, but what about a commit that comments out all that stuff (and it's use) for 1.5.0, and then revert it after release and bump it to 1.5.1 so we can kick the can a bit?

hughsie:
> Those defines essentially become a loose plugin ABI

Well, it's more than loose ABI, it's actual ABI. :) Galaxy brain moment: we don't actually need to export those IDs: https://github.com/fwupd/fwupd/pull/2453

PR #2453, opened 2020-10-07
Wip/hughsie/hsi

hughsie:
Type of pull request:
- [X] Code fix

### DIFF
--- a/docs/fwupd-docs.xml
+++ b/docs/fwupd-docs.xml
@@ -69,7 +69,9 @@
   </reference>
 
   <xi:include href="tutorial.xml"/>
+<!--
   <xi:include href="hsi.xml"/>
+-->
 
   <index id="api-index-full">
     <title>API Index</title>
--- a/libfwupd/fwupd-security-attr-private.h
+++ b/libfwupd/fwupd-security-attr-private.h
@@ -13,6 +13,37 @@
 
 G_BEGIN_DECLS
 
+#define FWUPD_SECURITY_ATTR_ID_ACPI_DMAR		"org.fwupd.hsi.AcpiDmar"		/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_ENCRYPTED_RAM		"org.fwupd.hsi.EncryptedRam"		/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_FWUPD_ATTESTATION	"org.fwupd.hsi.Fwupd.Attestation"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_FWUPD_PLUGINS		"org.fwupd.hsi.Fwupd.Plugins"		/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_FWUPD_UPDATES		"org.fwupd.hsi.Fwupd.Updates"		/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ENABLED	"org.fwupd.hsi.IntelBootguard.Enabled"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_VERIFIED	"org.fwupd.hsi.IntelBootguard.Verified"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ACM	"org.fwupd.hsi.IntelBootguard.Acm"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_POLICY	"org.fwupd.hsi.IntelBootguard.Policy"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_OTP	"org.fwupd.hsi.IntelBootguard.Otp"	/* Since: 1.5.0 */
+#define FWUPD_SECURITY_ATTR_ID_INTEL_CET_ENABLED	"org.fwupd.hsi.IntelCet.Enabled"	/* Since: 1.5.0 */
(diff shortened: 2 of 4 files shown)


superm1:
This certainly helps with the external client aspect, thanks.

However it's still (technically) a plugin ABI thing.
` libfwupdplugin/fu-security-attrs.h` is still installed into the filesystem, fwupdmgr and fwupdtool clients use it plugins can/do use it too.

src/fu-util-common.c
```diff
@@ -1633,7 +1633,7 @@ fu_util_security_attrs_to_string (GPtrArray *attrs, FuSecurityAttrToStringFlags
 
 			/* check for PCR0 not matching */
 			if (g_strcmp0 (fwupd_security_attr_get_appstream_id (attr),
-					FWUPD_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0) == 0 &&
+					FU_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0) == 0 &&
```
superm1, line 1636: So I think this specific one maybe we should stop special casing the client messaging.  It can be punted into https://github.com/fwupd/fwupd/wiki/Low-host-security-level and that page can link to  https://github.com/fwupd/fwupd/wiki/TPM-PCR0-differs-from-reconstruction

Then there is no longer a client dependency on attributes

hughsie:
> it's still (technically) a plugin ABI thing.

I'm happy marking this as unstable in the GTK_DOC if that helps. Maybe we should do what UPower does and require a -DI_KNOW_LIBFWUPDPLUGIN_IS_UNSTABLE for external use...

libfwupdplugin/fu-security-attrs.h
```diff
@@ -14,6 +14,37 @@
...
+#define FU_SECURITY_ATTR_ID_SPI_BLE			"org.fwupd.hsi.Spi.Ble"			/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_SPI_SMM_BWP			"org.fwupd.hsi.Spi.SmmBwp"		/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_SUSPEND_TO_IDLE		"org.fwupd.hsi.SuspendToIdle"		/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_SUSPEND_TO_RAM		"org.fwupd.hsi.SuspendToRam"		/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0	"org.fwupd.hsi.Tpm.ReconstructionPcr0"	/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_TPM_VERSION_20		"org.fwupd.hsi.Tpm.Version20"		/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_UEFI_SECUREBOOT		"org.fwupd.hsi.Uefi.SecureBoot"		/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_INTEL_DCI_ENABLED		"org.fwupd.hsi.IntelDci.Enabled"	/* Since: 1.5.0 */
+#define FU_SECURITY_ATTR_ID_INTEL_DCI_LOCKED		"org.fwupd.hsi.IntelDci.Locked"		/* Since: 1.5.0 */
+
```
superm1, line 47: How about if this whole section is moved into `libfwupdplugin/fu-security-attrs-private.h` and that's not installed into the filesystem?  It's then only usable by plugins in the tree, and we don't need to worry about it being supported forever.  Once my other concerns are assayed and more than Intel are on board we can make it installed in tree again?

hughsie: > moved into libfwupdplugin/fu-security-attrs-private.h

That's an even better idea. Will fix, but have a few hours of meetings first.

src/fu-util-common.c
superm1: Although I guess you still have `fu_security_attr_get_name` is used.. Hmm

superm1: I guess one approach is to rather than offer translations in the client by `fu_security_attr_get_name` to just leave appstream IDs.  Let the appstream ID decoder live in HSI docs (when they are live) or in wiki pages.

libfwupdplugin/fu-security-attrs.h
hughsie: > How about if this whole section is moved into libfwupdplugin/fu-security-attrs-private.h

That worked wonderfully, fixed.

src/fu-util-common.c
hughsie: >  I think this specific one maybe we should stop special casing the client messaging

Do you think now the PCR0 calculation has been fixed in fwupd we'll stop getting this? I still think it's somewhat useful to have translated content to cover the special cases.

superm1: I would like to think so, but we'll truly find out later. It's always a revert away to add/remove translations. How about for now the command line client uses the private file to get those strings and we can decide again later whether they stay.

hughsie: The fwupd-security-attr-private.h can't provide translations, it gets really messy when dealing with all the msgcontext stuff in a library.

> Then there is no longer a client dependency on attributes

I think fwupdmgr and fwupdtool can be considered internal enough, or are you worried about the case where a new daemon uses org.dtmf.hsi prefixes and the old client is expecting org.fwupd.hsi prefixes?

superm1: Oh I mean leave translations as is and just include the private header for now. We can remove them later when hsi goes stable if we decide that makes sense.

So yes keep it treated as internal for now.

hughsie: > and just include the private header for now

That's what I did, no?

superm1: I don't know, you haven't pushed anything :)

hughsie: Dude, you need me to do everything! :) Pushed now, sorry about that.

superm1 (approved):
OK this approach looks good for now for 1.5.0.

PR #2455, opened 2020-10-07
uefi: Make udisks2 errors more apparent

superm1:
When support for dynamically mounting disks was added for https://github.com/fwupd/fwupd/commit/25ba41579fb170cf3a29cb2753733172a54984cd
udisks2 became a harder dependency and it was less obvious to users.

Create devices but show an error in why devices aren't updatable if
it's not found.

Users can still configure ESP manually in `uefi.conf`

Fixes: #2444

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -443,10 +443,18 @@ fu_plugin_uefi_register_proxy_device (FuPlugin *plugin, FuDevice *device)
 {
 	FuPluginData *data = fu_plugin_get_data (plugin);
 	g_autoptr(FuUefiDevice) dev = fu_uefi_device_new_from_dev (device);
+	g_autoptr(GError) error_local = NULL;
 
 	/* load all configuration variables */
 	fu_plugin_uefi_load_config (plugin, FU_DEVICE (dev));
-	fu_uefi_device_set_esp (dev, data->esp);
+	if (data->esp == NULL)
+		data->esp = fu_common_get_esp_default (&error_local);
+	if (data->esp == NULL) {
+		fu_device_set_update_error (FU_DEVICE (dev), error_local->message);
+		fu_device_remove_flag (FU_DEVICE (dev), FWUPD_DEVICE_FLAG_UPDATABLE);
+	} else {
+		fu_uefi_device_set_esp (dev, data->esp);
+	}
 	fu_plugin_device_add (plugin, FU_DEVICE (dev));
 }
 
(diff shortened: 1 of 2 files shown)


plugins/uefi/fu-plugin-uefi.c
```diff
@@ -768,11 +777,14 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
...
 		fu_device_set_quirks (FU_DEVICE (dev), fu_plugin_get_quirks (plugin));
-		fu_uefi_device_set_esp (FU_UEFI_DEVICE (dev), data->esp);
+		if (data->esp != NULL)
+			fu_uefi_device_set_esp (FU_UEFI_DEVICE (dev), data->esp);
 		if (!fu_plugin_uefi_coldplug_device (plugin, dev, error))
 			return FALSE;
 		if (error_efivarfs != NULL) {
 			fu_device_set_update_error (FU_DEVICE (dev), error_efivarfs->message);
+		} else if (error_udisks2 != NULL) {
+			fu_device_set_update_error (FU_DEVICE (dev), error_udisks2->message);
```
hughsie, line 787: Not a criticism of this patch at all, but I do wonder if this is getting to the point where we should have actual API for a plugin to return a message to the user rather than either creating fake devices or setting the update error. It would certainly clean up a few plugins considerably.

e.g. an API like GetPlugins() where we return an array of `[ boolean:enabled, string:summary, string:error_message ]` which we use before we show the device tree in `fwupdmgr get-devices` or `fwupdmgr get-updates`.

superm1: Up until now we have been pretty prescriptive to not expose to the client that the implementation in the daemon is with plugins and they get disabled. So part of the problem can be older clients don't know how to show this information.

So if we're going to do such a thing maybe it's time to make a harder line in sand on versioning.

superm1: One more thought - there are a lot of plugins that get disabled and it's fully intentional. For example if you don't have a Dell system you won't have dell plugin, and if you have a uefi system you won't have coreboot.

hughsie: > So part of the problem can be older clients don't know how to show this information

Valid. Maybe in this case, server side, we can set the FuDevice update_error IF the device is not UPDATABLE and the FuPlugin has an error set?

> there are a lot of plugins that get disabled and it's fully intentional

I guess we either don't set an update message in that case (i.e. it's working as desired) or we make the `boolean:enabled` into an enum like _DISABLED_NOTIFY, _DISABLED_INTENTIONAL, ENABLED, etc.

superm1: Can we move the discussion to a new issue to figure out the solution? I think that this commit should backport directly into 1_4_X as is, and we can reswizzle for 1.5.1?

hughsie (approved):

PR #2459, opened 2020-10-08
cros-ec: RO/RW update with rw_sig support

### LINKED ISSUE #2454: cros-ec: Read Only/Read Write update with rw_sig checking support (plus honeybuns platform)

Current cros-ec plugin supports RO and RW update flow for devices that do not check the integrity of the RW partition (a process called rw_sig verification).

Servo Micro falls in that category, since it is a debug dongle, not a production device. In practice the lack of rw_sig check means the jump from RO to RW is not automatic. When it boots, it sits in RO, until something (ie, our updater) tells it to switch to RW.

rw_sig devices behave differently. They will automatically jump from RO to RW upon verifying the RW partition's signature. 
The honeybuns reference (gingerbread and quiche) will rw_sig.

Our current flow always boots into RO, then updates RW, then updates RO upon jumping to RW.

Since rw_sig devices will start from RW, and since important update-flow updates in RO may be required, update RO first  (this assumes that RO is not actually read-only, as is the current state of these prototypes), then jump to RO.

Once in RO, we need to carefully send the stay-in-ro message, which wasn't used properly in prior patch series. 
The stay-in-ro needs to be sent to cancel the RW_SIG and automatic jump to RW, and needs to happen right before the write phase.

**Additional context**

bleungatchromium:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

fixes #2454

### DIFF
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -21,6 +21,7 @@
 #define FLUSH_TIMEOUT_MS		10
 #define BULK_SEND_TIMEOUT_MS		2000
 #define BULK_RECV_TIMEOUT_MS		5000
+#define CROS_EC_REMOVE_DELAY_RE_ENUMERATE              20000
 
 #define UPDATE_DONE			0xB007AB1E
 #define UPDATE_EXTRA_CMD		0xB007AB1F
@@ -879,13 +880,13 @@ fu_cros_ec_usb_device_attach (FuDevice *device, GError **error)
 	FuCrosEcUsbDevice *self = FU_CROS_EC_USB_DEVICE (device);
 
 	if (self->in_bootloader && fu_device_has_custom_flag (device, "special")) {
-		fu_device_set_remove_delay (device, FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
+		fu_device_set_remove_delay (device, CROS_EC_REMOVE_DELAY_RE_ENUMERATE);
 		fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
 		fu_device_add_flag (device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
 		return TRUE;
 	}
 
-	fu_device_set_remove_delay (device, FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
+	fu_device_set_remove_delay (device, CROS_EC_REMOVE_DELAY_RE_ENUMERATE);
 	if (fu_device_has_custom_flag (device, "ro-written") &&
 	    !fu_device_has_custom_flag (device, "rw-written")) {
 		if (!fu_cros_ec_usb_device_reset_to_ro (device, error)) {
@@ -917,7 +918,7 @@ fu_cros_ec_usb_device_detach (FuDevice *device, GError **error)
 	} else if (self->targ.common.flash_protection != 0x0) {
 		/* in RW, and RO region is write protected, so jump to RO */
 		fu_device_set_custom_flags (device, "ro-written");
-		fu_device_set_remove_delay (device, FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
+		fu_device_set_remove_delay (device, CROS_EC_REMOVE_DELAY_RE_ENUMERATE);
 		if (!fu_cros_ec_usb_device_reset_to_ro (device, error))
 			return FALSE;
 
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -641,19 +641,12 @@ static gboolean
 fu_cros_ec_usb_device_reset_to_ro (FuDevice *device, GError **error)
 {
 	guint8 response;
-	guint16 subcommand = UPDATE_EXTRA_CMD_STAY_IN_RO;
+	guint16	subcommand = UPDATE_EXTRA_CMD_IMMEDIATE_RESET;
 	guint8 command_body[2]; /* Max command body size. */
 	gsize command_body_size = 0;
 	gsize response_size = 1;
 
-	/* send subcommand to remain in RO */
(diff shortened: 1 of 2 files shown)


plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -481,8 +482,17 @@ fu_cros_ec_usb_device_transfer_block (FuDevice *device, gpointer user_data,
...
 					    NULL,
 					    0, FALSE,
-					    NULL, error))
+					    NULL, error)) {
+		g_autoptr(GError) error_local = NULL;
+
+		/* flush all data from endpoint to recover in case of error */
+		if (!fu_device_retry (device, fu_cros_ec_usb_device_flush,
+				      SETUP_RETRY_CNT, NULL, &error_local)) {
+			g_debug ("failed to flush to idle");
```
superm1, line 491: if you're not actually showing the message from `error_local` here, I would just send `NULL` for `fu_device_retry`

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -494,15 +504,32 @@ fu_cros_ec_usb_device_transfer_block (FuDevice *device, gpointer user_data,
 						    NULL,
 						    0, FALSE,
 						    NULL, error)) {
+			g_autoptr(GError) error_local = NULL;
+			g_prefix_error (error, "failed at sending chunk: ");
+
+			/* flush all data from endpoint to recover in case of error */
+			if (!fu_device_retry (device, fu_cros_ec_usb_device_flush,
+					      SETUP_RETRY_CNT, NULL, &error_local)) {
+				g_debug ("failed to flush to idle");
```
superm1, line 513: same as other comment

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -494,15 +504,32 @@ fu_cros_ec_usb_device_transfer_block (FuDevice *device, gpointer user_data,
...
 					    (guint8 *)&reply, sizeof (reply),
-					    TRUE, &transfer_size, error))
+					    TRUE, &transfer_size, error)) {
+		g_autoptr(GError) error_local = NULL;
+
+		g_prefix_error (error, "failed at reply: ");
+		/* flush all data from endpoint to recover in case of error */
+		if (!fu_device_retry (device, fu_cros_ec_usb_device_flush,
+				      SETUP_RETRY_CNT, NULL, &error_local)) {
+			g_debug ("failed to flush to idle");
```
superm1, line 529: same as other comment

plugins/cros-ec/fu-cros-ec-usb-device.c
superm1: Also, since this is (essentially) the same code 3 times, does this want a static function to call into instead?

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -812,18 +902,29 @@ fu_cros_ec_usb_device_attach (FuDevice *device, GError **error)
...
+
+	if (fu_device_has_custom_flag (device, "rw-written") &&
+	    !fu_device_has_custom_flag (device, "ro-written"))
 		return TRUE;
 
-	fu_device_set_remove_delay (self, FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
-	if (!fu_cros_ec_usb_device_reset_to_ro (self, error))
-		return FALSE;
+	if (self->in_bootloader) {
+		g_print ("skipping immediate reboot in case of already in bootloader");
```
superm1, line 914: `g_debug` instead

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -812,18 +902,29 @@ fu_cros_ec_usb_device_attach (FuDevice *device, GError **error)
...
+		/* in RO so skip reboot */
+		return TRUE;
+	} else if (self->targ.common.flash_protection != 0x0) {
+		/* in RW, and RO region is write protected, so jump to RO */
+		fu_device_set_custom_flags (device, "ro-written");
+		fu_device_set_remove_delay (device, CROS_EC_REMOVE_DELAY_RE_ENUMERATE);
+		if (!fu_cros_ec_usb_device_reset_to_ro (device, error)) {
+			return FALSE;
+		}
```
hughsie, line 923: braces not required here, but otherwise generally looks good; Mario's comments are all valid.

plugins/cros-ec/fu-cros-ec-usb-device.c

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done.

plugins/cros-ec/fu-cros-ec-usb-device.c
bleungatchromium: Done. Thanks. This replaces 5 instances.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -494,15 +511,29 @@ fu_cros_ec_usb_device_transfer_block (FuDevice *device, gpointer user_data,
...
 			return FALSE;
 		}
 	}
 
 	/* get the reply */
 	if (!fu_cros_ec_usb_device_do_xfer (self, NULL, 0,
 					    (guint8 *)&reply, sizeof (reply),
-					    TRUE, &transfer_size, error))
+					    TRUE, &transfer_size, error)) {
+		g_autoptr(GError) error_local = NULL;
```
hughsie, line 528: are not these now unused?

bleungatchromium: Whoops. Good catch! Fixed!

PR #2461, opened 2020-10-09
bcm57xx: Improve reliability when flashing

meklort:
- Use pci function 0 instead of 1 when flashing firmware.
  In certain situations, the BCM5719 NVM controller can lockup if
  a function other than 0 is used to read from NVM word-by-word like
  the kernel driver does.
- Fix APE_MODE offset in BAR[2] to enable proper resetting of the APE.
- Remove unnededed NVRam lock when resetting the APE.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/bcm57xx/fu-bcm57xx-device.c
+++ b/plugins/bcm57xx/fu-bcm57xx-device.c
@@ -53,12 +53,12 @@ fu_bcm57xx_device_probe (FuUdevDevice *device, GError **error)
 	g_autofree gchar *fn = NULL;
 	g_autoptr(GPtrArray) ifaces = NULL;
 
-	/* only enumerate number 1 */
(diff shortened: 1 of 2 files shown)


meklort:
After additional testing of #2417, I ran into a pretty significant lockup of the NVM controller which would result in future flash attempts to fail. Switching from function 1 to function 0 works around the hardware lockup.
Note that the alternative fix is to patch the tg3 driver to use a burst read instead of word reads, however this would mean that unpached kernels will  still fail to flash. I'll look into adding the burst workaround to the tg3 driver itself, but it would be good to have this workaround still.

The other issue I was running into is that after flashing, the program would segfault as it attempted to reset the APE. This fixes the address used for APE_MOE.

hughsie:
Great, thanks!

PR #2462, opened 2020-10-09
fwupdtool: Enable the switch-branch client feature.

meklort:
This enabled fwupdtool to flash firmware with the switch-branch client feature requirement set.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2736,7 +2736,8 @@ main (int argc, char *argv[])
 		/* set our implemented feature set */
 		fu_engine_request_set_feature_flags (priv->request,
 						     FWUPD_FEATURE_FLAG_DETACH_ACTION |
-						     FWUPD_FEATURE_FLAG_UPDATE_ACTION);
+						     FWUPD_FEATURE_FLAG_UPDATE_ACTION |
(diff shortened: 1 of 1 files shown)


src/fu-tool.c
```diff
@@ -2736,7 +2736,8 @@ main (int argc, char *argv[])
 		/* set our implemented feature set */
 		fu_engine_request_set_feature_flags (priv->request,
 						     FWUPD_FEATURE_FLAG_DETACH_ACTION |
-						     FWUPD_FEATURE_FLAG_UPDATE_ACTION);
+						     FWUPD_FEATURE_FLAG_UPDATE_ACTION |
+						     FWUPD_FEATURE_FLAG_SWITCH_BRANCH);
```
meklort, line 2740: I'm not sure if this should be gated based on interactive or not. Feel free to fix if this is not right.

hughsie: I think @superm1 wanted us to implement the switch-branch warning dialog (with prompt) on upgrade before we added this flag. I can ponder how to do this next week.

superm1: Yeah

superm1: implemented same way as fwupdmgr in #2493

PR #2465, opened 2020-10-10
meson.build: update min meson version to 0.50.0.

ericonr:
The following feature is used:

 > * 0.50.0: {'install arg in configure_file'}

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('fwupd', 'c',
   version : '1.5.0',
   license : 'LGPL-2.1+',
-  meson_version : '>=0.47.0',
+  meson_version : '>=0.50.0',
   default_options : ['warning_level=2', 'c_std=c99'],
 )
 


hughsie:
I think the warning can be ignored; unfortunately upping the meson buildrequires would break compiling on RHEL and Ubuntu LTS.

ericonr:
Well, as long as it still works there :)

Would you prefer I leave this open as a reminder? Otherwise I can go ahead and close it.

hughsie:
> Would you prefer I leave this open as a reminder

Ya, lets keep this open.

ericonr:
do I have to do anything to avoid the bot closing it?

superm1:
I sorta think we shouldn't bother keeping this open, it will be a few years until it's time to stop worrying about RHEL and current Ubuntu LTS versions.  Just re-visit it then.

hughsie:
>  Just re-visit it then

Agree, we can reopen then.

PR #2466, opened 2020-10-10
Fix header names and remove check for <sys/errno.h>.

ericonr:
- <sys/poll.h> should be <poll.h>
- <sys/errno.h> should be <errno.h>

In the case of <errno.h>, no other source file has guards for including
it, so their detection was removed.

---

There are a few weird defines, such as `HAVE_IOCTL_H`, which aren't used for all source files - some have `#ifdef`s to include `<sys/ioctl.h>`, others include it unconditionally. Should this be the case? I can make a PR to make it consistent, if desired.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -11,7 +11,7 @@
 #include <fcntl.h>
 #include <string.h>
 #ifdef HAVE_ERRNO_H
-#include <sys/errno.h>
+#include <errno.h>
 #endif
 #ifdef HAVE_IOCTL_H
 #include <sys/ioctl.h>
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -10,9 +10,7 @@
 
 #include <fcntl.h>
 #include <string.h>
-#ifdef HAVE_ERRNO_H
-#include <sys/errno.h>
-#endif
+#include <errno.h>
```
superm1, line 13: The reason for the weird ifdef is to allow windows cross compilation.

superm1 (commented):
Can you just fix the paths but leave the weird ifdef?

libfwupdplugin/fu-udev-device.c
ericonr: I see, that (kinda?) makes sense!

ericonr: Fixed.

PR #2467, opened 2020-10-10
Add fu_device_msleep_with_progress helper()

hughsie:
This makes the console a bit more helpful than just 'hanging'...

### DIFF
--- a/plugins/bcm57xx/fu-bcm57xx-device.c
+++ b/plugins/bcm57xx/fu-bcm57xx-device.c
@@ -318,8 +318,7 @@ fu_bcm57xx_device_activate (FuDevice *device, GError **error)
 
 	/* wait for the device to restart before calling reload() */
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_BUSY);
-	fu_device_set_progress (device, 0);
-	g_usleep (G_USEC_PER_SEC * 5);
+	fu_device_sleep_with_progress (device, 5); /* seconds */
 	return TRUE;
 }
 
--- a/plugins/dfu/dfu-device.c
+++ b/plugins/dfu/dfu-device.c
@@ -1270,7 +1270,7 @@ dfu_device_probe (FuUsbDevice *device, GError **error)
 	/* hardware rom Jabra literally reboots if you try to retry a failed
 	 * write -- there's no way to avoid blocking the daemon like this... */
 	if (fu_device_has_custom_flag (FU_DEVICE (device), "attach-extra-reset"))
-		g_usleep (10 * G_USEC_PER_SEC);
+		fu_device_sleep_with_progress (FU_DEVICE (self), 10); /* seconds */
 
 	/* success */
 	return TRUE;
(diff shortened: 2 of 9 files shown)


plugins/vli/fu-vli-usbhub-msp430-device.c
```diff
@@ -149,8 +149,7 @@ fu_vli_usbhub_msp430_device_detach (FuDevice *device, GError **error)
 
 	/* avoid power instability by waiting T1 */
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
-	fu_device_set_progress (device, 0);
-	g_usleep (G_USEC_PER_SEC);
+	fu_device_msleep_with_progress (device, 4000);
```
superm1, line 152: Didn't you bump up the sleep to longer here

plugins/dfu/dfu-device.c
```diff
@@ -1270,7 +1270,7 @@ dfu_device_probe (FuUsbDevice *device, GError **error)
 	/* hardware rom Jabra literally reboots if you try to retry a failed
 	 * write -- there's no way to avoid blocking the daemon like this... */
 	if (fu_device_has_custom_flag (FU_DEVICE (device), "attach-extra-reset"))
-		g_usleep (10 * G_USEC_PER_SEC);
+		fu_device_msleep_with_progress (FU_DEVICE (self), 10000);
```
superm1, line 1273: Maybe 10* usec_sec

plugins/bcm57xx/fu-bcm57xx-device.c
```diff
@@ -318,8 +318,7 @@ fu_bcm57xx_device_activate (FuDevice *device, GError **error)
 
 	/* wait for the device to restart before calling reload() */
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_BUSY);
-	fu_device_set_progress (device, 0);
-	g_usleep (G_USEC_PER_SEC * 5);
+	fu_device_msleep_with_progress (device, 5000);
```
superm1, line 321: Maybe 5* usec_per_sec

plugins/vli/fu-vli-usbhub-msp430-device.c
hughsie: Eek, fixed.

plugins/dfu/dfu-device.c
hughsie: I called the method **m**sleep, as in miliseconds. Perhaps that's a mistake; perhaps it should be seconds -- we don't want to show the 0..100% progressbar for any less than one second anyway. Then it would be `fu_device_sleep_with_progress (FU_DEVICE (self), 10); /* seconds */`

superm1: Sounds good to me

hughsie: Done

superm1 (approved):

PR #2470, opened 2020-10-13
Offer to compile daemon without polkit support

superm1:
Based on a patch from Daniel Campello at
https://chromium-review.googlesource.com/c/chromiumos/third_party/fwupd/+/2354521

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -14,7 +14,9 @@
 #include <glib/gi18n.h>
 #include <glib-unix.h>
 #include <locale.h>
+#ifdef HAVE_POLKIT
 #include <polkit/polkit.h>
+#endif
 #include <stdio.h>
 #include <stdlib.h>
 #include <jcat.h>
@@ -33,13 +35,15 @@
 #include "fu-install-task.h"
 #include "fu-security-attrs-private.h"
 
+#ifdef HAVE_POLKIT
 #ifndef HAVE_POLKIT_0_114
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunused-function"
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(PolkitAuthorizationResult, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC(PolkitSubject, g_object_unref)
 #pragma clang diagnostic pop
-#endif
+#endif /* HAVE_POLKIT_0_114 */
+#endif /* HAVE_POLKIT */
 
 typedef enum {
 	FU_MAIN_MACHINE_KIND_PHYSICAL,
@@ -57,7 +61,9 @@ typedef struct {
 #if GLIB_CHECK_VERSION(2,63,3)
 	GMemoryMonitor		*memory_monitor;
 #endif
+#ifdef HAVE_POLKIT
 	PolkitAuthority		*authority;
+#endif
 	guint			 owner_id;
 	FuEngine		*engine;
 	gboolean		 update_in_progress;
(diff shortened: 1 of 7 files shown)


src/fu-main.c
```diff
@@ -653,6 +684,9 @@ fu_main_authorize_install_queue (FuMainAuthHelper *helper_ref)
 						      NULL,
 						      fu_main_authorize_install_cb,
 						      g_steal_pointer (&helper));
+#else
+		fu_main_authorize_install_cb (NULL, NULL, g_steal_pointer (&helper));
```
hughsie, line 688: does this mean that anything can do anything? I kinda guessed we might check uid=0 or something in the !polkit case

hughsie:
This task was literally on my TODO list for tomorrow, so many thanks!

src/fu-main.c
superm1: anything can do anything yes..

superm1:
> This task was literally on my TODO list for tomorrow, so many thanks!

No problem, feel free to iterate directly on the branch as you see fit too.

src/fu-main.c
hughsie: >  Would it (or should it be) 0?

As I understand it, the upstart launcher would be uid 0.

superm1: Pushed another commit.  Here's how it looks now:

```
$ fwupdmgr modify-remote lvfs Enabled false
Authenticating…          [-                                      ]
untrusted client process
$ sudo fwupdmgr modify-remote lvfs Enabled false
Authenticating…          [-                                      ]
Successfully modified remote
```

I think we should keep that in, hopefully @djcampello confirms it makes sense from ChromeOS standpoint too.

hughsie (approved):
Assuming this passes CI, and is squashed down, LGTM, thanks!

superm1:
@djcampello this is specifically for y'all, so if you can please review it before we merge.

src/fu-main.c
djcampello: Sorry for the delay... I imagine that if in first instance we will replace every occurrence of fwupdtool with fwupdmgr, these invocations are sandboxed and run as a different user than 0. See https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/master/sys-apps/fwupd/files/fwupdtool-update.conf#23

djcampello (approved):
I'm not sure that we will end up running as daemon given the sandoxing and existing per-plugin permissions requirements in place by the security team. If we do move to a daemon architecture this sounds good. We could also update the D-Bus file to restrict which user can issue dbus calls in place of the PK rules.

src/fu-main.c
superm1: My proposal on how to do it would be a little different; instead rely upon dbus activation and jail it a little different.  See what you would think of this.  It should work with the changes as landed in this PR right now.

1. Modify the dbus service file's `Exec` line to match your appropriate `minijail0` command.  Let daemon launch as `fwupd` user/group here.
That's this: https://github.com/fwupd/fwupd/blob/master/data/org.freedesktop.fwupd.service.in
2. Modify all minijail'ed `fwupdtool` calls from udev rules into `fwupdmgr` calls that launch as a `root` user that can't do anything other than talk to daemon.
3. Configure daemon to shut down after some period of time via fwupd.conf configuration (TBD).

So the daemon would then only start when a udev rule fires or a client launches it (manual `fwupdmgr` call or if there was potentially another client on system).  Updates would only be performed if a client was launched from a udev rule however.

djcampello: > So the daemon would then only start when a udev rule fires or a client launches it (manual fwupdmgr call or if there was potentially another client on system). 

Does fwupdmgr launches the daemon implicitly?

hughsie: As long as you have dbus activation enabled, yes.

superm1: On regular Linux trying to access the dbus system service provided by fwupd using a dbus client like `fwupdmgr` will start it.
It's called "dbus activation".  I'm not sure if it's supported in Chrome.
https://dbus.freedesktop.org/doc/system-activation.txt

djcampello: I incorporated this on my POC at crrev.com/c/2391972

PR #2478, opened 2020-10-15
Wip/hughsie/elantp

hughsie:
Type of pull request:
- [ ] Code fix

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2170,20 +2170,6 @@ fu_engine_install_release (FuEngine *self,
 		str = g_strdup_printf ("device version not updated on success, %s != %s",
 				       version_rel, fu_device_get_version (device));
 		fu_device_set_update_error (device, str);
-		if ((flags & FWUPD_INSTALL_FLAG_NO_HISTORY) == 0 &&
-		    !fu_history_modify_device (self->history, device, error))
-			return FALSE;
-		/* success */
-		return TRUE;
-	}
-
-	/* ensure the new version matched what we expected */
-	if (version_rel != NULL &&
-	    g_strcmp0 (fu_device_get_version (device), version_rel) != 0) {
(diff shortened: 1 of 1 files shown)


jinglewu:
it works.

src/fu-engine.c
```diff
@@ -311,7 +311,8 @@ fu_engine_get_release_version (FuEngine *self, FuDevice *dev, XbNode *rel, GErro
 		return g_strdup (version);
 
 	/* don't touch my version! */
-	if (fmt == FWUPD_VERSION_FORMAT_PLAIN)
+	if (fmt == FWUPD_VERSION_FORMAT_PLAIN ||
+	    fmt == FWUPD_VERSION_FORMAT_HEX)
```
superm1, line 315: Why doesn't version format hex parse as an integer properly?

hughsie: > 0x0003 gets transformed into 0x000003

Exactly, and then we do the dumb strcmp rather than a vercmp. I guess it's a UX thing; do we expect what the user sees in the terminal to match exactly what the metadinfo has? The purist in my would say yes, but if the hardware reports 0x0003 and it's `3` in the metainfo maybe that's just fine?

superm1: I think treating plain and hex the same might be asking for issues on other stuff that uses hex.  What if the other devices with hex do it differently right now, now we stop being able to compare the actual values.

So I think I would rather just render hex as dropping leading zeros, but still treat as integer.  We can also simplify what `get-releases` shows to do the same thing.

hughsie: Hang on, back a bit, I'm insane. The chunk above this test actually tests the version using fu_common_vercmp_full() which will do the right thing for hex anyway -- in the event it fails we set an update error. I think we can just nuke the section completely!

superm1 (approved):

PR #2479, opened 2020-10-15
Set _HAS_MULTIPLE_BRANCHES when release differs from device

hughsie:
Enable switch-branch any time a branch other than the default exists.

Based on a patch by Evan Lojewski <github@meklort.com>, many thanks.

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -1982,18 +1982,21 @@ fu_util_switch_branch (FuUtilPrivate *priv, gchar **values, GError **error)
 	if (rels == NULL)
 		return FALSE;
 
+	/* get all the unique branches */
+	for (guint i = 0; i < rels->len; i++) {
+		FwupdRelease *rel_tmp = g_ptr_array_index (rels, i);
+		const gchar *branch_tmp = fu_util_release_get_branch (rel_tmp);
+		if (g_ptr_array_find_with_equal_func (branches, branch_tmp,
+						      g_str_equal, NULL))
+			continue;
+		g_ptr_array_add (branches, g_strdup (branch_tmp));
+	}
+
 	/* branch name is optional */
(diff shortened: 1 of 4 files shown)


meklort:
This lets me run the switch-branch command.
Note that when I run switch-branch without specifying a branch, no branches are listed, however I don't know if the other options did that either:
```shell
Selected device: NetXtreme BCM5719 Gigabit Ethernet PCIe
Choose a branch:
0.	Cancel
0
Request canceled
```

hughsie:
> no branches are listed

Doh, looks like I forgot to populate the array. 10 minutes.

hughsie:
> no branches are listed

Can you try this commit pls ^^

meklort:
Latest commit works as expected
- If one branch is available, it is defaulted to
- If two branches are available, the list is printed for selection.

src/fu-util.c
```diff
@@ -1953,6 +1953,16 @@ fu_util_switch_branch_warning (FuUtilPrivate *priv,
 	return TRUE;
 }
 
+static const gchar *
+fu_util_get_branch_fallback (const gchar *nullable_branch)
+{
+	if (nullable_branch == NULL) {
+		/* TRANSLATORS: this is the default branch name when unset */
+		return _("default");
```
superm1, line 1961: this could be a different behavior of this will be different than `fwupdtool`.  I would say just pull the translation into the engine too in case it's launched from fwupdtool

hughsie: We can't push translated content over the D-Bus interface; we could push the fu_util_get_branch_fallback() into fu-util-common.c tho...

superm1: Yeah I think that works

hughsie: Done, it's nicer as an API too.

PR #2482, opened 2020-10-15
When compiled without PK require signed payloads for Install/Activate commands and root user for all other commands

superm1:
This can allow running daemon as a specific user and let client run as that same specific user and be considered the trusted user to communicate with the daemon.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-self-test.c
+++ b/src/fu-self-test.c
@@ -324,8 +324,13 @@ fu_engine_requirements_client_pass_func (gconstpointer user_data)
 	ret = fu_engine_check_requirements (engine, request, task,
 					    FWUPD_INSTALL_FLAG_NONE,
 					    &error);
+#ifdef HAVE_POLKIT
 	g_assert_no_error (error);
 	g_assert (ret);
+#else
+	g_assert_error (error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE);
+	g_assert_false (ret);
+#endif
 }
 
 static void
@@ -465,8 +470,13 @@ fu_engine_requirements_child_func (gconstpointer user_data)
 	ret = fu_engine_check_requirements (engine, request, task,
 					    FWUPD_INSTALL_FLAG_NONE,
 					    &error);
+#ifdef HAVE_POLKIT
 	g_assert_no_error (error);
 	g_assert (ret);
+#else
+	g_assert_error (error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE);
+	g_assert_false (ret);
+#endif
 }
 
 static void
@@ -560,8 +570,13 @@ fu_engine_requirements_func (gconstpointer user_data)
 	ret = fu_engine_check_requirements (engine, request, task,
 					    FWUPD_INSTALL_FLAG_NONE,
 					    &error);
+#ifdef HAVE_POLKIT
 	g_assert_no_error (error);
 	g_assert (ret);
+#else
+	g_assert_error (error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE);
+	g_assert_false (ret);
+#endif
 }
 
 static void
@@ -617,8 +632,13 @@ fu_engine_requirements_device_func (gconstpointer user_data)
 	ret = fu_engine_check_requirements (engine, request, task,
 					    FWUPD_INSTALL_FLAG_NONE,
 					    &error);
+#ifdef HAVE_POLKIT
 	g_assert_no_error (error);
 	g_assert (ret);
+#else
+	g_assert_error (error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE);
+	g_assert_false (ret);
+#endif
 }
 
 static void
@@ -664,8 +684,13 @@ fu_engine_requirements_device_plain_func (gconstpointer user_data)
 	ret = fu_engine_check_requirements (engine, request, task,
 					    FWUPD_INSTALL_FLAG_NONE,
 					    &error);
+#ifdef HAVE_POLKIT
 	g_assert_no_error (error);
 	g_assert (ret);
+#else
+	g_assert_error (error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE);
+	g_assert_false (ret);
+#endif
 }
 
 static void
@@ -780,8 +805,13 @@ fu_engine_requirements_other_device_func (gconstpointer user_data)
 	ret = fu_engine_check_requirements (engine, request, task,
 					    FWUPD_INSTALL_FLAG_NONE,
 					    &error);
+#ifdef HAVE_POLKIT
 	g_assert_no_error (error);
 	g_assert (ret);
+#else
+	g_assert_error (error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE);
+	g_assert_false (ret);
+#endif
 }
 
 static void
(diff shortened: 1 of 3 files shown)


hughsie:
This assumes the daemon is always running as UID? I'm not clear why the sysadmin would ever want to change this on a deployed system, shouldn't it just be an integer option in `meson_options.txt` and set at build time?

superm1:
I suppose that's one option, and the other one that comes up is to have the daemon look up the uid it's running as and trust that uid.

superm1:
Switched it to build time ^

hughsie:
> have the daemon look up the uid it's running as and trust that uid

That seems even simpler still, although I don't know if this will be sufficient for Google.

djcampello:
Thinking out loud: 
What if another process, say chrome (in ChromeOS it runs as the user chronos), wants to issue operations via libfwupd (not sure how the integration using the library is done at this point) instead of using upstart to spawn fwupdmgr with the same uid as the daemon. Should we configure this uid as chronos and then any execution of fwupdmgr needs also to be run as chronos? Should we have a trusted group instead of trusted uid?

superm1:
> Thinking out loud:
> What if another process, say chrome (in ChromeOS it runs as the user chronos), wants to issue operations via libfwupd (not sure how the integration using the library is done at this point) instead of using upstart to spawn fwupdmgr with the same uid as the daemon. Should we configure this uid as chronos and then any execution of fwupdmgr needs also to be run as chronos? Should we have a trusted group instead of trusted uid?

I think the part to consider is the threat model of a potentially untrusted process initiating the update, and what it can pass as data.  If you're allowing a process that runs as a user's context and that's OK for the threat model, I think it's better to just remove the check for an ID all together (and offer a compile time option that removes it).

In terms of whether a group can do it - `GetConnectionUnixUser` only gets you a `uid`.  I guess we can configure a group and then have some checking in the daemon to lookup if that `uid` is part of the "trusted" group.

hughsie:
> I guess we can configure a group and then have some checking in the daemon to lookup if that uid is part of the "trusted" group.

I think we're dangerously close to just reinventing PolicyKit now...

superm1:
Another passing thought I had on this.  We don't really have a mode for the daemon to reject "unsigned" CAB files.  If there is a signature in it and it doesn't match they fail.  If there is no signature though, it just accepts it.  In order for an OS to _potentially_ pass random payloads of data, it's probably worth inverting that and making all things HAVE to be signed from fwupdgmr.

superm1:
> My gut instinct would just be to reject all unsigned updates without PolicyKit. I also think Google needs to tell us what policy they want before we essentially reinvent libpolkit...

I actually think we should be doing that across the board, _even with_ PK.  We're potentially enabling a bad actor to be able to package a payload into an unsigned CAB and install it if PK rules would allow them to.

Thinking about the case of a device that doesn't verify it's own signature (like the current MST hubs).  I think we should shift that into only being a case that can be done via `$ fwupdtool install-blob`.  If you want to install stuff from a CAB it's got to be signed by something in `$sysconfdir/pki/fwupd`.

hughsie:
> device that doesn't verify it's own signature

That implies something like FU_DEVICE_FLAG_SIGNED_PAYLOAD, no? I'd be happy with that, although it might take some time to work out each one.

superm1:
>That implies something like FU_DEVICE_FLAG_SIGNED_PAYLOAD, no? I'd be happy with that, although it might take some time to work out each one.

Some devices I think you can work it out pretty easily because the protocol on LVFS knows it already.  Others you can't (say UEFI updates it depends on the OEM's implementation).  

But going back to the _peeling of onion_, it feels like even on signed/verified payloads it makes it harder for someone "trying" to attack it.

superm1:
If not wanting to make LVFS signature mandatory, how about adding a new PK prompt for explicitly untrusted CAB?

hughsie:
> how about adding a new PK prompt for explicitly untrusted CAB

Don't we have exactly that now? https://github.com/fwupd/fwupd/blob/master/policy/org.freedesktop.fwupd.policy.in#L16 -- the files with valid trust use the `-trusted` suffix and those without dont.

superm1:
> Don't we have exactly that now? https://github.com/fwupd/fwupd/blob/master/policy/org.freedesktop.fwupd.policy.in#L16 -- the files with valid trust use the -trusted suffix and those without dont.

Ah... yes you're right.  I didn't see it working properly because removing the `jcat` file it was still trusting, but that's because of the `asc` in there too.  Removing it too shows the PK prompt.  So potentially in the PK-less solution that's something else we could change too.

So a revised idea @djcampello is that we require matching uid for all actions that would normally use PK except for `install` and `activate`.  
That would mean drop this PR and new proposal:

Following activities would need root.  I don't think they would be used in a chromeos context anyway though.
* Sign data using client cert
* Set approved firmware
* Modify remote
* Verify update
* Modify daemon config
* Unlock device

These activities would not require root and any user could run them.
* Activate
* Install internal (trusted payload only)
* Install external (trusted payload only)

superm1:
Here is what experience looks like now.  First payload untrusted, second trusted.

```
$ fwupdmgr install ~/Downloads/tmp/firmware.cab --allow-reinstall 
untrusted release
$ fwupdmgr install ~/Downloads/xps_13_7390_1_6_0.cab --allow-reinstall 
Authenticating…          [       \                               ]
Installing on System Firmware…    |                              ]
Scheduling…              [***************************************]

An update requires a reboot to complete. Restart now? [y|N]: n
```

src/fu-install-task.c
```diff
@@ -418,6 +418,15 @@ fu_install_task_check_requirements (FuInstallTask *self,
 			return FALSE;
 		}
 	}
+#ifndef HAVE_POLKIT
+	if ((self->trust_flags & FWUPD_TRUST_FLAG_PAYLOAD) == 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INVALID_FILE,
+				     "untrusted release");
```
hughsie, line 426: maybe something more helpful, like `archive has not been signed and release is untrusted`?

superm1: Well you don't know that it's not signed, it might be signed but with the wrong signature.  So maybe `archive signature not trusted`?

hughsie: I also think this is in the wrong "layer", i.e. the logic is sound but it probably should live in `fu_engine_check_requirements` instead.

hughsie: > archive signature not trusted

maybe `archive signature missing or not trusted`?

src/fu-self-test.c
```diff
@@ -2917,7 +2980,13 @@ fu_plugin_composite_func (gconstpointer user_data)
 			g_ptr_array_add (install_tasks, g_steal_pointer (&task));
 		}
 	}
+#ifdef HAVE_POLKIT
 	g_assert_cmpint (install_tasks->len, ==, 3);
+#else
+	g_assert_cmpint (install_tasks->len, ==, 0);
+	g_test_skip("Compiling without PK support requires signed payloads");
```
hughsie, line 2987: space before bracket

hughsie (approved):
squash it down, and then LGTM, thanks.

djcampello:
Would root be able to install unsigned packages?

superm1:
>Would root be able to install unsigned packages?

No, not in the current implementation.
However root could use `fwupdtool install foo.cab` or `fwupdtool install-blob bar.blob`

Issue #2169, opened 2020-06-11
fwupd creates duplitate boot entries for Linux-Firmware-Updater on ThinkPad Carbon g7


crosser:
**Describe the bug**
On Lenovo ThinkPad Carbon gen 7 running Ubuntu focal, fwupd downloads and prepares installation of updates, but after reboot fwupdx64 does not run.

Ubuntu bug is [here](https://bugs.launchpad.net/ubuntu/+source/fwupd/+bug/1882357)

**Steps to Reproduce**
After fwupd run but before reboot, efi boot table contains two entries for the updater, boot entry #0 being added by fwupdmgr. After reboot there is one entry #2, that was always there. When the existing entry is selected by hand (`sudo efibootmgr -n 2`) then after reboot updater is run and installs updates successfully.

**Expected behavior**
efi boot table should be such as to load fwupdx64 on next boot.

**EFI boot table after running fwupdmgr**
```
BootNext: 0000
BootCurrent: 0001
Timeout: 0 seconds
BootOrder: 0001,0019,001A,001B,001C,001D,001E,001F,0020,0021,0022,0023,0024,0002,0000
Boot0000* Linux-Firmware-Updater HD(1,GPT,6ccce482-e2c2-48ca-991e-608bee5d38af,0x800,0x100000)/File(\EFI\ubuntu\shimx64.efi)\.f.w.u.p.d.x.6.4...e.f.i...
Boot0001* ubuntu HD(1,GPT,6ccce482-e2c2-48ca-991e-608bee5d38af,0x800,0x100000)/File(\EFI\ubuntu\shimx64.efi)
Boot0002* Linux-Firmware-Updater HD(1,GPT,6ccce482-e2c2-48ca-991e-608bee5d38af,0x800,0x100000)/File(\EFI\ubuntu\shimx64.efi)\.f.w.u.p.d.x.6.4...e.f.i...
Boot0010 Setup FvFile(721c8b66-426c-4e86-8e99-3457c46ab0b9)
Boot0011 Boot Menu FvFile(126a762d-5758-4fca-8531-201a7f57f850)
Boot0012 Diagnostic Splash Screen FvFile(a7d8d9a6-6ab0-4aeb-ad9d-163e59a7a380)
...
```

**"Normal" EFI boot table**
```
BootCurrent: 0001
Timeout: 0 seconds
BootOrder: 0001,0019,001A,001B,001C,001D,001E,001F,0020,0021,0022,0023,0024,0002
[...]

PR #2483, opened 2020-10-16
uefi: Don't overwrite BootNext and create new Boot#### when installin…

superm1:
…g 2 FW

Fixes: #2169

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-efivar.c
+++ b/libfwupdplugin/fu-efivar.c
@@ -456,3 +456,30 @@ fu_efivar_secure_boot_enabled (void)
 {
 	return fu_efivar_secure_boot_enabled_full (NULL);
 }
+
+/**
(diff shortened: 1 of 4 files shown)


hughsie:
That seems kinda fragile... I don't think it's actually addressing the bug, just working around it. A better workaround would be not to use that global `Dir` object, and another would be to set up bootnext on composite_finish() or whatever we called the vfunc.

superm1:
I don't think we can get a pointer to Dir without changes to efivar.

superm1:
In terms of composite - that's going to be a not good enough workaround. If you install one update, and then don't reboot your machine and install another you'll hit the same failure, but it won't be composite.

superm1:
close in favor of #2485

PR #2484, opened 2020-10-16
Wip/hughsie/vli

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
I think you need to look for and either show or no-op `FWUPD_DEVICE_FLAG_IS_COMPOSITE`

I wanted to ask though as well - why can't you just tell from topology something is composite?  You should be able to tell if it has a parent or child it's part of a composite device I would expect.

hughsie:
> I wanted to ask though as well - why can't you just tell from topology something is composite? 

You're right to question this, it's a hack. I've spent some time to do this properly, i.e. we count the devices _out_, and then count them _back in_. This means we do exactly the right thing without adding extra fields.

superm1 (approved):

PR #2485, opened 2020-10-16
uefi: Use fu_efivar_get_data() to fix setting BootNext correctly

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2169

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
100% untested, sorry.

superm1:
> 100% untested, sorry.

No worries, you turned that around really quick.  It unfortunately segfaults at 
```
			rc = efi_set_variable (*guid, name, opt, opt_size, attr, 0644);
```

It's because `guid` is no longer set (this was set previously by `efivar` when iterating)..  This particular function it's probably better to just switch to `fu_efivar_set_data`, but if switching that too is begging a bigger question now.
Why do we need to keep an efivar dep?  Maybe this should just a full out rip of efivar dependency and use the internal `fu-efivar` everywhere.  I'll rip it out of `dell`, it looks like it's just uefi and redfish plugin using it then.

hughsie:
>  It unfortunately segfaults

Fixed I think.

> it looks like it's just uefi and redfish plugin using it then

Yes, I can do that if you want. Redfish should be easy, UEFI might need a bit more time.

superm1:
>Fixed I think.

Fixes the segfault and makes it functional, but there is still a bug.  It's updating the attribute even when it doesn't need to.
`var_data_size` is reading as 178 and `opt_size` is 152.  (Presumably actual data is different too - but I didn't check)

superm1:
>Yes, I can do that if you want. Redfish should be easy, UEFI might need a bit more time.

Less dependencies I think is a net good thing, but it's your time, your call :)

hughsie:
> but it's your time, your call

Lets try this ^^^. The UEFI bits need a lot of testing, but I'm EOUTOFTIME for tonight.

libfwupdplugin/fu-efivar.c
```diff
@@ -322,6 +322,78 @@ fu_efivar_get_data (const gchar *guid, const gchar *name, guint8 **data,
...
+		return NULL;
+	return g_bytes_new_take (data, datasz);
+}
+
+/**
+ * fu_efivar_get_names:
+ * @guid: Globally unique identifier
+ * @error: A #GError
+ *
+ * Sets the data to a UEFI variable in NVRAM
```
superm1, line 356: wrong description

superm1 (commented):
given the amount of changes here, I don't know those symbols and these changes should come back to 1.4.7.

Need to update dependencies.xml to drop the efivar deps now too.

hughsie:
> given the amount of changes here, I don't know those symbols and these changes should come back to 1.4.7.

My gut instinct is to backport af3405dd655ea04cc287488919459fe75df1c7fb but leave the others in 1.5.0, indeed.

> Need to update dependencies.xml to drop the efivar deps now too.

I'm not sure; we still use libefiboot in the uefi plugin, and I think that's packaged in libefivar-devel in most distros. I think duplicating libefiboot would be not a trivial amount of work.

libfwupdplugin/fu-efivar.c
hughsie: Fixed.

superm1:
> I'm not sure; we still use libefiboot in the uefi plugin, and I think that's packaged in libefivar-devel in most distros. I think duplicating libefiboot would be not a trivial amount of work.

Oh in Debian and Ubuntu they're separate, `libefivar-dev` and `libefiboot-dev`

superm1:
> My gut instinct is to backport af3405d but leave the others in 1.5.0, indeed.

Another possibility is to put something like what I had in #2483 for 1_4_X (although without a function in libfwupdplugin) and then do it with this bigger change in master like this PR does.

hughsie:
> Oh in Debian and Ubuntu they're separate

I think less separate than you would hope:

    meson.build:310:2: ERROR: Could not generate cargs for efiboot:
    Package efivar was not found in the pkg-config search path.
    Perhaps you should add the directory containing `efivar.pc'
    to the PKG_CONFIG_PATH environment variable
    Package 'efivar', required by 'efiboot', not found

hughsie:
>  It's updating the attribute even when it doesn't need to

I get:

    08:31:44:0318 FuPluginUefi         Boot0002 -> Linux-Firmware-Updater : re-using existing boot entry

got any `--verbose` output showing the problems please? Maye also dump `var_data` and `opt` to console?

superm1:
>got any --verbose output showing the problems please? Maye also dump var_data and opt to console?

Rebuilt from this branch and all is working for me now.

superm1 (approved):

PR #2486, opened 2020-10-17
Wip/hughsie/hsi smbios chassis

hughsie:
Type of pull request:
- [X] Feature


libfwupdplugin/fu-smbios.h
```diff
@@ -20,12 +20,52 @@ FuSmbios	*fu_smbios_new			(void);
...
+#define FU_SMBIOS_CHASSIS_KIND_ADVANCED_TCA	0x1B
+#define FU_SMBIOS_CHASSIS_KIND_BLADE		0x1C
+#define FU_SMBIOS_CHASSIS_KIND_TABLET		0x1E
+#define FU_SMBIOS_CHASSIS_KIND_CONVERTIBLE	0x1F
+#define FU_SMBIOS_CHASSIS_KIND_DETACHABLE	0x20
+#define FU_SMBIOS_CHASSIS_KIND_IOT_GATEWAY	0x21
+#define FU_SMBIOS_CHASSIS_KIND_EMBEDDED_PC	0x22
+#define FU_SMBIOS_CHASSIS_KIND_MINI_PC		0x23
+#define FU_SMBIOS_CHASSIS_KIND_STICK_PC		0x24
+
```
superm1, line 58: This seems like it's asking for an enum rather than a list of devices.

src/fu-engine.c
```diff
@@ -5771,6 +5771,52 @@ fu_engine_ensure_security_attrs_supported (FuEngine *self)
...
+	g_autoptr(GError) error = NULL;
+
+	/* get chassis type from SMBIOS data */
+	val = fu_smbios_get_integer (self->smbios,
+				     FU_SMBIOS_STRUCTURE_TYPE_CHASSIS,
+				     0x05, &error);
+	if (val == G_MAXUINT) {
+		g_warning ("failed to get chassis type: %s", error->message);
+		return g_strdup ("HSI-INVALID:chassis");
+	}
```
superm1, line 5787: can't this be part of the switch/case?

superm1 (approved):

libfwupdplugin/fu-smbios.h
hughsie: Fair point, fixed.

src/fu-engine.c
hughsie: I think i'd rather handle failure with something setting the GError before we get into the "filtering" switch section.

PR #2491, opened 2020-10-20
Add fu_byte_array_set_size()

hughsie:
The GLib g_byte_array_set_size() function does not zero the contents if the
array size is larger, which leads to unpredictable output when using valgrind.

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -2001,6 +2001,24 @@ fu_byte_array_append_uint32 (GByteArray *array, guint32 data, FuEndianType endia
 }
 
 /**
+ * fu_byte_array_set_size:
+ * @array: a #GByteArray
+ * @length:  the new size of the GByteArray
+ *
+ * Sets the size of the GByteArray, expanding it with NULs if necessary.
+ *
+ * Since: 1.5.0
+ **/
+void
+fu_byte_array_set_size (GByteArray *array, guint length)
+{
+	guint oldlength = array->len;
+	g_byte_array_set_size (array, length);
+	if (length > oldlength)
+		memset (array->data + oldlength, 0x0, length - oldlength);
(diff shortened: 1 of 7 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -2000,6 +2000,25 @@ fu_byte_array_append_uint32 (GByteArray *array, guint32 data, FuEndianType endia
...
+ * @array: a #GByteArray
+ * @length:  the new size of the GByteArray
+ * @fillchar: the byte value to use when expanding the array
+ *
+ * Sets the size of the GByteArray, expanding it if necessary.
+ *
+ * Since: 1.5.0
+ **/
+void
+fu_byte_array_set_size (GByteArray *array, guint length, guint8 fillchar)
```
superm1, line 2014: Is there really a use case to fill the grown area with anything other than `NULL`?

hughsie: Flash is normally initialized with 0xff, but I agree in this case it's 100% redundant. Shall I remove `fillchar`?

superm1: Yes

hughsie: Done.

superm1 (approved):
To be honest, this feels like it should be a behavioral thing that `g_byte_array_set_size` should do to avoid problems with `valgrind` in the first place.  But as it doesn't, this PR looks good to me.

hughsie:
The definition of the GLib function is:

    GByteArray*
    g_byte_array_new (void)
    {
      return (GByteArray *)g_array_sized_new (FALSE, FALSE, 1, 0);
    }

It's frustrating it's not configurable, (e.g. `g_array_sized_new (FALSE, TRUE, ...)` but c'est la vie.

PR #2495, opened 2020-10-21
hailuck: Add an outline plugin for the Pinebook Pro laptop

hughsie:
Somewhat incomplete, untested, but 90% there.


superm1 (commented):
the sleeps all seem a big long, but nothing else stood out to me

plugins/hailuck/fu-plugin-hailuck.c
```diff
@@ -0,0 +1,21 @@
...
+
+#include "fu-hailuck-kbd-device.h"
+#include "fu-hailuck-kbd-firmware.h"
+
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+	fu_plugin_add_firmware_gtype (plugin, "hailuck", FU_TYPE_HAILUCK_KBD_FIRMWARE);
+	fu_plugin_set_device_gtype (plugin, FU_TYPE_HAILUCK_KBD_DEVICE);
```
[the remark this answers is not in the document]
hughsie: I think so; the TP is a logical child of the KBD device, so it's not something we ever want to build from a quirk file. What I might do is split up the Kbd device into bootloader and runtime objects as I think that would simplify things. Then I'd need both device types here...

plugins/hailuck/README.md
```diff
@@ -0,0 +1,31 @@
...
+---------------
+
+These devices use the standard USB DeviceInstanceId values, e.g.
+
+ * `USB\VID_0603&PID_1020`
+
+Vendor ID Security
+------------------
+
+The vendor ID is set from the USB vendor, in this instance set to `USB:0x0603`
```
superm1, line 31: I think need to add what interfaces are needed to this

hughsie: Agree, fixed, thanks.

plugins/hailuck/fu-plugin-hailuck.c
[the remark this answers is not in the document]
hughsie: I split this out like we thought.

plugins/hailuck/fu-hailuck-bl-device.c
```diff
@@ -0,0 +1,272 @@
...
+	fw_new = fu_hailuck_bl_device_dump_firmware (device, error);
+	return fu_common_bytes_compare (fw, fw_new, error);
+}
+
+static void
+fu_hailuck_bl_device_init (FuHaiLuckBlDevice *self)
+{
+	fu_device_set_firmware_size (FU_DEVICE (self), 0x4000);
+	fu_device_set_protocol (FU_DEVICE (self), "com.hailuck.kbd");
+	fu_device_set_name (FU_DEVICE (self), "Touchpad [bootloader]");
```
superm1, line 252: I thought this is a keyboard device

superm1 (approved):
Being this is a keyboard and mouse, does any care need to be made to not touch them while in use?  Messaging need to be sent for that?

plugins/hailuck/fu-hailuck-bl-device.c
hughsie: Good catch!

hughsie:
> does any care need to be made to not touch them while in use

No, they're completely non-functional.

PR #2497, opened 2020-10-21
trivial: match all elan devices in quirk

### LINKED ISSUE #2494: Auto detect Elan hid-i2c touchpad with VID 04f3 / PID 3XXX

superm1:
Here's what I had in mind.  Untested:

In probe routine, check whether they are actually i2c-hid touchpad
by looking at pid.

fixes #2494

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/elantp/elantp.quirk
+++ b/plugins/elantp/elantp.quirk
@@ -1,12 +1,4 @@
-[DeviceInstanceId=HIDRAW\VEN_04F3&DEV_3010]
-Plugin = elantp
-GType = FuElantpHidDevice
-
-[DeviceInstanceId=HIDRAW\VEN_04F3&DEV_30C5]
-Plugin = elantp
-GType = FuElantpHidDevice
-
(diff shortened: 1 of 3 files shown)


hughsie:
Hmm, so this works, but I get the additional warning in the daemon log:

    14:51:46:0001 FuEngine             failed to add udev device /sys/devices/pci0000:00/0000:00:14.0/usb1/1-6/1-6:1.0/0003:04F3:2219.0001/hidraw/hidraw0: failed to read HID ID: permission denied

....but then I worked out we need to filter in the HID probe, not the i2c recovery. Fixes pushed, thanks.

superm1:
LGTM

Issue #2391, opened 2020-09-14
TPM plugin should report version info in the LVFS uploaded report.


hughsie:
At the moment we don't know what mode the TPM is actually in.

PR #2502, opened 2020-10-22
Record the TPM version when reporting a UEFI update

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2391

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm/fu-plugin-tpm.c
+++ b/plugins/tpm/fu-plugin-tpm.c
@@ -29,9 +29,12 @@ void
 fu_plugin_device_added (FuPlugin *plugin, FuDevice *dev)
 {
 	FuPluginData *data = fu_plugin_get_data (plugin);
+	const gchar *family = fu_tpm_device_get_family (FU_TPM_DEVICE (dev));
+
 	data->has_tpm = TRUE;
-	if (g_strcmp0 (fu_tpm_device_get_family (FU_TPM_DEVICE (dev)), "2.0") == 0)
+	if (g_strcmp0 (family, "2.0") == 0)
 		data->has_tpm_v20 = TRUE;
+	fu_plugin_add_report_metadata (plugin, "TpmFamily", family);
 }
 
 void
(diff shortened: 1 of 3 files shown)


superm1 (commented):
I think you'll want to do the same for the Dell plugin since TPM plugin doesn't run on Dell machines normally.

plugins/tpm/fu-plugin-tpm.c
```diff
@@ -29,9 +29,12 @@ void
 fu_plugin_device_added (FuPlugin *plugin, FuDevice *dev)
 {
 	FuPluginData *data = fu_plugin_get_data (plugin);
+	const gchar *version = fu_tpm_device_get_family (FU_TPM_DEVICE (dev));
+
 	data->has_tpm = TRUE;
-	if (g_strcmp0 (fu_tpm_device_get_family (FU_TPM_DEVICE (dev)), "2.0") == 0)
+	if (g_strcmp0 (version, "2.0") == 0)
 		data->has_tpm_v20 = TRUE;
+	fu_plugin_add_report_metadata (plugin, "TpmVersion", version);
```
superm1, line 37: This is really TPM family not TPM version no?

hughsie: I have no idea why they call it a family. It's a string like "1.2" or "2.0"

hughsie: Maybe the report metadata should also be `TpmFamily` to avoid confusion?

superm1: >I have no idea why they call it a family. It's a string like "1.2" or "2.0"

Well it's indicating how it behaves, not a firmware version.

>Maybe the report metadata should also be TpmFamily to avoid confusion?

Yes

hughsie: Fair enough, agreed. I've pushed something that might pass muster.

superm1 (approved):

Issue #2376, opened 2020-09-10
vli: Hyper+Lenovo 07 TR Gen2


memily:
**Describe the bug**
We’re soooo, very, terribly, sorry…
 
As a sanity check we went back and tried to update/downgrade Proj 07, Lenovo Travel Gen2 (VL817+VL103Q4) and the BB/VL103 PD is not updating on that device either.  In theory the BB lives inside the VL817 hub and can only be accessed by the hub. Going along that assumption, fu-vli-usbhub-device.c, function “fu_vli_usbhub_device_update_v2()” should turn over control to "fu-vli-usbhub-pd-device.c" for writing pd fw, we think? Also hd2_fw_addr and hd2_fw_offset seem to be returning the wrong values for PD addr and offset (0x9000 and 0x2000). 

**Steps to Reproduce**
Attempt to fwupdtool update/downgrade install Lenovo TR Gen2 hub for sanity check, with manual power cycling. (Should we add flag "skips-restart" for PD child in quirk?)

**Expected behavior**
Hub+PD update...

**fwupd version information**
1.4.1 (dnf, most recent pull)

**fwupd device information**
Included updated hyper verbose log for reference as needed since it was same as TR Gen2.
Sorry! >x<
[verbose.txt](https://github.com/fwupd/fwupd/files/5199187/verbose.txt)

hughsie:
> fwupd version information
> 1.4.1

Are you sure you're using 1.4.1? If so you're missing out on a lot of fixes to the VLI plugin:

    git diff 1.4.1..1.4.6 ../plugins/vli/ | diffstat
     README.md                      |    3 
     fu-vli-common.c                |    4 +
     fu-vli-common.h                |    1 
     fu-vli-device.c                |    1 
     fu-vli-pd-device.c             |  167 +++++++++++++++++++++++++++++-------------------
     fu-vli-usbhub-device.c         |  216 ++++++++++++++++++++++++++++++++------------------------------
     fu-vli-usbhub-device.h         |   21 ------
     fu-vli-usbhub-i2c-common.h     |   12 ---
     fu-vli-usbhub-i2c-device.h     |   19 -----
     fu-vli-usbhub-msp430-device.c  |  192 +++++++++++++++++++++++++++++++++++++------------------
     fu-vli-usbhub-msp430-device.h  |   19 +++++
     fu-vli-usbhub-pd-device.c      |   18 +++++
[...]

PR #2506, opened 2020-10-23
vli:small hyper support edits

memily:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

HYPER edits to vli plugin, discussed in #2376

### DIFF
--- a/plugins/vli/fu-vli-pd-firmware.c
+++ b/plugins/vli/fu-vli-pd-firmware.c
@@ -48,6 +48,8 @@ fu_vli_pd_firmware_validate_header (FuVliPdFirmware *self)
 		return TRUE;
 	if (GUINT16_FROM_LE (self->hdr.vid) == 0x17EF)
 		return TRUE;
+	if (GUINT16_FROM_LE (self->hdr.vid) == 0x2D01)
+		return TRUE;
 	return FALSE;
 }
 
--- a/plugins/vli/meson.build
+++ b/plugins/vli/meson.build
@@ -4,6 +4,7 @@ install_data([
   'vli-pd.quirk',
   'vli-usbhub.quirk',
   'vli-usbhub-lenovo.quirk',
+  'vli-usbhub-hyper.quirk',
   ],
   install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
 )
--- a/plugins/vli/vli-usbhub-hyper.quirk
+++ b/plugins/vli/vli-usbhub-hyper.quirk
@@ -1,0 +1,11 @@
+# Hyper USB-C Hub
+[DeviceInstanceId=USB\VID_2D01&PID_0385]
+Plugin = vli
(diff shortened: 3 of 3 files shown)


plugins/vli/fu-vli-common.h
```diff
@@ -37,6 +37,7 @@ typedef enum {
 	FU_VLI_DEVICE_KIND_MSP430		= 0xf430, /* guessed */
 	FU_VLI_DEVICE_KIND_PS186		= 0xf186, /* guessed */
 	FU_VLI_DEVICE_KIND_RTD21XX		= 0xff00, /* guessed */
+	FU_VLI_DEVICE_KIND_HYPER		= 0x0817,
```
hughsie, line 40: this is just an enum value, I'll fix it up to 0xff01 for you.

plugins/vli/fu-vli-common.c
```diff
@@ -64,6 +64,10 @@ fu_vli_common_device_kind_to_string (FuVliDeviceKind device_kind)
 		return "PS186";
 	if (device_kind == FU_VLI_DEVICE_KIND_RTD21XX)
 		return "RTD21XX";
+	if (device_kind == FU_VLI_DEVICE_KIND_RTD21XX)
+		return "RTD21XX";
```
hughsie, line 68: seems repeated no?

plugins/vli/fu-vli-common.c
```diff
@@ -174,6 +180,8 @@ fu_vli_common_device_kind_get_size (FuVliDeviceKind device_kind)
 		return 0x20000 * 2;
 	if (device_kind == FU_VLI_DEVICE_KIND_PS186)
 		return 0x40000;
+	if (device_kind == FU_VLI_DEVICE_KIND_HYPER)
+		return 0x20000 * 2;
```
hughsie, line 184: Is hyper the deice being programmed over i2c? it's normally the "chip ID" of the i2c thing we use here e.g. PS186 rather than PARADE. If it's just a shared PD I don't think you need the FU_VLI_DEVICE_KIND_HYPER define at all.

plugins/vli/fu-vli-common.h
memily: Thank you!

plugins/vli/fu-vli-common.c
memily: Oops, looks like it yes, sorry.

plugins/vli/fu-vli-common.c
memily: We're not sure what line up Hyper has in mind for the future, but I believe it's safe to remove it since our intention was to cover as many bases --just in case.

hughsie: Okay, I've removed it; I think it's confusing to have an enum defined that's not actually used.

PR #2509, opened 2020-10-23
trivial: Split check for signed metadata when no polkit

superm1:
Checking this requirement at daemon startup meant that trust was
evaluated before releases could actually be checked.

It's only important to check at install time.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1441,6 +1441,21 @@ fu_engine_check_requirement (FuEngine *self,
 }
 
 gboolean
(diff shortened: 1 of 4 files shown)


superm1:
On the bright side, less special casing in the self tests.

superm1:
@djcampello indicated that other commit is no longer necessary.

djcampello (approved):

Issue #2510, opened 2020-10-25
bcm57xx plugin no longer detects updatable devices


meklort:
**Describe the bug**
As of commit f2a3e01f96cfea2ac641559eb3011c3927fde328, I am unable to detect updatable devices with the bcm57xx plugin.

**Steps to Reproduce**
Run 
```shell
fwupdtool get-devices
```

**Expected behavior**
fwupdtool should list devices that were properly detected by the bcm57xx plugin as updatable.

**fwupd version information**
master, built from source

**fwupd device information**
Note: The device is iniitally detected as updatable, but later on does not retain the updatable flag
```shell
meklort@talos2 ~/g/f/build ((f2a3e01f…)|BISECTING)> sudo ./src/fwupdtool get-devices --show-all-devices -v
02:38:25:0639 FuDebug              Verbose debugging enabled (on console 1)
02:38:25:0649 FuConfig             loading config values from /etc/fwupd/daemon.conf
02:38:25:0650 GLib-GIO             _g_io_module_get_default: Found default implementation local (GLocalVfs) for ‘gio-vfs’
02:38:25:0650 FuRemoteList         loading /share/fwupd/metainfo
02:38:25:0651 XbSilo               attempting to load /var/cache/fwupd/metainfo.xmlb
02:38:25:0651 XbSilo               file: 40b38459-9b28-c560-d6f6-71b35128a57a, current:26bd47fc-5125-89de-613d-bbfb7c947720, cached: (null)
02:38:25:0651 XbSilo               compiling /share/fwupd/metainfo/org.freedesktop.fwupd.remotes.lvfs-testing.metainfo.xml:ctime=1603593488.965879…
02:38:25:0651 XbSilo               compiling /share/fwupd/metainfo/org.freedesktop.fwupd.remotes.lvfs.metainfo.xml:ctime=1603593488.965879…
02:38:25:0659 FuRemoteList         loading remote from /etc/fwupd/remotes.d/evan_lojewski-embargo.conf
02:38:25:0659 Fwupd                converting Keyring value to Jcat
02:38:25:0659 FuRemoteList         skipping invalid file /etc/fwupd/remotes.d/vendor-directory.conf.rpmsave
02:38:25:0659 FuRemoteList         loading remote from /etc/fwupd/remotes.d/lvfs-testing.conf
02:38:25:0660 FuRemoteList         skipping invalid file /etc/fwupd/remotes.d/vendor.conf.rpmnew
[...]

PR #2511, opened 2020-10-25
Fix regression when filtering different priorities

hughsie:
We need to check if the device is in the remove timeout before unconditionally
replacing a device on insert.

Fixes https://github.com/fwupd/fwupd/issues/2510


hughsie:
Untested until Monday.

meklort:
I tried running this locally on my machine and can confirm that it fixes #2510

PR #2517, opened 2020-10-26
Enable the FWUPDATE_VERBOSE variable when deploying the update

hughsie:
This allows us to delete the previous log if debugging is now disabled.

### DIFF
--- a/plugins/uefi/fu-uefi-device.c
+++ b/plugins/uefi/fu-uefi-device.c
@@ -571,6 +571,10 @@ fu_uefi_device_write_firmware (FuDevice *device,
 	if (!fu_common_set_contents_bytes (fn, fixed_fw, error))
 		return FALSE;
 
+	/* delete the logs to save space; use fwupdate to debug the EFI binary */
+	fu_efivar_delete (FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_VERBOSE", NULL);
+	fu_efivar_delete (FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_DEBUG_LOG", NULL);
+
 	/* set the blob header shared with fwupd.efi */
 	if (!fu_uefi_device_write_update_info (self, fn, varname, self->fw_class, error))
 		return FALSE;


plugins/uefi/uefi.conf
```diff
@@ -1,5 +1,9 @@
 [uefi]
 
+# the EFI binary can output a lot of debugging to an NVRAM variable, but it is
+# slow and will eventually cause failure of the SPI device itself
```
superm1, line 4: That's a little overconfident.  How about _may_.

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -439,6 +440,10 @@ fu_plugin_uefi_load_config (FuPlugin *plugin, FuDevice *device)
 	fu_device_set_metadata_boolean (device,
 					"RequireShimForSecureBoot",
 					!disable_shim);
+
+	/* only used to debug the EFI binary itself */
+	efi_debug_log = fu_plugin_get_config_value_boolean (plugin, "EfiDebugLog");
```
superm1, line 445: Hmm rather than a new variable, why not just hook off `FWUPD_UEFI_VERBOSE` key that is already used to do verbose stuff and has the extra trigger from `fwupdtool --plugin-verbose=uefi`

superm1 (commented):
I think it's definitely step in the right direction to not use the NVRAM variable storage in 99% of the cases.

plugins/uefi/fu-plugin-uefi.c
hughsie: My first thought was that we should retain compatibility with the `fwupdate -D` and `-d` mode where QA users can just "_turn on_" extra debugging on a machine. I guess the alternate is we just explain to the user that the `-d` mode has been removed and that they should do fwupdmgr --enable-extra-debugging or something -- that would be a FuInstallFlag I guess, although I don't know if that's over-engineering the whole thing. Nowadays the EFI debugging is used less and less now we know where the bodies are hidden.

superm1: I guess your point being we need the debugging to be available from `fwupdgmr`?  I'd say that's overengineering.  If someone has a problem (which it's been so long since we needed it), send them over to `fwupdate` to install it or or `fwupdtool --plugin-verbose=uefi`.

hughsie: So maybe we simplify it further, and just unconditionally delete FWUPDATE_VERBOSE and FWUPDATE_DEBUG_LOG when deploying the update using fwupdtool/fwupd and leave it enabled and readable by fwupdate -- if you want to debug the efi binary you then have to use fwupdate and squirt the .cap file rather than using all the other layers.

superm1: Yes, that's great idea.

hughsie: Something like this^^?

superm1 (approved):

PR #2518, opened 2020-10-27
Fix probe warning for the Logitech Unifying device

hughsie:
The same plugin name was being added to the device from the quirk file more than
once, and so we enumerated the device *again* and tried to add a duplicate
device -- the device list correctly refusing to do so.

Check the plugin name does not already exist before adding it, and add a self
test to catch this for the future.


superm1:
> Check the plugin name does not already exist before adding it, and add a self
test to catch this for the future.

Do you have an idea when this started to happen?  It seems like it's preferable to apply to the fixes branches for earlier releases if applicable.

hughsie:
> Do you have an idea when this started to happen

I only noticed it on my 1.5.0 flatpak build; I think we could apply just the g_ptr_array_find lines to 1_4_X

superm1:
Sounds good to me.

PR #2519, opened 2020-10-27
Include the amount of NVRAM size in use in the LVFS failure report

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-efivar.c
+++ b/libfwupdplugin/fu-efivar.c
@@ -396,6 +396,51 @@ fu_efivar_get_names (const gchar *guid, GError **error)
 }
 
 /**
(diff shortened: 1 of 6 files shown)


superm1 (approved):
Presumably you want to _also_ collect the data from the uefi-dbx plugin?

hughsie:
> Presumably you want to also collect the data from the uefi-dbx plugin?

Ahh yes, good catch. I guess we can spinkle in a `fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "uefi");` in uefi-dbx

superm1:
LGTM

PR #2520, opened 2020-10-27
trivial: Convert update status in json state file to string

djcampello:
This makes for more clear debugging while looking at the json state file.

### DIFF
--- a/libfwupd/fwupd-device.c
+++ b/libfwupd/fwupd-device.c
@@ -2134,7 +2134,8 @@ fwupd_device_to_json (FwupdDevice *device, JsonBuilder *builder)
 	fwupd_device_json_add_int (builder, FWUPD_RESULT_KEY_INSTALL_DURATION, priv->install_duration);
 	fwupd_device_json_add_int (builder, FWUPD_RESULT_KEY_CREATED, priv->created);
 	fwupd_device_json_add_int (builder, FWUPD_RESULT_KEY_MODIFIED, priv->modified);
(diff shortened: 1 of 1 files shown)


superm1 (approved):

djcampello:
@shiyu-s this should help your debugging

hughsie:
Not so fast: https://gitlab.com/fwupd/lvfs-website/-/blob/master/lvfs/reports/routes.py#L257 -- the LVFS requires an integer. We can workaround this on the LVFS and convert it if required, but I'd like to know if we can do better showing the user something else.

superm1:
> Not so fast: https://gitlab.com/fwupd/lvfs-website/-/blob/master/lvfs/reports/routes.py#L257 -- the LVFS requires an integer. We can workaround this on the LVFS and convert it if required, but I'd like to know if we can do better showing the user something else.

Oh...  The other thing we can do is to update this in https://github.com/fwupd/fwupd/blob/master/src/fu-tool.c#L76 when it spits out the state file.

hughsie:
> when it spits out the state file.

I'd much rather have a command that reads in the JSON dump and shows the user something with translations and that kind of thing, if we want the user to read the actual machine-parsed json file.

djcampello:
> > when it spits out the state file.
> 
> I'd much rather have a command that reads in the JSON dump and shows the user something with translations and that kind of thing, if we want the user to read the actual machine-parsed json file.

I agree, I will drop this PR.

PR #2529, opened 2020-10-29
Use UDisks to find out if swap devices are encrypted

hughsie:
Using a heuristic is certainly not awesome.

(I'm also not sure how to test this in the self tests....)

### DIFF
--- a/plugins/linux-swap/fu-linux-swap.c
+++ b/plugins/linux-swap/fu-linux-swap.c
@@ -13,11 +13,82 @@
 
 struct _FuLinuxSwap {
 	GObject		 parent_instance;
-	gboolean	 encrypted;
-	gboolean	 enabled;
+	guint		 encrypted_cnt;
+	guint		 enabled_cnt;
 };
 
 G_DEFINE_TYPE (FuLinuxSwap, fu_linux_swap, G_TYPE_OBJECT)
+
+static gchar *
+fu_strdup_nospaces (const gchar *line)
+{
+	GString *str = g_string_new (NULL);
+	for (guint i = 0; line[i] != '\0' && !g_ascii_isspace (line[i]); i++)
+		g_string_append_c (str, line[i]);
+	return g_string_free (str, FALSE);
+}
+
+static gboolean
+fu_linux_swap_verify_partition (FuLinuxSwap *self, const gchar *fn, GError **error)
+{
+	g_autoptr(FuVolume) volume = NULL;
+
+	/* find the device */
+	volume = fu_common_get_volume_by_device (fn, error);
+	if (volume == NULL)
+		return FALSE;
+
+	/* this isn't technically encrypted, but isn't on disk in plaintext */
+	if (g_str_has_prefix (fn, "/dev/zram")) {
+		g_debug ("%s is zram, assuming encrypted", fn);
+		self->encrypted_cnt++;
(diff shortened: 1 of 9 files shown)


plugins/linux-swap/fu-linux-swap.c
```diff
@@ -13,46 +13,256 @@
...
+					  -1, NULL, error);
+	if (output == NULL)
+		return NULL;
+	devices = g_ptr_array_new_with_free_func (g_free);
+	g_variant_get (output, "(ao)", &iter);
+	while (g_variant_iter_next (iter, "&o", &obj))
+		g_ptr_array_add (devices, g_strdup (obj));
+
+	return g_steal_pointer (&devices);
+}
```
superm1, line 76: this should go into `fu-common.c` where we have already used udisks for siimilar tasks

superm1:
So if switching to using udisks2 for the backing, why not just have the plugin walk all the block devices udisks2 knows about and look for `http://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Swapspace.html`?  I guess I don't know if `file` swap shows in that though would be the one potential catch.

hughsie:
> I don't know if file swap shows in that though would be the one potential catch.

Exactly that, no swap files.

plugins/linux-swap/fu-linux-swap.c
hughsie: Refactoring now.

hughsie: Okay, how about something like that.

plugins/linux-swap/fu-linux-swap.c
```diff
@@ -13,46 +13,128 @@
...
+			if (g_strcmp0 (ty, "partition") == 0) {
+				self->enabled_cnt++;
+				if (!fu_linux_swap_verify_partition (self, fn, error))
+					return NULL;
+			} else if (g_strcmp0 (ty, "file") == 0) {
+				self->enabled_cnt++;
+				if (!fu_linux_swap_verify_file (self, fn, error))
+					return NULL;
+			} else {
+				g_warning ("unknown swap type: %s [%s]", ty, fn);
```
superm1, line 118: so with this refactor I was going to suggest all the swap should be evaluted when attributes are evaluated.  Although people don't usually change it, this isn't static information.

superm1:
New approach is great.

plugins/linux-swap/fu-linux-swap.c
hughsie: Agree. I think the FuLinuxSwap gets recreated whenever we change /proc/swaps, right?

superm1: Oh yeah it does.. I completely forgot we have `fu_plugin_linux_swap_changed_cb ` without that in front of me.  Yes carry on.

superm1 (approved):

hughsie:
@superm1 I see a Fedora crash in CI... is that a daemon segfault? Help!

superm1:
> @superm1 I see a Fedora crash in CI... is that a daemon segfault? Help!

Looks like you got it sorted, right?

And the Debian CI failure it's an upstream TSS bug: https://github.com/tpm2-software/tpm2-tss/issues/1876

PR #2534, opened 2020-10-30
uefi: a new option for uefi configuration to use UEFI removable path

woodrow-shen:
Per discussion of #2513, Ubuntu Core was going to use UEFI removable
path as default esp path, and it needs to change some parts around
getting the esp path and searching the shim app path. Also, a new option
"FallbacktoRemovablePath" is added into uefi.conf to be applied in this
case, and it will be false by default.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -241,7 +241,7 @@ fu_plugin_uefi_write_splash_data (FuPlugin *plugin,
 
 	/* save to a predicatable filename */
 	esp_path = fu_volume_get_mount_point (data->esp);
-	directory = fu_uefi_get_esp_path_for_os (esp_path);
+	directory = fu_uefi_get_esp_path_for_os (device, esp_path);
 	basename = g_strdup_printf ("fwupd-%s.cap", FU_EFIVAR_GUID_UX_CAPSULE);
 	fn = g_build_filename (directory, "fw", basename, NULL);
 	if (!fu_common_mkdir_parent (fn, error))
@@ -425,6 +425,7 @@ static void
 fu_plugin_uefi_load_config (FuPlugin *plugin, FuDevice *device)
 {
 	gboolean disable_shim;
+	gboolean fallback_removable_path;
 	guint64 sz_reqd = FU_UEFI_COMMON_REQUIRED_ESP_FREE_SPACE;
 	g_autofree gchar *require_esp_free_space = NULL;
 
@@ -439,6 +440,12 @@ fu_plugin_uefi_load_config (FuPlugin *plugin, FuDevice *device)
 	fu_device_set_metadata_boolean (device,
 					"RequireShimForSecureBoot",
 					!disable_shim);
+
+	/* check if using UEFI removeable path */
+	fallback_removable_path = fu_plugin_get_config_value_boolean (plugin, "FallbacktoRemovablePath");
+	fu_device_set_metadata_boolean (device,
+					"FallbacktoRemovablePath",
+					fallback_removable_path);
 }
 
 static void
--- a/plugins/uefi/fu-uefi-bootmgr.c
+++ b/plugins/uefi/fu-uefi-bootmgr.c
@@ -239,10 +239,11 @@ fu_uefi_copy_asset (const gchar *source, const gchar *target, GError **error)
 }
 
 gboolean
-fu_uefi_bootmgr_bootnext (const gchar *esp_path,
-			  const gchar *description,
-			  FuUefiBootmgrFlags flags,
-			  GError **error)
+fu_uefi_bootmgr_bootnext (FuDevice *device,
+			   const gchar *esp_path,
+			   const gchar *description,
+			   FuUefiBootmgrFlags flags,
+			   GError **error)
 {
 	const gchar *filepath;
 	gboolean use_fwup_path = TRUE;
@@ -273,14 +274,23 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
 	secure_boot = fu_efivar_secure_boot_enabled ();
 	if (secure_boot) {
 		/* test to make sure shim is there if we need it */
-		shim_app = fu_uefi_get_esp_app_path (esp_path, "shim", error);
+		shim_app = fu_uefi_get_esp_app_path (device, esp_path, "shim", error);
 		if (shim_app == NULL)
 			return FALSE;
+
+		/* try to fallback to use UEFI removable path if the shim path doesn't exist */
+		if (!g_file_test (shim_app, G_FILE_TEST_EXISTS)) {
+			if (fu_device_get_metadata_boolean (device, "FallbacktoRemovablePath")) {
+				shim_app = fu_uefi_get_esp_app_path (device, esp_path, "boot", error);
+				if (shim_app == NULL)
+					return FALSE;
+			}
+		}
 
 		if (g_file_test (shim_app, G_FILE_TEST_EXISTS)) {
 			/* use a custom copy of shim for firmware updates */
 			if (flags & FU_UEFI_BOOTMGR_FLAG_USE_SHIM_UNIQUE) {
-				shim_cpy = fu_uefi_get_esp_app_path (esp_path, "shimfwupd", error);
+				shim_cpy = fu_uefi_get_esp_app_path (device, esp_path, "shimfwupd", error);
 				if (shim_cpy == NULL)
 					return FALSE;
 				if (!fu_uefi_cmp_asset (shim_app, shim_cpy)) {
@@ -302,7 +312,7 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
 	}
 
 	/* test if correct asset in place */
-	target_app = fu_uefi_get_esp_app_path (esp_path, "fwupd", error);
+	target_app = fu_uefi_get_esp_app_path (device, esp_path, "fwupd", error);
 	if (target_app == NULL)
 		return FALSE;
 	if (!fu_uefi_cmp_asset (source_app, target_app)) {
--- a/plugins/uefi/fu-uefi-bootmgr.h
+++ b/plugins/uefi/fu-uefi-bootmgr.h
@@ -10,6 +10,8 @@
 #include <glib.h>
 #include <efivar.h>
 
+#include "fu-device.h"
+
 typedef enum {
 	FU_UEFI_BOOTMGR_FLAG_NONE		= 0,
 	FU_UEFI_BOOTMGR_FLAG_USE_SHIM_FOR_SB	= 1 << 0,
@@ -17,7 +19,8 @@ typedef enum {
 	FU_UEFI_BOOTMGR_FLAG_LAST
 } FuUefiBootmgrFlags;
 
-gboolean	 fu_uefi_bootmgr_bootnext	(const gchar		*esp_path,
+gboolean	 fu_uefi_bootmgr_bootnext	(FuDevice 		*device,
+						 const gchar		*esp_path,
 						 const gchar		*description,
 						 FuUefiBootmgrFlags	 flags,
 						 GError			**error);
--- a/plugins/uefi/fu-uefi-common.c
+++ b/plugins/uefi/fu-uefi-common.c
@@ -61,13 +61,16 @@ fu_uefi_bootmgr_get_suffix (GError **error)
 }
 
 gchar *
-fu_uefi_get_esp_app_path (const gchar *esp_path, const gchar *cmd, GError **error)
+fu_uefi_get_esp_app_path (FuDevice *device,
+			   const gchar *esp_path,
+			   const gchar *cmd,
+			   GError **error)
 {
 	const gchar *suffix = fu_uefi_bootmgr_get_suffix (error);
 	g_autofree gchar *base = NULL;
 	if (suffix == NULL)
 		return NULL;
-	base = fu_uefi_get_esp_path_for_os (esp_path);
+	base = fu_uefi_get_esp_path_for_os (device, esp_path);
 	return g_strdup_printf ("%s/%s%s.efi", base, cmd, suffix);
 }
 
@@ -219,11 +222,11 @@ fu_uefi_get_esrt_entry_paths (const gchar *esrt_path, GError **error)
 }
 
 gchar *
-fu_uefi_get_esp_path_for_os (const gchar *base)
+fu_uefi_get_esp_path_for_os (FuDevice *device, const gchar *base)
 {
 #ifndef EFI_OS_DIR
 	const gchar *os_release_id = NULL;
-	const gchar *id_like_id;
+	const gchar *id_like_id = NULL;
 	g_autofree gchar *esp_path = NULL;
 	g_autoptr(GError) error_local = NULL;
 	g_autoptr(GHashTable) os_release = fwupd_get_os_release (&error_local);
@@ -247,6 +250,12 @@ fu_uefi_get_esp_path_for_os (const gchar *base)
 			g_debug ("Using ID_LIKE key from os-release");
 			return g_steal_pointer (&id_like_path);
 		}
+	} 
+	/* try to fallback to use UEFI removable path if ID_LIKE path doesn't exist */
+	if (fu_device_get_metadata_boolean (device, "FallbacktoRemovablePath")) {
+		esp_path = g_build_filename (base, "EFI", "boot", NULL);
+		if (!g_file_test (esp_path, G_FILE_TEST_IS_DIR))
+			g_debug ("failed to fallback due to missing %s", esp_path);
 	}
 	return g_steal_pointer (&esp_path);
 #else
--- a/plugins/uefi/uefi.conf
+++ b/plugins/uefi/uefi.conf
@@ -10,3 +10,7 @@
 
 # amount of free space required on the ESP, for example using 0x2000000 for 32Mb
 #RequireESPFreeSpace=
+
+# with the UEFI removable path enabled, the default esp path is set to /EFI/boot
+# the shim EFI binary and presumably this is $ESP/EFI/boot/bootx64.efi
+#FallbacktoRemovablePath=false
--- a/plugins/uefi/fu-uefi-common.h
+++ b/plugins/uefi/fu-uefi-common.h
@@ -11,6 +11,8 @@
 
 #include "fwupd-common.h"
 
+#include "fu-device.h"
+
 #define EFI_CAPSULE_HEADER_FLAGS_PERSIST_ACROSS_RESET	0x00010000
 #define EFI_CAPSULE_HEADER_FLAGS_POPULATE_SYSTEM_TABLE	0x00020000
 #define EFI_CAPSULE_HEADER_FLAGS_INITIATE_RESET		0x00040000
@@ -58,7 +60,8 @@ typedef struct __attribute__((__packed__)) {
 /* the biggest size SPI part currently seen */
 #define FU_UEFI_COMMON_REQUIRED_ESP_FREE_SPACE		(32 * 1024 * 1024)
 
-gchar		*fu_uefi_get_esp_app_path	(const gchar	*esp_path,
+gchar		*fu_uefi_get_esp_app_path	(FuDevice       *device,
+						 const gchar	*esp_path,
 						 const gchar	*cmd,
 						 GError		**error);
 gchar		*fu_uefi_get_built_app_path	(GError		**error);
@@ -70,7 +73,8 @@ gboolean	 fu_uefi_get_bitmap_size	(const guint8	*buf,
 gboolean	 fu_uefi_get_framebuffer_size	(guint32	*width,
 						 guint32	*height,
 						 GError		**error);
-gchar		*fu_uefi_get_esp_path_for_os	(const gchar	*esp_path);
+gchar		*fu_uefi_get_esp_path_for_os	(FuDevice 	*device,
+						 const gchar	*esp_path);
 GPtrArray	*fu_uefi_get_esrt_entry_paths	(const gchar	*esrt_path,
 						 GError		**error);
 guint64		 fu_uefi_read_file_as_uint64	(const gchar	*path,
--- a/plugins/uefi/fu-uefi-device.c
+++ b/plugins/uefi/fu-uefi-device.c
@@ -560,7 +560,7 @@ fu_uefi_device_write_firmware (FuDevice *device,
 		return FALSE;
 
 	/* save the blob to the ESP */
-	directory = fu_uefi_get_esp_path_for_os (esp_path);
+	directory = fu_uefi_get_esp_path_for_os (device, esp_path);
 	basename = g_strdup_printf ("fwupd-%s.cap", self->fw_class);
 	fn = g_build_filename (directory, "fw", basename, NULL);
 	if (!fu_common_mkdir_parent (fn, error))
@@ -588,7 +588,7 @@ fu_uefi_device_write_firmware (FuDevice *device,
 	/* some legacy devices use the old name to deduplicate boot entries */
 	if (fu_device_has_custom_flag (device, "use-legacy-bootmgr-desc"))
 		bootmgr_desc = "Linux-Firmware-Updater";
-	if (!fu_uefi_bootmgr_bootnext (esp_path, bootmgr_desc, flags, error))
+	if (!fu_uefi_bootmgr_bootnext (device, esp_path, bootmgr_desc, flags, error))
 		return FALSE;
 
 	/* success! */


plugins/uefi/fu-plugin-uefi.c
```diff
@@ -241,7 +241,7 @@ fu_plugin_uefi_write_splash_data (FuPlugin *plugin,
 
 	/* save to a predicatable filename */
 	esp_path = fu_volume_get_mount_point (data->esp);
-	directory = fu_uefi_get_esp_path_for_os (esp_path);
+	directory = fu_uefi_get_esp_path_for_os (device, esp_path);
```
superm1, line 244: if passing device into this function no need to be passing `esp_path`.  It can be read from device metadata.

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -439,6 +440,12 @@ fu_plugin_uefi_load_config (FuPlugin *plugin, FuDevice *device)
 	fu_device_set_metadata_boolean (device,
 					"RequireShimForSecureBoot",
 					!disable_shim);
+
+	/* check if using UEFI removeable path */
+	enable_removable_path = fu_plugin_get_config_value_boolean (plugin, "EnableUefiRemovablePath");
+	fu_device_set_metadata_boolean (device,
+					"RequireRemovablePath",
```
superm1, line 447: My proposal was to fallback to this if shim doesn't exist so confined snap could work on other distros than UC.  Is there a reason not to do this?

plugins/uefi/fu-uefi-bootmgr.h
```diff
@@ -10,14 +10,17 @@
...
+
 typedef enum {
 	FU_UEFI_BOOTMGR_FLAG_NONE		= 0,
 	FU_UEFI_BOOTMGR_FLAG_USE_SHIM_FOR_SB	= 1 << 0,
 	FU_UEFI_BOOTMGR_FLAG_USE_SHIM_UNIQUE	= 1 << 1,
 	FU_UEFI_BOOTMGR_FLAG_LAST
 } FuUefiBootmgrFlags;
 
-gboolean	 fu_uefi_bootmgr_bootnext	(const gchar		*esp_path,
+gboolean	 fu_uefi_bootmgr_bootnext	(FuDevice 		*device,
```
superm1, line 22: Maybe this should be `FuUefiDevice` and stop passing esp_path and read it from metadata too.

plugins/uefi/uefi.conf
```diff
@@ -10,3 +10,7 @@
 
 # amount of free space required on the ESP, for example using 0x2000000 for 32Mb
 #RequireESPFreeSpace=
+
+# with the UEFI removable path enabled, the default esp path is set to /EFI/boot
+# the shim EFI bianry needs to be found at /EFI/boot/bootx64.efi
```
superm1, line 15: **binary** and presumably this is `$ESP/EFI/boot/bootx64.efi`?

plugins/uefi/fu-plugin-uefi.c
```diff
@@ -439,6 +440,12 @@ fu_plugin_uefi_load_config (FuPlugin *plugin, FuDevice *device)
 	fu_device_set_metadata_boolean (device,
 					"RequireShimForSecureBoot",
 					!disable_shim);
+
+	/* check if using UEFI removeable path */
+	enable_removable_path = fu_plugin_get_config_value_boolean (plugin, "EnableUefiRemovablePath");
```
superm1, line 445: if there isn't a fallback, this should rename to `UseUefiRemovablePath`

plugins/uefi/fu-plugin-uefi.c
woodrow-shen: I may not understand your point. If `esp_path` can be read form metadata, does that mean it is provided by uefi.conf? Or `EnableUefiRemovablePath` should be a string not boolean? Can you please specify more precisely?

plugins/uefi/fu-plugin-uefi.c
woodrow-shen: No, I tried to follow up your approach, but seems I misunderstand what you pointed to me when I thought a new option should be loaded here. Well..I'm not very familiar with these codes.

plugins/uefi/uefi.conf
woodrow-shen: Sure. will be fixed.

plugins/uefi/fu-plugin-uefi.c
woodrow-shen: just need to clarify the first comment I left and fix this.

plugins/uefi/fu-plugin-uefi.c
superm1: Actually this one is correct, never mind.

plugins/uefi/fu-uefi-bootmgr.c
```diff
@@ -273,14 +274,17 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
 	secure_boot = fu_efivar_secure_boot_enabled ();
 	if (secure_boot) {
 		/* test to make sure shim is there if we need it */
-		shim_app = fu_uefi_get_esp_app_path (esp_path, "shim", error);
+		if (fu_device_get_metadata_boolean (device, "RequireRemovablePath"))
+			shim_app = fu_uefi_get_esp_app_path (device, esp_path, "boot", error);
+		else
+			shim_app = fu_uefi_get_esp_app_path (device, esp_path, "shim", error);
```
superm1, line 280: Here's an untested alternative proposal that would meet my suggestion.

```
shim_app = fu_uefi_get_esp_app_path (device, esp_path, "shim", &error_local);

if (shim_app == NULL) {
        if (fu_device_get_metadata_boolean (device, "RequireRemovablePath")) {
                shim_app = fu_uefi_get_esp_app_path (device, esp_path, "boot", error);
        else
                g_propagate_error (error, g_steal_pointer (&error_local));
        if (shim_app == NULL)
                return FALSE;
}
```

superm1: And in doing that, you probably should rename `RequireRemovablePath` to `FallbacktoRemovablePath`

plugins/uefi/fu-uefi-common.c
```diff
@@ -235,6 +238,12 @@ fu_uefi_get_esp_path_for_os (const gchar *base)
 	}
 	if (os_release_id == NULL)
 		os_release_id = "unknown";
+	/* use UEFI removable path as default if the option is set to true */
+	if (fu_device_get_metadata_boolean (device, "RequireRemovablePath")) {
+		esp_path = g_build_filename (base, "EFI", "boot", NULL);
+		if (g_file_test (esp_path, G_FILE_TEST_IS_DIR))
+			return g_steal_pointer (&esp_path);
+	}
```
superm1, line 246: so rather than forcing the path, my suggestion is to put a test at the end after this failed: 
```		
if (g_file_test (id_like_path, G_FILE_TEST_IS_DIR)) {
```

And then that means if an OS path exists it uses that, and if it doesn't it uses removable path if the flag is turned on.

plugins/uefi/fu-plugin-uefi.c
superm1: Take a look at my two new comments and I think it should be clearer.

plugins/uefi/fu-uefi-bootmgr.h
superm1: Never mind this one.

woodrow-shen:
In Ubuntu Core, the content of /etc/os-release didn't have the field "ID_LIKE" so I put the fallback check after that.

plugins/uefi/fu-uefi-bootmgr.c
```diff
@@ -273,14 +274,23 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
 	secure_boot = fu_efivar_secure_boot_enabled ();
 	if (secure_boot) {
 		/* test to make sure shim is there if we need it */
-		shim_app = fu_uefi_get_esp_app_path (esp_path, "shim", error);
+		shim_app = fu_uefi_get_esp_app_path (device, esp_path, "shim", error);
```
superm1, line 277: you should be using `error_local` here, you want to discard this error in your fallback scenario.

plugins/uefi/fu-uefi-common.c
```diff
@@ -241,12 +244,17 @@ fu_uefi_get_esp_path_for_os (const gchar *base)
...
 	/* if ID key doesn't exist, try ID_LIKE */
 	id_like_id = g_hash_table_lookup (os_release, "ID_LIKE");
+	/* try to fallback to use UEFI removable path if ID_LIKE key doesn't exist */
 	if (id_like_id != NULL) {
 		g_autofree gchar* id_like_path = g_build_filename (base, "EFI", id_like_id, NULL);
 		if (g_file_test (id_like_path, G_FILE_TEST_IS_DIR)) {
 			g_debug ("Using ID_LIKE key from os-release");
 			return g_steal_pointer (&id_like_path);
 		}
+	} else if (fu_device_get_metadata_boolean (device, "FallbacktoRemovablePath")) {
```
superm1, line 254: just to double check - `/etc/os-release` is indeed missing for ubuntu core right?

plugins/uefi/fu-uefi-common.c
```diff
@@ -241,12 +244,17 @@ fu_uefi_get_esp_path_for_os (const gchar *base)
...
+		esp_path = g_build_filename (base, "EFI", "boot", NULL);
+		if (g_file_test (esp_path, G_FILE_TEST_IS_DIR))
+			return g_steal_pointer (&esp_path);
```
superm1, line 257: seems to me that this is not necessary, the next line (259) does it unconditionally.

plugins/uefi/uefi.conf
```diff
@@ -10,3 +10,7 @@
 
 # amount of free space required on the ESP, for example using 0x2000000 for 32Mb
 #RequireESPFreeSpace=
+
+# with the UEFI removable path enabled, the default esp path is set to /EFI/boot
+# the shim EFI bianry and presumably this is $ESP/EFI/boot/bootx64.efi
```
superm1, line 15: still spelled wrong here

plugins/uefi/fu-uefi-common.c
woodrow-shen: There is a cat from UC18:
```
NAME="Ubuntu Core"
VERSION="18"
ID=ubuntu-core
PRETTY_NAME="Ubuntu Core 18"
VERSION_ID="18"
HOME_URL="https://snapcraft.io/"
BUG_REPORT_URL="http://bugs.launchpad.net/snappy/"
```

plugins/uefi/uefi.conf
woodrow-shen: Awkward...I see.

plugins/uefi/fu-uefi-common.c
woodrow-shen: How about this?

plugins/uefi/fu-uefi-common.c
superm1, line 254: I think this specific clause needs to just be an `if` not a `else if`.  That way even if `ID_LIKE` key was set but it didn't exist, we fallback.  Right now if you have `ID_LIKE` key but file doesn't and you set option exist you would return `unknown`.
I think it should be:
1. Try `ID`
2. Try `ID_LIKE`
3. Check for removable path option
4. If option set, try removable path
5. Return unknown if all fail

superm1 (commented):
One minor change suggested and then LGTM.
Please also rebase on master to fix CI (there was a workaround to help the Debian CI failure there)

plugins/uefi/fu-uefi-common.c
woodrow-shen: Thanks for pointing out this pitfall. Fixed.

woodrow-shen:
> One minor change suggested and then LGTM.
> Please also rebase on master to fix CI (there was a workaround to help the Debian CI failure there)

Updated.

superm1 (approved):
LGTM

plugins/uefi/fu-uefi-bootmgr.c
```diff
@@ -273,14 +274,23 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
...
 		if (shim_app == NULL)
 			return FALSE;
 
+		/* try to fallback to use UEFI removable path if the shim path doesn't exist */
+		if (!g_file_test (shim_app, G_FILE_TEST_EXISTS)) {
+			if (fu_device_get_metadata_boolean (device, "FallbacktoRemovablePath")) {
```
hughsie, line 283: maybe this is a stupid question, but why do we need FallbacktoRemovablePath at all? I'd really like to avoid adding more configuration, especially when we want to use the same snap image for Ubuntu Core and real Ubuntu.

superm1: It's called the "Removable" path because it's intended for when booting removable media.  Abusing it for HDD/SSD on regular boot may be problematic based upon the UEFI implementation and how IBV/OEM choose to use it in their BDS implementation.  If we automatically fall back to this path all the time we're encouraging this behavior.

The only reason it's been used in Linux ecosystem is optional and for when there are bugs in UEFI implementation.  See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767037 for an example.

superm1: > especially when we want to use the same snap image for Ubuntu Core and real Ubuntu.

Regarding this point I would expect that the confined snap would have this option set to `true` and **nowhere** else.  If the confined snap gets used outside of Ubuntu core regular paths get used and on Ubuntu core this thing gets used.

hughsie: >  Abusing it for HDD/SSD on regular boot may be problematic

Can we push back on Ubuntu Core to stop doing that then?

woodrow-shen: I cited from our production engineering to express some points of this design: It allows booting as removable media (cdrom / sd-card / usb stick) but also as internal drive (i.e. flashed in factory) without any modifications of an EFI variables. Also it is intentional that neither fallback, .CSV, nor mock-manager are provided out of the box. As using any of those features will affect the PCR measurements against which FDE secret is sealed against. The miminise number of possible bootflows and PCR values, it is intentional that UC20 provides more rigid structure (and less featureful ESP) when compared to classic. The things we chose to not ship would mostly open up ways for users to brick their device instead of achieving similar results as on classic.

superm1: > It allows booting as removable media (cdrom / sd-card / usb stick) but also as internal drive (i.e. flashed in factory) 

By doing it this way, you are not exercising efi variables ever and if there _is_ a UEFI implementation BDS bug you won't encounter it until you tried to do a firmware update.  At which point you _might_ not be able to boot anymore.

>As using any of those features will affect the PCR measurements against which FDE secret is sealed against

I don't think it's a reasonable expectation that booting external media and booting internal media will have the same PCR values.  PCR4 will change based upon which entry is selected, and you no longer have a predictable entry.  Furthermore, you have no guarantee taking an external media between machines that all the PCRs will measure the same between machines.  If one has a dbx update installed and the other doesn't for example.

Furthermore, I don't think it makes sense to have FDE on an external media that you don't have a boot entry for.  You'll need to pick that entry from the boot list, or modify BootOrder to have that entry.  And now if you have multiple disks plugged in, or switch the port it's plugged into you break things!

> without any modifications of an EFI variables

It's impossible to do fwupd based updates unless efi variables are modified.

> Also it is intentional that neither fallback, .CSV, nor mock-manager are provided out of the box

This makes sense to me, but it doesn't mean you have to install the bootloader into /efi/boot.  It can still go into `/efi/ubuntu-core` and make an NVRAM entry.

>flashed in factory

Flashed "same" image in the factory doesn't need to mean just `dd if=foo.bin of=/dev/nvmen0`.  All reasonable factory processes offer the ability to write to NVRAM variables as well.  There are tools and APIs for Windows, and obviously Linux has `efibootmgr`.

woodrow-shen: Maybe I'm put in the difficult position between our design and your thoughts. If this PR can't be acceptable, the only thing I could do is to use `EFI_OS_DIR`, and unfortunately I will keep maintaining patches. Before that, let me discuss internally.

superm1: From linked bug recent discussion it sounds like there may be some appetite to using NVRAM variables instead.  Let's wait until conclusion there to revisit.  The actual code in this looks OK to me if this is the way we end up all agreeing to go.

xnox: unfortunately this is time criticial. We need this PR as is, even if in the future nvram variable writing comes in. I fear the nvram writing might only happen in 2021 if not 2022.

hughsie: > I fear the nvram writing might only happen in 2021 if not 2022

Would the nvram writing ever happen if we merged this PR?

xnox: > > I fear the nvram writing might only happen in 2021 if not 2022
> 
> Would the nvram writing ever happen if we merged this PR?

Yeah. Definatly. We still need bootentries to selftest / factory mode / etc. but backlog and priorities.

superm1: When this would happen in 2021/2022, would older systems also eventually be covered?  I think we would like to be able to remove this option from fwupd at some point to avoid incurring technical debt.

xnox: even if we add the bootentires, we will continue to use the UEFI removable path on disk. this is to stay long term.

superm1: So if you still keep using the UEFI removable path but add NVRAM entries that's another interesting failure you might cause.  There are UEFI BDS implementations that will do "de-duplication" of entries.  Depending upon their algorithm either the entry for UEFI removable path or the entry you added might get removed.

Really the safest thing to do is to follow what the rest of the OS ecosystem does and create your own path in the ESP, place your EFI applications there and create NVRAM entries.  Otherwise be prepared for the wild west of firmware behaviors.

@hughsie Since even if ubuntu core adds nvram options they'll keep using UEFI removable path from an upstream fwupd perspective I think from what's been discussed there are these options:
1. We take this patch as is today, but keep upstream default as disabled, it's only used in their confined snap.
2. We add UEFI removable path as a permanent fallback.  This might have interesting implications for live media that we should think about (_maybe_ able to install UEFI BIOS updates from live media if your media is FAT32 and not written as `iso9660` (from a `dd`))
3. We reject this and Ubuntu core carries (or hardcodes) the fallback in confined snap as a downstream patch.

hughsie: @superm1 I trust your judgement on this. I'm not super keen on new config options but also see the benefit of having the same codebase for both snaps.

superm1:
> 1. We take this patch as is today, but keep upstream default as disabled, it's only used in their confined snap.
> 2. We add UEFI removable path as a permanent fallback. This might have interesting implications for live media that we should think about (maybe able to install UEFI BIOS updates from live media if your media is FAT32 and not written as iso9660 (from a dd))
> 3. We reject this and Ubuntu core carries (or hardcodes) the fallback in confined snap as a downstream patch.

>@superm1 I trust your judgement on this. I'm not super keen on new config options but also see the benefit of having the same codebase for both snaps.

### Option 2 
I was originally leaning on option 2, but after some lots of deliberation I think this is way too dangerous.  My justification in not making an automatic implied fallback is that if you are relying upon a bootloader called `bootx64.efi`.  Who is to say that's a shim, it's a GRUB, it's a Microsoft bootloader or it's malware?  We have **no way** to know.  At least with the binary being named `shimx64.efi` we know it's something that _likely_ behaves like a shim.  IOW you can guarantee it's something that supports a second stage to load and might run our `fwupdx64.efi`.

So that leaves the implications of using 1 or what ubuntu core does if we select 3.
### Option 1
If we went with option 1 I'd like to think about what happens with this confined snap _outside_ of ubuntu core.  In general I think all of us should have a common goal that fwupd snap can eventually be fully confined and work with all the devices.  As in `uefi-fw-tools` goes away and our current `fwupd` classic snap becomes a confined snap.  So if we took this confined snap and put it outside of Ubuntu core, does it actually get access to the ESP from `snapd`?  Can it tell determine it's running on a Fedora host and hence it should use `/efi/fedora/shimx64.efi`?  Or will that host detection stuff fail and it will try to use `/efi/boot/bootx64.efi`?  On a Fedora host how are we to know what's actually in that binary?  It might have been another OS left the bootloader there and it's not a shim.  Or it can be a shim that has been revoked from BootHole.

### Option 3
Now if we went to option 3 I think that Ubuntu core would need to compile their snap with `EFI_OS_DIR` set to `boot` and then apply a 1 line patch like this:
```
diff --git a/plugins/uefi/fu-uefi-bootmgr.c b/plugins/uefi/fu-uefi-bootmgr.c
index ee739672..670dfa0b 100644
--- a/plugins/uefi/fu-uefi-bootmgr.c
+++ b/plugins/uefi/fu-uefi-bootmgr.c
@@ -273,7 +273,7 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
        secure_boot = fu_efivar_secure_boot_enabled ();
        if (secure_boot) {
                /* test to make sure shim is there if we need it */
-               shim_app = fu_uefi_get_esp_app_path (esp_path, "shim", error);
+               shim_app = fu_uefi_get_esp_app_path (esp_path, "boot", error);
                if (shim_app == NULL)
                        return FALSE;
```

### Dissent
This really comes down to the aspirations for that confined snap.  If our mutual goal is to have a confined snap that can do FW updates outside of Ubuntu core we should go with option 1 (merge this patch) and let `snapd` guys work out the details that let `snapd` tell fwupd what host it is running on and and so the fallback logic works even outside of ubuntu core.

If that's impossible or not desirable then we should reject this and they should adopt my suggestion for `EFI_OS_DIR` and 1 line patch.

**In either case I would strongly urge ubuntu-core to stop using this UEFI removable path in the future.**

woodrow-shen:
> > 1. We take this patch as is today, but keep upstream default as disabled, it's only used in their confined snap.
> > 2. We add UEFI removable path as a permanent fallback. This might have interesting implications for live media that we should think about (maybe able to install UEFI BIOS updates from live media if your media is FAT32 and not written as iso9660 (from a dd))
> > 3. We reject this and Ubuntu core carries (or hardcodes) the fallback in confined snap as a downstream patch.
> 
> > @superm1 I trust your judgement on this. I'm not super keen on new config options but also see the benefit of having the same codebase for both snaps.
> 
> ### Option 2
> 
> I was originally leaning on option 2, but after some lots of deliberation I think this is way too dangerous. My justification in not making an automatic implied fallback is that if you are relying upon a bootloader called `bootx64.efi`. Who is to say that's a shim, it's a GRUB, it's a Microsoft bootloader or it's malware? We have **no way** to know. At least with the binary being named `shimx64.efi` we know it's something that _likely_ behaves like a shim. IOW you can guarantee it's something that supports a second stage to load and might run our `fwupdx64.efi`.
> 
> So that leaves the implications of using 1 or what ubuntu core does if we select 3.
> ### Option 1
> 
> If we went with option 1 I'd like to think about what happens with this confined snap _outside_ of ubuntu core. In general I think all of us should have a common goal that fwupd snap can eventually be fully confined and work with all the devices. As in `uefi-fw-tools` goes away and our current `fwupd` classic snap becomes a confined snap. So if we took this confined snap and put it outside of Ubuntu core, does it actually get access to the ESP from `snapd`? Can it tell determine it's running on a Fedora host and hence it should use `/efi/fedora/shimx64.efi`? Or will that host detection stuff fail and it will try to use `/efi/boot/bootx64.efi`? On a Fedora host how are we to know what's actually in that binary? It might have been another OS left the bootloader there and it's not a shim. Or it can be a shim that has been revoked from BootHole.

Detecting the host distribution is available now for snapd, but the most controversial concern I can envisage is can we allow the apparmor rules of fwupd to access the paths (e.g. /pathto/efi/fedora/fw/ or /pathto/efi/fedora/fwupd*) that are not related to Ubuntu at all?!  It's definitely not easy to make a conclusion to us, from my point of view.

> ### Option 3
> 
> Now if we went to option 3 I think that Ubuntu core would need to compile their snap with `EFI_OS_DIR` set to `boot` and then apply a 1 line patch like this:
> 
> ```
> diff --git a/plugins/uefi/fu-uefi-bootmgr.c b/plugins/uefi/fu-uefi-bootmgr.c
> index ee739672..670dfa0b 100644
> --- a/plugins/uefi/fu-uefi-bootmgr.c
> +++ b/plugins/uefi/fu-uefi-bootmgr.c
> @@ -273,7 +273,7 @@ fu_uefi_bootmgr_bootnext (const gchar *esp_path,
>         secure_boot = fu_efivar_secure_boot_enabled ();
>         if (secure_boot) {
>                 /* test to make sure shim is there if we need it */
> -               shim_app = fu_uefi_get_esp_app_path (esp_path, "shim", error);
> +               shim_app = fu_uefi_get_esp_app_path (esp_path, "boot", error);
>                 if (shim_app == NULL)
>                         return FALSE;
> ```
> 
> ### Dissent
> 
> This really comes down to the aspirations for that confined snap. If our mutual goal is to have a confined snap that can do FW updates outside of Ubuntu core we should go with option 1 (merge this patch) and let `snapd` guys work out the details that let `snapd` tell fwupd what host it is running on and and so the fallback logic works even outside of ubuntu core.
> 
> If that's impossible or not desirable then we should reject this and they should adopt my suggestion for `EFI_OS_DIR` and 1 line patch.

Well, if it would happen, I think, much to our regret, that I'm still appreciating your help and conversation! I will try to keep sync with you for any work.

> 
> **In either case I would strongly urge ubuntu-core to stop using this UEFI removable path in the future.**

superm1:
> Detecting the host distribution is available now for snapd, but the most controversial concern I can envisage is can we allow the apparmor rules of fwupd to access the paths (e.g. /pathto/efi/fedora/fw/ or /pathto/efi/fedora/fwupd*) that are not related to Ubuntu at all?! It could be not easy to make a conclusion to us, from my point of view.

I think in general fwupd expects to have full access to at least read content on the ESP and to write certain content.  We've talked about reading and/or storing other things on the ESP later on so this may come up again later.
Do the apparmor rules need to be so strict?  It seems very arbitrary to say that snapd can write to `/pathto/efi/ubuntu/` but not to `/pathto/efi/fedora/`
Perhaps there are approaches like allowing write to `/pathto/efi/*/fwupd*.efi` and to to `/pathto/efi/*/fw` so that it doesn't need to be distribution specific but still protect other binaries in the directories.

superm1:
I'm going to close this PR as directionally we're not aligned on the confined snap working outside of ubuntu core.  If that changes, please re-open and we'll revisit it.

woodrow-shen:
@superm1 I'm trying to talk with security team to see if we're the same page, thanks.

jdstrand:
We have a couple of options in snapd to make this work for non-Ubuntu Core while retaining the same security properties on Ubuntu Core. Eg, on non-UC, we can add the more permissive glob rules to support fedora (from your above example, but would support any distro). With that in mind, I think this can be reopened?

superm1:
In that case I would say yes we should reopen. The branch that was associated with this was deleted so the reopen button is disabled. @woodrow-shen  please open another with the content that we had reviewed.

woodrow-shen:
Right...I'm cleaning up the branch from our repo. Let me create a new one.

woodrow-shen:
https://github.com/fwupd/fwupd/pull/2571

PR #2538, opened 2020-11-01
Updated fish completions to 1.5.0

exploide:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [x] Documentation

Just updated the fish completions to conform with v1.5.0.

By the way, I noticed in the `--help` output:

```
unblock-firmware [CHECKSUM]       Blocks a specific firmware from being installed
```

This is a copy and waste error, isn't it? Probably should be `Unblocks a...`


hughsie:
> By the way, I noticed in the --help output:

Eeek! Can you do a new PR please? Thanks!

PR #2539, opened 2020-11-01
trivial: fix description of unblock-firmware

exploide:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [x] Documentation

I don't know if I need to regenerate the `.po` files, too. Did not do this yet.


hughsie:
> I don't know if I need to regenerate the .po files, too

This is automatic, but I just synced transifex so the new sting should appear for the translators.

Issue #2543, opened 2020-11-03
Probably should de-emphasize snap


hughsie:
At the moment looking at the README the only installation method we suggest is Snap. This makes me feel somewhat uneasy, and has made people make more than a few people think that fwupd depends on Snap :(

![Screenshot_2020-11-03 Linux Mint pushes out its own Chromium build to help users avoid Canonical's Snap Store](https://user-images.githubusercontent.com/151380/98021809-aad54600-1dfc-11eb-9a4e-45aabdcf343b.png)
...from https://www.theregister.com/2020/11/02/linux_mint_chromium_build/

I think my general thought is that I don't think end users should ever compile fwupd from scratch; it's a complicated project with dozens of dependencies (and as many configuration options) and there's just too many things that can go wrong. Users should just have fwupd installed by their distro, hopefully by a packager who does understand all the config options. I don't think offering it as a snap absolves the downstream of doing the right thing and I **certainly** don't want to see a snap-only fwupd install method like Canonical is currently doing with Chromium.

My personal opinion is you're not supposed to replace low level parts of your OS yourself. If you install fwupd 1.5.1 on your Ubuntu LTS it'll probably install fine, but we've done exactly zero QA on that as a configuration and the software store frontend might work fine, or might explode in a ball of flames. The snap or flatpak version might be useful to update a specific device on the command line that needs a bleeding edge fwupd version (because the specific device support is so new) but I really don't think using the snap or flatpak should ever be seen as a replacement to the system version.

superm1:
>At the moment looking at the README the only installation method we suggest is Snap. This makes me feel somewhat uneasy, and has made people make more than a few people think that fwupd depends on Snap :(

Huh, I guess I haven't actually _read_ it in a while.

> certainly don't want to see a snap-only fwupd install method like Canonical is currently doing with Chromium.

Me either.

>I think my general thought is that I don't think end users should ever compile fwupd from scratch; it's a complicated project with dozens of dependencies (and as many configuration options) and there's just too many things that can go wrong. 

Yeah in general most people shouldn't be compiling.

>The snap or flatpak version might be useful to update a specific device on the command line that needs a bleeding edge fwupd version (because the specific device support is so new)

Now I've found snap is incredibly useful for ODMs to use, particularly because OSVs don't roll out newer versions in LTS release.

>but I really don't think using the snap or flatpak should ever be seen as a replacement to the system version.

[...]

PR #2544, opened 2020-11-03
Add note to the README about not using Snap or Flatpak versions

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2543

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,6 @@ fwupd
 [![Build Status](https://travis-ci.org/fwupd/fwupd.png?branch=master)](https://travis-ci.org/fwupd/fwupd)
 [![Coverity Scan Build Status](https://scan.coverity.com/projects/10744/badge.svg)](https://scan.coverity.com/projects/10744)
 
-[![Get it from the Snap Store](https://snapcraft.io/static/images/badges/en/snap-store-white.svg)](https://snapcraft.io/fwupd)
-
 This project aims to make updating firmware on Linux automatic, safe and reliable.
 
 Additional information is available at the website: https://fwupd.org/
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +11,20 @@ Additional information is available at the website: https://fwupd.org/
 
 The most up to date compilation instructions are available in the [Wiki](https://github.com/fwupd/fwupd/wiki/Compilation).
 
+**NOTE:** In most cases end users should never compile fwupd from scratch; it's a
+complicated project with dozens of dependencies (and as many configuration options)
+and there's just too many things that can go wrong.
+
+Users should just have fwupd installed and updated by their distro, hopefully
+by a packager who does understand all the config options.
```
superm1, line 19: no need to throw shade at packagers

README.md
```diff
@@ -13,6 +11,20 @@ Additional information is available at the website: https://fwupd.org/
...
+The distribution will have also done some testing with how fwupd interacts with
+other software on your system, for instance using GNOME Software.
+
+Installing fwupd from the [Snap Store](https://snapcraft.io/fwupd) or from
```
superm1, line 23: I sorta think we should kick people here instead: https://github.com/fwupd/fwupd/wiki/fwupd-snap

README.md
```diff
@@ -13,6 +11,20 @@ Additional information is available at the website: https://fwupd.org/
...
+[Flathub](https://flathub.org/) might be useful to update a specific device on
```
superm1, line 24: And here https://github.com/fwupd/fwupd/wiki/fwupd-flatpak

hughsie: Agree, fixed.

README.md
hughsie: Agree, fixed.

README.md
hughsie: hah, that was meant more with British sarcasm :) What about `Users should just have fwupd installed and updated by their distro, managed and tested by a trusted package maintainer."

superm1: I don't think bringing a trusted vs untrusted packager into the equation matters.  Most people see distributions as an amorphous blob that provides a Linux OS.

hughsie: Maybe `Users should just have fwupd installed and updated by their distro, managed and tested by the package maintainer.`?

superm1: That's better

hughsie: Okay, fixed.

superm1 (approved):

PR #2555, opened 2020-11-06
bcm57xx: Fix reading vendor/device ids from firmware.

meklort:
The PCI Vendor and Device ID locations located in firmware were mistakenly
swapped in the bcm5719-fw repository. As a result, the code here based on said
repository also has swapped IDs. This fixes the ids to reflect the
correct locations.

Relevant change: https://github.com/meklort/bcm5719-fw/pull/161

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Is this just cosmetic or actually will prevent firmware upgrade and downgrade in some way?

meklort:
I'm working on fixing the code in the oss firmware to make a fully valid firmware image (for use by RCS in the factory), and noticed that the hex dump had the vendor id / device ids swapped. After fixing it, fwupd started stating that the firmware had the incorrect vendor id.

So, this will be a functional issue once updated and fixed firmware images are being generated.

I'm thinking the best way to handle it is to make sure new fwupd cab files are restricted to the next fwupd version and to make sure a new release is uploaded before the next fwupd is available.

meklort:
Actually, if it makes sense, I can start packaging two metainfo.xml files in the .cab file - one for fwup 1.5.0 and 1.5.1 that has the version swapped, and one for greater than 1.5.1 with the correct format. That could allow for an easy transition before the next fwupd release.

hughsie:
> and one for greater than 1.5.1

Don't worry about 1.5.0 or 1.5.1 -- I can do a 1.5.2 for you next week. Just make the new firmware depend on 1.5.2 :)

meklort:
> Don't worry about 1.5.0 or 1.5.1 -- I can do a 1.5.2 for you next week. Just make the new firmware depend on 1.5.2 :)

Done. I've released a new version that restricts to 1.5.2, manually tested it, and also pushed it to testing.

hughsie:
> released a new version that restricts to 1.5.2

I'm just waiting to hear back from an ODM, and then I'll probably drop 1.5.2 early next week. Thanks.

PR #2559, opened 2020-11-06
Set the SMBIOS chassis type to portable if a DT battery exists

hughsie:
This allows getting the HSI result on a Pinebook Pro laptop.

### DIFF
--- a/libfwupdplugin/fu-smbios.c
+++ b/libfwupdplugin/fu-smbios.c
@@ -71,6 +71,15 @@ typedef struct {
 G_DEFINE_TYPE (FuSmbios, fu_smbios, G_TYPE_OBJECT)
 
 static void
+fu_smbios_convert_dt_value (FuSmbios *self, guint8 type, guint8 offset, guint8 value)
+{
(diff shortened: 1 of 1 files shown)


superm1 (commented):
This feels like an abuse of the chassis type.  Can't a phone also have a battery and not be a portable chassis?  Or a tablet have a battery?

Since this is DT - I think what we want to do is add something to DT to specify the form factor or chassis type.

hughsie:
> I think what we want to do is add something to DT to specify the form factor or chassis type.

I don't think this is going to work for all the devices in the field right now; I'd be happy to support a DT type and fallback to the battery heuristic. If there's already a DT attribute for this please let me know!

superm1:
I glanced through the spec and didn't see anything.  I guess i'm with that approach, but please submit a request to DT spec to add the field so we can start to rely upon that for the future!

hughsie:
> please submit a request to DT spec

I've banged some Tom Toms and I'll see where that leads. If something appears I'll fix up the code of course.

PR #2561, opened 2020-11-06
Wip/hughsie/fu hid device

hughsie:
Type of pull request:
- [X] Code fix

### DIFF
--- a/libfwupdplugin/fu-hid-device.c
+++ b/libfwupdplugin/fu-hid-device.c
@@ -33,6 +33,7 @@ typedef struct
 	FuUsbDevice		*usb_device;
 	guint8			 interface;
 	gboolean		 interface_autodetect;
+	FuHidDeviceFlags	 flags;
 } FuHidDevicePrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (FuHidDevice, fu_hid_device, FU_TYPE_USB_DEVICE)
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-hid-device.h
```diff
@@ -37,10 +37,13 @@ typedef enum {
 	FU_HID_DEVICE_FLAG_ALLOW_TRUNC		= 1 << 0,
 	FU_HID_DEVICE_FLAG_IS_FEATURE		= 1 << 1,
 	FU_HID_DEVICE_FLAG_RETRY_FAILURE	= 1 << 2,
+	FU_HID_DEVICE_FLAG_NO_BIND_ON_CLOSE	= 1 << 3,
```
superm1: why is this introduced in this commit?  (It's removed later)

libfwupdplugin/fu-hid-device.c
```diff
@@ -146,8 +147,19 @@ fu_hid_device_close (FuUsbDevice *device, GError **error)
...
-		g_prefix_error (error, "failed to release HID interface: ");
+	if (!g_usb_device_release_interface (usb_device, priv->interface, flags, &error_local)) {
+		if (g_error_matches (error_local,
+				     G_USB_DEVICE_ERROR,
+				     G_USB_DEVICE_ERROR_NO_DEVICE) ||
+		    g_error_matches (error_local,
+				     G_USB_DEVICE_ERROR,
+				     G_USB_DEVICE_ERROR_INTERNAL)) {
+			g_debug ("ignoring: %s", error_local->message);
+			return TRUE;
```
superm1, line 158: couldn't this possibly be a situation that the device was USB and unplugged though too?  Shouldn't it only be in certain contexts (such as you know it's about to reboot so it might go away).

libfwupdplugin/fu-hid-device.h
hughsie: Oops, rebase failure -- sorry. Fixed now.

libfwupdplugin/fu-hid-device.c
hughsie: This error happens when you have a HID device we use with a `FuDeviceLocker` and then get the HID device to reboot. The device returns from the SetReport, reboots -- then we try to call `libusb_release_interface()` on the interface that no longer exists, then `libusb_close()` -- the release of the interface failing as the interface doesn't actually exist any more. I did hope to use just `LIBUSB_ERROR_NO_DEVICE` but libusb returns `LIBUSB_ERROR_NOT_FOUND` which we unhelpfully map to `G_USB_DEVICE_ERROR_INTERNAL` rather than `G_USB_DEVICE_ERROR_NO_DEVICE` or something more helpful.

I'm not sure we can handle this in the plugin code.

superm1 (approved):

PR #2562, opened 2020-11-06
Add a flag to indicate if packages are supported

superm1:
Anyone can easily add this, but it makes it clearer that by default hand
build, snap, and flatpak are not checked by anyone.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -22,6 +22,7 @@ option('plugin_modem_manager', type : 'boolean', value : false, description : 'e
 option('plugin_msr', type : 'boolean', value : true, description : 'enable MSR support')
 option('plugin_flashrom', type : 'boolean', value : false, description : 'enable libflashrom support')
 option('plugin_coreboot', type : 'boolean', value : true, description : 'enable coreboot support')
+option('supported_build', type : 'boolean', value : false, description: 'distribution package with upstream support')
 option('systemd', type : 'boolean', value : true, description : 'enable systemd support')
 option('systemd_root_prefix', type: 'string', value: '', description: 'Directory to base systemd’s installation directories on')
 option('elogind', type : 'boolean', value : false, description : 'enable elogind support')
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -186,6 +186,7 @@ fu_util_start_engine (FuUtilPrivate *priv, FuEngineLoadFlags flags, GError **err
 			    fmt);
 	}
 	fu_util_show_plugin_warnings (priv);
+	fu_util_show_unsupported_warn ();
 	return TRUE;
 }
 
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -3120,6 +3120,9 @@ main (int argc, char *argv[])
 	/* show user-visible warnings from the plugins */
 	fu_util_show_plugin_warnings (priv);
 
+	/* show any unsupported warnings */
+	fu_util_show_unsupported_warn ();
+
 	/* we know the runtime daemon version now */
 	fwupd_client_set_user_agent_for_package (priv->client, "fwupdmgr", PACKAGE_VERSION);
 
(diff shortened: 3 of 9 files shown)


meson_options.txt
```diff
@@ -22,6 +22,7 @@ option('plugin_modem_manager', type : 'boolean', value : false, description : 'e
 option('plugin_msr', type : 'boolean', value : true, description : 'enable MSR support')
 option('plugin_flashrom', type : 'boolean', value : false, description : 'enable libflashrom support')
 option('plugin_coreboot', type : 'boolean', value : true, description : 'enable coreboot support')
+option('supported_build', type: 'boolean', value : false, description: 'Distribution package with formal support')
```
hughsie, line 25: maybe `s/formal support/upstream support`?

hughsie: lowercase `D` I think too.

src/fu-tool.c
```diff
@@ -177,14 +178,17 @@ fu_util_start_engine (FuUtilPrivate *priv, FuEngineLoadFlags flags, GError **err
 	if (!fu_engine_load (priv->engine, flags, error))
 		return FALSE;
 	if (fu_engine_get_tainted (priv->engine)) {
-		g_autofree gchar *fmt = NULL;
+		/* TRANSLATORS: the user is SOL for support... */
+		msg =  _("The daemon has loaded 3rd party code and "
+			"is no longer supported by the upstream developers!");
+		fu_util_show_daemon_warning (msg);
```
hughsie, line 184: I think this is a bit of a hack; I think it either needs to be a varargs `...` style function that can accept printf-style arguments or we just use `fu_util_term_format()` and a `GString` _accumulator_.

src/fu-util.c
```diff
@@ -3107,15 +3106,16 @@ main (int argc, char *argv[])
...
-			      "is no longer supported by the upstream developers!"));
-	}
+		/* TRANSLATORS: the user is SOL for support... */
+		msg =  _("The daemon has loaded 3rd party code and "
+			"is no longer supported by the upstream developers!");
+		fu_util_show_daemon_warning (msg);
+	}
+#ifndef SUPPORTED_BUILD
+	/* TRANSLATORS: the user is SOL for support... */
+	msg = _("This is not a distribution validated package, it is only best effort support");
```
hughsie, line 3116: Maybe `Use of this 3rd party package is not supported by the upstream developers.`?

hughsie:
A few niggles, but I think the concept is great.

src/fu-util.c
superm1: I really don't want to tell someone just because they used the snap or hand built they're getting ZERO help.  Is there somewhere in the middle?

hughsie: `Use of this 3rd party package is not really supported by the upstream developers` or `Use of this 3rd party package is not fully supported by the upstream developers`?

superm1: Hmm, I'm still not there with either of these messages.  If someone has gone through the trouble to successfully build 1.5.1, I would rather help them than an old Ubuntu 1.3.11.

Similarly with the snap, I can trust they got all the deps right because they're all in the snap already.

superm1: How about `This is not a distribution validated package, it may not work properly with other fwupd based clients`

superm1: Or even 
> This package is not validated by `$NAME`, it may not work properly with other fwupd based clients

Where `$NAME` comes from `/etc/os-release`.

hughsie: > This package is not validated by $NAME

I'm not sure that's true for Ubuntu; politically they want to push snaps everywhere. It's us who care. :(

superm1: OK so how about just:
`This package has not been validated, it may not work properly with other fwupd based clients`

hughsie: Maybe even shorter, `This package has not been validated, it may not work properly.`

hughsie (approved):

Issue #2564, opened 2020-11-07
Assertion failed when running installed tests


jtojnar:
**Describe the bug**
I am getting the following warning when running installed tests:

```
Nov 07 15:13:39 machine fwupd[763]: 15:13:39:0307 GLib                 g_strv_contains: assertion 'str != NULL' failed
```

Adding some traces, it originates from this line:

https://github.com/fwupd/fwupd/blob/01ba0c1eea3b0b550a12e96eef670e5e5ae9cf21/src/fu-engine.c#L1624

This is the hexdump of `/proc/cmdline`:

```
00000000: 636f 6e73 6f6c 653d 7474 7953 3020 7061  console=ttyS0 pa
00000010: 6e69 633d 3120 626f 6f74 2e70 616e 6963  nic=1 boot.panic
00000020: 5f6f 6e5f 6661 696c 206c 6f67 6c65 7665  _on_fail logleve
00000030: 6c3d 3720 6e65 742e 6966 6e61 6d65 733d  l=7 net.ifnames=
00000040: 3020 696e 6974 3d2f 6e69 782f 7374 6f72  0 init=/nix/stor
00000050: 652f 6671 3432 3439 7667 3372 7a67 6477  e/fq4249vg3rzgdw
00000060: 3670 6467 7872 7261 3035 3332 6762 3461  6pdgxrra0532gb4a
00000070: 3167 2d6e 6978 6f73 2d73 7973 7465 6d2d  1g-nixos-system-
00000080: 6d61 6368 696e 652d 3231 2e30 3370 7265  machine-21.03pre
00000090: 2d67 6974 2f69 6e69 7420 7265 6749 6e66  -git/init regInf
000000a0: 6f3d 2f6e 6978 2f73 746f 7265 2f6d 306c  o=/nix/store/m0l
000000b0: 7333 6b30 6972 7237 3178 3032 6731 6e7a  s3k0irr71x02g1nz
000000c0: 3033 6b36 6976 6b34 6867 3064 722d 636c  03k6ivk4hg0dr-cl
000000d0: 6f73 7572 652d 696e 666f 2f72 6567 6973  osure-info/regis
000000e0: 7472 6174 696f 6e20 636f 6e73 6f6c 653d  tration console=
000000f0: 7474 7953 3020 0a                        ttyS0 .
```

**Steps to Reproduce**
Run installed tests, read the logs.

[...]

PR #2565, opened 2020-11-07
trivial: fu-engine: skip empty cmdline commands

jtojnar:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

Closes: #2564


src/fu-engine.c
```diff
@@ -1620,6 +1620,8 @@ fu_engine_get_report_metadata_kernel_cmdline (GHashTable *hash, GError **error)
 		g_auto(GStrv) tokens = fu_common_strnsplit (buf, bufsz - 1, " ", -1);
 		g_autoptr(GString) cmdline_safe = g_string_new (NULL);
 		for (guint i = 0; tokens[i] != NULL; i++) {
+			if (strlen (tokens[i]) == 0)
```
superm1, line 1623: I think that C99 requires that the variables be declared at start of the block.  So this probably needs to be:

```
g_auto(Gstrv) kv = NULL;
if (strlen ...)
kv = g_strsplit( ..)
```

PR #2571, opened 2020-11-11
uefi: a new option for uefi configuration to use UEFI removable path

woodrow-shen:
Per discussion of #2513, Ubuntu Core was going to use UEFI removable
path as default esp path, and it needs to change some parts around
getting the esp path and searching the shim app path. Also, a new option
"FallbacktoRemovablePath" is added into uefi.conf to be applied in this
case, and it will be false by default.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -241,7 +241,7 @@ fu_plugin_uefi_write_splash_data (FuPlugin *plugin,
 
 	/* save to a predicatable filename */
 	esp_path = fu_volume_get_mount_point (data->esp);
-	directory = fu_uefi_get_esp_path_for_os (esp_path);
+	directory = fu_uefi_get_esp_path_for_os (device, esp_path);
 	basename = g_strdup_printf ("fwupd-%s.cap", FU_EFIVAR_GUID_UX_CAPSULE);
 	fn = g_build_filename (directory, "fw", basename, NULL);
 	if (!fu_common_mkdir_parent (fn, error))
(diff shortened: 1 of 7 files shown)


woodrow-shen:
What the security team gives the opinion regarding fwupd interface to support all classic distros:

Yeah, this is quite general. What we can do is in fwupdPermanentSlotAppArmor:

  <add current rules>

  if release.OnClassic
    <add above rules>

Eg, I suggest starting with something like this:

const fwupdPermanentSlotAppArmor = `
  ... all the current rules ...
`

const fwupdPermanentSlotAppArmorClassic = `
  # additional rules for fwupd on classic distro
  /boot/efi/EFI/*/fwupd*.efi rw,
  /boot/efi/EFI/*/fw/** rw,
`

func (iface *fwupdInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
  // Only apply slot snippet when running as application snap
  if !implicitSystemPermanentSlot(slot) {
    spec.AddSnippet(fwupdPermanentSlotAppArmor)
    // An application snap on classic also should have these rules
    if release.OnClassic {
      spec.AddSnippet(fwupdPermanentSlotAppArmorClassic)
    }
    ...
}

woodrow-shen:
So now it's more clear to say we're going to the **option 1**, is it correct?

superm1:
Thanks @woodrow-shen.

woodrow-shen:
@superm1 Great! I may be back to the open issue to summarize what's next TODO for me.

Issue #2578, opened 2020-11-13
regression with fwupdmgr switch-branch


meklort:
**Describe the bug**
Commit bf43978dfa1eeedf058dd3adc4d029b94d3bec59 introduces a regression when trying to flash the bcm5719 oss-firmware branch to the Talos II.

With the previous commit, `fwupdmgr switch-branch` properly detects that lvfs-testing has valid firmware that can be enabled.
When  bf43978dfa1eeedf058dd3adc4d029b94d3bec59 is used, "No supported devices" is reported instead.

**Steps to Reproduce**
- Flash the proprietary bcm5719 firmware
- Run Enable lvfs-testing
- Run fwupdmgr switch-branch

**Expected behavior**
Branch switching should be allowed, valid firmware should be found.

**fwupd version information**
built from latest soruce code. 

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
meklort@talos2 ~/g/f/build ((bf43978d…)) [1]> fwupdmgr get-devices --show-all-devices
WARNING: UEFI capsule updates not available or enabled
  See https://github.com/fwupd/fwupd/wiki/PluginFlag:capsules-unsupported for more information.
WARNING: Plugin dependencies missing
  See https://github.com/fwupd/fwupd/wiki/PluginFlag:failed-open for more information.
T2P9D01 REV 1.00
│
├─NetXtreme BCM5719 Gigabit Ethernet PCIe:
│     Device ID:          7de5ffdca08fa52d95fd4bb42aa5d07a4b35d2dd
│     Current version:    1.43
│     Vendor:             Broadcom Inc. and subsidiaries (PCI:0x14E4)
│     GUIDs:              ff247529-8b0a-5db7-98ef-8746aa018be6
│                         5d8b12bf-1973-58fc-8b31-3e50fe31954d
│                         3800738e-1413-5a18-846a-8782f5cc8c44
│                         30fe13b6-aa73-5d8c-a19f-c7b600f0117a
│     Device Flags:       • Updatable
│                         • Needs a reboot after installation
│                         • Cryptographic hash verification is available
│                         • Device will backup firmware before installing
│   
├─NetXtreme BCM5719 Gigabit Ethernet PCIe (4-port 1Gb Ethernet Adapter):
│     Device ID:          9e517c3cc54687578aa78274c9644219970a5b5b
│     Current version:    1.31
[...]

PR #2582, opened 2020-11-13
trivial: Fix a logic thinko to unbreak switch-branch

hughsie:
We still want to return TRUE, rather than falling through as if the requirement
type was unknown.

Fixes https://github.com/fwupd/fwupd/issues/2578

### DIFF
--- a/plugins/bcm57xx/fu-bcm57xx-device.c
+++ b/plugins/bcm57xx/fu-bcm57xx-device.c
@@ -57,8 +57,8 @@ fu_bcm57xx_device_probe (FuUdevDevice *device, GError **error)
 	g_autofree gchar *fn = NULL;
 	g_autoptr(GPtrArray) ifaces = NULL;
 
-	/* only enumerate number 0 */
-	if (fu_udev_device_get_number (device) != 0) {
+	/* only enumerate first device */
+	if (fu_udev_device_get_number (device) != 1) {
 		g_set_error_literal (error,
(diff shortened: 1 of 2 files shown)


superm1 (approved):

hughsie:
@meklort the Dell card doesn't work on PCI function 0; do we need to work out why or is 1 going to work for the RCS hardware too?

meklort:
@hughsie From the testing I've done, the BCM5719 has a lockup that happens when using function 1 *and* the ethtool api to flash, function 0 doesn't have this issue. The latest code in fwupd in both fwup and the bcmflash utility have this fixed, and so function 1 could be used with raw access.

What issue are you seeing with the Dell card on function 0? Last time I tried it, I thought it was working.

hughsie:
> and so function 1 could be used with raw access

Great, thanks for explaining!

> What issue are you seeing with the Dell card on function 0

I guess it might work, but the kernel doesn't present a "net" directory in the sysfs entry for `.0`, so fwupd never finds the ethtool interface name. It seems to work for all the other !0 PCI devices...

meklort:
> I guess it might work, but the kernel doesn't present a "net" directory in the sysfs entry for `.0`, so fwupd never finds the ethtool interface name. It seems to work for all the other !0 PCI devices...

Interesting, on my machine I see every interface in /sys/class/net/ even funciton 0 for the dell card.

PR #2587, opened 2020-11-15
Install modules-load configs in libdir

tinywrkb:
This change follows the convention that packaged distro/vendor configs should be installed to `/usr` while leaving `/etc` for admin/local system changes.

I believe this will break the Snap install and remove hooks, I'm not sure what to do there, can Snap pick up `libdir/modules-load.d` configs?.
Also, the deb spec (`contrib/fwupd.spec.in`) might need an update.  
If both considered merge blocking then I don't mind letting someone with experience with Snap and deb packages handle this PR.

### DIFF
--- a/plugins/msr/meson.build
+++ b/plugins/msr/meson.build
@@ -5,7 +5,7 @@ install_data(['msr.quirk'],
 )
 
 install_data(['fwupd-msr.conf'],
-  install_dir: join_paths(sysconfdir, 'modules-load.d')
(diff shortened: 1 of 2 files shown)


hughsie:
Both CI failures are merge blocking :) I've opened https://github.com/fwupd/fwupd/pull/2591 as a replacement to this; I don't think libdir is what you want in most cases. Luckily systemd provides a pkgconfig variable we can use.

tinywrkb:
@hughsie thank you for picking this up.

PR #2591, opened 2020-11-16
Install modules-load configs in the correct directory

hughsie:

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/debian/fwupd.install
+++ b/contrib/debian/fwupd.install
@@ -9,6 +9,7 @@ usr/share/polkit-1/*
 usr/share/locale
 usr/share/metainfo/*
 usr/libexec/fwupd/*
+usr/lib/modules-load.d/*.conf
 usr/share/man/man1/*
 lib/systemd/system/*
 lib/systemd/system-preset/*
--- a/plugins/msr/meson.build
+++ b/plugins/msr/meson.build
@@ -4,9 +4,11 @@ install_data(['msr.quirk'],
   install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
 )
 
+if get_option('systemd')
 install_data(['fwupd-msr.conf'],
-  install_dir: join_paths(sysconfdir, 'modules-load.d')
+  install_dir: systemd_modules_load_dir,
 )
+endif
 
 shared_module('fu_plugin_msr',
   fu_hash,
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -335,9 +335,9 @@ mkdir -p --mode=0700 $RPM_BUILD_ROOT%{_localstatedir}/lib/fwupd/gnupg
 %config(noreplace)%{_sysconfdir}/pki/fwupd
 %{_sysconfdir}/pki/fwupd-metadata
 %if 0%{?have_msr}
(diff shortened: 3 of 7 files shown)


hughsie:
@superm1 I've guessed the snap bits, please feel free to fix up any of the bits I've got wrong. Thanks.

plugins/msr/meson.build
```diff
@@ -4,9 +4,11 @@ install_data(['msr.quirk'],
   install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
 )
 
+if get_option('systemd')
 install_data(['fwupd-msr.conf'],
-  install_dir: join_paths(sysconfdir, 'modules-load.d')
+  install_dir: systemd_modules_load_dir,
 )
+endif
```
superm1, line 11: Hmm, so what do we do on a build _without_ systemd?  It just doesn't get to load the configuration file?  Should we have a fallback hardcoded path?

hughsie: > so what do we do on a build without systemd

I'm not sure I care; the module load stuff is implemented by systemd-udev these days anyway....

superm1: What about maybe emitting a warning in that else condition in this case so it's not a huge surprise for anyone?

hughsie: I guess that makes sense; the alternative is that we make the module-load-using plugins also depend on systemd...

superm1:
The snap stuff looks fine, PR is fine with me.

Regarding the CI failure, it's an HTTP 502 w/ installed-tests.  I wonder if we want to be catching that and at least in the context of self tests emitting something that we can catch so as to not fail tests?

plugins/msr/meson.build
superm1: I think making them depend on systemd is probably a bit much.  Those modules could be compiled into a kernel, or they could have been previously loaded by something else.

hughsie:
> it's an HTTP 502

Eeek, I see `Failed to download: Bad Gateway` -- is that a network issue on the LVFS, or in travis somewhere?

superm1:
> Eeek, I see Failed to download: Bad Gateway -- is that a network issue on the LVFS, or in travis somewhere?

I'm not sure where it actually is, but yeah that's what I was talking about.  If that's a travis network issue we shouldn't be failing CI.  But if it's an LVFS issue that we're getting early notification, maybe we should :)

As for how to figure out which is which, I guess cross reference time stamps with LVFS logs?

contrib/debian/fwupd.install
```diff
@@ -9,6 +9,7 @@ usr/share/polkit-1/*
 usr/share/locale
 usr/share/metainfo/*
 usr/libexec/fwupd/*
+usr/lib/modules-load.d/*.conf
```
superm1, line 12: Unfortunately I realize this probably needs some packaging transitional code since we put something into `/etc` previously.  Don't worry about blocking the PR on it. i'll open a separate issue to get to it later.

hughsie:
> As for how to figure out which is which, I guess cross reference time stamps with LVFS logs?

Nothing in the LVFS logs; I assume this is a Travis failure.

Issue #2553, opened 2020-11-05
UEFI ESP partition is in /boot/efi but not recognized


bin101:
**Describe the bug**
Every fwupd command responses with "UEFI ESP partition not detected or configured"

**Steps to Reproduce**
$ fwupdmgr refresh
WARNING: UEFI ESP partition not detected or configured

**Expected behavior**
The UEFI ESP partition should be detected

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version
client version:	1.5.0
compile-time dependency versions
	gusb:	0.3.5

daemon version:	1.5.0
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
pacman

**Additional questions**
- Operating system and version: Manjaro 20.2
- Have you tried rebooting? Yes
- Is this a regression? don't know

```shell
$ tree /boot/
/boot/
├── amd-ucode.img
├── efi
│   └── EFI
│       ├── boot
│       │   └── bootx64.efi
│       └── manjaro
│           ├── fw
│           │   ├── fwupd-3b8c8162-188c-46a4-aec9-be43f1d65697.cap
│           │   └── fwupd-603baf73-b997-45b5-86b4-2f981a008e18.cap
│           ├── fwupdx64.efi
│           └── grubx64.efi
├── grub
.
.
.
```

```shell
$ /usr/bin/fwupdtool esp-list --verbose                               
14:42:08:0585 FuDebug              Verbose debugging enabled (on console 1)
14:42:08:0592 FuCommon             device /org/freedesktop/UDisks2/block_devices/nvme0n1p2, type: ebd0a0a2-b9e5-4433-87c0-68b6b72699c7
14:42:08:0596 FuCommon             device /org/freedesktop/UDisks2/block_devices/nvme0n1p3, type: 0fc63daf-8483-4772-8e79-3d69d8477de4
14:42:08:0603 FuCommon             device /org/freedesktop/UDisks2/block_devices/nvme0n1p1, type: ebd0a0a2-b9e5-4433-87c0-68b6b72699c7
no volumes of type c12a7328-f81f-11d2-ba4b-00a0c93ec93b
```

**If I set OverrideESPMountPoint to /boot/efi the warning disappears.**

hughsie:
I think if you install 1.5.1 you'll get a link to the wiki page https://github.com/fwupd/fwupd/wiki/PluginFlag:esp-not-found

bin101:
[...]

PR #2593, opened 2020-11-16
trivial: uefi: don't let user's environment fail UEFI/Dell test suite

superm1:
Comment from https://github.com/fwupd/fwupd/issues/2553#issuecomment-728218741
makes it clear that test suite should ignore some of environment.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-plugin-uefi.c
+++ b/plugins/uefi/fu-plugin-uefi.c
@@ -620,6 +620,10 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 	g_autofree gchar *nvram_total_str = NULL;
 	g_autoptr(GError) error_local = NULL;
 
+	/* don't let user's environment influence test suite failures */
+	if (g_getenv ("FWUPD_UEFI_TEST") != NULL)
+		return TRUE;
+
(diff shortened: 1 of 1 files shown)


plugins/uefi/fu-plugin-uefi.c
```diff
@@ -620,6 +620,10 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 	g_autofree gchar *nvram_total_str = NULL;
 	g_autoptr(GError) error_local = NULL;
 
+	/* don't let user's environment influence test suite failures */
+	if (g_getenv ("FWUPD_UEFI_TEST") != NULL)
+		return TRUE;
```
hughsie, line 625: So this was being set accidentally?

superm1: Well we **intentionally** set that in the test suite for the dell plugin.  It's the only test suite that actually loads a real non-test plugin (IIRC).  So when that's set we want to make sure to ignore all startup checks.

hughsie: ahh, gotcha. LGTM then.

PR #2594, opened 2020-11-16
Enable github workflows

superm1:
Travis has been really slow lately.  See if migrating to Github can help.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (commented):
Yes, Travis is super irritating not allowing "back" which is my personal bugbear. I guess two followup questions:

 * Would it be possible for us to "run out of minutes" using github actions?
 * Should we migrate the CircleCI stuff too?

superm1:
Hmm looks like it's 2000 minutes a month. I'm not sure how much we end up using. That seems like 300ish sets of builds right now.

I wonder if we split up the docker image building separately from CI builds we could improve it.

No I don't think we should migrate all of circle.

hughsie:
> That seems like 300ish sets of builds right now

300 PRs a month would be fantastic :) I think it's probably fine for now; I'm sure one tweet could get some free credits.

> I wonder if we split up the docker image building separately from CI builds we could improve it.

We did this for gnome-softtware recently (I saw we, I have no part in it :) and it really sped up CI.

superm1:
Well 300 builds and 300 PR are different. think about how much a PR gets force pushed.

How did gnome softwares CI get split up like that? Some daily task somewhere that builds the docker images and pushes them to docker hub?

superm1:
Seems like maybe publishing the docker container to Github's container registry is the way to go about it.

>All data transferred out, when triggered by GitHub Actions, and data transferred in from any source is free. We determine you are downloading packages using GitHub Actions when you log in to GitHub Packages using a GITHUB_TOKEN.

PR #2597, opened 2020-11-17
libfwupd: Add fwupd_client_get_user_agent()

hughsie:
If we set a property, we should probably have a way to read it back...


superm1:
Hmm, I guess so.  LGTM.

PR #2599, opened 2020-11-17
libfwupd: Add fwupd_client_download_file()

hughsie:
I'm porting gnome-software to use this new API and this would be a very useful
thing to provide. No async API as we want to avoid writing temp files in most
cases -- this is just for legacy apps.

### DIFF
--- a/libfwupd/fwupd-client-sync.c
+++ b/libfwupd/fwupd-client-sync.c
@@ -1981,6 +1981,7 @@ fwupd_client_download_bytes (FwupdClient *self,
 	g_return_val_if_fail (url != NULL, NULL);
 	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), NULL);
 	g_return_val_if_fail (error == NULL || *error == NULL, NULL);
+	g_return_val_if_fail (fwupd_client_get_user_agent (self) != NULL, NULL);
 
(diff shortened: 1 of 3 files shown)


libfwupd/fwupd-client-sync.c
```diff
@@ -1997,6 +1997,56 @@ fwupd_client_download_bytes (FwupdClient *self,
...
+ * fwupd_client_download_file:
+ * @self: A #FwupdClient
+ * @url: the remote URL
+ * @file: a #GFile
+ * @flags: #FwupdClientDownloadFlags, e.g. %FWUPD_CLIENT_DOWNLOAD_FLAG_NONE
+ * @cancellable: the #GCancellable, or %NULL
+ * @error: the #GError, or %NULL
+ *
+ * Downloads data from a remote server. The fwupd_client_set_user_agent() function
+ * should be called before this method is used.
```
superm1, line 2010: Why not actually check for this and raise an error if it's not done?

hughsie: g_return_val_if_fail style, or GError?

hughsie: /me added the former, i guess it's a programmer error, not a client error

superm1 (approved):

PR #2601, opened 2020-11-18
Switch from libsoup to libcurl

hughsie:
The former drags on glib-networking and then gsettings-desktop-schemas, which
add over 5Mb to the minimal IoT and CoreOS composes. Everything already uses
libcurl (even NetworkManager!) and so this is an easy way to reduce image size.


libfwupd/fwupd-client.c
```diff
@@ -84,6 +88,18 @@ static guint signals [SIGNAL_LAST] = { 0 };
 G_DEFINE_TYPE_WITH_PRIVATE (FwupdClient, fwupd_client, G_TYPE_OBJECT)
 #define GET_PRIVATE(o) (fwupd_client_get_instance_private (o))
 
+static void
+fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
```
lucab, line 94: Should this early-return if `helper == NULL`, mimicking `g_free` behavior?

libfwupd/fwupd-client.c
```diff
@@ -284,22 +294,57 @@ fwupd_client_ensure_networking (FwupdClient *self, GError **error)
...
+				   curl_off_t dltotal,
+				   curl_off_t dlnow,
+				   curl_off_t ultotal,
+				   curl_off_t ulnow)
+{
+	FwupdClient *self = FWUPD_CLIENT (clientp);
+
+	/* calculate percentage */
+	if (dlnow <= dltotal) {
+		guint percentage = (guint) ((100 * dlnow) / dltotal);
```
lucab, line 313: `curl_off_t` is a signed type; perhaps it should have some non-negative clamping logic before casting to `guint`? Also, should division-by-0 a concern on the `dltotal` here?

libfwupd/fwupd-client.c
hughsie: I believe G_DEFINE_AUTOPTR_CLEANUP_FUNC does that for us.

libfwupd/fwupd-client.c
hughsie: Hmm, very valid. Also, we use the same function for uploads but don't look at ultotal at all.

hughsie: @lucab how about something like this?

    /* calculate percentage */
    if (dltotal >= 0 && dlnow >= 0 && dlnow <= dltotal) {
        guint percentage = (guint) ((100 * dlnow) / dltotal);
        g_debug ("download progress: %u%%", percentage);
        fwupd_client_set_percentage (self, percentage);
    } else if (ultotal >= 0 && ulnow >= 0 && ulnow <= ultotal) {
        guint percentage = (guint) ((100 * ulnow) / ultotal);
        g_debug ("upload progress: %u%%", percentage);
        fwupd_client_set_percentage (self, percentage);
    }

libfwupd/fwupd-client.c
lucab: Absolutely true, I should have double-checked the docs first, sorry for the noise.

hughsie: Not noise! I really appreciate the sanity checks!

libfwupd/fwupd-client.c
lucab: :+1: but I think you'd want a `> 0` on both `*total` conditions (instead of the `>= 0`).

hughsie: Fixed, ta.

snap/snapcraft.yaml
```diff
@@ -8,7 +8,7 @@ description: |
 
 grade: stable
 confinement: classic
-base: core18
+base: core20
```
superm1, line 11: I don't think it can go to core20 yet. I'll check out what is going on with snap. Do you know if the curl version you prescribed is actually required, or can it go older?

meson.build
```diff
@@ -208,7 +208,7 @@ endif
 libjcat = dependency('jcat', version : '>= 0.1.0', fallback : ['libjcat', 'libjcat_dep'])
 libjsonglib = dependency('json-glib-1.0', version : '>= 1.1.1')
 valgrind = dependency('valgrind', required: false)
-soup = dependency('libsoup-2.4', version : '>= 2.51.92')
+libcurl = dependency('libcurl', version : '>= 7.62.0')
```
superm1, line 211: Can it be older or is this just the first version you tried?

hughsie: It needs to be that for the CURLU support.

snap/snapcraft.yaml
hughsie: I'll admit I don't know the nuances of the snap base snap -- it's all a bit confusing.

superm1:
well this is unfortunately where the approach of split CI between build container and build job fails..  Adding new dependencies.  Will have a thought what can be done other than manually fixing the containers.

superm1:
> The former drags on glib-networking and then gsettings-desktop-schemas, which
add over 5Mb to the minimal IoT and CoreOS composes. Everything already uses
libcurl (even NetworkManager!) and so this is an easy way to reduce image size.

As I can tell, gusb is also pulling in libsoup.

snap/snapcraft.yaml
superm1: https://snapcraft.io/docs/base-snaps explains it a bit

superm1 (approved):
squash up my CI fixes and then LGTM.  We can also drop the libsoup in `dependencies.xml` after PR is merged (due to ABI check).

hughsie:
> squash up my CI fixes and then LGTM

Thanks for your help with this; I'd like to finish the gnome-software plugin work first -- we actually grab the SoupSession (using `g_object_get`, yuk!) in later gnome-software versions and I can use that as an excuse to port the plugin to the new async stuff we have.

hughsie:
If it's interesting, https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/538

superm1:
> If it's interesting, https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/538

Hmm so that means that moving to fwupd 1.5.2 and using "old" GS is off the table?

hughsie:
> old GS is off the table

I think so; it wasn't my finest moment of forward planning to use the GObject property like that. I think backporting the fwupd plugin changes will be fine, but we might want to hold off the libcurl stuff until 1.6.x even.

superm1:
As a thought, what about offering a compilation option of which one to use, with the intention that eventually soup is dropped?

superm1:
Or maybe use both and to "deprecate" the soup one for an ABI change (which really it sounds like it might be).

hughsie:
> Or maybe use both

The l33t hacker in me wants to dlopen the libsoup.so at runtime and just create the object that way, so there's no explicit library dep. It's kinda sucky the very same binary fwupd.rpm package gets used on coreos and the Fedora Workstation.

superm1:
> > Or maybe use both
> 
> The l33t hacker in me wants to dlopen the libsoup.so at runtime and just create the object that way, so there's no explicit library dep. It's kinda sucky the very same binary fwupd.rpm package gets used on coreos and the Fedora Workstation.

I guess I go back and forth - besides GS does any other fwupd client steal the soup session?  If not, as long as gnome-software "cleanly" aborts/blows up let's not make a bunch of noise about the small ABI break.  We can make sure that a new gnome-software and fwupd 1.5.2 flow into distros around the same time.

superm1:
The only other one that came to mind to maybe be problematic is the snap store (gnome-software fork IIRC?).  But they carry the fwupd library as a build time dependency.  So I'd think when they move up to 1.5.2 in their snap they can update their fwupd support at the same time.  CC @kenvandine FYI ^

kenvandine:
> The only other one that came to mind to maybe be problematic is the snap store (gnome-software fork IIRC?). But they carry the fwupd library as a build time dependency. So I'd think when they move up to 1.5.2 in their snap they can update their fwupd support at the same time. CC @kenvandine FYI ^

Thanks for the heads up!

superm1:
> If not, as long as gnome-software "cleanly" aborts/blows up

Running G-S master against this branch I don't see a functional issue.  I guess the user agent is going to be wrong, and if you had a proxy set it might not work properly.

hughsie:
> Running G-S master against this branch I don't see a functional issue

Does firmware still get downloaded tho? I might try the dlopen thing myself tomorrow and see if we can have the best of both worlds, i.e. dropping the dep and not breaking internal ABI.

superm1:
> Does firmware still get downloaded tho? I might try the dlopen thing myself tomorrow and see if we can have the best of both worlds, i.e. dropping the dep and not breaking internal ABI.

I only tried refreshing metadata since I didn't have something pending an update at the moment.

hughsie:
Answering my own question:

    14:05:56:0937 libsoup soup_session_send_message: assertion 'SOUP_IS_SESSION (session)' failed

I'm working on something horrific.

hughsie:
@superm1 try to review the above commit without being a little bit sick in your mouth.

libfwupd/fwupd-client.c
```diff
@@ -71,7 +80,7 @@ enum {
 	PROP_PERCENTAGE,
 	PROP_DAEMON_VERSION,
 	PROP_TAINTED,
-	PROP_SOUP_SESSION,
+	PROP_SOUP_SESSION, /* unused ABI */
```
superm1, line 83: this comment is wrong now..

libfwupd/fwupd-client.c
```diff
@@ -4240,6 +4250,54 @@ fwupd_client_upload_bytes_finish (FwupdClient *self, GAsyncResult *res, GError *
...
+	}
+
+	/* load the library at rumtime, leaking the module */
+	if (priv->soup_module == NULL) {
+		g_autofree gchar *fn = NULL;
+		const gchar *libdirs[] = {
+			FWUPD_LIBDIR,
+			"/usr/lib64",
+			"/usr/lib",
+			NULL };
```
superm1, line 4280: hardcoded paths?  Is there a cleaner way to do this?

libfwupd/fwupd-client.c
```diff
@@ -4240,6 +4250,54 @@ fwupd_client_upload_bytes_finish (FwupdClient *self, GAsyncResult *res, GError *
...
+		}
+	}
+	if (priv->soup_module == NULL) {
+		g_warning ("failed to find libsoup library");
+		return;
+	}
+	if (!g_module_symbol (priv->soup_module,
+			      "soup_session_new",
+			      (gpointer *) &func)) {
+		g_warning ("failed to find soup_session_get_type()");
```
superm1, line 4294: maybe unload the module in this error path?

libfwupd/fwupd-client.c
```diff
@@ -4240,6 +4250,54 @@ fwupd_client_upload_bytes_finish (FwupdClient *self, GAsyncResult *res, GError *
...
+	/* load the library at rumtime, leaking the module */
+	if (priv->soup_module == NULL) {
+		g_autofree gchar *fn = NULL;
+		const gchar *libdirs[] = {
+			FWUPD_LIBDIR,
+			"/usr/lib64",
+			"/usr/lib",
+			NULL };
+		for (guint i = 0; libdirs[i] != NULL && priv->soup_module == NULL; i++) {
+			fn = g_build_filename (libdirs[i], "libsoup-2.4.so.1", NULL);
```
superm1, line 4282: hardcoding this extension won't work in !linux

libfwupd/fwupd-client.c
hughsie: Fixed.

libfwupd/fwupd-client.c
hughsie: I thought I'd leave it be, otherwise we'll keep trying to open libsoup for every property read.

libfwupd/fwupd-client.c
hughsie: Valid, but I don't think it matters; this is just for gnome-software.

libfwupd/fwupd-client.c
hughsie: I originally just had FWUPD_LIBDIR but that's prefixed with the, well, prefix and then it breaks for when you're developing fwupd. Perhaps that's actually okay, and we just use FWUPD_LIBDIR which will be the right thing when installed systemwide.

libfwupd/fwupd-client.c
superm1: But don't you want to do to that?  What if someone didn't have libsoup installed and then installed it later at runtime?

libfwupd/fwupd-client.c
superm1: Maybe don't even bother compiling the code for mingw builds?

libfwupd/fwupd-client.c
superm1: What I was thinking was a `fu_common_get_path` type of solution, which will let this work well in the snap too, and also for developing fwupd.

libfwupd/fwupd-client.c
hughsie: Possible indeed. Fixed.

libfwupd/fwupd-client.c
hughsie: Yes, I'd feel a lot happier if all this evil was behind some kind of ifdef. Will fix.

libfwupd/fwupd-client.c
hughsie: fu_common_get_path is libfwupdplugin, not libfwupd  :(

superm1: ah..  Well in that case I guess yeah just make `FWUPD_LIBDIR` only.  Need to keep reminding myself it's just for gnome-software support, not general.  Snap-store will just need move at the same time they upgrade their library.

And `libfwupd3` can't come soon enough :)

hughsie: I actually think this would be fine for snap:

		tmp = g_getenv ("SNAP");
		if (tmp != NULL) {
			g_ptr_array_add (libdirs,
					 g_build_filename (tmp, FWUPD_LIBDIR, NULL));
		}

hughsie: although, maybe that's not required at all if ubuntu-software, ahem, gnome-software is using the system libfwupd and not the snapped version?

hughsie: I do like the simplicity of the FWUPD_LIBDIR-only approach -- it's just fixing the one problem

superm1: The snap store is fully confined and talks to the host daemon.  So expect the client library to move with the confined store itself.
@kenvandine am I right that snap store can just update interface at the same time the library version bumps?

hughsie: >  So expect the client library to move with the confined store itself

That seems sane; I think we don't need the snap-specific path then.

hughsie: I've made it a lot simpler, we can dig this back up if indeed snap needs the fix.

kenvandine: > The snap store is fully confined and talks to the host daemon. So expect the client library to move with the confined store itself.
> @kenvandine am I right that snap store can just update interface at the same time the library version bumps?

Yes, we build libfwupd into the snap and talk to the host service.  So no dependency on the lib on the host.  So FWUPD_LIBDIR should be fine.

libfwupd/fwupd-client.c
```diff
@@ -4471,12 +4517,12 @@ fwupd_client_class_init (FwupdClientClass *klass)
...
 	 *
-	 * The libsoup session.
+	 * The libsoup session, now unused.
 	 *
 	 * Since: 1.4.5
 	 */
-	pspec = g_param_spec_object ("soup-session", NULL, NULL, SOUP_TYPE_SESSION,
-				     G_PARAM_READWRITE | G_PARAM_STATIC_NAME);
+	pspec = g_param_spec_object ("soup-session", NULL, NULL, G_TYPE_OBJECT,
+				     G_PARAM_READABLE | G_PARAM_STATIC_NAME);
```
superm1, line 4525: isn't this also ABI breaking?  I guess _gnome-software didn't actually use it though, so don't make too much noise_?

superm1 (approved):
I already look forward to our ABI bump to drop this soup emulation code.  I think in the distro packages we may consider to change the defaults to not compile it once we get gnome-software new stuff landed in the big ones.

libfwupd/fwupd-client.c
hughsie: > so don't make too much noise

Right! Sssh! :)

Issue #2600, opened 2020-11-18
fwupd_client_get_devices never returns


aleixpol:
**Describe the bug**
We call fwupd_client_get_devices() from a worker thread. It never returns leaving the thread dangling.
There is no errors or warnings so I don't really know how to take an action.

```
#0 syscall() at /usr/lib/libc.so.6
#1 g_cond_wait() at /usr/lib/libglib-2.0.so.0
#2 ??() at /usr/lib/libglib-2.0.so.0
#3 g_main_loop_run() at /usr/lib/libglib-2.0.so.0
#4 fwupd_client_connect() at /home/apol/devel/playground/fwupd/libfwupd/fwupd-client-sync.c:98
#5 fwupd_client_get_devices() at /home/apol/devel/playground/fwupd/libfwupd/fwupd-client-sync.c:136
#6 FwupdBackend::checkForUpdates()::$_6::operator()() const() at /home/apol/devel/frameworks/discover/libdiscover/backends/FwupdBackend/FwupdBackend.cpp:419
#7 QtConcurrent::StoredFunctorCall0<_GPtrArray*, FwupdBackend::checkForUpdates()::$_6>::runFunctor()() at /home/apol/devel/kde5/include/QtConcurrent/qtconcurrentstoredfunctioncall.h:60
#8 QtConcurrent::RunFunctionTask<_GPtrArray*>::run() at /home/apol/devel/kde5/include/QtConcurrent/qtconcurrentrunbase.h:108
#9 QThreadPoolThread::run() at /home/apol/devel/frameworks/qt5/qtbase/src/corelib/thread/qthreadpool.cpp:100
#10 QThreadPrivate::start() at /home/apol/devel/frameworks/qt5/qtbase/src/corelib/thread/qthread_unix.cpp:329
#11 start_thread() at /usr/lib/libpthread.so.0
#12 clone() at /usr/lib/libc.so.6
```
https://bugs.kde.org/show_bug.cgi?id=428175

**Steps to Reproduce**
It's easy to reproduce using Discover, we could probably come up with a test app.

**Expected behavior**
Either getting the devices (which we used to get, so it's likely a regression) or an error message

**fwupd version information**
1.5.1
I also tried running from master, I get the same issue. Otherwise it was pacman (ArchLinux).

I'll be happy to provide any additional information required, although it seems like a problem with the library (?).

superm1:
Does the daemon crash mid thread perhaps?

[...]

PR #2605, opened 2020-11-19
Fix sync method when called from threads without a context

hughsie:
Set the thread-default context for the current thread before performing the
async operation.

Hopefully fixes https://github.com/fwupd/fwupd/issues/2600

### DIFF
--- a/libfwupd/fwupd-client-sync.c
+++ b/libfwupd/fwupd-client-sync.c
@@ -21,6 +21,7 @@ typedef struct {
 	gchar		*str;
 	GError		*error;
 	GPtrArray	*array;
+	GMainContext	*context;
 	GMainLoop	*loop;
 	GVariant	*val;
 	GHashTable	*hash;
@@ -45,6 +46,7 @@ fwupd_client_helper_free (FwupdClientHelper *helper)
 		g_object_unref (helper->device);
 	g_free (helper->str);
 	g_main_loop_unref (helper->loop);
+	g_main_context_unref (helper->context);
 	g_free (helper);
 }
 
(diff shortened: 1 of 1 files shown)


libfwupd/fwupd-client-sync.c
```diff
@@ -1357,6 +1408,7 @@ fwupd_client_refresh_remote (FwupdClient *self,
 					   fwupd_client_refresh_remote_cb,
 					   helper);
 	g_main_loop_run (helper->loop);
+	g_main_context_pop_thread_default (helper->context);
```
superm1, line 1411: this is missing the matching `g_main_context_push_thread_default (helper->context);`

hughsie: Eagle eyes! Fixed, thanks.

superm1: I wish I could blame my eyes, but I tried it and saw it broken :)

superm1 (approved):
LGTM, and it fixes the underlying issue in #2600 it seems too (at least the reproducer doesn't fail anymore).

aleixpol:
It does fix the issue, thank you very much!

Would you know under which version this will be fixed so I can tell our packagers to get that version? 1.5.2?

hughsie:
1.5.2 indeed, which I can do next week if you remind me closer to the time. This week EOUTOFTIME :)

aleixpol:
> 1.5.2 indeed, which I can do next week if you remind me closer to the time. This week EOUTOFTIME :)

No problem, will ping next week.

PR #2606, opened 2020-11-19
Only refresh remotes older than a specified cache age

hughsie:
Type of pull request:
- [X] Code fix

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -1129,8 +1129,9 @@ fu_util_download_metadata (FuUtilPrivate *priv, GError **error)
 			continue;
 		download_remote_enabled = TRUE;
 		g_print ("%s %s\n", _("Updating"), fwupd_remote_get_id (remote));
-		if (!fwupd_client_refresh_remote (priv->client, remote,
-						  priv->cancellable, error))
+		if (!fwupd_client_refresh_remote_full (priv->client, remote,
+						       10, /* seconds */
+						       priv->cancellable, error))
 			return FALSE;
 	}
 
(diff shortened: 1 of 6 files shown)


src/fu-util.c
```diff
@@ -1129,8 +1129,9 @@ fu_util_download_metadata (FuUtilPrivate *priv, GError **error)
 			continue;
 		download_remote_enabled = TRUE;
 		g_print ("%s %s\n", _("Updating"), fwupd_remote_get_id (remote));
-		if (!fwupd_client_refresh_remote (priv->client, remote,
-						  priv->cancellable, error))
+		if (!fwupd_client_refresh_remote_full (priv->client, remote,
+						       10, /* seconds */
+						       priv->cancellable, error))
```
superm1, line 1134: shouldn't lines 1101 to 1118 change too?

hughsie: Nah, that's checking the oldest remote, or would your proposal be that we just set the 24h limit for all remotes? If so, how would we know that all calls to fwupd_client_refresh_remote didn't do anything and we need to show the "Use --force to refresh again" message?

superm1: It could be the behavior is to return a nothing to do error if the remote is already updated.

hughsie: I'm really not sure about this patch; it's usable in gnome-software for this specific use case, but not really in the "_finding out what we actually did_" case. It's adding 80 lines to save 5 in one consumer; it seems way too heavy for the saving.

superm1: >It's adding 80 lines to save 5 in one consumer; it seems way too heavy for the saving.

Good perspective.  Axe it then!

PR #2614, opened 2020-11-24
modem-manager minor plugin fixes and updates

aleksander0m:
Type of pull request:
- [ X] Code fix

The MR includes some fixes to the modem-manager plugin, and some minor improvements.

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -111,6 +111,28 @@ fu_mm_device_get_port_qmi_ifnum (FuMmDevice *device)
 }
 
 static gboolean
+validate_firmware_update_method (MMModemFirmwareUpdateMethod methods, GError **error)
+{
+	static const MMModemFirmwareUpdateMethod supported_combinations[] = {
+		MM_MODEM_FIRMWARE_UPDATE_METHOD_FASTBOOT,
+		MM_MODEM_FIRMWARE_UPDATE_METHOD_QMI_PDC | MM_MODEM_FIRMWARE_UPDATE_METHOD_FASTBOOT,
+	};
+	g_autofree gchar *methods_str = NULL;
+
+	methods_str = mm_modem_firmware_update_method_build_string_from_mask (methods);
+	for (guint i = 0; i < G_N_ELEMENTS (supported_combinations); i++) {
+		if (supported_combinations[i] == methods) {
+			g_debug ("valid firmware update combination: %s", methods_str);
+			return TRUE;
+		}
+	}
+
+	g_set_error (error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED,
+		     "invalid firmware update combination: %s", methods_str);
+	return FALSE;
+}
+
+static gboolean
 fu_mm_device_probe_default (FuDevice *device, GError **error)
 {
 	FuMmDevice *self = FU_MM_DEVICE (device);
@@ -138,6 +160,10 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 				     "modem cannot be put in programming mode");
 		return FALSE;
 	}
+
+	/* make sure the combination is supported */
+	if (!validate_firmware_update_method (self->update_methods, error))
+		return FALSE;
 
 	/* various fastboot commands */
 	if (self->update_methods & MM_MODEM_FIRMWARE_UPDATE_METHOD_FASTBOOT) {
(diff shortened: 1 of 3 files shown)


plugins/modem-manager/fu-mm-device.c
```diff
@@ -840,9 +874,9 @@ fu_mm_device_udev_add_port (FuMmDevice	*self,
 {
 	g_return_if_fail (FU_IS_MM_DEVICE (self));
 
-	/* cdc-wdm ports always added unless one already set */
 	if (g_str_equal (subsystem, "usbmisc") &&
-	    (self->port_qmi == NULL)) {
+	    (self->port_qmi == NULL) &&
+	    (ifnum >= 0) && (ifnum == self->port_qmi_ifnum)) {
```
hughsie, line 879: no inner brackets pls

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -56,8 +56,22 @@ fu_mm_utils_get_udev_port_info (GUdevDevice	 *device,
 
 	if (out_port_ifnum != NULL)
 		*out_port_ifnum = port_ifnum;
-	if (out_device_sysfs_path != NULL)
-		*out_device_sysfs_path = g_steal_pointer (&device_sysfs_path);
+	if (out_device_sysfs_path != NULL) {
+		/* if device sysfs path was set for a previous port,
+		 * make sure they are all equal */
+		if (*out_device_sysfs_path != NULL) {
+			if (g_strcmp0 (device_sysfs_path, *out_device_sysfs_path) != 0) {
```
hughsie, line 63: It does feel a bit dodgy using out_device_sysfs_path like this

aleksander0m: Maybe I could always return the device sysfs path as output, and do the check in the caller side?

hughsie: Ya, that works too.

plugins/modem-manager/fu-mm-device.c
aleksander0m: Done!

plugins/modem-manager/fu-mm-utils.c
aleksander0m: Done!

plugins/modem-manager/fu-mm-device.c
```diff
@@ -216,13 +251,20 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
...
 		fu_mm_utils_get_port_info (self->port_at, &device_sysfs_path, &self->port_at_ifnum, NULL);
-	} else if (self->port_qmi != NULL) {
-		fu_mm_utils_get_port_info (self->port_qmi, &device_sysfs_path, NULL, NULL);
-	} else {
-		g_warn_if_reached ();
+	}
+	if (self->port_qmi != NULL) {
+		g_autofree gchar *qmi_device_sysfs_path = NULL;
+		fu_mm_utils_get_port_info (self->port_qmi, &qmi_device_sysfs_path, &self->port_qmi_ifnum, NULL);
+		if (!device_sysfs_path && qmi_device_sysfs_path) {
```
hughsie, line 260: we tend to use != NULL and == NULL for pointers

plugins/modem-manager/fu-mm-device.c
```diff
@@ -130,6 +154,17 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
...
 	}
+	firmware_update_str = mm_modem_firmware_update_method_build_string_from_mask (self->update_methods);
+
+	if (!validate_firmware_update_method (self->update_methods)) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "invalid firmware update combination: %s",
+			     firmware_update_str);
+		return FALSE;
```
hughsie, line 165: could you move this to validate_firmware_update_method() perhaps? Then it returns gboolean and GError like a lot of other things in fwupd. Ta. Then LGTM.

plugins/modem-manager/fu-mm-device.c
aleksander0m: I try to keep that in my mind all the time, looks like I missed this one :D Changed now

plugins/modem-manager/fu-mm-device.c
aleksander0m: That's a nice change, done now

aleksander0m: Done

plugins/modem-manager/fu-mm-device.c
aleksander0m: Done

superm1 (approved):

aleksander0m:
Not sure if I should be the one merging this when it gets approved, or if either of you should do it :D

superm1:
> Not sure if I should be the one merging this when it gets approved, or if either of you should do it :D

6 eggs one hand, half dozen in the other. The important part is another set of eyeballs saw it.

PR #2621, opened 2020-11-28
fu-main: Fix sender_features insert with wrong size

zeldin:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

`fu_main_daemon_method_call` stores a pointer to a `guint64` in the `priv->sender_features` hashtable.  However, wherever this pointer is extracted from the hashtable, it is casted to a `FwupdFeatureFlags*` and dereferenced.  Since `guint64` does not have the same size as `FwupdFeatureFlags`, this is undefined behaviour.  On a big endian system, the typical result will be that the higher order bits of the `guint64` are accessed, giving a `0` result.

### DIFF
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -1532,11 +1532,13 @@ fu_main_daemon_method_call (GDBusConnection *connection, const gchar *sender,
 		return;
 	}
 	if (g_strcmp0 (method_name, "SetFeatureFlags") == 0) {
-		guint64 feature_flags = 0;
-		g_variant_get (parameters, "(t)", &feature_flags);
-		g_debug ("Called %s(%" G_GUINT64_FORMAT ")", method_name, feature_flags);
+		FwupdFeatureFlags feature_flags;
+		guint64 feature_flags_u64 = 0;
+		g_variant_get (parameters, "(t)", &feature_flags_u64);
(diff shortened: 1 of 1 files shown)


hughsie:
@superm1 do you agree with this, or should we just typedef FwupdFeatureFlags to guint64 so it has the same size across all architectures?

hughsie:
Alternate proposal: https://github.com/fwupd/fwupd/pull/2623 -- @zeldin does that branch work for you as well?

zeldin:
@hughsie I'm not sure how to test.  Now that I have already installed the custom NIC firmware from the lvfs-testing branch, it seems like the `FWUPD_FEATURE_FLAG_SWITCH_BRANCH` flag is not needed anymore...

hughsie:
> I'm not sure how to test. 

`fwupdmgr reinstall` or `fwupdmgr downgrade` perhaps?

zeldin:
`fwupdmgr reinstall` works without the feature flag (tested without applying any of the patches).  I can't downgrade because the two previous releases do not work with 1.5.2.  Is there a way to revert to the original (vendor) firmware?

superm1:
I think the approach proposed makes sense considering ABI implications, but we probably need to better document that this needs to be fixed/changed next time we bump ABI.

hughsie:
I've added two things to `libfwupd/README.md` -- feel free to add any more that I've missed.

PR #2622, opened 2020-11-28
bcm57xx: Fix CRC check

zeldin:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

The function `fu_bcm57xx_nvram_crc` calls `GUINT32_FROM_BE` on a locally computed number, which makes absolutely no sense.  AFAICT this is an attempt to wall-paper over the fact that the call to `fu_common_read_uint32_safe` incorrectly uses `G_BIG_ENDIAN` instead of `G_LITTLE_ENDIAN` to read the (little endian) CRC value from the fw image.

This "worked" on little-endian, since `GUINT32_FROM_BE` would swap the CRC backwards, so that it compared equal to the CRC fetched from the image, which was also backwards.  But on big-endian `GUINT32_FROM_BE` naturally does nothing and so the comparison fails.

### DIFF
--- a/plugins/bcm57xx/fu-bcm57xx-common.c
+++ b/plugins/bcm57xx/fu-bcm57xx-common.c
@@ -14,7 +14,7 @@
 guint32
 fu_bcm57xx_nvram_crc (const guint8 *buf, gsize bufsz)
 {
-	return GUINT32_FROM_BE(fu_common_crc32 (buf, bufsz));
+	return GUINT32_SWAP_LE_BE (fu_common_crc32 (buf, bufsz));
 }
 
 gboolean
--- a/plugins/bcm57xx/fu-bcm57xx-common.c
+++ b/plugins/bcm57xx/fu-bcm57xx-common.c
@@ -14,7 +14,7 @@
 guint32
 fu_bcm57xx_nvram_crc (const guint8 *buf, gsize bufsz)
 {
-	return GUINT32_SWAP_LE_BE (fu_common_crc32 (buf, bufsz));
+	return fu_common_crc32 (buf, bufsz);
 }
 
 gboolean
@@ -27,7 +27,7 @@ fu_bcm57xx_verify_crc (GBytes *fw, GError **error)
 
 	/* expected */
 	if (!fu_common_read_uint32_safe (buf, bufsz, bufsz - sizeof(guint32),
-					 &crc_file, G_BIG_ENDIAN, error))
+					 &crc_file, G_LITTLE_ENDIAN, error))
 		return FALSE;
 
 	/* reality */
--- a/plugins/bcm57xx/fu-bcm57xx-dict-image.c
+++ b/plugins/bcm57xx/fu-bcm57xx-dict-image.c
@@ -71,7 +71,7 @@ fu_bcm57xx_dict_image_write (FuFirmwareImage *image, GError **error)
 
 	/* add CRC */
 	crc = fu_bcm57xx_nvram_crc (buf, bufsz);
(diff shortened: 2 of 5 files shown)


zeldin:
Gah, after installing the custom firmware, I now get

> failed to parse existing firmware: failed to parse header: invalid CRC, expected 0xbd3bc0ad got: 0xadc03bbd

from `fwupdmgr reinstall` _with_ the fixes.  Is it perhaps the case that the CRC _should_ be big endian (i.e. use of `G_BIG_ENDIAN` was correct, but use of `GUINT32_FROM_BE` was still wrong), but the upgrade image in lvfs mistakingly has the CRC in the wrong byte order?

What I don't get is how `fwupdmgr reinstall` could have worked on little endian with the old code...

zeldin:
Ah, now I think I see.   The write functions do

`fu_byte_array_append_uint32 (buf, fu_bcm57xx_nvram_crc (buf->data, buf->len), G_BIG_ENDIAN);`

So with the old (incorrect) implementation of `fu_bcm57xx_nvram_crc`, the CRC would be swapped twice on little endian, and then written as little endian.  Whereas on big endian it get swapped 0 times, and written as big endian.

hughsie:
@meklort looks sane to you too?

zeldin:
So, AFAICT, the CRC should always be little endian in the image, but the code uses G_BIG_ENDIAN in several locations to store, or compare against, the return value of `fu_bcm57xx_nvram_crc`.  Therefore, fu_bcm57xx_nvram_crc is expected to return the CRC backwards.  Rather than changing all the uses of G_BIG_ENDIAN, I modified the change to always swap the return value of `fu_bcm57xx_nvram_crc`.  While I still think this is weird, it should at least work everywhere without large changes to the code base.

meklort:
The change looks fine to me.

hughsie:
> it should at least work everywhere without large changes to the code base

Fwiw, I'd prefer the large changes if it makes it more logically correct and easily understood.

zeldin:
> Fwiw, I'd prefer the large changes if it makes it more logically correct and easily understood.

Ok.  I was a little worried that I wouldn't be able to test all the code paths I'd need to change, but I guess it should be safe to just look for all call sites of `fu_bcm57xx_nvram_crc` and change `G_BIG_ENDIAN` to `G_LITTLE_ENDIAN` in each, and it should be functionally equivalent.  I'll make this refactoring.

zeldin:
I tested this refactoring with `fwupdmgr reinstall`, and I did an ocular inspection of each code path that the new value didn't "leak" anywhere that I didn't update.  The exception is the error message, which now prints the CRC (both the calculated one and the one from the image) in the natural digit order rather than the order in which they appear in the image file.

hughsie:
> which now prints the CRC in the natural digit order

I think that's entirely correct BTW. Thanks for the comprehensive fix!

hughsie:
@meklort can you just verify that the mix of G_BIG_ENDIAN and G_LITTLE_ENDIAN is expected? Thanks.

meklort:
I've tested this on a LE system and it doesn't appear to break anything.

If my understanding is correct form @zeldin's comments, then the crc routine results in a different value returned for BE vs LE, and so no conversion is needed there. I'll need to do some testing here on BE for bcmflash, but I'll likely need to update it as well. In any case, I'm fine with the mix, although I agree it's looks like an artifact of the crc function.

zeldin:
> If my understanding is correct form @zeldin's comments, then the crc routine results in a different value returned for BE vs LE,
No, the CRC routine operates on bytes and returns the same value regardless of host endianness.  Endianness only comes into play when storing or retrieving the CRC to/from the image, which has fields with fixed endianness.

PR #2623, opened 2020-11-28
Ensure FwupdFeatureFlags is always 64 bits in size

hughsie:
Based on a patch from Marcus Comstedt <marcus@mc.pp.se>, many thanks.


hughsie:
@superm1 not sure if this is an ABI break or not... I'm guessing yes; ideas welcome.

PR #2631, opened 2020-12-01
trivial: fu-engine: If a device is using delayed activation don't flag failure

superm1:
Should hopefully fix this failure:
```
Plugin thunderbolt
UpdateError device version not updated on success, 43.00 != 40.00
VersionNew 43.00
VersionOld 40.00
```

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
@hughsie if you can check if you agree this should help the problems that showed up recently at https://fwupd.org/lvfs/firmware/3829/analytics/reports/3 causing a downgrade.

hughsie (approved):
Makes sense to me.

PR #2632, opened 2020-12-02
Fall back to the misc device for serio devices

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):

hughsie:
@superm1 the other thing i thought of was maybe we want to fall back to misc for *all* devices, but serio was the first one I found without a dev node, and I wasn't sure if falling back always would break another plugin without realizing. Your opinion welcome!

superm1:
I think this is fine.
 Can always change it later if we have something else that needs a misc node.  Leave a breadcrumb for future selves to think about in a comment.

PR #2633, opened 2020-12-02
Actually implement rescan in FuUdevDevice

hughsie:
This allows the plugin to force a reload of the device using a new GUdevDevice
object. This is required as the values are cached in the immutable object.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -1202,6 +1202,40 @@ fu_udev_device_open (FuDevice *device, GError **error)
 }
 
 static gboolean
+fu_udev_device_rescan (FuDevice *device, GError **error)
+{
+#ifdef HAVE_GUDEV
+	FuUdevDevice *self = FU_UDEV_DEVICE (device);
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	const gchar *sysfs_path;
+	g_autoptr(GUdevClient) udev_client = g_udev_client_new (NULL);
+	g_autoptr(GUdevDevice) udev_device = NULL;
+
+	/* never set */
+	if (priv->udev_device == NULL) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "rescan with no previous device");
+		return FALSE;
+	}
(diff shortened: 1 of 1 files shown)


superm1 (commented):
there are plugins that use udev and implement rescan already so I think you need to check for subclassed no?

superm1:
@ArchangeGabriel can you please check what's up with the TPM stuff 404'ing in CI?  Did a package or repo change?

hughsie:
@superm1 in `thunderbolt/fu-thunderbolt-device.c` there is already logic to handle rescan, and I don't really want to touch it is it's working. The TBT rescan will "overwrite" the baseclass rescan -- which I think means we don't break anything. Synaptics MST is the same; we want to keep the existing logic.

ArchangeGabriel:
> @ArchangeGabriel can you please check what's up with the TPM stuff 404'ing in CI? Did a package or repo change?

I don’t remember/see exactly at which step pacman cache gets updated in the CI, but this is a cache issue: pacman package list contains swtpm 0.5.1-1, but the mirror has been updated to 0.5.1-2 (released on 2020-12-02 04:41 UTC). It could be an unlucky timing issue: the package list was refreshed, and before it got to downloading the package that one has been replaced with a new version. Re-running should very likely fix.

superm1:
> I don’t remember/see exactly at which step pacman cache gets updated in the CI, but this is a cache issue: pacman package list contains swtpm 0.5.1-1, but the mirror has been updated to 0.5.1-2 (released on 2020-12-02 04:41 UTC). It could be an unlucky timing issue: the package list was refreshed, and before it got to downloading the package that one has been replaced with a new version. Re-running should very likely fix.

Is the pacman cache built in the docker container or in the scripts that run in the arch script?  If it's in the container, I think we should add a step to the arch script to refresh at that time.  We split up the container generation to be nightly and automatic and then run the CI on demand for the job.  So we always have a good "baseline" even if distros break (famous last words I guess...)

superm1:
I just re-ran it and it still fails.  I see it has `-S` in the call too.

superm1:
> The TBT rescan will "overwrite" the baseclass rescan

Ah OK.  Fine in this case then.

superm1 (approved):

PR #2638, opened 2020-12-03
Allow setting the GMainContext when used for sync methods

hughsie:
Use the 'default' main context by default so that we still get the signals
delivered via g_object_notify() but allow the library user to specify an
alternate GMainContext as required.

Using g_main_context_push_thread_default() works for PackageKit as signals are
returned using g_signal_emit() which is synchronous, and so doesn't use the
main context at all.

Use the same style of API which we did for libgusb all those years ago instead.

### DIFF
--- a/libfwupd/fwupd-client-sync.c
+++ b/libfwupd/fwupd-client-sync.c
@@ -51,11 +51,11 @@ fwupd_client_helper_free (FwupdClientHelper *helper)
 }
 
 static FwupdClientHelper *
-fwupd_client_helper_new (void)
+fwupd_client_helper_new (FwupdClient *self)
 {
 	FwupdClientHelper *helper;
 	helper = g_new0 (FwupdClientHelper, 1);
-	helper->context = g_main_context_new ();
+	helper->context = fwupd_client_get_main_context (self);
 	helper->loop = g_main_loop_new (helper->context, FALSE);
 	return helper;
 }
@@ -90,17 +90,16 @@ fwupd_client_connect_cb (GObject *source, GAsyncResult *res, gpointer user_data)
 gboolean
 fwupd_client_connect (FwupdClient *self, GCancellable *cancellable, GError **error)
 {
-	g_autoptr(FwupdClientHelper) helper = fwupd_client_helper_new ();
+	g_autoptr(FwupdClientHelper) helper = NULL;
 
 	g_return_val_if_fail (FWUPD_IS_CLIENT (self), FALSE);
 	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), FALSE);
 	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
 	/* call async version and run loop until complete */
-	g_main_context_push_thread_default (helper->context);
+	helper = fwupd_client_helper_new (self);
 	fwupd_client_connect_async (self, cancellable, fwupd_client_connect_cb, helper);
 	g_main_loop_run (helper->loop);
-	g_main_context_pop_thread_default (helper->context);
 	if (!helper->ret) {
 		g_propagate_error (error, g_steal_pointer (&helper->error));
 		return FALSE;
@@ -131,7 +130,7 @@ fwupd_client_get_devices_cb (GObject *source, GAsyncResult *res, gpointer user_d
 GPtrArray *
 fwupd_client_get_devices (FwupdClient *self, GCancellable *cancellable, GError **error)
 {
-	g_autoptr(FwupdClientHelper) helper = fwupd_client_helper_new ();
+	g_autoptr(FwupdClientHelper) helper = NULL;
 
 	g_return_val_if_fail (FWUPD_IS_CLIENT (self), NULL);
 	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), NULL);
@@ -142,11 +141,10 @@ fwupd_client_get_devices (FwupdClient *self, GCancellable *cancellable, GError *
 		return NULL;
 
 	/* call async version and run loop until complete */
-	g_main_context_push_thread_default (helper->context);
+	helper = fwupd_client_helper_new (self);
 	fwupd_client_get_devices_async (self, cancellable,
 					fwupd_client_get_devices_cb, helper);
 	g_main_loop_run (helper->loop);
-	g_main_context_pop_thread_default (helper->context);
 	if (helper->array == NULL) {
 		g_propagate_error (error, g_steal_pointer (&helper->error));
 		return NULL;
@@ -177,7 +175,7 @@ fwupd_client_get_plugins_cb (GObject *source, GAsyncResult *res, gpointer user_d
 GPtrArray *
 fwupd_client_get_plugins (FwupdClient *self, GCancellable *cancellable, GError **error)
 {
-	g_autoptr(FwupdClientHelper) helper = fwupd_client_helper_new ();
+	g_autoptr(FwupdClientHelper) helper = NULL;
 
 	g_return_val_if_fail (FWUPD_IS_CLIENT (self), NULL);
 	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), NULL);
@@ -188,11 +186,10 @@ fwupd_client_get_plugins (FwupdClient *self, GCancellable *cancellable, GError *
 		return NULL;
 
 	/* call async version and run loop until complete */
-	g_main_context_push_thread_default (helper->context);
+	helper = fwupd_client_helper_new (self);
 	fwupd_client_get_plugins_async (self, cancellable,
 					fwupd_client_get_plugins_cb, helper);
 	g_main_loop_run (helper->loop);
-	g_main_context_pop_thread_default (helper->context);
 	if (helper->array == NULL) {
 		g_propagate_error (error, g_steal_pointer (&helper->error));
 		return NULL;
@@ -223,7 +220,7 @@ fwupd_client_get_history_cb (GObject *source, GAsyncResult *res, gpointer user_d
 GPtrArray *
 fwupd_client_get_history (FwupdClient *self, GCancellable *cancellable, GError **error)
 {
-	g_autoptr(FwupdClientHelper) helper = fwupd_client_helper_new ();
+	g_autoptr(FwupdClientHelper) helper = NULL;
 
 	g_return_val_if_fail (FWUPD_IS_CLIENT (self), NULL);
 	g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), NULL);
@@ -234,11 +231,10 @@ fwupd_client_get_history (FwupdClient *self, GCancellable *cancellable, GError *
 		return NULL;
 
 	/* call async version and run loop until complete */
-	g_main_context_push_thread_default (helper->context);
+	helper = fwupd_client_helper_new (self);
 	fwupd_client_get_history_async (self, cancellable,
 					fwupd_client_get_history_cb, helper);
 	g_main_loop_run (helper->loop);
-	g_main_context_pop_thread_default (helper->context);
 	if (helper->array == NULL) {
 		g_propagate_error (error, g_steal_pointer (&helper->error));
 		return NULL;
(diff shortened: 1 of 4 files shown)


superm1:
@aleixpol Can you please check this still works properly for KDE discover?

aleixpol:
I just checked, this gets stuck again:
```
#0  0x00007ffff4d56d5d in syscall () at /usr/lib/libc.so.6
#1  0x00007ffff20af0a1 in g_cond_wait () at /usr/lib/libglib-2.0.so.0
#2  0x00007ffff2060f6f in  () at /usr/lib/libglib-2.0.so.0
#3  0x00007ffff2061ea8 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
#4  0x00007fffec2540e0 in fwupd_client_connect (self=0x55555602f1a0, cancellable=0x555556050820, error=0x7fffacffebc0) at ../libfwupd/fwupd-client-sync.c:102
#5  0x00007fffec2544bf in fwupd_client_get_devices (self=0x55555602f1a0, cancellable=0x555556050820, error=0x7fffacffebc0) at ../libfwupd/fwupd-client-sync.c:140
#6  0x00007fffec2ccb84 in FwupdBackend::checkForUpdates()::$_5::operator()() const (this=0x5555564188e8) at /home/apol/devel/frameworks/discover/libdiscover/backends/FwupdBackend/FwupdBackend.cpp:404
#7  0x00007fffec2ccae6 in QtConcurrent::StoredFunctorCall0<_GPtrArray*, FwupdBackend::checkForUpdates()::$_5>::runFunctor() (this=0x5555564188c0) at /home/apol/devel/kde5/include/QtConcurrent/qtconcurrentstoredfunctioncall.h:60
#8  0x00007fffec2d2542 in QtConcurrent::RunFunctionTask<_GPtrArray*>::run() (this=0x5555564188c0) at /home/apol/devel/kde5/include/QtConcurrent/qtconcurrentrunbase.h:108
#9  0x00007ffff5268032 in QThreadPoolThread::run() (this=0x555556425420) at /home/apol/devel/frameworks/qt5/qtbase/src/corelib/thread/qthreadpool.cpp:100
#10 0x00007ffff5264d41 in QThreadPrivate::start(void*) (arg=0x555556425420) at /home/apol/devel/frameworks/qt5/qtbase/src/corelib/thread/qthread_unix.cpp:329
#11 0x00007ffff49293e9 in start_thread () at /usr/lib/libpthread.so.0
#12 0x00007ffff4d5c293 in clone () at /usr/lib/libc.so.6
```

hughsie:
@aleixpol if you change `g_main_context_ref (g_main_context_default ())` to `g_main_context_ref_thread_default ()` does that work?

aleksander0m:
@hughsie this patch solves the issues I was having with `fwupdmgr`... but I can also see breaking @aleixpol 's tester from here: https://github.com/fwupd/fwupd/issues/2600#issuecomment-730021337

I changed it to g_main_context_ref_thread_default() and it's also stuck I think, @aleixpol should confirm if possible.

hughsie:
In the backtrace I see:

    #0  0x00007ffff72b980f in poll () at /lib64/libc.so.6
    #1  0x00007ffff777e576 in g_main_context_iterate.constprop () at /lib64/libglib-2.0.so.0
    #2  0x00007ffff7729bc3 in g_main_context_iteration () at /lib64/libglib-2.0.so.0
    #3  0x00007ffff7d23a2f in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib64/libQt5Core.so.5
    #4  0x00007ffff7cd764b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib64/libQt5Core.so.5
    #5  0x00007ffff7cdf010 in QCoreApplication::exec() () at /lib64/libQt5Core.so.5
    #6  0x000000000040378d in main ()

So QEventDispatcherGlib is manually iterating on g_main_context_iterate() -- is this the same one from g_main_context_default?

hughsie:
I don't think there's any way to fix this library side; something like this should work tho:

    fw->setFuture(QtConcurrent::run([&] {
        fwupd_client_set_main_context(client, g_main_context_get_thread_default());
        return fwupd_client_get_devices(client, cancellable, &error);
    }));

hughsie:
@aleixpol is that sane?

aleixpol:
> fwupd_client_set_main_context(client, g_main_context_get_thread_default());

If I do this, I get it locked at `
            g_autoptr(GPtrArray) releases = fwupd_client_get_releases(client, fwupd_device_get_id(device), m_cancellable, &error);`
This one runs on the main thread.

Also note that patching Discover means that it stops being backwards compatible (i.e. all deployments that get your latest release but not our latest release will remain broken).

hughsie:
>  This one runs on the main thread.

I don't think you want to run any DBus sync methods in the main thread; the UI could block for multiple seconds...

> patching Discover means that it stops being backwards compatible

I'm out of ideas, as I don't know how it ever worked between threads before. How does QEventDispatcherGlib interact with other libraries?

aleixpol:
I also have a patch that ports it to use async functions and works fine, the problem then is that I need to bump dependencies and distros won't allow it.

I guess I can try to issue a fix with something like `fwupd_client_set_main_context(client, g_main_context_get_thread_default());` and ifdef'd only for new versions and that does everything on the helper thread.

I don't know what the right fix is either.

superm1:
> I also have a patch that ports it to use async functions and works fine, the problem then is that I need to bump dependencies and distros won't allow it.

Which distros let you take a newer KDE discover _but not_ a newer fwupd?  This seems odd to me.

hughsie:
@aleixpol can you try this branch on unmodified discover please? I'm using a per-mainloop thread context by default to maintain the old behaviour, and allowing the library user to overwrite as required.

aleksander0m:
That last update makes my fwupdmgr test work (I can see progress/state updates) and it also doesn't block @aleixpol 's tester any more, I think it's looking good :D

aleixpol:
This definitely seems to improve the situation but now I have the problem that the flatpak backend never finishes. If I start Discover with just fwupd it works fine but as soon as we have both it gets all confused.

The backtrace looks like this, I could try to get debug symbols for flatpak if you think that would help but it feels like flatpak might be being left without an event loop to run on after this change?
```
#0  0x00007ffff4d5146f in poll () at /usr/lib/libc.so.6
#1  0x00007ffff20b67af in  () at /usr/lib/libglib-2.0.so.0
#2  0x00007ffff2061e43 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
#3  0x00007fff9a7445b3 in  () at /usr/lib/libflatpak.so.0
#4  0x00007fff9a745c8e in  () at /usr/lib/libflatpak.so.0
#5  0x00007fff9a70cc77 in  () at /usr/lib/libflatpak.so.0
#6  0x00007fff9a70d35d in  () at /usr/lib/libflatpak.so.0
#7  0x00007fff9a71c47c in flatpak_installation_fetch_remote_metadata_sync () at /usr/lib/libflatpak.so.0
#8  0x00007fff9a81d750 in FlatpakRunnables::fetchMetadata(_FlatpakInstallation*, FlatpakResource*) (installation=0x7fffac01c420, app=0x555556e5f320) at /home/apol/devel/frameworks/discover/libdiscover/backends/FlatpakBackend/FlatpakFetchDataJob.cpp:46
#9  0x00007fff9a816b47 in QtConcurrent::StoredFunctorCall2<QByteArray, QByteArray (*)(_FlatpakInstallation*, FlatpakResource*), _FlatpakInstallation*, FlatpakResource*>::runFunctor() (this=0x555556e5f1a0) at /home/apol/devel/kde5/include/QtConcurrent/qtconcurrentstoredfunctioncall.h:784
#10 0x00007fff9a816ad2 in QtConcurrent::RunFunctionTask<QByteArray>::run() (this=0x555556e5f1a0) at /home/apol/devel/kde5/include/QtConcurrent/qtconcurrentrunbase.h:108
#11 0x00007ffff5268032 in QThreadPoolThread::run() (this=0x555556602030) at /home/apol/devel/frameworks/qt5/qtbase/src/corelib/thread/qthreadpool.cpp:100
#12 0x00007ffff5264d41 in QThreadPrivate::start(void*) (arg=0x555556602030) at /home/apol/devel/frameworks/qt5/qtbase/src/corelib/thread/qthread_unix.cpp:329
#13 0x00007ffff49293e9 in start_thread () at /usr/lib/libpthread.so.0
#14 0x00007ffff4d5c293 in clone () at /usr/lib/libc.so.6
```

hughsie:
It looks like flatpak is using `g_main_context_ref_thread_default()` in the same way as this patch does, for instance in `flatpak_load_http_uri_once()` -- I don't understand why running things in different threads would affect a different GMainContext. You're not running them both at the same time from the same thread are you? My gut instinct is we merge this and then try to figure out the flatpak interaction async.

superm1:
@AdamWill see https://github.com/fwupd/fwupd/pull/2694, there is some other work going on..

mcatanzaro:
I'm nervous about this change since it violates one of the [GMainContext best practices](https://developer.gnome.org/programming-guidelines/stable/main-contexts.html.en): "Never iterate a context created outside the library, including the global-default or thread-default contexts. Otherwise, GSources created in the application may be dispatched when the application is not expecting it, causing re-entrancy problems for the application code."

I wrote more about this [here](https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1113#note_999813).

superm1:
@mcatanzaro can you see if your concerns still exist after the next change that was linked?

mcatanzaro:
That looks like an improvement, but it does not address my feedback.

Let's distinguish between two cases:

 * Synchronous API methods that wrap asynchronous implementations should run on an internal GMainContext that is never exposed to the application.
 * Application callbacks should execute on the GMainContext that is thread default at either (a) the time the source object (FwupdClient) is created, or (b) the time the async method is called. I believe (b) is more typical (notable exception: GIO), but it seems fwupd is trying to implement (a), which is totally fine. Although I never considered this before, I suppose (b) does not work at all when you need to emit signals during the async method call, and it looks like fwupd does that? So (a) looks like a good choice.

Since fwupd is taking approach (a), you can use fwupd_client_[get,set]_main_context() to control which GMainContext is used to execute application callbacks. But I do *not* think it is OK to *iterate* this GMainContext to implement synchronous methods. That should always happen on a different, internal-only GMainContext, not the FwupdClient's public GMainContext. That's why I don't like this MR, which allows applications to set the GMainContext used to implement sync methods.

Now, if my advice is followed and you create an internal GMainContext to implement sync methods, then you must be careful not to accidentally execute application callbacks (or emit signals) on the internal GMainContext; otherwise, the callback or signal will execute on an unexpected thread, which is typically a disaster. Instead, use g_main_context_invoke() to run code on the FwupdClient's public GMainContext, then execute the callback there.

P.S. This is hard and confused the heck out of me for longer than I care to admit. But I think the rules work very well in practice.

mcatanzaro:
> Although I never considered this before, I suppose (b) does not work at all when you need to emit signals during the async method call, and it looks like fwupd does that?

Ah, (b) does not work for *threadsafe objects* that need to emit signals during the async method call. Strategy (b) should work totally fine for most objects.

Anyway, my bottom line is that if you follow the rules in [Philip's guide](https://developer.gnome.org/programming-guidelines/stable/main-contexts.html.en), you won't have problems with callbacks or signals executing on unexpected times or on unexpected threads. When you do have a problem (you will, because programming is hard), it will be because one of the rules was not consistently followed.

mcatanzaro:
> Now, if my advice is followed and you create an internal GMainContext to implement sync methods, then you must be careful not to accidentally execute application callbacks (or emit signals) on the internal GMainContext; otherwise, the callback or signal will execute on an unexpected thread, which is typically a disaster. Instead, use g_main_context_invoke() to run code on the FwupdClient's public GMainContext, then execute the callback there.

Hm, I suppose that doesn't work if you need to emit a signal *before* your sync method call returns (because you will never iterate the public GMainContext yourself, and thus the signal would not be emitted until control returns to the application). In that case, you can use g_main_context_pop_thread_default() to temporarily unset the internal GMainContext to ensure the application doesn't wind up using it by mistake, then emit the signal, then use g_main_context_push_thread_default() to set your internal GMainContext again if you need to continue iterating it.

mcatanzaro:
My concerns were addressed in https://github.com/fwupd/fwupd/pull/2741.

PR #2640, opened 2020-12-03
Add a CI tester for running libfwupd with QT threads

superm1:
This only runs on Arch to avoid giving the extra dependencies to all
the distro CI builds.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/qt5-thread-test/meson.build
+++ b/contrib/ci/qt5-thread-test/meson.build
@@ -1,0 +1,23 @@
+project('qt5-thread-test', 'cpp',
+  license : 'LGPL-2.1+',
+)
+add_project_arguments('-fPIC', language : 'cpp')
+qt5core = dependency('Qt5Core')
+qt5concurrent = dependency('Qt5Concurrent')
+glib2 = dependency('glib-2.0')
+gio2 = dependency('gio-2.0')
+fwupd = dependency('fwupd')
+e = executable(
+  'qt-thread-test',
+  sources : [
+    'qt-thread-test.cpp'
+  ],
+  dependencies : [
+    qt5core,
+    qt5concurrent,
+    glib2,
+    gio2,
+    fwupd,
+  ],
+)
+test('qt-thread-test', e, timeout: 60)
--- a/contrib/ci/qt5-thread-test/qt-thread-test.cpp
+++ b/contrib/ci/qt5-thread-test/qt-thread-test.cpp
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2020 Aleix Pol <aleixpol@kde.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+extern "C" {
+#include <fwupd.h>
+}
+
+#include <QCoreApplication>
+#include <QtConcurrentRun>
+#include <QFutureWatcher>
+
+int main(int argc, char** argv)
+{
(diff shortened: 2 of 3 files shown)


superm1:
@aleixpol your tester didn't have a license.  Are you OK with it being licensed LGPL2 and included?

meson_options.txt
```diff
@@ -28,6 +28,7 @@ option('systemd', type : 'boolean', value : true, description : 'enable systemd
 option('systemd_root_prefix', type: 'string', value: '', description: 'Directory to base systemd’s installation directories on')
 option('elogind', type : 'boolean', value : false, description : 'enable elogind support')
 option('tests', type : 'boolean', value : true, description : 'enable tests')
+option('qt_tests', type: 'boolean', value: false, description: 'enable QT tests')
```
aleixpol, line 31: Qt is spelled with lower case t.

aleixpol (commented):
I'm happy to have my code included in LGPL2, thanks!

meson_options.txt
superm1: Thanks will fix it up.  I also need to relocated the code in the build system so that we don't have to add C++ compiler everywhere.

superm1:
I would think it makes sense to rebase https://github.com/fwupd/fwupd/pull/2638 on this, and whatever the outcome that is done there lands in the tester cpp file.

hughsie:
The thing is; I don't know how QEventDispatcherGlib actually works. I don't know if the QT5 test case code is valid or not...

contrib/ci/qt5-thread-test/meson.build
```diff
@@ -0,0 +1,23 @@
+project('qt5-thread-test', 'cpp',
+  license : 'LGPL-2.1+',
+)
+add_project_arguments('-fPIC', language : 'cpp')
+qt5core = dependency('Qt5Core')
+qt5concurrent = dependency('Qt5Concurrent')
+glib2 = dependency('glib-2.0')
+gio2 = dependency('gio-2.0')
+fwupd = dependency('fwupd')
```
hughsie, line 9: just to clarify; this is the installed fwupd version rather than the "just built" version, right?

superm1: Yeah it's explicitly run after install step.

contrib/ci/qt5-thread-test/qt-thread-test.cpp
```diff
@@ -0,0 +1,32 @@
...
+    auto cancellable = g_cancellable_new();
+    g_autoptr(GError) error = nullptr;
+
+    auto fw = new QFutureWatcher<GPtrArray*>(&app);
+    QObject::connect(fw, &QFutureWatcher<GPtrArray*>::finished, [fw]() {
+        QCoreApplication::exit(0);
+    });
+    fw->setFuture(QtConcurrent::run([&] {
+        return fwupd_client_get_devices(client, cancellable, &error);
+    }));
```
hughsie, line 29: if this hangs, how do we signify a CI failure? Is github actions going to kill our build after say 30 minutes? I'd much rather have a timer somewhere.

superm1: There is a timeout set on the test in meson, so a hang should be killed by the test runner I would expect

hughsie (approved):

PR #2653, opened 2020-12-08
Wip/hughsie/meson

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
I guess we'll probably want to double check gnome-firmware still subprojects properly with this change, but I think it's fine.

PR #2660, opened 2020-12-08
trivial: Add fu_common_bytes_new_offset()

hughsie:
This is a safer version of g_bytes_new_from_bytes() which returns a GError
if the offsets are invalid rather than emitting a critical warning.

This prevents a critical warning and potential crash when parsing invalid
bcm57xx firmware.

Type of pull request:
- [X] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi/fu-uefi-device.c
+++ b/plugins/uefi/fu-uefi-device.c
@@ -356,10 +356,10 @@ fu_uefi_device_fixup_firmware (FuDevice *device, GBytes *fw, GError **error)
 	/* ESRT header matches payload */
 	if (g_strcmp0 (fu_uefi_device_get_guid (self), guid_new) == 0) {
 		g_debug ("ESRT matches payload GUID");
-		return g_bytes_new_from_bytes (fw, 0, fw_length);
+		return g_bytes_ref (fw);
 	/* Type that doesn't require a header */
 	} else if (!self->requires_header) {
-		return g_bytes_new_from_bytes (fw, 0, fw_length);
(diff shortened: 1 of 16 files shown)


plugins/uefi/fu-uefi-device.c
```diff
@@ -356,10 +356,10 @@ fu_uefi_device_fixup_firmware (FuDevice *device, GBytes *fw, GError **error)
 	/* ESRT header matches payload */
 	if (g_strcmp0 (fu_uefi_device_get_guid (self), guid_new) == 0) {
 		g_debug ("ESRT matches payload GUID");
-		return g_bytes_new_from_bytes (fw, 0, fw_length);
+		return g_bytes_ref (fw);
 	/* Type that doesn't require a header */
 	} else if (!self->requires_header) {
-		return g_bytes_new_from_bytes (fw, 0, fw_length);
+		return g_bytes_ref (fw);
```
superm1, line 362: shouldn't this be a separate commit?

superm1 (approved):

plugins/uefi/fu-uefi-device.c
hughsie: I don't think so; I didn't convert g_bytes_new_from_bytes() here as it wasn't actually required.

superm1: Right, my point was this change could have been done separately and didn't need to come in the exact same commit.  Doesn't really matter for anything but backports and bisects.

PR #2662, opened 2020-12-09
trivial: Allow turning off the FMAP binary search when fuzzing

hughsie:
This speeds up the fuzzing task from 22ms to 33us.

Type of pull request:
- [X] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@bleungatchromium hey -- quick ask for related reasons -- do you have a *tiny* licence-free (e.g. LGPLv2+) FMAP file with zero contents? e.g. a FMAP file with just the `__FMAP__` header with one region with contents `foo` or something please?

Same question for `FuCrosEcFirmwareSection` -- the smallest file possible please.

superm1 (commented):
Although it speeds up fuzzing doesn't it defeat some of the point of fuzzing that you're turning off part of the codebase specifically for the fuzzing tests?

hughsie:
> Although it speeds up fuzzing doesn't it defeat some of the point of fuzzing

Yes, although I'm more worried about the "_read blob of data at offset to tell us the offset of a blob of data_" than "_memcmp this area_".

hughsie:
> do you have a tiny licence-free (e.g. LGPLv2+) FMAP file with zero contents

@bleungatchromium ignore that request; I just taught FuFmapFirmware how to write binary blobs too.

PR #2665, opened 2020-12-09
thunderbolt: Add maple ridge firmware parsing support

superm1:
CC @westeri, @YehezkelShB

This is based upon https://git.kernel.org/pub/scm/linux/kernel/git/westeri/thunderbolt.git/commit/?h=next&id=db0746e3399ee87ee5f957880811da16faa89fb8 but I found that parsing an image I have locally it's missing an ID 
0x1136.  So I think that's got to be added to kernel still too, no?

If you can please confirm the values I selected for these IDs.  Here's how a local binary parses.
```
FuThunderboltFirmwareUpdate:
Family:                 Maple Ridge
IsHost:                 true
IsNative:               true
DeviceId:               0x1136
VendorId:               0xd4
ModelId:                0xa58
FlashSize:              0x0
Generation:             0x4
Ports:                  0x2
HasPd:                  true
Section0:               0x4000
Section1:               0x4210
Section2:               0x4610
Section3:               0x22958
  FuFirmwareImage:
  Data:                 0x67000
```

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/thunderbolt/fu-thunderbolt-firmware.c
+++ b/plugins/thunderbolt/fu-thunderbolt-firmware.c
@@ -120,6 +120,8 @@ fu_thunderbolt_firmware_family_to_string (FuThunderboltFamily family)
 		return "Titan Ridge";
 	if (family == _FAMILY_BB)
 		return "BB";
+	if (family == _FAMILY_MR)
+		return "Maple Ridge";
 	return "Unknown";
 }
 
@@ -382,6 +384,13 @@ fu_thunderbolt_firmware_parse (FuFirmware *firmware,
 		{ 0x15EA, 3, _FAMILY_TR, 2 }, /* TR 4C */
 		{ 0x15EF, 3, _FAMILY_TR, 2 }, /* TR 4C device */
 		{ 0x15EE, 3, _FAMILY_BB, 0 }, /* BB device */
+		/* Maple ridge devices
+		 * NOTE: These are expected to be flashed via UEFI capsules *not* Thunderbolt plugin
+		 * Flashing via fwupd will require matching kernel work.
+		 * They're left here only for parsing the binaries
+		 */
+		{ 0x1136, 4, _FAMILY_MR, 2 },
+		{ 0x1137, 4, _FAMILY_MR, 2 },
 		{ 0 }
 	};
 
@@ -442,7 +451,7 @@ fu_thunderbolt_firmware_parse (FuFirmware *firmware,
 		g_set_error (error,
 			     FWUPD_ERROR,
 			     FWUPD_ERROR_NOT_SUPPORTED,
-			     "Unknown controller");
+			     "Unknown controller: %x", priv->device_id);
 		return FALSE;
 	}
 
--- a/plugins/thunderbolt/fu-thunderbolt-firmware.h
+++ b/plugins/thunderbolt/fu-thunderbolt-firmware.h
@@ -28,6 +28,7 @@ typedef enum {
 	_FAMILY_AR_C,
 	_FAMILY_TR,
 	_FAMILY_BB,
+	_FAMILY_MR,
 } FuThunderboltFamily;
 
 struct _FuThunderboltFirmwareClass


plugins/thunderbolt/fu-thunderbolt-firmware.c
```diff
@@ -382,6 +384,8 @@ fu_thunderbolt_firmware_parse (FuFirmware *firmware,
 		{ 0x15EA, 3, _FAMILY_TR, 2 }, /* TR 4C */
 		{ 0x15EF, 3, _FAMILY_TR, 2 }, /* TR 4C device */
 		{ 0x15EE, 3, _FAMILY_BB, 0 }, /* BB device */
+		{ 0x1136, 4, _FAMILY_MR, 2 }, /* MR DP */
+		{ 0x1137, 4, _FAMILY_MR, 2 }, /* MR 4C device */
```
hughsie, line 388: no idea on this bit unfortunately, but the rest looks sane.

hughsie:
I think it would be good to get this into the next RHEL release too...

plugins/thunderbolt/fu-thunderbolt-firmware.c
superm1: Yeah I think @westeri will need to confirm these and the descriptions for them.

superm1:
> I think it would be good to get this into the next RHEL release too...

With all the stuff the last 2 days you submitted (presumably caught by coverity) I sense a 1.5.4 in our near future to target to next RHEL release.

hughsie:
> With all the stuff the last 2 days you submitted

Ohh you noticed all that :)

> (presumably caught by coverity) 

Actually no, I've been using a new fuzzer called honggfuzz which I've wired up a test harness to test all the different firmware parser code. All the ->build vfuncs were so I can generate some high quality initial corpus data. Coverity is my plan for tomorrow, although it's not been super long since I ran that on master. I'll push the honggfuzz branch soon although it'll need fixups for some architectures I guess.

> I sense a 1.5.4 in our near future to target to next RHEL release.

Yes; keep the fixes coming in but sit on new features for a few days :)

Richard.

plugins/thunderbolt/fu-thunderbolt-firmware.c
westeri: Hi, actually in Windows they use UEFI Capsule update for MR so this is what we should do in Linux too . That's the reason the kernel side does not allow NVM update through the driver. I should have communicated this with you guys earlier, sorry about that.

superm1: @westeri oh right.  I knew that in the back of my mind.
Also, what about the missing 0x1136 in the kernel for that card though?  Should there be something similar to https://git.kernel.org/pub/scm/linux/kernel/git/westeri/thunderbolt.git/commit/?h=next&id=db0746e3399ee87ee5f957880811da16faa89fb8 for that ID?


@hughsie up to you, do we leave the IDs so that the parser can still understand them in case a non-UEFI MR case comes up later?  It would still require kernel code changes though to allow flashing.

hughsie: > so that the parser can still understand them in case a non-UEFI MR case

I think that's sensible, yes.

superm1: > I think that's sensible, yes.

OK let me change the PR to add a comment to the code for future selfs to know why we left it then.

hughsie (approved):

plugins/thunderbolt/fu-thunderbolt-firmware.c
westeri: Hi, 0x1136 is the PCIe downstream port of MR. I don't think it needs to be added to the TBT driver.

superm1: OK thanks!

PR #2666, opened 2020-12-09
Use honggfuzz to fuzz firmware rather than AFL

hughsie:
This has better multi-core performance and can run in persistent mode -- which
allows us to construct a test harness of all the parsers (which takes time) and
then just reuse the process for lots of different data.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


src/fu-engine.h
```diff
@@ -26,13 +26,18 @@ G_DECLARE_FINAL_TYPE (FuEngine, fu_engine, FU, ENGINE, GObject)
  * FuEngineLoadFlags:
  * @FU_ENGINE_LOAD_FLAG_NONE:		No flags set
  * @FU_ENGINE_LOAD_FLAG_READONLY_FS:	Ignore readonly filesystem errors
+ * @FU_ENGINE_LOAD_FLAG_NO_ENUMERATE:	Do not enumerate devices
+ * @FU_ENGINE_LOAD_FLAG_NO_REMOTES:	Do not enumerate remotes
+ * @FU_ENGINE_LOAD_FLAG_NO_HWINFO:	Do not load details about the hardware
```
superm1, line 31: These feel like they should also be fwupdtool arguments.

hughsie: I fixed this in a different way in https://github.com/fwupd/fwupd/pull/2668

hughsie:
@superm1 I'm happy with this now, re-review welcome, thanks.

meson.build
```diff
@@ -113,6 +112,14 @@ warning_flags = [
 cc = meson.get_compiler('c')
 add_project_arguments(cc.get_supported_arguments(warning_flags), language : 'c')
 
+# use honggfuzz for parallel persistent fuzzing
+if cc.has_function('HF_ITER')
+  honggfuzz = find_program('honggfuzz')
+  conf.set('HAVE_HF_ITER', '1')
+else
+  honggfuzz = find_program('honggfuzz', required : false)
```
superm1, line 120: I don't know if this works, but can you have:
`honggfuzz = find_program('honggfuzz', required : cc.has_function('HF_ITER')`

hughsie: I still need the HAVE_HF_ITER define tho

superm1: Like this maybe: 
```
honggfuzz = find_program('honggfuzz', required : cc.has_function('HF_ITER')
if honggfuzz.found()
   conf.set('HAVE_HF_ITER', '1')
endif
```

hughsie: Works for me! Fixed.

meson.build
```diff
@@ -113,6 +112,12 @@ warning_flags = [
 cc = meson.get_compiler('c')
 add_project_arguments(cc.get_supported_arguments(warning_flags), language : 'c')
 
+# use honggfuzz for parallel persistent fuzzing
+honggfuzz = find_program('honggfuzz', required: cc.has_function('HF_ITER'))
+if honggfuzz.found()
+  conf.set('HAVE_HF_ITER', '1')
+endif
+
```
superm1, line 120: you know I just realized something about doing it this way though.  If `honggfuzz` is _installed_ that will turn on fuzzing everywhere.  is that expected?  Or does this need an extra meson option?

hughsie: >  is that expected 

Yes, you can use honggfuzz without compiling with the custom clang compiler, but you can't use the persistent mode.

superm1: OK.

superm1 (approved):

PR #2667, opened 2020-12-10
Wip/hughsie/dos

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-ihex-firmware.c
+++ b/libfwupdplugin/fu-ihex-firmware.c
@@ -69,11 +69,24 @@ fu_ihex_firmware_record_new (guint ln, const gchar *line,
 
 	/* check starting token */
 	if (line[0] != ':') {
-		g_set_error (error,
-			     FWUPD_ERROR,
-			     FWUPD_ERROR_INVALID_FILE,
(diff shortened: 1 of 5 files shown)


chlee75 (commented):
@hughsie 

Your change is good to me.

I think that it need to be validated on HP DMC dock before commit.
However I can't do it because I returned those docks to HP.
Have you ever validated it?
If not or can't, I can ask HP to validate it on real DMC dock.

And I will be out of duty on Dec 11, I will be back to work on Dec 14.

hughsie:
@chlee -- I only have the Adicora one here -- I don't think that's DMC, right?

chlee75:
@hughsie 
Adicora is DMC. You can validate your change on Adicora dock.

hughsie:
> You can validate your change on Adicora dock

Thanks! Sure enough I ran the G5 device tests and these continued to pass. Thanks.

chlee75 (approved):
Code change is fine and you have already validated it.
Approve it.

superm1 (approved):

PR #2668, opened 2020-12-10
Wip/hughsie/fwupdtool non root2

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -6289,6 +6289,48 @@ fu_engine_ensure_client_certificate (FuEngine *self)
 	g_debug ("client certificate exists and working");
 }
 
+static void
+fu_engine_add_firmware_gtypes (FuEngine *self)
+{
+	fu_engine_add_firmware_gtype (self, "raw", FU_TYPE_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "dfu", FU_TYPE_DFU_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "fmap", FU_TYPE_FMAP_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "ihex", FU_TYPE_IHEX_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "srec", FU_TYPE_SREC_FIRMWARE);
+}
+
+/**
+ * fu_engine_load_firmware:
+ * @self: A #FuEngine
+ * @error: A #GError, or %NULL
+ *
+ * Load the plugins so that the firmware parser ONLY is ready to use.
+ *
+ * Returns: %TRUE for success
+ **/
+gboolean
+fu_engine_load_firmware (FuEngine *self, GError **error)
+{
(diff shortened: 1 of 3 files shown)


src/fu-engine.c
```diff
@@ -5929,14 +5929,10 @@ fu_engine_load_plugins (FuEngine *self, GError **error)
...
 				  self);
-
-		/* if loaded from fu_engine_load() open the plugin */
-		if (self->usb_ctx != NULL) {
-			if (!fu_plugin_open (plugin, filename, &error_local)) {
-				g_warning ("cannot load: %s", error_local->message);
-				fu_engine_add_plugin (self, plugin);
-				continue;
-			}
+		if (!fu_plugin_open (plugin, filename, &error_local)) {
```
superm1, line 5932: the check for `NULL` was specifically to support the `get-plugins` command for fwupdtool.

superm1: Here's the commit it came from: https://github.com/fwupd/fwupd/commit/8c71a3f2e98262d672f90fe0bcc0841e550083cc

src/fu-engine.c
```diff
@@ -6289,6 +6289,48 @@ fu_engine_ensure_client_certificate (FuEngine *self)
...
 
+static void
+fu_engine_add_firmware_gtypes (FuEngine *self)
+{
+	fu_engine_add_firmware_gtype (self, "raw", FU_TYPE_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "dfu", FU_TYPE_DFU_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "fmap", FU_TYPE_FMAP_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "ihex", FU_TYPE_IHEX_FIRMWARE);
+	fu_engine_add_firmware_gtype (self, "srec", FU_TYPE_SREC_FIRMWARE);
+}
```
superm1, line 6300: a static function only ever called only once for 5 different void functions?  Seems like it should just be in the caller function.

superm1: oh never mind, It's two code paths now, I missed that. sorry for the noise.

src/fu-tool.c
```diff
@@ -1748,7 +1748,7 @@ fu_util_get_firmware_types (FuUtilPrivate *priv, gchar **values, GError **error)
 	g_autoptr(GPtrArray) firmware_types = NULL;
 
 	/* load engine */
-	if (!fu_engine_load (priv->engine, FU_ENGINE_LOAD_FLAG_NO_ENUMERATE, error))
+	if (!fu_engine_load_firmware (priv->engine, error))
```
superm1, line 1751: why not have a flag that `FU_ENGINE_LOAD_FIRMWARE_ONLY` you detect and keep the function the same?

hughsie: Alternate proposal: https://github.com/fwupd/fwupd/pull/2670

Issue #2650, opened 2020-12-08
X1 Extreme Gen 2 fingerprint firmware update is stuck


HadiModarres:
fwupd stuck on upgrading fingerprint firmware ("Prometheus") on Thinkpad X1 Extreme Gen 2: 

> ^CTerminated
(base) hadi@hadi-ThinkPad-X1-Extreme-2nd:~$ fwupdmgr update -vv
(fwupdmgr:8784): GLib-DEBUG: 01:33:16.619: setenv()/putenv() are not thread-safe and should not be used after threads are created
• Thunderbolt Controller has the latest available firmware version
(fwupdmgr:8784): FuMain-DEBUG: 01:33:16.628: current version is 59.00: 59.00=same
Upgrade available for Prometheus from 10.01.3121519 to 10.01.3273255
Prometheus and all connected devices may not be usable while updating. Continue with update? [Y|n]: Y
Downloading 10.01.3273255 for Prometheus...
(fwupdmgr:8784): FuCommon-DEBUG: 01:33:21.221: creating path /home/hadi/.cache/fwupd
(fwupdmgr:8784): FuMain-DEBUG: 01:33:21.225: skpping download as file already exists
(fwupdmgr:8784): Fwupd-DEBUG: 01:33:21.226: Emitting ::status-changed() [decompressing]
Decompressing…           [-                                      ](fwupdmgr:8784): Fwupd-DEBUG: 01:33:21.228: Emitting ::status-changed() [idle]
Decompressing…           [***************************************]
(fwupdmgr:8784): Fwupd-DEBUG: 01:33:21.251: Emitting ::status-changed() [waiting-for-auth]
Authenticating…          [ -                                     ](fwupdmgr:8784): Fwupd-DEBUG: 01:33:21.253: Emitting ::status-changed() [idle]
Authenticating…          [***************************************]

The upgrade is stuck on Authenticating with no further info or messages as to why it's failing to go forward. Any help is appreciated

hughsie:
What distro? You ignored most of the details we need...

HadiModarres:
@hughsie oh sorry about that, it's Ubuntu 20.04

fwupdmgr version: 
`client version:	1.3.11
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.3.11
`

devices: 

> 20QVCTO1WW
│
├─Thunderbolt Controller:
│     Device ID:           607d06b8b13f22018453d6157e624da99ebb1340
[...]

PR #2669, opened 2020-12-10
synaptics-prometheus: Fix flashing a fingerprint reader that is in use

hughsie:
The fprint daemon only keeps the device open for 5 seconds and then releases it,
which seems like a small window to hit.

But! We're asking the user to authenticate with the same device we're about to
upgrade so a different part of the stack woke up the hardware just before we're
about to deploy an update onto it.

Just wait a couple of seconds to make sure the device is idle.

Fixes https://github.com/fwupd/fwupd/issues/2650

### DIFF
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -203,6 +203,8 @@ fwupd_device_flag_to_string (FwupdDeviceFlags device_flag)
 		return "has-multiple-branches";
 	if (device_flag == FWUPD_DEVICE_FLAG_BACKUP_BEFORE_INSTALL)
 		return "backup-before-install";
+	if (device_flag == FWUPD_DEVICE_FLAG_RETRY_OPEN)
+		return "retry-open";
 	if (device_flag == FWUPD_DEVICE_FLAG_UNKNOWN)
 		return "unknown";
 	return NULL;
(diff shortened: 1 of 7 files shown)


hughsie:
@HadiModarres this works for you?

superm1:
Do we have a similar problem at probe potentially too?  (I wonder if some sleep and retry 1-2 times in code is the better solution)

hughsie:
> Do we have a similar problem at probe potentially too

Not `probe`, but possibly `setup`. Do you think the error is coming from `g_usb_device_open()`? In which case should we add `G_USB_DEVICE_ERROR_ALREADY_OPEN` to the list fu_device_retry_add_recovery uses and then retry the open? Or we could handle this lower down the stack and try `FuDevice->open` perhaps.

superm1:
> Do you think the error is coming from g_usb_device_open()

That was my worry.  I don't think we have enough evidence of the call stack.  Maybe can this be repro'ed under a debugger with a breakpoint to see?

hughsie:
> Or we could handle this lower down the stack and try FuDevice->open perhaps

I tried a few things, and I think this is the most useful thing to provide.

superm1 (approved):
don't forget to add the new flag into gnome-firmware.

hughsie:
> don't forget to add the new flag into gnome-firmware.

I think this should probably depend on https://github.com/fwupd/fwupd/pull/2719 -- it's not useful to the end user in any way...

PR #2670, opened 2020-12-10
Allow using fwupdtool as non-root for firmware commands

hughsie:
Just avoid loading SMBIOS and re-generating remotes when all we want to use is
the firmware loader. It also speeds up the self tests somewhat too.


superm1 (approved):
Yes, like this much better, thanks!

PR #2673, opened 2020-12-11
dfu: Show a warning if we fail to release the interface

hughsie:
I'm not returning FALSE here as it might break plugins.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
honggfuzz 17 : coverity 1

superm1 (approved):

Issue #2676, opened 2020-12-14
fwupd 1.5.2 failed to detect UEFI ESP partition in some machines


lcp:
I got a report for openSUSE Tumbleweed that the user got the following warning after upgrading fwupd to 1.5.2:

WARNING: UEFI ESP partition not detected or configured

It turned out that fwupd thought the system disk is a removable disk. Here is the output of "udisksctl info --block-device /dev/sdb"

    /org/freedesktop/UDisks2/block_devices/sdb:
      org.freedesktop.UDisks2.Block:
        Configuration:              []
        CryptoBackingDevice:        '/'
        Device:                     /dev/sdb
        DeviceNumber:               2064
        Drive:                      '/org/freedesktop/UDisks2/drives/TEAM_L5_3D_240GB_AB20170727A0104534'
        HintAuto:                   true
        HintIconName:               
        HintIgnore:                 false
        HintName:                   
        HintPartitionable:          true
        HintSymbolicIconName:       
        HintSystem:                 false
        Id:                         
        IdLabel:                    
        IdType:                     
        IdUUID:                     
        IdUsage:                    
        IdVersion:                  
        MDRaid:                     '/'
        MDRaidMember:               '/'
        PreferredDevice:            /dev/sdb
        ReadOnly:                   false
        Size:                       240057409536
        Symlinks:                   /dev/disk/by-id/ata-TEAM_L5_3D_240GB_AB20170727A0104534
                                    /dev/disk/by-id/scsi-0ATA_TEAM_L5_3D_240GB_AB20170727A0104534
                                    /dev/disk/by-id/scsi-1ATA_TEAM_L5_3D_240GB_AB20170727A0104534
                                    /dev/disk/by-id/scsi-SATA_TEAM_L5_3D_240GBAB20170727A0104534
                                    /dev/disk/by-id/scsi-SATA_TEAM_L5_3D_240GB_AB20170727A0104534
                                    /dev/disk/by-path/pci-0000:06:00.0-ata-2
                                    /dev/disk/by-path/pci-0000:06:00.0-ata-2.0
                                    /dev/disk/by-path/pci-0000:06:00.0-scsi-5:0:0:0
        UserspaceMountOptions:      
      org.freedesktop.UDisks2.PartitionTable:
        Partitions:         /org/freedesktop/UDisks2/block_devices/sdb1
                            /org/freedesktop/UDisks2/block_devices/sdb3
                            /org/freedesktop/UDisks2/block_devices/sdb2
[...]

PR #2677, opened 2020-12-14
Do not use the Block.HintSystem boolean for ESP filtering

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2676

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -2437,6 +2437,7 @@ FuVolume *
 fu_common_get_esp_default (GError **error)
 {
 	const gchar *path_tmp;
+	gboolean has_internal = FALSE;
 	g_autoptr(GPtrArray) volumes_fstab = g_ptr_array_new ();
 	g_autoptr(GPtrArray) volumes_mtab = g_ptr_array_new ();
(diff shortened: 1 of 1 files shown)


superm1:
I don't think we should flat out remove it, but instead use it for preferential selection. As in if there is none it's okay, but if there are 3 pick the one with the flag.

hughsie:
> I don't think we should flat out remove it, but instead use it for preferential selection

Good idea; something like that^^?

libfwupdplugin/fu-common.c
```diff
@@ -2457,13 +2458,26 @@ fu_common_get_esp_default (GError **error)
...
+		FuVolume *vol = g_ptr_array_index (volumes, i);
+		g_autofree gchar *type = fu_volume_get_id_type (vol);
+		if (g_strcmp0 (type, "vfat") == 0 &&
+		    fu_volume_is_internal (vol)) {
+			has_internal = TRUE;
+			break;
+		}
+	}
+
 	/* only add in internal vfat partitions */
```
superm1, line 2474: I think this comment probably needs to be changed

superm1 (approved):
code changes look good, just update a comment

PR #2681, opened 2020-12-14
Lower libcurl requirement further still

hughsie:
This allows us to build on Ubuntu 16.04!

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -65,7 +65,9 @@ typedef struct {
 
 typedef struct {
 	CURL				*curl;
+#ifdef HAVE_LIBCURL_7_56_0
 	curl_mime			*mime;
+#endif
 	struct curl_slist		*headers;
 } FwupdCurlHelper;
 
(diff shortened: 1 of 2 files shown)


superm1:
> This allows us to build on Ubuntu 16.04!

Ha, I would expect we're going to have a few other deps that don't work.... but it's in that direction.

superm1 (approved):

hughsie:
> Ha, I would expect we're going to have a few other deps that don't work

I'm only compiling standalone fwupdtool enough for fuzzing, so maybe that's enough?

superm1:
> I'm only compiling standalone fwupdtool enough for fuzzing, so maybe that's enough?

You had a new enough gnutls and meson to do that though?

hughsie:
> You had a new enough gnutls 

I compiled jcat as a subproject with neither gpg nor pkcs7... :/ Not usable obviously, but good enough for a fuzzer....

> and meson to do that though?

`pip install meson` :/

PR #2682, opened 2020-12-14
trivial: Require <string.h> for strlen() and memcpy()

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
Any idea why this is needed now and wasn't previously?  A more stringent compiler?

hughsie:
> A more stringent compiler?

Exactly that; super new clang.

PR #2683, opened 2020-12-14
Wip/hughsie/oss fuzz

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -38,3 +38,4 @@ option('efi-libdir', type : 'string', description : 'path to the EFI lib directo
 option('efi-ldsdir', type : 'string', description : 'path to the EFI lds directory')
 option('efi-includedir', type : 'string', value : '/usr/include/efi', description : 'path to the EFI header directory')
 option('efi_os_dir', type: 'string', description : 'the name of OS directory in ESP')
+option('link_language', type : 'string', value : 'c', description : 'the linker to use for targets [c, cpp]')
--- a/libfwupd/fwupd-fuzz.cpp
+++ b/libfwupd/fwupd-fuzz.cpp
@@ -1,0 +1,7 @@
+/*
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+int unused_var = 42;
--- a/libfwupd/meson.build
+++ b/libfwupd/meson.build
@@ -34,22 +34,29 @@ libfwupd_deps = [
     libjsonglib,
 ]
 
+libfwupd_src = [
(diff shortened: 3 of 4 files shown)


meson_options.txt
```diff
@@ -38,3 +38,4 @@ option('efi-libdir', type : 'string', description : 'path to the EFI lib directo
 option('efi-ldsdir', type : 'string', description : 'path to the EFI lds directory')
 option('efi-includedir', type : 'string', value : '/usr/include/efi', description : 'path to the EFI header directory')
 option('efi_os_dir', type: 'string', description : 'the name of OS directory in ESP')
+option('link_language', type : 'string', value : 'c', description : 'the linker to use for targets [c, cpp]')
```
superm1, line 41: rather than a new option to force this, can you detect the compiler or environment in meson.build?

hughsie: > can you detect the compiler or environment in meson.build

I tried compiler, but just got `clang`. I think env vars are frowned upon for building, e.g. https://github.com/mesonbuild/meson/issues/9

superm1: Hmm OK.  Well it would be better not to have an option to activate a hack.  How about if you just use a cross compilation script and set the linker?  See https://github.com/mesonbuild/meson/pull/6207

superm1: Actually you might not even need a cross file looking at that, it seems like if you set `LD` in environment it should respect it in meson 0.53.0+

hughsie: Hmm, I tried `export LD=$CXX` but it seemed to not use the linker. I'll try a cross.ini and see if that can work.

hughsie: Hmm, no dice:

    [binaries]
    c_ld = 'clang++'

gave me

    stderr: clang-12: error: invalid linker name in argument '-fuse-ld=clang++'

superm1: Isn't that meaning it's a clang bug not a meson bug though?  Or is it maybe not in PATH and you need to give a full path?

hughsie: From the meson maintainer `Yep, wont work. It just passes -fuse-ld=yourthing to the linker.`

superm1 (approved):
begrudgingly I guess this works.  It would be nice for requesting meson to have a cleaner way to accomplish it for the future though.

hughsie:
>  It would be nice for requesting meson to have a cleaner way to accomplish it for the future though

Discussing with the meson maintainer now.

hughsie:
> begrudgingly I guess this works

Lets go with the kludge now, and I'll revert if the meson team find a better way. It seems `export LD=$CXX` works for cmake like we think it should, just not meson. :(

PR #2684, opened 2020-12-14
Correct overloaded use of --no-metadata-check

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -61,6 +61,7 @@ struct FuUtilPrivate {
 	FwupdInstallFlags	 flags;
 	FwupdClient		*client;
 	FuProgressbar		*progressbar;
+	gboolean		 no_remote_check;
 	gboolean		 no_metadata_check;
 	gboolean		 no_reboot_check;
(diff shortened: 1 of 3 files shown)


src/fu-util.c
```diff
@@ -2740,6 +2741,9 @@ main (int argc, char *argv[])
 		{ "no-metadata-check", '\0', 0, G_OPTION_ARG_NONE, &priv->no_metadata_check,
 			/* TRANSLATORS: command line option */
 			_("Do not check for old metadata"), NULL },
+		{ "no-remote-check", '\0', 0, G_OPTION_ARG_NONE, &priv->no_remote_check,
+			/* TRANSLATORS: command line option */
+			_("Do not check for download remotes to be enabled"), NULL },
```
hughsie, line 2746: Maybe `Do not check if download remotes should be enabled`?

superm1: yep that is better, adjusted

hughsie (approved):

Issue #2690, opened 2020-12-16
libfwupdplugin for 1.5.x hardcodes /lib64 path


Aerdan:
**Describe the bug**
`libfwupdplugin` expects `/lib64` to be unconditionally present, which in turn causes an assertion failure in its testsuite on 64-bit systems where `/lib64` does not exist.

**Steps to Reproduce**
1. Install Adélie Linux 1.0rc2.
2. Add [APK Fission](https://apkfission.net/install.html) with `adelie/current` distribution.
3. Install dependencies for `fwupd` package
4. Build fwupd
5. Run testsuite

**Expected behavior**
libfwupdplugin testsuite should pass

**fwupd version information**
1.5.3, building preparatory to packaging.

**Additional questions**
- [Adélie Linux 1.0rc](https://www.adelielinux.org)
- Problem did not exist in 1.4.2.

superm1:
Would you mind including a log to show exactly which test is failing?

Aerdan:
```
ERROR:../libfwupdplugin/fu-self-test.c:563:fu_common_firmware_builder_func: assertion failed (error == NULL): failed to build firmware: bwrap: Can't find source path /lib64: No such file or directory
```
[Full output](https://tpaste.us/BEWR)

superm1:
It looks like the error stems from this commit: b821557b35495e87b35d92e9f7948f399cf380d4

sroracle:
It should use `--ro-bind-try` instead of `--ro-bind`.

PR #2691, opened 2020-12-16
trivial: libfwupdplugin: don't fail to set up bwrap if /lib64 doesn't…

superm1:
… exist

Fixes: #2690

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -516,7 +516,7 @@ fu_common_firmware_builder (GBytes *bytes,
 	fu_common_add_argv (argv, "--die-with-parent");
 	fu_common_add_argv (argv, "--ro-bind /usr /usr");
 	fu_common_add_argv (argv, "--ro-bind /lib /lib");
-	fu_common_add_argv (argv, "--ro-bind /lib64 /lib64");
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -516,7 +516,8 @@ fu_common_firmware_builder (GBytes *bytes,
 	fu_common_add_argv (argv, "--die-with-parent");
 	fu_common_add_argv (argv, "--ro-bind /usr /usr");
 	fu_common_add_argv (argv, "--ro-bind /lib /lib");
-	fu_common_add_argv (argv, "--ro-bind /lib64 /lib64");
+	if (g_file_test ("/lib64", G_FILE_TEST_EXISTS))
+		fu_common_add_argv (argv, "--ro-bind /lib64 /lib64");
```
sroracle, line 520: From `bwrap(1)`:

```
       --ro-bind-try SRC DEST
           Equal to --ro-bind but ignores non-existent SRC
```

```suggestion
	fu_common_add_argv (argv, "--ro-bind-try /lib64 /lib64");
```

superm1: thanks for your suggestion

PR #2693, opened 2020-12-17
Added LabTop Mk IV to Flashrom quirk

Sean-StarLabs:

### DIFF
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -10,4 +10,4 @@ VersionFormat=quad
 # Star LabTop Mk IV
 [HwId=baf1d04e-fd16-5e6a-93cc-1c23d171f879]
 DeviceID=LabTop
-VersionFormat=triplet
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Can you please rebase on master so we don't end up with a merge commit?

plugins/flashrom/flashrom.quirk
```diff
@@ -6,3 +6,8 @@ DeviceId=librem15v3
 [HwId=52b68c34-6b31-5ecc-8a5c-de37e666ccd5]
 DeviceId=LT1000
 VersionFormat=quad
+
+# Star LabTop Mk IV
+[HwId=baf1d04e-fd16-5e6a-93cc-1c23d171f879]
+DeviceID=LabTop
+VersionFormat=triplet
```
superm1, line 13: is this really a triplet?  I'm just confirming since the other systems in this file don't use a triplet.

superm1:
Can you please squash?

PR #2694, opened 2020-12-17
Always return DBus signals in the GMainContext that created the FwupdClient

hughsie:
Without this patch the GDBusProxy will forever emit signals in the thread
default main context of the thread which called fwupd_client_connect_async()
which may be different to the thread context of the g_dbus_proxy_call().

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -47,8 +47,7 @@ typedef GObject		*(*FwupdClientObjectNewFunc)	(void);
 static void fwupd_client_finalize	 (GObject *object);
 
 typedef struct {
-	GMainContext			*main_ctx_sync;
-	GMainContext			*main_ctx_prxy;
+	GMainContext			*main_ctx;
 	FwupdStatus			 status;
 	gboolean			 tainted;
 	gboolean			 interactive;
@@ -283,7 +282,7 @@ fwupd_client_properties_changed_cb (GDBusProxy *proxy,
 	if (priv->idle_id != 0)
 		g_source_remove (priv->idle_id);
 	priv->idle_id = g_source_attach (g_steal_pointer (&source),
-					 priv->main_ctx_prxy);
+					 priv->main_ctx);
 }
 
 typedef struct {
@@ -357,7 +356,7 @@ fwupd_client_signal_cb (GDBusProxy *proxy,
 	if (priv->idle_id != 0)
 		g_source_remove (priv->idle_id);
 	priv->idle_id = g_source_attach (g_steal_pointer (&source),
-					 priv->main_ctx_prxy);
+					 priv->main_ctx);
 }
 
 /**
@@ -375,8 +374,8 @@ GMainContext *
 fwupd_client_get_main_context (FwupdClient *self)
 {
 	FwupdClientPrivate *priv = GET_PRIVATE (self);
-	if (priv->main_ctx_sync != NULL)
-		return g_main_context_ref (priv->main_ctx_sync);
+	if (priv->main_ctx != NULL)
+		return g_main_context_ref (priv->main_ctx);
 	return g_main_context_ref_thread_default ();
 }
 
@@ -385,7 +384,8 @@ fwupd_client_get_main_context (FwupdClient *self)
  * @self: A #FwupdClient
  * @main_ctx: (nullable): #GMainContext or %NULL
  *
- * Sets the internal #GMainContext to use for synchronous methods.
+ * Sets the internal #GMainContext to use for returning progress
+ * signals and for synchronous methods.
  *
  * Since: 1.5.3
  **/
@@ -393,11 +393,11 @@ void
 fwupd_client_set_main_context (FwupdClient *self, GMainContext *main_ctx)
 {
 	FwupdClientPrivate *priv = GET_PRIVATE (self);
-	if (main_ctx == priv->main_ctx_sync)
+	if (main_ctx == priv->main_ctx)
 		return;
-	g_clear_pointer (&priv->main_ctx_sync, g_main_context_unref);
+	g_clear_pointer (&priv->main_ctx, g_main_context_unref);
 	if (main_ctx != NULL)
-		priv->main_ctx_sync = g_main_context_ref (main_ctx);
+		priv->main_ctx = g_main_context_ref (main_ctx);
 }
 
 /**
@@ -2202,8 +2202,8 @@ fwupd_client_install_stream_async (FwupdClient *self,
  * Install firmware onto a specific device.
  *
  * NOTE: This method is thread-safe, but progress signals will be
- * emitted in the thread default main context of the thread which
- * called fwupd_client_new().
+ * emitted in the default main context, if not explicitly set with
+ * fwupd_client_set_main_context().
  *
  * Since: 1.5.0
  **/
@@ -2280,8 +2280,8 @@ fwupd_client_install_bytes_finish (FwupdClient *self, GAsyncResult *res, GError 
  * Install firmware onto a specific device.
  *
  * NOTE: This method is thread-safe, but progress signals will be
- * emitted in the thread default main context of the thread which
- * called fwupd_client_new().
+ * emitted in the default main context, if not explicitly set with
+ * fwupd_client_set_main_context().
  *
  * Since: 1.5.0
  **/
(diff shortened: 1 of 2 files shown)


hughsie:
With much help from @pwithnall -- many thanks to him!

superm1:
Looks like it breaks our QT thread thing..

hughsie:
> Looks like it breaks our QT thread thing..

Ya, I've engaged the experts who are looking at it now. :)

hughsie:
@pwithnall I'd appreciate another sanity check please - thanks.

hughsie:
@aleixpol does this break anything for you in the real app? I think it might help...

hughsie:
Gahh, no cigar:

    $ ./qt-thread-test
    GLib-CRITICAL **: 21:06:40.378: g_main_context_push_thread_default: assertion 'acquired_context' failed

hughsie:
Breaking on `g_main_context_acquire`:

    #0  g_main_context_acquire (context=context@entry=0x444150) at ../glib/gmain.c:3385
    #1  0x00007ffff79728e2 in g_main_context_push_thread_default (context=0x444150) at ../glib/gmain.c:821
    #2  0x00007ffff7d21a9d in QEventDispatcherGlibPrivate::QEventDispatcherGlibPrivate(_GMainContext*) (this=0x42e580, context=<optimized out>) at kernel/qeventdispatcher_glib.cpp:317
    #3  0x00007ffff7d21c65 in QEventDispatcherGlib::QEventDispatcherGlib(QObject*) (this=0x40d4d0, parent=0x0) at kernel/qeventdispatcher_glib.cpp:358
    #4  0x00007ffff7b3964f in QThreadPrivate::createEventDispatcher(QThreadData*) (data=<optimized out>) at thread/qthread_unix.cpp:258
    #5  0x00007ffff7b37f73 in QThreadData::createEventDispatcher() (this=0x42e680) at thread/qthread.cpp:126
    #6  0x00007ffff7cdab18 in QCoreApplicationPrivate::createEventDispatcher() (this=0x42f1c0) at kernel/qcoreapplication.cpp:525
    #7  QCoreApplicationPrivate::init() (this=<optimized out>) at kernel/qcoreapplication.cpp:834
    #8  0x00007ffff7cdb0b0 in QCoreApplication::QCoreApplication(int&, char**, int) (this=0x7fffffffdd00, argc=@0x7fffffffdcdc: 1, argv=0x7fffffffde68, _internal=331522)
        at kernel/qcoreapplication.h:86
    #9  0x00000000004024f2 in main(int, char**) (argc=1, argv=0x7fffffffde68) at ../qt-thread-test.cpp:17

I'm so far out of my comfort zone here.

superm1:
> ```
> GLib-CRITICAL **: 21:06:40.378: g_main_context_push_thread_default: assertion 'acquired_context' failed
> ```

huh?  How did that pass CI.  Shouldn't a critical make it abort?
> 1/1 qt-thread-test OK             0.02s

hughsie:
>  Shouldn't a critical make it abort?

Only if G_DEBUG=fatal-criticals is set; I'll try adding that to the CI now...

hughsie:
> 1/1 qt-thread-test OK             0.02s

Huh, no criticals... How odd. @superm1 do you get critical warnings or is my system cursed?

superm1:
> > 1/1 qt-thread-test OK             0.02s
> 
> Huh, no criticals... How odd. @superm1 do you get critical warnings or is my system cursed?

fails for me too..

pwithnall:
The latest version looks reasonable to me. I wonder if the Qt test needs a new `GMainContext` to be pushed as the new thread-default main context when it spawns its worker thread? Currently, there is actually only one `GMainContext` created in the test, and that’s the global default one, aiui. That means the worker thread tries to iterate or push that one at the same time as the main thread iterating it, and then bad things happen.

libfwupd/fwupd-client.c
```diff
@@ -491,6 +494,11 @@ fwupd_client_connect_get_bus_cb (GObject *source,
 		g_task_return_error (task, g_steal_pointer (&error));
 		return;
 	}
+
+	/* we want the GDBusProxy to emit signals in the thread default
+	 * main context of the thread which called fwupd_client_new()
+	 * not the first one that caused a fwupd_client_connect_async()... */
+	g_main_context_push_thread_default (priv->main_ctx_prxy);
```
pwithnall, line 501: I think the solution for the Qt problem is to actually drop this push/pop pair, and instead change the `fwupd_client_properties_changed_cb()` and `fwupd_client_signal_cb()` functions to `g_idle_add_full()` the signal emissions through to the `main_ctx_prxy`. That has the same effect as these lines, but would avoid the critical warning you’re seeing from this push line.

The critical warning happens because this `GMainContext` is being pushed in the worker thread, while also being iterated in the main thread.

pwithnall:
I wrote a test program (below) which is roughly equivalent to the Qt test program: it creates a `FwupdClient` in the main thread, and then calls `fwupd_client_get_devices()` on it in a worker thread. It replicates the critical warning you were seeing with the Qt test. If it looks OK to you, you may want to clean it up and merge it as a test so that it gets run by `meson test` (the Qt test currently doesn’t, presumably for dependency reasons).

```
/* gcc -ggdb -Wall -o test test.c $(pkg-config --cflags --libs gio-2.0 fwupd) */
#include <glib.h>
#include <fwupd.h>

typedef struct
{
  GApplication *app;
  FwupdClient *client;
  GThread *worker_thread;
} AppData;

static gpointer
thread_cb (gpointer user_data)
{
  AppData *data = user_data;
  g_autoptr(GError) local_error = NULL;
  g_autoptr(GPtrArray) devices = NULL;
  g_autoptr(GMainContext) context = g_main_context_new ();
  g_autoptr(GMainContextPusher) pusher = g_main_context_pusher_new (context);

  g_message ("Calling fwupd_client_get_devices() in thread %p with main context %p", g_thread_self (), g_main_context_get_thread_default ());
  devices = fwupd_client_get_devices (data->client, NULL, &local_error);
  g_application_release (data->app);

  return g_steal_pointer (&devices);
}

static gboolean
idle_cb (gpointer user_data)
{
  AppData *data = user_data;

  g_message ("idle_cb() in thread %p with main context %p", g_thread_self (), g_main_context_get_thread_default ());
  data->worker_thread = g_thread_new ("worker", thread_cb, data);

  return G_SOURCE_REMOVE;
}

static void
activate_cb (GApplication *app,
             gpointer      user_data)
{
  AppData *data = user_data;

  g_application_hold (data->app);
  g_idle_add (idle_cb, data);
}

int
main (void)
{
  g_autoptr(GApplication) app = g_application_new ("org.test.Test", 0);
  g_autoptr(FwupdClient) client = fwupd_client_new ();
  g_autoptr(GThread) thread = NULL;
  gint retval;
  AppData data = { app, client, NULL };

  g_message ("Created FwupdClient in thread %p with main context %p", g_thread_self (), g_main_context_get_thread_default ());
  g_signal_connect (app, "activate", G_CALLBACK (activate_cb), &data);
  retval = g_application_run (app, 0, NULL);
  g_thread_join (g_steal_pointer (&data.worker_thread));

  return retval;
}
```

libfwupd/fwupd-client.c
hughsie: > functions to g_idle_add_full() the signal emissions through to the main_ctx_prxy

Doesn't that imply g_idle_source_new(), g_source_set_callback() then g_source_attach() rather than g_idle_add_full()? I don't think you can specify the ctx for g_idle_add_full()

pwithnall: Yes, you’re right.

hughsie: Something like 50f149648a04072b5887de7011b3189fd8760d5b perhaps? I think it's sane to share the priv->idle_id for signals and properties, right? I'm just trying to avoid queuing up hundreds of events if there is nothing servicing the main context.

hughsie:
> I wrote a test program (below) which is roughly equivalent to the Qt test program

I'm not convinced it's the same; the QT program is automatically pop'ing the thread context in the thread handler, possibly "being helpful" -- your program actually works fine -- if you add a g_message after g_application_release then we see the function returns -- but for some reason the GApplication doesn't release properly. I assume g_application_release is thread safe?

pwithnall:
> I'm not convinced it's the same

Perhaps not 100% identically, but it did replicate the same critical warning and set of backtraces as you sent me from the Qt test.

Issue #2695, opened 2020-12-18
Add test for detecting the "AMI Test PK" in the HSI


dnoliver:
**Is your feature request related to a problem? Please describe.**

Several device manufacturers decide to ship the default "AMI Test PK" platform key instead of a Device Manufacturer specific one. In this cases, the owner of the device (user or organization) that wants to enable Secure Boot can do 2 things:
1. Ask the device manufacturer to ship a system with a ODM specific key instead of the "AMI Test PK"
2. Perform a full Secure Boot Customization as described in https://media.defense.gov/2020/Sep/15/2002497594/-1/-1/0/CTR-UEFI-Secure-Boot-Customization-UOO168873-20.PDF

I think the HSI could help detect devices that are shipping this key by default and recommend a remediation.

*NOTE to self*: I have no idea why my system keeps booting in Secure Boot mode with an already expired Platform Key

```shell
[~]$ mokutil --pk
[key 1]
SHA1 Fingerprint: a7:73:11:3b:af:af:51:29:aa:83:fd:09:12:e9:5d:a4:fa:55:5f:91
Certificate:
Data:
Version: 3 (0x2)
Serial Number:
(Negative)1b:ed:93:e2:59:4e:2b:60:be:6b:1f:01:c9:af:a6:37
Signature Algorithm: sha256WithRSAEncryption
Issuer: CN=DO NOT TRUST - AMI Test PK
Validity
Not Before: Apr 30 22:50:15 2013 GMT
Not After : Apr 30 22:50:14 2017 GMT
Subject: CN=DO NOT TRUST - AMI Test PK
Subject Public Key Info:
Public Key Algorithm: rsaEncryption
RSA Public-Key: (2048 bit)
Modulus:
00:9c:18:19:79:cd:2e:05:c9:f5:29:24:5b:f8:8e:
7f:40:9a:91:70:be:14:d9:fe:a2:d8:fd:33:60:e0:
6f:57:78:f4:0a:44:2a:d8:f5:cb:23:04:67:7f:25:
0f:cc:27:77:6b:2f:8e:e3:1c:f8:ad:9a:33:b8:0c:
26:a4:d4:41:a2:38:56:dc:eb:74:33:e0:61:09:bf:
79:a9:bb:7c:72:f8:0d:22:fa:4e:b8:2e:ed:a7:da:
e3:2f:13:9b:14:96:e8:4a:8a:80:cd:85:63:88:4f:
[...]

PR #2701, opened 2020-12-22
Check the AMI test key if not installed for HSI-1

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/2695

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


docs/hsi.xml
```diff
@@ -281,6 +281,34 @@
       </note>
     </refsect3>
 
+    <refsect3 id="org.fwupd.hsi.Uefi.PK.TestKey">
```
superm1, line 284: I think it's better to label this as "unsecure key".  The logic behind it being that if any evidence of another platform key that is not properly secured, even if not intended for testing was discovered it could be added to this as a disallow list.

hughsie: Good idea. Unsecure or insecure? I know there's a difference, but the first looks weird.

plugins/uefi-dbx/README.md
```diff
@@ -45,3 +45,22 @@ External interface access
...
+
+Several device manufacturers decide to ship the default "AMI Test PK"
+platform key instead of a Device Manufacturer specific one. This will
+cause an HSI-1 failure.
+
+External interface access
+-------------------------
+
+This plugin requires:
+* read/write access to `/sys/firmware/efi/efivars`
```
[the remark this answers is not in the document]
hughsie: Agree, thanks.

docs/hsi.xml
hughsie: Actually, all the HSI keys are in the "positive" so it needs to be `NotUnsecureKey` or maybe just `org.fwupd.hsi.Uefi.PK.Valid`

hughsie: Maybe just `org.fwupd.hsi.Uefi.PK` too; with states "valid" or "invalid"

superm1: Don't have a strong feeling between your last 3 comments.  Whichever sounds best to you is fine.

hughsie: I went with the latter...

meson.build
```diff
@@ -249,6 +249,10 @@ gcab = find_program('gcab', required : get_option('tests'))
 bashcomp = dependency('bash-completion', required: false)
 python3 = find_program('python3')
 
+if get_option('plugin_uefi_pk')
+  gnutls = dependency('gnutls', version : '>= 3.6.0')
```
superm1, line 253: Do you know if this was the same version needed for libjcat?  If possible we should align the version dependencies if they're skewed.

superm1 (commented):
Since this has it's own meson directive, I think it probably shouldn't share a directory with uefi-dbx.

meson.build
hughsie: Yes, I chose the same value deliberately.

hughsie:
> I think it probably shouldn't share a directory with uefi-dbx

If we do that, it means moving all this to libfwupdplugin:

    'fu-efi-image.c',
    'fu-efi-signature.c',
    'fu-efi-signature-common.c',
    'fu-efi-signature-list.c',
    'fu-efi-signature-parser.c',

superm1:
> > I think it probably shouldn't share a directory with uefi-dbx
> 
> If we do that, it means moving all this to libfwupdplugin:
> 
> ```
> 'fu-efi-image.c',
> 'fu-efi-signature.c',
> 'fu-efi-signature-common.c',
> 'fu-efi-signature-list.c',
> 'fu-efi-signature-parser.c',
> ```

By that logic, I would question should all 3 UEFI plugins collapse into `uefi`?  Or maybe this directory renames to `uefi-security`

hughsie:
> By that logic, I would question should all 3 UEFI plugins collapse into uefi?

I think the current uefi plugin should probably be called `uefi-capsule` or something; I don't think a single plugin for all the things UEFI provides is super maintainable, and also because UEFI-isms leak out to the other plugins too.

> Or maybe this directory renames to uefi-security

Yes, uefi-security is probably a better name if we put uefi_pk and uefi_dbx in the same logical directory. From, a maintainability point of view it's a tough one as ideally all plugins live in self-contained directories, but the cost of "maintaining" all those extra symbols in libfwupdplugin is nonzero. Maybe FuEfiSignature could be rewritten to subclass FuFirmwareImage and FuEfiSignatureList subclass FuFirmware? Then we get all the builder and fuzzing stuff "for free" at the cost of having to squint in with one eye shut.

superm1:
In the short term I'm fine with rename `uefi` -> `uefi-capsule` and `uefi-dbx` to `uefi-security`.  

>Maybe FuEfiSignature could be rewritten to subclass FuFirmwareImage and FuEfiSignatureList subclass FuFirmware? Then we get all the builder and fuzzing stuff "for free" at the cost of having to squint in with one eye shut.

The other thing you can do is collapse uefi-dbx plugin and uefi-pk plugin into a single `uefi-security` plugin rather than individual stubs like this.

PR #2705, opened 2020-12-26
Add plugin to update PixArt RF devices

hughsie:

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/pixart-rf/fu-pxi-device.c
+++ b/plugins/pixart-rf/fu-pxi-device.c
@@ -1,0 +1,626 @@
+/*
+ * Copyright (C) 2020 Jimmy Yu <Jimmy_yu@pixart.com>
+ * Copyright (C) 2020 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#ifdef HAVE_HIDRAW_H
+#include <linux/hidraw.h>
+#include <linux/input.h>
+#endif
+
+#include "fu-common.h"
+#include "fu-chunk.h"
+
+#include "fu-pxi-device.h"
+#include "fu-pxi-firmware.h"
+
+#define PXI_HID_DEV_OTA_INPUT_REPORT_ID		0x05
+#define PXI_HID_DEV_OTA_OUTPUT_REPORT_ID	0x06
+#define PXI_HID_DEV_OTA_FEATURE_REPORT_ID	0x07
+
+#define FU_PXI_DEVICE_CMD_FW_OTA_INIT		0x10u
+#define FU_PXI_DEVICE_CMD_FW_WRITE		0x17u
+#define FU_PXI_DEVICE_CMD_FW_UPGRADE		0x18u
+#define FU_PXI_DEVICE_CMD_FW_MCU_RESET		0x22u
+#define FU_PXI_DEVICE_CMD_FW_GET_INFO		0x23u
+#define FU_PXI_DEVICE_CMD_FW_OBJECT_CREATE	0x25u
+#define FU_PXI_DEVICE_CMD_FW_OTA_INIT_NEW	0x27u
+#define FU_PXI_DEVICE_CMD_FW_OTA_RETRANSMIT	0x28u
+#define FU_PXI_DEVICE_CMD_FW_OTA_DISCONNECT	0x29u
+
+#define FU_PXI_DEVICE_OBJECT_SIZE_MAX		4096	/* bytes */
+#define FU_PXI_DEVICE_OTA_BUF_SZ		32	/* bytes */
+#define FU_PXI_DEVICE_NOTTFY_RET_LEN		4	/* bytes */
+#define FU_PXI_DEVICE_FW_INFO_RET_LEN		8	/* bytes */
+
+/* OTA target selection */
+enum ota_process_setting {
+	OTA_MAIN_FW,				/* Main firmware */
+	OTA_HELPER_FW,				/* Helper firmware */
+	OTA_EXTERNAL_RESOURCE,			/* External resource */
+};
+
+/* OTA spec check result */
+enum ota_spec_check_result {
+	OTA_SPEC_CHECK_OK		= 1,	/* Spec check ok */
+	OTA_FW_OUT_OF_BOUNDS		= 2,	/* OTA firmware size out of bound */
+	OTA_PROCESS_ILLEGAL		= 3,	/* Illegal OTA process */
+	OTA_RECONNECT			= 4,	/* Inform OTA app do reconnect */
+	OTA_FW_IMG_VERSION_ERROR	= 5,	/* FW image file version check error */
+	OTA_SPEC_CHECK_MAX_NUM,			/* Max number of OTA driver defined error code */
+};
+
+/* OTA disconnect reason */
+enum ota_disconnect_reason {
+	OTA_CODE_JUMP			= 1,	/* OTA code jump */
+	OTA_UPDATE_DONE			= 2,	/* OTA update done */
+	OTA_RESET,				/* OTA reset */
+};
+
+struct _FuPxiDevice {
+	FuUdevDevice	 parent_instance;
+	guint8		 status;
+	guint8		 new_flow;
+	guint16		 offset;
+	guint16		 checksum;
+	guint32		 max_object_size;
+	guint16		 mtu_size;
+	guint16		 prn_threshold;
+	guint8		 spec_check_result;
+};
+
+G_DEFINE_TYPE (FuPxiDevice, fu_pxi_device, FU_TYPE_UDEV_DEVICE)
+
+static const gchar *
+fu_pxi_device_spec_check_result_to_string (guint8 spec_check_result)
+{
+	if (spec_check_result == OTA_SPEC_CHECK_OK)
+		return "ok";
+	if (spec_check_result == OTA_FW_OUT_OF_BOUNDS)
+		return "fw-out-of-bounds";
+	if (spec_check_result == OTA_PROCESS_ILLEGAL)
+		return "process-illegal";
+	if (spec_check_result == OTA_RECONNECT)
+		return "reconnect";
+	if (spec_check_result == OTA_FW_IMG_VERSION_ERROR)
+		return "fw-img-version-error";
+	return NULL;
+}
+
+static void
+fu_pxi_device_to_string (FuDevice *device, guint idt, GString *str)
+{
+	FuPxiDevice *self = FU_PXI_DEVICE (device);
+	fu_common_string_append_kx (str, idt, "Status", self->status);
+	fu_common_string_append_kx (str, idt, "NewFlow", self->new_flow);
+	fu_common_string_append_kx (str, idt, "CurrentObjectOffset", self->offset);
+	fu_common_string_append_kx (str, idt, "CurrentChecksum", self->checksum);
+	fu_common_string_append_kx (str, idt, "MaxObjectSize", self->max_object_size);
+	fu_common_string_append_kx (str, idt, "MtuSize", self->mtu_size);
+	fu_common_string_append_kx (str, idt, "PacketReceiptNotificationThreshold", self->prn_threshold);
+	fu_common_string_append_kv (str, idt, "SpecCheckResult",
+				    fu_pxi_device_spec_check_result_to_string (self->spec_check_result));
+}
+
+static FuFirmware *
+fu_pxi_device_prepare_firmware (FuDevice *device,
+				GBytes *fw,
+				FwupdInstallFlags flags,
+				GError **error)
+{
+	g_autoptr(FuFirmware) firmware = fu_pxi_rf_firmware_new ();
+	if (!fu_firmware_parse (firmware, fw, flags, error))
+		return NULL;
+	return g_steal_pointer (&firmware);
+}
+
+static gboolean
+fu_pxi_device_set_feature (FuPxiDevice *self, const guint8 *buf, guint bufsz, GError **error)
+{
+#ifdef HAVE_HIDRAW_H
+	if (g_getenv ("FWUPD_PIXART_RF_VERBOSE") != NULL)
+		fu_common_dump_raw (G_LOG_DOMAIN, "SetFeature", buf, bufsz);
+	return fu_udev_device_ioctl (FU_UDEV_DEVICE (self),
+				     HIDIOCSFEATURE(bufsz), (guint8 *) buf,
+				     NULL, error);
+#else
+	g_set_error_literal (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_NOT_SUPPORTED,
+			     "<linux/hidraw.h> not available");
+	return FALSE;
+#endif
+}
+
+static gboolean
+fu_pxi_device_get_feature (FuPxiDevice *self, guint8 *buf, guint bufsz, GError **error)
+{
+#ifdef HAVE_HIDRAW_H
+	if (!fu_udev_device_ioctl (FU_UDEV_DEVICE (self),
+				   HIDIOCGFEATURE(bufsz), buf,
+				   NULL, error)) {
+		return FALSE;
+	}
+	if (g_getenv ("FWUPD_PIXART_RF_VERBOSE") != NULL)
+		fu_common_dump_raw (G_LOG_DOMAIN, "GetFeature", buf, bufsz);
+	return TRUE;
+#else
+	g_set_error_literal (error,
+			     G_IO_ERROR,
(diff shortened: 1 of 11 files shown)


plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,515 @@
...
+	return checksum;
+}
+
+/* FIXME: is @port required? */
+static gboolean
+fu_pxi_device_wait_notify (FuPxiDevice *self,
+			   goffset port,
+			   guint8 *status,
+			   guint16 *checksum,
+			   GError **error)
```
hughsie, line 135: @jyu99 can you answer the FIXME please

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,515 @@
...
+					 req->data, req->len, error))
+		return FALSE;
+
+	/* reply */
+	if (!fu_udev_device_pread_full (FU_UDEV_DEVICE (self), 0x0,
+					res, sizeof(res), error))
+		return FALSE;
+	g_usleep (30 * 1000);
+
+	/* res seems unused */
```
hughsie, line 183: @jyu99 is that expected?

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,515 @@
...
+	return TRUE;
+}
+
+static gboolean
+fu_pxi_device_reset (FuPxiDevice *self, gsize fw_sz, GError **error)
+{
+	guint8 req[FU_PXI_DEVICE_OTA_BUF_SZ] = {
+		PXI_HID_DEV_OTA_OUTPUT_REPORT_ID,
+		FU_PXI_DEVICE_CMD_FW_MCU_RESET,
+		fw_sz, /* FIXME: this is size_t being sqeezed into uint8_t?! */
```
hughsie, line 257: @jyu99 is that right? It's a bit weird to squeeze a big number into just one byte...

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,515 @@
...
+		return FALSE;
+	if (!fu_common_read_uint16_safe (res, sizeof(res), 0xc,
+					 &self->mtu_size, G_LITTLE_ENDIAN, error))
+		return FALSE;
+	if (!fu_common_read_uint16_safe (res, sizeof(res), 0xe,
+					 &self->prn_threshold, G_LITTLE_ENDIAN, error))
+		return FALSE;
+	if (!fu_common_read_uint8_safe (res, sizeof(res), 0x10,
+					&self->spec_check_result, error))
+		return FALSE;
```
hughsie, line 332: @jyu99 is this offset correct?

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,515 @@
...
+
+	/* set current version */
+	version_str = g_strndup ((gchar *) req + 0x3, 5);
+	fu_device_set_version (FU_DEVICE (self), version_str);
+
+	/* add current checksum */
+	if (!fu_common_read_uint16_safe (req, sizeof(req), 0x8,
+					 &checksum, G_LITTLE_ENDIAN, error))
+		return FALSE;
+	checksum_str = g_strdup_printf ("0x%4x", checksum);
```
hughsie, line 471: @jyu99 so this is just a uint16_t byte sum, right? I don't know if this is "good enough" to use for verification. @superm1 what do you think?

plugins/pixart-rf/fu-pxi-device.c
jyu99: I will remove the delay thank your mention

plugins/pixart-rf/fu-pxi-device.c
jyu99: I will fix this bug , It should be a OTA_RESET in  previous

 enum ota_disconnect_reason
{
    OTA_CODE_JUMP = 1,          ///< OTA code jump
    OTA_UPDATE_DONE = 2,        ///< OTA update done
    OTA_RESET,                  ///< OTA reset
};

plugins/pixart-rf/fu-pxi-device.c
jyu99: Yes,our checksum is only unit16_byte

jyu99: our checksum is most

plugins/pixart-rf/fu-pxi-device.c
jyu99: Ok, I will check again base this commitm thank you

plugins/pixart-rf/fu-pxi-device.c
jyu99: Could I add previous `fu-pxi-device-ota.h`,becasue a lot of enum and structure  that  I need to use

jyu99:
[](plugins/pixart-rf/fu-pxi-device.c)

jyu99:
```
	fu_byte_array_append_uint8 (req, PXI_HID_DEV_OTA_OUTPUT_REPORT_ID);
	fu_byte_array_append_uint8 (req, FU_PXI_DEVICE_CMD_FW_OTA_INIT_NEW);
	fu_byte_array_append_uint32 (req, fw_sz, G_LITTLE_ENDIAN);
	fu_byte_array_append_uint8 (req, 0x0);	/* OTA setting */
	g_byte_array_append (req, fw_version, sizeof(fw_version));
	if (!fu_udev_device_pwrite_full (FU_UDEV_DEVICE (self), 0x0,
					 req->data, req->len, error))
```

Could use structure in previous `fu-pxi-device-ota.h ` to replace  array_append?

plugins/pixart-rf/fu-pxi-device.c
hughsie: It might be the hardware *needs* a read after the write; this is fine, but it would be nice to document this if so.

plugins/pixart-rf/fu-pxi-device.c
hughsie: > It should be a OTA_RESET

That makes sense. Just add this enum at the top of fu-pxi-device.c -- you don't need the enum in any other C files...

hughsie:
> Could use structure in previous fu-pxi-device-ota.h to replace array_append

No, I removed all the structure accessors as they are not endian safe; they're convenient, but they hide endian issues -- this means the code would do very unpredictable things on architectures like ppc64be or MIPS. It's much better to construct the buffers explicitly.

plugins/pixart-rf/fu-pxi-device.c
hughsie: I *think* the port can be zero in all cases, but didn't want to assume.

plugins/pixart-rf/fu-pxi-device.c
jyu99: Could I use below to replace append
```
struct _FuPxiDevice {
	FuUdevDevice	 parent_instance;
	ret_fw_ota_init_new_cmd	ret_fw_ota_init_new;
};
```
```
	if (!fu_memcpy_safe ((guint8 *) &self->ret_fw_ota_init_new, sizeof(ret_fw_ota_init_new_cmd), 0,		/* dst */
			     (guint8 *)  &evt.evt_param.ret_param.fw_ota_init_new, sizeof(ret_fw_ota_init_new_cmd), 0,			/* src */
			     sizeof(ret_fw_ota_init_new_cmd), error))
```

hughsie: No, because all the values like max_object_size are going to be wrong on big endian systems.

plugins/pixart-rf/fu-pxi-device.c
jyu99: I want reserve it because current device is BLE, but we will support other USB HID.

hughsie: No problem; we can just remove the FIXME then.

jyu99:
[https://github.com/jyu99/fwupd](url)
Hi 
I had fixed some code and flow , please help me review this commit  Thank you !!

plugins/pixart-rf/fu-pxi-device.c
hughsie: I've removed it, it'll make the security team nervous...

hughsie:
@jyu99 this looks good to me now. Can you confirm the version on this branch works on actual hardware again please. Thanks!

hughsie:
@jyu99 I've also added `com.pixart.rf` as a protocol to the LVFS now if you want to do an end-to-end test now.

jyu99:
I will confirm it again and try to end to end test

plugins/pixart-rf/meson.build
```diff
@@ -0,0 +1,25 @@
+cargs = ['-DG_LOG_DOMAIN="FuPluginPxiHid"']
```
superm1, line 1: Shouldn't this be `FuPluginPixartRf`?

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+#include "fu-pxi-firmware.h"
+
+#define PXI_HID_DEV_OTA_INPUT_REPORT_ID		0x05
+#define PXI_HID_DEV_OTA_OUTPUT_REPORT_ID	0x06
+#define PXI_HID_DEV_OTA_FEATURE_REPORT_ID	0x07
+
+#define FU_PXI_DEVICE_CMD_FW_OTA_INIT		0x10u
+#define FU_PXI_DEVICE_CMD_FW_WRITE		0x17u
+#define FU_PXI_DEVICE_CMD_FW_UPGRADE		0x18u
+#define FU_PXI_DEVICE_CMD_FW_MCU_RESET		0x22u	/* remove after windows OTA app ready */
```
superm1, line 28: Can this comment please be expanded?  Does this mean that the reset should go away when a future firmware comes that can handle the upgrade differently?  Leaving a comment like this makes me worry about maintainability.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+{
+	FuPxiDevice *self = FU_PXI_DEVICE (device);
+	fu_common_string_append_kx (str, idt, "Status", self->status);
+	fu_common_string_append_kx (str, idt, "NewFlow", self->new_flow);
+	fu_common_string_append_kx (str, idt, "CurrentObjectOffset", self->offset);
+	fu_common_string_append_kx (str, idt, "CurrentChecksum", self->checksum);
+	fu_common_string_append_kx (str, idt, "MaxObjectSize", self->max_object_size);
+	fu_common_string_append_kx (str, idt, "MtuSize", self->mtu_size);
+	fu_common_string_append_kx (str, idt, "PacketReceiptNotificationThreshold", self->prn_threshold);
+	fu_common_string_append_kx (str, idt, "SpecCheckResult", self->spec_check_result);
```
superm1, line 89: given there is actually an enum that decodes this, I think it makes sense to have a to_string helper for it that gives you messages.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+					&self->spec_check_result, error))
+		return FALSE;
+
+	/* sanity check */
+	if (self->spec_check_result != OTA_SPEC_CHECK_OK) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_READ,
+			     "FwInitNew Spec Check Fail 0x%02x",
+			     self->spec_check_result);
```
superm1, line 420: assuming the to string helper I mentioned gets made, it should be used here to.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+	/* send fw upgrade command */
+	fu_device_set_status (FU_DEVICE (self), FWUPD_STATUS_DEVICE_VERIFY);
+	if (!fu_udev_device_pwrite_full (FU_UDEV_DEVICE (self), 0,
+					 req->data, req->len, error))
+		return FALSE;
+	if (g_getenv ("FWUPD_PIXART_RF_VERBOSE") != NULL)
+		fu_common_dump_raw (G_LOG_DOMAIN, "fw upgrade", req->data, req->len);
+
+	/* read fw upgrade command result */
+	g_debug ("read fw upgrade result");
```
superm1, line 465: this seems like an unnecessary debug line

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+	return TRUE;
+}
+
+static void
+fu_pxi_device_init (FuPxiDevice *self)
+{
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_TRIPLET);
+	fu_device_set_vendor_id (FU_DEVICE (self), "USB:0x093A");
```
superm1, line 598: (depending if this goes in after the multiple vendor ID thingy, one of the PR's will need to update later)

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+	if (!fu_pxi_device_fw_get_info (self, error))
+		return FALSE;
+	return TRUE;
+}
+
+static void
+fu_pxi_device_init (FuPxiDevice *self)
+{
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
```
superm1, line 596: Is this flag really needed?  It seems like this is a straightforward USB device.  Am I wrong?

plugins/pixart-rf/meson.build
hughsie: Fixed.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -0,0 +1,611 @@
...
+					     opcode);
+				return FALSE;
+			}
+			prn = 0;
+		}
+	}
+
+	/* the last chunk */
+	self->checksum += checksum;
+	g_debug ("checksum 0x%04x, table checksum 0x%04x", checksum_device, self->checksum);
```
superm1, line 317: since you're already comparing the two of these and setting an error when they fail, it seems not useful to include

plugins/pixart-rf/fu-pxi-device.c
hughsie: @jyu99 can you explain please

plugins/pixart-rf/fu-pxi-device.c
hughsie: Agree, fixed.

plugins/pixart-rf/fu-pxi-device.c
hughsie: Fixed.

plugins/pixart-rf/fu-pxi-device.c
hughsie: Removed.

plugins/pixart-rf/fu-pxi-device.c
hughsie: I think the multiple vendor one needs more thinking (and certainly more self tests!) so lets try getting this one in first.

plugins/pixart-rf/fu-pxi-device.c
hughsie: > Am I wrong?

Of course not! I've removed it :)

plugins/pixart-rf/fu-pxi-device.c
hughsie: agree, removed.

superm1 (approved):

plugins/pixart-rf/fu-pxi-device.c
jyu99: It’s for early version using ,so I will remove it. Thank you suggestions

hughsie: you mean remove the _comment_? right?

jyu99: Yes.. remove the comment

PR #2707, opened 2020-12-28
Make libcurl optional for fuzzing

hughsie:
For fuzzing we want to exclude libcurl support as it depends on other very heavy
libraries like OpenSSL or libtasn which make the fuzzing binary much larger if
linked statically.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupd/fwupd-client.c
```diff
@@ -106,14 +108,14 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(CURLU, curl_url_cleanup)
...
+#ifdef HAVE_LIBCURL
 	if (helper->curl != NULL)
 		curl_easy_cleanup (helper->curl);
-#ifdef HAVE_LIBCURL_7_56_0
 	if (helper->mime != NULL)
 		curl_mime_free (helper->mime);
-#endif
 	if (helper->headers != NULL)
 		curl_slist_free_all (helper->headers);
+#endif
```
superm1, line 118: it's a bit funny to have functions like fwupd_client_curl_helper_free when you don't actually have curl support..

libfwupd/meson.build
```diff
@@ -125,8 +128,7 @@ if get_option('introspection')
     export_packages : 'fwupd',
     header : 'fwupd.h',
     dependencies : [
-      giounix,
-      libcurl,
+      libfwupd_deps,
```
superm1, line 131: this is actually adding more dependencies than needed

libfwupd/fwupd-client.c
hughsie: Totally valid, fixed.

libfwupd/meson.build
hughsie: Agree, good catch. Fixed.

meson.build
```diff
@@ -219,12 +219,12 @@ endif
...
 libjsonglib = dependency('json-glib-1.0', version : '>= 1.1.1')
 valgrind = dependency('valgrind', required: false)
-libcurl = dependency('libcurl', version : '>= 7.47.0')
-if libcurl.version().version_compare('>= 7.56.0')
-  conf.set('HAVE_LIBCURL_7_56_0', '1')
-endif
-if libcurl.version().version_compare('>= 7.62.0')
-  conf.set('HAVE_LIBCURL_7_62_0', '1')
+if get_option('curl')
+  libcurl = dependency('libcurl', version : '>= 7.56.0')
```
superm1, line 223: I guess the reason you let it go back to 7.47.0 was just for the version on the fuzzer right?  So that's why bumping back up to 7.56.0 is OK?

superm1 (approved):

meson.build
hughsie: Exactly that. The static binaries were huuuge with all the libcurl deps linked in, and "no libcurl" is better than "old libcurl".

PR #2716, opened 2021-01-03
Allow specifying more than one VendorID for a device

hughsie:
Asking the user for the UID mapping isn't working very well, as it requires lots
of manual handholding. It also doesn't work very well when the device vendor
does not actually have a PCI ID or if the vendor has split into two entities.

Just use the OUI address as an additional VendorID and match any of the device
IDs against any of the metadata-supplied values.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


libfwupd/fwupd-device.c
```diff
@@ -659,11 +660,13 @@ fwupd_device_set_vendor (FwupdDevice *device, const gchar *vendor)
  * fwupd_device_get_vendor_id:
  * @device: A #FwupdDevice
  *
- * Gets the device vendor ID.
+ * Gets the combined device vendor ID.
  *
  * Returns: the device vendor, e.g. 'USB:0x1234', or %NULL if unset
```
superm1, line 665: I think that this needs to be updated to show what a combined ID looks like

plugins/ata/fu-ata-device.c
```diff
@@ -394,28 +387,17 @@ fu_ata_device_parse_id (FuAtaDevice *self, const guint8 *buf, gsize sz, GError *
...
+	}
 
 	/* if not already set using the vendor block or a OUI quirk */
 	name = fu_ata_device_get_string (id, 27, 46);
 	if (name != NULL && !has_oui_quirk)
 		fu_ata_device_parse_vendor_name (self, name);
 
-	/* ask user to report data */
-	if (self->oui != 0x0 && !has_oui_quirk && self->unknown_oui_report) {
-		const gchar *url = "https://github.com/fwupd/fwupd/wiki/ATA-Disk:-OUI-Quirk-Required";
```
superm1, line 405: if this is merged, need to remember to update the wiki page about this.

src/fu-device-list.c
```diff
@@ -580,11 +580,14 @@ fu_device_list_replace (FuDeviceList *self, FuDeviceItem *item, FuDevice *device
...
-		g_debug ("copying old vendor ID %s to new device", vendor_id);
-		fu_device_set_vendor_id (device, vendor_id);
+	if (fu_device_get_vendor_ids (item->device)->len > 0 &&
+	    fu_device_get_vendor_ids(device)->len == 0) {
+		GPtrArray *vendor_ids = fu_device_get_vendor_ids (item->device);
+		for (guint i = 0; i < vendor_ids->len; i++) {
+			const gchar *vendor_id = g_ptr_array_index (vendor_ids, i);
+			g_debug ("copying old vendor ID %s to new device", vendor_id);
+			fu_device_add_vendor_id (device, vendor_id);
+		}
```
superm1, line 590: just wondering if there is possibly a bug here where there are two vendor IDs on donor and one on the new.  possibly these are the same and you just need the one from the donor?

plugins/ata/fu-ata-device.c
hughsie: Agree; I wasn't sure what your opinions would be.

src/fu-self-test.c
```diff
@@ -2194,7 +2194,7 @@ fu_device_list_compatible_func (gconstpointer user_data)
 	g_assert_cmpint (changed_cnt, ==, 1);
 
 	/* device2 should inherit the vendor ID and version from device1 */
-	g_assert_cmpstr (fu_device_get_vendor_id (device2), ==, "USB:0x20A0");
+	g_assert_true (fu_device_has_vendor_id (device2, "USB:0x20A0"));
```
superm1, line 2197: I think the self tests need to add some new corner cases for multiple vendor IDs and how the interactions with them work.

src/fu-device-list.c
hughsie: if they're the same they get deduped by fu_device_add_vendor_id -- or do I misunderstand?

superm1: But you never run the copy the ID code path above if the new device has a single ID set.  Right now it's a conservative, "don't destroy any IDs if something is set", "don't add IDs if something is set".  So I think it should be either replace all ID's or add to the existing IDs if some are set.

hughsie: Ohh I see what you mean, gotcha. I'm tempted to say "always copy vendor IDs" is probably the sanest behaviour; it's the same device afterall... i.e. we'd remove both conditionals in the `if` section.

superm1: Sounds good.

plugins/ata/fu-ata-device.c
superm1: Yeah just want to leave it on the PR as a reminder so if we merge this it's the last open item to do.

libfwupd/fwupd-device.c
hughsie: I used the `|` char as a logical OR; other variables like this already use `,` -- any preference?

src/fu-device-list.c
hughsie: Fixed.

libfwupd/fwupd-device.c
superm1: seems fine to me.

superm1 (approved):

libfwupd/fwupd-device.c
hughsie: Example updated.

src/fu-self-test.c
hughsie: I've added in one on either side.

plugins/ata/fu-ata-device.c
hughsie: Done!

PR #2719, opened 2021-01-04
Do not export useless device attributes to the client

hughsie:
That giant uint64_t isn't looking so big now, and we'll want to add even more
to it in the future. Split out some private flags that are never useful to the
client, although the #defines will have to remain until we break API again.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.h
+++ b/libfwupdplugin/fu-device.h
@@ -193,6 +193,34 @@ FuDevice	*fu_device_new				(void);
 #define fu_device_get_flashes_left(d)		fwupd_device_get_flashes_left(FWUPD_DEVICE(d))
 #define fu_device_get_install_duration(d)	fwupd_device_get_install_duration(FWUPD_DEVICE(d))
 
+/**
+ * FuDeviceInternalFlags:
+ * @FU_DEVICE_INTERNAL_FLAG_NONE:			No flags set
+ * @FU_DEVICE_INTERNAL_FLAG_NO_AUTO_INSTANCE_IDS:	Do not add instance IDs from the device baseclass
+ * @FU_DEVICE_INTERNAL_FLAG_ENSURE_SEMVER:		Ensure the version is a valid semantic version, e.g. numbers separated with dots
+ * @FU_DEVICE_INTERNAL_FLAG_ONLY_SUPPORTED:		Only devices supported in the metadata will be opened
+ * @FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME:		Set the device name from the metadata `name` if available
+ * @FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME_CATEGORY:	Set the device name from the metadata `category` if available
+ * @FU_DEVICE_INTERNAL_FLAG_MD_SET_VERFMT:		Set the device version format from the metadata if available
+ * @FU_DEVICE_INTERNAL_FLAG_MD_SET_ICON:		Set the device icon from the metadata if available
+ * @FU_DEVICE_INTERNAL_FLAG_RETRY_OPEN:			Retry the device open up to 5 times if it fails
+ *
+ * The device internal flags.
+ **/
+typedef enum {
+	FU_DEVICE_INTERNAL_FLAG_NONE			= 0,
+	FU_DEVICE_INTERNAL_FLAG_NO_AUTO_INSTANCE_IDS	= (1llu << 0),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_ENSURE_SEMVER		= (1llu << 1),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_ONLY_SUPPORTED		= (1llu << 2),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME		= (1llu << 3),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME_CATEGORY	= (1llu << 4),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_MD_SET_VERFMT		= (1llu << 5),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_MD_SET_ICON		= (1llu << 6),	/* Since: 1.5.5 */
+	FU_DEVICE_INTERNAL_FLAG_RETRY_OPEN		= (1llu << 7),	/* Since: 1.5.5 */
+	/*< private >*/
+	FU_DEVICE_INTERNAL_FLAG_UNKNOWN			= G_MAXUINT64,
+} FuDeviceInternalFlags;
+
 /* accessors */
 gchar		*fu_device_to_string			(FuDevice	*self);
 const gchar	*fu_device_get_alternate_id		(FuDevice	*self);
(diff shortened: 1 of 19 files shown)


hughsie:
@superm1 if you agree with the concept there are a few other #defines we should move too. I didn't want to get in *too deep* without some sanity checking.

superm1 (commented):
I've seen stuff like this happen in the linux kernel too when they outgrow structures.  What tends to happen is a new attribute is made such as if they had `flags`, they make `flags2`.

They're a lot more hard pressed on ABI though and will never break it.

I guess I would question whether there is ANY value in deprecating stuff in the middle of the structure before a future ABI bump?  Should it just happen "with" the ABI bump to change how flags work at that time?

hughsie:
> I would question whether there is ANY value in deprecating stuff in the middle of the structure before a future ABI bump

It was mainly for people writing plugins and looking at the code for "_the WAITING_FOR_REPLUG_" thing.

> whether there is ANY value in deprecating stuff in the middle of the structure before a future ABI bump

I was intending to reuse the we-know-these-are-unused  "slots" if we ran out of public flags before we bumped API.

superm1:
> I was intending to reuse the we-know-these-are-unused "slots" if we ran out of public flags before we bumped API.

Sounds risky to me as there are a bunch of fwupd clients not in tree that might decide to do things based on those flags.

hughsie:
> Sounds risky to me as there are a bunch of fwupd clients not in tree that might decide to do things based on those flags.

True, but I also don't think it's plausible a front end client is doing anything interesting with SET_NAME_CATEGORY or NO_AUTO_INSTANCE_IDS -- and ONLY_SUPPORTED never makes it out of the daemon either.

Lets go back a bit; what about the name? Is `PrivateFlag` right, or should this indicate that the flag result is somehow not trusted to leak from the server? `TrustedFlag` totally overlaps with `FwupdTrustFlags` Maybe we just stop storing and sending a bitfield and instead use (and send over DBus) a list of integers? That way we can scale way beyond 64 entries, not worry about "which" flag to set, at the expense of lookup time. Or something else where we store a list of guint64s and then fwupd "knows" which uint64_t entry to set/clear/check somehow...

superm1:
>  Is PrivateFlag right, or should this indicate that the flag result is somehow not trusted to leak from the server

How about `InternalFlag` or `DaemonDeviceFlag` or `InternalDeviceFlag`

>Maybe we just stop storing and sending a bitfield and instead use (and send over DBus) a list of integers? That way we can scale way beyond 64 entries, not worry about "which" flag to set, at the expense of lookup time. Or something else where we store a list of guint64s and then fwupd "knows" which uint64_t entry to set/clear/check somehow...

Seems heavy for such a small amount of data.  Are we really going to have so many things a client needs to know about devices? I would sooner say just have a second 64 bitfield to spill into and make the client fetch two integers rather than 15 requests for an integer.

hughsie:
> How about InternalFlag or DaemonDeviceFlag or InternalDeviceFlag

What about `FuDeviceAttr`? -- it's an attribute (and internal attribute) of the device rather than a flag that might be useful to, well, flag things in the UI client. Or `FuDeviceState` although I'm not sure all the internal ones would strictly be state. Maybe just FuDeviceFlags is simplest, and change the prefix of the internal defines from FWUPD_ to FU_...

If I'm worrying about nothing (we're at 43/63) then tell me so. We could just revisit this in 2 years time, although I do think the earlier we ponder the problem the less we paint ourselves into a corner.

superm1:
>  Maybe just FuDeviceFlags is simplest, and change the prefix of the internal defines from FWUPD_ to FU_...

Ooh.  I like this idea.

> If I'm worrying about nothing (we're at 43/63) then tell me so. We could just revisit this in 2 years time, although I do think the earlier we ponder the problem the less we paint ourselves into a corner.

It's hard to know what you don't know of course.  But this was started what 4 or 5 years ago and we are 43, presumably slowing down.  I can't imagine we get another 20 by the time ABI breaks.

But then again how soon do you want to break ABI? 🤲

libfwupdplugin/fu-device.h
```diff
@@ -193,6 +193,35 @@ FuDevice	*fu_device_new				(void);
...
+	FU_DEVICE_FLAG_NO_AUTO_INSTANCE_IDS	= (1llu << 19),	/* Since: 1.5.5 */
+	FU_DEVICE_FLAG_ENSURE_SEMVER		= (1llu << 21),	/* Since: 1.5.5 */
+	FU_DEVICE_FLAG_ONLY_SUPPORTED		= (1llu << 23),	/* Since: 1.5.5 */
+	FU_DEVICE_FLAG_MD_SET_NAME		= (1llu << 32),	/* Since: 1.5.5 */
+	FU_DEVICE_FLAG_MD_SET_NAME_CATEGORY	= (1llu << 33),	/* Since: 1.5.5 */
+	FU_DEVICE_FLAG_MD_SET_VERFMT		= (1llu << 34),	/* Since: 1.5.5 */
+	FU_DEVICE_FLAG_MD_SET_ICON		= (1llu << 41),	/* Since: 1.5.5 */
+	/*< private >*/
+	FU_DEVICE_FLAG_UNKNOWN			= G_MAXUINT64,
+} FuDeviceFlags;
```
superm1, line 220: while getting this started, is there a better way to make it more gtk-doc friendly?  I notice that the previous device flags take the comment on each line, which is really not useful.

https://fwupd.github.io/fwupd-fwupd-enums.html#FWUPD-DEVICE-FLAG-NO-GUID-MATCHING:CAPS

libfwupdplugin/fu-device.h
```diff
@@ -193,6 +193,35 @@ FuDevice	*fu_device_new				(void);
...
+ * @FU_DEVICE_FLAG_MD_SET_NAME:			Set the device name from the metadata `name` if available
+ * @FU_DEVICE_FLAG_MD_SET_NAME_CATEGORY:	Set the device name from the metadata `category` if available
+ * @FU_DEVICE_FLAG_MD_SET_VERFMT:		Set the device version format from the metadata if available
+ * @FU_DEVICE_FLAG_MD_SET_ICON:			Set the device icon from the metadata if available
+ *
+ * The device internal flags.
+ **/
+typedef enum {
+	FU_DEVICE_FLAG_NONE			= 0,
+	FU_DEVICE_FLAG_NO_AUTO_INSTANCE_IDS	= (1llu << 19),	/* Since: 1.5.5 */
```
superm1, line 211: Presumably you don't want to start at 19..

hughsie: That was actually deliberate, if a plugin author accidentally used the deprecated #define rather than the new one then it all magically just works. The other way would be to check if any of the new FU_DEVICE_FLAGS are >= `1 << 19` for a few months and critical warn if so so that plugin authors know they messed up.

hughsie:
> But then again how soon do you want to break ABI?

RHEL 10 perhaps!? i.e. a long time from now.

libfwupdplugin/fu-device.h
superm1: We could easily have a CI script that greps for stuff like `FWUPD_DEVICE_FLAG_MD_SET_ICON`  in the diff until ABI bump so we don't end up losing 18 more of these by "starting" at 19.

libfwupdplugin/fu-device.h
superm1: actually I think these `enum` will be fine - it's just the `#define` that do poorly.

libfwupdplugin/fu-device.h
hughsie: > We could easily have a CI script that greps for stuff like FWUPD_DEVICE_FLAG_MD_SET_ICON in the diff

This is an excellent idea. I've added this and renumbered the internal flags to start at 0 as expected.

superm1 (approved):

PR #2722, opened 2021-01-04
Allow updates to depend on device flags

hughsie:
This could be used, for instance to require a different device having no
pending updates. Another use case is to only deploy the new firmware onto
devices that have been tagged as supported or even not stuck in bootloader mode.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -205,6 +205,8 @@ fwupd_device_flag_to_string (FwupdDeviceFlags device_flag)
 		return "backup-before-install";
 	if (device_flag == FWUPD_DEVICE_FLAG_RETRY_OPEN)
 		return "retry-open";
+	if (device_flag == FWUPD_DEVICE_FLAG_HAS_UPGRADE)
+		return "has-upgrade";
 	if (device_flag == FWUPD_DEVICE_FLAG_UNKNOWN)
 		return "unknown";
 	return NULL;
@@ -311,6 +313,8 @@ fwupd_device_flag_from_string (const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_BACKUP_BEFORE_INSTALL;
 	if (g_strcmp0 (device_flag, "retry-open") == 0)
 		return FWUPD_DEVICE_FLAG_RETRY_OPEN;
+	if (g_strcmp0 (device_flag, "has-upgrade") == 0)
+		return FWUPD_DEVICE_FLAG_HAS_UPGRADE;
 	return FWUPD_DEVICE_FLAG_UNKNOWN;
 }
 
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -130,6 +130,7 @@ typedef enum {
  * @FWUPD_DEVICE_FLAG_BACKUP_BEFORE_INSTALL:	Device firmware should be saved before installing firmware
  * @FWUPD_DEVICE_FLAG_MD_SET_ICON:		Set the device icon from the metadata if available
  * @FWUPD_DEVICE_FLAG_RETRY_OPEN:		Retry the device open up to 5 times if it fails
+ * @FWUPD_DEVICE_FLAG_HAS_UPGRADE:		Device has an update that is pending
  *
  * The device flags.
  **/
@@ -177,6 +178,7 @@ typedef enum {
 #define FWUPD_DEVICE_FLAG_BACKUP_BEFORE_INSTALL	(1llu << 40)	/* Since: 1.5.0 */
 #define FWUPD_DEVICE_FLAG_MD_SET_ICON		(1llu << 41)	/* Since: 1.5.2 */
 #define FWUPD_DEVICE_FLAG_RETRY_OPEN		(1llu << 42)	/* Since: 1.5.5 */
+#define FWUPD_DEVICE_FLAG_HAS_UPGRADE		(1llu << 43)	/* Since: 1.5.5 */
 #define FWUPD_DEVICE_FLAG_UNKNOWN		G_MAXUINT64	/* Since: 0.7.3 */
 typedef guint64 FwupdDeviceFlags;
 
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1206,6 +1206,113 @@ fu_engine_check_requirement_vendor_id (FuEngine *self, XbNode *req,
 }
 
 static gboolean
+fu_engine_check_requirement_device_flag_str (FuEngine *self, FuDevice *device,
+					     const gchar *flag_str, GError **error)
+{
+	FwupdDeviceFlags flag;
+
+	/* is this a negated device flag */
+	if (g_str_has_prefix (flag_str, "~")) {
+		flag = fwupd_device_flag_from_string (flag_str + 1);
+		if (flag != FWUPD_DEVICE_FLAG_UNKNOWN) {
+			if (!fu_device_has_flag (device, flag))
+				return TRUE;
+			g_set_error (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "device [%s] flag %s is set",
+				     fu_device_get_id (device),
+				     flag_str + 1);
+			return FALSE;
+		}
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "unknown device flag, got %s",
+			     flag_str + 1);
+		return FALSE;
+	}
+
+	/* is this a known device flag */
+	flag = fwupd_device_flag_from_string (flag_str);
+	if (flag != FWUPD_DEVICE_FLAG_UNKNOWN) {
+		if (fu_device_has_flag (device, flag))
+			return TRUE;
+		g_set_error (error,
+			     FWUPD_ERROR,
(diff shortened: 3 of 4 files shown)


superm1:
>This could be used, for instance to require a different device having no pending update

I think this will greatly complicate the state machine and reproducing issues.

>Another use case is to only deploy the new firmware onto
devices that have been tagged as supported or even not stuck in bootloader mode.

Only deploying firmware to supported devices feels like a daemon configuration option.  It would be done for security purposes presumably to only allow firmware from a trusted source.

Devices not stuck in bootloader mode, I think I'd need to better understand the usecase.

hughsie:
> > This could be used, for instance to require a different device having no pending update
> I think this will greatly complicate the state machine and reproducing issues.

Maybe; on the other hand it allows the ODM to specify something more concretely than now -- there are two times in the recent past when I wanted to use this feature, both with the same OEM:

* The ODM has specified to the OEM that the ME update should only ever be deployed when the system firmware has no pending updates; I assume this is to cut down the text matrix when there are discrete EC, ME and BIOS updates for a specific system, some coming from different groups or ODMs entirely. Of course, the correct fix is to make the ME update depend on a specific BIOS update but the BIOS team are working async with the ME team and so the "latest" BIOS version fluctuates depending on if you have testing/stable/embargo configured on your system.

> Only deploying firmware to supported devices feels like a daemon configuration option.

Ya, that's probably a bad example.

> Devices not stuck in bootloader mode, I think I'd need to better understand the usecase.

Also a bad example. My point was more _you can match any device flag_.

The alternative I pondered was for that specific ME scenario:

    <firmware version="latest">1ff60ab2-3905-06a1-b476-0371f00c9e9b</firmware>

..although that seemed a bit too specific and encodes the "magic" version of latest to mean "_no newer releases available_". Other ideas welcome.

superm1:
> The ODM has specified to the OEM that the ME update should only ever be deployed when the system firmware has no pending updates; I assume this is to cut down the text matrix when there are discrete EC, ME and BIOS updates for a specific system, some coming from different groups or ODMs entirely. Of course, the correct fix is to make the ME update depend on a specific BIOS update but the BIOS team are working async with the ME team and so the "latest" BIOS version fluctuates depending on if you have testing/stable/embargo configured on your system.

Can't you accomplish the same thing by setting a version requirement though?  ME version whatever needs to have BIOS version W.X.Y.Z or later first.  Or is the problem actually **installing** the ME update and the BIOS update at the same boot?

superm1:
Also - this feels like an artifact of a process problem.  Maybe the right answer is to just limit the number of capsule updates that can be processed per boot to "1" on this subset  of machines.

hughsie:
>  Maybe the right answer is to just limit the number of capsule updates that can be processed per boot to "1" on this subset of machines

I don't think it's a UpdateCapsule implementation issue; I think it's more of the ODM saying to the OEM "_we only support ME updates when the latest system firmware has been deployed_" -- maybe the proper solution here is to push back and get the two teams to work together to co-ordinate releases.

kmauleon:
**installing** the ME update and the BIOS update at the same boot?

>> this might be it... for **some** reason and **some** cases BIOS and ME update can't be installed at the same time but successfully installed individually.... so i was wondering if it's a good idea to force them to install separately but with highest priority for BIOS/EC

superm1:
> this might be it... for some reason and some cases BIOS and ME update can't be installed at the same time but successfully installed individually.... so i was wondering if it's a good idea to force them to install separately but with highest priority for BIOS/EC

If this is it, I can't possibly understand how this works in Windows.  Windows offers zero controls over this in terms of setting relationships from firmware capsules.  It **should** be failing there too.
But anyway if it is this, then lets instead quirk the subset of affected systems to only allow installing one capsule at a time.

kmauleon:
> > this might be it... for some reason and some cases BIOS and ME update can't be installed at the same time but successfully installed individually.... so i was wondering if it's a good idea to force them to install separately but with highest priority for BIOS/EC
> 
> If this is it, I can't possibly understand how this works in Windows. Windows offers zero controls over this in terms of setting relationships from firmware capsules. It **should** be failing there too.
> But anyway if it is this, then lets instead quirk the subset of affected systems to only allow installing one capsule at a time.

i'm not sure either... firmware in windows and linux are just the same... we just convert it to capsule... if we are going to implement on particular systems only, would it be complicated rather than having it a standard implementation to all systems?
just a thought... but you guys understand it more than i do ^_^

superm1:
> if we are going to implement on particular systems only, would it be complicated rather than having it a standard implementation to all systems?

If there was a spec somewhere saying you should only be applying one capsule at a time, I'd say yes we should change that "default".  But this is the only time I've ever heard of multiple capsules failing to install at the same time, so it doesn't seem fair to other vendors and systems to penalize because of the bugs on one vendor and/or system.  This is why there is a "quirks" system.  Some vendor or system doesn't behave right, so change the behavior just for them.

kmauleon:
> > if we are going to implement on particular systems only, would it be complicated rather than having it a standard implementation to all systems?
> 
> If there was a spec somewhere saying you should only be applying one capsule at a time, I'd say yes we should change that "default". But this is the only time I've ever heard of multiple capsules failing to install at the same time, so it doesn't seem fair to other vendors and systems to penalize because of the bugs on one vendor and/or system. This is why there is a "quirks" system. Some vendor or system doesn't behave right, so change the behavior just for them.

oh sorry ... i thought this would be for lenovo... but i do understand the point.... so i guess a flag would be easier to implement for specific product/system.... thank you very much

hughsie:
If I understand correctly, the only time we schedule multiple updates is on the CLI; in gnome-software we offer the user to reboot straight-away after each update. Naively we could have a quirk that says "requires no pending uefi updates" but then the question then becomes how to detect a scheduled pending update (assuming fwupd exits on idle). Perhaps we can just look for `FU_EFIVAR_GUID_FWUPDATE` with any varname.

We already do a HWID check for Lenovo for the `use-legacy-bootmgr-desc` quirk; I'm not sure how we'd wire the internals to say "this release is not installable as there's already a pending update of a different component". We'd need to know this ahead of time so we can return the list from GetUpdates with the release set to `FWUPD_RELEASE_FLAG_BLOCKED_APPROVAL` or perhaps a new enum altogether. We'd have to choose the correct device (highest urgency value?) to prefer and it all gets a bit gnarly and complicated.

Maybe the simplest way is to return with FWUPD_ERROR_NOT_SUPPORTED and use FWUPD_UPDATE_STATE_FAILED_TRANSIENT if we schedule a update when the device is tagged with custom flag `no-capsule-coalesce` (or something like that) and `FU_EFIVAR_GUID_FWUPDATE-anything` is already set.

superm1:
> If I understand correctly, the only time we schedule multiple updates is on the CLI

I think if we're going to try to stop multiple capsules from installing, we need to stop it from _all_ clients.  Just because of how G-S behaves doesn't mean KDE discover or any other graphical client does it the same.

> how to detect a scheduled pending update

Honestly I thought we did this already.  Isn't this the purpose of the pending database?  And don't we check if you're on the same boot or not?

> prefer and it all gets a bit gnarly and complicated

I would suggest a plugin wide flag to say whether an update is pending.  Once you schedule an update (or start the daemon with a pending update and not rebooted) then you flag it.  And yes if you try to schedule a second one return a transient failure, or something along the lines of needing a reboot.

But before going down this path can we get a **definitive** reproducer and that this is not just a hypothetical process problem?  It sounds like it should be possible to take an affected machine, downgrade ME/BIOS try to run with both capsules.

hughsie:
> we need to stop it from all clients

True, my point was more of "_why don't we see this failure more often_"...

>  Isn't this the purpose of the pending database

I *think* that's only used for the pending stuff that requires activation, e.g. ATA type things.

> can we get a definitive reproducer

@kmauleon should I be able to reproduce this on my X1 7th Lenovo sent me? Or my P50?

hughsie:
Something like the 100% untested wip/hughsie/no-capsule-coalesce branch perhaps?

superm1:
> Something like the 100% untested wip/hughsie/no-capsule-coalesce branch perhaps?

Generally that direction seems sensible.  I would say though - a few comments:
*  you need to set the pending flag also after an update is scheduled, not just at coldplug
* other plugins that schedule for reboot like ATA might use use this too.  As an idea, any device that has needs reboot set could set it.

hughsie:
> you need to set the pending flag also after an update is scheduled, not just at coldplug

I think an inotify watch covers all bases and is the least likely to be forgotten. It also means it reflects immediately when using `efibootmgr`! :)

> other plugins that schedule for reboot like ATA might use use this too

Good idea, I can handle this in fu_engine_update_cleanup()

kmauleon:
> > we need to stop it from all clients
> 
> True, my point was more of "_why don't we see this failure more often_"...
> 
> > Isn't this the purpose of the pending database
> 
> I _think_ that's only used for the pending stuff that requires activation, e.g. ATA type things.
> 
> > can we get a definitive reproducer
> 
> @kmauleon should I be able to reproduce this on my X1 7th Lenovo sent me? Or my P50?

hi @hughsie you might be able to with X1 7th but not 100%.  I did encounter system hang up during multiple updates but successfully installed if individually updated....but this was intermittent during my test and when I brought it up to BIOS dev... they said they can't reproduce it and maybe just need to update firmware separately.... specifically with ME... sorry I burden you with this... if I could just at least do something at my end.... maybe a tag in the metainfo or something... please advise... thank you very much for your help and support

PR #2724, opened 2021-01-05
uefi: Rename to uefi-capsule

hughsie:
Rename the plugin to make it clearer of the scope.

Based on a patch from Mario Limonciello <mario.limonciello@dell.com>


hughsie:
@superm1 I think you need to add some kind of debian config file moved dh thing from memory...

hughsie:
I've fixed the conflict and force pushed. Modulo the snap failure (looks like gnutls can't be downloaded) looks good to you?

superm1 (approved):
yup looks good to me.

PR #2732, opened 2021-01-05
Correct some logic errors in daemon with handling of thunderbolt plugin

superm1:
The main issue is purely cosmetic, but it leads to erroneous failure reports.
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

CC @YehezkelShB 
@djcampello: some of this is ChromeOS specific fixes I found while fixing the generic issue that was flagging reports.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2237,7 +2237,7 @@ fu_engine_install_release (FuEngine *self,
 	if (version_rel != NULL &&
 	    fu_common_vercmp_full (version_orig, version_rel, fmt) != 0 &&
 	    fu_common_vercmp_full (version_orig, fu_device_get_version (device), fmt) == 0 &&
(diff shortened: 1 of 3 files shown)


src/fu-engine.c
```diff
@@ -2237,7 +2237,7 @@ fu_engine_install_release (FuEngine *self,
 	if (version_rel != NULL &&
 	    fu_common_vercmp_full (version_orig, version_rel, fmt) != 0 &&
 	    fu_common_vercmp_full (version_orig, fu_device_get_version (device), fmt) == 0 &&
-	    !fu_device_has_flag (device, FWUPD_DEVICE_FLAG_SKIPS_RESTART)) {
+	    !fu_device_has_flag (device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
```
hughsie, line 2240: all makes sense to me, but I'd love some kind of self tests for the delayed activation case; this is probably zccidentaly fixing ATA updates too..

PR #2734, opened 2021-01-06
Detect the AMD TSME encryption state for HSI

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -2236,26 +2236,49 @@ fu_common_cpuid (guint32 leaf,
  * Uses CPUID to discover the CPU vendor and check if it is Intel.
  *
  * Return value: %TRUE if the vendor was Intel.
+ * Deprecated: 1.5.5: Use fu_common_get_cpu_vendor() instead.
  *
  * Since: 1.5.0
  **/
 gboolean
 fu_common_is_cpu_intel (void)
 {
+	return fu_common_get_cpu_vendor () == FU_CPU_VENDOR_INTEL;
+}
+
+/**
+ * fu_common_get_cpu_vendor:
+ *
+ * Uses CPUID to discover the CPU vendor.
+ *
+ * Return value: a #FuCpuVendor, e.g. %FU_CPU_VENDOR_AMD if the vendor was AMD.
+ *
+ * Since: 1.5.5
+ **/
+FuCpuVendor
+fu_common_get_cpu_vendor (void)
+{
+#ifdef HAVE_CPUID_H
 	guint ebx = 0;
 	guint ecx = 0;
 	guint edx = 0;
 
-	if (!fu_common_cpuid (0x0, NULL, &ebx, &ecx, &edx, NULL))
-		return FALSE;
-#ifdef HAVE_CPUID_H
-	if (ebx == signature_INTEL_ebx &&
-	    edx == signature_INTEL_edx &&
-	    ecx == signature_INTEL_ecx) {
-		return TRUE;
+	if (fu_common_cpuid (0x0, NULL, &ebx, &ecx, &edx, NULL)) {
+		if (ebx == signature_INTEL_ebx &&
+		    edx == signature_INTEL_edx &&
+		    ecx == signature_INTEL_ecx) {
+			return FU_CPU_VENDOR_INTEL;
+		}
+		if (ebx == signature_AMD_ebx &&
+		    edx == signature_AMD_edx &&
+		    ecx == signature_AMD_ecx) {
+			return FU_CPU_VENDOR_AMD;
+		}
 	}
 #endif
-	return FALSE;
+
+	/* failed */
+	return FU_CPU_VENDOR_UNKNOWN;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -2258,6 +2258,34 @@ fu_common_is_cpu_intel (void)
...
+ * fu_common_is_cpu_amd:
+ *
+ * Uses CPUID to discover the CPU vendor and check if it is AMD.
+ *
+ * Return value: %TRUE if the vendor was AMD.
+ *
+ * Since: 1.5.5
+ **/
+gboolean
+fu_common_is_cpu_amd (void)
```
hughsie, line 2271: of course, this implies adding fu_common_is_cpu_riscv and fu_common_is_cpu_arm in the future. Maybe something like `fu_common_get_cpu_vendor()` might be better, with a `FuCpuVendor` enum with `FU_CPU_VENDOR_AMD` etc.

superm1: I think that approach is better, a lot less symbols to maintain for ABI.

plugins/msr/fu-plugin-msr.c
```diff
@@ -185,9 +223,41 @@ fu_plugin_add_security_attr_dci_locked (FuPlugin *plugin, FuSecurityAttrs *attrs
 	fwupd_security_attr_set_result (attr, FWUPD_SECURITY_ATTR_RESULT_LOCKED);
 }
 
+static void
+fu_plugin_add_security_attr_tsme_enabled (FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+	FuPluginData *priv = fu_plugin_get_data (plugin);
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+
+	/* this MSR is only valid for a subset of Intel CPUs */
```
superm1, line 232: AMD

libfwupdplugin/fu-common.c
hughsie: Agree; fixed.

plugins/msr/fu-plugin-msr.c
hughsie: Ta, fixed.

libfwupdplugin/fu-common.c
```diff
@@ -2236,26 +2236,49 @@ fu_common_cpuid (guint32 leaf,
...
+{
+	return fu_common_get_cpu_vendor () == FU_CPU_VENDOR_INTEL;
+}
+
+/**
+ * fu_common_get_cpu_vendor:
+ *
+ * Uses CPUID to discover the CPU vendor.
+ *
+ * Return value: %TRUE if the vendor was AMD.
```
superm1, line 2254: wrong comment now

libfwupdplugin/fu-common.c
```diff
@@ -2236,26 +2236,49 @@ fu_common_cpuid (guint32 leaf,
...
+FuCpuVendor
+fu_common_get_cpu_vendor (void)
 {
 	guint ebx = 0;
 	guint ecx = 0;
 	guint edx = 0;
 
-	if (!fu_common_cpuid (0x0, NULL, &ebx, &ecx, &edx, NULL))
-		return FALSE;
+	if (fu_common_cpuid (0x0, NULL, &ebx, &ecx, &edx, NULL)) {
```
superm1, line 2265: in the case of no `#ifdef` does the compiler let you do this?  I would think you need to have something in the clause.  IE switch it to 
```
if (!fu...)
   return FU_CPU_VENDOR_UNKNOWN;
#ifdef ...
//do stuff
#endif 
return  FU_CPU_VENDOR_UNKNOWN;
```

plugins/msr/fu-plugin-msr.c
```diff
@@ -41,12 +53,23 @@ gboolean
...
 {
 	FuPluginData *priv = fu_plugin_get_data (plugin);
+	guint eax = 0;
 	guint ecx = 0;
 
 	/* sdbg is supported: https://en.wikipedia.org/wiki/CPUID */
-	if (!fu_common_cpuid (0x01, NULL, NULL, &ecx, NULL, error))
-		return FALSE;
-	priv->ia32_debug_supported = ((ecx >> 11) & 0x1) > 0;
+	if (fu_common_get_cpu_vendor () != FU_CPU_VENDOR_INTEL) {
```
superm1, line 60: wait, don't you want to be checking **for** Intel?

hughsie: Indeed, doh.

libfwupdplugin/fu-common.c
hughsie: The compiler seems happy enough with the empy if clause, although I agree it's a bit wacky. I'll just move the entire function minus the last lines into the ifdef.

libfwupdplugin/fu-common.c
hughsie: Fixed.

superm1 (approved):

superm1:
Does HSI documentation need something added for AMD's TSME in the mentions?

hughsie:
> Does HSI documentation need something added for AMD's TSME in the mentions?

Yup! Will fix and merge, thanks.

PR #2735, opened 2021-01-06
Do not allow Lenovo hardware to install multiple capsules

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -232,6 +232,20 @@ fu_engine_progress_notify_cb (FuDevice *device, GParamSpec *pspec, FuEngine *sel
 }
 
 static void
+fu_engine_update_state_notify_cb (FuDevice *device, GParamSpec *pspec, FuEngine *self)
+{
+	/* update the plugin state for any device needing activation (like ATA) */
+	if (fu_device_get_update_state (device) == FWUPD_UPDATE_STATE_PENDING) {
+		FuPlugin *plugin;
+		plugin = fu_plugin_list_find_by_name (self->plugin_list,
+						      fu_device_get_plugin (device),
+						      NULL);
+		if (plugin != NULL)
+			fu_plugin_add_flag (plugin, FWUPD_PLUGIN_FLAG_PENDING_UPDATE);
+	}
+}
+
+static void
 fu_engine_status_notify_cb (FuDevice *device, GParamSpec *pspec, FuEngine *self)
 {
 	fu_engine_set_status (self, fu_device_get_status (device));
@@ -249,11 +263,16 @@ fu_engine_watch_device (FuEngine *self, FuDevice *device)
 		g_signal_handlers_disconnect_by_func (device_old,
 						      fu_engine_status_notify_cb,
 						      self);
+		g_signal_handlers_disconnect_by_func (device_old,
+						      fu_engine_update_state_notify_cb,
+						      self);
 	}
 	g_signal_connect (device, "notify::progress",
 			  G_CALLBACK (fu_engine_progress_notify_cb), self);
 	g_signal_connect (device, "notify::status",
 			  G_CALLBACK (fu_engine_status_notify_cb), self);
+	g_signal_connect (device, "notify::update-state",
+			  G_CALLBACK (fu_engine_update_state_notify_cb), self);
 }
 
 static void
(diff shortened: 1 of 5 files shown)


superm1 (approved):
The PR looks good for the Lenovo case.  I think you want to set `FWUPD_UPDATE_STATE_PENDING` for the stuff pending activation too though in the engine.

hughsie:
Something like this?

    -       if (fu_device_get_update_state (device) == FWUPD_UPDATE_STATE_PENDING) {
    +       if (fu_device_get_update_state (device) == FWUPD_UPDATE_STATE_PENDING ||
    +           fu_device_has_flag (device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {

superm1:
I was thinking directly in the plugins rather than the engine initially, but yeah I suppose that will work.

superm1:
Something else I realize needed now - if activate is called, those flags need to be cleared still.

hughsie:
> if activate is called, those flags need to be cleared still

Not so convinced, and it's a bit of long journey explaining why...

Rather than setting the plugin flag in `fu_engine_update_cleanup()` you could do:

    +       g_signal_connect (device, "notify::flags",
    +                         G_CALLBACK (fu_engine_flags_notify_cb), self);
    +       g_signal_connect (device, "notify::update-state",
    +                         G_CALLBACK (fu_engine_flags_notify_cb), self);

...and then check for the `FWUPD_UPDATE_STATE_PENDING` state or `FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION` flags there. That's a bit less magic than just at the end of update function and is probably a good move.

The next logical jump is to say "we could clear the flag if those device flags are not set" but you of course need to query all the active FuDevice objects for that specific plugin (e.g. `fu_device_list_get_active()`, then filter by plugin) and then clear the flag if none of them require activation or are pending. That would work.

But! Clearing the flag there undoes the work when we've restarted fwupd after deploying a UEFI update -- the plugin sets the flag when `FU_EFIVAR_GUID_EFI_GLOBAL:BootNext` and `FU_EFIVAR_GUID_FWUPDATE-*` exist -- none of the devices are set to `FWUPD_UPDATE_STATE_PENDING` as you might expect.

So, we can either refcount the flag as two things are setting it (ewww!) or we can untangle the mess a little. Or we can just avoid unsetting the flag on activation; which seems to be a _nice to have_ thing as activation is typically done just before we shutdown and then everything is killed anyway. Or, we go back a bit and define `FWUPD_PLUGIN_FLAG_PENDING_UPDATE` as nothing to do with activation and then this isn't a problem again. Other ideas welcome.

libfwupd/fwupd-device.c
```diff
@@ -74,6 +74,7 @@ enum {
 	PROP_PROTOCOL,
 	PROP_STATUS,
 	PROP_PARENT,
+	PROP_UPDATE_STATE,
```
superm1, line 77: With this type of change I almost need to wonder if we really need to be using the device flag for needs activation.  Should we deprecate it and become an update state?

src/fu-engine.c
```diff
@@ -2696,6 +2715,9 @@ fu_engine_activate (FuEngine *self, const gchar *device_id, GError **error)
 	if (!fu_plugin_runner_activate (plugin, device, error))
 		return FALSE;
 
+	/* unconditionally do this on success */
+	fu_device_remove_flag (device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION);
```
superm1, line 2719: I don't think this is correct, this is already done in `fu_plugin_runner_activate`

superm1:
>  typically done just before we shutdown and then everything is killed anyway

**Typically** is _relative_. It's also done at unplug for Dell's WD19 as well and logout for ChromeOS.

src/fu-engine.c
hughsie: Agree, dropped, thanks.

hughsie:
> do we even need a "public" daemon plugin flag in the first place

I think this is the right question, and I think the answer is no. New PRs coming up, thanks for the sanity check yet again.

hughsie:
I think https://github.com/fwupd/fwupd/pull/2757 is a much better solution.

PR #2736, opened 2021-01-07
Wip/hughsie/proxy mutex

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -56,7 +56,7 @@ typedef struct {
 	gchar				*host_product;
 	gchar				*host_machine_id;
 	gchar				*host_security_id;
-	GDBusConnection			*conn;
+	GMutex				 proxy_mutex;	/* for @proxy */
 	GDBusProxy			*proxy;
 	gchar				*user_agent;
 #ifdef SOUP_SESSION_COMPAT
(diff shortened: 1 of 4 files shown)


libfwupd/fwupd-client.c
```diff
@@ -535,7 +541,9 @@ fwupd_client_connect_async (FwupdClient *self, GCancellable *cancellable,
 	g_return_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable));
 
 	/* nothing to do */
+	g_mutex_lock (&priv->proxy_mutex);
```
pwithnall, line 544: Holding the lock until the end of `fwupd_client_connect_get_proxy_cb()` means that any other pending calls to `fwupd_client_connect_async()` will block on this line. i.e. The `*_async()` call will become blocking.

I think you probably want to unlock the mutex at end of this function, and lock it again in `fwupd_client_connect_get_bus_cb()` and `fwupd_client_connect_get_proxy_cb()` when you access the shared state in either of them. You’d then have to deal with what might have happened if another thread executed their `fwupd_client_connect_get_bus_cb()`/`fwupd_client_connect_get_proxy_cb()` callback first. i.e. `priv->conn` is unexpectedly not `NULL`, or `priv->proxy` is unexpectedly not `NULL`. In both cases it should be OK to free and drop the newly-returned result from `g_bus_get_finish()` and `g_dbus_proxy_new_finish()` and move on to the next step in the process.

hughsie: That makes sense; will give it a go, thanks!

hughsie: @pwithnall something like this?

libfwupd/fwupd-client.c
```diff
@@ -449,6 +449,13 @@ fwupd_client_connect_get_proxy_cb (GObject *source,
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GVariant) val = NULL;
 	g_autoptr(GVariant) val2 = NULL;
+	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->proxy_mutex);
+
+	/* another thread did this for us */
+	if (priv->proxy != NULL) {
+		g_task_return_boolean (task, TRUE);
```
pwithnall, line 456: You need to call `g_dbus_proxy_new_finish()` (and then drop the result) or the `g_dbus_proxy_new_async()` call will go unpaired.

Could do something like:
```
g_autoptr(GDBusProxy) proxy = NULL;
g_autoptr(GMutexLocker) locker = NULL;

proxy = g_dbus_proxy_new_finish (res, &error);

locker = g_mutex_locker_new (&priv->proxy_mutex);
if (priv->proxy != NULL)
  {
    g_task_return_boolean (task, TRUE);
    return;
  }
priv->proxy = g_steal_pointer (&proxy);

/* connect signals, etc. */
```

pwithnall (commented):
Nice simplification with `g_dbus_proxy_new_for_bus()`

libfwupd/fwupd-client.c
hughsie: > You need to call g_dbus_proxy_new_finish

Doh, of course. I've used your suggestion.

pwithnall:
lgtm now, fwiw

hughsie:
@superm1 don't delete the branch yet pls; someone else is testing with it and it would be confusing to explain to cherry pick from master.

PR #2738, opened 2021-01-07
Make building the plugins shared objects optional

hughsie:
When building the static fuzz targets we cannot link shared objects. We also
can make the list of built targets a lot smaller!

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -164,12 +164,19 @@ add_project_arguments('-D_GNU_SOURCE', language : 'c')
 if get_option('build') == 'all'
   build_standalone = true
   build_daemon = true
+  build_plugins = true
 elif get_option('build') == 'standalone'
   build_standalone = true
   build_daemon = false
+  build_plugins = true
 elif get_option('build') == 'library'
   build_standalone = false
   build_daemon = false
+  build_plugins = false
+elif get_option('build') == 'fuzzing'
+  build_standalone = true
+  build_daemon = false
+  build_plugins = false
 endif
 
 prefix = get_option('prefix')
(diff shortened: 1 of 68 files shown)


superm1 (commented):
can we have a CI target that does a build like this in "fuzzing" to make sure we don't regress the `meson.build` in the future?

hughsie:
> can we have a CI target that does a build like this in "fuzzing" to make sure we don't regress the meson.build in the future

Yes; this is what I'm planning to add for oss-fuzz; it's Ubuntu 16.04 and so it'll be a "new" target.

superm1:
> > can we have a CI target that does a build like this in "fuzzing" to make sure we don't regress the meson.build in the future
> 
> Yes; this is what I'm planning to add for oss-fuzz; it's Ubuntu 16.04 and so it'll be a "new" target.

Sounds good

superm1 (approved):
I think this PR is fine on it's own, but I would question don't you lose a lot of the value on the fuzzing if you're not actually fuzzing the plugins?  Wouldn't it be better to compile the plugins statically into the binary at that point?

hughsie:
> don't you lose a lot of the value on the fuzzing if you're not actually fuzzing the plugins

That's to come; e.g. lots of tiny static binaries each testing one firmware GType. At least that's the plan. Don't merge this yet and I'll try to get an example.

superm1:
> > don't you lose a lot of the value on the fuzzing if you're not actually fuzzing the plugins
> 
> That's to come; e.g. lots of tiny static binaries each testing one firmware GType. At least that's the plan. Don't merge this yet and I'll try to get an example.

But don't you want to fuzz actual plugin code too?  I guess that's a lot harder until we have something like https://github.com/martinpitt/umockdev/issues/82 to pretend to replay device interactions.

superm1:
>  lots of tiny static binaries each testing one firmware GType

Those are linked into the plugins though.  Which I guess comes back to my point - if you instead statically link all the plugins in, you can have a daemon that doesn't load modules and you should get all the firmware types included and plugin code included to allow more interesting fuzzing.

PR #2739, opened 2021-01-08
Wip/hughsie/g main context push thread default take 2

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -451,6 +451,10 @@ fwupd_client_connect_get_proxy_cb (GObject *source,
 	g_autoptr(GVariant) val = NULL;
 	g_autoptr(GVariant) val2 = NULL;
 	g_autoptr(GVariant) val3 = NULL;
+	g_autoptr(GVariant) val4 = NULL;
+	g_autoptr(GVariant) val5 = NULL;
+	g_autoptr(GVariant) val6 = NULL;
+	g_autoptr(GVariant) val7 = NULL;
 	g_autoptr(GMutexLocker) locker = NULL;
 
 	proxy = g_dbus_proxy_new_finish (res, &error);
@@ -481,18 +485,18 @@ fwupd_client_connect_get_proxy_cb (GObject *source,
 	val3 = g_dbus_proxy_get_cached_property (priv->proxy, "Status");
 	if (val3 != NULL)
 		fwupd_client_set_status (self, g_variant_get_uint32 (val3));
-	val2 = g_dbus_proxy_get_cached_property (priv->proxy, "Interactive");
-	if (val2 != NULL)
-		priv->interactive = g_variant_get_boolean (val2);
-	val = g_dbus_proxy_get_cached_property (priv->proxy, "HostProduct");
-	if (val != NULL)
-		fwupd_client_set_host_product (self, g_variant_get_string (val, NULL));
-	val = g_dbus_proxy_get_cached_property (priv->proxy, "HostMachineId");
-	if (val != NULL)
-		fwupd_client_set_host_machine_id (self, g_variant_get_string (val, NULL));
-	val = g_dbus_proxy_get_cached_property (priv->proxy, "HostSecurityId");
-	if (val != NULL)
-		fwupd_client_set_host_security_id (self, g_variant_get_string (val, NULL));
+	val4 = g_dbus_proxy_get_cached_property (priv->proxy, "Interactive");
+	if (val4 != NULL)
+		priv->interactive = g_variant_get_boolean (val4);
+	val5 = g_dbus_proxy_get_cached_property (priv->proxy, "HostProduct");
+	if (val5 != NULL)
+		fwupd_client_set_host_product (self, g_variant_get_string (val5, NULL));
+	val6 = g_dbus_proxy_get_cached_property (priv->proxy, "HostMachineId");
+	if (val6 != NULL)
+		fwupd_client_set_host_machine_id (self, g_variant_get_string (val6, NULL));
+	val7 = g_dbus_proxy_get_cached_property (priv->proxy, "HostSecurityId");
+	if (val7 != NULL)
+		fwupd_client_set_host_security_id (self, g_variant_get_string (val7, NULL));
 
 	/* success */
 	g_task_return_boolean (task, TRUE);
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -450,6 +450,7 @@ fwupd_client_connect_get_proxy_cb (GObject *source,
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GVariant) val = NULL;
 	g_autoptr(GVariant) val2 = NULL;
+	g_autoptr(GVariant) val3 = NULL;
 	g_autoptr(GMutexLocker) locker = NULL;
 
 	proxy = g_dbus_proxy_new_finish (res, &error);
@@ -477,6 +478,9 @@ fwupd_client_connect_get_proxy_cb (GObject *source,
 	val2 = g_dbus_proxy_get_cached_property (priv->proxy, "Tainted");
 	if (val2 != NULL)
 		priv->tainted = g_variant_get_boolean (val2);
+	val3 = g_dbus_proxy_get_cached_property (priv->proxy, "Status");
+	if (val3 != NULL)
+		fwupd_client_set_status (self, g_variant_get_uint32 (val3));
 	val2 = g_dbus_proxy_get_cached_property (priv->proxy, "Interactive");
 	if (val2 != NULL)
 		priv->interactive = g_variant_get_boolean (val2);
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -52,6 +52,9 @@ typedef struct {
 	gboolean			 tainted;
 	gboolean			 interactive;
 	guint				 percentage;
+	GMutex				 idle_mutex;	/* for @idle_id and @idle_sources */
+	guint				 idle_id;
+	GPtrArray			*idle_sources;	/* element-type FwupdClientContextHelper */
 	gchar				*daemon_version;
 	gchar				*host_product;
 	gchar				*host_machine_id;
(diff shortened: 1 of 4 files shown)


hughsie:
@pwithnall this is right up your street; thanks again!

hughsie:
Seems to fix the reported hang here: https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1113#note_1000844 -- When Adam is awake I'll get him to try with those openQA instances.

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,116 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+	gchar		*property_name;
+	guint		 signal_id;
+	FwupdDevice	*device;
+} FwupdClientContextHelper;
+
+static void
+fwupd_client_context_helper_free (FwupdClientContextHelper *helper)
+{
+	if (helper->device != NULL)
+		g_object_unref (helper->device);
```
pwithnall, line 137: Could simplify this with `g_clear_object (&helper->device)`

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,116 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+	FwupdClientPrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GMutexLocker) locker = NULL;
+
+	/* property */
+	if (helper->property_name != NULL)
+		g_object_notify (G_OBJECT (self), helper->property_name);
+
+	/* legacy signal name */
+	if (g_strcmp0 (helper->property_name, "status") == 0)
+		g_signal_emit (self, signals[SIGNAL_STATUS_CHANGED], 0, priv->status);
```
pwithnall, line 157: Is reading `priv->status` here thread-safe?

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,116 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->idle_id_mutex);
+	g_autoptr(GSource) source = g_idle_source_new ();
+
+	/* run in the correct GMainContext and thread */
+	g_source_set_callback (source,
+			       fwupd_client_context_idle_cb,
+			       helper,
+			       (GDestroyNotify) fwupd_client_context_helper_free);
+	if (priv->idle_id != 0)
+		g_source_remove (priv->idle_id);
```
pwithnall, line 182: This will cause the previously-pending signal emission to never happen, as the state from the `helper` object attached to the source being removed here is not stored anywhere else.

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,116 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+fwupd_client_object_notify (FwupdClient *self, const gchar *property_name)
+{
+	FwupdClientPrivate *priv = GET_PRIVATE (self);
+	FwupdClientContextHelper *helper = NULL;
+
+	/* shortcut */
+	if (priv->main_ctx == g_main_context_get_thread_default ()) {
+		g_object_notify (G_OBJECT (self), property_name);
+		if (g_strcmp0 (property_name, "status") == 0)
+			g_signal_emit (self, signals[SIGNAL_STATUS_CHANGED], 0, priv->status);
```
pwithnall, line 197: Duplicating the code from `fwupd_client_context_idle_cb()` here means it’s possibly going to end up going out of sync.

libfwupd/fwupd-client.c
```diff
@@ -305,19 +409,21 @@ fwupd_client_get_main_context (FwupdClient *self)
 /**
  * fwupd_client_set_main_context:
  * @self: A #FwupdClient
- * @main_ctx: #GMainContext
+ * @main_ctx: (nullable): #GMainContext or %NULL
```
pwithnall, line 412: Would be good to clarify what `NULL` means.

libfwupd/fwupd-client.c
```diff
@@ -491,6 +597,7 @@ fwupd_client_connect_get_proxy_cb (GObject *source,
 		fwupd_client_set_host_security_id (self, g_variant_get_string (val, NULL));
 
 	/* success */
+	fwupd_client_set_status (self, FWUPD_STATUS_IDLE);
```
pwithnall, line 600: This change looks unrelated

libfwupd/fwupd-client.c
```diff
@@ -2112,6 +2219,10 @@ fwupd_client_install_stream_async (FwupdClient *self,
  *
  * Install firmware onto a specific device.
  *
+ * NOTE: This method is thread-safe, but progress signals will be
+ * emitted in the default main context, if not explicitly set with
```
pwithnall, line 2223: Would be good to clarify what you mean by ‘default main context’. Do you mean the global default main context (`g_main_context_default()`), the thread default main context (`g_main_context_get_thread_default()`) or something else specific to `FwupdClient`?

libfwupd/fwupd-context-test.c
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright (C) 2020 Philip Withnall <philip@tecnocode.co.uk>
```
pwithnall, line 2:
```suggestion
 * Copyright (C) 2020 Philip Withnall <pwithnall@endlessos.org>
```

(I did it on work time)

libfwupd/fwupd-client.c
hughsie: Fixed, ta.

libfwupd/fwupd-client.c
hughsie: Do you think it should be stored in the helper?

libfwupd/fwupd-client.c
hughsie: That was by design; is that right? My logic was that if nothing was "servicing" the GMainContext then it's pointless just queing them all up for when the method returns.

libfwupd/fwupd-client.c
hughsie: Maybe we add something like fwupd_client_set_status_safe() and do it there?

libfwupd/fwupd-client.c
hughsie: It's actually completely related; it causes a property change event which I'm using in the context test to check the GSource is working as we expect. I guess the proper fix is to do something like:

    val = g_dbus_proxy_get_cached_property (proxy, "Status");
    fwupd_client_set_status (self, g_variant_get_uint32 (val));

as I think that's set after the proxy is set up.

libfwupd/fwupd-client.c
hughsie: > Do you mean the global default main context

Exactly that, will change to that wording.

libfwupd/fwupd-client.c
hughsie: Fixed, ta.

libfwupd/fwupd-client.c
hughsie: Yes, this works -- Status was missing completely!

libfwupd/fwupd-client.c
hughsie: This is what I did; better function name welcome.

libfwupd/fwupd-client.c
hughsie: The status is in reference to the daemon and shared state, rather than the transaction. All threads should have the same value and I think using a mutex for this legacy thing isn't required.

libfwupd/fwupd-client.c
pwithnall: I don’t think it’s right. It’s quite plausible for the `FwupdClient` thread to make two `notify` emissions in quick succession, fast enough that the main thread might not be able to process the first one before it’s removed and replaced with the second one.

If you want to do this properly, I’d suggest keeping a `GPtrArray` of the `GSource`s which are pending. Just store the pointers to the `GSource`s, don’t bother with the source IDs (they’re equivalent).

libfwupd/fwupd-client.c
pwithnall: Perhaps it should be a separate commit, with some explanation along the lines of what you’ve just said in the commit message.

libfwupd/fwupd-client.c
pwithnall: “All threads should have the same value” :rotating_light: Citation needed! :rotating_light: 

Storing it in the helper (so the value which is emitted in the signal is the value from when the signal was scheduled to be emitted) makes sense and should be simpler than reasoning about the thread-safety of `priv->status`.

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,123 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+static void
+fwupd_client_context_helper_free (FwupdClientContextHelper *helper)
+{
+	g_clear_object (&helper->device);
+	g_object_unref (helper->self);
+	g_free (helper->property_name);
+	g_free (helper);
+}
+
+/* main_ctx == g_main_context_get_thread_default() */
```
pwithnall, line 142: You could encode that as an assertion, for added safety:
```
g_return_if_fail (g_main_context_is_owner (priv->main_ctx));
```

Might be clearer to phrase the comment as `/* Always executed in the main context given by priv->main_ctx */`

libfwupd/fwupd-client.c
hughsie: I did exactly that :)

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,123 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+
+/* run callback in the correct thread */
+static void
+fwupd_client_object_notify (FwupdClient *self, const gchar *property_name)
+{
+	FwupdClientPrivate *priv = GET_PRIVATE (self);
+	FwupdClientContextHelper *helper = NULL;
+
+	/* shortcut */
+	if (priv->main_ctx == g_main_context_get_thread_default ()) {
```
pwithnall, line 203:
```suggestion
	if (g_main_context_is_owner (priv->main_ctx)) {
```
(See what `g_main_context_invoke()` does)

This is because of the difference between “what’s the default `GMainContext` in this thread at the moment” and “which `GMainContext` is currently being iterated in this thread”. One `GMainContext` might be the thread-default, but this function might actually be being called as part of a call chain from a `GSource` attached to a different (nested) `GMainContext` which is the one actually being iterated.

Put another way, one can think of `g_main_context_get_thread_default()` as being useful for deciding where to schedule something; and one can think of `g_main_context_is_owner()` as being useful for deciding where the currently-running code has been scheduled.

libfwupd/fwupd-client.c
```diff
@@ -121,13 +123,123 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+
+/* run callback in the correct thread */
+static void
+fwupd_client_signal_emit_device (FwupdClient *self, guint signal_id, FwupdDevice *device)
+{
+	FwupdClientPrivate *priv = GET_PRIVATE (self);
+	FwupdClientContextHelper *helper = NULL;
+
+	/* shortcut */
+	if (priv->main_ctx == g_main_context_get_thread_default ()) {
```
pwithnall, line 223: Same as the comment above.

libfwupd/fwupd-client.c
pwithnall: Sorry, I think I was reviewing a stale version of the branch at that point! :partying_face:

libfwupd/fwupd-client.c
hughsie: > keeping a GPtrArray of the GSources which are pending

How would I remove the GSource after the idle source has run? Surely a better way would be to store the FwupdClientContextHelper's in an array as then you have something in fwupd_client_context_idle_cb() to remove. i.e. There's only one idle_id like now, but when fwupd_client_context_idle_cb is called we might emit multiple signals...

pwithnall: Sure, that would work better

libfwupd/fwupd-client.c
hughsie: Agree, fixed, thanks.

libfwupd/fwupd-client.c
hughsie: TIL! Thanks, fixed.

libfwupd/fwupd-client.c
hughsie: Fixed.

libfwupd/fwupd-client.c
hughsie: > Sure, that would work better

**Narrator:** It did!

libfwupd/fwupd-client.c
pwithnall: Are you not going to store `priv->status` in the helper?

libfwupd/fwupd-client.c
```diff
@@ -121,13 +124,127 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
...
+fwupd_client_context_helper (FwupdClient *self, FwupdClientContextHelper *helper)
+{
+	FwupdClientPrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->idle_mutex);
+	g_autoptr(GSource) source = g_idle_source_new ();
+
+	/* run in the correct GMainContext and thread */
+	g_source_set_callback (source, fwupd_client_context_idle_cb, self, NULL);
+	if (priv->idle_id != 0)
+		g_source_remove (priv->idle_id);
```
pwithnall, line 195: If there’s already a source attached to the context (i.e. `priv->idle_id != 0`) then you could just return early from this function, don’t remove the existing source, and don’t add a new one. Would save a bit of locking churn on that `GMainContext`

hughsie: Agree. Something like this?

	FwupdClientPrivate *priv = GET_PRIVATE (self);
	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->idle_mutex);

	/* already a source attached to the context */
	if (priv->idle_id == 0) {
		g_autoptr(GSource) source = g_idle_source_new ();
		g_source_set_callback (source, fwupd_client_context_idle_cb, self, NULL);
		priv->idle_id = g_source_attach (g_steal_pointer (&source), priv->main_ctx);
	}

	/* run in the correct GMainContext and thread */
	g_ptr_array_add (priv->idle_sources, helper);

pwithnall: Yes, although I think the `already a source attached to the context` comment needs inverting (a source isn’t attached if that branch is taken)

hughsie: Ohh yes. indeed. thanks!

pwithnall (approved):
lgtm

Hopefully that’s the last of the fwupd main context issues and solves things for gnome-software. Thanks for taking the time to dig through it all :)

hughsie:
> Thanks for taking the time to dig through it all :)

No, quite the opposite! Thank you for explaining it to me!

Issue #2723, opened 2021-01-05
Need to check system clock before refreshing


hughsie:
An ODM got this:

    [root@localhost Desktop]# fwupdmgr refresh
    Fetching metadata https://fwupd.org/downloads/firmware-b3d1e87f047072094d6d5d46763941e2213819a9.xml.gz
    Failed to download https://fwupd.org/downloads/firmware-b3d1e87f047072094d6d5d46763941e2213819a9.xml.gz: SSL handshake failed

Using curl -v:

    curl: (60) SSL certificate problem: certificate is not yet valid

...the system date incorrectly, possibly even 1970-01-01. My suggestion is we check the system date is newer than the last build date of fwupd and otherwise warn the user with something like:

    [root@localhost Desktop]# fwupdmgr refresh
    Error: System clock is set incorrectly, please check

PR #2742, opened 2021-01-10
Show a console warning if the system clock is not set

hughsie:
If the date is wrong, the SSL certificate may not be in a valid date range and
the user gets a rather unhelpful 'SSL handshake failed' error message.

Detect the most common case and show a warning when starting fwupdmgr.

Fixes https://github.com/fwupd/fwupd/issues/2723

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -2702,6 +2702,7 @@ main (int argc, char *argv[])
 	gboolean verbose = FALSE;
 	gboolean version = FALSE;
 	g_autoptr(FuUtilPrivate) priv = g_new0 (FuUtilPrivate, 1);
+	g_autoptr(GDateTime) dt_now = g_date_time_new_now_utc ();
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GError) error_polkit = NULL;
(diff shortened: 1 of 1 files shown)


hughsie:
@superm1 better wording very welcome.

src/fu-util.c
```diff
@@ -3045,6 +3046,17 @@ main (int argc, char *argv[])
 		g_setenv ("DISABLE_SSL_STRICT", "1", TRUE);
 	}
 
+	/* this doesn't have to be precise (e.g. using the build-year) as we just
+	 * want to check the clock is not set to the default of 1970-01-01... */
```
superm1, line 3050: Won't ssl checks still fail if you have utc vs local or other similar wrong too though?

hughsie: They might, but it's unlikely. The SSL cert duration is ~3 months IIRC and so a few hours here or there isn't going to matter. What I'm trying to catch here is "no CMOS battery and no NTP"

superm1: Okay

PR #2743, opened 2021-01-11
Require ECC RAM for HSI-3

hughsie:
This provides some protection against attacks like RowHammer.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


docs/hsi.xml
```diff
@@ -742,6 +742,35 @@
         </para>
       </note>
     </refsect3>
+
+    <refsect3 id="org.fwupd.hsi.EccRam">
+      <title>DRAM memory error correction</title>
+      <para>
+        Memory error correction (ECC) is used by memory to correct accidental
```
superm1, line 749: There are multiple mitigations for rowhammer - ECC, LPDDR4 w/TRR (or DDR4 with TRR), shorter refresh intervals.  I think requiring ECC for HSI-3 is heavy handed given there are other mitigations too.

hughsie: I don't think you can get the TRR "enabled" bit in SMBIOS, right? I couldn't see anything in 7.18.2 that would hint at TRR. Maybe punt to HSI-4?

superm1: I don't think it will be in SMBIOS no..  I feel like what we really want is a check that tries to do a row hammer style attack and see if it works.  Not sure how difficult that is to do though..

hughsie: I tried to rowhammer my system (no ECC) for over 1h today and no luck.

hughsie:
Lets close until we can detect pTRR. A few people have raised concerns about this PR already.

PR #2747, opened 2021-01-12
Wip/hughsie/ps2 slowly

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


meson.build
```diff
@@ -249,8 +249,13 @@ gcab = find_program('gcab', required : get_option('tests'))
 bashcomp = dependency('bash-completion', required: false)
 python3 = find_program('python3')
 
-if get_option('plugin_uefi_pk')
+if get_option('gnutls')
```
superm1, line 252: As it's not at all obvious that the synaptics rmi plugin gets gimped without gnutls support, I think we need to make noise if this is turned off.  Perhaps we rename `plugin_synaptics` to `plugin_synaptics_mst` and then add another `plugin_synaptics_rmi` that can emit a warning?

meson_options.txt
```diff
@@ -8,6 +8,7 @@ option('lvfs', type : 'boolean', value : true, description : 'enable LVFS remote
 option('man', type : 'boolean', value : true, description : 'enable man pages')
 option('gudev', type : 'boolean', value : true, description : 'enable GUdev support')
 option('polkit', type: 'boolean', value : true, description : 'enable PolKit support in daemon')
+option('gnutls', type: 'boolean', value : true, description : 'enable GnuTLS support')
```
superm1, line 11: We probably want to make sure this is explicitly added to `dependencies.xml` again.  When jcat was added the direct dependencies for development packages on gnutls got dropped.  They're going to be important again now with this change.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
```diff
@@ -456,13 +358,23 @@ fu_synaptics_rmi_device_setup (FuDevice *device, GError **error)
 	g_autoptr(GByteArray) f01_product_id = NULL;
 	g_autoptr(GByteArray) f01_ds4 = NULL;
 
+	/* assume reset */
+	priv->in_backdoor = FALSE;
```
superm1, line 362: This is a funny name.  Is it more like bootloader mode/flash mode?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
```diff
@@ -118,19 +140,164 @@ fu_synaptics_rmi_v5_device_write_block (FuSynapticsRmiDevice *self,
...
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "failed to verify firmware");
+		return FALSE;
+	}
+#endif
+	/* IMPORTANT: the signature test is performed on the device *AS WELL*
+	 * and checking the signature here just verifies the firmware will
+	 * flash correctly rather than leave the device in bootloader mode */
+	return TRUE;
```
superm1, line 284: So basically you can write a malware or unsigned payload to the device, but if the signature doesn't match, it will just leave the device in bootloader mode rather than flash it?

Unless we automatically kick it back out of bootloader mode this seems like it enables a denial of service attack to have a fallback code.  
1. If a package is compiled without gnutls support
2. Bad payload is put into a CAB file for local installation (not from LVFS or any remote)
3. Process causes CAB file to be installed
4. Device gets stuck in bootloader mode.

We either need to make sure it gets back out of bootloader mode on it's own, or we should make gnutls a hard requirement for this plugin.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
```diff
@@ -151,6 +320,22 @@ fu_synaptics_rmi_v5_device_write_firmware (FuDevice *device,
 		g_prefix_error (error, "not idle: ");
 		return FALSE;
 	}
+	if (fu_synaptics_rmi_firmware_get_sig_size (rmi_firmware) == 0 &&
+	    fu_synaptics_rmi_device_get_sig_size (self) != 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
```
superm1, line 327: `FWUPD_ERROR_INVALID_FILE`?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
```diff
@@ -151,6 +320,22 @@ fu_synaptics_rmi_v5_device_write_firmware (FuDevice *device,
...
+				     "device secure but firmware not secure");
+		return FALSE;
+	}
+	if (fu_synaptics_rmi_firmware_get_sig_size (rmi_firmware) != 0 &&
+	    fu_synaptics_rmi_device_get_sig_size (self) == 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "device not secure but firmware secure");
```
superm1, line 336: How does this happen?  Pre-production?  Or should never happen?  Do we want a `--force` for this scenario maybe?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
```diff
@@ -118,19 +140,164 @@ fu_synaptics_rmi_v5_device_write_block (FuSynapticsRmiDevice *self,
...
+	}
+	ec = gnutls_pubkey_verify_hash2 (pub, GNUTLS_SIGN_RSA_SHA256,
+					 0, &hash, &sig);
+	if (ec < 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "failed to verify firmware");
+		return FALSE;
+	}
```
superm1, line 279: I wonder if this is code that we actually want in the daemon for future use.  It seems pretty useful to me to use gnutls to verify hashes for binaries, and something we might want to do later.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
hughsie: No, it's more than that, and I agree the name might give someone on the internet a nosebleed. It's basically kicking the hardware when it's in bootloader mode already so that it can squirt in some new firmware.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: > but if the signature doesn't match, it will just leave the device in bootloader mode

@blueue -- do you know this?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: It should never happen, but if you try to flash the wrong file (e.g. install-blob on the device) you hit this. We could add --force but I think it's probably never a good idea.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
superm1: Could you pick a better name perhaps for it based on everything else it can do while in this mode?  `flashing_mode`, `iep mode`, something like that?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
superm1: OK agree, then no --force.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: It solves the problem of the plugin needing gnutls... although I'm worried about hardcoding things like the exponent. Maybe this belongs in libjcat or something?

superm1: exponent can be a function argument perhaps?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: Agree, fixed, thanks.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
hughsie: @blueue what about `iep_mode` -- I think @superm1 is right and someone will misunderstand `in_backdoor` to mean something bad.

meson.build
hughsie: https://github.com/fwupd/fwupd/pull/2753

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
blueue: In my experiment, if the function secure_check returns FALSE, fwupd will call attach to let device be back to normal mode.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
blueue: That is great. I was considering the name of "backdoor" is not appropriate before, and I agree to @superm1 about iep_mode. Thanks.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: @blueue I think what mario was asking, was "_if you flash a device that requires an RSA signature test with a file that doesn't have a signature (e.g. if GnuTLS isn't available to test the file before we write it, or if you comment out the code) what happens to the end-user device?_"

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
hughsie: Switched to `iep_mode` now.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
blueue: If the device has RSA public key (you can call it requires RSA signature test) with firmware image file doesn't have a signature, we would treat this update process is invalid. We check this condition in write_firmware before call secure_check function, and return FALSE as failure. That means we won't proceed the firmware update in this condition.

superm1: But is the device remaining in bootloader mode at the end of that?  Or will the process return back to runtime mode on it's own?

blueue: The process would call attach function after got FALSE in return from secure_check, so the device would be back to runtime mode in this case.

superm1: `Well that actually makes my argument stronger.  If the daemon is not linked with gnutls and a unsigned or tampered payload is sent then this function would return `TRUE`.  This means the device gets stuck in bootloader mode since attach is only called in the `FALSE` condition?

If this is the case I think we really should only offer this plugin if the daemon links with gnutls to avoid such a situation.

blueue: Does the daemon is not lined with gnutls means HAVE_GNUTLS is not defined? It will return TRUE directly if my understanding is correct. If we got TRUE in return of this function, the process would be continued like erasing, flashing, and attaching. In this case, the verification on tool side is not activated, and we will rely on checking scheme on firmware side only (checking signature tool writes to firmware after writes data).

superm1: Yeah that's the exact scenario I want to make sure that we fully comprehend and it's safe.  If we make gnutls optional for this plugin we need to worry about what happens if someone doesn't have it.

So if payload is unsigned, will some other function call from fwupd later fail because of firmware side checking scheme?  And in that case does the device get attach called again?

blueue: We can know whether the payload is signed in setup stage because we can get signature flag and size in the header of payload if it is signed. We can know there is an unsigned payload here because it is without signature flag and size in write_firmware function, and return FALSE immediately. And the attach function would be called  after that.

superm1: Okay but what if it's a payload with a header that looks correct? Or it's signed with an untrusted key?

hughsie: I think it's probably best if I just make the plugin hard depend on gnutls.

meson_options.txt
hughsie: We never actually removed it: https://github.com/fwupd/fwupd/blob/master/contrib/ci/dependencies.xml#L510 I'm not sure if it was laziness or deliberate, but either way LGTM :)

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: @blueue could you write some gnutls example code to actually generate this signature please, and then at least we can write some unit tests for it before we work out where it needs to live. Thanks.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: Done!

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
```diff
@@ -8,8 +8,14 @@
 
 #include "config.h"
 
+#ifdef HAVE_GNUTLS
```
superm1, line 11: since it's a hard requirement now for the plugin shouldn't all the `#ifdef HAVE_GNUTLS` go away?

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
blueue: @superm1 Do you mean payload has signature with untrusted key?  We get public key modulus from device, and import a RSA public key by modulus and exponent. The public key and signature should be paired, so it turns verification failure if the signature is faked.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
blueue: @hughsie I found an API of gnutls that can use private key to sign hash data and get the related signature as well.
https://www.gnu.org/software/gnutls/reference/gnutls-abstract.html#gnutls-privkey-sign-data

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: Agree; done.

plugins/synaptics-rmi/fu-synaptics-rmi-v5-device.c
hughsie: I looked into this a bit over the weekend; I don't want libjcat to basically wrap the API of gnutls_pubkey_verify_hash2 -- instead I've moved the gnutls verification to synaptics-rmi-common. If another plugin starts doing the same verification we can move it out to a libfwupdplugin shared function but until then I think it's fine where it is.

plugins/synaptics-rmi/fu-synaptics-rmi-ps2-device.c
```diff
@@ -0,0 +1,947 @@
...
+							     &error_local)) {
+			g_debug ("failed write try #%u: %s",
+				 i, error_local->message);
+			continue;
+		}
+		success = TRUE;
+		break;
+	}
+	if (success == FALSE) {
+		g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "failed");
```
superm1, line 265: Would it be useful to save the final error from the for loop and propagate it?

superm1 (approved):

plugins/synaptics-rmi/fu-synaptics-rmi-ps2-device.c
hughsie: Well, it can only be a timeout, but sure, G_IO_ERROR_TIMED_OUT is a better error code for sure. Will fix.

superm1:
@hughsie with this in, do you have a lot more in mind for next release?  I'd like to make sure I can get 1.5.x into Debian's upcoming bullseye with their freeze around the corner.  So wondering if we can tag sometime this week so there's enough time.

hughsie:
> do you have a lot more in mind for next release

I did plan of shipping 1.5.5 and leaving master to settle a bit; this and the Lenovo changes need time to settle. There are always going to be more fixes and more plugins, so maybe do 1.5.5 and then we do a 1.5.6 in a couple of weeks time?

superm1:
OK i'll get 1.5.5 in, but at least cherry pick that patch for `Unknown`.

hughsie:
You might want 7619665d741237910aee36b28b1dfbad342db403 too

PR #2750, opened 2021-01-12
Check the return values of g_return_val_if_fail() in CI

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/check-null-false-returns.py
+++ b/contrib/ci/check-null-false-returns.py
@@ -1,0 +1,167 @@
+#!/usr/bin/python3
+# pylint: disable=invalid-name,missing-docstring,too-many-branches,too-many-statements
+#
+# Copyright (C) 2021 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+import glob
+import sys
+
+
+def parse_rvif(line):
+
+    # remove whitespace
+    line = line.replace("\t", "")
+    line = line.replace(" ", "")
+
+    # find value
+    line = line.replace(",", "|")
+    line = line.replace(")", "|")
+    return line.rsplit("|", maxsplit=2)[1]
+
+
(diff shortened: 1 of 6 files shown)


contrib/check-null-false-returns.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/python3
```
superm1, line 1: shouldn't this be in `contrib/ci` if it's for CI purposes.

hughsie: Yes! Fixed!

contrib/ci/check-null-false-returns.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # remove whitespace
+    line = line.replace("\t", "")
+    line = line.replace(" ", "")
+
+    # find value
+    line = line.replace(",", "|")
+    line = line.replace(")", "|")
+    return line.rsplit("|", maxsplit=2)[1]
+
```
superm1, line 22: I think you need to add something to strip comments no?  Because `/* foo bar * / g_return_val_if_fail(...);` is valid C isn't it?

superm1 (approved):
just a question around comments, otherwise looks good to me, good idea.  Seem it found a lot of interesting things already too!

For `developing.md` purpose, maybe we should have a helper thing in `contrib` that could wrap these in commit hooks too?

contrib/ci/check-null-false-returns.py
hughsie: Sure, but we're doing a rsplit, so unless the comment is something like `g_return_val_if_fail(...); /* should use foo_cb()` I think it's fine.

hughsie:
> maybe we should have a helper thing in contrib that could wrap these in commit hooks too

I'm not a huge fan of git commit hooks (and have zero experience) but if this is something you'd find useful please be my guest.

PR #2757, opened 2021-01-13
Wip/hughsie/no capsule coalesce2

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -762,6 +762,38 @@ fu_plugin_unlock (FuPlugin *plugin, FuDevice *device, GError **error)
 	return TRUE;
 }
 
+static void
+fu_plugin_uefi_update_state_notify_cb (GObject *object,
+				       GParamSpec *pspec,
+				       FuPlugin *plugin)
+{
+	FuDevice *device = FU_DEVICE (object);
+	GPtrArray *devices;
+	g_autofree gchar *msg = NULL;
+
+	/* device is not in needs-reboot state */
+	if (fu_device_get_update_state (device) != FWUPD_UPDATE_STATE_NEEDS_REBOOT)
+		return;
+
+	/* only do this on hardware that cannot coalesce multiple capsules */
+	if (!fu_plugin_has_custom_flag (plugin, "no-coalesce"))
+		return;
+
+	/* mark every other device for this plugin as non-updatable */
+	msg = g_strdup_printf ("Cannot update as %s [%s] needs reboot",
+			       fu_device_get_name (device),
+			       fu_device_get_id (device));
+	devices = fu_plugin_get_devices (plugin);
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *device_tmp = g_ptr_array_index (devices, i);
+		if (device_tmp == device)
+			continue;
+		fu_device_remove_flag (device_tmp, FWUPD_DEVICE_FLAG_UPDATABLE);
+		fu_device_add_flag (device_tmp, FWUPD_DEVICE_FLAG_UPDATABLE_HIDDEN);
+		fu_device_set_update_error (device_tmp, msg);
+	}
+}
+
 gboolean
 fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 {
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-plugin.c
```diff
@@ -507,6 +518,10 @@ fu_plugin_device_add (FuPlugin *self, FuDevice *device)
 		return;
 	}
 
+	/* add to array */
+	fu_plugin_ensure_devices (self);
+	g_ptr_array_add (priv->devices, g_object_ref (device));
```
superm1, line 523: are you missing the matching unref on remove?

hughsie: No, `g_ptr_array_new_with_free_func ((GDestroyNotify) g_object_unref);` speficies the function to use when deallocating or removing items from the array.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -762,6 +762,38 @@ fu_plugin_unlock (FuPlugin *plugin, FuDevice *device, GError **error)
...
+			       fu_device_get_name (device),
+			       fu_device_get_id (device));
+	devices = fu_plugin_get_devices (plugin);
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *device_tmp = g_ptr_array_index (devices, i);
+		if (device_tmp == device)
+			continue;
+		fu_device_remove_flag (device_tmp, FWUPD_DEVICE_FLAG_UPDATABLE);
+		fu_device_add_flag (device_tmp, FWUPD_DEVICE_FLAG_UPDATABLE_HIDDEN);
+		fu_device_set_update_message (device_tmp, msg);
```
superm1, line 793: shouldn't this be update error?

superm1 (commented):
Are there any plugins left tracking their own device cache in a `FuPluginData`?  If so those should probably be removed as a result of this change.

superm1:
I think this should be documented in `uefi-capsule/README.md` as it's a pretty big difference between Lenovo's implementation and everyone elses'.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: Hmm, I wasn't sure. It's not like the error was attempted and it failed, which I think is how we use update_error now.

superm1: Update message is shown when an update is done.  It's stuff like the WD19 does where it lets you know the update finishes next time you unplug.  I think of Update Error as any time it's not updatable, or failed to update here is why.

hughsie: Excellent, agreed -- fixed.

hughsie:
> Are there any plugins left tracking their own device cache in a FuPluginData

Good question. I think the only plugin doing this is synaptics_mst, and I didn't convert this as it's not strictly the same thing. fu_plugin_get_devices() is getting all the *added* devices, and synaptics_mst is using the device array for devices that might exist right now or might appear in the future during a rescan -- which is subtly different.

hughsie:
>  think this should be documented in uefi-capsule/README.md

Something like this perhaps?

	On some Lenovo hardware only one capsule is installable due to possible problems
	with the UpdateCapsule coalesce operation. As soon as one UEFI device has been
	scheduled for update the other UEFI devices found in the ESRT will be marked
	as `updatable-hidden` rather than `updatable`. Rebooting will restore them so
	they can be updated on next OS boot.

superm1:
I don't know what that HWID corresponds to, is it just some Lenovo hardware or all Lenovo hardware?  The text looks good otherwise for the capsule case.  Maybe also document the reason for the "Linux-Firmware-Updater" vs "Linux Firmware Updater" at same time though.

hughsie:
> is it just some Lenovo hardware or all Lenovo hardware

At the moment it's all Lenovo hardware, which is suboptimal as we only really want to match on specific thinkpad models

>  Maybe also document the reason for the "Linux-Firmware-Updater

Good idea, done.

superm1 (approved):

PR #2762, opened 2021-01-13
Make libarchive optional for fuzzing

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -214,7 +214,10 @@ endif
 libxmlb = dependency('xmlb', version : '>= 0.1.13', fallback : ['libxmlb', 'libxmlb_dep'])
 gusb = dependency('gusb', version : '>= 0.3.5', fallback : ['gusb', 'gusb_dep'])
 sqlite = dependency('sqlite3')
-libarchive = dependency('libarchive')
+if get_option('libarchive')
+  libarchive = dependency('libarchive')
+  conf.set('HAVE_LIBARCHIVE', '1')
+endif
 endif
 libjcat = dependency('jcat', version : '>= 0.1.0', fallback : ['libjcat', 'libjcat_dep'])
 libjsonglib = dependency('json-glib-1.0', version : '>= 1.1.1')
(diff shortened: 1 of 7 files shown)


superm1:
If you do merge something like https://github.com/fwupd/fwupd/pull/2738 then perhaps that could instead of a new build flag it could be `required: true` only if set to a non-fuzzing build.

meson.build
```diff
@@ -214,7 +214,10 @@ endif
 libxmlb = dependency('xmlb', version : '>= 0.1.13', fallback : ['libxmlb', 'libxmlb_dep'])
 gusb = dependency('gusb', version : '>= 0.3.5', fallback : ['gusb', 'gusb_dep'])
 sqlite = dependency('sqlite3')
-libarchive = dependency('libarchive')
+if get_option('libarchive')
+  libarchive = dependency('libarchive')
+  conf.set('HAVE_LIBARCHIVE', '1')
+endif
```
superm1, line 220: rather than a new option, how about instead make it like this:
```
libarchive = dependency('libarchive', required: false)
if libarchive.found()
    conf.set('HAVE_LIBARCHIVE', '1')
else
   warning('buliding without libarchive support')
endif
```

meson.build
```diff
@@ -480,11 +483,14 @@ if build_standalone
   plugin_deps += giounix
   plugin_deps += gmodule
   plugin_deps += gusb
-  plugin_deps += libarchive
   plugin_deps += gudev
   plugin_deps += libjsonglib
 endif
 
+if get_option('libarchive')
```
superm1, line 490: if you adopt my other suggestion, then this becomes
`if libarchive.found()`

meson.build
hughsie: I think the gentoo/arch crowd get very upset when you do this; https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

superm1: Then  I guess this should be correct PR then until #2738 goes in, in which case I think you really want to make it a requirement for anything but a fuzzing build.

superm1 (approved):

PR #2763, opened 2021-01-13
Increase the default maximum archive size to 1GB

zhj8407:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-config.c
+++ b/src/fu-config.c
@@ -301,7 +301,7 @@ fu_config_class_init (FuConfigClass *klass)
 static void
 fu_config_init (FuConfig *self)
 {
-	self->archive_size_max = 512 * 0x100000;
+	self->archive_size_max = G_MAXUINT64;
 	self->disabled_devices = g_ptr_array_new_with_free_func (g_free);
 	self->disabled_plugins = g_ptr_array_new_with_free_func (g_free);
 	self->approved_firmware = g_ptr_array_new_with_free_func (g_free);


superm1:
Per discussion in original issue we're probably better off dropping upper limit all together.  Set default in daemon conf to `-1`, and then update matching code to that behavior.  Can you rework that?

hughsie:
Isn't `0` no limit already?

zhj8407:
0 means the default. In this PR, I change the default to 1GB

superm1:
> Isn't `0` no limit already?

Nope.  0 means: https://github.com/fwupd/fwupd/blob/933e43005fb4f32da0a8c04f6a263f8e39d4ade2/src/fu-config.c#L304

So maybe that is what needs changing to 0, and then any code that uses it.

zhj8407:
Do you mean the max value of guint64? If I set the initial self->archive_size_max to 0, none of the cab files could be decompressed.
`	/* check the size is sane */
	if (gcab_cabinet_get_size (self->gcab_cabinet) > self->size_max) {
		g_autofree gchar *sz_val = g_format_size (gcab_cabinet_get_size (self->gcab_cabinet));
		g_autofree gchar *sz_max = g_format_size (self->size_max);
		g_set_error (error,
			     FWUPD_ERROR,
			     FWUPD_ERROR_INVALID_FILE,
			     "archive too large (%s, limit %s)",
			     sz_val, sz_max);
		return FALSE;
	}`

superm1:
Other code using this value will probably need to check if `size_max > 0` as a result of that change.

zhj8407:
Can we change the initial value to G_MAXUINT64? So, if the config is 0, that means unlimited size.
https://github.com/fwupd/fwupd/blob/924c7080fc1ab051e8ff1cc759f65534cf2094c8/src/fu-config.c#L304

In the current code, the archive_size_max and size_max are defined as unsigned int64. Not easy to handle the minus value.

superm1:
I think that seems like reasonable consolation, yes.

zhj8407:
Okay. I will do the rework.

zhj8407:
Seems like the checks were failed at
Installing test firmware...
Unknown                  [***************************************]
WARNING: The daemon has loaded 3rd party code and is no longer supported by the upstream developers!
WARNING: Firmware can not be updated in legacy BIOS mode
  See https://github.com/fwupd/fwupd/wiki/PluginFlag:legacy-bios for more information.
Message recipient disconnected from message bus without replying
No journal files were found.
-- No entries --
FAIL: fwupd/fwupdmgr.test (Child process exited with code 1)

zhj8407:
How can I retrigger the checks?

superm1:
If you force push, it will re-run the checks.

zhj8407:
Same failure in the test. Seems like setting the max_size to G_MAXUINT64 breaks some tests? I can not find detailed logs.

hughsie:
Alternate proposal: https://github.com/fwupd/fwupd/pull/2768

superm1:
Closing this one since it's superceded with a solution that fixes the daemon crash caused by it.

PR #2764, opened 2021-01-13
Make GUsb optional for fuzzing

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


meson_options.txt
```diff
@@ -8,6 +8,7 @@ option('lvfs', type : 'boolean', value : true, description : 'enable LVFS remote
 option('man', type : 'boolean', value : true, description : 'enable man pages')
 option('libarchive', type : 'boolean', value : true, description : 'enable libarchive support')
 option('gudev', type : 'boolean', value : true, description : 'enable GUdev support')
+option('gusb', type : 'boolean', value : true, description : 'enable GUsb support')
```
superm1, line 11: rather than `gusb`, maybe this should be `usb`.  And then it's `enable usb support`, and tears out everything related to usb rather than surgically the things that make references to `gusb`.

meson.build
```diff
@@ -212,7 +212,10 @@ else
   gudev = dependency('', required : false)
 endif
 libxmlb = dependency('xmlb', version : '>= 0.1.13', fallback : ['libxmlb', 'libxmlb_dep'])
-gusb = dependency('gusb', version : '>= 0.3.5', fallback : ['gusb', 'gusb_dep'])
+if get_option('gusb')
+  gusb = dependency('gusb', version : '>= 0.3.5', fallback : ['gusb', 'gusb_dep'])
+  conf.set('HAVE_GUSB', '1')
+endif
```
superm1, line 218: I think win32 build really doesn't make sense unless you have gusb - that's the only plugins we can do.
t probably should be set to fail out without usb support.

meson_options.txt
hughsie: The problem with that way is we've got a lot of references to FuUsbDevice in the various exported interfaces. This would mean the libfwupdplugin ABI/API is different depending on the compile setting, which is something I'd like to avoid.

meson.build
hughsie: Fixed, good idea.

superm1 (approved):

PR #2767, opened 2021-01-14
trivial: in the no updatable devices case, show error again

superm1:
Since we added a CPU and everything has a CPU the error stopped
showing up.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
It took me a while to get the logic, but yes, good catch, agreed.

Issue #2760, opened 2021-01-13
The default archive_size_max (512MB) is small for some Poly device's firmware.


zhj8407:
**Describe the bug**
The size of Poly Studio's firmware archive exceeds the default archive_size_max (512MB). 

**Steps to Reproduce**

1. Plug in the Poly Studio.
2. Run "sudo fwupdtool install poly_studio_fw_1_4_0_000770.bin.cab".
3. Error message "**archive too large (583.6 MB, limit 536.9 MB)**" is reported.

**Expected behavior**
The firmware upgrade can run successfully.

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version
client version:	1.3.11
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.3.11
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
apt install fwupd

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
$ fwupdmgr get-devices --show-all-devices
─Poly Studio:
│     Device ID:             ac8542369709970eaeabdca7d4f6e16aec31ad37
│     Current version:    2.7
│     Vendor:                 Poly, Inc. (USB:0x095D)
│     GUIDs:                  ef8d95ed-479a-5a0a-98cb-1799a11b7b25
│                                   d22c7eca-b220-5070-8844-c1f34ef56af0
│     Device Flags:        • Updatable
```

**Additional questions**
- Operating system and version: Ubuntu 20.04
- Have you tried rebooting? Yes
- Is this a regression? No

superm1:
Seems it's been set to 512MB for a while: https://github.com/fwupd/fwupd/commit/c7bbbc246b89df3d121e745a052322773a74e3ff

I guess we should bump up to 1GB?  Can you submit PR?

zhj8407:
Sure, no problem. I can submit the PR.

hughsie:
Could we do something smarter? e.g. set it to 50% of the total RAM or something?

superm1:
Lets see, the reason this was originally done was to prevent DoS the daemon with big files.
https://github.com/fwupd/fwupd/pull/353

[...]

PR #2768, opened 2021-01-14
Use a quarter of the RAM size as the archive maximum

hughsie:
It is impossible to choose a static default that is appropriate for both a tiny
ARM IoT device and a giant Xeon server.

Fixes https://github.com/fwupd/fwupd/issues/2760

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -19,6 +19,10 @@
 #include <shlwapi.h>
 #endif
 
+#ifdef _WIN32
+#include <sysinfoapi.h>
+#endif
+
 #ifdef HAVE_CPUID_H
 #include <cpuid.h>
 #endif
@@ -31,6 +35,7 @@
 #include <limits.h>
 #include <string.h>
 #include <stdlib.h>
+#include <unistd.h>
 
 #include "fwupd-error.h"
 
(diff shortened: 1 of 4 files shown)


hughsie:
@superm1 I see `Message recipient disconnected from message bus without replying` -- any ideas how we can get the backtrace (or verbose, or criticals) from inside the VM?

superm1:
> @superm1 I see `Message recipient disconnected from message bus without replying` -- any ideas how we can get the backtrace (or verbose, or criticals) from inside the VM?

Is that the daemon or the client crashing?  I can't tell for sure from that message.
This is the installed tests, so I would think just add `--verbose` to the individual failing test for `fwupd/fwupdmgr.test` if it's the client.  If it's actually the daemon crashing then modifying the systemd unit before this is run using `sed`: https://github.com/fwupd/fwupd/blob/master/contrib/ci/debian.sh#L77 to add verbose or critical checking.  That might not be a bad idea to do by default if CI is true.

superm1:
Here's the failure reproduced locally with the same test suite and the patch from https://github.com/fwupd/fwupd/pull/2763 (which makes it easier to reproduce):
```
Jan 14 11:40:59 supermario-XPS-13-7390 fwupd[222534]: 17:40:59:0820 FuMain               Called Install(08d460be0f1f9f128413f816022a6439e0078018,8)
Jan 14 11:40:59 supermario-XPS-13-7390 fwupd[222534]: 17:40:59:0820 FuMain               got option reason
Jan 14 11:40:59 supermario-XPS-13-7390 fwupd[222534]: 17:40:59:0820 GLib                 ../../../glib/gmem.c:108: failed to allocate 18446744073709551615 bytes
```

hughsie:
> ../../../glib/gmem.c:108: failed to allocate 18446744073709551615 bytes

That's uint64_t i.e. G_MAXUINT64, right? Can you get a `bt` from where that's exploding please? and maybe also print the value of fu_config_get_archive_size_max() and count in fu_common_get_contents_fd() -- I'm very confused!

hughsie:
I think I worked out the problem. I moved the -Dci bits to the wip/meson-Dci branch.

superm1 (approved):

superm1 (approved):

PR #2773, opened 2021-01-14
Wip/hughsie/flashrom

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -6,3 +6,8 @@ DeviceId=librem15v3
 [HwId=52b68c34-6b31-5ecc-8a5c-de37e666ccd5]
 DeviceId=LT1000
 VersionFormat=quad
+
+# Starlabs LabTop L4
+[HwId=baf1d04e-fd16-5e6a-93cc-1c23d171f879]
+DeviceId=StarlabsLabTopL4
+VersionFormat=triplet
--- a/plugins/flashrom/fu-plugin-flashrom.c
+++ b/plugins/flashrom/fu-plugin-flashrom.c
@@ -264,6 +264,7 @@ fu_plugin_update (FuPlugin *plugin,
 	}
 
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_WRITE);
+	fu_device_set_progress (device, 0); /* urgh */
 	rc = flashrom_image_write (data->flashctx, (void *) buf, sz, NULL /* refbuffer */);
 	if (rc != 0) {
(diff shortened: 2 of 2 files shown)


hughsie:
@Sean-StarLabs this, and then we'll need some coreboot fixes.

Sean-StarLabs:
@hughsie Awesome, I'll give it a test.

Is it worth adding the LOCKED flag to check BLE is disabled?

plugins/flashrom/flashrom.quirk
```diff
@@ -6,3 +6,8 @@ DeviceId=librem15v3
 [HwId=52b68c34-6b31-5ecc-8a5c-de37e666ccd5]
 DeviceId=LT1000
 VersionFormat=quad
+
+# Starlabs LabTop L4
+[HwId=baf1d04e-fd16-5e6a-93cc-1c23d171f879]
+DeviceId=StarlabsLabTopL4
+VersionFormat=triplet
```
superm1, line 13: triplet is correct?  others in file use quad, so just want to make sure.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -264,6 +264,7 @@ fu_plugin_update (FuPlugin *plugin,
 	}
 
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_WRITE);
+	fu_device_set_progress (device, 0); /* urgh */
```
superm1, line 268: we can't get anything out of the library?  Sounds like a development action for a future version.  Can we at least have time estimates that we can show?  Otherwise it looks hung.

plugins/flashrom/flashrom.quirk
hughsie: In this case, yes. I've got a follow up patch after this PR to make the coreboot stuff more sane.

plugins/flashrom/fu-plugin-flashrom.c
hughsie: At the moment, no, but I've approached flashrom with a proposal, although nobody (including me) actually wrote any code yet: https://www.mail-archive.com/flashrom@flashrom.org/msg13519.html

hughsie: **Narrator**: _He wrote some code_

hughsie:
> Is it worth adding the LOCKED flag to check BLE is disabled

I have a followup patch for this I need to polish a bit; then yes, we can use the SMBIOS info.

PR #2777, opened 2021-01-15
trivial: Show the device branch in fwupdmgr

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1217,6 +1217,13 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 		fu_common_string_append_kv (str, idt + 1, _("Vendor"), tmp2);
 	}
 
+	/* branch */
+	if (fwupd_device_get_branch (dev) != NULL) {
+		/* TRANSLATORS: the stream of firmware, e.g. nonfree or open-source */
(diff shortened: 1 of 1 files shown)


src/fu-util-common.c
```diff
@@ -1217,6 +1217,13 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 		fu_common_string_append_kv (str, idt + 1, _("Vendor"), tmp2);
 	}
 
+	/* branch */
+	if (fwupd_device_get_branch (dev) != NULL) {
+		/* TRANSLATORS: the stream of firmware, e.g. nonfree or open-source */
+		fu_common_string_append_kv (str, idt + 1, _("Branch"),
```
superm1, line 1223: As this is an attribute of the firmware installed not the device I wonder if you want to prefix this with "Firmware", IE `Firmware Branch`

hughsie: Or `Release Branch` perhaps?

superm1: Yeah that's good

hughsie: Fixed.

superm1 (approved):

PR #2779, opened 2021-01-15
Always return a valid semver from fu_common_version_ensure_semver()

hughsie:
Return NULL if not valid, and handle tilde and dash where required.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -1312,6 +1312,35 @@ fu_common_strstrip_func (void)
 }
 
 static void
+fu_common_version_semver_func (void)
+{
+	struct {
+		const gchar *old;
+		const gchar *new;
+	} map[] = {
+		{ "1.2.3",		"1.2.3" },
+		{ "1.2-3",		"1.2.3" },
+		{ "1~2-3",		"1.2.3" },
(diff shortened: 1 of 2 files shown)


libfwupdplugin/fu-self-test.c
```diff
@@ -1287,6 +1287,34 @@ fu_common_strstrip_func (void)
...
+		{ "1..2",		"1.2" },
+		{ "CBET1.2.3",		"1.2.3" },
+		{ "1.2.3alpha",		"1.2.3" },
+		{ "5",			"5" },
+		{ "\t5\n",		"5" },
+		{ "coreboot-unknown",	NULL },
+		{ "",			NULL },
+		{ " ",			NULL },
+		{ NULL,			NULL }
+	};
```
hughsie, line 1311: @superm1 any more edge cases you can think of?

superm1: Hex?

superm1 (commented):
Maybe ending in a -?

libfwupdplugin/fu-self-test.c
hughsie: Good idea, thanks.

PR #2780, opened 2021-01-15
flashrom: Do not allow flashing if BLE is enabled

hughsie:
If BLE is set flashrom isn't going to work and the user would get a super scary
warning.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/pci-bcr/fu-plugin-pci-bcr.c
+++ b/plugins/pci-bcr/fu-plugin-pci-bcr.c
@@ -30,6 +30,16 @@ fu_plugin_init (FuPlugin *plugin)
 	priv->bcr_addr = 0xdc;
 }
 
+static void
+fu_plugin_pci_bcr_set_updatable (FuPlugin *plugin, FuDevice *dev)
+{
+	FuPluginData *priv = fu_plugin_get_data (plugin);
+	if (priv->bcr & BCR_BLE) {
+		fu_device_remove_flag (dev, FWUPD_DEVICE_FLAG_UPDATABLE);
+		fu_device_set_update_error (dev, "BIOS locked");
+	}
+}
+
 void
 fu_plugin_device_registered (FuPlugin *plugin, FuDevice *dev)
 {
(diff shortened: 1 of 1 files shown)


plugins/pci-bcr/fu-plugin-pci-bcr.c
```diff
@@ -42,6 +52,14 @@ fu_plugin_device_registered (FuPlugin *plugin, FuDevice *dev)
 			priv->bcr_addr = tmp;
 		}
 	}
+	if (g_strcmp0 (fu_device_get_plugin (dev), "flashrom") == 0) {
```
superm1, line 55: is this going to inadvertanely affect UEFI based `main-system-firmware` too later very easily by accident?  That's not flashable from _within_ OS, but absolutely should still be marked updatable for handling the UEFI capsule.

I know right now you're checking for if a flashrom device is modified, but I think you need to check the _specific_ flashrom device is main-system-firmware.

hughsie: > but I think you need to check the specific flashrom device is main-system-firmware

I think we only support one flashrom device right now, and that unconditionally has `but I think you need to check the specific flashrom device is main-system-firmware` but I do get the point. What about something like this:

    -       if (g_strcmp0 (fu_device_get_plugin (dev), "flashrom") == 0) {
    +       if (g_strcmp0 (fu_device_get_plugin (dev), "flashrom") == 0 &&
    +           fu_device_has_instance_id (dev, "main-system-firmware")) {

superm1: >I think we only support one flashrom device right now, and that unconditionally has main-system-firmware

Right.  I'm just looking out for a future non-obvious bug.

>What about something like this:

Yes - exactly what I had in mind.

hughsie: Okay, fixed, thanks.

superm1 (approved):

PR #2781, opened 2021-01-15
Allow overriding SMBIOS data when using fwupdtool hwids

hughsie:
This is really useful when working out HwId quirk values for remote systems.
This is similar to what ComputerHardwareIds.exe allows:

    FWUPD_SMBIOS_DATA="Family=Foo|Manufacturer=Bar|ProductSku" fwupdtool hwids

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -1596,31 +1596,68 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
 }
 
 static gboolean
+fu_util_export_hwids (FuUtilPrivate *priv, gchar **values, GError **error)
+{
+	g_autoptr(FuHwids) hwids = fu_hwids_new ();
+	g_autoptr(FuSmbios) smbios = fu_smbios_new ();
+	g_autoptr(GKeyFile) kf = g_key_file_new ();
+	g_autoptr(GPtrArray) hwid_keys = NULL;
+
+	/* check args */
+	if (g_strv_length (values) != 1) {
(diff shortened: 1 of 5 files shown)


src/fu-tool.c
```diff
@@ -1614,6 +1615,17 @@ fu_util_hwids (FuUtilPrivate *priv, gchar **values, GError **error)
 		FU_HWIDS_KEY_BASEBOARD_PRODUCT,
 		NULL };
 
+	/* secret way to override SMBIOS for debugging */
```
superm1, line 1618: Why not an argument with an XML or JSON payload?  It seems like it's cleaner...

superm1: Similarly, maybe it's worth being able to dump "current" values as fwupd sees, so you can just change one?
I guess I don't know the underlying use for this - so if that's a lot of work without a lot of utility speak up.

hughsie: > Similarly, maybe it's worth being able to dump "current" values as fwupd sees

Makes sense. I added `import-hwids` and `export-hwids` although I'm not completely sure about the verbs in both. Better ideas welcome.

superm1: I think those are fine, but you probably want the old `hwids` command to rename to match too.  `print-hwids`?

hughsie: Nah, I merged the import hwids action into the existing one; it's easy enough to detect.

superm1 (approved):

PR #2783, opened 2021-01-18
Wip/hughsie/make images

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/po/make-images
+++ b/po/make-images
@@ -176,7 +176,8 @@ class Rasterizer:
                                  'height': height,
                                  'suffix': 'bmp.gz'})
                 filename = self.pattern.format_map(nameinfo)
-                render_one(lang, string, width, height, filename)
+                if not os.path.exists(filename):
+                    render_one(lang, string, width, height, filename)
         print("")
 
 if __name__ == '__main__':
--- a/po/make-images
+++ b/po/make-images
@@ -13,6 +13,8 @@ import os
 import sys
 import gettext
 import math
+import io
+import gzip
 import cairo
 import gi
 gi.require_version('Pango', '1.0')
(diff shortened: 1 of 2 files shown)


superm1 (approved):

superm1:
I wish there was a nice way to switch the image build into 'build' time rather than install time in the first place.  tbh it feels like a bit of a hack this way at least.

hughsie:
> I wish there was a nice way to switch the image build into 'build' time rather than install time in the first place

So, I spent a few hours trying to do this just now. It's kinda possible, but it gets a bit ugly:

 * Meson can't generate `en/foo.bmp` and also `fr/foo.bmp` as they are generated in the same source directory with the same basename: `Multiple producers for Ninja target bar.bmp`
 * You can't generate `en_foo.bmp` and then `configure_file` to install them as something different, as you get `ERROR: Inputs can only be strings or file objects` as even with `.full_path()` the configure phase runs before the build phase and so the file isn't there to be installed when required.
 * You can't use `run_command` as this would run unconditionally on build, and even checking for the filename slows everything down
 * We "skip" writing files like `en_GB/fwupd.bmp` if the source string is the same as `en` -- which really confuses meson if we try to later install something that's not been generated; e.g. `custom_target()` ran with success but didn't actually write anything. I guess we could write a 0-byte file in that case but it's yet another hack.

One alternative is that we stop installing into `/usr/share/locale/id/LC_IMAGES/fwupd-1024-768.bmp.gz` and instead install into something like `/usr/share/fwupd/ux/fwupd-en-1024-768.bmp.gz` -- it's the same amount of work to look up the image in `fu_plugin_uefi_capsule_get_splash_data()` but it does kinda feel like the right thing to do installing into `/usr/share/locale/<locale>/LC_IMAGES`. Same requirement for the 0-byte file also.

Other ideas very welcome.

superm1:
>Meson can't generate en/foo.bmp and also fr/foo.bmp as they are generated in the same source directory with the same basename: Multiple producers for Ninja target bar.bmp

How about changing the name to instead be `foo/en.bmp` and `foo/fr.bmp`?

>One alternative is that we stop installing into /usr/share/locale/id/LC_IMAGES/fwupd-1024-768.bmp.gz and instead install into something like /usr/share/fwupd/ux/fwupd-en-1024-768.bmp.gz -- it's the same amount of work to look up the image in fu_plugin_uefi_capsule_get_splash_data() but it does kinda feel like the right thing to do installing into /usr/share/locale/<locale>/LC_IMAGES. Same requirement for the 0-byte file also.

What about putting in `/usr/share/fwupd/ux` and symlink into `/usr/share/locale`?  Not sure that helps.

hughsie:
> How about changing the name to instead be foo/en.bmp and foo/fr.bmp?

Yes, I played with `'fwupd-@0@-@1@-@2@.bmp.gz'.format(language, size[0], size[1])` but that would be `/usr/share/locale/en/LC_IMAGES/fwupd-en-1920-1080.bmp.gz` which is pretty weird. Maybe that's actually okay -- no weirder than installing gzipped bitmaps to LC_IMAGES.

superm1:
I haven't heard of anything actually using `LC_IMAGES` before fwupd so that was weird to me too in the first place.  However I don't see any problem with adding the language into the filename, it certainly won't conflict with anything.

PR #2785, opened 2021-01-19
pci-bcr: Only mark the device non-updatable if WPD unset and BLE set

hughsie:
It's unusual, but if BIOS lock enable is enabled (so we cannot *change* the
value of BIOSWE) but the BIOS is already WE then we can write to the hardware
just fine.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
This seems like a firmware bug, but the logic sounds correct I suppose.

hughsie:
> This seems like a firmware bug

Yes, I agree. I reported it to the OEM, but there's firmware "in the wild " that does this.

PR #2786, opened 2021-01-19
coreboot: Refactor plugin to add metadata to flashrom devices

hughsie:
The coreboot plugin never actually gained the ability to write. As it stands a
coreboot system now adds *two* system-firmware devices (from both flashrom and
coreboot) which isn't ideal.

Just allow flashrom to enumerate quirked devices and add coreboot-specific
metadata as required when the device is registered. If we require some kind
of cbfs parsing then we can do that in FuFlashromDevice->prepare_firmware().

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -83,6 +85,39 @@ fu_plugin_flashrom_debug_cb (enum flashrom_log_level lvl, const char *fmt, va_li
...
+	const gchar *version;
+	const gchar *version_major;
+	const gchar *version_minor;
+
+	/* as-is */
+	version = fu_plugin_get_dmi_value (plugin, FU_HWIDS_KEY_BIOS_VERSION);
+	if (version != NULL) {
+		/* some Lenovo hardware requires a specific prefix for the EC,
+		 * so strip it before we use ensure-semver */
+		if (strlen (version) > 9 && g_str_has_prefix (version, "CBET"))
```
superm1, line 100: why did you take this off the quirk approach?  I thought it was Lenovo specific.

superm1:
TBH I think having a coreboot plugin just for adding a few metadata things now is a bit too heavy - why not just fold it ALL back into flashrom?

plugins/flashrom/fu-plugin-flashrom.c
hughsie: I did; I can add it back if you think it's worth it, but I really don't think any other vendor would have a CBET12345 prefix on firmware out of coincidence.

superm1: I guess that's true.  OK no concerns.

hughsie:
> why not just fold it ALL back into flashrom

Huh, I guess that's true. Will try.

hughsie:
> Will try.

^^

plugins/flashrom/README.md
```diff
@@ -17,9 +17,36 @@ This plugin supports the following protocol ID:
 
  * org.flashrom
 
+BIOS Version String
```
superm1, line 20: Probably need to prefix this with Coreboot

superm1 (approved):

plugins/flashrom/README.md
hughsie: Fixed.

PR #2787, opened 2021-01-19
Add flags to flashrom

Sean-StarLabs:
Set flashrom to build in deb packages

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
The reason this isn't done today is that flashrom isn't in main in Ubuntu. We need a MIR filed and approved in Ubuntu to turn it on by default.

Can you please file one?

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -21,6 +21,8 @@ fu_flashrom_device_init (FuFlashromDevice *self)
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_INTERNAL);
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE);
```
hughsie, line 25: I suppose the latter is true, indeed.

superm1: I think it's true for coreboot devices with multiple partitions, but all flashrom devices?  I don't know.

contrib/debian/rules
```diff
@@ -18,18 +18,7 @@ endif
...
-	CONFARGS += -Dplugin_flashrom=false
-else
-	TMPLDIR  := debian/fwupd-$(DEB_HOST_ARCH)-signed-template/usr/share/code-signing/fwupd-$(DEB_HOST_ARCH)-signed-template
-	ifneq ($(DEB_HOST_ARCH_CPU),ia64)
-		CONFARGS += -Dplugin_flashrom=true
-	else
-		CONFARGS += -Dplugin_flashrom=false
-	endif
-endif
+CONFARGS += -Dplugin_flashrom=true
```
superm1, line 21: Here is the MIR process: https://wiki.ubuntu.com/MainInclusionProcess
You'll need to get that approved before we can pull in this specific change.

plugins/flashrom/fu-flashrom-device.c
hughsie: I don't know of any machines that explode when you write to the SPI; some machines prevent writing the ME but flashrom won't try to (or be able to) write the SPI area in that case.

superm1: The BIOS code i'm not concerned about.  Ignoring ME - I think it comes down to whether or not the EC stores an image on the SPI and if that is being updated at the same time and has multiple banks and how it gets restarted.

I would feel more comfortable if this flag was added only on a HardwareId to match the expectations for a manufacturer and not generic.

hughsie: > I would feel more comfortable if this flag was added only on a HardwareId to match the expectations for a manufacturer and not generic.

Works for me; @Sean-StarLabs I think you know what to do.

Sean-StarLabs: @superm1 EC does store image on the SPI but it behaves the same way as SuperIO, that it doesn't take effect until reboot. No problems with removing the flag or making it specific, but I'm interested in what you think would go wrong?

superm1: If an external MCU such as an ME or EC is actively using the SPI chip at the same time as OS process you can get a race condition or corruption on the SPI.  This is why for other plugins we have things such as Synaptics MST will stop HDCP and set up quad mode before flashing.

superm1:
You still didn't back out the changes to `fu-flashrom-device.c `

superm1:
I think you might have rebased wrong, there's conflicts now

Sean-StarLabs:
> I think you might have rebased wrong, there's conflicts now

I sure did... want me to re-open for just adding the `AC` quirk? (Rather than leaving the whole thing open whilst waiting for Ubuntu)

superm1:
sure

contrib/debian/rules
Sean-StarLabs: It's done!
https://bugs.launchpad.net/ubuntu/+source/flashrom/+bug/1912371

PR #2788, opened 2021-01-19
Wip/hughsie/lockdown

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -41,6 +41,7 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "tpm");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "tpm_eventlog");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "dell");
+	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "linux_lockdown");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_CONFLICTS, "uefi"); /* old name */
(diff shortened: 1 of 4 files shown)


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -41,6 +41,7 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "tpm");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "tpm_eventlog");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "dell");
+	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "linux_lockdown");
```
superm1, line 44: I can see how this matters for flashrom and superio, but not for uefi capsule.  Care to elaborate?

superm1 (approved):
w/ UEFI capsule removed it LGTM

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: Selection bias; if lockdown is enabled the superio update won't be offered and we'll only get reports where the machine is not locked down.

superm1: So it's trying to get reports of machines that use uefi capsule + superio, but have lockdown enabled?  If that's going to be a big combination for the future the superio plugin should probably be ported to run in UEFI instead.

hughsie: > So it's trying to get reports of machines that use uefi capsule + superio, but have lockdown enabled

Exactly that.

> the superio plugin should probably be ported to run in UEFI instead

That's my plan. I might have some more headcount to work on fwupd in the future, and I thought this would be a good self-contained task. But before that, I need justification :)

superm1: OK..

PR #2789, opened 2021-01-19
Add the PCRs to the report metadata

hughsie:
Although we could reconstruct the PCRn metadata on the LVFS from the (already
included) event log, it's much more scalable if the client just provides the
data that we want to filter by.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
+++ b/plugins/tpm-eventlog/fu-plugin-tpm-eventlog.c
@@ -68,6 +68,18 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 		const gchar *csum = g_ptr_array_index (data->pcr0s, i);
 		fu_device_add_checksum (FU_DEVICE (dev), csum);
 	}
+	for (guint i = 0; i < data->pcr0s->len; i++) {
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Are you only wanting to add PCR0's or others?

hughsie:
> Are you only wanting to add PCR0's or others?

Well, I only wanted to filter by PCR0 on the LVFS, but I figured the others might be useful. If you think restrict to 0, that's fine by me too.

superm1:
>Well, I only wanted to filter by PCR0 on the LVFS, but I figured the others might be useful. If you think restrict to 0, that's fine by me too.

Right now it's only getting the PCR0's.  I don't know that getting the others will be super helpful, depends on what you want to do in the long term :)

hughsie:
> Right now it's only getting the PCR0

Ahh, yes, I understand now; agree. In which case my patch is wrong, instead of getting array-of-PCRs i'm actually getting array-of-PCR0 where there might be a SHA1 or SHA256 or both. Will respin.

hughsie:
@superm1 okay, code review FTW. New version uploaded.

superm1 (approved):

PR #2790, opened 2021-01-19
Install the UX data into a single .tar.gz file

hughsie:
This will provide much better compression (-20%) than gzip on them individually
and also allows us to build the capsule UX images as part of the build stage.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 not sure why I didn't think of this before; what do you think of this approach? There's some more simplification needed in make-images.py but I didn't want to sink time in it if you hated the idea.

superm1:
I love the idea!

superm1 (commented):
`usr/share/locale` can be removed from `contrib/debian/fwupd.install` IIRC now

plugins/uefi-capsule/make-images.py
```diff
@@ -6,177 +6,153 @@
...
+            continue
+        yield x
+    yield "en"
+
+
+def main(args) -> int:
+
+    domain = "fwupd"
+    gettext.textdomain(domain)
+    gettext.install(domain, args.localedir)
```
superm1, line 42: do you still need this?

plugins/uefi-capsule/meson.build
```diff
@@ -97,7 +97,22 @@ install_data(['uefi_capsule.conf'],
...
+custom_target('ux-capsule-tar-gz',
+  input : [
+    join_paths(meson.source_root(), 'po', 'LINGUAS'),
+  ],
+  output : 'ux-capsule.tar.gz',
+  command : [
+    python3.path(),
+    join_paths(meson.current_source_dir(), 'make-images.py'),
+    '--linguas', '@INPUT@',
+    '--localedir', localedir,
```
superm1, line 109: do you still need this?

plugins/uefi-capsule/make-images.py
hughsie: Urgh, yes, and no. If you ninja install then run the script then the .mo files exist and it works. From a clean checkout this gets run before the .mo files get created. From a 40,000ft view its insane we're creating a BMP file from a .mo file from a .po file, and I think we can skip out the middle step.

superm1 (approved):

hughsie:
Slight change; I switched from gzip to xz -- it took the archive size from 2.6Mb to 1.1Mb! I also added in a few more useful screen sizes.

PR #2795, opened 2021-01-20
Wip/hughsie/switch branch back to default

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2569,6 +2569,11 @@ fu_util_esp_list (FuUtilPrivate *priv, gchar **values, GError **error)
 	return TRUE;
 }
 
+static gboolean
+_g_str_equal0 (gconstpointer str1, gconstpointer str2)
+{
+	return g_strcmp0 (str1, str2) == 0;
+}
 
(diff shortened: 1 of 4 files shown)


hughsie:
@Sean-StarLabs FYI

Sean-StarLabs:
PR allows switching back to "default" branch. Without it, `No releases found`

src/fu-tool.c
```diff
@@ -2569,6 +2569,11 @@ fu_util_esp_list (FuUtilPrivate *priv, gchar **values, GError **error)
 	return TRUE;
 }
 
+static gboolean
+_g_str_equal0 (gconstpointer str1, gconstpointer str2)
+{
+	return g_strcmp0 (str1, str2) == 0;
+}
```
superm1, line 2576: to save a few lines of duplication maybe this shouldn't be static and in `fu-util-common.c`?

hughsie: I'm lobbying to get it in Glib. :)

PR #2797, opened 2021-01-21
Fix linking to fwupdplugin from out of tree

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

This is based upon work by @smcv at https://salsa.debian.org/efi-team/fwupd/-/merge_requests/4/


hughsie:
@superm1 what about something like this^^^? It's how other projects like GLib do it. If that's acceptable just fixup my commit into yours.

PR #2800, opened 2021-01-23
esp-list: allow external ESP again

sgn:
In fwupd 1.5.1 and before, we allowed ESP on external device.
From 56d816a5, (Fall back to FAT32 internal partitions for detecting
ESP, 2020-11-11), we started to only consider internal devices only.

While it would be desirable to only consider internal devices for
fallback esp partition, there're some setup that put ESP on external
device, e.g. full disk encryption with /boot on a USB.

Let's allow external ESP again.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -2471,6 +2471,7 @@ fu_util_prompt_for_volume (GError **error)
 {
 	FuVolume *volume;
 	guint idx;
+	gboolean is_fallback = FALSE;
 	g_autoptr(GPtrArray) volumes = NULL;
 	g_autoptr(GPtrArray) volumes_vfat = g_ptr_array_new ();
 	g_autoptr(GError) error_local = NULL;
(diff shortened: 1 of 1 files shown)


src/fu-tool.c
```diff
@@ -2471,27 +2471,29 @@ fu_util_prompt_for_volume (GError **error)
...
 	guint idx;
+	gboolean internal_vfat_only = FALSE;
 	g_autoptr(GPtrArray) volumes = NULL;
 	g_autoptr(GPtrArray) volumes_vfat = g_ptr_array_new ();
 	g_autoptr(GError) error_local = NULL;
 
 	/* exactly one */
 	volumes = fu_common_get_volumes_by_kind (FU_VOLUME_KIND_ESP, &error_local);
 	if (volumes == NULL) {
+		internal_vfat_only = TRUE;
```
hughsie, line 2482: I think something like `is_fallback = TRUE` might be clearer. With that change LGTM.

Issue #2801, opened 2021-01-23
Failed to run update on reboot


physkets:
**Describe the bug**
The system restarts after performing a `$ fwupdmgr update`, but does not perform the update. Upon checking the update history, this is what it says in the relevant section of the relevant firmware, the update of which was attempted:
```shell
$ fwupdmgr get-history
  │   Update State:       failed
  │   Update Error:       failed to run update on reboot
```

**Steps to Reproduce**
I am using the system with SecureBoot enabled, and my own keys. I have set `DisableShimForSecureBoot=true` in `/etc/fwupd/uefi_capsule.conf`.

**Expected behavior**
Upon restart, fwupd should initiate a UEFI flash.

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version
client version: 1.5.5
compile-time dependency versions
        gusb:   0.3.5

Unknown                  [***************************************]
daemon version: 1.5.5
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):  
`pacman`

**fwupd device information**

```shell
$ fwupdmgr get-devices --show-all-devices
20UECTO1WW
│
├─AMD Ryzen 7 PRO 4750U with Radeon Graphics:
│     Device ID:          4bde70ba4e39b28f9eab1628f9dd6e6244c03027
│     Vendor:             AMD
│     GUIDs:              b9a2dd81-159e-5537-a7db-e7101d164d3f
│                         997104da-ec11-5e04-9d94-ac5501fbf609
│                         1f2e0c2f-e078-5fcb-a8de-f78eff9a470a
│                         5fd5eb41-dd4a-5742-8448-861eaaa61cf8
│     Device Flags:       • Internal device
│   
├─SAMSUNG MZVLB512HBJQ-000L7:
│     Device ID:          71b677ca0f1bc2c5b804fa1d59e52064ce589293
│     Summary:            NVM Express Solid State Drive
│     Current version:    4M2QEXF7
│     Vendor:             Samsung Electronics Co Ltd (NVME:0x144D)
│     GUIDs:              0b4d773a-7ac3-58c1-a541-e22ef1cdfe02
│                         c9d531ea-ee7d-5562-8def-c64d0d144813
│                         6e54c992-d302-59ab-b454-2d26ddd63e6d
│                         47335265-a509-51f7-841e-1c94911af66b
│                         f4ceb2a8-09c4-55c0-af2a-13ccd63925b1
[...]

PR #2802, opened 2021-01-23
uefi-capsule: Check if the fwupd BootXXXX entry exists on failure

hughsie:
Some systems remove the BootXXXX entry we add (so we can run fwupdx64.efi) and
thus the firmware update does not run. Most commonly this failure is seen with
Lenovo systems that call the helpful option 'Boot Order Lock'.

Hopefully when we depend on the new kernel bios interface sysfs API in we can
check in ->prepare(), not after reboot, but until that we can mark the update
failure as transient as the user can actually fix the problem themselves.

Fixes https://github.com/fwupd/fwupd/issues/2801

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -16,6 +16,7 @@
 #include "fu-plugin-vfuncs.h"
 
 #include "fu-uefi-bgrt.h"
+#include "fu-uefi-bootmgr.h"
 #include "fu-uefi-common.h"
 #include "fu-uefi-device.h"
 #include "fu-efivar.h"
@@ -44,6 +45,10 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "linux_lockdown");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_CONFLICTS, "uefi"); /* old name */
 	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+
+	/* for the uploaded report */
+	if (fu_plugin_has_custom_flag (plugin, "use-legacy-bootmgr-desc"))
+		fu_plugin_add_report_metadata (plugin, "BootMgrDesc", "legacy");
 }
 
 void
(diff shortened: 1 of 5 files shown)


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -70,13 +71,30 @@ fu_plugin_get_results (FuPlugin *plugin, FuDevice *device, GError **error)
...
 	/* trivial case */
 	if (status == FU_UEFI_DEVICE_STATUS_SUCCESS) {
 		fu_device_set_update_state (device, FWUPD_UPDATE_STATE_SUCCESS);
 		return TRUE;
 	}
 
+	/* check if something rudely removed our BOOTXXXX entry */
+	if (!fu_uefi_bootmgr_verify_fwupd (&error_local)) {
+		/* Lenovo, grr */
+		if (fu_plugin_check_hwid (plugin, "6de5d951-d755-576b-bd09-c5cf66b27234")) {
```
superm1, line 85: should this be applied to the device via a quirk perhaps?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -70,13 +71,30 @@ fu_plugin_get_results (FuPlugin *plugin, FuDevice *device, GError **error)
...
+			g_prefix_error (&error_local,
+					"efibootmgr entry missing; "
```
superm1, line 87: `s/efibootmgr/Boot/`

physkets:
Won't this message only be displayed after the reboot, and only if the user checks `fwupdmgr get-history`? (Because even with the 'Boot Order Lock', the entry is added before the reboot, but removed at some point in the start-up, no?)

If so, it won't draw their attention to the problem immediately. Maybe an instruction in the new addition to the UEFI bug reporting format should suggest checking that?

hughsie:
> Won't this message only be displayed after the reboot, and only if the user checks fwupdmgr get-history?

Nah, the daemon calls get_results at startup for any device that's in `FWUPD_UPDATE_STATE_NEEDS_REBOOT` status (which the UEFI device will be).

> If so, it won't draw their attention to the problem immediately. 

I think the update message should be displayed in the get-devices output too.

> Maybe an instruction in the new addition to the UEFI bug reporting format should suggest checking that?

I added something; ideally I want to use the new kernel BIOS interface for finding this out before we schedule the update, but that's a few weeks away from being testable.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: Fixed, thanks.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: Good idea, I moved `use-legacy-bootmgr-desc` to the quirk file too.

superm1 (approved):

PR #2805, opened 2021-01-25
libfwupd: Add fwupd_release_add_location()

hughsie:
The metadata might want to pass more than one location URI to the client, for
instance if the file is available from more than one HTTP mirror.

Use the noun of location to match the AppStream <artifact> naming; this is the
last place where LVFS AppStream diverges from the official specification and
it would be good to bring fwupd back into line -- although the LVFS will have
to write both elements for a very long time.

See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

Also: we're not changing the format of the `Uri` GVariant key to preserve both
forward and backwards compatibility of the library. We can remove it when we
next break API.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-release.c
+++ b/libfwupd/fwupd-release.c
@@ -47,7 +47,7 @@ typedef struct {
 	gchar				*name_variant_suffix;
 	gchar				*summary;
 	gchar				*branch;
-	gchar				*uri;
+	GPtrArray			*locations;
 	gchar				*vendor;
 	gchar				*version;
 	gchar				*remote_id;
@@ -545,18 +545,21 @@ fwupd_release_get_metadata_item (FwupdRelease *release, const gchar *key)
  * fwupd_release_get_uri:
  * @release: A #FwupdRelease
  *
- * Gets the update uri.
+ * Gets the default update URI.
  *
- * Returns: the update uri, or %NULL if unset
+ * Returns: the update URI, or %NULL if unset
  *
  * Since: 0.9.3
+ * Deprecated: 1.5.6: Use fwupd_release_get_locations() instead.
  **/
 const gchar *
 fwupd_release_get_uri (FwupdRelease *release)
 {
 	FwupdReleasePrivate *priv = GET_PRIVATE (release);
 	g_return_val_if_fail (FWUPD_IS_RELEASE (release), NULL);
-	return priv->uri;
+	if (priv->locations->len == 0)
+		return NULL;
+	return (const gchar *) g_ptr_array_index (priv->locations, 0);
 }
 
 /**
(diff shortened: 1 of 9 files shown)


libfwupd/fwupd-release.c
```diff
@@ -545,36 +545,82 @@ fwupd_release_get_metadata_item (FwupdRelease *release, const gchar *key)
...
+ * Sets the update URI, i.e. where you can download the firmware from.
  *
  * Since: 0.9.3
+ * Deprecated: 1.5.6: Use fwupd_release_add_location() instead.
  **/
 void
 fwupd_release_set_uri (FwupdRelease *release, const gchar *uri)
+{
+	g_return_if_fail (FWUPD_IS_RELEASE (release));
+	fwupd_release_add_location (release, uri);
```
superm1, line 579: to really be compatible to the old behavior shouldn't this clear the old locations and just set the one?

hughsie: Yes, indeed. I couldn't see why CI failed; any ideas with that very welcome.

libfwupd/fwupd-self-test.c
```diff
@@ -426,6 +428,10 @@ fwupd_device_func (void)
 		"        \"deadbeef\"\n"
 		"      ],\n"
 		"      \"Size\" : 1024,\n"
+		"      \"Locations\" : [\n"
+		"        \"http://foo.com\",\n"
+		"        \"https://foo.com\"\n"
```
superm1, line 433: since you mention in the changelog that you can have some non-`http(`s)` prefixed URLs, probably should make sure to include one here too.

libfwupd/fwupd-release.c
```diff
@@ -545,36 +545,82 @@ fwupd_release_get_metadata_item (FwupdRelease *release, const gchar *key)
...
  * @release: A #FwupdRelease
  *
- * Gets the update uri.
+ * Gets the default update URI.
  *
- * Returns: the update uri, or %NULL if unset
+ * Returns: the update URI, or %NULL if unset
  *
  * Since: 0.9.3
+ * Deprecated: 1.5.6: Use fwupd_release_get_locations() instead.
```
superm1, line 553: you need to upgrade `src/fu-tool.c` too (And we probably need CI to be yelling about the fact you missed it)

superm1: presumably `libfwupd/fwupd-client.c` should be too?

libfwupd/fwupd-self-test.c
hughsie: Done, thanks.

libfwupd/fwupd-release.c
hughsie: > we probably need CI to be yelling about the fact you missed it

https://github.com/fwupd/fwupd/pull/2807

> presumably libfwupd/fwupd-client.c should be too?

All fixed, thanks.

libfwupd/fwupd-release.c
hughsie: Doh, CI fixed.

superm1 (approved):

PR #2808, opened 2021-01-26
Allow allow downloading from a list of possible URLs

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -70,6 +70,7 @@ typedef struct {
 
 #ifdef HAVE_LIBCURL
 typedef struct {
+	GPtrArray			*urls;
 	CURL				*curl;
 	curl_mime			*mime;
 	struct curl_slist		*headers;
@@ -118,6 +119,8 @@ fwupd_client_curl_helper_free (FwupdCurlHelper *helper)
 		curl_mime_free (helper->mime);
 	if (helper->headers != NULL)
 		curl_slist_free_all (helper->headers);
+	if (helper->urls != NULL)
+		g_ptr_array_unref (helper->urls);
 	g_free (helper);
 }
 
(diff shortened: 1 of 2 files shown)


libfwupd/fwupd-client.c
```diff
@@ -2473,7 +2476,7 @@ fwupd_client_install_release_download_cb (GObject *source, GAsyncResult *res, gp
 }
 
 static gboolean
-fwupd_client_is_url (const gchar *perhaps_url)
+fwupd_client_is_url_http (const gchar *perhaps_url)
```
superm1, line 2479: doesn't this function also handle directories?

hughsie: I don't think so; `fwupd_client_install_release_remote_cb` uses it as `fwupd_remote_get_kind (remote) == FWUPD_REMOTE_KIND_LOCAL && !fwupd_client_is_url_http (fwupd_release_get_uri (data->release)` i.e. a directory would return `FALSE` from `fwupd_client_is_url()`

libfwupd/fwupd-client.c
```diff
@@ -4228,56 +4231,131 @@ fwupd_client_download_write_callback_cb (char *ptr, size_t size, size_t nmemb, v
...
 	g_task_return_pointer (task,
-			       g_byte_array_free_to_bytes (g_steal_pointer (&buf)),
+			       g_steal_pointer (&blob),
 			       (GDestroyNotify) g_bytes_unref);
 }
 #endif
 
+/* private */
+void
+fwupd_client_download_bytes2_async (FwupdClient *self,
```
superm1, line 4319: presumably this should change with the ABI bump and not be the private one anymore.

superm1 (approved):

libfwupd/fwupd-client.c
hughsie: Yes, probably; let's see if we need any more parameters before we export it.

PR #2815, opened 2021-01-27
dfu: Add support for GD32VF103 as found in the Longan Nano

hughsie:
This bootloader is *weird* -- the chip ID is the first two bytes of the serial
number and the data is offset and encoded in UTF-8, not UTF-16.

The sector information is also wrong. Gah!

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/dfu-device.c
+++ b/plugins/dfu/dfu-device.c
@@ -1227,6 +1227,53 @@ dfu_device_open (FuUsbDevice *device, GError **error)
 		priv->status = DFU_STATUS_OK;
 	}
 
+	/* GD32VF103 encodes the serial number in UTF-8 (rather than UTF-16)
+	 * and also uses the first two bytes as the model identifier */
+	if (fu_device_has_custom_flag (FU_DEVICE (device), "gd32")) {
+#if G_USB_CHECK_VERSION(0,3,6)
+		GUsbDevice *usb_device = fu_usb_device_get_dev (device);
+		const guint8 *buf;
+		gsize bufsz = 0;
+		guint16 langid = G_USB_DEVICE_LANGID_ENGLISH_UNITED_STATES;
+		guint8 idx = g_usb_device_get_serial_number_index (usb_device);
+		g_autofree gchar *chip_id = NULL;
+		g_autofree gchar *serial_str = NULL;
+		g_autoptr(GBytes) serial_blob = NULL;
+		serial_blob = g_usb_device_get_string_descriptor_bytes (usb_device,
+									idx,
+									langid,
+									error);
+		if (serial_blob == NULL)
+			return FALSE;
+		if (g_getenv ("FWUPD_DFU_VERBOSE") != NULL)
+			fu_common_dump_bytes (G_LOG_DOMAIN, "GD32 serial", serial_blob);
+		buf = g_bytes_get_data (serial_blob, &bufsz);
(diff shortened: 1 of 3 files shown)


plugins/dfu/dfu-device.c
```diff
@@ -1227,6 +1228,51 @@ dfu_device_open (FuUsbDevice *device, GError **error)
 		priv->status = DFU_STATUS_OK;
 	}
 
+	/* GD32VF103 encodes the serial number in UTF-8 (rather than UTF-16)
+	 * and also uses the first two bytes as the model identifier */
+	if (fu_device_has_custom_flag (FU_DEVICE (device), "gd32")) {
+#if G_USB_CHECK_VERSION(0,3,6)
```
superm1, line 1234: probably want to set subprojects to 0.3.6

plugins/dfu/dfu-device.c
```diff
@@ -1227,6 +1228,51 @@ dfu_device_open (FuUsbDevice *device, GError **error)
...
+		dfu_device_set_chip_id (self, chip_id);
+
+		/* serial number follows */
+		serial_str = g_strndup ((const gchar *) buf + 2, bufsz - 2);
+		fu_device_set_serial (FU_DEVICE (device), serial_str);
+#else
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "GUsb version %s too old to support GD32",
```
superm1, line 1270: thinking like a user that would encounter this, perhaps you want to tell them the minimum version they need in the error?

plugins/dfu/dfu-device.c
hughsie: you okay with a git commit? 0.3.6 isn't out yet.

plugins/dfu/dfu-device.c
hughsie: something like

			     "GUsb version %s too old to support GD32, "
			     "fwupd needs to be rebuit against 0.3.6",

plugins/dfu/dfu-device.c
superm1: oh hmm, probably better to wait until 0.3.6 then.  don't want to other unstable /untested stuff to push out from subprojecting.

plugins/dfu/dfu-device.c
superm1: "...or later"

plugins/dfu/dfu-device.c
```diff
@@ -1216,6 +1216,7 @@ dfu_device_open (FuUsbDevice *device, GError **error)
 	DfuDevice *self = DFU_DEVICE (device);
 	DfuDevicePrivate *priv = GET_PRIVATE (self);
 	GPtrArray *targets = dfu_device_get_targets (self);
+	GUsbDevice *usb_device = fu_usb_device_get_dev (device);
```
superm1, line 1219: this seems it should go into the scope of the `if` loop

hughsie: Ack, moved.

plugins/dfu/dfu-device.c
hughsie: ahh yes, good catch. Fixed.

superm1 (approved):

PR #2816, opened 2021-01-27
Allow downloading firmware from IPFS

hughsie:
Only a small amount of the firmware on the LVFS will be available, and IPFS is
opt-in only, requiring the user to use --ipfs on the command line.

The IPFS daemon and command line client will need to be installed manually if
this option is used.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


src/fu-util.c
```diff
@@ -2778,6 +2780,9 @@ main (int argc, char *argv[])
 		{ "disable-ssl-strict", '\0', 0, G_OPTION_ARG_NONE, &priv->disable_ssl_strict,
 			/* TRANSLATORS: command line option */
 			_("Ignore SSL strict checks when downloading files"), NULL },
+		{ "ipfs", '\0', 0, G_OPTION_ARG_NONE, &enable_ipfs,
+			/* TRANSLATORS: command line option */
+			_("Use IPFS when downloading files"), NULL },
```
superm1, line 2785: Is the average fwupdmgr user going to know what IPFS is?

superm1 (commented):
I actually feel like we want this to be a config flag.
Here's my argument - if you're using IPFS you probably need to make sure that all clients use it.  Just in this project you'll have fwupdtool, fwupdmgr launched as a user, fwupdmgr refresh launched by systemd.

Gnome software won't use IPFS by default, and it would need changing.  I don't think all the clients adding support scales well as a result.  However - instead the daemon can just feed back to the client a URI with the correct URI.

src/fu-util.c
hughsie: It's one short google away -- and I don't think that they'll know what "Ignore SSL strict checks" would do either without some research.

hughsie:
> I actually feel like we want this to be a config flag.

Yes, perhaps. I wasn't sure if it was per-system, per-remote or per-client.

What does make this tricky is that not all firmware is going to be available over IPFS. I think we should _fallback_ to non IPFS where required, so something like `OnlyLocationProtocols=ipfs://` isn't going to work as the client would never get notified of new firmware from Dell.

> Gnome software won't use IPFS by default

Well, it will, as it has neither disable flags. I do agree it would suck to have to configure each client. What about a daemon option `PreferLocationProtocol=ipfs://` or something. That would change the sorting of the locations as added from the new `artifacts` tag and then gnome-software would try IPFS first, and then fall back to HTTP.

superm1:
>Yes, perhaps. I wasn't sure if it was per-system, per-remote or per-client.

Per system, but only if you can get a fallback to HTTPS perhaps?

> What does make this tricky is that not all firmware is going to be available over IPFS. I think we should _fallback_ to non IPFS where required, so something like `OnlyLocationProtocols=ipfs://` isn't going to work as the client would never get notified of new firmware from Dell.
> 
> >

Furthermore will you want to have a separate metadata release going onto IPFS?  Or will it be the same metadata that you just can't get a bunch of files from it?  I guess you could potentially have an "IPFS" remote on LVFS side too.

>Well, it will, as it has neither disable flags. I do agree it would suck to have to configure each client. What about a daemon option PreferLocationProtocol=ipfs:// or something. That would change the sorting of the locations as added from the new artifacts tag and then gnome-software would try IPFS first, and then fall back to HTTP.

That seems fine to me if keeping the same metadata for everyone.

hughsie:
>  Or will it be the same metadata that you just can't get a bunch of files from it

I think that; I don't think I want to try and keep any more metadata in sync. It also gets confusing when you'd have testing-ipfs and stable-ipfs.

> PreferLocationProtocol

Kinda sucks as a name tho; what about `ProtocolPriority=ipfs://,https://`. I don't think we need `DisabledProtocols` as installing `ipfs` is a manual thing and won't be available on 99% of all systems. I think returning the results from the daemon in IPFS->HTTP order by default would break old versions of libfwupd that are expecting just the first result as the default. Perhaps the fix for that is ordering them in the (new) libfwupd or maybe making the `Uri` fallback more smart so it only returns the HTTP entry.

superm1:
> Kinda sucks as a name tho; what about ProtocolPriority=ipfs://,https://. I don't think we need DisabledProtocols as installing ipfs is a manual thing and won't be available on 99% of all systems. I think returning the results from the daemon in IPFS->HTTP order by default would break old versions of libfwupd that are expecting just the first result as the default. Perhaps the fix for that is ordering them in the (new) libfwupd or maybe making the Uri fallback more smart so it only returns the HTTP entry.

I guess that means that actually clients do need newer libfwupd to use ipfs if we're "falling back" to https.  So maybe have gnome-software check compile time version of libfwupd and use the locations function if present, and the URI if not.

hughsie:
This works! set `127.0.0.1 fwupd.org` in /etc/hosts, and then:

Screenshot from 2021-02-01 09-46-37

superm1 (approved):

PR #2821, opened 2021-01-28
trivial: Add fu_common_uri_get_scheme() for future use

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -2854,3 +2854,26 @@ fu_common_crc32 (const guint8 *buf, gsize bufsz)
 {
 	return fu_common_crc32_full (buf, bufsz, 0xFFFFFFFF, 0xEDB88320);
 }
+
+/**
+ * fu_common_uri_get_scheme:
+ * @uri: valid URI, e.g. `https://foo.bar/baz`
+ *
+ * Returns the USI scheme for the given URI.
+ *
+ * Returns: scheme value, or %NULL if invalid, e.g. `https`
+ *
+ * Since: 1.5.6
+ **/
+gchar *
+fu_common_uri_get_scheme (const gchar *uri)
+{
+	gchar *tmp;
+
+	g_return_val_if_fail (uri != NULL, NULL);
+
+	tmp = g_strstr_len (uri, -1, ":");
+	if (tmp == NULL || tmp[0] == '\0')
+		return NULL;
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -2854,3 +2854,26 @@ fu_common_crc32 (const guint8 *buf, gsize bufsz)
...
+ **/
+gchar *
+fu_common_uri_get_scheme (const gchar *uri)
+{
+	gchar *tmp;
+
+	g_return_val_if_fail (uri != NULL, NULL);
+
+	tmp = g_strstr_len (uri, -1, ":");
+	if (tmp == NULL || tmp[0] == '\0')
```
superm1, line 2876: as it's a pointer that is returned, shouldn't it be `tmp == NULL || *tmp == '\0'`
(i'll admit functionality equivalent)

libfwupdplugin/fu-common.c
```diff
@@ -2854,3 +2854,26 @@ fu_common_crc32 (const guint8 *buf, gsize bufsz)
...
+		return NULL;
+	return g_strndup (uri, tmp - uri);
```
superm1, line 2878: I think you want to have this always return lowercase

libfwupdplugin/fu-self-test.c
```diff
@@ -350,6 +350,26 @@ fu_common_strsafe_func (void)
...
+fu_common_uri_scheme_func (void)
+{
+	struct {
+		const gchar *in;
+		const gchar *op;
+	} strs[] = {
+		{ "https://foo.bar/baz",	"https" },
+		{ "ftp://",			"ftp" },
+		{ "ftp:",			"ftp" },
+		{ "foobarbaz",			NULL },
```
superm1, line 363: I think you need a case insensitive test

libfwupdplugin/fu-common.c
hughsie: it's technically a `gchar *` pointer, not a gpointer.

libfwupdplugin/fu-common.c
hughsie: Good catch, agree.

libfwupdplugin/fu-self-test.c
hughsie: Fixed.

superm1 (approved):

PR #2824, opened 2021-01-28
Allow configuring URI scheme preferences

hughsie:
This would allow, for instance, only allowing IPFS downloads or preferring
https over http downloads.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
Ooops, this of course depends on https://github.com/fwupd/fwupd/pull/2821

superm1 (approved):

PR #2828, opened 2021-01-29
cros-ec: Make _do_xfer() outbuf const

hughsie:
The g_usb_device_bulk_transfer() function requires a mutable 'data' buffer, but
fwupd is sending file data and never wants the buffers modified.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


bleungatchromium (approved):
Thanks for the fix!

PR #2831, opened 2021-01-29
Wip/hughsie/dfu fu chunk

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 not tested on any actual devices yet, but I'm pretty happy with how this cleanup looks. Some sanity checks much appreciated tho. I'll fuzz for a few hours now too.

libfwupdplugin/fu-dfuse-firmware-image.c
```diff
@@ -0,0 +1,203 @@
...
+#include "fu-chunk-private.h"
+#include "fu-common.h"
+
+#include "fu-dfuse-firmware-image.h"
+
+static void fu_dfuse_firmware_image_finalize			 (GObject *object);
+
+typedef struct {
+	GPtrArray		*chunks;
+	gchar			 name[255];
```
superm1, line 30: this seems a bit wasteful.  can't you just create it as needed and check the bounds on set?

superm1 (approved):
some funky handling for name but nothing else stands out to me.

libfwupdplugin/fu-dfuse-firmware-image.c
hughsie: I can do better than that; the only reason we were copying the 255 bytes of junk was because we were round-tripping that nonfree dfuse file. Now we can build our own, we don't have to blindly copy the array as we know it's all NUL. I folded the name into the FuFirmwareImage::id and just moved the chunking there too -- removing half a dozen of the new symbols in the process! Review of that new code very welcome.

superm1 (approved):

Issue #2836, opened 2021-02-01
install_dell_bios_exe.py: ValueError: Namespace Fwupd not available


paulmenzel:
On Debian sid/unstable, trying to follow https://github.com/fwupd/firmware-dell/blob/master/README.md, the command fails.

```
$ dpkg -l fwupd
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-=================================
ii  fwupd          1.5.5-2      amd64        Firmware update daemon
$ fwupdmgr --version
client version:	1.5.5
compile-time dependency versions
	gusb:	0.3.5

daemon version:	1.5.5
$ sudo /usr/share/fwupd/install_dell_bios_exe.py XPS_13_9310_2-in-1_2.0.0.exe
Traceback (most recent call last):
  File "/usr/share/fwupd/install_dell_bios_exe.py", line 13, in <module>
    gi.require_version('Fwupd', '2.0')
  File "/usr/lib/python3/dist-packages/gi/__init__.py", line 126, in require_version
    raise ValueError('Namespace %s not available' % namespace)
ValueError: Namespace Fwupd not available
```

superm1:
Do you have the gir package installed too?

paulmenzel:
> Do you have the `gir1.2-fwupd-2.0` package installed too?

It was not installed. `apt install gir1.2-fwupd-2.0` fixed the issue. It’s not listed in the package metadata:

    Recommends: python3, bolt, dbus, secureboot-db, udisks2, fwupd-signed

There is no `Suggests` entry.

superm1:
Yeah we don't want it installed by default, so not in `Recommends` I think is the correct solution.  Suggests seems like a good consolation.

paulmenzel:
Sounds good.

Additionally, is there a way to catch this error, and print instructions to install the required packages?

superm1:
The package name is going to vary from OS to OS, I guess we can list a few things to try though maybe.

paulmenzel:
Indeed. Or link to some Wiki page or so.

PR #2837, opened 2021-02-01
trivial: Make it easier to know what packages to install to let

superm1:
helper scripts run.

Fixes: #2836

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [x] Documentation

### DIFF
--- a/contrib/firmware_packager/install_dell_bios_exe.py
+++ b/contrib/firmware_packager/install_dell_bios_exe.py
@@ -10,7 +10,11 @@ import sys
 import tempfile
 import gi
 
-gi.require_version('Fwupd', '2.0')
+try:
+    gi.require_version('Fwupd', '2.0')
+except ValueError:
+    print("Missing gobject-introspection packages.  Try to install gir1.2-fwupd-2.0.")
+    sys.exit(1)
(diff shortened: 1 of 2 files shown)


contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -10,7 +10,11 @@
 import tempfile
 import gi
 
-gi.require_version('Fwupd', '2.0')
+try:
+    gi.require_version('Fwupd', '2.0')
+except ValueError:
+    print("Missing gobject-introspection packages.  Try to install gir1.2-fwupd-2.0 or fwupd-devel.")
```
hughsie, line 16: I think for Fedora the typelib is in the main fwupd package:

    $ rpm -qf /usr/lib64/girepository-1.0/Fwupd-2.0.typelib
    fwupd-1.5.6-0.705.20210120git.fc33.x86_64

The GIR is indeed in the -devel package, but I don't think you need that for `gi`

superm1: OK dropped that.  I guess this really only is a Debian/Ubuntu problem then since it's not installed by default there.

hughsie (approved):

PR #2839, opened 2021-02-04
uefi-capsule: Add SBAT metadata to the fwupd EFI binary

hughsie:
The Secure Boot Advanced Targeting (SBAT) [0] is a Generation Number Based
Revocation mechanism that is meant to replace the DBX revocation file list.

Binaries must contain a .sbat data section that has a set entries, each of
them consisting of UTF-8 strings as comma separated values. Allow to embed
this information into the fwupd EFI binary at build time.

The SBAT metadata must contain at least two entries. One that defines the
SBAT version used and another one that defines the component generation.

Downstream users can add additional entries if have changes that make them
diverge from the upstream code and potentially add other vulnerabilities.

[0]: https://github.com/rhboot/shim/blob/sbat/SBAT.md


contrib/debian/rules
```diff
@@ -22,13 +22,25 @@ ifeq (yes,$(shell dpkg-vendor --derives-from Ubuntu && echo yes))
 	SB_STYLE := ubuntu
 	tar_name := fwupd_$(deb_version)_$(DEB_HOST_ARCH).tar.gz
 	CONFARGS += -Dplugin_flashrom=false
+	CONFARGS += -Defi_sbat_distro_id=ubuntu
+	CONFARGS += -Defi_sbat_distro_generation=1
+	CONFARGS += -Defi_sbat_distro_summary=Ubuntu
+	CONFARGS += -Defi_sbat_distro_pkgname=fwupd
+	CONFARGS += -Defi_sbat_distro_version=$(deb_version)
+	CONFARGS += -Defi_sbat_distro_url="https://packages.ubuntu.com/search?keywords=fwupd"
```
superm1, line 30: Better URL: https://launchpad.net/ubuntu/+source/fwupd

contrib/debian/rules
```diff
@@ -22,13 +22,25 @@ ifeq (yes,$(shell dpkg-vendor --derives-from Ubuntu && echo yes))
...
 		CONFARGS += -Dplugin_flashrom=true
 	else
 		CONFARGS += -Dplugin_flashrom=false
 	endif
+	CONFARGS += -Defi_sbat_distro_id=debian
+	CONFARGS += -Defi_sbat_distro_generation=1
+	CONFARGS += -Defi_sbat_distro_summary=Debian
+	CONFARGS += -Defi_sbat_distro_pkgname=fwupd
+	CONFARGS += -Defi_sbat_distro_version=$(deb_version)
+	CONFARGS += -Defi_sbat_distro_url="https://packages.debian.org/search?keywords=fwupd"
```
superm1, line 43: better URL: https://tracker.debian.org/pkg/fwupd

meson_options.txt
```diff
@@ -43,4 +43,12 @@ option('efi-libdir', type : 'string', description : 'path to the EFI lib directo
 option('efi-ldsdir', type : 'string', description : 'path to the EFI lds directory')
 option('efi-includedir', type : 'string', value : '/usr/include/efi', description : 'path to the EFI header directory')
 option('efi_os_dir', type: 'string', description : 'the name of OS directory in ESP')
+option('efi_sbat_spec_version', type : 'integer', value : 1, description : 'SBAT specification version')
```
superm1, line 46: spec is starting at 1?  I thought in @martinezjavier PR it was start 0

meson_options.txt
```diff
@@ -43,4 +43,12 @@ option('efi-libdir', type : 'string', description : 'path to the EFI lib directo
...
+option('efi_sbat_fwupd_generation', type : 'integer', value : 1, description : 'SBAT fwupd generation')
+option('efi_sbat_distro_id', type : 'string', value : '', description : 'SBAT distribution ID, e.g. fedora')
+option('efi_sbat_distro_generation', type : 'integer', value : 1, description : 'SBAT distribution generation')
```
superm1, line 49: starting at 1"

contrib/debian/rules
hughsie: Ack, ta.

contrib/debian/rules
hughsie: Ack.

contrib/debian/rules
```diff
@@ -22,13 +22,25 @@ ifeq (yes,$(shell dpkg-vendor --derives-from Ubuntu && echo yes))
...
+	CONFARGS += -Defi_sbat_distro_version=$(deb_version)
+	CONFARGS += -Defi_sbat_distro_url="https://packages.ubuntu.com/search?keywords=fwupd"
 else
 	TMPLDIR  := debian/fwupd-$(DEB_HOST_ARCH)-signed-template/usr/share/code-signing/fwupd-$(DEB_HOST_ARCH)-signed-template
 	ifneq ($(DEB_HOST_ARCH_CPU),ia64)
 		CONFARGS += -Dplugin_flashrom=true
 	else
 		CONFARGS += -Dplugin_flashrom=false
 	endif
+	CONFARGS += -Defi_sbat_distro_id=debian
```
superm1, line 38: I wonder if it's worth just parsing /etc/os-release directly in case a derivative decides to rebuild fwupd they don't end up with claiming to be debian.

meson_options.txt
hughsie: We went back and forth on this; I think it will indeed start at 1.

contrib/debian/rules
hughsie: What I want to happen is the spec changes to an AppStream ID, so rather than `fwupd.fedora` we'd have the java-style `org.fedoraproject.fwupd` if I can convince @vathpela that it's a good idea. And that, we can't get from the os-release file without some serious futzing.

plugins/uefi-capsule/efi/meson.build
```diff
@@ -154,6 +154,11 @@ so = custom_target('fwup.so',
                              efi_ldflags + ['@INPUT@'] +
                              ['-lefi', '-lgnuefi', libgcc_file_name])
 
+# sanity check the packager set this to *SOMETHING*
+if get_option('efi_sbat_distro_id') == '' and get_option('supported_build')
```
superm1, line 158: shouldn't they also need to set `efi_sbat_distro_generation` and the others explicitly?

hughsie: I was going to allow fallbacks to be pragmatic. e.g. falling back to `sbat_fwupd_generation` -- I also wasn't sure if `sbat_distro_pkgname` was always required; if it's a yocto build for instance there is no "package" just a blob of the finished image.

superm1: OK..

meson_options.txt
```diff
@@ -43,4 +43,12 @@ option('efi-libdir', type : 'string', description : 'path to the EFI lib directo
...
+option('efi_sbat_distro_summary', type : 'string', value : '', description : 'SBAT distribution summary, e.g. Fedora')
+option('efi_sbat_distro_pkgname', type : 'string', value : '', description : 'SBAT distribution package name, e.g. grub2')
+option('efi_sbat_distro_version', type : 'string', value : '', description : 'SBAT distribution version, e.g. 2.04-31.fc33')
+option('efi_sbat_distro_url', type : 'string', value : '', description : 'SBAT distribution URL, e.g. https://redhat.com/rpms/grub2')
```
superm1, line 53: this should be updated for "fwupd" rather than grub2 example

contrib/PKGBUILD
```diff
@@ -27,7 +27,14 @@ build() {
...
     fi
-    arch-meson -D b_lto=false $CI ../build -Dsupported_build=true
+    arch-meson -D b_lto=false $CI ../build \
+        -Defi_sbat_distro_id="arch" \
+        -Defi_sbat_distro_generation=1 \
+        -Defi_sbat_distro_summary="Arch Linux" \
+        -Defi_sbat_distro_pkgname="${pkgname}" \
+        -Defi_sbat_distro_version="$(pkgver)" \
+        -Defi_sbat_distro_url="https://archlinux.org/packages/community/x86_64/${pkgname}/" \
+        -Dsupported_build=true
```
hughsie, line 37: @superm1 what do you think these values should be for the snap package? Maybe we just don't write a distro section?

superm1: Well in the snap package it's actually pulling the Ubuntu signed binary at the moment...

contrib/debian/rules
vathpela: My concern here is that we need the `SBAT` UEFI variable to be /small/, and I already think the strings we've currently got are pushing it.  If we wanted to use something more verbose in the `.sbat` section, we probably need a mapping table somewhere.  Which means we want some canonical source of the aliases that wind up in `SBAT` and all that goes with that.

superm1: If we want it to be small - why bother including a URL in the SBAT field?  That is certainly the biggest part right now, and people should be able to encode that metadata elsewhere from the actual binary itself.

meson_options.txt
vathpela: I was also thinking it would be 0... I don't think we've landed on an official answer yet.

contrib/PKGBUILD
hughsie: Ahh, well that makes things simpler then! Thanks.

meson_options.txt
```diff
@@ -43,4 +43,12 @@ option('efi-libdir', type : 'string', description : 'path to the EFI lib directo
 option('efi-ldsdir', type : 'string', description : 'path to the EFI lds directory')
 option('efi-includedir', type : 'string', value : '/usr/include/efi', description : 'path to the EFI header directory')
 option('efi_os_dir', type: 'string', description : 'the name of OS directory in ESP')
+option('efi_sbat_spec_version', type : 'integer', value : 1, description : 'SBAT specification version')
+option('efi_sbat_fwupd_generation', type : 'integer', value : 1, description : 'SBAT fwupd generation')
+option('efi_sbat_distro_id', type : 'string', value : '', description : 'SBAT distribution ID, e.g. fedora')
+option('efi_sbat_distro_generation', type : 'integer', value : 1, description : 'SBAT distribution generation')
+option('efi_sbat_distro_summary', type : 'string', value : '', description : 'SBAT distribution summary, e.g. Fedora')
+option('efi_sbat_distro_pkgname', type : 'string', value : '', description : 'SBAT distribution package name, e.g. grub2')
+option('efi_sbat_distro_version', type : 'string', value : '', description : 'SBAT distribution version, e.g. 2.04-31.fc33')
```
hughsie, line 52: @vathpela is this supposed to be `fwupd-1.5.6.fc33.x86_64` or `fwupd-1.5.6.fc33`?

meson_options.txt
hughsie: Indeed, thanks.

contrib/debian/rules
hughsie: > we need the SBAT UEFI variable to be /small

The URLs are the biggest strings by far.

vathpela: Those aren't in the variable at all.

hughsie: Ohh, I see, sorry. My brain assumed the entire .sbat section would be in the variable, but that's not the case at all.

superm1: Mine made the same conclusion until I re-read the comment above.

meson_options.txt
superm1: Unless you apply different patches to different architectures I would think it should be the latter.

plugins/uefi-capsule/efi/generate_binary.py
```diff
@@ -72,6 +128,57 @@ def _run_genpeimg(args):
...
+        "--project-name",
+        help="SBAT project name",
+    )
+    parser.add_argument(
+        "--project-version",
+        help="SBAT project version",
+    )
+    parser.add_argument(
+        "--sbat-version",
+        default=0,
```
superm1, line 141: whatever the conclusion is for default to 0 or 1 index this might need updating.

plugins/uefi-capsule/efi/generate_binary.py
```diff
@@ -72,6 +128,57 @@ def _run_genpeimg(args):
...
+        type=int,
+        help="SBAT version",
+    )
+    parser.add_argument(
+        "--sbat-generation",
+        default=0,
```
superm1, line 147: whatever the conclusion is for default to 0 or 1 index this might need updating.

contrib/debian/rules
hughsie: > My concern here is that we need the SBAT UEFI variable to be /small/

Could we define three letter shorthands for the different distros? e.g. RHT = Red Hat, FED = Fedora etc? That obviously needs a "registry" of sorts, but I don't think that's always a bad thing.

meson_options.txt
hughsie: From a 40,000ft view, a specification version 0 doesn't inspire confidence. It certainly says "_prerelease, not ready_" to me.

meson_options.txt
martinezjavier: I would also expect the latter.

contrib/debian/rules
martinezjavier: I think something that could be done is to mention in the spec that the `component_name` should be the existing vendor EFI directory name and the component name. 

i.e `fedora.grub2`, `redhat.fwupd` and so on, which is what @vathpela already has in his proposal. This seems to be a middle ground between a full reverse DNS identification string and just an arbitrary string for the distro. The EFI vendor directory should be exclusive anyways or otherwise two distros would write their EFI binaries in the same directory in the ESP.

meson_options.txt
martinezjavier: I think that agree with @hughsie on this. Specially for the `sbat` entry.

contrib/debian/rules
hughsie: > component_name should be the existing vendor EFI directory name

If we have that in the spec, we can just reuse the existing `-Defi_os_dir` define.

plugins/uefi-capsule/efi/generate_binary.py
hughsie: They decided: revocations start at 0, version starts at 1.

contrib/debian/rules
hughsie: > If we have that in the spec, we can just reuse the existing -Defi_os_dir define.

I've respun the patch to use that.

superm1 (approved):

martinezjavier (approved):

PR #2845, opened 2021-02-05
trivial: Add support for oss-fuzz

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
This took way too long to complete. The complexity is having to build everything static, and to create as tiny fuzzing binaries as possible. It's not possible to use fwupd's meson support without doing horrible, horrible things.

hughsie:
Phew

	INFO: performing bad build checks for /tmp/not-out/dfuse_fuzzer
	INFO: performing bad build checks for /tmp/not-out/bcm57xx_fuzzer
	INFO: performing bad build checks for /tmp/not-out/ihex_fuzzer
	INFO: performing bad build checks for /tmp/not-out/fmap_fuzzer
	INFO: performing bad build checks for /tmp/not-out/srec_fuzzer
	INFO: performing bad build checks for /tmp/not-out/synaprom_fuzzer
	Check build passed.

contrib/ci/oss-fuzz.py
```diff
@@ -0,0 +1,431 @@
...
+        "fu-firmware-common",
+        "fu-firmware-image",
+        "fu-dfu-firmware",
+        "fu-fmap-firmware",
+        "fu-dfuse-firmware",
+        "fu-srec-firmware",
+        "fu-ihex-firmware",
+        "fu-quirks",
+        "fu-volume",
+    ]:
```
superm1, line 311: If there is going to be a different build system for oss-fuzz, can it at least browse directories to find the source files it compiles rather than hardcoding in _another_ place?

superm1:
> This took way too long to complete. The complexity is having to build everything static, and to create as tiny fuzzing binaries as possible. It's not possible to use fwupd's meson support without doing horrible, horrible things.

😭

contrib/ci/oss-fuzz.py
```diff
@@ -0,0 +1,431 @@
...
+from typing import Dict, Optional, List, Union
+
+
+class Builder:
+    def __init__(self) -> None:
+
+        self.cc = self._ensure_environ("CC", "gcc")
+        self.cxx = self._ensure_environ("CXX", "g++")
+        self.builddir = self._ensure_environ("WORK", os.path.realpath("_build"))
+        self.installdir = self._ensure_environ("OUT", os.path.realpath("_build/out"))
```
superm1, line 23: we have `.gitignore` for `build` already, why not re-use that?

superm1 (commented):
is it worth having a CI target that runs less frequently than with every build (probably takes too long) to check that this thing keeps working?

contrib/ci/oss-fuzz.py
hughsie: I do wonder if we can use `meson introspect` -- I do agree it sucks to have the list in two places.

contrib/ci/oss-fuzz.py
hughsie: Fair point, will fix.

hughsie:
> is it worth having a CI target

If I can get this in master, and then merge the oss-fuzz side, I believe we can then hook into the Google CI system so we get  failures.

contrib/ci/oss-fuzz.py
hughsie: I had a play with this; `meson introspect meson.build --projectinfo` gives us the project version (which we don't actually need, as FWUPD_CHECK_VERSION always returns `1`), but even the AST output doesn't give you the variables like `libfwupd_src`. The important bit is we don't want to build everything like the meson files does; only maybe 30-50% of the sources. I guess there are two things we could do here:

 * We have something like this in the meson.build:

	+++ libfwupd/meson.build
	@@ -40,12 +40,12 @@ endif
	 libfwupd_src = [
	   'fwupd-client.c',
	   'fwupd-client-sync.c',
	-  'fwupd-common.c',
	+  'fwupd-common.c',         # used for fuzzers
	   'fwupd-device.c',
	-  'fwupd-enums.c',
	-  'fwupd-error.c',
	+  'fwupd-enums.c',          # used for fuzzers
	+  'fwupd-error.c',          # used for fuzzers
	   'fwupd-security-attr.c',
	-  'fwupd-release.c',
	+  'fwupd-release.c',        # used for fuzzers
	   'fwupd-plugin.c',
	   'fwupd-remote.c',
	 ]

..and then just read the meson.build file and create the file list based on the source lines with the magic comment suffix.

* We source a file like libfwupd/fuzzing.py and libfwupdplugin/fuzzing.py to get the filelists -- although this isn't awesome as you still need to remember to edit the file list in two places.

hughsie: > with the magic comment

That actually worked quite well. I've pushed a new version that does that.

PR #2853, opened 2021-02-08
Integrate oss-fuzz with the CI

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,3 +37,25 @@ jobs:
         run: |
           echo $GITHUB_WORKSPACE
           docker run --privileged -e CI=true -t -v $GITHUB_WORKSPACE:/github/workspace docker.pkg.github.com/fwupd/fwupd/fwupd-${{matrix.os}}:latest
+
+  fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'fwupd'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'fwupd'
+        fuzz-seconds: 300
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
--- a/libfwupdplugin/fu-firmware-common.c
+++ b/libfwupdplugin/fu-firmware-common.c
@@ -10,6 +10,7 @@
 
 #include <string.h>
 
+#include "fu-common.h"
 #include "fu-firmware-common.h"
 
 /**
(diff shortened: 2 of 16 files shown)


hughsie:
I did not expect oss-fuzz to find that first bug _quite_ that fast...

.github/workflows/main.yml
```diff
@@ -37,3 +37,25 @@ jobs:
...
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'fwupd'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'fwupd'
+        fuzz-seconds: 300
```
superm1, line 54: so it basically has a 300 second time for every job?

.github/workflows/main.yml
```diff
@@ -37,3 +37,25 @@ jobs:
         run: |
           echo $GITHUB_WORKSPACE
           docker run --privileged -e CI=true -t -v $GITHUB_WORKSPACE:/github/workspace docker.pkg.github.com/fwupd/fwupd/fwupd-${{matrix.os}}:latest
+
+  fuzzing:
+    runs-on: ubuntu-latest
```
superm1, line 42: Do you want this only running on master perhaps?

.github/workflows/main.yml
hughsie: Yup, the default was 10 minutes but I thought that was a bit too long to wait for each PR. Less is fine too, even 60 seconds would find any glaring issues.

.github/workflows/main.yml
hughsie: I wanted it on each branch so that the plugin submitter would have to fix fuzzing issues *before* we merged to master and it became *our* problem.

superm1: So the expectation is that plugin submitter integrates the plugin into oss-fuzz as _part_ of the submission?

.github/workflows/main.yml
superm1: Hear out my idea.  How about if we have 10 minutes on master nightly but then something short only on the code that _changed_?  Is that crazy?

libfwupdplugin/fu-firmware-common.c
```diff
@@ -121,3 +122,158 @@ fu_firmware_strparse_uint32 (const gchar *data)
...
+
+/**
+ * fu_firmware_strparse_uint24_safe:
+ * @data: destination buffer
+ * @datasz: size of @data, typcally the same as `strlen(data)`
+ * @offset: offset in chars into @data to read
+ * @value: (out) (nullable): parsed value
+ * @error: A #GError or %NULL
+ *
+ * Parses a base 16 number from a string.
```
superm1, line 227: this description should be changed for each function.  It's same right now.

superm1: IE describe the need for uint24 vs uint16 vs uint8

libfwupdplugin/fu-firmware-common.h
```diff
@@ -8,8 +8,39 @@
...
 
+G_DEPRECATED_FOR(fu_firmware_strparse_uint4_safe)
 guint8		 fu_firmware_strparse_uint4		(const gchar	*data);
+G_DEPRECATED_FOR(fu_firmware_strparse_uint8_safe)
 guint8		 fu_firmware_strparse_uint8		(const gchar	*data);
+G_DEPRECATED_FOR(fu_firmware_strparse_uint16_safe)
 guint16		 fu_firmware_strparse_uint16		(const gchar	*data);
+G_DEPRECATED_FOR(fu_firmware_strparse_uint24_safe)
 guint32		 fu_firmware_strparse_uint24		(const gchar	*data);
+G_DEPRECATED_FOR(fu_firmware_strparse_uint32_safe)
```
superm1, line 19: add this to the list of stuff to drop with ABI bump

.github/workflows/main.yml
hughsie: >  How about if we have 10 minutes on master nightly

We get several hours on master twice a day; the oss-fuzz bot has been opening AddressSanitizer bugs against fwupd all morning :) The CI version of oss-fuzz is meant as a quick check.

.github/workflows/main.yml
hughsie: I think for anything that's doing more than _safe() stuff in the FuXxxxxFirmware we should expect the author to implement ->build() and ->write() so that we (they?) can construct minimal fuzzing targets.

.github/workflows/main.yml
hughsie: I meant to add; if you want cc on those bugs let me know and I'll submit a pr to oss-fuzz so that you're on the list too. Slight warning, the only way to opt out is another PR...

libfwupdplugin/fu-firmware-common.h
hughsie: Fixed, thanks.

.github/workflows/main.yml
superm1: >We get several hours on master twice a day; the oss-fuzz bot has been opening AddressSanitizer bugs against fwupd all morning :) The CI version of oss-fuzz is meant as a quick check.

Oh if it's already running like that, yeah I think what you're doing here makes a lot more sense.

>if you want cc on those bugs

You can leave me off for now.

.github/workflows/main.yml
superm1: I guess that should be documented in https://fwupd.org/lvfs/docs/consulting/info as well?

libfwupdplugin/fu-firmware-common.c
hughsie: Something like:

     * Parses a base 16 number from a string of 6 characters in length.
     * The returned @value will range from from 0 to 0xffffff.

superm1: Yes that is exactly what I think we need in these functions

.github/workflows/main.yml
hughsie: I think adding it to https://github.com/fwupd/fwupd/wiki/New-plugin-checklist is probably better -- maybe something like:

  [X] Implement ->write() and include a fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass

superm1: Oh yeah.  Sounds good

libfwupdplugin/fu-firmware-common.c
hughsie: Okay, fixed, thanks.

.github/workflows/main.yml
hughsie: Done.

superm1 (approved):

PR #2855, opened 2021-02-09
Wip/hughsie/fu provider slowly

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/README.md
+++ b/libfwupdplugin/README.md
@@ -3,3 +3,4 @@ Planned API/ABI changes for next release
 
  * Remove fu_common_is_cpu_intel()
  * Remove fu_firmware_strparse_uintXX()
+ * Remove fu_plugin_get_usb_context() and fu_plugin_set_usb_context()
--- a/libfwupdplugin/README.md
+++ b/libfwupdplugin/README.md
@@ -4,3 +4,5 @@ Planned API/ABI changes for next release
  * Remove fu_common_is_cpu_intel()
  * Remove fu_firmware_strparse_uintXX()
  * Remove fu_plugin_get_usb_context() and fu_plugin_set_usb_context()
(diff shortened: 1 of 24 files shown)


superm1 (commented):
Are you hard pressed on `provider`?  How about `interface_provider` or `hardware_interface`?

hughsie:
> Are you hard pressed on provider? 

Nah, but I did spend 10 minutes with a thesaurus this morning.

> How about interface_provider or hardware_interface?

Both too wordy. I pondered FuSubsys but it's weird passing in `subsystems` _into_ a `FuUdevSubsys`. I guess the other logical one is `FuUsbBackend` but I'm not sure that's better. What I'm trying to convey is that it's a mechanism for how we get "this device has appeared" and "this device has gone away". `FuUsbInterface` is already overloaded and `FuUdevFacet` isn't super clear either. Maybe it's a factory. At the risk of sounding like Java, how about `FuUdevFactory` or FuUdevDeviceFactory? I'm not sure the metaphor holds 100% as a factory doesn't come and take the things it produces away again.

superm1:
How about just `interface`?  or `hwif`?  Factory seems better than provider though too.

superm1:
or `enumerator`?

hughsie:
> or `enumerator`?

I think that's the verb, not the noun. Interface is going to cause all kinds of problems with GType. 😐

hughsie:
FuUsbMethod?

superm1:
I think backend and method are my top two for now then.

hughsie:
I went with `backend` as it's better than provider.

superm1:
OK LGTM

hughsie:
I'll do some testing with the device-tests tomorrow and make sure we've not broken anything.

PR #2856, opened 2021-02-09
trivial: Use json-glib from master to get a fuzzing crash fix

hughsie:
See https://gitlab.gnome.org/GNOME/json-glib/-/commit/7711cbc7f8d85d909bd4ccaa0c4c0f8d306f872f

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/oss-fuzz.py
+++ b/contrib/ci/oss-fuzz.py
@@ -217,7 +217,6 @@ def _build(bld: Builder) -> None:
     src = bld.checkout_source(
         "json-glib",
         url="https://gitlab.gnome.org/GNOME/json-glib.git",
-        commit="c30c988ac3b0d9e7c332232e3c3969818749b239",
     )
     bld.build_meson_project(
         src,


superm1 (commented):
I guess this is mostly a philosophical question if you want a deterministic fuzz run or fuzz essentially ToT for all the projects

hughsie:
That's a good question. I'm erring on master for JSON-glib, glib and libxmlb as we know all upstream teams and I think it would be useful to know that something we depend on has regressed. On the logic, if fwupd is crashing in the field it's us who look like Muppets, even if it's some dep that's done something dumb.

superm1:
> That's a good question. I'm erring on master for JSON-glib, glib and libxmlb as we know all upstream teams and I think it would be useful to know that something we depend on has regressed. On the logic, if fwupd is crashing in the field it's us who look like Muppets, even if it's some dep that's done something dumb.

I tend to agree - but the problem you run into though is what if a distro has an old snapshot with the crasher?  Like you still need a feedback loop to document somewhere that there was a vulnerability and people need to pick it up right?

hughsie:
> what if a distro has an old snapshot

I don't think it's our job to ensure the distro(s) included  the fix. I guess a compromise would be to target the *branch* being shipped by something fairly aggressive, e.g. F33.

hughsie:
I'll merge this; on the logic that there is no stable branch for json-glib-1-6 -- I'm happy to continue discussing the stable branch for glib tho, as I think that's the one that might bite us the hardest.

PR #2866, opened 2021-02-10
1_4_X: uefi-capsule: Add SBAT metadata to the fwupd EFI binary

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


superm1:
Backported from 378b00018b42bdaefe50e70fa546065a41ee0e42 with the following modifications:
     - No support for supported build
     - Plugin did not rename yet
     - Update debian/rules for newer CONFARGS building from master

superm1:
@hughsie I think you should tag 1_4_X and 1_3_X releases at the same time as 1.5.x so all the distros on older stuff can grab the new tarball to add sbat to their binaries too.

hughsie:
> I think you should tag 1_4_X and 1_3_X releases

Agree; lets get 1.5.6 out next week and see if the SBAT stuff actually works in practice before I tag the others.

superm1:
@hughsie hopefully I got the backports right for 1_4_X, 1_3_X and 1_2_X done correctly.  let me know if you want them squashed or if it's better to leave as individual commits so it's easier to match what changed as part of the backport with a diffing tool.

superm1:
@hughsie can you look at these again?  I think we need to get point releases out relatively soon.

hughsie (approved):

PR #2869, opened 2021-02-10
pixart: Add support for more devices

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@jyu99 I've rebased the changes and fixed up a lot of issues. Lets carry on discussion on this PR.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -599,10 +615,25 @@ static gboolean
...
+	if (!fu_pxi_device_get_raw_info (self, &hid_raw_info ,error))
+		return FALSE;
 	if (!fu_pxi_device_fw_ota_init (self, error))
 		return FALSE;
 	if (!fu_pxi_device_fw_get_info (self, error))
 		return FALSE;
+
+	/* extra GUID */
+	g_string_ascii_up (ble_name);
+	fu_common_string_replace (ble_name, " ", "_");
```
superm1, line 631: is there any other filtering that might be needed for this string?

superm1 (approved):
Just one nit from me, else looks fine.

plugins/pixart-rf/fu-pxi-device.c
hughsie: Do we have any ideas for this? Quite a few plugins do this now. I think the baseline ought to be ASCII, no spaces, uppercase.

superm1: yeah just with the all the fuzzing stuff you've done lately in mind, it's definitely "untrusted" data.  I think that's spot on, ASCII only, no spaces.  uppercase we need to double check all the other plugins that they're not using lowercase.

hughsie: > we need to double check all the other plugins

At least one is. We can't change the instance IDs now, the GUIDs are essentially _API_.

superm1: then just got with uppercase for everything new I guess

jyu99:
If we have more than one device to update at same time, the fwupd is execute update task concurrent by multi-thread, or just execute the task sequentially ?

hughsie:
> fwupd execute updating task concurrent

No, all of fwupd is (currently) single threaded. All updates are always done sequentially.

jyu99:
Hi  I have a problem about GUID

superm1:
>Hi I have a problem about GUID

What problem?

PR #2871, opened 2021-02-11
Wip/hughsie/oss fuzz crc

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-firmware.h
```diff
@@ -48,12 +48,16 @@ struct _FuFirmwareClass
  * @FU_FIRMWARE_FLAG_NONE:			No flags set
  * @FU_FIRMWARE_FLAG_DEDUPE_ID:			Dedupe imges by ID
  * @FU_FIRMWARE_FLAG_DEDUPE_IDX:		Dedupe imges by IDX
+ * @FU_FIRMWARE_FLAG_HAS_CHECKSUM:		Has a CRC or checksum to test internal consistency
+ * @FU_FIRMWARE_FLAG_HAS_VID_PID:		Has a vendor or product check in the firmware
```
superm1, line 52: I don't think you can prove it's a product / vendor **check**.  You can only prove that the firmware image contains an identifier of some sort.  Whether or not it's validated, that's a different story.

hughsie: Well, I figured `HAS` kinda implies that the information is there, rather than we verify it.

superm1: But the comment says `check`

hughsie: Ha, sorry! What about `Has a vendor or product ID in the firmware`

superm1: +1

hughsie: Fixed, thanks.

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -21,153 +21,34 @@ typedef struct {
...
-	gboolean fmap_found = FALSE;
-
-	if (offset == NULL) {
-		g_set_error_literal (error,
-				     G_IO_ERROR,
-				     G_IO_ERROR_INVALID_DATA,
-				     "offset return not valid");
-		return FALSE;
-	}
+#ifdef HAVE_MEMMEM
```
superm1, line 28: probably need @bleungatchromium to comment if this is OK in chromeos to depend on that.

superm1 (approved):

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Alright. Let me double check both architectures we build to make sure we have this.

hughsie: It's really just depending on _GNU_SOURCE and glibc. Thanks!

bleungatchromium:
Hmm. Something went wrong with this change. I have memmem, but it fails to find the FMAP in my firmware file.

`attempted to read 0x38 bytes at offset 0xfffcfa80 from buffer of 0x80000`

hughsie:
> but it fails to find the FMAP in my firmware file

Doh. We need a self test for that, I can add one. @bleungatchromium I've fixed the issue I think, can you retry with ^^ please.

bleungatchromium:
> > but it fails to find the FMAP in my firmware file
> 
> Doh. We need a self test for that, I can add one. @bleungatchromium I've fixed the issue I think, can you retry with ^^ please.

Works now. Thanks!

hughsie:
> Works now

I'll add the test then commit. Thanks!

bleungatchromium (approved):

superm1:
@bleungatchromium BTW the fact that oss-fuzz caught the crasher in the fmap code leading to this simplification probably means y'all should probably drop the use of library this was previously based upon previously and go all in on fwupd! 😝

hughsie:
@bleungatchromium @superm1 can you just check that last commit again pls; I've added some tests and fixed some bugs in the process :o

bleungatchromium:
> @bleungatchromium @superm1 can you just check that last commit again pls; I've added some tests and fixed some bugs in the process :o

Something else went wrong.

`file size incorrect, expected 0x80000 got 0x4fa80`

libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -205,26 +90,25 @@ fu_fmap_firmware_parse (FuFirmware *firmware,
...
+	offset = priv->offset;
 	if (!fu_memcpy_safe ((guint8 *) &fmap, sizeof(fmap), 0x0,	/* dst */
-			     buf, bufsz, offset,			/* src */
+			     buf, bufsz, priv->offset,			/* src */
 			     sizeof(fmap), error))
 		return FALSE;
 	priv->base = GUINT64_FROM_LE (fmap.base);
 
-	if (GUINT32_FROM_LE (fmap.size) != bufsz) {
+	if (GUINT32_FROM_LE (fmap.size) != bufsz - priv->offset) {
```
bleungatchromium, line 105: Not sure this is correct...

This error pops on this version of the change:
`file size incorrect, expected 0x80000 got 0x4fa80`

bleungatchromium (changes requested):

libfwupdplugin/fu-fmap-firmware.c
hughsie: Can you send me a link of an image that's valid and that FuFmapFirmware fails to parse please? I've added one test, and we probably need a few more.

hughsie: Also; you sure you tested the latest version? I got that too on the interim version of the commit that I repushed.

bleungatchromium: I'm at commit 40a3bb786916b93f79339de5c0db5782ac3cdd21.

This binary has the issue too. 

https://fwupd.org/downloads/b67bda9d75094f7d534e706557ce0a5ff973e9a3543d39eb1ba78b833ec410a6-servo_micro_v2.4.17.cab

bleungatchromium: I can't share with you the .bin I tested before, but the servo_micro_v2.4.17 has a similar issue. Here's the exact error from the binary I just linked to:

`file size incorrect, expected 0x20000 got 0x165c0`

hughsie: Hmm, so there's two bugs. Is the FuFmapArea.offset supposed to be relative to the start of the .bin or relative to the offset that __FMAP__ is specified at? The .bin would suggest the former. The FuFmap.size also seems to specify the file size, not the size of the __FMAT__ section...

bleungatchromium: https://www.chromium.org/chromium-os/firmware-porting-guide/fmap

yeah, fmap_size is the full size of the file.

plugins/cros-ec/fu-cros-ec-firmware.h
```diff
@@ -7,7 +7,7 @@
 #pragma once
 
 #include "fu-firmware.h"
-#include "fu-fmap-firmware.h"
+#include "fu-fmap-firmware-private.h"
```
superm1, line 10: mm this doesn't feel right that a plugin should rely upon a private header.  Remember plugin ABI is supposed to be public..

libfwupdplugin/fu-fmap-firmware.c
bleungatchromium: Area.offsets are from the start of the .bin.

plugins/cros-ec/fu-cros-ec-firmware.h
hughsie: I wasn't sure what to do here; exporting FuFmap and FuFmapArea C structs seemed like the wrong thing to do. Maybe what we had before was the least bad thing afterall?

libfwupdplugin/fu-fmap-firmware.c
hughsie: Okay, the issue was in the way we wrote the FMAP file; I've changed the reading back to what we had before and than can now pass it's new self tests and also parse the file we know should work. I'll re-push when we've worked out the -private thing.

plugins/cros-ec/fu-cros-ec-firmware.h
hughsie: I've undone the -private change. If we do anything it shouldn't be part of this crash-fix bug.

libfwupdplugin/fu-fmap-firmware.c
hughsie: @bleungatchromium looks good to you now?

bleungatchromium (approved):
Looks good. Passes on my side too.

PR #2874, opened 2021-02-11
fwupd port for BSD distros

Asiderr:
Draft PR of fwupd port for BSD distros.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -6232,7 +6232,7 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 	guint backend_cnt = 0;
 	g_autoptr(GPtrArray) checksums_approved = NULL;
 	g_autoptr(GPtrArray) checksums_blocked = NULL;
-#ifdef __linux__
+#ifndef _WIN32
 	g_autoptr(GError) error_local = NULL;
 #endif
 


contrib/generate-version-script.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
superm1, line 1: I seem to recall we had a good reason for doing it the way we do, @hughsie do you remember?

libfwupdplugin/fu-common.c
```diff
@@ -2256,7 +2256,8 @@ fu_byte_array_set_size (GByteArray *array, guint length)
 gboolean
 fu_common_kernel_locked_down (void)
 {
-#ifndef _WIN32
+#if !defined(_WIN32) && !defined(__FreeBSD__) && !defined(__DragonflyBSD__) \
```
superm1, line 2259: If these will all be the same for WIN32 and BSD I wonder if it's worth making a meson.build level define for this kind of thing?

meson_options.txt
```diff
@@ -1,5 +1,6 @@
 option('build', type : 'combo', choices : ['all', 'standalone', 'library'], value : 'all', description : 'build type')
 option('agent', type : 'boolean', value : true, description : 'enable the fwupd agent')
+option('bsd', type : 'boolean', value : false, description : 'disable linux specific options')
```
superm1, line 3: I don't think we want an option for this - it should all be detectable at build time.

plugins/meson.build
```diff
@@ -13,9 +13,12 @@ subdir('nitrokey')
 subdir('pci-bcr')
 subdir('pci-mei')
 subdir('test')
-subdir('upower')
 subdir('pixart-rf')
 
+if not get_option('bsd')
```
superm1, line 18: assuming you successfully detect at build time, it should be guarded behind that flag rather than an option

contrib/generate-version-script.py
superm1: Ah I remember now!
https://blogs.gnome.org/mcatanzaro/2018/02/16/on-python-shebangs/

libfwupdplugin/fu-common.c
Asiderr: There are some WIN32 specific lines, that are not related to BSD distros, but most of the WIN32 ifdefs apply to BSD as well.

hughsie: Can we just do `ifdef __linux__` or something? That's basically what we're asking...

meson_options.txt
hughsie: Agree.

libfwupdplugin/fu-common.c
Asiderr: Sure, it will be more readable.

contrib/generate-version-script.py
superm1: A better solution is to instead make the call to `generate-version-script.py` use your intended python3 like we do here: https://github.com/fwupd/fwupd/blob/master/meson.build#L371

src/fu-engine.c
```diff
@@ -6287,7 +6287,7 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 	FuQuirksLoadFlags quirks_flags = FU_QUIRKS_LOAD_FLAG_NONE;
 	g_autoptr(GPtrArray) checksums_approved = NULL;
 	g_autoptr(GPtrArray) checksums_blocked = NULL;
-#ifndef _WIN32
+#ifndef __linux__
```
superm1, line 6290: shouldn't this be `#ifdef`  not `#ifndef`

src/fu-engine.c
```diff
@@ -6299,7 +6299,7 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 		return TRUE;
 
 /* TODO: Read registry key [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography] "MachineGuid" */
-#ifndef _WIN32
+#ifndef __linux__
```
superm1, line 6302: shouldn't this be `#ifdef` not `#ifndef`

src/fu-self-test.c
```diff
@@ -583,7 +583,7 @@ fu_engine_requirements_device_func (gconstpointer user_data)
 		"    <firmware compare=\"ge\" version=\"1.2.3\"/>"
 		"    <firmware compare=\"eq\" version=\"4.5.6\">bootloader</firmware>"
 		"    <firmware compare=\"regex\" version=\"USB:0xFFFF|DMI:Lenovo\">vendor-id</firmware>"
-#ifndef _WIN32
+#ifndef __linux__
```
superm1, line 586: shouldn't this be `#ifdef` not `#ifndef`

plugins/meson.build
superm1: And if you can't detect from meson that it's a linux build, we should instead offer a meson option for this plugin rather than one for "bsd"

superm1: According to docs https://mesonbuild.com/Cross-compilation.html you can use `host_machine.system()`

superm1: I've pulled out some of the stuff related to this to https://github.com/fwupd/fwupd/pull/2896 while I was looking at it.

Asiderr: @superm1 I'll change this in the close future. I was trying to build fwupd on DragonflyBSD today and I've pushed WIP commit to the upstream PR by mistake. I will fix following problems in the next push force.

superm1: Something like what I did landed on master.  So please rebase all your work on master, and you shouldn't need to have a `get_option('bsd')` anymore.

hughsie:
@Asiderr I think we're nearly there! Rebase against master, and then a few more things to fix up.

libfwupd/fwupd-common.c
```diff
@@ -169,6 +169,22 @@ fwupd_get_os_release (GError **error)
...
+			     g_strdup("OS"),
+			     g_strdup("NetBSD"));
+	return hash;
+#endif
+
+#ifdef __OpenBSD__
+	hash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+	g_hash_table_insert (hash,
+			     g_strdup("OS"),
+			     g_strdup("OpenBSD"));
```
hughsie, line 184: is there no data file we can read? :(

src/fu-engine.c
Asiderr: Fixed.

src/fu-engine.c
```diff
@@ -6354,7 +6354,7 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 		return TRUE;
 
 /* TODO: Read registry key [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography] "MachineGuid" */
-#ifndef _WIN32
+#ifdef __linux__
```
superm1, line 6357: is there anything else similar that can be obtained from `*BSD`?

libfwupd/fwupd-common.c
hughsie: also, space before the brackets please

src/fu-engine.c
Asiderr: @superm1 I found out how we can recognize if dbus is enabled. The `machine-id` exists only if dbus is running, so we can check if the file is present during meson configuration. In addition, I will add the one more `fns` path (`/var/db/dbus/machine-id`) to `fwupd_build_machine_id`.

superm1: Sounds good.

libfwupd/fwupd-common.c
```diff
@@ -358,6 +374,7 @@ fwupd_build_machine_id (const gchar *salt, GError **error)
 	fns[1] = g_build_filename (FWUPD_LOCALSTATEDIR, "lib", "dbus", "machine-id", NULL);
 	fns[2] = g_strdup ("/etc/machine-id");
 	fns[3] = g_strdup ("/var/lib/dbus/machine-id");
+	fns[4] = g_strdup ("/var/db/dbus/machine-id");
```
hughsie, line 377: you need to bump `g_auto(GStrv) fns = g_new0 (gchar *, 5);` if you do this.

meson.build
```diff
@@ -242,10 +242,29 @@ if build_daemon
...
+  ]
+  foreach arg: machine_id_paths
+    if fs.is_file(arg)
+      dbus_enabled = true
+      break
+    endif
+  endforeach
+  if not dbus_enabled and not host_machine.system() == 'windows'
+    error('D-Bus is required for the daemon to function properly')
+  endif
```
hughsie, line 267: This isn't really checking if dbus is enabled, this is just checking to location of the machine-id file. I wonder if we're going to all this trouble in meson.build we should just set something in config.h and then use that rather than searching in fwupd again.

Asiderr: @hughsie I've push that idea as separate commit.

Asiderr: @hughsie I implemented this idea in the last commit.

libfwupd/fwupd-common.c
```diff
@@ -330,7 +346,7 @@ fwupd_build_user_agent (const gchar *package_name, const gchar *package_version)
 
 /**
  * fwupd_build_machine_id:
- * @salt: The salt, or %NULL for none
+ * @fn: The path of machine-id
```
superm1, line 349: this is a _weak_ ABI change.  The function is exported to all clients to be able to potentially use.  Even though the data type didn't change, the behavior of it changed.  The filename is now used for the salt, and it actually is required.

I think we actually need a new symbol for this new function behavior and to deprecate the old one with the next ABI bump.

meson.build
```diff
@@ -242,10 +242,28 @@ if build_daemon
...
-  if udevdir == ''
+  if udevdir == '' and host_machine.system() == 'linux'
     udev = dependency('udev')
     udevdir = udev.get_pkgconfig_variable('udevdir')
   endif
+  # Check if machine-id exists
+  fs = import('fs')
+  machine_id_paths = [
+    '/var/local/lib/dbus/machine-id',
+    '/var/local/etc/machine-id"',
```
superm1, line 253: these paths should not be hardcoded, but should instead reference the arguments like previously.  build a path using them.

libfwupd/fwupd-common.c
```diff
@@ -538,7 +534,7 @@ fwupd_build_history_report_json (GPtrArray *devices, GError **error)
 	g_return_val_if_fail (error == NULL || *error == NULL, NULL);
 
 	/* get a hash that represents the machine */
-	machine_id = fwupd_build_machine_id ("fwupd", error);
```
superm1, line 541: this change broke the fuzzing target . So fuzzing target will need to set this to something too.

hughsie:
I merged a good chunk of this PR in https://github.com/fwupd/fwupd/pull/3046 -- I think we can drop the machine_id changes on this PR, I think I've changed my mind and we really do want to do this just at runtime.

Asiderr:
@hughsie I've added some minor fixes that enable building fwupd in the *BSDs.

superm1 (dismissed):

plugins/dfu/meson.build
```diff
@@ -1,4 +1,4 @@
-if get_option('gusb')
+if get_option('gusb') and host_machine.system() == 'linux'
```
superm1, line 1: Wait, why doesn't this work on win32?

Asiderr: @superm1 It doesn't work on *BSDs. [G_USB_DEVICE_LANGID_ENGLISH_UNITED_STATES](https://github.com/fwupd/fwupd/blob/f3c64adb6c6d4ad6b2045682c48cdd3bbf3a85a7/plugins/dfu/fu-dfu-device.c#L1232) is undeclared. I will add win32 to this condition.

superm1: Hmm, are you perhaps using too old of a gusb version?  That code seems pretty straightforward.
https://github.com/hughsie/libgusb/blob/master/gusb/gusb-device.h#L129

superm1: Looks like 0.3.6 or newer is needed. https://github.com/hughsie/libgusb/commit/38d1f8b3af5e76b548b5701f39ac9dd8a74bde9e

Asiderr: @superm1 My bad. I had old gusb and that was the problem.

hughsie (approved):

Issue #2873, opened 2021-02-11
plugins/uefi-capsule/efi/meson.build:159:4: ERROR: Problem encountered: -Defi_os_dir is unset


joakim-tjernlund:
**Describe the bug**
Just tried to biuld master (been a few weeks since I did that) and got:
```
Configuring vendor.conf using configuration
Configuring vendor-directory.conf using configuration
Configuring fwupdtool using configuration
Configuring fwupdmgr using configuration
Configuring 85-fwupd using configuration
Configuring org.freedesktop.fwupd.service using configuration
Program /usr/bin/vapigen-0.48 found: YES
Program help2man found: YES
Compiler for C supports arguments -fcf-protection: YES 
Program help2man found: YES
Has header "/usr/include/efi/x86_64/efibind.h" : YES 
Message: efi-libdir: "/usr/lib64"
Message: efi-ldsdir: "/usr/lib64"
Message: efi-includedir: "/usr/include/efi"

plugins/uefi-capsule/efi/meson.build:159:4: ERROR: Problem encountered: -Defi_os_dir is unset, see plugins/uefi-capsule/README.md
```

I figure fwupd should not require this by default?

joakim-tjernlund:
if I unset supported_build I get this annoying msg:
```
WARNING: This package has not been validated, it may not work properly.
```

How can I keep supported_build? What should I set efi_os_dir to(without breaking anything)?

hughsie:
> How can I keep supported_build

You either need `supported_build=false` or you need to define `efi_os_dir` -- it's like that by design for exactly this reason.

joakim-tjernlund:
> > How can I keep supported_build
> 
> You either need `supported_build=false` or you need to define `efi_os_dir` -- it's like that by design for exactly this reason.

OK, but I don't know this stuff. What should I set efi_os_dir to without breaking anything ?

superm1:
what do you use on the esp?  `/boot/efi/$FOO/`.  It should be whatever `$FOO` evaluates to.

joakim-tjernlund:
OK, I guess it should be BOOT then:
```
cd /boot/EFI/BOOT/
# >  ls
./  ../  BOOTX64.EFI*

joakim-tjernlund:
Somewhat odd I have to specified that now.
Maybe you should consider dropping this warning instead:
    WARNING: This package has not been validated, it may not work properly.

it is the only reason for me to have supported_build=true

superm1:
[...]

PR #2876, opened 2021-02-12
trivial: Untie efi_os_dir from the SBAT distro ID

hughsie:
We happily fallback with an unset -Defi_os_dir and it gets confusing explaining
why the EFI dir needs to be set on a non-supported system.

In practice they'll probably have the same eventual value on most supported
distributions.

Fixes https://github.com/fwupd/fwupd/issues/2873

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


meson_options.txt
```diff
@@ -44,6 +44,7 @@ option('efi-ldsdir', type : 'string', description : 'path to the EFI lds directo
 option('efi-includedir', type : 'string', value : '/usr/include/efi', description : 'path to the EFI header directory')
 option('efi_os_dir', type: 'string', description : 'the name of OS directory in ESP, e.g. fedora')
 option('efi_sbat_fwupd_generation', type : 'integer', value : 0, description : 'SBAT fwupd generation')
+option('efi_sbat_distro_id', type : 'string', value : '', description : 'SBAT distribution ID, e.g. fedora')
```
superm1, line 47: I think both `*os_dir`  values need the word `hardcoded` adjective added to them now to make it clearer why you would use them

hughsie: I'm not super keen changing the meson define name as that's going to break (silently?) distros using the old name to override the default -- or do you mean the description?

superm1: Just the description field I mean.

hughsie: Ohh, something like `the hardcoded name of OS directory in ESP, e.g. fedora`?

superm1: Exactly

hughsie: Fixed.

superm1 (approved):

PR #2883, opened 2021-02-15
Progressbar display fix

bubbleguuum:


hughsie:
Can you squash it down to one commit please. Then lgtm.

bubbleguuum:
Done. Had to ninja learn what squash is.

PR #2884, opened 2021-02-15
fu-udev-device.c: Fix ioctl error checking

hardboprobot:
fu_udev_device_ioctl() interprets the return value of ioctl() as the
error cause and generates a misleading error message (permission
denied).
Fix it to use errno for error checking and reporting.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Can you drop the filename prefix to the commit message please, then LGTM (which means you can rebase and merge your own patch :)

hughsie (approved):

PR #2888, opened 2021-02-16
Allow building without SBAT metadata

hughsie:
Two reasons:

 * It seems a bit antisocial to hard-require all this data without fair warning
 * The aarch64 pesign crashes when trying to sign the binary with SBAT metadata


martinezjavier (approved):
LGTM.

PR #2889, opened 2021-02-16
Wip/bluez backend

hardboprobot:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
I'll commit a few fixes to the branch and then we can re-review. I hope that's okay.

libfwupdplugin/fu-ble-device.c
```diff
@@ -0,0 +1,120 @@
...
+	FuBleDevicePrivate *priv = GET_PRIVATE (self);
+
+	return priv->name;
+}
+
+/**
+ * fu_ble_device_get_address:
+ * @self: A #FuBleDevice
+ *
+ * Gets the address of the device.
```
superm1, line 57: I think an example expected syntax would be good to leave here in documentation

libfwupdplugin/fu-ble-device.c
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * An object that represents a Bluetooth LE device.
+ *
+ * See also: #FuDevice
+ */
+
+typedef struct {
+	gchar		*name;
+	gchar		*address;
+	gchar		*adapter;
```
superm1, line 27: I don't actually see adapter used.  Is that an oversight/mistake or not yet implemented?

libfwupdplugin/fu-ble-device.c
```diff
@@ -0,0 +1,120 @@
...
+#define GET_PRIVATE(o) (fu_ble_device_get_instance_private (o))
+
+/**
+ * fu_ble_device_get_name:
+ * @self: A #FuBleDevice
+ *
+ * Gets the name of the device.
+ *
+ * Returns: The name of the device.
+ *
```
superm1, line 42: Can you contrast how this is different than the name that the parent `FuDevice` has?

src/fu-bluez-backend.c
```diff
@@ -0,0 +1,345 @@
...
+
+struct _FuBluezBackend {
+	FuBackend		 parent_instance;
+	GDBusConnection		*connection;
+	GHashTable		*devices;	/* platform_id : * FuDevice */
+};
+
+G_DEFINE_TYPE (FuBluezBackend, fu_bluez_backend, FU_TYPE_BACKEND)
+
+
```
superm1, line 24: should just be one line of whitespace

src/fu-bluez-backend.c
```diff
@@ -0,0 +1,345 @@
...
+			continue;
+		}
+		if (g_strcmp0 (prop_name, "Name") == 0) {
+			name = g_strdup (g_variant_get_string(prop_val, NULL));
+			continue;
+		}
+		if (g_strcmp0 (prop_name, "Connected") == 0) {
+			connected = g_variant_get_boolean(prop_val);
+			continue;
+		}
```
superm1, line 57: did I miss it - how does this get updated?

src/fu-engine.c
```diff
@@ -52,13 +52,15 @@
...
 #include "fu-udev-device-private.h"
+#include "fu-ble-device.h"
 
 #ifdef HAVE_GUDEV
 #include "fu-udev-backend.h"
 #endif
 #ifdef HAVE_GUSB
 #include "fu-usb-backend.h"
 #endif
+#include "fu-bluez-backend.h"
```
superm1, line 63: I think we actually want to have the bluez backend as optional.  We won't have bluez in Windows for example.

hughsie: I'm splitting this up right now :)

libfwupdplugin/fu-ble-device.c
hughsie: Done.

libfwupdplugin/fu-ble-device.c
hughsie: yet implemented, and I added it

libfwupdplugin/fu-ble-device.c
hughsie: It's not! I've switched it to just use the parent name

src/fu-bluez-backend.c
hughsie: Fixed.

src/fu-bluez-backend.c
hughsie: There's a big fat:

	/*
	 * TODO: Subscribe DBus signals.
	 */

:)

src/fu-engine.c
hughsie: So, we don't actually *need* to make it optional; everything is done over DBus. For win32 we'll just fail to coldplug, which might be fine. Maybe we don't want to compile it at all, opinions welcome.

libfwupdplugin/fu-ble-device.c
```diff
@@ -22,50 +22,186 @@
...
 	FuBleDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (FU_IS_BLE_DEVICE (self), NULL);
+	return priv->adapter;
+}
 
-	return priv->address;
+/**
+ * fu_ble_device_set_adapter:
+ * @self: A #FuBleDevice
+ * @adapter: (nullable): The adapter, e.g. `FIXME:better-example-please`
```
superm1, line 96: (just leaving this as todo)

hughsie: OOps, fixed.

libfwupdplugin/fu-ble-device.c
```diff
@@ -22,50 +22,186 @@
...
+				     "not supported");
+		return NULL;
+	}
+	return klass->read (self, uuid, error);
+}
+
+/**
+ * fu_ble_device_read_string:
+ * @self: A #FuBleDevice
+ * @uuid: The UUID, e.g. `FIXME:better-example-please`
```
superm1, line 145: (leave as an open comment with todo)

libfwupdplugin/fu-ble-device.c
```diff
@@ -22,50 +22,186 @@
  */
 
 typedef struct {
-	gchar		*name;
-	gchar		*address;
-	gchar		*adapter;
-	GHashTable	*characteristics;	/* utf8 : utf8 */
+	gchar			*name;
```
superm1, line 25: I still don't see a need to keep a name attribute, get it from the parent `FuDevice`

libfwupdplugin/fu-ble-device.h
```diff
@@ -15,21 +16,30 @@ G_DECLARE_DERIVABLE_TYPE (FuBleDevice, fu_ble_device, FU, BLE_DEVICE, FuDevice)
...
-
-FuBleDevice		*fu_ble_device_new		(const gchar *name,
-							 const gchar *addr,
-							 GHashTable *characteristics);
-const gchar		*fu_ble_device_get_name 	(FuBleDevice *self);
-const gchar		*fu_ble_device_get_address 	(FuBleDevice *self);
+FuBleDevice		*fu_ble_device_new		(void);
+const gchar		*fu_ble_device_get_address 	(FuBleDevice	*self);
+void			 fu_ble_device_set_address	(FuBleDevice	*self,
+							 const gchar	*name);
```
superm1, line 32: this should be `addr` no?

libfwupdplugin/fu-ble-device.h
```diff
@@ -15,21 +16,30 @@ G_DECLARE_DERIVABLE_TYPE (FuBleDevice, fu_ble_device, FU, BLE_DEVICE, FuDevice)
...
+const gchar		*fu_ble_device_get_adapter 	(FuBleDevice	*self);
+void			 fu_ble_device_set_adapter	(FuBleDevice	*self,
+							 const gchar	*name);
```
superm1, line 35: should be `adapter`

hughsie: Good catch, fixed.

libfwupdplugin/fu-ble-device.c
hughsie: Fixed, thanks.

libfwupdplugin/fu-ble-device.h
hughsie: Indeed! Fixed.

src/fu-engine.c
superm1: Well aside from Win32, what about BSD, ChromeOS?  Do they use bluez?  I'm wondering if we want to make it clearer with a meson option `-Dbluetooth=true` and only compile plugins that use bluez when that's set.

src/fu-bluez-device.c
```diff
@@ -0,0 +1,304 @@
...
+
+G_DEFINE_TYPE (FuBluezDevice, fu_bluez_device, FU_TYPE_BLE_DEVICE)
+
+#define GET_PRIVATE(o) (fu_bluez_device_get_instance_private (o))
+
+/**
+ * fu_bluez_device_add_uuid_path:
+ * @self: A #FuBluezDevice
+ * @uuid: (nullable): The UUID, e.g. `FIXME:better-example-please`
+ * @uuid: (nullable): The path, e.g. `FIXME:better-example-please`
```
superm1, line 42: Just a thought to prevent us missing fixing these in CI - I wonder if we want a grep for `FIXME` in `git diff` failing a job

src/fu-bluez-device.c
```diff
@@ -0,0 +1,304 @@
...
+
+	g_return_if_fail (modalias != NULL);
+
+	/* usb:v0461p4EEFd0001 */
+	if (g_str_has_prefix (modalias, "usb:")) {
+		subsys = "USB";
+		fu_firmware_strparse_uint16_safe (modalias, modaliaslen, 5, &vid, NULL);
+		fu_firmware_strparse_uint16_safe (modalias, modaliaslen, 10, &pid, NULL);
+		fu_firmware_strparse_uint16_safe (modalias, modaliaslen, 15, &rev, NULL);
+
```
superm1, line 76: I'm confused, why does a bluez device have a usb modalias?

src/fu-bluez-backend.c
```diff
@@ -0,0 +1,196 @@
...
+		if (g_strcmp0 (prop_name, "Modalias") == 0) {
+			fu_bluez_device_set_modalias (dev, g_variant_get_string (prop_val, NULL));
+			continue;
+		}
+		if (g_strcmp0 (prop_name, "Connected") == 0) {
+			if (g_variant_get_boolean (prop_val))
+				fu_device_add_flag (FU_DEVICE (dev), FWUPD_DEVICE_FLAG_CONNECTED);
+			continue;
+		}
+	}
```
superm1, line 71: How do we avoid a situation that the device is connected but disconnects (out of range for example)?

src/fu-bluez-device.c
hughsie: ¯\_(ツ)_/¯ -- it just does!

	14:56:53:0705 FuBackend            Address: 'C4:30:18:6F:B5:49'
	14:56:53:0705 FuBackend            AddressType: 'public'
	14:56:53:0705 FuBackend            Name: 'LG Monitor(49)'
	14:56:53:0705 FuBackend            Alias: 'LG Monitor(49)'
	14:56:53:0705 FuBackend            Class: uint32 2360324
	14:56:53:0705 FuBackend            Icon: 'audio-card'
	14:56:53:0706 FuBackend            Paired: false
	14:56:53:0706 FuBackend            Trusted: false
	14:56:53:0706 FuBackend            Blocked: false
	14:56:53:0706 FuBackend            LegacyPairing: false
	14:56:53:0706 FuBackend            RSSI: int16 -45
	14:56:53:0706 FuBackend            Connected: false
	14:56:53:0706 FuBackend            UUIDs: ['0000110d-0000-1000-8000-00805f9b34fb', '0000110b-0000-1000-8000-00805f9b34fb', '0000110a-0000-1000-8000-00805f9b34fb', '0000110e-0000-1000-8000-00805f9b34fb', '0000110f-0000-1000-8000-00805f9b34fb', '00001101-0000-1000-8000-00805f9b34fb']
	14:56:53:0706 FuBackend            Modalias: 'bluetooth:v000ApFFFFdFFFF'
	14:56:53:0706 FuBackend            Adapter: objectpath '/org/bluez/hci0'
	14:56:53:0706 FuBackend            TxPower: int16 8
	14:56:53:0706 FuBackend            ServicesResolved: false
	14:56:53:0706 FuBackend            FuBluezDevice:
	LG Monitor(49)
	  InstanceIds:          BLE\VID_000A&PID_FFFF&REV_FFFF
	  InstanceIds:          BLE\VID_000A&PID_FFFF
	  Flags:                no-guid-matching
	  VendorId:             BLE:000A
	  Icon:                 audio-card
	  Address:              C4:30:18:6F:B5:49
	  Adapter:              /org/bluez/hci0

	14:56:53:0706 FuBackend            Address: 'D2:F6:46:4B:FC:C5'
	14:56:53:0706 FuBackend            AddressType: 'random'
	14:56:53:0706 FuBackend            Name: 'HP 910 White BT Keyboard'
	14:56:53:0706 FuBackend            Alias: 'HP 910 White BT Keyboard'
	14:56:53:0706 FuBackend            Appearance: uint16 961
	14:56:53:0706 FuBackend            Icon: 'input-keyboard'
	14:56:53:0706 FuBackend            Paired: true
	14:56:53:0706 FuBackend            Trusted: true
	14:56:53:0706 FuBackend            Blocked: false
	14:56:53:0706 FuBackend            LegacyPairing: false
	14:56:53:0706 FuBackend            Connected: false
	14:56:53:0706 FuBackend            UUIDs: ['00001800-0000-1000-8000-00805f9b34fb', '00001801-0000-1000-8000-00805f9b34fb', '0000180a-0000-1000-8000-00805f9b34fb', '0000180f-0000-1000-8000-00805f9b34fb', '00001812-0000-1000-8000-00805f9b34fb', '0000ff00-0000-1000-8000-00805f9b34fb']
	14:56:53:0706 FuBackend            Modalias: 'usb:v0461p4EEEd0001'
	14:56:53:0706 FuBackend            Adapter: objectpath '/org/bluez/hci0'
	14:56:53:0706 FuBackend            ServicesResolved: false
	14:56:53:0706 FuBackend            WakeAllowed: true
	14:56:53:0706 FuBackend            FuBluezDevice:
	HP 910 White BT Keyboard
	  InstanceIds:          USB\VID_0461&PID_4EEE&REV_0001
	  InstanceIds:          USB\VID_0461&PID_4EEE
	  Flags:                no-guid-matching
	  VendorId:             USB:0461
	  Icon:                 input-keyboard
	  Address:              D2:F6:46:4B:FC:C5
	  Adapter:              /org/bluez/hci0

src/fu-bluez-backend.c
hughsie: I think we'll have to create a proxy for this (for each device?) and then *set* the properties based on a requery. I'll leave this to @hardboprobot tho, I'm nearly out of time today.

src/fu-bluez-backend.c
hardboprobot: The whole thing is only a skeleton for now to sketch ideas on, so consider most missing parts are things that will either be done next or removed. But thanks for pointing them out, as some of them might also be wrong ;)

Also, thanks for the huge overhaul and for filling in all the fwupd-specific things I missed. There's a lot of fwupd internals I'm still not familiar with.

src/fu-bluez-backend.c
hardboprobot: Device changes such as connection and disconnection for whatever reason will be ideally updated aynchronously by catching the appropriate DBus signals, so technically that's not a problem. Now, what to do if this happens in the middle of an update is something to be defined. What's the policy for other (wired) device types?

superm1: So with hardwired devices we will get events that they're plugged/unplugged and when you run `fwupdmgr get-devices` (or a simlar dbus call) you will only see the devices presently connected to the system.  There is no messaging or concept of a "saved" device.  There was an idea posted a while back https://github.com/fwupd/fwupd/discussions/1418 which talked about remembering unconnected USB devices.

If the device gets disconnected mid-update then the USB stack gets upset and an error bubbles up.

src/fu-bluez-device.c
superm1: Is that Instance ID matching your USB dongle perhaps?

hardboprobot: I got the same modalias here for the Primax keyboard.

src/fu-bluez-backend.c
hardboprobot: I think that same method can be implemented for BLE. AFAICT all queries are cached so, by default, `fu_bluez_backend_coldplug()` will get all the devices known to the system regardless of whether they're connected or not, but we can query and filter that list keeping only the connected ones to make it similar to the behavior with wired devices.

Prototyping the signal handling for listening to this kind of events should be easy, I can work on that next. Then I have to think of ways of making all of this work more smoothly (ie. not relying so much on synchronous calls).

hughsie: > not relying so much on synchronous calls

At the moment I think that's fine. If we're blocking for multiple seconds frequently we can do something more clever, but don't stress about the sync-ness at the moment.

superm1: >fu_bluez_backend_coldplug

Since signals aren't yet there - is there also a dbus signal that we'll get when a device is **paired** after coldplug?  I think that's the most common way that devices will work that I am worried about.  System is already booted up(`bluetoothd` and `fwupd` running) and then they go into the UX to pair something.  `bluetoothd` knows about it, but `fwupd` needs to learn about it after coldplug was already done.

hughsie: > but fwupd needs to learn about it after coldplug was already done

I think we need to connect to all GDBusProxy's of all objects. I think it only makes sense to `add` on connected and `remove` on disconnected, otherwise we have a lot more state to worry about.

hardboprobot: You can listen for connection and disconnection events without having to go through all objects. A way of doing this is keeping a dbus connection to listen to all signals globally (just an idea, I tested it and it works fine).

About checking for pairing after coldplug, if necessary: I wrote a test program that notifies any device _discovery_ or _removal_. If you want to check specifically for _pairing_ I'll have to find a way of doing it (I'm very new to the Bluez api). In any case, whatever bluetoothctl can do, we can do too.

superm1: OK - If that's the case, then I don't think there is any reason to keep a connected attribute in the first place.

hughsie: > You can listen for connection and disconnection events

I don't think _connected_ is enough. I think it needs to be _paired_ too. I think GDBusObjectManager helps you with that. I can have a play too if that helps.

hughsie: @hardboprobot I've converted FuBluezBackend to use GDBusObjectManager -- I think it makes a lot more sense -- and also means we can match on devices that are connected **and** paired. I'll remove the new device flag now too.

src/fu-engine.c
hughsie: We could certainly add -Dbluez but I think it needs to be conditional on the implementation, not the abstract functionality. I think if an alternate bluetooth implementation provides enumeration, and a UUID read/write I think devices would work just as well.

libfwupdplugin/fu-ble-device.c
hughsie: Fixed!

src/fu-bluez-device.c
hughsie: I'm not so sure. Wed just physiologically just switch to `XXX`, `TODO` or something else.

superm1: Assuming this only applies to `master`, how about using this thing: https://probot.github.io/apps/todo/ with the keywords `FIXME` and `TODO`

src/fu-bluez-backend.c
hughsie: Fixed.

src/fu-bluez-device.c
hughsie: Not convinced dude; too many false positives in the existing code. Plus we use FIXME in one of the metainfo files to avoid people uploading it as-is to the LVFS...

src/fu-engine.c
superm1: OK, I think in that case `-Dbluez` is a good idea.

hughsie: Done

contrib/ci/build_windows.sh
```diff
@@ -20,6 +20,7 @@ meson .. \
     -Dplugin_uefi_capsule=false \
     -Dplugin_redfish=false \
     -Dplugin_altos=false \
+    -Dplugin_bluez=false \
```
superm1, line 23: should be `-Dbluez=false`

src/fu-engine.c
```diff
@@ -52,13 +52,17 @@
 #include "fu-security-attrs-private.h"
 #include "fu-smbios-private.h"
 #include "fu-udev-device-private.h"
+#include "fu-ble-device.h"
```
superm1, line 55: I don't think this needs to be included here

src/fu-bluez-device.h
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+#pragma once
+
+#include <gio/gio.h>
+
+#include "fu-ble-device.h"
+
+#define FU_TYPE_BLUEZ_DEVICE (fu_bluez_device_get_type ())
+G_DECLARE_FINAL_TYPE (FuBluezDevice, fu_bluez_device, FU, BLUEZ_DEVICE, FuBleDevice)
```
superm1, line 14: This should be `G_DECLARE_DERIVABLE_TYPE` no?

plugins/test/meson.build
```diff
@@ -46,3 +50,26 @@ shared_module('fu_plugin_invalid',
...
+  sources : [
+    'fu-plugin-test-ble.c',
+    'fu-test-ble-device.c',
+  ],
+  include_directories : [
+    root_incdir,
+    fwupd_incdir,
+    fwupdplugin_incdir,
+  ],
+  install : true,
```
superm1, line 65: this should only be set if `install_dummy` is set to true

plugins/test/meson.build
```diff
@@ -5,6 +5,10 @@ if get_option('plugin_dummy')
   install_dummy = true
 endif
 
+install_data(['test-ble.quirk'],
+  install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
+)
+
```
superm1, line 11: Is there another way to key off knowing this is a ble device without having to quirk?  I would think there should be something we can look at in udev for example?

plugins/test/meson.build
```diff
@@ -46,3 +50,26 @@ shared_module('fu_plugin_invalid',
     plugin_deps,
   ],
 )
+
+shared_module('fu_plugin_test_ble',
```
superm1, line 54: should this have `if get_option('bluez')` around it since we don't actually have another bluetooth implementation at the time?

plugins/test/meson.build
```diff
@@ -5,6 +5,10 @@ if get_option('plugin_dummy')
   install_dummy = true
 endif
 
+install_data(['test-ble.quirk'],
```
superm1, line 8: same comment here - should this have `if get_option('bluez')` around it?

src/fu-bluez-device.h
hughsie: Yes, and it's worse than that. @hardboprobot pointed out we can't have the FuBleDevice and FuBluezDevice split as otherwise the inorporate design doesn't work; you can't call methods on FuBluezDevice if you only subclass FuBleDevice. I've merged them back together as Google told us on Thursday that they'd be providing a bluez-compatible DBUS API for Chrome.

plugins/test/meson.build
hughsie: Maybe all BLE devices should all be BLE/ rather than USB/ afterall?

plugins/test/meson.build
hughsie: Not sure. At the moment the ble_test isn't actually a test, it's more to play with how we'd design a real plugin in the future.

plugins/test/meson.build
hughsie: yes, this should **certainly** have it; i'll add it now.

plugins/test/meson.build
superm1: I think that would certainly make it a lot clearer yes.

plugins/test/meson.build
hughsie: Thinking slowly; it's pointless to build this if we can't use it. fixed.

plugins/test/meson.build
hughsie: I went with BLUETOOTH as it's not BLE specific at all.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -0,0 +1,551 @@
...
+		}
+	}
+}
+
+/*
+ * Returns the value of a property of an object specified by its path as
+ * a GVariant, or NULL if the property wasn't found.
+ */
+static GVariant *
+fu_bluez_backend_get_property (const gchar *obj_path,
```
hardboprobot: I removed the `backend` reference from these function names in https://github.com/fwupd/fwupd/pull/2889/commits/d6ee70acd60bde531b6ba1b7db4db9ab23b91eca

hughsie: Excellent. Can you fixup the first commit with this (and maybe rebase against master too) and force push I think we're good to go.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -0,0 +1,551 @@
...
+}
+
+/*
+ * Populates the {uuid : object_path} entries of a device for all its
+ * characteristics.
+ *
+ * TODO: Extend to services and descriptors too?
+ */
+static void
+fu_bluez_backend_add_device_uuids (FuBluezDevice *self, GVariant *uuids)
```
hardboprobot: The uuids parameter is not used in the function: uuids are fetched from the ObjectManager. Fixed in https://github.com/fwupd/fwupd/commit/7f62e8095621460a3b3f77a7a43bdf486e288040

libfwupdplugin/fu-bluez-device.c
```diff
@@ -0,0 +1,551 @@
...
+}
+
+static gboolean
+fu_bluez_device_setup (FuDevice *device, GError **error)
+{
+	FuBluezDevice *self = FU_BLUEZ_DEVICE (device);
+	FuBluezDevicePrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GVariant) val_uuids = NULL;
+
+	val_uuids = g_dbus_proxy_get_cached_property (priv->proxy, "UUIDs");
```
hardboprobot: fu_bluez_backend_add_device doesn't need a list of uuids as a parameter. Besides, the "UUIDs" property here list only the Service UUIDs. Fixed in https://github.com/fwupd/fwupd/commit/7f62e8095621460a3b3f77a7a43bdf486e288040

libfwupdplugin/fu-bluez-device.c
hardboprobot: Sure, I'll squash the other fix I did too (https://github.com/fwupd/fwupd/commit/7f62e8095621460a3b3f77a7a43bdf486e288040)

hardboprobot: Done

libfwupdplugin/fu-bluez-device.c
```diff
@@ -0,0 +1,548 @@
...
+#define DEFAULT_PROXY_TIMEOUT	5000
+
+/**
+ * SECTION:fu-bluez-device
+ * @short_description: a Bluez Bluetooth LE device
+ *
+ * An object that represents a Bluez Bluetooth LE device.
+ *
+ * See also: #FuBluezDevice
+ */
```
superm1, line 27: Is there any reason this is BLE only?  I would think bluez will provide both LE and full energy devices.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -0,0 +1,548 @@
...
+		}
+	}
+}
+
+/*
+ * Returns the value of a property of an object specified by its path as
+ * a GVariant, or NULL if the property wasn't found.
+ */
+static GVariant *
+fu_bluez_device_get_ble_property (const gchar *obj_path,
```
superm1, line 127: similarly here, shouldn't this just be something like `get_bt_property`?

plugins/test/meson.build
```diff
@@ -46,3 +52,26 @@ shared_module('fu_plugin_invalid',
     plugin_deps,
   ],
 )
+
+shared_module('fu_plugin_test_ble',
```
superm1, line 56: Should this plugin rename to `test_bluetooth` or `test_bluez`?

libfwupdplugin/fu-bluez-device.c
hardboprobot: Bluez supports both Bluetooth classic and BLE, but both stacks are very different from a programming point of view. This device implements a BLE device specifically. Maybe the naming should reflect both things: that this is a BLE device and that uses Bluez as a backend.

We decided to focus on the Bluez-specific devices for now, as those are the use cases we're going to be dealing with in the near future. Ideally, this should be redesigned to be stack-agnostic.

superm1: >that this is a BLE device and that uses Bluez as a backend.

Yes, I think it's a bit confusing otherwise.

libfwupdplugin/fu-bluez-device.c
hardboprobot: Honestly, I don't have enough experience with Bluetooth Classic to foresee how much of the current backend+device can be generalized to both BT Classic and BLE, but that's definitely in the roadmap.

plugins/test/meson.build
superm1: Given other comment, this name selection seems accurate.  However I want to point out it's still pointless to compile this plugin when you're compiling without `-Dbluez=true`.  I think this plugin shared module declaration should have a guard around that still.

libfwupdplugin/fu-bluez-device.c
hughsie: I think we can just `s/LE/`

libfwupdplugin/fu-bluez-device.c
hardboprobot: Fixed

plugins/test/meson.build
hardboprobot: Fixed

superm1 (approved):

PR #2890, opened 2021-02-16
Add fu_firmware_set_version_raw()

hughsie:
A few plugins need this now, so move it into the base class.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-firmware.c
+++ b/libfwupdplugin/fu-firmware.c
@@ -24,6 +24,7 @@ typedef struct {
 	FuFirmwareFlags			 flags;
 	GPtrArray			*images;	/* FuFirmwareImage */
 	gchar				*version;
+	guint64				 version_raw;
 } FuFirmwarePrivate;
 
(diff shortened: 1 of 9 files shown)


superm1 (commented):
Can you please better explain why any plugins _need_ this?  Is it just a debugging feature?  I don't see a _need_ from the diff.

hughsie:
>  Is it just a debugging feature? I don't see a need from the diff.

Fair point. Debugging yes, but main reason is the ccgx and pixart plugins need this in the other two PRs to be able to *build* firmware of a specific version.

superm1:
>Fair point. Debugging yes, but main reason is the ccgx and pixart plugins need this in the other two PRs to be able to build firmware of a specific version.

Haven't looked at the others yet, so that makes more sense knowing there is an actual consumer of it.  I think from a plugin developer perspective there needs to be better explanation of the different between this and regular `version` and whether this is mandatory or optional.

hughsie:
> there needs to be better explanation of the different between this and regular version

I've added gtk-doc comments in both functions; how about that?

superm1 (approved):

PR #2892, opened 2021-02-16
contrib/PKGBUILD: Ensure pkgver gets expanded as an bash variable

Foxboron:
Signed-off-by: Morten Linderud <morten@linderud.pw>

Type of pull request:
- [x] Code fix


FFY00 (approved):
:+1:

During the build process, `pkgver()` is run and `$pkgver` is set. `pkgver()` shouldn't be used directly, `$pkgver` should be used instead.

PR #2893, opened 2021-02-16
Do not fail to start the daemon for a single backend setup failure

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5580,6 +5580,8 @@ fu_engine_plugin_recoldplug_cb (FuPlugin *plugin, FuEngine *self)
 		for (guint i = 0; i < self->backends->len; i++) {
 			FuBackend *backend = g_ptr_array_index (self->backends, i);
 			g_autoptr(GError) error_local = NULL;
+			if (!fu_backend_get_enabled (backend))
+				continue;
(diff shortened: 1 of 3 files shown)


src/fu-engine.c
```diff
@@ -6385,8 +6387,13 @@ fu_engine_load (FuEngine *self, FuEngineLoadFlags flags, GError **error)
 	/* set up backends */
 	for (guint i = 0; i < self->backends->len; i++) {
 		FuBackend *backend = g_ptr_array_index (self->backends, i);
-		if (!fu_backend_setup (backend, error))
-			return FALSE;
+		g_autoptr(GError) error_backend = NULL;
```
superm1, line 6390: I think you need to special case 1 backend. If compiled with one backend it should be fatal no?

hughsie: Would it be fair to say that we should return a failure if they _all_ fail? In the case of win32 with just gusb we'd fail as you suggest, and for Linux embedded it's unlikely we'd fail on embedded without bluez.

superm1: Yeah I think that's reasonable.

hughsie: Okay, something like that^^?

superm1 (approved):

PR #2895, opened 2021-02-17
plugins: rts54hub: optimized rtd21xx ISP flow

AnyProblem:
Optimized and modified rtd21xx ISP flow

Signed-off-by: RickyWu <ricky_wu@realtek.com>

Type of pull request:
- [ *] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
+++ b/plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
@@ -29,8 +29,8 @@ G_DEFINE_TYPE (FuRts54hubRtd21xxDevice, fu_rts54hub_rtd21xx_device, FU_TYPE_DEVI
 #define UC_FOREGROUND_OPCODE		0x33
 #define UC_FOREGROUND_ISP_DATA_OPCODE	0x34
 
-#define ISP_DATA_BLOCKSIZE		30
-#define ISP_PACKET_SIZE			32
+#define ISP_DATA_BLOCKSIZE		256
+#define ISP_PACKET_SIZE			257
 
 typedef enum {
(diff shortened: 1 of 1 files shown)


plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -357,6 +368,7 @@ fu_rts54hub_rtd21xx_device_detach (FuDevice *device, GError **error)
 static gboolean
 fu_rts54hub_rtd21xx_device_attach (FuDevice *device, GError **error)
 {
+#if 0
```
hughsie, line 371: How does the rtd21xx get back into runtime mode?

AnyProblem: this i2c_write is for changing slave addr 0x3A(ISP mode ) back to 0x6A(normal mode)
But now I move this flow to after read version so it will work fine,
and after update FW are going to reset, it will be in normal mode and work fine too

In original case, after update FW and reset, in that time slave addr is 0x6A(Normal mode) but FWUPD will call fu_rts54hub_rtd21xx_device_attach it will use 0x3A to send i2c_write, it always be stalled by rtd21xx

superm1: @AnyProblem I think the point is this shouldn't be committed as `#if 0` code - that's indicative of a problem that needs to be explained.  so if it's intended behavior then the code should be removed instead and a comment left explaining why it's not needed in it's place.

AnyProblem: @superm1 OK, I got it, I will remove #if 0 code and re-push again

PR #2896, opened 2021-02-17
Refine plugins/meson.build

superm1:
I find this file incredibly difficult to parse, especially with nested `if`s.

I've sorted it, indented it and adjusted the ones that are Linux specific to be in their own grouping.  We can now make more  concious decisions which plugins will be Linux only initially.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Being honest, I think putting the plugin deps and requirements in the plugins/meson.build is probably an antipattern. Swapping a gudev dep to a gusb dep should involve just changing the plugin/foo/* files, not rearranging the order in plugins/meson.build

Counterproposal in https://github.com/fwupd/fwupd/pull/2898 -- if you think we need the 2-space indent in each plugins meson.build just say, I can easily add it.

superm1:
>Counterproposal in #2898 

OK let's chat there.

>  if you think we need the 2-space indent in each plugins meson.build just say, I can easily add it.

I don't think it's as important there, no.

PR #2898, opened 2021-02-18
trivial: Rename meson option tpm->plugin_tpm

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


plugins/synaptics-rmi/meson.build
```diff
@@ -1,3 +1,4 @@
+if get_option('gudev') and get_option('plugin_synaptics_rmi')
```
superm1, line 1: this should follow the pattern of showing an error if plugin flag set but not gudev set

plugins/meson.build
```diff
@@ -1,149 +1,66 @@
...
 subdir('ep963x')
+subdir('fastboot')
+subdir('flashrom')
+subdir('fresco-pd')
+subdir('goodix-moc')
+subdir('hailuck')
 subdir('iommu')
+subdir('jabra')
 subdir('linux-lockdown')
 subdir('linux-sleep')
```
superm1, line 29: I still think that we should make it clear that a few plugins like `linux-lockdown` and `linux-sleep` are truly linux only like my original PR had.

superm1 (commented):
overall I think this is a better approach, yes

plugins/meson.build
hughsie: Ohh, I missed that bit. Got a list offhand of the ones you worked out were linux specific?

superm1: These were the ones I made out, feel free to disagree or add anything else you think though.

```
  subdir('acpi-dmar')       #/sys/firmware/acpi
  subdir('acpi-facp')       #/sys/firmware/acpi
  subdir('iommu')           #linux kernel iommu
  subdir('linux-lockdown')
  subdir('linux-sleep')
  subdir('linux-swap')
  subdir('linux-tainted')
  subdir('pci-bcr')         #pci
  subdir('pci-mei')         #pci
  subdir('upower')
  subdir('pixart-rf')       #linux kernel hidraw
```

hughsie: Do we want to run acpi-facp-self-test on !linux? I'm erring on yes.

plugins/synaptics-rmi/meson.build
hughsie: Agree, fixed.

superm1 (approved):
Lgtm now

PR #2899, opened 2021-02-18
Call the superclass directly from subclassed devices

hughsie:
This allows a device subclass to call the parent method after doing an initial
action, or even deliberately not call the *generic* parent method at all.

It also simplifies the plugins; you no longer have to remember what the plugin
is deriving from and accidentally clobber the wrong superclass method.

Type of pull request:
- [X] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/README.md
+++ b/libfwupdplugin/README.md
@@ -6,3 +6,6 @@ Planned API/ABI changes for next release
  * Remove fu_plugin_get_usb_context() and fu_plugin_set_usb_context()
  * Remove fu_plugin_runner_usb_device_added(), fu_plugin_runner_udev_device_added()
    and fu_plugin_runner_udev_device_changed()
+ * Remove FuHidDevice->open() and FuHidDevice->close()
+ * Remove FuUsbDevice->probe(), FuUsbDevice->open() and FuUsbDevice->close()
+ * Remove FuUdevDevice->to_string(), FuUdevDevice->probe(), FuUdevDevice->open() and FuUdevDevice->close()
--- a/libfwupdplugin/fu-hid-device.c
+++ b/libfwupdplugin/fu-hid-device.c
@@ -78,14 +78,18 @@ fu_hid_device_set_property (GObject *object, guint prop_id,
 }
 
 static gboolean
-fu_hid_device_open (FuUsbDevice *device, GError **error)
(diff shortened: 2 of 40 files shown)


libfwupdplugin/fu-hid-device.h
```diff
@@ -16,11 +16,7 @@ G_DECLARE_DERIVABLE_TYPE (FuHidDevice, fu_hid_device, FU, HID_DEVICE, FuUsbDevic
 struct _FuHidDeviceClass
 {
 	FuUsbDeviceClass parent_class;
-	gboolean	 (*open)		(FuHidDevice		*device,
-						 GError			**error);
-	gboolean	 (*close)		(FuHidDevice		*device,
-						 GError			**error);
-	gpointer	__reserved[29];
+	gpointer	__reserved[31];
```
superm1, line 19: ABI change no?

libfwupdplugin/fu-udev-device.h
```diff
@@ -23,16 +23,7 @@ G_DECLARE_DERIVABLE_TYPE (FuUdevDevice, fu_udev_device, FU, UDEV_DEVICE, FuDevic
...
-							 GError		**error);
-	gboolean	 (*open)			(FuUdevDevice	*device,
-							 GError		**error);
-	gboolean	 (*close)			(FuUdevDevice	*device,
-							 GError		**error);
-	void		 (*to_string)			(FuUdevDevice	*self,
-							 guint		 indent,
-							 GString	*str);
-	gpointer	__reserved[28];
+	gpointer	__reserved[31];
```
superm1, line 26: same comment as `fu-hid-device.h` - ABI change no?

libfwupdplugin/fu-hid-device.h
hughsie: Yes, absolutely. I pondered leaving the code so that we called `->close()` like before for any plugins in development (maybe with `g_warning()`) and also pondered just leaving the vfuncs with an `/* unused */` comment. Ideas?

hughsie: I added them back with g_warning() and added them to README.txt for the next time we bump the API. You're completely correct, again :)

superm1 (approved):

PR #2904, opened 2021-02-18
1.4.x: Disable Travis CI, and enable Github CI instead

superm1:
I think this will be more stable for something that doesn't change much and if it works here well, can bring it back to 1.3.x. and 1.2.x.

All the builds will use the same containers this means though, so "dropping" dependencies will have a negative effect.  That doesn't happen to us very frequently though.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
I think you could squash a few of these together, but I can't look a gift horse in the mouth. Thanks for doing this!

PR #2909, opened 2021-02-19
trivial: install any missing dependencies in containers

superm1:
some of these might not be present if the containers got out of
sync.  it will be a no-op at worst.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
This is mostly to try to future proof CI when this eventually branches to 1_5_X and we keep using the same containers but make some small changes.

hughsie (approved):

Issue #2910, opened 2021-02-19
Setting efi_sbat_distro vars corrupts EFI program


tohojo:
**Describe the bug**
I've been having some trouble with the Arch package of fwupd, see https://bugs.archlinux.org/task/69694

The issue I'm seeing is that when trying to launch the EFI binary, I get an "unsupported program" error from systemd-boot or the EFI shell (the system firmware just fails with no error message, so that's why I resorted to those options for launching).

In the bug report linked above we narrowed it down to the `efi_sbat_distro` config. Specifically, the Arch PKGBUILD passes these parameters to 'meson build':

```
        -D efi_sbat_distro_id='arch' \
        -D efi_sbat_distro_summary='Arch Linux' \
        -D efi_sbat_distro_pkgname=${pkgname} \
        -D efi_sbat_distro_version=${pkgver} \
        -D efi_sbat_distro_url="https://archlinux.org/packages/community/x86_64/${pkgname}/"
```

When I recompile the package without those parameters the EFI binary works fine, but with them I get the "unsupported program" error. So I'm opening this issue on request of the distro maintainer.

superm1:
@hughsie  Is this the stuff that was fixed by https://github.com/fwupd/fwupd/pull/2902 perhaps?  And then also maybe the root cause of https://github.com/fwupd/firmware-dell/issues/53?

tohojo:
Sorry, I should have mentioned, we did take a detour around #2902 and that was part of the issue why things didn't work (see discussion in the bug I linked above). However, even when backporting commit ec52942bc5ad9e80c8bf905e6d0a0ccf218acc1f (which the Arch package does), the issue remains when the `efi_sbat_distro` vars are set...

superm1:
@martinezjavier  ^

superm1:
@martinezjavier is this perhaps an ordering problem?  Does the `sbat` section need to be placed at the end?  Something like this (untested - no idea if this still or now works):

```
diff --git a/plugins/uefi-capsule/efi/generate_binary.py b/plugins/uefi-capsule/efi/generate_binary.py
index 06a43725..bd8b47bf 100755
--- a/plugins/uefi-capsule/efi/generate_binary.py
+++ b/plugins/uefi-capsule/efi/generate_binary.py
@@ -84,8 +84,6 @@ def _run_objcopy(args):
             "-j",
             ".text",
             "-j",
-            ".sbat",
-            "-j",
             ".sdata",
             "-j",
             ".data",
[...]

PR #2913, opened 2021-02-20
uefi-capsule: Align the SBAT section

hughsie:
Possibly fixes https://github.com/fwupd/fwupd/issues/2910

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/efi/generate_binary.py
+++ b/plugins/uefi-capsule/efi/generate_binary.py
@@ -74,6 +74,45 @@ def _run_objcopy_sbat(args, tfd):
         subprocess.run(argv, check=True)
 
 
+def _get_sbat_objcopy_argv(args):
+
+    _SBAT_SECTION_ALIGNMENT = 512
+    _SBAT_SECTION_VMA = 0x2000
+
+    if not args.sbat_distro_id:
+        return []
+
+    argv = [
+        "-j",
+        ".sbat",
+    ]
+    has_sbat_in_lds = False
+    if args.linker_script:
+        with open(args.linker_script, "r") as flinker:
+            has_sbat_in_lds = flinker.read().find(".sbat") != -1
+        if not has_sbat_in_lds:
+            print(
+                "No .sbat in {}, assuming alignment {} and vma 0x{:04x}",
+                args.linker_script,
+                _SBAT_SECTION_ALIGNMENT,
+                _SBAT_SECTION_VMA,
+            )
+    else:
+        print(
+            "No lds specified, assuming alignment {} and vma 0x{:04x}",
+            _SBAT_SECTION_ALIGNMENT,
+            _SBAT_SECTION_VMA,
+        )
+    if not has_sbat_in_lds:
+        argv += [
+            "--set-section-alignment",
+            ".sbat={}".format(_SBAT_SECTION_ALIGNMENT),
+            "--change-section-vma",
+            ".sbat=0x{:04x}".format(_SBAT_SECTION_VMA),
+        ]
+    return argv
+
+
 def _run_objcopy(args):
 
     with tempfile.NamedTemporaryFile() as tfd:
@@ -84,8 +123,6 @@ def _run_objcopy(args):
             "-j",
             ".text",
             "-j",
-            ".sbat",
-            "-j",
             ".sdata",
             "-j",
             ".data",
@@ -95,6 +132,9 @@ def _run_objcopy(args):
             ".dynsym",
             "-j",
             ".rel*",
+        ]
+        argv += _get_sbat_objcopy_argv(args)
+        argv += [
             tfd.name,
             args.outfile,
         ]
@@ -193,6 +233,11 @@ if __name__ == "__main__":
         "--sbat-distro-url",
         default=None,
         help="SBAT distribution URL",
+    )
+    parser.add_argument(
+        "--linker-script",
+        default=None,
+        help="Linker script, e.g. /usr/lib/gnuefi/x64/efi.lds",
     )
     parser.add_argument(
         "infile",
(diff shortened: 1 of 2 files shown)


hughsie:
With this:

	$ readpe plugins/uefi-capsule/efi/fwupdx64.efi -s
	############
	# SECTIONS #
	############
	╒══════════╤═════════════╤════════════╤═════════════════╤════════════════╤═══════════════════╤═════════╤═══════════╕
	│ Name     │ Raw Addr.   │ Raw Size   │ Virtual Addr.   │ Virtual Size   │ Characteristics   │ Flags   │   Entropy │
	╞══════════╪═════════════╪════════════╪═════════════════╪════════════════╪═══════════════════╪═════════╪═══════════╡
	│ .sbat    │ 0x400       │ 0x200      │ 0x0             │ 0xdf           │ 0x40a00000        │ R--     │ 3.08394   │

plugins/uefi-capsule/efi/generate_binary.py
```diff
@@ -95,6 +93,15 @@ def _run_objcopy(args):
...
             "-j",
             ".rel*",
+        ]
+        if args.sbat_distro_id:
+            argv += [
+                "-j",
+                ".sbat",
+                "--set-section-alignment",
+                ".sbat=512",
+            ]
```
martinezjavier, line 103: I think we always want to do the explicit alignment and not only when sbat_distro_id is defined, but I let @vathpela and @ chrisccoulson to confirm that’s the case.

hughsie: I moved the `-j .sbat` too

martinezjavier: Ah, right. Looks good to me.

martinezjavier (approved):

plugins/uefi-capsule/efi/generate_binary.py
```diff
@@ -95,6 +93,17 @@ def _run_objcopy(args):
...
+        ]
+        if args.sbat_distro_id:
+            argv += [
+                "-j",
+                ".sbat",
+                "--set-section-alignment",
+                ".sbat=512",
+                "--change-section-vma",
+                ".sbat=0x2000",
+            ]
```
martinezjavier, line 105: The changes looks good to me and it leads to a binary with proper alignment and a VMA that doesn't cause issues for `edk2`. I would rather prefer this small change than pulling all the linker scripts, but wasn't sure if it was a safe enough option to make these assumptions about the default `gnu-efi` linker script.

As mentioned in the issue, if these values are not good for any distro then probably we should provide some meson options to set them if that was the case. I would like to know @vathpela @chrisccoulson option on this.

plugins/uefi-capsule/efi/generate_binary.py
```diff
@@ -95,6 +93,17 @@ def _run_objcopy(args):
             ".dynsym",
             "-j",
             ".rel*",
+        ]
+        if args.sbat_distro_id:
+            argv += [
+                "-j",
+                ".sbat",
+                "--set-section-alignment",
+                ".sbat=512",
```
martinezjavier, line 102: Maybe add a comment here about why this value is chosen? Something like:

```
# The PE sections must begin on a multiple of the file
# alignment, which usually is 512 bytes.
```

plugins/uefi-capsule/efi/generate_binary.py
```diff
@@ -95,6 +93,17 @@ def _run_objcopy(args):
...
+                "--change-section-vma",
+                ".sbat=0x2000",
```
martinezjavier, line 104: same for this to not be a magic value, i.e:

```
# The gnu-efi linker script efi.lds places the .text section at the
# virtual memory address 0x4000. Place the .sbat section below this.
```

chrisccoulson: I don't think the linker script specifies the virtual address of the .text section though - it only specifies page alignment. I think the address is determined by the linker based on that alignment and the size of the preceding sections. In the example in https://github.com/fwupd/fwupd/pull/2914#issuecomment-782780063, the .text section would be placed at 0x3000 if the .eh_frame section were 284 bytes smaller.

martinezjavier: @chrisccoulson is there a fixed address that you think could be safe to use? Since as you mentioned the sections are page aligned by the `efi.lds` linker script and there are a couple of sections before `.text` (that are discarded when creating the PE binary), so at the very least `.text` would be above `0x1000` IIUC.

plugins/uefi-capsule/efi/generate_binary.py
chrisccoulson: This will probably create a working amd64 binary, but the aarch64 linker script places the .text section first. I've just cross-compiled shim for aarch64 and the .text section ends up at 0x0 in the ELF binary and 0x1000 in the final PE binary, so this would require a different solution for aarch64.

martinezjavier: Urgh, how complicated... so that header that edk2 places at VMA 0x0 is not added for aarch64 ?

martinezjavier: This is the reason why I think all this logic has to be handled in the system linker scripts instead of each project having to deal with it.

chrisccoulson: I was confused about how the .text section at offset 0x0 in the ELF binary ended up at 0x1000 in the final PE binary, but it's because gnu-efi defines the PE header in assembly at the start of the ELF .text section to work around the lack of native support in binutils. So this isn't going to work on aarch64 at all right now :/

hughsie: I guess that's the nail in the coffin of this PR. Thanks for the explanation!

PR #2914, opened 2021-02-20
uefi-capsule: Ensure SBAT metadata is added correctly

chrisccoulson:
The current approach of adding SBAT metadata after linking is creating
an image that is badly formed in 2 ways:
- The SBAT section's file offset and size are not a multiple of the
file alignment.
- The SBAT section has a virtual address of zero. EDK2 loads the header
here, and so it gets rejected.

This changes the approach to match shim, where an object file is
created with a .sbat section and then the linker takes care of placing
the section at a more appropriate virtual address.

Unfortunately, this requires shim's linker scripts to be imported (minus
the .vendor_cert section). I'm not sure if there's a better way to do this.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@vathpela is there a better way to do this? The linker scripts are kinda fugly.

hughsie:
This PR also built just fine on the aarch64 Fedora builders, so many thanks for tracking that down properly.

martinezjavier:
> @vathpela is there a better way to do this? The linker scripts are kinda fugly.

I spent some time today trying to explore alternative solutions to this problem that wouldn't require pulling all these linker scripts. I didn't find any other solution that would be more reliable to what @chrisccoulson proposed, but I'll just share my findings anyways to evaluate all the possible options.

The first issue mentioned about the file alignment can be solved by using `--set-section-alignment` as @hughsie did in PR #2913. But we still have the second issue about the `.sbat` section VMA being zero that causes issues with `edk2`.

The problem is what @chrisccoulson mentioned in #2910, that if a linker script isn't used to specify the output `.sbat` section address, then the `objcopy` invocation will have to make assumptions about the sections layout created by the default linker script from `gnu-efi` (i.e: `/usr/lib/gnuefi/x64/efi.lds` in Fedora).

The `.efi.lds` script places the `.text` section at address `0x4000` and all the sections that come before `.text` section are discarded when creating the `fwupdx64.efi` PE binary. This means that there is a gap in the virtual address space between `0x0` and `0x4000` that in theory could be used to place other sections.

These are the sections from the `fwup.so` that are copied to the PE binary:

```
$ objdump -h build/plugins/uefi-capsule/efi/fwup.so 

build/plugins/uefi-capsule/efi/fwup.so:     file format elf64-x86-64

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .hash         000000d8  0000000000000000  0000000000000000  00001000  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  1 .gnu.hash     000000f4  00000000000000d8  00000000000000d8  000010d8  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  2 .eh_frame     0000211c  0000000000001000  0000000000001000  00002000  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .text         00007d10  0000000000004000  0000000000004000  00005000  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  4 .plt          00000020  000000000000bd10  000000000000bd10  0000cd10  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  5 .reloc        0000000a  000000000000c000  000000000000c000  0000d000  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  6 .data         00002d28  000000000000d000  000000000000d000  0000e000  2**5
                  CONTENTS, ALLOC, LOAD, DATA
  7 .dynamic      00000140  0000000000010000  0000000000010000  00011000  2**3
                  CONTENTS, ALLOC, LOAD, DATA
  8 .dynsym       00000348  0000000000012000  0000000000012000  00013000  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  9 .rela         00000e70  0000000000011000  0000000000011000  00012000  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
 10 .rela.plt     00000018  0000000000011e70  0000000000011e70  00012e70  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
 11 .dynstr       000001bb  0000000000013000  0000000000013000  00014000  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
 12 .gcc_except_table 000000d7  0000000000014000  0000000000014000  00015000  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
 13 .comment      0000005c  0000000000000000  0000000000000000  000150d7  2**0
                  CONTENTS, READONLY
 14 .debug_line   000019ed  0000000000000000  0000000000000000  00015133  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 15 .debug_info   00007abf  0000000000000000  0000000000000000  00016b20  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 16 .debug_abbrev 00000de1  0000000000000000  0000000000000000  0001e5df  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 17 .debug_aranges 000000f0  0000000000000000  0000000000000000  0001f3c0  2**4
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 18 .debug_str    0000f181  0000000000000000  0000000000000000  0001f4b0  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 19 .debug_loc    00002fe6  0000000000000000  0000000000000000  0002e631  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 20 .debug_ranges 00000230  0000000000000000  0000000000000000  00031617  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
 21 .debug_macro  00002944  0000000000000000  0000000000000000  00031847  2**0
                  CONTENTS, READONLY, DEBUGGING, OCTETS
```

see how `.hash`, `gnu.hash` and `.eh_frame` are before `.text` but this are not included in the `objcopy` invocation in the `generate_binary.py` script.

These are the sections that end in the `fwupdx64.efi` binary:

```
objdump -h build/plugins/uefi-capsule/efi/fwupdx64.efi 

build/plugins/uefi-capsule/efi/fwupdx64.efi:     file format pei-x86-64

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .sbat         000000ed  0000000000000000  0000000000000000  000002c8  2**0
                  CONTENTS, READONLY
  1 .text         00007d10  0000000000004000  0000000000004000  00000600  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  2 .reloc        0000000a  000000000000c000  000000000000c000  00008400  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .data         00002d28  000000000000d000  000000000000d000  00008600  2**5
                  CONTENTS, ALLOC, LOAD, DATA
  4 .dynamic      00000140  0000000000010000  0000000000010000  0000b400  2**3
                  CONTENTS, ALLOC, LOAD, DATA
  5 .rela         00000e70  0000000000011000  0000000000011000  0000b600  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  6 .rela.plt     00000018  0000000000011e70  0000000000011e70  0000c670  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  7 .dynsym       00000348  0000000000012000  0000000000012000  0000ca00  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
```

The alignment `2**0` and VMA `0x0` for the `.sbat` section are wrong, and as @chrisccoulson mentioned trying to execute this on `edk2/ovmf` will fail:

```
FS0:\EFI\fedora\> fwupdx64.efi
Command Error Status: Unsupported
```

Now, we could place the `.sbat` section in an address lower than `0x4000`, for example `0x2000`. And also align the section when copying, i.e:

```
diff --git a/plugins/uefi-capsule/efi/generate_binary.py b/plugins/uefi-capsule/efi/generate_binary.py
index 06a43725021..605c90ca4f2 100755
--- a/plugins/uefi-capsule/efi/generate_binary.py
+++ b/plugins/uefi-capsule/efi/generate_binary.py
@@ -95,6 +95,10 @@ def _run_objcopy(args):
             ".dynsym",
             "-j",
             ".rel*",
+            "--set-section-alignment",
+            ".sbat=512",
+            "--change-section-vma",
+            ".sbat=0x2000",
             tfd.name,
             args.outfile,
         ]
```

This will result in the following sections in the `fwupdx64.efi` binary:

```
objdump -h build/plugins/uefi-capsule/efi/fwupdx64.efi 

build/plugins/uefi-capsule/efi/fwupdx64.efi:     file format pei-x86-64

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .sbat         000000ed  0000000000002000  0000000000002000  00000400  2**9
                  CONTENTS, READONLY
  1 .text         00007d10  0000000000004000  0000000000004000  00000600  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  2 .reloc        0000000a  000000000000c000  000000000000c000  00008400  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .data         00002d28  000000000000d000  000000000000d000  00008600  2**5
                  CONTENTS, ALLOC, LOAD, DATA
  4 .dynamic      00000140  0000000000010000  0000000000010000  0000b400  2**3
                  CONTENTS, ALLOC, LOAD, DATA
  5 .rela         00000e70  0000000000011000  0000000000011000  0000b600  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  6 .rela.plt     00000018  0000000000011e70  0000000000011e70  0000c670  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  7 .dynsym       00000348  0000000000012000  0000000000012000  0000ca00  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
```

In this case both the alignment (`2**9`) and VMA (`0x2000`) won't cause issues. And this binary can be executed by `ovmf`:

```
FS0:\EFI\fedora\> fwupdx64.efi
WARNING: No updates to process.  Called in error?
```

A third option would be to patch the default linker script in `gnu-efi` to be aware of the `.sbat` section. The advantage is that we wouldn't have to have bespoke linker scripts in `fwupd` nor make assumptions of the sections layout, but the disadvantage is that `fwupd` will require collaboration from the distros in order to build proper EFI binaries.

plugins/uefi-capsule/efi/generate_sbat.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2021 Javier Martinez Canillas <javierm@redhat.com>
+# Copyright (C) 2021 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+#
+# pylint: disable=missing-docstring, invalid-name
+
```
martinezjavier, line 9: I think that would be better to split this in two different patches. One that creates an ELF object file with the `.sbat` section and another one that adds the custom linker scripts and uses those instead of `efi.lds` from `gnu-efi`.

plugins/uefi-capsule/efi/lds/elf_ia64_efi.lds
```diff
@@ -0,0 +1,93 @@
+OUTPUT_FORMAT("elf64-ia64-little")
+OUTPUT_ARCH(ia64)
```
martinezjavier, line 2: I think this linker script can be dropped, since `fwupd` doesn't support `ia64`.

plugins/uefi-capsule/efi/lds/elf_ia32_efi.lds
```diff
@@ -0,0 +1,87 @@
+OUTPUT_FORMAT("elf32-i386", "elf32-i386", "elf32-i386")
+OUTPUT_ARCH(i386)
```
martinezjavier, line 2: Is this linker script really needed ?

hughsie: Agree, we don't build the efi binary for this arch.

plugins/uefi-capsule/efi/lds/elf_ia64_efi.lds
hughsie: Agree.

plugins/uefi-capsule/efi/meson.build
```diff
@@ -31,38 +31,37 @@ if not have_gnu_efi
...
   endif
   if cmd.returncode() != 0
-    error('Cannot find @0@'.format(arch_lds))
+    error('Cannot find @0@'.format(arch_crt))
   endif
 endif
 
 message('efi-libdir: "@0@"'.format(efi_libdir))
-message('efi-ldsdir: "@0@"'.format(efi_ldsdir))
+message('efi-crtdir: "@0@"'.format(efi_crtdir))
```
hughsie, line 64: Is there a reason for this change? I personally think `ldsdir` is a more helpful name. I also think changing the meson config option is kinda antisocial for distros.

hughsie: I've split out something friendlier in https://github.com/fwupd/fwupd/pull/2921

PR #2919, opened 2021-02-22
Wip/hughsie/no guid matching

hughsie:
Type of pull request:
- [X] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 any ideas here? This is making my head explode.

libfwupdplugin/fu-device.c
```diff
@@ -218,6 +222,12 @@ fu_device_add_internal_flag (FuDevice *self, FuDeviceInternalFlags flag)
 	FuDevicePrivate *priv = GET_PRIVATE (self);
 	g_return_if_fail (FU_IS_DEVICE (self));
 	priv->internal_flags |= flag;
+
+	/* backwards compat */
+	if (flag & FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID)
+		fu_device_add_flag (self, FWUPD_DEVICE_FLAG_NO_GUID_MATCHING);
```
superm1, line 228: is this really the right match to add to the device side?  It seems like an opposite behavior.

hughsie: my original idea was to force the plugin to either opt-in with FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or opt-out with FWUPD_DEVICE_FLAG_NO_GUID_MATCHING -- although now I'm checking all three before the g_warning. I agree it's confusing. I think we can just drop these.

hughsie: Removed.

plugins/dell/fu-self-test.c
```diff
@@ -82,9 +82,8 @@ fu_plugin_dell_tpm_func (gconstpointer user_data)
 	g_autoptr(GPtrArray) devices = NULL;
 
 #ifdef HAVE_GETUID
-	if (tpm_server_running == NULL &&
-	    (getuid () != 0 || geteuid () != 0)) {
-		g_test_skip ("TPM tests require simulated TPM2.0 running or need root access with physical TPM");
+	if (tpm_server_running == NULL) {
+		g_test_skip ("TPM tests require simulated TPM2.0 running");
```
superm1, line 86: this doesn't seem correct to change - it is supposed to work with a real tpm OR with root.

hughsie: Ya, it was more of a way of testing if it was a physical TPM issue or a logic issue. Given the arch virtual TPM failed too I guess I need to debug this a bit more. EOUTOFTIME for today.

hughsie: Worked it out.

superm1:
@hughsie I think 764221e14b2bef664c9ff10ae718d1f6a6442e34 should go to 1.5.7, but the other commit to 1.5.8.

libfwupdplugin/fu-device.h
```diff
@@ -204,6 +204,8 @@ FuDevice	*fu_device_new				(void);
  * @FU_DEVICE_INTERNAL_FLAG_MD_SET_VERFMT:		Set the device version format from the metadata if available
  * @FU_DEVICE_INTERNAL_FLAG_MD_SET_ICON:		Set the device icon from the metadata if available
  * @FU_DEVICE_INTERNAL_FLAG_RETRY_OPEN:			Retry the device open up to 5 times if it fails
+ * @FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID:		Match GUIDs on device replug
+ * @FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID:	Match GUIDs on device coldplug
```
superm1, line 208: For any future plugin authors I think that this needs an explicit paragraph in documentation contrasting default behavior to this and the other flag.  It's all super confusing otherwise.

superm1:
On my side, it seems to make a lot of noise about devices I have.

```
Loading…                 [ -                                     ]19:16:14:0059 FuEngine             device  RTS5487 in Dell dock [d706b144f2b81e7242f731391be6356ea8a9cc94] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
Loading…                 [  -                                    ]19:16:14:0101 FuEngine             device  WD19TB [981a41a15b4c9689a2d75cc26b87bd4dbee87b67] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
19:16:14:0102 FuEngine             device  Package level of Dell dock [499ab652f9b6f9aa0f1ff5f11b4154b63c323b2a] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
19:16:14:0102 FuEngine             device  VMM5331 in Dell dock [ac15723d34d11b8bee8a7e0e2fb74ba055b85deb] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
19:16:14:0104 FuEngine             device  Thunderbolt controller in Dell dock [ef6906e96ae76ab8d5d2f4f8dfd3c337d2c0e79c] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
19:16:14:0115 FuEngine             device  RTS5413 in Dell dock [374167aba0ff3cd1678af3296ca10dadea48db5f] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
19:16:17:0038 FuEngine             device  Thunderbolt host controller [f78dd6bb609522e31f1ff2ff819827108606d18a] does not set either FWUPD_DEVICE_FLAG_NO_GUID_MATCHING, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID or FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID, so setting both internal flags!
```

hughsie:
> it seems to make a lot of noise about devices I have

Excellent! Can you choose the appropriate flags and add to the branch please? 🙂

src/fu-device-list.c
```diff
@@ -723,7 +723,7 @@ fu_device_list_add (FuDeviceList *self, FuDevice *device)
 	if (item != NULL &&
 	    g_strcmp0 (fu_device_get_protocol (item->device),
 		       fu_device_get_protocol (device)) == 0) {
-		if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_NO_GUID_MATCHING)) {
+		if (fu_device_has_internal_flag (device, FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID)) {
 			if (fu_device_get_priority (device) < fu_device_get_priority (item->device)) {
 				g_debug ("ignoring device %s [%s:%s] as better device %s [%s:%s] already exists",
 					 fu_device_get_id (device),
```
superm1, line 729: is this logic actually right?

superm1:
To be honest I don't understand why to have `FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID` and still use `FWUPD_DEVICE_FLAG_NO_GUID_MATCHING` both.  

It seems like we should instead retire all stuff that has `FWUPD_DEVICE_FLAG_NO_GUID_MATCHING`.
Here's a proposal:
* By default nothing sets anything and that gets you the behavior that you get lots of matching
* If a device detaches or marks that it will replug you set `FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID`

Then each plugin doesn't need to hardcode it's behavior but instead you're specifically looking at the time you would match stuff which is replug only and you are more deterministic.  Am I wrong?

superm1:
> > it seems to make a lot of noise about devices I have
> 
> Excellent! Can you choose the appropriate flags and add to the branch please? 🙂

I tried at least from my understanding.

superm1:
> To be honest I don't understand why to have `FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID` and still use `FWUPD_DEVICE_FLAG_NO_GUID_MATCHING` both.
> 
> It seems like we should instead retire all stuff that has `FWUPD_DEVICE_FLAG_NO_GUID_MATCHING`.
> Here's a proposal:
> 
> * By default nothing sets anything and that gets you the behavior that you get lots of matching
> * If a device detaches or marks that it will replug you set `FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID`
> 
> Then each plugin doesn't need to hardcode it's behavior but instead you're specifically looking at the time you would match stuff which is replug only and you are more deterministic. Am I wrong?

And just to illustrate the problem with needing to set it for each plugin none of the source files in these plugins seem to set anything (yes some of them don't have device are just security attrs, but several are real deviceful plugins)
```
ccgx      
dfu-csr  
ep963x     
goodix-moc 
linux-lockdown  
logind          
msr        
pci-bcr
redfish   
steelseries
synaptics-prometheus
thelio-io
upower
acpi-dmar 
ebitdo  
fastboot   
linux-sleep     
pci-mei    
rts54hid
superio            
synaptics-rmi         
acpi-facp 
dell-esrt  
elantp   
flashrom
iommu       
linux-swap
pixart-rf
synaptics-cxaudio  
uefi-pk
wacom-raw
altos     
bios     
cros-ec
emmc     
fresco-pd
linux-tainted  
platform-integrity
solokey
uefi-recovery
wacom-usb
```

src/fu-device-list.c
hughsie: Yup, it's doing the checks there based on GUIDs.

hughsie:
> And just to illustrate the problem with needing to set it for each plugin none of the source files in these plugins

I think that's the problem. Those plugins obviously "work" but that's because we're matching everything everywhere. For instance, plug in two docks each with a cxaudio device and we'll only show one. I sorta think we should just rip the bandaid off, remove plugin references to NO_GUID_MATCHING and just review each plugin if it actually needs _REPLUG_MATCH_GUID. As you saw from dell-dock it's not always obvious and I think we're hiding bugs by just blindly matching everwhere for all devices. I think auto-setting REPLUG_MATCH_GUID on detach is also much **much** safer than assuming just to match all the time. I'll spin up a new PR that plays with that idea.

hughsie:
Alternate proposal which I like a lot better: https://github.com/fwupd/fwupd/pull/2936

FWIW, Google also requested more info in the README about what actually happens, so I guess 2 birds, one stone.

PR #2921, opened 2021-02-22
uefi-capsule: Ensure SBAT metadata is added correctly

hughsie:
The current approach of adding SBAT metadata after linking is creating
an image that is badly formed in 2 ways:

 * The SBAT section's file offset and size are not a multiple of the
   file alignment.

 * The SBAT section has a virtual address of zero. EDK2 loads the header
   here, and so it gets rejected.

This changes the approach to match shim, where an object file is
created with a .sbat section and then the linker takes care of placing
the section at a more appropriate virtual address.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -229,13 +229,11 @@ or server machines.
 %if 0%{?have_uefi}
     -Dplugin_uefi_capsule=true \
     -Dplugin_uefi_pk=true \
-%ifarch x86_64
     -Defi_sbat_distro_id="fedora" \
     -Defi_sbat_distro_summary="The Fedora Project" \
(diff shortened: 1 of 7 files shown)


martinezjavier (approved):

superm1 (approved):
Can we please include in a code comment and commit message a link to the gnu-efi submission to add sbat region to their linker scripts?  That should be useful for something to watch when it's time to bump the gnu-efi dependency and drop the linker scripts?

hughsie:
> include in a code comment and commit message a link to the gnu-efi submission

Done. https://github.com/vathpela/gnu-efi/pull/14 if you're following along! :)

PR #2924, opened 2021-02-22
Drop unused heap pages after startup has completed

hughsie:
Although glibc calls this when memory gets low automatically, we can invoke it
manually when we've loaded all plugins and coldplugged all devices.

This has no effect when using proper heap benchmarks like massif, but random
Reddit users just looking at the RSS column in an 'easy-to-use memory profiler'
will see a much lower numbers.

Where this feature might actually be useful is after a libxmlb metadata rebuild
where we dirtied a lot of heap just parsing XML.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -14,6 +14,7 @@
 #include <glib/gi18n.h>
 #include <glib-unix.h>
 #include <locale.h>
+#include <malloc.h>
 #ifdef HAVE_POLKIT
 #include <polkit/polkit.h>
 #endif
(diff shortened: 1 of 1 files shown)


hughsie:
Before:

    root       37103  1.3  0.1 587572 52668 pts/0    Sl+  20:49   0:00 ./src/fwupd --verbose

After:

    root       37676  9.2  0.0 563444 28400 pts/0    Sl+  20:54   0:00 ./src/fwupd --verbose

superm1 (approved):

PR #2932, opened 2021-02-23
dfu: Modernize the coding style to match the other plugings

hughsie:
Use a Fu prefix for source objects and files and use @self.

No logic changes.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
it's obviously a lot more work, but I also wonder if something should be done to break the static library -> csr thing.  IE should some of dfu be in libfwupdplugin?

superm1 (approved):

hughsie:
> should some of dfu be in libfwupdplugin

That's a good question. From memory the CSR plugin is only using the DFU state and status enums, and the corresponding to_string() methods. My gut instinct is we wait until a 3rd thing requires them and then split them out then.

hughsie:
Of course, then I wonder if the DFU_STATE_ enums should be renamed to FU_DFU_STATE_ enums as part of this patch to avoid a mass rename _twice_, so to speak.

hughsie:
> I wonder if the DFU_STATE_ enums should be renamed to FU_DFU_STATE_

I did this.

PR #2934, opened 2021-02-24
Add a backend ID to the device

hughsie:
This is typically a Linux sysfs path or USB platform ID and is used in a
different way to the physical ID. The physical ID is only set for some devices
after setup() and depends on the subsystem list, and this would not be defined
for devices that do not match a plugin.

This also fixes an regression where the FuDeviceList fails to match the new
FuUdevDevice device in fu_device_list_get_by_guids_removed() and instead
silently gets 'fixed up' only if FWUPD_DEVICE_FLAG_NO_GUID_MATCHING is not set.

This also allows us to move the various backends device caches to FuBackend as
we now have a suitable ID that is for just the backend to use.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -39,6 +39,7 @@ typedef struct {
 	gchar				*equivalent_id;
 	gchar				*physical_id;
 	gchar				*logical_id;
+	gchar				*backend_id;
 	gchar				*proxy_guid;
 	FuDevice			*alternate;
 	FuDevice			*proxy;		/* noref */
@@ -76,6 +77,7 @@ enum {
 	PROP_PROGRESS,
 	PROP_PHYSICAL_ID,
 	PROP_LOGICAL_ID,
+	PROP_BACKEND_ID,
 	PROP_QUIRKS,
 	PROP_PROXY,
 	PROP_LAST
@@ -99,6 +101,9 @@ fu_device_get_property (GObject *object, guint prop_id,
 		break;
 	case PROP_LOGICAL_ID:
 		g_value_set_string (value, priv->logical_id);
+		break;
+	case PROP_BACKEND_ID:
+		g_value_set_string (value, priv->backend_id);
 		break;
 	case PROP_QUIRKS:
 		g_value_set_object (value, priv->quirks);
@@ -126,6 +131,9 @@ fu_device_set_property (GObject *object, guint prop_id,
 		break;
 	case PROP_LOGICAL_ID:
 		fu_device_set_logical_id (self, g_value_get_string (value));
+		break;
+	case PROP_BACKEND_ID:
+		fu_device_set_backend_id (self, g_value_get_string (value));
 		break;
 	case PROP_QUIRKS:
 		fu_device_set_quirks (self, g_value_get_object (value));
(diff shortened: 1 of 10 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -2049,6 +2057,55 @@ fu_device_set_logical_id (FuDevice *self, const gchar *logical_id)
...
+	g_return_if_fail (FU_IS_DEVICE (self));
+
+	/* not changed */
+	if (g_strcmp0 (priv->backend_id, backend_id) == 0)
+		return;
+
+	g_free (priv->backend_id);
+	priv->backend_id = g_strdup (backend_id);
+	priv->device_id_valid = FALSE;
+	g_object_notify (G_OBJECT (self), "logical-id");
```
superm1, line 2106: `backend-id` no?

superm1 (commented):
Since backends are compiled into the daemon and are a fixed quantity, why is this a string?  Shouldn't it be an enum?

hughsie:
> Since backends are compiled into the daemon and are a fixed quantity, why is this a string

Ohh, it's an ID for the _device_, not for the backend.

libfwupdplugin/fu-device.c
hughsie: Yes! Good catch.

libfwupdplugin/fu-device.c
```diff
@@ -2049,6 +2057,55 @@ fu_device_set_logical_id (FuDevice *self, const gchar *logical_id)
...
+ * Since: 1.5.8
+ **/
+const gchar *
+fu_device_get_backend_id (FuDevice *self)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (FU_IS_DEVICE (self), NULL);
+	if (priv->backend_id != NULL)
+		return priv->backend_id;
+	return priv->physical_id;
```
hardboprobot, line 2079: Can a device not be managed by a backend at all?

superm1:
> Ohh, it's an ID for the device, not for the backend.

Oh I see... OK.

src/fu-bluez-backend.c
```diff
@@ -58,11 +56,11 @@ fu_bluez_backend_object_properties_changed (FuBluezBackend *self, GDBusProxy *pr
 
 	/* create device */
 	dev = g_object_new (FU_TYPE_BLUEZ_DEVICE,
+			    "device-id", path,
```
hardboprobot, line 59: "backend-id"?

src/fu-udev-backend.c
```diff
@@ -29,9 +28,6 @@ fu_udev_backend_device_add (FuUdevBackend *self, GUdevDevice *udev_device)
 	g_autoptr(FuUdevDevice) device = fu_udev_device_new (udev_device);
 
 	/* success */
-	g_hash_table_insert (self->devices,
-			     g_strdup (g_udev_device_get_sysfs_path (udev_device)),
-			     g_object_ref (device));
 	fu_backend_device_added (FU_BACKEND (self), FU_DEVICE (device));
```
hardboprobot, line 31: Shouldn't this set the device backend-id before calling `fu_backend_device_added`?

src/fu-usb-backend.c
```diff
@@ -26,13 +25,9 @@ fu_usb_backend_device_added_cb (GUsbContext *ctx,
...
 			       FuBackend *backend)
 {
-	FuUsbBackend *self = FU_USB_BACKEND (backend);
 	g_autoptr(FuUsbDevice) device = fu_usb_device_new (usb_device);
 
 	/* success */
-	g_hash_table_insert (self->devices,
-			     g_strdup (g_usb_device_get_platform_id (usb_device)),
-			     g_object_ref (device));
 	fu_backend_device_added (backend, FU_DEVICE (device));
```
hardboprobot, line 31: Same as for the udev backend, doesn't backend-id need to be set before calling `fu_backend_device_added`?

src/fu-bluez-backend.c
hughsie: Doh, thanks, fixed.

src/fu-udev-backend.c
hughsie: It's set in fu_udev_device_new() -> fu_udev_device_set_dev() -> fu_device_set_backend_id()

src/fu-usb-backend.c
hughsie: The *physical* ID is set by fu_usb_device_new() -> fu_usb_device_set_dev() which we fall back to in fu_device_get_backend_id()

libfwupdplugin/fu-device.c
hughsie: A FuDevice doesn't have to be created by a backend; in fact most are created inside the plugins -- although they do `->incorporate()` the backend_id I suppose. The fallback logic here was really just trying to avoid the common case of the physical ID also being the backend ID, which would save a few g_strdup's at startup. Maybe it's not worth it, if it's confusing things.

src/fu-udev-backend.c
hardboprobot: Right, thanks!

libfwupdplugin/fu-device.c
hardboprobot: Well, what I understood from this code is that a device may or may not have a backend_id but it will have a physical_id. I was confused because a few days ago I learnt that device discovery starts in a backend (at least the base device type that is later incorporated into a specific plugin device type).
I don't really care too much about this as long as it works, but yeah, someone that reads this for the first time might reach the same conclusion that I reached.

hardboprobot (approved):

Issue #2915, opened 2021-02-21
Fails to recognize a second Logitech Unifying Receiver


thx1111:
Arch Linux
fwupd 1.5.5-1

$ fwupdmgr --version
client version: 1.5.5
compile-time dependency versions
        gusb:   0.3.5

Unknown                  [***************************************]
daemon version: 1.5.5

$ solaar show       
Unifying Receiver
  Device path  : /dev/hidraw0
  USB id       : 046d:C52B
  Serial       : xxxxxxxx
    Firmware   : 12.10.B0032
    Bootloader : 02.14
    Other      : AA.AA
  Has 5 paired device(s) out of a maximum of 6.
  Notifications: wireless (0x000100)
  Device activity counters: 3=70, 5=192
...
Unifying Receiver
  Device path  : /dev/hidraw7
  USB id       : 046d:C52B
  Serial       : yyyyyyyy
    Firmware   : 12.01.B0019
    Bootloader : 02.14
  Has 0 paired device(s) out of a maximum of 6.
  Notifications: wireless (0x000100)
  Device activity counters: (empty)

$ fwupdmgr get-devices --show-all-devices
...
[ only the first Unifying Receiver is shown ]

$ lsusb
Bus 007 Device 004: ID 046d:c52b Logitech, Inc. Unifying Receiver
Bus 003 Device 006: ID 046d:c52b Logitech, Inc. Unifying Receiver

Of course, either device can be unplugged to work with the remaining device, so maybe this is more of a feature request.  In contrast, for instance, though, multiple hard drives are detected without problem.

Why is more than one Unifying Receiver plugged, you ask?  I simply wanted to update and configure the second receiver before moving it to another machine.  Instead, I had to remove the configured device that connected the keyboard and mouse I was using, and dig-out a wired keyboard and mouse to use while the second receiver was getting the update.  It's just inconvenient.

hughsie:
Hmm, this worked for me the last time I checked. Can you attach the output of `sudo fwupdtool get-devices --plugins logitech_hidpp --verbose` when both are installed please.

thx1111:
Thanks.  Hmm - I expect the part of interest may be:
```
...
15:04:18:0608 FuDevice             using b9d17735cbfcab51db9a9872394bd3210a4148c7 for DEVPATH=/devices/pci0000:00/0000:00:1d.1/usb7/7-2/7-2:1.2                             
[...]

PR #2936, opened 2021-02-24
Add two new flags to opt-in to GUID matching

hughsie:
Add two new flags to control if GUID matching is enabled for either coldplug or
replug. It is far too easy to forget to set FWUPD_DEVICE_FLAG_NO_GUID_MATCHING
for new plugins, and without it it all works really well *until* you have two
devices of the same type installed at the same time and then one 'disappears'
for hard to explain reasons. Typically we only need it for replug anyway!

Add two private device flags to explicitly opt-in to this rarely-required
behaviour, with the default to just use the physical and logical IDs. Also
document the update behavior for each plugin to explain why the flags is used.

This allows you to have two identical Unifying plugged in without one of them
being hidden from the user, at the same time allowing a HIDRAW<->USB transition
when going to and from bootloader and runtime modes.

This removes the workaround added in 99eb3f06b651c34a01acef20d1fde96e4996fffe.

Fixes https://github.com/fwupd/fwupd/issues/2915

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -176,6 +176,8 @@ fu_device_internal_flag_to_string (FuDeviceInternalFlags flag)
 		return "ensure-semver";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_RETRY_OPEN)
 		return "retry-open";
+	if (flag == FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID)
+		return "replug-match-guid";
 	return NULL;
 }
 
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -2591,7 +2591,7 @@ fu_device_add_string (FuDevice *self, guint idt, GString *str)
 		}
 		if (tmp2->len > 0)
 			g_string_truncate (tmp2, tmp2->len - 1);
-		fu_common_string_append_kv (str, idt + 1, "PrivateFlags", tmp2->str);
+		fu_common_string_append_kv (str, idt + 1, "InternalFlags", tmp2->str);
 	}
 
 	/* subclassed */
--- a/plugins/dell-dock/README.md
+++ b/plugins/dell-dock/README.md
@@ -51,8 +51,7 @@ These devices use several different generation schemes, e.g.
 Update Behavior
 ---------------
 
-For devices implementing "passive flow" the update is deployed at runtime and
-then actually written the next time the dock is attached.
+All devices will be updated the next time the usb-c plug from the dock is unplugged from the host.
 
 Vendor ID Security
 ------------------
--- a/plugins/dell-dock/README.md
+++ b/plugins/dell-dock/README.md
@@ -48,6 +48,12 @@ These devices use several different generation schemes, e.g.
  * MST Hub: `MST-panamera-vmm5331-259`
  * Thunderbolt Controller: `TBT-00d4b070`
 
+Update Behavior
+---------------
+
+For devices implementing "passive flow" the update is deployed at runtime and
+then actually written the next time the dock is attached.
+
 Vendor ID Security
 ------------------
 
--- a/plugins/thunderbolt/README.md
+++ b/plugins/thunderbolt/README.md
@@ -44,6 +44,21 @@ The retimer index is oriented around the physical connection within
 the machine.  It is important as multiple controllers may otherwise
 identify identically.
 
+Update Behavior
+---------------
+
+For most devices the firmware is written to the device at runtime and the
+update is applied immediately. Once complete the controller may reboot
+which may cause all connected USB and TBT devices to be reenumerated.
+
+For some devices and circumstances (such as the Dell WD19 with a new enough
+kernel) the device will remain functional for the duration of the update.
+The update will complete on unplug.
+
+If a user sets `DelayedActivation` configuration option then the update will
+be staged but not completed until `activate` is separately called such as at
+logout or shutdown.
+
 Vendor ID Security
 ------------------
 
--- a/contrib/ci/fedora.sh
+++ b/contrib/ci/fedora.sh
@@ -12,6 +12,7 @@ deprecated="FWUPD_DEVICE_FLAG_NO_AUTO_INSTANCE_IDS
             FWUPD_DEVICE_FLAG_ONLY_SUPPORTED
             FWUPD_DEVICE_FLAG_MD_SET_NAME
             FWUPD_DEVICE_FLAG_MD_SET_VERFMT
+            FWUPD_DEVICE_FLAG_NO_GUID_MATCHING
             FWUPD_DEVICE_FLAG_MD_SET_ICON"
 for val in $deprecated; do
     if grep -- $val plugins/*/*.c ; then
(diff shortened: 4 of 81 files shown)


hughsie:
If you're testing this you also need the commit from https://github.com/fwupd/fwupd/pull/2934

libfwupdplugin/fu-device.c
```diff
@@ -2906,6 +2916,16 @@ fu_device_detach (FuDevice *self, GError **error)
...
 		return TRUE;
 
+	/* fixup old plugins */
+	if (priv->has_counterpart &&
+	    !fu_device_has_internal_flag (self, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID)) {
+		g_warning ("device %s [%s] did not set FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID "
+			   "so setting it for backwards compat -- this will stop soon!",
+			   fu_device_get_name (self),
+			   fu_device_get_id (self));
+		fu_device_add_internal_flag (self, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
```
superm1, line 2926: the number of plugins that use counterpart is a fixed number, why not just fix those immediately and set an error on detach?

hughsie: I think I fixed them all. There are actually plugins using counterpart for other reasons, e.g. the vli plugin uses it to find the USB-2 device that corresponds to the USB-3 device. Maybe dropping this chunk makes sense.

plugins/dell-dock/README.md
```diff
@@ -48,6 +48,12 @@ These devices use several different generation schemes, e.g.
  * MST Hub: `MST-panamera-vmm5331-259`
  * Thunderbolt Controller: `TBT-00d4b070`
 
+Update Behavior
+---------------
+
+For devices implementing "passive flow" the update is deployed at runtime and
+then actually written the next time the dock is attached.
```
superm1, line 55: Come to think of it, NOT using passive was only pre-production devices.  I wonder if it's worth just ripping out the code for non-passive update now.

superm1:
> This removes the workaround added in 99eb3f0.

If doing that - it sounds to me like the devices it helped need `FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID` set don't they?

superm1:
> If doing that - it sounds to me like the devices it helped need FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID set don't they?

The case that needed it was `uefi_capsule` + `nvme` both using the same GUID.

plugins/dell-dock/README.md
hughsie: I think ripping out code for immediate update (on the assumption they're not what was ever sold in shops) is the best plan. Do you mind doing that and updating the README please? You can probably describe the delayed write thing better than me.

hughsie:
> The case that needed it was uefi_capsule + nvme both using the same GUID.

So, if we actually do this explicit opt-in case I don't actually think anything needs the FU_DEVICE_INTERNAL_FLAG_COLDPLUG_MATCH_GUID behaviour. Certainly not NVMe or UEFI-capsule anymore, and I think the COLDPLUG_MATCH_GUID was really a _load-bearing-bug_ until I worked out the problem in https://github.com/fwupd/fwupd/commit/253e6e4f17fd70661a1f5a7e0cb48503d11cb12e#diff-ceb6643e2c2b41f6ff57f2f230d996792efff6e0da680581d6d445a38f445dceR6000

plugins/dell-dock/README.md
superm1: The main thing needed for passive flow was a new enough EC and a new enough hub.  The earliest hub in public release was 01.47 (https://fwupd.org/lvfs/devices/com.dell.wd198ceeeffd.firmware) and minimum EC was 00.00.00.23 (torn out here https://github.com/fwupd/fwupd/commit/ead0deaa255f08db0d777deb97f650d10d545eb5#diff-2d320a289c64c6066499373624e1d9b944d4e13b0eccd3068dad84fda2afd804).

I'll finish tearing out the rest then.

plugins/thunderbolt/README.md
```diff
@@ -44,6 +44,12 @@ The retimer index is oriented around the physical connection within
 the machine.  It is important as multiple controllers may otherwise
 identify identically.
 
+Update Behavior
+---------------
+
+The firmware is written to the device at runtime. Once complete the controller
+may reboot which may cause all connected USB devices to be reenumerated.
```
superm1, line 51: This probably needs some elaboration since flashing WD19 and config option change it.

superm1 (approved):
just that documentation note for thunderbolt and then LGTM

plugins/thunderbolt/README.md
hughsie: Got anything more please? You understand this much better than me.

hughsie:
> and then LGTM

I'll also spend the better part of tmorrow running the device tests for all the hardware in my office...

superm1:
Just need to fixup the commit message for e4d9c29bef45273fdc92d9a9402ef9d8cf7211f4 to be one new GUID

hughsie:
> Just

I'm about 2/3rds through testing all the hardware in my lab. Will fix up before pushing, thanks.

plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
```diff
@@ -611,6 +611,7 @@ fu_logitech_hidpp_peripheral_setup (FuDevice *device, GError **error)
 	if (idx != 0x00) {
 		self->is_updatable = TRUE;
 		fu_device_add_flag (FU_DEVICE (device), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+		fu_device_set_protocol (FU_DEVICE (device), "com.logitech.unifyingsigned");
```
superm1, line 614: this seems like an incorrect change - isn't the protocol signed only when it has idx `HIDPP_FEATURE_DFU_CONTROL_SIGNED` set?

superm1: and furthermore the protocol was actually originally set in `fu_logitech_hidpp_peripheral_init`

hughsie: Yes, but you only get `HIDPP_FEATURE_DFU_CONTROL` or `HIDPP_FEATURE_DFU_CONTROL_SIGNED` when you're in runtime mode. If you're stuck in recovery mode (e.g. you unplugged during update) you only get HIDPP_FEATURE_DFU which doesn't tell you what protocol we should be verifying. I guess the real fix would be to change `fu_device_set_protocol()` into `fu_device_add_protocol()` but that's a lot of work for something that I don't think is very common.

hughsie: I'll split this commit out as a new PR as it's more controversial it seems.

thx1111:
Thanks Richard and Mario.  I tested this on Arch fwupd-git 1.5.7.r13.g0b6c67a3-1.  It recognized both receivers properly and also performed an upgrade properly on the correct second receiver.

hughsie:
Woo, thanks for confirming! It was a pretty solid few days work, and I think I've tested every hit of hardware in my office validating the invasive change didn't break anything :)

PR #2942, opened 2021-02-25
Wip/hughsie/logitech fixes

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
@@ -693,6 +693,7 @@ fu_logitech_hidpp_peripheral_detach (FuDevice *device, GError **error)
 			g_prefix_error (error, "failed to put device into DFU mode: ");
 			return FALSE;
 		}
+		g_usleep (200 * 1000);
 		return fu_logitech_hidpp_peripheral_setup (FU_DEVICE (self), error);
 	}
 


plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
```diff
@@ -688,6 +689,7 @@ fu_logitech_hidpp_peripheral_detach (FuDevice *device, GError **error)
 			g_prefix_error (error, "failed to put device into DFU mode: ");
 			return FALSE;
 		}
+		g_usleep (200 * 1000);
```
FFY00, line 692: What is the context of this sleep?
It seems it is when putting the device in DFU, right? If so, seems reasonable.

Please note that there are 2 different hardware variants and one of them takes significantly longer to validate cryptographic signatures, so if we are waiting for the device to wake up after flashing, that needs to be taken into account. But that should already be implemented and I think this code is about putting it in DFU, so not relevant, just thought I would note this in case.

FFY00 (commented):
Do you have a bug report?

FFY00:
> You can't actually tell if the bootloader supports signed or unsigned updates,

I think you should? By looking at the firmware version.

> but Logitech started moving from unsigned to signed peripherals in 2016 and I
think it's now safe to assume the latter over 5 years later...

Eh... I wouldn't be too sure about that... I receive a significant amount of bug reports in libratbag for older devices. And it's not like 5 years is too big of a time frame to drop support, IMO.

hughsie:
> > You can't actually tell if the bootloader supports signed or unsigned updates,
> I think you should? By looking at the firmware version.

Is there a hard and fast rule? I know there's one for unifying receivers, but peripherals?

> > but Logitech started moving from unsigned to signed peripherals in 2016 and I
> > think it's now safe to assume the latter over 5 years later...
> Eh... I wouldn't be too sure about that... I receive a significant amount of bug reports in libratbag for older devices

Okay, that's a really useful datapoint, thanks.

FFY00:
> Is there a hard and fast rule? I know there's one for unifying receivers, but peripherals?

I think so, though you'd probably need to get the firmware versions from Logitech. Or perhaps ask them to push both firmware updates (signed and unsigned) with a regex, as I think they already do for receivers.

plugins/logitech-hidpp/fu-logitech-hidpp-peripheral.c
```diff
@@ -611,6 +611,7 @@ fu_logitech_hidpp_peripheral_setup (FuDevice *device, GError **error)
 	if (idx != 0x00) {
 		self->is_updatable = TRUE;
 		fu_device_add_flag (FU_DEVICE (device), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+		fu_device_set_protocol (FU_DEVICE (device), "com.logitech.unifyingsigned");
```
superm1, line 614: in `fu-logitech-hidpp-bootloader` there is a regex to tell which bootloaders have signed images or not, can some similar heuristic exist here?

hughsie: I'm wondering if we need to rip off the bandaid and add fu_device_add_protocol...

hughsie: > rip off the bandaid

Yeooooowch! https://github.com/fwupd/fwupd/pull/2957

PR #2944, opened 2021-02-25
Support for GATT characteristic signals/notifications

hardboprobot:
This adds three methods to bluez devices:

  - fu_bluez_device_connect_signal: to register a callback for a signal
    on a BlueZ object.
  - fu_bluez_device_start_notify: to enable notifications for property
    changes in a specific object.
  - fu_bluez_device_stop_notify: to disable notifications for property
    changes in a specific object.

NOTE that currently this is implemented for GATT characteristics only,
but can be extended to other types of objects.

Listening for property changes in a bluez characteristic requires a
long-lived connection or proxy, so this also refactors bluez device
UUIDs into a simple object that keeps the path, the proxy and the signal
id in case a callback is registered for this UUID.

An additional advantage of this is that fwupd no longer creates a
throwaway proxy object for every read and write UUID operation.

API
===

Subclasses of FuBluezDevice don't need to handle the FuBluezUuid
directly and can reference a UUID using the UUID string only:

    #define READ_WRITE_CHAR "0000ff01-0000-1000-8000-00805f9b34fb"

    fu_bluez_device_connect_signal (bluez_dev,
                                    READ_WRITE_CHAR,
                                    "g-properties-changed",
                                    fu_pxi_ble_device_readwrite_notify_cb,
                                    error);

    fu_bluez_device_start_notify (bluez_dev,
                                  READ_WRITE_CHAR,
                                  error);

Limitations
===========

  - Bluez device UUIDs are still for characteristics only.
  - Only one signal/callback per UUID.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-bluez-device.c
+++ b/libfwupdplugin/fu-bluez-device.c
@@ -30,8 +30,16 @@ typedef struct {
 	FuBluezDevice		 parent_instance;
 	GDBusObjectManager	*object_manager;
 	GDBusProxy		*proxy;
-	GHashTable		*uuid_paths;	/* utf8 : utf8 */
+	GHashTable		*uuids;		/* utf8 : FuBluezDeviceUuidHelper */
 } FuBluezDevicePrivate;
+
+typedef struct {
+	FuBluezDevice		*self;
+	gchar			*uuid;
+	gchar			*path;
+	gulong			 signal_id;
+	GDBusProxy		*proxy;
+} FuBluezDeviceUuidHelper;
 
 enum {
 	PROP_0,
@@ -40,20 +48,115 @@ enum {
 	PROP_LAST
 };
 
+enum {
+	SIGNAL_CHANGED,
+	SIGNAL_LAST
+};
+
+static guint signals[SIGNAL_LAST] = { 0 };
+
 G_DEFINE_TYPE_WITH_PRIVATE (FuBluezDevice, fu_bluez_device, FU_TYPE_DEVICE)
 
 #define GET_PRIVATE(o) (fu_bluez_device_get_instance_private (o))
 
 static void
+fu_bluez_uuid_free (FuBluezDeviceUuidHelper *uuid_helper)
+{
+	if (uuid_helper->path != NULL)
+		g_free (uuid_helper->path);
+	if (uuid_helper->proxy != NULL)
+		g_object_unref (uuid_helper->proxy);
+	g_free (uuid_helper->uuid);
+	g_object_unref (uuid_helper->self);
+	g_free (uuid_helper);
+}
+
+/*
+ * Looks up a UUID in the FuBluezDevice uuids table.
+ */
+static FuBluezDeviceUuidHelper *
+fu_bluez_device_get_uuid_helper (FuBluezDevice *self,
+				 const gchar *uuid,
+				 GError **error)
+{
+	FuBluezDevicePrivate *priv = GET_PRIVATE (self);
+	FuBluezDeviceUuidHelper *uuid_helper;
+
+	uuid_helper = g_hash_table_lookup (priv->uuids, uuid);
+	if (uuid_helper == NULL) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_NOT_SUPPORTED,
+			     "UUID %s not supported", uuid);
+		return NULL;
+	}
+
+	return uuid_helper;
+}
+
+static void
+fu_bluez_device_signal_cb (GDBusProxy *proxy,
+			   GVariant *changed_properties,
+			   GStrv invalidated_properties,
+			   FuBluezDeviceUuidHelper *uuid_helper)
+{
+	g_signal_emit (uuid_helper->self, signals[SIGNAL_CHANGED], 0, uuid_helper->uuid);
+}
+
+/*
+ * Builds the GDBusProxy of the BlueZ object identified by a UUID
+ * string. If the object doesn't have a dedicated proxy yet, this
+ * creates it and saves it in the FuBluezDeviceUuidHelper object.
+ *
+ * NOTE: Currently limited to GATT characteristics.
+ */
+static gboolean
+fu_bluez_device_ensure_uuid_helper_proxy (FuBluezDeviceUuidHelper *uuid_helper,
+					  GError **error)
+{
+	if (uuid_helper->proxy != NULL)
+		return TRUE;
+	uuid_helper->proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
+						     G_DBUS_PROXY_FLAGS_NONE,
+						     NULL,
+						     "org.bluez",
+						     uuid_helper->path,
+						     "org.bluez.GattCharacteristic1",
+						     NULL, error);
+	if (uuid_helper->proxy == NULL) {
+		g_prefix_error (error, "Failed to create GDBusProxy for uuid_helper: ");
+		return FALSE;
+	}
+	g_dbus_proxy_set_default_timeout (uuid_helper->proxy, DEFAULT_PROXY_TIMEOUT);
+	uuid_helper->signal_id = g_signal_connect (uuid_helper->proxy,
+						   "g-properties-changed",
+						   G_CALLBACK (fu_bluez_device_signal_cb),
+						   uuid_helper);
+	if (uuid_helper->signal_id <= 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_NOT_SUPPORTED,
+			     "cannot connect to signal of UUID %s",
+			     uuid_helper->uuid);
+		return FALSE;
+	}
+	return TRUE;
+}
+
+static void
 fu_bluez_device_add_uuid_path (FuBluezDevice *self, const gchar *uuid, const gchar *path)
 {
 	FuBluezDevicePrivate *priv = GET_PRIVATE (self);
+	FuBluezDeviceUuidHelper *uuid_helper;
 	g_return_if_fail (FU_IS_BLUEZ_DEVICE (self));
 	g_return_if_fail (uuid != NULL);
 	g_return_if_fail (path != NULL);
-	g_hash_table_insert (priv->uuid_paths,
-			     g_strdup (uuid),
-			     g_strdup (path));
+
+	uuid_helper = g_new0 (FuBluezDeviceUuidHelper, 1);
+	uuid_helper->self = g_object_ref (self);
+	uuid_helper->uuid = g_strdup (uuid);
+	uuid_helper->path = g_strdup (path);
+	g_hash_table_insert (priv->uuids, g_strdup (uuid), uuid_helper);
 }
 
 static void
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-bluez-device.c
```diff
@@ -45,15 +51,87 @@ G_DEFINE_TYPE_WITH_PRIVATE (FuBluezDevice, fu_bluez_device, FU_TYPE_DEVICE)
...
 	g_return_if_fail (path != NULL);
-	g_hash_table_insert (priv->uuid_paths,
-			     g_strdup (uuid),
-			     g_strdup (path));
+
+	uuid = g_new0 (FuBluezUuid, 1);
+	uuid->path = g_strdup (path);
+	g_hash_table_insert (priv->uuids,
+			     g_strdup (uuid_str),
+			     uuid);
```
hughsie, line 134: nitpick: I think this can all be one line

libfwupdplugin/fu-bluez-device.c
```diff
@@ -107,14 +185,14 @@ fu_bluez_device_to_string (FuDevice *device, guint idt, GString *str)
 	FuBluezDevice *self = FU_BLUEZ_DEVICE (device);
 	FuBluezDevicePrivate *priv = GET_PRIVATE (self);
 
-	if (priv->uuid_paths != NULL) {
+	if (priv->uuids != NULL) {
 		GHashTableIter iter;
 		gpointer key, value;
-		g_hash_table_iter_init (&iter, priv->uuid_paths);
+		g_hash_table_iter_init (&iter, priv->uuids);
 		while (g_hash_table_iter_next (&iter, &key, &value)) {
```
hughsie, line 192: I think adding `const FuBluezUuid *uuid = (const FuBluezUuid *) value;` here will be more useful for the future.

hughsie: maybe drop the consts too. Not sure.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -276,37 +354,18 @@ fu_bluez_device_probe (FuDevice *device, GError **error)
...
-					       "org.bluez",
-					       path,
-					       "org.bluez.GattCharacteristic1",
-					       NULL, error);
-	if (proxy == NULL) {
-		g_prefix_error (error, "Failed to connect GattCharacteristic1: ");
-		return NULL;
-	}
-	g_dbus_proxy_set_default_timeout (proxy, DEFAULT_PROXY_TIMEOUT);
+	proxy = fu_bluez_device_get_uuid_proxy (self, uuid_str, error);
```
hughsie, line 366: this is much better than recreating the proxy every time BTW, good job.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -340,7 +399,7 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
 /**
  * fu_bluez_device_read_string:
  * @self: A #FuBluezDevice
- * @uuid: The UUID, e.g. `00cde35c-7062-11eb-9439-0242ac130002`
+ * @uuid_str: The UUID, e.g. `00cde35c-7062-11eb-9439-0242ac130002`
```
hughsie, line 402: I kinda think the name should remain `uuid` -- and maybe call the Uuid object a `FuBluezDeviceHelper *helper` or `* uuid_obj` instead.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -350,18 +409,140 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
...
+ *
+ * Returns: %TRUE if the signal connection was successful.
+ *
+ * Since: 1.5.7
+ **/
+gboolean
+fu_bluez_device_connect_signal (FuBluezDevice *self,
+				const gchar *uuid_str,
+				const gchar *signal_name,
+				gpointer callback,
```
hughsie, line 440: I was going to say you also need `gpointer callback_userdata` here so you can pass a pointer to `@callback` but I'm wondering if just emitting a signal from the `FuBluezDevice` itself might be more useful to the client. e.g. `g_signal_emit (self, "signal", name)`

libfwupdplugin/fu-bluez-device.c
```diff
@@ -350,18 +409,140 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
...
+ *
+ * Since: 1.5.7
+ **/
+gboolean
+fu_bluez_device_start_notify (FuBluezDevice *self,
+			      const gchar *uuid_str,
+			      GError **error)
+{
+	GDBusProxy *proxy;
+	g_autoptr(GVariant) ret = NULL;
```
hughsie, line 486: maybe `retval`, `ret` is normally for booleans.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -350,18 +409,140 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
...
+			     G_IO_ERROR_NOT_SUPPORTED,
+			     "Can't connect to signal %s of UUID %s", signal_name, uuid_str);
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+/**
+ * fu_bluez_device_start_notify:
```
hughsie, line 469: do we need this? couldn't we call it automatically when calling `fu_bluez_device_connect_signal`?

libfwupdplugin/fu-bluez-device.c
```diff
@@ -350,18 +409,140 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
...
+ * @callback: The handler for this signal
+ * @error: A #GError, or %NULL
+ *
+ * Connects to a signal of a DBus object identified by a UUID and
+ * registers a callback to handle it.
+ * If successful, it saves the signal id in the uuid object.
+ *
+ * Returns: %TRUE if the signal connection was successful.
+ *
+ * Since: 1.5.7
```
superm1, line 434: 1.5.8

libfwupdplugin/fu-bluez-device.c
superm1: I agree, it's one less thing that every plugin needs to do too then.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -350,18 +409,140 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
...
+ * fu_bluez_device_stop_notify:
+ * @uuid_str: The UUID, e.g. `00cde35c-7062-11eb-9439-0242ac130002`
+ * @error: A #GError, or %NULL
+ *
+ * Disables notifications for property changes in a UUID (StopNotify
+ * method).
+ *
+ * Returns: %TRUE if the method call completed successfully.
+ *
+ * Since: 1.5.7
```
superm1, line 515: 1.5.8

libfwupdplugin/fu-bluez-device.c
hardboprobot: I'm not sure I'm understanding that. You mean that instead of a client calling `fu_bluez_device_connect_signal` to register a callback directly, FuBluezDevice should collect all BlueZ signals and then emit signals for the clients to connect to?

libfwupdplugin/fu-bluez-device.c
hardboprobot: Yeah, we could. I just though it was more orthogonal to have the "register callback" and the "start/stop notifications" functionalities as separate things. Just thinking ahead, maybe someone will need to "silence" notifications selectively for a period of time and then enable them back again. Or maybe not, we'll see :)

libfwupdplugin/fu-bluez-device.c
hughsie: No, I think we'd just have `fu_bluez_device_connect_signal (FuBluezDevice *self, const gchar *uuid_str)` which would automatically call `fu_bluez_device_start_notify (self, uuid_str)` -- the signal would be sent out as a standard `g_signal_emit()` rather than by calling a callback.

libfwupdplugin/fu-bluez-device.c
hughsie: Yes, I agree we need _start() and _stop() -- I just think we should _start() by default.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -350,18 +409,140 @@ fu_bluez_device_read (FuBluezDevice *self, const gchar *uuid, GError **error)
...
 	if (buf == NULL)
 		return NULL;
 	return (gchar *) g_byte_array_free (buf, FALSE);
 }
 
+/**
+ * fu_bluez_device_connect_signal:
+ * @self: A #FuBluezDevice
+ * @uuid_str: The UUID, e.g. `00cde35c-7062-11eb-9439-0242ac130002`
+ * @signal_name: The signal to connect, e.g. `g-properties-changed`
```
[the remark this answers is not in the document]
hardboprobot: Mmmm... now that I look at it, GATT characteristics only implement the "PropertiesChanged" signal, so no. I'll change it.

libfwupdplugin/fu-bluez-device.c
hardboprobot: Sure, we can do that.

libfwupdplugin/fu-bluez-device.c
hardboprobot: Mmm... I don't know if we're thinking about the same use case, sorry I still don't see it.

What I had in mind was:

A device plugin (for the Pixart keyboard, for example) writes a characteristic and it has to wait for that characteristic to update its value before reading the ACK or result back.
The device setup calls `fu_bluez_device_connect_signal` to attach a callback to the PropertiesChanged signal of that characteristic and enable notifications.
When the value is updated, the callback function updates internal data or notifies some other waiting thread that the data is ready.

How can we do this without a device-specific callback?

By the way, we can talk about this in the irc channel if it's faster/easier

hughsie:
> I noticed you're not enabling the notifications

I flip flopped on that. I think it's better to not do it by default and be explicit. For instance, we might want the proxy for a write and not want signals.

hughsie:
> Shall I proceed to squash the commits and fix the conflict with master?

Sounds like a plan! :)

libfwupdplugin/fu-bluez-device.c
```diff
@@ -433,6 +505,80 @@ fu_bluez_device_write (FuBluezDevice *self,
...
+fu_bluez_device_notify_start (FuBluezDevice *self, const gchar *uuid, GError **error)
+{
+	FuBluezDeviceUuidHelper *uuid_helper;
+	g_autoptr(GVariant) retval = NULL;
+
+	uuid_helper = fu_bluez_device_get_uuid_helper (self, uuid, error);
+	if (uuid_helper == NULL)
+		return FALSE;
+	if (uuid_helper->signal_id > 0)
+		return TRUE;
```
hughsie, line 530: ohh i think we need to remove these two lines now -- the signal is unconditionally wired up at proxy creation, so these can't fail.

hughsie (approved):
Thanks!

PR #2959, opened 2021-03-01
thunderbolt: if doing disconnected update flow, lie about the version

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -2270,6 +2270,13 @@ fu_device_add_flag (FuDevice *self, FwupdDeviceFlags flag)
 	if (flag & FWUPD_DEVICE_FLAG_INSTALL_ALL_RELEASES)
 		flag |= FWUPD_DEVICE_FLAG_VERSION_CHECK_REQUIRED;
 	fwupd_device_add_flag (FWUPD_DEVICE (self), flag);
+
+	/* activatable devices shouldn't be allowed to update again until activated */
+	/* don't let devices be updated until activated */
+	if (flag & FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION) {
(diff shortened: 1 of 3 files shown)


superm1:
I guess with the CI failure I'll need to plumb a little more in, but the other approach is that the daemon just automatically lies about the version when something goes into `needs-activation`.  Can tear out some code in other `dell-dock` plugins when this happens.

hughsie:
I'm not convinced about faking the version. Can't we just skip the version checks if FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION is set? I like the setting of the FuFirmware version tho, feel free to push that bit.

superm1:
> Can't we just skip the version checks if `FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION` is set?

You mean a client side fix?  We'll still have gnome-software and every other client offering firmware upgrades won't we?

hughsie:
>  We'll still have gnome-software and every other client offering firmware upgrades won't we

Ohh, indeed. Can we take away `UPDATABLE` in that case?

superm1:
> Ohh, indeed. Can we take away UPDATABLE in that case?

Like in the daemon if `FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION` gets set we take away `FWUPD_DEVICE_FLAG_UPDATABLE` and then when activated we put it back?

hughsie:
Yes, something like that. I think we want to take away FWUPD_DEVICE_FLAG_UPDATABLE, and add FWUPD_DEVICE_FLAG_UPDATABLE_HIDDEN when FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION gets set. Maybe even in `fu_device_add_flag`....

hughsie (approved):

PR #2960, opened 2021-03-02
Allow enabling plugins only matching a specific HwId

hughsie:
At the moment plugins are doing this a few different ways; either looping
through the HwIds manually (e.g. flashrom) or setting a custom flag that is
checked in fu_plugin_setup (e.g. uefi-recovery).

Define a standard 'Plugin' HwId quirk to simplify plugins.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 I added a new flag as the `FWUPD_PLUGIN_FLAG_NO_HARDWARE` flag is more of a "_result_" than a "_behaviour_". If you have any better ideas I'm open for suggestions.

superm1:
This approach seems good to me, I'm happy with it.

PR #2961, opened 2021-03-02
Wip/hughsie/quirk prefixes

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I think https://github.com/fwupd/fwupd/pull/2960 is a prerequisite for this PR.

hughsie:
@superm1 okay, rebased against master again, so ready for review.

libfwupdplugin/fu-quirks.c
```diff
@@ -66,23 +68,28 @@ fu_quirks_build_group_key (const gchar *group)
 	for (guint i = 0; guid_prefixes[i] != NULL; i++) {
 		if (g_str_has_prefix (group, guid_prefixes[i])) {
 			gsize len = strlen (guid_prefixes[i]);
+			g_warning ("using %s in quirk files is deprecated!",
+				   guid_prefixes[i]);
```
superm1, line 72: maybe a good thing to add in the ABI bump TODO?  Or some other line in the sand (1.6.x?)

superm1 (approved):

libfwupdplugin/fu-quirks.c
hughsie: I flip flop on this; on one hand it's dead code, on the other hand the codepath here is not in the fast path (it's only run when the quirks get updated) and so it's not hurting performance. Removing it means that we turn an ugly red warning on the console into a silent failure (we'd convert `Guid=cc4cbfa9-bf9d-540b-b92b-172ce31013c1` into `a9eab53b-893e-5f5f-9b9e-bb7b9fd5b9bb` which would be an absolute **arse** to debug why the GUID isn't matching.

PR #2962, opened 2021-03-02
Wip/hughsie/pixart

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -1109,6 +1109,85 @@ fu_udev_device_set_physical_id (FuUdevDevice *self, const gchar *subsystems, GEr
 }
 
 /**
+ * fu_udev_device_set_logical_id:
+ * @self: A #FuUdevDevice
+ * @subsystem: A subsystem string, e.g. `pci,usb`
+ * @error: A #GError, or %NULL
+ *
+ * Sets the logical ID from the device subsystem. Plugins should choose the
+ * subsystem that most relevant in the udev tree, for instance choosing 'hid'
+ * over 'usb' for a mouse device.
+ *
+ * Returns: %TRUE if the logical device was set.
+ *
+ * Since: 1.5.8
+ **/
+gboolean
+fu_udev_device_set_logical_id (FuUdevDevice *self, const gchar *subsystem, GError **error)
+{
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	const gchar *tmp;
+	g_autofree gchar *logical_id = NULL;
+	g_autoptr(GUdevDevice) udev_device = NULL;
+
+	g_return_val_if_fail (FU_IS_UDEV_DEVICE (self), FALSE);
+	g_return_val_if_fail (subsystem != NULL, FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+	/* nothing to do */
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -1108,6 +1108,93 @@ fu_udev_device_set_physical_id (FuUdevDevice *self, const gchar *subsystems, GEr
...
 
+/**
+ * fu_udev_device_set_logical_id:
+ * @self: A #FuUdevDevice
+ * @subsystems: A subsystem string, e.g. `pci,usb`
+ * @error: A #GError, or %NULL
+ *
+ * Sets the logical ID from the device subsystem. Plugins should choose the
+ * subsystem that is "deepest" in the udev tree, for instance choosing 'usb'
+ * over 'pci' for a mouse device.
```
superm1, line 1119: your examples here aren't used, maybe make it choosing `hid` over `usb`

superm1 (approved):

hughsie:
Lets see if PixArt gets back to us about why the HID report is sometimes prefixed. I don't like papering over bugs.

superm1:
>Lets see if PixArt gets back to us about why the HID report is sometimes prefixed. I don't like papering over bugs.

Yeah that does look weird.

jyu99:
Hi 

I found the some information about HIDGETFEATURE
[https://www.kernel.org/doc/Documentation/hid/hidraw.txt](url)

and here

[https://stackoverflow.com/questions/62333326/zero-byte-disappear-when-get-hid-feature-report-using-ioctl-in-linux](url)

 It's seem like skip the first byte, I guess default  first byte is report number.

But in your case. not only skip the first byte, then system will keep report number automatically

hughsie:
> This commit: bluez/bluez@35a2c50 present in BlueZ

Good detective work!

> I guess we should include some code to detect the BlueZ version.

I agree. It needs to be a runtime daemon-version test rather than a built-time test as the user is free to update fwupd and bluez independently. I suspect the best place might be in fu-bluez-backend.c and then set a "backend-version" string property and fu_bluez_device_get_backend_version() in FuBluezDevice. Actually, it might make sense in FuDevice (the parent class) too -- we could do the same for the udev version and gusb version -- the latter actually being useful in bugreports. Could you do that in a new PR please?

hughsie:
> FuPxiDevice is not a FuBluezDevice

Urgh, I forgot that bit. Can we do both? i.e. do the bluez-backend thing (which I think we need anyway for this kind of thing) and then we just copy and paste the code into FuPxiDevice. Or we do the bluez-backend thing and then do something like my hack i suggested -- but instead of shuffling the bytes along for new versions, prepending the bytes for old versions....

hughsie:
I think we'll need to support the older bluez as it's in the LTS releases of various things (including RHEL IIRC)

superm1:
>I think we'll need to support the older bluez as it's in the LTS releases of various things (including RHEL IIRC)

But also there is the possibility to have a runtime check on the bluez version and disable plugins if it's too old.  We do this in other plugins (modem manager comes to mind).

libfwupdplugin/fu-udev-device.c
hughsie: Fixed. I've also done `s\subsystems\subsystem` as I don't think the fallback makes sense for the logical ID. You know what you're looking for, unlike the physical ID.

PR #2964, opened 2021-03-02
elantp: Only call ->detach() when writing a firmware blob

hughsie:
The detach is actually destructive (!) and cannot be reattached unless actually
writing a new firmware image.

Type of pull request:
- [ X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@jinglewu

superm1:
Can we also validate the signature on the image perhaps in the write routine?  It seems especially important now..

hughsie:
Isn't that what we do in prepare?

superm1:
> Isn't that what we do in prepare?

I only saw some verification it applies to the HW, nothing for signature checking.  If you don't do signature checking, this can be a DOS to provide a binary that is generated to match the hardware but not be real..

hughsie:
@jinglewu is there some signature or checksum in the file? Thanks.

jinglewu:
@hughsie Yes, there have a signature, but not have checksum.
https://github.com/torvalds/linux/blob/master/drivers/input/mouse/elan_i2c_core.c#L615
https://github.com/torvalds/linux/blob/master/drivers/input/mouse/elan_i2c_core.c#L139

hughsie:
> Yes, there have a signature

Okay, that's useful, thanks. I've added that as a follow up commit.

superm1:
> @hughsie Yes, there have a signature, but not have checksum.
> https://github.com/torvalds/linux/blob/master/drivers/input/mouse/elan_i2c_core.c#L615
> https://github.com/torvalds/linux/blob/master/drivers/input/mouse/elan_i2c_core.c#L139

Maybe a silly question - but if there is a sysfs API for this driver, why aren't we using that in the first place to flash?

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -622,6 +628,7 @@ fu_elantp_i2c_device_init (FuElantpI2cDevice *self)
 {
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_INTERNAL);
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
```
superm1, line 631: I think we need to discuss this - is the touchpad _now_ non-functional while in this state?  I guess if it's recovery mode it was probably _already_ non-functional, but want to confirm.  Otherwise we have a problem for GUI clients...

hughsie: It was already dead. In i2c mode it's essentially a dumb bit of glass that doesn't "do" anything.

hughsie:
> why aren't we using that in the first place to flash

As I understand it the kernel support is incomplete and fwupd will support many more devices.

superm1:
> > why aren't we using that in the first place to flash
> 
> As I understand it the kernel support is incomplete and fwupd will support many more devices.

That's a sensible reason, OK thanks.

PR #2973, opened 2021-03-04
pixart: Filter the wrong report ID

jyu99:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/pixart-rf/fu-pxi-device.c
+++ b/plugins/pixart-rf/fu-pxi-device.c
@@ -36,6 +36,8 @@
 #define FU_PXI_DEVICE_NOTTFY_RET_LEN		4	/* bytes */
 #define FU_PXI_DEVICE_FW_INFO_RET_LEN		8	/* bytes */
 
+#define FU_PXI_DEVICE_NOTIFY_TIMEOUT_MS		5000
+
 /* OTA target selection */
 enum ota_process_setting {
 	OTA_MAIN_FW,				/* Main firmware */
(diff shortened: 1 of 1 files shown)


plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -255,10 +255,21 @@ fu_pxi_device_wait_notify (FuPxiDevice *self,
...
+
+		if (!fu_udev_device_pread_full (FU_UDEV_DEVICE (self),
+						port, res, (FU_PXI_DEVICE_NOTTFY_RET_LEN + 1) - port,
+						error))
+			return FALSE;
+		if (res[0x0] == PXI_HID_DEV_OTA_INPUT_REPORT_ID)
+			break;
+
+		/* delay for next poll report */
+		g_usleep (5);
```
hardboprobot, line 270: Is this necessary? I think the `pread()` should block here, so `fu_udev_device_pread_full()` will wait for new data to be received. Anyway, I'm going to test this version.

hughsie: We should also have a limit to the amount of time spent retrying, and aborting after say 5 seconds. Other pugins do this already, search for `GTimer`.

hardboprobot: Tested ok. Still, when moving the mouse during the upgrade I sometimes get an ioctl I/O error when trying to reset it:

```
Restarting device…       [  -                                    ]failed to reset: ioctl error: Input/output error
```
I don't think there's a lot the plugin can do to fix this, off the top of my head, since the reset command is a simple HIDIOCSFEATURE ioctl call. Maybe it's a timing issue and the device isn't responding to that feature report id in the instants right after the FW flashing? Anyway, it doesn't look catastrophic to me, as it happens under worst conditions (me moving the mouse all the way through). Still ugly, but not catastrophic.

I tested the same code removing the `g_usleep` call. Definitely not needed.

hardboprobot: The Android updater waits for a value change in a characteristic property after issuing the "FW_UPGRADE" command and before sending the "RESET" command, so it looks like that might have something to do with it. The pixart_rf plugin simply waits for the result of the FU_PXI_DEVICE_CMD_FW_UPGRADE command, I don't know if that's equivalent to the Android version.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -251,23 +253,26 @@ fu_pxi_device_wait_notify (FuPxiDevice *self,
...
 			   GError **error)
 {
-	guint8 res[FU_PXI_DEVICE_OTA_BUF_SZ] = {
-		PXI_HID_DEV_OTA_INPUT_REPORT_ID,
-		0x0,
-	};
-
-	/* skip the wrong reprot id ,and keep polling until result is correct */
-	while(1) {
+	g_autoptr(GTimer) timer = g_timer_new();
```
hughsie, line 256: space before the bracket pls

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -251,23 +253,26 @@ fu_pxi_device_wait_notify (FuPxiDevice *self,
...
+	guint8 res[FU_PXI_DEVICE_OTA_BUF_SZ] = {0};
```
hughsie, line 257: space either size of the 0 pls

hughsie (approved):
two nits, and then pleas also squash it down to one commit please

PR #2981, opened 2021-03-07
Add FuEfiFirmware to parse EFI file volumes

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
Other things pending, e.g. FFS2 -- but I wanted to get the basics in place first.

libfwupdplugin/fu-efi-firmware-volume.c
```diff
@@ -0,0 +1,386 @@
...
+	g_autoptr(GBytes) blob = NULL;
+
+	/* check if this is _FIT */
+	if (!fu_common_read_uint32_safe (buf, bufsz, offset,
+					 &sig, G_LITTLE_ENDIAN, error)) {
+		g_prefix_error (error, "failed to read start signature: ");
+		return FALSE;
+	}
+	if (sig == FU_EFI_FIT_SIGNATURE) {
+		*offset_val += 0x150000;
```
superm1, line 76: I think this magic number should be a define.

superm1 (commented):
why commit the binary files to tree?  it seems that they can be generated by your XML.

libfwupdplugin/fu-efi-firmware-volume.c
```diff
@@ -0,0 +1,386 @@
...
+	for (guint i = 0; i < 0x10; i++) {
+		if (buf[offset + FU_EFI_FIRMWARE_VOLUME_OFFSET_ZERO_VECTOR + i] != 0x0) {
+			g_set_error (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "reserved section invalid @0x%x",
+				     (guint) offset + FU_EFI_FIRMWARE_VOLUME_OFFSET_ZERO_VECTOR + i);
+			return FALSE;
+		}
+	}
```
superm1, line 104: does the spec actually say that it should be zero?  Usually reserved means don't care - it should be zero, but a future version of a spec can encroach into this area and so looking for zero might mean a failure later.

libfwupdplugin/fu-efi-firmware-volume.c
```diff
@@ -0,0 +1,386 @@
...
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "invalid volume length");
+		return FALSE;
+	}
+	if (!fu_common_read_uint32_safe (buf, bufsz,
+					 offset + FU_EFI_FIRMWARE_VOLUME_OFFSET_ATTRS,
+					 &attrs, G_LITTLE_ENDIAN, error))
+		return FALSE;
```
superm1, line 133: don't you want to prefix the error here?

hughsie: Agree, fixed.

libfwupdplugin/fu-efi-firmware-volume.c
hughsie: You're completely correct; Apple firmware seems to put some kind of CRC here. I'll remove this check.

libfwupdplugin/fu-efi-firmware-volume.c
hughsie: Agree, fixed.

hughsie:
> why commit the binary files to tree

Good question; the oss-fuzz.py doesn't actually do a build, and so the generated files wouldn't be created. I do wonder if that should be created using a build-fuzzing-samples.sh or something, and that can be called from meson and from the oss-fuzz.py thing.

hughsie:
I'll pull the FFS2 stuff out, that's not baked very well yet.

superm1:
> Good question; the oss-fuzz.py doesn't actually do a build, and so the generated files wouldn't be created. I do wonder if that should be created using a build-fuzzing-samples.sh or something, and that can be called from meson and from the oss-fuzz.py thing.

Yes, I think having a separate build target of some sort to make them makes most sense.

libfwupdplugin/fu-efi-firmware-volume.c
```diff
@@ -0,0 +1,377 @@
...
+	if (!fu_common_read_uint16_safe (buf, bufsz,
+					 offset + FU_EFI_FIRMWARE_VOLUME_OFFSET_EXT_HDR,
+					 &ext_hdr, G_LITTLE_ENDIAN, error)) {
+		g_prefix_error (error, "failed to read ext_hdr: ");
+		return FALSE;
+	}
+	if (!fu_common_read_uint8_safe (buf, bufsz,
+					offset + FU_EFI_FIRMWARE_VOLUME_OFFSET_REVISION,
+					&revision, error))
+		return FALSE;
```
superm1, line 153: missing a prefix error here too

hughsie:
Galaxy brain wave: I don't like the design this headed in: The FuEfiFirmwareVolume should be parsing "one thing" and not adding the images as different volumes. It's kinda working around that there doesn't exist a FuBiosFirmware that reads in the BIOS section of the IFD and spits out the different FVs. At the moment I only care about the FVs in the BIOS image, but what if we did actually want to look at the _FIT_ in the future. New patch incoming, probably tomorrow.

superm1:
> FuBiosFirmware that reads in the BIOS section of the IFD and spits out the different FVs

Well not to mention that you can actually have UEFI and firmware volumes on a lot of architectures that aren't Intel.

hughsie:
> you can actually have UEFI and firmware volumes on a lot of architectures that aren't Intel

But isn't the BIOS blob in the IFD Intel specific? e.g. _jump 16MiB, read FV, read FV, read FV, read _FIT__ seems really Intel specific to me.

superm1:
> > you can actually have UEFI and firmware volumes on a lot of architectures that aren't Intel
> 
> But isn't the BIOS blob in the IFD Intel specific? e.g. _jump 16MiB, read FV, read FV, read FV, read _FIT__ seems really Intel specific to me.

Yes, but this PR is about parsing EFI FV's.  I'm in agreement I think you need a level of abstraction up, but you need one in the middle too.  Maybe something like this:

FuPlatformFw
-> FuIntelFIT

Then you have an opportunity to have alternatives to `FuIntelFit`.  And multiple types can reference `FuEfiVolume`.

PR #2986, opened 2021-03-09
pixart: Add suggestion feature

jyu99:
Type of pull request:
- [x] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/pixart-rf/README.md
+++ b/plugins/pixart-rf/README.md
@@ -26,6 +26,7 @@ Pixart Imaging, Inc and Primax Electronics, Ltd, e.g.
  * `HIDRAW\VEN_093A&DEV_2801`
  * `HIDRAW\VEN_0461&DEV_4EEF`
  * `HIDRAW\VEN_0461&DEV_4EEF&NAME_${NAME}`
+ * `HIDRAW\VEN_0461&DEV_4EEF&MODEL_${MODEL_NAME}`
 
 Additionaly, a custom GUID values including the name is used, e.g.
 
--- a/plugins/pixart-rf/fu-pxi-device.c
+++ b/plugins/pixart-rf/fu-pxi-device.c
@@ -30,6 +30,7 @@
 #define FU_PXI_DEVICE_CMD_FW_OTA_INIT_NEW	0x27u
 #define FU_PXI_DEVICE_CMD_FW_OTA_RETRANSMIT	0x28u
 #define FU_PXI_DEVICE_CMD_FW_OTA_DISCONNECT	0x29u
+#define FU_PXI_DEVICE_CMD_FW_OTA_GET_MODEL	0x2bu
 
 #define FU_PXI_DEVICE_OBJECT_SIZE_MAX		4096	/* bytes */
 #define FU_PXI_DEVICE_OTA_BUF_SZ		512	/* bytes */
@@ -37,6 +38,7 @@
 #define FU_PXI_DEVICE_FW_INFO_RET_LEN		8	/* bytes */
 
 #define FU_PXI_DEVICE_NOTIFY_TIMEOUT_MS		5000
+#define FU_PXI_DEVICE_UNDEFINED_MODEL_STR	"Undefined"
 
 /* OTA target selection */
 enum ota_process_setting {
@@ -52,6 +54,7 @@ enum ota_spec_check_result {
 	OTA_PROCESS_ILLEGAL		= 3,	/* Illegal OTA process */
 	OTA_RECONNECT			= 4,	/* Inform OTA app do reconnect */
 	OTA_FW_IMG_VERSION_ERROR	= 5,	/* FW image file version check error */
+	OTA_DEVICE_LOW_BATTERY		= 6,	/* Device is under low battery */
 	OTA_SPEC_CHECK_MAX_NUM,			/* Max number of OTA driver defined error code */
 };
 
@@ -72,6 +75,7 @@ struct _FuPxiDevice {
 	guint16		 mtu_size;
 	guint16		 prn_threshold;
 	guint8		 spec_check_result;
+	guint8 		 model_name[FU_PXI_DEVICE_MODEL_NAME_LEN];
 };
 
 G_DEFINE_TYPE (FuPxiDevice, fu_pxi_device, FU_TYPE_UDEV_DEVICE)
@@ -102,6 +106,8 @@ fu_pxi_device_spec_check_result_to_string (guint8 spec_check_result)
 		return "reconnect";
 	if (spec_check_result == OTA_FW_IMG_VERSION_ERROR)
 		return "fw-img-version-error";
+	if (spec_check_result == OTA_DEVICE_LOW_BATTERY)
+		return "device battery is too low";
 	return NULL;
 }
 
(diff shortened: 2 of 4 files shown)


plugins/pixart-rf/README.md
```diff
@@ -26,6 +26,7 @@ Pixart Imaging, Inc and Primax Electronics, Ltd, e.g.
  * `HIDRAW\VEN_093A&DEV_2801`
  * `HIDRAW\VEN_0461&DEV_4EEF`
  * `HIDRAW\VEN_0461&DEV_4EEF&NAME_${NAME}`
+  * `HIDRAW\VEN_0461&DEV_4EEF&MODEL_${MODLE_NAME}`
```
hardboprobot, line 29: s/MODLE_NAME/MODEL_NAME

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -72,6 +74,7 @@ struct _FuPxiDevice {
 	guint16		 mtu_size;
 	guint16		 prn_threshold;
 	guint8		 spec_check_result;
+	guint8 		 model_name[12];
```
hardboprobot, line 77: nit: The literal `12` appears in many places in this file. I'd rather have that defined as a macro instead.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -131,8 +137,36 @@ fu_pxi_device_prepare_firmware (FuDevice *device,
...
+	}
+	if (g_getenv ("FWUPD_PIXART_RF_VERBOSE") != NULL) {
+
+		fu_common_dump_raw (G_LOG_DOMAIN, "self->model_name",
+				    self->model_name, 12);
+		fu_common_dump_raw (G_LOG_DOMAIN, "model_name",
+				    model_name, 12);
+	}
+
+	for(guint idx = 0; idx < 12; idx++) {
```
hardboprobot, line 158: nit: If the strings are known to be null-terminated, `g_strcmp0` can be used instead of this loop

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -601,8 +635,13 @@ fu_pxi_device_fw_get_info (FuPxiDevice *self, GError **error)
 	if (!fu_pxi_device_set_feature (self, req, error))
 		return FALSE;
 
+
+	/* delay for BLE device read command */
+	g_usleep (10 * 1000);
```
hardboprobot, line 640: This wait is only used before certain `fu_pxi_device_get_feature` calls. Is that because the device takes longer to reply to these specific commands?

jyu99: fw receive command and then will set result in global var. It could get the wrong data,if result is not ready.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -52,6 +53,7 @@ enum ota_spec_check_result {
 	OTA_PROCESS_ILLEGAL		= 3,	/* Illegal OTA process */
 	OTA_RECONNECT			= 4,	/* Inform OTA app do reconnect */
 	OTA_FW_IMG_VERSION_ERROR	= 5,	/* FW image file version check error */
+	OTA_DEVICE_LOW_BATTERY		= 6,	/* Device is under low battery */
```
hardboprobot, line 56: @hughsie I don't think the new battery level API is useful for this device. It will do the battery check internally instead of returning a battery level. On the other hand, the device is not put in "bootloader" mode and the upload won't start if the battery is too low.

For completeness, maybe the `FU_PXI_DEVICE_CMD_FW_OTA_INIT_NEW` request can be moved or copied to `fu_pxi_device_prepare_firmware` and then do something like `fu_device_set_battery_level (dev, 1)` if the battery check fails, but I don't know if it's worth the trouble.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -131,8 +137,36 @@ fu_pxi_device_prepare_firmware (FuDevice *device,
...
+
 	if (!fu_firmware_parse (firmware, fw, flags, error))
 		return NULL;
+
+	/* check is compatible with hardware */
+	if (!fu_pxi_firmware_get_model_name (FU_PXI_FIRMWARE (firmware), model_name, error)) {
+		return NULL;
+	}
+	if (g_getenv ("FWUPD_PIXART_RF_VERBOSE") != NULL) {
+
```
hughsie, line 151: pls remove extra whitespace

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -131,8 +137,36 @@ fu_pxi_device_prepare_firmware (FuDevice *device,
...
+		fu_common_dump_raw (G_LOG_DOMAIN, "self->model_name",
+				    self->model_name, 12);
+		fu_common_dump_raw (G_LOG_DOMAIN, "model_name",
+				    model_name, 12);
```
hughsie, line 155: that 12 really needs a #define

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -628,6 +667,49 @@ fu_pxi_device_fw_get_info (FuPxiDevice *self, GError **error)
...
+		return FALSE;
+	}
+
+	/* get model from res */
+	if (!fu_memcpy_safe (self->model_name, sizeof(self->model_name), 0x0,	/* dst */
+			     (guint8 *) res, sizeof(res), 0x6,			/* src */
+			     sizeof(self->model_name), error))
+		return FALSE;
+
+
```
hughsie, line 708: just one space pls

plugins/pixart-rf/fu-pxi-firmware.c
```diff
@@ -14,66 +14,119 @@
...
 	for (guint32 i = 0x0; i < sizeof(tag); i++) {
 		guint8 tmp = 0;
-		if (!fu_common_read_uint8_safe (fw_header, sizeof(fw_header),
+		if (!fu_common_read_uint8_safe (fw_header_tmp, sizeof(fw_header_tmp),
 						i + PIXART_RF_FW_HEADER_TAG_OFFSET,
 						&tmp, error))
 			return FALSE;
 		if (tmp != tag[i]) {
-			header_ok = FALSE;
+			return FALSE;
```
hughsie, line 83: if you're returning FALSE you also need to set the `error`

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -122,6 +128,7 @@ fu_pxi_device_to_string (FuDevice *device, guint idt, GString *str)
 	fu_common_string_append_kx (str, idt, "PacketReceiptNotificationThreshold", self->prn_threshold);
 	fu_common_string_append_kv (str, idt, "SpecCheckResult",
 				    fu_pxi_device_spec_check_result_to_string (self->spec_check_result));
+	fu_common_string_append_kv (str, idt, "ModelName", g_strndup ((gchar *) self->model_name, FU_PXI_DEVICE_MODEL_NAME_LEN));
```
hughsie, line 131: you need to do something sucky like:

   g_autofree gchar *model_name = NULL;
   g_strndup ((gchar *) self->model_name, FU_PXI_DEVICE_MODEL_NAME_LEN)
   fu_common_string_append_kv (str, idt, "ModelName", model_name);

..otherwise you leak memory.

plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -131,8 +138,45 @@ fu_pxi_device_prepare_firmware (FuDevice *device,
...
 	g_autoptr(FuFirmware) firmware = fu_pxi_firmware_new ();
+	FuPxiDevice *self = FU_PXI_DEVICE (device);
+	guint8 model_name[FU_PXI_DEVICE_MODEL_NAME_LEN];
+
 	if (!fu_firmware_parse (firmware, fw, flags, error))
 		return NULL;
+
+	/* check is compatible with hardware */
+	if (!fu_pxi_firmware_get_model_name (FU_PXI_FIRMWARE (firmware), model_name, error)) {
+		return NULL;
```
hughsie, line 149: no need for `{}`'s

hughsie:
Merged in another PR, thanks!

PR #2988, opened 2021-03-09
Remove unused, unsafe and deprecated functions from libfwupdplugin

hughsie:
Keeping *internal* API and ABI compatibility makes working with an already
complex codebase more mentally demanding than it needs to be.

Remember: plugins should be in-tree and upstream! If your out of tree plugin
stops working then it should be upstream.

The public-facing libfwupd will remain API and ABI stable for obvious reasons.


hughsie:
@superm1 -- I agonised last night for a few hours about a way to fix the FuFirmware design issues without breaking ABI and this morning came to the conclusion that we're making a rod for our own back trying to maintain compatibility for what was supposed to be an internal interface. I turned the "what API is deprecated" document into a "how to port to the new API" document.

hughsie:
FYI, deleting FuFirmwareImage and just using FuFirmware is `1103 insertions(+), 1513 deletions(-)`

libfwupdplugin/README.md
```diff
@@ -1,12 +1,25 @@
...
- * Remove FuHidDevice->open() and FuHidDevice->close()
- * Remove FuUsbDevice->probe(), FuUsbDevice->open() and FuUsbDevice->close()
- * Remove FuUdevDevice->to_string(), FuUdevDevice->probe(), FuUdevDevice->open() and FuUdevDevice->close()
- * Remove fu_device_get_protocol() and fu_device_set_protocol()
+libfwupdplugin
+==============
+
+This library is deliberately API and ABI unstable. Think of it like the kernel,
+where keeping unused, unsafe and deprecated functions around forever is a
+maintainance burden.
```
superm1, line 6: This is too strong of a statement.  I don't really want to see plugin API/ABI bumping every release or anything.  Can we meet somewhere in the middle?  We only bump API/ABI for plugins on major jumps?  IE fwupd 1.5.x to 1.6.?  Because otherwise we're going to have a harder time bringing this into LTS releases for the future.

superm1 (changes requested):
Need to bump packaging as well for libfwupdplugin1 -> libfwupdplugin2

libfwupdplugin/README.md
hughsie: > We only bump API/ABI for plugins on major jumps?

I guess that's a compromise.

> we're going to have a harder time bringing this into LTS releases

Can you elaborate a bit please? For RHEL it would be acceptable as it's an internal interface, not designed to be used by other things.

hughsie:
> Need to bump packaging as well for libfwupdplugin1 -> libfwupdplugin2

I think I did this.

libfwupdplugin/README.md
hughsie: So then the next question becomes, if we branch for 1.6.0 now, at what point do we post-date the branch? e.g. do we want these things in 1.5.8?

* QUBES? [yes?]
* fu_common_write_uint16_safe[yes? seems safe and useful]
* battery_level [yes?]
* fu-ifd-firmware [no? on the logic that nothing uses it yet, and it might all change]
* fu_plugin_add_possible_quirk_key and the quirk prefix removal [yes? on the logic i want stuff easy to backport]
* FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID [yes? on the logic it fixes real bugs and has been tested really well...]

hughsie: Okay, I've pushed a 1_5_X branch now. I'll fix up master in a new PR now.

hughsie: I've softened the README -- does that look better?

PR #2992, opened 2021-03-09
backport uefi-capsule: Specify the section flags for .sbat for 1_2_X

superm1:
When using "objcopy -O binary" to generate AArch64 EFI images, it
silently drops the sections without "alloc" or "load" or the sections
with "unload", and this caused the content of .sbat was skipped in the
final EFI image.

This commit sets the common read-only data section flags to .sbat to
make sure the content will be copied.

Signed-off-by: Gary Lin <glin@suse.com>

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
hmmm, @hughsie we don't have `abi` job or `build-arch` job on 1_2_X.  I can't seem to merge.

PR #2993, opened 2021-03-09
Wip/hughsie/fu firmware ng

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/synaptics-mst/fu-self-test.c
```diff
@@ -138,6 +138,7 @@ main (int argc, char **argv)
 	g_log_set_fatal_mask (NULL, G_LOG_LEVEL_ERROR | G_LOG_LEVEL_CRITICAL);
 
 	g_assert_cmpint (g_mkdir_with_parents ("/tmp/fwupd-self-test/var/lib/fwupd", 0755), ==, 0);
+	g_test_expect_message ("FuQuirks", G_LOG_LEVEL_WARNING, "invalid key names*");
```
superm1, line 141: if it's transient, then shouldn't this cause failures transiently as well?  This type of code is deterministic.

hughsie: I'll drop this for now; I'll have a deeper look later.

superm1 (approved):

PR #2994, opened 2021-03-10
Some fix for let firmware update be successful and improve the performance.

blueue:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/synaptics-rmi/fu-synaptics-rmi-device.c
+++ b/plugins/synaptics-rmi/fu-synaptics-rmi-device.c
@@ -187,6 +187,13 @@ fu_synaptics_rmi_device_set_iepmode (FuSynapticsRmiDevice *self, gboolean iepmod
 }
 
 gboolean
+fu_synaptics_rmi_device_get_iepmode (FuSynapticsRmiDevice *self)
+{
+	FuSynapticsRmiDevicePrivate *priv = GET_PRIVATE (self);
+	return priv->in_iep_mode;
+}
+
+gboolean
 fu_synaptics_rmi_device_write_bus_select (FuSynapticsRmiDevice *self, guint8 bus, GError **error)
 {
 	FuSynapticsRmiDeviceClass *klass_rmi = FU_SYNAPTICS_RMI_DEVICE_GET_CLASS (self);
@@ -386,6 +393,9 @@ fu_synaptics_rmi_device_setup (FuDevice *device, GError **error)
 	/* set page */
 	if (!fu_synaptics_rmi_device_set_page (self, 0, error))
 		return FALSE;
+
+	/* Force entering iep mode again. */
+	fu_synaptics_rmi_device_set_iepmode (self, FALSE);
 	if (!fu_synaptics_rmi_device_enter_iep_mode (self, error))
 		return FALSE;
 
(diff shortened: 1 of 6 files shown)


plugins/synaptics-rmi/fu-synaptics-rmi-device.c
```diff
@@ -386,6 +393,9 @@ fu_synaptics_rmi_device_setup (FuDevice *device, GError **error)
 	/* set page */
 	if (!fu_synaptics_rmi_device_set_page (self, 0, error))
 		return FALSE;
+
+	/* Force entering iep mode again. */
+	fu_synaptics_rmi_device_set_iepmode (self, FALSE);
 	if (!fu_synaptics_rmi_device_enter_iep_mode (self, error))
```
superm1, line 399: why doesn't the call for `fu_synaptics_rmi_device_enter_iep_mode` carry the void call for `fu_synaptics_rmi_device_set_iepmode`?

plugins/synaptics-rmi/fu-synaptics-rmi-firmware.c
```diff
@@ -510,6 +511,12 @@ fu_synaptics_rmi_firmware_get_sig_size (FuSynapticsRmiFirmware *self)
 	return self->sig_size;
 }
 
+guint32
+fu_synaptics_rmi_firmware_get_firmware_size (FuSynapticsRmiFirmware *self)
+{
+	return self->firmware_size;
+}
+
```
superm1, line 519: I wonder if it's going to be better to rebase this on https://github.com/fwupd/fwupd/pull/2998 when that goes in.

plugins/synaptics-rmi/fu-synaptics-rmi-device.c
blueue: It will determine whether it needs issuing entering iep mode command really to device depends on in_iep_mode in fu_synaptics_rmi_device_enter_iep_mode(), and device will be back to non-iep mode by itself in some cases without any notification in application layer. I put fu_synaptics_rmi_device_set_iepmode() to make in_iep_mode be FALSE to let entering iep mode command be issued.

hughsie:
Can you please split this commit apart, it's mixing at least three different things into one commit. Additionally, I'm now very confused about what IEP mode actually is. It seems we have a tristate now, i.e. "runtime, bootloader, iep" and if that's correct I don't think trying to squash them into a single boolean is working well.

blueue:
Apologize for mixing lots of changes into one commit, but they are the all necessary changes to make firmware updating be successful. I can explain the change if there is a question.

In PS/2 Synaptics device, we need switch device driver from **psmouse** to **serio_raw** because we can communicate with device by serio_raw driver only. 
The entering bootloader mode for Synaptics touchpad is **writing bootloader ID** to device and i**ssuing enable_flash_prog command** to device. That is what we do in _fu_synaptics_rmi_v5_device_detach()_.
The entering runtime mode for Synaptics touchpad is **writing RESET command** to device. That is what we do in _fu_synaptics_rmi_device_reset()_. 
Then we need switch device driver from **serio_raw** to **psmouse** to let device be ready to report packet. 

IEP mode is a special mode in bootloader mode of Synaptics touchpad firmware. The device would report different value from the same register address depends on whether it is in IEP mode or not, and taking different action by itself with the same command in the same register address as well. However, we cannot get IEP mode status of device in application layer, so there are lots of changes of assigning in_iep_mode to be FALSE to issue the **entering IEP mode command** again by experience.

hughsie:
Alternate proposal in https://github.com/fwupd/fwupd/pull/3043

superm1:
closing since #3043 merged

PR #2996, opened 2021-03-10
trivial: Add fu_common_align_up() for future usage

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -3206,3 +3206,38 @@ fu_common_uri_get_scheme (const gchar *uri)
 		return NULL;
 	return g_utf8_strdown (uri, tmp - uri);
 }
+
+/**
+ * fu_common_align_up:
+ * @value: value to align
+ * @alignment: align to this power of 2
+ *
+ * Align a value to a power of 2 boundary, where @alignment is the bit position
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -3189,3 +3189,30 @@ fu_common_uri_get_scheme (const gchar *uri)
...
+
+/**
+ * fu_common_align_up:
+ * @value: value to align
+ * @alignment: align to this power of 2
+ *
+ * Align a value to a power of 2 boundary, where @alignment is the bit position
+ * to align to. If @alignment is zero then @value is always returned unchanged.
+ *
+ * Returns: aligned value, which will be the same as @value if already aligned
```
superm1, line 3201: I think you might need a check that you aren't getting bigger than `G_MAXSIZE` as a result of this rounding up.

superm1:
> i'm not 100% on the name, so other ideas welcome

seems fine by me name wise

libfwupdplugin/fu-common.c
hughsie: Good idea. Fixed.

superm1 (approved):

PR #2998, opened 2021-03-10
trivial: Add fu_firmware_get_size()

hughsie:
The idea here is to return the size of the firmware, including the header,
footer or other encapsulation. It would be expected that this value would
include the alignment if provided.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1 (dismissed):
I think `container_size` or `binary_size` is a better, but code looks OK.

hughsie:
> I think container_size or binary_size is a better

I think given how RMI wants to do the same thing, I think that kind of tips the balance in the shorter name.

PR #3001, opened 2021-03-10
pixart: Support building firmware with the model_name metadata

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


plugins/pixart-rf/fu-pxi-firmware.c
```diff
@@ -155,6 +171,17 @@ fu_pxi_firmware_write (FuFirmware *firmware, GError **error)
 			     (guint) version_raw);
 		return NULL;
 	}
+	if (self->model_name != NULL) {
+		gsize model_namesz = MIN (strlen (self->model_name),
```
hardboprobot, line 175: Are we sure `self->model_name` is null-terminated? I'm now looking at `fu_pxi_firmware_parse()` again and `model_name` is simply a 12-bit copy of a portion of the FW tag. I'm inspecting one of the test binaries and the model_name _is_ properly terminated, but I think the code should guard against that anyway.

Then again, the model in the FW binaries is not prone to change, so it's not a priority. We can go with this for now and make the proper string-checking changes later.

hughsie: `self->model_name` has to be, as `g_strndup` ensures this.

hardboprobot: Oh you're right!

hardboprobot, line 175: Shouldn't this always be `FU_PXI_DEVICE_MODEL_NAME_LEN` regardless of the length of `self->model_name`?

hughsie: I don't thjink so. If `self->model_name` is `Test\0` and we try reading past item 5 in the memcpy then we're going to at best get junk and at worse get an asan error.

hardboprobot: What I meant was that `fu_pxi_firmware_parse` expects `fw_header` to contain a 12-bit model_name field, and if the model_name string here is shorter than that, maybe the rest of the 12 bits should be padding.

But I haven't tested it and, frankly, I don't understand some things about `fu_firmware_parse` (for example, does it take the whole binary blob as an argument or only a fraction of it?). Besides, I don't know all the possible use cases of this, so don't take my suggestion as a blocker.

hughsie: Ahh, I see what you mean. I think the (incorrectly named) `fw_header[PIXART_RF_FW_HEADER_SIZE] = { 0x0 };` is set to 0x0, so it's predictable even if the string is shorter.

hardboprobot (approved):

PR #3005, opened 2021-03-10
solokey: Add support for writing firmware

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


src/fuzzing/firmware/solokey.json
```diff
@@ -1 +1 @@
-{"firmware": "OjA0NDAwMDAwM0RFRjIwRjA4MAo6MDAwMDAwMDFGRgo=", "signature": "aGVsbG8gd29ybGQ="}
+{"firmware":"OjA0NDAwMDAwM0RFRjIwRjA4MAo6MDAwMDAwMDFGRgo=","signature":"aGVsbG8gd29ybGQ="}
```
superm1, line 1: that intentional that no newline?

hughsie: Not at all, fixed, thanks.

PR #3006, opened 2021-03-10
Some fixups for Thunderbolt

superm1:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -1549,7 +1549,12 @@ fu_util_activate (FuUtilPrivate *priv, gchar **values, GError **error)
 		return FALSE;
 
 	/* load engine */
-	if (!fu_util_start_engine (priv, FU_ENGINE_LOAD_FLAG_READONLY, error))
+	if (!fu_util_start_engine (priv,
+				   FU_ENGINE_LOAD_FLAG_READONLY |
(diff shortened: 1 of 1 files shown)


plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -776,7 +776,8 @@ fu_thunderbolt_device_write_firmware (FuDevice *device,
 	}
 
 	/* whether to wait for a device replug or not */
-	if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER &&
```
djcampello, line 779: I am not entirely sure this is always the case. This might be true only on the NDA case.

djcampello (approved):

djcampello:
I guess I approved both changes by mistake. See my previous comment

plugins/thunderbolt/fu-thunderbolt-device.c
superm1: OK I'll take just the other one until we can confirm what happens in device attached case.

hughsie (approved):
Mea culpa.

PR #3013, opened 2021-03-11
cros-ec: Allow a failure to write a section, do another write

### LINKED ISSUE #3012: cros-ec-usb-device: Retry write process in the event of failures due to interrupted usb connection
**Describe the bug**
During development of quiche, I've observed that the RO firmware's USB endpoint will occasionally reset briefly upon rebooting into it. This is enough to interrupt a write in progress, which causes the cros-ec plugin to fail write, and bail out entirely.

**Steps to Reproduce**
Steps to reproduce the behavior.
Flash quiche from RW -> RO -> RW several times. Eventually should fail like this:
04:49:33:0957 FuEngine             Emitting PropertyChanged('Status'='device-write')
Writing?                 [                     \                 ]04:49:33:0957 FuPluginCrosEc       trimmed 584 trailing bytes
04:49:33:0957 FuPluginCrosEc       sending 0x2fdb8 bytes to 0x10000
04:49:33:0970 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0970 FuPluginCrosEc       failed to flush to idle
04:49:33:0970 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0970 FuPluginCrosEc       failed to flush to idle
04:49:33:0970 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0970 FuPluginCrosEc       failed to flush to idle
04:49:33:0971 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0971 FuPluginCrosEc       failed to flush to idle
04:49:33:0971 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0971 FuPluginCrosEc       failed to flush to idle
04:49:33:0972 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0972 FuPluginCrosEc       failed to flush to idle
04:49:33:0972 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0972 FuPluginCrosEc       failed to flush to idle
04:49:33:0972 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0972 FuPluginCrosEc       failed to flush to idle
04:49:33:0972 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0972 FuPluginCrosEc       failed to flush to idle
04:49:33:0973 FuDevice             fu_device_retry_full: assertion 'error != NULL' failed
04:49:33:0973 FuPluginCrosEc       failed to flush to idle

**Expected behavior**
Retry. Actually, just rerunning fwupdtool install-blob again with the device sitting idle in the RO section continues the update 

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
Self built as a part of chromiumos.

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
$ fwupdmgr get-devices --show-all-devices
```

**Additional questions**
- Operating system and version: Chrome OS
- Have you tried rebooting? Not relevant.
- Is this a regression? Nope.

bleungatchromium:
The RO partition's USB endpoint has been observed to occasionally hiccup
(usb disconnect and reenumeration) upon reboot and initial attach, causing a
failure to update a section. Rather that making this fatal, let's set
ANOTHER_WRITE_REQUIRED and allow us to handle the new USB device after
this event.

fixes #3012

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -811,10 +811,21 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
 		FuCrosEcFirmwareSection *section = g_ptr_array_index (sections, i);
 
 		if (section->ustatus == FU_CROS_EC_FW_NEEDED) {
+			g_autoptr(GError) error_local = NULL;
+
 			if (!fu_cros_ec_usb_device_transfer_section (device,
 								     firmware,
 								     section,
-								     error)) {
+								     &error_local)) {
+				if (g_error_matches (error_local,
+						     G_USB_DEVICE_ERROR,
+						     G_USB_DEVICE_ERROR_NOT_SUPPORTED)) {
+					g_debug ("failed to transfer section, trying another write, ignoring error: %s",
+						 error_local->message);
+					fu_device_add_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED);
+					return TRUE;
+				}
+				g_propagate_error (error, g_steal_pointer (&error_local));
 				return FALSE;
 			}
 			num_txed_sections++;


plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -806,7 +806,10 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
 								     firmware,
 								     section,
 								     error)) {
-				return FALSE;
+				g_debug ("failed to transfer section, trying another write\n");
+				fu_device_add_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED);
+				g_clear_error (error);
```
superm1, line 811: typically what you should do if an error is expected to not be critical is have a local scope `g_autoptr error_local` and then use that.  However I'd suspect you also only want to do this `another_write_required` if it's a certain error type - so maybe catch that and check?

hughsie: I pushed a few fixes to wip/hughsie/cros_ec-error but i really want to see that GError->message first

bleungatchromium: Hi @superm1,

I check now for G_USB_DEVICE_ERROR_NOT_SUPPORTED and clear the error only on that case.

Other kinds of failures we will proceed to fail here.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -815,6 +815,13 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
 								     firmware,
 								     section,
 								     error)) {
+				if ((*error)->code == G_USB_DEVICE_ERROR_NOT_SUPPORTED) {
+					g_debug ("failed to transfer section, trying another write, ignoring error: %s",
+						 (*error)->message);
+					fu_device_add_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED);
```
superm1, line 821: @hughsie I wonder if we actually want a different flag for this. My logic is this another write required is behind used differently than others do. This is another write required based on device come and go. Other times we use same device but make another pass.

I want to make sure we acknowledge the difference in the daemon and don't paper over this some day by changing behavior coming in and out of write with a replug.

bleungatchromium: FWIW, the way that I'm using ANOTHER_WRITE_REQUIRED is always in conjunction with a device reboot or some kind of event that results in a reenumeration.

It's possible I'm the only one using it in this way, as last year I filed #2297, where it was broken and wouldn't actually survive the go-around in the detaching and reenumerating case.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -815,6 +815,13 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
 								     firmware,
 								     section,
 								     error)) {
+				if ((*error)->code == G_USB_DEVICE_ERROR_NOT_SUPPORTED) {
+					g_debug ("failed to transfer section, trying another write, ignoring error: %s",
+						 (*error)->message);
```
hughsie, line 820: ew. Can you add `g_autoptr(GError) error_local = NULL` and then use that instead? We don't actually enforce that `error != NULL` and this would be explotastic.

superm1: g_error_matches

bleungatchromium: Thanks. Uploaded a new version.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -811,10 +811,21 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
...
+								     &error_local)) {
+				if (g_error_matches (error_local,
+						     G_USB_DEVICE_ERROR,
+						     G_USB_DEVICE_ERROR_NOT_SUPPORTED)) {
+					g_debug ("failed to transfer section, trying another write, ignoring error: %s",
+						 error_local->message);
+					fu_device_add_flag (device, FWUPD_DEVICE_FLAG_ANOTHER_WRITE_REQUIRED);
+					return TRUE;
+				}
+				g_propagate_error (error,error_local);
```
hughsie, line 828: you need the slightly ungainly: `g_propagate_error (error, g_steal_pointer (&error_local));` -- g_propagate_error was designed for another era...

bleungatchromium: Gotcha. I see the pattern elsewhere. Will do.

hughsie (approved):

PR #3031, opened 2021-03-15
Continuous integration for FreeBSD

pkonkol:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -64,6 +64,11 @@ jobs:
     runs-on: macos-latest
     name: Build fwupd from FreeBSD ports
     steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Find tag
+      id: tagger
+      uses: jimschubert/query-tag-action@v1
     - name: Build
       id: test
       uses: vmactions/freebsd-vm@v0.1.2
@@ -87,6 +92,8 @@ jobs:
           sed -i .old "s/#GH_TAGNAME=.*$/GH_TAGNAME=\t${GITHUB_SHA}/" Makefile
           sed -i .old "s/GH_ACCOUNT=.*$/GH_ACCOUNT=\t${GITHUB_REPOSITORY_OWNER}/" Makefile
           sed -i .old "s/GH_PROJECT=.*$/GH_PROJECT=\t${REPOSITORY_NAME}/" Makefile
+          sed -i .old "s/DISTVERSION=.*$/DISTVERSION=${{steps.tagger.outputs.tag}}/" Makefile
+          cat Makefile
           make makesum
           make package
           make install
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,51 +75,26 @@ jobs:
             vala gtk-doc json-glib gpgme gnutls sqlite3 curl gcab libarchive
         sync: rsync
         run: |
-          pwd
-          env
           mkdir -p /usr/local/etc/pkg/repos/
           cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
           sed -i .old 's|url: "pkg+http://pkg.FreeBSD.org/${ABI}/quarterly"|url: "pkg+http://pkg.FreeBSD.org/${ABI}/latest"|' /usr/local/etc/pkg/repos/FreeBSD.conf
           pkg install -y libtasn1
           pkg upgrade -y libtasn1
-          pkg info libtasn1
           cd /usr
           git clone https://github.com/3mdeb/freebsd-ports.git --depth 1 -b fwupd-port ports
           cd /usr/ports/sysutils/fwupd
-          ls
           REPOSITORY_NAME=$(echo "${GITHUB_REPOSITORY}" | awk -F / '{print $2}')
           sed -i .old "s/#GH_TAGNAME=.*$/GH_TAGNAME=\t${GITHUB_SHA}/" Makefile
           sed -i .old "s/GH_ACCOUNT=.*$/GH_ACCOUNT=\t${GITHUB_REPOSITORY_OWNER}/" Makefile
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -1,61 +1,100 @@
...
         run: |
-          echo $GITHUB_WORKSPACE
-          docker run --privileged -e CI=true -t -v $GITHUB_WORKSPACE:/github/workspace docker.pkg.github.com/fwupd/fwupd/fwupd-${{matrix.os}}:latest
+          mkdir -p /usr/local/etc/pkg/repos/
+          cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+          sed -i .old 's|url: "pkg+http://pkg.FreeBSD.org/${ABI}/quarterly"|url: "pkg+http://pkg.FreeBSD.org/${ABI}/latest"|' /usr/local/etc/pkg/repos/FreeBSD.conf
+          pkg install -y libtasn1
+          pkg upgrade -y libtasn1
+          cd /usr
+          git clone https://github.com/3mdeb/freebsd-ports.git --depth 1 -b fwupd-port ports
```
superm1, line 84: A few comments I notice about this:
1. It seems like `sysutils/fwupd/files/patch-libfwupd_meson.build` should be submitted to apply upstream.  The other ones too for `#ifdef` should each be reviewed too.
2. `DISTVERSION` I doubt you want to hardcode.  For the purpose of CI you should probably be passing this in as an environment variable or with `sed` or something. 
3. Overall I think it would be more flexible if we could keep an "upstream" copy of this BSD Makefile in the fwupd tree and you regularly sync it with the downstream tree.  Otherwise we may end up with a situation that we break the BSD build by hard bumping some dep but not being able to get the downstream BSD Makefile updated.  If we can keep it (or a template of it) upstream then we can just submit the changes to make sure CI keeps working for BSD..

.github/workflows/main.yml
```diff
@@ -1,61 +1,100 @@
 name: Continuous Integration
 on:
   push:
-    branches: [ master, 1_4_X ]
+    branches: [ master, 1_4_X, wip_bsd_ci ]
   pull_request:
-    branches: [ master, 1_4_X ]
+    branches: [ master, 1_4_X, wip_bsd_ci ]
```
superm1, line 6: obviously this is for POC, and after you've got it working this goes away and we only do it on master.

.github/workflows/main.yml
```diff
@@ -1,61 +1,100 @@
...
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-      - name: Refresh dependencies
-        run: sudo apt update
-      - name: Install dependencies
-        run: ./contrib/ci/generate_dependencies.py | sudo xargs apt install -y
-      - name: Check ABI
-        run: ./contrib/ci/check-abi $(git describe --abbrev=0 --tags) $(git rev-parse HEAD)
+  #   abi:
```
superm1, line 9: also, obviously we wouldn't turn off other Github actions jobs.

superm1 (changes requested):

hughsie:
@pkonkol is this somethingyou still want upstream? There's quite some work to do :)

pkonkol:
@hughsie Yes. It's still work in progress, not yet ready for upstreaming but I will update that. This PR is directly related to @Asiderr previous work on moving fwupd to BSD systems.

.github/workflows/main.yml
pkonkol: 1. @Asiderr was the patch finally merged into fwupd/fwupd? By my understanding, the patches were for development purposes and now should be upstreamed.
2. I will update that to use current tag as DISTVERSION
3. My idea was that the CI would use the official freebsd-ports set to specific release (eg. 2020Q4) which would be updated manually. @Asiderr is working on upstreaming fwupd Makefile to freebsd/freebsd-ports and AFAIK it should be done soon.

.github/workflows/main.yml
```diff
@@ -1,61 +1,107 @@
...
         run: |
-          echo $GITHUB_WORKSPACE
-          docker run --privileged -e CI=true -t -v $GITHUB_WORKSPACE:/github/workspace docker.pkg.github.com/fwupd/fwupd/fwupd-${{matrix.os}}:latest
+          mkdir -p /usr/local/etc/pkg/repos/
+          cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+          sed -i .old 's|url: "pkg+http://pkg.FreeBSD.org/${ABI}/quarterly"|url: "pkg+http://pkg.FreeBSD.org/${ABI}/latest"|' /usr/local/etc/pkg/repos/FreeBSD.conf
+          pkg install -y libtasn1
+          pkg upgrade -y libtasn1
+          cd /usr
+          git clone https://github.com/3mdeb/freebsd-ports.git --depth 1 -b fwupd-port ports
```
Asiderr, line 89: @superm1 @hughsie @SergiiDmytruk We should push this forward in the near future. More and more FreeBSD-related changes are coming to fwupd, so a proper CI would be really useful. The upstream of the [fwupd port](https://reviews.freebsd.org/D29332) will take some more time, so at the start we should base CI on the [freebsd-ports fork](https://github.com/3mdeb/freebsd-ports.git), which will contain the fwupd package. 
The FreeBSD related patches are already merged in the fwupd, so this PR needs significant changes. I will address this in the coming days.

hughsie: @Asiderr can you create a wip/freebsd branch on the upstream fwupd repo please, and then we can all just commit fixes on before we squash everything up for commit. If you just **add** the freesbd CI job there we can see what's left to upstream.

Asiderr:
@superm1 @hughsie Moved to https://github.com/fwupd/fwupd/pull/3274

PR #3038, opened 2021-03-17
Make intel-spi actually useful

hughsie:
If we need to export the EFI symbols into libfwupdplugin for other plugins we
can do that when we need them.

### DIFF
--- a/docs/hsi.xml
+++ b/docs/hsi.xml
@@ -393,6 +393,22 @@
       </note>
     </refsect3>
 
+    <refsect3 id="org.fwupd.hsi.Spi.Descriptor">
+      <title>Read-only SPI Descriptor</title>
+      <para>
+        The SPI descriptor must always be read only from all other regions.
+        Additionally on Intel architectures the FLOCKDN register must be set to
+        prevent configuration registers in the SPI BAR from being changed.
+      </para>
+      <itemizedlist>
+        <listitem>
+          <para>
+            For HSI-1 this should be read only from all regions <emphasis>v1.6.0</emphasis>
+          </para>
+        </listitem>
+      </itemizedlist>
+    </refsect3>
+
     <refsect3 id="org.fwupd.hsi.Tpm.Version20">
       <title>TPM 2.0 Present</title>
       <para>
--- a/plugins/intel-spi/fu-plugin-intel-spi.c
+++ b/plugins/intel-spi/fu-plugin-intel-spi.c
@@ -8,6 +8,8 @@
 
 #include "fu-plugin-vfuncs.h"
 
+#include "fu-intel-spi-device.h"
+
 #include "fu-ifd-firmware.h"
 #include "fu-ifd-bios.h"
 
@@ -19,6 +21,7 @@
 void
 fu_plugin_init (FuPlugin *plugin)
 {
+	FuContext *ctx = fu_plugin_get_context (plugin);
 	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
 	fu_plugin_add_firmware_gtype (plugin, NULL, FU_TYPE_EFI_FIRMWARE_FILE);
 	fu_plugin_add_firmware_gtype (plugin, NULL, FU_TYPE_EFI_FIRMWARE_FILESYSTEM);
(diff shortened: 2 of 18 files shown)


hughsie:
This PR escalated quickly. It needs some review, but the simple rename I've split out as https://github.com/fwupd/fwupd/pull/3049 which I think I'll merge as soon as CI finishes.

hughsie:
Needless to say, this only works with secure boot turned off (`iopl`, **and** `/dev/mem`) but I think it'll allow us a playground to explore what we do actually need from the kernel. The long term plan is to get a good chunk of the config-descriptor bits there.

docs/hsi.xml
```diff
@@ -393,6 +393,22 @@
       </note>
     </refsect3>
 
+    <refsect3 id="org.fwupd.hsi.Spi.Descriptor">
+      <title>Read-only SPI Descriptor</title>
+      <para>
+        The SPI descriptor must always be read only from all other IFD regions.
```
superm1, line 399: This is written to reference "Intel flash descriptor" but I think you want it in HSI as a generic requirement, it will vary from different implementations, no?

libfwupdplugin/fu-device.c
```diff
@@ -3427,6 +3427,27 @@ fu_device_report_metadata_post (FuDevice *self)
...
+ * fu_device_add_security_attrs:
+ * @self: A #FuDevice
+ * @attrs: A #FuSecurityAttrs
+ *
+ * Adds HSI security attributes.
+ *
+ * Since: 1.6.0
+ **/
+void
+fu_device_add_security_attrs (FuDevice *self, FuSecurityAttrs *attrs)
```
superm1, line 3440: so if we're having individual device security attributes instead of plugin based, that means probably re-checking all the other uses to see if there is a better way to do it too.

plugins/intel-spi/fu-plugin-intel-spi.c
```diff
@@ -26,4 +28,10 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_firmware_gtype (plugin, NULL, FU_TYPE_EFI_FIRMWARE_VOLUME);
 	fu_plugin_add_firmware_gtype (plugin, NULL, FU_TYPE_IFD_BIOS);
 	fu_plugin_add_firmware_gtype (plugin, NULL, FU_TYPE_IFD_FIRMWARE);
+	fu_plugin_add_udev_subsystem (plugin, "pci");
```
superm1, line 31: at least for now, I would think you want to have a check if lockdown is set to only initiate all these attributes if lockdown _isn't_ set.

libfwupdplugin/fu-device.c
hughsie: I did a review of the existing plugins, and I think CPU is the only one that made sense.

docs/hsi.xml
hughsie: Fixed.

plugins/intel-spi/fu-plugin-intel-spi.c
hughsie: Added, good idea.

docs/hsi.xml
```diff
@@ -393,6 +393,22 @@
       </note>
     </refsect3>
 
+    <refsect3 id="org.fwupd.hsi.Spi.Descriptor">
+      <title>Read-only SPI Descriptor</title>
+      <para>
+        The SPI descriptor must always be read only from all other regions.
+        Additionally the FLOCKDN register must be set to prevent configuration
```
superm1, line 400: I think you need something like ""Additionally on Intel architectures..."

plugins/intel-spi/intel-spi.quirk
```diff
@@ -0,0 +1,7 @@
+[PCI\VEN_8086&DEV_A150]
+Plugin = intel_spi
+Guid = INTEL_SPI_CHIPSET\SUNRISE_POINT
+
+[INTEL_SPI_CHIPSET\SUNRISE_POINT]
+IntelSpiKind = pch
+IntelSpiBar = 0xfe010000
```
superm1, line 7: seems a bit light - is this intentional?

hughsie: keep going, the file gets massive!

superm1: It the plugin is development only (effectively) maybe it should be a complete quirk list and just not installed by default?

hughsie: define "complete"? I'm a bit lost, sorry.

superm1: I mean _a more complete list_ of the intel chipsets.

docs/hsi.xml
hughsie: Agree, thanks.

plugins/intel-spi/intel-spi.quirk
hughsie: Ohh! I'm missing a commit, will fixup.

hughsie: Fixed -- sorry about that.

plugins/intel-spi/generate-quirk.py
```diff
@@ -0,0 +1,76 @@
...
+        self.bios_cntl = bios_cntl
+        self.spibar_proxy = spibar_proxy
+        self.spibar = spibar
+        self.flags = flags
+
+
+if __name__ == "__main__":
+
+    if len(sys.argv) != 2:
+        print("required: /path/to/chipset_enable.c")
```
superm1, line 23: this is kernel source right?  Just a thought - what about a daily or weekly CI job that is checking it against a file at https://github.com/torvalds/linux/ (or similar) and making noise when something is added?  Or even automatically _committing_ when something is added?

hughsie: Nah, this is from flashrom. I'm not 100% in agreement with the auto-sync yet, I'm pondering using flashrom *and* chipsec and take the superset of hardware.

PR #3052, opened 2021-03-22
Add four new category types for new firmware to use

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -670,6 +670,31 @@ fu_util_release_get_name (FwupdRelease *release)
 			 * at system bootup */
 			return g_strdup_printf (_("%s CPU Microcode Update"), name);
 		}
+		if (g_strcmp0 (cat, "X-Battery") == 0) {
+			/* TRANSLATORS: battery refers to the system power source */
+			return g_strdup_printf (_("%s Battery Update"), name);
+		}
+		if (g_strcmp0 (cat, "X-Camera") == 0) {
+			/* TRANSLATORS: camera can refer to the laptop internal
+			 * camera in the bezel or external USB webcam */
+			return g_strdup_printf (_("%s Camera Update"), name);
+		}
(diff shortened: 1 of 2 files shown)


src/fu-util-common.c
```diff
@@ -670,6 +670,24 @@ fu_util_release_get_name (FwupdRelease *release)
 			 * at system bootup */
 			return g_strdup_printf (_("%s CPU Microcode Update"), name);
 		}
+		if (g_strcmp0 (cat, "X-Battery") == 0) {
+			/* TRANSLATORS: battery refers to the system laptop power source */
```
superm1, line 674: I think that many things have BMUs, they don't have to be laptops :)

src/fu-util-common.c
```diff
@@ -670,6 +670,24 @@ fu_util_release_get_name (FwupdRelease *release)
...
+			 * camera in the bezel or external USB webcam */
+			return g_strdup_printf (_("%s Camera Update"), name);
+		}
+		if (g_strcmp0 (cat, "X-TPM") == 0) {
+			/* TRANSLATORS: TPM refers to a Trusted Platform Module */
+			return g_strdup_printf (_("%s TPM Update"), name);
+		}
+		if (g_strcmp0 (cat, "X-Touchpad") == 0) {
+			/* TRANSLATORS: TouchPad refers to the flat pointing
+			 * device on a latop */
```
superm1, line 688: again doesn't need to be a laptop here.

superm1 (approved):
just nits on the comments, no concerns.

src/fu-util-common.c
hughsie: Agreed, thanks.

src/fu-util-common.c
hughsie: Fixed.

PR #3066, opened 2021-03-25
trivial: cros-ec: Fix error string to use inlen

bleungatchromium:
Looks like this error message was copy/pasted from above.
Since we're testing whether actual != inlen, the error message should
output inlen.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):

hughsie:
Related, can we get a device-test .json for this device please?

bleungatchromium:
> Related, can we get a device-test .json for this device please?

Do you have an example of that?

hughsie:
Something like this: https://github.com/fwupd/fwupd/blob/master/data/device-tests/devices/hughski-colorhug2.json

bleungatchromium:
> Something like this: https://github.com/fwupd/fwupd/blob/master/data/device-tests/devices/hughski-colorhug2.json

Ok, thanks.

We're not ready to release firmware for test for quiche (close though), but I can certainly put something together like that for the servo-micro board I sent you guys last year.

PR #3069, opened 2021-03-26
pixart-rf: Retry the SetReport after a successfull firmware update

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
> even without the fix

It seems to fix my ~1/5 failure and if it's not hurting you I think merge this.

hardboprobot (approved):

PR #3075, opened 2021-03-26
Offer the user to refresh after enabling

hughsie:
Some users are enabling updates-testing and then wondering why there are no more
firmware updates available. It appears after a 'fwupdmgr refresh' but that's
not obvious to end users who don't know about how this stuff is built.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -478,30 +478,6 @@ fu_util_modify_remote_warning (FuUtilPrivate *priv, FwupdRemote *remote, GError 
 	return TRUE;
 }
 
-static gboolean
-fu_util_modify_remote (FuUtilPrivate *priv,
-		       const gchar *remote_id,
-		       const gchar *key,
-		       const gchar *value,
-		       GError **error)
-{
-	g_autoptr(FwupdRemote) remote = NULL;
-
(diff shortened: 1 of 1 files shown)


superm1 (approved):

src/fu-util.c
```diff
@@ -1758,32 +1749,74 @@ fu_util_remote_modify (FuUtilPrivate *priv, gchar **values, GError **error)
...
+					 priv->cancellable, error))
+		return FALSE;
+
+	/* ask for permission to refresh */
+	if (fwupd_remote_get_kind (remote) != FWUPD_REMOTE_KIND_DOWNLOAD) {
+		/* TRANSLATORS: success message */
+		g_print ("%s\n", _("Successfully enabled remote"));
+		return TRUE;
+	}
+	if (!priv->assume_yes) {
```
superm1, line 1778: I think this needs `priv->no_metadata_check` or `priv->no_remote_check` to check also.  Otherwise now if you call this command from a non-interactive context it will hang IIRC.

hughsie: Agree, added.

PR #3078, opened 2021-03-29
Ask the user to confirm all CLI actions

hughsie:
For some users typing 'fwupdmgr update' just deploys all updates to all hardware,
and that might come as a shock. Ask the user to confirm all actions explicitly
and show the update notes in a warning box rather than plain text on the console.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -394,48 +394,93 @@ fu_util_update_reboot (GError **error)
 }
 
 gboolean
-fu_util_prompt_warning (FwupdDevice *device, const gchar *machine, GError **error)
+fu_util_prompt_warning (FwupdDevice *device,
+			FwupdRelease *release,
+			const gchar *machine,
+			GError **error)
 {
 	FwupdDeviceFlags flags;
-	g_autofree gchar *str = NULL;
+	const gchar *desc_tmp;
+	gint vercmp;
+	g_autoptr(GString) title = g_string_new (NULL);
+	g_autoptr(GString) str = g_string_new (NULL);
 
-	/* device is already in bootloader mode */
+	/* up, down, or re-install */
+	vercmp = fu_common_vercmp_full (fwupd_release_get_version (release),
+				        fu_device_get_version (device),
+				        fwupd_device_get_version_format (device));
+	if (vercmp < 0) {
+		/* TRANSLATORS: message letting the user know an downgrade is available
+		 * %1 is the device name and %2 and %3 are version strings */
+		g_string_append_printf (title, _("Downgrade %s from %s to %s?"),
+					fwupd_device_get_name (device),
+					fwupd_device_get_version (device),
+					fwupd_release_get_version (release));
+	} else if (vercmp > 0) {
+		/* TRANSLATORS: message letting the user know an upgrade is available
+		 * %1 is the device name and %2 and %3 are version strings */
+		g_string_append_printf (title, _("Upgrade %s from %s to %s?"),
+					fwupd_device_get_name (device),
+					fwupd_device_get_version (device),
+					fwupd_release_get_version (release));
+	} else {
+		/* TRANSLATORS: message letting the user know an upgrade is available
+		 * %1 is the device name and %2 is a version string */
+		g_string_append_printf (title, _("Reinstall %s to %s?"),
+					fwupd_device_get_name (device),
+					fwupd_release_get_version (release));
+	}
+
+	/* description is optional */
+	desc_tmp = fwupd_release_get_description (release);
+	if (desc_tmp != NULL) {
+		g_autofree gchar *desc = fu_util_convert_description (desc_tmp, NULL);
+		if (desc != NULL)
+			g_string_append_printf (str, "\n%s", desc);
+	}
+
+	/* device is not already in bootloader mode so show warning */
 	flags = fwupd_device_get_flags (device);
-	if (flags & FWUPD_DEVICE_FLAG_IS_BOOTLOADER)
-		return TRUE;
+	if ((flags & FWUPD_DEVICE_FLAG_IS_BOOTLOADER) == 0) {
 
-	/* device may reboot */
-	if ((flags & FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE) == 0) {
-		/* TRANSLATORS: warn the user before updating, %1 is a device name */
-		str = g_strdup_printf (_("%s and all connected devices may not be usable while updating."),
-					fwupd_device_get_name (device));
-	/* device can get bricked */
-	} else if ((flags & FWUPD_DEVICE_FLAG_SELF_RECOVERY) == 0) {
-		/* external device */
-		if ((flags & FWUPD_DEVICE_FLAG_INTERNAL) == 0) {
+		/* device may reboot */
+		if ((flags & FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE) == 0) {
+			g_string_append (str, "\n\n");
 			/* TRANSLATORS: warn the user before updating, %1 is a device name */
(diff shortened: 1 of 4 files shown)


hughsie:
Screenshot from 2021-03-29 12-55-33

Screenshot from 2021-03-29 12-56-01

hardboprobot (approved):
That's awesome! Although I can't test the previous scenario of having many different FW options for a device because I refreshed the remote earlier today and now I can only flash the keyboard with the LVFS FW for the proper layout. Everything else looks great.

src/fu-util.c
```diff
@@ -1514,6 +1499,10 @@ fu_util_update_device_with_release (FuUtilPrivate *priv,
 				    FwupdRelease *rel,
 				    GError **error)
 {
+	if (!priv->assume_yes) {
```
superm1, line 1502: This needs another check for interactive.  Otherwise I expect this breaks people who script `fwupdmgr update`

src/fu-util-common.c
```diff
@@ -393,6 +393,63 @@ fu_util_update_reboot (GError **error)
...
+					fwupd_release_get_version (release));
+	} else {
+		/* TRANSLATORS: message letting the user know an upgrade is available
+		 * %1 is the device name and %2 is a version string */
+		g_string_append_printf (str, _("Reinstall %s to %s:"),
+					fwupd_device_get_name (device),
+					fwupd_release_get_version (release));
+	}
+
+	/* no description provided */
```
superm1, line 428: I don't follow this comment, what does it really mean?

src/fu-util-common.c
```diff
@@ -393,6 +393,63 @@ fu_util_update_reboot (GError **error)
...
+		g_autofree gchar *desc = fu_util_convert_description (desc_tmp, NULL);
+		if (desc != NULL)
+			g_string_append_printf (str, "%s", desc);
+		fu_util_warning_box (str->str, 80);
+	}
+
+	/* ask for confirmation */
+	g_print ("\n%s [Y|n]: ",
+		/* TRANSLATORS: prompt to apply the update */
+		_("Perform operation?"));
```
superm1, line 440: Don't we also have a prompt when the device will reboot?  So this will manifest as two prompts?  I think we need to move the reboot message into this function so we only have one prompt.

src/fu-util.c
hughsie: Maybe we just use `no_safety_check`?

src/fu-util-common.c
hughsie: It's supposed to be `description is optional` and the logic was wonky, fixed.

src/fu-util-common.c
hughsie: True, willfix.

hughsie: Combining them makes the PR much nicer too... Updated.

src/fu-util-common.c
```diff
@@ -394,48 +394,92 @@ fu_util_update_reboot (GError **error)
...
+		g_autofree gchar *desc = fu_util_convert_description (desc_tmp, NULL);
+		if (desc != NULL)
+			g_string_append_printf (str, "\n%s", desc);
+	}
+
+	/* device is not already in bootloader mode so show warning */
 	flags = fwupd_device_get_flags (device);
-	if (flags & FWUPD_DEVICE_FLAG_IS_BOOTLOADER)
-		return TRUE;
+	if ((flags & FWUPD_DEVICE_FLAG_IS_BOOTLOADER) == 0) {
```
superm1, line 443: didn't you invert the logic here?

superm1 (commented):
Is there a possibility this turns into a lot of prompts with composite updates or several groups of updates?  I wonder if the whole transaction needs to be calculated - showing the prompt before any device updates and only once.

hughsie:
I see your point about composite updates. The prompt for every top level update was intentional. Will ponder.

hughsie:
> I see your point about composite updates

I pondered this for some time. I do think we want to show the prompt for each update for a composite device. The CLI is being used by the 1% "power users" and I do think it makes sense for them to review each and every update even for a composite device. We already allow users in the GUI to do this, as you have to click a different button on each of the composite device items that have updates.

superm1:
>I pondered this for some time. I do think we want to show the prompt for each update for a composite device. The CLI is being used by the 1% "power users" and I do think it makes sense for them to review each and every update even for a composite device. We already allow users in the GUI to do this, as you have to click a different button on each of the composite device items that have updates.

The problem here is that some composite updates if you decline a single one of them can have unpredictable failures.  If you perform a composite update today, everything contained in it is updated.  This goes both for GUI and command line.

hughsie:
> If you perform a composite update today, everything contained in it is updated. 

If you do `fwupdmgr install foo.cab` that's still true with this patch; we just continue to install *everything* that has requirements that are satisfied.

> This goes both for GUI and command line.

You sure? GNOME Software shows updates per-device, rather than per-release. I think if there are requirements intra-device for composite updates then that needs to be expressed as requirements rather than hoping that they get installed together.

I'm banging this drum again today as someone **new** just emailed me to say they were unhappy they updated the ME firmware when they just wanted to update the unifying receiver. They can't downgrade as the ME is version_lowest to the new version :/

superm1:
> You sure? GNOME Software shows updates per-device, rather than per-release. I think if there are requirements intra-device for composite updates then that needs to be expressed as requirements rather than hoping that they get installed together.

Yes I know it shows updates per-device, but the daemon will take the CAB and process all of the updates contained within the CAB no matter which device you end up updating.  If we don't keep this behavior, I expect problems with being able to update dell-dock.

>I'm banging this drum again today as someone new just emailed me to say they were unhappy they updated the ME firmware when they just wanted to update the unifying receiver. They can't downgrade as the ME is version_lowest to the new version :/

I"m fine with the concept of this branch, but I think it needs to just not prompt more than once for a composite device.  Look and see if it's got a parent, child or silbing, and if so only one message for all of those.

hughsie:
> but the daemon will take the CAB and process all of the updates contained within the CAB no matter which device you end up updating

Honestly, I forgot that, apologies.

>  Look and see if it's got a parent, child or silbing, and if so only one message for all of those.

Give me a few mins.

hughsie:
> Honestly, I forgot that, apologies.

Ahh, I see  how this works now: `fu_main_get_device_family` adds the parent and children for the device when the ID is specified (e.g. not `FWUPD_DEVICE_ID_ANY`). I do wonder if we want a `family_id` or maybe `collection_id` on each FuDevice that identifies devices in the same physical plastic package. I think the `physical_id` isn't the right thing in all cases.

Then the frontend could show all the release notes for all the things we're going to install as part of the "transaction" and *then* ask the user to confirm all of them just once.

superm1:
> > Honestly, I forgot that, apologies.
> 
> Ahh, I see how this works now: `fu_main_get_device_family` adds the parent and children for the device when the ID is specified (e.g. not `FWUPD_DEVICE_ID_ANY`). I do wonder if we want a `family_id` or maybe `collection_id` on each FuDevice that identifies devices in the same physical plastic package. I think the `physical_id` isn't the right thing in all cases.
> 

Yup!

> Then the frontend could show all the release notes for all the things we're going to install as part of the "transaction" and _then_ ask the user to confirm all of them just once.

Instead of individually prompting in general, this could be another way to do it too for everything in the `fwupdmgr update` call.  But it might not really be better than the current - and possibly way worse, obviously people aren't reading what's in the update call at the moment if there are still reports of MEFW update at the same time as Unifying updates and those are surprises.

hughsie:
@superm1 I think this works much better now ^^^ it's using the release checksum to know what other firmware is packaged in the same archive.

PR #3079, opened 2021-03-29
Simplify asking the user about reviews

hughsie:
Although a 4-state question encodes all the possibilities entirely correctly,
it's not easy for end users to understand. Replace the 2x2 logic by two boolean
questions.

This also has the advantage that we only offer to make the upload automatic if
the initial upload actually succeeded.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -39,14 +39,6 @@
 #define EXIT_NOTHING_TO_DO		2
 
 typedef enum {
-	FU_UTIL_HISTORY_DO_NOTHING,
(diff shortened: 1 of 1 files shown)


hughsie:
> does this change also need to use assume yes

a good question, but I don't think so; I kinda see assume-yes as "do everything you need to update automatically" not "do all actions even the optional ones"

superm1 (approved):
since there is `priv->no_unreported_check` earlier in the function, looks good to me.

PR #3080, opened 2021-03-29
Fix build on musl.

ericonr:
malloc_trim isn't a linux function, it's a GNU extension to malloc.  We
can check for it in meson.build, which avoids hardcoding platform
knowledge.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -318,6 +318,12 @@ endif
 if cc.has_header('fnmatch.h')
   conf.set('HAVE_FNMATCH_H', '1')
 endif
+if cc.has_header('malloc.h')
+  conf.set('HAVE_MALLOC_H', '1')
+  if cc.has_function('malloc_trim', prefix: '#include <malloc.h>')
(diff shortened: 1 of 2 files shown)


hughsie (approved):

ericonr:
@superm1 do you prefer a stable version or can it be a rolling release distro? For the first, there's Alpine, and for the latter, there's Void Linux (what I use) and Alpine Edge. I'd be happy to set them up.

hughsie:
I think Void makes a lot of sense personally.

superm1:
I agree - we catch more issues way ahead of it hitting the stable distros with rolling distro.

hughsie:
@ericonr could you add it to https://github.com/fwupd/fwupd/blob/master/.github/workflows/main.yml perhaps?

ericonr:
Yes, I will try to cook up a PR later today or tomorrow :)

PR #3085, opened 2021-03-30
Wip/hughsie/1 1 x sbat

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
Lets see how much CI explodes. Needless to say, this is not what I wanted to do, but RHEL z-stream isn't going to allow a rebase.

superm1:
I think you'll want to backport the CI changes from 1.2.x

martinezjavier (approved):
The backport looks good to me. Thanks for doing this.

hughsie:
I'm really wondering if we want to do this upstream, it's a bag of hacks. I think maybe we just do this as a downstream patch for z-stream RHEL.

PR #3086, opened 2021-03-30
Add fu_device_set_battery_threshold()

hughsie:
This allows a device to set a custom 'check battery level is X%' value from
either the plugin or from a quirk.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -23,6 +23,8 @@
 #define FU_DEVICE_RETRY_OPEN_COUNT			5
 #define FU_DEVICE_RETRY_OPEN_DELAY			500 /* ms */
 
+#define FU_DEVICE_DEFAULT_BATTERY_THRESHOLD		10 /* % */
+
 /**
  * SECTION:fu-device
  * @short_description: a physical or logical device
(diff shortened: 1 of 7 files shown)


hughsie:
> I think we should also export this key into fwupdmgr

And `battery_level` by association?

superm1:
> > I think we should also export this key into fwupdmgr
> 
> And `battery_level` by association?

yeah

hughsie:
So I started doing this, and then stopped. We'd only show the battery level and threshold for peripheral devices. We could of course set the level and threshold on the esrt-derived devices although then it gets a bit meta as to what's responsible for updating them, and also the distinction between "system" and "device" power level breaks down.

hughsie:
Counterproposal. We remove `updatable` (and use `updatable-hidden`) and set an `update_error` when the battery level is below the threshold.

superm1:
> esrt-derived devices

Basically for all uefi devices and whatever identifies as `main-system-firmware`

>Counterproposal. We remove updatable (and use updatable-hidden) and set an update_error when the battery level is below the threshold.

As long as fwupd catches and reacts to the events from upower indicating current battery that seems fine.

hughsie:
I've reverted the bit changing the upower threshold, that needs some more thinking about. Does this one commit look good? I'd like to commit this and then rebase on that for the fu_device_inhibit() next step.

PR #3088, opened 2021-03-30
[WIP] Add Void Linux musl CI.

ericonr:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

I'm a bit lost, the CI setup seems very tightly coupled. I'm copying from arch; do I need to add things to `dependencies.xml` and use `generate_dependencies.py`? Or can I just list packages to install elsewhere?

### DIFF
--- a/contrib/ci/Dockerfile-void-musl.in
+++ b/contrib/ci/Dockerfile-void-musl.in
@@ -1,0 +1,10 @@
+FROM ghcr.io/void-linux/void-linux:latest-full-x86_64-musl
+%%%OS%%%
+ENV LANG en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+ENV CI_NETWORK true
+RUN echo fubar > /etc/machine-id
+RUN xbps-install -Suy xbps && xbps-install -uy
+%%%INSTALL_DEPENDENCIES_COMMAND%%%
+WORKDIR /github/workspace
+CMD ["./contrib/ci/void.sh"]
--- a/contrib/ci/void.sh
+++ b/contrib/ci/void.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -e
+set -x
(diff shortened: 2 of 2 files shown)


superm1:
> I'm copying from arch; do I need to add things to dependencies.xml and use generate_dependencies.py? Or can I just list packages to install elsewhere?

Ideally you'll add all your dependencies in that XML file.  It's one place to choke then when we introduce or remove a new dependency.  If that doesn't make sense for your package manager though, then skip it.

hughsie:
Doesn't it need something to add it the list of github actions?

ericonr:
> Doesn't it need something to add it the list of github actions?

I believe so, the other distros are also listed there. I just haven't touched those files yet.

> Ideally you'll add all your dependencies in that XML file. It's one place to choke then when we introduce or remove a new dependency. If that doesn't make sense for your package manager though, then skip it.

Ok, that makes sense.

hughsie:
@ericonr anything more you want to submit? Someone asked about the status of this just today and I wasn't sure what to say. Thanks.

ericonr:
@hughsie I believe I still need to add a lot of stuff for this to be a useful PR, I just haven't had the time to sit down and figure the whole thing out, sorry.

If anyone comes across this PR and wants to help or add another CI target, you're welcome to do so :)

superm1:
So I did some of the dance to make this sorta work with CI by getting a dockerfile published into our CI through some force pushes of this PR #3553.

It fails though with:
```
/home/runner/work/fwupd/fwupd
standard_init_linux.go:228: exec user process caused: no such file or directory
Error: Process completed with exit code 1.
```

Which I would guess is something like /bin/bash not existing in void's container?  Can you finish up the shell script?
Here is the container as published.
https://github.com/fwupd/fwupd/packages/916833

PR #3090, opened 2021-03-31
Wip/hughsie/fu device inhibit

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -46,6 +46,7 @@ typedef struct {
 	FuDevice			*alternate;
 	FuDevice			*proxy;		/* noref */
 	FuQuirks			*quirks;
+	GHashTable			*inhibits;	/* (nullable) */
 	GHashTable			*metadata;	/* (nullable) */
 	GRWLock				 metadata_mutex;
 	GPtrArray			*parent_guids;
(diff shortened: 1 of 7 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -2524,6 +2663,19 @@ fu_device_sleep_with_progress (FuDevice *self, guint delay_secs)
...
+fu_device_ensure_battery_inhibit (FuDevice *self)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	if (priv->battery_level == 0)
+		return;
+	if (priv->battery_level >= fu_device_get_battery_threshold (self)) {
+		fu_device_uninhibit (self, "battery");
+		return;
+	}
+	fu_device_inhibit (self, "battery", "Battery level is too low");
```
superm1, line 2676: I wonder if you should prefix with the device name to disambiguate from host?

superm1 (approved):

libfwupdplugin/fu-device.c
hughsie: I don't think we need to, it's per device rather than per host. The _other_ system one would have a prefix however.

Issue #3089, opened 2021-03-31
clang CI failures


superm1:
I tried to take a look at the recent CI failures with `clang`.  It seems it's related to the environment flags getting set.
I can reproduce with the following
```
export CC="clang"
export CFLAGS="-g -O2 -ffile-prefix-map=/github/workspace=. -flto=auto -ffat-lto-objects -fstack-protector-strong -Wformat -Werror=format-security"
export LDFLAGS="-flto=auto -Wl,-z,relro"
export CPPFLAGS="-Wdate-time -D_FORTIFY_SOURCE=2"
```

However taking out link time optimization related things it seems to go away.
```
export CC="clang"
export CFLAGS="-g -O2 -ffile-prefix-map=/github/workspace=. -fstack-protector-strong -Wformat -Werror=format-security"
export LDFLAGS="-Wl,-z,relro"
export CPPFLAGS="-Wdate-time -D_FORTIFY_SOURCE=2"
```

Actually comparing the last successful build log (https://circleci.com/api/v1.1/project/github/fwupd/fwupd/21821/output/103/0?file=true&allocation-id=605c9e7193279b76180d8f79-0-build%2F2F82869C) It seems that those new `CFLAGS` and `LDFLAGS` getting introduced is exactly what caused the problems.

This was intentionally introduced in Ubuntu 21.04 (See https://wiki.ubuntu.com/ToolChain/LTO and https://www.phoronix.com/scan.php?page=news_item&px=Ubuntu-21.04-LTO-Packages)

In that Wiki page it's explicitly mentioned 
`GCC and clang use different LTO options. The injected flags use the GCC variants.`

So I think in our clang CI script we probably need to be doing something to not pickup the `gcc` based flags.

PR #3091, opened 2021-03-31
trivial: ci: disable LTO for clang (fixes: #3089)

superm1:
We'll see if this is enough to get it disabled.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
Thanks for looking into this.

PR #3092, opened 2021-03-31
flashrom: Update plugin to support multiple devices

djcampello:
This change moves many of the logic at the plugin level to the device
level in order to support multiple devices updatable via libflashrom.

### DIFF
--- a/plugins/flashrom/fu-flashrom-device.c
+++ b/plugins/flashrom/fu-flashrom-device.c
@@ -8,8 +8,14 @@
 
 #include "fu-flashrom-device.h"
 
+#include <libflashrom.h>
+
 struct _FuFlashromDevice {
-	FuDevice		 parent_instance;
+	FuDevice			 parent_instance;
+	gsize				 flash_size;
+	struct flashrom_flashctx	*flashctx;
+	struct flashrom_layout		*layout;
+	struct flashrom_programmer	*flashprog;
 };
 
 G_DEFINE_TYPE (FuFlashromDevice, fu_flashrom_device, FU_TYPE_DEVICE)
@@ -30,6 +36,16 @@ fu_flashrom_device_init (FuFlashromDevice *self)
 	fu_device_add_icon (FU_DEVICE (self), "computer");
 }
 
+static void
+fu_flashrom_device_finalize (GObject *object)
+{
+	FuFlashromDevice *self = FU_FLASHROM_DEVICE (object);
+	flashrom_layout_release (self->layout);
+	flashrom_programmer_shutdown (self->flashprog);
+	flashrom_flash_release (self->flashctx);
+	G_OBJECT_CLASS (fu_flashrom_device_parent_class)->finalize (object);
+}
+
 static gboolean
 fu_flashrom_device_set_quirk_kv (FuDevice *device,
 				 const gchar *key,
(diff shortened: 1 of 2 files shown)


plugins/flashrom/fu-flashrom-device.c
```diff
@@ -48,11 +72,206 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
...
+	gint rc;
+
+	if (flashrom_init (SELFCHECK_TRUE)) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "flashrom initialization error");
+		return FALSE;
+	}
+	flashrom_set_log_callback (fu_flashrom_device_debug_cb);
```
hughsie, line 119: isn't this global?

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -48,11 +72,206 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
...
+	return 0;
+}
+
+static gboolean
+fu_flashrom_device_setup (FuDevice *device, GError **error)
+{
+	FuFlashromDevicePrivate *priv = GET_PRIVATE (FU_FLASHROM_DEVICE (device));
+	gint rc;
+
+	if (flashrom_init (SELFCHECK_TRUE)) {
```
superm1, line 112: won't it be problematic to call `flashrom_init` multiple times?  I would think you want to once at plugin init, but not at a device level.

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -48,11 +72,206 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
...
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "flashrom initialization error");
+		return FALSE;
+	}
+	flashrom_set_log_callback (fu_flashrom_device_debug_cb);
+	if (flashrom_programmer_init (&priv->flashprog, "internal", NULL)) {
```
superm1, line 120: I think moving this to a device context you probably also want to be able to disambiguate the flasher to use too and not just internal, no?

plugins/flashrom/fu-flashrom-device.c
hughsie: Agree.

plugins/flashrom/fu-flashrom-device.c
djcampello: @sammc @bleungatchromium do you have any insights on this?

plugins/flashrom/fu-flashrom-device.c
djcampello: @sammc @bleungatchromium do you have any insights on this?

I tested this with more than one device (i.e., calling init at least twice) and everything worked as expected buy I'm not sure if it is the right thing to do.

plugins/flashrom/fu-flashrom-device.c
djcampello: I have that on a follow up change! yes!

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -48,11 +72,206 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
...
+				   FwupdInstallFlags flags,
+				   GError **error)
+{
+	FuFlashromDevicePrivate *priv = GET_PRIVATE (FU_FLASHROM_DEVICE (device));
+	gsize sz = 0;
+	gint rc;
+	const guint8 *buf;
+	g_autoptr(GBytes) blob_fw;
+
+	blob_fw = fu_firmware_get_bytes (firmware, error);
```
superm1, line 208: combine these lines to fix the CI failure.

djcampello: Thanks! Done

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -48,11 +72,206 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
...
+			  FwupdInstallFlags flags,
+			  GError **error)
+{
+	FuFlashromDevicePrivate *priv = GET_PRIVATE (FU_FLASHROM_DEVICE (device));
+	g_autofree gchar *firmware_orig = NULL;
+	g_autofree gchar *basename = NULL;
+
+	/* if the original firmware doesn't exist, grab it now */
+	basename = g_strdup_printf ("flashrom-%s.bin", fu_device_get_id (device));
+	firmware_orig = g_build_filename (FWUPD_LOCALSTATEDIR, "lib", "fwupd",
```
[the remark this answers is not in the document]
hughsie: IIRC, the location was originally chosen to be able to do an emergency manual recovery in case the flashing failed. I'm not really sure it's a cache.

plugins/flashrom/fu-flashrom-device.c
djcampello: PTAL. I moved it to fu_plugin_startup() so we can fail (init is void)

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -30,6 +42,16 @@ fu_flashrom_device_init (FuFlashromDevice *self)
 	fu_device_add_icon (FU_DEVICE (self), "computer");
 }
 
+static void
+fu_flashrom_device_finalize (GObject *object)
+{
+	FuFlashromDevicePrivate *priv = GET_PRIVATE (FU_FLASHROM_DEVICE (object));
+	G_OBJECT_CLASS (fu_flashrom_device_parent_class)->finalize (object);
```
hughsie, line 49: I think the G_OBJECT_CLASS() is normally last, no? I don't know if it matters, but I think it's destroying the parent object and so the FuFlashromDevicePrivate would be invalid.

djcampello: Done, Thanks!

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -8,11 +8,23 @@
...
-G_DEFINE_TYPE (FuFlashromDevice, fu_flashrom_device, FU_TYPE_DEVICE)
+typedef struct {
+	FuFlashromDevice		 parent_instance;
+	gsize				 flash_size;
+	struct flashrom_flashctx	*flashctx;
+	struct flashrom_layout		*layout;
+	struct flashrom_programmer	*flashprog;
+} FuFlashromDevicePrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE (FuFlashromDevice, fu_flashrom_device, FU_TYPE_DEVICE)
```
hughsie, line 25: Hangon, back a bit. You don't need to use G_DEFINE_TYPE_WITH_PRIVATE at all. The device is G_DECLARE_FINAL_TYPE, and so you can just stuff everything in the `G_DECLARE_FINAL_TYPE` above. Then you can use `self->flash_size` in the code rather than use `flash_size`

plugins/flashrom/fu-flashrom-device.c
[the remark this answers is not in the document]
superm1: Okay but we should still switch to the key rather than extra code building paths.

plugins/flashrom/fu-flashrom-device.c
djcampello: Fixed

plugins/flashrom/fu-flashrom-device.c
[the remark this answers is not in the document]
djcampello: Updated

hughsie (approved):

PR #3105, opened 2021-04-05
dfu-tool: Avoid runtime warning (overriding error)

djcb:
We were seeing:

```
,----
| (dfu-tool:1139827): GLib-WARNING **: 17:34:42.671: GError set over the top of a previous GError or uninitialized memory.
| This indicates a bug in someone's code. You must ensure an error is NULL before it's set.
| The overwriting error message was: no device matches for 1234:abcd
`----
``
This is because we were attempting to overwrite libgusb's error with one
for fwupd, and glib rightfully complains. Since we're likely more
interested in the fwupd error (since it's specifically
FWUPD_ERROR_NOT_FOUND), let's ignore the error from gusb.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/fu-dfu-tool.c
+++ b/plugins/dfu/fu-dfu-tool.c
@@ -222,11 +222,9 @@ fu_dfu_tool_get_default_device (FuDfuTool *self, GError **error)
 							    (guint16) pid,
 							    error);
 		if (usb_device == NULL) {
-			g_set_error (error,
-				     FWUPD_ERROR,
-				     FWUPD_ERROR_NOT_FOUND,
-				     "no device matches for %04x:%04x",
-				     (guint) vid, (guint) pid);
+			g_prefix_error (error,
+					"no device matches for %04x:%04x: ",
+					(guint) vid, (guint) pid);
 			return NULL;
 		}
 		device = fu_dfu_device_new (usb_device);

### CODE REVIEW

**Review by superm1 — CHANGES_REQUESTED** [review]
Although this fixes the problem, I don't think it's the correct solution.  Instead the below `g_set_error` should be a `g_prefix_error` with the content that was being set on top of the old error.

**Comment by djcb** [comment]
> Although this fixes the problem, I don't think it's the correct solution. Instead the below `g_set_error` should be a `g_prefix_error` with the content that was being set on top of the old error.

Thanks for the quick review! 

I considered `g_prefix_error`, but that would propogate the raw g-usb-error rather than the fwupd domain/error-code. If that's okay we can do that of course. Or alternatively, something a little more complicated to include the gubs error message in the new error. What do you think?

**Comment by superm1** [comment]
> > Although this fixes the problem, I don't think it's the correct solution. Instead the below `g_set_error` should be a `g_prefix_error` with the content that was being set on top of the old error.
> 
> Thanks for the quick review!
> 
> I considered `g_prefix_error`, but that would propogate the raw g-usb-error rather than the fwupd domain/error-code. If that's okay we can do that of course. Or alternatively, something a little more complicated to include the gubs error message in the new error. What do you think?

We do the prefixing and propagating of other domains/error codes elsewhere too.  (For example: MST https://github.com/fwupd/fwupd/blob/5c9b1fcc81942d6c3f3333a73faf8667c57507c8/plugins/synaptics-mst/fu-synaptics-mst-connection.c#L68)  I'd say this is OK to do here.  If we decide as a project to change that to make sure that all these codes map out I think we can pick a single choke point for the errors (such as library API or tool).

**Comment by djcb** [comment]
> ) I'd say this is OK to do here. If we decide as a project to change that to make sure that all these codes map out I think we can pick a single choke point for the errors (such as library API or tool).

Okay, I'll do that, but note that the error will then look something like:
 ```
no device matches for 0034:0035Failed to find device 0034:0035
```

superm1:
> no device matches for 0034:0035: Failed to find device 0034:0035

Maybe in that case the prefix should just be `no devices matches: `

djcb:
> > no device matches for 0034:0035: Failed to find device 0034:0035
> 
> Maybe in that case the prefix should just be `no devices matches: `

I guess it depends on the specific error libgusb returns, but `g_usb_context_find_by_vid_pid` does not document that (at least in my version). Or what would you recommend?

superm1 (approved):

superm1:
> I guess it depends on the specific error libgusb returns, but g_usb_context_find_by_vid_pid does not document that (at least in my version). Or what would you recommend?

In that case let's just leave the _potential_ double VID/PID print.  This is error scenario after all.  Too much information to help solve an error won't be a bad thing.

djcb:
Okay, thanks!

Issue #3106, opened 2021-04-05
needs-activation flag getting applied when irrelevant


superm1:
The WD19 connected to my system has gone through lots of upgrades and downgrades, but I noticed recently that it's claiming to be in a `needs-activation` state even though I've not done anything.

```$ sudo fwupdtool get-devices --plugins dell-dock --verbose
18:17:54:0614 FuDebug              Verbose debugging enabled (on console 1)
18:17:54:0619 FuMain               Failed to stop daemon: failed to find fwupd.service: GDBus.Error:org.freedesktop.systemd1.NoSuchUnit: Unit fwupd.service not loaded.
18:17:54:0620 FuConfig             loading config values from /usr/local/etc/fwupd/daemon.conf
18:17:54:0620 FuConfig             using autodetected max archive size 18.4 EB
18:17:54:0621 GLib-GIO             _g_io_module_get_default: Found default implementation local (GLocalVfs) for ‘gio-vfs’
18:17:54:0621 FuRemoteList         loading /usr/local/share/fwupd/metainfo
18:17:54:0621 XbSilo               attempting to load /var/local/cache/fwupd/metainfo.xmlb
18:17:54:0621 XbSilo               file: 97729952-2cf5-04c0-28e7-26d2026f0675, current:97729952-2cf5-04c0-28e7-26d2026f0675, cached: (null)
18:17:54:0621 XbSilo               loading silo with file contents
18:17:54:0622 FuRemoteList         loading remote from /usr/local/etc/fwupd/remotes.d/vendor.conf
18:17:54:0622 FuRemoteList         loading remote from /usr/local/etc/fwupd/remotes.d/lvfs-testing.conf
18:17:54:0622 FuRemoteList         loading remote from /usr/local/etc/fwupd/remotes.d/vendor-directory.conf
18:17:54:0622 FuRemoteList         loading remote from /usr/local/etc/fwupd/remotes.d/dell-esrt.conf
18:17:54:0622 FuRemoteList         loading remote from /usr/local/etc/fwupd/remotes.d/fwupd-tests.conf
18:17:54:0622 FuRemoteList         loading remote from /usr/local/etc/fwupd/remotes.d/lvfs.conf
18:17:54:0623 FuRemoteList         ordering lvfs-testing=lvfs+1
18:17:54:0623 FuRemoteList         ignoring unfound remote fwupd
[...]

PR #3107, opened 2021-04-05
fu-engine: Don't inherit activation flags for devices

superm1:
This was originally added for the WD19 which can actually advertise
it's own activation state from the hardware.

Fixes: #3106

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -198,6 +198,8 @@ fu_device_internal_flag_to_string (FuDeviceInternalFlags flag)
 		return "retry-open";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID)
 		return "replug-match-guid";
+	if (flag == FU_DEVICE_INTERNAL_FLAG_INHERIT_ACTIVATION)
+		return "inherit-activation";
 	return NULL;
 }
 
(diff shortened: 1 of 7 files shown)


superm1:
if this is the direction we go, it should probably backport to older branches too.

hughsie:
Wasn't this added for ATA devices too?

superm1:
The original commit message was specifically around the time we did dell-dock delayed activation (https://github.com/fwupd/fwupd/commit/96a0dd56063791dc03b1fcd738f6721b4420629d).

We did discuss it might have merit for ATA too.  The other way this can be spun is to make a new internal flag that to indicate whether or not to use history for activation inheritance and then opt ata in (or dell dock out).

hughsie:
>  The other way this can be spun is to make a new internal flag that to indicate whether or not to use history for activation inheritance and then opt ata in

I think this is exactly what we want.

superm1:
> > The other way this can be spun is to make a new internal flag that to indicate whether or not to use history for activation inheritance and then opt ata in
> 
> I think this is exactly what we want.

OK pushed that.  Let me know if you think this should still backport older, and i'll update the `Since` from 1.6.0 to something older (gotta see how far back this can come).

hughsie:
I think we want this in 1.5.x too. Thanks

hughsie (approved):

PR #3108, opened 2021-04-06
Wip/analogix

hardboprobot:
Type of pull request:
  - [x]  Fill out README.md with update protocol
  - [ ] Fill out README.md with any custom quirks and flags
  - [ ] Implement FuFirmware->write() and include at least one fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


plugins/analogix/fu-analogix-common.c
```diff
@@ -0,0 +1,231 @@
+/*
+ * Copyright (C) 2021 Xiaotian Cui <xtcui@analogixsemi.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+#include "fu-analogix-common.h"
+
+guint64
+hex_str_to_dec (const gchar *str, guint8 len)
```
superm1, line 9: this should be a static function declaration, and the symbol name should have a proper namespace for the plugin

plugins/analogix/fu-analogix-common.c
```diff
@@ -0,0 +1,231 @@
...
+		return 0;
+	src_str = g_strndup (str, len);
+
+	return g_ascii_strtoull (src_str, NULL, 16);
+}
+
+gboolean
+parse_fw_hex_file (const guint8 *fw_src, guint32 src_fw_size,
+		   AnxImgHeader *out_header, guint8 *out_binary)
+{
```
superm1, line 23: this function is only used by the `fu-analogix-firmware.c` as I can tell.  should it be moved there and then made into a static function?

hughsie: I think the plan needs to be "_subclass FuIhexFirmware and use the tokenization there_"

plugins/analogix/fu-analogix-common.c
```diff
@@ -0,0 +1,231 @@
...
+	if (len > 8 || str == NULL)
+		return 0;
+	src_str = g_strndup (str, len);
+
+	return g_ascii_strtoull (src_str, NULL, 16);
+}
+
+gboolean
+parse_fw_hex_file (const guint8 *fw_src, guint32 src_fw_size,
+		   AnxImgHeader *out_header, guint8 *out_binary)
```
superm1, line 22: I would wonder do you actually want to pass a `GError` into this function and set up the error on the failure scenarios?  Then you can tell _what_ the reason for failed hex parsing was.  Otherwise it's going to mean stepping through with a debugger in any failures in the future.

plugins/analogix/fu-analogix-common.c
```diff
@@ -0,0 +1,231 @@
...
+		 img_header->fw_payload_len);
+	g_debug ("Secure OCM TX start: 0x%x, len:0x%x",
+		 img_header->secure_tx_start_addr,
+		 img_header->secure_tx_payload_len);
+	g_debug ("Secure OCM RX start: 0x%x, len:0x%x",
+		 img_header->secure_rx_start_addr,
+		 img_header->secure_rx_payload_len);
+	g_debug ("Custom start: 0x%x, len:0x%x",
+		 img_header->custom_start_addr,
+		 img_header->custom_payload_len);
```
superm1, line 229: If this ends up moving in to `fu-analogix-firmware.c` then I would think all this information should get ended up added into the `to_string` output.

plugins/analogix/fu-analogix-device.c
```diff
@@ -0,0 +1,556 @@
...
+
+	/* check size */
+	if (out_len > 64) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_DATA,
+			     "output buffer too large");
+		return FALSE;
+	}
+	g_return_val_if_fail (out != NULL, FALSE);
```
superm1, line 103: shouldn't this be right at the start of the function?

plugins/analogix/fu-analogix-device.c
```diff
@@ -0,0 +1,556 @@
...
+			    guint16 val0code,
+			    guint16 index,
+			    guint8 *out,
+			    gsize out_len,
+			    GError **error)
+{
+	gsize actual_len = 0;
+
+	/* check size */
+	if (out_len > 64) {
```
superm1, line 96: everywhere this function is called this is a hardcoded value.  I think this should be a `g_return_val_if_false` indicating programming error if anything.

plugins/analogix/fu-analogix-device.c
hardboprobot: Indeed

plugins/analogix/fu-analogix-device.c
hardboprobot: Yeah, I'll find a saner way of checking that.

plugins/analogix/fu-analogix-device.c
```diff
@@ -0,0 +1,503 @@
...
+		if (status == UPDATE_STATUS_ERROR) {
+			g_set_error_literal (error,
+					     FWUPD_ERROR,
+					     FWUPD_ERROR_NOT_FOUND,
+					     "update status was error");
+			return FALSE;
+		}
+
+		/* assumed success */
+		if (status != UPDATE_STATUS_INVALID)
```
superm1, line 147: this seems wrong - there is a status for start and finish, won't you mark it prematurely done if you end up getting `*_START`?

plugins/analogix/fu-analogix-device.c
```diff
@@ -0,0 +1,503 @@
...
+}
+
+static void
+fu_analogix_device_init (FuAnalogixDevice *self)
+{
+	fu_device_add_protocol (FU_DEVICE (self), "com.analogix.bb");
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE);
+	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_PAIR);
+	fu_device_set_summary (FU_DEVICE (self), "Phoenix-Lite");
```
superm1, line 488: Shouldn't this be in the quirk file, as presumably the protocol might be used by multiple devices eventually?

hughsie: Removed.

plugins/analogix/fu-analogix-device.c
hughsie: I've renamed it `fu_analogix_device_get_update_status` and now check the value in the caller.

LuckyCandice:
@hughsie , can you help me check if I have the permission to commit the codes to this branch? Thank you!
$ git push origin wip/analogix
Username for 'https://github.com': LuckyCandice
Password for 'https://LuckyCandice@github.com':
remote: Permission to fwupd/fwupd.git denied to LuckyCandice.
**fatal: unable to access 'https://github.com/fwupd/fwupd.git/': The requested URL returned error: 403**

hughsie:
@LuckyCandice open .git/config and change:

    https://github.com/fwupd/fwupd.git

to:

    git@github.com:fwupd/fwupd.git

...then try pushing again.

LuckyCandice:
@hughsie ,thank you!
But still has error, can you help? Thank you!
$git push origin wip/analogix
ERROR: Permission to fwupd/fwupd.git denied to LuckyCandice.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

hughsie:
@LuckyCandice ahh, you have to accept the role by clicking on the link: https://github.com/fwupd/fwupd/invitations

LuckyCandice:
@hughsie , Yes, it works, thanks for your great help.

hughsie:
@LuckyCandice I'm coming to the realisation that the firmware format is actually standard IntelHex, with the different offsets in the binary output representing the different sections. Would that be accurate? In which case we could make it a *lot* simpler...

LuckyCandice:
@hughsie Yes, you are right. Our firmware file is Intel Hex format, and it may includes multiple sections or only one section.

hughsie:
@LuckyCandice is there an example I can download from somewhere? Thanks.

LuckyCandice:
@hughsie Sorry, we have no firmware file on any public server until now. I may upload it on LVFS in future days.

hughsie:
> Sorry, we have no firmware file on any public server until now

Okay, then can you email it to me then please? We can't approve something we can't test, and the firmware parsing code is really fragile at the moment.

hughsie:
@LuckyCandice I think you're going to have to describe the binary format and provide some kind of test binary, the current code is just too complicated and fragile to merge. The AnxImgHeader structure doesn't actually seem to be a header, and seems to be a kind of wrapper to export offsets and constants from the firmware to the device.

LuckyCandice:
> @LuckyCandice , @hughsie squashed all the commits into one a while ago, I think you smashed that by committing from an unsync'ed branch. Also, there are .swp files and source code with diff markup in the latest commits.

@hardboprobot I saw it, I made a mistake and restored it.

hughsie (approved):
Lets get this into master, and any fixups we can do in a new PR.

Issue #3084, opened 2021-03-30
Lenovo ThinkPad P1 Gen 2 is requested to get 10% of battery for updating bios but fails to flash at 11% because of the battery level


ErwanAliasr1:
**Describe the bug**
Lenovo ThinkPad P1 Gen 2 (20QUS58100) is said to need 10% of battery at least to get the bios updated.
At 11%, the firmware refuses to flash as battery is too low

**Steps to Reproduce**
- Lenovo ThinkPad P1 Gen 2 with 1.36 bios
- Put battery at 11%
- Install the bios 1.38
- Reboot
- Fw flash fail because of not enough battery

**Expected behavior**
The fwuptool does request a minimal battery level to get the fw being installed and host rebooted but this does seems in sync with the fw expectations.

**fwupd version information**
Please provide the version of the daemon and client.
```shell
client version:	1.3.11
compile-time dependency versions
	gusb:	0.3.4
	efivar:	37
daemon version:	1.3.11

```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
- via apt

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
$ fwupdmgr get-devices --show-all-devices
20QUS58100
?
??Thunderbolt Controller:
?     Device ID:           61b6e470639f64427b47ae7e101a778a2ecca334
?     Summary:             Unmatched performance for high-speed I/O
?     Current version:     59.00
?     Vendor:              Lenovo (TBT:0x0109)
?     GUIDs:               02ca1877-1054-5273-9f47-586b27f9abc3 ? TBT-01091813
?                          6075eaff-25d6-5a4f-91f8-6f565a644e72 ? TBT-01091813-controller0-0
?     Device Flags:        ? Internal device
?                          ? Updatable
?                          ? Requires AC power
?                          ? Device stages updates
?   
??Dell Universal Dock D6000:
?     Device ID:           62d5ed7ac4bfd0f40b57140818519eb35784fb8c
?     Current version:     31.27
?     Vendor:              DisplayLink (USB:0x17E9)
?     Serial Number:       1907028815
?     GUIDs:               5405d978-fd0d-5890-be13-593f009cc292 ? USB\VID_17E9&PID_6006&REV_3127
?                          1b3b6e05-bd4d-5880-b68d-c7c00b6681f8 ? USB\VID_17E9&PID_6006
?     Device Flags:        ? Updatable
?   
[...]

PR #3109, opened 2021-04-06
Set the system battery state and level on the FuContext shared state

hughsie:
This allows plugins to set the battery power state of the *machine* which means
we can automatically inhibit devices with FWUPD_DEVICE_FLAG_REQUIRE_AC set.

It also allows to set the BatteryThreshold to 25% for Lenovo hardware, and we
can override other vendors with further quirks as required.

Fixes https://github.com/fwupd/fwupd/issues/3084

### DIFF
--- a/libfwupdplugin/fu-context.c
+++ b/libfwupdplugin/fu-context.c
@@ -8,7 +8,6 @@
 
 #include "config.h"
 
-#include "fu-common.h"
 #include "fu-context-private.h"
 #include "fu-hwids.h"
 #include "fu-smbios-private.h"
@@ -29,11 +28,22 @@ typedef struct {
 	GHashTable		*compile_versions;
 	GPtrArray		*udev_subsystems;
 	GHashTable		*firmware_gtypes;
+	FuBatteryState		 battery_state;
+	guint			 battery_level;
+	guint			 battery_threshold;
 } FuContextPrivate;
 
 enum {
 	SIGNAL_SECURITY_CHANGED,
 	SIGNAL_LAST
+};
+
+enum {
+	PROP_0,
+	PROP_BATTERY_STATE,
+	PROP_BATTERY_LEVEL,
+	PROP_BATTERY_THRESHOLD,
+	PROP_LAST
 };
 
 static guint signals[SIGNAL_LAST] = { 0 };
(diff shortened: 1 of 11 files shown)


libfwupdplugin/fu-context.c
```diff
@@ -546,6 +557,184 @@ fu_context_load_quirks (FuContext *self, FuQuirksLoadFlags flags, GError **error
...
+		return "charging";
+	if (battery_state == FU_CONTEXT_BATTERY_STATE_DISCHARGING)
+		return "discharging";
+	if (battery_state == FU_CONTEXT_BATTERY_STATE_EMPTY)
+		return "empty";
+	if (battery_state == FU_CONTEXT_BATTERY_STATE_FULLY_CHARGED)
+		return "fully-charged";
+	return NULL;
+}
+
```
superm1, line 593: spurious new line

libfwupdplugin/fu-context.c
```diff
@@ -546,6 +557,184 @@ fu_context_load_quirks (FuContext *self, FuQuirksLoadFlags flags, GError **error
...
+	g_object_notify (G_OBJECT (self), "battery-state");
+}
+
+/**
+ * fu_context_get_battery_level:
+ * @self: A #FuContext
+ *
+ * Gets the system battery level in percent.
+ *
+ * Returns: percentage value, or `0` for unknown
```
superm1, line 623: I actually wonder if this type and return should be a signed integer.  0 can mean a lot of things:
* Dead battery
* No battery
* One battery is 0 and another is something else?

src/fu-engine.c
```diff
@@ -225,10 +225,32 @@ fu_engine_watch_device (FuEngine *self, FuDevice *device)
 			  G_CALLBACK (fu_engine_status_notify_cb), self);
 }
 
+static void
+fu_engine_ensure_device_battery_inhibit (FuEngine *self, FuDevice *device)
+{
+	if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_REQUIRE_AC) &&
+	    (fu_context_get_battery_state (self->ctx) == FU_CONTEXT_BATTERY_STATE_DISCHARGING ||
```
superm1, line 232: Not sure I agree with this.  You can have an AC adapter plugged in that doesn't provide enough power so the system is effectively discharging, but really slowly.

plugins/upower/fu-plugin-upower.c
```diff
@@ -27,128 +25,131 @@ void
...
 		return TRUE;
+	if (flags & FWUPD_INSTALL_FLAG_IGNORE_POWER)
+		return TRUE;
 
-	/* determine if operating on AC or battery */
-	if (fu_plugin_upower_check_on_battery (plugin) &&
-	    (flags & FWUPD_INSTALL_FLAG_IGNORE_POWER) == 0) {
+	/* not charging */
+	if (fu_context_get_battery_state (ctx) == FU_CONTEXT_BATTERY_STATE_DISCHARGING ||
+	    fu_context_get_battery_state (ctx) == FU_CONTEXT_BATTERY_STATE_EMPTY) {
```
superm1, line 152: can't battery state also be empty if the system is a desktop or doesn't support a battery?

superm1 (commented):
what happens when there are two batteries on the system?  does upower just return info on the first one?  Or an average of the two?

hughsie:
> Or an average of the two?

Yes, it's the Display device that is the "composite" device used for system power notifications, e.g. in the shell.

libfwupdplugin/fu-context.c
hughsie: What about if we define a `FU_DEVICE_BATTERY_LEVEL_INVALID` as `G_MAXUINT32` or something?

hughsie: or maybe `101` as another option

src/fu-engine.c
hughsie: From a UPower point of view if the AC adapter is connected it's either _charging_ or _charged_. I don't think we handle the _plugged in but discharging_ case at all in the UI.

plugins/upower/fu-plugin-upower.c
hughsie: If there's no battery then the state would be `FU_CONTEXT_BATTERY_STATE_UNKNOWN`.

libfwupdplugin/fu-context.c
superm1: Yes, definitely should define something for this meaning.  No strong feeling across those 3, you pick.

libfwupdplugin/fu-context.c
hughsie: Fixed.

libfwupdplugin/fu-context.c
hughsie: Okay, fixed, let me know what you think :)

src/fu-engine.c
superm1: FWIW Windows doesn't really handle this well either.  I think it's most likely fine, but wanted to mention the terms were not strictly true :)

superm1 (dismissed):

src/fu-engine.c
hughsie: Out of morbid curiosity, is there any way to detect this case from the OS without just noticing that the percentage is actually going down over time?

PR #3119, opened 2021-04-08
Flashrom: add lspcon-i2c-spi support

djcampello:
Adds a new device type to the flashrom plugin for devices requiring the lspcon-i2c-spi programmer.

### DIFF
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -32,3 +32,14 @@ VersionFormat=number
 [0ee5867c-93f0-5fb4-adf1-9d686ea1183a]
 Branch=coreboot
 VersionFormat=number
+
+# match all devices with this udev subsystem
+[I2C]
+Plugin = flashrom
+
+# Parade PS175
+[1AF80175:00]
+Plugin=flashrom
+FlashromProgrammer=lspcon_i2c_spi
+Name=PS175
+Vendor=Parade
--- a/plugins/flashrom/fu-flashrom-device.c
+++ b/plugins/flashrom/fu-flashrom-device.c
@@ -19,7 +19,7 @@ typedef struct {
 	struct flashrom_programmer	*flashprog;
 } FuFlashromDevicePrivate;
 
-G_DEFINE_TYPE_WITH_PRIVATE (FuFlashromDevice, fu_flashrom_device, FU_TYPE_DEVICE)
+G_DEFINE_TYPE_WITH_PRIVATE (FuFlashromDevice, fu_flashrom_device, FU_TYPE_UDEV_DEVICE)
 
 #define GET_PRIVATE(o) (fu_flashrom_device_get_instance_private (o))
 
@@ -103,9 +103,25 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
 }
 
 static gboolean
-fu_flashrom_device_prepare (FuDevice *device,
-			    FwupdInstallFlags flags,
-			    GError **error)
+fu_flashrom_device_probe (FuDevice *device, GError **error)
+{
+	const gchar *dev_name = NULL;
+	g_autofree gchar *path = NULL;
+
+	/* FuUdevDevice->probe */
+	if (!FU_DEVICE_CLASS (fu_flashrom_device_parent_class)->probe (device, error))
+		return FALSE;
+
+	path = g_strdup_printf ("DEVNAME=%s",
+				fu_udev_device_get_sysfs_path (FU_UDEV_DEVICE (device)));
+	fu_device_set_physical_id (device, path);
+	dev_name = fu_udev_device_get_sysfs_attr (FU_UDEV_DEVICE (device), "name", error);
+	fu_device_add_instance_id_full (device, dev_name, FU_DEVICE_INSTANCE_FLAG_ONLY_QUIRKS);
+	return TRUE;
+}
+
+static gboolean
+fu_flashrom_device_open (FuDevice *device, GError **error)
 {
 	FuFlashromDevicePrivate *priv = GET_PRIVATE (FU_FLASHROM_DEVICE (device));
 	gint rc;
@@ -161,9 +177,7 @@ fu_flashrom_device_prepare (FuDevice *device,
 }
 
 static gboolean
-fu_flashrom_device_cleanup (FuDevice *device,
-			    FwupdInstallFlags flags,
-			    GError **error)
+fu_flashrom_device_close (FuDevice *device, GError **error)
 {
 	FuFlashromDevicePrivate *priv = GET_PRIVATE (FU_FLASHROM_DEVICE (device));
 	flashrom_flash_release (priv->flashctx);
@@ -178,6 +192,7 @@ fu_flashrom_device_class_init (FuFlashromDeviceClass *klass)
 	FuDeviceClass *klass_device = FU_DEVICE_CLASS (klass);
 	object_class->finalize = fu_flashrom_device_finalize;
 	klass_device->set_quirk_kv = fu_flashrom_device_set_quirk_kv;
-	klass_device->prepare = fu_flashrom_device_prepare;
-	klass_device->cleanup = fu_flashrom_device_cleanup;
+	klass_device->probe = fu_flashrom_device_probe;
+	klass_device->open = fu_flashrom_device_open;
+	klass_device->close = fu_flashrom_device_close;
 }
--- a/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
+++ b/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
@@ -175,6 +175,96 @@ fu_flashrom_lspcon_i2c_spi_device_setup (FuDevice *device, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_flashrom_lspcon_i2c_spi_device_write_firmware (FuDevice *device,
+						  FuFirmware *firmware,
+						  FwupdInstallFlags flags,
+						  GError **error)
+{
+	FuFlashromDevice *parent = FU_FLASHROM_DEVICE (device);
+	FuFlashromLspconI2cSpiDevice *self = FU_FLASHROM_LSPCON_I2C_SPI_DEVICE (device);
+	struct flashrom_flashctx *flashctx = fu_flashrom_device_get_flashctx (parent);
+	gsize flash_size = fu_flashrom_device_get_flash_size (parent);
+	g_autofree guint8 *newcontents = g_malloc0 (flash_size);
+	const guint8 target_partition = self->active_partition == 1 ? 2 : 1;
+	const gsize region_size = entries[target_partition].end -
+				  entries[target_partition].start + 1;
+	gsize sz = 0;
+	gint rc;
+	const guint8 *buf;
+	g_autoptr(GBytes) blob_fw = fu_firmware_get_bytes (firmware, error);
+	if (blob_fw == NULL)
+		return FALSE;
+
+	buf = g_bytes_get_data (blob_fw, &sz);
+
+	if (sz != region_size) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "invalid image size 0x%x, expected 0x%x",
+			     (guint) sz, (guint) flash_size);
+		return FALSE;
+	}
+	if (!fu_memcpy_safe (newcontents, flash_size,
+			     entries[target_partition].start, buf, sz, 0,
+			     region_size, error))
+		return FALSE;
+
+	flashrom_flag_set (flashctx, FLASHROM_FLAG_VERIFY_AFTER_WRITE, TRUE);
+	flashrom_layout_set (flashctx, &layout);
+
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_WRITE);
+	fu_device_set_progress (device, 0); /* urgh */
+
+	/* only include target_partition on layout */
+	entries[0].included = FALSE;
+	entries[3].included = FALSE;
+	entries[4].included = FALSE;
+	entries[self->active_partition].included = FALSE;
+	entries[target_partition].included = TRUE;
+
+	/* write target_partition */
+	rc = flashrom_image_write (flashctx, (void *) newcontents, flash_size,
+				   NULL /* refbuffer */);
+	if (rc != 0) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_WRITE,
+			     "image write failed, err=%i", rc);
+		return FALSE;
+	}
+
+	/* only include flag area on layout */
+	entries[0].included = TRUE;
+	entries[target_partition].included = FALSE;
+
+	/* Flag area is header bytes (0x55, 0xAA) followed by the bank ID to
+	 * boot from (1 or 2) and the two's complement inverse of that bank ID
+	 * (0 or 0xFF). We only write bytes 2 and 3, assuming the header is
+	 * already valid. */
+	newcontents[2] = (gint8) target_partition;
+	newcontents[3] = (gint8) -target_partition + 1;
+
+	/* libflashrom fails to verify a write of a layout smaller than the
+	 * block size. */
+	flashrom_flag_set (flashctx, FLASHROM_FLAG_VERIFY_AFTER_WRITE, FALSE);
+
+	/* write target_partition */
+	rc = flashrom_image_write (flashctx, (void *) newcontents, flash_size,
+				   NULL /* refbuffer */);
+	if (rc != 0) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_WRITE,
+			     "flag write failed, err=%i", rc);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
 static void
 fu_flashrom_lspcon_i2c_spi_device_class_init (FuFlashromLspconI2cSpiDeviceClass *klass)
 {
@@ -182,4 +272,5 @@ fu_flashrom_lspcon_i2c_spi_device_class_init (FuFlashromLspconI2cSpiDeviceClass 
 	klass_device->probe = fu_flashrom_lspcon_i2c_spi_device_probe;
 	klass_device->open = fu_flashrom_lspcon_i2c_spi_device_open;
 	klass_device->setup = fu_flashrom_lspcon_i2c_spi_device_setup;
+	klass_device->write_firmware = fu_flashrom_lspcon_i2c_spi_device_write_firmware;
 }
--- a/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
+++ b/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2021 Daniel Campello <campello@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-flashrom-device.h"
+#include "fu-flashrom-lspcon-i2c-spi-device.h"
+
+#include <libflashrom.h>
+
+#define I2C_PATH_REGEX "/i2c-([0-9]+)/"
+#define HID_LENGTH 8
+#define DEVICE_GUID_FORMAT "FLASHROM-LSPCON-I2C-SPI\\VEN_%s&DEV_%s"
+
+struct _FuFlashromLspconI2cSpiDevice {
+	FuFlashromDevice	  parent_instance;
+	gint			  bus_number;
+};
+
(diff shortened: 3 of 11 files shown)


plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+
+	path = fu_udev_device_get_sysfs_path (FU_UDEV_DEVICE (device));
+	self->bus_number = fu_flashrom_lspcon_i2c_spi_device_get_bus_number_from_path (path, error);
+	if (self->bus_number == -1) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "%s failed to get bus number for " \
+			     "device under %s.",
+			     __func__,
```
hughsie, line 153: not the __func__ thing please

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+	return TRUE;
+}
+
+static gint
+fu_flashrom_lspcon_i2c_spi_device_get_bus_number_from_path (const gchar *path,
+							    GError **error)
+{
+	gint bus_number = -1;
+	GRegex *regex = NULL;
+	GMatchInfo *info;
```
hughsie, line 104: both g_autoptr()

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+fu_flashrom_lspcon_i2c_spi_device_prepare (FuDevice *device,
+					   FwupdInstallFlags flags,
+					   GError **error)
+{
+	FuFlashromDevice *parent = FU_FLASHROM_DEVICE (device);
+	FuFlashromLspconI2cSpiDevice *self = FU_FLASHROM_LSPCON_I2C_SPI_DEVICE (device);
+
+	/* flashrom_programmer_init() mutates the programmer_args string. */
+	fu_flashrom_device_set_programmer_args (parent,
+						g_strdup_printf	("bus=%d", self->bus_number));
```
hughsie, line 86: no g_strdup here, you'll need to use a g_autofree temp I'm afraid.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+};
+
+G_DEFINE_TYPE (FuFlashromLspconI2cSpiDevice, fu_flashrom_lspcon_i2c_spi_device,
+	       FU_TYPE_FLASHROM_DEVICE)
+
+struct romentry {
+	uint32_t start;
+	uint32_t end;
+	bool included;
+	const char *name;
```
hughsie, line 31: gchar, and right aligned please

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+	guint8				 active_partition;
+};
+
+G_DEFINE_TYPE (FuFlashromLspconI2cSpiDevice, fu_flashrom_lspcon_i2c_spi_device,
+	       FU_TYPE_FLASHROM_DEVICE)
+
+struct romentry {
+	uint32_t start;
+	uint32_t end;
+	bool included;
```
hughsie, line 30: gboolean

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+	FuFlashromDevice		 parent_instance;
+	gint				 bus_number;
+	guint8				 active_partition;
+};
+
+G_DEFINE_TYPE (FuFlashromLspconI2cSpiDevice, fu_flashrom_lspcon_i2c_spi_device,
+	       FU_TYPE_FLASHROM_DEVICE)
+
+struct romentry {
+	uint32_t start;
```
hughsie, line 28: guint32

plugins/flashrom/flashrom.quirk
```diff
@@ -32,3 +32,15 @@ VersionFormat=number
...
+# match all devices with this udev subsystem
+[I2C]
+Plugin = flashrom
+
+# Parade PS175
+[1AF80175:00]
+Plugin=flashrom
+FlashromProgrammer=lspcon_i2c_spi
+Name=PS175
+Protocol=com.parade.ps175
```
hughsie, line 45: you sure? that's not a protocol on the LVFS.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -39,7 +40,8 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "linux_lockdown");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_CONFLICTS, "coreboot"); /* obsoleted */
-	fu_plugin_add_flag (plugin, FWUPD_PLUGIN_FLAG_REQUIRE_HWID);
+	fu_plugin_set_device_gtype (plugin, FU_TYPE_FLASHROM_LSPCON_I2C_SPI_DEVICE);
```
superm1, line 43: I think if you're adding this code, don't we also need one for internal?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+			     "device under %s.",
+			     __func__,
+			     path);
+		return FALSE;
+	}
+
+	hw_id = fu_udev_device_get_sysfs_attr (FU_UDEV_DEVICE (device), "name", error);
+	vid = g_strndup (hw_id, HID_LENGTH / 2);
+	pid = g_strndup (&hw_id[HID_LENGTH / 2], HID_LENGTH / 2);
+	fu_device_add_vendor_id (device, g_strdup_printf ("I2C:%s", vid));
```
hughsie, line 161: not a `DMI` ID? isn't this system firmware?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+	g_match_info_free (info);
+	g_regex_unref (regex);
+
+	return bus_number;
+}
+
+static gboolean
+fu_flashrom_lspcon_i2c_spi_device_setup (FuDevice *device, GError **error)
+{
+	FuFlashromDevice *parent = FU_FLASHROM_DEVICE (device);
```
hughsie, line 124: Why parent? I don't see any children.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+
+	if (self->active_partition != 1 && self->active_partition != 2) {
+		g_set_error_literal (error, FWUPD_ERROR, FWUPD_ERROR_READ,
+				     "failed to read active partition");
+		return FALSE;
+	}
+
+	/* find the current version for the active partition */
+	addr = entries[self->active_partition].start + VERSION_ADDR_OFFSET;
+	fu_device_set_version (device, g_strdup_printf ("%d.%d", contents[addr],
```
hughsie, line 205: need g_autofree temp var I'm afraid/

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+							contents[addr + 2]));
+
+	/* FuFlashromDevice->cleanup */
+	if (!FU_DEVICE_CLASS
+	    (fu_flashrom_lspcon_i2c_spi_device_parent_class)->cleanup (device,
```
hughsie, line 210: cleanup? In setup()?

plugins/flashrom/fu-plugin-flashrom.c
djcampello: Internal is called for coldplug method on the plugin

hughsie: It's a bit more tricky than that, it needs to be g_type_ensure if it's being created from a quirk. I think fu_plugin_**set**_device_gtype means we can't create internal ones anymore too.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "invalid programmer");
+		return FALSE;
+	}
+
+	path = fu_udev_device_get_sysfs_path (FU_UDEV_DEVICE (device));
+	self->bus_number = fu_flashrom_lspcon_i2c_spi_device_get_bus_number_from_path (path, error);
+	if (self->bus_number == -1) {
+		g_set_error (error,
```
superm1, line 148: isn't error going to be set by `fu_flashrom_lspcon_i2c_spi_device_get_bus_number_from_path`

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: I know it is ugly... I may want to create another function on flashrom_device that can be called here and on flashrom_device->cleanup... Basically we need to call the cleanup (i.e., shutting down programmer, etc) every time we finish accessing the flash because libflashrom does not support more than one context (i.e., a internal programmer context and a lspcon_i2c_spi programmer context) at a time...

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "%s failed to get bus number for " \
+			     "device under %s.",
+			     __func__,
+			     path);
+		return FALSE;
+	}
+
+	hw_id = fu_udev_device_get_sysfs_attr (FU_UDEV_DEVICE (device), "name", error);
```
superm1, line 158: you should check for `hw_id` is `NULL` here

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+
+	guid = fwupd_guid_hash_string (g_strdup_printf (DEVICE_GUID_FORMAT, vid, pid));
+	if (guid == NULL) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "%s failed to generate guid.",
+			     __func__);
+		return FALSE;
+	}
```
superm1, line 171: how does this scenario actually happen?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
hughsie: > I may want to create another function on flashrom_device

Is that not FuDevice->open() and ->close()?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+		return FALSE;
+
+	/* set up flashrom layout */
+	flashctx = fu_flashrom_device_get_flashctx (parent);
+	flashrom_layout_set (flashctx, &layout);
+
+	/* read the current flash contents */
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_READ);
+	flash_size = fu_flashrom_device_get_flash_size (parent);
+	contents = g_malloc0 (flash_size);
```
superm1, line 187: I2C isn't that fastest bus.
it seems like you really don't need to read the whole flash, would it be faster to only read up to what data you really need?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+	addr = entries[self->active_partition].start + VERSION_ADDR_OFFSET;
+	fu_device_set_version (device, g_strdup_printf ("%d.%d", contents[addr],
+							contents[addr + 2]));
+
+	/* FuFlashromDevice->cleanup */
+	if (!FU_DEVICE_CLASS
+	    (fu_flashrom_lspcon_i2c_spi_device_parent_class)->cleanup (device,
+								       FWUPD_INSTALL_FLAG_NONE,
+								       error))
+		return FALSE;
```
superm1, line 213: does this maybe want to be in a `close` vfunc?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: If I use g_autoptr() do I need the g_match_info_free and g_regex_unref below?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,315 @@
...
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "invalid image size 0x%x, expected 0x%x",
+			     (guint) sz, (guint) flash_size);
+		return FALSE;
+	}
+	if (!fu_memcpy_safe (newcontents, flash_size,
+			     entries[target_partition].start, buf, sz, 0,
+			     region_size, error)) {
+		return FALSE;
+	}
```
superm1, line 253: no need for braces on a one line return

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: not in every case, but I can move this above

plugins/flashrom/fu-plugin-flashrom.c
djcampello: In my testing both device types coexist...

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: The amount of data read is limited by the layout set above. I guess I could add some additional regions with only the version area and read those... Let me experiment with that..

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
hughsie: Nope, it does those automatically when the variable goes out of scope.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: This is an internal component firmware. system firmware should be handled by fu_flashrom_internal_device

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
superm1: I mean don't bust your butt over it especially if we're talking milliseconds of difference in probe time.  I just want to avoid a 30s probe that could be a 3s probe.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
superm1: I think it would be better to follow the style of having -1 mean error and set error properly in that function.

hughsie: I'd much rather have the function return a gboolean TRUE/FALSE and then pass &self->bus_number as an out parameter...

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: Added additional entries on layout to just read the version section.

plugins/flashrom/flashrom.quirk
```diff
@@ -32,3 +32,14 @@ VersionFormat=number
...
+# match all devices with this udev subsystem
+[I2C]
+Plugin = flashrom
+
+# Parade PS175
+[1AF80175:00]
+Plugin=flashrom
+FlashromProgrammer=lspcon_i2c_spi
+Name=PS175
+Vendor=Parade
```
superm1, line 45: It's not clear to me  - how is this I2C device actually getting probed?  Is there some kernel driver that is exporting that it exists?  Maybe some other documentation / links in README.md would be useful

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -103,9 +103,24 @@ fu_flashrom_device_set_quirk_kv (FuDevice *device,
...
+{
+	const gchar *dev_name = NULL;
+	const gchar *path = NULL;
+
+	/* FuUdevDevice->probe */
+	if (!FU_DEVICE_CLASS (fu_flashrom_device_parent_class)->probe (device, error))
+		return FALSE;
+
+	path = fu_udev_device_get_sysfs_path (FU_UDEV_DEVICE (device));
+	fu_device_set_physical_id (device, g_strdup_printf ("DEVNAME=%s", path));
```
superm1, line 116: memory leak, this should be `g_autofree`'d

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -0,0 +1,273 @@
...
+#include "config.h"
+
+#include "fu-flashrom-device.h"
+#include "fu-flashrom-lspcon-i2c-spi-device.h"
+
+#include <libflashrom.h>
+
+#define I2C_PATH_REGEX "/i2c-([0-9]+)/"
+#define HID_LENGTH 8
+#define DEVICE_GUID_FORMAT "FLASHROM-LSPCON-I2C-SPI\\VEN_%s&DEV_%s"
```
superm1, line 16: this should come into README.md too.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
djcampello: I moved the logic into ->probe() as it was short enough..

plugins/flashrom/flashrom.quirk
djcampello: To be honest I am not sure. Here is the sysfs device on my system and its modalias:
```
# cat /sys/devices/pci0000:00/0000:00:15.2/i2c_designware.1/i2c-7/i2c-1AF80175:00/modalias
acpi:1AF80175:
```

superm1: OK, so it's programmatically described in the ACPI tables as being something hanging off that HID controller.  That makes somewhat more sense.

superm1:
> Do you know how can I debug the CI failure on fedora? It looks like a SEGFAULT while exiting?

I'd guess if it's failing while exiting there is a memory problem somewhere that isn't being cleaned up properly, or not in the correct order.  Trying running it with `valgrind`

plugins/flashrom/flashrom.quirk
superm1: I guess the one thing I want to make sure though, that ACPI ID - it's only ever used with Parade?  Like they've reserved that namespace.  If not, please make sure you get the firmware fixed to a namespace only for Parade.

djcampello: That id is a combo of vid pid (4 characters each) which I understand is specific to Parade.

superm1:
> > > Do you know how can I debug the CI failure on fedora? It looks like a SEGFAULT while exiting?
> > 
> > 
> > I'd guess if it's failing while exiting there is a memory problem somewhere that isn't being cleaned up properly, or not in the correct order. Trying running it with `valgrind`
> 
> What I don't get is why it only happens on the fedora ci, cannot get the same failure on my system so I don't know how to repro.

More aggressively configured compiler?  We run CI across a variety of common distros for this reason that sometimes you don't catch problems just running in one place.

In the logs I noticed for CI this:
` 90008 is not a valid triplet (guessed number)`
Which I guess you should figure out where is coming from, that's output as a warning, which I didn't expect, and maybe this is from error path.

superm1:
Something else sticking out to me - it's actually coldplugging in CI a BIOS device - which we shouldn't be seeing...

```
19:24:04:0989 FuDevice             using dbee8bd3b1ae0316ad143336155651eedb495a0e for flashrom:bios
19:24:04:0989 FuPlugin             emit added from flashrom: dbee8bd3b1ae0316ad143336155651eedb495a0e
```

djcampello:
I think that with the last push all comments have been addressed. :)

superm1 (approved):

djcampello:
Thank you for the review!

PR #3120, opened 2021-04-08
Check with CI that all python code was formatted with black

superm1:
My thought here is that if something intentionally really doesn't match we can make an exception at that time.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/fedora.sh
+++ b/contrib/ci/fedora.sh
@@ -26,6 +26,12 @@ if ! find . -name '*.sh' | xargs shellcheck --severity=error -e SC2068; then
     exit 1
 fi
 
+# check python code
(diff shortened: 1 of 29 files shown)


hughsie:
It's lots of churn, but it's probably the right thing to do.

hughsie (approved):

superm1:
I've seen in some projects in `CONTRIBUTING` where they ask you to run a setup script that creates some pre-commit hooks in your local repo.  Then that script will run all the style checks that they also run in CI to help prevent pointless churn.  I feel like this and the `TRUE`/`FALSE` check would be pretty valuable to move into such an environment.

PR #3124, opened 2021-04-09
Write BMP data directly without using PIL

hughsie:
This allows us to drop a build-time dep.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -8,7 +8,6 @@ extraction:
       - python3-gi
       - libarchive-tools
       - libcogl-pango-dev
-      - python3-pil
       - python3-pip
       - python3-setuptools
       - python3-wheel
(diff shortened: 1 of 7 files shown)


hughsie:
Some background: fwupd is the only installed-by-default package in RHEL that needed python3-pillow and I really don't want to have to be responsible an additonal security-critical python library when we really don't need to.

hughsie:
I assume the ABI checker failed as we removed a dep? I think that's fine to ignore if that's the case.

superm1 (approved):

hroncok:
> fwupd is the only installed-by-default package in RHEL that needed python3-pillow

Not "installed-by-default". The only package in RHEL 9.

PR #3127, opened 2021-04-09
Add fu_bytes_get_data_safe() to check for zero sized data

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3122

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -3307,6 +3307,38 @@ fu_battery_state_to_string (FuBatteryState battery_state)
 }
 
 /**
+ * fu_bytes_get_data_safe:
+ * @bytes: a #GBytes
+ * @bufsz: (out) (optional): location to return size of byte data
+ * @error: A #GError or %NULL
+ *
+ * Get the byte data in the #GBytes. This data should not be modified.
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -3306,6 +3306,38 @@ fu_battery_state_to_string (FuBatteryState battery_state)
...
+ **/
+const guint8 *
+fu_bytes_get_data_safe (GBytes *bytes, gsize *bufsz, GError **error)
+{
+	const guint8 *buf = g_bytes_get_data (bytes, bufsz);
+	if (buf == NULL) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_INVALID_DATA,
+			     "invalid data");
```
superm1, line 3335: is it worth including `(size: %s)` in the error message perhaps?

hughsie: won't size always be zero?

superm1 (approved):

libfwupdplugin/fu-common.c
superm1: Ah right never mind

PR #3129, opened 2021-04-10
Add more things to the precommit list

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,9 +1,13 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v3.4.0
     hooks:
     -   id: no-commit-to-branch
         args: [--branch, master, --pattern, 1_.*_X]
+    -   id: check-added-large-files
+    -   id: check-byte-order-marker
+    -   id: check-executables-have-shebangs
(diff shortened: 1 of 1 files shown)


superm1 (approved):

superm1:
The one I really want is something like uncrustify or similar linters, but that's a ton of work to sort out

hughsie:
> The one I really want is something like uncrustify

Right, I agree. The problem is I use "style checking" as a way of **really** looking at the new code. i.e. thinking "_would I be happy if I wrote it this way_" so that we really do reviews, rather than just skimming and approving. Maybe that's insane. One thing I'm not super keen on is reformatting the entire codebase as then we loose all ability to do `git blame` which I for one find really useful.

superm1:
> > The one I really want is something like uncrustify
> 
> Right, I agree. The problem is I use "style checking" as a way of **really** looking at the new code. i.e. thinking "_would I be happy if I wrote it this way_" so that we really do reviews, rather than just skimming and approving. Maybe that's insane. One thing I'm not super keen on is reformatting the entire codebase as then we loose all ability to do `git blame` which I for one find really useful.

Yeah if we do end up adding style checking pre commit my thought is it should only run on the diff. That might be even harder though.

Issue #3019, opened 2021-03-11
Establish locking mechanism to prevent two simultaneous instances of fwupdtool


djcampello:
**Is your feature request related to a problem? Please describe.**
If two asynchronous instances of fwupdtool try to access the database these leads to issues

**Describe the solution you'd like**
Simple file locking mechanism should suffice

**Describe alternatives you've considered**
File locking external to fwupd wrapping fwupdtool invocations

hughsie:
>  If two asynchronous instances of fwupdtool try to access the database these leads to issues

What are you trying to do? The database was really designed for multiple readers and one writer (the daemon) -- although I don't see a problem with two things trying to write to sqlite at the same time.

>  File locking external to fwupd wrapping fwupdtool invocations

Hmm, I think the solution here is to use the daemon which will serialize access to both the database, and more importantly **the hardware** too.

djcampello:
I had some SQLite errors showing up when a `fwupdtool get-devices` executed while a `fwupdtool update` was also executing. My guess is that the later was writing the database to mark the device as "pending for activation" when the former executed. 

I agree with the daemon solving the problem. While this is not the case for us, I am looking to have a temporary solution to serialize multiple pending fwupdtool update executions (all triggered from upstart). I was thinking on a file lock that I could wait on.

hughsie:
> when a fwupdtool get-devices executed while a fwupdtool update was also executing

I really don't think that's safe at all. Probing the hardware while it's being updated isn't tested in any way and is probably a Bad Idea™.

> I was thinking on a file lock

I think we can do something better than that. Why don't we make fwupdtool also "own" the org.freedesktop.fwupd DBus interface for the entire time it's running? The benefits is that it would ensure the daemon quits when safe, and it would wait for other fwupdtool instances to quit before continuing to the user initiated action.

djcampello:
If a fwupdtool instance is running and the user launches another one, will it fail or hang until it can acquire the DBus interface?

hughsie:
> will it fail or hang until it can acquire the DBus interface

I believe you can configure GDBus in either way. I suspect blocking with a message is what the user would expect.

superm1:
> I believe you can configure GDBus in either way. I suspect blocking with a message is what the user would expect.

We do require you to stop `fwupd.service` if you start `fwupdtool`, and so that jives pretty well with that too.

PR #3131, opened 2021-04-11
Quit the running fwupd instance when running fwupdtool

hughsie:
Also, this ensures we can't run more than one fwupdtool instance at a time.

Fixes https://github.com/fwupd/fwupd/issues/3019

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -1791,6 +1791,10 @@ fu_main_on_name_lost_cb (GDBusConnection *connection,
 			 gpointer user_data)
 {
 	FuMainPrivate *priv = (FuMainPrivate *) user_data;
+	if (priv->update_in_progress) {
+		g_warning ("name lost during a firmware update, ignoring");
+		return;
+	}
 	g_warning ("another service has claimed the dbus name %s", name);
 	g_main_loop_quit (priv->loop);
 }
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -65,6 +65,9 @@ struct FuUtilPrivate {
 	FwupdInstallFlags	 flags;
 	gboolean		 show_all;
 	gboolean		 disable_ssl_strict;
+	GThread			*dbus_thread;
+	guint			 dbus_owner_id;
+	gint			 dbus_status;
 	/* only valid in update and downgrade */
 	FuUtilOperation		 current_operation;
 	FwupdDevice		*current_device;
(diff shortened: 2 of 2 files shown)


hughsie:
It's not super awesome just having a "dead thread" to handle the DBus ownership but it seems to work...

superm1 (commented):
Why not just a lockfile?

hughsie:
> Why not just a lockfile

Because I want the first new `fwupdtool` instance to quit the _idle_ `fwupd` instance.

superm1:
> > Why not just a lockfile
> 
> Because I want the first new `fwupdtool` instance to quit the _idle_ `fwupd` instance.

Don't we already do that? https://github.com/fwupd/fwupd/blob/master/src/fu-tool.c#L176

hughsie:
> Don't we already do that? https://github.com/fwupd/fwupd/blob/master/src/fu-tool.c#L176

On one hand, I forgot all about that. On the other hand, what happens if we ask the systemd-controlled instance to manually stop during a firmware update? My gut says it probably does SIGQUIT and then after a few seconds does SIGKILL, and that's probably not what we want given a lot of the firmware updates take a lot longer than that. I guess my proposal would be to remove the calls to fu_systemd_unit_stop() and then let the daemon handle the replacement when it knows it's safe.

superm1:
>My gut says it probably does SIGQUIT and then after a few seconds does SIGKILL, 

Actually this behavior is entirely controllable by `KillMode=` and `KillSignal=`. 

> and that's probably not what we want given a lot of the firmware updates take a lot longer than that. I guess my proposal would 
> be to remove the calls to fu_systemd_unit_stop() and then let the daemon handle the replacement when it knows it's safe.

Keep in mind someone launching `fwupdtool` as root while `fwupd.service` is running is a pretty advanced thing to do.  Don't want to over-engineer the solution.

The original report was two `fwupdtool` processes conflicting, which you might argue is better solved by it instead becoming two `fwupdgmr`  processes talking to one fwupd daemon process....

hughsie:
> The original report was two fwupdtool processes conflicting

So maybe the fix here is to make fwupdtool claim the org.freedesktop.fwupd**tool** name -- which won't affect the fwupd process at all (other than the systemd service stopping) and would prevent the 2nd instance from running. I totally agree about using the daemon; it's almost like we designed something that provides this exact system-state-sharing functionality already...

superm1:
Well back to my point about simple solution - if it's just two fwupdtool processes what's wrong with a simple lockfile? Or even exclusive access to the sqlite3 database. If you have exclusive access then the second process should fail immediately.

hughsie:
> what's wrong with a simple lockfile

Isn't a lockfile far from simple when we're talking about making it work on win32 as well? Although, I guess the same point can be made about win32 where there's no DBus daemon.... If you have any pointers on how to implement a lockfile (ideally with GIO/Glib, rather than Unix primitives) I'm happy to give this a go.

superm1:
Sqlite3 has this stuff natively:
https://sqlite.org/lockingv3.html

hughsie:
I'm not totally convinced about using sqlite locking. What about https://github.com/fwupd/fwupd/pull/3147 ?

superm1:
> I'm not totally convinced about using sqlite locking. What about #3147 ?

Seems fine to me.

PR #3134, opened 2021-04-12
flashrom/lspcon: avoid parsing sysfs paths

tari:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

Parsing bus numbers out of sysfs paths may be fragile, so use udev to find the I2C bus matching a LSPCON device. Support for specifying an I2C bus by path rather than number is also required in flashrom, implemented at https://review.coreboot.org/c/flashrom/+/51967; the flashrom dependency will require an update to provide this support once merged upstream.

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -1747,6 +1747,50 @@ fu_udev_device_get_devtype (FuUdevDevice *self)
 #endif
 }
 
+/**
+ * fu_udev_device_get_siblings_with_subsystem
+ * @self: a #FuUdevDevice
+ * @subsystem: the name of a udev subsystem
+ *
+ * Get a list of devices that are siblings of self and have the
+ * provided subsystem.
+ *
+ * Returns (element-type FuUdevDevice) (transfer full): devices
+ *
+ * Since: 1.6.0
+ */
+GPtrArray *
+fu_udev_device_get_siblings_with_subsystem (FuUdevDevice *self,
+					    const gchar *const subsystem)
+{
+	g_autoptr(GPtrArray) out = g_ptr_array_new_with_free_func (g_object_unref);
+
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GUdevDevice) udev_parent = g_udev_device_get_parent (priv->udev_device);
+	const gchar *udev_parent_path = g_udev_device_get_sysfs_path (udev_parent);
+	g_autoptr(GUdevClient) udev_client = g_udev_client_new (NULL);
+
+	g_autoptr(GList) enumerated = g_udev_client_query_by_subsystem (udev_client, subsystem);
+	for (GList *element = enumerated; element != NULL; element = element->next) {
+		g_autoptr(GUdevDevice) enumerated_device = element->data;
+		g_autoptr(GUdevDevice) enumerated_parent =
+			g_udev_device_get_parent (enumerated_device);
+		const gchar *enumerated_parent_path =
+			g_udev_device_get_sysfs_path (enumerated_parent);
+
+		/* if the sysfs path of self's parent is the same as that of the
+		 * located device's parent, they are siblings */
+		if (g_strcmp0 (udev_parent_path, enumerated_parent_path) == 0) {
+			g_ptr_array_add (out,
+					 fu_udev_device_new (g_steal_pointer (&enumerated_device)));
+		}
+	}
+#endif
+
+	return g_steal_pointer(&out);
+}
+
 static void
 fu_udev_device_get_property (GObject *object, guint prop_id,
 			    GValue *value, GParamSpec *pspec)
--- a/libfwupdplugin/fu-udev-device.h
+++ b/libfwupdplugin/fu-udev-device.h
@@ -118,3 +118,5 @@ gboolean	 fu_udev_device_write_sysfs		(FuUdevDevice	*self,
 							 GError		**error)
 							 G_GNUC_WARN_UNUSED_RESULT;
 const gchar	*fu_udev_device_get_devtype		(FuUdevDevice	*self);
+GPtrArray 	*fu_udev_device_get_siblings_with_subsystem (FuUdevDevice *self,
+							     const gchar *subsystem);
--- a/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
+++ b/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
@@ -11,13 +11,12 @@
 
 #include <libflashrom.h>
 
-#define I2C_PATH_REGEX "/i2c-([0-9]+)/"
 #define HID_LENGTH 8
 #define DEVICE_GUID_FORMAT "FLASHROM-LSPCON-I2C-SPI\\VEN_%s&DEV_%s"
 
 struct _FuFlashromLspconI2cSpiDevice {
 	FuFlashromDevice	  parent_instance;
-	gint			  bus_number;
+	gchar			 *bus_path;
 	guint8			  active_partition;
 };
 
@@ -49,6 +48,15 @@ static struct flashrom_layout layout = {
 	.entries = entries, .num_entries = sizeof(entries) / sizeof(entries[0])
 };
 
+static void
+fu_flashrom_lspcon_i2c_spi_device_finalize(GObject *object)
+{
+	FuFlashromLspconI2cSpiDevice *self =
+		FU_FLASHROM_LSPCON_I2C_SPI_DEVICE (object);
+
+	g_free (self->bus_path);
+	G_OBJECT_CLASS (fu_flashrom_lspcon_i2c_spi_device_parent_class)->finalize (object);
+}
 
 static void
 fu_flashrom_lspcon_i2c_spi_device_init (FuFlashromLspconI2cSpiDevice *self)
@@ -64,9 +72,7 @@ fu_flashrom_lspcon_i2c_spi_device_probe (FuDevice *device, GError **error)
 	FuFlashromDevice *flashrom_device = FU_FLASHROM_DEVICE (device);
 	FuDeviceClass *klass =
 		FU_DEVICE_CLASS (fu_flashrom_lspcon_i2c_spi_device_parent_class);
-	g_autoptr(GRegex) regex = NULL;
-	g_autoptr(GMatchInfo) info = NULL;
-	const gchar *path = NULL;
+	g_autoptr(GPtrArray) sibling_buses = NULL;
 
 	/* FuFlashromDevice->probe */
 	if (!klass->probe (device, error))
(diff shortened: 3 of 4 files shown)


superm1:
@hughsie the CI failure seems to be caused by the help2man changes.  I think we need to configure the CI jobs to not let libjcat build man pages when subproject'ed.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -64,9 +72,12 @@ fu_flashrom_lspcon_i2c_spi_device_probe (FuDevice *device, GError **error)
...
 	FuDeviceClass *klass =
 		FU_DEVICE_CLASS (fu_flashrom_lspcon_i2c_spi_device_parent_class);
-	g_autoptr(GRegex) regex = NULL;
-	g_autoptr(GMatchInfo) info = NULL;
-	const gchar *path = NULL;
+	g_autoptr(GUdevDevice) self_device = fu_udev_device_get_dev (FU_UDEV_DEVICE(device));
+	g_autoptr(GUdevDevice) self_parent_device = g_udev_device_get_parent (self_device);
+	const gchar *self_parent_path = g_udev_device_get_sysfs_path (self_parent_device);
+	g_autoptr(GUdevClient) udev_client = g_udev_client_new (NULL);
+	g_autoptr(GUdevEnumerator) udev_enumerator = g_udev_enumerator_new (udev_client);
```
superm1, line 79: I think I would prefer to see direct usage of `GUdev` land in `fu-udev-device.c` as a helper function of some sort (maybe `fu_udev_get_sibling` with some arguments of what subsystem to find siblings of?)

superm1:
> https://review.coreboot.org/c/flashrom/+/51967

In that PR can you add some sort of way for fwupd to determine that it's interacting with a new enough libflashrom or not?  Otherwise we're going to potentially have non-obvious failures that are difficult to root cause.

tari:
> > https://review.coreboot.org/c/flashrom/+/51967
> 
> In that PR can you add some sort of way for fwupd to determine that it's interacting with a new enough libflashrom or not? Otherwise we're going to potentially have non-obvious failures that are difficult to root cause.

That's difficult to do because the existing implementation in flashrom doesn't assign much structure to its error codes; in particular both misconfiguration and I/O errors could return -1 as `SPI_ERROR_GENERIC` or `EPERM`. We know that if programmer init succeeds with this option then it's supported, so we can really only try it and see what happens.

However, I don't think this is a significant problem- flashrom will bail out and print a message if given programmer options that aren't handled, which from a library consumer's perspective is the same as if you requested a programmer that doesn't exist: `flashrom_programmer_init_fails` and it prints a message. Since we similarly don't have a way to check if flashrom is new enough to even support the LSPCON programmer, I think it's okay to assume it supports the `devpath` option.

An alternative might be to try extracting the bus number if init with `devpath` fails, but that sounds like a worse option.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
hughsie: Yup, 100% agreed; all this needs to live in fu-udev-device.c

hughsie:
>  I think we need to configure the CI jobs to not let libjcat build man pages when subprojec

I just added it back as a dep for arch, and opened up https://github.com/hughsie/libjcat/pull/51

superm1:
>  Since we similarly don't have a way to check if flashrom is new enough to even support the LSPCON programmer, I think it's okay to assume it supports the devpath option.

Ah yes - OK then.  This does seem like a logical future improvement though that I would suggest.  Set up something that lets you query supported programmers and compare that to the requested one.  As more flashrom devices get supported I think it could easily become quite important to watch that the locally built flashrom was appropriately compiled with that support.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: Added fu_udev_device_get_siblings_with_subsystem.

libfwupdplugin/fu-udev-device.c
```diff
@@ -1747,6 +1747,51 @@ fu_udev_device_get_devtype (FuUdevDevice *self)
...
+	g_autoptr(GPtrArray)
+		out = g_ptr_array_new_with_free_func (g_object_unref);
+
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GUdevDevice) self_parent = g_udev_device_get_parent (priv->udev_device);
+	const gchar *self_parent_path = g_udev_device_get_sysfs_path (self_parent);
+	g_autoptr(GUdevClient) udev_client = g_udev_client_new (NULL);
+
+	g_autoptr(GList) enumerated = g_udev_client_query_by_subsystem(udev_client, subsystem);
```
superm1, line 1775: missing space after `g_udev_client_query_by_subsystem`

tari: Added.

libfwupdplugin/fu-udev-device.c
```diff
@@ -1747,6 +1747,51 @@ fu_udev_device_get_devtype (FuUdevDevice *self)
 #endif
 }
 
+/**
+ * fu_udev_device_get_siblings_with_subsystem
+ * @self : a #FuUdevDevice
```
hughsie, line 1752: no space before the `:` please, else the scanner gets upset

libfwupdplugin/fu-udev-device.c
```diff
@@ -1747,6 +1747,51 @@ fu_udev_device_get_devtype (FuUdevDevice *self)
...
+ * Returns (element-type FuUdevDevice) (transfer full): devices
+ *
+ * Since: 1.6.0
+ */
+GPtrArray *
+fu_udev_device_get_siblings_with_subsystem (FuUdevDevice *self,
+					    const gchar *const subsystem)
+{
+	g_autoptr(GPtrArray)
+		out = g_ptr_array_new_with_free_func (g_object_unref);
```
hughsie, line 1767: i think just a long line here, I'm fine with the column limit being broken when required

libfwupdplugin/fu-udev-device.c
```diff
@@ -1747,6 +1747,51 @@ fu_udev_device_get_devtype (FuUdevDevice *self)
...
+
+#ifdef HAVE_GUDEV
+	FuUdevDevicePrivate *priv = GET_PRIVATE (self);
+	g_autoptr(GUdevDevice) self_parent = g_udev_device_get_parent (priv->udev_device);
```
hughsie, line 1771: i think `udev_parent` is a better name for this

libfwupdplugin/fu-udev-device.c
```diff
@@ -1747,6 +1747,51 @@ fu_udev_device_get_devtype (FuUdevDevice *self)
...
+	const gchar *self_parent_path = g_udev_device_get_sysfs_path (self_parent);
```
hughsie, line 1772: and udev_parent_path

libfwupdplugin/fu-udev-device.h
```diff
@@ -118,3 +118,5 @@ gboolean	 fu_udev_device_write_sysfs		(FuUdevDevice	*self,
 							 GError		**error)
 							 G_GNUC_WARN_UNUSED_RESULT;
 const gchar	*fu_udev_device_get_devtype		(FuUdevDevice	*self);
+GPtrArray 	*fu_udev_device_get_siblings_with_subsystem (FuUdevDevice *self,
```
hughsie, line 121: siblings or children? I think we use the latter in other places, but I don't care either way too much.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -81,15 +87,34 @@ fu_flashrom_lspcon_i2c_spi_device_probe (FuDevice *device, GError **error)
...
-		return TRUE;
+	/* locate an I2C bus that is a sibling of this device, assumed to be
+	 * the bus the device is connected to */
+	sibling_buses = fu_udev_device_get_siblings_with_subsystem (FU_UDEV_DEVICE (self), "i2c-dev");
+	for (guint i = 0; i < sibling_buses->len; i++) {
+		FuUdevDevice *bus_device = g_ptr_array_index (sibling_buses, i);
+		const gchar *bus_path = fu_udev_device_get_device_file (bus_device);
+
+		if (bus_path == NULL) {
+			g_info ("Bus %s has no device file! Ignoring.",
```
hughsie, line 98: lowercase B and no full stop please, maybe something like `ignoring as %s has no device file` perhaps?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -81,15 +87,34 @@ fu_flashrom_lspcon_i2c_spi_device_probe (FuDevice *device, GError **error)
...
+			        fu_udev_device_get_sysfs_path (bus_device));
+			continue;
+		}
+
+		self->bus_path = g_strdup (bus_path);
+		if (i < (sibling_buses->len - 1)) {
+			g_info ("Ignoring %u additional sibling buses",
+				sibling_buses->len - i - 1);
+		}
```
hughsie, line 107: i'm not really sure what this is doing TBH.

libfwupdplugin/fu-udev-device.h
tari: This is specifically finding siblings by looking only for devices with the same parent as self. Getting children with a given subsystem would be a different operation.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: Modified slightly to hopefully be clearer. There may be more than one valid sibling bus, but we only care about one so the point is to note that there are additional buses that might have been chosen but weren't.

libfwupdplugin/fu-udev-device.h
hughsie: Gah, of course. Sorry for the noise.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -81,15 +87,36 @@ fu_flashrom_lspcon_i2c_spi_device_probe (FuDevice *device, GError **error)
...
+		if (bus_path == NULL) {
+			g_info ("ignoring bus %s with no device file",
+			        fu_udev_device_get_sysfs_path (bus_device));
+			continue;
+		}
+
+		self->bus_path = g_strdup (bus_path);
+		/* if there are other buses, ignore them (because we already
+		 * have one) but note they exist */
+		if (i < (sibling_buses->len - 1)) {
```
hughsie, line 106: no inner brackets needed here

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -81,15 +87,36 @@ fu_flashrom_lspcon_i2c_spi_device_probe (FuDevice *device, GError **error)
...
+			g_info ("%u additional buses found but ignored",
```
hughsie, line 107: I think this is g_debug, not g_info -- the latter would make it into the journal and that's a sure way to get a support case raised from an otherwise unaffected concerned customer :)

hughsie (dismissed):
two nits, and then LGTM, thanks.

hughsie (approved):

djcampello:
Please make sure that the flashrom changes make all the way to Chromium OS. As is this merge is breaking Chromium OS

djcampello:
This is actually not even upstream: https://review.coreboot.org/c/flashrom/+/51967 Please revert this change until that lands in both upstream and Chromium OS flashrom

hughsie:
The fact that it _wasn't upstream_ was a funky thing to omit from the initial review...

hughsie:
@djcampello can you put a PR together with the commits you want to revert?

djcampello:
https://github.com/fwupd/fwupd/pull/3150

PR #3147, opened 2021-04-14
Use an OFD lock to prevent more than one fwupdtool process

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3019

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -65,6 +65,7 @@ struct FuUtilPrivate {
 	FwupdInstallFlags	 flags;
 	gboolean		 show_all;
 	gboolean		 disable_ssl_strict;
+	gint			 lock_fd;
 	/* only valid in update and downgrade */
 	FuUtilOperation		 current_operation;
 	FwupdDevice		*current_device;
@@ -262,6 +263,8 @@ fu_util_private_free (FuUtilPrivate *priv)
 		g_object_unref (priv->progressbar);
 	if (priv->context != NULL)
 		g_option_context_free (priv->context);
+	if (priv->lock_fd != 0)
+		g_close (priv->lock_fd, NULL);
 	g_free (priv->current_message);
 	g_free (priv);
 }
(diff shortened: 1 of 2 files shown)


src/fu-tool.c
```diff
@@ -2781,6 +2784,48 @@ fu_util_switch_branch (FuUtilPrivate *priv, gchar **values, GError **error)
...
+	struct flock lockp = {
+		.l_type = F_WRLCK,
+		.l_whence = SEEK_SET,
+	};
+	g_autofree gchar *localstatedir = NULL;
+	g_autofree gchar *lockfn = NULL;
+
+	/* open file */
+	localstatedir = fu_common_get_path (FU_PATH_KIND_LOCALSTATEDIR);
+	lockfn = g_build_filename (localstatedir, "run", "fwuptool", NULL);
```
superm1, line 2800: `s/fwuptool/fwupdtool/`

src/fu-tool.c
```diff
@@ -2781,6 +2784,48 @@ fu_util_switch_branch (FuUtilPrivate *priv, gchar **values, GError **error)
...
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "failed to open %s",
+			     lockfn);
+		return FALSE;
+	}
+
+	/* write lock */
+	if (fcntl (priv->lock_fd, F_OFD_SETLK, &lockp) < 0) {
```
superm1, line 2814: I suppose it's not actually required to set `F_UNLCK` on close because the descriptor releases, right?

hughsie: Yup, that's why I chose ofd rather than a file lock.

superm1: Sorry, I should fully complete my thoughts - if the fwupdtool crashes I want to make sure that we don't end up with dangling descriptors, which I don't know if that's a thing.

superm1 (approved):

src/fu-tool.c
hughsie: Thanks, fixed. I think the tool should probably have been called fwuptool and fwupmgr but we can't go back in time now :)

djcampello:
Does it make sense to allow an option to busy wait for the lock to be released? I'm not sure is a great idea as there could be a stale lock but I'm trying to figure out the best way to serialize concurrent executions...

hughsie:
> Does it make sense to allow an option to busy wait for the lock to be released

From a "_fwupdtool is a development tool_" point of view I think exiting straight away makes sense. From a "_launching fwupdtool from an upstart script_" point of view maybe it does make sense to just wait for the lock to clear. If you need the latter I'd be fine reviewing it, but I've got tons of stuff on my plate for the next few days.

I chose OFD locks as it's impossible for them to be stale as the kernel cleans up after whatever userspace does.

superm1:
I almost wonder if the upstart fwupdtool scenario should have it's _own_ lock and that behavior.

PR #3152, opened 2021-04-15
trivial: Hard limit the fuzz duration to 10 minutes

hughsie:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,6 +55,7 @@ jobs:
 
   fuzzing:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     steps:
     - name: Build Fuzzers
       id: build


hughsie:
@superm1 I think the oss-fuzz service isn't respecting the fuzz-seconds any more for some reason. I guess this PR shows the `timeout-minutes: 10` is working...

superm1:
Good.. In that case I wonder if we should just align it closer to what we want oss-fuzz timeout to be.

hughsie:
The oss-fuzz timeout is 5 minutes, so I figured allowing 5 minutes **extra** for setup&teardown seemed plenty.

superm1:
> The oss-fuzz timeout is 5 minutes, so I figured allowing 5 minutes **extra** for setup&teardown seemed plenty.

Hmm actually it looks like the building and teardown might take longer than I expected.
Looking at https://github.com/fwupd/fwupd/pull/3152/checks?check_run_id=2353819811
It was 7 minutes to get to the point everything was built, and then the fuzzer only ran for 3 minutes before timing out.

So maybe you want closer to 15 then.

superm1 (approved):

PR #3159, opened 2021-04-19
fu-tool.c: Use traditional UNIX record locks if OFD is not available

Asiderr:
Open file description locks are Linux-specific. If fwupd is not built
for Linux, it should use the traditional UNIX record locks (F_SETLK).

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -352,6 +352,9 @@ endif
 if cc.has_header_symbol('fcntl.h', 'F_WRLCK')
   conf.set('HAVE_WRLCK', '1')
 endif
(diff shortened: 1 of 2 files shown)


hughsie:
Do we have to manually unlock F_SETLK?

Asiderr:
@hughsie Lock is automatically released when `g_close` is called.
Ps. CI failed due to some LFVS server problems:
`failed to download file: Could not resolve host: cdn.fwupd.org`

hughsie (approved):

superm1 (approved):

PR #3162, opened 2021-04-20
flashrom/lspcon: probe device to determine actual active partition

tari:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

While the PS175 will attempt to boot from the flash bank specified by the flag bytes, it does not always actually boot the specified bank because it also performs integrity checks on the target image. If those fail, it will instead boot from bank 3 which should always be treated as read-only.

In all cases, the device exposes a register on I2C that says which bank it actually booted from. Update the plugin to probe which bank the device has actually booted, rather than assuming it booted the one the flags say it should.

In support of this, replace the mutable static flash layout with making copies of a "pristine" layout. The mutable layout can easily cause unintended flash regions to be included in operations; making a copy of a layout that includes nothing and turning on the desired regions should be much easier to use correctly.

### DIFF
--- a/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
+++ b/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
@@ -9,7 +9,12 @@
 #include "fu-flashrom-device.h"
 #include "fu-flashrom-lspcon-i2c-spi-device.h"
 
+#include <errno.h>
+#include <fcntl.h>
+#include <glib/gstdio.h>
 #include <libflashrom.h>
+#include <linux/i2c-dev.h>
+#include <sys/ioctl.h>
 
 #define I2C_PATH_REGEX "/i2c-([0-9]+)/"
 #define HID_LENGTH 8
@@ -37,12 +42,13 @@ struct flashrom_layout {
 	gsize			 num_entries;
 };
 
-static const struct romentry ENTRIES_TEMPLATE[5] = {
+static const struct romentry ENTRIES_TEMPLATE[6] = {
 	{ .start = 0x00002, .end = 0x00003, .included = FALSE, .name = "FLAG" },
 	{ .start = 0x10000, .end = 0x1ffff, .included = FALSE, .name = "PAR1" },
 	{ .start = 0x20000, .end = 0x2ffff, .included = FALSE, .name = "PAR2" },
 	{ .start = 0x15000, .end = 0x15002, .included = FALSE, .name = "VER1" },
 	{ .start = 0x25000, .end = 0x25002, .included = FALSE, .name = "VER2" },
+	{ .start = 0x35000, .end = 0x35002, .included = FALSE, .name = "VERBOOT" },
 };
 
 typedef struct flashrom_layout FlashromLayout;
@@ -118,12 +124,51 @@ fu_flashrom_lspcon_i2c_spi_device_open (FuDevice *device,
 	FuDeviceClass *klass =
 		FU_DEVICE_CLASS (fu_flashrom_lspcon_i2c_spi_device_parent_class);
 	g_autofree gchar *temp = NULL;
+	g_autofree gchar *bus_path = NULL;
+	gint bus_fd = -1;
 
 	/* flashrom_programmer_init() mutates the programmer_args string. */
 	temp = g_strdup_printf ("bus=%d", self->bus_number);
 	fu_flashrom_device_set_programmer_args (flashrom_device, temp);
 
+	/* open the bus, not the device represented by self */
+	bus_path = g_strdup_printf ("/dev/i2c-%d", self->bus_number);
+	g_debug ("communicating with device on %s", bus_path);
+	if ((bus_fd = g_open (bus_path, O_RDWR, 0)) == -1) {
+		g_set_error (error, G_IO_ERROR, g_io_error_from_errno (errno),
+			     "failed to open %s read-write", bus_path);
+		return FALSE;
+	}
+	fu_udev_device_set_fd (FU_UDEV_DEVICE (self), bus_fd);
+	fu_udev_device_set_flags (FU_UDEV_DEVICE (self), FU_UDEV_DEVICE_FLAG_NONE);
+
 	return klass->open (device, error);
+}
+
+static gboolean
+probe_active_flash_partition (FuFlashromLspconI2cSpiDevice *self,
+			      guint8 *partition,
+			      GError **error)
+{
+	guint8 data;
+
+	/* read register 0x0e on page 5, which is set to the currently-running
+	 * flash partition number */
+	if (!fu_udev_device_ioctl (FU_UDEV_DEVICE (self),
+				   I2C_SLAVE, (guint8 *) (0x9a >> 1), NULL, error)) {
+		g_prefix_error (error, "failed to set I2C slave address: ");
+		return FALSE;
+	}
+	if (!fu_udev_device_pwrite (FU_UDEV_DEVICE (self), 0, 0x0e, error)) {
+		g_prefix_error (error, "failed to write register address: ");
+		return FALSE;
+	}
+	if (!fu_udev_device_pread (FU_UDEV_DEVICE (self), 0, &data, error)) {
+		g_prefix_error (error, "failed to read register value: ");
+		return FALSE;
+	}
+	*partition = data;
+	return TRUE;
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -141,17 +222,9 @@ fu_flashrom_lspcon_i2c_spi_device_set_version (FuDevice *device, GError **error)
...
-		g_set_error_literal (error, FWUPD_ERROR, FWUPD_ERROR_READ,
-				     "failed to read active partition");
-		return FALSE;
-	}
-
-	/* find the current version for the active partition */
-	addr = entries[self->active_partition + 2].start;
+	/* extract the active partition's version number */
+	addr = layout->entries[self->active_partition + 2].start;
+	g_return_val_if_fail(addr < (flash_size - 2), FALSE);
```
hughsie, line 227: i'd be much happier with a g_set_error here if it's a plausible runtime problem.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -117,19 +188,29 @@ fu_flashrom_lspcon_i2c_spi_device_set_version (FuDevice *device, GError **error)
...
+	gint active_partition;
 
 	/* set up flashrom layout */
 	flashctx = fu_flashrom_device_get_flashctx (flashrom_device);
-	flashrom_layout_set (flashctx, &layout);
+	flashrom_layout_set (flashctx, layout);
+
+	/* get the active partition */
+	if ((active_partition = probe_active_flash_partition (self, error)) == -1)
+		return FALSE;
```
hughsie, line 202: ew, can probe_active_flash_partition please return gboolean and then just pass active_partition by reference as the second parameter.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -101,14 +123,63 @@ fu_flashrom_lspcon_i2c_spi_device_open (FuDevice *device,
...
+{
+	gint bus_fd = fu_udev_device_get_fd (FU_UDEV_DEVICE (self));
+	guint8 data = 0x0e;
+
+	/* device must have been opened */
+	g_return_val_if_fail (bus_fd != -1, -1);
+
+	/* read register 0x0e on page 5, which is set to the currently-running
+	 * flash partition number */
+	if (ioctl (bus_fd, I2C_SLAVE, 0x9a >> 1) == -1) {
```
hughsie, line 158: fu_udev_device_ioctl?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -101,14 +123,63 @@ fu_flashrom_lspcon_i2c_spi_device_open (FuDevice *device,
...
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     g_io_error_from_errno (errno),
+				     "failed to set I2C slave address");
+		return -1;
+	}
+
+	if (write (bus_fd, &data, 1) != 1) {
```
hughsie, line 166: fu_udev_device_pwrite?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -101,14 +123,63 @@ fu_flashrom_lspcon_i2c_spi_device_open (FuDevice *device,
...
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     g_io_error_from_errno (errno),
+				     "failed to write register address");
+		return -1;
+	}
+	if (read (bus_fd, &data, 1) != 1) {
```
hughsie, line 173: fu_udev_device_pread?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -101,14 +123,63 @@ fu_flashrom_lspcon_i2c_spi_device_open (FuDevice *device,
...
 	g_autofree gchar *temp = NULL;
+	g_autofree gchar *bus_path = NULL;
+	gint bus_fd = -1;
 
 	/* flashrom_programmer_init() mutates the programmer_args string. */
 	temp = g_strdup_printf ("bus=%d", self->bus_number);
 	fu_flashrom_device_set_programmer_args (flashrom_device, temp);
 
+	/* open the bus, not the device represented by self */
+	bus_path = g_strdup_printf("/dev/i2c-%d", self->bus_number);
```
hughsie, line 134: space before bracket

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -37,17 +42,34 @@ struct flashrom_layout {
...
 
-static struct flashrom_layout layout = {
-	.entries = entries, .num_entries = sizeof(entries) / sizeof(entries[0])
-};
+typedef struct flashrom_layout FlashromLayout;
+
+static FlashromLayout *create_flash_layout(void)
+{
+	struct romentry *entries = g_malloc (sizeof (ENTRIES_TEMPLATE));
+	struct flashrom_layout *out = g_malloc (sizeof (FlashromLayout));
```
hughsie, line 59: g_new?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -37,17 +42,34 @@ struct flashrom_layout {
...
+
+	memcpy (entries, ENTRIES_TEMPLATE, sizeof (ENTRIES_TEMPLATE));
```
hughsie, line 61: g_memdup?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: The partition number is already validated and the address is static, so this is just defensive against getting the addresses wrong (which would be a bug).

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: That does make rather more sense.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: Kind of awkward casting the bus address to a pointer, but done.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: Didn't think pwrite would work since `i2c-dev` isn't really seekable, but turns out it does.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -101,14 +122,53 @@ fu_flashrom_lspcon_i2c_spi_device_open (FuDevice *device,
...
+			      guint8 *partition,
+			      GError **error)
+{
+	guint8 data;
+
+	/* read register 0x0e on page 5, which is set to the currently-running
+	 * flash partition number */
+	if (!fu_udev_device_ioctl (FU_UDEV_DEVICE (self),
+				   I2C_SLAVE, (guint8 *) (0x9a >> 1), NULL, error)) {
+		g_prefix_error (error, "failed to set I2C slave address");
```
hughsie, line 157: THis is literally a prefix, you need:

 		g_prefix_error (error, "failed to set I2C slave address: ");

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
hughsie: then space before bracket please.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -37,17 +42,33 @@ struct flashrom_layout {
...
+{
+	struct flashrom_layout *out = g_new (FlashromLayout, 1);
+	struct romentry *entries = g_memdup (ENTRIES_TEMPLATE, sizeof (ENTRIES_TEMPLATE));
+
+	out->entries = entries;
+	out->num_entries = sizeof (ENTRIES_TEMPLATE) / sizeof (*ENTRIES_TEMPLATE);
+	return out;
+}
+
+static void dispose_flash_layout(FlashromLayout *layout) {
```
hughsie, line 66: `static void` on a new line pls

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -37,17 +42,33 @@ struct flashrom_layout {
...
+	{ .start = 0x25000, .end = 0x25002, .included = FALSE, .name = "VER2" },
+	{ .start = 0x35000, .end = 0x35002, .included = FALSE, .name = "VERBOOT" },
 };
 
-static struct flashrom_layout layout = {
-	.entries = entries, .num_entries = sizeof(entries) / sizeof(entries[0])
-};
+typedef struct flashrom_layout FlashromLayout;
+
+static FlashromLayout *create_flash_layout(void)
```
hughsie, line 56: space before the bracket

hughsie: and `static FlashromLayout` on a new line

hughsie (dismissed):
Two nits then LGTM, thanks.

hughsie (approved):

PR #3169, opened 2021-04-21
1_5_X: Add an option to avoid building UEFI binary

superm1:
This will allow packagers to choose to distribute the binary from
the fwupd-efi subproject instead.

This goes with https://github.com/fwupd/fwupd/pull/3153, and assuming it's merged will be what can go into 1.5.x to allow packagers to optionally choose to use subproject instead.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -51,3 +51,4 @@ option('efi_sbat_distro_summary', type : 'string', value : '', description : 'SB
 option('efi_sbat_distro_pkgname', type : 'string', value : '', description : 'SBAT distribution package name, e.g. fwupd')
 option('efi_sbat_distro_version', type : 'string', value : '', description : 'SBAT distribution version, e.g. fwupd-1.5.6.fc33')
 option('efi_sbat_distro_url', type : 'string', value : '', description : 'SBAT distribution URL, e.g. https://src.fedoraproject.org/rpms/fwupd')
+option('efi_binary', type: 'boolean', value: 'true', description : 'build included UEFI binary if missing')
--- a/plugins/uefi-capsule/meson.build
+++ b/plugins/uefi-capsule/meson.build
@@ -1,5 +1,11 @@
 if get_option('plugin_uefi_capsule')
-subdir('efi')
+
+if get_option('efi_binary')
+  efi_binary = dependency('fwupd-efi', required: false)
+  if not efi_binary.found()
+    subdir('efi')
+  endif
+endif
 
 cargs = ['-DG_LOG_DOMAIN="FuPluginUefiCapsule"']
 


meson_options.txt
```diff
@@ -51,3 +51,4 @@ option('efi_sbat_distro_summary', type : 'string', value : '', description : 'SB
 option('efi_sbat_distro_pkgname', type : 'string', value : '', description : 'SBAT distribution package name, e.g. fwupd')
 option('efi_sbat_distro_version', type : 'string', value : '', description : 'SBAT distribution version, e.g. fwupd-1.5.6.fc33')
 option('efi_sbat_distro_url', type : 'string', value : '', description : 'SBAT distribution URL, e.g. https://src.fedoraproject.org/rpms/fwupd')
+option('efi_binary', type: 'boolean', value: 'true', description : 'build included UEFI binary')
```
hughsie, line 54: maybe `plugin_uefi_capsule_efi_binary`? It might not be the only EFI binary we every use, vis-a-vis superio. Same for the description.

superm1: In 1_5_X?  I would think such a thing wouldn't backport that far.

hughsie: Hmm, maybe not. Maybe one big switch for "all the efi stuff" is okay

hughsie (approved):
lets get a a fwupd-efi release out before merging

plugins/uefi-capsule/meson.build
```diff
@@ -1,5 +1,11 @@
 if get_option('plugin_uefi_capsule')
-subdir('efi')
+
+if get_option('efi_binary')
+  efi_binary = dependency ('fwupd-efi', required: false)
```
hughsie, line 4: does meson not complain about the space after dependency?

hughsie:
> lets get a a fwupd-efi release out before merging

....and we're done. Feel free to merge whenever.

plugins/uefi-capsule/meson.build
hughsie: It might be safer just to rely on efi_binary for 1.5.x afterall?

superm1: >does meson not complain about the space after dependency?

I don't think meson cares, but I can certainly fixup the style :)

>It might be safer just to rely on efi_binary for 1.5.x afterall?

Can you elaborate by what you mean?  I didn't add any subproject for 1.5.x, it's entirely a user decision whether or not to use it.  If you think it's better to add subproject to all the older releases like we did for master, that's certainly possible.  I just thought it might be a bit much for a stable release when realistically only some packagers will change this on stable releases.

hughsie: The `dependency ('fwupd-efi` is checking the .pc file, no? Isn't it strictly a runtime requirement, not a build time one?

superm1: Right, it does check the pc file at build time.  So at build time - if the PC file is there and efi_binary was set to true, don't build the binary.  If it's not there build the binary. 

The runtime check we already have in place, so no concerns whether or not binary is built for 1.5.x (will need to check this on older versions too though when backporting)

hughsie: Ahh, so it's a way of ignoring the efi_binary=true if the binary really is already installed. I don't know if that's cunning or insane.

superm1: Cunning!

PR #3173, opened 2021-04-22
Add a composite ID that is used to identify all devices in the same physical package

hughsie:
Type of pull request:
- [X] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -833,10 +833,26 @@ fu_device_get_root (FuDevice *self)
 	return g_object_ref (self);
 }
 
+static void
+fu_device_set_composite_id (FuDevice *self, const gchar *composite_id)
+{
+	GPtrArray *children;
+
+	/* subclassed simple setter */
+	fwupd_device_set_composite_id (FWUPD_DEVICE (self), composite_id);
+
+	/* all children */
+	children = fu_device_get_children (self);
+	for (guint i = 0; i < children->len; i++) {
+		FuDevice *child_tmp = g_ptr_array_index (children, i);
+		fu_device_set_composite_id (child_tmp, composite_id);
+	}
+}
+
 /**
  * fu_device_set_parent:
  * @self: A #FuDevice
- * @parent: A #FuDevice
+ * @parent: (nullable): A #FuDevice
  *
  * Sets any parent device. An parent device is logically "above" the current
  * device and this may be reflected in client tools.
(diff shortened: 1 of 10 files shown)


hughsie:
I've got a local rebased confirm-actions branch that uses this functionality to do the right thing and show the releases from the same archive on the composite device.

src/fu-main.c
```diff
@@ -843,7 +802,9 @@ fu_main_install_with_helper (FuMainAuthHelper *helper_ref, GError **error)
 		if (devices_possible == NULL)
 			return FALSE;
 	} else {
-		devices_possible = fu_main_get_device_family (helper, error);
```
hardboprobot, line 846: Won't this break the current cases using ParentGuid?

hughsie: I don't think so; ParentGuid is a GUID we use to set the parent FuDevice on the child FuDevice. The composite_id is really like a surname of the oldest grandparent. This means you can do GetDevices to return all the devices, and use the ParentDeviceId to construct the ID -- or use the Composite ID to find all the devices that are either children, parents, grandparents or siblings of any level.

libfwupdplugin/fu-device.c
```diff
@@ -851,6 +867,9 @@ fu_device_set_parent (FuDevice *self, FuDevice *parent)
 {
 	g_return_if_fail (FU_IS_DEVICE (self));
 
+	/* set the composite ID on the siblings, children and grandchildren */
+	fu_device_set_composite_id (parent, fu_device_get_composite_id (parent));
```
hardboprobot, line 871: Why `fu_device_set_composite_id (parent, fu_device_get_composite_id (parent));` instead of `fu_device_set_composite_id (self, fu_device_get_composite_id (parent));` ?

This won't set the composite id on the device the first time it's called because `parent` is still not set as the device parent. Is this just a typo?

hughsie: > This won't set the composite id on the device the first time it's called because parent is still not set as the device parent

Hmm, true; **good catch**. I originally set the parent so we could set the composite ID of the other siblings of `self` -- but this isn't needed as it's by definition already been set. Fixed, thanks.

hardboprobot (approved):

hughsie:
@superm1 if you have a Dell docking station handy I'd appreciate if you can check this doesn't regress anything -- I'm at the wrong desk today. There is a slight behaviour change (we include great grandchildren, children, siblings and parents, rather than just the latter three) but I don't think that will affect anything.

superm1:
> @superm1 if you have a Dell docking station handy I'd appreciate if you can check this doesn't regress anything -- I'm at the wrong desk today. There is a slight behaviour change (we include great grandchildren, children, siblings and parents, rather than just the latter three) but I don't think that will affect anything.

@CragW may be able to check..

hughsie:
Okay, l forgot you don't work for Dell any more, sorry! :) I'll merge this, and we can fix up any issues -- as this unlocks another PR I'd like to push soon. The Dell WD19 dock is tested before the tarball is tagged IIRC.

CragW:
I just checked WD19's family tree, looks good.

hughsie:
@CragW many thanks for that dude, much appreciated.

PR #3180, opened 2021-04-23
dfu: Fix up some overzealous replacement when renaming the DFU files

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
Well it's nice that it wasn't _broken_ at least by the overzealous replacements previously :)

PR #3188, opened 2021-04-26
flashrom: support multiple device GTypes on I2C

tari:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [x] Documentation

In order to support multiple kinds of I2C devices, the flashrom plugin must
provide its own backend_device_added function to determine the correct
specialized GType for a given device because there is no standard way for
fwupd to identify I2C devices beyond simply matching a udev subsystem.
In order to support additional GTypes, the plugin does additional probing
based on sysfs name which is suitable for the current LSPCON support and
planned future devices.

While support for various different kinds of device would be better implemented
in different plugins to achieve logical separation of devices that share only
an underlying library, libflashrom as it currently exists does not support
such a use case: document that reasoning in the plugin's README.

### DIFF
--- a/plugins/flashrom/README.md
+++ b/plugins/flashrom/README.md
@@ -4,7 +4,16 @@ Flashrom
 Introduction
 ------------
 
-This plugin uses `flashrom` to update the system firmware.
+This plugin uses `libflashrom` to update various types of device:
+
+ * system firmware
+ * Parade Technologies DisplayPort-to-HDMI converters
+
+Though these devices are very dissimilar, they must currently be handled in a
+single plugin because the underlying APIs in `libflashrom` are not reentrant:
+it has internal global state that means it cannot be safely used from multiple
+points in a single process, as it may be if multiple plugins were to use the
+library.
 
 Firmware Format
 ---------------
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -40,6 +40,7 @@ Plugin = flashrom
 # Parade PS175
 [1AF80175:00]
 Plugin=flashrom
+GType=FuFlashromLspconI2cSpiDevice
 FlashromProgrammer=lspcon_i2c_spi
 Name=PS175
 Vendor=Parade
--- a/plugins/flashrom/fu-plugin-flashrom.c
+++ b/plugins/flashrom/fu-plugin-flashrom.c
@@ -195,6 +195,47 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 }
 
 gboolean
+fu_plugin_backend_device_added (FuPlugin *plugin, FuDevice *device, GError **error)
+{
+	FuUdevDevice *udev_device;
+	const gchar *dev_name;
+	GType specialized_type;
+	g_autoptr(FuDevice) specialized_device = NULL;
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	/* doesn't handle non-udev devices */
+	if (!FU_IS_UDEV_DEVICE (device))
+		return TRUE;
+	udev_device = FU_UDEV_DEVICE (device);
+
+	/* only care about i2c devices */
+	if (g_strcmp0 (fu_udev_device_get_subsystem (udev_device), "i2c") != 0)
+		return TRUE;
+
+	/* find device quirks by sysfs name */
+	dev_name = fu_udev_device_get_sysfs_attr (FU_UDEV_DEVICE (device), "name", NULL);
+	if (dev_name != NULL) {
+		fu_device_add_instance_id_full (device, dev_name,
+						FU_DEVICE_INSTANCE_FLAG_ONLY_QUIRKS);
+	}
+
+	/* construct an instance of the specialized device type from quirks */
+	if ((specialized_type = fu_device_get_specialized_gtype (device)) == G_TYPE_INVALID) {
+		g_debug ("no GType set for %s: ignoring", dev_name);
+		return TRUE;
+	}
+	specialized_device = g_object_new (specialized_type, NULL);
+	fu_device_incorporate (specialized_device, device);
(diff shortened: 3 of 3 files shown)


hughsie:
> While support for various different kinds of device would be better implemented in different plugins

I'm kinda wondering at what point implementing the entire erase/write and read in a dedicated fwupd plugin would just be a lot less code and complexity. There's certainly an impedance mismatch between the flashrom API and fwupd, and even trivial stuff like "give me a percentage completion" has not been merged months after being submitted for review. I'm getting increasing anxious about building on top of unstable foundations.

I think now is probably the appropriate time to start pushing back on adding workarounds and complexity, especially when the "future device" code isn't proposed yet.

tari:
> especially when the "future device" code isn't proposed yet.

In this particular instance, the "future device" is one that is currently supported by flashrom, and for which I'm working on fwupd support.

tari:
In thinking about the practicalities of avoiding flashrom, I think complexity of such a thing would be more challenging than that of the flashrom solution. The advantage of using flashrom is that it already understands how to program a large number of flash devices, so for devices that basically provide an interface over which commands will be tunneled to their flash (true of both system firmware and LSPCON, as things stand) a lot of complexity is handled by flashrom, and it's frequently updated to add support for new chips.

I appreciate your frustration with a seeming lack of activity from upstream flashrom though, as I have similar experiences with them such as in the support I've implemented for #3150 that's not yet merged.

hughsie:
> The advantage of using flashrom is that it already understands how to program a large number of flash devices

I don't think the generic DediProg-style flashing tools are super interesting to support from a fwupd point of view. The *internal* one obviously is super interesting, as there's a lot of work that's gone into flashrom to make it work on a huge number of motherboards -- with all the mess of ECs getting in the way of writing. The other ones; not _so_ convinced.

I do wonder how much LSPCON code is left in flashrom with much of the romentry logic duplicated in fwupd.

> it's frequently updated to add support for new chips

Dude, we do a release every few weeks -- it's exhausting trying to keep up with all the different vendors and OEMs requesting new fwupd versions :)

tari:
> I don't think the generic DediProg-style flashing tools are super interesting to support from a fwupd point of view. The internal one obviously is super interesting

I was more referring to specific flash chips rather than programming adapters, which often have various quirks around how they handle certain commands, which also leads to:

> it's frequently updated to add support for new chips

Maintaining a separate list of chip quirks would be duplication of effort (or maybe the goal would be to completely replace flashrom for general use?), and it may be difficult to realize when support for new chips is required: a given device might suddenly be built with a different chip and nobody is thinking about the implications of a BOM change on firmware updates, meaning fwupd starts failing on some devices for reasons that might not even be visible without physically opening the device up.

hughsie:
> Maintaining a separate list of chip quirks would be duplication of effort 

It kinda already is. https://github.com/fwupd/fwupd/blob/master/plugins/intel-spi/intel-spi.quirk and https://github.com/fwupd/fwupd/blob/master/plugins/dfu/dfu.quirk to name but two.

> (or maybe the goal would be to completely replace flashrom for general use?)

I'm not totally happy building the success of fwupd on top of a project that I can't even get trivial patches into. There's also this weird power play between Google and the "old" flashrom developers and it's not something that I want to get myself caught up with. As my wife says: _not my circus, not my monkeys_.

> meaning fwupd starts failing on some devices for reasons that might not even be visible

At least for the VLI plugin we do a JEDEC probe of the chip so we can know the right command for SPI erase, for instance: https://github.com/fwupd/fwupd/blob/master/plugins/vli/vli-usbhub.quirk#L61

I think my point generally is that we should just "make things magically work" with the minimum amount of code, and the minimum number of dependencies. The more "layers" of the cake, the more places where an imperfect or leaky interface causes real world bugs. I'd apportion a good part of the success of fwupd is the fact that you install it, and it works. No configuration, no device enrolment etc.

djcampello:
Sorry I'm late, isn't this functionality implemented by FlashromProgrammer entry on quirk?

tari:
That handles setting a different programmer to use in flashrom, but does not accomodate another FlashromDevice subclass (beyond the one that currently exists) that needs to do its own work.

djcampello:
It does.... You just need to add a check during init that the programmer matches the programmer supported by the new FlashromDevice. See https://github.com/fwupd/fwupd/blob/master/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c#L99

Something like: 
```
if (g_strcmp0 (fu_flashrom_device_get_programmer_name (flashrom_device),
		       "<my_supported_programmer>") != 0) {
		g_set_error_literal (error,
				     FWUPD_ERROR,
				     FWUPD_ERROR_NOT_SUPPORTED,
				     "invalid programmer");
```

They way I see it we can have a 1:1 mapping between flashrom programmer and FlashromDevice subclass by simply matching on the FlashromProgrammer quirk entry.

tari:
Currently the fwupd engine creates device instances, and because it doesn't automatically load quirks for I2C devices beyond the subsystem match (compared to USB devices where it will match the VID/PID automatically), by the time the plugin gets a device it's too late to specify the desired GType- the engine has already created an instance of the plugin's default device type.

This change is moving quirk matching for I2C earlier in the process and replacing the engine's default choice, so we can specify more than one GType for I2C devices in this plugin.

tari:
I've discussed these concerns with @djcampello and it seems like we can use multiple plugins to avoid complicating the flashrom one, but it may also want an API to wrap flashrom to ensure it's used safely. Fortunately that would be a temporary measure, as there is ongoing work (largely by [aklm@](https://review.coreboot.org/q/author:aklm%2540chromium.org)) that will also addres the safety issues around libflashrom but may take ~months to reach the point that the current issues are fixed.

I'm going to close this and #3207 and work on splitting the current `flashrom` plugin into two, then add the new device that I want to support in another plugin.

PR #3189, opened 2021-04-26
Use libproxy to get the system proxy setting for a given URL

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -15,6 +15,9 @@
 #ifdef HAVE_GIO_UNIX
 #include <gio/gunixfdlist.h>
 #endif
+#ifdef HAVE_LIBPROXY
+#include <proxy.h>
+#endif
 
 #include <fcntl.h>
 #include <string.h>
@@ -74,6 +77,9 @@ typedef struct {
 	CURL				*curl;
 	curl_mime			*mime;
 	struct curl_slist		*headers;
+#ifdef HAVE_LIBPROXY
+	pxProxyFactory			*libproxy_factory;
+#endif
 } FwupdCurlHelper;
 #endif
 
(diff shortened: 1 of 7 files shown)


hughsie:
@superm1 I'm not really super what the snap package requires... Could you have a look please? I could just turn off libproxy support there if you don't think it's relevant.

snap/snapcraft.yaml
```diff
@@ -231,6 +232,7 @@ parts:
       - libgpgme11
       - libjson-glib-1.0-0
       - libpolkit-gobject-1-0
+      - libproxy
```
superm1, line 235: I expect this should be `libproxy1v5`

libfwupd/fwupd-client.c
```diff
@@ -569,16 +601,10 @@ fwupd_client_curl_new (FwupdClient *self, GError **error)
...
-	if (http_proxy == NULL)
-		http_proxy = g_getenv ("http_proxy");
-	if (http_proxy == NULL)
-		http_proxy = g_getenv ("HTTP_PROXY");
-	if (http_proxy != NULL && strlen (http_proxy) > 0)
-		curl_easy_setopt (helper->curl, CURLOPT_PROXY, http_proxy);
+#ifdef HAVE_LIBPROXY
+	/* use libproxy to get system-wide proxy settings */
+	helper->libproxy_factory = px_proxy_factory_new ();
+#endif
```
superm1, line 607: I think this should be an `#else` to get all the stuff from environment if you don't have libproxy though, no?  Otherwise how can Windows ever work in an environment needing proxy?

superm1: oh nvm, it's done in the helper now, sorry for the noise.

superm1 (approved):
once the snap passes, LGTM.  I think it's just that package name I mentioned is wrong in it.

snap/snapcraft.yaml
hughsie: Will try, thanks!

PR #3201, opened 2021-04-28
trivial: Don't restart pesign in the example spec file

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
I guess this actually needs to go into fwupd-efi packaging too, huh?

hughsie:
> I guess this actually needs to go into fwupd-efi packaging too, huh?

Not sure; I'm trying to work out what they actually _do_.

PR #3207, opened 2021-04-29
flashrom: allow devices to control when the programmer is opened

tari:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

Some devices need better control over when the flashrom programmer is opened,
because opening the programmer and probing the flash may involve the device
changing mode, such as entering a bootloader mode which should be done in
detach() rather than open(). Add a flag for flashrom devices allowing them
to indicate if opening the device should automatically connect the programmer,
otherwise allowing them to explicitly connect at a device-controlled time.

This change supports a new device I'm currently working on support for, where connecting flashrom to it puts it into a bootloader mode in which its normal functionality is unavailable.

### DIFF
--- a/plugins/flashrom/fu-flashrom-device.c
+++ b/plugins/flashrom/fu-flashrom-device.c
@@ -16,11 +16,28 @@ typedef struct {
 	gsize				 flash_size;
 	struct flashrom_flashctx	*flashctx;
 	struct flashrom_programmer	*flashprog;
+	FuFlashromDeviceFlags 		flags;
 } FuFlashromDevicePrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE (FuFlashromDevice, fu_flashrom_device, FU_TYPE_UDEV_DEVICE)
 
 #define GET_PRIVATE(o) (fu_flashrom_device_get_instance_private (o))
+
+FuFlashromDeviceFlags
+fu_flashrom_device_get_flags (FuFlashromDevice *self)
+{
+	FuFlashromDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (FU_IS_FLASHROM_DEVICE (self), FU_FLASHROM_DEVICE_FLAG_NONE);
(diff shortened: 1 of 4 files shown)


plugins/flashrom/fu-flashrom-device.c
```diff
@@ -186,12 +218,39 @@ fu_flashrom_device_open (FuDevice *device, GError **error)
...
-fu_flashrom_device_close (FuDevice *device, GError **error)
+fu_flashrom_device_open (FuDevice *device, GError **error)
+{
+	FuFlashromDevice *self = FU_FLASHROM_DEVICE (device);
+
+	if ((fu_flashrom_device_get_flags (self)
+		& FU_FLASHROM_DEVICE_FLAG_OPEN_PROGRAMMER) != 0) {
+		if (!fu_flashrom_device_open_programmer (self, error))
+			return FALSE;
+	}
```
hughsie, line 229: this breaks the `internal` programmer, no?

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -152,6 +181,9 @@ fu_flashrom_device_open (FuDevice *device, GError **error)
 		return FALSE;
 	}
 	rc = flashrom_flash_probe (&priv->flashctx, priv->flashprog, NULL);
+	if (rc != 0)
+		/* clean up flashprog */
+		fu_flashrom_device_close_programmer (self);
```
hughsie, line 186: Doesn't fwupd call `close()` even if `open()` fails?

hughsie:
> where connecting flashrom to it puts it into a bootloader mode

Dude, this is really getting to the point where a custom fwupd plugin is going to be a **lot** less complex.

tari:
As noted in https://github.com/fwupd/fwupd/pull/3188#issuecomment-832361740, I'm going to try to avoid the complexity of multiple device support in this plugin by breaking it up and adding my new device in another plugin.

PR #3210, opened 2021-04-30
modem-manager: add generic support for PCI based modems

aleksander0m:
No longer rely on the modems being USB based, we can also support PCI based devices with the same protocols.

Type of pull request:
- [X ] Feature

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -144,6 +144,7 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 	guint n_ports = 0;
 	g_autoptr(MMFirmwareUpdateSettings) update_settings = NULL;
 	g_autofree gchar *device_sysfs_path = NULL;
+	g_autofree gchar *device_bus = NULL;
 
 	/* inhibition uid is the modem interface 'Device' property, which may
 	 * be the device sysfs path or a different user-provided id */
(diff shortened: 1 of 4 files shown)


plugins/modem-manager/fu-mm-device.c
```diff
@@ -285,15 +297,32 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
...
 			g_warning ("failed to set vendor ID: %s", error_local->message);
 		} else {
-			g_autofree gchar *vendor_id = g_strdup_printf ("USB:0x%s", g_strchomp (value));
-			fu_device_add_vendor_id (device, vendor_id);
+			guint64 value_int;
+
+			/* note: the string value may be prefixed with '0x' (e.g. when reading
+			 * the PCI 'vendor' attribute, or not prefixed with anything, as in the
+			 * USB 'idVendor' attribute. */
+			value_int = g_ascii_strtoull (value_str, NULL, 16);
```
hughsie, line 319: does fu_common_strtoull help you here?

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -12,59 +12,133 @@
 
 #include "fu-mm-utils.h"
 
+static gchar *
+find_device_bus_subsystem (GUdevDevice *device)
```
hughsie, line 16: does g_udev_device_get_parent_with_subsystem() work here instead?

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -12,59 +12,133 @@
...
+				break;
+			}
+
+			/* check next parent */
+			next = g_udev_device_get_parent (parent);
+			g_set_object (&parent, next);
 		}
+	}
+	/* PCI device */
+	else if (g_strcmp0 (device_bus, "PCI") == 0) {
```
hughsie, line 95: } on the else please

plugins/modem-manager/fu-plugin-modem-manager.c
```diff
@@ -149,21 +149,22 @@ fu_plugin_mm_udev_uevent_cb (GUdevClient	*udev,
...
 
 	/* ignore if loading port info fails */
-	if (!fu_mm_utils_get_udev_port_info (device, &device_sysfs_path, &ifnum, NULL))
+	if (!fu_mm_utils_get_udev_port_info (device, &device_bus, &device_sysfs_path, &ifnum, NULL))
 		return TRUE;
 
-	/* ignore all events for ports not owned by our device */
-	if (g_strcmp0 (device_sysfs_path, priv->inhibited->physical_id) != 0)
+	/* ignore non-USB and non-PCI events */
+	if ((g_strcmp0 (device_bus, "USB") != 0) && (g_strcmp0 (device_bus, "PCI") != 0))
```
hughsie, line 163: I don't think you need the extra braces.

plugins/modem-manager/fu-mm-utils.c
aleksander0m: I'm not totally convinced, because `g_udev_device_get_parent_with_subsystem()` would require us to run it one by one for all the bus subsystems we may expect, and then we would need to decide which of them is closer in the hierarchy to the port GUdevDevice (as e.g. a USB device may be hanging off PCI as well). This `find_device_bus_subsystem()` will instead iterate parents until the first subsystem of the expected buses is found (e.g. on a USB device hanging off PCI we would find USB first).

plugins/modem-manager/fu-mm-device.c
aleksander0m: No, it doesn't, because the `value_str` is always in base16, so we shouldn't guess the base. `fu_common_strtoull()` would read the string as base 10 if it doesn't have the `0x` prefix.

plugins/modem-manager/fu-mm-utils.c
aleksander0m: ok

hughsie (approved):

Issue #3197, opened 2021-04-27
fwupd crashes on startup


gwduvalljr:
Dell XPS 13 9310
Ubuntu 20.04
Crash Dump Attached

I am happy to provide anything else needed.

Error is:
```
fwupd.service - Firmware update daemon
     Loaded: loaded (/lib/systemd/system/fwupd.service; static; vendor preset: enabled)
     Active: failed (Result: core-dump) since Tue 2021-04-27 15:01:39 EDT; 5min ago
       Docs: https://fwupd.org/
    Process: 14289 ExecStart=/usr/libexec/fwupd/fwupd (code=dumped, signal=SEGV)
   Main PID: 14289 (code=dumped, signal=SEGV)

Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Failed to read oom_kill field of memory.events cgroup attribute: No such file or directory
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Child 14289 belongs to fwupd.service.
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Main process exited, code=dumped, status=11/SEGV
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Failed with result 'core-dump'.
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Service will not restart (restart setting)
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Changed start -> failed
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Job 4570 fwupd.service/start finished, result=failed
Apr 27 15:01:39 garlandXPS13 systemd[1]: Failed to start Firmware update daemon.
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Unit entered failed state.
Apr 27 15:01:39 garlandXPS13 systemd[1]: fwupd.service: Control group is empty.

```
[_usr_libexec_fwupd_fwupd.0.zip](https://github.com/fwupd/fwupd/files/6387021/_usr_libexec_fwupd_fwupd.0.zip)

hughsie:
You didn't fill in any of the requested details in the issue template. Backtrack please.

gwduvalljr:
When I first clicked "New Issue" I wasn't shown a template. Just now found them. Interesting. Anyway...

**Describe the bug**
fwupd crashes on start

**Steps to Reproduce**
[...]

PR #3212, opened 2021-04-30
Check the versions of libfwupd and libfwupdplugin at startup

hughsie:
This prevents super-hard-to-debug crashes like we saw in #3197 where the user
was mixing PPAs and official versions.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1 (approved):

superm1:
I think this approach works fine, but a "lighter" one "might" be to just do debian/control depending upon binary Version.

I think the lighter approach would mean double checking debian policy its allowed. This heavier one I think will cover better though.

PR #3213, opened 2021-04-30
Wip/hughsie/win32

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -593,21 +593,26 @@ fu_plugin_quirks_device_func (void)
 static void fu_common_kernel_lockdown_func (void)
 {
 	gboolean ret;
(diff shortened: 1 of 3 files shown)


hughsie:
Also needs https://github.com/hughsie/libxmlb/pull/89

hughsie:
Frustratingly I can't reproduce the `ci/circleci: build-windows` failure, even in a new VM of f34 with the same deps as specified in the circleci yml file.

superm1:
> Frustratingly I can't reproduce the `ci/circleci: build-windows` failure, even in a new VM of f34 with the same deps as specified in the circleci yml file.

that is a very frustrating.  To me I an suspecting some sort of missing configuration of a path or so that is not getting set up in the minimal container but that a VM would setup.
can you try to do it in a docker container perhaps?

superm1 (approved):

PR #3216, opened 2021-04-30
Fix wacom_usb plugin

jigpu:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

This set of commits fixes some issues with the wacom_usb plugin, e.g. an inability to flash and incorrect parsing of the version numbers. It also adds an additional assertion to prevent a class of issues in the future.


hughsie (approved):
Looks great, thanks!

Issue #3187, opened 2021-04-25
fwupd automatically claims some USB devices and attempts bulk transfers, thus preventing programs like fastboot from working


sandy-8925:
**Describe the bug**
In recent versions of fwupd (probably February/March onwards), fwupd automatically connects to USB devices and seems to attempt USB bulk transfers. However, this fails for some devices (e.g Android device in fastboot mode), and this prevents other programs like fastboot from working.

**Steps to Reproduce**
- Run fwupd
- Following error is observed in journalctl logs: 
`fwupd[2248]: 21:08:27:0117 FuEngine             failed to add device usb:03:00:04: failed to do bulk transfer: USB error on device 8087:0fff : Input/Output Error [-1]`

- After a few seconds, attempt to use fastboot to flash an image to an Android device partition (e.g fastboot flash boot boot.img)
- fastboot is stuck in disk sleep mode
- Most likely, fwupd failed to release the USB device after whatever it was trying to do failed

**Expected behavior**
Fastboot should succeed, because fwupd shouldn't try to connect to random devices and leave them hanging.

**fwupd version information**
client version: 1.5.9
compile-time dependency versions
        gusb:   0.3.6

daemon version: 1.5.9

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
`pacman`

**Additional questions**
- Operating system and version: Arch Linux, latest software
- Have you tried rebooting? Yes

hughsie:
> failed to do bulk transfer

This comes from `plugins/fastboot/fu-fastboot-device.c` -- which certainly does `g_usb_device_release_interface()` on `->close`.

> fwupd shouldn't try to connect to random devices

We're explictly connecting to fastboot devices specified with `USB\CLASS_FF&SUBCLASS_42&PROT_03`

>  and leave them hanging

When you say that `fastboot is stuck in disk sleep mode` what does that mean? The device is in deep sleep (as in no USB interface) or is fwupd not disconnecting from fastboot properly? We do query the device as soon as it appears in fastboot mode, but only the common stuff like `product`, `serialno` etc.

sandy-8925:
fwupd is not disconnecting from fastboot (on the device) properly, and when I attempt to run a fastboot command from my computer, the fastboot process on my computer was stuck in disk sleep mode, because it was waiting for fwupd to relinquish it's claim on the device.

This is very frustrating, because fastboot commands just failed with no clue as to what was going on.

[...]

PR #3218, opened 2021-05-01
Force the device locker to ->close() an aborted ->open()

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3187

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device-locker.c
+++ b/libfwupdplugin/fu-device-locker.c
@@ -202,8 +202,14 @@ fu_device_locker_new_full (gpointer device,
 	self->close_func = close_func;
 
 	/* open device */
-	if (!self->open_func (device, error))
+	if (!self->open_func (device, error)) {
+		g_autoptr(GError) error_local = NULL;
+		if (!self->close_func (device, &error_local)) {
+			g_debug ("ignoring close error on aborted open: %s",
+				 error_local->message);
+		}
 		return NULL;
+	}
 
 	/* success */
 	self->device_open = TRUE;
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -695,17 +695,19 @@ fu_device_locker_func (void)
 }
 
 static gboolean
-_fail_open_cb (GObject *device, GError **error)
+_fail_open_cb (FuDevice *device, GError **error)
 {
+	fu_device_set_metadata_boolean (device, "Test::Open", TRUE);
 	g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED, "fail");
 	return FALSE;
 }
 
 static gboolean
-_fail_close_cb (GObject *device, GError **error)
+_fail_close_cb (FuDevice *device, GError **error)
 {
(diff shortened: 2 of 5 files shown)


hughsie:
100% untested.

libfwupdplugin/fu-device-locker.c
```diff
@@ -202,8 +202,14 @@ fu_device_locker_new_full (gpointer device,
 	self->close_func = close_func;
 
 	/* open device */
-	if (!self->open_func (device, error))
+	if (!self->open_func (device, error)) {
+		g_autoptr(GError) error_local = NULL;
+		if (!self->close_func (device, &error_local)) {
+			g_debug ("ignoring close error on aborted open: %s",
+				 error_local->message);
+		}
```
superm1, line 210: devil's advocate here - why shouldn't this be part of finalizing the device?

hughsie: I did think about that, but then we also encourage retrying an action if it fails. e.g. if we retry ->open 3 times, then fail in setup() (in the last step of `fu_device_open()`, possibly after doing something like opening a USB device) and finalize, we'd expect to get 3 ->close's rather than just one.

libfwupdplugin/fu-self-test.c
```diff
@@ -695,28 +695,34 @@ fu_device_locker_func (void)
 }
 
 static gboolean
-_fail_open_cb (GObject *device, GError **error)
+_fail_open_cb (FuDevice *device, GError **error)
 {
+	fu_device_set_metadata_boolean (device, "isLocked", TRUE);
```
superm1, line 700: Shouldn't this be heavier than a metadata item? I feel like it should be a private property with methods to call.

hughsie: `FU_DEVICE_INTERNAL_FLAG_LOCKED` seems sane, no?

superm1: seems good

hughsie: > FU_DEVICE_INTERNAL_FLAG_LOCKED seems sane, no?

Turns out, the answer was indeed: **no**. Let me explain: The "is-locked" is a property of the locker, not of the device. It's quite valid to (and plugins do) call something like:

    locker1 = fu_device_locker_new_full (device, _open_cb, _close_cb, error)
    locker2 = fu_device_locker_new_full (device, _detach_cb, _attach_cb, error)

So if we set the flag on the device for the first locker, then it would be invalid for the second. I then pondered `fu_device_locker_is_open()` but all that really tells you is whether `fu_device_locker_close` has been manually called, as if the `->setup()` failed in the locker open the locker returns NULL. Notably, we also call `->close_func` on locker finalize, so we could have just moved up the `self->device_open = TRUE;` but in this case I think handling it locally is better as we get the `g_debug("ignoring close error on aborted open")` rather than the `g_warning()` on finalize.

hughsie: I did the pragmatic thing. New version below.

PR #3220, opened 2021-05-03
plugin: rts54hub: rtd21xx add new update mode

AnyProblem:
Signed-off-by: RickyWu <ricky_wu@realtek.com>

Type of pull request:
- [ ] Code fix
- [ *] Feature
- [ ] Documentation

### DIFF
--- a/plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
+++ b/plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
@@ -18,25 +18,45 @@ struct _FuRts54hubRtd21xxDevice
 	guint8			 target_addr;
 	guint8			 i2c_speed;
 	guint8			 register_addr_len;
+	guint8			 fwupdate_mode;
 };
 
 G_DEFINE_TYPE (FuRts54hubRtd21xxDevice, fu_rts54hub_rtd21xx_device, FU_TYPE_DEVICE)
 
 #define I2C_DELAY_AFTER_SEND		5000	/* us */
 
-#define UC_FOREGROUND_SLAVE_ADDR	0x3A
+#define UC_ISP_SLAVE_ADDR		0x3A
 #define UC_FOREGROUND_STATUS		0x31
 #define UC_FOREGROUND_OPCODE		0x33
 #define UC_FOREGROUND_ISP_DATA_OPCODE	0x34
-
-#define ISP_DATA_BLOCKSIZE		256
+#define ISP_FOREGROUND_DATA_BLOCKSIZE	256
 #define ISP_PACKET_SIZE			257
+/* Background Cmd */
+#define UC_BACKGROUND_OPCODE		0x31
+#define UC_BACKGROUND_ISP_DATA_OPCODE	0x32
+#define ISP_BACKGROUND_DATA_BLOCKSIZE	32
 
 typedef enum {
 	ISP_STATUS_BUSY			 = 0xBB,	/* host must wait for device */
 	ISP_STATUS_IDLE_SUCCESS		 = 0x11,	/* previous command was OK */
 	ISP_STATUS_IDLE_FAILURE		 = 0x12,	/* previous command failed */
 } IspStatus;
+
+typedef enum {
+    FW_UPDATE_FOREGROUND_MODE = 0x01,
+    FW_UPDATE_BACKGROUND_MODE,
+    FW_UPDATE_LAST_MODE,
+} FuRt21xxFwUpdaMode;
+
+typedef enum {
+	ISP_BG_CMD_FW_UPDATE_START		= 0x01,
+	ISP_BG_CMD_FW_UPDATE_ISP_DONE	= 0x02,
+	ISP_BG_CMD_GET_FW_INFO		= 0x03,
+	ISP_BG_CMD_FW_UPDATE_EXIT		= 0x04,
+	ISP_BG_CMD_GET_PROJECT_ID_ADDR	= 0x05,
+	ISP_BG_CMD_SYNC_IDENTIFY_CODE	= 0x06,
+
+} IsBackgroundCmd;
 
 typedef enum {
 	ISP_CMD_ENTER_FW_UPDATE		= 0x01,
@@ -47,7 +67,7 @@ typedef enum {
 	ISP_CMD_FW_UPDATE_ISP_DONE	= 0x06,
 	ISP_CMD_FW_UPDATE_EXIT		= 0x07,
 	ISP_CMD_FW_UPDATE_RESET		= 0x08,
-} IspCmd;
+} IspForegroundCmd;
 
 typedef enum {
 	VENDOR_CMD_DISABLE		= 0x00,
@@ -62,6 +82,7 @@ fu_rts54hub_rtd21xx_device_to_string (FuDevice *module, guint idt, GString *str)
 	fu_common_string_append_kx (str, idt, "TargetAddr", self->target_addr);
 	fu_common_string_append_kx (str, idt, "I2cSpeed", self->i2c_speed);
 	fu_common_string_append_kx (str, idt, "RegisterAddrLen", self->register_addr_len);
+	fu_common_string_append_kx (str, idt, "Rts21xxFwUpdateMode", self->fwupdate_mode);
 }
 
 static FuRts54HubDevice *
@@ -118,6 +139,20 @@ fu_rts54hub_rtd21xx_device_set_quirk_kv (FuDevice *device,
 		guint64 tmp = fu_common_strtoull (value);
 		if (tmp <= 0xff) {
 			self->register_addr_len = tmp;
+			return TRUE;
+		}
+		g_set_error_literal (error,
+				     G_IO_ERROR,
+				     G_IO_ERROR_INVALID_DATA,
+				     "invalid register address length");
+		return FALSE;
+	}
+
+	/* load fwupdate mode from quirks */
+	if (g_strcmp0 (key, "Rts21xxFwUpdateMode") == 0) {
+		guint64 tmp = fu_common_strtoull (value);
+		if (tmp < FW_UPDATE_LAST_MODE) {
+			self->fwupdate_mode = tmp;
 			return TRUE;
 		}
 		g_set_error_literal (error,
@@ -193,6 +228,33 @@ fu_rts54hub_rtd21xx_device_i2c_read (FuRts54hubRtd21xxDevice *self,
 	return TRUE;
 }
 
+static void
+fu_rts54hub_rtd21xx_device_exit_background(FuRts54hubRtd21xxDevice *self, GError **error)
+{
+
+	guint8 buf[] = { ISP_BG_CMD_FW_UPDATE_EXIT };
+
+	if (!fu_rts54hub_rtd21xx_device_i2c_write (self,
+						   UC_ISP_SLAVE_ADDR,
+						   UC_BACKGROUND_OPCODE,
+						   buf, sizeof(buf),
+						   error))
+		g_prefix_error (error, "failed to ISP_BG_CMD_FW_UPDATE_EXIT: ");
+}
+
+static void
+fu_rts54hub_rtd21xx_device_exit_foreground(FuRts54hubRtd21xxDevice *self, GError **error)
+{
+	guint8 buf[] = { ISP_CMD_FW_UPDATE_EXIT };
+
+	if (!fu_rts54hub_rtd21xx_device_i2c_write (self,
+						   UC_ISP_SLAVE_ADDR,
+						   UC_FOREGROUND_OPCODE,
+						   buf, sizeof(buf),
+						   error))
+		g_prefix_error (error, "failed to ISP_BG_CMD_FW_UPDATE_EXIT: ");
+}
+
 static gboolean
 fu_rts54hub_rtd21xx_device_read_status_raw (FuRts54hubRtd21xxDevice *self,
 					    guint8 *status,
@@ -200,7 +262,7 @@ fu_rts54hub_rtd21xx_device_read_status_raw (FuRts54hubRtd21xxDevice *self,
 {
 	guint8 buf = 0x00;
 	if (!fu_rts54hub_rtd21xx_device_i2c_read (self,
-						  UC_FOREGROUND_SLAVE_ADDR,
+						  UC_ISP_SLAVE_ADDR,
 						  UC_FOREGROUND_STATUS,
 						  &buf, sizeof(buf),
 						  error))
@@ -226,6 +288,14 @@ fu_rts54hub_rtd21xx_device_read_status_cb (FuDevice *device,
 				"status was 0x%02x", status);
 		return FALSE;
 	}
+	else if(status == ISP_STATUS_IDLE_FAILURE)
+	{
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "status was 0x%02x", status);
+		return FALSE;
+	}
 	return TRUE;
 }
 
@@ -240,15 +310,16 @@ fu_rts54hub_rtd21xx_device_read_status (FuRts54hubRtd21xxDevice *self,
 }
 
 static gboolean
-fu_rts54hub_rtd21xx_ensure_version_unlocked (FuRts54hubRtd21xxDevice *self,
+fu_rts54hub_rtd21xx_ensure_version_foreground (FuRts54hubRtd21xxDevice *self,
 					     GError **error)
 {
 	guint8 buf_rep[7] = { 0x00 };
 	guint8 buf_req[] = { ISP_CMD_GET_FW_INFO };
 	guint8 buf[] = { ISP_CMD_FW_UPDATE_RESET };
 	g_autofree gchar *version = NULL;
+
 	if (!fu_rts54hub_rtd21xx_device_i2c_write (self,
-						   UC_FOREGROUND_SLAVE_ADDR,
+						   UC_ISP_SLAVE_ADDR,
 						   UC_FOREGROUND_OPCODE,
 						   buf_req, sizeof(buf_req),
 						   error)) {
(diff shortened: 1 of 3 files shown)


plugins/rts54hub/rts54hub.quirk
```diff
@@ -21,3 +21,22 @@ FirmwareSize = 0x30000
...
+Children = FuRts54hubRtd21xxDevice|USB\VID_0BDA&PID_5420&I2C_01
+[USB\VID_0BDA&PID_5420&I2C_01]
+Plugin = rts54hub
+Name = HDMI Converter1
+Flags = updatable
+FirmwareSize = 0x70000
+Rts54TargetAddr = 0x20
+Rts54I2cSpeed = 0x2
+Rts54RegisterAddrLen = 0x04
+Rts21xxFwUpdateMode = 2
```
superm1, line 42: is there a minimum firmware version necessary for this feature to work?  Does it need to be probed to check?

plugins/rts54hub/rts54hub.quirk
```diff
@@ -21,3 +21,22 @@ FirmwareSize = 0x30000
 Rts54TargetAddr = 0x20
 Rts54I2cSpeed = 0x2
 Rts54RegisterAddrLen = 0x04
+Rts21xxFwUpdateMode = 1
```
superm1, line 24: Is this always going to be a binary decision of 1 or 2?  If so, why not just having something like `Rts21xxBackgroundUpdate=False`

plugins/rts54hub/rts54hub.quirk
```diff
@@ -21,3 +21,22 @@ FirmwareSize = 0x30000
...
+[USB\VID_0BDA&PID_5420]
+Plugin = rts54hub
+GType = FuRts54HubDevice
+Vendor = Lenovo
+FirmwareSizeMin = 0x20000
+FirmwareSizeMax = 0x70000
+Children = FuRts54hubRtd21xxDevice|USB\VID_0BDA&PID_5420&I2C_01
+[USB\VID_0BDA&PID_5420&I2C_01]
+Plugin = rts54hub
+Name = HDMI Converter1
```
superm1, line 36: seems like an extra 1 at end

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -388,56 +522,245 @@ fu_rts54hub_rtd21xx_device_write_firmware (FuDevice *device,
...
 						  error)) {
 		g_prefix_error (error, "failed to read project ID: ");
+		fu_rts54hub_rtd21xx_device_exit_foreground(self, error);
 		return FALSE;
 	}
 	if (read_buf[0] != ISP_STATUS_IDLE_SUCCESS) {
 		g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA,
 			     "failed project ID with error 0x%02x: ",
 			     read_buf[0]);
+		fu_rts54hub_rtd21xx_device_exit_foreground(self, error);
```
superm1, line 763: style nit - make sure that you have space before the `(` (This is all over this change you made this error)

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -193,14 +228,41 @@ fu_rts54hub_rtd21xx_device_i2c_read (FuRts54hubRtd21xxDevice *self,
...
+
+	if (!fu_rts54hub_rtd21xx_device_i2c_write (self,
+						   UC_ISP_SLAVE_ADDR,
+						   UC_BACKGROUND_OPCODE,
+						   buf, sizeof(buf),
+						   error))
+		g_prefix_error (error, "failed to ISP_BG_CMD_FW_UPDATE_EXIT: ");
+}
+
+static void
```
superm1, line 245: if this is a void function, why are you passing error?  It seems like it's just going to cause double use of error variable.  I would suggest make this boolean and the use local error everywhere or make it void and don't have an error variable.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -226,6 +288,14 @@ fu_rts54hub_rtd21xx_device_read_status_cb (FuDevice *device,
 				"status was 0x%02x", status);
 		return FALSE;
 	}
+	else if(status == ISP_STATUS_IDLE_FAILURE)
+	{
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "status was 0x%02x", status);
```
superm1, line 296: I think this can be a literal error, you're matching a single value

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -18,26 +18,46 @@ struct _FuRts54hubRtd21xxDevice
 	guint8			 target_addr;
 	guint8			 i2c_speed;
 	guint8			 register_addr_len;
+	guint8			 fwupdate_mode;
```
superm1, line 21: I suggest making this boolean.  For example `gboolean background_update`

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -541,6 +874,22 @@ fu_rts54hub_rtd21xx_device_write_firmware (FuDevice *device,
...
+					   GError **error)
+{
+
+	FuRts54hubRtd21xxDevice *self = FU_RTS54HUB_RTD21XX_DEVICE (device);
+	if(self->fwupdate_mode == FW_UPDATE_BACKGROUND_MODE)
+		return fu_rts54hub_rtd21xx_device_write_firmware_background(device, firmware, flags, error);
+	else
+		return fu_rts54hub_rtd21xx_device_write_firmware_foreground(device, firmware, flags, error);
+	return TRUE;
+}
```
superm1, line 891: this else block seems redundant.  Might I suggest a different way?


```
return (self->fwupdate_mode == FW_UPDATE_BACKGROUND_MODE) ? fu_rts54hub_rtd21xx_device_write_firmware_background (..) : fu_rts54hub_rtd21xx_device_write_firmware_foreground (...);
```

That will be more readable

plugins/rts54hub/rts54hub.quirk
AnyProblem: This feature will controlled by project ID in rts54hub.quirk
if the new project need background update will add the project ID and Rts21xxFwUpdateMode together to rts54hub.quirk

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -388,56 +525,245 @@ fu_rts54hub_rtd21xx_device_write_firmware (FuDevice *device,
...
-	if (locker == NULL)
+	if (locker == NULL) {
+		return fu_rts54hub_rtd21xx_device_exit_background (self, error);
+		return FALSE;
+	}
+
+	/* simple image */
+	fw = fu_firmware_get_bytes (firmware, error);
+	if (fw == NULL) {
+		fu_rts54hub_rtd21xx_device_exit_background (self, error);
```
hughsie, line 536: we can't just add these in here -- it's all very chaotic. Can you describe what foreground and background modes actually are? Could we do the same in attach and detach and avoid all these sprinkled calls?

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -282,15 +356,78 @@ fu_rts54hub_rtd21xx_ensure_version_unlocked (FuRts54hubRtd21xxDevice *self,
...
+						   UC_ISP_SLAVE_ADDR,
+						   UC_BACKGROUND_OPCODE,
+						   buf_req, sizeof(buf_req),
+						   error)) {
+		g_prefix_error (error, "failed to get version number: ");
+		return FALSE;
+	}
+
+	/* wait for device ready */
+	g_usleep (300000);
```
superm1, line 378: would this be better done as a retry until it works rather than fixed timeout?

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -282,15 +356,78 @@ fu_rts54hub_rtd21xx_ensure_version_unlocked (FuRts54hubRtd21xxDevice *self,
...
+
 static gboolean
 fu_rts54hub_rtd21xx_device_detach_raw (FuRts54hubRtd21xxDevice *self, GError **error)
 {
 	guint8 buf = 0x03;
+
+	if (self->fwupdate_mode == FW_UPDATE_BACKGROUND_MODE)
+		buf = 0x02;
+	else
+		buf = 0x03;
```
superm1, line 423: you seem to set this default in the initialization.  I would suggest pick one or the other.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -388,56 +525,245 @@ fu_rts54hub_rtd21xx_device_write_firmware (FuDevice *device,
...
+	if (locker == NULL) {
+		fu_rts54hub_rtd21xx_device_exit_foreground (self, error);
 		return FALSE;
+	}
 
 	/* simple image */
 	fw = fu_firmware_get_bytes (firmware, error);
-	if (fw == NULL)
+	if (fw == NULL) {
+		fu_rts54hub_rtd21xx_device_exit_foreground (self, error);
```
superm1, line 715: can't this be done on device close instead?

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: There's no error. Status vaue is ISP_STATUS_BUSY(0xBB) or ISP_STATUS_IDLE_FAILURE(0x12) should return FASLE.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: The "guint8"  has better expansibility. If there is another burning method, it can also be supported.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: After sending the command, it takes time to wait for interrupt reception and FW processing to complete. 
Removing delay and directly using retry will cause the interrupt to be triggered all the time, resulting in incorrect reading value

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: OK ~~

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
hughsie: I think it should be `FuRt21xxFwUpdaMode` then, rather than guint8

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
hughsie: I'm very unclear what all the foreground/background stuff does TBH.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: Background mode means that the device can work normally in the process of fw update. After burning, the device cannot be restarted.
Foreground mode means the device can't work  in the process of fw update, after FW Update it will restart. If  add exit process  to attach, you need to judge whether FW update has been performed or not, If there is a burn, you need to send resetcommand. If you do not perform burn, you need to send the command to exit FW update mode.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: hi @hughsie ， Background/foreground using different Command int the process of fw update.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: yes , FuRt21xxFwUpdaMode  is better. 
We'll  use FuRt21xxFwUpdaMode instead. thanks.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
superm1: If that's the case - this is basically the meaning of [FWUPD-DEVICE-FLAG-USABLE-DURING-UPDATE](https://fwupd.github.io/fwupd-fwupd-enums.html#FWUPD-DEVICE-FLAG-USABLE-DURING-UPDATE:CAPS).  Just drop the private member and set/use that flag in the plugin.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
superm1: thanks for clarification.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
nwj1027: OK

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
superm1: But in this block you explicitly match the case `ISP_STATUS_IDLE_FAILURE`, so just use `g_set_error_literal` instead of `g_set_error`

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -368,9 +447,54 @@ fu_rts54hub_rtd21xx_device_detach (FuDevice *device, GError **error)
...
+	} else {
+		(fu_device_has_flag (device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) ?
+			(buf[0] = ISP_CMD_FW_UPDATE_EXIT) : (buf[0] = ISP_CMD_FW_UPDATE_RESET);
+
+		if (!fu_rts54hub_rtd21xx_device_i2c_write (self,
+							   UC_ISP_SLAVE_ADDR,
+							   UC_FOREGROUND_OPCODE,
+							   buf, sizeof (buf),
+							   error))
+			g_prefix_error (error, "failed to failed to attach: : ");
```
superm1, line 481: You're missing the `FALSE` return here.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -368,9 +447,54 @@ fu_rts54hub_rtd21xx_device_detach (FuDevice *device, GError **error)
...
+	if (locker == NULL)
+		return FALSE;
+	if(self->fwupdate_mode == FW_UPDATE_BACKGROUND_MODE){
+		buf[0] = ISP_BG_CMD_FW_UPDATE_EXIT ;
+		if (!fu_rts54hub_rtd21xx_device_i2c_write (self,
+							   UC_ISP_SLAVE_ADDR,
+							   UC_BACKGROUND_OPCODE,
+							   buf, sizeof (buf),
+							   error))
+			g_prefix_error (error, "failed to attach: ");
```
superm1, line 469: You're missing the `FALSE` return here.

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -395,14 +519,163 @@ fu_rts54hub_rtd21xx_device_write_firmware (FuDevice *device,
...
+	}
+
+	/* exit background-fw mode */
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
+	fu_device_set_progress (device, 0);
+	if (!fu_rts54hub_rtd21xx_device_read_status (self, NULL, error))
+		return FALSE;
+
+	/* success */
+	return TRUE;
```
superm1, line 641: nit: suggested code simplification `fu_rts54hub_rtd21xx_device_read_status (self, NULL, error)`

plugins/rts54hub/fu-rts54hub-rtd21xx-device.c
```diff
@@ -395,14 +519,163 @@ fu_rts54hub_rtd21xx_device_write_firmware (FuDevice *device,
...
+						   UC_BACKGROUND_OPCODE,
+						   write_buf, 1,
+						   error)) {
+		g_prefix_error (error, "failed update finish cmd: ");
+		return FALSE;
+	}
+
+	/* exit background-fw mode */
+	fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
+	fu_device_set_progress (device, 0);
```
superm1, line 636: shouldn't this go to 100% at the end?

nwj1027: Yes ，it should be 100%.

nwj1027: hi @superm1 ， 
for Background fwupdate, it won’t reset after finish wirting fw. The new FW will not run until it is replug again. 
Shoud I Remove the “fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);” and add 	“fu_device_add_flag (device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);”

superm1: Wait for replug will mean that it's going to disappear and reappear on it's own.  Does the device reboot on it's own, or it's next power cycle?

nwj1027: Next power cycle. The device will not reboot on it's own.

hughsie:
@AnyProblem alternate proposal in https://github.com/fwupd/fwupd/pull/3235 -- I think that's a lot clearer having a new GType.

PR #3225, opened 2021-05-05
Switch lock directory from /var/run to /run/lock

djcampello:
This allows for proper locking between executions of fwupdtool in
Chromium OS minijail environment. It is needed after commit
9cf5f8f7ff7b93473eabf21e8482bb8feace039a was introduced.

Type of pull request:
- [ X ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1253,6 +1253,9 @@ fu_common_get_path (FuPathKind path_kind)
 			return g_build_filename (tmp, NULL);
 		basedir = fu_common_get_path (FU_PATH_KIND_LOCALSTATEDIR);
 		return g_build_filename (basedir, "cache", PACKAGE_NAME, NULL);
+	/* /run/lock */
+	case FU_PATH_KIND_LOCKDIR:
+		return g_strdup ("/run/lock");
 	case FU_PATH_KIND_OFFLINE_TRIGGER:
 		tmp = g_getenv ("FWUPD_OFFLINE_TRIGGER");
 		if (tmp != NULL)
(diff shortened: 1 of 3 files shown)


meson_options.txt
```diff
@@ -40,6 +40,7 @@ option('elogind', type : 'boolean', value : false, description : 'enable elogind
 option('tests', type : 'boolean', value : true, description : 'enable tests')
 option('soup_session_compat', type : 'boolean', value : true, description : 'enable SoupSession runtime compatibility support')
 option('curl', type : 'boolean', value : true, description : 'enable libcurl support')
+option('lockdir', type: 'string', value: '', description: 'Directory for fwupdtool lock')
```
hughsie, line 43: does ChromeOS use something different to /run/lock?

djcampello: no, do you want me to remove the meson option and hard code it?

hughsie: I think that's best; I think /run/lock is pretty standard.

djcampello: Done

djcampello: Should I remove the FU_PATH_KIND_LOCKDIR too?

hughsie: I'm not sure. I think it makes sense to keep, in case the snap needs to override it. @superm1

superm1: Isn't /var/run a symlink these days?

djcampello: It is, but I got complains about using it. Also, for the minijail sandbox to be effective, the lock should exist under /run/lock (or /var/run/lock) instead of plain /run so the tool does not have access to the whole contents of /run... crrev.com/c/2873506 for context

hughsie (approved):

djcampello:
Thank you!

PR #3229, opened 2021-05-07
chromeos build flag for conditionally-compiled code

hardboprobot:
Add a meson option to enable ChromeOS-specific code that is compiled-out
depending on a preprocessor macro.

Fix fwupd_get_os_release to support ChromeOS, as it doesn't have
os-release files.

Currently, any fwupdtool operation that starts the engine with the FU_ENGINE_LOAD_FLAG_REMOTES flag fails in ChromeOS:

```
$ fwupdtool get-devices
Failed to load remotes: No os-release found
```
This is a way (not the only one) to fix that. Maybe the ifdef'd code can be in fu-engine.c instead.
I chose OS_CHROMEOS as a macro based on other examples from the ChromeOS ecosystem: https://chromium.googlesource.com/chromiumos/third_party/libcamera/+/185574b741f0eca6c6225f4dec6be847aa73738f%5E%21/#F0

Type of pull request:
- [x] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-common.c
+++ b/libfwupd/fwupd-common.c
@@ -153,7 +153,12 @@ GHashTable *
 fwupd_get_os_release (GError **error)
 {
 	const gchar *filename = NULL;
-	const gchar *paths[] = { "/etc/os-release", "/usr/lib/os-release", NULL };
+	const gchar *paths[] = {
+		"/etc/os-release",
+		"/usr/lib/os-release",
+		"/etc/lsb-release",
+		NULL, /* last entry */
+	};
 	g_autofree gchar *buf = NULL;
 	g_auto(GStrv) lines = NULL;
 	g_autoptr(GHashTable) hash = NULL;


superm1:
I would say find another way to detect chrome os and don't add a meson option. This is a > pretty heavy way to do things. 

An extra fall back that finds an extra chrome specific path that complied for everyone seems fine to me.

djcampello:
Could you check if this error is happening on 1.6.0 ? I cannot reproduce on my side. Also note that fwupdtool runs on a sandboxed environment and it won't have access to any of these files: https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/third_party/chromiumos-overlay/sys-apps/fwupd/files/fwupdtool-update.conf;l=26

djcampello:
Could yo add verbose output so I can see at what stage we differ?

djcampello:
Actually my bad... /etc is in the sandbox and it most be allowed so the tool can read the fwupd configuration

```
localhost ~ # minijail0 --uts -e -l -p -N -v -P /mnt/empty -b / -b /proc -r -t -b /dev,,1 -k run,/run,tmpfs -k /var,/var,tmpfs -b /var/cache/fwupd,,1 -b /var/lib/fwupd,,1 -b /run/lock,,1 -b /sys,,1 -u fwupd -g fwupd -G /usr/bin/find / -name fwupd 2>/dev/null
/var/lib/fwupd
/var/cache/fwupd
/usr/libexec/fwupd
/usr/libexec/fwupd/fwupd
/usr/share/fwupd
/etc/fwupd
/etc/pki/fwupd
```

Also, I do have /etc/os-release on my system.

djcampello:
I'm actually not sure if the file is expected to be missing on some environments.. It seems to me that you should have it: https://chromium.googlesource.com/chromiumos/docs/+/HEAD/os_config.md#os-release

hughsie:
> I think we can scrap this PR then

Okay, feel free to re-open if things change. I certainly think something like `README-chromeos.md` upstream would make sense to document all of this stuff.

djcampello:
Circling back on this. /etc/os-release is only currently available on official Chrome OS builds and not on Chromium OS builds which might be the reason with @hardboprobot was hitting this problem

superm1:
Is there a better way to detect on chromium builds? Or can you change that so it's part of base layout even in chromium?

hughsie:
I really think this should be part of Chromium builds too...

djcampello:
This was attempted but it currently causes issues and had to be reverted: crrev.com/c/2892356

djcampello:
It seems that it might be relanded soon... I'll keep you updated here! https://bugs.chromium.org/p/chromium/issues/detail?id=1208870#c20

PR #3237, opened 2021-05-09
Add support for the PHAT table

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
100% untested with an actual PHAT blob. I'll wait to try this on actual hardware before marking it ready for review.

superm1:
> 100% untested with an actual PHAT blob. I'll wait to try this on actual hardware before marking it ready for review.

Presumably whatever prompted this - they should be able to provide a demo blob from a real system?

hughsie:
> they should be able to provide a demo blob from a real system

Ya, Intel are sorting it out.

plugins/acpi-phat/README.md
```diff
@@ -0,0 +1,20 @@
+Platform Health Assessment Table
```
superm1, line 1: I think you want to actually link back to the ACPI spec (IE here: https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html#platform-health-assessment-table)

hughsie:
> <offset>0x54b</offset>

I think all of these should probably disappear from the report attribute too.

plugins/acpi-phat/README.md
hughsie: Fixed.

hughsie:
> I think all of these should probably disappear from the report attribute too.

I did this too.

PR #3239, opened 2021-05-10
Wip/hughsie/gtk doc stage1

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
The backstory here is that I'm trying to migrate away from gtk-doc to gi-docgen and the latter is a lot more strict, but produces beautiful docs. We can't really migrate until there's a devhelp release with this commit https://gitlab.gnome.org/GNOME/devhelp/-/merge_requests/20 so phase2 might be a few more months.

PR #3241, opened 2021-05-10
Port from gktdoc to gi-docgen

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -1,12 +1,82 @@
-gnome.gtkdoc(
-  'fwupd',
-  src_dir : [
-    join_paths(meson.source_root(), 'libfwupd'),
-    join_paths(meson.source_root(), 'libfwupdplugin'),
-    join_paths(meson.build_root(), 'libfwupd'),
-    join_paths(meson.build_root(), 'libfwupdplugin'),
-    join_paths(meson.current_source_dir()),
-  ],
-  main_xml : 'fwupd-docs.xml',
-  install : true
-)
+if gidocgen.found()
+  toml_conf = configuration_data()
+  toml_conf.set('version', meson.project_version())
+
+  fwupd_toml = configure_file(
+    input: 'fwupd.toml.in',
+    output: 'fwupd.toml',
+    configuration: toml_conf
+  )
+
+  fwupdplugin_toml = configure_file(
+    input: 'fwupdplugin.toml.in',
+    output: 'fwupdplugin.toml',
+    configuration: toml_conf
+  )
+
+  custom_target('doc-fwupd',
+    input: [
+      fwupd_toml,
+      fwupd_gir[0],
+    ],
+    output: 'libfwupd',
+    command: [
+      gidocgen,
+      'generate',
+      '--quiet',
+      '--add-include-path=@0@'.format(meson.current_build_dir() / '../libfwupd'),
+      '--config=@INPUT0@',
+      '--output-dir=@OUTPUT@',
+      '--no-namespace-dir',
+      '--content-dir=@0@'.format(meson.current_source_dir()),
+      '@INPUT1@',
+    ],
+    depends: [
+      fwupd_gir[0],
+    ],
+    build_by_default: true,
+    install: true,
+    install_dir: join_paths(datadir, 'doc', 'fwupd'),
+  )
+
+  custom_target('doc-fwupdplugin',
+    input: [
+      fwupdplugin_toml,
+      fwupdplugin_gir[0],
+    ],
+    output: 'libfwupdplugin',
+    command: [
+      gidocgen,
+      'generate',
+      '--quiet',
+      '--add-include-path=@0@'.format(meson.current_build_dir() / '../libfwupd'),
+      '--add-include-path=@0@'.format(meson.current_build_dir() / '../libfwupdplugin'),
+      '--config=@INPUT0@',
+      '--output-dir=@OUTPUT@',
+      '--no-namespace-dir',
+      '--content-dir=@0@'.format(meson.current_source_dir()),
+      '@INPUT1@',
+    ],
+    depends: [
+      fwupdplugin_gir[0],
+    ],
+    build_by_default: true,
+    install: true,
+    install_dir: join_paths(datadir, 'doc', 'fwupd'),
+  )
+elif gtkdocscan.found()
+  gnome.gtkdoc(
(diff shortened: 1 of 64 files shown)


subprojects/gi-docgen.wrap
```diff
@@ -0,0 +1,6 @@
+[wrap-git]
+directory=gi-docgen
+url=https://gitlab.gnome.org/GNOME/gi-docgen.git
+push-url=ssh://git@gitlab.gnome.org:GNOME/gi-docgen.git
```
superm1, line 4: is push-url really needed?

libfwupd/meson.build
```diff
@@ -135,6 +135,7 @@ if get_option('introspection')
     includes : [
       'Gio-2.0',
       'GObject-2.0',
+#      'Json-1.0',
```
superm1, line 138: if not needed, maybe just drop all together?

libfwupdplugin/meson.build
```diff
@@ -203,6 +203,8 @@ if get_option('introspection') and get_option('gusb')
     includes : [
       'Gio-2.0',
       'GObject-2.0',
+      'GUsb-1.0',
+#      'Xmlb-2.0',
```
superm1, line 207: drop it instead of comment out?

meson.build
```diff
@@ -529,10 +529,12 @@ endif
...
-if get_option('gtkdoc')
-  gtkdocscan = find_program('gtkdoc-scan', required : true)
-  subdir('docs')
-endif
+gidocgen_dep = dependency(
+  'gi-docgen', version: '>= 2021.1',
+  fallback: ['gi-docgen', 'dummy_dep'],
+  required: get_option('gtkdoc'),
+)
+
```
superm1, line 537: shouldn't this be only if gtkdoc is set to true? Otherwise it will cause it to subproject on all builds

docs/meson.build
```diff
@@ -1,16 +1,66 @@
...
+gidocgen = find_program('gi-docgen')
+
+custom_target('doc-fwupd',
+  input: [
+    fwupd_toml,
+    fwupd_gir[0],
+  ],
+  output: 'libfwupd',
+  command: [
+    gidocgen,
```
superm1, line 25: shouldn't this be `gidocgen_dep` or similar?

libfwupd/meson.build
hughsie: We do need it, but including it breaks the vapi build further down. It's on the list to debug, although it looks like a vala issue really.

libfwupdplugin/meson.build
hughsie: Again, we need it, but we need to depend on a version of libxml that has this bug fixed: https://github.com/hughsie/libxmlb/pull/90

docs/meson.build
hughsie: Nah, `gidocgen_dep` is the .pc file target, `gidocgen` is the binary. The former is only really needed to create the subproject.

hughsie: > But how are you going to find the binary if you didn't have it pre-installed

No idea; I don't have it locally so I guess it's finding it from the subpackage somehow.

libfwupd/meson.build
hughsie: Fixed.

docs/meson.build
hughsie: Fixed.

libfwupdplugin/meson.build
hughsie: I merged a better version of this in https://github.com/fwupd/fwupd/pull/3283

superm1:
I played with it a bit, did some cleanup and this stuff is phenomenal.  I love that it links back to live source links too.  If we wrap this with a bow and just build it at release time and publish to fwupd.github.io, I fell like that's more than useful enough for anyone to use.

So crazy idea - what if we just stop making any distro packages for documentation?  Then we aren't stuck waiting for a stable release of gi-docgen to end up anywhere in repositories.  We just use it at release time and that's it.  Also we don't have to worry about building documentation for all our CI targets, we just build it for the one that publishes the docs.  (Ubuntu x86-64 IIRC).

All we'll need is a landing page to drop down to either `/libfwupd/index.html` or `/libfwupdplugin/index.html`.

hughsie:
> what if we just stop making any distro packages for documentation

We need it for the devhelp file unforntately.

superm1:
> > what if we just stop making any distro packages for documentation
> 
> We need it for the devhelp file unforntately.

What's that?

superm1:
Furthermore what's to stop this devhelp file from being a html file with links to fwupd.github.io and lvfs.readthedocs.com?

superm1:
Looking in Debian about this:
>Until GI-DocGen becomes stable, Debian packages that use it for their documentation should use a vendored
copy (as allowed by Policy §4.13), and should not have a Build-Depends on the gi-docgen binary package.

And the way that GTK does it is like this:
```
meson setup -Dgtk_doc=true --force-fallback-for gi-docgen
```

Which isn't going to work with automatic github release tagging and tarball generation.  So if we did something like that, we would have to have a CI job go and swap out the `.tar.gz` with one with all that probably.

hughsie:
>>     We need it for the devhelp file unforntately.
> What's that?

Screenshot from 2021-06-04 17-16-11

>  So if we did something like that, we would have to have a CI job go and swap out the .tar.gz with one with all that probably

Ya, it's not ideal it's a subproject. Ideally we'd have a distro release we could depend on instead but gi-docgen is still in alpha mode.

superm1:
Another crazy idea... How about support gtk doc and gidoc both? If you've got the subprojected gidocgen use it, otherwise use gtk?

Then "everyone" wins.

hughsie:
That's not crazy at all -- we'd loose some of the deep linking as the syntax is slightly different, but for the transition that's probably fine. I can work on that on Monday if that helps; today I'm besieged with hundreds of emails after 7 days of PTO.

superm1:
Lose the deep linking in which direction? I think keep it for gi and lose for gtk if we go that way.

superm1:
@hughsie I think it's in a good state now for review with all the fallback magic.

hughsie:
Pushed as https://github.com/fwupd/fwupd/pull/3335

PR #3235, opened 2021-05-08
rts54hub: Allow updating the RTD21xx devices in background mode

hughsie:
Original code by Ricky Wu <ricky_wu@realtek.com>

Type of pull request:
- [ ] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [ ] Code fix
- [X] Feature
- [ ] Documentation

PR #3245, opened 2021-05-11
Modify rts54hub rt2xxx foreground

nwj1027:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/rts54hub/fu-rts54hub-rtd21xx-foreground.c
+++ b/plugins/rts54hub/fu-rts54hub-rtd21xx-foreground.c
@@ -134,7 +134,7 @@ fu_rts54hub_rtd21xx_foreground_attach (FuDevice *device, GError **error)
 	if (locker == NULL)
 		return FALSE;
 
-	if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
+	if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_IS_BOOTLOADER)) {
 		write_buf[0] = ISP_CMD_FW_UPDATE_EXIT;
 		if (!fu_rts54hub_rtd21xx_device_i2c_write (FU_RTS54HUB_RTD21XX_DEVICE (self),
 						   UC_ISP_SLAVE_ADDR,
@@ -164,7 +164,7 @@ fu_rts54hub_rtd21xx_foreground_attach (FuDevice *device, GError **error)
 		/* the device needs some time to restart with the new firmware before
 		* it can be queried again */
 		fu_device_sleep_with_progress (device, 60);
-		fu_device_remove_flag (device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE);
+		fu_device_remove_flag (device, FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
 	}
 	/* success */
 	return TRUE;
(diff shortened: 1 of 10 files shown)


plugins/rts54hub/fu-rts54hub-rtd21xx-foreground.c
```diff
@@ -289,6 +302,7 @@ fu_rts54hub_rtd21xx_foreground_write_firmware (FuDevice *device,
 						     NULL, error))
 		return FALSE;
 
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE);
```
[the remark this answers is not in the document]
nwj1027: Done。

plugins/rts54hub/fu-rts54hub-rtd21xx-foreground.c
```diff
@@ -126,33 +126,46 @@ fu_rts54hub_rtd21xx_foreground_attach (FuDevice *device, GError **error)
...
+							   write_buf, 1,
+							   error)) {
+			g_prefix_error (error, "exit foreground-fw mode: ");
+			return FALSE;
+		}
 
+		/* the device needs some time to restart with the new firmware before
+		* it can be queried again */
+		fu_device_sleep_with_progress (device, 60);
+		fu_device_remove_flag (device, FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
```
hughsie, line 167: rather than hardcoding this, is it possible for setup() to either add or remove the flag depending on a query result?

plugins/rts54hub/fu-rts54hub-rtd21xx-foreground.c
```diff
@@ -289,6 +302,7 @@ fu_rts54hub_rtd21xx_foreground_write_firmware (FuDevice *device,
 						     NULL, error))
 		return FALSE;
 
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
```
hughsie, line 305: ya, i think this is something we want to set during ->probe() or ->setup()

plugins/rts54hub/fu-rts54hub-rtd21xx-foreground.c
hughsie: Sure, no problem. Most other plugins detect what "mode" the device is in during `fu_rts54hub_rtd21xx_foreground_setup()` -- but I guess in this case maybe you're right -- as `fu_rts54hub_rtd21xx_foreground_attach()` is being called both at setup time (because of the FuDeviceLocker) and **also** after the firmware update has been run.

Are you using the presence of `FWUPD_DEVICE_FLAG_IS_BOOTLOADER` to chose what to run during attach? Because if so, might it be a lot simpler if you had something like this in `fu_rts54hub_rtd21xx_foreground_setup()`:

	locker = fu_device_locker_new_full (device,
					    (FuDeviceLockerFunc) fu_device_detach,
					    (FuDeviceLockerFunc) fu_device_update_exit,
					    error);

and then `fu_device_update_exit()` would use `ISP_CMD_FW_UPDATE_EXIT` and the existing `fu_device_attach_for_setup()` (used after upgrade) would use just `ISP_CMD_FW_UPDATE_RESET`.

hughsie: Exactly like that, thanks.

hughsie:
I've pulled in the fixes into https://github.com/fwupd/fwupd/pull/3235, thanks.

PR #3246, opened 2021-05-11
Do not write changes to /etc/fwupd/remotes.d

hughsie:
Write to a mutable version of the remote in /var/lib/fwupd/remotes.d
rather than writing to /etc/fwupd/remotes.d which may be immutable on
some systems. This also allows distributors to turn off features without
modifying the core fwupd package.

The values in /var/lib override the ones in /etc.

### DIFF
--- a/libfwupd/fwupd-remote.c
+++ b/libfwupd/fwupd-remote.c
@@ -190,9 +190,11 @@ fwupd_remote_set_id (FwupdRemote *self, const gchar *id)
 /**
  * fwupd_remote_set_filename_source:
  * @self: a #FwupdRemote
- * @filename_source: (nullable): filename
+ * @filename_source: (nullable): The mutable filename
  *
- * Sets the source filename. This is typically a file in `/etc/fwupd/remotes/`.
+ * Sets the filename of the mutable metadata source.
+ * This is typically a location in /var/lib/fwupd/remotes.d rather than the
+ * possibly-read-only location in /etc/fwupd/remotes.d
  *
  * Since: 1.6.1
  **/
--- a/src/fu-remote-list.c
+++ b/src/fu-remote-list.c
@@ -168,6 +168,7 @@ fu_remote_list_add_for_path (FuRemoteList *self, const gchar *path, GError **err
 		return FALSE;
 	while ((tmp = g_dir_read_name (dir)) != NULL) {
 		g_autofree gchar *filename = g_build_filename (path_remotes, tmp, NULL);
+		g_autofree gchar *filename_mut = NULL;
 		g_autoptr(FwupdRemote) remote = fwupd_remote_new ();
 		g_autofree gchar *localstatedir = NULL;
 		g_autofree gchar *remotesdir = NULL;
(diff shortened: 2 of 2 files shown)


libfwupd/fwupd-remote.c
```diff
@@ -185,7 +187,18 @@ fwupd_remote_set_id (FwupdRemote *self, const gchar *id)
...
+ * @self: a #FwupdRemote
+ * @filename_source: (nullable): The mutable filename
+ *
+ * Sets the filename of the mutable metadata source.
+ * This is typically a location in /var/lib/fwupd/remotes.d rather than the
+ * possibly-read-only location in /etc/fwupd/remotes.d
+ *
+ * Since: 1.6.1
+ **/
+void
```
superm1, line 201: so going this approach a single field can't be modified, but rather you need to fully duplicate the file, right?

superm1:
> Write to a mutable version of the remote in /var/lib/fwupd/remotes.d
> rather than writing to /etc/fwupd/remotes.d which may be immutable on
> some systems. This also allows distributors to turn off features without
> modifying the core fwupd package.
> 
> The values in /var/lib override the ones in /etc.

Which OSes have /etc/ as immutable?  What do people do for systemd units?  I always thought that etc was the right place even on atomic operating systems.

superm1:
> > Which OSes have /etc/ as immutable
> 
> http://0pointer.net/blog/projects/stateless.html is the answer to the question, but the thing I'm trying to fix here is that Pop!_OS is currently shipping a custom fwupd package just so they can set `ReportURI=` i.e. turn off the LVFS reports. Doing this with an override just means the OS (and not the fuwpd package) has to ship a `/usr/lib/fwupd/remotes.d/lvfs.conf` of:
> 
> ```
> [fwupd Remote]
> ReportURI=
> ```

Sorry but that seems like it's a bit of an overreach to go out of our ways to enable them to do.
Reports are _already_ opt-in, and they don't even want to allow people take the **opportunity** to opt-in?

hughsie:
@jackpot51 can you provide some more info here? Mario kinda has a point...

johannbg:
> So it almost sounds like the design that we want here is that the configurations are thrown into /use as immutable, and then let /etc or /var override them in a layering progressive order.

Files in /usr are just considered immutable from the "admin/user" not the "vendor/distribution" so that wont work if I understand you correctly since you still want/need conf.d directorie(s) under /usr for the main configuration file(s) as well so that vendors/distributions can easily overwrite the main configuration(s) even if those main configuration files are small since main configuration file can a) grow over time as more options get added and b) if you dont do that you are somewhat forcing vendors having to replace the entire main configuration(s) file(s) and having to maintain it when in fact they might just want to adjust a single entry but otherwise accept all other current and future defaults right.

superm1:
But if they're layered, then a distro wants to change they drop the change in /etc as part of package just like now.

johannbg:
> But if they're layered, then a distro wants to change they drop the change in /etc as part of package just like now.

/etc is admin only territory so those distro's that are doing that are doing it wrong as in they should be dropping it into /usr not /etc since those distributions are in the role of the vendor but it probably depends on the distribution.

Basically you only have one source of truth in a modern os and that source of truth is /usr everything else can be (re)created from there.

hughsie:
> When using the command line the opt in prompt is fine

Okay, given that lets close this. We can reuse all the other changes in the future if we want to make /etc/ immutable or something.

PR #3249, opened 2021-05-14
Fix dependencies of the ModemManager plugin

aleksander0m:
Trivial changes to rework the MM plugin dependencies.


hughsie (approved):

aleksander0m:
Oh wait, we do need MM_REQUIRED_VERSION, but for fwupd itself. Let me fix that.

hughsie (approved):

PR #3251, opened 2021-05-14
Follow HTTP 3XX redirects when downloading files

hughsie:
This allows us to punt some smaller files onto the CDN.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):

superm1:
What's going to happen to older releases though?

hughsie:
> What's going to happen to older releases though?

It's not going to work :) What happens with older releases is we try to parse the 302 output as a cab file (which fails the checksum) which is about as successful as you might expect. This is more for any cleverness we chose to do in the far future.

PR #3257, opened 2021-05-17
Split flashrom users into multiple plugins

tari:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

It's currently surprising for multiple completely different devices to be supported by a single `flashrom` plugin (coreboot-based system firmware, and Parade LSPCONs), and adding support for more devices in the `flashrom` plugin would require additional undesirable complexity (like #3207, #3188). This change moves flashrom support code from `FlashromDevice` (which is limited to the single plugin and somewhat brittle) to a new `FuFlashromContext` that allows users to use flashrom more flexibly without compromising safety as well as papering over some of libflashrom's problems (which are being fixed upstream, but will remain as they are for an unknown amount of time).

With flashrom support code moving out of the `flashrom` plugin, the existing PS175 LSPCON support is moved to a new `parade_lspcon` plugin and future devices can also be placed in new plugins.


hughsie (changes requested):
I'm not happy to hard depend on libflashrom in the public API like this. It's not really possible for regular human users (non-Google employees) to get patches into the flashrom project. It's taken 5 months of continually rebasing to fix conflicts and getting approval to just get the most basic of 0..100% _percentage reporting_ of the transaction, and it doesn't look likely that patch will ever get merged. More generally, I'm certainly not going to pin the success of the fwupd project on the health of the flashrom community and adding hundreds of lines to to work around actual or perceived design issues in flashrom isn't something I want to maintain for the next 10 years.

It would be so much simpler, and orders of magnitude more maintainable to just reimplement the low level commands like lspcon_i2c_spi_write_register() and lspcon_i2c_spi_disable_protection() in a plugins/lspcon-i2c -- and then FuLspconI2cParade can either use, or can derive from the parent lspcon. I'm pretty sure it would be **less** lines of code, not more, and also allows us to use our existing primitives for things like reporting errors where they occur, rather than having to screen scrape things from the global `flashrom_set_log_callback()` queue.

I know it's probably easier for you to add all this extra helper code to fwupd and *just use libflashrom* but it's going to **me** maintaining this stack for the next 10 years, and I think this is another one of the impedance mismatched abstraction layers that we've tried so hard to avoid so far.

I'm sorry this isn't the review comment you were hoping for.

hughsie:
adding @bleungatchromium for sanity check

tari:
@djcampello as well, as I've discussed this with him.

superm1:
> I'm not happy to hard depend on libflashrom in the public API like this. It's not really possible for regular human users (non-Google employees) to get patches into the flashrom project

Is this something that Google can help with perhaps?  Putting some weight behind it might make this more palatable?

meson_options.txt
```diff
@@ -7,6 +7,7 @@ option('introspection', type : 'boolean', value : true, description : 'generate
 option('lvfs', type : 'boolean', value : true, description : 'enable LVFS remotes')
 option('man', type : 'boolean', value : true, description : 'enable man pages')
 option('libarchive', type : 'boolean', value : true, description : 'enable libarchive support')
+option('libflashrom', type : 'boolean', value : true, description : 'enable libflashrom support')
```
superm1, line 10: I find it hard to make the argument of default of this to be `true` when `plugin_flashrom` is default to `false`.

Furthermore, if this PR ends up directionally being correct, I think that `plugin_flashrom` should just become `flashrom`.  In switching to `flashrom` it would be "support for flashrom based devices".  Similar to how we have support for gusb/guedev/bluez but don't prefix them with `lib`

plugins/parade-lspcon/meson.build
```diff
@@ -0,0 +1,36 @@
+if get_option('plugin_parade_lspcon')
+if not get_option('libflashrom')
+  error('libflashrom is required for plugin_parade_lspcon')
+endif
```
superm1, line 4: is there not any extra check needed for the specific version of libflashrom?

plugins/parade-lspcon/meson.build
```diff
@@ -0,0 +1,36 @@
+if get_option('plugin_parade_lspcon')
```
superm1, line 1: Doesn't this also have an implied dependency on gudev?

hughsie:
>  Putting some weight behind it might make this more palatable

The lack of progress reporting is just one example. We'd be having the same argument about libfwupdate if we never absorbed that years ago. All these extra layers on layers just to write a few bytes to an file descriptor is crazy. Seeing things like this in the flashrom source:

    int realtek_mst_i2c_spi_init(void)
    {
	if (get_params(&reset, &enter_isp))
		return SPI_GENERIC_ERROR;
	if (enter_isp) {
		ret |= realtek_mst_i2c_spi_enter_isp_mode(fd);
		if (ret)
			return ret;
	}

...just really shows the impedance mismatch between the two interfaces. Trying to work around things like that with wrappers is a maintenance nightmare and almost impossible to make both easy-to-understand and reliable.

PR #3258, opened 2021-05-17
Enable building uefi-capsule plugin on FreeBSD 

SergiiDmytruk:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

FreeBSD has Linux-like `<efivar-dp.h>` header, but it doesn't define type constants and it's not included implicitly by `<efivar.h>`.

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-devpath.c
+++ b/plugins/uefi-capsule/fu-uefi-devpath.c
@@ -13,6 +13,14 @@
 
 #include "fwupd-error.h"
 
+#ifndef EFIDP_END_TYPE
+#define EFIDP_END_TYPE      0x7f
+#endif
+
+#ifndef EFIDP_END_ENTIRE
+#define EFIDP_END_ENTIRE    0xff
+#endif
+
 typedef struct {
 	guint8	 type;
 	guint8	 subtype;
--- a/plugins/uefi-capsule/fu-uefi-devpath.c
+++ b/plugins/uefi-capsule/fu-uefi-devpath.c
@@ -7,6 +7,7 @@
 #include "config.h"
 
 #include <efivar.h>
+#include <efivar-dp.h>
 
 #include "fu-common.h"
 #include "fu-uefi-devpath.h"
--- a/plugins/uefi-capsule/fu-uefi-devpath.c
+++ b/plugins/uefi-capsule/fu-uefi-devpath.c
@@ -6,19 +6,20 @@
 
 #include "config.h"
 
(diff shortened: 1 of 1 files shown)


superm1 (commented):
I think would prefer to do ifdef here between Linux and freebsd to be more predicable in case anything changes in the future for the headers in either case and breaks one another.

Also can you add a CI target for bsd?

SergiiDmytruk:
> I think would prefer to do ifdef here between Linux and freebsd to be more predicable in case anything changes in the future for the headers in either case and breaks one another.

Done in the latest commit.

> Also can you add a CI target for bsd?

There are quite [some changes](https://reviews.freebsd.org/D29332) needed for a successful build, but it is [possible](https://blog.3mdeb.com/2021/2021-03-15-fwupd-bsd-packages-and-ci/). It depends on whether you're OK with depending on those unmerged changes, which are even more WIP for dependencies of uefi-capsule.

superm1:
> > I think would prefer to do ifdef here between Linux and freebsd to be more predicable in case anything changes in the future for the headers in either case and breaks one another.
> 
> Done in the latest commit.

Thanks

> 
> > Also can you add a CI target for bsd?
> 
> There are quite [some changes](https://reviews.freebsd.org/D29332) needed for a successful build, but it is [possible](https://blog.3mdeb.com/2021/2021-03-15-fwupd-bsd-packages-and-ci/). It depends on whether you're OK with depending on those unmerged changes, which are even more WIP for dependencies of uefi-capsule.

I just skimmed, but a majority of the focus seems to be on the fwupd daemon side of things. Is it pretty heavy on the EFI app too? If it's just depending on a master branch somewhere for CI purposes on bsd only thats different than random hacky patch. So I guess I mean it depends on what we're looking at :)

My main goal is that we have some way to tell that a change for BSD breaks Linux build or vice versa.

superm1 (approved):

SergiiDmytruk:
I missed that someone is already adding CI in https://github.com/fwupd/fwupd/pull/3031 for the same ports fork, not sure what's the trouble with it.

> I just skimmed, but a majority of the focus seems to be on the fwupd daemon side of things. Is it pretty heavy on the EFI app too?

That PR hasn't yet been updated with the changes which make this plugin build (it just started to build successfully, hence the PRs here and to `fwupd-efi`).

superm1:
Maybe for now, can you get the FreeBSD builds in fwupd-efi going?  That will be a good start to make sure the EFI application is building outside of linux.

PR #3269, opened 2021-05-19
pixart-rf: Fix uuid naming way

jyu99:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/pixart-rf/fu-pxi-device.c
+++ b/plugins/pixart-rf/fu-pxi-device.c
@@ -887,7 +887,7 @@ fu_pxi_device_setup_guid (FuPxiDevice *self, GError **error)
 		devid2 = g_strdup_printf ("HIDRAW\\VEN_%04X&DEV_%04X&MODEL_%s",
 					 (guint) hid_raw_info.vendor,
 					 (guint) hid_raw_info.product,
-					 dev_name->str);
+					 model_name->str);
 		fu_device_add_instance_id (FU_DEVICE (self), devid2);
 	}
 #endif


plugins/pixart-rf/fu-pxi-device.c
```diff
@@ -887,7 +887,7 @@ fu_pxi_device_setup_guid (FuPxiDevice *self, GError **error)
 		devid2 = g_strdup_printf ("HIDRAW\\VEN_%04X&DEV_%04X&MODEL_%s",
 					 (guint) hid_raw_info.vendor,
 					 (guint) hid_raw_info.product,
-					 dev_name->str);
+					 model_name->str);
```
hughsie, line 890: Why not add both? i.e. create a `devid3` with `model_name`? Then you can choose the most correct GUID in the firmware file.

jyu99: I think it was duplicate with devid in code. That is my fault . I didn't notice this part.
So create devid3 is the better way?

hughsie: If in doubt, more GUIDs are better. If you add the extra one, be sure to update README.md too.

jyu99: After I check model name. it already existed in README. So I think fix code  is better than adding new device id.

hughsie: @hardboprobot what do you think here?

hardboprobot: IIRC, the existing FW releases match the GUID based on the vendor/device/revision ids, so this change shouldn't break compatibility with them. If there aren't any other possible issues that I might be overlooking, then I'm ok with it.

hughsie (approved):

PR #3274, opened 2021-05-20
Add FreeBSD CI

Asiderr:
Signed-off-by: Piotr Konkol <piotr.konkol@3mdeb.com>

Related to: https://github.com/fwupd/fwupd/pull/3031

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


.github/workflows/main.yml
```diff
@@ -53,6 +53,47 @@ jobs:
...
+        mem: 8192
+        prepare: |
+          pkg install -y git python3 glib meson pkgconf gobject-introspection \
+            vala gtk-doc json-glib gpgme gnutls sqlite3 curl gcab libarchive \
+            help2man libelf libgpg-error gettext-tools gtk-update-icon-cache \
+            atk pango
+        sync: rsync
+        run: |
+          cd /usr
+          git clone https://github.com/3mdeb/freebsd-ports.git --depth 1 -b fwupd ports
```
superm1, line 79: Just some comments on this:
1. how come gusb can't subproject?  If you could rely on subproject, then you can upstream those patches and depend on a newer tag or release potentially.
2. It would be nice if at least for CI that `sysutils/fwupd/*` was in upstream `contrib/`.  When we drop/add new options then that thing can be directly patched as part of CI.  Otherwise if we remove an option or change the default, CI might break for FreeBSD since we can't directly change it.  That's why we keep debian packaging and fedora packaging upstream for example.
3. Is that libjcat patch and libxmlb really needed?  They both look odd to me.

.github/workflows/main.yml
```diff
@@ -53,6 +53,47 @@ jobs:
...
+          cd /usr/ports/sysutils/fwupd
+          REPOSITORY_NAME=$(echo "${GITHUB_REPOSITORY}" | awk -F / '{print $2}')
+          sed -i .old "s/#GH_TAGNAME=.*$/GH_TAGNAME=\t${GITHUB_SHA}/" Makefile
+          sed -i .old "s/GH_ACCOUNT=.*$/GH_ACCOUNT=\t${GITHUB_REPOSITORY_OWNER}/" Makefile
+          sed -i .old "s/DISTVERSION=.*$/DISTVERSION=${{steps.tagger.outputs.tag}}/" Makefile
+          make makesum
+          make package
+          make install
+          cp /usr/ports/sysutils/fwupd/work/pkg/fwupd*.txz ~/work/fwupd/fwupd/ || exit 1
```
superm1, line 88: does this artifact name end up with freebsd in the title hopefully?  If we're going to save artifacts as part of the release we ideally want to make sure we can delineate which are BSD.

.github/workflows/main.yml
Asiderr: @superm1 1. More information on this topic https://github.com/hughsie/libgusb/pull/10
2. Totally agree. After clone the freebsd-ports fwupd files could be replaced with the one from the `contrib/`.
3. Yes, otherwise meson configuration fails:
libxmlb:
```
Configuring config.h using configuration
Program python found: NO

meson.build:152:0: ERROR: python3 not found
```
libjcat
```
Program python (xml.etree.ElementTree, pkg_resources) found: NO

libjcat/meson.build:142:2: ERROR: python3 not found
```

.github/workflows/main.yml
Asiderr: @superm1 Unfortunately, no. I will change the name to fwupd-freebsd-${commit-sha}-${tag}.txz

.github/workflows/main.yml
```diff
@@ -53,6 +53,47 @@ jobs:
...
+      id: test
+      uses: vmactions/freebsd-vm@v0.1.3
+      with:
+        usesh: true
+        mem: 8192
+        prepare: |
+          pkg install -y git python3 glib meson pkgconf gobject-introspection \
+            vala gtk-doc json-glib gpgme gnutls sqlite3 curl gcab libarchive \
+            help2man libelf libgpg-error gettext-tools gtk-update-icon-cache \
+            atk pango
```
hughsie, line 75: do we need atk too?

.github/workflows/main.yml
```diff
@@ -53,6 +53,47 @@ jobs:
...
+    - name: Build
+      id: test
+      uses: vmactions/freebsd-vm@v0.1.3
+      with:
+        usesh: true
+        mem: 8192
+        prepare: |
+          pkg install -y git python3 glib meson pkgconf gobject-introspection \
+            vala gtk-doc json-glib gpgme gnutls sqlite3 curl gcab libarchive \
+            help2man libelf libgpg-error gettext-tools gtk-update-icon-cache \
```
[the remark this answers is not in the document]
Asiderr: fwupd doesn't rely on them, but they are needed to create the FreeBSD package. If they are not installed manually, the build system will compile these packages from the source (which takes a long time) during the fwupd build.

.github/workflows/main.yml
Asiderr: As above.

.github/workflows/main.yml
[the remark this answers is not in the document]
hughsie: Fair enough, thanks for the rationale.

.github/workflows/main.yml
hughsie: > meson.build:152:0: ERROR: python3 not found

Most unexpected. We should fix that upstream certainly.

contrib/freebsd/Makefile
```diff
@@ -0,0 +1,53 @@
+# Created by: Norbert Kamiński <norbert.kaminski@3mdeb.com>
+# $FreeBSD$
+
+PORTNAME=	fwupd
+DISTVERSION=	1.5.9
```
superm1, line 5: Don't hardcode this to anything upstream. Leave it as a value that is filled in with sed or awk.

contrib/freebsd/Makefile
```diff
@@ -0,0 +1,53 @@
...
+GH_TAGNAME=	58fbbc2
```
superm1, line 6: Same

contrib/freebsd/Makefile
```diff
@@ -0,0 +1,53 @@
...
+		-Dplugin_nvme=false \
+		-Dplugin_redfish=false \
+		-Dplugin_synaptics_mst=false \
+		-Dplugin_synaptics_rmi=false \
+		-Dplugin_thunderbolt=false \
+		-Dplugin_tpm=false \
+		-Dplugin_uefi_capsule=false \
+		-Dpolkit=false \
+		-Dsystemd=false \
+		-Dtests=false
```
superm1, line 51: Can these be left on and CI pass? We ideally want them too.

contrib/freebsd/pkg-plist
```diff
@@ -0,0 +1,189 @@
...
+%%ETCDIR%%/remotes.d/lvfs.conf
+%%ETCDIR%%/remotes.d/vendor-directory.conf
+%%ETCDIR%%/remotes.d/vendor.conf
+etc/pki/fwupd-metadata/GPG-KEY-Linux-Foundation-Metadata
+etc/pki/fwupd-metadata/GPG-KEY-Linux-Vendor-Firmware-Service
+etc/pki/fwupd-metadata/LVFS-CA.pem
+etc/pki/fwupd/GPG-KEY-Linux-Foundation-Firmware
+etc/pki/fwupd/GPG-KEY-Linux-Vendor-Firmware-Service
+etc/pki/fwupd/LVFS-CA.pem
+include/fwupd-1/fwupd.h
```
superm1, line 15: Any chance you can use wildcards in this file?

hughsie:
@Asiderr do you want to do anything cleverer than in https://github.com/hughsie/libgusb/pull/49 -- I think adding FreeBSD CI to our deps is probably a long term good plan.

Asiderr:
@hugsie currently, I'm testing [those changes](https://github.com/hughsie/libgusb/pull/49) and I'll leave review/approval. After that, I need to focus on the USB reset problem (it's pretty urgent for us).

.github/workflows/main.yml
hughsie: Okay, you can use subproject versions of master for everything now it it just works without patches.

superm1: Until releases are tagged I guess that means moving subprojects/*.wrap over to master for those needed.

hughsie: hang on for now; i was going to do new releases of all the deps next week to keep this sane

superm1 (commented):
can this rebased now?

contrib/freebsd/Makefile
Asiderr: Values set during the build.

contrib/freebsd/Makefile
Asiderr: Values set during the build.

contrib/freebsd/files/patch-meson.build
```diff
@@ -0,0 +1,11 @@
+--- meson.build.orig	2021-05-31 11:30:38.106093380 +0200
++++ meson.build	2021-05-31 11:31:11.774182473 +0200
+@@ -252,7 +252,7 @@
+ libgcab = dependency('libgcab-1.0', version : '>= 1.0', fallback : ['gcab', 'gcab_dep'])
+ gcab = find_program('gcab', required : get_option('tests'))
+ bashcomp = dependency('bash-completion', required: false)
+-python3 = find_program('python3')
++python3 = find_program('python3.8')
```
superm1, line 8: any cleaner way to do this?  Maybe something like this upstream?
```
if !python3.found()
    python3 = find_program('python3.8')
endif
```

Or instead have a for loop of options (python3, python3.8, python3.9) or something?

contrib/freebsd/distinfo
```diff
@@ -0,0 +1,3 @@
+TIMESTAMP = 1622115022
```
superm1, line 1: I don't see any sed into this file, does it need some?

contrib/freebsd/pkg-plist
Asiderr: The file is not required to build the package (if there is no pkg-plist, the list will be generated during the build).

superm1 (commented):
I think as part of the addition of this job, the other freebsd built job can be removed.  We don't need both.

contrib/freebsd/distinfo
Asiderr: There is no need, it's overwritten during the build. `make makesum` generates distinfo for the current commit:
https://github.com/fwupd/fwupd/pull/3274/commits/cd5c8eb8389f0094b1e5abe7ef94793c3047c7b4#diff-bdce2284300df67bb504fd10f5eb71470dbe6d3c074ff78fc6ccaaf405e6e008R54

contrib/freebsd/files/patch-meson.build
Asiderr: Added.

meson.build
```diff
@@ -252,7 +252,7 @@ libm = cc.find_library('m', required: false)
 libgcab = dependency('libgcab-1.0', version : '>= 1.0', fallback : ['gcab', 'gcab_dep'])
 gcab = find_program('gcab', required : get_option('tests'))
 bashcomp = dependency('bash-completion', required: false)
-python3 = find_program('python3')
+python3 = find_program('python3.8', 'python3', 'python3.9')
```
superm1, line 255: Any reason it needs to be 3.8 first then 3?  I would think it should be 3 first.

contrib/freebsd/distinfo
superm1: Then perhaps just drop the file alltogether if it's generated during make command.

.github/workflows/main.yml
```diff
@@ -78,17 +78,34 @@ jobs:
...
       id: test
       uses: vmactions/freebsd-vm@v0.1.4
       with:
         usesh: true
         mem: 8192
         prepare: |
-          pkg install -y git gtar python3 glib meson pkgconf gobject-introspection \
-            json-glib gnutls gtk-doc vala sqlite3 curl gcab libarchive libelf \
-            gettext-tools gnutls usbids polkit
-        run: ./contrib/ci/build_freebsd.sh
```
superm1, line 94: maybe we should drop this file too as part of this PR.

meson.build
Asiderr: Yes it's important. In the FreeBSD 13.0 python3 is related to the 3.7 version and there is no py37-setuptool module in the pkg  (It conflicts with py38-setuptool when it's built from the source). So we have to check first if there is python3.8 to provide required dependencies.

Asiderr: Yes it's important. In the FreeBSD 13.0 python3 is related to the 3.7 version and there is no py37-setuptool module in the pkg  (It conflicts with py38-setuptool when it's built from the source). So we have to check first if there is python3.8 to provide required dependencies.

.github/workflows/main.yml
Asiderr: Deleted.

.github/workflows/main.yml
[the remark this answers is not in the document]
hughsie: Still help2man? Nothing uses it anymore, right?

hughsie:
Looks good to me, @superm1 any changes you wanted to see before we merge?

superm1:
> Looks good to me, @superm1 any changes you wanted to see before we merge?

Just drop `contrib/freebsd/distinfo` and then it's good to me.

contrib/ci/build_freebsd_package.sh
```diff
@@ -0,0 +1,58 @@
...
+set -e
+mkdir -p /usr/local/etc/pkg/repos/
+# Fix meson flag problem https://www.mail-archive.com/freebsd-ports@freebsd.org/msg86617.html
+cp /etc/pkg/FreeBSD.conf /usr/local/etc/pkg/repos/FreeBSD.conf
+sed -i .old 's|url: "pkg+http://pkg.FreeBSD.org/${ABI}/quarterly"|url: "pkg+http://pkg.FreeBSD.org/${ABI}/latest"|' \
+/usr/local/etc/pkg/repos/FreeBSD.conf
+pkg install -y meson
+pkg upgrade -y meson
+cd /usr
+git clone https://github.com/3mdeb/freebsd-ports.git --depth 1 -b fwupd ports
```
hughsie, line 44: this bit does worry me slightly; can't we get this from freebsd upstream? Ideally this would all be self contained.

Asiderr: It will be changed to the upstream repo when fwupd and its dependencies will come into freebsd-ports.

Asiderr:
@superm1 @hughsie I've droped distifo and now we are generating the `plist` file in the runtime. If it's not generated, artifact does not install the files.

hughsie (approved):

Whissi:
This is causing a regression for us, Gentoo: https://bugs.gentoo.org/796143

PR #3275, opened 2021-05-20
Also use the FirmwareBaseURI for the detach and update images

hughsie:
If a company has mirrored the entire LVFS repo for privacy reasons
they'll certainly want to download the images from the same place.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -548,8 +548,17 @@ fu_engine_set_release_from_appstream (FuEngine *self,
 	if (tmp != NULL)
 		fwupd_release_set_detach_caption (rel, tmp);
 	tmp = xb_node_query_text (component, "screenshots/screenshot/image", NULL);
-	if (tmp != NULL)
-		fwupd_release_set_detach_image (rel, tmp);
+	if (tmp != NULL) {
+		if (remote != NULL) {
+			g_autofree gchar *img = NULL;
+			img = fwupd_remote_build_firmware_uri (remote, tmp, error);
+			if (img == NULL)
+				return FALSE;
+			fwupd_release_set_detach_image (rel, img);
+		} else {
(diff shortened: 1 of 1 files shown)


hughsie:
@bleungatchromium

hughsie:
Also, we'll need to add the .png files to the list of things in the PULP_MANIFEST if this is decided to be a good idea.

bleungatchromium:
@djcampello

hughsie:
Gagh, sorry, correct Google person indeed! 🙂

src/fu-engine.c
```diff
@@ -548,17 +548,35 @@ fu_engine_set_release_from_appstream (FuEngine *self,
...
 	tmp = xb_node_query_text (component, "custom/value[@key='LVFS::UpdateMessage']", NULL);
 	if (tmp != NULL)
 		fwupd_release_set_update_message (rel, tmp);
 	tmp = xb_node_query_text (component, "custom/value[@key='LVFS::UpdateImage']", NULL);
-	if (tmp != NULL)
-		fwupd_release_set_update_image (rel, tmp);
+	if (tmp != NULL) {
+		if (remote != NULL) {
+			g_autofree gchar *img = NULL;
+			img = fwupd_remote_build_firmware_uri (remote, tmp, error);
```
superm1, line 572: You can combine with the previous line

superm1 (approved):

src/fu-engine.c
hughsie: It made it a little wide for my liking.

hughsie: We need clang format to settle arguments like this once and for all 🙂

hughsie:
LVFS side in https://gitlab.com/fwupd/lvfs-website/-/merge_requests/985

PR #3284, opened 2021-05-22
Wip/hughsie/kenv

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -23,6 +23,10 @@
 #include <sysinfoapi.h>
 #endif
 
+#ifdef HAVE_KENV_H
+#include <kenv.h>
+#endif
+
 #ifdef HAVE_CPUID_H
 #include <cpuid.h>
 #endif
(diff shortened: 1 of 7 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -3282,6 +3286,44 @@ fu_common_align_up (gsize value, guint8 alignment)
 	return value_new;
 }
 
+/**
+ * fu_common_kenv:
```
superm1, line 3290: How about fu_common_bsd_kenv to make it clearer to future selfs this is only intended for use with bsd?

superm1 (dismissed):

libfwupdplugin/fu-common.c
hughsie: is kenv used on macos too? that's the bit i wasn't sure of.

hughsie: Yes, this seems used on OS-X too. Maybe a better approach would be to have a libfwupdplugin/fu-kenv.[c|h] with a symbol like fu_kenv_get_string()? That might be better for the long term.

superm1: sounds good

hughsie: Done.

superm1 (approved):

PR #3287, opened 2021-05-24
ESRT access on FreeBSD

SergiiDmytruk:
Type of pull request:
- [ ] New plugin
- [ ] Code fix
- [x] Feature
- [ ] Documentation

***

This introduces new `GObject` for accessing ESRT in a platform-specific way. The implementation is provided in separate source files, which otherwise would consist of a lot of `#ifdef`s.

FreeBSD implementation targets [this upstream patch](https://reviews.freebsd.org/D30104).

Had to disable several tests on FreeBSD which rely on sysfs mocking as being Linux-specific. FreebBSD's version uses `sysctl()` which is not as straightforward to mock.

### DIFF
--- a/plugins/uefi-capsule/fu-self-test.c
+++ b/plugins/uefi-capsule/fu-self-test.c
@@ -144,15 +144,22 @@ fu_uefi_bitmap_func (void)
 	g_assert_cmpint (height, ==, 24);
 }
 
+#ifdef __linux__
+
 static void
 fu_uefi_device_func (void)
 {
-	g_autofree gchar *fn = NULL;
 	g_autoptr(FuUefiDevice) dev = NULL;
 	g_autoptr(GError) error = NULL;
+	g_autoptr(FuUefiEsrt) esrt = fu_uefi_esrt_new ();
+	FuUefiEsrtEntry *entry = NULL;
 
-	fn = g_build_filename (TESTDATADIR, "efi/esrt/entries/entry0", NULL);
-	dev = fu_uefi_device_new_from_entry (fn, &error);
+	g_assert_true (fu_uefi_esrt_setup (esrt, &error));
+	g_assert_no_error (error);
+	g_assert_true (fu_uefi_esrt_get_entry_count (esrt) > 0);
+
+	entry = fu_uefi_esrt_get_entry (esrt, 0);
+	dev = fu_uefi_device_new_from_entry (entry, &error);
 	g_assert_nonnull (dev);
 	g_assert_no_error (error);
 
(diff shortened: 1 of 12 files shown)


plugins/uefi-capsule/fu-self-test.c
```diff
@@ -264,8 +277,11 @@ main (int argc, char **argv)
 	g_test_add_func ("/uefi/bgrt", fu_uefi_bgrt_func);
 	g_test_add_func ("/uefi/framebuffer", fu_uefi_framebuffer_func);
 	g_test_add_func ("/uefi/bitmap", fu_uefi_bitmap_func);
+	/* on linux ESRT data can be faked easily */
+#ifdef __linux__
```
hughsie, line 281: i think still compile the tests, but ifndef __linux__ then do a `g_test_skip` and then `return` at the start.

plugins/uefi-capsule/fu-uefi-device.c
```diff
@@ -754,30 +755,27 @@ fu_uefi_device_class_init (FuUefiDeviceClass *klass)
...
-	self->kind = fu_uefi_read_file_as_uint64 (entry_path, "fw_type");
-	self->fw_version = fu_uefi_read_file_as_uint64 (entry_path, "fw_version");
-	self->last_attempt_status = fu_uefi_read_file_as_uint64 (entry_path, "last_attempt_status");
-	self->last_attempt_version = fu_uefi_read_file_as_uint64 (entry_path, "last_attempt_version");
-	self->fw_version_lowest = fu_uefi_read_file_as_uint64 (entry_path, "lowest_supported_fw_version");
+	self->fw_class = fu_uefi_esrt_entry_get_class (entry);
+
+	self->capsule_flags = fu_uefi_esrt_entry_get_capsule_flags (entry);
+	self->kind = fu_uefi_esrt_entry_get_kind (entry);
+	self->fw_version = fu_uefi_esrt_entry_get_version (entry);
```
hughsie, line 775: could we just store a ref to the FuUefiEsrtEntry?

plugins/uefi-capsule/fu-uefi-esrt-freebsd.c
```diff
@@ -0,0 +1,208 @@
...
+
+FuUefiEsrt *
+fu_uefi_esrt_new (void)
+{
+	FuUefiEsrt *self = g_object_new (FU_TYPE_UEFI_ESRT, NULL);
+	return FU_UEFI_ESRT (self);
+}
+
+guint
+fu_uefi_esrt_get_entry_count (FuUefiEsrt *self)
```
hughsie, line 126: I think just returning self->entries (i.e. a GPtrArray *) would make this a lot simpler to use

plugins/uefi-capsule/meson.build
```diff
@@ -10,6 +10,8 @@ endif
 install_data(['uefi-capsule.quirk'],
   install_dir: join_paths(datadir, 'fwupd', 'quirks.d'))
 
+esrt_impl_source = 'fu-uefi-esrt-@0@.c'.format(host_machine.system())
```
hughsie, line 13: what happens if host_machine.system() == windows? I think we need to do the `if`, `elif`, `else` -> `error()` pattern here I'm afraid.

superm1: Or add dummy code for Windows for now.  I don't think it's going to be impossible to get the same info on Windows....

plugins/uefi-capsule/fu-uefi-device.c
SergiiDmytruk: This is just one of three constructors, storing a ref would affect usage of the data a lot.

hughsie: >  storing a ref would affect usage of the data a lot

How do you mean? I think it would make things a lot simpler.

plugins/uefi-capsule/fu-uefi-esrt-linux.c
```diff
@@ -0,0 +1,203 @@
...
+{
+	FuUefiEsrtEntry *self = g_object_new (FU_TYPE_UEFI_ESRT_ENTRY, NULL);
+	self->path = path;
+	return FU_UEFI_ESRT_ENTRY (self);
+}
+
+static void
+fu_uefi_esrt_entry_init (FuUefiEsrtEntry *self)
+{
+	self->path = NULL;
```
superm1, line 38: @hughsie is this actually needed?  Or does it init NULL automatically?

hughsie: It's not needed

SergiiDmytruk: I removed this, [the docs](https://developer.gnome.org/gobject/stable/gobject-Type-Information.html#GInstanceInitFunc) say (if that's documentation for this function):

    The extended members of instance are guaranteed to have been filled with zeros before this function is called.

plugins/uefi-capsule/fu-uefi-device.c
SergiiDmytruk: You'd have to check how device was constructed before accessing its data, so every data access would become conditional. I assumed you suggest to store the ref instead of copying its data or maybe I misunderstood?

hughsie: I think fu_uefi_esrt_setup() should read out all the properties on the ESRT node and allocate strings. Then `fu_uefi_esrt_entry_get_class()` should return a `const gchar *` -- i.e. all the failable i/o should be done in _setup() and then the getters become inexpensive accessors into the data.

hughsie: It might make sense to define a parent GObject with all the logic, and then subclass from that for the linux/freebsd versions just with a ->setup() vfunc. If you'd like me to set that up for you with an example, can you push your changes to origin (i.e. I'll make you a committer) and then I can edit your branch too.

SergiiDmytruk: Thanks, you can push to my branch anyway, because "Allow edits by maintainers" is set for this pull request. Pushed `wip/3mdeb/freebsd-esrt` branch if that's more convenient.

> If you'd like me to set that up for you with an example

That would be helpful. You could also point to a similar case in the sources.

hughsie: Okay, I got carried away and ended up adding a bit more scaffolding as I don't think ESRT will be the only case like this. Alternative proposal in https://github.com/fwupd/fwupd/pull/3289 -- comments welcome.

SergiiDmytruk:
Merged as part of #3289.

PR #3289, opened 2021-05-25
Wip/hughsie/esrt

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-backend.c
+++ b/libfwupdplugin/fu-backend.c
@@ -34,6 +34,15 @@ static guint signals[SIGNAL_LAST] = { 0 };
 G_DEFINE_TYPE_WITH_PRIVATE (FuBackend, fu_backend, G_TYPE_OBJECT)
 #define GET_PRIVATE(o) (fu_backend_get_instance_private (o))
 
+/**
+ * fu_backend_device_added:
+ * @self: a #FuBackend
+ * @device: a device
+ *
+ * Emits a signal that indicates the device has been added.
+ *
+ * Since: 1.6.1
+ **/
 void
 fu_backend_device_added (FuBackend *self, FuDevice *device)
 {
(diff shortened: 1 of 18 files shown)


hughsie:
@SergiiDmytruk can you test ^^ on FreeBSD please? I can't get it to compile as there doesn't seem to be `efiboot` available.

superm1:
Hmm, I guess the question I want to ask - do we really want plugins themselves to be introducing the concept of multiple backends?  It feels like it might be overloading the definition of backend.  Or should we keep that abstraction in libfwupdplugin directly.

SergiiDmytruk (dismissed):

SergiiDmytruk:
I've added a commit to fix FreeBSD. 3mdeb's fork of ports doesn't yet have [my changes](https://github.com/SergiiDmytruk/freebsd-ports/commits/fwupd) with `libefiboot` that can be linked to (mind that my fork lacks updates from 3mdeb's fork, maybe I should rebase).

hughsie:
> I've added a commit to fix FreeBSD

Thanks, I've merged that into my commit.

>  3mdeb's fork of ports doesn't yet have my changes with libefiboot that can be linked to

I did wonder how much of libefiboot we actually use nowadays.

hughsie:
> do we really want plugins themselves to be introducing the concept of multiple backends?

I also share that concern. I actually started with a FuUefiEsrt factory like @SergiiDmytruk suggested, but it was 70% the same code as FuBackend.

> Or should we keep that abstraction in libfwupdplugin directly.

So, that was my longer term plan. i.e. write the plugins to subclass FuBackend to build a device factory, and then if later we want to move the plugin coldplug into a more generic backend instantiated from the engine (to be used on multiple plugins) then the amount of changes should be minimal. i.e. we're converging to a more standard factory design longer term.

superm1:
Yeah, I guess we don't realistically have a need for more plugins to access the ESRT.

OK fine by me.

PR #3291, opened 2021-05-25
Add FreeBSD to the CI matrix

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1:
Nice, 10 minutes instead of 17.

.github/workflows/main.yml
```diff
@@ -75,3 +75,21 @@ jobs:
...
+      id: test
+      uses: vmactions/freebsd-vm@v0.1.4
+      with:
+        usesh: true
+        mem: 8192
+        prepare: |
+          pkg install -y git gtar python3 glib meson pkgconf gobject-introspection \
+            json-glib gnutls gtk-doc vala sqlite3 curl gcab libarchive libelf \
+            gettext-tools gnutls usbids polkit
+        sync: rsync
```
superm1, line 94: if not saving artifacts, I don't know that you need to rsync stuff out do you?

hughsie: Not a clue; I cargo culted that without understanding what it did. @Asiderr ?!

contrib/ci/build_freebsd.sh
```diff
@@ -0,0 +1,27 @@
...
+    -Dplugin_emmc=false \
+    -Dplugin_nvme=false \
+    -Dplugin_synaptics_mst=false \
+    -Dplugin_synaptics_rmi=false \
+    -Dplugin_thunderbolt=false \
+    -Dplugin_tpm=false \
+    -Dplugin_uefi_capsule=false \
+    -Dsystemd=false \
+    $@
+ninja -v || bash
```
superm1, line 24: I think you want to `||` to `sh`, since `bash` isn't in freebsd.

hughsie:
> 10 minutes instead of 17

I'd be less if we can 'pkg install' the subprojects rather than building all of them; @SergiiDmytruk is pushing the deps to the ports tree possible now?

.github/workflows/main.yml
Asiderr: Yes, it's needed only when you're saving artifacts.

contrib/ci/build_freebsd.sh
hughsie: Agreed, thanks.

Asiderr:
@hughsie Yes, the subprojects packages will be upstreamed with fwupd in the one patch to the fwupd-ports, but it will take a while. I will update following diff in the near future: https://reviews.freebsd.org/D29332

PR #3303, opened 2021-05-29
Short term documentation improvements

superm1:
These help the gtk-doc documentation as well.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [x] Documentation

### DIFF
--- a/libfwupd/fwupd-common.h
+++ b/libfwupd/fwupd-common.h
@@ -10,10 +10,35 @@
 
 G_BEGIN_DECLS
 
+/**
+ * FWUPD_DBUS_PATH:
+ *
+ * The dbus path
+ *
+ **/
 #define FWUPD_DBUS_PATH			"/"
+/**
+ * FWUPD_DBUS_SERVICE:
+ *
+ * The dbus service
+ *
+ **/
 #define FWUPD_DBUS_SERVICE		"org.freedesktop.fwupd"
+/**
+ * FWUPD_DBUS_INTERFACE:
+ *
+ * The dbus interface
+ *
+ **/
+
 #define FWUPD_DBUS_INTERFACE		"org.freedesktop.fwupd"
 
+/**
+ * FWUPD_DEVICE_ID_ANY:
+ *
+ * Wildcard used for matching all device ids in fwupd
+ *
+ **/
 #define FWUPD_DEVICE_ID_ANY		"*"
 
 /**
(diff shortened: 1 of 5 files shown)


libfwupd/fwupd-common.h
```diff
@@ -10,10 +10,35 @@
...
+ **/
 #define FWUPD_DBUS_SERVICE		"org.freedesktop.fwupd"
+/**
+ * FWUPD_DBUS_INTERFACE:
+ *
+ * The dbus interface
+ *
+ **/
+
 #define FWUPD_DBUS_INTERFACE		"org.freedesktop.fwupd"
```
hughsie, line 34: I think we should probably take this opportunity to expand out all these -- making them at least into single sentences (e.g. with a full stop at the end) -- for instance, `The main D-Bus interface which is exported by the fwupd daemon which clients can call methods on and register for signals.` I know it's a ton of work, and I'm happy to help with this when I get back.

libfwupd/fwupd-enums-private.h
```diff
@@ -10,61 +10,500 @@
...
+ * Result key to represent ParentDeviceId
+ *
+ * Type of 's'
+ *
+ **/
+#define FWUPD_RESULT_KEY_PARENT_DEVICE_ID	"ParentDeviceId"
+/**
+ * FWUPD_RESULT_KEY_COMPOSITE_ID:
+ *
+ * Result key to represent CompositeId
```
hughsie, line 88: I think we should probably mention things like fwupd_device_get_composite_id() here too.

libfwupd/fwupd-enums-private.h
```diff
@@ -10,61 +10,500 @@
...
+ * Type of 's'
+ *
+ **/
+#define FWUPD_RESULT_KEY_COMPOSITE_ID		"CompositeId"
+/**
+ * FWUPD_RESULT_KEY_FILENAME:
+ *
+ * Result key to represent Filename
+ *
+ * Type of 's'
```
hughsie, line 99: Maybe `The D-Bus type signature string is `s` i.e. a string`

libfwupd/fwupd-enums-private.h
superm1: On the result key?  That's going to be a lot of work in what I think is little value.  It's going to start precedent of trying to remember everywhere that a data type if used.

In this I think it would be _better_ to modify `fwupd_device_get_composite_id` (and similar functions) to point to their types in the constants when used.  This is of course also a lot of work, but I think more useful work.

hughsie: Yes, agreed.

hughsie (dismissed):

PR #3312, opened 2021-06-03
Add support for USB4 module in dell dock

CragW:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

TODO:
1. Update Dock EC lowest version from development to production
2. Upcoming commit to add support for firmware update over USB protocol

### DIFF
--- a/contrib/vscode/build.sh
+++ b/contrib/vscode/build.sh
@@ -8,7 +8,7 @@ if [ -z "$ROOT" ]; then
 fi
 
 # build in tree
-rm -rf build ${ROOT}/dist
+sudo rm -rf build ${ROOT}/dist
 meson build --prefix=${ROOT}/dist -Dsystemd=false -Dudevdir=${ROOT}/dist
 ninja -C build install
 
@@ -16,7 +16,7 @@ ninja -C build install
 TEMPLATE=${SOURCE}/launcher.sh
 sed "s,#ROOT#,${ROOT},; s,#EXECUTABLE#,libexec/fwupd/fwupd," \
         ${TEMPLATE} > ${ROOT}/dist/fwupd.sh
-sed "s,#ROOT#,${ROOT},; s,#EXECUTABLE#,libexec/fwupd/fwupdtool," \
+sed "s,#ROOT#,${ROOT},; s,#EXECUTABLE#,bin/fwupdtool," \
         ${TEMPLATE} > ${ROOT}/dist/fwupdtool.sh
 sed "s,#ROOT#,${ROOT},; s,#EXECUTABLE#,bin/fwupdmgr," \
         ${TEMPLATE} > ${ROOT}/dist/fwupdmgr.sh
--- a/contrib/vscode/launcher.sh
+++ b/contrib/vscode/launcher.sh
@@ -2,6 +2,7 @@
 export ROOT=#ROOT#
 export FWUPD_LOCALSTATEDIR=${ROOT}/dist
 export FWUPD_SYSCONFDIR=${ROOT}/dist/etc
+export LD_LIBRARY_PATH=${ROOT}/dist/lib/$(gcc -dumpmachine)
 if [ -n "${DEBUG}" ]; then
         DEBUG="gdbserver localhost:9091"
 fi
--- a/plugins/dell-dock/README.md
+++ b/plugins/dell-dock/README.md
@@ -47,6 +47,7 @@ These devices use several different generation schemes, e.g.
  * Update Level: `USB\VID_413C&PID_B06E&hub&status`
  * MST Hub: `MST-panamera-vmm5331-259`
  * Thunderbolt Controller: `TBT-00d4b070`
+ * USB4 Controller: `TBT-00d4b071`
 
 Update Behavior
 ---------------
--- a/plugins/dell-dock/dell-dock.quirk
+++ b/plugins/dell-dock/dell-dock.quirk
@@ -63,7 +63,6 @@ Icon = dock-usb
 FirmwareSizeMin = 0x1FFC0
 FirmwareSizeMax = 0x20000
 Flags = require-ac,dual-image,self-recovery,usable-during-update
-Children = FuDellDockStatus|USB\VID_413C&PID_B06E&hub&status,FuDellDockMst|MST-panamera-vmm5331-259
 DellDockUnlockTarget = 1
 DellDockBoardMin = 6
 DellDockVersionLowest = 01.00.00.00
@@ -72,6 +71,17 @@ InstallDuration = 60
 
 # Representation of overall dock update
 [USB\VID_413C&PID_B06E&hub&status]
+Name = Package level of Dell dock
+Summary = A representation of dock update status
+Plugin = dell_dock
+Vendor = Dell Inc.
+Flags = self-recovery,usable-during-update
+FirmwareSize = 24
+InstallDuration = 5
+DellDockBlobVersionOffset = 0x14
+
+# Representation of overall dock update
+[USB\VID_413C&PID_B06E&hub&salomon_mlk_status]
 Name = Package level of Dell dock
 Summary = A representation of dock update status
 Plugin = dell_dock
@@ -115,3 +125,16 @@ DellDockUnlockTarget = 10
 DellDockHubVersionLowest = 1.31
 DellDockBlobMajorOffset = 0x400a
 DellDockBlobMinorOffset = 0x4009
+
+# USB4 controller
+[TBT-00d4b071]
+Name = USB4 controller in Dell dock
+Summary = USB4 controller
+Vendor = Dell Inc.
+VendorId = TBT:0x00D4
+ParentGuid = USB\VID_413C&PID_B06E&hub&embedded
+FirmwareSizeMin=0x40000
+FirmwareSizeMax=0x80000
+Flags = skips-restart,require-ac,dual-image
+Icon = thunderbolt
+InstallDuration = 46
--- a/plugins/dell-dock/fu-dell-dock-i2c-ec.c
+++ b/plugins/dell-dock/fu-dell-dock-i2c-ec.c
@@ -106,6 +106,7 @@ typedef enum {
 	MODULE_TYPE_130_UNIVERSAL,
 	MODULE_TYPE_240_TRIN,
 	MODULE_TYPE_210_DUAL,
+	MODULE_TYPE_130_USB4,
 } FuDellDockDockModule;
 
 typedef struct __attribute__ ((packed)) {
@@ -176,6 +177,13 @@ fu_dell_dock_ec_set_board (FuDevice *device)
 		fu_device_set_summary (device, summary);
 }
 
+gboolean
+fu_dell_dock_module_is_usb4 (FuDevice *device)
+{
+	FuDellDockEc *self = FU_DELL_DOCK_EC (device);
+	return self->data->module_type == MODULE_TYPE_130_USB4;
+}
+
 const gchar *
 fu_dell_dock_ec_get_module_type (FuDevice *device)
 {
@@ -195,6 +203,8 @@ fu_dell_dock_ec_get_module_type (FuDevice *device)
 		return "240 (Trinity)";
 	case MODULE_TYPE_210_DUAL:
 		return "210 (Dual)";
+	case MODULE_TYPE_130_USB4:
+		return "130 (TBT4)";
 	default:
 		return "unknown";
 	}
@@ -421,7 +431,8 @@ fu_dell_dock_ec_get_dock_info (FuDevice *device,
 			g_debug ("\tParsed version %s", self->mst_version);
 		} else if (map->device_type == FU_DELL_DOCK_DEVICETYPE_TBT &&
 			   (self->data->module_type == MODULE_TYPE_130_TBT ||
-			    self->data->module_type == MODULE_TYPE_45_TBT)) {
+			    self->data->module_type == MODULE_TYPE_45_TBT  ||
+			    self->data->module_type == MODULE_TYPE_130_USB4)) {
 			/* guard against invalid Thunderbolt version read from EC */
 			if (!fu_dell_dock_test_valid_byte (device_entry[i].version.version_8, 2)) {
 				g_warning ("[EC bug] EC read invalid Thunderbolt version %08x",
@@ -455,23 +466,11 @@ fu_dell_dock_ec_get_dock_info (FuDevice *device,
 
 	/* Thunderbolt SKU takes a little longer */
 	if (self->data->module_type == MODULE_TYPE_130_TBT ||
-	    self->data->module_type == MODULE_TYPE_45_TBT) {
+	    self->data->module_type == MODULE_TYPE_45_TBT  ||
+	    self->data->module_type == MODULE_TYPE_130_USB4) {
 		guint64 tmp = fu_device_get_install_duration (device);
 		fu_device_set_install_duration (device, tmp + 20);
 	}
-
-	/* minimum EC version this code will support */
-	if (fu_common_vercmp_full (self->ec_version,
-				   self->ec_minimum_version,
-				   FWUPD_VERSION_FORMAT_QUAD) < 0) {
-		g_set_error (error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED,
-			     "dock containing EC version %s is not supported",
-			     self->ec_version);
-		return FALSE;
-	}
-
-	fu_device_set_version_lowest (device, self->ec_minimum_version);
-
 
 	/* Determine if the passive flow should be used when flashing */
 	hub_version = fu_device_get_version (fu_device_get_proxy (device));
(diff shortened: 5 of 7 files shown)


plugins/dell-dock/README.md
```diff
@@ -71,6 +72,7 @@ This plugin uses the following plugin-specific quirks:
 | `DellDockBlobVersionOffset`  | The offset of the ASCII representation of a version string in a payload | 1.1.3                 |
 | `DellDockBoardMin`           | The minimum board revision required to safely operate the plugin        | 1.1.3                 |
 | `DellDockVersionLowest`      | The minimum component version required to safely operate the plugin     | 1.1.3                 |
+| `DellDockUSB4VersionLowest`  | The minimum component version required to use USB4 module               | 1.6.0                 |
```
CragW, line 75: How to manage this if we backport to older branch e.g. 1.3.X?

plugins/dell-dock/dell-dock.quirk
```diff
@@ -63,10 +63,10 @@ Icon = dock-usb
 FirmwareSizeMin = 0x1FFC0
 FirmwareSizeMax = 0x20000
 Flags = require-ac,dual-image,self-recovery,usable-during-update
-Children = FuDellDockStatus|USB\VID_413C&PID_B06E&hub&status,FuDellDockMst|MST-panamera-vmm5331-259
 DellDockUnlockTarget = 1
 DellDockBoardMin = 6
 DellDockVersionLowest = 01.00.00.00
+DellDockUSB4VersionLowest = 00.00.00.01
```
CragW, line 69: In development stage

contrib/vscode/build.sh
```diff
@@ -8,15 +8,15 @@ if [ -z "$ROOT" ]; then
 fi
 
 # build in tree
-rm -rf build ${ROOT}/dist
+sudo rm -rf build ${ROOT}/dist
```
superm1, line 11: can you please prefix the commit message for this file with `trivial:`

plugins/dell-dock/README.md
superm1, line 75: this should be 1.6.1

superm1: >How to manage this if we backport to older branch e.g. 1.3.X?

This information is most useful for building a CAB and what version you put as minimum in the XML.  I think when we backport we can adjust it as far back as we want, but the thing to keep in mind is that all the "in between" versions won't be able to handle it.  So if you put 1.3.13 then 1.4.x, 1.5.x and 1.6.0 really **won't** be able to handle it.

plugins/dell-dock/dell-dock.quirk
```diff
@@ -115,3 +126,20 @@ DellDockUnlockTarget = 10
...
+ParentGuid = USB\VID_413C&PID_B06E&hub&embedded
+FirmwareSizeMin=0x40000
+FirmwareSizeMax=0x80000
+Flags = skips-restart,require-ac,dual-image
+Icon = thunderbolt
+InstallDuration = 26
+DellDockUnlockTarget = 10
+DellDockHubVersionLowest = 1.31
+DellDockBlobMajorOffset = 0x400a
+DellDockBlobMinorOffset = 0x4009
```
superm1, line 145: I don't think you want these keys - they're only used for I2C based updater, no?

plugins/dell-dock/dell-dock.quirk
superm1: Shouldn't this version be later than 01.00.00.00 though?

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -52,6 +52,18 @@ typedef enum {
...
 
+typedef enum {
+	DOCK_MODE_CAP_TBT        = 1 << 0,
+	DOCK_MODE_CAP_DP         = 1 << 1,
+	DOCK_MODE_CAP_USB3       = 1 << 2,
+	DOCK_MODE_CAP_DOCKING    = 1 << 3,
+	DOCK_MODE_CAP_EXT_POWER  = 1 << 4,
+	DOCK_MODE_CAP_USB4       = 1 << 5,
+	DOCK_MODE_CAP_USB2       = 1 << 6,
+	DOCK_MODE_CAP_SOUCE_BACK = 1 << 7,
```
superm1, line 63: is this a misspelling?

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -106,6 +118,7 @@ typedef enum {
 	MODULE_TYPE_130_UNIVERSAL,
 	MODULE_TYPE_240_TRIN,
 	MODULE_TYPE_210_DUAL,
+	MODULE_TYPE_130_USB4,
```
superm1, line 121: There is no differentiation for greater or lesser voltages connected to USB4 module?

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -204,14 +227,19 @@ gboolean
...
+		( DOCK_MODE_CAP_TBT | DOCK_MODE_CAP_USB4 );
 
 	/* check for TBT module type */
-	if (self->data->module_type != MODULE_TYPE_130_TBT &&
-	    self->data->module_type != MODULE_TYPE_45_TBT)
-		return FALSE;
+	switch (self->data->module_type){
+		case MODULE_TYPE_130_TBT:
+		case MODULE_TYPE_45_TBT:
+		case MODULE_TYPE_130_USB4:
```
superm1, line 237: I'm not convinced this case should be here - isn't this code going to cause an I2C endpoint to be made?  You don't want that for the USB4 module.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -455,22 +485,28 @@ fu_dell_dock_ec_get_dock_info (FuDevice *device,
...
-	    self->data->module_type == MODULE_TYPE_45_TBT) {
+	    self->data->module_type == MODULE_TYPE_45_TBT  ||
+	    self->data->module_type == MODULE_TYPE_130_USB4) {
 		guint64 tmp = fu_device_get_install_duration (device);
 		fu_device_set_install_duration (device, tmp + 20);
 	}
 
 	/* minimum EC version this code will support */
+	if (self->data->module_type == MODULE_TYPE_130_USB4)
+		opt_ec_minimum_version = self->ec_minimum_version_usb4;
```
superm1, line 496: I'm actually not convinced that all the code for the minimum EC version for USB4 module makes sense to check here.  If you have a version without USB4 support doesn't that mean it won't know about USB4 enums/types/etc?  

Where it _might_ be useful to check is during the write function.

If the user is trying to install an "older" EC image and you have USB4 module then it shouldn't be allowed to install. However I think that idea falls apart if the EC image used doesn't start at _least_ with the latest image that is available for WD19 today.  Development versions of 00.xx.yy.zz it breaks.

So in summary; I think you should drop this idea for now, and consider to add it back in for safety later.

superm1: One more thought coming to mind though - if development ECs are around less than 01.00.00.00 for USB4 module, then maybe the minimum version check should be **removed all together** when the module is USB4..

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: it's custom that notebook is capable of sourcing back power to dock

plugins/dell-dock/dell-dock.quirk
CragW: they start from 0.x  :)

plugins/dell-dock/README.md
CragW: ok, in next patch.

CragW: Thanks for the explanation, in order to support early branch tool I think XML will keep it open not saying a specific version.

contrib/vscode/build.sh
CragW: ok, in next patch.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
superm1: So it's just missing an 'r'

CragW: Thanks for the catch! :)

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: This ID just indicates the type of cable and the wattage designation just identifies the maximum wattage that is supported by the cable.

plugins/dell-dock/dell-dock.quirk
CragW: yes, i2c-tbt.c isn't needed. I will remove the offset pattern.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: That's right, I don't need i2c device for usb4 regardless what the port status is.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: > If you have a version without USB4 support doesn't that mean it won't know about USB4 enums/types/etc? Where it might be useful to check is during the write function.

I think `dock info` in general is well supported. The check for `ec_minimum_version` is more useful in `write` func. to prevent user accidentally downgrade to USB4's development version via `--force`.

> if development ECs are around less than 01.00.00.00 for USB4 module, then maybe the minimum version check should be removed all together when the module is USB4..

At this time I think `ec_minimum_version` in `write` func. is useful for `none-USB4` only. Depends on the need of USB4 hardware if there's a denial to downgrade to a particular version then we can add it back.

superm1: > I think dock info in general is well supported. The check for ec_minimum_version is more useful in write func. to prevent user accidentally downgrade to USB4's development version via --force.

So perhaps you can switch the logic around to move the check for minimum version to write func and set up logic something like this:
1. Check for force and ignore all checks if so.
2. If not force: module type is USB4, ignore checks if so.
3. If not force and not usb4 module check If target version is less than minimum version AND current version is greater than minimum version
 
And then later when USB4 EC is stable and closer to launch this logic an be adjusted again.

plugins/dell-dock/dell-dock.quirk
CragW: Removed `DellDockUSB4VersionLowest` as needn't.

contrib/vscode/build.sh
CragW: Done.

plugins/dell-dock/README.md
CragW: Removed `DellDockUSB4VersionLowest` as needn't.

plugins/dell-dock/dell-dock.quirk
CragW: Done.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: Removed for now.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: Done.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: Done, thanks for the flow suggestion.

contrib/vscode/launcher.sh
```diff
@@ -2,6 +2,7 @@
 export ROOT=#ROOT#
 export FWUPD_LOCALSTATEDIR=${ROOT}/dist
 export FWUPD_SYSCONFDIR=${ROOT}/dist/etc
+export LD_LIBRARY_PATH=${ROOT}/dist/lib/x86_64-linux-gnu
```
superm1, line 5: can this instead be

```
export LD_LIBRARY_PATH=${ROOT}/dist/lib/x86_64-linux-gnu/$(gcc -dumpmachine)
```

plugins/dell-dock/README.md
```diff
@@ -47,6 +47,7 @@ These devices use several different generation schemes, e.g.
  * Update Level: `USB\VID_413C&PID_B06E&hub&status`
  * MST Hub: `MST-panamera-vmm5331-259`
  * Thunderbolt Controller: `TBT-00d4b070`
+ * Thunderbolt Controller (USB4): `TBT-00d4b071`
```
superm1, line 50: nit: I would suggest just call this "USB4 Controller"

plugins/dell-dock/dell-dock.quirk
```diff
@@ -115,3 +125,18 @@ DellDockUnlockTarget = 10
...
+Vendor = Dell Inc.
+VendorId = TBT:0x00D4
+ParentGuid = USB\VID_413C&PID_B06E&hub&embedded
+FirmwareSizeMin=0x40000
+FirmwareSizeMax=0x80000
+Flags = skips-restart,require-ac,dual-image
+Icon = thunderbolt
+InstallDuration = 26
+DellDockUnlockTarget = 10
+DellDockHubVersionLowest = 1.31
```
superm1, line 142: I don't think these keys are used either.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -781,9 +780,23 @@ fu_dell_dock_ec_write_fw (FuDevice *device,
...
 	data = g_bytes_get_data (fw, &fw_size);
-	write_size =  (fw_size / HIDI2C_MAX_WRITE) >= 1 ? HIDI2C_MAX_WRITE : fw_size;
-
 	dynamic_version = g_strndup ((gchar *) data + self->blob_version_offset, 11);
+
+	/* meet the minimum EC version */
+	if ((flags & FWUPD_INSTALL_FLAG_FORCE) == 0) {
+		if (self->data->module_type != MODULE_TYPE_130_USB4) {
+			if (fu_common_vercmp_full (dynamic_version, self->ec_minimum_version,
+				   FWUPD_VERSION_FORMAT_QUAD) < 0) {
```
superm1, line 789: new logic looks good, but can you align the text better?  project has it doing things like this instead:

```
g_set_error (error, 
	     FWUPD_ERROR, 
	     FWUPD_ERROR_NOT_SUPPORTED,
	     "New EC version %s is less than minimum required %s",
	     dynamic_version, self->ec_minimum_version);
```
when it's a long one.

superm1, line 789: likewise, align the `FWUPD_VERSION_FORMAT_QUAD` to `dynamic_version`

superm1 (commented):
just some changes and nits left

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -781,9 +780,23 @@ fu_dell_dock_ec_write_fw (FuDevice *device,
...
 		return FALSE;
 	data = g_bytes_get_data (fw, &fw_size);
-	write_size =  (fw_size / HIDI2C_MAX_WRITE) >= 1 ? HIDI2C_MAX_WRITE : fw_size;
-
 	dynamic_version = g_strndup ((gchar *) data + self->blob_version_offset, 11);
+
+	/* meet the minimum EC version */
+	if ((flags & FWUPD_INSTALL_FLAG_FORCE) == 0) {
+		if (self->data->module_type != MODULE_TYPE_130_USB4) {
+			if (fu_common_vercmp_full (dynamic_version, self->ec_minimum_version,
```
superm1, line 788: since it's really just one case, maybe combine these "3" if statements with an `&&`?

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -781,9 +780,23 @@ fu_dell_dock_ec_write_fw (FuDevice *device,
 	if (fw == NULL)
 		return FALSE;
 	data = g_bytes_get_data (fw, &fw_size);
-	write_size =  (fw_size / HIDI2C_MAX_WRITE) >= 1 ? HIDI2C_MAX_WRITE : fw_size;
-
```
superm1, line 785: I don't think you need to move these lines do you?

contrib/vscode/launcher.sh
CragW: Thanks!

plugins/dell-dock/dell-dock.quirk
```diff
@@ -115,3 +125,16 @@ DellDockUnlockTarget = 10
...
+Name = USB4 controller in Dell dock
+Summary = USB4 controller
+Vendor = Dell Inc.
+VendorId = TBT:0x00D4
+ParentGuid = USB\VID_413C&PID_B06E&hub&embedded
+FirmwareSizeMin=0x40000
+FirmwareSizeMax=0x80000
+Flags = skips-restart,require-ac,dual-image
+Icon = thunderbolt
+InstallDuration = 46
```
CragW, line 140: I thought the extended duration for `tbt` is a buffer therefore the `InstallDuration` here changed to `46` (previously `26` takes account of the `20` sec tolerance)

```
	/* Thunderbolt SKU takes a little longer */
	if (self->data->module_type == MODULE_TYPE_130_TBT ||
	    self->data->module_type == MODULE_TYPE_45_TBT  ||
	    self->data->module_type == MODULE_TYPE_130_USB4) {
		guint64 tmp = fu_device_get_install_duration (device);
		fu_device_set_install_duration (device, tmp + 20);
	}

```

superm1 (approved):

PR #3316, opened 2021-06-05
More fixes for UEFI Capsule plugin on FreeBSD

SergiiDmytruk:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

***

* Use kenv variable to determine UEFI support (SMBIOS doesn't provide the information).
* Run Linux-specific checks only on Linux.
* Work around `bsdisks` not implementing parts of  UDisks2 (`Manager` object is missing, `Type` field contains FS type name).
* Several small corrections of related code.

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -15,6 +15,11 @@
 #include "fu-device-metadata.h"
 #include "fu-plugin-vfuncs.h"
 
+#ifdef __FreeBSD__
+#include "fu-common-version.h"
+#include "fu-kenv.h"
+#endif
+
 #include "fu-uefi-backend.h"
 #include "fu-uefi-bgrt.h"
 #include "fu-uefi-bootmgr.h"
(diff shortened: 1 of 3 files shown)


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -638,6 +667,20 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
...
+	if (!uefi_supported) {
+		g_error_free (error_local);
+		error_local = NULL;
+
+		/* check kenv for UEFI information */
+		uefi_supported = fu_plugin_uefi_capsule_kenv_enabled (&error_local);
+	}
+#endif
+
+	if (!uefi_supported) {
```
superm1, line 683: I think this new flow is really hard to follow with ifdef in the else blocks.  Can we please Make some static functions, or bring `fu_plugin_uefi_capsule_kenv_enabled` into `fu_plugin_uefi_capsule_smbios_enabled` perhaps?  It seems to me like this `fu_plugin_startup` could avoid all changes if you put all the BSD logic into that function

hughsie: I think splitting it out as fu-uefi-capsule-common-linux.c and fu-uefi-capsule-common-freebsd.c might make this a lot easier to maintain. I can help on Monday with this if that would be useful.

SergiiDmytruk: @hughsie, new files with functions that look something like this?
```c
gboolean fu_uefi_platform_startup_extra(void); // Linux: nothing, FreeBSD: check kenv
gboolean fu_uefi_platform_startup_fallback(void); // Linux: ignore smbios failure, FreeBSD: nothing
gboolean fu_uefi_platform_coldplug_extra(void); // Linux: check efivarsfs, FreeBSD: nothing
```

hughsie: My proposal would be something like https://github.com/fwupd/fwupd/pull/3318 -- I've merged the obviously correct easy ones directly to master thanks.

PR #3317, opened 2021-06-07
libfwupdplugin: Implement fu-efivar-freebsd.c

Asiderr:
Signed-off-by: Michał Kopeć <michal.kopec@3mdeb.com>

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-efivar-freebsd.c
+++ b/libfwupdplugin/fu-efivar-freebsd.c
@@ -1,0 +1,418 @@
+/*
+ * Copyright (C) 2021 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2021 Norbert Kamiński <norbert.kaminski@3mdeb.com>
+ * Copyright (C) 2021 Michał Kopeć <michal.kopec@3mdeb.com>
+ * Copyright (C) 2021 Sergii Dmytruk <sergii.dmytruk@3mdeb.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <errno.h>
+#include <fcntl.h>
+#include <string.h>
+#include <efivar.h>
+
+#include "fu-common.h"
+#include "fu-efivar.h"
+
+#include "fwupd-error.h"
+
+/**
+ * fu_efivar_supported:
+ * @error: #GError
+ *
+ * Determines if the kernel supports EFI variables
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_supported (GError **error)
+{
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+	if (efi_variables_supported () == 0) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_NOT_SUPPORTED,
+			     "kernel efivars support missing");
+		return FALSE;
+	}
+	return TRUE;
+}
+
+/**
+ * fu_efivar_space_used:
+ * @error: (nullable): optional return location for an error
+ *
+ * Gets the total size used by all EFI variables. This may be less than the size reported by the
+ * kernel as some (hopefully small) variables are hidden from userspace.
+ *
+ * Returns: total allocated size of all visible variables, or %G_MAXUINT64 on error
+ *
+ * Since: 1.6.1
+ **/
+guint64
+fu_efivar_space_used (GError **error)
+{
+	g_return_val_if_fail (error == NULL || *error == NULL, G_MAXUINT64);
+
+	guint64 total = 0;
+	efi_guid_t *guid = NULL;
+	char *name = NULL;
+
+	while (efi_get_next_variable_name (&guid, &name)) {
+		size_t size = 0;
+		if (efi_get_variable_size (*guid, name, &size) < 0) {
+			g_set_error (error,
+					G_IO_ERROR,
+					G_IO_ERROR_FAILED,
+					"failed to get efivar size");
+			return G_MAXUINT64;
+		}
+		total += size;
+	}
+
+	/* success */
+	return total;
+}
+
+/**
+ * fu_efivar_set_data:
+ * @guid: Globally unique identifier
+ * @name: Variable name
+ * @data: Data to set
+ * @sz: size of @data
+ * @attr: Attributes
+ * @error: (nullable): optional return location for an error
+ *
+ * Sets the data to a UEFI variable in NVRAM
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_set_data (const gchar *guids, const gchar *name, const guint8 *data,
+		     gsize sz, guint32 attr, GError **error)
+{
+	g_return_val_if_fail (guids != NULL, FALSE);
+	g_return_val_if_fail (name != NULL, FALSE);
+	g_return_val_if_fail (data != NULL, FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+	efi_guid_t guid;
+	efi_str_to_guid (guids, &guid);
+
+	if (efi_set_variable (guid, name, data, sz, attr) != 0) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "failed to write data to efivar %s", name);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+/**
+ * fu_efivar_delete:
+ * @guid: Globally unique identifier
+ * @name: Variable name
+ * @error: #GError
+ *
+ * Removes a variable from NVRAM
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_delete (const gchar *guids, const gchar *name, GError **error)
+{
+	g_return_val_if_fail (guids != NULL, FALSE);
+	g_return_val_if_fail (name != NULL, FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+	efi_guid_t guid;
+	efi_str_to_guid (guids, &guid);
+
+	if (efi_del_variable (guid, name) == 0)
+		return TRUE;
+
+	g_set_error (error,
+		     G_IO_ERROR,
+		     G_IO_ERROR_INVALID_DATA,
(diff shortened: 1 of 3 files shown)


libfwupdplugin/meson.build
```diff
@@ -125,12 +124,19 @@ library_deps = [
   valgrind,
   libjcat,
   platform_deps,
+  efivar,
```
hughsie, line 127: this needs to be surrounded by `host_machine.system() == 'freebsd'` no?

meson.build
```diff
@@ -249,6 +249,7 @@ if build_daemon
   endif
 endif
 libm = cc.find_library('m', required: false)
+efivar = cc.find_library('efivar', required: false)
```
hughsie, line 252: host_machine.system() == 'freebsd' and then drop the `required: false` i think

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,402 @@
...
+ * fu_efivar_get_path, fu_efivar_get_filename:
+ * Linux-only functions, kept here temporarily for compatibility with
+ * plugins that haven't been adapted yet
+ */
+
+static gchar *
+fu_efivar_get_path (void)
+{
+	g_autofree gchar *sysfsfwdir = fu_common_get_path (FU_PATH_KIND_SYSFSDIR_FW);
+	return g_build_filename (sysfsfwdir, "efi", "efivars", NULL);
```
hughsie, line 32: I think just return NULL perhaps?

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,402 @@
...
+{
+	g_return_val_if_fail (guids != NULL, FALSE);
+
+	/* any name */
+	if (name == NULL)
+		return fu_efivar_exists_guid (guids);
+
+	if (!fu_efivar_get_data (guids, name, NULL, NULL, NULL, NULL))
+		return FALSE;
+	return TRUE;
```
hughsie, line 218: just `return fu_efivar_get_data (...` perhaps?

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,402 @@
...
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_get_data (const gchar *guids, const gchar *name, guint8 **data,
+		       gsize *data_sz, guint32 *attr, GError **error)
+{
+	efi_guid_t guid;
+	efi_str_to_guid (guids, &guid);
+
+	if (efi_get_variable (guid, name, data, data_sz, attr) < 0)
```
hughsie, line 243: same

libfwupdplugin/meson.build
mkopec: Done.

meson.build
mkopec: Done.

libfwupdplugin/fu-efivar-freebsd.c
mkopec: Right, of course. Done.

libfwupdplugin/fu-efivar-freebsd.c
mkopec: Also done.

libfwupdplugin/fu-efivar-freebsd.c
mkopec: Fixed.

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_supported (GError **error)
+{
+	if (efi_variables_supported () == 0)
+		return FALSE;
```
hughsie, line 54: I think you need a g_set_error here

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_get_data (const gchar *guids, const gchar *name, guint8 **data,
+		       gsize *data_sz, guint32 *attr, GError **error)
+{
+	efi_guid_t guid;
+	efi_str_to_guid (guids, &guid);
+
+	return !(efi_get_variable (guid, name, data, data_sz, attr))
```
hughsie, line 239: remove outer brackets perhaps? or check for != 0 perhaps?

libfwupdplugin/meson.build
SergiiDmytruk: @hughsie, I've seen there is `platform_deps`. Should it be used in this case or it's not an appropriate usage?

hughsie: I guess you could use that instead. I don't mind either way.

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+ **/
+gboolean
+fu_efivar_set_data (const gchar *guids, const gchar *name, const guint8 *data,
+		     gsize sz, guint32 attr, GError **error)
+{
+	efi_guid_t guid;
+	efi_str_to_guid (guids, &guid);
+
+	if (efi_set_variable (guid, name, data, sz, attr) != 0) {
+		g_prefix_error (error, "failed to write data to efivar: ");
```
SergiiDmytruk, line 112: I think `g_prefix_error()` isn't appropriate here, on input `error` or `*error` should be `NULL`, so an error should just be set in this case. You also might want to add
```c
g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
to the functions that accept an error and set it, because most of them have the check and some of new functions are missing it.

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+ **/
+gboolean
+fu_efivar_delete (const gchar *guids, const gchar *name, GError **error)
+{
+	efi_guid_t guid;
+	efi_str_to_guid (guids, &guid);
+
+	if (efi_del_variable (guid, name) == 0)
+		return TRUE;
+	return FALSE;
```
SergiiDmytruk, line 140: Error message is not being set.

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 1.6.1
+ **/
+gboolean
+fu_efivar_delete_with_glob (const gchar *guids, const gchar *name_glob, GError **error)
+{
+	efi_guid_t guid_to_delete;
+	efi_str_to_guid (guids, &guid_to_delete);
```
SergiiDmytruk, line 159: The call should probably follow all declarations.

SergiiDmytruk (changes requested):

libfwupdplugin/meson.build
```diff
@@ -131,6 +130,13 @@ if get_option('libarchive')
   library_deps += libarchive
 endif
 
+if host_machine.system() == 'freebsd'
+  fwupdplugin_src += 'fu-efivar-freebsd.c'
+  library_deps += efivar
```
superm1, line 135: isn't this `platform_deps`?

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+#include "fu-efivar.h"
+
+#include "fwupd-error.h"
+
+/**
+ * fu_efivar_get_path, fu_efivar_get_filename:
+ * Linux-only functions, kept here temporarily for compatibility with
+ * plugins that haven't been adapted yet
+ */
+
```
superm1, line 27: these are static functions, I don't think you need them in the freebsd versions.

libfwupdplugin/fu-efivar-freebsd.c
superm1: I agree - this should should be `g_set_error`

superm1 (commented):
so something interesting that _might_ happen here is that we have different "public" symbols in different implementations.  I wonder if we need something to avoid problems or confusion happening?

libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -0,0 +1,396 @@
...
+	guint64 total = 0;
+	efi_guid_t *guid = NULL;
+	char *name = NULL;
+
+	g_return_val_if_fail (error == NULL || *error == NULL, G_MAXUINT64);
+
+	while (efi_get_next_variable_name (&guid, &name)) {
+		size_t size = 0;
+		if (efi_get_variable_size (*guid, name, &size) < 0)
+			return G_MAXUINT64;
```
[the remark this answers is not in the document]
mkopec: Added a g_set_error

libfwupdplugin/fu-efivar-freebsd.c
mkopec: Added a g_set_error.

libfwupdplugin/fu-efivar-freebsd.c
mkopec: Fixed.

libfwupdplugin/meson.build
mkopec: Moved `efivar` to `platform_deps`.

libfwupdplugin/fu-efivar-freebsd.c
mkopec: You're right. Fixed.

hughsie:
> is that we have different "public" symbols in different implementations

The way I did that in 694e774c2a3f98fd790a412d591cee28e19e0f28 was to use a _impl suffix and call the correct private implementation from that public symbol. I agree it sucks to have to duplicate the gtk-doc comments like that.

SergiiDmytruk (approved):
Approving to remove request for changes.

libfwupdplugin/fu-efivar-freebsd.c
mkopec: Changed to a `!= 0` check.

hughsie:
Give me 30 minutes with this branch in a dark room.

hughsie:
Alternate proposal in https://github.com/fwupd/fwupd/pull/3330 -- it's the same code, just shuffled around a bit.

hughsie:
I merged the other one, thanks @Asiderr !

Asiderr:
@hughsie Thanks are due to @mkopec :)

PR #3326, opened 2021-06-08
Use the parent physical ID to autoset the device parent

hughsie:
Of course, this only works if the parent and child both *export* a
FuDevice of the same physical kind, e.g. USB.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device-private.h
+++ b/libfwupdplugin/fu-device-private.h
@@ -17,6 +17,9 @@ FuDeviceInternalFlags fu_device_internal_flag_from_string (const gchar	*flag);
 GPtrArray	*fu_device_get_parent_guids		(FuDevice	*self);
 gboolean	 fu_device_has_parent_guid		(FuDevice	*self,
 							 const gchar	*guid);
+GPtrArray	*fu_device_get_parent_physical_ids	(FuDevice	*self);
+gboolean	 fu_device_has_parent_physical_id	(FuDevice	*self,
+							 const gchar	*physical_id);
 void		 fu_device_set_parent			(FuDevice	*self,
 							 FuDevice	*parent);
 gint		 fu_device_get_order			(FuDevice	*self);
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -51,6 +51,7 @@ typedef struct {
 	GRWLock				 metadata_mutex;
 	GPtrArray			*parent_guids;
 	GRWLock				 parent_guids_mutex;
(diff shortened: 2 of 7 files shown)


hughsie:
> how common is that really?

Basically, any non Dell docking station.

> It's basically saying "these USB devices are connected to these USB hubs".

Yes.

superm1:
I don't think you want this behavior by default though.  You're going to potentially end up with Logitech USB receivers that are part of composite devices of their parent device.  Or USB peripherals connected to a dock that get included in the parent.  I feel like it should be an extra flag to do this behavior on an "opt-in" basis.

memily:
Thank you for waiting, we have English wording questions: 
By physical id do you mean top of the bus stack, root hub (assuming usb up-port, down-port)?
And by platform id do you mean the id of devices on i2c bus?

A tree structure makes sense, but we don't see how physical id will help initiate cross plugin update. Our point of confusion is --after grouping all the chips on a composite device together for enumerating-- how will each chip get matched into its respective vendor's plugin for update? We had assumed this was done in quirk files, if that helps.

superm1 (commented):
I don't know that it makes sense to add all these new symbols into the plugin ABI when realistically it's only the engine using them.

Maybe just the `add_parent_physical_id`, and put the other two in the engine as static functions?

hughsie:
> I don't know that it makes sense to add all these new symbols into the plugin ABI

I'm pretty sure that if the daemon needs to use a symbol in libfwupdplugin then it needs to be exported.

superm1:
> > I don't know that it makes sense to add all these new symbols into the plugin ABI
> 
> I'm pretty sure that if the daemon needs to use a symbol in libfwupdplugin then it needs to be exported.

Right but those could be static functions in the daemon instead. Like in `fu-engine.c`

hughsie:
> Right but those could be static functions in the daemon instead. Like in fu-engine.c

    libfwupdplugin/fu-device-private.h |    2 --
    libfwupdplugin/fu-device.c         |   20 ++------------------
    libfwupdplugin/fwupdplugin.map     |    1 -
    src/fu-engine.c                    |   16 +++++++++++++++-
    4 files changed, 17 insertions(+), 22 deletions(-)

This is because we need to check if the ID exists in fu_device_add_parent_physical_id -- I think keeping the _has_parent_id is probably best.

superm1 (approved):

hughsie:
Okay, excellent; I'll dig out a lot of hardware on Monday and remove some `ParentGuid`s.

PR #3330, opened 2021-06-09
libfwupdplugin: Implement fu-efivar-freebsd.c

hughsie:

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/meson.build
+++ b/libfwupdplugin/meson.build
@@ -37,11 +37,14 @@ fwupdplugin_src = [
 ]
 
 if host_machine.system() == 'linux'
-  fwupdplugin_src += 'fu-common-linux.c'  # fuzzing
+  fwupdplugin_src += 'fu-common-linux.c'    # fuzzing
+  fwupdplugin_src += 'fu-efivar-linux.c'
 elif host_machine.system() == 'freebsd'
   fwupdplugin_src += 'fu-common-freebsd.c'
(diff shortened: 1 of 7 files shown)


libfwupdplugin/meson.build
```diff
@@ -37,11 +37,14 @@ fwupdplugin_src = [
...
 if host_machine.system() == 'linux'
-  fwupdplugin_src += 'fu-common-linux.c'  # fuzzing
+  fwupdplugin_src += 'fu-common-linux.c'    # fuzzing
+  fwupdplugin_src += 'fu-efivar-linux.c'
 elif host_machine.system() == 'freebsd'
   fwupdplugin_src += 'fu-common-freebsd.c'
+  fwupdplugin_src += 'fu-efivar-freebsd.c'
 elif host_machine.system() == 'windows'
   fwupdplugin_src += 'fu-common-windows.c'
+  fwupdplugin_src += 'fu-efivar-windows.c'  # fuzzing
```
superm1, line 47: by splitting this way I don't think this code will ever be fuzzed

hughsie: That was intentional; i.e. we don't actually link giounix into the fuzzer binary. We certainly _could_, but the efivar parsing isn't something that lends itself to fuzzing.

superm1 (approved):

PR #3332, opened 2021-06-09
trivial: Fix the LGTM build

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


.lgtm.yml
```diff
@@ -26,7 +22,9 @@ extraction:
     - "make install"
     - "export PKG_CONFIG_PATH=$LGTM_WORKSPACE/installdir/usr/lib/pkgconfig:$PKG_CONFIG_PATH"
     - "export LD_LIBRARY_PATH=$LGTM_WORKSPACE/installdir/usr/lib:$LD_LIBRARY_PATH"
+    configure:
+      command:
+      - "meson setup _lgtm_build_dir -Defi_binary=false -Dplugin_uefi_capsule_splash=false"
```
superm1, line 27: It seems like the arguments you did are right, but it's not rebuilding on force push or something since I see the old command called.  I'd suggest just merge with those arguments (and drop the DO NOT COMMIT commit)

hughsie: Agree. Thanks for the sanity check.

PR #3327, opened 2021-06-09
dell-dock: add usb protocol for USB4 device

CragW:
Type of pull request:
- [ ] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [ ] Code fix
- [x] Feature
- [ ] Documentation

PR #3333, opened 2021-06-09
trivial: don't set priority between thunderbolt and dell-dock plugins

superm1:
The dell-dock plugin has a check whether or not to create the I2C based
child device based upon whether thunderbolt link is active during probe.

So there will never be a situation that daemon needs to de-duplicate and
set priority between the two plugins.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

Stemmed from this conversation: https://github.com/fwupd/fwupd/pull/3327#discussion_r648245102
CC @CragW


hughsie (approved):
I trust your judgement here dude.

PR #3334, opened 2021-06-09
Obtain firmware major and minor versions from SMBIOS

mkopec:
Add EC firmware major and minor versions to the list of hwids obtained from SMBIOS.
This will be used to add support for EC updating using the flashrom plugin.

Signed-off-by: Michał Kopeć <michal.kopec@3mdeb.com>

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie (approved):

superm1:
Doesn't this need some documentation expansion? Gtkdoc or otherwise?

hughsie:
I think *all* the keys need documentation. talking of which, how close are we to the gi-docgen stuff being merged? Then I can polish up stuff like this as part of that too.

PR #3339, opened 2021-06-10
Draft: Add tuxec plugin

mkopec:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

This is a proposal for a new plugin for flashing the embedded controller in Tuxedo laptops. It's based on (it's a copy of) the flashrom plugin, but is separate from it for the following reasons:
- The flashrom plugin (currently) does not support more than one internal device - and the machines will also have BIOS updates applied via flashrom,
- I tried creating a device class similar to how `fu-flashrom-lspcon-i2c-device` does it, but could not get it to load instead of `fu-flashrom-internal-device` - and eventually we'd have to have both of them load at the same time, which doesn't seem possible with the current quirk system,
- The EC flashing procedure is very specific and needs manual intervention by the user, which seemed easier to implement in a separate plugin (not implemented yet)

Let me know whether this is a good / sane approach, or something completely different would be preferable.

Currently all it does is create a `FuTuxecInternalDevice` and displays the current EC FW version.


plugins/tuxec/tuxec.quirk
```diff
@@ -0,0 +1,3 @@
+# Tuxedo InifinityBook S15 Gen6
+[60f53465-e8fc-5122-b79b-f7b03f063037]
+Plugin=tuxec
```
superm1, line 3: missing newline

plugins/tuxec/meson.build
```diff
@@ -0,0 +1,35 @@
+if get_option('plugin_tuxec')
```
superm1, line 1: I think this has a flashrom dependency too (ie `plugin_flashrom`)? right?

plugins/tuxec/fu-tuxec-internal-device.c
```diff
@@ -0,0 +1,158 @@
...
+	       FU_TYPE_TUXEC_DEVICE)
+
+static void
+fu_tuxec_internal_device_init (FuTuxecInternalDevice *self)
+{
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_INTERNAL);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
+	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
+	fu_device_add_instance_id (FU_DEVICE (self), "main-system-firmware");
```
superm1, line 28: I think this is usually reserved for BIOS, we're going to end up with two instance ID's of main-system-firmware possibly no?

hughsie:
How many lines of code is the flashrom part?

plugins/tuxec/fu-plugin-tuxec.c
```diff
@@ -0,0 +1,191 @@
...
+#define SELFCHECK_TRUE 1
+
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	FuContext *ctx = fu_plugin_get_context (plugin);
+
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "linux_lockdown");
+	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_CONFLICTS, "coreboot"); /* obsoleted */
```
superm1, line 41: I don't think this is needed on "new" plugins now right?

meson_options.txt
```diff
@@ -32,6 +32,7 @@ option('plugin_msr', type : 'boolean', value : true, description : 'enable MSR s
 option('plugin_flashrom', type : 'boolean', value : false, description : 'enable libflashrom support')
 option('plugin_platform_integrity', type : 'boolean', value : false, description : 'enable platform integrity support')
 option('plugin_intel_spi', type : 'boolean', value : false, description : 'enable Intel SPI support')
+option('plugin_tuxec', type : 'boolean', value : false, description : 'enable Tuxedo EC support')
```
superm1, line 35: can you make sure that at least one CI job is running this?

superm1:
>The EC flashing procedure is very specific and needs manual intervention by the user, which seemed easier to implement in a separate plugin (not implemented yet)

Yikes... Can you elaborate what that means manual intervention?

meson_options.txt
hughsie: I think back a bit; this is waaay too much copy and paste code.

plugins/flashrom/meson.build
```diff
@@ -1,4 +1,4 @@
-if get_option('plugin_flashrom')
+if get_option('plugin_flashrom') or get_option('plugin_tuxec')
```
superm1, line 1: although this is functional - I don't think it's the right thing to do.  If someone compiles with flashrom turned off but tuxec they'll still get flashrom.  What you should do is add a check in tuxec/meson.build that flashrom is turned on or off, and report an error if flashrom is off but tuxec is on.

meson_options.txt
macpijan: @hughsie I agree. That was an experiment to make something work, but it does not look like a correct path. 

We will try with the superio plugin approach as suggested.

PR #3343, opened 2021-06-10
Update ESRT access implementation for FreeBSD

SergiiDmytruk:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

***

Upstream kernel [patch](https://reviews.freebsd.org/D30104) was rejected in its original form. `sysctl()` interface was changed to `ioctl()` and it won't compile now without updated patch applied. Actually patch update wasn't yet sent upstream, but you can see it [here](https://github.com/3mdeb/freebsd-src/pull/2). We believe that's basically final form of the API.

I've also faced some issues with efivars, which weren't noticed during review. Maybe cherry-pick it separately.

### DIFF
--- a/libfwupdplugin/fu-efivar-freebsd.c
+++ b/libfwupdplugin/fu-efivar-freebsd.c
@@ -11,6 +11,7 @@
 
 #include <efivar.h>
 
+#include "fu-common.h"
 #include "fu-efivar-impl.h"
 
 #include "fwupd-error.h"
@@ -55,7 +56,7 @@ fu_efivar_delete_with_glob_impl (const gchar *guid, const gchar *name_glob, GErr
 	efi_str_to_guid (guid, &guid_to_delete);
 
 	while (efi_get_next_variable_name (&guidt, &name)) {
-		if (memcmp (&guid_to_delete, guidt, sizeof (guidt)) != 0)
+		if (memcmp (&guid_to_delete, guidt, sizeof (guid_to_delete)) != 0)
 			continue;
 		if (!fu_common_fnmatch (name, name_glob))
 			continue;
@@ -105,7 +106,6 @@ fu_efivar_get_data_impl (const gchar *guid, const gchar *name, guint8 **data,
 GPtrArray *
 fu_efivar_get_names_impl (const gchar *guid, GError **error)
 {
-	const gchar *name_guid;
 	g_autoptr(GPtrArray) names = g_ptr_array_new_with_free_func (g_free);
 	efi_guid_t *guidt = NULL;
 	gchar *name = NULL;
@@ -172,7 +172,7 @@ fu_efivar_set_data_impl (const gchar *guid, const gchar *name, const guint8 *dat
 	efi_guid_t guidt;
 	efi_str_to_guid (guid, &guidt);
 
-	if (efi_set_variable (guidt, name, data, sz, attr) != 0) {
+	if (efi_set_variable (guidt, name, (guint8 *)data, sz, attr) != 0) {
 		g_set_error (error,
 			     G_IO_ERROR,
 			     G_IO_ERROR_FAILED,
--- a/meson.build
+++ b/meson.build
@@ -281,7 +281,7 @@ if get_option('default_library') != 'static'
     platform_deps += cc.find_library('shlwapi')
   endif
   if host_machine.system() == 'freebsd'
-    platform_deps += cc.find_library('efivar')
+    platform_deps += dependency('efivar')
   endif
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -366,6 +366,12 @@ if cc.has_function('pwrite', args : '-D_XOPEN_SOURCE')
   conf.set('HAVE_PWRITE', '1')
 endif
 
+if host_machine.system() == 'freebsd'
+  if cc.has_type('struct efi_esrt_entry_v1', prefix: '#include <sys/types.h>\n#include <sys/efiio.h>')
+    conf.set('HAVE_FREEBSD_ESRT', '1')
+  endif
+endif
+
 if build_standalone and get_option('plugin_tpm')
   tpm2tss = dependency('tss2-esys', version : '>= 2.0')
   conf.set('HAVE_TSS2', '1')
--- a/plugins/uefi-capsule/fu-uefi-backend-freebsd.c
+++ b/plugins/uefi-capsule/fu-uefi-backend-freebsd.c
@@ -9,7 +9,16 @@
 
 #include <fwupdplugin.h>
 #include <sys/types.h>
-#include <sys/sysctl.h>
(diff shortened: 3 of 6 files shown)


libfwupdplugin/fu-efivar-freebsd.c
```diff
@@ -55,7 +56,7 @@ fu_efivar_delete_with_glob_impl (const gchar *guid, const gchar *name_glob, GErr
 	efi_str_to_guid (guid, &guid_to_delete);
 
 	while (efi_get_next_variable_name (&guidt, &name)) {
-		if (memcmp (&guid_to_delete, guidt, sizeof (guidt)) != 0)
+		if (memcmp (&guid_to_delete, guidt, sizeof (guid_to_delete)) != 0)
```
hughsie, line 59: isn't this the same?

plugins/uefi-capsule/fu-uefi-backend-freebsd.c
```diff
@@ -23,67 +29,42 @@ struct _FuUefiBackend {
...
-	g_autofree gchar *fw_class_fn = NULL;
 	g_autofree gchar *id = NULL;
-	g_autofree gchar *sysctl_name = NULL;
+	g_autofree gchar *phys_id = NULL;
+	uint32_t status;
 
-	sysctl_name = g_strdup_printf ("hw.efi.esrt.entry%u", (guint) idx);
-	fw_class_fn = g_build_path (".", sysctl_name, "fw_class", NULL);
-	fw_class = fu_uefi_backend_get_sysctl_string (fw_class_fn);
+	uuid_to_string (&entry->fw_class, &fw_class, &status);
```
hughsie, line 41: fwupd_guid_from_string?

plugins/uefi-capsule/fu-uefi-backend-freebsd.c
```diff
@@ -92,7 +73,10 @@ fu_uefi_backend_setup (FuBackend *backend, GError **error)
 {
 	g_autofree gchar *efi_ver = fu_kenv_get_string ("efi-version", error);
 	if (efi_ver == NULL) {
-		g_prefix_error (error, "System does not support UEFI mode, no efi-version kenv: ");
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "System does not support UEFI mode, no efi-version kenv");
```
hughsie, line 79: error is already set by fu_kenv_get_string

libfwupdplugin/fu-efivar-freebsd.c
SergiiDmytruk: No, `guidt` is a pointer, while `guid_to_delete` isn't.

plugins/uefi-capsule/fu-uefi-backend-freebsd.c
SergiiDmytruk: `entry->fw_class` isn't an array of bytes, it's of type `uuid`:
```c
struct uuid {
        uint32_t        time_low;
        uint16_t        time_mid; 
        uint16_t        time_hi_and_version;
        uint8_t         clock_seq_hi_and_reserved;
        uint8_t         clock_seq_low;
        uint8_t         node[_UUID_NODE_LEN]; 
};
```

libfwupdplugin/fu-efivar-freebsd.c
hughsie: Ack, gotcha, thanks.

plugins/uefi-capsule/fu-uefi-backend-freebsd.c
hughsie: Related question; where is efi_esrt_entry_v1 defined?

meson.build
```diff
@@ -281,7 +281,7 @@ if get_option('default_library') != 'static'
     platform_deps += cc.find_library('shlwapi')
   endif
   if host_machine.system() == 'freebsd'
-    platform_deps += cc.find_library('efivar')
+    platform_deps += dependency('efivar')
   endif
```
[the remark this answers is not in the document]
SergiiDmytruk: This is the top-level `meson.build`. `cc.find_library` found base-system library (in `/usr/lib`), which isn't complete, instead of discovering a port of `libefivar` (in `/usr/local/lib`).

plugins/uefi-capsule/fu-uefi-backend-freebsd.c
SergiiDmytruk: [Here](https://github.com/3mdeb/freebsd-src/pull/2/files#diff-572d772ebf30e43ba6125d73894f6312242f94a7d1627c06fc3ee4601947ee1eR137), it's a new structure.

plugins/uefi-capsule/fu-uefi-backend-freebsd.c
SergiiDmytruk: Right. What's weird, I don't remember touching this function.

meson.build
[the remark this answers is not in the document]
superm1: So just fix the top level one to find the port?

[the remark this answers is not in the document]
SergiiDmytruk: This **is** the top-level one :)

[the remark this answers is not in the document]
superm1: > This **is** the top-level one :)

🤦sorry about that!

superm1 (dismissed):

hughsie:
Is the CI failure transient? Should I retry it?

SergiiDmytruk:
CI doesn't have required declarations in headers, so it should fail. We could replace EFI-related system headers on CI to make build pass.

hughsie:
> CI doesn't have required declarations in headers, so it should fail

We can't just leave CI failing. Can we detect the declarations in meson and #ifdef around it's use please.

hughsie:
@SergiiDmytruk I added one commit and fixed up a few of yours, I hope that was okay. It's not failing with:

    ld: error: undefined symbol: __builtin_va_arg_pack

A toolchain issue perhaps?

SergiiDmytruk:
> A toolchain issue perhaps?

Yes, looks like it's build with Clang instead of GCC. `efivar` depends on GCC-specific `__builtin_va_arg_pack` builtin function.

SergiiDmytruk:
@hughsie, you once asked how much of `libefiboot` is being used and I later found list of functions in my notes:
```
efi_loadopt_is_valid
efi_loadopt_desc
efi_loadopt_attr_set
efi_loadopt_create
efi_generate_file_device_path
```

hughsie:
@vathpela I do wonder if it would make sense to copy these into fwupd -- any comments?

hughsie (approved):

PR #3344, opened 2021-06-11
flashrom: probe PS175 version via DPCD, not direct flash access

tari:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

---

The LSPCON mostly appears to work normally while flashing, but some functions
such as hotplug detection appear to stop working. Avoid opening the flashrom
programmer for the device unless actually flashing it, and change to reading
the device firmware version from DPCD to achieve that. A new quirk is used
to determine the device over which to access DPCD, keyed by system HWID family.

cc @djcampello

### DIFF
--- a/plugins/flashrom/README.md
+++ b/plugins/flashrom/README.md
@@ -72,6 +72,7 @@ This plugin uses the following plugin-specific quirks:
 | Quirk                  | Description                                 | Minimum fwupd version |
 |------------------------|---------------------------------------------|-----------------------|
 |`FlashromProgrammer`    | Used to specify the libflashrom programmer to be used.     | 1.5.9                 |
+|`ParadeLspconAuxDeviceName` | sysfs name of the drm_dp_aux_dev device over which to read LSPCON version | 1.6.1 |
 
 
 External interface access
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -43,3 +43,6 @@ Plugin=flashrom
 FlashromProgrammer=lspcon_i2c_spi
 Name=PS175
 Vendor=Parade
+
+[FLASHROM-LSPCON-I2C-SPI\HwidFamily=Google_Hatch]
+ParadeLspconAuxDeviceName=DPDDC-B
--- a/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
+++ b/plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
@@ -15,6 +15,7 @@
 #include <libflashrom.h>
 #include <linux/i2c-dev.h>
 #include <sys/ioctl.h>
+#include <libfwupdplugin/fu-hwids.h>
 
 #define I2C_PATH_REGEX "/i2c-([0-9]+)/"
 #define HID_LENGTH 8
@@ -23,6 +24,7 @@ struct _FuFlashromLspconI2cSpiDevice {
 	FuFlashromDevice	  parent_instance;
 	gint			  bus_number;
 	guint8			  active_partition;
+	gchar			 *aux_device_name;
 };
 
 G_DEFINE_TYPE (FuFlashromLspconI2cSpiDevice, fu_flashrom_lspcon_i2c_spi_device,
@@ -77,8 +79,18 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(FlashromLayout, dispose_flash_layout);
 static void
 fu_flashrom_lspcon_i2c_spi_device_init (FuFlashromLspconI2cSpiDevice *self)
 {
+	self->aux_device_name = NULL;
+
 	fu_device_add_flag (FU_DEVICE (self), FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_set_version_format (FU_DEVICE (self), FWUPD_VERSION_FORMAT_PAIR);
+}
+
+static void
+fu_flashrom_lspcon_i2c_spi_device_finalize (GObject *object)
+{
+	FuFlashromLspconI2cSpiDevice *self = FU_FLASHROM_LSPCON_I2C_SPI_DEVICE (object);
+
+	g_free (self->aux_device_name);
 }
 
 static gboolean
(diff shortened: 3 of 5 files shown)


plugins/flashrom/README.md
```diff
@@ -72,6 +72,7 @@ This plugin uses the following plugin-specific quirks:
 | Quirk                  | Description                                 | Minimum fwupd version |
 |------------------------|---------------------------------------------|-----------------------|
 |`FlashromProgrammer`    | Used to specify the libflashrom programmer to be used.     | 1.5.9                 |
+|`FlashromLspconAuxDeviceName` | sysfs name of the drm_dp_aux_dev device over which to read LSPCON version | 1.6.1 |
```
hughsie, line 75: This doesn't feel very flashrom, no?

plugins/flashrom/flashrom.quirk
```diff
@@ -43,3 +43,6 @@ Plugin=flashrom
 FlashromProgrammer=lspcon_i2c_spi
 Name=PS175
 Vendor=Parade
+
+[FLASHROM-LSPCON-I2C-SPI\HwidFamily=Google_Hatch]
+FlashromLspconAuxDeviceName=DPDDC-B
```
hughsie, line 48: needs newline at the end

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -246,9 +305,31 @@ fu_flashrom_lspcon_i2c_spi_device_setup (FuDevice *device, GError **error)
...
+
 	return fu_flashrom_lspcon_i2c_spi_device_set_version (device, error);
 }
 
+static gboolean
+fu_flashrom_lspcon_i2c_spi_device_detach (FuDevice *device, GError **error)
+{
+	return FU_DEVICE_CLASS (fu_flashrom_lspcon_i2c_spi_device_parent_class)
+		->open (device, error);
+}
```
hughsie, line 324: isn't this a bit weird? open for detach? I'm not sure how much stuff is going to explode, as sometimes we call ->open(), ->detach(), ->close() etc.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -328,6 +409,9 @@ fu_flashrom_lspcon_i2c_spi_device_write_firmware (FuDevice *device,
 		return FALSE;
 	}
 
+	/* device (re)loads its firmware at power-on */
+	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_NEEDS_SHUTDOWN);
```
superm1, line 413: Does warm reboot help it too?  Or it really needs _shutdown_?

plugins/flashrom/flashrom.quirk
superm1, line 48: Any chance at publishing this somewhere more firmware-y rather than having to hardcode in `.quirk` eventually?  It feels like something that could go in ACPI or DT.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -175,56 +210,80 @@ probe_active_flash_partition (FuFlashromLspconI2cSpiDevice *self,
...
 
-	/* extract the active partition's version number */
-	addr = layout->entries[self->active_partition + 2].start;
-	g_return_val_if_fail (addr < (flash_size - 2), FALSE);
-	version = g_strdup_printf ("%d.%d", contents[addr], contents[addr + 2]);
+	/* DPCD address 0x50A, 0x50B: branch device firmware
+	 * major and minor revision */
+	if ((aux_device_locker = fu_device_locker_new (aux_device, error)) == NULL)
+		return FALSE;
+	if (!fu_udev_device_pread_full (aux_device, 0x50a, version_buf,
```
superm1, line 268: is there a check you can do to make sure you're really talking to the LSPCON before you go and believe this offset means version?  Some sort of ID byte or anything?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
```diff
@@ -175,56 +210,80 @@ probe_active_flash_partition (FuFlashromLspconI2cSpiDevice *self,
...
+		return FALSE;
+	version = g_strdup_printf ("%d.%d", version_buf[0], version_buf[1]);
 	fu_device_set_version (device, version);
 
 	return TRUE;
+#else
+	g_set_error_literal (error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED,
+			     "udev support is required to probe firmware version");
+	return FALSE;
+#endif /* HAVE_GUDEV */
```
superm1, line 279: since this is adding a harder suggestion on `gudev`, I would suggest perhaps to add a check if `gudev` was found into `plugins/flashrom/meson.build` and emit a `warning()` if it's turned off.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: It really needs shutdown, determined by experiment.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: These bytes are defined to be version bytes by the DisplayPort specification, but I could also check the OUI at 0x500.

plugins/flashrom/flashrom.quirk
tari: It's a little weird, because ACPI already specifies which I2C bus the LSPCON is connected to but we need to use the DisplayPort aux channel to access DPCD- the LSPCON appears in two places, one of which is a control/debug interface and the other being mostly transparent.

That aside, updating firmware would be non-trivial since there are devices in the wild and no current plan to issue any updates (which would require re-validating).

plugins/flashrom/README.md
tari: Changed to `ParadeLspcon`. I had chosen to use the plugin name prefix, but clearly specifying the device type seems fine too.

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
tari: I agree it's weird, but the alternative is to move where calls to `flashrom_programmer_init` and friends live. Given `fu_flashrom_device_open` doesn't chain up and the `open` method on the LSPCON device is changed here to avoid chaining up to `FlashromDevice`, this seems robust.

plugins/flashrom/flashrom.quirk
superm1: But would it be possible to add another field to "new" systems that could pick this up automatically?

plugins/flashrom/fu-flashrom-lspcon-i2c-spi-device.c
superm1: OK.  I do think checking the OUI to know you're talking to the right lspcon is a good general safety thing then at least so when it comes time to rev the plugin for another parade part someone makes a conscious check on all the bits from this part to another.

tari: Added OUI check.

plugins/flashrom/flashrom.quirk
tari: It probably could be, but I'm not sure exactly how that might look. Hanging something off the graphics pipe doesn't really work because that gets probed via PCI, but it might be possible to encode something like this quirk in a spare attribute on the I2C device.

hughsie:
@tari we can close this one now?

tari:
I'd still like to get this merged, since it fixes an actual bug with the existing plugin and what it does remains relevant if the device stops using flashrom.

tari:
Closing this in favor of the same feature in #3476.

PR #3348, opened 2021-06-14
trivial: modem-manager: don't convert instance IDs to GUIDs in probe

superm1:
This already happens during setup, there shouldn't be a need to depend
on the private header to accomplish this.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -11,7 +11,6 @@
 #include "fu-io-channel.h"
 #include "fu-archive.h"
 #include "fu-mm-device.h"
-#include "fu-device-private.h"
 #include "fu-mm-utils.h"
 #include "fu-qmi-pdc-updater.h"
 
@@ -325,9 +324,6 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 			}
 		}
 	}
-
-	/* convert the instance IDs to GUIDs */
-	fu_device_convert_instance_ids (device);
 
 	return TRUE;
 }


aleksander0m (commented):
In a block a bit earlier we add all the `device_ids` retrieved from `mm_firmware_update_settings_get_device_ids()` into the device, with `fu_device_add_instance_id()`.

The instance ids added are strings in the following form:
 * `USB\VID_413C&PID_81D7&REV_0318&CARRIER_VODAFONE`
 * `USB\VID_413C&PID_81D7&REV_0318`
 * `USB\VID_413C&PID_81D7`
 * `USB\VID_413C`

When you say "This already happens during setup", do you mean the operation will be performed *after* these `fu_device_add_instance_id()` calls? Or that it happened *before*?

If it happened before, I guess we still need to have that call after the custom `fu_device_add_instance_id()` calls right?

If it happens after, I guess it may not be needed as you suggest.

hughsie:
> If it happens after, I guess it may not be needed as you suggest.

`fu_device_convert_instance_ids()` gets automatically called by the `->setup()` vfunc. The question @superm1 is really asking, is "does the ModemManager really need the instance IDs converted to GUIDs *before* the device setup gets called. Seeing as there's no `->setup()` being used by FuMmDevice I suspect the answer is no, and we can remove that code.

@aleksander0m can you tell me easily if this branch breaks enumeration or update for any of the devices you handy? I can test too, but I'm in the wrong office today.

aleksander0m:
> @aleksander0m can you tell me easily if this branch breaks enumeration or update for any of the devices you handy? I can test too, but I'm in the wrong office today.

I have in my TODO list the test with the SDX55 module, I hope to do that this week if I have some free time (I said this last week also though...). While doing that, I can test this change as well. If you're in a hurry to merge this, please do so as your reasoning seems to be the correct one :)

superm1:
> > @aleksander0m can you tell me easily if this branch breaks enumeration or update for any of the devices you handy? I can test too, but I'm in the wrong office today.
> 
> I have in my TODO list the test with the SDX55 module, I hope to do that this week if I have some free time (I said this last week also though...). While doing that, I can test this change as well. If you're in a hurry to merge this, please do so as your reasoning seems to be the correct one :)

We're not in a big hurry to merge it, so checking with hardware when you can is fine.  As Richard elaborated when he read my thought process, I think it should work fine, but it's better to check it.

aleksander0m:
I tested this patch, and unfortunately it didn't work as expected. The WWAN device is no longer listed as available, and the daemon log spits:
```
12:01:04:0928 FuPlugin             coldplug(modem_manager)
12:01:04:0931 FuPluginMm           ModemManager 1.17.1 is available
12:01:04:0932 FuPluginMm           valid firmware update combination: mbim-qdu
12:01:04:0932 FuDevice             using bba54f76de66958e5d70964b6ddd2d1e6c7dae9e for /sys/devices/pci0000:00/0000:00:1c.0/0000:08:00.0
12:01:04:0932 FuPlugin             emit added from modem_manager: bba54f76de66958e5d70964b6ddd2d1e6c7dae9e
12:01:04:0932 FuEngine             no GUIDs for device Qualcomm Snapdragon X55 5G [bba54f76de66958e5d70964b6ddd2d1e6c7dae9e]
```

hughsie:
@aleksander0m can you try https://github.com/fwupd/fwupd/pull/3363 instead please.

aleksander0m:
> @aleksander0m can you try #3363 instead please.

That one works

PR #3363, opened 2021-06-15
modem-manager: Do not convert instance IDs to GUIDs manually

hughsie:
This already happens during setup, there shouldn't be a need to depend
on the private header to accomplish this.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


aleksander0m (approved):
These changes didn't break anything

PR #3373, opened 2021-06-16
modem-manager: implement Quectel EM120/160 firmware upgrade support

ivanmikh:
Introduces support for Quectel EM120/160 PCIe modems as well as Firehose
update method to modem-manager plugin.

Related changes have been recently merged into ModemManager master (https://gitlab.freedesktop.org/mobile-broadband/ModemManager/-/merge_requests/553).

The code is mainly based on @aleksander0m implementation of the Firehose update method that he did last year for Quectel (https://github.com/aleksander0m/fwupd/tree/aleksander/next-20201203). It was adapted to work with the new upstream mhi_wwan driver.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -13,6 +13,10 @@
 #include "fu-mm-utils.h"
 #include "fu-qmi-pdc-updater.h"
 
+#if MM_CHECK_VERSION(1, 17, 2)
+#include "fu-firehose-updater.h"
+#endif
+
 /* Amount of time for the modem to boot in fastboot mode. */
 #define FU_MM_DEVICE_REMOVE_DELAY_RE_ENUMERATE	20000	/* ms */
 
@@ -52,6 +56,13 @@ struct _FuMmDevice {
 	FuQmiPdcUpdater			*qmi_pdc_updater;
 	GArray				*qmi_pdc_active_id;
 	guint				 attach_idle;
+
+	/* firehose update handling */
+	gchar				*port_qcdm;
+	gchar				*port_edl;
+#if MM_CHECK_VERSION(1, 17, 2)
+	FuFirehoseUpdater	*firehose_updater;
+#endif
 };
 
 enum {
@@ -71,6 +82,8 @@ fu_mm_device_to_string (FuDevice *device, guint idt, GString *str)
 		fu_common_string_append_kv (str, idt, "AtPort", self->port_at);
 	if (self->port_qmi != NULL)
 		fu_common_string_append_kv (str, idt, "QmiPort", self->port_qmi);
+	if (self->port_qcdm != NULL)
+		fu_common_string_append_kv (str, idt, "QcdmPort", self->port_qcdm);
 }
 
 const gchar *
@@ -114,6 +127,9 @@ validate_firmware_update_method (MMModemFirmwareUpdateMethod methods, GError **e
 	static const MMModemFirmwareUpdateMethod supported_combinations[] = {
 		MM_MODEM_FIRMWARE_UPDATE_METHOD_FASTBOOT,
 		MM_MODEM_FIRMWARE_UPDATE_METHOD_QMI_PDC | MM_MODEM_FIRMWARE_UPDATE_METHOD_FASTBOOT,
+#if MM_CHECK_VERSION (1, 17, 2)
+		MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE,
+#endif
 	};
 	g_autofree gchar *methods_str = NULL;
 
@@ -227,6 +243,18 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 		if (fu_device_get_protocols (device)->len == 0)
 			fu_device_add_protocol (device, "com.qualcomm.qmi_pdc");
 	}
+#if MM_CHECK_VERSION (1, 17, 2)
+	if (self->update_methods & MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE) {
+		for (guint i = 0; i < n_ports; i++) {
+			if (ports[i].type == MM_MODEM_PORT_TYPE_QCDM) {
+				self->port_qcdm = g_strdup_printf ("/dev/%s", ports[i].name);
+				break;
+			}
+		}
+		fu_device_add_protocol (device, "com.qualcomm.firehose");
+	}
+#endif
+
 	mm_modem_port_info_array_free (ports, n_ports);
 
 	/* an at port is required for fastboot */
@@ -248,6 +276,18 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 				     "failed to find QMI port");
 		return FALSE;
 	}
+
+#if MM_CHECK_VERSION (1, 17, 2)
+	/* a qcdm port is required for firehose */
+	if ((self->update_methods & MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE) &&
+		(self->port_qcdm == NULL)) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOT_SUPPORTED,
+				     "failed to find QCDM port");
+		return FALSE;
+	}
+#endif
 
 	if (self->port_at != NULL) {
 		fu_mm_utils_get_port_info (self->port_at, &device_bus, &device_sysfs_path, &self->port_at_ifnum, NULL);
@@ -276,6 +316,28 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 		}
 	}
 
+	if (self->port_qcdm != NULL) {
+		g_autofree gchar *qcdm_device_sysfs_path = NULL;
+		g_autofree gchar *qcdm_device_bus = NULL;
+		fu_mm_utils_get_port_info (self->port_qcdm, &qcdm_device_bus, &qcdm_device_sysfs_path, NULL, NULL);
+		if (device_sysfs_path == NULL && qcdm_device_sysfs_path != NULL) {
+			device_sysfs_path = g_steal_pointer (&qcdm_device_sysfs_path);
+		} else if (g_strcmp0 (device_sysfs_path, qcdm_device_sysfs_path) != 0) {
+			g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+				     "mismatched device sysfs path: %s != %s",
+				     device_sysfs_path, qcdm_device_sysfs_path);
+			return FALSE;
+		}
+		if (device_bus == NULL && qcdm_device_bus != NULL) {
+			device_bus = g_steal_pointer (&qcdm_device_bus);
+		} else if (g_strcmp0 (device_bus, qcdm_device_bus) != 0) {
+			g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+				     "mismatched device bus: %s != %s",
+				     device_bus, qcdm_device_bus);
+			return FALSE;
+		}
+	}
+
 	/* if no device sysfs file, error out */
 	if (device_sysfs_path == NULL || device_bus == NULL) {
 		g_set_error_literal (error,
@@ -294,7 +356,7 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
 	fu_device_set_version (device, version);
 	for (guint i = 0; device_ids[i] != NULL; i++)
 		fu_device_add_instance_id (device, device_ids[i]);
-	if (fu_device_get_vendor_ids (device) == NULL) {
+	if (fu_device_get_vendor_ids (device) == NULL || fu_device_get_vendor_ids (device)->len == 0) {
 		g_autofree gchar *path = NULL;
 		g_autofree gchar *value_str = NULL;
 		g_autoptr(GError) error_local = NULL;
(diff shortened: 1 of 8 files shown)


aleksander0m:
It's in my TODO list now, thanks!

P.S.: I have a EM120 that I can still use for testing this, in addition to reviewing.

hughsie:
I see the Fedora CI failing. Do we need some meson checks for MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE perhaps?

aleksander0m:
> I see the Fedora CI failing. Do we need some meson checks for MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE perhaps?

That enum value was recently added to MM git master; I'll bump the development version so that we can use `MM_CHECK_VERSION()` in order to match that enum without needing to fully bump the MM min version in fwupd

aleksander0m:
> > I see the Fedora CI failing. Do we need some meson checks for MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE perhaps?
> 
> That enum value was recently added to MM git master; I'll bump the development version so that we can use `MM_CHECK_VERSION()` in order to match that enum without needing to fully bump the MM min version in fwupd

Ok, done. The code matching the FIREHOSE method can now use `#if MM_CHECK_VERSION (1, 17, 2)...`

ivanmikh:
> It's in my TODO list now, thanks!
> 
> P.S.: I have a EM120 that I can still use for testing this, in addition to reviewing.

Just one note: to test firmware upgrade, you need to put prog_firehose_sdx24.mbn into your /lib/firmware/qcom folder, because with the new mhi_wwan driver flashing of firehose binary is done by the kernel (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/bus/mhi/pci_generic.c#n299). This binary usually does not change, so if you still have firmware packages for EM120, you can take from there. Hopefully, we will add it to linux-firmware later.

ivanmikh:
> > > I see the Fedora CI failing. Do we need some meson checks for MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE perhaps?
> > 
> > 
> > That enum value was recently added to MM git master; I'll bump the development version so that we can use `MM_CHECK_VERSION()` in order to match that enum without needing to fully bump the MM min version in fwupd
> 
> Ok, done. The code matching the FIREHOSE method can now use `#if MM_CHECK_VERSION (1, 17, 2)...`

Added guard macros for FIREHOSE code. Should compile now

plugins/modem-manager/fu-mm-device.c
```diff
@@ -227,6 +238,19 @@ fu_mm_device_probe_default (FuDevice *device, GError **error)
...
+#if MM_CHECK_VERSION (1, 17, 2)
+	if (self->update_methods & MM_MODEM_FIRMWARE_UPDATE_METHOD_FIREHOSE) {
+		for (guint i = 0; i < n_ports; i++) {
+			if (ports[i].type == MM_MODEM_PORT_TYPE_QCDM) {
+				self->port_qcdm = g_strdup_printf ("/dev/%s", ports[i].name);
+				break;
+			}
+		}
+		fu_device_add_protocol (device, "com.qualcomm.firehose");
+		g_warning("firehose modem: qcdm port = %s", self->port_qcdm);
```
superm1, line 250: warning?

ivanmikh: oops, overlooked that. will remove

aleksander0m:
> > It's in my TODO list now, thanks!
> > P.S.: I have a EM120 that I can still use for testing this, in addition to reviewing.
> 
> Just one note: to test firmware upgrade, you need to put prog_firehose_sdx24.mbn into your /lib/firmware/qcom folder, because with the new mhi_wwan driver flashing of firehose binary is done by the kernel (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/bus/mhi/pci_generic.c#n299). This binary usually does not change, so if you still have firmware packages for EM120, you can take it from there. Hopefully, we will add it to the linux-firmware in the future.

Didn't review yet, but we have a problem already here. Do you mean the `prog_firehose_sdx24.mbn` is not being provided by the CAB file and installed by fwupd, and instead that file is expected to exist in _/lib/firmware/qcom_ instead? And if it doesn't, the upgrade will fail?

I'm afraid that should be a prerequisite to have this merge request merged. _Hopefully, we will add it to the linux-firmware in the future_ should not be a solution for the problem; having that file integrated must be a prerequisite for this merge request to go on.

Have you attempted to integrate the file in linux-firmware already? Is there a request for that online already? If not, please do so right away, and we'll work on this merge request after that.

hughsie:
> And if it doesn't, the upgrade will fail

We should certainly check for that file existing, and if not use fu_device_inhibit().

> having that file integrated must be a prerequisite for this merge request to go on

Well, I think we can continue to review it, but I agree we should wait to _merge_ this PR until it's upstream.

aleksander0m:
> > And if it doesn't, the upgrade will fail
> 
> We should certainly check for that file existing, and if not use fu_device_inhibit().
> 

And that would make the device not show in the list of devices, right?

> > having that file integrated must be a prerequisite for this merge request to go on
> 
> Well, I think we can continue to review it, but I agree we should wait to _merge_ this PR until it's upstream.

Oh yes, that's what I meant here. I still plan to review it, especially since I wrote it originally :) But for merging, I would definitely suggest to wait until the linux-firmware bits are resolved, because that is the only way not to lose traction on these changes (i.e. we want the full solution upstream, not just bits and pieces that then the manufacturer/distributor can fix by themselves only for their own setups). We had a great success already with the new wwan subsystem in the kernel (instead of per-manufacturer custom mhi_uci drivers), and I would like to believe that that was in part because we were very stubborn not to allow merging support for those non-upstreamed drivers.

To be fully clear, I wouldn't want to use this logic in ModemManager and fwupd exclusively in a Lenovo-sponsored "custom" Ubuntu-based installation for which we don't have sources anywhere, I would like to be able to use it in my Arch Linux, and I would like everyone to be able to use it in other distributions like Fedora or even a "real" Ubuntu.

hughsie:
> And that would make the device not show in the list of devices, right?

It would show if you added `--show-all` but not by default.

> I would like to believe that that was in part because we were very stubborn

Agreed. Stick to the plan, you know what you're doing; I trust your judgement here.

ivanmikh:
> Didn't review yet, but we have a problem already here. Do you mean the `prog_firehose_sdx24.mbn` is not being provided by the CAB file and installed by fwupd, and instead that file is expected to exist in `/lib/firmware/qcom` instead? And if it doesn't, the upgrade will fail?

Actually, we do include `prog_firehose_sdx24.mbn` in CAB file. I was thinking, would it be a good solution for fwupd to copy this binary file to the `/lib/firmware/qcom` during .cab file decompression?

The problem with linux-firmware is that `prog_firehose_sdx24.mbn` is a Qualcomm binary that needs to be uploaded to qcom directory. Qualcomm is not planning to do that, and I am not sure if we are allowed to do that or is it going to be accepted.

mrhpearson:
Just jumping in from a Lenovo perspective, and apologies in advance if I say anything stupid.

Lenovo want the firmware update to work for this modem on any distro - "Ubuntu only" is definitely not an intent here (we've been testing this modem on Fedora too). I'll support solutions that makes this available for any distro.
Note - WWAN has been a really tough nut to crack and Quectel have been fantastic to work with and do things in the right way upstream - I want to support that and get this out for their modems and our users and fully functional to the same standards Windows users have.

I'll will raise this internally to see if we can help get Qualcomm to add the firmware to linux-firmware as that sounds like the best solution. Those discussions always take a long time and I'd love fwupd support to be available for these modems and Linux users in the meantime. The alternative is Windows, a non-LVFS updater or no updates - all of which are far worse. Not trying to short-cut correct solutions here - just working with the realities of the ecosystem.

I'm open to suggestions of how to solve this in the short term (or if worst-case Qualcomm won't upstream their firmware). I like Ivan's solution of packaging it in the cab if that is acceptable - that seems vendor and distro agnostic to me.

If that isn't possible then let me know alternate suggestions. I have some ideas but they're all Lenovo specific which seems less  ideal (but are distro agnostic)

Thanks
Mark

aleksander0m:
> Just jumping in from a Lenovo perspective, and apologies in advance if I say anything stupid.
> 

Thanks for your comments; truly appreciated!

> Lenovo want the firmware update to work for this modem on any distro - "Ubuntu only" is definitely not an intent here (we've been testing this modem on Fedora too). I'll support solutions that makes this available for any distro.

It's great to hear that :+1: 

> Note - WWAN has been a really tough nut to crack and Quectel have been fantastic to work with and do things in the right way upstream - I want to support that and get this out for their modems and our users and fully functional to the same standards Windows users have.
> 

I was hired by Quectel to write the original version of this update method, at that time the integration was done with a non-upstreamed kernel driver, and the way the prog_firehose was passed down to the module was different (using BHI instead of /lib/firmware). I can also say that Quectel engineers have been extremely professional in supporting all this, and they still are, I have 0 complaints :)

> I'll will raise this internally to see if we can help get Qualcomm to add the firmware to linux-firmware as that sounds like the best solution. Those discussions always take a long time and I'd love fwupd support to be available for these modems and Linux users in the meantime. The alternative is Windows, a non-LVFS updater or no updates - all of which are far worse. Not trying to short-cut correct solutions here - just working with the realities of the ecosystem.
> 
> I'm open to suggestions of how to solve this in the short term (or if worst-case Qualcomm won't upstream their firmware). I like Ivan's solution of packaging it in the cab if that is acceptable - that seems vendor and distro agnostic to me.
> 

Yes. If that is an option, I would prefer that as well. I raised this issue to @hughsie a while ago when I got notified how this was going to be used instead of using BHI, and I'm not sure to which extent this was discussed after that. I have no idea if temporarily placing a file in /lib/firmware/qcom by fwupd makes sense or not; maybe it does make sense and we just need to be the first ones to do it. It is true that the prog_firehose file is not a "firmware file" as linux-firmware is used to contain; it's just a temporary utility that we need to flash in the module in order to trigger a more robust and complete firmware update method (firehose). Being such a strange "firmware file" it may make sense to just temporarily copy it there during the upgrade, and remove it once we're done (or just once we've downloaded it to the module). I think we need @hughsie and @superm1 to weight in here.

> If that isn't possible then let me know alternate suggestions. I have some ideas but they're all Lenovo specific which seems less ideal (but are distro agnostic)

I sent you an email with some further discussion and requests in this regard; hope you don't mind.

superm1:
`/lib/firmware` isn't guaranteed to be writable.  If it's got to go to that directory, would this work instead perhaps? https://github.com/fwupd/fwupd/pull/3397

This can let you reconfigure the directory temporarily in a plugin, do something, and then configure it back.

aleksander0m:
> `/lib/firmware` isn't guaranteed to be writable. If it's got to go to that directory, would this work instead perhaps? #3397
> 
> This can let you reconfigure the directory temporarily in a plugin, do something, and then configure it back.

This is definitely something to investigate!

ivanmikh:
> `/lib/firmware` isn't guaranteed to be writable. If it's got to go to that directory, would this work instead perhaps? #3397
> 
> This can let you reconfigure the directory temporarily in a plugin, do something, and then configure it back.

This looks like a really good alternative! Is it something we can consider adding to the fwupd?

Another option could be to actually change the path in the mhi_pci_generic module itself. As @aleksander0m mentioned, it is not a real firmware, so if we can propose a better location for this firehose binary, there are good chances that mhi-next maintainer will accept it. (https://git.kernel.org/pub/scm/linux/kernel/git/mani/mhi.git/tree/drivers/bus/mhi/pci_generic.c?h=mhi-next#n301)

superm1:
> This looks like a really good alternative! Is it something we can consider adding to the fwupd?

I think if we can see it working, I don't see why not.  Go ahead and experiment :)

>so if we can propose a better location for this firehose binary, there are good chances that mhi-next maintainer will accept it

How big is the binary?   If it's small enough you can do a sysfs loader.  If it's too big, you can perhaps do a nvmem loader?

Those are way cleaner options from a kernel perspective compared to loading up from `/lib/firmware`  in my mind.

ivanmikh:
> I think if we can see it working, I don't see why not. Go ahead and experiment :)

Thanks, I will try it on my Ubuntu laptop. I don't have any other setup, but hope it should be enough for initial testing

> How big is the binary? If it's small enough you can do a sysfs loader. If it's too big, you can perhaps do a nvmem loader?

it's 272 kb currently. Is it small enough for a sysfs loader?

hughsie:
> I have no idea if temporarily placing a file in /lib/firmware/qcom by fwupd makes sense or not

I think it's weird; but it's not totally crazy. I think modifying the loaderdir to point at something fwupd-controlled (and possibly temporary) is probably a better idea -- and then change it back to whatever it was before. Although we have to make sure we keep things secure and probably only do the loader thing for signed firmware or something.

> it's 272 kb currently. Is it small enough for a sysfs loader?

I think that's fine.

superm1:
If it's more than PAGE_SIZE I'm not sure you'll be able to do it with sysfs. The issue is the kernel code has to do a copy from userspace, and I don't think it's able to do that much data in one swoop.

Nvmem probably makes more sense for that size of data.

ivanmikh:
so modifying `/sys/module/firmware_class/parameters/path` did not work for me. it seems that the problem is that `FW_LOADER` is builtin on most systems, and to change firmware_class pass, kernel boot parameters need to be modified.

hughsie:
> kernel boot parameters need to be modified

Ew, so we don't want to do that. Maybe we can do both? e.g. try to change /sys/module/firmware_class/parameters/path and if that fails (to write?, to exist?) then try writing the firmware into the distro-set location.

ivanmikh:
`/sys/module/firmware_class/parameters/path` exists in the system and writing to it does not yield any errors. its just not used by the module, if its builtin.

i guess the only option left is copying to /lib/firmware directly. I will try to implement it and test.

by the way, is there any plugin vfunc that works with firmware bytes before detaching a device?

hughsie:
> by the way, is there any plugin vfunc

->prepare_firmware() is what you want I think.

ivanmikh:
will be merged in #3422

PR #3374, opened 2021-06-16
Allow probing Synaptics MST w/ amdgpu on newer enough kernels

superm1:
I've found that the original problem we had has been fixed in current kernels.  It appears to be fixed by https://github.com/torvalds/linux/commit/8ae5b1d78d4acbe9755570f26703962877f9108a

Running with an AMD Renoir system I can find that it probes properly as far back as kernel 5.8 (but can't check much earlier due to lack of kernel support for Renoir).
```
~/fwupd$ sudo fwupdtool get-devices --plugins synaptics-mst
20Y10015US
│
└─VMM5322:
      Device ID:          81905eb4059e79f42c9f19fdcb195beeae5f38fa
      Summary:            Multi-Stream Transport Device
      Current version:    5.05.05
      Vendor:             Synaptics (DRM_DP_AUX_DEV:0x06CB)
      GUIDs:              e6b5bc25-d512-51b0-b23c-25fdf483b15d ← MST-595
                          f15aa55c-9cd5-5942-85ae-a6bf8740b96c ← MST-panamera
                          e0b67325-e8ff-5171-93e9-aaf098589166 ← MST-panamera-vmm5322-595
                          50f72ec0-b147-56d7-b1a4-6af9256b2d6a ← MST-panamera-595
      Device Flags:       • Updatable
                          • Device stages updates
```

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -319,6 +319,7 @@ done
 %config(noreplace)%{_sysconfdir}/fwupd/uefi_capsule.conf
 %endif
 %config(noreplace)%{_sysconfdir}/fwupd/redfish.conf
+%config(noreplace)%{_sysconfdir}/fwupd/synaptics_mst.conf
 %config(noreplace)%{_sysconfdir}/fwupd/thunderbolt.conf
 %dir %{_libexecdir}/fwupd
 %{_libexecdir}/fwupd/fwupd
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -21,6 +21,10 @@
 #include <cpuid.h>
 #endif
 
+#ifdef HAVE_UTSNAME_H
+#include <sys/utsname.h>
+#endif
+
 #ifdef HAVE_LIBARCHIVE
 #include <archive_entry.h>
 #include <archive.h>
@@ -34,6 +38,7 @@
 #include "fwupd-error.h"
 
 #include "fu-common-private.h"
+#include "fu-common-version.h"
 #include "fu-firmware.h"
 #include "fu-volume-private.h"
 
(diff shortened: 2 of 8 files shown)


superm1:
CC @mrhpearson as an FYI, as this will allow TBT3 Gen2 dock to update MST controller from AMD systems.

libfwupdplugin/fu-common-version.c
```diff
@@ -545,3 +549,41 @@ fu_common_vercmp_full (const gchar *version_a,
...
+ * fu_common_version_kernel:
+ * @minimum_kernel: The minimum kernel version to check against
+ * @error: (nullable): optional return location for an error
+ *
+ * Determines if the system is running at least a certain required kernel version
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_common_version_kernel (const gchar *minimum_kernel, GError **error)
```
hughsie, line 563: I feel like this should have `check` or `verify` in the name somewhere. Maybe `fu_common_check_kernel_version` is more obvious?

superm1: Sure, I'll shift that.  Should it live in this file or `fu-common.c` with such a change?

libfwupdplugin/fu-common-version.c
```diff
@@ -545,3 +549,41 @@ fu_common_vercmp_full (const gchar *version_a,
 	}
 	return fu_common_vercmp_safe (version_a, version_b);
 }
+
+/**
+ * fu_common_version_kernel:
+ * @minimum_kernel: The minimum kernel version to check against
```
hughsie, line 555: `(not nullable)`

libfwupdplugin/fu-common-version.c
```diff
@@ -545,3 +549,41 @@ fu_common_vercmp_full (const gchar *version_a,
...
+ * Determines if the system is running at least a certain required kernel version
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_common_version_kernel (const gchar *minimum_kernel, GError **error)
+{
+#ifdef HAVE_UTSNAME_H
+	struct utsname name_tmp;
+
```
hughsie, line 567: needs a return_val_if_fail for minimum_kernel==NULL

libfwupdplugin/fu-common-version.c
```diff
@@ -545,3 +549,41 @@ fu_common_vercmp_full (const gchar *version_a,
...
+	memset (&name_tmp, 0, sizeof(struct utsname));
+	if (uname (&name_tmp) < 0) {
+		g_debug ("Failed to read current kernel version");
+		return TRUE;
```
hughsie, line 571: TRUE? Wouldn't False be more appropriate here if we're checking for something that requires a specific kernel version?

libfwupdplugin/fu-common-version.c
```diff
@@ -545,3 +549,41 @@ fu_common_vercmp_full (const gchar *version_a,
...
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "kernel %s doesn't meet minimum %s",
+			     name_tmp.release, minimum_kernel);
+		return FALSE;
+	}
+#else
+	g_debug ("platform doesn't support checking for minimum Linux kernel %s",
+		 minimum_kernel);
```
hughsie, line 585: I think this should be g_set_error and FALSE personally....

libfwupdplugin/fu-common-version.c
hughsie: Ahh, i missed that.  fu-common.c certainly, thanks.

libfwupdplugin/fu-common-version.c
superm1: yeah I'll adjust.  I'm not sure why we had it like that previously in thunderbolt.

hughsie (approved):

PR #3378, opened 2021-06-17
libfwupd: Show unconverted GUIDs when debugging

hughsie:
Add a suffix to any GUIDs that are currently unconverted InstanceIDs.

This makes plugin development much easier as you can now see why GUID
quirks in the subclassed ->setup() are not matching.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-device.c
+++ b/libfwupd/fwupd-device.c
@@ -2584,6 +2584,10 @@ fwupd_device_to_string (FwupdDevice *self)
 {
 	FwupdDevicePrivate *priv = GET_PRIVATE (self);
 	GString *str;
+	GHashTableIter iter;
+	gpointer key = NULL;
+	gpointer value = NULL;
+	g_autoptr(GHashTable) instance_ids_by_guid = NULL;
 
 	g_return_val_if_fail (FWUPD_IS_DEVICE (self), NULL);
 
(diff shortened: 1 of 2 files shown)


libfwupd/fwupd-device.c
```diff
@@ -2600,32 +2604,34 @@ fwupd_device_to_string (FwupdDevice *self)
...
+	g_hash_table_iter_init (&iter, instance_ids_by_guid);
+	while (g_hash_table_iter_next (&iter, &key, &value)) {
+		const gchar *guid = (const gchar *) key;
+		const gchar *instance_id = (const gchar *) value;
+		g_autofree gchar *tmp = NULL;
+		if (fwupd_device_has_guid (self, guid)) {
+			tmp = g_strdup_printf ("%s <- %s", guid, instance_id);
+		} else {
+			tmp = g_strdup_printf ("%s <- %s (!)", guid, instance_id);
 		}
```
superm1, line 2625: nit; i don't think you need braces here

superm1 (approved):

libfwupd/fwupd-device.c
hughsie: I did wonder about:

		tmp = g_strdup_printf ("%s <- %s%s", guid, instance_id,
				       fwupd_device_has_guid (self, guid) ? "" : " (!)");

hughsie: maybe just a GString would be better here.

superm1: Yeah either a GString or that are better.

hughsie: I reworked it a bit to be more efficient too. Looks better?

superm1 (dismissed):

PR #3381, opened 2021-06-17
Wip/hughsie/fu usb device setup

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-usb-device.c
+++ b/libfwupdplugin/fu-usb-device.c
@@ -160,11 +160,10 @@ fu_usb_device_query_hub (FuUsbDevice *self, GError **error)
 static gboolean
 fu_usb_device_open (FuDevice *device, GError **error)
 {
+#ifdef HAVE_GUSB
 	FuUsbDevice *self = FU_USB_DEVICE (device);
 	FuUsbDevicePrivate *priv = GET_PRIVATE (self);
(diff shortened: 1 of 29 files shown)


superm1 (dismissed):
@CragW you should double check dell-dock with this change.

hughsie:
Ahh, I broke dell dock as it failed the regex:

	klass_device->setup = fu_dell_dock_hid_get_hub_version;

Can I move `fu_dell_dock_hid_get_hub_version` into `fu-dell-dock-hid.c` and rename it `fu_dell_dock_hub_setup`?

superm1:
> Ahh, I broke dell dock as it failed the regex:
> 
> ```
> klass_device->setup = fu_dell_dock_hid_get_hub_version;
> ```
> 
> Can I move `fu_dell_dock_hid_get_hub_version` into `fu-dell-dock-hid.c` and rename it `fu_dell_dock_hub_setup`?

Glad I mentioned it.  I don't think it should move out of `fu-dell-dock-hid.c`.  I think maybe add a new `fu_dell_dock_hub_setup` which calls the existing function.

hughsie:
>  I think maybe add a new fu_dell_dock_hub_setup which calls the existing function.

Good idea. I did that, and also reviewed all the other USB/HID-using plugins.

PR #3382, opened 2021-06-17
trivial: Allow plugins to use fu_device_convert_instance_ids()

hughsie:
This may be required if the plugin has a _custom_ ->setup() vfunc and
needs the quirks added before the actual ->setup() vfunc returns --
for example in the FuVliDeviceClass.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device-private.h
+++ b/libfwupdplugin/fu-device-private.h
@@ -29,7 +29,6 @@ gboolean	 fu_device_ensure_id			(FuDevice	*self,
 							 G_GNUC_WARN_UNUSED_RESULT;
 void		 fu_device_incorporate_from_component	(FuDevice	*device,
 							 XbNode		*component);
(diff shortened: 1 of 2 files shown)


superm1:
> This may be required if the plugin has a custom ->setup() vfunc and
needs the quirks added before the actual ->setup() vfunc returns --
for example in the FuVliDeviceClass.

I'm a little worried about this causing confusion in the future for what setup will do vs not do.  Do we actually want a new "phase" for convert GUIDs?

hughsie:
>  Do we actually want a new "phase" for convert GUIDs?

That's actually way less magic, and probably a good idea. So we'd have `probe(), open(), setup(), <convert instance IDs here>, the-new-thing(), close()`. Names for `the-new-thing` don't immediately come to mind. maybe `->ready()`

superm1:
Yeah I like ready

Issue #3366, opened 2021-06-15
Lenovo TBT3 Gen2 dock probe errors


superm1:
**Describe the bug**
Notice the following emitted from daemon.  This dock is connected to non-TBT3 host (Lenovo P14s AMD).
```
16:33:31:0138 FuPluginVliUsbhub    cannot create MSP430 I?C device: failed to read versions: failed to read I2C: endpoint stalled or request not supported
16:33:31:0183 FuEngine             failed to add device usb:06:00:01:05: failed to get version: endpoint stalled or request not supported
```

**Steps to Reproduce**
```
sudo fwupdtool get-devices --plugins vli
```

**Expected behavior**
No errors

**fwupd version information**
```
1.6.1-g12
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):source

[vli.txt](https://github.com/fwupd/fwupd/files/6657329/vli.txt)

hughsie:
I think that's by design; IIUC the MSP device enumerates find when you do it using the USB3 side of the VLI chip, but not on the USB-2 side.

superm1:
Then shouldn't the errors downgrade? They're loud and error level..

hughsie:
Hmm good question. Will investigate whether we can use a more specific GUID in the morning.

hughsie:
We can, but there are a lot of other things to fix before it works :)

PR #3383, opened 2021-06-17
vli: Be more specific when matching the MSP430

hughsie:
Add a FuDevice->ready() vfunc that is run after ->setup()
This allows plugins to perform setup actions that require quirk matches
after the instance IDs have been converted. This allows us to also
remove FuVliDevice->setup() as the ordering chain has been broken.

There are two USB-3 hubs internally (with the same VID&PID), and the
MSP430 is only attached to the top-level one.

Fixes https://github.com/fwupd/fwupd/issues/3366

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -3618,6 +3618,12 @@ fu_device_setup (FuDevice *self, GError **error)
 	/* convert the instance IDs to GUIDs */
 	fu_device_convert_instance_ids (self);
 
+	/* subclassed */
+	if (klass->ready != NULL) {
+		if (!klass->ready (self, error))
+			return FALSE;
+	}
+
 	priv->done_setup = TRUE;
 	return TRUE;
 }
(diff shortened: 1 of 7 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -3618,6 +3618,12 @@ fu_device_setup (FuDevice *self, GError **error)
 	/* convert the instance IDs to GUIDs */
 	fu_device_convert_instance_ids (self);
 
+	/* subclassed */
+	if (klass->ready != NULL) {
+		if (!klass->ready (self, error))
+			return FALSE;
+	}
+
```
superm1, line 3626: I was kind of thinking it would be a new device phase all together rather than part of setup.  IE 

`fu_device_ready` calls this exact code and has a check for `ready` that could be used potentially.

It's a lot more spaghetti though, so let me know if you don't think that is valuable.

hughsie: I actually did this first, but it's a lot more state machinery we didn't actually need, so I removed it.

hughsie: Really, ->ready is just a ->setup_phase_2

superm1: OK..

superm1 (approved):

PR #3390, opened 2021-06-18
Allow adding and removing custom flags on devices

hughsie:
The CustomFlags feature is a bit of a hack where we just join the flags
and store in the device metadata section as a string. This makes it
inefficient to check if just one flag exists as we have to split the
string to a temporary array each time.

Rather than adding to the hack by splitting, appending (if not exists)
then joining again, store the custom flags in an actual array.

This means we have to change the fu_device_get_custom_flags() ABI
slightly, but this isn't something plugins should really be using.

This allows us to support negating custom properties (e.g. ~hint) and
also allows quirks to append custom values without duplicating them on
each GUID match, e.g.

	[USB\VID_17EF&PID_307F]
	Plugin = customflag1
	[USB\VID_17EF&PID_307F&HUB_0002]
	Flags = customflag2

...would result in `customflag1,customflag2` which is the same as you'd
get from an enumerated device flag doing the same thing.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -51,6 +51,7 @@ typedef struct {
 	GRWLock				 metadata_mutex;
 	GPtrArray			*parent_guids;
 	GRWLock				 parent_guids_mutex;
+	GPtrArray			*custom_flags;	/* (nullable) */
 	guint				 remove_delay;	/* ms */
 	guint				 progress;
 	guint				 battery_level;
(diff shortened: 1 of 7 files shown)


hughsie:
@bleungatchromium I don't think this breaks cros-ec as the behaviour of `fu_device_set_custom_flags()` is to **overwrite** like before. But now `fu_device_add_custom_flag()` and `fu_device_remove_custom_flag()` exist I wondered if it would be easier to use -- I pushed a test branch wip/hughsie/custom-flags-cros-ec which is 100% untested. If you want to push to that please do, but I'll not test or suggest a PR until this one is merged.

libfwupdplugin/fu-device.c
```diff
@@ -1177,6 +1178,128 @@ fu_device_add_child_by_kv (FuDevice *self, const gchar *str, GError **error)
...
+	return FALSE;
+}
+
+static gboolean
+fu_device_maybe_remove_flag (FuDevice *self, const gchar *hint)
+{
+	FwupdDeviceFlags flag;
+	FuDeviceInternalFlags internal_flag;
+
+	/* this a negated device flag */
```
superm1, line 1207: aren't you missing something like:
`g_str_has_prefix (hint, "~")`

libfwupdplugin/fu-device.c
```diff
@@ -2531,11 +2627,14 @@ fu_device_set_custom_flags (FuDevice *self, const gchar *custom_flags)
  *
  * Since: 1.1.0
```
superm1, line 2628: shouldn't this bump up to 1.6.2 since the qualifiers are changing on the return type?

hughsie: I was hoping to ignore that given that nothing now in-tree uses it.

superm1:
So we added custom flags at the time that we were super limited on number of device flags (IE before we had the idea of internal flags).  Another idea would be to just scrap them all together and put custom flags into internal flags slots.  The downside there is of course that LOTS of internal flag slots get taken up pretty quickly and definitions end up getting overloaded between plugins.

hughsie:
>  Another idea would be to just scrap them all together and put custom flags into internal flags slots

Maybe that's better. Will prototype something.

hughsie:
> Will prototype something.

I think https://github.com/fwupd/fwupd/pull/3392 is a much better idea. It essentially allows plugins to register a bitmask and a *place* to put the value, which gives each plugin it's own (or multiple!) 64 bit spaces.

PR #3392, opened 2021-06-18
Allow adding and removing custom flags on devices

hughsie:
The CustomFlags feature is a bit of a hack where we just join the flags
and store in the device metadata section as a string. This makes it
inefficient to check if just one flag exists as we have to split the
string to a temporary array each time.

Rather than adding to the hack by splitting, appending (if not exists)
then joining again, store the flags in the plugin privdata directly.

This allows us to support negating custom properties (e.g. ~hint) and
also allows quirks to append custom values without duplicating them on
each GUID match, e.g.

[USB\VID_17EF&PID_307F]
Plugin = customflag1
[USB\VID_17EF&PID_307F&HUB_0002]
Flags = customflag2

...would result in customflag1,customflag2 which is the same as you'd
get from an enumerated device flag doing the same thing.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -70,6 +70,8 @@ typedef struct {
 	GPtrArray			*retry_recs;	/* of FuDeviceRetryRecovery */
 	guint				 retry_delay;
 	FuDeviceInternalFlags		 internal_flags;
+	guint64				 private_flags;
+	GPtrArray			*private_flag_items;	/* (nullable) */
 } FuDevicePrivate;
 
 typedef struct {
@@ -311,6 +313,92 @@ fu_device_has_internal_flag (FuDevice *self, FuDeviceInternalFlags flag)
 	FuDevicePrivate *priv = GET_PRIVATE (self);
 	g_return_val_if_fail (FU_IS_DEVICE (self), FALSE);
 	return (priv->internal_flags & flag) > 0;
+}
+
+/**
+ * fu_device_add_private_flag:
+ * @self: a #FuDevice
+ * @flag: a device flag
+ *
+ * Adds a private flag that can be used by the plugin for any purpose.
+ *
+ * Since: 1.6.2
+ **/
+void
+fu_device_add_private_flag (FuDevice *self, guint64 flag)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_if_fail (FU_IS_DEVICE (self));
+	priv->private_flags |= flag;
+}
+
+/**
+ * fu_device_remove_private_flag:
+ * @self: a #FuDevice
+ * @flag: a device flag
+ *
+ * Removes a private flag that can be used by the plugin for any purpose.
+ *
+ * Since: 1.6.2
+ **/
+void
+fu_device_remove_private_flag (FuDevice *self, guint64 flag)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_if_fail (FU_IS_DEVICE (self));
+	priv->private_flags &= ~flag;
+}
+
+/**
+ * fu_device_has_private_flag:
+ * @self: a #FuDevice
+ * @flag: a device flag
+ *
+ * Tests for a private flag that can be used by the plugin for any purpose.
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_device_has_private_flag (FuDevice *self, guint64 flag)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (FU_IS_DEVICE (self), FALSE);
+	return (priv->private_flags & flag) > 0;
+}
+
+/**
+ * fu_device_get_private_flags:
+ * @self: a #FuDevice
+ *
+ * Returns all the private flags that can be used by the plugin for any purpose.
+ *
+ * Returns: flags
+ *
+ * Since: 1.6.2
+ **/
+guint64
+fu_device_get_private_flags (FuDevice *self)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_val_if_fail (FU_IS_DEVICE (self), G_MAXUINT64);
+	return priv->private_flags;
+}
+
+/**
+ * fu_device_set_private_flags:
+ * @self: a #FuDevice
+ * @flag: flags
+ *
+ * Sets private flags that can be used by the plugin for any purpose.
+ *
+ * Since: 1.6.2
+ **/
+void
+fu_device_set_private_flags (FuDevice *self, guint64 flag)
+{
+	FuDevicePrivate *priv = GET_PRIVATE (self);
+	g_return_if_fail (FU_IS_DEVICE (self));
+	priv->private_flags = flag;
 }
 
 /**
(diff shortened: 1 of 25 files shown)


hughsie:
@superm1 if you like the idea and the API let me know and I'll convert everything else.

libfwupdplugin/fu-device.c
```diff
@@ -2472,30 +2473,114 @@ fu_device_add_flag (FuDevice *self, FwupdDeviceFlags flag)
...
+ * @value: an integer value
+ * @value_str: a string that represents @value
+ *
+ * Registers a custom device flag that can be used from quirk files.
+ *
+ * Since: 1.6.2
+ **/
+void
+fu_device_register_custom_flag (FuDevice *self,
+				guint64 *bitfield,
```
superm1, line 2516: Hmm, I think it's a little bit weird to send in a bitfield pointer that gets written to.  Why not instead just have a guint64 bitfield in the local device that you assign meaning to each bit when you make the custom flag?  It's very similar to what you're doing already but the ownership would be with the `fu-device`.

hughsie: Then we'd have `fu_device_add_flag()`, `fu_device_add_internal_flag()` and presumably something like `fu_device_add_private_flag` which gets a bit confusing, no? We'd also need `fu_device_remove_private_flag()`, `fu_device_get_private_flags()` etc too. My idea of using the bitfield pointer is that we can use the thing _already_ being used by the plugin.

superm1: But what happens if you need to refer to the private flags _between_ two plugins?  Maybe I answered my own Q...  that makes the stronger argument for an internal flag in that case.

hughsie: > But what happens if you need to refer to the private flags between two plugins

I don't think we want to allow that.

superm1: Okay this approach looks good to me then. I think the doc on the register function needs a little more meat to explain how it works though.

hughsie: I'll port some more plugins and see if the approach works there. Thanks for the sanity check.

hughsie: >  Why not instead just have a guint64 bitfield in the local device

I think you might be right.

hughsie: Narrator: _he was right_.

hughsie:
@bleungatchromium I think I've probably broken `cros_ec_usb` -- any chance you could test this branch and fix up the failure? I think you'll need to sprinkle in a few `fu_device_remove_private_flag()` calls in the right place but it's late here and I'm nowhere near the right desk to test on actual hardware. I also think `in_bootloader` could be a private flag now and make things even simpler.

bleungatchromium:
Ok. I'll take a look at it. Thanks!

libfwupdplugin/fu-self-test.c
```diff
@@ -1390,6 +1390,36 @@ fu_device_inhibit_func (void)
...
+
+	fu_device_set_custom_flags (device, "foo");
+	g_assert_cmpint (fu_device_get_private_flags (device), ==, TEST_FLAG_FOO);
+	fu_device_set_custom_flags (device, "bar");
+	g_assert_cmpint (fu_device_get_private_flags (device), ==, TEST_FLAG_FOO | TEST_FLAG_BAR);
+	fu_device_set_custom_flags (device, "~bar");
+	g_assert_cmpint (fu_device_get_private_flags (device), ==, TEST_FLAG_FOO);
+	fu_device_set_custom_flags (device, "baz");
+	g_assert_cmpint (fu_device_get_private_flags (device), ==, TEST_FLAG_FOO);
+
```
superm1, line 1413: I think you want another test case of trying to use an unregistered flag `baz`.

plugins/ccgx/fu-ccgx-dmc-device.c
```diff
@@ -541,10 +549,9 @@ static gboolean
 fu_ccgx_dmc_device_attach (FuDevice *device, GError **error)
 {
 	FuCcgxDmcDevice *self = FU_CCGX_DMC_DEVICE (device);
-	gboolean manual_replug = FALSE;
+	gboolean manual_replug;
 
-	if (fu_device_has_custom_flag (device, "has-manual-replug"))
-		manual_replug = TRUE;
+	manual_replug = fu_device_has_private_flag (device, FU_CCGX_DMC_DEVICE_FLAG_HAS_MANUAL_REPLUG);
```
superm1, line 554: nit: just combine the lines, no?

hughsie: I thought it was a bit long.

libfwupdplugin/fu-self-test.c
hughsie: like the above? Or did you expect the baz value to appear in the to_string output, e.g. `PrivateFlags:         foo,unknown:4\n` etc?

plugins/jabra/fu-plugin-jabra.c
```diff
@@ -33,7 +33,7 @@ fu_plugin_update_cleanup (FuPlugin *plugin,
 
 	/* check for a property on the *dfu* FuDevice, which is also why we
 	 * can't just rely on using FuDevice->cleanup() */
-	if (!fu_device_has_custom_flag (device, "attach-extra-reset"))
+	if (!fu_device_has_vendor_id (device, "USB:0x0B0E"))
```
superm1, line 36: Perhaps should this become an internal plugin flag since it needs to go across two plugins?

hughsie: Yes, this did feel a bit of hack. Move it to a ATTACH_EXTRA_RESET flag perhaps?

libfwupdplugin/fu-self-test.c
superm1: I think there should be some error here somehow - you shouldn't be able to add a private flag when you haven't registered it.  Maybe actually it should be an assertion.

plugins/jabra/fu-plugin-jabra.c
superm1: Yup

hughsie: Fixed, indirectly.

libfwupdplugin/fu-self-test.c
hughsie: I went with critical warning, as then you can test for it in the self tests.

superm1 (dismissed):

superm1:
(Of course need to rebase on master for the jabra stuff - but the changes are otherwise good to me now)

hughsie:
I think @bleungatchromium needs to sprinkle some love onto cros-ec, then I'll rebase. Thanks!

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -856,19 +860,18 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
...
-		fu_device_set_custom_flags (device, "ro-written,rw-written");
+		if (fu_device_has_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_RO_WRITTEN))
+			fu_device_add_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_RW_WRITTEN);
+	} else if (fu_device_has_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_RW_WRITTEN)) {
+		fu_device_add_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_RO_WRITTEN);
 	} else {
-		fu_device_set_custom_flags (device, "ro-written");
+		fu_device_remove_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_RW_WRITTEN);
+		fu_device_add_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_RO_WRITTEN);
 	}
```
bleungatchromium, line 870: I think this whole section can be simplified a bit. I'll work on this.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -760,7 +762,9 @@ fu_cros_ec_usb_device_write_firmware (FuDevice *device,
 	FuCrosEcFirmware *cros_ec_firmware = FU_CROS_EC_FIRMWARE (firmware);
 	gint num_txed_sections = 0;
 
-	if (fu_device_has_custom_flag (device, "rebooting-to-ro")) {
+	fu_device_remove_private_flag (device, FU_CROS_EC_USB_DEVICE_FLAG_SPECIAL);
```
bleungatchromium, line 765: I'm trying to figure out how the "special" custom flag worked before, and if this is the right place to remove it.

bleungatchromium: Interesting. Many of these flags previously got swept up and removed implicitly by the way the custom flags overwrote each other. Now that private flags are a bit more sticky, we can be more explicit with this one and the rebooting to ro.

I'll find the right place to manipulate each of these transitory flags.

Thanks for doing this, @hughsie! This makes this whole thing a lot easier to follow, at the end of all of this.

superm1: Awesome!  Go ahead and rebase the branch and force push after you're done.

hughsie: Nah, @bleungatchromium just commit a "fixup" commit to the branch and I'll rebase and squash. I know how to resolve the conflict that will happen. :)

bleungatchromium:
I pushed one commit onto the branch. Not sure how to get it to update here though.

https://github.com/fwupd/fwupd/commit/8484665dac7ee463c924c86c932938d68d59f2a3

superm1:
> I pushed one commit onto the branch. Not sure how to get it to update here though.
> 
> [8484665](https://github.com/fwupd/fwupd/commit/8484665dac7ee463c924c86c932938d68d59f2a3)

Oh you pushed to a new branch.  I just cherry-picked it here and squashed it.  Hopefully I got the rebase right, @hughsie can double check it.

superm1:
> I pushed one commit onto the branch. Not sure how to get it to update here though.
> 
> [8484665](https://github.com/fwupd/fwupd/commit/8484665dac7ee463c924c86c932938d68d59f2a3)

@bleungatchromium BTW, I noticed that pre-commit CI failed on yours for spelling errors, make sure your update your local environment by running `./contrib/setup` to get a local pre-commit hook setup.

bleungatchromium:
> > I pushed one commit onto the branch. Not sure how to get it to update here though.
> > [8484665](https://github.com/fwupd/fwupd/commit/8484665dac7ee463c924c86c932938d68d59f2a3)
> 
> Oh you pushed to a new branch. I just cherry-picked it here and squashed it. Hopefully I got the rebase right, @hughsie can double check it.

Oh I see what I did. I'll get it right next time. Thanks!

PR #3394, opened 2021-06-18
Bump minimum required meson version to correct build warning

jigpu:
Sets the minimum required version to 0.50.0 since we are using features
which are not available prior to that release:

    WARNING: Project specifies a minimum meson_version '>=0.47.0' but uses
    features which were added in newer versions:
     * 0.49.0: {'/ with string arguments'}
     * 0.50.0: {'install arg in configure_file'}

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('fwupd', 'c',
   version : '1.6.2',
   license : 'LGPL-2.1+',
-  meson_version : '>=0.47.0',
+  meson_version : '>=0.50.0',
   default_options : ['warning_level=2', 'c_std=c99'],
 )
 


hughsie:
I think the warning is bogus; it works with the lower meson versions just fine. We've got ifdefs in the code so we can still compile on old Ubuntu versions and old RHEL releases. Thanks tho!

superm1:
The only thing I'm not sure about is gi-docgen now..  I think it really does require a newer version.  We might double check that..

hughsie:
gi-gocgen does indeed need 0.55.0

superm1:
It seems we're actually not guarding some of the new stuff either anymore.

I see ` * 0.50.0: {'install arg in configure_file'}` which is in a lot of places.

superm1:
OK so with current arguments (meson 0.47) the first error that comes up with default options is:

```
|
|Executing subproject fwupd-efi.
|
Couldn't use fallback subproject in subprojects/fwupd-efi for the dependency fwupd-efi
Reason: Meson version is 0.47.0 but project requires >=0.53.0.

meson.build:406:4: ERROR:  Native dependency 'fwupd-efi' not found
```

So if I try with `-Defi_binary=false`
```
|
|Executing subproject gi-docgen.
|
Couldn't use fallback subproject in subprojects/gi-docgen for the dependency gi-docgen
Reason: Meson version is 0.47.0 but project requires >= 0.55.0.

meson.build:555:2: ERROR:  Native dependency 'gi-docgen' not found

A full log can be found at /home/supermario/src/fwupd/build/meson-logs/meson-log.txt
```

So I need to do `-Defi_binary=false -Ddocs=none` in order to compile successfully with older meson..

hughsie:
So maybe it's best to just merge this and anything super old (looking at you RHEL 8) can just patch downstream if required.

superm1 (approved):

PR #3397, opened 2021-06-18
Add a new path for configuring runtime firmware load path

superm1:
The kernel allows configuring this, and certain plugins will use this
path to be able to load firmware in the kernel driver through runtime,
but to flash to the device.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.h
+++ b/libfwupdplugin/fu-common.h
@@ -67,6 +67,7 @@ typedef guint FuEndianType;
  * @FU_PATH_KIND_SYSFSDIR_SECURITY:	The sysfs security location (IE /sys/kernel/security)
  * @FU_PATH_KIND_ACPI_TABLES:		The location of the ACPI tables
  * @FU_PATH_KIND_LOCKDIR:		The lock directory (IE /run/lock)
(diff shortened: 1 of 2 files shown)


libfwupdplugin/fu-common.h
```diff
@@ -67,6 +67,7 @@ typedef guint FuEndianType;
  * @FU_PATH_KIND_SYSFSDIR_SECURITY:	The sysfs security location (IE /sys/kernel/security)
  * @FU_PATH_KIND_ACPI_TABLES:		The location of the ACPI tables
  * @FU_PATH_KIND_LOCKDIR:		The lock directory (IE /run/lock)
+ * @FU_PATH_KIND_FIRMWARE_SEARCH:	The path to configure the kernel policy for runtime loading other than /lib/firmware (IE /sys/module/firmware_class/parameters/path)
```
hughsie, line 70: This seems super Linux-specific, no?

superm1: Yes it is. Let's see if it's useful for modem manager plugin for now first before deciding what to do with it.

superm1:
will come in via #3422

PR #3401, opened 2021-06-21
Add support for soft-requirements that can be ignored with --force

hughsie:
This allows OEMs to suggest that certain components are updated to specific
versions, but allowing end-users to override this as required.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1645,6 +1645,28 @@ fu_engine_check_trust (FuInstallTask *task, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_engine_check_soft_requirement (FuEngine *self,
+				  FuEngineRequest *request,
+				  XbNode *req,
+				  FuDevice *device,
+				  FwupdInstallFlags flags,
+				  GError **error)
+{
+	g_autoptr(GError) error_local = NULL;
+	if (!fu_engine_check_requirement (self, request, req, device,
+					  flags, &error_local)) {
+		if (flags & FWUPD_INSTALL_FLAG_FORCE) {
+			g_debug ("ignoring soft-requirement due to --force: %s",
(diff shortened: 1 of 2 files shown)


superm1:
Can you give a better idea of what sort of requirements this is used for?  It seems like possibly a slippery slope to allow these kinds of overrides.  The metadata is either right or wrong..

hughsie:
> Can you give a better idea of what sort of requirements this is used for

Sure. The specific example here is from a tier 1 server OEM who release a set of updates once per quarter. It's only suggested that the end user updates from the previous n-3 quarter releases (and not older releases), and it's only supported that updates are installed in a specific order. In the field admins however want to override this, either due to security policy requirements or to work around hardware issues.

By default I only want admins to do the updates as the OEM suggests, (and has _tested_), but in the real world we need an escape hatch of --force. The OEM promises to add actual `<requirements>` if one update really does depend on another.

superm1:
OK... So on the server side can we have some enforcement that it doesn't get used for more types of things perhaps, so we can control the flow?

superm1 (approved):
I think realistically if updates like this are going to be made available, this feature should be backported to stable release branches as well.

hughsie:
> So on the server side can we have some enforcement that it doesn't get used for more types of things perhaps, so we can control the flow?

I guess we could have a `bool Vendor::can_use_soft_deps` but I don't think it's going to be abused; I mean if it is we could add it then, but it seems overkill.

> this feature should be backported to stable release branches as well

Sure, I can do that. I was going to just dep on 1.6.2 for this OEM, as they'll need modifications to the Redfish plugin too. Don't merge this until I get the LVFS up and working and we can test it on real hardware.

PR #3408, opened 2021-06-22
Add "lenovo-firmware" plugin

superm1:
This plugin is added specifically to determine whether Boot Order Lock is enabled and to prevent uefi-capsule updates from running when it is enabled.  I see it as a bare minimum right now, but I think there are a lot more things that could be done by Lenovo with it depending upon the direction that they want to go.

It uses thinklmi (which is part of kernel 5.14).  While coming up with it I found a few problems that I submitted patches for:
* https://patchwork.kernel.org/project/platform-driver-x86/patch/20210622175516.10100-1-mario.limonciello@amd.com/
* https://patchwork.kernel.org/project/platform-driver-x86/patch/20210622182613.1657-1-mario.limonciello@amd.com/

I don't think we should merge this plugin until those are merged, particularly the second one because the value read from `current_value` changes from what is in the kernel module today.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

New plugin checklist:

these are N/A for a simple plugin like this::
- [x] Fill out `README.md` with update protocol
- [x] Fill out `README.md` with any custom quirks and flags
- [x] Fill out `README.md` with the vendor ID security value
- [x] Implement `FuFirmware->write()` and include at least one fuzzer testcase in `src/fuzzing/firmware` for any custom `FuFirmware` subclass
- [x] Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc).

CC @mrhpearson


plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
```diff
@@ -0,0 +1,88 @@
...
+
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_BETTER_THAN, "bios");
```
hughsie, line 15: What's this for?

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
```diff
@@ -0,0 +1,88 @@
...
+		g_prefix_error (error, "failed to get current value: ");
+		return FALSE;
+	}
+	g_strchomp (buf);
+
+	if (g_strcmp0 (buf, "Enable") != 0) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_NOTHING_TO_DO,
+				     "not locked");
```
hughsie, line 69: I'm not enamoured with returning success and value like this if we don't have to. Can't we return FALSE for error, and just pass a `gboolean *enabled (out)` or something?

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
```diff
@@ -0,0 +1,88 @@
...
+
+gboolean
+fu_plugin_startup (FuPlugin *plugin, GError **error)
+{
+	g_autofree gchar *sysfsfwdir = NULL;
+	g_autofree gchar *thinklmidir = NULL;
+
+	/* already exists */
+	sysfsfwdir = fu_common_get_path (FU_PATH_KIND_SYSFSDIR_FW_ATTRIB);
+	thinklmidir = g_build_filename (sysfsfwdir, "thinklmi", NULL);
```
hughsie, line 26: do we have to "watch" this? i.e. can you change it at runtime?

plugins/meson.build
```diff
@@ -28,6 +28,7 @@ subdir('hailuck')
 subdir('intel-spi')
 subdir('iommu')
 subdir('jabra')
+subdir('lenovo-firmware')
```
hughsie, line 31: why the `-firmware`? maybe just `lenovo`? or `lenovo-thinklmi` perhaps?

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
superm1: You can certainly _change_ it in `thinklmi` runtime, but whether it's effective before the next boot, that's a question for @mrhpearson to find out the answer.

I notice that thinklmi is missing a `reboot_required` attribute, which I think if things aren't effective until the next boot the driver probably should add if posisble.

plugins/meson.build
superm1: I don't want to get in the business of just `lenovo` because what happens when they have some other thing, do they end up with both a `lenovo` and `lenovo-widget` plugin?

I specifically chose `lenovo-firmware` because I think there are other interesting things they could do when they're on their own systems which might not be `thinklmi` based.  But if that's not really the eventual direction I would concede to `lenovo-thinklmi`

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
superm1: sure

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
superm1: My original thought of looking what mode it's operating in, but it's really not needed.  I'll drop it.

plugins/meson.build
hughsie: What's lmi out of interest?

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
mrhpearson: Hi,
A reboot is required I'm afraid.

We don't have the reboot_required attribute yet as the support isn't provided by the BIOS. At the moment everything needs a reboot (that's certainly been my experience in testing).

I thought about adding it to the driver - but was worried that SW shouldn't really be doing something that the BIOS should provide. I'm hoping the BIOS team add it in the future. If you think that's incorrect and we should add the attribute and set it whenever something is changed happy to do that.

mrhpearson:
Thanks for doing this Mario - looks great
As an aside - I want to check out the 2nd patch on a couple of systems just to check it's working as expected there
Mark

superm1:
> Thanks for doing this Mario - looks great
> As an aside - I want to check out the 2nd patch on a couple of systems just to check it's working as expected there
> Mark

Yeah, as mentioned on LKML, please check v2 of the series instead.  I found a problem with the patches with some testing.

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
superm1: @mrhpearson just to make sure you didn't miss this - can you please check explicitly if changing it requires a reboot to be effective or it's immediate?  If it's reboot required, I think you need to also populate
`/sys/class/firmware-attributes/*/attributes/pending_reboot` on any attribute change and this plugin should look for it.

superm1: >A reboot is required I'm afraid.

No problem, that was actually my expectation too.

>We don't have the reboot_required attribute yet as the support isn't provided by the BIOS. At the moment everything needs a reboot (that's certainly been my experience in testing).
> I thought about adding it to the driver - but was worried that SW shouldn't really be doing something that the BIOS should provide. I'm hoping the BIOS team add it in the future. If you think that's incorrect and we should add the attribute and set it whenever something is changed happy to do that.

Yes - I think it's important.  If you look at this from a userspace software perspective, userspace software should be able to say something along the lines of, this is what is set for the next boot, and options might not apply yet.
I think that fwupd should look `reboot_pending` and prevent updates until the value goes from 1->0 from the next boot to be safe.

FWIW Dell BIOS doesn't provide it either, and dell-wmi-sysman just caches it based on there being a change.  I think this is perfectly fine to do in `think-lmi` as well.

superm1: Also, IIRC `dell-wmi-sysman` will emit a `CHANGE` `uevent` that userspace can monitor for.   to @hughsie's point, basically if that exists then we don't need to poll the file or anything.  We monitor for that change uevent and if it comes through then we run a callback to add inhibits.

plugins/meson.build
superm1: If I was gonna guess, "Lenovo Management Interface"?

mrhpearson: I think it's 'Linux Management Interface', as the team doing it wanted to differentiate it from 'Windows Management Interface' for WMI - at least I think I saw that on a slide deck somewhere :)

plugins/lenovo-firmware/fu-plugin-lenovo-firmware.c
mrhpearson: Ack - thanks. I'll look at adding that then. Thanks for the notes :)
I'll try and get that done this week but I've got a couple of half day events coming up, end of school year stuff and a vaccine jab (which last time knocked me out for a day) so it might slide into next week.

plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if (!fu_plugin_lenovo_firmware_pending_change (&pending, &error_pending)) {
+		g_debug ("%s", error_pending->message);
+		return;
+	}
+
+	if (locked)
+		fu_device_inhibit (device, "uefi-capsule-bootorder", "BootOrder is locked in firmware setup");
+	if (pending)
+		fu_device_inhibit (device, "uefi-capsule-pending-reboot", "UEFI BIOS settings update pending reboot");
```
hughsie, line 105: looks a bit wiiide, maybe split onto multiple lines and use {}?

plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
```diff
@@ -0,0 +1,106 @@
...
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+}
+
+gboolean
+fu_plugin_startup (FuPlugin *plugin, GError **error)
+{
+	g_autofree gchar *sysfsfwdir = NULL;
+	g_autofree gchar *thinklmidir = NULL;
+
+	/* already exists */
```
hughsie, line 23: do we need a kernel_version check here? No is a valid answer.

hughsie (approved):
LGTM, a few nits -- I don't mind if you want to get this in and then fix up the monitor stuff later.

plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
superm1: No, older kernel and not a Lenovo system just won't run the plugin, should abort startup.

plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
superm1: split up

plugins/lenovo-thinklmi/meson.build
```diff
@@ -0,0 +1,28 @@
+if get_option('plugin_uefi_capsule')
+
+cargs = ['-DG_LOG_DOMAIN="FuPluginLenovoThinkLmi"']
```
hughsie, line 3: do we maybe want to mock this up for a self test? e.g. set up a `tests/thinklmi/attributes/BootOrderLock/current_value`

superm1: done

plugins/lenovo-thinklmi/meson.build
```diff
@@ -0,0 +1,61 @@
...
+if get_option('tests')
+  testdatadir = join_paths(meson.current_source_dir(), 'tests')
+  testdatadirs = environment()
+  testdatadirs.set('G_TEST_SRCDIR', meson.current_source_dir())
+  testdatadirs.set('G_TEST_BUILDDIR', meson.current_build_dir())
+  testdatadirs.set('FWUPD_LOCALSTATEDIR', '/tmp/fwupd-self-test/var')
+  cargs += '-DTESTDATADIR="' + testdatadir + '"'
+  cargs += '-DPLUGINBUILDDIR="' + meson.current_build_dir() + '"'
+  e = executable(
+    'lenovo-self-test',
```
hughsie, line 38: this should be lenovo-thinklmi-self-test

hughsie (approved):
One tiny nit then LGTM, thanks.

PR #3413, opened 2021-06-23
Add plugin for ECs that use custom protocols

SergiiDmytruk:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

***

- [x] Fill out `README.md` with update protocol
- [x] Fill out `README.md` with any custom quirks and flags
- [x] Fill out `README.md` with the vendor ID security value
- [ ] Implement `FuFirmware->write()` and include at least one fuzzer testcase in `src/fuzzing/firmware` for any custom `FuFirmware` subclass.

***

This plugin copies structure of SuperIO plugin. The only implementation provided targets Tuxedo laptops that use IT5570. Previous attempt to support this EC was in #3339.

I didn't get where to document information about CI, but this is hardware-dependent plugin, which can't be tested easily except for building.


macpijan:
According to the discussion here: https://review.coreboot.org/c/flashrom/+/55715
the following plugin structure should work for much more Clevo laptops, and also
for more devices which use the ITE EC framework for EC firmware development.

hughsie:
This looks much better, really good. One question; why a new plugin and not just make changes to the existing superio plugin? It seems much the same. Would you like me to have a go at merging them for you?

plugins/ec/README.md
```diff
@@ -0,0 +1,61 @@
...
+Configuration
+-------------
+
+The following parameters can be specified in `/etc/fwupd/ec.conf`.
+
+**AutoloadAction** parameter can be set to `none`, `disable`, `seton` or
+`setoff`. `none` is the default.
+
+**DoNotRequireAC** parameter is boolean (`true` or `false`).  `false` is the
+default.
```
hughsie, line 39: can you explain these both a bit please? e.g. DoNotRequireAC is confusing as we already allow the user to --force the update on battery power.

SergiiDmytruk: I don't really know anything about `AutoloadAction` except that we disable it according to update instructions.

`DoNotRequireAC` is for internal check which I added due to this error:
```
23:16:36:0205 FuPluginUpower       failed to query power type
```
So builtin check doesn't work in this case (`upowerd` is up and running).

SergiiDmytruk:
> One question; why a new plugin and not just make changes to the existing superio plugin? It seems much the same.

There are definitely similarities, yet it didn't seem like a good fit in this case and I thought it won't pass review. To make it work as part of the superio plugin, one would have to generalize or duplicate the simplest device functions and the rest would be explicitly disabled or just ignored. At the end it would look a lot like two plugins in one.

> Would you like me to have a go at merging them for you?

If you think it makes sense, then please give it a try.

plugins/ec/meson.build
```diff
@@ -0,0 +1,31 @@
...
+install_data(['ec.quirk'],
+  install_dir: join_paths(datadir, 'fwupd', 'quirks.d')
+)
+
+shared_module('fu_plugin_ec',
+  fu_hash,
+  sources : [
+    'fu-plugin-ec.c',
+    'fu-ec-device.c',
+    'fu-ec-it5570-device.c',
```
hughsie, line 13: Is this protocol specific to the it5570, or is it common to al the it55XX chips?

macpijan: @hughsie 
If you follow the discussion I linked above (https://review.coreboot.org/c/flashrom/+/55715) it may be common (at least some parts of it) to even more ITE chips (provided that they are using this class of firmware - described from the ITE EC framework). So it is more firmware framework-specific rather than a fully hardware feature.

```
Well, Tuxedo sells Clevo notebooks. Clevo uses ITE EC with ITE EC firmware framework. This code is *NOT* tuxedo-specific nor Clevo-specific, but ITE-specific and can be used with other ITE ECs as well (I know, because I know the original ITE tool - yes, a single one ;-) ). So, please do not artificially restrict this to Tuxedo. Thanks!
```

At that point we know for sure that the IT5570 needs different read/write functions (no idea if just 5570 or different 55xx chips as well). The other functions from `plugins/ec/fu-ec-it5570-device.c` can be **probably** shared between more devices.

hughsie: Okay, thanks. It looks common to other IT55xx to my eyes. I'll have a pull request for you in a couple of hours :)

plugins/ec/README.md
hughsie: Okay, we don't want to workaround the upower failure in the plugin. I'll remove that.

Can you find out more what AutoloadAction is actually for? I don't know if it's supposed to be per-machine (i.e. a DMI match) or per-chipset, i.e. in the quirk file. I feel like we don't want a config entry for this.

macpijan: @hughsie Don't know much more what it does currently, but it looks to me like it's rather chipset-specific. And for the IT5570 should be set to disable according to the official update instructions.

macpijan: Quirk is rather a better way than config indeed.

plugins/ec/meson.build
hughsie: @macpijan can you have a look at wip/hughsie/superio please? If that works, please create a new pull request then we can review in detail. If it doesn't can you work out why, add a FIXUP commit to the branch and then do the PR thing please? I think using superio is a much nicer thing.

macpijan: @hughsie Thank you, @SergiiDmytruk will test and report.

SergiiDmytruk:
Closing in favor of #3420.

PR #3414, opened 2021-06-23
Updating powerd plugin components

g2pop:
This change brings the powerd plugin up to date with local changes


plugins/powerd/meson.build
```diff
@@ -0,0 +1,26 @@
+if host_machine.system() == 'linux'
+  cargs = ['-DG_LOG_DOMAIN="FuPluginPowerd"']
+
+shared_module('fu_plugin_cros_powerd',
```
hughsie, line 4: i think either the module needs to be called fu_plugin_powerd or the directory needs to be cros-powerd. I guess it depends on whether powerd is CrOS specific or not.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,91 @@
...
+	GDBusProxy		*proxy;		/* nullable */
+};
+
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+	fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+	fu_plugin_alloc_data (plugin, sizeof (FuPluginData));
+        
+        return;
```
hughsie, line 23: you don't need a return from void

plugins/powerd/meson.build
g2pop: That makes sense. I think we've settled on just fu-plugin-powerd.

plugins/powerd/fu-plugin-powerd.c
g2pop: Thank you, I went back and adjusted this.

hughsie: If it helps, do you want me to make you a committer on the main repo, and then you can push a wip/g2pop/powerd branch? if it's WIP we can scribble all over it and then when it's ready for review we can squash it into a sensible commit and merge.

g2pop: Thank you, right now though, we still have more foundational work to do. I think we'll try to make the new plugin functional while on a fork, and then push it to a branch on the main repo for you to review at a higher level.

hughsie: Okay, sounds like a plan. Yell if you need any help or advice.

PR #3420, opened 2021-06-24
Add IT5570 to the superio plugin

SergiiDmytruk:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

***

The provided implementation targets Tuxedo laptops that use IT5570. Latest attempt to support this EC was in #3413.

### DIFF
--- a/plugins/superio/fu-superio-it55-device.c
+++ b/plugins/superio/fu-superio-it55-device.c
@@ -1,0 +1,637 @@
+/*
+ * Copyright (C) 2021, TUXEDO Computers GmbH
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <string.h>
+
+#include <fwupdplugin.h>
+
+#include "fu-superio-common.h"
+#include "fu-superio-it55-device.h"
+
+/* ROM of IT5570 consists of 64KB blocks. Blocks can be further subdivided in
+ * 256-byte chunks, which is especially visible when erasing the ROM. This is
+ * because in case of erasure offset within a block is specified in chunks (even
+ * though erasure is done one kilobyte at a time).
+ *
+ * Accessing ROM requires entering a special mode, which should be always left
+ * to restore normal operation of EC (handling of buttons, keyboard, etc.). */
+
+#define SIO_CMD_EC_WRITE_BLOCK		0x02
+#define SIO_CMD_EC_READ_BLOCK		0x03
+#define SIO_CMD_EC_ERASE_KBYTE		0x05
+#define SIO_CMD_EC_WRITE_1ST_KBYTE	0x06
+#define EC_ROM_ACCESS_ON_1		0xDE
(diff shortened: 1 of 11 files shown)


plugins/superio/fu-superio-it55-device.c
```diff
@@ -0,0 +1,680 @@
...
+
+	if (!fu_superio_device_ec_write_cmd (self, SIO_CMD_EC_ID, error))
+		goto error;
+	if (!fu_superio_device_ec_read_data (self, &tmp, error))
+		goto error;
+	is_it5570 = (tmp != 0x00 && tmp != 0xff);
+
+	/* leave ROM access mode */
+	if (!fu_superio_device_ec_write_cmd (self, EC_ROM_ACCESS_OFF, error))
+		return FALSE;
```
superm1, line 86: shouldn't this be set after you've checked if it's the right part?  If it's not the right part, you're shooting in the dark

hughsie:
@SergiiDmytruk the $54,000,000 question -- does it still work on actual hardware? :)

plugins/superio/fu-superio-it55-device.c
hughsie: I *think* it's okay, as we call the parent setup, which calls `fu_superio_device_check_id()` before the subclassed setup.

SergiiDmytruk: > shouldn't this be set after you've checked if it's the right part? If it's not the right part, you're shooting in the dark

@superm1 `0xf0` requires entering this mode to succeed, so there isn't many options. Apparently this mode is common across some ECs (maybe for all `IT55*`, but only `IT5570` does 1KB erasure and requires weird separate write of the first kilobyte).

> I _think_ it's okay, as we call the parent setup, which calls `fu_superio_device_check_id()` before the subclassed setup.

@hughsie The check in `fu_superio_device_check_id()` is disabled because it uses `->port`, which isn't defined for this EC.

SergiiDmytruk:
> does it still work on actual hardware? :)

After some fixes, it does. Most notably timeout in this case should be ~650ms and not ~250ms, but there were also wrong function calls and some stuff with quirks.

plugins/superio/fu-superio-it55-device.c
hughsie: buuuut, are we not allow-listing the models with the hwids in the quirk file anyway?

SergiiDmytruk: Yes, correct. So the commands should be applicable for the EC as long as information in the quirks is accurate.

superm1: If we're allowlisting all the machines by quirks the only value for it is if someone tries to add a machine to quirk list to see "if it works". In that regard I'd say this check unnecessary.

SergiiDmytruk: Dropped the check.

superm1 (dismissed):
no concerns by me

PR #3421, opened 2021-06-25
Add support for optionally installing UEFI updates from GRUB

superm1:
Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

@hughsie can you please check and see if this works on Fedora or there are details that are done differently?


plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -292,6 +293,165 @@ fu_uefi_copy_asset (const gchar *source, const gchar *target, GError **error)
...
+				     G_IO_ERROR_FAILED,
+#endif
+				     "chmod failed for %s", grub_target);
+			return FALSE;
+		}
+
+		/* refresh GRUB configuration */
+//Fedora needs
+// grub2-mkconfig -o "$(readlink -e /etc/grub2.conf)"
+//IIUC
```
hughsie, line 345: @martinezjavier is this true?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -292,6 +293,165 @@ fu_uefi_copy_asset (const gchar *source, const gchar *target, GError **error)
...
+			g_debug ("%s", output);
+		} else {
+			g_warning ("unable to refresh GRUB configuration");
+		}
+	}
+	grub_reboot = fu_common_find_program_in_path ("grub-reboot", NULL);
+	if (grub_reboot == NULL)
+		grub_reboot = fu_common_find_program_in_path ("grub2-reboot", NULL);
+	if (grub_reboot != NULL) {
+		g_autofree gchar *cmd = g_strdup_printf ("%s fwupd", grub_reboot);
```
hughsie, line 361: @martinezjavier can we do this by label?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -292,6 +293,165 @@ fu_uefi_copy_asset (const gchar *source, const gchar *target, GError **error)
...
+			    const gchar *target_app,
+			    GError **error)
+{
+	g_autofree gchar *grub_reboot = NULL;
+	g_autofree gchar *grub_target = NULL;
+	g_autofree gchar *sysconfdir = NULL;
+
+	/* verify GRUB snippet exists */
+	sysconfdir = fu_common_get_path (FU_PATH_KIND_SYSCONFDIR);
+	grub_target = g_build_filename (sysconfdir, "grub.d", "35_fwupd", NULL);
```
martinezjavier, line 308: I wonder if generating a `/etc/grub.d/35_fwupd` snippet at runtime is the best approach for this.

What if this is instead a static file that's just shipped by `fwupd` but that's just a no-op if `EnableGrubChainLoad` isn't enabled in `/etc/fwupd/uefi_capsule.conf`? After all, the GRUB snippets are just shell scripts so there's no point on replacing strings IMO (i.e: could just do a `chainloader $EFI_PATH`).

Something like [30_uefi-firmware](https://git.savannah.gnu.org/cgit/grub.git/tree/util/grub.d/30_uefi-firmware.in):

```
EFI_VARS_DIR=/sys/firmware/efi/efivars
EFI_GLOBAL_VARIABLE=8be4df61-93ca-11d2-aa0d-00e098032b8c
OS_INDICATIONS="$EFI_VARS_DIR/OsIndicationsSupported-$EFI_GLOBAL_VARIABLE"

if [ -e "$OS_INDICATIONS" ] && \
   [ "$(( $(printf 0x%x \'"$(cat $OS_INDICATIONS | cut -b5)"\') & 1 ))" = 1 ]; then
  LABEL="UEFI Firmware Settings"

  gettext_printf "Adding boot menu entry for UEFI Firmware Settings ...\n" >&2

  cat << EOF
menuentry '$LABEL' \$menuentry_id_option 'uefi-firmware' {
	fwsetup
}
EOF
fi
```

Another option is to make the `fwupd` snippet even more static and just `source` from it a variable that's contained in another file (i.e: `/boot/grub2/fwupd.cfg`). That's what we do in Fedora to [support ACL](https://github.com/rhboot/grub2/blob/fedora-35/util/grub.d/01_users.in) of boot entries for example:

```
cat << EOF
if [ -f \${prefix}/user.cfg ]; then
  source \${prefix}/user.cfg
  if [ -n "\${GRUB2_PASSWORD}" ]; then
    set superusers="root"
    export superusers
    password_pbkdf2 root \${GRUB2_PASSWORD}
  fi
fi
EOF
```

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -292,6 +293,165 @@ fu_uefi_copy_asset (const gchar *source, const gchar *target, GError **error)
...
+			return FALSE;
+
+		/* replace ESP path with what we detected */
+		str = g_string_new (buf);
+		fu_common_string_replace (str, "##EFIPATH##", target_app);
+		fu_common_string_replace (str, esp_path, "");
+		fu_common_string_replace (str, "##ESP##", esp_path);
+		if (!g_file_set_contents (grub_target, str->str, -1, error))
+			return FALSE;
+		if (chmod (grub_target, 0755) == -1) {
```
martinezjavier, line 330: Another advantage of not generating the snippet at runtime is that this won't be needed, since the file mode bits will be set at install time.

plugins/uefi-capsule/fu-uefi-bootmgr.c
martinezjavier: It is true. But I wonder if Fedora's grub2 package should just ship `update-grub`, which is just a stub for this command.

plugins/uefi-capsule/fu-uefi-bootmgr.c
martinezjavier: Yes, either menu entry id or label work.

martinezjavier:
@superm1 @hughsie I think this change makes sense, since the `chainloader` command in GRUB (on most distros) have support to call shim's verification using the `SHIM_LOCK` protocol, this would even work with Secure Boot enabled.

I know is a draft but would be good to explain the rationale for this change. My educated guess is that this is to workaround when shim is broken and isn't able to parse the EFI `BootXXXX` entries' arguments (which I believe is currently the case with the latest release).

hughsie:
> My educated guess is that this is to workaround when shim is broken

Exactly that.

plugins/uefi-capsule/fu-uefi-bootmgr.c
superm1: At least in Ubuntu it doesn't use `/etc/grub2.conf` though I think - it just goes right to something like `/boot/grub/grub.conf`

superm1: So maybe the right answer is a heuristic to check a list of paths to see if they exist and just use `grub2-mkconfig` with them?

plugins/uefi-capsule/fu-uefi-bootmgr.c
superm1: > Another option is to make the fwupd snippet even more static and just source from it a variable that's contained in another file (i.e: /boot/grub2/fwupd.cfg). 

I like this idea a lot.  When I get a chance I'll update the PR for it (unless @hughsie beats me to it).  Then we can also support systems that have read-only rootfs.  The fwupd package can just start shipping `/etc/grub.d/35_fwupd` and source the settings from somewhere actually writable.

>That's what we do in Fedora to support ACL of boot entries for example:

BTW TBH this sounds like it makes sense generally in GRUB, not just in Fedora.  Yall tried to upstream it?

superm1:
> (which I believe is currently the case with the latest release).

BTW - do you have a RHBZ# tracking this in Fedora?  This has come up enough lately that we should have a boilerplate response to say there is a known SB problem, here's some links about it.
I know Ubuntu has: https://bugs.launchpad.net/ubuntu/+source/shim/+bug/1929471
And this PR is in progress to fix things: https://github.com/rhboot/shim/pull/381

plugins/uefi-capsule/fu-uefi-bootmgr.c
hughsie: > unless @hughsie beats me to it

/me is neck deep in Redfish this week...

plugins/uefi-capsule/fu-uefi-bootmgr.c
superm1: Just checked on Ubuntu they do put it in `/boot/grub/grub.cfg` directly.
```
$ cat /usr/sbin/update-grub
#!/bin/sh
set -e
exec grub-mkconfig -o /boot/grub/grub.cfg "$@"
```

martinezjavier: Yes, I think we should do the same in Fedora and just update the GRUB config directly instead of the symlink in `/etc`. But I still wonder if is worth to also ship an update-grub wrapper script... @vathpela what do you think ?

plugins/uefi-capsule/fu-uefi-bootmgr.c
martinezjavier: > BTW TBH this sounds like it makes sense generally in GRUB, not just in Fedora. Yall tried to upstream it?

Yes, it's in our backlog for upstreaming (along with the [grub-set-password](https://github.com/rhboot/grub2/blob/fedora-35/util/grub-set-password.8) user-space tool we ship) but it's just that wasn't prioritized over the other downstream patches we are still carrying.

plugins/uefi-capsule/fu-uefi-bootmgr.c
superm1: Oh it's a cross filesystem symlink on Fedora?  In that case perhaps we should just do `grub-mkconfig -o /boot/grub/grub.cfg` across the board to make this as compatible as possible across distros (including old releases).

plugins/uefi-capsule/fu-uefi-bootmgr.c
superm1: I added something to match what you suggested (untested though so far)

plugins/uefi-capsule/fu-uefi-bootmgr.c
superm1: Can you re-check what I've pushed to the branch now?  Hopefully this works on Debian/Ubuntu/Fedora all.

hughsie:
I had to push a couple of commits to get it to work on Fedora, but then disaster struck on reboot:

PXL_20210629_112903821

and then pressing a key and trying to boot back into Fedora (all of them):

PXL_20210629_113003802 MP

...at that point it booted fine into Windows 10, which then spent 10 minutes doing an update, then rebooted, which then let me boot the Fedora entry again. It's entirely possible I broke something with my fixups, but the failure mode wasn't any better than shim...

superm1:
> I had to push a couple of commits to get it to work on Fedora, but then disaster struck on reboot:
> 
> PXL_20210629_112903821
> 
> (but /boot/efi/EFI/fedora/fwupdx64.efi does exist...) and then pressing a key and trying to boot back into Fedora (all of them):
> 
> PXL_20210629_113003802 MP
> 
> ...at that point it booted fine into Windows 10, which then spent 10 minutes doing an update, then rebooted, which then let me boot the Fedora entry again. It's entirely possible I broke something with my fixups, but the failure mode wasn't any better than shim...

I think it's because I was previously chomping the /boot prefix and forgot to do that when changing things around in this last force push or two.

superm1:
> I think it's because I was previously chomping the /boot prefix and forgot to do that when changing things around in this last force push or two.

That last fixup commit should fix it.

superm1:
The only other idea I had which I wanted to discuss was dropping the conffile alltogether for this.  Detect `BootCurrent` is GRUB or not.  If it is then use GRUB.  If it's not, then use old method.

plugins/cpu/fu-cpu-helper-cet-common.c
```diff
@@ -9,20 +9,29 @@
...
+#endif
+
 static void
 fu_cpu_helper_cet_testfn_fptr (void)
 {
 }
 
 static void
-__attribute__ ((noinline, noclone))
+__attribute__((__noinline__)) ATTRIBUTE_NOCLONE
```
hughsie, line 27: ?? Got any background?

hughsie: I think it specifically needs noinline, noclone for CET to be detected... Maybe drop this bit and we can check it in another PR?

superm1: yeah I'll move it over to another PR.  If noclone is needed, then we probably need to turn off CET checker when using clang.

hughsie:
@superm1 I've rebased this and fix up the conflicts. Can you check it still works? If so, LGTM for now. I'll probably split it up a bit later into difference subclasses, but that's not your concern for this PR IMO.

contrib/debian/rules
```diff
@@ -64,6 +64,9 @@ override_dh_install:
 	[ ! -d debian/tmp/usr/lib/modules-load.d ] || dh_install -pfwupd usr/lib/modules-load.d
 	[ ! -d debian/tmp/lib/modules-load.d ] || dh_install -pfwupd lib/modules-load.d
 	dh_missing -a --fail-missing
+	#enable GRUB instead of shim
+	[ ! -f debian/tmp/etc/fwupd/uefi_capsule.conf ] || sed -i 's,^#EnableGrubChainLoad=.*,EnableGrubChainLoad=true,' \
```
superm1, line 68: I guess the question I have is whether we want this behavior turned on by default anywhere?  I feel like we shouldn't pick and choose packaged distros, it should be all or none.

hughsie: My gut instinct is "none" and have it easily enable-able/testable for users having problems with shim. Maybe change the default when it's been tested a bit.

superm1: OK.  So eventually I think we also need to worry about the priority paths with COD, NVRAM, and GRUB.  Steady state my gut tells me we should have the order be to try these:
COD, GRUB, NVRAM

hughsie: I think CoD is the least well tested at the moment;lets release 1.6.2 defaulting to NVRAM and ask people to test chainloading in the release notes and report back.

superm1: Yeah I'm just saying eventually that is where we want to be.  Another release or two.

hughsie: I'll drop the Debian bits that change the default and rebase it for you then,

hughsie:
Closing in favour of https://github.com/fwupd/fwupd/pull/3421 -- both grub and CoD are making changes to the same code now.

PR #3422, opened 2021-06-25
Add firehose update with fw copy

ivanmikh:
Introduces support for firehose update and qcdm port handling for modem-manager plugin.

Same as #3373 , but adds copying of firehose binary into /lib/firmware/qcom folder and some minor bug fixes.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1175,6 +1175,12 @@ fu_common_get_path (FuPathKind path_kind)
 		if (tmp != NULL)
 			return g_strdup (tmp);
 		return g_strdup ("/sys/firmware/acpi/tables");
+	/* /sys/module/firmware_class/parameters/path */
+	case FU_PATH_KIND_FIRMWARE_SEARCH:
+		tmp = g_getenv ("FWUPD_FIRMWARESEARCH");
+		if (tmp != NULL)
+			return g_strdup (tmp);
+		return g_strdup ("/sys/module/firmware_class/parameters/path");
 	/* /etc */
 	case FU_PATH_KIND_SYSCONFDIR:
 		tmp = g_getenv ("FWUPD_SYSCONFDIR");
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -3384,3 +3384,118 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
 {
 	xb_builder_node_insert_text (bn, key, value ? "true" : "false", NULL);
 }
+
+/**
+ * fu_common_get_firmware_search_path:
+ * @error: (nullable): optional return location for an error
+ *
+ * Reads the FU_PATH_KIND_FIRMWARE_SEARCH and
+ * returns its contents
+ *
+ * Returns: a pointer to a gchar array
+ *
+ * Since: 1.6.2
+ **/
+gchar *
+fu_common_get_firmware_search_path (GError **error)
+{
+	gsize sz = 0;
+	g_autofree gchar *sys_fw_search_path = NULL;
+	g_autofree gchar *contents = NULL;
+
+	sys_fw_search_path = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	if (!g_file_get_contents(sys_fw_search_path, &contents, &sz, error))
+		return NULL;
+
+	/* remove newline character */
+	if (contents != NULL && sz > 0 && contents[sz - 1] == '\n')
+		contents[sz - 1] = 0;
+
+	g_debug ("read firmware search path (%" G_GSIZE_FORMAT "): %s", sz, contents);
+
+	return g_steal_pointer (&contents);
+}
+
+/**
+ * fu_common_set_firmware_search_path:
+ * @path: NUL-terminated string
+ * @error: (nullable): optional return location for an error
+ *
+ * Writes path to the FU_PATH_KIND_FIRMWARE_SEARCH
+ *
+ * Returns: %TRUE if successful
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_common_set_firmware_search_path (const gchar *path, GError **error)
+{
+#if GLIB_CHECK_VERSION(2,66,0)
+	g_autofree gchar *sys_fw_search_path_prm = NULL;
+
+	g_return_val_if_fail (path != NULL, FALSE);
+	g_return_val_if_fail (strlen (path) < PATH_MAX, FALSE);
+
+	sys_fw_search_path_prm = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+
+	g_debug ("writing firmware search path (%" G_GSIZE_FORMAT "): %s", strlen (path), path);
+
+	return g_file_set_contents_full (sys_fw_search_path_prm, path, strlen (path),
+				G_FILE_SET_CONTENTS_NONE, 0644, error);
+#else
+	FILE *fd;
+	gsize res;
+	g_autofree gchar *sys_fw_search_path_prm = NULL;
+
+	g_return_val_if_fail (path != NULL, FALSE);
+	g_return_val_if_fail (strlen (path) < PATH_MAX, FALSE);
+
+	sys_fw_search_path_prm = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	/* g_file_set_contents will try to create backup files in sysfs, so use fopen here */
+	fd = fopen (sys_fw_search_path_prm, "w");
+	if (fd == NULL) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_PERMISSION_DENIED,
+			     "Failed to open %s: %s",
+			     sys_fw_search_path_prm,
+			     g_strerror (errno));
+		return FALSE;
+	}
+
+	g_debug ("writing firmware search path (%" G_GSIZE_FORMAT "): %s", strlen (path), path);
+
+	res = fwrite (path, sizeof (gchar), strlen (path), fd);
+
+	fclose (fd);
+
+	if (res != strlen (path)) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_WRITE,
+			     "Failed to write firmware search path: %s",
+			     g_strerror (errno));
+		return FALSE;
+	}
+
+	return TRUE;
+#endif
+}
+
+/**
+ * fu_common_reset_firmware_search_path:
+ * @error: (nullable): optional return location for an error
+ *
+ * Resets the FU_PATH_KIND_FIRMWARE_SEARCH to an empty string
+ *
+ * Returns: %TRUE if successful
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_common_reset_firmware_search_path (GError **error)
+{
+	const gchar *contents = " ";
+
+	return fu_common_set_firmware_search_path (contents, error);
+}
--- a/plugins/modem-manager/fu-firehose-updater.c
+++ b/plugins/modem-manager/fu-firehose-updater.c
@@ -1,0 +1,748 @@
+/*
+ * Copyright (C) 2020 Aleksander Morgado <aleksander@aleksander.es>
+ * Copyright (C) 2021 Quectel Wireless Solutions Co., Ltd.
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <string.h>
+#include <sys/ioctl.h>
+
+#include "fu-firehose-updater.h"
+
+/* Maximum amount of non-"response" (e.g. "log") XML messages that can be
+ * received from the module when expecting a "response". This is just a safe
+ * upper limit to avoid reading forever. */
+#define MAX_RECV_MESSAGES                                       100
+
+/* When initializing the conversation with the firehose interpreter, the
+ * first step is to receive and process a bunch of messages sent by the
+ * module. The initial timeout to receive the first message is longer in case
+ * the module needs some initialization time itself; all the messages after
+ * the first one are expected to be received much quicker. The default timeout
+ * value should not be extremely long because the initialization phase ends
+ * when we don't receive more messages, so it's expected that the timeout will
+ * fully elapse after the last message sent by the module. */
+#define INITIALIZE_INITIAL_TIMEOUT_MS                           3000
+#define INITIALIZE_TIMEOUT_MS                                   250
+
+/* Maximum amount of time to wait for a message from the module. */
+#define DEFAULT_RECV_TIMEOUT_MS                                 15000
+
+/* The first configure attempt sent to the module will include all the defaults
+ * listed below. If the module replies with a NAK specifying a different
+ * (shorter) max payload size to use, the second configure attempt will be done
+ * with that new suggested max payload size value. Only 2 configure attempts are
+ * therefore expected. */
+#define MAX_CONFIGURE_ATTEMPTS                                  2
+
+/* Defaults for the firehose configuration step. The max payload size to target
+ * in bytes may end up being a different if the module requests a shorter one.
+ */
+#define CONFIGURE_MEMORY_NAME                                   "nand"
+#define CONFIGURE_VERBOSE                                       0
+#define CONFIGURE_ALWAYS_VALIDATE                               0
+#define CONFIGURE_MAX_DIGEST_TABLE_SIZE_IN_BYTES                2048
+#define CONFIGURE_MAX_PAYLOAD_SIZE_TO_TARGET_IN_BYTES           8192
+#define CONFIGURE_ZLP_AWARE_HOST                                1
+#define CONFIGURE_SKIP_STORAGE_INIT                             0
+
+enum {
+	SIGNAL_WRITE_PERCENTAGE,
+	SIGNAL_LAST
+};
+
+static guint signals [SIGNAL_LAST] = { 0 };
+
+struct _FuFirehoseUpdater {
+	GObject		 parent_instance;
+	gchar		*port;
+	FuIOChannel	*io_channel;
+};
+
+G_DEFINE_TYPE (FuFirehoseUpdater, fu_firehose_updater, G_TYPE_OBJECT)
+
+static void
+fu_firehose_updater_log_message (const gchar *action, GBytes *msg)
+{
+	const gchar *msg_data;
+	gsize msg_size;
+	g_autofree gchar *msg_strsafe = NULL;
+
+	if (g_getenv ("FWUPD_MODEM_MANAGER_VERBOSE") == NULL)
+		return;
+
+	msg_data = (const gchar *) g_bytes_get_data (msg, &msg_size);
+	if (msg_size > G_MAXINT)
+		return;
+
+	msg_strsafe = fu_common_strsafe (msg_data, msg_size);
+
+	g_debug ("%s: %.*s", action, (gint) msg_size, msg_strsafe);
+}
+
+static gboolean
+validate_program_action (XbNode *program, FuArchive *archive, GError **error)
+{
+	const gchar *filename_attr;
+	GBytes *file;
+	gsize file_size;
+	guint64 computed_num_partition_sectors;
+	guint64 num_partition_sectors;
+	guint64 sector_size_in_bytes;
+
+	filename_attr = xb_node_get_attr (program, "filename");
+	if (filename_attr == NULL) {
+		g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+			     "Missing 'filename' attribute in 'program' action");
+		return FALSE;
+	}
+
+	/* contents of the CAB file are flat, no subdirectories; look for the
+	 * exact filename */
+	file = fu_archive_lookup_by_fn (archive, filename_attr, error);
+	if (file == NULL)
+		return FALSE;
+
+	file_size = g_bytes_get_size (file);
+
+	num_partition_sectors = xb_node_get_attr_as_uint (program, "num_partition_sectors");
+	if (num_partition_sectors == G_MAXUINT64) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "Missing 'num_partition_sectors' attribute in 'program' action for filename '%s'",
+			     filename_attr);
+		return FALSE;
+	}
+	sector_size_in_bytes = xb_node_get_attr_as_uint (program, "SECTOR_SIZE_IN_BYTES");
+	if (sector_size_in_bytes == G_MAXUINT64) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "Missing 'SECTOR_SIZE_IN_BYTES' attribute in 'program' action for filename '%s'",
+			     filename_attr);
+		return FALSE;
+	}
+
+	computed_num_partition_sectors = file_size / sector_size_in_bytes;
+	if ((file_size % sector_size_in_bytes) != 0)
+		computed_num_partition_sectors++;
+
+	if (computed_num_partition_sectors != num_partition_sectors) {
+		g_set_error (error,
+			     G_IO_ERROR,
+			     G_IO_ERROR_FAILED,
+			     "Invalid 'num_partition_sectors' in 'program' action for filename '%s': "
+			     "expected %" G_GUINT64_FORMAT " instead of %" G_GUINT64_FORMAT " bytes",
+			     filename_attr, computed_num_partition_sectors, num_partition_sectors);
+		return FALSE;
+	}
+
+	xb_node_set_data (program, "fwupd:ProgramFile", file);
+	return TRUE;
+}
+
+gboolean
+fu_firehose_validate_rawprogram (GBytes *rawprogram, FuArchive *archive,
+                                 XbSilo **out_silo, GPtrArray **out_action_nodes, GError **error)
+{
+	g_autoptr(XbBuilder) builder = xb_builder_new ();
+	g_autoptr(XbBuilderSource) source = xb_builder_source_new ();
+	g_autoptr(XbSilo) silo = NULL;
+	g_autoptr(XbNode) data_node = NULL;
+	g_autoptr(GPtrArray) action_nodes = NULL;
+
+	if (!xb_builder_source_load_bytes (source, rawprogram, XB_BUILDER_SOURCE_FLAG_NONE, error))
+		return FALSE;
+	xb_builder_import_source (builder, source);
+	silo = xb_builder_compile (builder, XB_BUILDER_COMPILE_FLAG_NONE, NULL, error);
+	if (silo == NULL)
+		return FALSE;
+
+	data_node = xb_silo_get_root (silo);
+	action_nodes = xb_node_get_children (data_node);
+	if (action_nodes == NULL || action_nodes->len == 0) {
+		g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+			     "No actions given");
+		return FALSE;
+	}
+
+	for (guint i = 0; i < action_nodes->len; i++) {
+		XbNode *n = g_ptr_array_index (action_nodes, i);
+		if ((g_strcmp0 (xb_node_get_element (n), "program") == 0) &&
+		    !validate_program_action (n, archive, error)) {
+			return FALSE;
+		}
+	}
+
+	*out_silo = g_steal_pointer (&silo);
+	*out_action_nodes = g_steal_pointer (&action_nodes);
+	return TRUE;
+}
+
+gboolean
+fu_firehose_updater_open (FuFirehoseUpdater *self, GError **error)
+{
+	g_debug ("opening firehose port...");
+	self->io_channel = fu_io_channel_new_file (self->port, error);
+	return (self->io_channel != NULL);
+}
+
+gboolean
+fu_firehose_updater_close (FuFirehoseUpdater *self, GError **error)
+{
+	g_debug ("closing firehose port...");
+	if (!fu_io_channel_shutdown (self->io_channel, error))
+		return FALSE;
+	g_clear_object (&self->io_channel);
+	return TRUE;
+}
+
+static gboolean
+fu_firehose_updater_check_operation_result (XbNode *node, gboolean *out_rawmode)
+{
+	g_warn_if_fail (g_strcmp0 (xb_node_get_element (node), "response") == 0);
+	if (g_strcmp0 (xb_node_get_attr (node, "value"), "ACK") != 0)
+		return FALSE;
+	if (out_rawmode)
+		*out_rawmode = (g_strcmp0 (xb_node_get_attr (node, "rawmode"), "true") == 0);
+	return TRUE;
+}
+
+static gboolean
+fu_firehose_updater_process_response (GBytes *rsp_bytes, XbSilo **out_silo,
+				      XbNode **out_response_node, GError **error)
+{
+	g_autoptr(XbBuilder) builder = xb_builder_new ();
+	g_autoptr(XbBuilderSource) source = xb_builder_source_new ();
+	g_autoptr(XbSilo) silo = NULL;
+	g_autoptr(XbNode) data_node = NULL;
+	g_autoptr(GPtrArray) action_nodes = NULL;
+
+	if (!xb_builder_source_load_bytes (source, rsp_bytes, XB_BUILDER_SOURCE_FLAG_NONE, error))
+		return FALSE;
+	xb_builder_import_source (builder, source);
+	silo = xb_builder_compile (builder, XB_BUILDER_COMPILE_FLAG_NONE, NULL, error);
+	if (silo == NULL)
+		return FALSE;
+
+	data_node = xb_silo_get_root (silo);
+	if (data_node == NULL) {
+		g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+				     "Missing root data node");
+		return FALSE;
+	}
+
+	action_nodes = xb_node_get_children (data_node);
+	if (action_nodes != NULL) {
+		for (guint j = 0; j < action_nodes->len; j++) {
+			XbNode *node = g_ptr_array_index (action_nodes, j);
+
+			if (g_strcmp0 (xb_node_get_element (node), "response") == 0) {
+				if (out_silo)
+					*out_silo = g_steal_pointer (&silo);
+				if (out_response_node)
+					*out_response_node = g_object_ref (node);
+				return TRUE;
+			}
+
+			if (g_strcmp0 (xb_node_get_element (node), "log") == 0) {
+				const gchar *value_attr = xb_node_get_attr (node, "value");
+				if (value_attr)
+					g_debug ("device log: %s", value_attr);
+			}
+		}
+	}
+
+	if (out_silo != NULL)
+		*out_silo = NULL;
+	if (out_response_node != NULL)
+		*out_response_node = NULL;
+	return TRUE;
+}
+
+static gboolean
+fu_firehose_updater_send_and_receive (FuFirehoseUpdater *self, GByteArray *take_cmd_bytearray,
+				      XbSilo **out_silo, XbNode **out_response_node, GError **error)
+{
+	if (take_cmd_bytearray) {
+		const gchar *cmd_header = "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>\n<data>\n";
+		const gchar *cmd_trailer = "</data>";
+		g_autoptr(GBytes) cmd_bytes = NULL;
+
+		g_byte_array_prepend (take_cmd_bytearray, (const guint8 *)cmd_header, strlen (cmd_header));
+		g_byte_array_append (take_cmd_bytearray, (const guint8 *)cmd_trailer, strlen (cmd_trailer));
+		cmd_bytes = g_byte_array_free_to_bytes (take_cmd_bytearray);
+
+		fu_firehose_updater_log_message ("writing", cmd_bytes);
+		if (!fu_io_channel_write_bytes (self->io_channel, cmd_bytes, 1500,
+						FU_IO_CHANNEL_FLAG_FLUSH_INPUT, error)) {
+			g_prefix_error (error, "Failed to write command: ");
+			return FALSE;
+		}
+	}
+
+	for (guint i = 0; i < MAX_RECV_MESSAGES; i++) {
+		g_autoptr(GBytes) rsp_bytes = NULL;
+		g_autoptr(XbSilo) silo = NULL;
+		g_autoptr(XbNode) response_node = NULL;
+
+		rsp_bytes = fu_io_channel_read_bytes (self->io_channel, -1, DEFAULT_RECV_TIMEOUT_MS,
+						      FU_IO_CHANNEL_FLAG_SINGLE_SHOT, error);
+		if (rsp_bytes == NULL) {
+			g_prefix_error (error, "Failed to read XML message: ");
+			return FALSE;
+		}
+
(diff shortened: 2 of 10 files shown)


plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+
+/* Maximum amount of time to wait for a message from the module. */
+#define DEFAULT_RECV_TIMEOUT_MS 15000
+
+/* The first configure attempt sent to the module will include all the defaults
+ * listed below. If the module replies with a NAK specifying a different
+ * (shorter) max payload size to use, the second configure attempt will be done
+ * with that new suggested max payload size value. Only 2 configure attempts are
+ * therefore expected. */
+#define MAX_CONFIGURE_ATTEMPTS 2
```
hughsie, line 39: can you align the value like you did below pls

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+	gsize msg_size;
+
+	if (g_getenv ("FWUPD_MODEM_MANAGER_VERBOSE") == NULL)
+		return;
+
+	msg_data = (const gchar *) g_bytes_get_data (msg, &msg_size);
+	if (msg_size > G_MAXINT)
+		return;
+
+	g_debug ("%s:\n%.*s", action, (gint) msg_size, msg_data);
```
hughsie, line 80: are we sure msg_data is a string? Would fu_common_strsafe be a good idea?

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+	g_autoptr(XbNode) data_node = NULL;
+	g_autoptr(GPtrArray) action_nodes = NULL;
+
+	if (!xb_builder_source_load_bytes (source, rsp_bytes, XB_BUILDER_SOURCE_FLAG_NONE, error))
+		return FALSE;
+	xb_builder_import_source (builder, source);
+	silo = xb_builder_compile (builder, XB_BUILDER_COMPILE_FLAG_NONE, NULL, error);
+	if (silo == NULL) {
+		return FALSE;
+	}
```
hughsie, line 230: no braces pls

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+
+	data_node = xb_silo_get_root (silo);
+	if (data_node == NULL) {
+		g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+				     "Missing root data node");
+		return FALSE;
+	}
+
+	action_nodes = xb_node_get_children (data_node);
+	if (action_nodes) {
```
hughsie, line 240: != NULL pls

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+			return FALSE;
+		}
+	}
+
+	for (int i = 0; i < MAX_RECV_MESSAGES; i++) {
+		g_autoptr(GBytes) rsp_bytes = NULL;
+		g_autoptr(XbSilo) silo = NULL;
+		g_autoptr(XbNode) response_node = NULL;
+
+		rsp_bytes = fu_io_channel_read_bytes (self->io_channel, -1, DEFAULT_RECV_TIMEOUT_MS, FU_IO_CHANNEL_FLAG_SINGLE_SHOT, error);
```
hughsie, line 293: split onto multiple lines pls

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+		     "Didn't get any response in the last %d messages", MAX_RECV_MESSAGES);
+	return FALSE;
+}
+
+static gboolean
+fu_firehose_updater_initialize (FuFirehoseUpdater *self, GError **error)
+{
+	guint n_msg = 0;
+
+	for (int i = 0; i < MAX_RECV_MESSAGES; i++) {
```
hughsie, line 324: guint i?

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+					   " SkipStorageInit=\"%d\""
+					   "/>",
+					   CONFIGURE_MEMORY_NAME,
+					   CONFIGURE_VERBOSE,
+					   CONFIGURE_ALWAYS_VALIDATE,
+					   CONFIGURE_MAX_DIGEST_TABLE_SIZE_IN_BYTES,
+					   max_payload_size,
+					   CONFIGURE_ZLP_AWARE_HOST,
+					   CONFIGURE_SKIP_STORAGE_INIT);
+		cmd_bytearray = g_byte_array_new_take ((guint8 *)cmd_str, strlen (cmd_str));
```
hughsie, line 378: would g_string_free_to_bytes be less clunky to use here? Then you can also append each attr in the element,rather than a super long list of args.

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+	}
+
+	if (!fu_firehose_updater_check_operation_result (rsp_node, NULL)) {
+		g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+				     "Reset operation failed");
+		return FALSE;
+	}
+
+	/* read out all of the remaining messages. otherwise modem won't go into reset */
+	while (fu_firehose_updater_send_and_receive (self, NULL, &rsp_silo, &rsp_node, error));
```
hughsie, line 433: do we want to have some kind of limit? with by number or by time?

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+	g_autofree guint8 *last_block = NULL;
+
+	/* helper for the last block, which may need padding up to the next
+	 * sector size */
+	g_warn_if_fail (payload_size % sector_size == 0);
+	last_block = g_malloc0 (payload_size);
+
+	n_blocks = (guint) (program_file_size / payload_size);
+	if (program_file_size % payload_size != 0)
+		n_blocks++;
```
hughsie, line 455: isn't FuChunk going to be more helpful here?

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+				     "Failed to validate program file '%s' command: "
+				     "requested sector size bigger (%" G_GUINT64_FORMAT " bytes) "
+				     "than maximum payload size agreed with device (%u bytes)",
+				     program_filename, program_sector_size_in_bytes, max_payload_size);
+			return FALSE;
+		}
+	}
+
+	for (guint i = 0; i < action_nodes->len; i++) {
+		XbNode *node = g_ptr_array_index (action_nodes, i);
```
hughsie, line 547: this function is kinda huge, can we split it up for each node perhaps?

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+		const gchar *program_filename = NULL;
+		GBytes *program_file = NULL;
+		guint64 program_sector_size_in_bytes = 0;
+
+		action = xb_node_get_element (node);
+
+		if (g_strcmp0 (action, "program") != 0)
+			continue;
+
+		program_file = xb_node_get_data (node, "fwupd:ProgramFile");
```
hughsie, line 527: maybe check for NULL here?

hughsie: as in, return with an error rather than trying to go on...

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+		g_debug ("sending program file '%s' (%zu bytes)", program_filename, g_bytes_get_size (program_file));
+		if (!fu_firehose_updater_send_program_file (self, program_filename, program_file,
+							    payload_size, program_sector_size, error)) {
+			g_prefix_error (error, "Failed to send program file '%s': ",
+					program_filename);
+			return FALSE;
+		}
+
+		g_clear_object (&rsp_node);
+		g_clear_object (&rsp_silo);
```
hughsie, line 607: this is normally an indication we should split this up as well...

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+
+	return TRUE;
+}
+
+gboolean
+fu_firehose_updater_write (FuFirehoseUpdater *self, XbSilo *silo, GPtrArray *action_nodes, GError **error)
+{
+	guint max_payload_size;
+	gboolean result;
+	g_autoptr(GError) reset_error = NULL;
```
hughsie, line 642: I think error_local is the convention

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,699 @@
...
+}
+
+static void
+fu_firehose_updater_class_init (FuFirehoseUpdaterClass *klass)
+{
+	GObjectClass *object_class = G_OBJECT_CLASS (klass);
+	object_class->finalize = fu_firehose_updater_finalize;
+
+	signals [SIGNAL_WRITE_PROGRESS] =
+		g_signal_new ("write-progress",
```
hughsie, line 687: i think everywhere else it's called `percentage` no?

plugins/modem-manager/fu-firehose-updater.h
```diff
@@ -0,0 +1,37 @@
...
+
+/* helpers */
+
+gboolean		fu_firehose_validate_rawprogram	(GBytes		*rawprogram,
+							 FuArchive	*archive,
+							 XbSilo		**out_silo,
+							 GPtrArray	**out_action_nodes,
+							 GError		**error);
+
+gboolean fu_firehose_updater_reset (FuFirehoseUpdater *self, GError **error);
```
hughsie, line 35: alignment is odd

plugins/modem-manager/fu-firehose-updater.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (C) 2020 Aleksander Morgado <aleksander@aleksander.es>
+ * Copyright (C) 2021 Quectel Wireless Solutions Co., Ltd.
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#ifndef __FU_FIREHOSE_UPDATER_H
+#define __FU_FIREHOSE_UPDATER_H
```
hughsie, line 9: just use `#once` please

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -151,3 +151,39 @@ fu_mm_utils_get_port_info (const gchar	 *path,
 
 	return fu_mm_utils_get_udev_port_info (dev, out_device_bus, out_device_sysfs_path, out_port_usb_ifnum, error);
 }
+
+gboolean
+fu_mm_utils_find_device_file (const gchar	*device_sysfs_path,
+			   const gchar	*subsystem,
+			   gchar		**out_device_file,
+			   GError		**error)
```
hughsie, line 159: weird indenting

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -151,3 +151,39 @@ fu_mm_utils_get_port_info (const gchar	 *path,
...
+	GList *devices;
+	g_autoptr(GUdevClient) client = NULL;
+	g_autofree gchar *device_file = NULL;
+
+	client = g_udev_client_new (NULL);
+	devices = g_udev_client_query_by_subsystem (client, subsystem);
+	for (GList *l = devices; l != NULL; l = g_list_next (l)) {
+		if (g_str_has_prefix (g_udev_device_get_sysfs_path (G_UDEV_DEVICE (l->data)), device_sysfs_path)) {
+			device_file = g_strdup (g_udev_device_get_device_file (l->data));
+			if (device_file)
```
hughsie, line 170: != NULL

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -151,3 +151,39 @@ fu_mm_utils_get_port_info (const gchar	 *path,
...
 	return fu_mm_utils_get_udev_port_info (dev, out_device_bus, out_device_sysfs_path, out_port_usb_ifnum, error);
 }
+
+gboolean
+fu_mm_utils_find_device_file (const gchar	*device_sysfs_path,
+			   const gchar	*subsystem,
+			   gchar		**out_device_file,
+			   GError		**error)
+{
+	GList *devices;
```
hughsie, line 161: can we use g_autolist here?

plugins/modem-manager/fu-mm-utils.h
```diff
@@ -20,5 +20,9 @@ gboolean	fu_mm_utils_get_port_info	(const gchar	 *path,
 						 gchar		**device_sysfs_path,
 						 gint		 *port_usb_ifnum,
 						 GError		**error);
+gboolean	fu_mm_utils_find_device_file (const gchar	*device_sysfs_path,
```
hughsie, line 23: aligning is weird

superm1 (commented):
I thought this was going to first try to use `/sys/module/firmware_class/parameters/path`?  Does that not work?  Even if it's compiled into the kernel, I still see that parameter.  I really think that's the preferable way to go.

ivanmikh:
> I thought this was going to first try to use `/sys/module/firmware_class/parameters/path`? Does that not work? Even if it's compiled into the kernel, I still see that parameter. I really think that's the preferable way to go.

Yeah, I tried that approach first. Parameter is there, but modifying it does not seem to make any difference for the module. I tried different paths, but firmware module was not looking up that path.

superm1:
> > I thought this was going to first try to use `/sys/module/firmware_class/parameters/path`? Does that not work? Even if it's compiled into the kernel, I still see that parameter. I really think that's the preferable way to go.
> 
> Yeah, I tried that approach first. Parameter is there, but modifying it does not seem to make any difference for the module. I tried different paths, but firmware module was not looking up that path.

Can you check the kernel code and with maintainer for a bug with a hardcoded path or something? The kernel documentation seems pretty clear to me that path is first precedence if it's set. If it's not working I think this needs some printk's for figuring out why.

It really is tons cleaner than hardcoded path, so until I hear it's impossible to solve for xyz I think it should be preferred.

ivanmikh:
> > > I thought this was going to first try to use `/sys/module/firmware_class/parameters/path`? Does that not work? Even if it's compiled into the kernel, I still see that parameter. I really think that's the preferable way to go.
> > 
> > 
> > Yeah, I tried that approach first. Parameter is there, but modifying it does not seem to make any difference for the module. I tried different paths, but firmware module was not looking up that path.
> 
> Can you check the kernel code and with maintainer for a bug with a hardcoded path or something? The kernel documentation seems pretty clear to me that path is first precedence if it's set. If it's not working I think this needs some printk's for figuring out why.
> 
> It really is tons cleaner than hardcoded path, so until I hear it's impossible to solve for xyz I think it should be preferred.

Okay, it was my my fault, I was supposed to remove a newline character from the path (e.g. use echo -n). Thanks for pushing it.

So, which path I should use to copy firmware to? Is there any fwupd runtime directory I can use?

hughsie:
> Is there any fwupd runtime directory I can use?

Does FU_PATH_KIND_CACHEDIR_PKG work?

superm1:
> Okay, it was my my fault, I was supposed to remove a newline character from the path (e.g. use echo -n). Thanks for pushing it.

Whew, glad to hear that.

> So, which path I should use to copy firmware to? Is there any fwupd runtime directory I can use?

`FU_PATH_KIND_LOCALSTATEDIR_PKG`

superm1:
Seems @hughsie and I stepped on each other, I think it depends on how ephemeral the files will be.  Is it really just for flashing once, or will it get used again from time to time and not change?  I'd say cache if it's just quickly used, and state if it will be used again some time.

superm1:
Description of all of them is available here BTW: https://fwupd.github.io/libfwupdplugin/enum.PathKind.html

ivanmikh:
> Seems @hughsie and I stepped on each other, I think it depends on how ephemeral the files will be. Is it really just for flashing once, or will it get used again from time to time and not change? I'd say cache if it's just quickly used, and state if it will be used again some time.

It is only used once during fw upgrade/downgrade process. It comes as a part of fw cab archive, so it might change with updates. I guess cache is more appropriate here. I will try FU_PATH_KIND_CACHEDIR_PKG and update this merge request.

superm1:
I say you should probably make a subdirectory for this plugin in that directory. That way if we ever have two plugins using the directory they don't both fight for a filename in there.

plugins/modem-manager/fu-mm-utils.c
```diff
@@ -151,3 +151,40 @@ fu_mm_utils_get_port_info (const gchar	 *path,
 
 	return fu_mm_utils_get_udev_port_info (dev, out_device_bus, out_device_sysfs_path, out_port_usb_ifnum, error);
 }
+
+gboolean
+fu_mm_utils_find_device_file (const gchar	*device_sysfs_path,
```
superm1, line 156: this function can it be moved into fu-udev-device?

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,748 @@
...
+}
+
+gboolean
+fu_firehose_updater_open (FuFirehoseUpdater *self, GError **error)
+{
+	g_debug ("opening firehose port...");
+	self->io_channel = fu_io_channel_new_file (self->port, error);
+	if (self->io_channel == NULL)
+		return FALSE;
+	return TRUE;
```
superm1, line 193: maybe just `return (self->io_channel != NULL)`

plugins/modem-manager/fu-firehose-updater.c
```diff
@@ -0,0 +1,748 @@
...
+	}
+
+	if (!fu_firehose_updater_check_operation_result (rsp_node, NULL)) {
+		g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+				     "Reset operation failed");
+		return FALSE;
+	}
+
+	/* read out all of the remaining messages. otherwise modem won't go into reset */
+	while (--recv_cnt && fu_firehose_updater_send_and_receive (self, NULL, &rsp_silo, &rsp_node, error));
```
superm1, line 438: if one of these send/receives fails you need to return false.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -679,6 +864,178 @@ fu_mm_device_write_firmware_qmi_pdc (FuDevice *device, GBytes *fw, GArray **acti
...
+		return FALSE;
+	}
+
+	/* write mm_fw_dir to a search path parameter of the firmware_loader module */
+	sys_fw_search_path_prm = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	/* g_file_set_contents will try to create backup files in sysfs, so use fopen here */
+	fd = fopen (sys_fw_search_path_prm, "w");
+	if (fd == NULL) {
+		g_set_error (error,
+					FWUPD_ERROR,
```
superm1, line 939: whitespace seems off here

plugins/modem-manager/fu-mm-device.c
```diff
@@ -679,6 +864,178 @@ fu_mm_device_write_firmware_qmi_pdc (FuDevice *device, GBytes *fw, GArray **acti
...
+			     FWUPD_ERROR_INTERNAL,
+			     "Failed to create '%s': %s",
+			     mm_fw_dir, g_strerror (errno));
+		return FALSE;
+	}
+
+	/* write mm_fw_dir to a search path parameter of the firmware_loader module */
+	sys_fw_search_path_prm = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	/* g_file_set_contents will try to create backup files in sysfs, so use fopen here */
+	fd = fopen (sys_fw_search_path_prm, "w");
```
superm1, line 936: I think you want to read the contents first don't you?  So it can be restored later if it's set to something else

plugins/modem-manager/fu-mm-utils.c
ivanmikh: I can move it, but is it okay if it does not use FuUdevDevice type? I can only use FuDevice as a first parameter instead of passing physical id

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: it should fail actually. we are trying to read out all of the remaining messages in the modem's buffer, so we are reading until the send/receive function returns `FALSE`. that means modem is now ready to go into reset and function has finished successfully.
recv_cnt there is only to protect against infinite loop. that can only happen if something goes wrong with the modem. usually there are only 2 or 3 messages left.

i will add g_warn_if_fail (recv_cnt > 0) though

superm1: If you're waiting and expecting until it returns false, you need a local error variable. Not the one passed.

ivanmikh: makes sense, will fix

plugins/modem-manager/fu-mm-utils.c
ivanmikh: if I am adding new functions to libfwupdplugin, should I use `Since: 1.6.3`?

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: it is only used to print xml string, but I have added fu_common_strsafe just to be safe

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: fixed

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: fixed

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: fixed

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: missed it, fixed

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: switched to GString

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: added counter

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: did not know about it, rewrote the function with FuChunk

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: refactored into several function, should look good now

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: fixed

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: thanks, fixed

plugins/modem-manager/fu-firehose-updater.h
ivanmikh: fixed

plugins/modem-manager/fu-firehose-updater.h
ivanmikh: fixed

plugins/modem-manager/fu-mm-utils.c
ivanmikh: tried to use g_autolist, but got seg_fault error. not sure why exactly, switched back to GList for now

plugins/modem-manager/fu-mm-utils.c
ivanmikh: left it in fu-mm-utils.c for now

plugins/modem-manager/fu-firehose-updater.c
ivanmikh: replaced with NULL, since it does not matter if that function returns error or not

plugins/modem-manager/fu-mm-device.c
ivanmikh: done. also added two new functions to fu-common to read/write `FU_PATH_KIND_FIRMWARE_SEARCH`

libfwupdplugin/fu-common.c
```diff
@@ -3384,3 +3384,90 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+	if (fd == NULL) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_PERMISSION_DENIED,
+			     "Failed to open %s: %s",
+			     sys_fw_search_path_prm,
+			     g_strerror (errno));
+		return FALSE;
+	}
+	res = fwrite (path, sizeof (gchar), strlen (path), fd);
```
aleksander0m, line 3460: g_file_set_contents() maybe, instead of all this custom code?

plugins/modem-manager/fu-mm-device.c
```diff
@@ -931,6 +1116,170 @@ fu_mm_device_write_firmware_mbim_qdu (FuDevice *device, GBytes *fw, GError **err
...
+static gboolean
+fu_mm_setup_firmware_dir (FuMmDevice *self, GError **error)
+{
+	g_autofree gchar *cachedir = NULL;
+	g_autofree gchar *mm_fw_dir = NULL;
+
+	/* create a directory to store firmware files for modem-manager plugin */
+	cachedir = fu_common_get_path (FU_PATH_KIND_CACHEDIR_PKG);
+	mm_fw_dir = g_build_filename (cachedir, "modem-manager", "firmware", NULL);
+	if (g_mkdir_with_parents (mm_fw_dir, 0755) == -1) {
```
aleksander0m, line 1174: Wondering, `0700` could be better here?

aleksander0m (changes requested):

aleksander0m:
Given that I have the module, I would definitely like to test this myself. @ivanmikh can I use the same original CAB files I prepared a while ago, or should I use some new ones generated by you? If so, what are the changes done in the CAB files?

ivanmikh:
> Given that I have the module, I would definitely like to test this myself. @ivanmikh can I use the same original CAB files I prepared a while ago, or should I use some new ones generated by you? If so, what are the changes done in the CAB files?

Yes, there are some changes with CAB files. I will forward you an email with the details

libfwupdplugin/fu-common.c
```diff
@@ -3384,3 +3384,90 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
 {
 	xb_builder_node_insert_text (bn, key, value ? "true" : "false", NULL);
 }
+
+/**
+ * fu_common_read_firmware_search_path:
+ * @error: (nullable): optional return location for an error
+ *
+ * Reads the FU_PATH_KIND_FIRMWARE_SEARCH and
+ * returns it's contents
```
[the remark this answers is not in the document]
ivanmikh: fixed, thanks

libfwupdplugin/fu-common.c
```diff
@@ -3384,3 +3384,90 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_PERMISSION_DENIED,
+			     "Failed to open %s: %s",
+			     sys_fw_search_path_prm,
+			     g_strerror (errno));
+		return NULL;
+	}
+	res = fread (path, sizeof (gchar), PATH_MAX, fd);
+
+	g_debug ("%s(): read %lu bytes: %s", __func__, res, path);
```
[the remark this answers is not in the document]
ivanmikh: makes sense, change to `"read firmware search path: %s", path`

libfwupdplugin/fu-common.c
```diff
@@ -3384,3 +3384,90 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+{
+	FILE *fd;
+	gsize res = 0;
+	g_autofree gchar *sys_fw_search_path_prm = NULL;
+	g_autofree gchar *path = g_malloc0 (PATH_MAX);
+
+	/* write mm_fw_dir to a search path parameter of the firmware_loader module */
+	sys_fw_search_path_prm = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	/* g_file_set_contents will try to create backup files in sysfs, so use fopen here */
+	fd = fopen (sys_fw_search_path_prm, "r");
```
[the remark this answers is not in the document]
ivanmikh: oh yeah, definitely better, thanks

libfwupdplugin/fu-common.c
```diff
@@ -3384,3 +3384,90 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+	}
+	res = fwrite (path, sizeof (gchar), strlen (path), fd);
+
+	fclose (fd);
+
+	if (res != strlen (path)) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_WRITE,
+			     "Failed to write firmware search path");
```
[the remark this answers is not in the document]
ivanmikh: added

libfwupdplugin/fu-common.c
ivanmikh: `g_file_set_contents()` will try to create backup files during writing, which is not allowed for sysfs. I think it should be possible with `g_file_set_contents_full` and `G_FILE_SET_CONTENTS_NONE`, but its only available in the latest version of GLib.

plugins/modem-manager/fu-mm-device.c
ivanmikh: probably more secure, yeah. changed to `0700`

superm1:
> -    fu_common_read_firmware_search_path;
> -    fu_common_write_firmware_search_path;

This CI failure means needing to add to `libfwupdplugin.map` in the right place

hughsie:
>       fu_common_read_firmware_search_path;
>        fu_common_write_firmware_search_path;

Wouldn't that be more standard to be fu_common_get_firmware_search_path and fu_common_set_firmware_search_path?

libfwupdplugin/fu-common.c
aleksander0m: Maybe there's a different preferred way to do this in fwupd?

ivanmikh:
Changed function names to set/get and moved them to the right location in the `libfwupdplugin.map`

libfwupdplugin/fu-common.c
ivanmikh: i could not find anything similar in the codebase. maybe @hughsie and @superm1 can suggest something

superm1: My opinion, how about check the glib version with a macro and use `g_file_set_contents_full` if 2.66 or later.  And then use this code if older version.  Eventually some day we bump up minimum glib version requirement and can drop the dead code.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -931,6 +1116,170 @@ fu_mm_device_write_firmware_mbim_qdu (FuDevice *device, GBytes *fw, GError **err
...
+	archive = fu_archive_new (fw, FU_ARCHIVE_FLAG_IGNORE_PATH, error);
+	if (archive == NULL)
+		return FALSE;
+
+	old_fw_path = fu_common_get_firmware_search_path (NULL);
+	/* firehose modems that switch to the EDL mode over QCDM port require
+	 * firehose binary to be present in the firmware-loader search path. */
+	if (!fu_mm_setup_firmware_dir (self, error))
+		return FALSE;
+	if (!fu_mm_copy_firehose_prog(self, archive, error))
```
superm1, line 1237: nit: missing space in `g(`

plugins/modem-manager/fu-mm-device.c
```diff
@@ -931,6 +1116,170 @@ fu_mm_device_write_firmware_mbim_qdu (FuDevice *device, GBytes *fw, GError **err
...
+		return FALSE;
```
superm1, line 1238: I think you need to restore the old path in case of failure here.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -931,6 +1116,170 @@ fu_mm_device_write_firmware_mbim_qdu (FuDevice *device, GBytes *fw, GError **err
...
+
+	/* flag as restart because the QCDM and MHI/BHI operations switch the
+	 * device into FP execution environment */
+	fu_device_set_status (FU_DEVICE (self), FWUPD_STATUS_DEVICE_RESTART);
+	/* switch to embedded downloader (EDL) execution environment */
+	res = fu_mm_device_qcdm_switch_to_edl (FU_DEVICE(self), error);
+	/* restore the contents of firmware search path, if any */
+	if (old_fw_path != NULL && strlen (old_fw_path) > 0 &&
```
superm1, line 1246: even if the old path was length zero don't you still need to restore it?  That seems like the common scenario, no path set so it's zero length and you restore it back to zero length.

libfwupdplugin/fu-common.c
ivanmikh: okay, I have added `g_file_set_contents_full` with `GLIB_CHECK_VERSION(2,66,0)`

plugins/modem-manager/fu-mm-device.c
ivanmikh: fixed, thanks

plugins/modem-manager/fu-mm-device.c
ivanmikh: you are right, fixed it

plugins/modem-manager/fu-mm-device.c
ivanmikh: i agree, it is gonna be more logical here. removed `if`

libfwupdplugin/fu-common.c
```diff
@@ -3378,3 +3384,102 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+ **/
+gchar *
+fu_common_get_firmware_search_path (GError **error)
+{
+	gsize sz = 0;
+	g_autofree gchar *sys_fw_search_path = NULL;
+	g_autofree gchar *contents = NULL;
+
+	sys_fw_search_path = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	g_file_get_contents(sys_fw_search_path, &contents, &sz, error);
```
hughsie, line 3407: if this returns FALSE, return FALSE. Also space before the bracket.

libfwupdplugin/fu-common.c
```diff
@@ -3378,3 +3384,102 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+ * Returns: %TRUE if successful
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_common_set_firmware_search_path (const gchar *path, GError **error)
+{
+#if GLIB_CHECK_VERSION(2,66,0)
+	g_autofree gchar *sys_fw_search_path_prm = NULL;
+	gsize len = strlen (path);
```
hughsie, line 3434: you need to do the strlen after the g_return_val_if_fail

libfwupdplugin/fu-common.c
```diff
@@ -3378,3 +3384,102 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+
+	g_return_val_if_fail (path != NULL, FALSE);
+
+	if (len == 0)
+		len = 1;
```
hughsie, line 3439: eigh?

libfwupdplugin/fu-common.c
```diff
@@ -3378,3 +3384,102 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+
+	sys_fw_search_path_prm = fu_common_get_path (FU_PATH_KIND_FIRMWARE_SEARCH);
+	if (!g_file_set_contents_full (sys_fw_search_path_prm, path, len,
+				G_FILE_SET_CONTENTS_NONE, 0644, error))
+		return FALSE;
+
+	return TRUE;
```
hughsie, line 3446: maybe just return g_file_set_contents_full (...

libfwupdplugin/fu-common.c
ivanmikh: If we pass an empty string to the function, meaning that we want to erase the contents of the firmware search path, strlen will return 0 and `g_file_set_contents_full` will not change the contents of the file. Hence len is changed to 1 so 0 is written to the file.
I couldn't think of any other way to overcome this

ivanmikh: i guess i can add the len parameter to the function arguments and do this checking in the calling function

ivanmikh: or maybe add another function to clear the firmware search path. please suggest what would be a better solution here

libfwupdplugin/fu-common.c
ivanmikh: fixed

libfwupdplugin/fu-common.c
ivanmikh: fixed

libfwupdplugin/fu-common.c
ivanmikh: fixed

libfwupdplugin/fu-common.c
hughsie: Yes, I think another function to unset would be much better -- e.g. fu_common_reset_firmware_search_path which also means we can use it from a FuDeviceLocker and do the right thing on the error path in all cases.

ivanmikh: So I added `fu_common_reset_firmware_search_path` to FuCommon and used FuDeviceLocker for setting up and restoring firmware path during firehose update

libfwupdplugin/fu-common.c
```diff
@@ -3384,3 +3384,119 @@ fu_xmlb_builder_insert_kb (XbBuilderNode *bn, const gchar *key, gboolean value)
...
+ * @error: (nullable): optional return location for an error
+ *
+ * Writes to the FU_PATH_KIND_FIRMWARE_SEARCH
+ *
+ * Returns: %TRUE if successful
+ *
+ * Since: 1.6.2
+ **/
+gboolean
+fu_common_set_firmware_search_path (const gchar *path, gsize path_sz, GError **error)
```
hughsie, line 3432: Can we just document the "path is NUL-terminated, and then we can drop the path_sz argument. We can so the strlen in this function, and drop it from the callers. With that change I think the PR looks good to me and is probably ready to be merged. Good work!

aleksander0m: I still want to test this work myself, but I don't have much free time at the moment, so my test should definitely not block merging all this if it's already reviewed :D

ivanmikh: @hughsie added the comment and dropped the path_sz argument. Thank you for all of the valuable feedback!

@aleksander0m thanks! it seems there is a change request still pending
image

aleksander0m (approved):

libfwupdplugin/fu-common.c
aleksander0m: Just approved the changes according to what you did w.r.t. my last review

hughsie:
Okay, lets press the button. Any follow up fixes can be done as a new PR. Thanks all!

libfwupdplugin/fu-common.c
ivanmikh: thanks!

PR #3425, opened 2021-06-25
kinetic-dp: New plugin to update FW for Kinetic's DisplayPort converter chips

jeffrey-lin-kinetic:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

New plugin checklist:
- [ ] Fill out README.md with any custom quirks and flags
- [ ] Fill out README.md with the vendor ID security value
- [ ] Implement FuFirmware->write() and include at least one fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass


superm1:
Can you please run `./contrib/setup` to install the pre-commit hooks locally?  Then run `git commit --amend` and it will show you all of the things that were failing in pre-commit CI.

superm1:
Fedora CI failed, you need to add the file into `.spec.in`:

```
RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/lib/debug/usr/lib64/fwupd-plugins-3/libfu_plugin_kinetic_dp.so-1.6.2-0.1alpha.fc34.x86_64.debug
   /usr/lib64/fwupd-plugins-3/libfu_plugin_kinetic_dp.so
```

plugins/kinetic-dp/fu-kinetic-dp-aux-dpcd.c
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (C) 2021 Jeffrey Lin <jlin@kinet-ic.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "fu-kinetic-dp-aux-dpcd.h"
+
+#include "fwupd-error.h"
```
superm1, line 9: this should be `<fwupdplugin.h>` IIRC

plugins/kinetic-dp/fu-kinetic-dp-aux-dpcd.c
```diff
@@ -0,0 +1,76 @@
...
+				 GError **error)
+{
+	if (buf_size < DPCD_SIZE_IEEE_OUI)
+	{
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Buffer size [%u] is too small to read IEEE OUI",
+			     buf_size);
+
```
superm1, line 24: extra unneeded whitespace

plugins/kinetic-dp/fu-kinetic-dp-aux-dpcd.c
```diff
@@ -0,0 +1,76 @@
...
+					   GError **error)
+{
+	if (buf_size < DPCD_SIZE_BRANCH_DEV_ID_STR)
+	{
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INTERNAL,
+			     "Buffer size [%u] is too small to read branch ID string",
+			     buf_size);
+
```
superm1, line 62: extra unneeded whitespace

superm1:
I fixed up the nits I mentioned, please run `./contrib/setup` locally on your end though so that future work you can pass pre-commit CI immediately.

plugins/kinetic-dp/fu-plugin-kinetic-dp.c
```diff
@@ -0,0 +1,156 @@
...
+#include "fu-kinetic-dp-device.h"
+#include "fu-kinetic-dp-firmware.h"
+
+#define FU_KINETIC_DP_DRM_REPLUG_DELAY	5	/* sec */
+
+struct FuPluginData {
+	GPtrArray	*devices;
+	guint		drm_changed_id;
+};
+
```
superm1, line 22: so the previous codebase this was plugin was partially based upon had a safety check for `amdgpu`.  Does that apply here?  Is it possible that one of these DP converter chips is connected to an `amdgpu` with an older kernel?  Or is it only in fixed hardware designs (IE soldered down to the board and running in stuff like ChromeOS)?  

@hughsie since we now _potentially_ have 3 plugins potentially accessing DP aux, I think it's worth moving the safety check "into the daemon" for that somewhere. Thoughts?

plugins/kinetic-dp/fu-kinetic-dp-firmware.c
```diff
@@ -0,0 +1,259 @@
...
+	isp_drv_payload = g_bytes_new_from_bytes (fw, HEADER_LEN_ISP_DRV_SIZE, isp_drv_payload_size);
+	isp_drv_img = fu_firmware_new_from_bytes (isp_drv_payload);
+	fu_firmware_set_idx (isp_drv_img, FU_KT_FW_IMG_IDX_ISP_DRV);
+
+	fu_firmware_add_image (firmware, isp_drv_img);
+
+	/* Add App FW as a new image into firmware */
+	app_fw_payload = g_bytes_new_from_bytes (fw, HEADER_LEN_ISP_DRV_SIZE + isp_drv_payload_size, app_fw_payload_size);
+	buf = g_bytes_get_data (app_fw_payload, &bufsz);
+	if (!kt_dp_get_chip_id_from_fw_buf (buf, bufsz, &chip_id, &fw_bin_flag)) {
```
superm1, line 220: you need to send error in this function and set it on the FALSE scenario

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,853 @@
...
+}
+
+static gboolean
+fu_kinetic_dp_secure_aux_isp_send_isp_drv (FuKineticDpConnection *connection,
+					   gboolean is_app_mode,
+					   const guint8 *isp_drv_data,
+					   guint32 isp_drv_len,
+					   GError **error)
+{
+	g_message ("Sending ISP driver payload... started");
```
superm1, line 436: please don't use g_message for debugging code.  I guess this was during bringup, at this point since you have nice error handling I would guess that most of these `g_message` can either go away or downgrade to `g_debug` (up to you on what's actually still useful)

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,853 @@
...
+		if (flash_size < 2048)    // One bank size in Jaguar is 1024KB
+			g_message ("Flash Size: %d KB, Dual Bank is not supported!", flash_size);
+		else
+			g_message ("Flash Size: %d KB", flash_size);
+	} else {
+		if (flash_id)
+			g_message ("(SPI flash not supported)");
+		else
+			g_message ("(SPI flash not connected)");
+	}
```
superm1, line 468: these cases should be returning `FALSE` and setting `error`

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,853 @@
...
+		if (isp_procd_size < isp_total_data_size) {
+			isp_procd_size += progress_inc;
+			isp_payload_procd_size += progress_inc;
+		}
+
+		// Wait 50ms
+		g_usleep (50 * 1000);
+	}
+
+	g_prefix_error (error, "Installing images... timed-out!");
```
superm1, line 635: don't end this line with` !`, it should be: `timed-out: ` so the prefix looks correct on the error.

superm1: And actually this should probably be a `g_set_error`, error never got set.

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,853 @@
...
+		}
+
+		if (status != (cmd_id | DPCD_KT_CONFIRMATION_BIT)) {	// Target response
+			if (status == cmd_id) {				// Confirmation bit if cleared
+				isp_payload_procd_size += (isp_total_data_size - isp_procd_size);
+				g_message ("Programming F/W payload... done!");
+
+				return TRUE;
+			} else {
+				g_prefix_error (error, "Installing images... failed!");
```
superm1, line 620: this should probably be a `g_set_error`, error never got set.

superm1 (changes requested):
some general feedback:
* There are a lot of `g_message` commands, none of these should stay.  They should either be removed (now that you've finished bringup) or they should downgrade to `g_debug`.  Up to you on which one, but please don't leave them.
* There are various TODOs left, do you want any of these to block merging?  Or they're just nice to do?
* Please update comments style to `/* */` instead of `//`
* All of your instances of `g_prefix_error` that end in `!` should correct to end with `: ` so that they prefix properly

plugins/kinetic-dp/fu-plugin-kinetic-dp.c
jeffrey-lin-kinetic: Currently we can't confirm  what kind of GPU would be used on our customer's product.
But if this is a known common issue related to kernel version, I think we should also include this safety check as well.

plugins/kinetic-dp/fu-kinetic-dp-firmware.c
jeffrey-lin-kinetic: > some general feedback:
> 
> * There are a lot of `g_message` commands, none of these should stay.  They should either be removed (now that you've finished bringup) or they should downgrade to `g_debug`.  Up to you on which one, but please don't leave them.
> * There are various TODOs left, do you want any of these to block merging?  Or they're just nice to do?
> * Please update comments style to `/* */` instead of `//`
> * All of your instances of `g_prefix_error` that end in `!` should correct to end with `: ` so that they prefix properly

I've fixed most of the feedback in the new commit.

And I don't want the TODO list to block merging.
I think I would have some issues to implement after I can do online test.
As you said, some of the TODOs are just nice to do.
I'll complete them if the implementation details are determined.

Thanks.

jeffrey-lin-kinetic:
> some general feedback:
> 
> * There are a lot of `g_message` commands, none of these should stay.  They should either be removed (now that you've finished bringup) or they should downgrade to `g_debug`.  Up to you on which one, but please don't leave them.
> * There are various TODOs left, do you want any of these to block merging?  Or they're just nice to do?
> * Please update comments style to `/* */` instead of `//`
> * All of your instances of `g_prefix_error` that end in `!` should correct to end with `: ` so that they prefix properly

I've fixed most of your feedback in the new commit.

And I don't want the TODO list to block merging.
I think I would have some issues to implement after I can do online test.
As you said, some of the TODOs are just nice to do.
I'll complete them if the implementation details are determined.

Thanks.

superm1:
Looks like freebsd CI needs to turn this off by default:
> plugins/kinetic-dp/meson.build:3:2: ERROR: Problem encountered: gudev is required for plugin_kinetic_dp

plugins/kinetic-dp/fu-kinetic-dp-firmware.c
```diff
@@ -218,7 +219,10 @@ fu_kinetic_dp_firmware_parse (FuFirmware *firmware,
 	app_fw_payload = g_bytes_new_from_bytes (fw, HEADER_LEN_ISP_DRV_SIZE + isp_drv_payload_size, app_fw_payload_size);
 	buf = g_bytes_get_data (app_fw_payload, &bufsz);
 	if (!kt_dp_get_chip_id_from_fw_buf (buf, bufsz, &chip_id, &fw_bin_flag)) {
-		g_prefix_error (error, "No valid chip ID is found in the firmware: ");
+		g_set_error_literal(error,
```
superm1, line 222: missing a space before `(`

superm1 (dismissed):
just a whitespace nit, and fix the freebsd CI (turn off this plugin by default for freebsd) and then we're good.

plugins/kinetic-dp/fu-plugin-kinetic-dp.c
superm1: OK, we need to pull this into the daemon then separately from this PR and modify to make sure all plugins using DP aux do the safety check from the daemon.

plugins/kinetic-dp/fu-kinetic-dp-aux-isp.c
```diff
@@ -0,0 +1,305 @@
...
+	{KT_CHIP_JAGUAR_5000,	KT_FW_STATE_RUN_IROM,	{'5', '0', '1', '0', 'I', 'R'},	6},
+	{KT_CHIP_JAGUAR_5000,	KT_FW_STATE_RUN_APP,	{'K', 'T', '5', '0', 'X', '0'},	6},
+	/* Mustang MCDP52x0 */
+	{KT_CHIP_MUSTANG_5200,	KT_FW_STATE_RUN_IROM,	{'5', '2', '1', '0', 'I', 'R'},	6},
+	{KT_CHIP_MUSTANG_5200,	KT_FW_STATE_RUN_APP,	{'K', 'T', '5', '2', 'X', '0'},	6},
+};
+
+static KtDpDevInfo dp_dev_infos[MAX_DEV_NUM];
+static KtChipId dp_root_dev_chip_id = KT_CHIP_NONE;
+static KtFwRunState dp_root_dev_state = KT_FW_STATE_RUN_NONE;
```
hughsie, line 33: What's all this global device state? This looks fishy to me.

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,873 @@
...
+static guint32 isp_payload_procd_size;
+static guint32 isp_procd_size;
+static guint32 isp_total_data_size;
+
+static guint16 read_flash_prog_time;
+static guint16 flash_id;
+static guint16 flash_size;
+
+static gboolean is_isp_secure_auth_mode;
+
```
hughsie, line 53: this is all fishy too; it needs to go into the priv data struct

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,873 @@
...
+static guint16
+_gen_crc16 (guint16 accum, guint8 data_in)
```
hughsie, line 55: remove this and use fu_common_crc16 please

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,873 @@
...
+	}
+
+	if (!fu_kinetic_dp_secure_aux_isp_read_dpcd_reply_data_reg (connection, reply_data, sizeof (reply_data), &read_len, error)) {
+		g_prefix_error (error, "Failed to read flash ID and size: ");
+		return FALSE;
+	}
+
+	flash_id = (guint16)reply_data[0] << 8 | reply_data[1];
+	flash_size = (guint16)reply_data[2] << 8 | reply_data[3];
+	read_flash_prog_time = (guint16)reply_data[4] << 8 | reply_data[5];
```
hughsie, line 435: Can you use fu_common_read_uint16_safe here please.

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,873 @@
...
+	guint8 pl_size_data[12] = {0};
+
+	g_debug ("Entering F/W loading mode...");
+
+	/* Send payload size to DPCD_MCA_REPLY_DATA_REG */
+	*(guint32 *)pl_size_data = fu_kinetic_dp_firmware_get_esm_payload_size (firmware);
+	*(guint32 *)&pl_size_data[4] = fu_kinetic_dp_firmware_get_arm_app_code_size (firmware);
+	*(guint16 *)&pl_size_data[8] = (guint16)fu_kinetic_dp_firmware_get_app_init_data_size (firmware);
+	*(guint16 *)&pl_size_data[10] = (guint16)fu_kinetic_dp_firmware_get_cmdb_block_size (firmware) |
+					(fu_kinetic_dp_firmware_get_is_fw_esm_xip_enabled (firmware) << 15);
```
hughsie, line 509: Please use fu_common_read_uint32_safe() -- that's not even remotely safe on non-LE systems

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,873 @@
...
+	payload_data = g_bytes_get_data (app, &payload_len);
+	if (!fu_kinetic_dp_secure_aux_isp_send_fw_payload (connection, firmware_self, payload_data, payload_len, error))
+		goto SECURE_AUX_ISP_END;
+
+	/* Install FW images */
+	ret = fu_kinetic_dp_secure_aux_isp_install_fw_images (connection, error);
+
+SECURE_AUX_ISP_END:
+	/* Send reset command */
+	fu_kinetic_dp_secure_aux_isp_send_reset_command (connection, error);
```
hughsie, line 847: you can't re-use the error here if it's set. Maybe just a g_warning and error_local?

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,873 @@
...
+
+	cmd_id = (guint8)target_port;
+	if (!fu_kinetic_dp_connection_write (connection, DPCD_ADDR_FLOAT_PARAM_REG, &cmd_id, sizeof (cmd_id), error))
+		return FALSE;
+
+	ret = fu_kinetic_dp_secure_aux_isp_send_kt_prop_cmd (connection, KT_DPCD_CMD_ENABLE_AUX_FORWARD, 1000, 20, &status, error);
+
+	/* Clear CMD_STATUS_REG */
+	cmd_id = KT_DPCD_CMD_STS_NONE;
+	fu_kinetic_dp_connection_write (connection, DPCD_ADDR_FLOAT_CMD_STATUS_REG, &cmd_id, sizeof (cmd_id), error);
```
hughsie, line 727: also reusing the GError here

hughsie (changes requested):
We need to move out all that static stuff.I Ive added a few more "have-to-fixes" but I agree we can do the TODOs later.

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
jeffrey-lin-kinetic: iIn our implementation, I think CRC16 is calculated with CRC-16-CCITT polynomial.
Is fu_common_crc16() based on CRC-16-IBM polynomial?
I'm not very sure if these two are different CRC-16 algorithms.

hughsie: This should help you https://github.com/fwupd/fwupd/pull/3452 -- feel free to cherry pick that commit into your branch.

jeffrey-lin-kinetic: These days I've studied for CRC-16 things, it's more complicated than what I could imagine.
I tried to find some reference codes to compare the outputs among Kinetic's CRC-16 function, fu_common_crc16(), and other reference implementations(CRC-16, CRC16-CCITT).
I ported the CCITT function from a library called LibCRC.
But I found that the output of fu_common_crc16(init value: 0xFFFF, polynomial: 0x12) is not the same as the output of libcrc.
My references are as below links:
http://srecord.sourceforge.net/crc16-ccitt.html
https://reveng.sourceforge.io/crc-catalogue/16.htm
[LibCRC]
https://www.lammertbies.nl/comm/info/crc-calculation
https://www.libcrc.org/

According to one of above articles, there're many incorrect CRC-16 implementations on the web.
And there're so many variants(parameters) to implement.
May I know how do you implement fu_common_crc16() function and what's your reference?

Back to the PR, I think our CRC-16 implementation just works as a checksum mechanism while transmitting the payload.
I don't think it's a standard CRC16-CCITT implementation(I don't know the history either...).
I can't just replace it with fu_common_crc16() because of the backward-compatible issue in Windows tools released to our customers.
So my first thought is to rename our CRC16 implementation as a proprietary CRC-16 "static" function.
Would it be acceptable?

Thanks a lot.

hughsie: > So my first thought is to rename our CRC16 implementation as a proprietary CRC-16 "static" function

Yes, that's probably the best plan. If you can add some documentation explaining that I'd appreciate it.

superm1:
I rebased and squashed it on master to fix the conflicts.

plugins/kinetic-dp/fu-kinetic-dp-aux-isp.c
```diff
@@ -0,0 +1,300 @@
...
+								"Unknown"};
+
+const gchar *
+fu_kinetic_dp_aux_isp_get_chip_id_str(KtChipId chip_id)
+{
+	if (chip_id == KT_CHIP_JAGUAR_5000)
+		return "KTM50X0";
+	if (chip_id == KT_CHIP_MUSTANG_5200)
+		return "KTM52X0";
+	return "";
```
superm1, line 52: Maybe rather than "", `unknown` might be better?

plugins/kinetic-dp/fu-kinetic-dp-aux-isp.c
```diff
@@ -0,0 +1,300 @@
...
+    {KT_CHIP_MUSTANG_5200, KT_FW_STATE_RUN_IROM, {'5', '2', '1', '0', 'I', 'R'}, 6},
+    {KT_CHIP_MUSTANG_5200, KT_FW_STATE_RUN_APP, {'K', 'T', '5', '2', 'X', '0'}, 6},
+};
+
+static KtDpDevInfo dp_dev_infos[MAX_DEV_NUM];
+static const gchar *kt_dp_fw_run_state_strs[KT_FW_STATE_NUM] = {"iROM",
+								"App",
+								"Boot-Code",
+								"Unknown"};
+
```
superm1, line 44: can this be moved into `fu_kinetic_dp_aux_isp_get_fw_run_state_str` instead with a switch/case or a bunch of `if`'s?

plugins/kinetic-dp/fu-kinetic-dp-common.c
```diff
@@ -0,0 +1,41 @@
...
+	return NULL;
+}
+
+const gchar *
+fu_kinetic_dp_family_to_string(FuKineticDpFamily family)
+{
+	if (family == FU_KINETIC_DP_FAMILY_MUSTANG)
+		return "MUSTANG";
+	if (family == FU_KINETIC_DP_FAMILY_JAGUAR)
+		return "JAGUAR";
```
superm1, line 29: Do you want to change these to not be codenames?  I saw the other commit you changed them.

plugins/kinetic-dp/fu-kinetic-dp-common.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+FuKineticDpFamily
+fu_kinetic_dp_chip_id_to_family(KtChipId chip_id)
+{
+	if (chip_id == KT_CHIP_MUSTANG_5200)
+		return FU_KINETIC_DP_FAMILY_MUSTANG;
+	if (chip_id == KT_CHIP_JAGUAR_5000)
+		return FU_KINETIC_DP_FAMILY_JAGUAR;
+	return FU_KINETIC_DP_FAMILY_UNKNOWN;
```
superm1, line 40: Do you want to change these to not be codenames? I saw the other commit you changed them.

plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c
```diff
@@ -0,0 +1,1246 @@
...
+			/* maximum length of each AUX write transaction is 16 bytes */
+			guint32 aux_wr_size = (chunk_remaian_len >= 16) ? 16 : chunk_remaian_len;
+			if (!fu_kinetic_dp_connection_write(connection,
+							    DPCD_ADDR_KT_AUX_WIN + chunk_offset,
+							    remain_payload + chunk_offset,
+							    aux_wr_size,
+							    error)) {
+				g_prefix_error(
+				    error,
+				    "failed to AUX write at payload 0x%lX: ",
```
superm1, line 430: this is breaking CI:
```
../plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c: In function ‘fu_kinetic_dp_secure_aux_isp_send_payload’:
../plugins/kinetic-dp/fu-kinetic-dp-secure-aux-isp.c:430:45: error: format ‘%lX’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘long long unsigned int’ [-Werror=format=]
  430 |         "failed to AUX write at payload 0x%lX: ",
      |                                           ~~^
      |                                             |
      |                                             long unsigned int
      |                                           %llX
  431 |         (guint64)((remain_payload + chunk_offset) - payload));
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |         |
      |         long long unsigned int
```

hughsie: Fixed.

plugins/kinetic-dp/fu-kinetic-dp-aux-isp.c
hughsie: There still looks like some global state remaining.

plugins/kinetic-dp/fu-kinetic-dp-aux-isp.c
hughsie: Or `NULL`

hughsie:
@jeffrey-lin-kinetic it's looking much better. There is one thing to fix and a couple of things to check, and then I think we can merge this.

hughsie:
@jeffrey-lin-kinetic there are a few unanswered questions and then a tiny bit more work to get this upstream. Is this still what you want to do?

hughsie:
I've rebased this against main, but I think we should probably close this PR in the new year if there's no interest in merging it.

superm1:
@jeffrey-lin-kinetic ^, it's so close to the finish line, you sure you don't want it?

PR #3427, opened 2021-06-25
trivial: fix win32 build

superm1:
Use the gusb that we build rather than the system one.

Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Looks good, but won't this break when the new libgusb appears in Fedora and the subproject no longer gets built?

superm1:
> Looks good, but won't this break when the new libgusb appears in Fedora and the subproject no longer gets built?

Same problem as libxmlb I think..

superm1:
> Looks good, but won't this break when the new libgusb appears in Fedora and the subproject no longer gets built?

Yeah I think that's true.  Can we maybe just force to always use subproject libgusb?

hughsie:
> Same problem as libxmlb I think

Indeed. I'm not sure I want to be generating that file automatically but it might be safer.

hughsie:
>  Can we maybe just force to always use subproject libgusb

ohh for win32, yes! that should work.

PR #3429, opened 2021-06-27
redfish: Allow discovery of Redfish BMCs specified by VID-PID or MAC

hughsie:
Also, allow overriding the data using the config file.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/redfish/fu-plugin-redfish.c
```diff
@@ -45,23 +46,137 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
...
+	if (g_bytes_get_size (indications) != 4) {
+		g_set_error (error,
+			     FWUPD_ERROR,
+			     FWUPD_ERROR_INVALID_FILE,
+			     "invalid value for %s, got %" G_GSIZE_FORMAT " bytes",
+			     REDFISH_EFI_INFORMATION_INDICATIONS,
+			     g_bytes_get_size (indications));
+		return FALSE;
+	}
+	memcpy (&indications_le, g_bytes_get_data (indications, NULL), 4);
```
superm1, line 74: isn't this going to matter what architecture it's run on?

plugins/redfish/fu-redfish-network.c
```diff
@@ -0,0 +1,222 @@
...
+{
+	g_autoptr(GDBusProxy) proxy = NULL;
+	g_autoptr(GVariant) addr_data = NULL;
+
+	proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
+					       G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS,
+					       NULL,
+					       "org.freedesktop.NetworkManager",
+					       object_path,
+					       "org.freedesktop.NetworkManager.IP4Config",
```
superm1, line 34: network manager on a server?  Is this super common?

plugins/redfish/fu-self-test.c
```diff
@@ -27,11 +28,47 @@ fu_test_redfish_common_func (void)
 	g_assert_cmpstr (maca, ==, "00:01:02:03:04:05");
 }
 
+static void
+fu_test_redfish_network_mac_addr_func (void)
+{
+	g_autofree gchar *ip_addr = NULL;
+	g_autoptr(GError) error = NULL;
+
```
superm1, line 36: looking at the CI failures, I think if you're going to have a hard requirement on NM, then you're going to need to let the test suite check for NM and skip tests if it's not there

plugins/redfish/fu-self-test.c
```diff
@@ -27,11 +28,47 @@ fu_test_redfish_common_func (void)
...
 int
 main (int argc, char **argv)
 {
+	g_autoptr(GError) error = NULL;
+	g_setenv ("FWUPD_REDFISH_VERBOSE", "1", TRUE);
 	g_test_init (&argc, &argv, NULL);
 	g_log_set_fatal_mask (NULL, G_LOG_LEVEL_ERROR | G_LOG_LEVEL_CRITICAL);
 	g_test_add_func ("/redfish/common", fu_test_redfish_common_func);
+	g_test_add_func ("/redfish/network{mac_addr}", fu_test_redfish_network_mac_addr_func);
+	g_test_add_func ("/redfish/network{vid_pid}", fu_test_redfish_network_vid_pid_func);
```
superm1, line 72: I think you want a test that shows that config files can override smbios values

plugins/redfish/fu-plugin-redfish.c
hughsie: It is, good catch. Fixed.

plugins/redfish/fu-redfish-network.c
hughsie: On RHEL, yes, it's a hard dep. On others not sure, and I'd be totally okay if someone wanted to do a netlink-style implementation if they wanted; I just don't want that person to be me :)

plugins/redfish/fu-self-test.c
hughsie: Agreed, fixed.

plugins/redfish/fu-self-test.c
hughsie: With fresh eyes, the whole plugin is woefully undertested. Will work on that some more today.

hughsie: RIP my morning.

hughsie: Also, I've not finished yet; more tests in progress. Other ideas welcome.

superm1: How does redfish.py get launched?  I didn't notice anything connecting the two.

plugins/redfish/tests/redfish.py
```diff
@@ -0,0 +1,125 @@
...
+@app.route("/redfish/v1/UpdateService/FirmwareInventory/BIOS")
+def firmware_inventory_bios():
+
+    res = {
+        "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/BIOS",
+        "@odata.type": "#SoftwareInventory.v1_2_3.SoftwareInventory",
+        "Id": "BIOS",
+        "LowestSupportedVersion": "P79 v1.10",
+        "Manufacturer": "Contoso",
+        "Name": "Contoso BIOS Firmware",
```
superm1, line 113: I really hope this didn't come from a real system...

plugins/redfish/fu-self-test.c
hughsie: At the moment, manually. I'm not sure how to wire it up for autostart, or if it makes sense to install tons of flask deps on the CI system.

plugins/redfish/tests/redfish.py
hughsie: Nah, I'm using Redfish-Mockup-Server as the reference data until real hardware arrives.

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -45,23 +46,139 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
...
+{
+	FuPluginData *data = fu_plugin_get_data (plugin);
+	FuContext *ctx = fu_plugin_get_context (plugin);
+	const gchar *smbios_data_fn;
+	g_autofree gchar *hostname = NULL;
+	g_autoptr(FuRedfishSmbios) redfish_smbios = fu_redfish_smbios_new ();
+	g_autoptr(GBytes) smbios_data = NULL;
+
+	/* is optional if not in self tests */
+	smbios_data_fn = g_getenv ("FWUPD_REDFISH_SMBIOS_DATA");
```
superm1, line 112: should this maybe a fu-firmware override instead of redfish specific override?  Presumably the whole daemon should be able to run with fake smbios tables

plugins/redfish/fu-self-test.c
superm1: at least one CI target I think we should install them and auto-start it.  I think in terms of auto-starting, maybe just do what we do for the TPM server on arch?  run it in the background before test suite starts?

plugins/redfish/fu-plugin-redfish.c
hughsie: Yes, I need to write a `FuSmbios->write()` handler before we can do that. It's probably one for a rainy day stuck in a cabin.

plugins/redfish/fu-self-test.c
hughsie: Done, waiting for CI to see if it works.

hughsie: It works!

hughsie:
I just tried this on actual hardware; with the fixups it actually works...

plugins/redfish/README.md
```diff
@@ -23,8 +23,15 @@ This plugin supports the following protocol ID:
...
-without modification. Devices without GUIDs are not supported.
+These devices use the provided GUID provided in the `SoftwareId` property
+without modification if it is a valid GUID. If the property is not a GUID then
+the vendor instance ID is used instead:
+
+* `REDFISH\\VENDOR_${Manufacturer}&SOFTWAREID_${SoftwareId}`
+
+Additionally, this Instance ID is added for quirk and parent matching:
+
+* `REDFISH\VENDOR_${Manufacturer}&ID_${Id}`
```
superm1, line 34: what does `Id` mean in this context?

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -8,69 +8,179 @@
...
 	FuPluginData *data = fu_plugin_get_data (plugin);
-	GPtrArray *devices;
+	g_autoptr(GPtrArray) devices = NULL;
+	g_autoptr(GError) error_local = NULL;
 
 	/* get the list of devices */
-	if (!fu_redfish_client_coldplug (data->client, error))
+	if (!fu_backend_coldplug (FU_BACKEND (data->backend), &error_local)) {
+		if (g_error_matches (error_local, FWUPD_ERROR, FWUPD_ERROR_AUTH_FAILED))
+			fu_plugin_add_flag (plugin, FWUPD_PLUGIN_FLAG_UNLOCK_REQUIRED);
```
superm1, line 30: you don't want to fail coldplug though in this instance do you?  Don't you want to allow an unlock?

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -8,69 +8,179 @@
...
-	g_autoptr(GBytes) smbios_data = NULL;
+	g_autofree gchar *username = NULL;
+	g_autoptr(GError) error_uefi = NULL;
 
 	/* optional */
-	smbios_data = fu_context_get_smbios_data (ctx, REDFISH_SMBIOS_TABLE_TYPE);
+	if (!fu_redfish_plugin_discover_smbios_table (plugin, error))
+		return FALSE;
+	if (!fu_redfish_plugin_discover_uefi_credentials (plugin, &error_uefi)) {
+		g_debug ("failed to get username and password automatically: %s",
```
superm1, line 168: Isn't this useful to show to the user somehow?  It would be good to be able to tell whether you need to fill out a conffile or if there is a problem.

plugins/redfish/fu-redfish-backend.c
```diff
@@ -0,0 +1,391 @@
...
+	GHashTable		*request_cache;		/* str:GByteArray */
+};
+
+G_DEFINE_TYPE (FuRedfishBackend, fu_redfish_backend, FU_TYPE_BACKEND)
+
+const gchar *
+fu_redfish_backend_kind_to_string (FuRedfishBackendKind kind)
+{
+	if (kind == FU_REDFISH_BACKEND_UPDATE_KIND_LEGACY)
+		return "mooo";
```
superm1, line 41: ?

plugins/redfish/fu-redfish-backend.c
```diff
@@ -0,0 +1,391 @@
...
+	if (kind == FU_REDFISH_BACKEND_UPDATE_KIND_MULTIPART)
+		return "multipart";
+//	if (kind == FU_REDFISH_BACKEND_UPDATE_KIND_OEM_LENOVO)
+//		return "oem-lenovo";
```
superm1, line 45: what's this?

plugins/redfish/fu-redfish-backend.c
```diff
@@ -0,0 +1,391 @@
...
+{
+	g_autoptr(FuDevice) dev = fu_redfish_device_new_from_object (self, member);
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	locker = fu_device_locker_new (dev, error);
+	if (locker == NULL)
+		return FALSE;
+	if (self->max_image_size != 0)
+		fu_device_set_firmware_size_max (dev, (guint64) self->max_image_size);
+	//fu_device_add_private_flag (dev, FU_REDFISH_DEVICE_FLAG_USE_MULTIPART);
```
superm1, line 99: how do you know when this is needed?

plugins/redfish/fu-redfish-device.c
```diff
@@ -0,0 +1,739 @@
...
+	/* sometimes an array, sometimes an object! */
+	if (json_object_has_member (json_obj, "PCIeFunctions")) {
+		JsonNode *pcie_functions = json_object_get_member (json_obj, "PCIeFunctions");
+		if (JSON_NODE_HOLDS_OBJECT (pcie_functions)) {
+			JsonObject *obj = json_node_get_object (pcie_functions);
+			if (json_object_has_member (obj, "@odata.id")) {
+				const gchar *id = json_object_get_string_member (obj, "@odata.id");
+				if (!fu_redfish_device_probe_related_pcie_functions (self, id, error))
+					return FALSE;
+			}
```
superm1, line 177: this seems like a duplicate block of code -> static function?

plugins/redfish/fu-self-test.c
```diff
@@ -8,7 +8,50 @@
...
+	g_assert (ret);
+
+	self->plugin = fu_plugin_new (ctx);
+	pluginfn = g_build_filename (PLUGINBUILDDIR,
+				     "libfu_plugin_redfish." G_MODULE_SUFFIX,
+				     NULL);
+	ret = fu_plugin_open (self->plugin, pluginfn, &error);
+	g_assert_no_error (error);
+	g_assert (ret);
+	ret = fu_plugin_runner_startup (self->plugin, &error);
```
superm1, line 44: how do you make sure it tries to use the `redfish.py` to make the self test results reproducible even on real hardware and not let smbios tables from the hardware influence the test?  Perhaps need to override the SMBIOS tables to `/dev/null`?

src/fu-util-common.c
```diff
@@ -1263,8 +1263,10 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 	if (tmp != NULL) {
 		g_autofree gchar *desc = NULL;
 		desc = fu_util_convert_description (tmp, NULL);
-		/* TRANSLATORS: multiline description of device */
-		fu_common_string_append_kv (str, idt + 1, _("Description"), desc);
+		if (desc == NULL) {
+			/* TRANSLATORS: multiline description of device */
+			fu_common_string_append_kv (str, idt + 1, _("Description"), desc);
```
superm1, line 1268: you mean `tmp` here

plugins/redfish/README.md
superm1: Yes, It certainly wasn't clear to me until you explained it.

hughsie: Okay, fixed, thanks.

plugins/redfish/fu-plugin-redfish.c
hughsie: I think I chose the wrong plugin flag. What I really wanted was something like FWUPD_PLUGIN_FLAG_AUTH_REQUIRED. The logic then that "fwupdmgr get-plugins" would clearly show why it's failed to coldplug with some nice red text.

plugins/redfish/fu-plugin-redfish.c
hughsie: The UEFI credentials have been deprecated apparently, and so failure to have them set isn't a real problem. It's only a problem if we can't get them from anywhere else, e.g. from the conf file or from KCS -- the latter bit I've not written yet.

plugins/redfish/fu-redfish-backend.c
hughsie: HA, so that slipped through. Fixed.

plugins/redfish/fu-redfish-backend.c
hughsie: I might need to explicitly say "use this lenovo variant of Redfish for the multipart push. I'm trying to avoid it by doing fallbacks in the right place. It might be better to use a private flag for this. Not sure yet.

hughsie: Also notable is that I'm pushing the Lenovo team to align the XCC-variant of Redfish into conforming with actual redfish.

plugins/redfish/fu-redfish-backend.c
hughsie: My heuristic so far is USE_MULTIPART if MultipartHttpPushUri is defined, or use legacy if not. I'm not sure if that's good enough or if we'll have to quirk that in the conf file.

plugins/redfish/fu-redfish-device.c
hughsie: Nah, it's subtly different unfortunately. I might make some convenience helpers for getting arrays of @odata.id objects if you think this is a good idea.

plugins/redfish/fu-self-test.c
hughsie: I'm setting FWUPD_REDFISH_SMBIOS_DATA to get a static SMBIOS table, no?

plugins/redfish/fu-plugin-redfish.c
hughsie: Fixed.

plugins/redfish/fu-redfish-backend.c
hughsie: Removed as we're pushing everything to conform to DSP2062 instead

plugins/redfish/fu-redfish-backend.c
hughsie: Removed.

hughsie:
@superm1 can you do one more round of review on this please; I'm still working out a few kinks but it seems to mainly work well. Thanks!

plugins/redfish/fu-redfish-device.c
```diff
@@ -512,19 +512,106 @@ fu_redfish_device_get_backend (FuRedfishDevice *self)
...
+	if (g_strcmp0 (message_id, "Update.1.1.UpdateSuccessful") == 0) {
+		fu_device_set_status (FU_DEVICE (self), FWUPD_STATUS_IDLE);
+		return TRUE;
+	}
+	if (g_strcmp0 (message_id, "Update.1.1.InstallingOnComponent") == 0) {
+		fu_device_set_status (FU_DEVICE (self), FWUPD_STATUS_DEVICE_WRITE);
+		return TRUE;
+	}
+	g_warning ("no idea how to process MessageId %s", message_id);
+	return TRUE;
```
superm1, line 605: shouldn't this be FALSE and an error?

hughsie: it's not critical if we don't know what to do with the message, it could be something like HpeFirmwareUpdateRegistry.1.0.CheckLicenseExpiration or something. Maybe it's best if that function just doens't return anything and we drop the GError?

superm1: seems fine to me

plugins/redfish/fu-self-test.c
superm1: Ah I missed that, yes that's good

plugins/redfish/fu-redfish-device.c
hughsie: Okay, fixed.

superm1 (commented):
nothing else stands out to me

PR #3430, opened 2021-06-28
Wip/tari/realtek mst

hughsie:
Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@tari do we have any firmware on the LVFS we can use for a device-test? e.g. https://github.com/fwupd/fwupd/blob/master/data/device-tests/devices/bizlink-no-sku-vli-tier1.json -- that can be a follow up when ready, merging shouldn't block it.

tari:
We don't have a firmware on LVFS (and I'm unsure if we're allowed to do so), but the (draft) Chromium package that includes the firmware [has sources available](https://chromium-review.googlesource.com/c/chromiumos/overlays/board-overlays/+/2966630) and the [firmware files it refers to](https://gsdview.appspot.com/chromeos-localmirror/distfiles/realtek-rtd2142-firmware-3.3.tar.xz) are available publicly.

plugins/realtek-mst/fu-realtek-mst-device.c
```diff
@@ -0,0 +1,863 @@
...
+			if (i2c_buses->len > 1) {
+				g_debug ("ignoring %u additional i2c-dev under %s",
+					 i2c_buses->len - 1,
+					 fu_udev_device_get_sysfs_path (i2c_device));
+			}
+
+			bus_device = g_ptr_array_index (i2c_buses, 0);
+			g_debug ("Found I2C bus at %s, using this device",
+				 fu_udev_device_get_sysfs_path (bus_device));
+//			fu_udev_device_set_dev (FU_UDEV_DEVICE (self),	//FIXME export?
```
tari, line 181: This is now exported, right?

hughsie:
@Asiderr how am I supposed to add meson flags to the FreeBSD CI? It's not really practical to modify sysutils/fwupd/Makefile in a different repo every time we add a plugin that doesn't build on FreeBSD... For the moment I'm just using `sed`...

tari (approved):

plugins/realtek-mst/fu-realtek-mst-device.c
hughsie: Good catch, fixed, thanks.

Asiderr:
@hughsie The files in the FreeBSD build system are replaced with the files placed in the `contrib/freebsd` lines 48-51:
https://github.com/fwupd/fwupd/blob/master/contrib/ci/build_freebsd_package.sh#L49. So you can add new meson flag to the Makefile in the `fwupd` repo: https://github.com/fwupd/fwupd/blob/master/contrib/freebsd/Makefile.

hughsie:
@AsciiWolf like this https://github.com/fwupd/fwupd/pull/3431 -- if so, sorry for the noise!

PR #3442, opened 2021-07-01
Compile with the flashrom plugin in Ubuntu now that flashrom is being…

GlenPickle:
Compile with the flashrom plugin in Ubuntu now that flashrom is being included in main (https://bugs.launchpad.net/ubuntu/+source/flashrom/+bug/1912371)

Type of pull request:
- contrib/debian/rules change due to dependency changes in Ubuntu

### DIFF
--- a/contrib/debian/rules
+++ b/contrib/debian/rules
@@ -16,14 +16,10 @@ ifneq ($(CI),)
 	CONFARGS += --werror
 endif
 
-ifeq (yes,$(shell dpkg-vendor --derives-from Ubuntu && echo yes))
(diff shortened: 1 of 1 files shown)


superm1:
Can you please reference the MIR in Ubuntu in your commit message? I see it open but it will be good for future cross reference.

superm1 (approved):
Code is good, just add reference to commit message and we're good to go.

GlenPickle:
Added the link to the LaunchPad bug in an amended commit as well as the comment for this PR.

superm1 (approved):

PR #3444, opened 2021-07-02
Move amdgpu safety check into the engine

superm1:
There are now multiple plugins using drm_dp_aux_dev interface which
may potentially be combined with an amdgpu.  Prevent exercising this
interface with any plugin using DP aux unless a new enough kernel is
installed.

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -6107,6 +6107,43 @@ fu_engine_backend_device_removed_cb (FuBackend *backend, FuDevice *device, FuEng
 	}
 }
 
+static gboolean
+fu_engine_check_device_safe (FuEngine *self, FuDevice *device, GError **error)
+{
+#ifdef __linux__
+	/* see https://github.com/fwupd/fwupd/issues/1121 for more details */
+	if (FU_IS_UDEV_DEVICE (device) &&
+	    g_strcmp0 (fu_udev_device_get_subsystem (FU_UDEV_DEVICE (device)), "drm_dp_aux_dev") == 0) {
+		gsize bufsz = 0;
+		const gchar *minimum_kernel;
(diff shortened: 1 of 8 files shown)


src/fu-engine.c
```diff
@@ -6119,6 +6156,14 @@ fu_engine_backend_device_added_cb (FuBackend *backend, FuDevice *device, FuEngin
 		g_debug ("%s added %s", fu_backend_get_name (backend), str);
 	}
 
+	/* engine safety checks */
+	if (!fu_engine_check_device_safe (self, device, &error_local)) {
```
hughsie, line 6160: Isn't this quite a big change in behaviour? Before we wouldn't probe or setup the device as the plugin was disabled, and with this it seems we'd probe, setup and only ignore the device when it's being added. IIRC wasn't it the initialisation that caused the kernel to lockup? This seems like the sort of thing you want to do in `fu_engine_load_plugins` and set `FWUPD_PLUGIN_FLAG_DISABLED` with `FWUPD_PLUGIN_FLAG_KERNEL_TOO_OLD` or something.

superm1: It should be before the plugin's probe function runs, look specifically it's before fu_device_probe runs.

I can certainly see the argument moving it to where you suggested though too.

hughsie: want me to have a go, or you already done? :)

superm1: Go ahead :)

hughsie: Alternate proposal in https://github.com/fwupd/fwupd/pull/3453

PR #3445, opened 2021-07-02
plugins: rts54hub: update rts54hub.quirk

AnyProblem:
add Acer Dock ID

Type of pull request:
- [ ] Code fix
- [*] Feature
- [ ] Documentation

### DIFF
--- a/plugins/rts54hub/rts54hub.quirk
+++ b/plugins/rts54hub/rts54hub.quirk
@@ -43,7 +43,6 @@ Rts54RegisterAddrLen = 0x04
 [USB\VID_2BEF&PID_1009]
 Plugin = rts54hub
 GType = FuRts54HubDevice
-Vendor = Realtek
 Name = Acer D501 Dock USB Hub
 FirmwareSizeMin = 0x10000
 FirmwareSizeMax = 0x40000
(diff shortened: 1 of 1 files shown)


plugins/rts54hub/rts54hub.quirk
```diff
@@ -38,3 +38,24 @@ FirmwareSizeMax = 0x90000
 Rts54TargetAddr = 0x20
 Rts54I2cSpeed = 0x2
 Rts54RegisterAddrLen = 0x04
+
+# Acer D501 Dock
+[USB\VID_2BEF&PID_1009]
+Plugin = rts54hub
+GType = FuRts54HubDevice
+Vendor = Realtek
```
hughsie, line 46: Not Acer?

AnyProblem: This VID is QSI( Quanta Storage Inc) which is ODM of Acer Dock
Acer do not customized ID on this project

hughsie: Okay, so why do we need the Realtek override? Usually we just set the vendor automatically based on the VID.

AnyProblem: okay, so just don't set the Vendor default is automatic?

hughsie: Yes, I think you can just remove the `Vendor=` line completely.

Issue #3441, opened 2021-07-01
Please add a null-update UEFI capsule update option


steve-mcintyre:
Testing that shim calls fwupd ok is hard, with no way to do this on a VM (for example). It would be Really Nice (TM) if thrre was a way to provoke fwupd to create a null/stubbed-out EFI update so that we can validate this works.

I think we don't need to actually call UpdateCapsule(), just start the fwupd.efi program and say "Congratulations! Here is how you called me: <stuff>".

AFAICS fwupd is about the only user of the load options handling in shim, and it's in a fragile code path inside the shim binary. It would be lovely to be able to validate that at least the basics are working here, for reliable CI and testing.

superm1:
There is already a message that shows up, but it's not there for a long time:
https://github.com/fwupd/fwupd-efi/blob/main/efi/fwupdate.c#L564

There is also already the ability to fake an ESRT via an environment variable.  How about a flow like this:

1. dd if=/dev/zero of=foo.bin bs=1
2. FWUPD_SYSFSFWDIR=directory fwupdtool install-blob foo.bin  $GUID
3. Reboot, and check that you see that message.

All we need to do then is give some clearer directions on all of that, and add a longer delay in fwupd-efi in the no capsules case.

BTW if you're not already aware, we're also considering this: https://github.com/fwupd/fwupd/pull/3421 which will take shim arguments junk entirely out of the equation if you use GRUB...

hughsie:
> fwupdtool install-blob foo.bin $GUID

Don't we need something to add a fake device with that GUID? e.g. like a fake ESRTR entry?

superm1:
Yeah, we use a fake esrt entry by `FWUPD_SYSFSFWDIR`, no?

steve-mcintyre:
> Yeah, we use a fake esrt entry by `FWUPD_SYSFSFWDIR`, no?

So what I'm asking for is already available then? If so, can you give me a 100% spelled-out example of how to use it please? :-)

I'm feeling dumb - I have no idea what $GUID might work, etc.

hughsie:
> I'm feeling dumb - I have no idea what $GUID might work, etc.

You're certainly not dumb, this is much too hard to test. What about something like https://github.com/fwupd/fwupd/pull/3447

steve-mcintyre:
looks bang-on, thanks!

PR #3447, opened 2021-07-02
Add a test UEFI device that can be used to test capsules

hughsie:
Enable the 'test' plugin in /etc/fwupd/daemon.conf and then use
something like:

    fwupdmgr install plugins/test/example/firmware.cab

Fixes https://github.com/fwupd/fwupd/issues/3441

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/meson.build
+++ b/plugins/uefi-capsule/meson.build
@@ -185,4 +185,24 @@ if get_option('tests')
   )
   test('uefi-self-test', e, env: env)
 endif
+
+# to use these do `sudo systemctl edit fwupd.service` and set
+# Environment="FWUPD_SYSFSFWDIR=/usr/share/installed-tests/fwupd"
+install_data([
+    'tests/efi/esrt/entries/entry0/capsule_flags',
+    'tests/efi/esrt/entries/entry0/fw_class',
+    'tests/efi/esrt/entries/entry0/fw_type',
+    'tests/efi/esrt/entries/entry0/fw_version',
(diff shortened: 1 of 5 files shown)


superm1 (commented):
To be honest I don't like this approach. It means your host system is still involved, and the esrt may not be present, the UEFI plugin may disable as a result.

I think it's better to fake a test mode only esrt.

A helper script can be distributed to make it easy.

hughsie:
So would the recommendation be to run fwupd with something like `Environment="FWUPD_SYSFSFWDIR=/usr/share/fwupd/tests/firmware"` by using `sudo systemctl edit fwupd.service`?

superm1:
> So would the recommendation be to run fwupd with something like `Environment="FWUPD_SYSFSFWDIR=/usr/share/fwupd/tests/firmware"` by using `sudo systemctl edit fwupd.service`? 

I think that makes perfect sense yes.

plugins/uefi-capsule/meson.build
```diff
@@ -189,4 +189,24 @@ if get_option('tests')
   )
   test('uefi-self-test', e)
 endif
+
+# to use these do `sudo systemctl edit fwupd.service` and set
+# Environment="FWUPD_SYSFSFWDIR=/usr/share/installed-tests/fwupd"
```
superm1, line 194: I think we want to add a test to gnome-desktop-tests to call out this thing too if the device was detected (IE this was done)

PR #3452, opened 2021-07-03
Allow calculating the CRC-16 with the CCITT polynomial

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.h
+++ b/libfwupdplugin/fu-common.h
@@ -404,6 +404,10 @@ guint8		 fu_common_crc8			(const guint8	*buf,
 						 gsize		 bufsz);
 guint16		 fu_common_crc16		(const guint8	*buf,
 						 gsize		 bufsz);
+guint16		 fu_common_crc16_full		(const guint8	*buf,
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-common.h
```diff
@@ -404,6 +404,10 @@ guint8		 fu_common_crc8			(const guint8	*buf,
 						 gsize		 bufsz);
 guint16		 fu_common_crc16		(const guint8	*buf,
 						 gsize		 bufsz);
+guint16		 fu_common_crc16_full		(const guint8	*buf,
```
superm1, line 407: Maybe also switch code to use this one and mark the other as deprecated and to remove/rename on next abi bump?

superm1 (approved):

libfwupdplugin/fu-common.h
hughsie: Not convinced, I think the fff/IBM is the common case tbh. This matches the crc32 API too.

PR #3453, opened 2021-07-03
Move amdgpu safety check into the plugin

hughsie:
There are now multiple plugins using drm_dp_aux_dev interface which
may potentially be combined with an amdgpu. Prevent exercising this
interface with any plugin using DP aux unless a new enough kernel is
installed.

Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -1303,6 +1303,62 @@ fu_common_string_uncamelcase (const gchar *str)
 	return g_string_free (tmp, FALSE);
 }
 
+
+static gboolean
+fu_plugin_check_amdgpu_dpaux (FuPlugin *self, GError **error)
+{
+#ifdef __linux__
+	gsize bufsz = 0;
+	g_autofree gchar *buf = NULL;
+	g_auto(GStrv) lines = NULL;
+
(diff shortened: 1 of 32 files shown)


libfwupdplugin/fu-plugin.c
```diff
@@ -1303,6 +1303,62 @@ fu_common_string_uncamelcase (const gchar *str)
...
+
+	/* no module support in the kernel, we can't test for amdgpu module */
+	if (!g_file_test ("/proc/modules", G_FILE_TEST_EXISTS))
+		return TRUE;
+	if (!g_file_get_contents ("/proc/modules", &buf, &bufsz, error))
+		return FALSE;
+	lines = g_strsplit (buf, "\n", -1);
+	for (guint i = 0; lines[i] != NULL; i++) {
+		if (g_str_has_prefix (lines[i], "amdgpu ")) {
+			/* this was 3 years ago! */
```
superm1, line 1323: This won't age well, it's better to leave a date in the comment

superm1 (approved):

libfwupdplugin/fu-plugin.c
hughsie: Fair point, changed.

PR #3456, opened 2021-07-06
trivial: Fallback to text if the multiline description was invalid XML

hughsie:
Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1263,6 +1263,8 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 	if (tmp != NULL) {
 		g_autofree gchar *desc = NULL;
 		desc = fu_util_convert_description (tmp, NULL);
+		if (desc == NULL)
+			desc = g_strdup (tmp);
 		/* TRANSLATORS: multiline description of device */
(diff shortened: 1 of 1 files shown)


src/fu-util-common.c
```diff
@@ -1263,8 +1263,10 @@ fu_util_device_to_string (FwupdDevice *dev, guint idt)
 	if (tmp != NULL) {
 		g_autofree gchar *desc = NULL;
 		desc = fu_util_convert_description (tmp, NULL);
-		/* TRANSLATORS: multiline description of device */
-		fu_common_string_append_kv (str, idt + 1, _("Description"), desc);
+		if (desc == NULL) {
+			/* TRANSLATORS: multiline description of device */
+			fu_common_string_append_kv (str, idt + 1, _("Description"), desc);
```
superm1, line 1268: Shouldn't this be tmp?

superm1 (changes requested):

src/fu-util-common.c
hughsie: Gah, that was all sorts of wrong.  Fixed, thanks.

superm1 (approved):

PR #3459, opened 2021-07-07
Export the build timestamp from plugins

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
I'm not exactly convinced with the `build_timestamp` name. It's meant to be the "`date and time the firmware was built`" and seems to be hugely important to server vendors for some reason.

superm1 (commented):
Everywhere it's `build_timestamp` I think it should explicitly be `fw_build_timestamp` to disambiguate manufacturing date and show that it's the firmware's build date

hughsie:
> fw_build_timestamp

I'm not sure about that; we have `firmware` rather than `fw` everywhere in the API now. `firmware_build_timestamp` is kinda long. Some other ideas:

 * build_timestamp
 * build_version
 * build_created
 * version_timestamp
 * version_date
 * version_created
 * version_built
 * version_release
 * version_build
 * version_created

Other ideas certainly welcome.

superm1:
> * version_created

That one is my favorite of those.  Or how about `firmware_build_date`

hughsie:
> version_created

Actually, that's probably the most correct answer given we already have a `created` property on the release.

	/**
	 * fwupd_release_get_created:
	 * @self: a #FwupdRelease
	 *
	 * Gets when the update was created.
	 *
	 * Returns: UTC timestamp in UNIX format, or 0 if unset
	 *
	 * Since: 1.4.0
	 **/

hughsie:
> given we already have a created property on the release

Actually that's not quite true; the created property seems to be the timestamp of the FuRelease creation, i.e. when the object was allocated. It might be confusing to have two integer properties with the same name and different semantics.

What about `version_date`? I think that's clearer.

superm1:
I personally think it's worth biting the bullet and having a longer variable name than trying to understand what two somewhat related words mean

hughsie:
> having a longer variable name

Fair point. The problem I have with `firmware_build_date` is that we don't have `firmware_version` or `firmware_serial` -- the fact that it's of the firmware is implicit. Maybe `version_build_date`, `version_build_timestamp` or `version_built_date` as a compromise?

superm1:
Yup any of those last 3 sound good

PR #3462, opened 2021-07-09
RFC implement CapsuleUpdate on-disk

apalos:
Hi!

This is the basic code for implementing CapsuleUpdate on disk.  There's still missing code for the overall functionality, but this is mostly what I had in mind.

Keep in mind U-Boot does support the functionality without needing to perform a runtime setvariable for OsIndications. 
So if we can detect U-Boot is the firmware we are running on, we can completely skip setting the variable.

The code that's currently missing is:
1. Detect u-boot is running
2. Copy the capsule on the special 'EFI\UpdateCapsule' dir.

We also need to discuss which functionality we want first. Currently the newly added capsuleupdate on disk takes precedence, unless the firmware doesn't support it, or a config falg is set.

Signed-off-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>

Type of pull request:
- [ ] Code fix
- [x] Feature
- [ ] Documentation


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -370,6 +370,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
 	return fu_plugin_uefi_capsule_write_splash_data (plugin, device, image_bmp, error);
 }
 
+static gboolean
+fu_capsule_update_on_disk (FuDevice *device, GError **error)
```
superm1, line 374: I think this function would be better to have an out pointer for the actual result, and then save the TRUE/FALSE for success/failure in setting it

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -370,6 +370,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
...
+{
+	g_autofree guint64 *os_indications_supported = NULL;
+	g_autofree guint64 *os_indications = NULL;
+	guint32 attr;
+	gsize os_ind_size = 0;
+
+	if (fu_device_get_metadata_boolean (device, "DisableCapsuleUpdateOnDisk"))
```
superm1, line 381: is your idea here that you would set this from a conffile?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -370,6 +370,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
...
+		return FALSE;
+
+	/* check for capsuleupdate ondisk */
+	if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndicationsSupported",
+				 (guint8 **) &os_indications_supported,
+				 &os_ind_size, &attr, error))
+		g_debug ("Can't Read OsIndicationsSupported");
```
superm1, line 388: you'll need to return false here and prefix error

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -370,6 +370,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
...
+	if (*os_indications_supported & EFI_OS_INDICATIONS_FILE_CAPSULE_DELIVERY_SUPPORTED) {
+		if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndications",
+				 (guint8 **) &os_indications,
+				 &os_ind_size, &attr, error)) {
+			*os_indications = 0;
+			g_debug ("Failed to read OsIndications");
```
superm1, line 394: you'll need to return false here and prefix the error

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -402,6 +446,9 @@ fu_plugin_update (FuPlugin *plugin,
 	str = _("Installing firmware update…");
 	g_assert (str != NULL);
 
+	if (fu_capsule_update_on_disk(device, error))
+			return TRUE;
+
```
superm1, line 451: I think you should detect during coldplug if capsules on disk is supported, and then install separately.  It shouldn't be a runtime decision, OsIndications will be set before OS loads.

superm1 (commented):
Your whitespace and comments doesn't match the project style, you'll want to adjust those.

In general the concept seems good to me, but the actual location of initial checks should be coldplug, and writing assets should be in `fu-uefi-device.c` in write vfunc.

superm1:
>Detect u-boot is running

How can this be detected?  By looking at `BootCurrent` and checking it's in `BootOrder`?

>Copy the capsule on the special 'EFI\UpdateCapsule' dir.

There is a copy asset function you should use for this.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: sure

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: Yea. Basically this becomes the default, unless you:
1. disable it
2. firmware doesnt support it

But I am mostly looking into feedback here, since this might break existing devices that support capsuleupdate on disk (but that should be really rare).
Alternatively I could look into getting rid of it and try to locate the protocol for runtime capsule update.  If that's not supported we could fallback to the on-disk variant

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: returning false already is the case. Effectively the next if won't succeed,, so you'll end up at the bottom. What do you mean by prefix the error?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: No you don't. OsIndications is the OS responsibility. If it's not defined by the firmware the OS can just create it.  The reason you need to read it, is OR it with any existing values.  If it's not there, you just set it to 0

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: That would be OsIndicationsSupported, but yea I think I get the idea

apalos:
> > Detect u-boot is running
> 
> How can this be detected? By looking at `BootCurrent` and checking it's in `BootOrder`?
> 

No those would be set to GRUB2/OS loader/the kernel directly depending on how you boot.
We could try to pick in smbios tables, but then again those can be disabled.  

The point here is pretty simple.  U-boot doesnt always support setvariable at runtime.  It does support updating via a capsule though even if OsIndications isn't set.  So in that case we only need to copy the firmware.

> > Copy the capsule on the special 'EFI\UpdateCapsule' dir.
> 
> There is a copy asset function you should use for this.

yea that was what I was looking for, since you already have code for copying a capsule. Pointers please?

hughsie:
@apalos I've invited you to the project; can you take your existing code and push to wip/apalos/capsule-on-disk and then we can all pile in with fixes and help you out a bit.

apalos:
> @apalos I've invited you to the project; can you take your existing code and push to wip/apalos/capsule-on-disk and then we can all pile in with fixes and help you out a bit.

Thanks @hughsie, done!

superm1:
Okay let's close this PR and have a new one based in that branch then too.

apalos:
> Okay let's close this PR and have a new one based in that branch then too.

I've already pushed the branch, so closing this

superm1:
Open up another though still, even if draft.

PR #3465, opened 2021-07-09
trivial: Print the GUIDs in the order they were added

hughsie:
This fixes the self tests on old GLib versions as the order of the
GHashTableIter is not guaranteed to be stable.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-device.c
+++ b/libfwupd/fwupd-device.c
@@ -2660,6 +2660,30 @@ fwupd_device_verstr_raw (guint64 value_raw)
 	return g_strdup_printf ("0x%08x", (guint) value_raw);
 }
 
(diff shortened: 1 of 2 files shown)


hughsie:
Found on RHEL8...

superm1 (commented):
This is a lot of extra code to work around an older glib behavior. Is it maybe worth skipping the test in older glib instead of all this technical debt?

hughsie:
> This is a lot of extra code to work around an older glib behavior

The order of the hash keys is entirely unspecified. It's completely plausible that a future _new_ glib could change the order once again.

superm1 (approved):

PR #3474, opened 2021-07-11
trivial: override support for fwupd build by runtime environment vari…

CragW:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

If a specific fwupd build went through certain level of validation cycle thereafter pursuing to remove the warning message of untested, it is now available by adding a custom environment variable at runtime. 

I've tested it with adding env `FWUPD_SUPPORTED` to `fwupd-command` from a snap build locally.

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -2137,6 +2137,9 @@ fu_util_show_unsupported_warn (void)
 {
 #ifndef SUPPORTED_BUILD
 	g_autofree gchar *fmt = NULL;
+
+	if (g_getenv ("FWUPD_SUPPORTED") != NULL)
+                return;
 	/* TRANSLATORS: this is a prefix on the console */
 	fmt = fu_util_term_format (_("WARNING:"), FU_UTIL_TERM_COLOR_YELLOW);
 	/* TRANSLATORS: unsupported build of the package */


hughsie:
I think in this context, `This package has not been validated` was meant as "validated by upstream" rather than "validated by the user". A question; what's setting `FWUPD_SUPPORTED`?

superm1:
Actually the message means it's not validated by a human. We turn it off on distro builds with the thought process that a distro maintainer has done some level of validation on it.

In Crag's case a wrapper script using the fwupd snap for a specific use would set it.

CragW:
My use case would add modification to the fwupd snap downloaded from store, afterwards repack it. The modified build will go through some validation cycles ultimately release to users.

hughsie:
> My use case would add modification to the fwupd snap downloaded from store, afterwards repack it

If you're modifying and rebuilding it, could you just add `-Dsupported_build=true` to the build arguments? If it's actually been tested then that's fine.

CragW:
I have the feeling that altering the build configuration `-Dsupported_build=true` creates another fwupd build locally instead using the stable build from snap store (upstream). I think the modification in the wrapper script has its majority to remain fwupdate unchanged, so the fwupdate elements are at last reliable.

hughsie:
Can you give me the big picture? why are you trying to disable the warning?

CragW:
The fwupd snap is part of [Dell's deliverable](https://www.dell.com/support/home/en-us/drivers/driversdetails?driverid=9x0f4&oscode=ubt18&productcode=dell-wd19-130w-dock) to end users, it's embedded in the executable. The fwupdate utility goes through some level of testing on Dell hardware systems ahead of release, nevertheless the warning message brings user confusion.
```
This package has not been validated, it may not work properly
```
I'm looking to improve user experience in this area.

hughsie:
Okay, that makes sense. I have no worries; @superm1 what do you think of the env var name?

superm1 (approved):

PR #3476, opened 2021-07-12
Move Parade PS175 support to a new plugin

tari:
Type of pull request:
- [x] Code fix
- [ ] Feature
- [ ] Documentation

This new plugin works largely the same as the existing support, but doesn't use libflashrom for device communication. Also includes the probing changes from #3344.

cc @djcampello


hughsie:
@tari this looks great, thanks. I've done some style fixups, and I'll do the review now -- but this looks **so** much better than the flashrom version IMHO.

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,829 @@
...
+};
+
+G_DEFINE_TYPE (FuParadeLspconDevice, fu_parade_lspcon_device, FU_TYPE_I2C_DEVICE)
+
+static void
+fu_parade_lspcon_device_init (FuParadeLspconDevice *self)
+{
+	FuDevice *device = FU_DEVICE (self);
+	fu_device_set_vendor (device, "Parade Technologies");
+	fu_device_add_vendor_id (device, "PCI:0x1AF8");
```
hughsie, line 81: Is PCI the right prefix here? I guess it's all we've probably got for Parade...

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,829 @@
...
+static gboolean
+fu_parade_lspcon_read_register (FuParadeLspconDevice *self,
+				guint8 register_addr,
+				guint8 *value,
+				GError **error)
+{
+	FuI2cDevice *i2c_device = FU_I2C_DEVICE (self);
+	if (!fu_i2c_device_write (i2c_device, register_addr, error))
+		return FALSE;
+	return fu_i2c_device_read (i2c_device, value, error);
```
hughsie, line 240: is this a standard i2c thing? i.e. should we rename `fu_parade_lspcon_read_register()` to `fu_i2c_device_read_register()`?

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,829 @@
...
+			     GError **error)
+{
+	FuDevice *device = FU_DEVICE (self);
+	FuI2cDevice *i2c_device = FU_I2C_DEVICE (self);
+	const gsize target_len = len;
+
+	while (len > 0) {
+		/* page 7 reads always start from the base of the mapped window- we'll
+		 * read the whole page then pull out the parts we care about, using the
+		 * full page everywhere except possibly in the first and last reads */
```
hughsie, line 302: Does FuChunk's `fu_chunk_array_new` help here? You can set the page size in that.

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,829 @@
...
+	}
+
+	/* reset clt2SPI, required before write */
+	if (!fu_parade_lspcon_write_register (self, REG_ADDR_CLT2SPI, 0x20, error))
+		return FALSE;
+	g_usleep (100 * G_TIME_SPAN_MILLISECOND);
+	if (!fu_parade_lspcon_write_register (self, REG_ADDR_CLT2SPI, 0, error))
+		return FALSE;
+
+	for (gsize bytes_written = 0; bytes_written < data_len; bytes_written += 256) {
```
hughsie, line 455: I think we should use FuChunk here

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,829 @@
...
+			     "found multiple DP aux devices with name \"%s\"",
+			     self->aux_device_name);
+		return FALSE;
+	}
+	aux_device = fu_udev_device_new (g_steal_pointer (&aux_devices->data));
+	g_debug ("using aux dev %s", fu_udev_device_get_sysfs_path (aux_device));
+
+	/* the following open() requires the device have IDs set */
+	if (!fu_udev_device_set_physical_id (aux_device, "drm_dp_aux_dev", error))
+		return FALSE;
```
hughsie, line 610: Is there any way to split this out? IIRC, i think the new fu-realtek-mst-device.c does something similar?

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,829 @@
...
+
+	/* open device to read version from DPCD */
+	if ((aux_device_locker = fu_device_locker_new (aux_device, error)) == NULL)
+		return FALSE;
+
+	/* DPCD address 00500-00502: device OUI */
+	if (!fu_udev_device_pread_full (aux_device, 0x500, (guint8 *) &oui, 3, error))
+		return FALSE;
+	oui = GUINT32_FROM_BE(oui) >> 8;
+	if (oui != 0x001CF8) {
```
hughsie, line 620: I think you want a `fu_device_add_vendor_id ("OUI:%06x")` here -- that's what we have for ATA devices.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -41,7 +40,6 @@ fu_plugin_init (FuPlugin *plugin)
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_METADATA_SOURCE, "linux_lockdown");
 	fu_plugin_add_rule (plugin, FU_PLUGIN_RULE_CONFLICTS, "coreboot"); /* obsoleted */
 	fu_plugin_add_flag (plugin, FWUPD_PLUGIN_FLAG_REQUIRE_HWID);
-	fu_plugin_add_device_gtype (plugin, FU_TYPE_FLASHROM_LSPCON_I2C_SPI_DEVICE);
 	fu_plugin_add_udev_subsystem (plugin, "i2c");
```
djcampello, line 43: This line should also go away... Following the opposite from fe073a304ddd4db10c6ca2de75a649cad65e1a9a

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: The device name that firmware provides (referred to by the quirks file) is the concatenation of this ID and the device name, so it seems appropriate to include here.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: This is a common pattern, but not standardized anywhere. It might be reasonable to provide, though a name like `fu_i2c_device_write_read` might be better since it doesn't imply any particular semantics.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: I had trouble with chunks here where the `address` of every chunk was 0 and it ended up writing at the wrong addresses, but I can have another go at it.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: The Realtek version uses the attached I2C tunnel whereas this one reads DPCD directly on the aux device; they're similar in that they need to find a device with given subsystem and sysfs name, but otherwise not.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: We don't have a way to read anywhere but from the beginning of a page, so chunking won't help us significantly here.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: Turns out the difference between page and packet sizes is significant with regard to chunk addresses.

plugins/flashrom/fu-plugin-flashrom.c
tari: Good catch; done.

tari:
Dealt with comments, and fixed a regression in verifying the boot flags.

plugins/parade-lspcon/fu-parade-lspcon-device.c
hughsie: Okay, I'll leave it until there's another user that does the same thing; rule of three and all that.

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,832 @@
...
+			return FALSE;
+
+		/* write data to page 7 memory window */
+		guard = fu_parade_lspcon_i2c_address_guard_new (self,
+								I2C_ADDR_PAGE7,
+								error);
+		/* page write is prefixed with an offset:
+		 * we always start from offset 0 */
+		write_data[0] = 0;
+		memcpy(write_data + 1, fu_chunk_get_data (chunk), chunk_size);
```
hughsie, line 474: Can you use `fu_memcpy_safe` here please.

plugins/parade-lspcon/fu-parade-lspcon-device.c
```diff
@@ -0,0 +1,832 @@
...
+			return FALSE;
+		guard = fu_parade_lspcon_i2c_address_guard_new (self,
+								I2C_ADDR_PAGE7,
+								error);
+		if (guard == NULL)
+			return FALSE;
+		if (!fu_i2c_device_read_full (i2c_device, page_data, 256, error))
+			return FALSE;
+
+		memcpy (data, page_data + page_data_start, page_data_take);
```
hughsie, line 318: Definately use `fu_memcpy_safe` here please. I think it's safe but with the helper I'm not asking the various security teams to think. Then LGTM I think.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: Done.

plugins/parade-lspcon/fu-parade-lspcon-device.c
tari: Done.

tari:
Squashed fixups, verified still working. Noticed the DP aux device was emitting some warnings so set its context before opening to fix them. Seems good to go on my end.

hughsie (approved):

PR #3481, opened 2021-07-13
emmc: add check that "name" sysattr is not NULL

daniil-lunev:
At this point it has to be present, but better have the check for
    correctness sake.
    Follow up of https://github.com/fwupd/fwupd/commit/64fa633e3029ff619a2de86e85e7fbb60cc40225


superm1 (commented):
Actually in this case you need to set error before returning false.

daniil-lunev:
Updated. I should have just replaced the function in the original CL and preserved the return, was a bit rushing. PTAL

PR #3482, opened 2021-07-13
implement capsuleupdate on-disk

apalos:
WIP for implementing capsuleupdate on-disk proposed on #3462 

Signed-off-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>

Type of pull request:
- [ ] Code fix
- [x ] Feature
- [ ] Documentation


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -366,6 +366,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
 	return fu_plugin_uefi_capsule_write_splash_data (plugin, device, image_bmp, error);
 }
 
+static gboolean
+fu_capsule_update_on_disk (FuDevice *device, GError **error)
```
superm1, line 370: This function should move into `fu_uefi_device_write_firmware`

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -366,6 +366,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
...
+{
+	g_autofree guint64 *os_indications_supported = NULL;
+	g_autofree guint64 *os_indications = NULL;
+	guint32 attr;
+	gsize os_ind_size = 0;
+
+	if (fu_device_get_metadata_boolean (device, "DisableCapsuleUpdateOnDisk"))
+		return FALSE;
```
superm1, line 378: you need to set error if you're returning false.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -366,6 +366,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
...
+		g_debug ("Can't Read OsIndicationsSupported");
+	if (*os_indications_supported & EFI_OS_INDICATIONS_FILE_CAPSULE_DELIVERY_SUPPORTED) {
+		if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndications",
+				 (guint8 **) &os_indications,
+				 &os_ind_size, &attr, error)) {
+			*os_indications = 0;
+			g_debug ("Failed to read OsIndications");
+		}
+		//if u-boot
+		//cp capsule;
```
superm1, line 393: where does it need to get set?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -366,6 +366,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
...
+	if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndicationsSupported",
+				 (guint8 **) &os_indications_supported,
+				 &os_ind_size, &attr, error))
+		g_debug ("Can't Read OsIndicationsSupported");
+	if (*os_indications_supported & EFI_OS_INDICATIONS_FILE_CAPSULE_DELIVERY_SUPPORTED) {
+		if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndications",
+				 (guint8 **) &os_indications,
+				 &os_ind_size, &attr, error)) {
+			*os_indications = 0;
+			g_debug ("Failed to read OsIndications");
```
superm1, line 390:
```
g_prefix_error (error, "failed to read OsIndications: ");
return FALSE;
```

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -366,6 +366,50 @@ fu_plugin_uefi_capsule_update_splash (FuPlugin *plugin, FuDevice *device, GError
 	return fu_plugin_uefi_capsule_write_splash_data (plugin, device, image_bmp, error);
 }
 
+static gboolean
+fu_capsule_update_on_disk (FuDevice *device, GError **error)
+{
+	g_autofree guint64 *os_indications_supported = NULL;
+	g_autofree guint64 *os_indications = NULL;
```
superm1, line 373: no need to autofree integers

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: We've discussed this on the original PR.  OsIndications is supposed to allow the OS to request the firmware to take a specific action or enable certain features.  It's not an error we should break out of, since we'll create the variable regardless.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: I am not sure I am following, plugins/uefi-capsule/fu-uefi-bootmgr.c does a similar thing while reading BootOrder.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: You mean the u-boot detection part? Since it's only need to be tested once we can run a scan during coldplug?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: ok

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: I can move it, but why rename it?  The function only needs to copy the firmware on a special directory, which is what capsule update on-disk is supposed to do for the OS. On he next reboot the firmware will pick it up and do the writing. The _write part can be misleading.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -433,6 +481,12 @@ fu_plugin_uefi_capsule_load_config (FuPlugin *plugin, FuDevice *device)
 	fu_device_set_metadata_boolean (device,
 					"FallbacktoRemovablePath",
 					fallback_removable_path);
+
+	/* delivery of Capsules via file on Mass Storage device */
+	capsule_on_disk = fu_plugin_get_config_value_boolean (plugin, "DisableCapsuleUpdateOnDisk");
```
hughsie, line 486: i'm confused why we'd want to disable it if the platform explicitly says "this is supported" -- is this in case the firmware gets the OS_INDICATIONS value wrong?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: if you tell me how to do the commented out bits in meta-code (i.e. abstract) then I can fill in some of the fwupd bits.

hughsie (changes requested):

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: Because the firmware might support 2 delivery methods.  Runtime capsuleupdates (which is what currently works on fwupd) and capsuleupdate on-disk.  With the way the patch is structured the on-disk takes precedence over the existing method, so a user might want to explicitly disable it

apalos:
@hughsie If you could take care of the 'cp capsule', it would be great. Basically we need a function that copies the capsule to <ESP>/EFI/UpdateCapsule

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: @hughsie, I think I sorted out the copy as well.  I'll update it and you can have a look

hughsie: Yes, but dont' stress. I was going to reorganise the plugin a bit to make it cleaner to add this support. The "detect" u-boot bit would be more useful tbh.

apalos: @hughsie what @superm1 proposed actually solved all my questions. fu-uefi-device.c (which I completely missed) had all the infrastructure I needed.  So I've basically split the function into two parts as I suggested.  How does this work now? Do I push an update or force push it?

plugins/uefi-capsule/fu-uefi-device.c
```diff
@@ -537,6 +537,70 @@ fu_uefi_device_cleanup (FuDevice *device,
...
+static gboolean
+fu_capsule_update_on_disk (FuDevice *device, GError **error)
+{
+	g_autofree guint64 *os_indications = NULL;
+	guint32 attr;
+	gsize os_ind_size = 0;
+
+	if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndications",
+			 (guint8 **) &os_indications,
+			 &os_ind_size, &attr, error)) {
```
superm1, line 585: you need to use a local error here or return false in this if loop result

plugins/uefi-capsule/fu-uefi-device.c
```diff
@@ -567,11 +632,15 @@ fu_uefi_device_write_firmware (FuDevice *device,
...
 		return FALSE;
-
+	/* check if we want a capsule update on-disk */
+	on_disk = fu_check_capsuleupdate_ondisk(device, error);
 	/* save the blob to the ESP */
 	directory = fu_uefi_get_esp_path_for_os (device, esp_path);
 	basename = g_strdup_printf ("fwupd-%s.cap", self->fw_class);
-	fn = g_build_filename (directory, "fw", basename, NULL);
+	if (on_disk)
+		fn = g_build_filename (directory, "EFI/UpdateCapsule", basename, NULL);
```
superm1, line 641: is this path actually right?  won't it end up in a distro specific path (for example /efi/ubuntu/efi/updatecapsule)?

plugins/uefi-capsule/fu-uefi-device.c
apalos: Since this isn't an error that should prevent us from running, I'll set a local error and ignore it

plugins/uefi-capsule/fu-uefi-device.c
apalos: Yes I think it is. The UEFI spec says:
"The directory \EFI\UpdateCapsule is checked for capsules only within the EFI system partition on the device specified in the active boot option determine by reference to BootNext variable or BootOrdervariable processing. The active Boot Variable is the variable with highest priority BootNext or within BootOrder that refers to a device found to be present. Boot variables in BootOrder but referring to devices not present are ignored when determining active boot variable"

So your BootNext or Boot#### variables define the directory you should look into for EFI/UpdateCapsule/<capsules>

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: Done

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: Actually I am going to change this, since I don't like having it as a config option.  The runtime capability of the capsule update is described on the RT_PROPERTIES table and even if it's not the QueryCapsuleCapabilities will return EFI_UNSUPPORTED.  So I'll get rid of the config option, try the runtime capsule update first and that is unsupported, we can try the on-disk one

hughsie: > try the runtime capsule update first and that is unsupported, we can try the on-disk one

That doesn't sound very explicit and I think will make it difficult to debug at scale. I would much rather it's explicit.

plugins/uefi-capsule/fu-uefi-device.c
apalos: I managed to set this up on an rpi4 for testing. I misunderstood your original point. I assumed "/efi/ubuntu' was part of were the OS mounted the ESP.  This is indeed wrong, I'll fix it

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
apalos: It would involve at least understanding if the firmware suppots a runtime capsule update service. 
That's fine, I changed the logic slightly on the entire patchset.  I think we are safe to keep things as config options for now. 

What I did is: 
1. Make this argument called EnableCapsuleUpdateOnDisk, so the user has to explicitly enable it now and we wont break existing installations
2. The setvariable problem is a bit weird, since U-Boot has no way of letting us know if ignores OsIndications.  So I added another config option for the user to enable 

I understand the goal should be that this works out of the box, but the setvariable problem is weird. I am currently working on it, and I do have a fix, but keep in mind there might be devices running U-Boot that will never support it

apalos: @hughsie I've tested the whole thing on an RPI4. What's missing from the current code is fixing `fu_uefi_backend_check_efivarfs_rw`, we need to disable that check if setvariable isn't supported at runtime.  In that case the efiavarfs is mount as ro.  
So I was thinking, maybe it's better to rename IgnoreOsIndications to SetVariableMissing or something like that and skip all the calls to it ?

hughsie: > maybe it's better to rename IgnoreOsIndications to SetVariableMissing or something

Ahh, is the IgnoreOsIndications the way of the firmware basically saying "I don't support RT->SetVariable"? Maybe just quirking the "ignore NOT_SUPPORTED" in fwupd would be a quicker approach for now?

apalos: > > maybe it's better to rename IgnoreOsIndications to SetVariableMissing or something
> 
> Ahh, is the IgnoreOsIndications the way of the firmware basically saying "I don't support RT->SetVariable"? 

Yes, but we only use it to skip setting OsIndications, hence the naming.

> Maybe just quirking the "ignore NOT_SUPPORTED" in fwupd would be a quicker approach for now?

Yea and probably better, since we'll be able to skip setting variables automatically, instead of adding a config option.  I have no idea how to do that though :)

hughsie:
Closing in favour of https://github.com/fwupd/fwupd/pull/3524

PR #3488, opened 2021-07-13
Support devices that need an action during the update process

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
The "_return error and hope the client resubmits the firmware again_" is clunky. There are two affected plugins, and about to be one more.

hughsie:
@superm1 the other way to do this is create an exported FwupdDeviceEvent object and then have a DeviceEvent signal with `asv` -- which is probably less fragile than what I have here. I'm not sure if that's architecture astronauting or not.

superm1 (commented):
so my take on this, if you're going to be biting the bullet to allow interactive stuff you may as well let the client block on it, and retry things.  
IE If it's an unplug, have the daemon send an unplug event back to the client and client will retry it's action.
If it's a press a button type thing, daemon sends a "let me know when you've done action" message and then user does it.

Otherwise it's just fragile duct tape not much different than what exists right now.

superm1:
> so my take on this, if you're going to be biting the bullet to allow interactive stuff you may as well let the client block on it, and retry things.
> IE If it's an unplug, have the daemon send an unplug event back to the client and client will retry it's action.
> If it's a press a button type thing, daemon sends a "let me know when you've done action" message and then user does it.
> 
> Otherwise it's just fragile duct tape not much different than what exists right now.

Which I guess is your counterproposal idea, and I think is probably a better way to go.  I think then we get to say what kind of events clients can do, wait for, etc.  IOW when we suddenly have to plug in an ethernet cable for an update to complete, we can actually plumb the events for said things :)

hughsie:
>  IOW when we suddenly have to plug in an ethernet cable for an update to complete, we can actually plumb the events for said things

Yes, it's a whole lot more extensible than this PR. I'll have a rethink and prototype something tomorrow. Thanks for the sanity check.

PR #3489, opened 2021-07-13
Inhibit suspend upstream

binaryNewts:
This is an extension of the powerd plugin skeleton, which has not yet landed. It serves to create and destroy /run/lock/power_override/fwupd.lock, which will hold the contents of getpid () and stop the device from being suspended while the file exists. The file will be created and written just before the update is put into motion and be destroyed once the update
finishes.

Type of pull request:
- [x] New plugin
- [x] Feature

### DIFF
--- a/plugins/powerd/fu-plugin-powerd.c
+++ b/plugins/powerd/fu-plugin-powerd.c
@@ -20,6 +20,43 @@ fu_plugin_init(FuPlugin *plugin)
 	fu_plugin_alloc_data(plugin, sizeof(FuPluginData));
 }
 
+static gboolean
+fu_plugin_powerd_create_suspend_file(GError **error)
+{
+	g_autofree gchar *lockdir = NULL;
+	g_autofree gchar *inhibitsuspend_filename = NULL;
+	g_autofree gchar *getpid_str = NULL;
+
+	lockdir = fu_common_get_path(FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_filename = g_build_filename(lockdir, "power_override", "fwupd.lock", NULL);
+	getpid_str = g_strdup_printf("%d", getpid());
+	if (!g_file_set_contents(inhibitsuspend_filename, getpid_str, -1, error)) {
+		g_prefix_error(error, "lock file unable to be created");
+		return FALSE;
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_plugin_powerd_delete_suspend_file(GError **error)
+{
+	g_autoptr(GError) local_error = NULL;
+	g_autofree gchar *lockdir = NULL;
+	g_autoptr(GFile) inhibitsuspend_file = NULL;
+
+	lockdir = fu_common_get_path(FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_file =
+	    g_file_new_build_filename(lockdir, "power_override", "fwupd.lock", NULL);
+	if (!g_file_delete(inhibitsuspend_file, NULL, &local_error) &&
+	    !g_error_matches(local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND)) {
+		g_propagate_prefixed_error(error,
+					   g_steal_pointer(&local_error),
+					   "lock file unable to be deleted");
+		return FALSE;
+	}
+	return TRUE;
+}
+
 void
 fu_plugin_destroy(FuPlugin *plugin)
 {
@@ -70,6 +107,9 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 	FuPluginData *data = fu_plugin_get_data(plugin);
 	g_autofree gchar *name_owner = NULL;
 
+	if (!fu_plugin_powerd_delete_suspend_file(error))
+		return FALSE;
+
 	/* establish proxy for method call to powerd */
 	data->proxy = g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SYSTEM,
 						    G_DBUS_PROXY_FLAGS_NONE,
@@ -98,4 +138,16 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 	g_timeout_add_seconds(5, G_SOURCE_FUNC(fu_plugin_powerd_refresh_cb), plugin);
 
 	return TRUE;
+}
+
+gboolean
+fu_plugin_update_prepare(FuPlugin *plugin, FwupdInstallFlags flags, FuDevice *dev, GError **error)
+{
+	return fu_plugin_powerd_create_suspend_file(error);
+}
+
+gboolean
+fu_plugin_update_cleanup(FuPlugin *plugin, FwupdInstallFlags flags, FuDevice *dev, GError **error)
+{
+	return fu_plugin_powerd_delete_suspend_file(error);
 }


binaryNewts:
@bleungatchromium @djcampello

bleungatchromium (changes requested):
Please remove BUG=, TEST=, and Change-Id: fields in the commit message.

These are meaningful and required for the chromiumos gerrit submission, but they have no meaning for github.

Remove these, but keep them stashed away for when you cherry-pick to chromiumos.

bleungatchromium (commented):
Is it worthwhile to remove the inhibit file at startup or init just in case the daemon (or tool) errors out or crashes before fu_plugin_update_cleanup is called?

Or is the way the inhibit file set up (with a pid) it should be harmless?

bleungatchromium (approved):

plugins/powerd/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Vendor ID Security
+------------------
+
+This protocol does not create a device and thus requires no vendor ID set.
+
+External interface access
+-------------------------
+
+This plugin requires access to a dbus interface that is to be discussed.
```
superm1, line 18: this needs to be updated

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include <stdio.h>
+
```
superm1, line 14: nit: you can remove this whitespace

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,67 @@
...
+void
+fu_plugin_init (FuPlugin *plugin)
+{
+        fu_plugin_set_build_hash (plugin, FU_BUILD_HASH);
+}
+
+gboolean
+fu_plugin_startup (FuPlugin *plugin, GError **error)
+{
+        return TRUE;
```
superm1, line 26: this needs to actually communicate to a bus to make sure that there is something on the other end

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,67 @@
...
+{
+        g_autofree gchar *lockdir = NULL;
+        g_autofree gchar *inhibitsuspend_path = NULL;
+
+        lockdir = fu_common_get_path (FU_PATH_KIND_LOCKDIR);
+        inhibitsuspend_path = g_build_filename (lockdir, "power_override", "fwupd.lock", NULL);
+        FILE *inhibit_suspend_file;
+        inhibit_suspend_file = fopen (inhibitsuspend_path, "w");
+        fprintf (inhibit_suspend_file, "%d", getpid());
+        fclose (inhibit_suspend_file);
```
superm1, line 40: use `g_file_set_contents` instead?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,67 @@
...
+}
+
+gboolean
+fu_plugin_startup (FuPlugin *plugin, GError **error)
+{
+        return TRUE;
+}
+
+static void
+fu_plugin_inhibit_suspend ()
```
superm1, line 30: if you switch to `g_file_set_contents`, this should change return type to `gboolean` and you pass `error`.

superm1 (changes requested):
make sure that CI passes, and we should have a more functional plugin before merging (IE should communicate with dbus and disable on systems without powerd)

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,62 @@
...
+
+static gboolean
+fu_plugin_inhibit_suspend (GError **error)
+{
+        g_autofree gchar *lockdir = NULL;
+        g_autofree gchar *inhibitsuspend_filename = NULL;
+
+        lockdir = fu_common_get_path (FU_PATH_KIND_LOCKDIR);
+        inhibitsuspend_filename = g_build_filename (lockdir, "power_override", "fwupd.lock", NULL);
+        g_file_set_contents (inhibitsuspend_filename, getpid(), sizeof(pid_t), error);
```
hughsie, line 37: you're missing a return here.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -0,0 +1,62 @@
...
+                          FuDevice *dev,
+                          GError **error)
+{
+        g_autofree gchar *lockdir = NULL;
+        g_autofree gchar *inhibitsuspend_path = NULL;
+
+        lockdir = fu_common_get_path (FU_PATH_KIND_LOCKDIR);
+        inhibitsuspend_path = g_build_filename (lockdir, "power_override", "fwupd.lock", NULL);
+
+        return (g_remove (inhibitsuspend_path) == 0);
```
hughsie, line 61: you need to set an error if returning FALSE. Maybe use g_file_delete() instead?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -74,6 +74,21 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
...
 
+static gboolean
+fu_plugin_inhibit_suspend (GError **error)
+{
+	g_autofree gchar *lockdir = NULL;
+	g_autofree gchar *inhibitsuspend_filename = NULL;
+        g_autofree gchar *getpid_str = NULL;
+
+	lockdir = fu_common_get_path (FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_filename =
```
bleungatchromium, line 85: Style nit. If you find you've run out of space because of 80-cols, you can intent as follows to line up the arguments with the (

```
	inhibitsuspend_filename = g_build_filename (lockdir,
						    "power_override",
						    "fwupd.lock",
						    NULL);

```

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -74,6 +74,21 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
...
+	    g_build_filename (lockdir, "power_override", "fwupd.lock", NULL);
+        getpid_str = g_strdup_printf ("%d", getpid ());
+	return g_file_set_contents (
```
bleungatchromium, line 88: Same styling comment as above regarding indenting to line up arguments with the (

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -133,5 +148,21 @@ fu_plugin_update_prepare (FuPlugin *plugin,
...
+fu_plugin_update_cleanup (FuPlugin *plugin,
+			  FwupdInstallFlags flags,
+			  FuDevice *dev,
+			  GError **error)
+{
+	g_autofree gchar *lockdir = NULL;
+	g_autofree GFile *inhibitsuspend_file = NULL;
+
+	lockdir = fu_common_get_path (FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_file =
```
bleungatchromium, line 164: Same styling comment as above. Line up comments with the (.

plugins/powerd/fu-plugin-powerd.c
djcampello: I think this may be coming from applying clang-format on https://github.com/fwupd/fwupd/issues/3505

djcampello: Actually is this true? I was hoping that ``BinPackArguments: false`` and ``BinPackParameters: false`` would deal with this.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -74,6 +74,21 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_plugin_inhibit_suspend (GError **error)
+{
+	g_autofree gchar *lockdir = NULL;
+	g_autofree gchar *inhibitsuspend_filename = NULL;
+        g_autofree gchar *getpid_str = NULL;
```
superm1, line 82: this whitespace looks wrong

superm1 (commented):
What happens if the update fails?  

I suspect the daemon needs to either:
1. Be changed to call update cleanup even if the update fails.
2. Add a new plugin vfunc that gets called in update_error.

superm1:
It seems like there is weird mix of tabs and spaces in this commit.  If your editor supports it, I'd suggest you turn on visible whitespace.  For example on vscode it shows -> for tabs and `····` for spaces.  So it's a lot easier to catch these types of problems.

If we do actually come up with a clang format for the project (like #3505 suggests) then these things will be less common, but for now I would suggest you check and fix it on your end.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -133,5 +152,38 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 			     MINIMUM_BATTERY_PERCENTAGE_FALLBACK);
 		return FALSE;
 	}
-	return TRUE;
+	return fu_plugin_inhibit_suspend (error);
+}
+
+gboolean
+fu_plugin_update_cleanup (GError **error)
```
bleungatchromium, line 159: I think you need to rename this function.
Given the context, I think you meant to call this `fu_plugin_delete_suspend_file`.

Although you might want to call it `fu_plugin_powerd_delete_suspend_file` to make it more specific.

bleungatchromium (changes requested):
Looks like you didn't implement `fu_plugin_delete_suspend_file`.

plugins/powerd/fu-plugin-powerd.c
binaryNewts: I'll get on that
I also forgot to do some checks and this doesn't build; I probably won't get changes to fix that in today.

binaryNewts:
> What happens if the update fails?
> 
> I suspect the daemon needs to either:
> 
> 1. Be changed to call update cleanup even if the update fails.
> 2. Add a new plugin vfunc that gets called in update_error.

As far as execution for this goes, I'm trying to do the second of these two. I'm having trouble figuring out where to call the new vfunc from, though. It seems as though they're run (automatically, maybe? unsure) from src/fu-engine.c, bu the update_error functions I could find were located in libfwupd/fwupd-device.c, where I don't have access to things like device that I'd want to be able to pass into the vfunc. Can I get some guidance on this?

hughsie:
> I'm trying to do the second of these two

I think I prefer the 1st of those two; running composite_cleanup on failure seems like a sensible thing to do anyway...

binaryNewts:
> I think I prefer the 1st of those two; running composite_cleanup on failure seems like a sensible thing to do anyway...

Why would we run composite_cleanup and not update_cleanup?
Also, I've got some concerns having to do with running either of those two in case of failure. Is there any chance that a plugin would have content in either of those two functions that should only be run in case of a completed update? I'm worried that a current or future plugin might have something in a cleanup function that would not execute properly or cause an error if the update does not fully complete.

hughsie:
> Why would we run composite_cleanup and not update_cleanup?

Maybe we should run both? The idea for the composite version is that it's typically doing:

 * composite_prepare -> switch dock device into high power mode
 * detach(1)
 * write_firmware(1)
 * attach(1)
 * cleanup(1)
 * detach(2)
 * write_firmware(2)
 * attach(2)
 * cleanup(2)
 * detach(3)
 * write_firmware(3)
 * attach(3)
 * cleanup(3)
 * composite_cleanup -> switch dock device into normal power mode

I don't think any of the `_cleanup()` methods run with the expectation that anything actually worked; they're cleaning up afterall.

superm1:
Dell dock assumes it was a successful update. It puts some devices into state to do more on unplug.

If composite cleanup is called all the time, I think it might need some thought if that is the right approach still. Cc @CragW

hughsie:
> Dell dock assumes it was a successful update

I had no idea, apologies for the misinformation. _Could_ it look at the update_state for that?

superm1:
> > Dell dock assumes it was a successful update
> 
> I had no idea, apologies for the misinformation. _Could_ it look at the update_state for that?

Yeah it could. But crag will need to decide if that's the right approach or not.

binaryNewts:
@hughsie would it be alright for me to try to make a new vfunc in the meantime and then perhaps migrate later to running the cleanups?

And if so, I still need help in the same capacity as earlier:
> I'm having trouble figuring out where to call the new vfunc from, though. It seems as though they're run (automatically, maybe? unsure) from src/fu-engine.c, bu the update_error functions I could find were located in libfwupd/fwupd-device.c, where I don't have access to things like device that I'd want to be able to pass into the vfunc. Can I get some guidance on this?

hughsie:
> would it be alright for me to try to make a new vfunc

I'm not overly keen on a new vfunc if we can fix an old one.

binaryNewts:
@superm1 it looks like composite_cleanup already gets called on error in fu_engine_install_tasks () (https://github.com/fwupd/fwupd/blob/master/src/fu-engine.c#L2115)
and update_cleanup is called on error in fu_engine_update () (https://github.com/fwupd/fwupd/blob/master/src/fu-engine.c#L3091)

superm1:
> @superm1 it looks like composite_cleanup already gets called on error in fu_engine_install_tasks () (https://github.com/fwupd/fwupd/blob/master/src/fu-engine.c#L2115)
> and update_cleanup is called on error in fu_engine_update () (https://github.com/fwupd/fwupd/blob/master/src/fu-engine.c#L3091)

TIL, thanks.

superm1:
pre-commit fails, can you please re-format changed code using clang?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -61,6 +79,19 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_plugin_inhibit_suspend(GError **error)
```
djcampello, line 83: Can we make this fu_plugin_powerd_create_suspend_file() and move it above the other function for symmetry

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -31,6 +45,10 @@ fu_plugin_destroy (FuPlugin *plugin)
 gboolean
 fu_plugin_startup (FuPlugin *plugin, GError **error)
 {
+	if (!fu_plugin_powerd_delete_suspend_file(error)) {
+		g_prefix_error(error, "lock file unable to be deleted");
+		return FALSE;
+	}
 	FuPluginData *data = fu_plugin_get_data (plugin);
```
djcampello, line 52: Declarations must go before any other statements

superm1: I was going to say why didn't something in CI catch this, but it did :)
```
../plugins/powerd/fu-plugin-powerd.c:52:2: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
```

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -97,5 +129,11 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 	/* start a timer to repeatedly run fu_plugin_powerd_refresh_cb at a set interval */
 	g_timeout_add_seconds(5, G_SOURCE_FUNC(fu_plugin_powerd_refresh_cb), plugin);
 
-	return TRUE;
+	return fu_plugin_powerd_create_suspend_file(error);
```
djcampello, line 132: This should not happen on startup but update_prepare

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -67,10 +94,17 @@ fu_plugin_powerd_refresh_cb(FuPlugin *plugin)
...
+	if (!file_absent) {
+		g_prefix_error(error, "lock file unable to be deleted");
+		return FALSE;
+	}
+
+	FuPluginData *data = NULL;
 	g_autofree gchar *name_owner = NULL;
 
 	/* establish proxy for method call to powerd */
+	data = fu_plugin_get_data(plugin);
```
djcampello, line 107: Why do you break this apart?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -67,10 +94,17 @@ fu_plugin_powerd_refresh_cb(FuPlugin *plugin)
...
 fu_plugin_startup(FuPlugin *plugin, GError **error)
 {
-	FuPluginData *data = fu_plugin_get_data(plugin);
+	gboolean file_absent = fu_plugin_powerd_create_suspend_file(error);
+	if (!file_absent) {
+		g_prefix_error(error, "lock file unable to be deleted");
+		return FALSE;
+	}
+
+	FuPluginData *data = NULL;
```
djcampello, line 103: Mixed declaration

djcampello (approved):

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -69,6 +96,11 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 {
 	FuPluginData *data = fu_plugin_get_data(plugin);
 	g_autofree gchar *name_owner = NULL;
+	gboolean file_absent = fu_plugin_powerd_create_suspend_file(error);
+	if (!file_absent) {
+		g_prefix_error(error, "lock file unable to be deleted");
+		return FALSE;
+	}
```
hughsie, line 103: *create* on startup? I can' kinda see the need for `fu_plugin_powerd_delete_suspend_file` but I can't see the need for the opposite. I'd just drop these 5 lines completely; if fwupd crashes half way thought updating you have bigger problems then a stale lock file..

binaryNewts: Agh that was supposed to be delete yeah. Is there a problem with having it there? Despite having bigger problems I feel like that is still a problem. Also, this is also useful for testing, when updates may purposefully stopped in the middle.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -69,6 +96,11 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 {
 	FuPluginData *data = fu_plugin_get_data(plugin);
 	g_autofree gchar *name_owner = NULL;
+	gboolean file_absent = fu_plugin_powerd_create_suspend_file(error);
+	if (!file_absent) {
+		g_prefix_error(error, "lock file unable to be deleted");
```
djcampello, line 101: Why do we prefix the error here and not in cleanup?

djcampello: We could have it as:
```
if (!fu_plugin_powerd_delete_suspend_file(error))
    return FALSE;
```

binaryNewts: I moved the error to the function itself, and added one in for create (didn't see the suggestion until just now). How do you feel about having it that way?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -20,6 +20,40 @@ fu_plugin_init(FuPlugin *plugin)
...
+	g_autofree gchar *lockdir = NULL;
+	g_autofree gchar *inhibitsuspend_filename = NULL;
+	g_autofree gchar *getpid_str = NULL;
+        gboolean file_present = FALSE;
+
+	lockdir = fu_common_get_path(FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_filename = g_build_filename(lockdir, "power_override", "fwupd.lock", NULL);
+	getpid_str = g_strdup_printf("%d", getpid());
+	file_present = g_file_set_contents(inhibitsuspend_filename, getpid_str, -1, error);
+        if (!file_present)
```
djcampello, line 35: Wrong formatting... Lets get the `contrib/setup` figured out so this stops happening.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -20,82 +20,137 @@ fu_plugin_init(FuPlugin *plugin)
...
 	}
 
-	/* start a timer to repeatedly run fu_plugin_powerd_refresh_cb at a set interval */
-	g_timeout_add_seconds(5, G_SOURCE_FUNC(fu_plugin_powerd_refresh_cb), plugin);
+	gboolean fu_plugin_startup(FuPlugin * plugin, GError * *error)
+	{
+		FuPluginData *data = fu_plugin_get_data(plugin);
+		g_autofree gchar *name_owner = NULL;
+		gboolean file_absent = fu_plugin_powerd_delete_suspend_file(error);
+		if (!file_absent)
```
djcampello, line 109: Just do `if (!fu_plugin_powerd_delete_suspend_file(error))`

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -69,6 +104,8 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 {
 	FuPluginData *data = fu_plugin_get_data(plugin);
 	g_autofree gchar *name_owner = NULL;
+	if (!fu_plugin_powerd_delete_suspend_file(error))
```
djcampello, line 107: nit picking: I would add an empty line between declarations and this

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -20,6 +20,41 @@ fu_plugin_init(FuPlugin *plugin)
...
+{
+	g_autofree gchar *lockdir = NULL;
+	g_autoptr(GFile) inhibitsuspend_file = NULL;
+	gboolean file_absent = TRUE;
+
+	lockdir = fu_common_get_path(FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_file =
+	    g_file_new_build_filename(lockdir, "power_override", "fwupd.lock", NULL);
+	if (g_file_query_exists(inhibitsuspend_file, NULL)) {
+		file_absent = g_file_delete(inhibitsuspend_file, NULL, error);
```
djcampello, line 51: What about (here and above): 
```
...{
if (!g_file_delete(inhibitsuspend_file, NULL, error)) {
    g_prefix_error(error, "lock file unable to be deleted");
    return FALSE;
}
}
return TRUE;
```

binaryNewts: If g_file_delete () can't delete, it returns false, but in the case that the file doesn't exist, it also raises an error. I added a check so that, if it returns false and the error is due to something other than the file not existing already, it will return false.

Documentation did suggest that using the query was not as good as just deleting and dealing with the error if it arises (in case something happens to the file between the check and the command).

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -20,6 +20,41 @@ fu_plugin_init(FuPlugin *plugin)
...
+	g_autoptr(GError) local_error = NULL;
+	g_autofree gchar *lockdir = NULL;
+	g_autoptr(GFile) inhibitsuspend_file = NULL;
+
+	lockdir = fu_common_get_path(FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_file =
+	    g_file_new_build_filename(lockdir, "power_override", "fwupd.lock", NULL);
+	if (!g_file_delete(inhibitsuspend_file, NULL, &local_error) &&
+	    !g_error_matches(local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND)) {
+		g_prefix_error(error, "lock file unable to be deleted");
```
djcampello, line 52: error here does not have the contents of local_error... I would revert to the previous logic as it may be easier to do the error handling

djcampello: I just read the Documentation cabeat... Then you can leave this as is but update error with local_error?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -20,6 +20,41 @@ fu_plugin_init(FuPlugin *plugin)
...
+	g_autoptr(GError) local_error = NULL;
+	g_autofree gchar *lockdir = NULL;
+	g_autoptr(GFile) inhibitsuspend_file = NULL;
+
+	lockdir = fu_common_get_path(FU_PATH_KIND_LOCKDIR);
+	inhibitsuspend_file =
+	    g_file_new_build_filename(lockdir, "power_override", "fwupd.lock", NULL);
+	if (!g_file_delete(inhibitsuspend_file, NULL, &local_error) &&
+	    !g_error_matches(local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND)) {
+		g_prefix_error(error, "%s", local_error->message);
```
djcampello, line 52: g_propagate_prefixed_error(error, local_error, "lock file unable to be deleted");

djcampello: Actually I think you need a steal pointer: 
		g_propagate_prefixed_error(error, g_steal_pointer(&error_local), "lock file unable to be deleted");

djcampello (approved):

PR #3491, opened 2021-07-14
emmc: Filter out rpmb and boot block devices

daniil-lunev:
The move to query sysattr for the name inadvertently changed the extent of
devices accepted by emmc probe. udev rules setup ID_NAME only for main
block device, but not for rpmb or boot, while 'name' sysattr is present
for all of them for they point to the same parent. The original check
for name was brittle either way and relied on a side effect of udev
subsystem. This CL introduces a proper filtering check that picks up only
the main devices. That is important because an attempt to perform an
update not on the main device returns a cryptic timeout error.

### DIFF
--- a/plugins/emmc/fu-emmc-device.c
+++ b/plugins/emmc/fu-emmc-device.c
@@ -140,6 +140,7 @@ fu_emmc_device_probe (FuDevice *device, GError **error)
 	g_autofree gchar *man_oem = NULL;
 	g_autofree gchar *man_oem_name = NULL;
 	g_autofree gchar *vendor_id = NULL;
+	g_autoptr(GRegex) dev_regex = NULL;
 
 	/* FuUdevDevice->probe */
 	if (!FU_DEVICE_CLASS (fu_emmc_device_parent_class)->probe (device, error))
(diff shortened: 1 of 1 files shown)


plugins/emmc/fu-emmc-device.c
```diff
@@ -164,6 +165,17 @@ fu_emmc_device_probe (FuDevice *device, GError **error)
 		return FALSE;
 	}
 
+	/* ignore *rpmb and *boot* mmc block devices */
+	dev_regex = g_regex_new("mmcblk\\d$", 0, 0, NULL);
+	if (!g_regex_match(dev_regex, g_udev_device_get_name (udev_device), 0, NULL)) {
```
hughsie, line 170: Space before the bracket please. Also do we need to check for name!=Null?

daniil-lunev: Fixed.
Checking for null is not strictly necessary  (https://fossies.org/linux/glib/glib/gregex.c line 1923), but better to have it, yes.

hughsie (approved):

PR #3492, opened 2021-07-14
Allow the daemon to request interactive action from the end user

hughsie:
The "return error and hope the client resubmits the firmware again"
pattern is clunky. There are two plugins doing this now, and about to
be one more.

This adds FwupdRequest which provides a structured way of asking the
user to perform an action, e.g. to replug the device or to press a
special key or button.

This replaces much of the UpdateMessage and UpdateImage API although
it is still used internally. Clients capable of processing the new
DeviceRequest signal should add REQUESTS to their feature flags.

Also, this allows us go back to the old meaning of _NEEDS_BOOTLOADER,
which was "needs rebooting into a bootloader mode" rather than the
slightly weird "user needs to do something and resubmit request".

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I like this a lot more than https://github.com/fwupd/fwupd/pull/3488

hardboprobot (approved):
Tested on the logitech-hidpp plugin (private repo) and it works as expected.

src/fu-engine.c
```diff
@@ -2848,6 +2882,29 @@ fu_engine_update_detach (FuEngine *self, const gchar *device_id, GError **error)
...
+	/* support older clients without the ability to do immediate requests */
+	if ((feature_flags & FWUPD_FEATURE_FLAG_REQUESTS) == 0 &&
+	    fu_device_get_request_cnt (device, FWUPD_REQUEST_KIND_IMMEDIATE) > 0) {
+
+		/* fallback to something sane */
+		if (fu_device_get_update_message (device) == NULL) {
+			g_autofree gchar *tmp = NULL;
+			tmp = g_strdup_printf ("Device %s needs to manually be put in update mode",
+					       fu_device_get_name (device));
+			fu_device_set_update_message (device, tmp);
```
superm1, line 2895: Is this true that all immediate requests are to go into update mode?  Or would there also be requests that are power cycle the device and the plugin continues to verify it?

hughsie: This is for `<screenshots><screenshot><caption>` which I think we specify specifically is for detach into bootloader. Perhaps we should just nuke the `Device %s needs to manually be put in update mode` prefix and put the entire string in the metadata?

superm1: Yeah that's better

hughsie: Okay, removed.

hughsie:
@superm1 I'm not 100% on the `request` noun; it seemed more descriptive than `event`. Other ideas welcome.

plugins/system76-launch/fu-system76-launch-device.c
```diff
@@ -96,31 +96,65 @@ fu_system76_launch_device_setup (FuDevice *device, GError **error)
...
+	fu_device_emit_request (device, request);
+
+	/* poll for the user-unlock */
+	fu_device_set_progress (device, 0);
+	do {
+		g_usleep (G_USEC_PER_SEC);
+		if (!fu_system76_launch_device_reset (device, &rc, error))
+			return FALSE;
+	} while (rc != 0 &&
+		 g_timer_elapsed (timer, NULL) * 1000.f < FU_DEVICE_REMOVE_DELAY_USER_REPLUG);
```
superm1, line 150: I wonder if we want a libfwupdplugin set of code to use here for the timer and loop in all plugins rather than them re-inventing their own thing?

superm1:
>@superm1 I'm not 100% on the request noun; it seemed more descriptive than event. Other ideas welcome.

User Request?  Client Request?  UserAction?  

>I like this a lot more than #3488

Yeah me to.

plugins/system76-launch/fu-system76-launch-device.c
hughsie: I kinda err on the rule of 3 for stuff like this; once three plugins are doing the same thing then add a library call and convert them. Until then it feels a bit like architecture astronauting.

hughsie:
>  Client Request

That gets confusing as it's talking about the client as in the thing that's initiated the action, not the PkClient thing.

>  UserAction

I don't think Action conveys that we need the user to _do_ something. Maybe `Request` is indeed best...

plugins/system76-launch/fu-system76-launch-device.c
superm1: Didn't you just say that exactly 3 were gonna need it?  :)

hughsie: Ohh, sorry. Three plugins that need to use fu_device_emit_request.

PR #3495, opened 2021-07-14
Markdown lint support

superm1:
So I think it's going to be a lot of busy work fixing up the rest, but this is probably the way to go to keep codacity analysis happy.

Type of pull request:
- [ ] Code fix
- [ ] Feature
- [x] Documentation


superm1:
Pre-commit CI goes from 20-30s up to 1m due to setup.  I'm not sure if maybe https://github.com/DavidAnson/markdownlint-cli2 is better, might be faster.

superm1:
Also, can we maybe turn off codacity looking at markdown?  I'm not sure we want to be so pedantic to their style.  I'd rather we control it to our own that we can have a conf file for (like the pre-commit I did, but maybe we still need more tweaking)

hughsie:
> Pre-commit CI goes from 20-30s up to 1m due to setup

I think that's fine.

> Also, can we maybe turn off codacity looking at markdown?

I think if it's being checked in precommit I think that's fine. I think we should turn off `MD013/line-length Line length` too.

FFY00:
> Pre-commit CI goes from 20-30s up to 1m due to setup.

But this is just in CI, which should be fine. Locally, pre-commit will only check modified files if you have the hook.

hughsie:
@superm1 you happy with the changes? I hope it's okay I fixed up the remaining issues.

.markdownlint.yaml
```diff
@@ -0,0 +1,2 @@
+MD013:
+  line_length: 1000
```
superm1, line 2: I don't think we need both yaml and json, it should be one or the other.

.pre-commit-config.yaml
```diff
@@ -48,3 +48,8 @@ repos:
       language: system
       entry: shellcheck --severity=error -e SC2068
       types: [shell]
+-   repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.27.1
+    hooks:
+    - id: markdownlint
+      args: ['--fix', '--config', '.markdownlint.yaml', '--ignore', '.github']
```
superm1, line 55: I don't think you need --config if you use json.
Also --ignore on .github is that right?

superm1 (commented):
Just a nit on the configuration so we don't end up with two copies in different formats

.markdownlint.yaml
hughsie: TIL about the existing json file. I've removed the YAML one.

.pre-commit-config.yaml
hughsie: The .github one is required as the pull_request_template.md isn't really a proper document; it has no title and some of the rules don't apply. If you _want_ to make it a proper document that's fine too.

FFY00: Btw, you can exclude directly in pre-commit, which will prevent pre-commit from trying to run the hook on those files.

https://github.com/pypa/build/blob/5a3c5243a2135236f207da3a4b6401e1561f952d/.pre-commit-config.yaml#L11

hughsie: oh, specific to markdownlint-cli? Got an example?

superm1: We run on --all-files for CI. So I think the exclusion you're doing is fine

Issue #3270, opened 2021-05-20
Compiling with test support fails


djcampello:
data/tests/missing-hwid/meson.build:1:0: ERROR: Unknown variable "gcab".

Adding to data/tests/colorhug/meson.build the following (extracted from meson.build) resolves the issue:
gcab = find_program('gcab', required : get_option('tests'))

The Meson build system
Version: 0.55.3

Probably unrelated but also present in the logs (regardless of test feature):
WARNING: Project targeting '>=0.47.0' but tried to use feature introduced in '0.50.0': install arg in configure_file.

superm1:
>Adding to data/tests/colorhug/meson.build the following (extracted from meson.build) resolves the issue:
gcab = find_program('gcab', required : get_option('tests'))

That's a bit confusing that this actually helps.  By it being in top level meson.build, it's supposed to be used elsewhere too.  Can you provide the full configure flags used to cause it so we can repro it?

>WARNING: Project targeting '>=0.47.0' but tried to use feature introduced in '0.50.0': install arg in configure_file.

This is expected; we intentionally check which version is in use and only turn those features on newer versions.

djcampello:
Hope the following works:

sys-apps/fwupd-9999::chromiumos was built with the following:
USE="abi_x86_64 bluetooth elibc_glibc flashrom gnutls gusb kernel_linux minimal nvme pkcs7 python_single_target_python3_6 synaptics test thunderbolt userland_GNU -agent -amt -archive -cros_host -cros_workon_tree_ -dell -elogind -gpg -gtk-doc -introspection -man -nls -policykit -python_single_target_python2_7 -python_single_target_python3_7 -systemd -uefi"

https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/master/sys-apps/fwupd/fwupd-9999.ebuild

hughsie:
>  By it being in top level meson.build, it's supposed to be used elsewhere too

That was also my conclusion. Weird!

superm1:
> Hope the following works:
> 
> sys-apps/fwupd-9999::chromiumos was built with the following:
[...]

PR #3497, opened 2021-07-14
fu-engine: Preserve NEEDS_REBOOT on successful update

daniil-lunev:
Currently the SUCCESS state overrides NEEDS_REBOOT, thus making the
devices considered for successive get-upgrades within same boot. The
change preserves the NEEDS_REBOOT state. It also adds a missed
'changed' signal, which otherwise would skip updating the persistent
state.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2502,9 +2502,6 @@ fu_engine_install_release (FuEngine *self,
 		fu_device_set_update_error (device, str);
 	}
 
-	/* success */
-	fu_engine_emit_changed (self);
-
 	return TRUE;
 }
 
@@ -2693,8 +2690,13 @@ fu_engine_install (FuEngine *self,
 			return FALSE;
 	}
 
-	/* success */
-	fu_device_set_update_state (device, FWUPD_UPDATE_STATE_SUCCESS);
+	/* mark success unless needs a reboot */
+	if (fu_device_get_update_state (device) != FWUPD_UPDATE_STATE_NEEDS_REBOOT)
+		fu_device_set_update_state (device, FWUPD_UPDATE_STATE_SUCCESS);
+
+	/* make the UI update */
+	fu_engine_emit_changed (self);
+
 	return TRUE;
 }
 
--- a/src/fu-self-test.c
+++ b/src/fu-self-test.c
@@ -1887,6 +1887,8 @@ fu_engine_history_inherit (gconstpointer user_data)
 	FuTest *self = (FuTest *) user_data;
 	gboolean ret;
 	g_autofree gchar *filename = NULL;
+	g_autofree gchar *localstatedir = NULL;
+	g_autofree gchar *history_db = NULL;
 	g_autoptr(FuDevice) device = fu_device_new_with_context (self->ctx);
 	g_autoptr(FuEngine) engine = fu_engine_new (FU_APP_FLAGS_NONE);
 	g_autoptr(FuInstallTask) task = NULL;
@@ -1896,6 +1898,11 @@ fu_engine_history_inherit (gconstpointer user_data)
 	g_autoptr(XbNode) component = NULL;
 	g_autoptr(XbSilo) silo_empty = xb_silo_new ();
 	g_autoptr(XbSilo) silo = NULL;
+
+	/* delete history */
+	localstatedir = fu_common_get_path (FU_PATH_KIND_LOCALSTATEDIR_PKG);
+	history_db = g_build_filename (localstatedir, "pending.db", NULL);
+	g_unlink (history_db);
 
 	/* no metadata in daemon */
 	fu_engine_set_silo (engine, silo_empty);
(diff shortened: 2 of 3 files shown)


hughsie:
Could you add a self test for this please, just so we never regress in the future? i.e. add something like a `FWUPD_DEVICE_FLAG_NEEDS_REBOOT` to `plugins/test/fu-plugin-test.c` and then add some extra checks in `fu_engine_history_func` in `src/fu-self-test.c`?

hughsie:
I think https://github.com/fwupd/fwupd/pull/3502 is the proper fix, no?

daniil-lunev:
Re: test - I have some troubles with actually compiling in the test on chromeos with 
"data/tests/colorhug/meson.build:1:0: ERROR: Unknown variable "gcab"." Daniel mentioned that such an issue has happened already at some point?
Re: fix - The state would have been overwritten unconditionally to success without the fix - at least in the in-memory state. The unknown state is not a concern for this particular case.

hughsie:
> ERROR: Unknown variable "gcab"."

I think you need `meson configure -Dtests=true` and also to include gcab.

Richard.

superm1:
> > ERROR: Unknown variable "gcab"."
> 
> I think you need `meson configure -Dtests=true` and also to include gcab.
> 
> Richard.

I think it's the same failure as @djcampello had turning on test support that we can't repro outside of chromeos build environment:
https://github.com/fwupd/fwupd/issues/3270

daniil-lunev:
Ok,  I have figured out all things. I could make the test the way you suggested thus I just created a new test for the thing specifically. PTAL

src/fu-engine.c
```diff
@@ -2691,8 +2691,14 @@ fu_engine_install (FuEngine *self,
 			return FALSE;
 	}
 
-	/* success */
-	fu_device_set_update_state (device, FWUPD_UPDATE_STATE_SUCCESS);
+	/* mark success unless needs a reboot */
+	if (fu_device_get_update_state (device) != FWUPD_UPDATE_STATE_NEEDS_REBOOT) {
```
superm1, line 2695: just a nit here: no need for braces for one line

src/fu-engine.c
```diff
@@ -2691,8 +2691,14 @@ fu_engine_install (FuEngine *self,
...
+		fu_device_set_update_state (device, FWUPD_UPDATE_STATE_SUCCESS);
+	}
+
+	/* make the UI update */
+	fu_engine_emit_changed (self);
```
superm1, line 2700: I think you want to remove this from `fu_engine_install_release` since you're moving it here.

src/fu-self-test.c
```diff
@@ -1998,6 +1998,89 @@ fu_engine_history_inherit (gconstpointer user_data)
...
+
+	/* check the device requires an activation */
+	g_assert_true (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_NEEDS_REBOOT));
+	g_assert_cmpint (fu_device_get_update_state (device), ==, FWUPD_UPDATE_STATE_NEEDS_REBOOT);
+	g_assert_cmpstr (fu_device_get_version (device), ==, "1.2.2");
+
+	/* delete files */
+	localstatedir = fu_common_get_path (FU_PATH_KIND_LOCALSTATEDIR_PKG);
+	history_db = g_build_filename (localstatedir, "pending.db", NULL);
+	g_unlink (history_db);
```
superm1, line 2081: Can you avoid even having a history database by using `FU_ENGINE_LOAD_FLAG_READONLY`?

src/fu-engine.c
daniil-lunev: Done

src/fu-engine.c
daniil-lunev: Done

src/fu-self-test.c
daniil-lunev: If I add the flag to the engine and remove the deletion, somehow it still fails exactly the same way on the next test as with NONE flag - by failing to find a device specified in history. May be I am missing something.

superm1: OK in that regard, I would rather move the cleanup to the start of the function to make this more deterministic.  This is how `fu_history_func` does it as well.

daniil-lunev: I am not sure if that would achieve the necessary effect for the problem comes from the added tests leaving some state which affects running the following one and thus being required to clean after itself, though I am not sure why that is not a problem in some other cases. I guess generically that is a problem of the tests not being hermetic, which is a problem in its own turn, but should be addressed separately.

superm1: > not being hermetic

Which is why I'm proposing you remove it at the beginning of the test instead.  The other tests that need it will do the same thing.

daniil-lunev: Ah, I see, you mean clean it up in the following test, I misunderstood. Done.

superm1 (dismissed):
Squash to one commit and it LGTM

daniil-lunev:
Done

PR #3500, opened 2021-07-15
Wip/hughsie/as json

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@djcampello this lets you do `fwupdmgr get-remotes --json`:

	{
	  "Remotes" : [
	    {
	      "Id" : "lvfs-testing",
	      "Kind" : "download",
	      "KeyringKind" : "jcat",
	      "ReportUri" : "https://fwupd.org/lvfs/firmware/report",
	      "MetadataUri" : "https://cdn.fwupd.org/downloads/firmware-testing.xml.gz",
	      "MetadataUriSig" : "https://cdn.fwupd.org/downloads/firmware-testing.xml.gz.jcat",
	      "Title" : "Linux Vendor Firmware Service (testing)",
	      "FilenameCache" : "/var/lib/fwupd/remotes.d/lvfs-testing/metadata.xml.gz",
	      "FilenameCacheSig" : "/var/lib/fwupd/remotes.d/lvfs-testing/metadata.xml.gz.jcat",
	      "FilenameSource" : "/etc/fwupd/remotes.d/lvfs-testing.conf",
	      "Enabled" : "false",
	      "ApprovalRequired" : "false",
	      "AutomaticReports" : "false",
	      "AutomaticSecurityReports" : "false",
	      "Priority" : 1,
	      "Mtime" : -1
	    },
	    ...

djcampello:
This looks good. 
@g2pop we can use this to adapt our test checking for plugin presence with `fwupdmgr get-plugins --json`

superm1 (commented):
rather than calling fwupdmgr binary, how about just making fwupdagent a symlink for now?

hughsie:
> how about just making fwupdagent a symlink for now

You need to add the --json argument unfortunately. Actually, maybe fwupdmgr can look at the argv0? i.e. will the argv0 be fwupdagent in this case?

superm1:
> > how about just making fwupdagent a symlink for now
> 
> You need to add the --json argument unfortunately. Actually, maybe fwupdmgr can look at the argv0? i.e. will the argv0 be fwupdagent in this case?

Yes, that seems like a better way to go if that works.  If not, then perhaps fwupdagent can be a simple shell script instead of another C app?

Also, I think various command completion, and man pages need to be updated for this change too.

hughsie:
The symlink thing works. Do you think that we should remove the bash completion and man pages? Both are still valid really.

superm1:
> The symlink thing works. Do you think that we should remove the bash completion and man pages? Both are still valid really.

Well I was going to suggest a single man page and bash completion file that work for both commands.

superm1:
--json needs to be added to fwupdmgr completions too.

hughsie:
I've fixed up the fwupdagent man page to indicate it's deprecated. I don't think we want the symlink the bash completion, I don't want people to start using `fwupdagent get-plugins` for instance.

src/meson.build
```diff
@@ -94,34 +94,13 @@ fwupdmgr = executable(
...
-    fwupd,
-    fwupdplugin,
-  ],
-  install : true,
-  install_dir : bindir
-)
+# for compatibility
+meson.add_install_script(
+  'sh', '-c',
+  'ln -fs fwupdmgr @0@@1@'.format(
```
superm1, line 101: did you try to just have it just store symlink in the tree and install the symlink instead of making it explicitly at install time?  I feel like this approach you did might be more fragile with packaging.

superm1 (approved):
other than the question on the symlink installation, looks great, thanks

src/meson.build
hughsie: I tried this; it gets horrible when trying to install into destdirs and things like that. See https://github.com/mesonbuild/meson/issues/1602

PR #3502, opened 2021-07-15
Never write 'unknown' update state on success

hughsie:
This means we don't get asked to submit a report.

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-device.c
+++ b/libfwupd/fwupd-device.c
@@ -78,6 +78,7 @@ enum {
 	PROP_PARENT,
 	PROP_UPDATE_STATE,
 	PROP_UPDATE_MESSAGE,
+	PROP_UPDATE_ERROR,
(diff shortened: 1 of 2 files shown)


superm1:
This feels like the wrong solution. Why is the status ever unknown?

hughsie:
I think it starts off in unknown, no?

superm1:
But shouldn't it get updates elsewhere?

superm1:
Let me elaborate.  `fu_engine_install_release` is only called from `fu_engine_install`, which calls that at the very end of the function _already_.

hughsie:
>  So if you're calling it here, it should no longer be called there

Hmm. Indeed. Will ponder, thanks.

hughsie:
Okay, I like this one much more ^^

superm1 (approved):

PR #3503, opened 2021-07-16
update-capsule: Move FuUefiDevice to a derivable for future use

hughsie:
Type of pull request:
- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -462,7 +462,7 @@ fu_plugin_uefi_capsule_coldplug_device (FuPlugin *plugin, FuUefiDevice *dev, GEr
 		fu_device_add_private_flag (FU_DEVICE (dev),
 					    FU_UEFI_DEVICE_FLAG_USE_LEGACY_BOOTMGR_DESC);
 	}
-	if (fu_plugin_has_custom_flag (plugin, "boot-order-lock")) {
+	if (fu_plugin_has_custom_flag (plugin, "supports-boot-order-lock")) {
 		fu_device_add_private_flag (FU_DEVICE (dev),
 					    FU_UEFI_DEVICE_FLAG_SUPPORTS_BOOT_ORDER_LOCK);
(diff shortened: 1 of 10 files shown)


hughsie:
The plan here would be to have FuUefiShimDevice and FuUefiCodDevice in the future. I'm not sure about the shim name, maybe FuUefiBootmgrDevice is more correct?

hughsie:
@apalos for visibility

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -513,6 +462,10 @@ fu_plugin_uefi_capsule_coldplug_device (FuPlugin *plugin, FuUefiDevice *dev, GEr
 		fu_device_add_private_flag (FU_DEVICE (dev),
 					    FU_UEFI_DEVICE_FLAG_USE_LEGACY_BOOTMGR_DESC);
 	}
+	if (fu_plugin_has_custom_flag (plugin, "boot-order-lock")) {
```
superm1, line 465: should this be `supports-boot-order-lock`?

hughsie: Nah, the existing plugin one doesn't have this -- we could certainly change the quirk file to match if you think it would be clearer.

superm1, line 465: and shouldn't it be using the private flag not the custom flag?

hughsie: **fu_plugin** - not fu_device

hughsie: I've made them both `supports-boot-order-lock` as it's confusing to have both.

apalos:
I don't know much about fwupd, but I've skimmed through the patches and they look reasonable for supporting the functionality we need for capsule update on disk.

Acked-by: Ilias Apalodimas <Ilias.Apalodimas@linaro.org>

Issue #3460, opened 2021-07-08
ThinkPad Thunderbolt 3 Workstation Dock Gen 1 & 2; firmware incompatible, got VL820Q8, expected VL820Q7


josiah47:
**Describe the bug**
sudo fwupdmgr update
results in error firmware incompatible, got VL820Q8, expected VL820Q7

**Steps to Reproduce**
sudo fwupdmgr get-updates
sudo fwupdmgr update
Restarting device…       [-                                      ]Devices with no available firmware updates: 
 • Thunderbolt host controller
 • UEFI Device Firmware
 • UEFI Device Firmware
 • USB2.0 Hub
 • USB2.0 Hub
 • USB3.0 Hub
Upgrade available for ThinkPad Thunderbolt 3 Dock from 56.00 to 62.00
ThinkPad Thunderbolt 3 Dock and all connected devices may not be usable while updating. Continue with update? [Y|n]: y
Downloading…             [***************************************]
Decompressing…           [***************************************]
Authenticating…          [***************************************]
Authenticating…          [***************************************]
Updating ThinkPad Thunderbolt 3 Dock USB Audio…******************]
Restarting device…       [***************************************]
Updating USB3.1 Hub…     [************************************** ]
Restarting device…       [************************************** ]firmware incompatible, got VL820Q8, expected VL820Q7

**Expected behavior**
Firmware to be applied

**fwupd version information**
Please provide the version of the daemon and client.
```shell
$ fwupdmgr --version
```
fwupdmgr --version
client version:	1.6.0
compile-time dependency versions
	gusb:	0.3.6

Restarting device…       [-                                      ]daemon version:	1.6.0

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
`sudo fwupdmgr update

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
$ fwupdmgr get-devices --show-all-devices
```
fwupdmgr get-devices --show-all-devices
Restarting device…       [-                                      ]20SUS34700
│
├─ThinkPad Thunderbolt 3 Dock:
│ │   Device ID:          11ef1614a9d17196a5e780378194750b11296379
│ │   Current version:    56.00
│ │   Vendor:             Lenovo (THUNDERBOLT:0x0108|TBT:0x0108)
│ │   GUIDs:              45c1814e-eb83-5b03-8997-ef8d4d8ade29
│ │                       03b97e38-f4b8-5a19-917c-28f95ca29768
[...]

PR #3504, opened 2021-07-16
vli: Fix a regression when installing VL820Q7 firmware

hughsie:
This regressed in c255034574c725583afac4a6c7eed901b8ac2d7f which caused
ThinkPad Thunderbolt 3 Workstation Dock updates to fail.

Fixes https://github.com/fwupd/fwupd/issues/3460

Type of pull request:
- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
these seem like totally different offsets being read?  but no concerns.

hughsie:
Ya, I just used the old working one.

PR #3513, opened 2021-07-19
Wip/hughsie/redfish rhel

hughsie:
Type of pull request:

- [X] Code fix
- [X] Feature
- [ ] Documentation


plugins/redfish/fu-plugin-redfish.c
```diff
@@ -122,19 +140,47 @@ fu_redfish_plugin_discover_smbios_table (FuPlugin *plugin, GError **error)
...
+		device_path = fu_redfish_plugin_discover_device_path (plugin);
+		if (device_path != NULL) {
+			device = fu_redfish_network_device_for_path (device_path, error);
+			if (device == NULL)
+				return FALSE;
+		}
+	}
+
+	/* autoconnect device if required */
+	if (device != NULL) {
```
superm1, line 171: shouldn't the device == NULL be a failure here that you return false?

hughsie: I don't think so; I think if you just specify an IP address in the SMBIOS blob and not a MAC or VID/PID. I've got a fixup patch I can push on Thursday that removes the RedfishNetworkDevicePath as Lenovo pointed out the VID/PID parsing in the plugin was wonky. Until then I'm in daddy mode for a couple of days.

hughsie:
I confirmed this works on a Lenovo server with RHEL 8.5 installed.

PR #568, opened 2018-06-20
Add clang formating

superm1:
In an effort to try to keep style consistent (and be helpful to new submitters) I want to keep a .clang-format that does things at least mostly right.  Personally I'd like to use it with VS code's automatic clang format on save functionality.

There are some things I know clang format wont' be able to get right (such as `#define` block alignment; see https://stackoverflow.com/questions/38620019/can-clang-format-align-a-block-of-defines-for-me) and for those I'd like to just use `//clang-format off //clang-format` on wrappers (see https://philwrightdev.wordpress.com/2014/10/01/making-clang-format-ignore-sections/)

So here's a first pass at a `.clang-format` and an example of what it does to one of the plugins.  I used https://zed0.co.uk/clang-format-configurator/ to build it and I didn't use any //clang-format controls to block it from running on anything.

Thoughts?

My hope is that we can come up with something that looks good enough, run it on the whole project to get good starting points and then adopt exceptions when needed.

PR #3514, opened 2021-07-20
Wip/djcampello/clang format

djcampello:
Reviving https://github.com/fwupd/fwupd/pull/568

> In an effort to try to keep style consistent (and be helpful to new submitters) I want to keep a .clang-format that does things at least mostly right. Personally I'd like to use it with VS code's automatic clang format on save functionality.
> 
> There are some things I know clang format wont' be able to get right (such as #define block alignment; see https://stackoverflow.com/questions/38620019/can-clang-format-align-a-block-of-defines-for-me) and for those I'd like to just use //clang-format off //clang-format on wrappers (see https://philwrightdev.wordpress.com/2014/10/01/making-clang-format-ignore-sections/)
> 
> So here's a first pass at a .clang-format and an example of what it does to one of the plugins. I used https://zed0.co.uk/clang-format-configurator/ to build it and I didn't use any //clang-format controls to block it from running on anything.
> 
> Thoughts?
> 
> My hope is that we can come up with something that looks good enough, run it on the whole project to get good starting points and then adopt exceptions when needed.


superm1:
If we do end up going in this direction, I think we'll want to possibly glue it to pre-commit, but that is going to also mean biting the bullet of a project wide commit of either:
1. fixing up formatting for areas we didn't follow our _own style_ or were inconsistent.
2. various headers added to code to exclude running the reformatter

superm1:
>So here's the problem. Clang format does have a way to correct the structs but it will have two limitations:

>it also impacts other code in functions. I think this is actually OK - it does a pretty good job.
>Pointer alignment gets out of wack. It's fixed with https://reviews.llvm.org/D27651 but that's not yet merged into clang-format upstream even.
>You can see what happens with that turned on in above new sample.

> It seems this has been fixed with https://reviews.llvm.org/D103245

Can you check what version of clang format that's in?  We may want to only run the pre-commit and hook with a new enough version.

plugins/dell/fu-plugin-dell.h
```diff
@@ -34,8 +37,8 @@ struct tpm_status {
...
-#define TPM_EN_MASK	0x0001
-#define TPM_OWN_MASK	0x0004
-#define TPM_TYPE_MASK	0x0F00
-#define TPM_1_2_MODE	0x0001
-#define TPM_2_0_MODE	0x0002
+#define TPM_EN_MASK 0x0001
+#define TPM_OWN_MASK 0x0004
+#define TPM_TYPE_MASK 0x0F00
+#define TPM_1_2_MODE 0x0001
+#define TPM_2_0_MODE 0x0002
```
YehezkelShB, line 44: This requires `clang-format off` to protect the current formatting, isn't it?

djcampello: I just check that this is now supported on clang-format-9.0 with AlignConsecutiveMacros: true

> > So here's the problem. Clang format does have a way to correct the structs but it will have two limitations:
> 
> > it also impacts other code in functions. I think this is actually OK - it does a pretty good job.
> > Pointer alignment gets out of wack. It's fixed with https://reviews.llvm.org/D27651 but that's not yet merged into clang-format upstream even.
> > You can see what happens with that turned on in above new sample.
> 
> > It seems this has been fixed with https://reviews.llvm.org/D103245
> 
> Can you check what version of clang format that's in? We may want to only run the pre-commit and hook with a new enough version.

This should be on clang-format-13 onwards

djcampello:
@bleungatchromium @g2pop @binaryNewts for visibility

.clang-format
```diff
@@ -0,0 +1,43 @@
...
+AlignOperands: 'true'
+AlignTrailingComments: 'true'
+AllowAllParametersOfDeclarationOnNextLine: 'false'
+AllowShortBlocksOnASingleLine: 'false'
+AllowShortCaseLabelsOnASingleLine: 'false'
+AllowShortFunctionsOnASingleLine: 'Inline'
+AllowShortIfStatementsOnASingleLine: 'false'
+AlwaysBreakAfterReturnType: 'All'
+BinPackParameters: 'false'
+BinPackArguments: 'true'
```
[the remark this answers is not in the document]
djcampello: never mind. This is probably better and more consistent with current state

contrib/reformat-code.py
```diff
@@ -0,0 +1,73 @@
...
+#
+
+import argparse
+import os
+import sys
+import subprocess
+
+CLANG_FORMATTERS = [
+    "clang-format-13",
+    "clang-format",
```
superm1, line 15: I think we explicitly only want 13+ right?

contrib/reformat-code.py
```diff
@@ -0,0 +1,73 @@
...
+        sys.exit(1)
+    lines = None
+    with open(f, "r") as rfd:
+        lines = rfd.readlines()
+    with open(f, "w") as wfd:
+        for line in lines:
+            for fixup in FIXUPS:
+                if fixup in line:
+                    line = line.replace(fixup, FIXUPS[fixup])
+            wfd.write(line)
```
superm1, line 55: is this going to cause the file time to change every time?  I wonder if the file should just be cached and only spit out if a fixup was run.

contrib/reformat-code.py
djcampello: We do. I believe normally the second one should not exist. At our Google environment we do have another installation coming from our repos and installed at `/usr/bin/clang-format`.

contrib/reformat-code.py
djcampello: Does this matter? If the file does not change git would not stage it

superm1: Doesn't it cause mtime changes to open and write it out even if it's identical?

djcampello: As an example: `touch COMMITMENT; git add COMMITMENT` yields no staged changes

hughsie: hmm, but it makes my editor reload the file. can we add a save = True if `if fixup in line` and then only write if that's set?

plugins/dell/fu-dell-smi.c
```diff
@@ -27,7 +27,7 @@ static void
 _dell_smi_obj_free (FuDellSmiObj *obj)
 {
 	dell_smi_obj_free (obj->smi);
-	g_free(obj);
+	g_free (obj);
```
hughsie, line 30: Lets drop the reformatting for this PR. We can apply manually it on new plugins or when we refactor files.

contrib/reformat-code.py
djcampello: Sadly, the clang-format will introduce the space on every run for the script to remove it after. I'll try to get around this by not having clang-format to replace the file in place

plugins/dell/fu-dell-smi.c
djcampello: Done

contrib/reformat-code.py
djcampello: Done

superm1 (dismissed):
This looks good to me, but something else I want to see happen (_perhaps separate from this PR_) is the ability to have pre-commit call this python script with an argument for the file to examine (IIRC that's how pre-commit works), and also the ability to skip it if it's making too many mistakes (maybe an environment variable?)

As seen in the example commit (that was reverted) although the style is mostly right, we don't want to have commits to change the whole project overnight and should do it over time.  This will also help refine the `.clang-format` and `reformat-code.py` for any style changes we miss from the example run.

IOW I want it to run only on changed/new files, so it would need to be excluded from CI.  This can be accomplished by modifying `.github/workflows/main.yaml` to remove the pre-commit lines for it.

djcampello:
Should I squash all of these in one to merge?

hughsie:
> Should I squash all of these in one to merge?

Please.

hughsie:
I'm not so fussed about the precommit bit; maybe we can test a little on the existing code.

superm1:
> I'm not so fussed about the precommit bit; maybe we can test a little on the existing code.

Yeah, sure let's worry about that separately.

superm1 (approved):

djcampello:
It looks like we can accomplish a partial format with `git diff | clang-format-diff-11 -p1`... I'm not sure how can we go about doing partial fixups

djcampello:
Even better: `git clang-format`

djcampello:
`git diff | clang-format-diff-11 -p1 | sed '/^+/s/g_autoptr (/g_autoptr(/g;/^+/s/sizeof (/sizeof(/g;/^+/s/g_auto (/g_auto{/g' | patch -p0`

superm1:
> `git diff | clang-format-diff-11 -p1 | sed '/^+/s/g_autoptr (/g_autoptr(/g;/^+/s/sizeof (/sizeof(/g;/^+/s/g_auto (/g_auto{/g' | patch -p0`

Nice

PR #3515, opened 2021-07-20
Wip/hughsie/powerd

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/powerd/fu-plugin-powerd.c
+++ b/plugins/powerd/fu-plugin-powerd.c
@@ -9,8 +9,6 @@
 
 #include <fwupdplugin.h>
 
-#define MINIMUM_BATTERY_PERCENTAGE_FALLBACK 10
-
 struct FuPluginData {
 	GDBusProxy	*proxy; /* nullable */
 };
@@ -68,6 +66,7 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 			  FuDevice *device,
 			  GError **error)
 {
+	FuContext *ctx = fu_plugin_get_context (plugin);
 	FuPluginData *data = fu_plugin_get_data (plugin);
 	guint32 power_type = 0;
 	guint32 current_state = 0;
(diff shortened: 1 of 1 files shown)


bleungatchromium:
@g2pop and @binaryNewts as a fyi.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -34,40 +32,28 @@ gboolean
...
-	data->proxy = g_dbus_proxy_new_sync (connection,
-					     G_DBUS_PROXY_FLAGS_NONE,
-					     NULL,
-					     "org.chromium.PowerManager",
-					     "/org/chromium/PowerManager",
-					     "org.chromium.PowerManager",
-					     NULL,
-					     error);
+	/* establish proxy for method call to powerd */
+	data->proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
```
bleungatchromium, line 38: @g2pop and @binaryNewts, was there a reason to use the `g_dbus_proxy_new` instead of `g_dbus_proxy_new_for_bus_sync` ? 

I admit I'm not 100% familiar with the difference.

hughsie: The only use of getting the connection separately is if you want to do low level filtering, or watch for the closed signal -- but for a system bus that's realistically never going to happen. From a speed point of view it's probably unmeasurable, as we'll just be getting a copy of the singleton instance -- it's just less LoC.

g2pop: Right, like @hughsie implied, g_dbus_proxy_new_for_bus_sync () basically combines the previous two functions, g_bus_get_sync () (connecting to a bus) and g_dbus_proxy_new_sync () (making a proxy for it). It was easier to see the steps and know what was happening when they were separated (like in https://github.com/fwupd/fwupd/blob/HEAD/libfwupdplugin/fu-common-linux.c#L34), but there's no real difference except for what @hughsie mentioned.

djcampello (approved):

bleungatchromium (approved):

djcampello:
I should have not merged this as is...

It is failing since it should compare against FU_BATTERY_VALUE_INVALID before:

```
failed to update_prepare using powerd: Cannot install update when system battery is not at least 101% unless forced
```

bleungatchromium:
> I should have not merged this as is...
> 
> It is failing since it should compare against FU_BATTERY_VALUE_INVALID before:
> 
> ```
> failed to update_prepare using powerd: Cannot install update when system battery is not at least 101% unless forced
> ```

Yeah, it looks like we need to actually `fu_context_set_battery_threshold` in the _setup phase, like upower does.

@g2pop, can you help take care of that?

PR #3517, opened 2021-07-21
SMBIOS: try reading from /sys/class/dmi if direct access fails

### LINKED ISSUE #3511: Reading SMBIOS information requires root
When fwupd is not running as root on Linux, it cannot read Hwid values from SMBIOS because Linux restricts access to the SMBIOS to the root user. This gracefully fails, but devices that need to key quirks on Hwid values will not work.
Concretely, the `parade_lspcon` plugin needs some Hwid values to know how to talk to a device and on Chromium OS the daemon does not run as root so it can't complete probing the device.

Linux exposes a subset of parsed values from SMBIOS in file in `/sys/class/dmi/id` which are mostly world-readable (some device-specific IDs are still restricted), so I'd like to add support for reading values from there if direct access to the SMBIOS tables fails.

tari:
The raw SMBIOS tables that Linux exposes in /sys/firmware/dmi
are restricted to being readable by root only. If running as
non-root access is still permitted by fields that have been
pre-parsed by the kernel in /sys/class/dmi, most of which are
world-readable. This allows the daemon to load most HWIDs even
if running as a non-root user, as is done on Chromium OS.

Fixes #3511.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

cc @djcampello


libfwupdplugin/fu-smbios.c
```diff
@@ -567,8 +709,18 @@ fu_smbios_setup (FuSmbios *self, GError **error)
 
 	/* DMI */
 	path = g_build_filename (sysfsfwdir, "dmi", "tables", NULL);
-	if (g_file_test (path, G_FILE_TEST_EXISTS))
-		return fu_smbios_setup_from_path (self, path, error);
+	if (g_file_test (path, G_FILE_TEST_EXISTS)) {
+		if (fu_smbios_setup_from_path (self, path, error))
```
superm1, line 713: I think you want to actually use a local error here.

libfwupdplugin/fu-smbios.c
```diff
@@ -567,8 +709,18 @@ fu_smbios_setup (FuSmbios *self, GError **error)
...
+			return TRUE;
+		/* try values the kernel parsed if failed to read raw tables; these
+		 * are world-readable whereas the raw tables are not */
+		if (g_file_test (path_dmi_class, G_FILE_TEST_IS_DIR)) {
+			g_debug ("trying to read %s: %s", path_dmi_class, (*error)->message);
+			g_clear_error (error);
+			return fu_smbios_setup_from_kernel (self, path_dmi_class, error);
+		}
+		return FALSE;
```
superm1, line 722: and if you use a local error, you'll need to propagate it here

libfwupdplugin/fu-smbios.c
```diff
@@ -341,6 +341,147 @@ fu_smbios_setup_from_file (FuSmbios *self, const gchar *filename, GError **error
...
+		item->buf = g_byte_array_new ();
+		item->strings = g_ptr_array_new_with_free_func (g_free);
+		g_ptr_array_add (self->items, item);
+	}
+
+	/* parse every known field from the corresponding file */
+	for (gsize i = 0; i < (sizeof (KERNEL_DMI_FIELDS) / sizeof (*KERNEL_DMI_FIELDS)); i++) {
+		const struct kernel_dmi_field *field = &KERNEL_DMI_FIELDS[i];
+		g_autofree gchar *path = g_build_filename (class_path, field->name, NULL);
+		GError *local_error = NULL;
```
superm1, line 454: this should be a `g_autoptr`, and then no need to clear the error each time in the error scenarios.

libfwupdplugin/fu-smbios.c
tari: I'd prefer if we could chain errors, but added a local error to prefix the original one and indicate what was tried.

libfwupdplugin/fu-smbios.c
tari: Done.

hughsie:
Looks good, but can we add some tests for this please? i.e. duplicate a /sys/class/dmi in data/tests and then read out the right values?

tari:
Added a test exercising string and integer values.

PR #3519, opened 2021-07-21
upower: Move battery_threshold setting to fu-engine

djcampello:


plugins/powerd/fu-plugin-powerd.c
```diff
@@ -57,6 +62,28 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 			     g_dbus_proxy_get_name (data->proxy));
 		return FALSE;
 	}
+
+	battery_str = fu_plugin_get_config_value (plugin, "BatteryThreshold");
+	if (battery_str == NULL) {
```
hughsie, line 67: Why the config file override?

djcampello: This was basically copied as is from fu_plugin_upower.c. I can remove if needed

hughsie: Are you copying from upower plugin from master?

hughsie: Also, maybe we should move the fu_context_set_battery_threshold() stuff into the daemon? It seems a bit silly to duplicate it.

bleungatchromium: Hey @djcampello, it looks like the version of fu-plugin-upower.c you've brought this over from doesn't have this latest commit:
https://github.com/fwupd/fwupd/commit/6f831903926dcdf201328fb450dd17c5ef0d8ecf

The most recent one is cleaner.

djcampello: I just fixed it on the recent push... Where would you think this should go in the daemon?

hughsie (approved):

Issue #2900, opened 2021-02-18
EFI Capsule On Disk Support


jackpot51:
Capsule On Disk is defined in the UEFI Specification, section 8.5.5, Delivery of Capsules via file on Mass Storage device. It is being used on many new designs, such that UpdateCapsule often is a wrapper around it. This is due to memory persistence issues when unlocking the flash and resetting.

A new OsIndications bit is defined in the specification:

```
#define EFI_OS_INDICATIONS_FILE_CAPSULE_DELIVERY_SUPPORTED 0x0000000000000004
```

The OS does the following:
- Check the bit in OsIndicationsSupported, use UpdateCapsule if that is supported instead
- Put signed capsule in \EFI\UpdateCapsule\(anything)
- Set the bit in OsIndications
- Reboot

The firmware then takes over and does the following:
- Detects the bit in OsIndications, unlocks, and resets
- Processes the capsules in case ignored alphabetic order
- Sets Capsule0000, Capsule0001, etc. to an EFI_STATUS for each capsule as it finishes
- Deletes the capsule when it is finished
- Reboots to the OS

The OS can now check the Capsule0000 and so on variables for the result of firmware flashing

superm1:
So in this case, we actually wouldn't use `fwupdx64.efi` for updates in your mind of that bit is supported?

jackpot51:
Right, and it would reduce the number of reboots by one on supported systems. The reason to support this, for me at least, is we are planning to have System76 open firmware products use capsule updates, but do not want to include the extra code to support UpdateCapsule when it would just be a wrapper around capsule on disk.

superm1:
Sounds pretty sensible to me.  Some implementation suggestions for when you submit a PR for this:
1. The spec says it's up to the OS to decide what to name the files on the ESP, so I would say we want the prefix of `fwupd` included in the file name going to to the ESP.
2. I think we'll want to have a boolean in the device attribute to indicate UEFI devices are picking this (IE here: https://github.com/fwupd/fwupd/blob/master/plugins/uefi-capsule/fu-uefi-device.c#L39)
3. You'll want to include debugging in the `to_string` here: https://github.com/fwupd/fwupd/blob/master/plugins/uefi-capsule/fu-uefi-device.c#L126 to indicate this is being used

martinezjavier:
[...]

PR #3524, opened 2021-07-22
uefi-capsule: Add support for CapsuleOnDisk

hughsie:
Based on a patch by Ilias Apalodimas <ilias.apalodimas@linaro.org>, many thanks.

Fixes https://github.com/fwupd/fwupd/issues/2900

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -15,7 +15,7 @@
 #include "fu-uefi-bgrt.h"
 #include "fu-uefi-bootmgr.h"
 #include "fu-uefi-common.h"
-#include "fu-uefi-device.h"
+#include "fu-uefi-cod-device.h"
```
hughsie, line 18: @superm1 maybe fu-uefi-disk-device.h might be clearer? Or more unclear perhaps?

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+static gboolean
+fu_uefi_cod_device_write_firmware (FuDevice *device,
+				   FuFirmware *firmware,
+				   FwupdInstallFlags flags,
+				   GError **error)
+{
+	FuUefiDevice *self = FU_UEFI_DEVICE (device);
+	guint64 os_indications = EFI_OS_INDICATIONS_FILE_CAPSULE_DELIVERY_SUPPORTED;
```
apalos, line 103: you need to read OsIndications before seeting it and OR it

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,157 @@
...
+		return FALSE;
+	basename = g_strdup_printf ("fwupd-%s.cap", fu_uefi_device_get_guid (self));
+	cod_path = g_build_filename (esp_path, "UpdateCapsule", basename, NULL);
+	if (!fu_common_mkdir_parent (cod_path, error))
+		return FALSE;
+	if (!fu_common_set_contents_bytes (cod_path, fw, error))
+		return FALSE;
+
+	/* assume we can set the runtime variable */
+	if (!fu_efivar_set_data (FU_EFIVAR_GUID_EFI_GLOBAL,
```
apalos, line 130: if the efivarfs is mounted as r/o we need to ignore the setvariable call and just copy the capsule.  
U-Boot has a config option that ignores OsIndications before applying the update.
On the next reboot the Capsule0000 efi variable will be missing, so that's our hint the update didn't run at all.

plugins/uefi-capsule/fu-uefi-backend-linux.c
```diff
@@ -70,15 +76,53 @@ fu_uefi_backend_device_new (const gchar *path)
...
+	g_autofree guint8 *buf = NULL;
+
+	if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "OsIndicationsSupported",
+				 &buf, &bufsz, NULL, error)) {
+		g_prefix_error (error, "failed to read EFI variable: ");
+		return FALSE;
+	}
+	if (!fu_memcpy_safe ((guint8 *) &value, sizeof(value), 0x0,	/* dst */
+			      buf, bufsz, 0x0, sizeof(value), error))
+		return FALSE;
```
apalos, line 100: Why do we need to copy the value here? Can't we just use buf?

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
superm1: cod is fine

plugins/uefi-capsule/fu-uefi-backend-linux.c
```diff
@@ -70,15 +76,53 @@ fu_uefi_backend_device_new (const gchar *path)
 	return g_steal_pointer (&dev);
 }
 
+void
+fu_uefi_backend_set_enable_cod (FuBackend *backend, gboolean enable_cod)
+{
+	FuUefiBackend *self = FU_UEFI_BACKEND (backend);
+	self->enable_cod = !enable_cod;
```
superm1, line 83: inverting a negative we get from the config file makes this confusing, can we just set the disable instead?

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,157 @@
...
+	gsize bufsz = 0;
+	guint32 status = 0;
+	guint32 total_size = 0;
+	g_autofree gchar *err_msg = NULL;
+	g_autofree gchar *guidstr = NULL;
+	g_autofree gchar *version_str = NULL;
+	g_autofree guint8 *buf = NULL;
+
+	/* lets assume we just did one update */
+	if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, "Capsule0000",
```
superm1, line 36: seems like a broad assumption, would it make more sense to store the variable name that was used into the history database or something?

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,157 @@
...
+				     "EFI_CAPSULE_RESULT_VARIABLE_HEADER too small");
+		return FALSE;
+	}
+
+	/* verify guid */
+	if (!fu_memcpy_safe (guid, sizeof(guid), 0x0,		/* dst */
+			     buf, bufsz, 0x08, sizeof(guid), error))
+		return FALSE;
+	guidstr = fwupd_guid_to_string (&guid, FWUPD_GUID_FLAG_MIXED_ENDIAN);
+	if (g_strcmp0 (guidstr, fu_uefi_device_get_guid (device_uefi)) != 0) {
```
superm1, line 58: this seems to be asking for a struct definition of what this header format is supposed to be looking like.  then you can do the check with `sizeof(CapsuleResultVariableHeader)` and access from `header->guid` and such.

plugins/uefi-capsule/fu-uefi-cod-device.c
apalos: You can do easier than that. 
Start a loop and append the 'i' to Capsule.  So you'll read
Capsule0001
Capsule0002 etc, 
Once you fail to read a variable, you know your parsing needs to stop

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,157 @@
...
+	if (total_size < 0x3A) {
+		g_set_error_literal (error,
+				     FWUPD_ERROR,
+				     FWUPD_ERROR_INTERNAL,
+				     "EFI_CAPSULE_RESULT_VARIABLE_HEADER too small");
+		return FALSE;
+	}
+
+	/* verify guid */
+	if (!fu_memcpy_safe (guid, sizeof(guid), 0x0,		/* dst */
```
superm1, line 54: if you already verified the size is big enough to contain the GUID, do you really need the same memory copy function?

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: Fixed, thanks.

plugins/uefi-capsule/fu-uefi-backend-linux.c
hughsie: I was wrong, memcpy isn't endian-clean. Fixed.

plugins/uefi-capsule/fu-uefi-backend-linux.c
hughsie: whoops, that inversion was a typo.

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: That's not endian clean tho; although I admit in this context LE is probably the only thing we care about.

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: No, but is saves the security team worrying that it's a random memcpy. i.e. for ~3 integer comparisons we don't even need to think about "is this safe" even if we remove the total length ready in the future.

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: I've added this check.

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: I've added something like this; can you verify it does the right thing? We might need to add another error in the `g_error_matches()` block.

apalos: @hughsie, sure I'll try to run the whole thing tomorrow, but it seems to be doing the right thing

apalos: You can also read the CapsuleLast variable (which will contain the name of the last applied capsule e.g Capsule0001) and figure out the size of 'idx'. The later is more 'efish' but both will work.

I did manage to run the whole thing and the ro/rw scanning of efivarfs is working correctly. The capsules are installed and executed and Capsule0000 is creted.  Here's the get-history output 

`Loading…                 [-                                      ]07:01:34:0315 FuEngine             failed to add device /sys/devices/platform/emmc2bus/fe340000.emmc2/mmc_host/mmc0/mmc0:0007/block/mmcb )
Unknown Product
│
└─Unknown Firmware:
  │   Device ID:          a45df35ac0e948ee180fe216a5f703f32dda163f
  │   Previous version:   0
  │   Update State:       Success
  │   Last modified:      2021-07-23 06:41
  │   GUID:               e2bb9c06-70e9-4b14-97a3-5a7913176e3f
  │   Device Flags:       • Internal device
  │                       • Updatable
  │                       • System requires external power source
  │                       • Needs a reboot after installation
  │                       • Device is usable for the duration of the update
  │ 
  └─  New version:      1.0.0
        License:          Unknown
`

apalos:
@hughsie one of the things that I missed on my initial patchset due to my setup is the BootNext potential usage.  The EFI spec says 
`The directory \EFI\UpdateCapsule is checked for capsules only within the EFI system partition on the device specified in the active boot option determine by reference to BootNext variable or BootOrdervariable processing. The active Boot Variable is the variable with highest priority BootNext or within BootOrder that refers to a device found to be present. Boot variables in BootOrder but referring to devices not present are ignored when determining active boot variable. `

So what this means is if you Boot####(0000, 0001 etc) variables point to the ESP partition that contains the 'EFI/Updatecapsule' directory, then your capsule update on-disk will run on next reboot and we don't have to do anything.  However if the Boot#### variables point outside that directory, you need to create a Boot#### variable that points to it (even if if doesn't contain any executable) and set BootNext to that boot variable.

Given the SetVariable limitations in U-Boot, I think we need to document that.  It's up to you if you want to add the extra functionality, but we could always do it in a follow up patch?

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: CapsuleLast contains the string Capsule0001? Could you please upload a tarball with the exact Capsule0001 and CapsuleLast contents and I'll add them to git and add some tests.

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,190 @@
...
+	g_autofree gchar *err_msg = NULL;
+	g_autofree gchar *guidstr = NULL;
+	g_autofree gchar *name = NULL;
+	g_autofree gchar *version_str = NULL;
+	g_autofree guint8 *buf = NULL;
+
+	/* lets assume we just did one update */
+	name = g_strdup_printf ("Capsule%04u", idx);
+	if (!fu_efivar_get_data (FU_EFIVAR_GUID_EFI_GLOBAL, name,
+				 &buf, &bufsz, NULL, error)) {
```
apalos, line 39: FU_EFIVAR_GUID_EFI_GLOBAL is not valid for Capsule#### we'll need EFI_CAPSULE_REPORT_GUID (from the EFI spec)

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,190 @@
...
+	fu_device_set_update_error (device, err_msg);
+	return TRUE;
+}
+
+static gboolean
+fu_uefi_cod_device_get_results (FuDevice *device, GError **error)
+{
+	/* this should be enough for any system */
+	for (guint i = 0; i < 1024; i++) {
+		g_autoptr(GError) error_local = NULL;
```
apalos, line 103: The CapsuleLast will have the last capsule we updated, so instead of running the loop from 0 here, we must read CapsuleLast (which will have somelike like Capsule0002, and start from 2

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,190 @@
...
+	}
+	fu_device_set_update_error (device, err_msg);
+	return TRUE;
+}
+
+static gboolean
+fu_uefi_cod_device_get_results (FuDevice *device, GError **error)
+{
+	/* this should be enough for any system */
+	for (guint i = 0; i < 1024; i++) {
```
apalos, line 102: There's a r/o CapsuleMax variable we ca read

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: Fixed, thanks.

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: I generated some for the tests.

apalos:
> #define STRLEN(s) (sizeof(s)/sizeof(s[0]) -1) would work?

But if your concern here is code generation, just use sizeof() instead of strlen() and it will be computed on compile time

hughsie:
@apalos I think this addresses most of your concerns, right?

hughsie:
@superm1 this also has the advantage that you can test deploying a capsule using either CoD, grub or shim using fwupdtool now.

apalos:
> @apalos I think this addresses most of your concerns, right?

Yea, looks good on a first look.  I'll try testing it within the week and let you know

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -682,6 +721,17 @@ fu_plugin_coldplug (FuPlugin *plugin, GError **error)
 		}
 	}
 
+	/* firmware may lie */
+	if (!fu_plugin_get_config_value_boolean(plugin, "DisableCapsuleUpdateOnDisk")) {
+		g_autoptr(GError) error_cod = NULL;
+		if (!fu_backend_uefi_check_cod_support(&error_cod)) {
+			g_debug("failed to check CapsuleOnDisk support: %s", error_cod->message);
```
superm1, line 728: nit: In the scenario it's not supported, this message will emit `failed to check CapsuleOnDisk support: Capsule-on-Disk is not supported` which seems wrong.
It should be instead `not using CapsuleOnDisk support: %s`

plugins/uefi-capsule/fu-uefi-grub-device.c
```diff
@@ -0,0 +1,201 @@
+/*
+ * Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
```
superm1, line 2: 2021, and I think I was part of this one too

plugins/uefi-capsule/fu-uefi-grub-device.c
```diff
@@ -0,0 +1,201 @@
...
+			  NULL,
+			  G_SPAWN_DEFAULT,
+			  NULL,
+			  NULL,
+			  &output,
+			  NULL,
+			  NULL,
+			  error))
+		return FALSE;
+	g_debug("%s", output);
```
superm1, line 91: do you think we need this debugging still?  I was pretty useful early on, but I think we only really care about logging for failures.

plugins/uefi-capsule/fu-uefi-grub-device.c
```diff
@@ -0,0 +1,201 @@
...
+
+	/* make fwupd default */
+	argv_reboot[0] = grub_reboot;
+	if (!g_spawn_sync(NULL,
```
superm1, line 95: maybe just return `return g_spawn_sync`?

plugins/dell/fu-self-test.c
```diff
@@ -553,6 +553,7 @@ main (int argc, char **argv)
 	g_setenv ("FWUPD_UEFI_ESP_PATH", sysfsdir, TRUE);
 	g_setenv ("FWUPD_UEFI_TEST", "1", TRUE);
 	g_setenv ("FWUPD_DELL_FAKE_SMBIOS", "1", FALSE);
+	g_setenv("FWUPD_IGNORE_EFIVARFS", "1", FALSE);
```
superm1, line 556: why a new variable?  can't we just flag the existing `FWUPD_UEFI_TEST`?

plugins/uefi-capsule/fu-uefi-backend.c
```diff
@@ -0,0 +1,83 @@
...
+	device = g_object_new(priv->device_gtype, "fw-class", guid, NULL);
+	fu_device_set_version_format(FU_DEVICE(device), FWUPD_VERSION_FORMAT_NUMBER);
+	return device;
+}
+
+static void
+fu_uefi_backend_init(FuUefiBackend *self)
+{
+	FuUefiBackendPrivate *priv = GET_PRIVATE(self);
+	priv->device_gtype = FU_TYPE_UEFI_SHIM_DEVICE;
```
superm1, line 77: I think this should be `FU_TYPE_UEFI_NVRAM_DEVICE`.  If secure boot is turned off it will not be using shim.  Shim is only used for SB.

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,236 @@
...
+	FuUefiDevice *device_uefi = FU_UEFI_DEVICE(device);
+	fwupd_guid_t guid = {0x0};
+	gsize bufsz = 0;
+	guint32 status = 0;
+	guint32 total_size = 0;
+	g_autofree gchar *guidstr = NULL;
+	g_autofree gchar *name = NULL;
+	g_autofree guint8 *buf = NULL;
+
+	/* lets assume we just did one update */
```
superm1, line 34: comment seems wrong?

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -0,0 +1,236 @@
...
+	name = g_strdup_printf("Capsule%04u", idx);
+	g_debug("attempting to read %s", name);
```
superm1, line 36: rather than attempting to read, how about just prefix the error of `failed to read %s`

plugins/uefi-capsule/fu-uefi-tool.c
```diff
@@ -62,55 +65,121 @@ main (int argc, char *argv[])
...
+	     /* TRANSLATORS: command argument: uppercase, spaces->dashes */
+	     C_("A single GUID", "GUID")},
+	    {"type",
+	     't',
+	     G_OPTION_FLAG_NONE,
+	     G_OPTION_ARG_STRING,
+	     &type,
+	     /* TRANSLATORS: command line option */
+	     _("Set the device type"),
+	     "shim|grub|cod"},
```
superm1, line 173: update to `nvram` instead of `shim` if you make the other changes

plugins/uefi-capsule/fu-uefi-tool.c
```diff
@@ -62,55 +65,121 @@ main (int argc, char *argv[])
...
+	     _("Apply firmware updates"),
+	     /* TRANSLATORS: command argument: uppercase, spaces->dashes */
+	     C_("A single GUID", "GUID")},
+	    {"type",
+	     't',
+	     G_OPTION_FLAG_NONE,
+	     G_OPTION_ARG_STRING,
+	     &type,
+	     /* TRANSLATORS: command line option */
+	     _("Set the device type"),
```
superm1, line 172: Maybe when presenting this it should be "device update method"

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: Fixed, ta.

plugins/uefi-capsule/fu-uefi-grub-device.c
hughsie: I'm really sorry, I've added you first in that file.

plugins/uefi-capsule/fu-uefi-grub-device.c
hughsie: I've hidden it with a FWUPD_UPDATE_CAPSULE_VERBOSE

plugins/uefi-capsule/fu-uefi-grub-device.c
hughsie: Fixed.

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: Fixed

plugins/uefi-capsule/fu-uefi-cod-device.c
hughsie: Fixed.

plugins/uefi-capsule/fu-uefi-tool.c
hughsie: Fixed, thanks.

plugins/uefi-capsule/fu-uefi-backend.c
hughsie: Agreed, changed.

plugins/uefi-capsule/fu-uefi-tool.c
hughsie: Fixed.

plugins/dell/fu-self-test.c
hughsie: Agreed, fixed.

plugins/uefi-capsule/fu-uefi-grub-device.c
```diff
@@ -0,0 +1,199 @@
...
+
+	/* delete the logs to save space; use fwupdate to debug the EFI binary */
+	if (fu_efivar_exists(FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_VERBOSE")) {
+		if (!fu_efivar_delete(FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_VERBOSE", error))
+			return FALSE;
+	}
+	if (fu_efivar_exists(FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_DEBUG_LOG")) {
+		if (!fu_efivar_delete(FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_DEBUG_LOG", error))
+			return FALSE;
+	}
```
superm1, line 161: I think you want these to only run in non self tests (IE move them later in the function).

superm1 (dismissed):
Just one problem, and then LGTM.  It would be good to make a call out to the OEMs to test master in the various scenarios once this is merged.

plugins/uefi-capsule/fu-uefi-grub-device.c
hughsie: Or move the tests for FWUPD_UEFI_TEST up? Do you think we want to run fu_uefi_device_write_update_info() and fu_uefi_device_fixup_firmware() for example?

hughsie:
I/we/you actually need to test nvram, grub and @apalos needs to test CoD on this branch; but I'm away from home until Thurs and don't want a broken laptop just yet :)

superm1:
> I/we/you actually need to test nvram, grub and @apalos needs to test CoD on this branch; but I'm away from home until Thurs and don't want a broken laptop just yet :)

With the change I pushed (and various other changes I've opened PR's for with unrelated issues) it works for GRUB on a system for me.

superm1 (dismissed):
LGTM and GRUB mode tests well for me now.

plugins/uefi-capsule/fu-uefi-grub-device.c
hughsie: Fixed, thanks.

apalos:
@hughsie on a quick pass this seems fine. I'll test it on monday and give you an update!

PR #3526, opened 2021-07-22
trivial: powerd: reformat code with contrib/reformat-code.py

djcampello:
cc @bleungatchromium @g2pop @binaryNewts


plugins/powerd/fu-plugin-powerd.c
```diff
@@ -31,28 +31,21 @@ fu_plugin_destroy (FuPlugin *plugin)
 gboolean
 fu_plugin_startup (FuPlugin *plugin, GError **error)
 {
-	FuPluginData *data = fu_plugin_get_data (plugin);
+	FuPluginData     *data = fu_plugin_get_data (plugin);
```
hughsie, line 34: is that right? I thought we only block aligned in structs?

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -31,28 +31,21 @@ fu_plugin_destroy (FuPlugin *plugin)
...
-						     G_DBUS_PROXY_FLAGS_NONE,
-						     NULL,
-						     "org.chromium.PowerManager",
-						     "/org/chromium/PowerManager",
-						     "org.chromium.PowerManager",
-						     NULL,
-						     error);
+	data->proxy = g_dbus_proxy_new_for_bus_sync (
+	    G_BUS_TYPE_SYSTEM, G_DBUS_PROXY_FLAGS_NONE, NULL, "org.chromium.PowerManager",
+	    "/org/chromium/PowerManager", "org.chromium.PowerManager", NULL, error);
```
hughsie, line 40: not sure i agree that is clearer....

plugins/powerd/fu-plugin-powerd.c
djcampello: AlignConsecutiveDeclarations is global in the sense that affects structs and blocks of code the same

hughsie: There's no way of making it non-global?

djcampello: not that I can find on https://clang.llvm.org/docs/ClangFormatStyleOptions.html :(

plugins/powerd/fu-plugin-powerd.c
djcampello: done

plugins/powerd/fu-plugin-powerd.c
hughsie: maybe AlignConsecutiveDeclarations=false makes more sense then perhaps?

djcampello: done

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -61,10 +62,7 @@ fu_plugin_startup (FuPlugin *plugin, GError **error)
 }
 
 gboolean
-fu_plugin_update_prepare (FuPlugin *plugin,
-			  FwupdInstallFlags flags,
-			  FuDevice *device,
-			  GError **error)
+fu_plugin_update_prepare (FuPlugin *plugin, FwupdInstallFlags flags, FuDevice *device, GError **error)
```
hughsie, line 65: that's 102 chars wide -- didn't we standardize on 80?

djcampello: This was left for avoiding huge changes to existent code... The piece responsible of this is the `PenaltyExcessCharacter: '1'` I'll remove now to go back to under 80

djcampello: done

hughsie (approved):

PR #3527, opened 2021-07-22
Fall back to the proxy GType vfuncs

hughsie:
This allows us to nest the GTypes and use the correct device for
progress and status updates.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I just worked this out; if we do the api-break thing (where FuProgress is passed as a parameter) we don't need to do this risky PR to get the status+progress sent to the daemon correctly.

PR #3528, opened 2021-07-22
powerd: Adding case for devices that ignore power

g2pop:
This adds a condition in the power plugin that permits updates for devices that don't take
any power conditions into account.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/powerd/fu-plugin-powerd.c
+++ b/plugins/powerd/fu-plugin-powerd.c
@@ -86,6 +86,10 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 		return FALSE;
 	}
 
+	/* permit updates when the device does not care about power conditions */
+	if (flags & FWUPD_INSTALL_FLAG_IGNORE_POWER)
+		return TRUE;
+
 	/* parse and use for battery-check conditions */
 	g_variant_get (powerd_response,
 		       "(uud)",


plugins/powerd/fu-plugin-powerd.c
```diff
@@ -86,6 +86,11 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 		return FALSE;
 	}
 
+	/* permit updates when the device does not care about power conditions */
+	if (flags & FWUPD_INSTALL_DEVICE_IGNORE_POWER){
+		return TRUE;
+	}
```
hughsie, line 92: no need for braces here

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -86,6 +86,10 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 		return FALSE;
 	}
 
+	/* permit updates when the device does not care about power conditions */
+	if (flags & FWUPD_INSTALL_DEVICE_IGNORE_POWER)
```
hughsie, line 90: do you mean `FWUPD_INSTALL_FLAG_IGNORE_POWER`? Does this even compile?

bleungatchromium (changes requested):
The upower plugin also has this quick check at the beginning of `fu_plugin_update_prepare`:

```
	/* not all devices need this */
	if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_REQUIRE_AC))
		return TRUE;
```

We should add this too for this plugin.

djcampello:
> The upower plugin also has this quick check at the beginning of `fu_plugin_update_prepare`:
> 
> ```
> 	/* not all devices need this */
> 	if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_REQUIRE_AC))
> 		return TRUE;
> ```
> 
> We should add this too for this plugin.

I guess this depends on the meaning for `FWUPD_DEVICE_FLAG_REQUIRE_AC`. If the meaning is that you need to have AC connected to the device to run the update: for plugins that do not care about AC they will still care about battery level. We don't want to succeed fast before we realize that we are running on 5% battery.

g2pop:
If I am reading  correctly, in both upower and here: https://github.com/fwupd/fwupd/blob/master/src/fu-engine.c#L287, if AC Power is required, it looks like updates only passes when the device is charging and when its battery level is above the threshold. If it's at 99%, but its discharging, the update will be blocked. And like @djcampello said, if no AC power is required, without checking battery after, we could pass a device that only has 5% battery power. I'm not sure if we want to use this logic, @bleungatchromium

bleungatchromium (approved):

hughsie:
> we could pass a device that only has 5% battery power

This is why it's important to update the power level of the FuContext; fwupd handles the device inhibits automatically and won't let this happen.

djcampello:
I don't think the FuContext is helping according the logic in https://github.com/fwupd/fwupd/blob/f2fd6b07170bb5fea52af2851abb0c162a0d073b/src/fu-engine.c#L287

It seems the logic only care about this when the flag is set. I would suggest removing the check for the flag on the second if and instead of adding this check to powerd plugin, to remove it from upower

WDYT?

hughsie:
> WDYT?

Seems sane to me.

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -86,6 +86,10 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 		return FALSE;
 	}
 
+	/* permit updates when the device does not care about power conditions */
```
djcampello, line 89: This should be done before calling on dbus. We don't care about what powerd says if the --ignore-power is set

PR #3529, opened 2021-07-22
Fix battery threshold handling when AC is *not* required

djcampello:
Also applying .clang-format to only the changed lines via: 
```
git diff -U0 | clang-format-diff-11 -p1 | sed '/^+/s/g_autoptr (/g_autoptr(/g;/^+/s/sizeof (/sizeof(/g;/^+/s/g_auto (/g_auto{/g' | patch -p0
```


djcampello:
cc @bleungatchromium @g2pop @binaryNewts

hughsie:
Hmm, I think the root issue here is that we currently have one flag which we're using for two things: e.g. FWUPD_DEVICE_FLAG_REQUIRE_AC is being used for "_I need to be on AC power and have at least this much battery_" -- which is probably what most plug-ins are expecting. We could split the flag up into:

    FWUPD_DEVICE_FLAG_REQUIRE_AC
    FWUPD_DEVICE_FLAG_REQUIRE_BATTERY_LEVEL

but I can't really think of a situation where you'd want just one and not the other. Comments welcome.

djcampello:
My take is that normally no one should want to initiate an update if there is a risk that fwupd will go away in the middle of an update because the system is being fourcefully powerd off due to lack of power. If anything I think the `FWUPD_DEVICE_FLAG_REQUIRE_BATTERY_LEVEL` should be `FWUPD_DEVICE_FLAG_IGNORE_BATTERY_LEVEL` which this is somehow already implemented as `--ignore-power`

hughsie:
> normally no one should want to initiate an update if there is a risk that fwupd will go away in the middle of an update

That's completely sane. So are we saying going forward that all updates must have above the critical battery level, and only select updates that actually require the AC adapter to be plugged in would set `FWUPD_DEVICE_FLAG_REQUIRE_AC`?

djcampello:
I think so! I think that internal components particularly would want to require AC and devices that are known to have very long update times.

Docks are particularly interesting in the sense that if they require AC to perform an update but they themselves are the power source, the handling becomes complicated.

plugins/upower/fu-plugin-upower.c
```diff
@@ -117,15 +117,14 @@ fu_plugin_update_prepare (FuPlugin *plugin,
...
 	/* not charging */
-	if (fu_context_get_battery_state (ctx) == FU_BATTERY_STATE_DISCHARGING ||
-	    fu_context_get_battery_state (ctx) == FU_BATTERY_STATE_EMPTY) {
+	if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_REQUIRE_AC) &&
+	    (fu_context_get_battery_state (ctx) ==
+		 FU_BATTERY_STATE_DISCHARGING ||
+	     fu_context_get_battery_state (ctx) == FU_BATTERY_STATE_EMPTY)) {
 		g_set_error_literal (error,
 				     FWUPD_ERROR,
 				     FWUPD_ERROR_AC_POWER_REQUIRED,
```
hughsie, line 130: do we want to move this into the engine perhaps?

djcampello: This again would require that powerd keeps the state up-to-date as it changes in the system via DBus signal handling and `fu_context_set_battery_state()`. At the moment powerd does not support said signal. @binaryNewts for additional details

hughsie: Ohh I think that's exactly what we want. It's much better to not allow the user to schedule the update than to "fail" after downloading and trying to deploy.

djcampello: I agree, but due to time constraints we may want to treat this as a future improvement :(

hughsie: Could you do a poll every minute using g_timer_add() and update the state/level then?

djcampello: @g2pop could you take a look at this and send a follow up change?

g2pop: Sure, I'll check it out.

djcampello:
I updated the commit pushing the logic to the engine from upower. We will move powerd once @g2pop figures out the timer based implementation.

hughsie (approved):

PR #3530, opened 2021-07-22
powerd: Moving ignore power case before method call

g2pop:
This moves the condition that permits updates for devices that don't account for power so that the plugin passes updates before a method call is made.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/powerd/fu-plugin-powerd.c
+++ b/plugins/powerd/fu-plugin-powerd.c
@@ -74,6 +74,10 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 	gdouble current_level = 0;
 	g_autoptr(GVariant) powerd_response = NULL;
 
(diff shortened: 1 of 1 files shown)


plugins/powerd/fu-plugin-powerd.c
```diff
@@ -73,6 +73,10 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 	gdouble current_level = 0;
 	g_autoptr(GVariant) powerd_response = NULL;
 
+	/* permit updates when the device does not care about power conditions */
```
djcampello, line 77: rebase on top of previous change so you can remove these lines from the current location on this change

djcampello: Thanks!

djcampello (approved):

Issue #3533, opened 2021-07-23
Lenovo T480 UEFI update 0.1.22 failed


konstiak:
**Describe the bug**
Installing FW update failed.

sudo fwupdmgr update
No upgrades for 20L5000AMC System Firmware, current is 0.1.36: 0.1.36=same, 0.1.35=older, 0.1.34=older, 0.1.33=older, 0.1.31=older
No upgrades for UEFI Device Firmware, current is 184.86.3909: 184.86.3909=same, 184.83.3874=older, 184.82.3838=older, 184.80.3746=older, 184.79.3722=older
Downloading 0.1.22 for UEFI Device Firmware...
Fetching firmware https://fwupd.org/downloads/41476a3066b9b459bca87936f63235c52f13ac047306dfb06d4d4782aa400b64-Lenovo-ThinkPad-T480-ECFirmware-1.22.cab
Downloading…             [*                                      ] Less than oneDownloading…             [**                                     ] Less than oneDownloading…             [**                                     ] Less than oneDownloading…             [***                                    ] Less than oneDownloading…             [****                                   ] Less than oneDownloading…             [*****                                  ] Less than oneDownloading…             [******                                 ] Less than oneDownloading…             [*******                                ] Less than oneDownloading…             [********                               ] Less than oneDownloading…             [***************************************]
failed to compile a783bf5c5c1c3b5e5e69e351f97059795198e31f:prefix=components: Chyba na riadku 46 znak 2: Dokument musí začínať prvkom (napr. <book>)

(translated: Error on row 46 character 2: Document has to start with element (e.g.<book>))

**Steps to Reproduce**
sudo fwupdmgr update

No upgrades for 20L5000AMC System Firmware, current is 0.1.36: 0.1.36=same, 0.1.35=older, 0.1.34=older, 0.1.33=older, 0.1.31=older
No upgrades for UEFI Device Firmware, current is 184.86.3909: 184.86.3909=same, 184.83.3874=older, 184.82.3838=older, 184.80.3746=older, 184.79.3722=older
Downloading 0.1.22 for UEFI Device Firmware...
Fetching firmware https://fwupd.org/downloads/41476a3066b9b459bca87936f63235c52f13ac047306dfb06d4d4782aa400b64-Lenovo-ThinkPad-T480-ECFirmware-1.22.cab
[...]

PR #3536, opened 2021-07-23
Do not assume the metainfo file is NUL-terminated

hughsie:
This was only true by accident. We'll need another fix for the LVFS to
add the missing NUL to restore compatibility for older clients.

Fixes https://github.com/fwupd/fwupd/issues/3533

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
The LVFS fix: https://gitlab.com/fwupd/lvfs-website/-/merge_requests/1050

PR #3538, opened 2021-07-23
reformat-code.py: add argumentless mode for formatting diffs

djcampello:
This is basically implementing `git diff -U0 | clang-format-diff-11 -p1 | sed '/^+/s/g_autoptr (/g_autoptr(/g;/^+/s/sizeof (/sizeof(/g;/^+/s/g_auto (/g_auto{/g' | patch -p0`

### DIFF
--- a/contrib/reformat-code.py
+++ b/contrib/reformat-code.py
@@ -11,8 +11,13 @@ import sys
 import subprocess
 
 CLANG_FORMATTERS = [
+    "clang-format-11",
     "clang-format-13",
     "clang-format",
+]
+CLANG_DIFF_FORMATTERS = [
+    "clang-format-diff-11",
+    "clang-format-diff-13",
 ]
 FIXUPS = {"g_autoptr (": "g_autoptr(", "sizeof (": "sizeof(", "g_auto (": "g_auto("}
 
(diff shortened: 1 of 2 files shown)


hughsie:
Could we add something in a README.md on how new contributors are supposed to use this? Thanks.

contrib/reformat-code.py
```diff
@@ -11,31 +11,36 @@
 import subprocess
 
 CLANG_FORMATTERS = [
+    "clang-format-11",
     "clang-format-13",
     "clang-format",
 ]
+CLANG_DIFF_FORMATTERS = [
+    "clang-format-diff-11",
```
superm1, line 19: Do you actually want to add 11 or not? I thought there was features needed from 13.

superm1:
> Could we add something in a README.md on how new contributors are supposed to use this? Thanks.

Since it's running on diffs it might be worth using with precommit now too.

hughsie:
> Since it's running on diffs it might be worth using with precommit now too.

Still not a fan; can we give it a few more iterations?

contrib/reformat-code.py
djcampello: On https://github.com/fwupd/fwupd/pull/3526 we removed the align consecutive declarations that required 13 as it not only affects structs but all declarations

djcampello:
Added some text to CONTRIBUTING.md

hughsie (approved):

PR #3549, opened 2021-07-27
Wip/hughsie/libcurl

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -675,6 +675,35 @@ fu_util_display_current_message (FuUtilPrivate *priv)
 }
 
 static gboolean
+fu_util_download(FuUtilPrivate *priv, gchar **values, GError **error)
+{
+	g_autofree gchar *basename = NULL;
+	g_autoptr(GBytes) blob = NULL;
+
+	/* one argument required */
+	if (g_strv_length(values) != 1) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_ARGS,
+				    "Invalid arguments");
+		return FALSE;
+	}
+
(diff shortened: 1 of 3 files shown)


src/fu-util.c
```diff
@@ -674,6 +674,35 @@ fu_util_display_current_message (FuUtilPrivate *priv)
 	}
 }
 
+static gboolean
+fu_util_download(FuUtilPrivate *priv, gchar **values, GError **error)
+{
+	g_autofree gchar *basename = NULL;
+	g_autoptr(GBytes) blob = NULL;
+
+	/* handle both forms */
```
superm1, line 683: comment seems wrong

src/fu-util.c
```diff
@@ -3344,6 +3373,13 @@ main (int argc, char *argv[])
 		     /* TRANSLATORS: command description */
 		     _("Get all enabled plugins registered with the system"),
 		     fu_util_get_plugins);
+	fu_util_cmd_array_add(cmd_array,
+			      "download",
+			      /* TRANSLATORS: command argument: uppercase, spaces->dashes */
+			      _("LOCATION"),
+			      /* TRANSLATORS: command description */
+			      _("Download a file using the client"),
```
superm1, line 3381: I don't think "using the client" is really needed.  Isn't that a detail that doesn't need to be exported "how" it's downloaded?

src/fu-util.c
hughsie: Fixed, thanks.

src/fu-util.c
hughsie: True, fixed.

superm1 (approved):

PR #3552, opened 2021-07-28
trivial: don't offer to upload reports for 0 length report URIs

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-remote.c
+++ b/libfwupd/fwupd-remote.c
@@ -380,6 +380,12 @@ fwupd_remote_set_report_uri (FwupdRemote *self, const gchar *report_uri)
 {
 	FwupdRemotePrivate *priv = GET_PRIVATE (self);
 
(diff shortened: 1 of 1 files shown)


superm1:
The other alternative that probably fixes this is to adjust fwupd-remote.c and check the length of the string being set and make the failure mode return NULL there.

hughsie:
That's what I was going to suggest as a counter argument. It saves all the mess of checking at the caller for NULL or "". I can have a go at this on Friday if that helps.

superm1:
> That's what I was going to suggest as a counter argument. It saves all the mess of checking at the caller for NULL or "". I can have a go at this on Friday if that helps.

just adjusted it

hughsie:
Alternate proposal in https://github.com/fwupd/fwupd/pull/3557

superm1:
Sure, looks better.

PR #3553, opened 2021-07-28
Add VOID target

superm1:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/void.sh
+++ b/contrib/ci/void.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -e
+set -x
+
+#install dependencies
+xbps-install -Suy python3
+./contrib/ci/generate_dependencies.py | xargs xbps-install -y
+
+#clone test firmware
+if [ "$CI_NETWORK" = "true" ]; then
+	./contrib/ci/get_test_firmware.sh
+	export G_TEST_SRCDIR=`pwd`/fwupd-test-firmware/installed-tests
+fi
+
+#build
+rm -rf build
+meson build \
+ -Dman=false \
+ -Ddocs=none \
+ -Dgusb:tests=false \
+ -Dgcab:docs=false \
+ -Dconsolekit=false \
+ -Dsystemd=false \
+ -Db_lto=false \
+ -Delogind=true
+ninja -C build test -v
--- a/.github/workflows/create_containers.yml
+++ b/.github/workflows/create_containers.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [fedora, debian-x86_64, arch, debian-i386]
+        os: [fedora, debian-x86_64, arch, debian-i386, void]
 
     steps:
       - name: Check out the repo
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [fedora, debian-x86_64, arch, debian-i386]
+        os: [fedora, debian-x86_64, arch, debian-i386, void]
     steps:
       - uses: actions/checkout@v2
       - name: Docker login
--- a/contrib/ci/Dockerfile-void.in
+++ b/contrib/ci/Dockerfile-void.in
@@ -1,0 +1,10 @@
+FROM ghcr.io/void-linux/void-linux:latest-full-x86_64-musl
+%%%OS%%%
+ENV LANG en_US.UTF-8
+ENV LC_ALL en_US.UTF-8
+ENV CI_NETWORK true
+RUN echo fubar > /etc/machine-id
+RUN xbps-install -Suy xbps && xbps-install -uy
+%%%INSTALL_DEPENDENCIES_COMMAND%%%
+WORKDIR /github/workspace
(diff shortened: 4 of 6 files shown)


superm1:
CC @ericonr If you can please review it

contrib/ci/void.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+set -e
+set -x
+
+#install dependencies
+xbps-install -Suy python3
+./contrib/ci/generate_dependencies.py | xargs xbps-install -Suy
```
paper42, line 7: -S and -u are unnecessary here, the repositories were synced and the system was updated before in the Dockerfile

superm1: The Dockerfile is only updated nightly.  Is it possible they're out of sync?

superm1: Although I guess the previous line installing python also did the sync and update.  So probably at most need it in dockerfile and when python installs and that's it.

paper42: You are probably right, removing it from the second line shouldn't break anything. Anyway, xbps-install -Suy python3 will update **only** python3, nothing else.

superm1: Fixed 

My experience was python3 wasn't in the container by default. I think after this is merged and the first nightly container builds it won't get needed anymore.

paper42: What I am saying doesn't mean that you don't need to install python3, but that xbps-install -Suy python3 will not update the whole system. So add a line before this one with xbps-install -Suy.

I still don't think it's necessary to update the system again here. The Dockerfile is ran once per build right? If yes, then it doesn't matter that you are pulling a 24 hour old image, because in the next step you `RUN xbps-install -Suy`, so the system is up to date and -Su is not needed in this file.

superm1: No the Dockerfile is used to generate the container once per night, and so `xbps-install -Suy` runs once per 24 hours.  From the CI this is the only command run:
https://github.com/fwupd/fwupd/blob/master/.github/workflows/main.yml#L55

Which effectively means it runs `void.sh` and that's it.

hughsie (approved):
I have no authority here! :)

contrib/ci/void.sh
paper42: That makes sense, thanks for explaining. You still need to run `xbps-install -Suy` somewhere though.

superm1: I'll merge the PR as it is and then tomorrow after we have a successful run making the container will tear out python3 (which it shouldn't be needed anymore - should end up in container) and then can move the `-Suy` around

ericonr:
@superm1 thanks a lot for finishing this! I couldn't find time to sit down and figure out the whole thing. Really appreciate it.

superm1:
> @superm1 thanks a lot for finishing this! I couldn't find time to sit down and figure out the whole thing. Really appreciate it.

Sure thing. Something else we do with other distros that might be helpful is keep a copy of packaging stuff in contrib/ and call the real packaging.

It helps you as a downstream identify issues or packaging changes sooner.

If you want to do that feel free to submit in.

superm1:
@paper42 @ericonr the container isn't generating properly each night.  I've tried a few times to fix it, but keep flopping.
Can you take a look?
Here is the failure log:
https://github.com/fwupd/fwupd/runs/3206669569?check_suite_focus=true

And here is what is called to build it:
https://github.com/fwupd/fwupd/blob/master/.github/workflows/create_containers.yml
https://github.com/fwupd/fwupd/blob/master/contrib/ci/generate_docker.py
https://github.com/fwupd/fwupd/blob/master/contrib/ci/Dockerfile-void.in

Presumably you can check out master and try to do the same thing and repro it too.

ericonr:
I believe it's what was said here:

https://github.com/fwupd/fwupd/pull/3553#discussion_r678455173

You are trying to install something from the repositories that depends on updated packages, which leads to some weird corner cases in XBPS.

I changed the `RUN` line to be `RUN xbps-install -Suy xbps && xbps-install -uy && xbps-install -y \` and it seems to be going through fine.

I believe CI runners are hosted in the US, right (if you know that)? We could move it to use the US mirror for a speed up too.

superm1:
> I believe it's what was said here:
> 
> https://github.com/fwupd/fwupd/pull/3553#discussion_r678455173
> 
> You are trying to install something from the repositories that depends on updated packages, which leads to some weird corner cases in XBPS.
> 
> I changed the `RUN` line to be `RUN xbps-install -Suy xbps && xbps-install -uy && xbps-install -y \` and it seems to be going through fine.
> 
> I believe CI runners are hosted in the US, right (if you know that)? We could move it to use the US mirror for a speed up too.

Thanks! Can you add a PR with that fix?
I don't know where GitHub hosts them, if it's US then yes sure!

ericonr:
Sorry for the delay, will send in a bit.

ericonr:
Ah, you implemented already. Thanks.

dkwo:
@superm1 We currently use `-Dsupported_build=true -Dconsolekit=false -Dintrospection=true -Dsystemd=false -Dplugin_altos=false -Delogind=true -Dtests=false` and a couple of new dependencies that should be added are `gi-docgen fwupd-efi`.

PR #3566, opened 2021-07-29
realtek-mst: Add a warning that extreme caution should be used

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/plugins/realtek-mst/README.md
+++ b/plugins/realtek-mst/README.md
@@ -14,6 +14,9 @@ identified partially through the presence of Realtek's OUI in the Branch
 Device OUI fields of DPCD (DisplayPort Configuration Data), they do not have
 unique Device Identification strings.
 
+This plugin was neither written, verified, supported or endorsed by Realtek
+Semiconductor Corp.
+
 ## Firmware Format
 
 The daemon will decompress the cabinet archive and extract a firmware blob in


plugins/realtek-mst/README.md
```diff
@@ -5,6 +5,9 @@
 This plugin updates the firmware of DisplayPort MST hub devices made by Realtek,
 such as the RTD2141b and RTD2142.
 
+This plugin was neither written by, verified or endorsed by Realtek Semiconductor
+Corp, and so should be used with extreme caution.
```
superm1, line 9: how is this any different than most of the other plugins?  Why is it special to receive such a message in README?

hughsie: Realtek expressed concern.

superm1: So instead, can't they review the plugin and endorse it?  It feels like it's 2 steps away from person reports "there is a problem with that tool" to realtek and they say "we only support `realtek-mst-flasher.so`".

superm1: Specifically r8152 comes to mind...

hughsie: I'll poke them offline.

tari: The implication that the correctness of the plugin ("extreme caution") depends on Realtek's endorsement seems weird- what about "it is not supported by Realtek," which would also short-circuit @superm1's hypothetical?

hughsie: So `This plugin was neither written by, verified, supported or endorsed by Realtek Semiconductor Corp.`? Or still too wordy?

tari: Could remove the first "by", but I'm happy with that.

hughsie: > So instead, can't they review the plugin and endorse it

That'd be a _no_ from them.

tari (approved):
The commit summary might be edited since the README no longer recommends extreme caution, but content changes LGTM.

Realtek-William:
We saw that the discussion in Github will only add a warning documentation file.
We hope Google can use another word to replace “realtek”  word of the plugin’s name, instead of only adding a documentation to illustrate.

PR #3567, opened 2021-07-29
trivial: Ignore the NUL suffix to fix loading new firmware with old GLib versions

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-cabinet.c
+++ b/libfwupdplugin/fu-cabinet.c
@@ -453,6 +453,16 @@ fu_cabinet_set_container_checksum_cb (XbBuilderFixup *builder_fixup,
 	return TRUE;
 }
 
+static gboolean
+fu_common_bytes_has_nul_suffix(GBytes *data)
+{
+	gsize bufsz = 0;
+	const guint8 *buf = g_bytes_get_data(data, &bufsz);
+	if (bufsz > 0 && buf[bufsz - 1] == '\0')
+		return TRUE;
+	return FALSE;
+}
+
 /* adds each GCabFile to the silo */
 static gboolean
 fu_cabinet_build_silo_file (FuCabinet *self,
@@ -461,6 +471,7 @@ fu_cabinet_build_silo_file (FuCabinet *self,
 			    GError **error)
 {
 	GBytes *blob;
+	g_autoptr(GBytes) blob_fixed = NULL;
 	g_autoptr(GError) error_local = NULL;
 	g_autoptr(XbBuilderSource) source = xb_builder_source_new ();
 	g_autoptr(XbBuilderNode) bn_info = xb_builder_node_new ("info");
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Can you detect what version of glib drops the suffix?   Then we can leave some comments and do a version check to get rid of the technical debt some day when the minimum version of glib is bumped.

hughsie:
The honest answer is that I didn't manage to work it out. It works for some firmware but not others.

superm1:
> The honest answer is that I didn't manage to work it out. It works for some firmware but not others.

Can you determine what function from glib is introducing this?  Also - I wonder if this is better done in libxmlb?

hughsie:
>  Also - I wonder if this is better done in libxmlb

Then wouldn't we have to hard depend on that new libxmlb version?

superm1:
> Then wouldn't we have to hard depend on that new libxmlb version?

Well I mean it depends on the impact from this bug... what's happening and what's breaking?  I don't think I understand that bigger picture of the impact.

hughsie:
> what's happening and what's breaking

Okay, I think I understand what's happened now:

 * I switched on compression on the LVFS for cabinet archives
 * some time passes
 * a few files fail to install, users report bugs about `Error on line 46 char 2: Document must begin with an element` which the root cause is ae48e02ac23537acda110191d801c30c82e8f70e where we were assuming the document was NUL terminated and treating it as a string, but now the blob is compressed, the uncompressed length is no longer 0 (which was being read as \0) -- this fixes the root issue for new fwupd versions
 * some time passes
 * I then patch the LVFS, with https://gitlab.com/fwupd/lvfs-website/-/commit/eeb217c6923e1521473e7f8f4658624eb960b124 which forces a NUL byte at the end of each metainfo file which "fixes" the issue for older versions of fwupd which assume the metainfo data is NUL terminated
 * this then breaks **newer** fwupd parsing **newer** firmware as the XML parser knows that the data is a specific number of bytes long (and thus tries to parse NUL as markup), rather than relying on the string being NUL terminated.

So we have a few choices.

 * We revert the fwupd patch and use the bug that fwupd assumes a NUL byte. This breaks if you do get-details on any non-LVFS .cab file that is compressed, so isn't ideal.
 * We revert the lvfs-website NUL-adding patch and tell people affected to upgrade to the bleeding edge fwupd
 * We disable cab compression (which has reduced download size by 50% on some firmware) and then resign everything signed in the last two weeks or so -- and rely on the accidental NUL-termination non-compression buglet
 * We leave the lvfs-website and fwupd fixes in and merge this PR with a fixed up description -- after-all `xmllint` ignores the trailing NUL and that's probably good enough rationale
 * We revert the lvfs-website patch adding the extra NUL and we add some kind of hidden padding in python-cabarchive so that we arrange the archive in a specific way -- but see the first point.
 * Other ideas very welcome!

superm1:
Wow, OK.  I think that
>We leave the lvfs-website and fwupd fixes in and merge this PR with a fixed up description -- after-all xmllint ignores the trailing NUL and that's probably good enough rationale

Is probably the right answer then.

hughsie:
> Wow, OK

Yes, I was really surprised this worked before. The luck was that we always write the jcat blob after the metainfo.xml file and so we don't get into the situation where you can push the raw CFDATA into the GLib XML parser. I still don't think that would be a security bug as you only get a "<IHH" buffer which isn't enough to exploit expat/libxml.

superm1 (approved):

PR #3572, opened 2021-07-30
Some adjustments to make contrib/reformat-code.py more friendly

superm1:
This should let someone who failed to set it up the first time before they committed easily run it.

Workflow I envision:
1. Person does some cool new feature but forgot to run contrib/setup first.
2. We tell them to fix it up, adjust for pre-commit CI failure
3. They run `./contrib/setup`, but their commit is still unhappy because pre-commit runs "during" commit.
4. They run `./contrib/reformat-code.py master` and their code is reformatted.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/reformat-code.py
+++ b/contrib/reformat-code.py
@@ -8,12 +8,27 @@
 import os
 import sys
 import subprocess
+import argparse
 
 CLANG_DIFF_FORMATTERS = [
     "clang-format-diff-11",
     "clang-format-diff-13",
     "clang-format-diff",
 ]
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Reformat C code to match project style",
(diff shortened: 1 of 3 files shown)


contrib/reformat-code.py
```diff
@@ -16,6 +17,21 @@
...
+        epilog="Call with no argument to reformat uncommitted code.",
+    )
+    parser.add_argument(
+        "commit", nargs="*", default="", help="Reformat all changes since this commit"
+    )
+    parser.add_argument(
+        "--debug", action="store_true", help="Display all launched commands"
+    )
+    args = parser.parse_args()
+    return args
```
hughsie, line 32: just return directly?

hughsie:
One nit, and then LGTM, thanks.

hughsie (approved):

PR #3578, opened 2021-07-30
figure out what's going on with freebsd

superm1:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/build_freebsd_package.sh
+++ b/contrib/ci/build_freebsd_package.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-
+set -x
 GITHUB_SHA=
 GITHUB_REPOSITORY=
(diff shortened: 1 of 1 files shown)


superm1:
@Asiderr, can you help?  I tried to look if the file got moved around, but I appear to be out of my depth and something that is outside of fwupd broke FreeBSD CI.

Asiderr:
@superm1 Sure, I'll look into it by this Tuesday. I was AFK this weekend, so I need to catch up on my current work

hughsie:
The FreeBSD CI isn't blocking, and we can disable it if required. It does seem fragile.

hughsie:
I've disabled this for now, we can revert and fix when we know what the problem is.

Issue #3554, opened 2021-07-28
1_5_X: No Prometheus IOTA Config on HP laptop with version 1.5.11


andch-nn:
**Describe the bug**
Use fwupdmgr get-devices to list all devices on HP laptop, Promethues IOTA config will miss in 1.5.11-0ubuntu1~21.04.1.

**Steps to Reproduce**
1. Enable proposed package in Ubuntu Hirsute
2. sudo apt install fwupd
3. fwupdmgr get-devices

**Expected behavior**
├─Prometheus:
│ │ Device ID: 0fef0a0c55f6442bffaebd774ae771341c89571b
│ │ Summary: Fingerprint reader
│ │ Current version: 10.01.3273255
│ │ Vendor: Synaptics (USB:0x06CB)
│ │ Install Duration: 2 seconds
│ │ GUIDs: 816af83b-646c-5033-ae19-26391eeb675d
│ │ 8b06355d-849b-5125-85ec-6d02a20cf1ac
│ │ Device Flags: • Updatable
│ │ • Supported on remote server
│ │ • Cryptographic hash verification is available
│ │
│ └─Prometheus IOTA Config:
│ Device ID: 40b77e58384a02768e3ce90ed74c93ca2311327d
│ Current version: 0021
│ Minimum Version: 0021
│ Vendor: Synaptics (USB:0x06CB)
│ GUIDs: 0e58a9cb-f338-567b-a7ca-2a821eb8f3c2
│ 7b93dd19-18d1-5a0c-a32d-79a5d27e2b38
│ Device Flags: • Updatable
│ • Supported on remote server

**fwupd version information**
1.5.11.

```shell
fwupdmgr --version
```
client version: 1.5.11
compile-time dependency versions
        gusb:   0.3.5

daemon version: 1.5.11
Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
apt
**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
fwupdmgr get-devices --show-all-devices
```
HP ZBook Fury 15 G7 Mobile Workstation
│
├─Thunderbolt host controller:
│     Device ID:          22ec5d304a254d9e72059a7661ad2e09dbb49d2f
│     Summary:            Unmatched performance for high-speed I/O
│     Current version:    62.00
│     Vendor:             HP Inc. (THUNDERBOLT:0x00F0, TBT:0x00F0)
[...]

PR #3580, opened 2021-07-31
Allow devices to only accept version upgrades

hughsie:
See https://github.com/fwupd/fwupd/issues/3554 for details.

Type of pull request:

- [X] Code fix
- [X] Feature
- [ ] Documentation


superm1 (approved):
I think this way works really well and looks good, but something else to consider - the device can just set version lowest as something bigger than the current version.  IE current version + 1 or something.

PR #3582, opened 2021-08-02
Wip/hughsie/lenovo

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -393,7 +393,9 @@ done
 %dir %{_datadir}/fwupd/quirks.d
 %{_datadir}/fwupd/quirks.d/*.quirk
 %{_datadir}/doc/fwupd/builder/README.md
+%if 0%{?have_uefi}
 %{_sysconfdir}/grub.d/35_fwupd
+%endif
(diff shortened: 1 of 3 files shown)


hughsie:
@superm1 I've merged this to avoid delaying the release. We can make it less broad when Lenovo tell us what's happened.

superm1:
> @superm1 I've merged this to avoid delaying the release. We can make it less broad when Lenovo tell us what's happened.

How about making ux capsule opt in instead?

hughsie:
> How about making ux capsule opt in instead?

I don't think any OEMs would actually opt-in and test.

superm1:
> > How about making ux capsule opt in instead?
> 
> I don't think any OEMs would actually opt-in and test.

I seem to recall that you sent an inquiry out if anyone wants to use it and someone wanted to. If they really want to use it they can?

hughsie:
> and someone wanted to

IIRC, that was Lenovo. :/

superm1:
> > and someone wanted to
> 
> IIRC, that was Lenovo. :/

So all the more reason make it opt in for when they're ready! :)

Issue #3568, opened 2021-07-29
FakeDevice update time


binaryNewts:
**Is your feature request related to a problem? Please describe.**
When doing tests on fwupd that require testing of something while an update is running, it's helpful to have something that takes a couple seconds to update. The FakeDevice integrated webcam provided, however, takes about a second to update, which is not very much time to do testing during.

**Describe the solution you'd like**
It would be nice to have either
- a way to set the amount of time that the device sleeps for
- a flag that would stop the update somewhere in the middle and continue when prompted

**Additional context**
Here's where the amount of time is currently set
https://github.com/fwupd/fwupd/blob/master/plugins/test/fu-plugin-test.c#L164

(@djcampello @bleungatchromium @g2pop)

hughsie:
> a way to set the amount of time that the device sleeps for

Could you parse the payload? e.g. parse it as a json or XML file with `<delay_seconds>130<delay_seconds>`

> a flag that would stop the update somewhere in the middle

Something like DeviceRequest() would work well for this.

superm1:
I think parsing the fake device payload is a great idea.
On that note - We are currently parsing `FWUPD_PLUGIN_TEST` to determine what tests to run.  If you shift this over to add delays from the payload, please adjust the plugin to get this information from the payload too.

bleungatchromium:
Hi @hughsie and @superm1 

Do you have an example where similar values are pulled from a payload metadata?

I assume this is coming from https://fwupd.org/downloads/3fab34cfa1ef97238fb24c5e40a979bc544bb2b0967b863e43e7d58e0d9a923f-fakedevice124.cab right?

So the idea is that if we want to run a test and customize the fake delay, we'd source a different .cab file with an extra field to tell the plugin how long to spin?

superm1:
>Do you have an example where similar values are pulled from a payload metadata?

Right now `data/installed-tests/fakedevice124.bin` is just ASCII, and a value.  The idea being proposed is to make that more.
```
<firmware>
<version>0x1020004</version>
<test>foo</test>
<delay>0</delay>
</firmware>
```

If that turns out to not scale as well as we were thinking, you can certainly say BAD IDEA, and then maybe an environment variable in the fwupd environment that the test plugin picks up is better/easier.

hughsie:
I think setting the environment variables once during self test startup is fine, but I think changing them during self test runtime is going to make it a bit like spaghetti code that's really hard to debug.

hughsie:
[...]

PR #3586, opened 2021-08-02
test env variables

binaryNewts:
Create environment variables for the fake camera so that it can have a user-determined update time for each state. If no input is given, it should default to its current wait time.

in reference to https://github.com/fwupd/fwupd/issues/3568

### DIFF
--- a/plugins/test/fu-plugin-test.c
+++ b/plugins/test/fu-plugin-test.c
@@ -152,6 +152,13 @@ fu_plugin_update (FuPlugin *plugin,
 	const gchar *test = g_getenv ("FWUPD_PLUGIN_TEST");
 	gboolean requires_activation = g_strcmp0 (test, "requires-activation") == 0;
 	gboolean requires_reboot = g_strcmp0 (test, "requires-reboot") == 0;
+	guint64 decompress_ms = 100;
+	const gchar *decompress_ms_str = g_getenv("FWUPD_PLUGIN_TEST_DECOMPRESS_MS");
+	guint64 write_ms = 100;
+	const gchar *write_ms_str = g_getenv("FWUPD_PLUGIN_TEST_WRITE_MS");
+	guint64 verify_ms = 100;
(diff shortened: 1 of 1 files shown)


plugins/test/fu-plugin-test.c
```diff
@@ -160,19 +160,37 @@ fu_plugin_update (FuPlugin *plugin,
...
 	}
 	fu_device_set_status (device, FWUPD_STATUS_DECOMPRESSING);
-	for (guint i = 1; i <= 100; i++) {
-		g_usleep (1000);
-		fu_device_set_progress (device, i);
+	guint64 decompress_ms = 100;
+	const gchar *decompress_ms_str = g_getenv("FWUPD_PLUGIN_TEST_DECOMPRESS_MS");
+	if (decompress_ms_str != NULL) {
+		decompress_ms = g_ascii_strtoll(decompress_ms_str, NULL, 10);
+	}
```
hughsie, line 167: @djcampello is it possible to specify "no braces for single line if's" in the clang-format somehow?

djcampello: My understanding is that this goes beyond what `clang-format` can do but it may be possible to achieve with `clang-tidy`

plugins/test/fu-plugin-test.c
```diff
@@ -160,19 +160,34 @@ fu_plugin_update (FuPlugin *plugin,
 		return FALSE;
 	}
 	fu_device_set_status (device, FWUPD_STATUS_DECOMPRESSING);
-	for (guint i = 1; i <= 100; i++) {
-		g_usleep (1000);
-		fu_device_set_progress (device, i);
+	guint64 decompress_ms = 100;
```
djcampello, line 163: Please fix the mix ue of declarations / code

hughsie:
Alternate proposal in https://github.com/fwupd/fwupd/pull/3594 -- does that work?

binaryNewts:
works perfect!

PR #3588, opened 2021-08-03
redfish: Disable the uefi_capsule plugin if coldplug succeeded

hughsie:
We don't want to show the big warning about the missing ESRT on server
hardware that is managed by a BMC:

    WARNING: UEFI capsule updates not available or enabled in firmware setup
      See https://github.com/fwupd/fwupd/wiki/PluginFlag:capsules-unsupported for more information.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
There isn't any spec saying it has to be done this way, so at least until we see an EC or BMC with redfish and ESRT together it makes sense to me.

hughsie:
Right, I did consider a quirk, but it seemed sufficiently unlikely. Thanks for the sanity check.

PR #3592, opened 2021-08-03
Fall back to the parent if the child does not have a battery level set

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@hardboprobot would this be useful for you?

superm1 (commented):
I think this should be an explicit opt in behavior. If you have a parent and child that both have batteries but one can't properly report for some reason this could be reporting improper information.

hughsie:
Fixed.

hardboprobot (approved):

PR #3594, opened 2021-08-03
test: Allow setting plugin parameters using an environment variable

hughsie:
Based on a patch by Twain Byrnes <binarynewts@google.com>, many thanks.

Use this to test:

sudo FWUPD_TEST_PLUGIN_XML="<config><delay_decompress_ms>100</delay_decompress_ms></config>" \
./src/fwupdtool --plugins test get-devices -v

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/test/fu-plugin-test.c
+++ b/plugins/test/fu-plugin-test.c
@@ -9,7 +9,9 @@
 #include <fwupdplugin.h>
 
 struct FuPluginData {
-	GMutex			 mutex;
+	guint delay_decompress_ms;
+	guint delay_write_ms;
+	guint delay_verify_ms;
 };
 
 void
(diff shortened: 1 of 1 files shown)


hughsie:
@binaryNewts does this work instead?

binaryNewts:
I'm trying to test this out for myself but am having some troubles.

I run `FWUPD_TEST_PLUGIN_XML="<delay_decompress_ms>100000</delay_decompress_ms>" ` and then `/usr/bin/fwupdmgr update b585990a-003e-5270-89d5-3705a17f9a4` (or `fwupdtool` instead of `/usr/bin/fwupdmgr`).
On the second command, I keep getting an error: `Invalid arguments` and then it terminates.
When I run with fwupdtool, I get to full bars for decompressing and loading before it breaks, but it stops working immediately for fwupdmgr.

Do you know what could be causing this?

hughsie:
Doh, the commit message is wrong. I think it needs to be:

    sudo FWUPD_TEST_PLUGIN_XML="<config><delay_decompress_ms>100</delay_decompress_ms></config>"
./src/fwupdtool --plugins test get-devices -v

If you want to use the daemon, you'll need to do something like:

    sudo FWUPD_TEST_PLUGIN_XML="<config><delay_decompress_ms>100</delay_decompress_ms></config>"
./src/fwupd -v

and then do:

    fwupdmgr install foo.cab

...in another tab.

bleungatchromium (approved):
@binaryNewts  and I tested this in a ChromeOS environment and we got it working with both the tool and the daemon + fwupdmgr.

The "get-devices" test isn't super useful because I don't think the decompress, write or verify delays are ever hit when we do a simple get-devices.

An update, or a reinstall, yes.

That's what we were able to test on our side. We could stick into the env that bit of xml and it worked.

PR #3595, opened 2021-08-04
trivial: if usb4 device device presents then activate it exclusively

CragW:
Successful usb4 authenticate will reboot the dock immediately, dock
relevant devices are gone. If EC activate already in queue will
eventually get to failure as device missing (disconnected).

Same behavior is seen with EC activate which reboots the dock right
away therefore only one of the two devices can exercise 'activate'
action.

Running fwupdate activate With presence of usb4 device, if EC has
NEED_ACTIVATION flag from EC query the flag will be transfered to
usb4 device therefore only usb4 gets updated.

Unplug dock cable is the only way to complete dock EC firmware update.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
>Same behavior is seen with EC activate which reboots the dock right
away therefore only one of the two devices can exercise 'activate'
action.

So when the USB4 does an online activation, does that actually also cause a PD event indicating disconnect of the cable and EC activation kicks off too?  Or does it stay pending after this happens?

plugins/dell-dock/fu-dell-dock-usb-usb4.c
```diff
@@ -413,11 +417,19 @@ fu_dell_dock_usb4_hub_nvm_write (FuDevice *device, const guint8 *buf,
...
+		return FALSE;
+
 	if (!fu_dell_dock_usb4_hub_operation (device, OP_NVM_AUTH_WRITE,
 					      NULL, error)) {
 		g_prefix_error (error, "NVM authenticate failed: ");
 		return FALSE;
 	}
+
+	if (!fu_device_locker_close(locker, error))
+		return FALSE;
```
superm1, line 431: I don't think you need to explicitly call close do you?  Shouldn't it happen automatically from using `g_autoptr`?

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -129,11 +129,22 @@ fu_plugin_backend_device_added (FuPlugin *plugin,
 	/* GR controller internal USB HUB */
 	if ((guint) fu_usb_device_get_vid (FU_USB_DEVICE (device)) == GR_USB_VID &&
 	    (guint) fu_usb_device_get_pid (FU_USB_DEVICE (device)) == GR_USB_PID) {
+		GPtrArray *devices = fu_plugin_get_devices(plugin);
```
superm1, line 132: this feels too racy to be calling it within `backend_device_added`.  It means the order of events will matter now on which devices have probed first (GR or Hub2).  Can you run it somewhere else?

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -129,11 +129,22 @@ fu_plugin_backend_device_added (FuPlugin *plugin,
...
 		locker = fu_device_locker_new (FU_DEVICE (usb4_dev), error);
 		if (locker == NULL)
 			return FALSE;
+
+		/* steal needs_activation if EC has pending update */
+		for (guint i = 0; i < devices->len; i++) {
+			FuDevice *device_tmp = g_ptr_array_index(devices, i);
+			if (FU_IS_DELL_DOCK_EC(device_tmp)) {
+				fu_dell_dock_steal_activation(FU_DEVICE(device_tmp),
+							      FU_DEVICE(usb4_dev));
```
superm1, line 145: in general since this is only called in one place and one time, I would say no need to make a static function for it.  Just do the stuff directly "here".

superm1 (commented):
The early return on online activation makes sense to me, but I think we need to think more about the stealing activation flag:
1. when to do that
2. what happens to the EC's activation if you call USB4 online activation

plugins/dell-dock/fu-dell-dock-usb-usb4.c
CragW: > 1. when to do that
Will move it to _registered, thanks.

> 2. what happens to the EC's activation if you call USB4 online activation
EC's activation needs to wait until the dock cable is unplugged, no conflict for now but I will test it again when EC's activation for USB4 back normal. E.g. remove USB4 online activation.

CragW:
> The early return on online activation makes sense to me, but I think we need to think more about the stealing activation flag:
> 
> 1. when to do that

Will move it to _registered, thanks.

> 2. what happens to the EC's activation if you call USB4 online activation

EC's activation needs to wait until the dock cable is unplugged, no conflict for now but I will test it again when EC's activation for USB4 back normal. E.g. remove USB4 online activation.

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -174,7 +174,7 @@ fu_plugin_backend_device_added (FuPlugin *plugin,
 void
 fu_plugin_device_registered (FuPlugin *plugin, FuDevice *device)
 {
-	/* thunderbolt plugin */
+	/* usb4 device from thunderbolt plugin */
```
superm1, line 177: nit: seems wrong - this is actually for TBT3 device and USB4 device both..

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -184,6 +184,25 @@ fu_plugin_device_registered (FuPlugin *plugin, FuDevice *device)
...
 	}
 
+	/* steal needs_activation if EC has pending update */
+	if (g_strcmp0(fu_device_get_plugin(device), "dell_dock") == 0 &&
+	    FU_IS_DELL_DOCK_USB4(device)) {
+		GPtrArray *devices = fu_plugin_get_devices(plugin);
+		for (guint i = 0; i < devices->len; i++) {
+			FuDevice *device_tmp = g_ptr_array_index(devices, i);
+			if (FU_IS_DELL_DOCK_EC(device_tmp) &&
+			    fu_device_has_flag(device_tmp, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
```
superm1, line 194: I'm confused - if the EC has an activation pending that doesn't mean USB4 also has an activation pending.  If you have an update that is EC + package version only, but not USB4 NVM, doesn't that mean this will leave USB4 NVM in needs activation state EVERY SINGLE TIME?

I would think it's better to use history database to set this information, rather than mistakingly taking it from EC needs-activation.

Otherwise this scenario happens:
1. Plug in dock
2. It enumerates.
2. Issue update that is EC and package only
3. Notice that EC is in needs activation.
4. Restart `fwupd.service` or  reboot (don't unplug dock)
5. Notice that USB4 is in needs activation.

superm1 (commented):
I have no concerns to the  `fu-dell-dock-usb-usb4.c` changes, but the stealing activation I need more convincing.  Please split to two commits and we can cherry-pick the first if need be.

plugins/dell-dock/fu-plugin-dell-dock.c
superm1: I think specifically what is missing is is a check that usb4 device already had needs activation flag.  Then you just "remove" the EC's needs-activation to prevent the problem, instead of moving it over every time.

CragW: > I think specifically what is missing is is a check that usb4 device already had needs activation flag. Then you just "remove" the EC's needs-activation to prevent the problem, instead of moving it over every time.

For running `fwupd` with daemon that is okay as `needs_activation` is tagged to usb4 device. But for `fwupdtool` that is blindsided from `history` doesn't know that usb4 has `needs_activation` flag nor the `update_state`, ended up ignored usb4.

plugins/dell-dock/fu-dell-dock-usb-usb4.c
```diff
@@ -510,7 +520,7 @@ fu_dell_dock_usb4_write_fw (FuDevice *device, FuFirmware *firmware,
 		return FALSE;
 
 	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION);
-	fu_device_set_version (device, fw_version);
+	fu_device_set_version(device, fw_version);
```
CragW, line 523: @superm1 do you think it's okay to set a predicted version here as is yet authenticated? I know it's already in this way for a while. :)

superm1: I think we've done this for some other devices too to intentionally prevent the same update getting staged twice while it's pending activation.  I don't see a problem with it, especially if you can't read needs activation on in between fwupdtool runs.

plugins/dell-dock/fu-plugin-dell-dock.c
superm1: Why doesn't it come from history database? That sounds like the bug this is working around to me...

superm1: Try adding `FU_DEVICE_INTERNAL_FLAG_INHERIT_ACTIVATION` to the device.  If that works - it's way cleaner, and you can just remove the activation flag from EC device when both are set.

CragW: Adding `FU_DEVICE_INTERNAL_FLAG_INHERIT_ACTIVATION` brings the flag to `usb4` device, thank you. 

Can we shift `fu_engine_device_inherit_history` earlier than `fu_engine_plugin_device_register`? I wanted to address the mutual exclusive problem in `_registered` but the flag comes at the very end.

```
  /* notify all plugins about this new device */
  if (!fu_device_has_flag (device, FWUPD_DEVICE_FLAG_REGISTERED))
	  fu_engine_plugin_device_register (self, device);
  ...
  ...
  
  /* sometimes inherit flags from recent history */
  fu_engine_device_inherit_history (self, device);

```

superm1: I don't see a problem with swapping the order here.

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -184,6 +184,24 @@ fu_plugin_device_registered (FuPlugin *plugin, FuDevice *device)
...
+	    FU_IS_DELL_DOCK_USB4(device) &&
+	    fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
+		GPtrArray *devices = fu_plugin_get_devices(plugin);
+		for (guint i = 0; i < devices->len; i++) {
+			FuDevice *device_tmp = g_ptr_array_index(devices, i);
+			if (FU_IS_DELL_DOCK_EC(device_tmp) &&
+			    fu_device_has_flag(device_tmp, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION)) {
+				fu_device_remove_flag(FU_DEVICE(device_tmp),
+						      FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION);
+				g_debug("activate for %s is suspended by %s",
```
superm1, line 198: How does this flag ever come back now? I wonder if we are actually talking about inhibited but for activation not update.

superm1: But actually looking at it, I see that might be problematic because inhibitition is usually called when something is put into needs-activate state already.  I think what you've done is OK.

hughsie:
Crag, you can merge this whenever you like. Thanks.

superm1:
> It seems merged in one, would probably a bad idea?

I guess you hit squash and merge rather than rebase and merge.  Oh well, not worth reverting and splitting now.

CragW:
> > It seems merged in one, would probably a bad idea?
> 
> I guess you hit squash and merge rather than rebase and merge. Oh well, not worth reverting and splitting now.

Yup, hit too soon. OK.

PR #3605, opened 2021-08-05
Allow a device to specify that it must be involved in a composite update

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -235,6 +235,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "use-parent-for-open";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_BATTERY)
 		return "use-parent-for-battery";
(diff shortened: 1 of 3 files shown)


superm1:
It feels like this PR is a workaround for a badly crafted composite CAB during development. Is that right?

hughsie:
> It feels like this PR is a workaround for a badly crafted composite CAB during development. Is that right?

Yes, it's when the user downgraded the child device version manually (e.g. using install-blob) and then we "repair" the device by upgrading to any version. But because the main firmware isn't included we flash the child device and make the situation worse.

superm1:
So then perhaps should install blob be blocked entirely for this class of device?

hughsie:
> So then perhaps should install blob be blocked entirely for this class of device?

I think that would make development much harder. Install-blob is supposed to be a "I really know what I'm doing" action. I think you can get into the same situation if you install the correct firmware on a standard device, if you remove the physical device before the main firmware has been written. I agree it's an edge case.

hughsie:
@hardboprobot do you concur with this? I'm kinda guessing this works for you.

hughsie:
@hardboprobot do you think we should close this PR?

PR #3611, opened 2021-08-06
redfish: Add a per-vendor quirk to set all devices as requiring a reboot

hughsie:
Some versions of XCC do not send the Base.1.10.ResetRequired message.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (dismissed):
is it really going to be all devices? can't BMC reload some without a reboot?  If so, no concerns.

hughsie:
> is it really going to be all devices

I *think* so. I've asked Lenovo to verify.

PR #3612, opened 2021-08-06
Precommit checks for license

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/check-license.py
+++ b/contrib/ci/check-license.py
@@ -1,0 +1,42 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2021 Richard Hughes <richard@hughsie.com>
+# Copyright (C) 2021 Mario Limonciello <superm1@gmail.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+import glob
+import sys
+
+
+def __get_license(fn: str) -> str:
+    with open(fn, "r") as f:
+        for line in f.read().split("\n"):
+            if line.find("SPDX-License-Identifier:") > 0:
+                return line.split(":")[1]
+    return ""
+
+
+def test_files() -> int:
+
+    rc: int = 0
+
+    for fn in glob.glob("**/*.[c|h|py|sh]", recursive=True):
+        if "meson-private" in fn:
+            continue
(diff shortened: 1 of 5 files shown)


contrib/ci/check-license.py
```diff
@@ -0,0 +1,44 @@
...
+    for fn in glob.glob("**/*.[c|h|py|sh]", recursive=True):
+        if "meson-private" in fn:
+            continue
+        license = __get_license(fn)
+        if not license:
+            print("{} does not specify a license".format(fn))
+            rc = 1
+            continue
+        if not "GPL" in license:
+            print("{} does not contain LGPL or GPL ({})".format(fn, license))
```
hughsie, line 35: Are we not LGPL2+ everywhere?

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -3,21 +3,7 @@
...
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program; if not, write to the Free Software
- * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ * SPDX-License-Identifier: GPL-2+
```
hughsie, line 6: I think this should be LGPLv2, no?

superm1: You wrote it originally, so I guess that's actually our intent..

contrib/ci/check-license.py
superm1: Here are the exceptions:
1. EMMC plugin was leveraged from GPL code, I don't think we can just reclassify it.
2. BCM57xx is GPL-2+
3. CPU plugin is GPL-2+
4. Intel SPI plugin is GPL-2+
5. Flashrom is GPL-2

hughsie (approved):

PR #3613, opened 2021-08-06
Allow using a global percentage completion

hughsie:
and start all over again.

By creating a FuProgress helper object we can create a truely global
percentage completion of the entire transaction, which was always the
grand plan before things got complicated with composite devices.

Using FuProgress we can encode at the plugin level "the erase takes 50%
of the time, the write takes 40% and the read takes 10%". This means
we can have a progressbar which goes up at a consistent speed, even when
multiple devices are being updated or when there are multiple phases
required.

This does mean a fairly major break to the FuDevice API so that we can
pass around the FuProgress objects (or children there-of), but we just
broken plugin ABI and it's a good time to do this kind of thing.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 lots more work to make the plugins emit global progress, but I wanted to get some initial feedback.

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,819 @@
+/*
+ * Copyright (C) 2009 Richard Hughes <richard@hughsie.com>
```
superm1, line 2: Back to the future!

superm1:
> Using FuProgress we can encode at the plugin level "the erase takes 50%
of the time, the write takes 40% and the read takes 10%". This means
we can have a progressbar which goes up at a consistent speed, even when
multiple devices are being updated or when there are multiple phases
required.

To be honest, I don't really like having an object that is not part of the device to represent the progress.    Why can't the daemon reconcile the progress of all the devices to build out the transaction rather than passing an object around like this?  The same concepts of explaining the usage of time can happen, but as a representation of the device.  The daemon can add up all the devices in the transaction and do the math to determine total install duration for the devices and assign percentages to report to a frontend based upon that.

If my device takes 30 seconds to update, and 30% of that is typically erase, the devices could certainly indicate in some metadata how long the phases take and timers can run using `InstallDuration` in between those timers to reduce spurts if the times vary sometimes by a second or two.

libfwupdplugin/fu-progress.c
hughsie: Yep! I actually wrote most of this code for zif, a loooong time ago: https://github.com/hughsie/zif/tree/master/libzif :)

hughsie:
> To be honest, I don't really like having an object that is not part of the device to represent the progress. 

Dude, I asked for your feedback, I appreciate you being honest! The first version of this patch actually had FuProgress as an internal object inside FuDevice, and fu_device_set_progress() just proxied to that in the common case. In the uncommon case we'd just fu_device_get_progress_obj() and then do the usual fu_progress_set_steps() type things.

This however got messy really fast. For instance, what's responsible for actually setting (or connecting to) the progress object for each device? If there's a composite device with children, is the parent supposed to set the progress_obj for the child if the child is updated at the same time? What happens if the child restarts, id the device-list responsible for setting up the progress signals correctly? What happens if both restart? Is the engine supposed to set up the signal connections again for both devices in the tree? After a night of pondering I ripped out all the complex code and decided that passing the transaction state explicitly around was so much cleaner. You don't have to worry _which_ FuDevice is actually doing the action (e.g. parent, proxy, etc) or if it's a runtime or bootloader, you just use the transaction state. See below.

> and do the math to determine total install duration for the devices and assign percentages to report to a frontend based upon that.

You can do that, but it turns the percentage completion into a lie. It just becomes the interpolated sum of all the install durations -- and that doesn't exactly inspire confidence from end users when the percentage goes from 0..100 but the first device in the composite update actually failed to replug and we waited 30 seconds even though the entire update (for both devices) was only supposed to take 20 seconds.

> devices could certainly indicate in some metadata how long the phases take 

In reality, I don't think that works. e.g. for a mobile broadband device you have to flash 4 partitions (of wildly different sizes) and for some composite devices you do erase,write,read on one device and just a write on the child device.

Maybe my mistake was the *name*. Rather than FuProgress, would something like FuState or FuTransaction be clearer that it represents the slice of the current transaction. I don't know if reporting things back to the client like "sane safely cancel the action" makes sense using this same object and FuProgress would be an odd name for that.

superm1:
> > To be honest, I don't really like having an object that is not part of the device to represent the progress.
> 
> Dude, I asked for your feedback, I appreciate you being honest! The first version of this patch actually had FuProgress as an internal object inside FuDevice, and fu_device_set_progress() just proxied to that in the common case. In the uncommon case we'd just fu_device_get_progress_obj() and then do the usual fu_progress_set_steps() type things.
> 
> This however got messy really fast. For instance, what's responsible for actually setting (or connecting to) the progress object for each device? 

See I was thinking that FuDeviceList could easily iterate across the devices and update the progress.  We're single threaded, nothing writes updates in parallel.  Even if we go multi-threaded some day, I don't really see us updating devices in parallel, maybe just handling dbus in one thread and device updates in another.  The device updates would otherwise never be deterministic.

>If there's a composite device with children, is the parent supposed to set the progress_obj for the child if the child is updated at the same time? 

Why can't the child set it's own progress?  If it's because of a long blocking transaction then can't the FuDeviceList take the InstallDuration and just update progress as time progresses?  I guess I don't see why composite devices need to be any different than discrete devices in this regard.  As long as the Install Duration is accuratish for each composite device it shouldn't matter.

> What happens if the child restarts, id the device-list responsible for setting up the progress signals correctly? What happens if both restart?

Yeah shouldn't it be handled as part of replug and incorporate?  You can copy the progress of the individual donor device over, and then reconnect the signals in replug.

> Is the engine supposed to set up the signal connections again for both devices in the tree? After a night of pondering I ripped out all the complex code and decided that passing the transaction state explicitly around was so much cleaner. You don't have to worry _which_ FuDevice is actually doing the action (e.g. parent, proxy, etc) or if it's a runtime or bootloader, you just use the transaction state. See below.
> 
> > and do the math to determine total install duration for the devices and assign percentages to report to a frontend based upon that.
> 
> You can do that, but it turns the percentage completion into a lie. 

Well it then means percentage completion of _this_ device.  Which might be a very small subset of the total.  The daemon doesn't need to tell the client about this.  It just needs to say "Erasing foo device", "Writing foo device" and have a number that goes up.

>It just becomes the interpolated sum of all the install durations -- and that doesn't exactly inspire confidence from end users when the percentage goes from 0..100 but the first device in the composite update actually failed to replug and we waited 30 seconds even though the entire update (for both devices) was only supposed to take 20 seconds.

But if something fails to replug you would presumably the replug and check the device version as some step.  So if you have 3 devices each taking 30 seconds total and reading a version maybe 1 second and the "second device" fails to replug then your percentage will probably go up to 65% and be pegged there until a timeout.  Isn't that exactly what _should_ happen? 
> 
> > devices could certainly indicate in some metadata how long the phases take
> 
> In reality, I don't think that works. e.g. for a mobile broadband device you have to flash 4 partitions (of wildly different sizes) and for some composite devices you do erase,write,read on one device and just a write on the child device.
> 
> Maybe my mistake was the _name_. Rather than FuProgress, would something like FuState or FuTransaction be clearer that it represents the slice of the current transaction. I don't know if reporting things back to the client like "sane safely cancel the action" makes sense using this same object and FuProgress would be an odd name for that.

I don't think there is ever a safe point of return when the client starts a transaction unless every device in the transaction has `FWUPD_DEVICE_FLAG_DUAL_IMAGE`.

hughsie:
> I was thinking that FuDeviceList could easily iterate across the devices and update the progress

Sorry for the delay, I wanted to try some new ideas before replying. Yes, this works, but fails for non-obvious reasons. When we push the device percentage to 0 (to get the spinning back and forth fwupd UI for something like detach) you have to keep track of the "highest seen percentage value" and then you need to reset *that* back to 0% at the end of the transaction, which gets tricky as the FuDevice might be part of the old device (e.g. bootloader) and isn't actually active in the device list anymore. If you instead decouple the FuDevice progress with a FuDeviceLIst-specific progress property it all gets crazy very quickly.

> can't the FuDeviceList take the InstallDuration and just update progress as time progresses

The key problem here is threefold:

 * The InstallDuration is wildly inaccurate, if provided at all (as it's optional)
 * We told vendors to include the time needed for device replug, which includes the USB reenumeration step (which differs wildly on different machines/hubs) possibly multiple times
* The update time depends on the device firmware state. So If i have a modem firmware with three partitions, on a very old factory version I might need to update all three partitions and on an updated-last-week device I might only need to flash the telecom data. The difference in time would be say 5 minutes when including the OS partition to less than 2 seconds. This means until we know the state of the old firmware and the new firmware we're not in a position to change the install time.

I see the install time as the "worst case scenario" which is probably per-firmware in most cases. The device state (e.g. writing, verifying) is probably still a property of the device, but I don't think the percentage completion is a property of the device, but a property of the action being done. For instance, would the device "progress" _make sense_ as an abstract property if there was no transaction going on?

> I don't think there is ever a safe point of return when the client starts a transaction

Yes, you're probably correct. I'll drop that crazy cancellation idea! :)

There's one point here which might be worth thinking about. At the moment we provide the `FwupdInstallFlags` to `write_firmware` and `prepare_firmware` so they can change their behaviour based on what the user has specified. What happens if we want that information in `detach`, `activate` or `composite_prepare`? What happens if a plugin needs to change device deploy behaviour based on a `FwupdFeatureFlags` in the future?

Maybe *all* the vfuncs should have a `FuState` parameter which includes the transaction state. Or `FuTransaction`, or `FuTask` or whatever. Then we don't need to break API/ABI *again* in the future to include something that makes sense to leak to-or-from the FuDevice action without creating it as writable property with unclear ownership on the device itself.

superm1:
>Sorry for the delay, I wanted to try some new ideas before replying. Yes, this works, but fails for non-obvious reasons. When we push the device percentage to 0 (to get the spinning back and forth fwupd UI for something like detach) you have to keep track of the "highest seen percentage value" and then you need to reset that back to 0% at the end of the transaction, which gets tricky as the FuDevice might be part of the old device (e.g. bootloader) and isn't actually active in the device list anymore. If you instead decouple the FuDevice progress with a FuDeviceLIst-specific progress property it all gets crazy very quickly.

Now I would argue if you're going to be changing things around the whole "bouncing around because we don't have an accurate value" either needs to go away or change to be represented as a timer that is increasing with time.

>The InstallDuration is wildly inaccurate, if provided at all (as it's optional)

As I recall the intention was "do I sit and watch it finish (<10 seconds) or do I go grab a snack"?

>The update time depends on the device firmware state. So If i have a modem firmware with three partitions, on a very old factory version I might need to update all three partitions and on an updated-last-week device I might only need to flash the telecom data. The difference in time would be say 5 minutes when including the OS partition to less than 2 seconds. This means until we know the state of the old firmware and the new firmware we're not in a position to change the install time.

Great point - this is true for Synaptics MST as well.

>I see the install time as the "worst case scenario" which is probably per-firmware in most cases. The device state (e.g. writing, verifying) is probably still a property of the device, but I don't think the percentage completion is a property of the device, but a property of the action being done. For instance, would the device "progress" make sense as an abstract property if there was no transaction going on?

You're almost arguing towards creating a model for the updater that the daemon can use to know how long things take on average.  Sounds crazy at first, but actually, why not?  If you can let the daemon record how long different phases take and then spit that into some XML that can get compiled into a database.   You can potentially crowdsource data, get more accurate numbers and all that fun stuff too.

>Yes, you're probably correct. I'll drop that crazy cancellation idea! :)

Phew.

>Maybe all the vfuncs should have a FuState parameter which includes the transaction state. Or FuTransaction, or FuTask or whatever.

It comes down to the model, do plugins write firmware and track the state of their individual device, or do they need to keep track of the state of everything in the transaction for some reason?  8 times out of 10 I think it's the former.

hughsie:
> bouncing around because we don't have an accurate value

Yes, this is the firmware equivalent of  ¯\_(ツ)_/¯ -- and you're right to say it's not good enough.

> If you can let the daemon record how long different phases take

Yes, we could certainly do this, but you'd also need to actually add the information about what each phase is doing, e.g. adding the fu_progress_step() commands in the right place so we can transition from one state to another between erase,write,read. We'd also need something exactly like FuProgress that we could connect to and read some kind of json/xml blob out of the root node that captures the entire set of tasks and subtasks. We have all the data in FuProgress for that kind of thing already, we just don't export that. Will ponder.

hughsie:
I've split out a super-simple little step in https://github.com/fwupd/fwupd/pull/3627

PR #3619, opened 2021-08-09
modem-manager: Fix the report that failed to get device after update

Jarvis-Jiang-G:
Module may not have been re-probed and exposed in MM when the FW version was successfully obtained, which leads to the report that failed to get device after update. In fact, FW has been uodated successfully, it needs to add 150s to wait for the module to be re-probed and exposed in MM.

Signed-off-by: Jarvis Jiang <jarvis.w.jiang@gmail.com>

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/modem-manager/fu-mm-device.c
```diff
@@ -1109,6 +1109,8 @@ fu_mm_device_write_firmware_mbim_qdu (FuDevice *device, GBytes *fw, GError **err
 		return FALSE;
 
 	fu_device_set_status (device, FWUPD_STATUS_DEVICE_READ);
+	fu_device_set_remove_delay (device, MAX_WAIT_TIME_SECS * 1000);
+	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
```
hughsie, line 1113: we've changed out style guide -- can you remove the spaces before the brackets on these two added lines please and then precommit will be happy. Then LGTM, thanks.

Jarvis-Jiang-G: ok, thank you for reminding me .

superm1: Would suggest you run contrib/setup on your side to set up pre-commit rules for the future.

PR #3620, opened 2021-08-09
Fix cleanup and plugin to have the correct args order

hughsie:
All the other vfuncs have 'plugin, device, flags' but prepare and
cleanup vfuncs being 'plugin, flags, device' order has been triggering
my OCD for the last few years.

We've just broken the symbol names, so it's the right time to fix this.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin-private.h
+++ b/libfwupdplugin/fu-plugin-private.h
@@ -44,10 +44,10 @@ gboolean	 fu_plugin_runner_coldplug_cleanup	(FuPlugin	*self,
 							 GError		**error)
 							 G_GNUC_WARN_UNUSED_RESULT;
 gboolean
-fu_plugin_runner_prepare(FuPlugin *self, FwupdInstallFlags flags, FuDevice *device, GError **error)
+fu_plugin_runner_prepare(FuPlugin *self, FuDevice *device, FwupdInstallFlags flags, GError **error)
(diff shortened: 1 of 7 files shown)


superm1 (commented):
Fix the CI failure and it is fine to me with the abi change

superm1:
I fixed the CI failure.  Something else that is coming to mind though - as we're making these ABI breaking changes and it's going to eventually flow into the documentation we need to think about what happens on documentation.  Should we have an ABI indication there somehow?
IE if someone tries to go refer to https://fwupd.github.io and they see some pages pointing back to older ABI version generated documentation too?

hughsie:
> we need to think about what happens on documentation

The way GNOME does this is have per-branch documentation pages. I don't know if that makes so much sense for fwupd as we encourage vendors to only submit new plugins targeting master, and rarely backport entire new plugins to stable branches.

superm1:
> > we need to think about what happens on documentation
> 
> The way GNOME does this is have per-branch documentation pages. I don't know if that makes so much sense for fwupd as we encourage vendors to only submit new plugins targeting master, and rarely backport entire new plugins to stable branches.

I guess the worry I have here is the ABI breakage between master and 1.6.x means that the documentation doesn't reflect the code for someone developing a new plugin until 1.7.0 is released.  Pretty big corner case though I guess if this really breaks someone.

Issue #3622, opened 2021-08-10
Automatically inhibit devices with a low or critical battery level


hardboprobot:
**Is your feature request related to a problem? Please describe.**
`get-devices` doesn't show the user when an updatable device (a peripheral) has a critical battery level that's below its required threshold for updatability.

**Describe the solution you'd like**
Ideally, when the battery level of a device is below the defined threshold, it should be automatically inhibited and `get-devices` should show that information to the user.

PR #3623, opened 2021-08-10
Recalculate the device inhibits when the flags property changes

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3622

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -73,6 +73,7 @@ typedef struct {
 	FuDeviceInternalFlags		 internal_flags;
 	guint64				 private_flags;
 	GPtrArray			*private_flag_items;	/* (nullable) */
+	gulong notify_flags_handler_id;
 } FuDevicePrivate;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -73,6 +73,7 @@ typedef struct {
 	FuDeviceInternalFlags		 internal_flags;
 	guint64				 private_flags;
 	GPtrArray			*private_flag_items;	/* (nullable) */
+	gulong notify_flags_handler_id;
```
hardboprobot, line 76: Proper tabulation here

hardboprobot (commented):
It works as expected. Only a nit,  the client now reports `Device <whatever> [aa6ab44818f70a9f9e763d914adb9e55c7cf819d] does not currently allow updates`. Could we add the reason (the Update Error string) to that message in `fu_install_task_check_requirements`?

hughsie:
@hardboprobot does that ^^^ work like you expected?

hardboprobot (approved):

Issue #3625, opened 2021-08-11
Watch /etc/fwupd/uefi_capsule.conf for changes and restart the daemon as required


c0deright:
**Is your feature request related to a problem? Please describe.**

* fwupdmgr v1.3.11 did work with 2 ESPs without having to use `OverrideESPMountPoint`. It just used the `/boot/efi` mount point.
* after updating to v1.5.11 the 2 ESPs resulted in the "UEFI ESP partition not detected or configured" error so I modified `/etc/fwupd/uefi_capsule.conf` and put `OverrideESPMountPoint=/boot/efi` in it
* I didn't know there was a daemon involved that needed to be restarted so I opened https://github.com/fwupd/fwupd/issues/3621 which could have been prevented if
  * a) there was a clear notice in `uefi_capsule.conf` to restart the daemon after modifications to it
  * b) the daemon running in the background would check the mtime of `uefi_capsule.conf` every time a userspace program like `fwupdmgr` is called and then would emit a warning like _Config file changed since daemon start. please restart the fwupd daemon_ when the mtime is newer than the time of the process start.
  * c) the daemon would automatically SIGHUP and reload the config. This might not be what everybody is expecting and should be considered before implementing.

PR #3626, opened 2021-08-11
Restart the daemon if any of the the plugin config files are modified

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3625

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5762,6 +5762,14 @@ fu_engine_plugin_rules_changed_cb (FuPlugin *plugin, gpointer user_data)
 }
 
 static void
+fu_engine_plugin_config_changed_cb(FuPlugin *plugin, gpointer user_data)
+{
+	FuEngine *self = FU_ENGINE(user_data);
+	g_info("config file for %s changed, sending SHUTDOWN", fu_plugin_get_name(plugin));
+	fu_engine_set_status(self, FWUPD_STATUS_SHUTDOWN);
+}
+
+static void
 fu_engine_context_security_changed_cb (FuContext *ctx, gpointer user_data)
 {
 	FuEngine *self = FU_ENGINE (user_data);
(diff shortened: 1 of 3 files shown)


superm1:
Is it a crazy idea to instead unload and reload plugin? Given that is what happens for the other configuration files in the daemon just wondering if it makes sense rather than take down the whole daemon.

hughsie:
> Is it a crazy idea to instead unload and reload plugin

I did ponder that, but there's just too much going on. e.g. recoldplugs, handling registration of other plugins etc -- it's much more safe to just quit when it's safe and then start up from scratch again. I don't think it's the sort of thing that's going to happen very often in reality.

src/fu-engine.c
```diff
@@ -5761,6 +5761,13 @@ fu_engine_plugin_rules_changed_cb (FuPlugin *plugin, gpointer user_data)
 	}
 }
 
+static void
+fu_engine_plugin_config_changed_cb(FuPlugin *plugin, gpointer user_data)
+{
+	FuEngine *self = FU_ENGINE(user_data);
+	fu_engine_set_status(self, FWUPD_STATUS_SHUTDOWN);
```
superm1, line 5768: I think you need a non-verbose message here for the reason to shut down the daemon for the journal.

superm1 (dismissed):
one nit and then looks okay

superm1:
> > Is it a crazy idea to instead unload and reload plugin
> 
> I did ponder that, but there's just too much going on. e.g. recoldplugs, handling registration of other plugins etc -- it's much more safe to just quit when it's safe and then start up from scratch again. I don't think it's the sort of thing that's going to happen very often in reality.

very true, sounds good.

superm1 (approved):

PR #3627, opened 2021-08-11
Allow using a per-device global percentage completion

hughsie:
and start all over again. Some plugins go 0..100% for write, others
go 0..100% for erase, then again for write, then *again* for verify.

By creating a helper object we can easily split up the progress of the
specific task, e.g. write_firmware().

We can encode at the plugin level "the erase takes 50% of the time, the
write takes 40% and the read takes 10%". This means we can have a
progressbar which goes up just once at a consistent speed.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 if you agree with the FuProgress name/API then I'll also include the other plugins where we have multiple phases of erase/write/verify.

libfwupdplugin/fu-progress.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <gio/gio.h>
+
+#define FU_TYPE_PROGRESS (fu_progress_get_type())
+G_DECLARE_DERIVABLE_TYPE(FuProgress, fu_progress, FU, PROGRESS, GObject)
+
+struct _FuProgressClass {
+	GObjectClass parent_class;
+	/* signals */
+	void (*percentage_changed)(FuProgress *self, guint value);
```
superm1, line 17: do you also want a signal for step changed?

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,714 @@
+/*
+ * Copyright (C) 2009 Richard Hughes <richard@hughsie.com>
```
superm1, line 2: fix the date here

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,714 @@
...
+ * fu_progress_set_custom_steps_full:
+ * @self: A #FuProgress
+ * @id: the code location
+ * @value: A step weighting variable argument array
+ *
+ * This sets the step weighting, which you will want to do if one action
+ * will take a bigger chunk of time than another.
+ *
+ * All the values must add up to 100, and the list must end with -1.
+ * Do not use this function directly, instead use the fu_progress_set_custom_steps() macro.
```
superm1, line 316: should it be in a private header to prevent that?

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,714 @@
...
+ *
+ * Since: 1.7.0
+ **/
+void
+fu_progress_set_custom_steps_full(FuProgress *self, const gchar *id, guint value, ...)
```
superm1, line 321: I don't know about locking into this API of steps are percent of the device update vs time of the device update.  Maybe allow funcs for both to determine later?

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,714 @@
...
+{
+	FuProgressPrivate *priv = GET_PRIVATE(self);
+	va_list args;
+	guint total = value;
```
superm1, line 325: doesn't this mean that you double count the first one because total is +=?

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,714 @@
...
+fu_progress_sleep(FuProgress *self, guint delay_secs)
+{
+	gulong delay_us_pc = (delay_secs * G_USEC_PER_SEC) / 100;
+
+	g_return_if_fail(FU_IS_PROGRESS(self));
+	g_return_if_fail(delay_secs > 0);
+
+	fu_progress_set_percentage(self, 0);
+	for (guint i = 0; i < 100; i++) {
+		g_usleep(delay_us_pc);
```
superm1, line 654: So is the idea that all plugins that have sleep statements will instead have a fu_progress_sleep?

plugins/colorhug/fu-colorhug-device.c
```diff
@@ -423,9 +423,13 @@ fu_colorhug_device_write_firmware (FuDevice *device,
 				   GError **error)
 {
 	FuColorhugDevice *self = FU_COLORHUG_DEVICE (device);
+	FuProgress *progress = fu_device_get_progress_helper(device);
 	g_autoptr(GBytes) fw = NULL;
 	g_autoptr(GPtrArray) chunks = NULL;
 
+	/* progress */
+	fu_progress_set_custom_steps(progress, 19 /* erase */, 45 /* write */, 36 /* read */, -1);
```
superm1, line 431: so in my mind I was thinking that every time the device entered into a certain state (say erase or write) you were going to be able to map out a new step.  Is that how you're thinking too?  And things with multiple partitions or multiple erase steps as long as there is a state change it keeps working?

IE set status to erase, write, erase, write, read would be 5 steps

libfwupdplugin/fu-progress.h
hughsie: I don't _think_ so. If we needed one later we can easily add it.

libfwupdplugin/fu-progress.c
hughsie: So, I'm kinda hoping that percentage complete and time-complete will be the same thing, at least if everything uses FuProgress like I'm imagining. At least when you use FWUPD_PROGRESS_PROFILE=1 we use the _time_ between each step to suggest the `fu_progress_set_custom_steps()` values.

libfwupdplugin/fu-progress.c
hughsie: As I understand `va_arg` -- or rather, _if_ I understand it then only the `...` args get included.

libfwupdplugin/fu-progress.c
hughsie: I think so. Then we know how the time maps to the percentage. I'm also pondering making it be milliseconds instead, as whole seconds doesn't seem useful enough for most devices.

plugins/colorhug/fu-colorhug-device.c
hughsie: Yes, that's exactly what I think this does; i.e. for 2 partitions each with E,W,R you'd have one parent progress with two steps, and then two children each with three steps, so you've got 6 non-similar "splits" in your 0..100% scale.

libfwupdplugin/fu-progress.c
hughsie: Hmm, this didn't work as well as I hoped. As it's a pre-processor directive (as we need G_STRLOC) you need to be able to find the `fu_progress_set_custom_steps_full` prototype when compiling the code. The alternative is we do something like:

	fu_progress_set_custom_steps(progress, G_STRLOC, 19 /* erase */, 45 /* write */, 36 /* read */, -1);

But that's kinda _nrrrgh_. Other ideas welcome. The reason we save the G_STRLOC is so we can tell the user _which_ `fu_progress_set_custom_steps()` needs fixing up when profiling, or when the user used to many `fu_progress_step_done()`'s

superm1: Maybe add individual steps (`fu_progress_add_step`)

libfwupdplugin/fu-progress.c
superm1: OK..

libfwupdplugin/fu-device.c
```diff
@@ -4731,10 +4751,18 @@ fu_device_init (FuDevice *self)
...
 	priv->possible_plugins = g_ptr_array_new_with_free_func (g_free);
 	priv->retry_recs = g_ptr_array_new_with_free_func (g_free);
+	priv->progress = fu_progress_new();
 	g_rw_lock_init (&priv->parent_guids_mutex);
 	g_rw_lock_init (&priv->metadata_mutex);
 	priv->notify_flags_handler_id =
 	    g_signal_connect(self, "notify::flags", G_CALLBACK(fu_device_flags_notify_cb), NULL);
+
+	/* allows more verbose profiling for plugin development */
+	fu_progress_set_profile(priv->progress, g_getenv("FWUPD_PROGRESS_PROFILE") != NULL);
```
superm1, line 4761: Some suggestions to consider here...
1. Maybe make `fwupdtool` always set this.  People do development using fwupdtool, and get this data.
2. If a plugin doesn't set custom steps, then assume that it needs to be profiled.

libfwupdplugin/fu-progress.c
superm1: sounds g=fine

libfwupdplugin/fu-device.c
hughsie: Yes, I agree that we need to set the profiling on more aggressively. Maybe we also tie it to the daemon --verbose option?

> If a plugin doesn't set custom steps, then assume that it needs to be profiled.

Maybe. That needs some thinking about because the non-custom step count is typically set before a loop of some sort. Will ponder.

libfwupdplugin/fu-progress.c
hughsie: I was _dreaming_ about this, so it's probably about time to go on holiday. Anyway. Where I think this might make sense is doing something like:

	fu_progress_set_custom_steps(progress, 10 /* erase */, 70 /* write */, 20 /* verify */, -1);

	/* erase phase */
	fu_device_set_status(device,FWUPD_STATUS_DEVICE_ERASE)
	...

	/* write phase */
	fu_device_set_status(device,FWUPD_STATUS_DEVICE_WRITE)
	...

	/* verify phase */
	fu_device_set_status(device,FWUPD_STATUS_DEVICE_VERIFY)
	...

and instead have someting like:

	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_ERASE, 1);
	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 7);
	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_VERIFY, 2);

	/* erase phase */
	...

	/* write phase */
	...

	/* verify phase */
	...

this means we decouple the "must add up to 100%" thing and the numbers become just scaled factors. It also means the device state gets set automatically at each stage.

hughsie: I added this as a fixup commit -- I think it's much nicer. Comments welcome.

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,774 @@
...
+{
+	FuProgressPrivate *priv = GET_PRIVATE(self);
+	g_return_val_if_fail(FU_IS_PROGRESS(self), NULL);
+	return priv->id;
+}
+
+/**
+ * fu_progress_set_id:
+ * @self: a #FuProgress
+ * @id: progress ID, normally `G_STRLOC`
```
superm1, line 107: the main reason for this is to tell which plugin you're profiling right?  Considering this wouldn't it make more sense to just use `fu_plugin_get_name`?

superm1: (And if so, then this can be more implicit during initialization)

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,774 @@
...
+}
+
+/**
+ * fu_progress_set_percentage:
+ * @self: a #FuProgress
+ * @percentage: value between 0% and 100%
+ *
+ * Sets the progress percentage complete.
+ *
+ * NOTE: this must be above what was previously set, or it will be rejected.
```
superm1, line 210: Rather than a warning level rejection, how about making it an error and this return a boolean?

libfwupdplugin/fu-progress.c
hughsie: > the main reason for this is to tell which plugin you're profiling right

Actually it's more granular than that. So when you're profiling if you have a FuProgress with `75%, 25%` and then a child one of the first that has `10%, 90%` the profiler tells you actually _the parent one needs to be 30%, 70%_ and _the child one needs to be 50%, 50%_ -- i.e. you need more than the plugin name to know which one the profiler is referring to.

libfwupdplugin/fu-progress.c
hughsie: I also pondered this. I think in the case where the plugin author got the FuProgress stuff wrong we still want to update the firmware on the device, i.e. it's better we do the #1 purpose of the tool and nag on the console than complain to the user that can't actually do anything with the failure.

hughsie:
Moving to https://github.com/fwupd/fwupd/pull/3678 as I've managed to confuse the hell out of github with all my git foo.

PR #3628, opened 2021-08-11
migrate powerd plugin from method call to signal

binaryNewts:
Allow fwupd to use the BatteryStatePoll signal instead of the GetBatteryState method call so that fwupd does not need to run a timer itself and is automatically updated on power connect/disconnect.

Type of pull request:

- [x] Code fix
- [x] Feature
- [ ] Documentation


plugins/powerd/fu-plugin-powerd.c
```diff
@@ -65,40 +65,40 @@ fu_plugin_destroy(FuPlugin *plugin)
 		g_object_unref(data->proxy);
 }
 
-static gboolean
-fu_plugin_powerd_refresh_cb(FuPlugin *plugin)
+static void
+fu_plugin_powerd_rescan(GVariant *parameters, FuPlugin *plugin)
 {
 	FuPluginData *data = fu_plugin_get_data(plugin);
```
superm1, line 71:
```
../plugins/powerd/fu-plugin-powerd.c:71:16: error: unused variable ‘data’ [-Werror=unused-variable]
```

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -65,40 +65,39 @@ fu_plugin_destroy(FuPlugin *plugin)
...
-						 G_DBUS_CALL_FLAGS_NONE,
-						 -1,
-						 NULL,
-						 G_SOURCE_REMOVE);
-	if (powerd_response == NULL) {
-		g_debug("battery information was not loaded");
-		return G_SOURCE_REMOVE;
+
+	if (parameters == NULL) {
+		g_debug("BatteryStatePoll signal has no contents");
```
superm1, line 77: is this a programming error, or caused by an older version of powerd?

superm1: It seems like a programming error, and it should be  `g_return_val_if_null` type of thing

djcampello: I think this is defensive programming and not actually expected to happen

superm1: In that case - I would suggest `g_return_if_fail` for it

plugins/powerd/fu-plugin-powerd.c
```diff
@@ -65,40 +65,39 @@ fu_plugin_destroy(FuPlugin *plugin)
 		g_object_unref(data->proxy);
 }
 
-static gboolean
-fu_plugin_powerd_refresh_cb(FuPlugin *plugin)
+static void
+fu_plugin_powerd_rescan(GVariant *parameters, FuPlugin *plugin)
```
superm1, line 69: I don't really like having the GVariant first.  Nearly every other use that passes the plugin as an argument, the plugin is the first argument.  Would you mind swapping them?

binaryNewts: yeah fs
would you also like it to be swapped so plugin is first on the callback function?

djcampello: You cannot do that... as plugin is holding the gpointer user_data expected in the signature of the callback

binaryNewts: I could swap g_signal_connect() for g_signal_connect_swapped() with the same arguments and it would just put the user_data first (to my understanding of the function)

djcampello: fair enough, I think I rather the g_signal_connect() use to be consistent with other parts of the code (as the upower plugin) but I'll defer to Mario if he thinks otherwise

superm1: I agree with @djcampello let's stay consistent with other parts of the code.

djcampello (approved):

PR #3636, opened 2021-08-13
Add support for synaptics prometheus 009a

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/synaptics-prometheus/synaptics-prometheus.quirk
+++ b/plugins/synaptics-prometheus/synaptics-prometheus.quirk
@@ -6,6 +6,10 @@ InstallDuration = 2
 Plugin = synaptics_prometheus
 InstallDuration = 2
 
+[USB\VID_06CB&PID_009A]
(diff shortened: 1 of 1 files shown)


superm1:
@vhuag something I noticed is the existing quirk file has:
```
[USB\VID_06CB&PID_00A9]
```

Whereas this device that users are reporting is:
```
[USB\VID_06CB&PID_009A]
```

Those are too close to similar.  I need to see if this was just a typo initially.  Can you check if 00A9 is a real device and should be in the quirk file?

superm1:
It seems this alone isn't enough to make the device work.

vhuag:
hello
0x9a is an older sensor that can't be updated by the current plugin, and adding the quirk file isn't enough by the way.

superm1:
> hello
> 0x9a is an older sensor that can't be updated by the current plugin, and adding the quirk file isn't enough by the way.

Thanks for confirming that.

PR #3638, opened 2021-08-13
Forcefully set checksums found in cabinet files to lowercase

superm1:
Archives uploaded to LVFS will have checksums written by hexdigest
and be set to lowercase, but hand generated archives this may not
be true.

For maximum compatibility with the most fwupd versions they should
be written in lowercase, but in case they aren't, convert to lowercase.

Reference https://github.com/fwupd/firmware-lenovo-thinkpad/issues/145

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-cabinet.c
+++ b/libfwupdplugin/fu-cabinet.c
@@ -453,6 +453,46 @@ fu_cabinet_set_container_checksum_cb (XbBuilderFixup *builder_fixup,
 	return TRUE;
 }
 
+static void
+fu_cabinet_fixup_checksum_children(XbBuilderNode *bn,
+				   const gchar *element,
+				   const gchar *attr_name,
+				   const gchar *attr_value)
+{
+	GPtrArray *bcs = xb_builder_node_get_children(bn);
+	for (guint i = 0; i < bcs->len; i++) {
+		XbBuilderNode *bc = g_ptr_array_index(bcs, i);
+		if (g_strcmp0(xb_builder_node_get_element(bc), element) != 0)
+			continue;
+		if (attr_value == NULL ||
+		    g_strcmp0(xb_builder_node_get_attr(bc, attr_name), attr_value) == 0) {
+			const gchar *tmp = xb_builder_node_get_text(bc);
+			if (tmp != NULL) {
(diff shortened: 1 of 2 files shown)


hughsie:
I think this kind of fixup should be done at silo generation time.

libfwupdplugin/fu-cabinet.c
```diff
@@ -453,6 +453,31 @@ fu_cabinet_set_container_checksum_cb (XbBuilderFixup *builder_fixup,
...
+fu_cabinet_set_lowercase_checksum_cb(XbBuilderFixup *builder_fixup,
+				     XbBuilderNode *bn,
+				     gpointer user_data,
+				     GError **error)
+{
+	g_autoptr(XbBuilderNode) csum = NULL;
+
+	/* verify it exists */
+	if (g_strcmp0(xb_builder_node_get_element(bn), "artifact") == 0)
+		csum = _xb_builder_node_get_child_by_element_attr(bn, "checksum", "type", "sha256");
```
hughsie, line 466: I think you want SHA1 (and others?) too. This might mean splitting out something like `fu_cabinet_set_lowercase_checksum` as a function to use in both places.

superm1: See if you think what I did accomplishes that (a little bit different)

libfwupdplugin/fu-cabinet.c
```diff
@@ -453,6 +454,33 @@ fu_cabinet_set_container_checksum_cb (XbBuilderFixup *builder_fixup,
...
+				     gpointer user_data,
+				     GError **error)
+{
+	g_autoptr(XbBuilderNode) csum = NULL;
+
+	/* verify it exists */
+	if (g_strcmp0(xb_builder_node_get_element(bn), "artifact") == 0)
+		/* don't care whether it's sha256, sha1 or something else so don't check for
+		 * specific value */
+		csum = _xb_builder_node_get_child_by_element_attr(bn, "checksum", "type", NULL);
```
hughsie, line 469: Isn't this just fixing up the "first" child? What if md5 was first listed?

superm1: Adjusted it again.

hughsie (approved):

PR #3646, opened 2021-08-17
Various shell and static analysis improvements

superm1:
I was glancing at codacity and fixed a few things that stood out, and also upped shellcheck to warning.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/redfish/tests/redfish.py
+++ b/plugins/redfish/tests/redfish.py
@@ -7,7 +7,7 @@
 
 import json
 
-from flask import Flask, Response, request, g
+from flask import Flask, Response, request
 
 app = Flask(__name__)
 
@@ -21,7 +21,7 @@ def _failure(msg: str, status=400):
     res = {
         "error": {"message": msg},
     }
-    return Response(response=json.dumps(msg), status=401, mimetype="application/json")
+    return Response(response=json.dumps(res), status=401, mimetype="application/json")
 
 
 @app.route("/redfish/v1/")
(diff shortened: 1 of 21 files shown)


plugins/redfish/tests/redfish.py
```diff
@@ -18,9 +18,6 @@
 
 
 def _failure(msg: str, status=400):
-    res = {
-        "error": {"message": msg},
-    }
     return Response(response=json.dumps(msg), status=401, mimetype="application/json")
```
hughsie, line 21: I actually think this should have been json.dumps(res) -- does that look plausible?

plugins/redfish/tests/redfish.py
```diff
@@ -97,7 +94,6 @@ def firmware_inventory_bmc():
         "Manufacturer": "Lenovo",
         "Name": "Lenovo BMC Firmware",
         "RelatedItem": [{"@odata.id": "/redfish/v1/Managers/BMC"}],
-        "ReleaseDate": "2017-08-22T12:00:00",
```
hughsie, line 100: What was the problem with this?

hughsie (approved):
two questions, but looks generally okay. Thanks.

plugins/redfish/tests/redfish.py
superm1: duplicate `ReleaseDate` key

plugins/redfish/tests/redfish.py
superm1: Yeah, looks more reasonable to me.  Will switch.
```
>>> from flask import Response
>>> import json
>>> msg='foo'
>>> Response(response=json.dumps(msg), status=401, mimetype="application/json")
<Response 5 bytes [401 UNAUTHORIZED]>
>>> res = {
...         "error": {"message": msg},
...     }
>>> Response(response=json.dumps(res), status=401, mimetype="application/json")
<Response 29 bytes [401 UNAUTHORIZED]>
```

PR #3650, opened 2021-08-18
FreeBSD CI fix

Asiderr:
The changes fix two problems:
* The extension of the generated artifact is `.pkg`, not `.txz` 
* The generated `pkg-plist` was not taken into account during the package build
The build must be cleaned when the new `pkg-plist` appears. Build system creates stage directories based on the `pkg-plist`, which are used to create the package. 

PR fixes https://github.com/fwupd/fwupd/pull/3578.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,5 +109,5 @@ jobs:
       with:
         name: Binary package
         path: |
-          fwupd*.txz
+          fwupd*.pkg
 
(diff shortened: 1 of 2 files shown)


superm1 (approved):

Asiderr:
@superm1 Sorry for the late reply (vacation time), I will make it up in the coming days.

PR #3665, opened 2021-08-23
Tree Wide: set the format for the whole tree

superm1:
See https://github.com/fwupd/fwupd/discussions/3662 for context 

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


meson.build
```diff
@@ -12,9 +12,9 @@ fwupd_minor_version = varr[1]
 fwupd_micro_version = varr[2]
 
 conf = configuration_data()
-conf.set('FWUPD_MAJOR_VERSION', fwupd_major_version)
-conf.set('FWUPD_MINOR_VERSION', fwupd_minor_version)
-conf.set('FWUPD_MICRO_VERSION', fwupd_micro_version)
+conf.set('MESON_MAJOR_VERSION', fwupd_major_version)
+conf.set('MESON_MINOR_VERSION', fwupd_minor_version)
+conf.set('MESON_MICRO_VERSION', fwupd_micro_version)
```
hughsie, line 17: That doesn't feel right; it's not the meson version, it's the fwupd project version. Maybe drop the MESON_ prefix?

hughsie:
I think we also need some README contents to tell people how to reformat their branch which was created before these changes to the new format. e.g. There are currently about half a dozen branches "in flight" that will need fixing up (by us?) to continue review.

superm1:
> I think we also need some README contents to tell people how to reformat their branch which was created before these changes to the new format. e.g. There are currently about half a dozen branches "in flight" that will need fixing up (by us?) to continue review.

Don't we already have that as part of `CONTRIBUTING.md`?  I'm not sure what else is needed content wise and where it should live.

hughsie (approved):

PR #3666, opened 2021-08-24
dell-dock: handle activation until both ec and usb4 devices are probed

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


CragW:
previously the `needs_activation` flag is only addressed when device is `usb4`, this commit adds `ec` to fix the probing sequence issue.

superm1 (approved):
go ahead and cherry-pick this to earlier branches too so you don't need to wait for 1.7.x for this behavior.

PR #3668, opened 2021-08-25
dell-dock: add device GUID earlier even setup yet done

CragW:
This commit resolves a warning message at runtime:
- no GUIDs for device ...

fu_device_add_child emits a child-added signal afterwards the engine
try to add given device and verify whether GUIDs is set.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -66,6 +66,7 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 	g_autoptr(FuDellDockEc) ec_device = NULL;
 	g_autoptr(FuDellDockMst) mst_device = NULL;
 	g_autoptr(FuDellDockStatus) status_device = NULL;
+	FuContext *ctx = fu_plugin_get_context(plugin);
 
 	/* create ec endpoint */
 	ec_device = fu_dell_dock_ec_new(proxy);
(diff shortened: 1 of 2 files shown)


plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -74,19 +75,26 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 
 	/* create mst endpoint */
 	mst_device = fu_dell_dock_mst_new();
+	dev_guid = DELL_DOCK_VM5331_INSTANCE_ID;
+	if (!fwupd_guid_is_valid(dev_guid))
```
superm1, line 79: isn't this always going to be the case?  It's a hardcoded string

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -74,19 +75,26 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
...
 
 	/* create package version endpoint */
 	status_device = fu_dell_dock_status_new();
 	if (fu_dell_dock_module_is_usb4(FU_DEVICE(ec_device)))
-		instance = "USB\\VID_413C&PID_B06E&hub&salomon_mlk_status";
+		dev_guid = DELL_DOCK_DOCK2_INSTANCE_ID;
 	else
-		instance = "USB\\VID_413C&PID_B06E&hub&status";
+		dev_guid = DELL_DOCK_DOCK1_INSTANCE_ID;
+	if (!fwupd_guid_is_valid(dev_guid))
```
superm1, line 93: same over here.  Maybe it makes sense to just set `dev_guid` to `fwupd_guid_hash_string(foo)` in both cases

CragW: yes, will always true. Removed.  I added back the instance ID as it will be displayed in the client device list for better understanding of the relationship between GUID and sourced instance ID.

superm1 (approved):

PR #3641, opened 2021-08-16
dfu: Introduct a quirk to allow zero polltimeout in dfuDNLOAD.

zhj8407:
Some Poly usb devices report zero in the bwPollTimeout field of
GET_STATUS request. The host can issue the next DFU_DNLOAD
request immediately without any delay.

Introduced a private flag to skip the default DNLOAD timeout
(5ms) fix. It could remarkably reduce the firmware downloading
time taking into account the large firmware (more than 500MB).

Type of pull request:

- [ ] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [X] Code fix
- [ ] Feature
- [ ] Documentation

PR #3678, opened 2021-08-28
Allow using a per-device global percentage completion

hughsie:
and start all over again. Some plugins go 0..100% for write, others
go 0..100% for erase, then again for write, then *again* for verify.

By creating a helper object we can easily split up the progress of the
specific task, e.g. write_firmware().

We can encode at the plugin level "the erase takes 50% of the time, the
write takes 40% and the read takes 10%". This means we can have a
progressbar which goes up just once at a consistent speed.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -54,7 +54,6 @@ typedef struct {
 	GRWLock parent_guids_mutex;
 	GPtrArray *parent_physical_ids; /* (nullable) */
 	guint remove_delay;		/* ms */
-	guint progress;
 	guint battery_level;
 	guint battery_threshold;
 	guint request_cnts[FWUPD_REQUEST_KIND_LAST];
@@ -90,7 +89,6 @@ typedef struct {
 
 enum {
 	PROP_0,
-	PROP_PROGRESS,
 	PROP_BATTERY_LEVEL,
 	PROP_BATTERY_THRESHOLD,
 	PROP_PHYSICAL_ID,
@@ -115,9 +113,6 @@ fu_device_get_property(GObject *object, guint prop_id, GValue *value, GParamSpec
 	FuDevice *self = FU_DEVICE(object);
 	FuDevicePrivate *priv = GET_PRIVATE(self);
 	switch (prop_id) {
-	case PROP_PROGRESS:
-		g_value_set_uint(value, priv->progress);
-		break;
 	case PROP_BATTERY_LEVEL:
 		g_value_set_uint(value, priv->battery_level);
 		break;
@@ -153,9 +148,6 @@ fu_device_set_property(GObject *object, guint prop_id, const GValue *value, GPar
 {
 	FuDevice *self = FU_DEVICE(object);
 	switch (prop_id) {
-	case PROP_PROGRESS:
-		fu_device_set_progress(self, g_value_get_uint(value));
-		break;
 	case PROP_BATTERY_LEVEL:
 		fu_device_set_battery_level(self, g_value_get_uint(value));
 		break;
@@ -3066,123 +3058,6 @@ fu_device_set_remove_delay(FuDevice *self, guint remove_delay)
 }
 
 /**
- * fu_device_get_status:
- * @self: a #FuDevice
- *
- * Returns what the device is currently doing.
- *
- * Returns: the status value, e.g. %FWUPD_STATUS_DEVICE_WRITE
- *
- * Since: 1.0.3
- **/
-FwupdStatus
-fu_device_get_status(FuDevice *self)
-{
-	g_return_val_if_fail(FU_IS_DEVICE(self), 0);
-	return fwupd_device_get_status(FWUPD_DEVICE(self));
-}
-
-/**
- * fu_device_set_status:
- * @self: a #FuDevice
- * @status: the status value, e.g. %FWUPD_STATUS_DEVICE_WRITE
- *
- * Sets what the device is currently doing.
- *
- * Since: 1.0.3
- **/
-void
-fu_device_set_status(FuDevice *self, FwupdStatus status)
-{
-	g_return_if_fail(FU_IS_DEVICE(self));
-	fwupd_device_set_status(FWUPD_DEVICE(self), status);
-}
-
-/**
- * fu_device_get_progress:
- * @self: a #FuDevice
- *
- * Returns the progress completion.
- *
- * Returns: value in percent
- *
- * Since: 1.0.3
- **/
-guint
-fu_device_get_progress(FuDevice *self)
-{
-	FuDevicePrivate *priv = GET_PRIVATE(self);
-	g_return_val_if_fail(FU_IS_DEVICE(self), 0);
-	return priv->progress;
-}
-
-/**
- * fu_device_set_progress:
- * @self: a #FuDevice
- * @progress: the progress percentage value
- *
- * Sets the progress completion.
- *
- * Since: 1.0.3
- **/
-void
-fu_device_set_progress(FuDevice *self, guint progress)
-{
-	FuDevicePrivate *priv = GET_PRIVATE(self);
-	g_return_if_fail(FU_IS_DEVICE(self));
-	if (priv->progress == progress)
-		return;
-	priv->progress = progress;
-	g_object_notify(G_OBJECT(self), "progress");
-}
-
-/**
- * fu_device_set_progress_full:
- * @self: a #FuDevice
- * @progress_done: the bytes already done
- * @progress_total: the total number of bytes
- *
- * Sets the progress completion using the raw progress values.
- *
- * Since: 1.0.3
- **/
-void
-fu_device_set_progress_full(FuDevice *self, gsize progress_done, gsize progress_total)
-{
-	gdouble percentage = 0.f;
-	g_return_if_fail(FU_IS_DEVICE(self));
-	if (progress_total > 0)
-		percentage = (100.f * (gdouble)progress_done) / (gdouble)progress_total;
-	fu_device_set_progress(self, (guint)percentage);
-}
-
-/**
- * fu_device_sleep_with_progress:
- * @self: a #FuDevice
- * @delay_secs: the delay in seconds
- *
- * Sleeps, setting the device progress from 0..100% as time continues.
- * The value is gven in whole seconds as it does not make sense to show the
- * progressbar advancing so quickly for durations of less than one second.
- *
- * Since: 1.5.0
- **/
-void
-fu_device_sleep_with_progress(FuDevice *self, guint delay_secs)
-{
-	gulong delay_us_pc = (delay_secs * G_USEC_PER_SEC) / 100;
-
-	g_return_if_fail(FU_IS_DEVICE(self));
-	g_return_if_fail(delay_secs > 0);
-
-	fu_device_set_progress(self, 0);
-	for (guint i = 0; i < 100; i++) {
-		g_usleep(delay_us_pc);
-		fu_device_set_progress(self, i + 1);
-	}
-}
-
-/**
  * fu_device_set_update_state:
  * @self: a #FuDevice
  * @update_state: the state, e.g. %FWUPD_UPDATE_STATE_PENDING
@@ -3547,6 +3422,7 @@ fu_device_get_results(FuDevice *self, GError **error)
  * fu_device_write_firmware:
  * @self: a #FuDevice
  * @fw: firmware blob
+ * @progress: a #FuProgress
  * @flags: install flags, e.g. %FWUPD_INSTALL_FLAG_FORCE
  * @error: (nullable): optional return location for an error
  *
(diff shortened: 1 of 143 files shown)


hughsie:
@superm1 I've got to do a whole ton of actual flashing on hardware to get actual numbers for the fu_progress_add_step() calls but in my limited testing this works so much better than what we have now. You get a progressbar that goes from 0..100% for each device, rather than pinging back and forwards. It also means if you use a proxy device like in the VLI hubs the percentage and status changes can actually be updated in the transaction. I've fixed a huge number of percentage and status values in this change too. Lots of work still to do, but I think this is a step in the right direction.

hughsie:
@bleungatchromium this is the start of the single-progressbar thing we were discussing all those moons ago.

hughsie:
@superm1 I wanted to show you the difference between the old "`just go 0..100`" to using FuProgress -- this is updating a Wacom tablet. It happens to have an internal child that gets updated too, but to the user it's just **one** device that's being updated:

OLD:
old

This PR:
v2

superm1:
> @superm1 I wanted to show you the difference between the old "`just go 0..100`" to using FuProgress -- this is updating a Wacom tablet. It happens to have an internal child that gets updated too, but to the user it's just **one** device that's being updated:
> 
> OLD:
> old
> 
> This PR:
> v2

wow, that is a significant improvement in experience :)

hughsie:
> wow, that is a significant improvement in experience :)

it's even more dramatic for devices that used to do 0..100% for erase, then write, then verify. It's much nicer for the end user.

hughsie:
@CragW can you please test the dell dock device with this branch please, and also attach a full --verbose output please. I've guessed the values and it would be good to sanity check on actual hardware.

hughsie:
@superm1 are you able to do the same for a emmc device please? I don't have anything updatable here.

hughsie:
@Sean-StarLabs can you do the same thing for superio please? I do have hardware here but I'm trying to share the load a little. Thanks.

hughsie:
FYI, what I'm looking for is something like:

    09:58:00:0915 FuProgress           raw timing data was { 0.002, 21.413, 0.595 } -- steps were set as [ 10 80 10 ] but should have been [ 1 97 2 ] at ../plugins/foo/fu-foo-device.c:3172

superm1:
> @superm1 are you able to do the same for a emmc device please? I don't have anything updatable here.

Sorry I don't have any anymore. TBH I think you want this part of a feedback loop. IOW it should be part of the database and send up with a history report to LVFS if the steps are wrong do that they can be fixed in the future either by metadata or code changes.

hughsie:
>> are you able to do the same for a emmc device
> Sorry I don't have any anymore

No worries, I can pester someone in the ChromeOS group.

> send up with a history report to LVFS

Will ponder. I've added something in the code so we get notified by developers, but I think reporting accurate percentage for unknown devices and also merging in progress data from firmware files is probably phase 2. This PR is already huuuuge.

Sean-StarLabs:
> @Sean-StarLabs can you do the same thing for superio please? I do have hardware here but I'm trying to share the load a little. Thanks.

Doesn't seem to play ball with an IT5570. I'll run it on an IT8987 tomorrow morning.

hughsie:
@Sean-StarLabs I did IT89 just now on the Star Lite. The easiest way I found to do it was `ninja install` then `sudo ./src/fwupdtool --plugins superio install-blob --verbose I2_V14.bin`

src/fu-tool.c
```diff
@@ -837,6 +838,9 @@ fu_util_install_blob(FuUtilPrivate *priv, gchar **values, GError **error)
 		return FALSE;
 	}
 
+	/* progress */
+	fu_progress_set_profile(progress, g_getenv("FWUPD_VERBOSE") != NULL);
+
```
superm1, line 843: although I definitely agree we want this for verbose mode, what about adding a new `--profile` option that also does it?  My thought then it will be less verbose for a less technical contributor to say my values were way off and pull numbers out.

libfwupdplugin/fu-device.h
```diff
@@ -504,9 +507,9 @@ fu_device_get_firmware_size_min(FuDevice *self);
 guint64
 fu_device_get_firmware_size_max(FuDevice *self);
 guint
-fu_device_get_progress(FuDevice *self);
+fu_device_get_progress(FuDevice *self) G_DEPRECATED_FOR(fu_progress_get_percentage);
 void
-fu_device_set_progress(FuDevice *self, guint progress);
+fu_device_set_progress(FuDevice *self, guint progress) G_DEPRECATED_FOR(fu_progress_set_percentage);
```
superm1, line 512: didn't we break ABI already in 1.7.0 but not release yet?  Why not just axe the `fu_device_get_progress` and `fu_device_set_progress` _now_?

hughsie: It was more to avoid upsetting the people working on new plugins (some public, some still secret) that are being worked on behind the scenes. Perhaps we can rip out the symbols before we do the 1.7.0 release?

src/fu-tool.c
hughsie: Maybe? I'm not a far of more options generally, and in the *plugin development* case we'll be using `fwupdtool --plugins foobarbaz get-devices --verbose` anyway.

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,905 @@
...
+	}
+	g_string_append(str, "]");
+	if (!close_enough) {
+#ifdef SUPPORTED_BUILD
+		g_debug("%s at %s", str->str, priv->id);
+#else
+		if (priv->flags & FU_PROGRESS_FLAG_GUESSED) {
+			g_warning("%s at %s", str->str, priv->id);
+			g_warning("If not already fixed on master (please check!), please report "
+				  "this warning to us as a GitHub issue. Thanks!");
```
superm1, line 718: I think putting this into a wiki page and better explaining what this is and what we want in the issue report will be useful.  Maybe even a github bug template for it.

libfwupdplugin/fu-progress.c
```diff
@@ -0,0 +1,905 @@
...
+		FuProgressStep *step = g_ptr_array_index(priv->steps, i);
+		g_string_append_printf(str, "%.0f ", step->profile / division);
+
+		/* this is sufficiently different to what we guessed */
+		if (fabs((gdouble)step->value - step->profile / division) > 5)
+			close_enough = FALSE;
+	}
+	g_string_append(str, "]");
+	if (!close_enough) {
+#ifdef SUPPORTED_BUILD
```
superm1, line 712: Nice idea, I like it.

src/fu-tool.c
superm1: OK - maybe this should be a phase 2 type of idea then if you want a feedback loop or anything like that.

src/fu-device-list.c
```diff
@@ -554,6 +554,14 @@ fu_device_list_item_finalized_cb(gpointer data, GObject *where_the_object_was)
 static void
 fu_device_list_item_set_device(FuDeviceItem *item, FuDevice *device)
 {
+	/* TODO: move the progress to the write_firmware vfunc and break ABI */
```
superm1, line 557: this one too, I think should just be done now.  The folks working on plugins can easily be helped when they submit their PR or rebase on master.

libfwupdplugin/fu-self-test.c
```diff
@@ -3401,6 +3401,224 @@ fu_ifd_image_xml_func(void)
 	g_assert_cmpstr(csum1, ==, csum2);
 }
 
+typedef struct {
+	guint last_percentage;
+	guint updates;
```
superm1, line 3406: both of these variables could just be 8 bits wide

superm1 (commented):
I feel like there still needs to be a way to read this stuff in from a quirk file.  I know there are some more complex cases, but I feel like the cases like synaptics RMI for example should be something that can read in from a quirk and then the write firmware vfunc can populate them into the progress steps from that data.

libfwupdplugin/fu-self-test.c
hughsie: I think keep last_percentage as a guint as then there's no type conversion (i.e. we can detect if we get wonky values)

libfwupdplugin/fu-progress.c
hughsie: Fixed.

hughsie:
> I feel like there still needs to be a way to read this stuff in from a quirk file

This is pretty much required for the detach, and attach steps too; the plugin isn't in control of that FuProgress so we need a way to merge in the parent steps somehow. Will ponder.

src/fu-device-list.c
hughsie: I've done this. Phew!

libfwupdplugin/fu-device.h
hughsie: I ripped them out.

hughsie:
> I feel like there still needs to be a way to read this stuff in from a quirk file

A quirk file didn't work as you needed to know things like "is the bootloader signed" so I implemented a vfunc for the plugin to set the appropriate values.

plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
```diff
@@ -449,6 +456,7 @@ fu_logitech_bulkcontroller_device_write_firmware(FuDevice *device,
 		g_prefix_error(error, "error in writing start transfer packet: ");
 		return FALSE;
 	}
+	fu_progress_step(progress);
```
superm1, line 459: with this new way of doing it, would it make sense to get into a habit of commenting where it's supposed to go?  It might make tracking down mismatches easier later?
```
 fu_progress_step(progress); /* erase */
 fu_progress_step(progress); /* write */
 fu_progress_step(progress); /* busy */
```

libfwupdplugin/fu-device.c
```diff
@@ -3633,7 +3513,7 @@ fu_device_prepare_firmware(FuDevice *self, GBytes *fw, FwupdInstallFlags flags,
 
 	/* optionally subclassed */
 	if (klass->prepare_firmware != NULL) {
-		fu_device_set_status(self, FWUPD_STATUS_DECOMPRESSING);
+		//		fu_progress_set_status(priv->progress, FWUPD_STATUS_DECOMPRESSING);
```
superm1, line 3516: seems like a leftover TODO

plugins/flashrom/fu-flashrom-internal-device.c
```diff
@@ -117,6 +119,12 @@ fu_flashrom_internal_device_write_firmware(FuDevice *device,
 	if (blob_fw == NULL)
 		return FALSE;
 
+	/* progress */
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED);
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 90);
```
superm1, line 125: flashrom doesn't have an erase step?

plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
hughsie: we could do, but the fu_progress_step() signifies the *end* of the action. So maybe `fu_progress_step(progress); /* busy done */` -- or maybe `fu_progress_step()` is the wrong verb, perhaps `fu_progress_step_done()` or `fu_progress_done()` is more obvious.

libfwupdplugin/fu-device.c
hughsie: Indeed, fixed.

plugins/flashrom/fu-flashrom-internal-device.c
hughsie: It does, but it's hidden in `flashrom_image_write()` -- which is yet another frustration with the libflashrom API...

plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
superm1: Yeah I think `step_done` makes more sense.

hughsie: I changed it to `step_done`.

superm1 (approved):

hughsie:
Don't merge yet; I need to test with a ton more hardware on Monday.

superm1:
> Don't merge yet; I need to test with a ton more hardware on Monday.

sure, and need to squash it too.  I just wanted to signal I don't need to check it unless you've got material changes you want me to check again.

plugins/dfu/fu-dfu-device.c
```diff
@@ -1971,7 +1948,4 @@ fu_dfu_device_init(FuDfuDevice *self)
 					FU_DFU_DEVICE_FLAG_NO_BUS_RESET_ATTACH,
 					"no-bus-reset-attach");
 	fu_device_register_private_flag(FU_DEVICE(self), FU_DFU_DEVICE_FLAG_GD32, "gd32");
-	fu_device_register_private_flag(FU_DEVICE(self),
```
zhj8407, line 1974: Sorry for the late comment. Is there any reason to remove this custom flag? It was introduced in https://github.com/fwupd/fwupd/pull/3641/commits/442cd0db4cc7a663aa871756d2ed9ac2009c99b0 for fixing a slow downloading issue with some Poly devices.

hughsie: Removing this was totally unintentional! I'm so sorry! I'll add it back straight away.

hughsie: Okay, I've committed this back into main. I'll do a new upstream tarball this week to include this. I have no idea what happened here, mea culpa.

zhj8407: Thanks!

PR #3682, opened 2021-08-30
Converting security attribute into json and write to hsi_history table.

smallorange:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

Converting all security attributes to JSON and write them into hsi_security tables to record the changing of HSI valuer.
This commit consists of two-part:
1. Add a hsi_history table and schema.
2. JSON conversion and DB writing.

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5908,6 +5908,8 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 	GPtrArray *plugins = fu_plugin_list_get_all(self->plugin_list);
 	g_autoptr(GPtrArray) devices = fu_device_list_get_all(self->device_list);
 	g_autoptr(GPtrArray) items = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autofree gchar *data = NULL;
 
 	/* already valid */
 	if (self->host_security_id != NULL)
@@ -5952,6 +5954,20 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 	/* distil into one simple string */
 	g_free(self->host_security_id);
 	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self);
+
+	/* Convert Security attribute to json string */
+	data = fu_security_attrs_to_json_string(self->host_security_attrs, &error);
+
+	/* Store string to db */
+	if (data == NULL) {
+		g_warning("Fail to convert security attributes to string: %s", error->message);
+	} else {
+		if (fu_history_add_security_attribute(self->history,
+						      data,
+						      self->host_security_id,
+						      &error) == FALSE)
+			g_warning("Fail to write security attribute to DB: %s", error->message);
+	}
 }
 
 const gchar *
--- a/src/fu-history.c
+++ b/src/fu-history.c
@@ -20,7 +20,7 @@
 #include "fu-history.h"
 #include "fu-mutex.h"
 
-#define FU_HISTORY_CURRENT_SCHEMA_VERSION 6
+#define FU_HISTORY_CURRENT_SCHEMA_VERSION 7
 
 static void
 fu_history_finalize(GObject *object);
@@ -183,6 +183,10 @@ fu_history_create_database(FuHistory *self, GError **error)
 			  "checksum TEXT);"
 			  "CREATE TABLE IF NOT EXISTS blocked_firmware ("
 			  "checksum TEXT);"
+			  "CREATE TABLE IF NOT EXISTS hsi_history ("
+			  "timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,"
+			  "hsi_details TEXT DEFAULT NULL,"
+			  "hsi_score TEXT DEFAULT NULL);"
 			  "COMMIT;",
 			  NULL,
 			  NULL,
@@ -307,6 +311,29 @@ fu_history_migrate_database_v5(FuHistory *self, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_history_migrate_database_v6(FuHistory *self, GError **error)
+{
+	gint rc;
+	rc = sqlite3_exec(self->db,
+			  "CREATE TABLE IF NOT EXISTS hsi_history ("
+			  "timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,"
+			  "hsi_details TEXT DEFAULT NULL,"
+			  "hsi_score TEXT DEFAULT NULL);",
+			  NULL,
+			  NULL,
+			  NULL);
+	if (rc != SQLITE_OK) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INTERNAL,
+			    "Failed to create table: %s",
+			    sqlite3_errmsg(self->db));
+		return FALSE;
+	}
+	return TRUE;
+}
+
 /* returns 0 if database is not initialized */
 static guint
 fu_history_get_schema_version(FuHistory *self)
@@ -362,6 +389,10 @@ fu_history_create_or_migrate(FuHistory *self, guint schema_ver, GError **error)
 	/* fall through */
 	case 5:
 		if (!fu_history_migrate_database_v5(self, error))
+			return FALSE;
+	/* fall through */
+	case 6:
+		if (!fu_history_migrate_database_v6(self, error))
 			return FALSE;
 		break;
 	default:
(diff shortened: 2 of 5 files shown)


src/fu-history.c
```diff
@@ -307,6 +311,28 @@ fu_history_migrate_database_v5(FuHistory *self, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_history_migrate_database_v6(FuHistory *self, GError **error)
+{
+	gint rc;
+	rc = sqlite3_exec(self->db,
+			  "CREATE TABLE IF NOT EXISTS his_history ("
```
superm1, line 319: wrong table name

src/fu-history.c
```diff
@@ -1283,6 +1313,44 @@ fu_history_add_blocked_firmware(FuHistory *self, const gchar *checksum, GError *
...
+	locker = g_rw_lock_writer_locker_new(&self->db_mutex);
+	g_return_val_if_fail(locker != NULL, FALSE);
+	rc = sqlite3_prepare_v2(self->db,
+				"INSERT INTO hsi_history (hsi_details, hsi_score)"
+				"VALUES (?1, ?2)",
+				-1,
+				&stmt,
+				NULL);
+	if (rc != SQLITE_OK) {
+		g_warning("write error %s", sqlite3_errmsg(self->db));
```
superm1, line 1341: is the warning really needed?  the same thing is returned as an error

src/fu-engine.c
```diff
@@ -5900,6 +5900,8 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 	GPtrArray *plugins = fu_plugin_list_get_all(self->plugin_list);
 	g_autoptr(GPtrArray) devices = fu_device_list_get_all(self->device_list);
 	g_autoptr(GPtrArray) items = NULL;
+	g_autoptr(GError) error = NULL;
+	gchar *data = NULL;
```
superm1, line 5904: this should be a g_autofree

src/fu-engine.c
```diff
@@ -5944,6 +5946,14 @@ fu_engine_ensure_security_attrs(FuEngine *self)
...
 	g_free(self->host_security_id);
 	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self);
+
+	/* Convert Security attribute to json string */
+	data = fu_security_attrs_to_json_string(self->host_security_attrs);
+
+	/* Store string to db */
+	fu_history_add_security_attribute(self->history, data, self->host_security_id, &error);
+
+	g_free(data);
```
superm1, line 5956: drop when you make the variable `g_autofree`

src/fu-security-attr.c
```diff
@@ -247,3 +250,61 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+	{
+	    "SecurityAttrs": {
+		"Attrs": [
+		    {
+			"name": "aaa",
+			"value": "bbb"
+		    }
+		]
+	    }
+	}
```
superm1, line 267: I think this should be added as a documentation comment for the function explaining this information, or linking elsewhere.

src/fu-security-attr.c
```diff
@@ -247,3 +250,61 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+	*/
+	gchar *data = NULL;
```
superm1, line 269: this should be a g_autofree

src/fu-security-attr.c
```diff
@@ -247,3 +250,61 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+	json_generator_set_root(json_generator, json_root);
+	data = json_generator_to_data(json_generator, NULL);
+	if (data == NULL) {
+		g_warning("Converting to Json string error");
+		return NULL;
+	}
+	json_node_free(json_root);
+	g_object_unref(json_generator);
+	g_object_unref(builder);
+	return data;
```
superm1, line 287: you should steal the pointer here when you make it g_autofree

src/fu-security-attr.c
```diff
@@ -247,3 +250,61 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+			"value": "bbb"
+		    }
+		]
+	    }
+	}
+	*/
+	gchar *data = NULL;
+	JsonNode *json_root = NULL;
+	JsonGenerator *json_generator = NULL;
+	JsonBuilder *builder = json_builder_new();
```
superm1, line 272: these should be g_autoptr

src/fu-security-attr.c
```diff
@@ -247,3 +250,61 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+	fu_security_attrs_to_json(attrs, builder);
+	/* export as a string */
+	json_root = json_builder_get_root(builder);
+	json_generator = json_generator_new();
+	json_generator_set_pretty(json_generator, TRUE);
+	json_generator_set_root(json_generator, json_root);
+	data = json_generator_to_data(json_generator, NULL);
+	if (data == NULL) {
+		g_warning("Converting to Json string error");
```
superm1, line 281: rather than a warning, you should pass error into this function and set an error if this scenario happens (`data` is NULL) and let the caller handle it.

src/fu-history.c
smallorange: I'll fix it.

src/fu-history.c
smallorange: It is not necessary. I'll remove it.

src/fu-engine.c
smallorange: OK! I'll modify all of them.

src/fu-security-attr.c
smallorange: I'll move them before the function and describe the function and JSON format.

src/fu-security-attr.c
smallorange: Thanks. I'll add this part of the codes.

src/fu-engine.c
```diff
@@ -5944,6 +5946,16 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 	/* distil into one simple string */
 	g_free(self->host_security_id);
 	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self);
+
+	/* Convert Security attribute to json string */
+	data = fu_security_attrs_to_json_string(self->host_security_attrs, &error);
```
superm1, line 5951: you need to catch the NULL condition here and display a warning with the error at this time.

src/fu-engine.c
```diff
@@ -5944,6 +5946,16 @@ fu_engine_ensure_security_attrs(FuEngine *self)
...
+
+	/* Store string to db */
+	if (data != NULL)
+		fu_history_add_security_attribute(self->history,
```
superm1, line 5955: you should check the return of this function and display the error if it failed as a warning

src/fu-security-attr.c
```diff
@@ -247,3 +250,72 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+ *             {
+ *                  "name": "aaa",
+ *                  "value": "bbb"
+ *             }
+ *         ]
+ *     }
+ *  }
+ *
+ * Returns: A string and NULL on fail.
+ *
```
superm1, line 274: This is missing a `Since:` for 1.7.0

src/fu-security-attr.c
```diff
@@ -247,3 +250,72 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+ */
+gchar *
+fu_security_attrs_to_json_string(FuSecurityAttrs *attrs, GError **error)
+{
+	g_autofree gchar *data = NULL;
+	JsonNode *json_root = NULL;
```
superm1, line 280: can this also be a g_autoptr?  If not, then it needs to be cleaned up in the failing case below

src/fu-engine.c
smallorange: Sure, I'll display a warning message here if data is NULL.

src/fu-engine.c
smallorange: Sure, An DB writing fail message will be added here.

src/fu-security-attr.c
smallorange: OK. I'll declare it using g_autoptr.

src/fu-security-attr.c
smallorange: I'll add this. Thanks.

superm1 (approved):

hughsie (changes requested):
Can you rebase and squash into one commit please.

src/fu-security-attr.c
```diff
@@ -247,3 +250,72 @@ fu_security_attr_get_result(FwupdSecurityAttr *attr)
...
+	json_generator_set_root(json_generator, json_root);
+	data = json_generator_to_data(json_generator, NULL);
+	if (data == NULL) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INTERNAL,
+			    "Failed to convert security attribute to json.");
+		return NULL;
+	}
+	return (gchar *)g_steal_pointer(&data);
```
hughsie, line 298: i don't think you need the cast

smallorange: I'll remove it.

src/fu-history.c
```diff
@@ -307,6 +311,28 @@ fu_history_migrate_database_v5(FuHistory *self, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_history_migrate_database_v6(FuHistory *self, GError **error)
+{
+	gint rc;
+	rc = sqlite3_exec(self->db,
+			  "CREATE TABLE IF NOT EXISTS hsi_history ("
+			  "last timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,"
```
hughsie, line 320: out of interest, what's the `last` keyword here?

src/fu-history.c
```diff
@@ -307,6 +311,28 @@ fu_history_migrate_database_v5(FuHistory *self, GError **error)
...
+			  "hsi_details TEXT DEFAULT NULL);",
```
hughsie, line 321: Don't you need `hsi_score` here too?

src/fu-history.c
smallorange: It's a column name.

src/fu-history.c
smallorange: Ops. I'll correct this.

src/fu-history.c
hughsie: Ohh, I think you might be copying my typo of `created timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP` -- I think the `timestamp` is probably duplicated for no reason. I think you probably want `timestamp TIMESTAMP...` for the column to be called `timestamp`.

smallorange: You are right!  Let me correct it.

smallorange:
timestamp is good. Let me revise it.

hughsie (approved):

PR #3689, opened 2021-08-31
logitech-hidpp: Add support for updating Bolt receivers

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/logitech-hidpp/fu-logitech-hidpp-device.c
```diff
@@ -32,6 +33,15 @@
...
 #define FU_LOGITECH_HIDPP_DEVICE_FLAG_REBIND_ATTACH (1 << 2)
 
+/**
+ * FU_LOGITECH_HIDPP_DEVICE_FLAG_ADD_RADIO:
+ *
+ * The device should add a softdevice (index 0x5), typically a radio.
+ *
+ * Since: 1.7.0
+ */
+#define FU_LOGITECH_HIDPP_DEVICE_FLAG_ADD_RADIO (1 << 5)
```
superm1, line 43: why the jump from 2 to 5?

hughsie: I'm chunking a mega commit -- the 3 and 4 will appear soon

plugins/logitech-hidpp/fu-logitech-hidpp-radio.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+static void
+fu_logitech_hidpp_radio_init(FuLogitechHidPpRadio *self)
+{
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
+	fu_device_add_internal_flag(FU_DEVICE(self),
+				    FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_BATTERY);
+	fu_device_add_protocol(FU_DEVICE(self), "com.logitech.unifyingsigned");
```
superm1, line 91: is radio always unsigned?

hughsie: yes, radio is always signed.

plugins/logitech-hidpp/fu-logitech-hidpp-device.c
```diff
@@ -351,13 +361,42 @@ fu_logitech_hidpp_device_feature_get_idx(FuLogitechHidPpDevice *self, guint16 fe
...
+	FuContext *ctx = fu_device_get_context(FU_DEVICE(self));
+	g_autofree gchar *instance_id = NULL;
+	g_autofree gchar *radio_version = NULL;
+	g_autoptr(FuLogitechHidPpRadio) radio = NULL;
+
+	radio_version = g_strdup_printf("0x%.4x", build);
+	radio = fu_logitech_hidpp_radio_new(ctx, entity);
+	fu_device_set_physical_id(FU_DEVICE(radio), fu_device_get_physical_id(FU_DEVICE(self)));
+
+	fu_device_set_logical_id(FU_DEVICE(radio), priv->model_id);
```
superm1, line 380: don't you have a case here that `priv->model_id` could potentially be `NULL` and make a bad instance ID for a "new" device not yet supported?  To be safe, I think you should guard against this when creating children.

plugins/logitech-hidpp/fu-logitech-hidpp-device.c
superm1: ah okay.

plugins/logitech-hidpp/fu-logitech-hidpp-device.c
hughsie: Yes, good point. It works now, but without a check we're causing our future selves some head scratching. I've added an explicit check now.

PR #3694, opened 2021-09-01
logitech-hidpp: Add support for updating Bolt peripheral firmware

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
This doesn't actually 100% work at the moment, but I think it's the way we want to do this. BLE still pending.

data/device-tests/devices/logitech-mxmaster3b.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "Logitech MX Master 3B",
```
hardboprobot, line 2: This and the file name should be made generic, I guess.

plugins/logitech-hidpp/README.md
```diff
@@ -74,6 +75,17 @@ the bootloader and runtime modes are treated as the same device.
 The Bolt receiver enumerates as a hidraw device both in runtime and
 bootloader mode, but with different HIDRAW devIDs.
 
+Peripherals paired to a receiver are enumerated as separate hidraw
+devices but those device files can't be used for DFU. Instead, all the
+DFU-related messages need to be piped through the hidraw device file of
+the receiver. They are polled and queried by the receiver and listed as
+its children. Note that this will likely change once the Logitech Linux
+driver supports Bolt devices.
```
superm1, line 83: hopefully not meaning that there is a regression coming down the road!

data/device-tests/devices/logitech-mxmaster3b.json
hughsie: Fixed.

plugins/logitech-hidpp/README.md
hughsie: Fingers crossed...

plugins/logitech-hidpp/fu-logitech-hidpp-device.c
```diff
@@ -676,6 +679,17 @@ fu_logitech_hidpp_device_probe(FuDevice *device, GError **error)
...
 	fu_device_add_vendor_id(device, "USB:0x046D");
 
+	/*
+	 * Devices connected to Bolt receivers share the same physical
+	 * id. Make them unique by using the pairing slot as the logical
+	 * id.
+	 */
+	if (priv->hidpp_id != HIDPP_DEVICE_ID_UNSET) {
+		g_autoptr(GString) id_str = g_string_new(NULL);
+		g_string_append_printf(id_str, "%d", priv->hidpp_id);
```
hughsie, line 689: maybe something like `"HIDPP_ID=%d"` for our future selves?

hardboprobot: Now that I think about it, `hidpp_id` is kind of a misnomer and it might lead to further confusion in the future. In the specs it's referred to as the device index, so I think it's better to rename the `priv->hidpp_id` attribute to `device_idx` and document its uses and possible values. New commit added.

hughsie:
Please!

hardboprobot (approved):

Issue #3701, opened 2021-09-03
Dell WD19: fwupd uses synaptics_mst plugin to probe Dell WD19's VMM5331 mst device


CragW:
**Describe the bug**
I'm seeing a reported issue that has error prints `failed to set quirk key` from the console output, it seems `fwupd` wrongly used `synaptics_mst` plugin to probe dock's `VMM5331` device. 

I was told Issue is reproducible on Chrome OS R96 14192 (fwupd 1.6.3) as well as Ubuntu 20.04 (snap variant 1.6.2), still trying to recreate by myself.

```
01:36:36:0239 FuDevice             failed to set quirk key DellDockUnlockTarget=9: quirk key not supported
01:36:36:0239 FuDevice             failed to set quirk key DellDockInstallDurationI2C=360: quirk key not supported
01:36:36:0239 FuDevice             failed to set quirk key DellDockBlobMajorOffset=0x18400: quirk key not supported
01:36:36:0239 FuDevice             failed to set quirk key DellDockBlobMinorOffset=0x18401: quirk key not supported
01:36:36:0239 FuDevice             failed to set quirk key DellDockBlobBuildOffset=0x18402: quirk key not supported
```
![X10_Dorset_R94_WD19TBT](https://user-images.githubusercontent.com/30334596/131947805-11be9cde-7e0b-4fac-9837-56dd0bd5fa86.png)

**Steps to Reproduce**
fwupdtool get-devices

**Expected behavior**
In `dell-dock` plugin already set the priority for `dell-dock` over `synaptics_mst` on `VMM5331` mst device specifically, device shouldn't be added by `synaptics_msc` plugin.

**fwupd version information**
`fwupdtool --version`
client version: 1.6.3

**Dock SKU**
Please mention which module is installed in your WD19.

- [x] WD19 (Single-C)
- [x] WD19TB (Thunderbolt)
- [ ] WD19DC (Dual-C)

**Verbose daemon logs**
[8h4n0g3_all_ng.log](https://github.com/fwupd/fwupd/files/7103251/8h4n0g3_all_ng.log)

```
09:24:23:0480 FuMain               ADDED:
FuSynapticsMstDevice:
VMM5331 in Dell dock
  DeviceId:             585c346c009da3cc19a400fb59a4ed8d0cbd0151
  CompositeId:          b5556eda4eb2315f490b0f72884f8c891c055274
  Guid:                 b3908108-9d2a-5f8e-83d7-729707550d27 <- MST-259
[...]

PR #3705, opened 2021-09-03
synaptics-mst: Only probe devices that have opted-in

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3701

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1219,6 +1219,13 @@ fu_common_get_path(FuPathKind path_kind)
 		if (tmp != NULL)
 			return g_build_filename(tmp, FWUPD_DATADIR, PACKAGE_NAME, NULL);
 		return g_build_filename(FWUPD_DATADIR, PACKAGE_NAME, NULL);
+	/* /usr/share/fwupd/quirks.d */
+	case FU_PATH_KIND_DATADIR_QUIRKS:
+		tmp = g_getenv("FWUPD_DATADIR_QUIRKS");
+		if (tmp != NULL)
+			return g_strdup(tmp);
+		basedir = fu_common_get_path(FU_PATH_KIND_DATADIR_PKG);
+		return g_build_filename(basedir, "quirks.d", NULL);
 	/* /usr/libexec/fwupd/efi */
 	case FU_PATH_KIND_EFIAPPDIR:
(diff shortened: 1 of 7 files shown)


superm1 (approved):

CragW (approved):

hughsie:
Urgh, I have to fix up the tests to actually load the quirks. Will ponder and fix on Monday. One idea is to move all the quirks in the plugins to individual `quirk.d` directories and then override `FWUPD_DATADIR` in the self tests to be the plugin source directory. This might be a nice tidy-up as some plugins have quite a few quirk files now.

The other way is to do a `ninja install` before a `ninja test` although this seems a bit of a hack.

Other other, other way is to inject an extra fake path in `fu_quirks_add_quirks_for_path`, perhaps reading a `FWUPD_QUIRK_FILE` env var, or a `FWUPD_EXTRA_QUIRK_PATH` or something. Other ideas welcome, or a sanity check on what I just wrote :)

superm1:
> Urgh, I have to fix up the tests to actually load the quirks. Will ponder and fix on Monday. One idea is to move all the quirks in the plugins to individual `quirk.d` directories and then override `FWUPD_DATADIR` in the self tests to be the plugin source directory. This might be a nice tidy-up as some plugins have quite a few quirk files now.

Why do they need to be moved?  Doesn't the quirk detection already look for `.quirk` extension?  Just set the env variable to the plugin source and call it good.

> 
> The other way is to do a `ninja install` before a `ninja test` although this seems a bit of a hack.

No, the tests should be able to run without installing.  I agree this is a hack and to be avoided.

> 
> Other other, other way is to inject an extra fake path in `fu_quirks_add_quirks_for_path`, perhaps reading a `FWUPD_QUIRK_FILE` env var, or a `FWUPD_EXTRA_QUIRK_PATH` or something. Other ideas welcome, or a sanity check on what I just wrote :)

If setting the data directory fails, sure another extra test specific environment variable is another way to do it.

hughsie:
>  Just set the env variable to the plugin source and call it good.

We have to define things like `FU_PATH_KIND_DATADIR_QUIRKS` -- but that's probably a good thing to do anyway.

superm1 (approved):

PR #3706, opened 2021-09-03
Run self tests as nobody for at least one CI target

superm1:
Beyond the failures that happened in ATA and NVME plugins, other parts in the codebase are making writes into `data/tests` or `plugins/synaptics-mst/tests`.  We shouldn't be allowing this at all.  Fix those and then enforce that everything touched is in `build/` or `/tmp` during the run.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/synaptics-mst/fu-synaptics-mst-device.c
+++ b/plugins/synaptics-mst/fu-synaptics-mst-device.c
@@ -79,15 +79,17 @@ fu_synaptics_mst_device_finalize(GObject *object)
 static void
 fu_synaptics_mst_device_init(FuSynapticsMstDevice *self)
 {
+	FuUdevDeviceFlags flags =
+	    FU_UDEV_DEVICE_FLAG_OPEN_READ | FU_UDEV_DEVICE_FLAG_VENDOR_FROM_PARENT;
+	if (fu_udev_device_get_dev(FU_UDEV_DEVICE(self)) != NULL)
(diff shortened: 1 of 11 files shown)


plugins/synaptics-mst/fu-synaptics-mst-device.c
```diff
@@ -79,15 +79,17 @@ fu_synaptics_mst_device_finalize(GObject *object)
 static void
 fu_synaptics_mst_device_init(FuSynapticsMstDevice *self)
 {
+	FuUdevDeviceFlags flags =
+	    FU_UDEV_DEVICE_FLAG_OPEN_READ | FU_UDEV_DEVICE_FLAG_VENDOR_FROM_PARENT;
+	if (fu_udev_device_get_dev(FU_UDEV_DEVICE(self)) != NULL)
+		flags |= FU_UDEV_DEVICE_FLAG_OPEN_WRITE;
```
hughsie, line 85: Is this safe? Do we ever set the udev dev after the mst device has been set up?

superm1: We shouldn't be at least..  Synaptics MST device inherits from a udev device, and that's the thing we're setting flags on

hughsie (approved):
Lets wait for CI, but LGTM, thanks for figuring this one out.

PR #3711, opened 2021-09-05
Allow using a filename when using set-approved-firmware

hughsie:
See https://gitlab.com/fwupd/lvfs-website/-/merge_requests/924 for more
discussion.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -2439,12 +2439,21 @@ fu_util_set_approved_firmware(FuUtilPrivate *priv, gchar **values, GError **erro
 		g_set_error_literal(error,
 				    FWUPD_ERROR,
 				    FWUPD_ERROR_INVALID_ARGS,
-				    "Invalid arguments: list of checksums expected");
+				    "Invalid arguments: filename or list of checksums expected");
 		return FALSE;
 	}
 
+	/* filename */
+	if (g_file_test(values[0], G_FILE_TEST_EXISTS)) {
+		g_autofree gchar *data = NULL;
+		if (!g_file_get_contents(values[0], &data, NULL, error))
+			return FALSE;
+		checksums = g_strsplit(data, "\n", -1);
+	} else {
+		checksums = g_strsplit(values[0], ",", -1);
+	}
+
 	/* call into daemon */
-	checksums = g_strsplit(values[0], ",", -1);
 	return fwupd_client_set_approved_firmware(priv->client,
 						  checksums,
 						  priv->cancellable,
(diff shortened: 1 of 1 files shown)


superm1 (commented):
The idea being its just a raw list of strings of checksums not a conf file?

What's the workflow for using it/managing it?

hughsie:
> The idea being its just a raw list of strings of checksums not a conf file?

Yes, but this PR was more of an invitation to discuss all this. My idea was that "central IT" however you want to define that would download a allow-list of firmware from a central location (ansible etc) and then "sync" fwupd using set-approved-firmware. Using a huge string on the command line doesn't really work much past a dozen or so.

> What's the workflow for using it/managing it?

Internally the allowlist is stored in the DB, so maybe it makes sense to expose an "append" interface, or even a SetApprovedChecksum() RemoveApprovedChecksum interface. The problem with the latter is that you're not sure *what was set before* and so it's not just like you can just fire a command once per day with the latest approved firmwares.

superm1:
>Yes, but this PR was more of an invitation to discuss all this. My idea was that "central IT" however you want to define that would download a allow-list of firmware from a central location (ansible etc) and then "sync" fwupd using set-approved-firmware. Using a huge string on the command line doesn't really work much past a dozen or so.

Right, I agree it shouldn't just be a giant list on a command line.

>Internally the allowlist is stored in the DB, so maybe it makes sense to expose an "append" interface, or even a SetApprovedChecksum() RemoveApprovedChecksum interface. The problem with the latter is that you're not sure what was set before and so it's not just like you can just fire a command once per day with the latest approved firmwares.

Does it _need_ to be on the DB though?  If the flow we end up enabling is a conffile or a raw list of checksums, then fwupd could potentially just parse this at startup and mark all those things, but admittedly it will make startup slower the longer it goes on.

hughsie:
> Does it need to be on the DB though?

Of course, we could let other tools write into the db, but then it becomes AABI and not some internal implementation detail. I also think it's the "_too hard, we give up_" style of API design.

>  but admittedly it will make startup slower the longer it goes on.

Well, it would be maximum 3000 lines at the moment; not anything to worry about from a performance point of view. Perhaps we could avoid reloading the file if the mtime hasn't changed or something?

superm1:
>Of course, we could let other tools write into the db, but then it becomes AABI and not some internal implementation detail. I also think it's the "too hard, we give up" style of API design.

Oh I didn't mean let other software access the DB. I meant that don't store it at all.  Just in memory.  You load the file every time you start the daemon or the file changes.  Then you don't have to keep track of what is changing every time it's pushed down by "central management".

>Well, it would be maximum 3000 lines at the moment; not anything to worry about from a performance point of view. Perhaps we could avoid reloading the file if the mtime hasn't changed or something?

Well if you do what I say that it loads every time then there might be an impact.

hughsie:
>  Just in memory

Not sure I want to store thousands of strings in RSS, even as a hash. I think we can certainly do something with libxmlb here, storing as a silo, and querying using XPath. That's probably better than the sqlite database too.

PR #3716, opened 2021-09-06
Wip/hughsie/native impi

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/redfish/fu-ipmi-device.c
+++ b/plugins/redfish/fu-ipmi-device.c
@@ -1,0 +1,593 @@
+/*
+ * Copyright (C) 2021 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include <fcntl.h>
+#include <glib/gstdio.h>
+#include <linux/ipmi.h>
+#include <linux/ipmi_msgdefs.h>
+#include <poll.h>
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "fu-ipmi-device.h"
+
+#define FU_IPMI_DEVICE_TIMEOUT 1500 /* ms */
+
+/* not defined in linux/ipmi_msgdefs.h */
+#define IPMI_SET_USER_ACCESS   0x43
+#define IPMI_SET_USER_NAME     0x45
+#define IPMI_GET_USER_NAME     0x46
+#define IPMI_SET_USER_PASSWORD 0x47
+
+#define IPMI_PASSWORD_DISABLE_USER  0x00
+#define IPMI_PASSWORD_ENABLE_USER   0x01
+#define IPMI_PASSWORD_SET_PASSWORD  0x02
+#define IPMI_PASSWORD_TEST_PASSWORD 0x03
+
+struct _FuIpmiDevice {
+	FuUdevDevice parent_instance;
+	glong seq;
+	guint8 device_id;
+	guint8 device_rev;
+	guint8 version_ipmi;
+};
+
+G_DEFINE_TYPE(FuIpmiDevice, fu_ipmi_device, FU_TYPE_UDEV_DEVICE)
+
+static void
+fu_ipmi_device_to_string(FuDevice *device, guint idt, GString *str)
+{
+	FuIpmiDevice *self = FU_IPMI_DEVICE(device);
+	fu_common_string_append_kx(str, idt, "DeviceId", self->device_id);
+	fu_common_string_append_kx(str, idt, "DeviceRev", self->device_rev);
+	fu_common_string_append_kx(str, idt, "VersionIpmi", self->version_ipmi);
+}
+
+static gboolean
+fu_ipmi_device_send(FuIpmiDevice *self,
+		    guint8 netfn,
+		    guint8 cmd,
+		    const guint8 *buf,
+		    gsize bufsz,
+		    GError **error)
+{
+	g_autofree guint8 *buf2 = fu_memdup_safe(buf, bufsz, NULL);
+	struct ipmi_system_interface_addr addr = {.addr_type = IPMI_SYSTEM_INTERFACE_ADDR_TYPE,
+						  .channel = IPMI_BMC_CHANNEL};
+	struct ipmi_req req = {
+	    .addr = (guint8 *)&addr,
+	    .addr_len = sizeof(addr),
+	    .msgid = self->seq++,
+	    .msg.data = buf2,
+	    .msg.data_len = (guint16)bufsz,
+	    .msg.netfn = netfn,
+	    .msg.cmd = cmd,
+	};
+	if (g_getenv("FWUPD_REDFISH_VERBOSE") != NULL)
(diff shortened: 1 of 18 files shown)


hughsie:
FYI: untested as a whole, but I've tested each bit. I'll be able to test this tomorrow.

hughsie:
With a few little fixups this works beautifully on real server hardware.

plugins/redfish/fu-ipmi-device.c
```diff
@@ -0,0 +1,573 @@
...
+	if (g_getenv("FWUPD_REDFISH_VERBOSE") != NULL)
+		fu_common_dump_raw(G_LOG_DOMAIN, "ipmi-recv", buf, *len);
+	if (netfn != NULL)
+		*netfn = recv.msg.netfn;
+	if (cmd != NULL)
+		*cmd = recv.msg.cmd;
+	if (seq != NULL)
+		*seq = recv.msgid;
+	if (len != NULL)
+		*len = (gsize)recv.msg.data_len;
```
superm1, line 116: should any of these set an error?

plugins/redfish/fu-ipmi-device.c
```diff
@@ -0,0 +1,573 @@
...
+		}
+		fu_device_set_version(device, str->str);
+		bcd = resp[4] & 0x0f;
+		bcd += 10 * (resp[4] >> 4);
+		self->version_ipmi = bcd;
+	} else {
+		g_set_error(error,
+			    G_IO_ERROR,
+			    G_IO_ERROR_NOT_SUPPORTED,
+			    "failed to retrieve Device ID from IPMI");
```
superm1, line 347: including resp_length in the error perhaps?

plugins/redfish/fu-ipmi-device.c
hughsie: I don't _think_ so. `fu_ipmi_device_recv` just gets the response, no matter the contents.

plugins/redfish/fu-ipmi-device.c
hughsie: Good idea, fixed.

plugins/redfish/fu-ipmi-device.c
```diff
@@ -0,0 +1,573 @@
...
+	FuDeviceClass *klass_device = FU_DEVICE_CLASS(klass);
+	klass_device->setup = fu_ipmi_device_setup;
+	klass_device->to_string = fu_ipmi_device_to_string;
+}
+
+FuIpmiDevice *
+fu_ipmi_device_new(FuContext *ctx)
+{
+	FuIpmiDevice *self;
+	self = g_object_new(FU_TYPE_IPMI_DEVICE, "context", ctx, "device-file", "/dev/ipmi0", NULL);
```
superm1, line 571: can't this also be `/dev/ipmidev/0`?  Or are those just symlinks?

hughsie: I don't see any of `/dev/ipmidev` at all on RHEL/Fedora.

plugins/redfish/fu-ipmi-device.c
```diff
@@ -0,0 +1,573 @@
...
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_ipmi_device_init(FuIpmiDevice *self)
+{
+	fu_device_set_physical_id(FU_DEVICE(self), "/dev/ipmi0");
```
superm1, line 552: can this come from udev instead?  If not, do we need to also make sure that we load `impi-si` and `ipmi-devintf`?

plugins/redfish/fu-ipmi-device.c
superm1: Okay.

plugins/redfish/fu-ipmi-device.c
hughsie: not really; there's no `udev_device` at all in FuUdevDevice, it's literally a hardcoded `device-file` we're using.

plugins/redfish/fu-ipmi-device.c
superm1: I don't know the context of it - but I see a few different things tried in ipmitool...
https://github.com/ipmitool/ipmitool/blob/d674bfdc9a2447706ea9b0b943c98bcb0bbe80d2/src/plugins/open/open.c#L102

plugins/redfish/fu-ipmi-device.c
hughsie: > make sure that we load impi-si and ipmi-devintf

They seem loaded by default on Fedora and RHEL, is that a downstream thing?

plugins/redfish/fu-ipmi-device.c
hughsie: I had no idea they existed. I don't think I want to support more than one BMC device (!!!) but I guess it makes sense to search for /dev/ipmi/0 and /dev/ipmidev/0 too

plugins/redfish/fu-ipmi-device.c
superm1: What is `CONFIG_IPMI_DEVICE_INTERFACE` set to?  I don't see anything for modaliases for the ipmi-devintf module in https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/drivers/char/ipmi/ipmi_devintf.c so unless it's compiled into the kernel I would think that module needs to get loaded _somehow_.

hughsie: > What is CONFIG_IPMI_DEVICE_INTERFACE set to

CONFIG_IPMI_DEVICE_INTERFACE=m

Maybe it's safer to manually load impi-si and ipmi-devintf? I'll do just that.

hughsie: Fixed.

plugins/redfish/fu-ipmi-device.c
hughsie: Fixed.

superm1 (dismissed):

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -181,6 +185,89 @@ fu_redfish_plugin_discover_smbios_table(FuPlugin *plugin, GError **error)
...
+	}
+	return g_string_free(str, FALSE);
+}
+
+static gboolean
+fu_redfish_plugin_ipmi_create_user(FuPlugin *plugin, GError **error)
+{
+	FuPluginData *data = fu_plugin_get_data(plugin);
+	const gchar *username_fwupd = "fwupd";
+	guint8 user_id = 0x04;
```
tcap26, line 208: looks like you hardcoded the user_id = 0x04 ..... you probably need to add some logic to handle cases where 0x04 is used by other users, where you need to pick another channel number for user_id.

hughsie: Yup, certainly a valid criticism. I guess the best thing to do at startup is:

 * look from `user_id` 2 to 63 for a user called fwupd
 * if found, report an error as we didn't save the generated password and can't delete the user
 * if not found, create a fwupd user in the first unused slot and continue to enumerate

@tcap26 does that sound [more] sane?

PR #3717, opened 2021-09-07
trivial: system firmware may primarily use sha256 bank instead sha1

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Fixes:
tpm_evt_log_nulled_sha1

### DIFF
--- a/plugins/tpm-eventlog/fu-tpm-eventlog-device.c
+++ b/plugins/tpm-eventlog/fu-tpm-eventlog-device.c
@@ -48,7 +48,13 @@ fu_tpm_eventlog_device_report_metadata(FuTpmEventlogDevice *self)
 	for (guint i = 0; i < self->items->len; i++) {
 		FuTpmEventlogItem *item = g_ptr_array_index(self->items, i);
 		g_autofree gchar *blobstr = fu_tpm_eventlog_blobstr(item->blob);
-		g_autofree gchar *checksum = fu_tpm_eventlog_strhex(item->checksum_sha1);
+		g_autofree gchar *checksum = NULL;
+		if (item->checksum_sha1 != NULL)
+			checksum = fu_tpm_eventlog_strhex(item->checksum_sha1);
+		else if (item->checksum_sha256 != NULL)
+			checksum = fu_tpm_eventlog_strhex(item->checksum_sha256);
+		else
+			continue;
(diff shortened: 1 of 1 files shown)


plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -48,7 +48,11 @@ fu_tpm_eventlog_device_report_metadata(FuTpmEventlogDevice *self)
 	for (guint i = 0; i < self->items->len; i++) {
 		FuTpmEventlogItem *item = g_ptr_array_index(self->items, i);
 		g_autofree gchar *blobstr = fu_tpm_eventlog_blobstr(item->blob);
-		g_autofree gchar *checksum = fu_tpm_eventlog_strhex(item->checksum_sha1);
+		g_autofree gchar *checksum = NULL;
+		if (item->checksum_sha1)
+			checksum = fu_tpm_eventlog_strhex(item->checksum_sha1);
+		else if (item->checksum_sha256)
+			checksum = fu_tpm_eventlog_strhex(item->checksum_sha256);
```
hughsie, line 55: ohh good catch. Can you add a:

    else
        continue;

and then we can cover ourselves for the future.

plugins/tpm-eventlog/fu-tpm-eventlog-device.c
```diff
@@ -48,7 +48,13 @@ fu_tpm_eventlog_device_report_metadata(FuTpmEventlogDevice *self)
 	for (guint i = 0; i < self->items->len; i++) {
 		FuTpmEventlogItem *item = g_ptr_array_index(self->items, i);
 		g_autofree gchar *blobstr = fu_tpm_eventlog_blobstr(item->blob);
-		g_autofree gchar *checksum = fu_tpm_eventlog_strhex(item->checksum_sha1);
+		g_autofree gchar *checksum = NULL;
+		if (item->checksum_sha1)
```
hughsie, line 52: actually, can you also add `!= NULL` to both `if`s please -- I think these are pointers, right?

CragW: Ah that's right, okay will add to prevent headache.

hughsie (approved):

PR #3721, opened 2021-09-07
trivial: create urlmap for each gi-docgen target rather than sharing

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
I guess that's a symlink, right? Precommit is failing, but after that LGTM. Thanks.

PR #3730, opened 2021-09-09
plugins/superio: Add IT5570 on Port 0x4e

Sean-StarLabs:
Add standard IT5570 that uses Port 0x4e.

### DIFF
--- a/plugins/superio/superio.quirk
+++ b/plugins/superio/superio.quirk
@@ -72,3 +72,7 @@ SuperioControlPort = 0x66
 SuperioDataPort = 0x62
 SuperioAutoloadAction = disable
 SuperioTimeout = 650
+
+[SuperIO-IT5570-4e]
+SuperioId = 0x5570
+SuperioPort = 0x4e


plugins/superio/superio.quirk
```diff
@@ -72,3 +72,7 @@ SuperioControlPort = 0x66
 SuperioDataPort = 0x62
 SuperioAutoloadAction = disable
 SuperioTimeout = 650
+
+[SuperIO-IT5570-4e]
```
hughsie, line 76: what's the `-4e` all about?

Sean-StarLabs: [[SuperIO-IT5570] was taken (and doesn't work - in this case at least)

hughsie: ...but doesn't the plugin need to read that specific key? e.g. https://github.com/fwupd/fwupd/blob/master/plugins/superio/fu-superio-device.c#L276

Sean-StarLabs: Shall I change it to strip -`02e` and `-4e`?

I think I'll annoy Tuxedo(?) if I change the existing one...

hughsie: So the port isn't specific to the SuperIO chip? Is the port actually specific to the platform (e.g. DMI?)

Sean-StarLabs: Nope, I've seen IT8987E's using 0x2e.

For coreboot at least, it's declared in the board under 1f.0 and there's no way 
to see what that is.

_Might_ be able to query the chip ID on both ports and use whichever one matches?

hughsie: What's the worst case consequence of accessing the *wrong* one?

Sean-StarLabs: Writing to it? Or query chip ID?

hughsie: The latter, although I think I've answered by own question...

Sean-StarLabs: Shame, I was going to say `Hold my beer` and wait for it to start smoking...

IT5570 doesn't cause any issues, would have to check IT8987 tomorrow.

Sean-StarLabs: Doesn't cause in issue on IT8987 either

hughsie: Okay, lets just make this a bit simpler and explicit. Would something like https://github.com/fwupd/fwupd/pull/3734 work?

Issue #3732, opened 2021-09-10
Dell dock: proxy device is stalled


CragW:
**Describe the bug**
Failed to install mst firmware, error message is seen: `Failed to query active bank: failed after 5 retries: failed to SetReport: endpoint stalled or request not supported`

<img width="449" alt="Screenshot from 2021-09-10 15-54-30" src="https://user-images.githubusercontent.com/30334596/132821515-c0b9ad69-aea2-471d-a1d7-53248fcf740b.png">

**Steps to Reproduce**
`# fwupdtool install dell-dock.cab -v --allow-reinstall`

**Expected behavior**
mst firmware is applied successfully, no error message.

**fwupd version information**
client version:	1.6.1-174-g15b668c0c
compile-time dependency versions
	gusb:	0.3.5

**Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):**
source

**Verbose daemon logs**
[15b668c0c-proxy-device-failed.log](https://github.com/fwupd/fwupd/files/7142104/15b668c0c-proxy-device-failed.log)

**Additional info**
- Operating system and version: Ubuntu 21.04
- Is this a regression? yes
- Commit 15b668c0c is causing this failure.
- Issue is unseen with previous commit a0fd32f41, verbose log [a0fd32f41-proxy-device-passed.log](https://github.com/fwupd/fwupd/files/7142140/a0fd32f41-proxy-device-passed.log)

hughsie:
Okay, super-uncool, sorry about the regression. Lets make the proxy fallback an opt-in flag. Give me 10 minutes.

PR #3733, opened 2021-09-10
Fix a regression in flashing the Dell dock

hughsie:
This fixes a regression caused by 15b668c, we should have been more careful.

Fixes https://github.com/fwupd/fwupd/issues/3732

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.h
+++ b/libfwupdplugin/fu-device.h
@@ -378,6 +378,15 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_BATTERY (1ull << 17)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK:
+ *
+ * Use parent for the battery level and threshold.
+ *
+ * Since: 1.6.4
+ */
+#define FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK (1ull << 18)
+
 /* accessors */
 gchar *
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-device.h
```diff
@@ -378,6 +378,15 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_BATTERY (1ull << 17)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK:
+ *
+ * Use parent for the battery level and threshold.
+ *
+ * Since: 1.6.4
```
hughsie, line 386: this is because we want to backport it

superm1 (approved):

hughsie:
Lets see if @CragW can confirm that it fixes the issue.

CragW (commented):
Pls wait a second, doing more test cycles.

superm1:
I think this is a correct commit, but maybe there is a secondary problem.

Crag, I'd suggest to run `fwupdtool` with `--verbose` on each run so you have the extra logging to point at where.

CragW:
> I think this is a correct commit, but maybe there is a secondary problem.

Agree with you, this commit looks good to me. Let's go ahead and I will back here or new bug with further findings.

> Crag, I'd suggest to run `fwupdtool` with `--verbose` on each run so you have the extra logging to point at where.

Yeah, I should run verbose on each cycle...

CragW:
I've just finished a couple more cycles and everything went fine, cheers.

PR #3743, opened 2021-09-13
plugins/flashrom: Fix logic in reset cmos

Sean-StarLabs:
ioperm returns -1 for an error and 0 for success

### DIFF
--- a/plugins/flashrom/fu-flashrom-cmos.c
+++ b/plugins/flashrom/fu-flashrom-cmos.c
@@ -34,7 +34,7 @@ gboolean
 fu_flashrom_cmos_reset(GError **error)
 {
 	/* Call ioperm() to grant us access to ports 0x70 and 0x71 */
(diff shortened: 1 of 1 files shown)


plugins/flashrom/fu-flashrom-cmos.c
```diff
@@ -34,7 +34,7 @@ gboolean
 fu_flashrom_cmos_reset(GError **error)
 {
 	/* Call ioperm() to grant us access to ports 0x70 and 0x71 */
-	if (!ioperm(RTC_BASE_PORT, 2, TRUE)) {
+	if (ioperm(RTC_BASE_PORT, 2, TRUE) == -1) {
```
hughsie, line 37: maybe `< 0` is more appropriate?

Sean-StarLabs: Done

hughsie (approved):

PR #3744, opened 2021-09-13
trivial: Add translations for more device categories

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
> what happens on older versions that see these new categories

They fall back to X-Device automatically. It's how I can be fearless adding new strings :)

PR #3748, opened 2021-09-14
Wip/hardboprobot/logitech hidpp fix devids

hardboprobot:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-device.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-device.c
@@ -673,8 +673,12 @@ fu_logitech_hidpp_device_probe(FuDevice *device, GError **error)
 	FuLogitechHidPpDevice *self = FU_HIDPP_DEVICE(device);
 	FuLogitechHidPpDevicePrivate *priv = GET_PRIVATE(self);
 
-	/* FuUdevDevice->probe */
-	if (!fu_device_has_private_flag(device, FU_LOGITECH_HIDPP_DEVICE_FLAG_BOLT_PERIPHERAL)) {
+	/*
+	 * FuUdevDevice->probe except for paired devices. We don't want
+	 * paired devices to inherit the logical ids of the receiver.
+	 */
+	if (priv->device_idx == HIDPP_DEVICE_IDX_UNSET ||
+	    priv->device_idx == HIDPP_DEVICE_IDX_BLE) {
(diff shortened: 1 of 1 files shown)


plugins/logitech-hidpp/fu-logitech-hidpp-device.c
```diff
@@ -354,7 +355,14 @@ fu_logitech_hidpp_device_create_radio_child(FuLogitechHidPpDevice *self,
...
 	radio = fu_logitech_hidpp_radio_new(ctx, entity);
 	fu_device_set_physical_id(FU_DEVICE(radio), fu_device_get_physical_id(FU_DEVICE(self)));
-	fu_device_set_logical_id(FU_DEVICE(radio), priv->model_id);
+	/*
+	 * Use the parent logical id as well as the model id for the
+	 * logical id of the radio child device. This allows the radio
+	 * devices of two devices of the same type (same device type,
+	 * BLE mode) to coexist correctly.
+	 */
+	logical_id = g_strdup_printf("%s_%s", priv->model_id, fu_device_get_logical_id(FU_DEVICE(self)));
```
hughsie, line 364: I think the convention we have in other plugins is `PARENTVALUE-SOMETHINGELSE`, i.e. the other way around with a dash.

hughsie:
Also precommit seems upset. :/

plugins/logitech-hidpp/fu-logitech-hidpp-device.c
hardboprobot: Done, I fixed both commits, actually. Setting the proper logical id (HID_UNIQ) for BLE devices is trickier because the private flags aren't loaded yet when `fu_logitech_hidpp_device_probe` runs, so you can't tell whether a device is in BLE mode or not (although maybe there's a definite way to figure that out, maybe querying udev). For now, I moved the logical_id setup for BLE devices to `fu_logitech_hidpp_device_setup`

hughsie: > I moved the logical_id setup for BLE devices to fu_logitech_hidpp_device_setup

I think that works -- the device ID will change at that point, but it's before the device is added to the daemon and so I don't think that is matters. Thanks.

hughsie (approved):

PR #3750, opened 2021-09-14
update void linux ci target

dkwo:
i tried to update the ci integration for void linux.

the current packaged version is here
https://github.com/void-linux/void-packages/blob/master/srcpkgs/fwupd/template
should you need to correct something

it would be great to run a regular xbps-src build,
which would show cross compilation issues.

### DIFF
--- a/contrib/ci/dependencies.xml
+++ b/contrib/ci/dependencies.xml
@@ -1186,6 +1186,11 @@
       <package variant="x86_64" />
     </distro>
   </dependency>
+  <dependency type="build" id="gi-docgen">
+    <distro id="void">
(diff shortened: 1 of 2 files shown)


superm1 (approved):

superm1:
In the future some suggestions:
* I don't know why this is set in your package - you should probably leave the default.
`-Dplugin_altos=false`

* `xbps-src build` - Yes, please do that. it would be better to track breakage with your real packaging rather than just a script ran in your environment.  We do this with other distros and it helps us identify things like fedora's spec file needs something added, or anything that is different.

PR #3751, opened 2021-09-14
Convert the stand-alone device test utility to a fwupdmgr subcommand

hughsie:
This allows us to do a few things:

 * Remove the runtime dep on Python 3, which is tricky for ChromeOS
 * Test composite devices more efficiently, only writing once per test

In the future, with a little more work, it'll also allow us to
automatically upload signed reports for successful device tests.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@bleungatchromium this may be of interest.

src/fu-util.c
```diff
@@ -673,6 +675,222 @@ fu_util_display_current_message(FuUtilPrivate *priv)
...
+	/* download file if required */
+	basename = json_object_get_string_member(json_obj, "file");
+	url = g_build_filename("https://fwupd.org/downloads", basename, NULL);
+	filename = fu_util_download_if_required(priv, url, error);
+	if (filename == NULL)
+		return FALSE;
+
+	/* install file */
+	priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_OLDER;
+	priv->flags |= FWUPD_INSTALL_FLAG_ALLOW_REINSTALL;
```
superm1, line 771: does this really want to be set all the times?  Wouldn't it make sense to have tests that skip particular firmwares that are already installed or older too?

src/fu-util.c
```diff
@@ -673,6 +675,222 @@ fu_util_display_current_message(FuUtilPrivate *priv)
...
+{
+	JsonArray *json_array;
+	const gchar *basename;
+	g_autofree gchar *url = NULL;
+	g_autofree gchar *filename = NULL;
+	g_autoptr(GBytes) fw = NULL;
+
+	/* download file if required */
+	basename = json_object_get_string_member(json_obj, "file");
+	url = g_build_filename("https://fwupd.org/downloads", basename, NULL);
```
superm1, line 764: rather than hardcode this in fwupdmgr, can it be an argument or config file option?  I would think fetching prefix from the remote configuration makes perfect sense for example.

superm1 (commented):
do the device tests want to get installed now into the filesystem as part of fwupd-tests packages or anything like that?

src/fu-util.c
hughsie: Maybe we just put the URL in the .JSON file? It doesn't have to be the LVFS, and in some cases it's better if it's not...

superm1: Yeah, that's a better idea.

hughsie: Fixed.

hughsie:
> do the device tests want to get installed now into the filesystem

Done.

src/fu-util.c
hughsie: I tried this. It gets really hairy when you're testing and you install one of the two component firmwares (e.g. using install-blob). The combination of ALLOW_OLDER and ALLOW_REINSTALL is really a FORCE_-lite_ and basically tells the daemon JFDI which I think is entirely appropriate for device tests.

bleungatchromium:
@djcampello as well.

Thanks for doing this! I'll give this a try.

djcampello:
I just tried this on a Chromebook:

```
localhost ~ # fwupdmgr device-test /usr/share/fwupd/device-tests/hp-dock-g5.json
Downloading?             [***************************************]
Authenticating?          [***************************************]
Waiting?                 [***************************************] Less than one minute remaining?
Downloading?             [***************************************] Less than one minute remaining?
Authenticating?          [***************************************]
Waiting?                 [***************************************] Less than one minute remaining?
/usr/share/fwupd/device-tests/hp-dock-g5.json: OK!
```

It looks good. We might want some sort of flag to specify a mirror to use as a BaseURI for the url in the json files.

hughsie:
It looks like I need to do another pass at getting the progressbar to one line per action, but 22 steps forward and one step back.

    Downloading?             [***************************************]
    Authenticating?          [***************************************]

> We might want some sort of flag to specify a mirror to use as a BaseURI for the url in the json files.

Hmm, I hadn't thought of that. I guess my naive idea would be a --base-uri argument but I don't really like adding options that only operate on one sub-command for a very-specific non-upstream use-case. Would a semi-hidden `FWUPD_DEVICE_TESTS_BASE_URI` environment variable do?

djcampello:
> Would a semi-hidden `FWUPD_DEVICE_TESTS_BASE_URI` environment variable do?

I think that should work!

superm1:
I wonder if we should add something documentation to fwupd.github.io a central page for all our "hidden" environment variables what they do...

hughsie:
Or is that something we want in the man page? Maybe docs page is easier to keep up to date.

hughsie:
> documentation to fwupd.github.io

Split out as https://github.com/fwupd/fwupd/issues/3760

hughsie:
@djcampello can you test this new version? `FWUPD_DEVICE_TESTS_BASE_URI="https://foo.bar.com/"` should work.

djcampello:
It does work! :)

PR #3752, opened 2021-09-15
Add new API for splitting an untrusted string

hughsie:
Using fu_common_strnsplit() has the drawback that a malicious user (or
a fuzzer!) could create a file with 5,000,000 newlines, and then pass
that into any parser that tokenizes into lines. This causes millions of
tiny allocations and quickly dirties hundreds of megabytes of RSS due
to heap overheads.

Rather than splitting a huge array and then processing each line, set
up a callback to process each line and only allocate the next string if
the token was parsed correctly. This means that we only dup the buffer
just once before we start parsing rather than allocating everything and
then failing at the first hurdle.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38696

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1942,6 +1942,79 @@ fu_common_strnsplit(const gchar *str, gsize sz, const gchar *delimiter, gint max
 }
 
 /**
+ * fu_common_strnsplit_full:
+ * @str: a string to split
+ * @sz: size of @str, or -1 for unknown
+ * @delimiter: a string which specifies the places at which to split the string
+ * @callback: (scope call): a #FuCommonStrsplitFunc.
+ * @user_data: user data
+ * @error: (nullable): optional return location for an error
+ *
+ * Splits the string, calling the given function for each
+ * of the tokens found. If any @callback returns %FALSE scanning is aborted.
+ *
+ * Use this function in preference to fu_common_strnsplit() when the input file is untrusted,
+ * and you don't want to allocate a GStrv with billions of one byte items.
+ *
+ * Returns: %TRUE if no @callback returned FALSE
+ *
+ * Since: 1.7.0
+ */
+gboolean
+fu_common_strnsplit_full(const gchar *str,
+			 gssize sz,
+			 const gchar *delimiter,
+			 FuCommonStrsplitFunc callback,
+			 gpointer user_data,
+			 GError **error)
+{
+	gsize delimiter_sz;
+	gsize str_sz;
+	guint found_idx = 0;
+	guint token_idx = 0;
+
+	g_return_val_if_fail(str != NULL, FALSE);
+	g_return_val_if_fail(delimiter != NULL && delimiter[0] != '\0', FALSE);
+	g_return_val_if_fail(callback != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	/* make known */
+	str_sz = sz != -1 ? (gsize)sz : strlen(str);
+	delimiter_sz = strlen(delimiter);
+
+	/* cannot split */
+	if (delimiter_sz > str_sz) {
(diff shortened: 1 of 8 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -1941,6 +1941,95 @@ fu_common_strnsplit(const gchar *str, gsize sz, const gchar *delimiter, gint max
...
+
+	/* make known */
+	str_sz = sz != -1 ? (gsize)sz : strlen(str);
+	delimiter_sz = strlen(delimiter);
+	if (delimiter_sz == 0)
+		return TRUE;
+
+	/* can never match */
+	if (delimiter_sz > str_sz)
+		return TRUE;
```
superm1, line 1991: shouldn't these be failures - IE failure to split that the caller should know about?

plugins/ccgx/fu-ccgx-firmware.c
```diff
@@ -287,6 +296,60 @@ fu_ccgx_firmware_parse_md_block(FuCcgxFirmware *self, FwupdInstallFlags flags, G
...
+} FuCcgxFirmwareTokenHelper;
+
+static gboolean
+fu_ccgx_firmware_tokenize_cb(GString *token, guint token_idx, gpointer user_data, GError **error)
+{
+	FuCcgxFirmwareTokenHelper *helper = (FuCcgxFirmwareTokenHelper *)user_data;
+	FuCcgxFirmware *self = FU_CCGX_FIRMWARE(helper->self);
+
+	/* remove WIN32 line endings */
+	g_strdelimit(token->str, "\r\x1a", '\0');
```
superm1, line 311: isn't this only going to effectively remove EOF?

plugins/wacom-usb/fu-wac-firmware.c
```diff
@@ -18,244 +18,264 @@ struct _FuWacFirmware {
...
-					g_debug("header_fw%u_addr: 0x%x", j, hdr->addr);
-					g_debug("header_fw%u_sz:   0x%x", j, hdr->sz);
-					g_ptr_array_add(header_infos, g_steal_pointer(&hdr));
-				}
-				continue;
+	FuWacFirmwareTokenHelper *helper = (FuWacFirmwareTokenHelper *)user_data;
+	g_autofree gchar *cmd = NULL;
+
+	/* remove WIN32 line endings */
+	g_strdelimit(token->str, "\r\x1a", '\0');
```
superm1, line 44: same comment as other plugin, this is removing the EOF, I'd think you want to remove all `\r`?

libfwupdplugin/fu-common.c
hughsie: I think `delimiter_sz == 0` should probably be a return_val_if_fail programmer critical. I'm not sure about "delimiter_sz > str_sz" -- you can argue it both ways.

plugins/ccgx/fu-ccgx-firmware.c
hughsie: Nah, it's per-token, i.e. operating on the allocated `line`, not the file.

superm1: Then shouldn't it be looking for just carriage return not EOF delimiter?

hughsie: Isn't the last token going to include the EOF?

libfwupdplugin/fu-common.c
hughsie: Fixed both.

plugins/ccgx/fu-ccgx-firmware.c
superm1: oh right - in that case the srec plugin probably needs to match this too.

hughsie: Indeed, fixed, good catch.

PR #3763, opened 2021-09-16
Wip/hardboprobot/logitech hidpp fixes

hardboprobot:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-hidpp/fu-logitech-hidpp-runtime-bolt.c
+++ b/plugins/logitech-hidpp/fu-logitech-hidpp-runtime-bolt.c
@@ -25,7 +25,6 @@ fu_logitech_hidpp_runtime_bolt_detach(FuDevice *device, FuProgress *progress, GE
 	FuLogitechHidPpRuntime *self = FU_HIDPP_RUNTIME(device);
 	g_autoptr(FuLogitechHidPpHidppMsg) msg = fu_logitech_hidpp_msg_new();
 	g_autoptr(GError) error_local = NULL;
-	g_autoptr(FwupdRequest) request = fwupd_request_new();
 
 	msg->report_id = HIDPP_REPORT_ID_LONG;
 	msg->device_id = HIDPP_DEVICE_IDX_RECEIVER;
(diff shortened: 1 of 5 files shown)


plugins/logitech-hidpp/fu-logitech-hidpp-runtime-bolt.c
```diff
@@ -51,14 +51,6 @@ fu_logitech_hidpp_runtime_bolt_detach(FuDevice *device, FuProgress *progress, GE
...
-	if (fu_device_get_update_message(device) == NULL) {
-		g_autofree gchar *str = NULL;
-		str = g_strdup_printf("%s needs to be manually restarted to complete the update. "
-				      "Please unplug it and plug it back again.",
-				      fu_device_get_name(device));
-		fu_device_set_update_message(device, str);
-	}
 	fwupd_request_set_kind(request, FWUPD_REQUEST_KIND_IMMEDIATE);
 	fwupd_request_set_id(request, FWUPD_REQUEST_ID_REMOVE_REPLUG);
 	fwupd_request_set_message(request, fu_device_get_update_message(device));
```
hughsie, line 56: don't we need the above for this line?

hardboprobot: I didn't notice any errors whatsoever, but you're right, that line should be wiped too. I also need to fix the same thing in fu-logitech-hidpp-device.c

plugins/logitech-hidpp/fu-logitech-hidpp-runtime-bolt.c
```diff
@@ -51,17 +51,8 @@ fu_logitech_hidpp_runtime_bolt_detach(FuDevice *device, FuProgress *progress, GE
...
-	if (fu_device_get_update_message(device) == NULL) {
-		g_autofree gchar *str = NULL;
-		str = g_strdup_printf("%s needs to be manually restarted to complete the update. "
-				      "Please unplug it and plug it back again.",
-				      fu_device_get_name(device));
-		fu_device_set_update_message(device, str);
-	}
 	fwupd_request_set_kind(request, FWUPD_REQUEST_KIND_IMMEDIATE);
 	fwupd_request_set_id(request, FWUPD_REQUEST_ID_REMOVE_REPLUG);
-	fwupd_request_set_message(request, fu_device_get_update_message(device));
```
hughsie, line 64: don't we need this? I did kinda assume that we'd construct the gchar *str and then set it as the request message.

hardboprobot: Well, the initial idea was to disable this message for some specific devices, and that's defined by the "no-request-required" private flag. But I think the Bolt receiver (or at least this type of Bolt receiver) isn't supposed to ask the user for a replug ever. The other alternative would be to define that private flag in `fu_logitech_hidpp_runtime_bolt_init` as well (right now it's only defined in `fu_logitech_hidpp_device_init`) and parameterize this per device, but I'd have to check if that'll be necessary with the Logitech people.

hughsie: My bolt receivers on my desk need replugging, so we probably need a way to ask the device "do you need a replug" -- I had assumed that logitech would have different VID/PIDs for the different behaviour. Or am I completely mistaken and all "available in the shops" bolt receivers will no longer need a replug? If that's the case then we need to remove the fwupd_request_set_id() and fwupd_request_set_kind() lines too.

hardboprobot: Here's the catch: all the Bolt receivers I tested, both the "prototype" ones and the ones that came with the M650 and the M750 seem to be the same hardware, and they needed a replug up to firmware MPR05.00_B0009. Starting with firmware MPR05.01_B0010 they don't need that anymore. But let me check with them first. Definitely, the most flexible way to do this is to do it just like in fu-logitech-hidpp-device.c

hardboprobot: Ok, confirmed, Bolt receivers won't ever need user interaction

hughsie (approved):

PR #3764, opened 2021-09-16
redfish: Do not hardcode the KCS user_id to be 0x4

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/redfish/fu-plugin-redfish.c
+++ b/plugins/redfish/fu-plugin-redfish.c
@@ -205,11 +205,10 @@ fu_redfish_plugin_ipmi_create_user(FuPlugin *plugin, GError **error)
 {
 	FuPluginData *data = fu_plugin_get_data(plugin);
 	const gchar *username_fwupd = "fwupd";
-	guint8 user_id = 0x04;
+	guint8 user_id = G_MAXUINT8;
 	g_autofree gchar *password_new = fu_common_generate_password(15);
 	g_autofree gchar *password_tmp = fu_common_generate_password(15);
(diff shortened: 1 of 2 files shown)


plugins/redfish/fu-plugin-redfish.c
```diff
@@ -221,14 +220,28 @@ fu_redfish_plugin_ipmi_create_user(FuPlugin *plugin, GError **error)
...
-	/* check the slot is clear */
-	username = fu_ipmi_device_get_user_password(device, 0x04, NULL);
-	if (username != NULL) {
-		g_set_error(error,
-			    FWUPD_ERROR,
-			    FWUPD_ERROR_NOT_SUPPORTED,
-			    "cannot create fwupd user with account %s already existing",
-			    username);
+	/* check for existing user, and if not then remember the first spare slot */
+	for (guint8 i = 3; i < 63; i++) {
```
superm1, line 224: these numbers seem somewhat arbitrary and hardcoded, are they implementation specific or speced somewhere?

hughsie: That's an excellent question. Will ask the OEM.

hughsie: Okay, so 0 is reserved, and 1 is typically the factory-set admin account, so 2 is more appropriate. The field "on the wire" is one byte, so 0xFF is a more appropriate maximum. I've fixed up the commit with those changes and tested on real hardware.

superm1 (approved):

PR #3765, opened 2021-09-16
Fix probing the Dell TPM (Fixes: #3759)

superm1:
Since commit 078beaf, device de-duplication between plugins doesn't
work anymore.  Both the TPM and Dell plugin create a device with an
identical physical ID and no logical ID.  This makes the device list
think whichever plugin probes second is just updating the first.

To avoid this problem, set a logical ID in the Dell plugin, ensuring
that the device ID built from physical + logical is different from the
TPM plugin.

The side effect of this is that both devices will export into the device
list, but the TPM plugin's devices will not be upgradable.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I think this is sane, its not like the dell plugin needs to watch for ->registered() and do an inhibit on the "other" TPM as it's not updatable anyway. The "other TPM" isn't going to be in the get-devices list by default and so I don't think it's confusing to the user.

hughsie (approved):

PR #3768, opened 2021-09-17
HSI Change log implementation.

smallorange:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

This commit can be used to search for the changes from previous and current HSI security attributes. If the current HSI score is different from the previous HSI score, the change information will be made and store the changes to the DB. Once the user wants to know or the information will be displayed to the GUI application, the changes can access from the DB immediately.

1. The DB column of hsi_histroy table is modified to
   a. hsi_score is changed to integer.
   b. change and version columns are added to the table.
2. Json of SecurityAttrs is modified from array to Key/value structure to ensure the JSON object can act as a HashTable.
3. Add a change searching mechanism to search the change when HSI value is not the same as the previous one.

Changes in v1
1. JSON format is simplified to be suitable for display.
2. Code clean up
3. Add a "flag to string array" converter.

### DIFF
--- a/libfwupd/fwupd-security-attr.c
+++ b/libfwupd/fwupd-security-attr.c
@@ -866,6 +866,37 @@ fwupd_security_attr_to_json(FwupdSecurityAttr *self, JsonBuilder *builder)
 }
 
 /**
+ * fwupd_security_attr_flag_to_string_array:
+ *
+ * @self: a single FwupdSecurityAttr
+ *
+ * Convert flags to string and store them in GPtrArray.
+ *
+ * Since: 1.7.0
+ *
+ */
+
+GPtrArray *
+fwupd_security_attr_flag_to_string_array(FwupdSecurityAttr *self)
+{
+	GPtrArray *flag_array = NULL;
+	FwupdSecurityAttrPrivate *priv = GET_PRIVATE(self);
+	if (priv->flags != FWUPD_SECURITY_ATTR_FLAG_NONE) {
+		flag_array = g_ptr_array_new();
+		for (guint i = 0; i < 64; i++) {
+			if ((priv->flags & ((guint64)1 << i)) == 0)
+				continue;
+			g_ptr_array_add(
+			    flag_array,
+			    g_strdup(fwupd_security_attr_flag_to_string((guint64)1 << i)));
+		}
+	} else {
+		return NULL;
+	}
+	return flag_array;
+}
+
+/**
  * fwupd_security_attr_to_string:
  * @self: a #FwupdSecurityAttr
  *
--- a/libfwupd/fwupd-security-attr.c
+++ b/libfwupd/fwupd-security-attr.c
@@ -865,7 +865,6 @@ fwupd_security_attr_to_json(FwupdSecurityAttr *self, JsonBuilder *builder)
 	}
 }
 
-
 /**
  * fwupd_security_attr_to_string:
  * @self: a #FwupdSecurityAttr
--- a/libfwupd/fwupd-security-attr.c
+++ b/libfwupd/fwupd-security-attr.c
@@ -865,36 +865,6 @@ fwupd_security_attr_to_json(FwupdSecurityAttr *self, JsonBuilder *builder)
 	}
 }
 
-/**
- * fwupd_security_attr_flag_to_string_array:
- *
- * @self: a single FwupdSecurityAttr
- *
- * Convert flags to string and store them in GPtrArray.
- *
- * Since: 1.7.0
- *
- */
-
-GPtrArray *
-fwupd_security_attr_flag_to_string_array(FwupdSecurityAttr *self)
-{
-	GPtrArray *flag_array = NULL;
-	FwupdSecurityAttrPrivate *priv = GET_PRIVATE(self);
-	if (priv->flags != FWUPD_SECURITY_ATTR_FLAG_NONE) {
-		flag_array = g_ptr_array_new();
-		for (guint i = 0; i < 64; i++) {
-			if ((priv->flags & ((guint64)1 << i)) == 0)
-				continue;
-			g_ptr_array_add(
-			    flag_array,
-			    g_strdup(fwupd_security_attr_flag_to_string((guint64)1 << i)));
-		}
-	} else {
-		return NULL;
-	}
-	return flag_array;
-}
 
 /**
  * fwupd_security_attr_to_string:
--- a/libfwupd/fwupd-security-attr.h
+++ b/libfwupd/fwupd-security-attr.h
@@ -174,8 +174,6 @@ const gchar *
 fwupd_security_attr_flag_to_suffix(FwupdSecurityAttrFlags flag);
 const gchar *
 fwupd_security_attr_result_to_string(FwupdSecurityAttrResult result);
-GPtrArray *
-fwupd_security_attr_flag_to_string_array(FwupdSecurityAttr *self);
 FwupdSecurityAttr *
 fwupd_security_attr_from_variant(GVariant *value);
 GPtrArray *
--- a/libfwupd/fwupd-security-attr.h
+++ b/libfwupd/fwupd-security-attr.h
@@ -174,6 +174,7 @@ const gchar *
 fwupd_security_attr_flag_to_suffix(FwupdSecurityAttrFlags flag);
 const gchar *
 fwupd_security_attr_result_to_string(FwupdSecurityAttrResult result);
+
 FwupdSecurityAttr *
 fwupd_security_attr_from_variant(GVariant *value);
 GPtrArray *
--- a/libfwupd/fwupd-security-attr.h
+++ b/libfwupd/fwupd-security-attr.h
@@ -174,7 +174,8 @@ const gchar *
 fwupd_security_attr_flag_to_suffix(FwupdSecurityAttrFlags flag);
 const gchar *
 fwupd_security_attr_result_to_string(FwupdSecurityAttrResult result);
-
+GPtrArray *
+fwupd_security_attr_flag_to_string_array(FwupdSecurityAttr *self);
 FwupdSecurityAttr *
 fwupd_security_attr_from_variant(GVariant *value);
 GPtrArray *
--- a/libfwupd/fwupd.map
+++ b/libfwupd/fwupd.map
@@ -707,5 +707,6 @@ LIBFWUPD_1.7.0 {
     fwupd_security_attr_add_guids;
     fwupd_security_attr_get_guids;
     fwupd_security_attr_has_guid;
+    fwupd_security_attr_flag_to_string_array;
   local: *;
 } LIBFWUPD_1.6.2;
--- a/libfwupd/fwupd.map
+++ b/libfwupd/fwupd.map
@@ -707,6 +707,5 @@ LIBFWUPD_1.7.0 {
     fwupd_security_attr_add_guids;
     fwupd_security_attr_get_guids;
     fwupd_security_attr_has_guid;
-    fwupd_security_attr_flag_to_string_array;
   local: *;
 } LIBFWUPD_1.6.2;
--- a/libfwupdplugin/fu-security-attrs.c
+++ b/libfwupdplugin/fu-security-attrs.c
@@ -154,10 +154,9 @@ fu_security_attrs_remove_all(FuSecurityAttrs *self)
  * fu_security_attrs_calculate_hsi:
  * @self: a #FuSecurityAttrs
  * @flags: HSI attribute flags
+ * @hsi_number: (out) (nullable): the HSI number
  *
  * Calculates the HSI string from the appended attributes.
- *
- * If hsi_uint_ret is NULL, the guint type hsi will not be returned.
  *
  * Returns: (transfer full): a string or %NULL
  *
@@ -167,9 +166,9 @@ fu_security_attrs_remove_all(FuSecurityAttrs *self)
 gchar *
 fu_security_attrs_calculate_hsi(FuSecurityAttrs *self,
 				FuSecurityAttrsFlags flags,
-				guint *hsi_guint_ret)
+				guint *hsi_number)
 {
-	guint hsi_number = 0;
+	guint hsi_number_tmp = 0;
 	FwupdSecurityAttrFlags attr_flags = FWUPD_SECURITY_ATTR_FLAG_NONE;
 	GString *str = g_string_new("HSI:");
 	const FwupdSecurityAttrFlags hpi_suffixes[] = {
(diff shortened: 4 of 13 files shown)


libfwupd/fwupd.map
```diff
@@ -707,5 +707,6 @@ LIBFWUPD_1.7.0 {
     fwupd_security_attr_add_guids;
     fwupd_security_attr_get_guids;
     fwupd_security_attr_has_guid;
+    fwupd_security_attr_flag_to_string_array;
```
superm1, line 710: This is the wrong place for it (see CI failure)

src/fu-history.c
```diff
@@ -1331,12 +1383,13 @@ fu_history_add_security_attribute(FuHistory *self,
...
 	g_return_val_if_fail(locker != NULL, FALSE);
-	rc = sqlite3_prepare_v2(self->db,
-				"INSERT INTO hsi_history (hsi_details, hsi_score)"
-				"VALUES (?1, ?2)",
-				-1,
-				&stmt,
-				NULL);
+	rc = sqlite3_prepare_v2(
+	    self->db,
+	    "INSERT INTO hsi_history (hsi_details, hsi_score, fwupd_version, changes)"
```
superm1, line 1388: doesn't changing this mean a database migration?

src/fu-history.c
```diff
@@ -1346,10 +1399,58 @@ fu_history_add_security_attribute(FuHistory *self,
...
+							       sqlite3_column_text(stmt, i));
+					*ret_json_attr = g_string_free(tmp_str, FALSE);
+				}
+				break;
+			case (SQLITE_INTEGER):
+				if (i == 0)
+					*ret_hsi = sqlite3_column_int(stmt, i);
+				break;
+			default:
+				g_warning("Mismatch HSI history column format");
```
superm1, line 1444: rather than a warning here, I'd say you should plumb `error` into `fu_history_get_last_hsi_details`

libfwupd/fwupd.map
smallorange: Yes, I am trying to resolve this. I may let this function live in src/.

src/fu-history.c
smallorange: I modified the table schema and try to move the data to the new schema.

src/fu-history.c
smallorange: I'll modify this. Thank you.

superm1:
CI fails for pre-commit and style hooks.
Did you set up pre-commit and run `./contrib/setup`?

libfwupdplugin/fu-security-attrs.c
```diff
@@ -157,12 +157,17 @@ fu_security_attrs_remove_all(FuSecurityAttrs *self)
  *
  * Calculates the HSI string from the appended attributes.
  *
+ * If hsi_uint_ret is NULL, the guint type hsi will not be returned.
```
hughsie, line 160: I think you want instead (a few lines further up) `@hsi_number: (out) (nullable): the HSI number`

libfwupdplugin/fu-security-attrs.c
```diff
@@ -157,12 +157,17 @@ fu_security_attrs_remove_all(FuSecurityAttrs *self)
...
  * Since: 1.5.0
+ *
  **/
 gchar *
-fu_security_attrs_calculate_hsi(FuSecurityAttrs *self, FuSecurityAttrsFlags flags)
+fu_security_attrs_calculate_hsi(FuSecurityAttrs *self,
+				FuSecurityAttrsFlags flags,
+				guint *hsi_guint_ret)
 {
 	guint hsi_number = 0;
```
hughsie, line 172: rename this `hsi_number_tmp` and use `hsi_number` in the function header.

libfwupdplugin/fu-security-attrs.c
```diff
@@ -212,6 +217,10 @@ fu_security_attrs_calculate_hsi(FuSecurityAttrs *self, FuSecurityAttrsFlags flag
 		attr_flags |= fwupd_security_attr_get_flags(attr);
 	}
 
+	if (hsi_guint_ret != NULL) {
+		*hsi_guint_ret = hsi_number;
```
hughsie, line 221: no need for curly braces on one line

libfwupd/fwupd-security-attr.c
```diff
@@ -865,6 +865,7 @@ fwupd_security_attr_to_json(FwupdSecurityAttr *self, JsonBuilder *builder)
 	}
 }
 
+
```
hughsie, line 868: remove this pls

libfwupd/fwupd-security-attr.h
```diff
@@ -174,7 +174,6 @@ const gchar *
 fwupd_security_attr_flag_to_suffix(FwupdSecurityAttrFlags flag);
 const gchar *
 fwupd_security_attr_result_to_string(FwupdSecurityAttrResult result);
-
```
hughsie, line 177: try to avoid making changes in unrelated functions please

src/fu-engine.c
```diff
@@ -5863,7 +5863,7 @@ fu_engine_ensure_security_attrs_tainted(FuEngine *self)
 }
 
 static gchar *
-fu_engine_attrs_calculate_hsi_for_chassis(FuEngine *self)
+fu_engine_attrs_calculate_hsi_for_chassis(FuEngine *self, guint *hsi_value)
```
hughsie, line 5866: `hsi_number` perhaps?

src/fu-engine.c
```diff
@@ -5953,8 +5960,14 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 
 	/* distil into one simple string */
 	g_free(self->host_security_id);
-	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self);
+	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self, &hsi_value);
 
+	if (fu_history_get_last_hsi_details(self->history, &previous_hsi, &last_json_attr) ==
+	    TRUE) {
```
hughsie, line 5966: no need for == TRUE

src/fu-security-attr.c
```diff
@@ -301,17 +305,269 @@ fu_security_attrs_to_json(FuSecurityAttrs *attrs, JsonBuilder *builder)
...
+ * Since: 1.7.0
+ *
+ */
+static GPtrArray *
+fu_security_attr_flag_to_string_array(FwupdSecurityAttr *attr)
+{
+	GPtrArray *flag_array = NULL;
+	FwupdSecurityAttrFlags flags = fwupd_security_attr_get_flags(attr);
+	if (flags != FWUPD_SECURITY_ATTR_FLAG_NONE) {
+		flag_array = g_ptr_array_new();
```
hughsie, line 389: this needs to be `g_ptr_array_new_with_free_func(g_free)` to deallocate the stuff you're g_strdup'ing.

src/fu-security-attr.c
```diff
@@ -301,17 +305,269 @@ fu_security_attrs_to_json(FuSecurityAttrs *attrs, JsonBuilder *builder)
...
+				fwupd_security_attr_get_appstream_id(attr)),
+			    result_builder);
+		} else {
+			/* Miss- A new AppStreamID */
+			fu_security_attr_deep_object_compare(attr, NULL, result_builder);
+		}
+	}
+	removed_keys = g_hash_table_get_keys(found_list);
+	if (removed_keys != NULL) {
+		/* Removed from current */
```
hughsie, line 557: single line trivial comments like this should be lowercase please

smallorange:
Ops! There were some mistakes on git here. Let me clean up and fix my branch.

libfwupdplugin/fu-security-attrs.c
smallorange: Thanks for correcting this. I'll revise it.

libfwupdplugin/fu-security-attrs.c
smallorange: OK. I had modified all of them.

libfwupdplugin/fu-security-attrs.c
smallorange: removed.

libfwupd/fwupd-security-attr.c
smallorange: I had cleaned up the unnecessary changes.

src/fu-engine.c
smallorange: ok

src/fu-engine.c
smallorange: Got it and removed it.

src/fu-security-attr.c
smallorange: Thanks for reminding me this. I have corrected this.

src/fu-security-attr.c
smallorange: ok

src/fu-engine.c
```diff
@@ -5953,20 +5960,37 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 
 	/* distil into one simple string */
 	g_free(self->host_security_id);
-	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self);
+	self->host_security_id = fu_engine_attrs_calculate_hsi_for_chassis(self, &hsi_number);
+
+	if (fu_history_get_last_hsi_details(self->history,
+					    &previous_hsi,
+					    &last_json_attr,
+					    &error)) {
```
hughsie, line 5968: I think it's better to reverse this; it's more common to see `if (!fu_history_get_last_hsi_details` e.g. where you handle the failure case first.

src/fu-engine.c
```diff
@@ -5953,20 +5960,37 @@ fu_engine_ensure_security_attrs(FuEngine *self)
...
+		if (previous_hsi != G_MAXUINT) {
+			if (fu_security_attrs_compare_hsi_score(previous_hsi, hsi_number))
+				diff_result =
```
hughsie, line 5971: you can move the g_autofree gchar *diff_result = NULL; into this scope

hughsie: Also, we're not print the text or anything?

src/fu-history.c
```diff
@@ -334,6 +336,48 @@ fu_history_migrate_database_v6(FuHistory *self, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_history_migrate_database_v7(FuHistory *self, GError **error)
+{
+	gint rc;
+	/* rename table */
+	rc = sqlite3_exec(self->db,
+			  "ALTER TABLE hsi_history RENAME TO hsi_history_tmp",
```
hughsie, line 345: do we really need to do this to alter one column type?

src/fu-security-attr.c
```diff
@@ -301,17 +305,281 @@ fu_security_attrs_to_json(FuSecurityAttrs *attrs, JsonBuilder *builder)
...
+	    json_object_get_string_member(previous_json_obj, FWUPD_RESULT_KEY_APPSTREAM_ID));
+	json_builder_begin_object(result_builder);
+	json_builder_set_member_name(result_builder, "removed");
+	json_builder_begin_object(result_builder);
+	fu_security_attr_dup_json(previous_json_obj, result_builder);
+	json_builder_end_object(result_builder);
+	json_builder_end_object(result_builder);
+}
+
+gchar *
```
hughsie, line 517: i think some function documentation would help a lot here, i.e. what's it *doing*?

src/fu-security-attr.c
```diff
@@ -301,17 +305,281 @@ fu_security_attrs_to_json(FuSecurityAttrs *attrs, JsonBuilder *builder)
...
+				previous_security_attrs,
+				fwupd_security_attr_get_appstream_id(attr)),
+			    result_builder);
+		} else {
+			/* miss- a new AppStreamId */
+			fu_security_attr_deep_object_compare(attr, NULL, result_builder);
+		}
+	}
+	removed_keys = g_hash_table_get_keys(found_list);
+	if (removed_keys != NULL) {
```
hughsie, line 568: I don't think you need this `if`

src/fu-security-attr.c
```diff
@@ -301,17 +305,281 @@ fu_security_attrs_to_json(FuSecurityAttrs *attrs, JsonBuilder *builder)
...
+	g_autoptr(GHashTable) found_list = g_hash_table_new(g_str_hash, g_str_equal);
+	g_autoptr(GList) member_list = NULL;
+	g_autoptr(GList) removed_keys = NULL;
+	JsonNode *json_root = NULL;
+	JsonObject *json_obj = NULL;
+	JsonObject *previous_security_attrs = NULL;
+
+	if (last_hsi_detail == NULL) {
+		json_null = g_strdup("{\"SecurityAttributes\":{}}");
+		json_parser_load_from_data(parser, json_null, -1, NULL);
```
hughsie, line 536: this feels a bit like a hack; is there a better way to handle this initial case?

src/fu-engine.c
smallorange: Ok. Let me double-check the code and handle the failure case first.

src/fu-engine.c
smallorange: diff_result will be referred at line 5991 (fu_history_add_security_attribute()) so I should let it stay in current scope.
Yes, you pointed this out. I'll print some text to notice user HSI number was changed here.

src/fu-history.c
smallorange: Unfortunately, sqlite does not allow to switch data types using a ALTER command. The table should be rename and create a new table with the modified schema then copy all the records to the new table. Finally, the renamed old table will be dropped.

src/fu-security-attr.c
smallorange: I'll put some descriptions in the function document. Thanks for pointing this out.

src/fu-security-attr.c
smallorange: Ops. I'll remove this.

src/fu-security-attr.c
smallorange: I'll enhance this part. Thanks.

hughsie:
I copied the core comparison bits out into https://github.com/fwupd/fwupd/pull/3913 -- and I think we should try and do the diff in runtime until we know that it's causing a bottleneck. Rather than precomputing each diff, I just decided to not write the HSI result if has the same attrs -- which means in the common case of < 5 system changes it's instantaneous to recompute. The diff stuff you did lives on in e4b097c3dd3aaf1890bed2d141d0b0a01e32c413 and query stuff in 17b6d79728789cfc1150797eaa0076f259a3a641. It also means we get results for people that have old versions of fwupd _with_ the attributes, but `without` the changelog column -- it means we can hit the ground running with lots of existing data.

PR #3770, opened 2021-09-17
trivial: restore s390x CI

superm1:
The next release of Debian is out, hopefully s390x is workable again
so we can have big endian coverage in CI.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Nope, seems that it's still not fixed in Debian:
```
 libgcrypt20-dev:s390x : Depends: libgpg-error-dev:s390x but it is not installable
```

Is s390x (or any other big endian architecture) more accessible and in better shape in fedora?

hughsie:
> Is s390x (or any other big endian architecture) more accessible and in better shape in fedora?

Maybe; the only way I run it is by SSHing to one of the actual machines. I've never actually set up a virtualized version before.

Issue #3771, opened 2021-09-17
Can no longer use detach as a FuDeviceLockerFunc


hughsie:
@hughsie Here it is:

```
~/fwupd/build/src$ sudo G_DEBUG=fatal-criticals gdb --args ./fwupdtool install 15_ec.cab --allow-reinstall --force -v
[...]
12:21:44:0696 XbSilo               ignoring for OR statement: indexed string 'a9898afa-ef58-48fe-b09a-f476afce467c' was unfound
12:21:44:0696 XbSilo               ignoring for OR statement: indexed string '832af090-2ef9-7c47-8f6d-b405c8c7f156' was unfound
WARNING: This package has not been validated, it may not work properly.
12:21:44:0701 FuCommon             reading 15_ec.cab with 132310 bytes
12:21:44:0701 FuEngine             Emitting PropertyChanged('Status'='decompressing')
Decompressing…           [ -                                     ]12:21:44:0701 FuCabinet            failed to verify firmware.metainfo.xml: no JcatItem
12:21:44:0701 FuCabinet            processing file: firmware.metainfo.xml
12:21:44:0701 XbSilo               compiling 8e5e7c029f8633bbeaa5a31c33dba67a9f93239f:filename=firmware.metainfo.xml:prefix=components…
12:21:44:0701 FuCabinet            processing release: 1.07.02TR1
12:21:44:0701 FuEngine             Emitting PropertyChanged('Status'='idle')
Decompressing…           [***************************************]
12:21:44:0701 FuMain               first pass requirement on acb9e599f8d594d7c6cee1482b615d3611f56c69:com.tuxedo.NS5xMU.ec.firmware failed: No supported devices found
12:21:44:0701 FuMain               first pass requirement on 4bde70ba4e39b28f9eab1628f9dd6e6244c03027:com.tuxedo.NS5xMU.ec.firmware failed: No supported devices found
12:21:44:0701 FuMain               first pass requirement on 5792b48846ce271fab11c4a545f7a3df0d36e00a:com.tuxedo.NS5xMU.ec.firmware failed: No supported devices found
12:21:44:0701 FuMain               first pass requirement on e421b2fc248391f6fe3e55ddbb3c9043be068bd0:com.tuxedo.NS5xMU.ec.firmware failed: No supported devices found
12:21:44:0701 FuEngine             requirement 1.0.1 ge 1.7.0 -> org.freedesktop.fwupd passed
[...]

PR #3772, opened 2021-09-17
Restore the ABI for fu_device_detach() and provide new symbols

hughsie:
Quite a few plugins are using a FuDeviceLocker to detach then attach in
the error path, and finding them isn't easy as we explicitly cast to a
FuDeviceLockerFunc.

For sanity, just provide both symbols so we can do the right thing in
both cases. It seems like a sensible thing to allow.

Fixes https://github.com/fwupd/fwupd/issues/3771

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


Asiderr (approved):
EC update works correctly on Tuxedo15

Issue #3760, opened 2021-09-16
Add documentation to fwupd.github.io for all our "hidden" environment variables 


hughsie:
We've got a lot of hidden variables for debugging. Start documenting them!

superm1:
Should this go to dedicated document?  IE something like `https://fwupd.github.io/libfwupdplugin/debugging.html`?

PR #3773, opened 2021-09-17
Document some of the environment variables used by fwupd

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3760

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [x] Documentation

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -1,0 +1,84 @@
+---
+title: Environment Variables
+---
+
+When running fwupd reads some variables from your environment and changes some
+behavior. This might be useful for debugging, or to make fwupd run somewhere
+with a non-standard filesystem layout.
+
+## fwupdmgr and fwupdtool
+
+* `DISABLE_SSL_STRICT` disables strict SSL certificate checking, which may make
+  downloading files work when using some antisocial corporate firewalls.
+* `FWUPD_CURL_VERBOSE` shows more information when downloading files
+* `FWUPD_DEVICE_TESTS_BASE_URI` sets the base URI when downloading firmware for the device-tests
+* `FWUPD_SUPPORTED` overrides the `-Dsupported_build` meson option at runtime
+* `FWUPD_VERBOSE` is set when running `--verbose`
+* `FWUPD_XMLB_VERBOSE` can be set to show Xmlb silo regeneration
+* standard glibc variables like `LANG` are also honored for CLI tools that are translated
+* libcurl respects the session proxy, e.g. `http_proxy`, `all_proxy`, `sftp_proxy` and `no_proxy`
+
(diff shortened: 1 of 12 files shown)


docs/env.md
```diff
@@ -0,0 +1,78 @@
...
+* `FU_HID_DEVICE_VERBOSE` shows HID traffic
+* `FU_SREC_FIRMWARE_VERBOSE` shows more information about parsing
+* `FU_UDEV_DEVICE_DEBUG` shows more information about UDEV devices, including parents
+* `FU_USB_DEVICE_DEBUG` shows more information about USB devices
+* `FWUPD_DEVICE_LIST_VERBOSE` display devices being added and removed from the list
+* `FWUPD_PROBE_VERBOSE` dump the detected devices to the console, even if not supported by fwupd
+
+## Plugins
+
+Most plugins read a plugin-specific key to increase verbosity, e.g. `FWUPD_DELL_VERBOSE`.
```
superm1, line 33: I think explaining how the option for plugin verbose sets this makes sense

superm1 (commented):
missing proxy variables for HTTPS/HTTP

superm1:
I think probably should be referenced from `fwupdplugin.toml.in` and/or `fwupd.toml.in` right?

docs/env.md
hughsie: Added

hughsie:
> missing proxy variables for HTTPS/HTTP, and also LANG variable

They're not really self-consumed variables, but I guess they're useful and I've briefly documented them.

> I think probably should be referenced from fwupdplugin.toml.in and/or fwupd.toml.in right?

I'm not sure. fwupd docs are aimed at *I'm building a GUI for fwpud* and fwupdplugin is aimed at `I'm building a plugin for fwupd` and we don't really have a website aimed at end users wanting to debug things.I suppose it's leaning more on the 2nd, although a markdown document seems a safe first step.

superm1:
>I'm not sure. fwupd docs are aimed at I'm building a GUI for fwpud and fwupdplugin is aimed at I'm building a plugin for fwupd and we don't really have a website aimed at end users wanting to debug things.I suppose it's leaning more on the 2nd, although a markdown document seems a safe first step.

I was thinking the second as well.  The logic being most of the people using this will come from the plugin development side to figure out how to debug problems.  So if you put it near the tutorial (and probably also link from the tutorial) that might help.

superm1 (dismissed):
LGTM, I think discoverability needs to go up.

hughsie:
> I think discoverability needs to go up.

I found and fixed the bug which caused the titles to not show up on the index, and so included the env vars in the docs.

PR #3774, opened 2021-09-18
Add FuCfuPayload and FuCfuOffer

hughsie:
We can use these in the future elantp plugin, and I'm sure we'll need
them again in the future for an *actual* CFU plugin.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-cfu-offer.c
```diff
@@ -0,0 +1,561 @@
...
+		return FALSE;
+
+	/* version */
+	if (!fu_common_read_uint32_safe(buf, bufsz, 0x4, &tmp32, G_LITTLE_ENDIAN, error))
+		return FALSE;
+	fu_firmware_set_version_raw(firmware, tmp32);
+	if (!fu_common_read_uint32_safe(buf, bufsz, 0x8, &priv->hw_variant, G_LITTLE_ENDIAN, error))
+		return FALSE;
+
+	/* product info */
```
superm1, line 449: aren't there vendor specific bytes that go inbetween?

hughsie: It's technically *reserved*.

	typedef struct
	{
		struct
		{
			UINT8 segmentNumber;
			UINT8 reserved0 : 6;
			UINT8 forceImmediateReset : 1;
			UINT8 forceIgnoreVersion : 1;
			UINT8 componentId;
			UINT8 token;
		} componentInfo;
		UINT32 version;
		UINT32 hwVariantMask;
		struct
		{
			UINT8 protocolRevision : 4;
			UINT8 bank : 2;
			UINT8 reserved0 : 2;
			UINT8 milestone : 3;
			UINT8 reserved1 : 5;
			UINT16 productId;
		} productInfo;
	} FWUPDATE_OFFER_COMMAND;

I think if if the vendor is using the reserved area they shouldn't be, but if they are the new-plugin can superclass FuCfuOffer into something else and then call the parent->parse vfunc after/before the custom parsing.

superm1: OK sounds good.

PR #3775, opened 2021-09-19
cfu: Add an initial outline CFU plugin

hughsie:
There's no actual hardware to test this against yet, but this is how I
would lay out a plugin if there was.

We still need to work out a generic encapsulation for the offer and
payload (for each component and bank) so this can work with LVFS and
fwupd.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


plugins/cfu/README.md
```diff
@@ -0,0 +1,35 @@
+# Component Firmware update
+
+## Introduction
+
+CFU is a protocol from Microsoft to make it easy to install firmware on HID devices.
+
+This protocol is unique in that it requires has a pre-download phase before sending the firmware to
+the microcontroller. This is so the device can check if the firmware is required and compatible.
+CFU also requires devices to be able to transfer the entire new transfer mode in runtime mode.
+
```
superm1, line 10: since there are some specs for this from the web, link to them in readme?

plugins/cfu/fu-cfu-device.c
```diff
@@ -0,0 +1,91 @@
...
+		offset += 0x8;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_cfu_device_init(FuCfuDevice *self)
+{
```
superm1, line 82: all of these should have `FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE`

plugins/cfu/fu-cfu-module.c
```diff
@@ -0,0 +1,110 @@
...
+	/* bank */
+	if (!fu_common_read_uint8_safe(buf, bufsz, offset + 0x4, &tmp, error))
+		return FALSE;
+	self->bank = tmp & 0b11;
+	instance_id2 = g_strdup_printf("USB\\VID_%04X&PID_%04X&CID_%02X&BANK_%01X",
+				       fu_udev_device_get_vendor(FU_UDEV_DEVICE(parent)),
+				       fu_udev_device_get_model(FU_UDEV_DEVICE(parent)),
+				       self->component_id,
+				       self->bank);
+	fu_device_add_instance_id(FU_DEVICE(self), instance_id2);
```
superm1, line 68: This is a bit odd to me - why would each bank have it's own instance ID?
Also if more than one bank is available the device, should apply `FWUPD_DEVICE_FLAG_DUAL_IMAGE`

plugins/cfu/fu-cfu-module.c
```diff
@@ -0,0 +1,110 @@
...
+
+	/* version */
+	if (!fu_common_read_uint32_safe(buf, bufsz, offset, &version_raw, G_LITTLE_ENDIAN, error))
+		return FALSE;
+	fu_device_set_version_raw(FU_DEVICE(self), version_raw);
+	version = fu_common_version_from_uint32(version_raw,
+						fu_device_get_version_format(FU_DEVICE(self)));
+
+	/* logical ID */
```
superm1, line 77: the spec has callouts for some vendor specific field usages (https://docs.microsoft.com/en-us/windows-hardware/drivers/cfu/cfu-specification#5-cfu-protocol-packet-format).  We probably need to be thinking about how those fit into the architecture.  IE perhaps CFU should _not_ be it's own plugin, but a type of device that other plugins can inherit from or a library function, especially if they want to use some of those vendor specific usages in a specific way.

plugins/cfu/fu-cfu-module.c
```diff
@@ -0,0 +1,110 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_cfu_module_init(FuCfuModule *self)
+{
+	fu_device_add_protocol(FU_DEVICE(self), "com.microsoft.cfu");
+	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_QUAD);
```
superm1, line 89: From spec this is wrong:
* This specification does not mandate any specific format for the firmware version. See section Firmware Version for guidelines.

plugins/cfu/README.md
```diff
@@ -0,0 +1,35 @@
...
+* `USB\VID_17EF&PID_7226`
+
+## Update Behavior
+
+The device has to support runtime updates and does not have a detach-into-bootloader mode -- but
+after the install has completed the device still has to reboot into the new firmware.
+
+## Vendor ID Security
+
+The vendor ID is set from the USB vendor, in this instance set to `USB:0x17EF`
```
superm1, line 31: I don't see why CFU has to be USB

plugins/cfu/README.md
hughsie: Done.

plugins/cfu/fu-cfu-device.c
hughsie: Agreed, fixed.

plugins/cfu/fu-cfu-module.c
hughsie: > why would each bank have it's own instance ID

Assuming there's more than one bag (and that is making an assumption...) The firmware _is_ specific to the bank. e.g. like the wacom-usb payload where there are different entry addresses for each bank firmware. In some cases there's only one bank (of the bank firmware is the same somehow) so I went with adding both to be most flexible.

I think `FWUPD_DEVICE_FLAG_DUAL_IMAGE` might have to be a quirk; the only way we could award that is if we "see" both banks in the GET_VERSION command. I'm guessing we don't actually know if there is:

 * a device with only one bank and the new firmware is loaded into RAM for the CFU process
 * a device with two banks and only the running one is reported, and the new firmware is supposed to be the *other* one.

plugins/cfu/fu-cfu-module.c
hughsie: > perhaps CFU should not be it's own plugin, but a type of device that other plugins can inherit from

That's a good idea, but maybe when we have at least *one* user of the CFU code -- unlike now where we don't even have that. Perhaps CFU could be implemented as a GObject interface rather than a concrete object, but without a consumer it feels like a ton of [exported] code. This plugin was meant more as a theoretical "how would we implement CFU" rather than something that I'm desperate to merge for new hardware support. Maybe making it derivable at this stage and adding the GtkDoc markup might be a good interim solution.

plugins/cfu/fu-cfu-module.c
hughsie: Agreed, but I wanted to at least choose a sane default. Given we need a VID/PID entry in the quirk I figured we could override the `VersionFormat` there, and that's why I added the instance IDs before reading and processing the version `ui32_t`.

plugins/cfu/README.md
hughsie: Indeed, valid point -- and the same with the Instance IDs too. I guess well get HIDRAW for both thanks to the `FuUdevDevice->probe()` -- I've changed the docs and code accordingly.

PR #3780, opened 2021-09-21
Wip/hughsie/bcm57xx

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -728,7 +728,13 @@ fu_plugin_device_write_firmware(FuPlugin *self,
 		g_autofree gchar *fn = NULL;
 		g_autofree gchar *localstatedir = NULL;
 
-		fw_old = fu_device_dump_firmware(device, progress, error);
+		/* progress */
+		fu_progress_set_id(progress, G_STRLOC);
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-plugin.c
```diff
@@ -728,7 +728,13 @@ fu_plugin_device_write_firmware(FuPlugin *self,
 		g_autofree gchar *fn = NULL;
 		g_autofree gchar *localstatedir = NULL;
 
-		fw_old = fu_device_dump_firmware(device, progress, error);
+		/* progress */
+		fu_progress_set_id(progress, G_STRLOC);
+		fu_progress_add_flag(progress, FU_PROGRESS_FLAG_NO_PROFILE);
+		fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_READ, 25);
+		fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 75);
```
superm1, line 735: no erase?

hughsie: This is for the case where we back up the firmware engine-side before we do the write process. The R-E-W cylce will be overridden by the ->write_firmware().

superm1 (approved):

PR #3783, opened 2021-09-23
modem-manager: Add the logic of inhibiting modemmanager device in mbim-qdu

Jarvis-Jiang-G:
Type of pull request:

- [x] Code fix
- [x] Feature
- [ ] Documentation

https://gitlab.freedesktop.org/mobile-broadband/libmbim/-/issues/10
Due to the issue that mbim-proxy allow other transactions while a multi-fragment transaction is ongoing, which cause the 
firmware upgrade with QDU fail when modem is connected. 

In reality, it may make sense to have the modem not doing anything during the upgrade, so the inhibition logic need to be added.

Signed-off-by: Jarvis Jiang <jarvis.w.jiang@gmail.com>


aleksander0m:
I've added this to my list of TODO things; will go through it in the next weeks.

PR #3788, opened 2021-09-24
Lower the GUsb dep to allow compiling on RHEL-8

hughsie:
This is something that I keep being asked to do...

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
The mass-reformat broke all my downstream patches, but they should probably be upstream anyway...

superm1 (approved):

Issue #3356, opened 2021-06-15
fwupd changes to support USB4 retimer firmware update when no devices are attached


RajmohanMani:
New kernel changes to support USB4 retimer firmware has merged in Mika's USB4 next branch as below.
https://git.kernel.org/pub/scm/linux/kernel/git/westeri/thunderbolt.git/log/?h=next

fwupd needs to be updated to add support for USB4 retimer fw update using the latest kernel changes.

fwupd should be modified to make use of the new USB4 retimer firmware update kernle changes, when no devices are attached.

More context on how the kernel changes can be used by user space to enable USB4 retimer enumeration (from Mika):

   # echo 1 > 0-0/usb4_port1/offline
   # echo 1 > 0-0/usb4_port1/rescan
   # dd if=new_nvm.bin of=0-0:1.1/nvm_nonactiveX/nvmem
   # echo 2 > 0-0:1.1/nvm_authenticate
   # echo 0 > 0-0/usb4_port1/offline

Now the port is functional again. Whenever user logs out the upgrade app can do this:

   # echo 1 > 0-0/usb4_port1/offline
   # echo 1 > 0-0/usb4_port1/rescan
   # echo 3 > 0-0:1.1/nvm_authenticate
   # echo 0 > 0-0/usb4_port1/offline

and it runs the authentication only."

This feature tracks the work that need to be done in fwupd/fwupdtool and related UI scripts, so that USB4 retimer firmware update is functional with the upstream solution.

superm1:
>when no devices are attached

How do you determine no devices attached and this flow is needed?

>echo 1 > 0-0/usb4_port1/offline

Will this take the port offline?  Does that mean the device plugged in during this time would be non-functional?  What if a device was plugged in when calling this?

superm1:
>How do you determine no devices attached and this flow is needed?

Seem to me that it's `link attribute returns "none"`.  This sounds relatively straightforward now.  Both during enumeration and writing though the check will need to be performed since a user can start with nothing plugged in, or start with something plugged in and later change it.

RajmohanMani:
Yes Mario.
The link sysfs entry should return "none" as the status. Only then, "offline" mode can be entered for that port.
This status needs to be checked, just before trying to do a retimer firmware update on that USB4 port.

superm1:
Does kernel automatically do this and check for firmware version when driver loads? Or will it be on demand?  I'm wondering during initial query from fwupd if value is cached by kernel or not.

[...]

PR #3791, opened 2021-09-24
thunderbolt: fix host controller probing

djcampello:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -271,15 +271,22 @@ fu_thunderbolt_device_probe(FuDevice *device, GError **error)
 	if (!FU_DEVICE_CLASS(fu_thunderbolt_device_parent_class)->probe(device, error))
 		return FALSE;
 
+	self->devpath = g_strdup(fu_udev_device_get_sysfs_path(FU_UDEV_DEVICE(device)));
+
 	/* device */
 	if (g_strcmp0(tmp, "thunderbolt_device") == 0) {
+		g_autofree gchar *parent_name =
+		    fu_udev_device_get_parent_name(FU_UDEV_DEVICE(self));
+		/* determine if host controller or not */
+		if (parent_name != NULL && g_str_has_prefix(parent_name, "domain"))
+			self->device_type = FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER;
 		tmp = fu_udev_device_get_sysfs_attr(FU_UDEV_DEVICE(device), "unique_id", NULL);
 		if (tmp != NULL)
 			fu_device_set_physical_id(device, tmp);
 		/* retimer */
 	} else if (g_strcmp0(tmp, "thunderbolt_retimer") == 0) {
 		self->device_type = FU_THUNDERBOLT_DEVICE_TYPE_RETIMER;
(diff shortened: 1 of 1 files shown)


superm1:
This seems to break self tests, can you check them?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -271,15 +271,22 @@ fu_thunderbolt_device_probe(FuDevice *device, GError **error)
...
 
+	self->devpath = g_strdup(fu_udev_device_get_sysfs_path(FU_UDEV_DEVICE(device)));
+
 	/* device */
 	if (g_strcmp0(tmp, "thunderbolt_device") == 0) {
+		g_autofree gchar *parent_name =
+		    fu_udev_device_get_parent_name(FU_UDEV_DEVICE(self));
+		/* determine if host controller or not */
+		if (parent_name != NULL && g_str_has_prefix(parent_name, "domain"))
+			self->device_type = FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER;
```
superm1, line 282: doesn't this exact same code already run in `fu_thunderbolt_device_setup_controller`?

djcampello: I am moving it on this patch earlier so I can stop failing on device_setup if we cannot read the version

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -356,10 +356,9 @@ fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 	fu_device_set_name(device, tmp);
 
 	/* set vendor string */
-	tmp = fu_udev_device_get_sysfs_attr(FU_UDEV_DEVICE(self), "vendor_name", error);
-	if (tmp == NULL)
-		return FALSE;
-	fu_device_set_vendor(device, tmp);
+	tmp = fu_udev_device_get_sysfs_attr(FU_UDEV_DEVICE(self), "vendor_name", NULL);
+	if (tmp != NULL)
+		fu_device_set_vendor(device, tmp);
```
superm1, line 361: Won't this be problematic for targeting firmware if we don't have a vendor ID set?  IIRC that's the reason it was mandatory.

superm1:
> How do you manually run them?

`ninja -C build test`

plugins/thunderbolt/fu-thunderbolt-device.c
djcampello: What I'm trying to accomplish is to have a host controller show up as a device.. It does not necessarily needs to be an updatable device. This is to enable the use case in https://github.com/fwupd/fwupd/issues/3356 where we would use the host controller to enable the children retimers for probing.

cc @RajmohanMani @krankun

djcampello:
> ninja -C build test

Now I need to figure out this in the context of Chrome OS portage

djcampello:
> This seems to break self tests, can you check them?

Fixed now

plugins/thunderbolt/fu-thunderbolt-device.c
djcampello: Isn't this just a string for information purposes? The vendor id is set a few lines below: 386/390

hughsie: Yup, Vendor is just for the UI. The vendor-id is the important one.

superm1 (approved):

PR #3798, opened 2021-09-24
synaptics-cape: Add interrupt endpoint support

CNXT-Simon:
Add new option can receive HID report over Interrupt In endpoint
instead of control endpoint. This new option is enabled by default and
can be turned off by setting 'InReportViaInterrupt = off' in quirk file.

This commit also fixes a false error within software reset function.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/synaptics-cape/README.md
+++ b/plugins/synaptics-cape/README.md
@@ -32,9 +32,10 @@ device will reset when the new firmware has been written.
 
 The vendor ID is set from the USB vendor, in this instance set to `USB:0x1395`
 
-## Quirk Use
+### Plugin-specific flags
 
-This plugin uses no plugin-specific quirks.
+* use-in-report-interrupt: some devices will support IN_REPORT that allow host communicate with
+  device over interrupt instead of control endpoint, since: 1.7.0
 
 ## External Interface Access
 
--- a/plugins/synaptics-cape/fu-synaptics-cape-device.c
+++ b/plugins/synaptics-cape/fu-synaptics-cape-device.c
@@ -20,15 +20,14 @@
 #define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_RETRY_TIMEOUT	300   /* ms */
 #define FU_SYNAPTICS_CAPE_DEVICE_USB_RESET_DELAY_MS	3000  /* ms */
 
-/* define CAPE command constant values and macro */
+/* defines CAPE command constant values and macro */
 #define FU_SYNAPTICS_CAPE_DEVICE_GOLEM_REPORT_ID 1 /* HID report id */
 
 #define FU_SYNAPTICS_CAPE_CMD_MAX_DATA_LEN    13 /* number of guint32 */
 #define FU_SYNAPTICS_CAPE_CMD_WRITE_DATAL_LEN 8	 /* number of guint32 */
 #define FU_SYNAPTICS_CAPE_WORD_IN_BYTES	      4	 /* bytes */
 
-#define FU_SYNAPTICS_CAPE_CMD_APP_ID(a, b, c, d)                                                   \
-	((((a)-0x20) << 8) | (((b)-0x20) << 14) | (((c)-0x20) << 20) | (((d)-0x20) << 26))
+#define FU_SYNAPTICS_CAPE_CMD_APP_ID_CTRL 0xb32d2300u
 
 /* CAPE command return codes */
 #define FU_SYNAPTICS_CAPE_MODULE_RC_GENERIC_FAILURE	(-1025)
@@ -45,13 +44,15 @@
 
 #define FU_SYNAPTICS_CMD_GET_FLAG 0x100 /* GET flag */
 
+#define FU_SYNAPTICS_CAPE_FM3_HID_INTR_IN_EP 0x83
+
 /* CAPE message structure, Little endian */
 typedef struct __attribute__((packed)) {
 	gint16 data_len : 16; /* data length in dwords */
-	guint16 cmd_id : 15;  /* Command id */
-	guint16 reply : 1;    /* Host want a reply from device, 1 = true */
-	guint32 module_id;    /* Module id */
-	guint32 data[FU_SYNAPTICS_CAPE_CMD_MAX_DATA_LEN]; /* Command data */
+	guint16 cmd_id : 15;  /* command id */
+	guint16 reply : 1;    /* host want a reply from device, 1 = true */
+	guint32 module_id;    /* module id */
+	guint32 data[FU_SYNAPTICS_CAPE_CMD_MAX_DATA_LEN]; /* command data */
 } FuCapCmd;
 
 /* CAPE HID report structure */
@@ -60,7 +61,7 @@ typedef struct __attribute__((packed)) {
 	FuCapCmd cmd;
 } FuCapCmdHidReport;
 
-/* CAPE Commands */
+/* CAPE commands */
 typedef enum {
 	FU_SYNAPTICS_CMD_FW_UPDATE_START = 0xC8,	  /* notifies firmware update started */
 	FU_SYNAPTICS_CMD_FW_UPDATE_WRITE = 0xC9,	  /* updates firmware data */
@@ -70,15 +71,22 @@ typedef enum {
 	FU_SYNAPTICS_CMD_GET_VERSION = 0x103,		  /* gets cur firmware version */
 } FuCommand;
 
-/* CAPE Fuupd device structure */
+/* CAPE fwupd device structure */
 struct _FuSynapticsCapeDevice {
 	FuHidDevice parent_instance;
-	guint32 ActivePartition; /* active partition, either 1 or 2 */
+	guint32 active_partition; /* active partition, either 1 or 2 */
 };
+
+/**
+ * FU_SYNAPTICS_CAPE_DEVICE_FLAG_USE_IN_REPORT_INTERRUPT:
+ *
+ * gets HID REPORT via Interrupt instead of Control endpoint.
+ */
+#define FU_SYNAPTICS_CAPE_DEVICE_FLAG_USE_IN_REPORT_INTERRUPT (1 << 0)
 
 G_DEFINE_TYPE(FuSynapticsCapeDevice, fu_synaptics_cape_device, FU_TYPE_HID_DEVICE)
 
-/* Sends SET_REPORT to device */
+/* sends SET_REPORT to device */
 static gboolean
 fu_synaptics_cape_device_set_report(FuSynapticsCapeDevice *self,
 				    const FuCapCmdHidReport *data,
@@ -100,7 +108,7 @@ fu_synaptics_cape_device_set_report(FuSynapticsCapeDevice *self,
 					error);
 }
 
-/* Gets data from device via GET_REPORT */
+/* gets HID report over control ep */
 static gboolean
 fu_synaptics_cape_device_get_report(FuSynapticsCapeDevice *self,
 				    FuCapCmdHidReport *data,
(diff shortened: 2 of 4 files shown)


CNXT-Simon:
Hi @hughsie and @superm1,

For synaptics-cape plugin, I added new option can receive hid report over interrupt in ep. Could you help to review my code? 

Thanks
Simon

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -28,7 +28,8 @@
 #define FU_SYNAPTICS_CAPE_WORD_IN_BYTES	      4	 /* bytes */
 
 #define FU_SYNAPTICS_CAPE_CMD_APP_ID(a, b, c, d)                                                   \
-	((((a)-0x20) << 8) | (((b)-0x20) << 14) | (((c)-0x20) << 20) | (((d)-0x20) << 26))
+	((guint32)((((a)-0x20) << 8) | (((b)-0x20) << 14) | (((c)-0x20) << 20) |                   \
+		   (((d)-0x20) << 26)))
```
hughsie, line 32: To be honest, this confused me for the first review. Maybe just include the final value as a 0x1234 integer?

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -70,10 +73,11 @@ typedef enum {
 	FU_SYNAPTICS_CMD_GET_VERSION = 0x103,		  /* gets cur firmware version */
 } FuCommand;
 
-/* CAPE Fuupd device structure */
+/* CAPE Fwupd device structure */
 struct _FuSynapticsCapeDevice {
 	FuHidDevice parent_instance;
 	guint32 ActivePartition; /* active partition, either 1 or 2 */
+	gboolean InReportViaInterrupt; /* host can get hid report via INTERRUPT IN EP*/
```
hughsie, line 80: maybe call this in_report_via_interrupt?

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -289,7 +369,39 @@ fu_synaptics_cape_device_to_string(FuDevice *device, guint idt, GString *str)
...
+				      const gchar *value,
+				      GError **error)
+{
+	FuSynapticsCapeDevice *self = FU_SYNAPTICS_CAPE_DEVICE(device);
+	if (g_strcmp0(key, "InReportViaInterrupt") == 0) {
+		if ((g_strcmp0(value, "on") == 0) || (g_strcmp0(value, "true") == 0) ||
+		    (g_strcmp0(value, "yes") == 0) || (g_strcmp0(value, "1") == 0)) {
+			self->InReportViaInterrupt = TRUE;
+		} else if ((g_strcmp0(value, "off") == 0) || (g_strcmp0(value, "false") == 0) ||
+			   (g_strcmp0(value, "no") == 0) || (g_strcmp0(value, "0") == 0)) {
```
hughsie, line 388: If it's just one bit of information, I think registering a private flag is probably a better idea than using a quirk key-value. Then you can do something like `Flags = in-report-via-interrupt` in the quirk file.

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -418,6 +530,8 @@ fu_synaptics_cape_device_setup(FuDevice *device, GError **error)
 		return FALSE;
 	}
 
+	g_debug("InReportViaInterrupt: %s", self->InReportViaInterrupt ? "yes" : "no");
```
hughsie, line 533: not this.

plugins/synaptics-cape/README.md
```diff
@@ -34,7 +34,13 @@ The vendor ID is set from the USB vendor, in this instance set to `USB:0x1395`
...
 ## Quirk Use
 
-This plugin uses no plugin-specific quirks.
+This plugin uses the following plugin-specific quirks:
+
+### InReportViaInterrupt
+
+If firmware can reply command over interrupt IN EP.
+
+Since: 1.6.5
```
superm1, line 43: I guess we want to backport this to 1.6.x that means?  Or did you not remember to update this to 1.7.0?

plugins/synaptics-cape/fu-synaptics-cape-device.c
CNXT-Simon: Do you mean something like `#define FUSYNAPTICS_CAPE_APP_ID_CTRL  0xaabbcc00 ` ?

hughsie: Yes, maybe. I think that looks a ton less scary than that macro :)

plugins/synaptics-cape/fu-synaptics-cape-device.c
CNXT-Simon: I am okay with it, but I am a little confused. Last time, I was told to change variable 'active_partition' from snake case to camel case.

plugins/synaptics-cape/fu-synaptics-cape-device.c
CNXT-Simon: Thanks, will change it to flag instead.

plugins/synaptics-cape/fu-synaptics-cape-device.c
CNXT-Simon: okay. 
BTW, can tool dump all flag value and friendly name in log?

plugins/synaptics-cape/fu-synaptics-cape-device.c
hughsie: Yes, my fault. What I meant was the variable name should be active_partition and the printable debug _title_ should be ActivePartition. No biggie.

plugins/synaptics-cape/README.md
CNXT-Simon: The github show the latest version number is 1.6.4 released few hours ago. Therefore, I update this to 1.6.5.  Will update it to 1.7.0.

plugins/synaptics-cape/fu-synaptics-cape-device.c
hughsie: Yes, if you register a private flag then the fu_device_to_string() knows how to print it.

plugins/synaptics-cape/fu-synaptics-cape-device.c
CNXT-Simon: Okay, I got it. will update both variables to snake case.

plugins/synaptics-cape/fu-synaptics-cape-device.c
superm1: Does the macro really need as many parenthesis as it has?  Maybe just taking some off would make it more appealing too.
```
(a-0x20) << 8 | \
(b-0x20) << 14 | \
(c-0x20) << 20 | \
(d-0x20) << 6  \
```

Isn't that bad if the order of operations works as I thought it should.

hughsie: My brain is trying to work out what happens if a,b,c,d are uint8_t's, or int32_t's

superm1: Also - what happens if any of those values are less than `0x20`?  It seems like the macro might only be safe for values 0x20 or greater.  In which case maybe back to @hughsie's point it should be a function.  Maybe something like this:

```
static inline guint32 fu_synaptics_cape_app_id_ctrl (guint a, guint b, guint c, guint d)
{
       g_return_val_if_fail (a > 0x20, 0);
       g_return_val_if_fail (b > 0x20, 0);
       g_return_val_if_fail (c > 0x20, 0);
       g_return_val_if_fail (d > 0x20, 0);
       return (a-0x20) << 8 |  (b-0x20) << 14 |  (c-0x20) << 20 | (d-0x20) << 26;
}
```

plugins/synaptics-cape/README.md
superm1: OK thanks.
Yeah we should only set it to 1.6.x if this plugin will backport to 1_6_x, which I don't think it will.

plugins/synaptics-cape/fu-synaptics-cape-device.c
hughsie: But we only ever do `FU_SYNAPTICS_CAPE_CMD_APP_ID('C', 'T', 'R', 'L')` -- just hardcode it.

superm1: Your point that if they want to add a second command other than CTRL then introduce a function then?

CNXT-Simon: No, there is no other than CTRL for firmware update case.

plugins/synaptics-cape/synaptics-cape.quirk
```diff
@@ -1,77 +1,99 @@
...
+Flags = use-in-report-interrupt
 
 [USB\VID_1395&PID_0293]
 Guid = SYNAPTICS_CAPE\CX31993
+Flags = use-in-report-interrupt
 
 # USB audio codec Dongle
 [SYNAPTICS_CAPE\CX31993]
 Plugin = synaptics_cape
+InReportViaInterrupt = on
```
hughsie, line 96: ??

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -100,7 +108,7 @@ fu_synaptics_cape_device_set_report(FuSynapticsCapeDevice *self,
 					error);
 }
 
-/* Gets data from device via GET_REPORT */
+/* Gets HID report over control ep */
```
hughsie, line 111: NOT: lowercase `Gets` please

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -70,12 +71,19 @@ typedef enum {
 	FU_SYNAPTICS_CMD_GET_VERSION = 0x103,		  /* gets cur firmware version */
 } FuCommand;
 
-/* CAPE Fuupd device structure */
+/* CAPE Fwupd device structure */
```
hughsie, line 74: we normally do fwupd as all lower case

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -27,8 +27,7 @@
 #define FU_SYNAPTICS_CAPE_CMD_WRITE_DATAL_LEN 8	 /* number of guint32 */
 #define FU_SYNAPTICS_CAPE_WORD_IN_BYTES	      4	 /* bytes */
 
-#define FU_SYNAPTICS_CAPE_CMD_APP_ID(a, b, c, d)                                                   \
-	((((a)-0x20) << 8) | (((b)-0x20) << 14) | (((c)-0x20) << 20) | (((d)-0x20) << 26))
+#define FU_SYNAPTICS_CAPE_CMD_APP_ID_CTRL (0xb32d2300u)
```
hughsie, line 30: no brackets required

plugins/synaptics-cape/README.md
```diff
@@ -32,9 +32,9 @@ device will reset when the new firmware has been written.
 
 The vendor ID is set from the USB vendor, in this instance set to `USB:0x1395`
 
-## Quirk Use
+### Plugin-specific flags
 
-This plugin uses no plugin-specific quirks.
+* use-in-report-interrupt: some devices will support IN_REPORT that allow host communicate with device over interrupt instead of control endpoint, since: 1.7.0
```
hughsie, line 37: isn't this line a little long? I think you can break it and use two spaces.

hughsie (approved):

PR #3810, opened 2021-09-28
Wait and verify after firmware upgrade finish

vcdmp:
Type of pull request:

- [ x] Code fix
- [ ] Feature
- [ ] Documentation
Fixed following issues discovered during integration testing:
1. Explicitly set device to USB mode. Appliance/BYOD mode not supported
2. Retry if device is not in IDLE state
3. Added dedicated thread to receive updates from the device once image file is pushed
4. Added Wait & Verify logic to ensure device not exposed prematurely

### DIFF
--- a/plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
+++ b/plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
@@ -20,7 +20,6 @@
 /* Payload size limited to 8k for both interfaces */
 #define UPD_PACKET_HEADER_SIZE	      (2 * sizeof(guint32))
 #define SYNC_PACKET_HEADER_SIZE	      (3 * sizeof(guint32))
-#define WRITE_TIME_OUT		      100
 #define HASH_TIMEOUT		      30000
 #define MAX_DATA_SIZE		      8192 /* 8k */
 #define PAYLOAD_SIZE		      MAX_DATA_SIZE - UPD_PACKET_HEADER_SIZE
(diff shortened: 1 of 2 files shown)


plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
```diff
@@ -734,3 +745,198 @@ fu_logitech_bulkcontroller_device_class_init(FuLogitechBulkcontrollerDeviceClass
...
+
+static void
+fu_logitech_bulkcontroller_device_startlistening_sync(FuDevice *device)
+{
+	FuLogitechBulkcontrollerDevice *self = FU_LOGITECH_BULKCONTROLLER_DEVICE(device);
+
+	self->is_sync_transfer_in_progress = TRUE;
+	self->sync_thread = g_thread_new("Sync Transfer thread",
+					 &fu_logitech_bulkcontroller_device_sync_thread,
+					 device);
```
hughsie, line 930: I still don't see why you need a thread. We've got multiple `GMainLoop`s and multiple `GMainContext`s, and I don't see how it makes anything possible that you can't do before.

hughsie:
Merging these all into one commit makes it very difficult to review. Can you split out the 4 changes please? I'd still like to see the daemon manage the device replug, which means querying the device to find out if it's going to restart.

plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
vcdmp: Will explore GMainLoops and multiple GMainContexts. Any additional pointers/examples for these objects?

vcdmp: Exploring it. We have some success with GMainLoop, but still seeing some issues. Will reach out to you as needed.

PR #3813, opened 2021-09-29
plugins/{uefi-capsule,flashrom}: Dont add device if it cant be used

Sean-StarLabs:
For devices that support two plugins, dont add both if they cant be used.


hughsie:
Give us the bigger picture: what's the problem, what's the goal? Then we can review. i.e. could we just have the `(priv->bcr & BCR_WPD) == 0 && (priv->bcr & BCR_BLE) > 0) return FALSE` login in the flashrom plugin?

Sean-StarLabs:
So, all of our stuff now you see two "System Firmware" type devices, one flashrom and one capsule.

If they're on AMI, they'll constantly see an error of "BIOS Locked" so people are disabling BIOS Lock. The theory is to only show the useful one, i.e.
If your BIOS is locked, hide flashrom and show capsule
If it's unlocked, show flashrom and hide capsule

superm1:
Another way to approach this - can you query the active branch of firmware?  Then some logic can be put in place to say mark a device not updatable or fail coldplug if it's on the wrong branch.

Sean-StarLabs:
> Another way to approach this - can you query the active branch of firmware? Then some logic can be put in place to say mark a device not updatable or fail coldplug if it's on the wrong branch.

coreboot firmware's branch name is `coreboot`. AMI's is `default`, which would presumably break a few things if we checked that. But, that made me think we can use DMI.

Might still need to be tied to a quirk to account for someone not enabling BLE and still using capsules...

bcr
```diff
@@ -0,0 +1,4057 @@
+plugins/intel-spi/fu-pci-device.c:	val |= priv->bus << 16;
```
superm1, line 1: don't include this file

superm1 (commented):
I don't think this is the right approach - we shouldn't be putting hacks like this in either plugin natively.  I think what you're looking at is maybe a plugin that only runs on these machines and receives registered hooks and switches things around.

hughsie:
@Sean-StarLabs you still need this?

Sean-StarLabs:
Yes, but it's low down on my to-do list so I'll re-open it later!

PR #3814, opened 2021-09-29
plugins/flashrom: Reset CMOS aftering writing region

Sean-StarLabs:
Resetting the CMOS on branch switch will cause a lock on reboot which
requires hard power off.

### DIFF
--- a/plugins/flashrom/fu-flashrom-internal-device.c
+++ b/plugins/flashrom/fu-flashrom-internal-device.c
@@ -125,15 +125,6 @@ fu_flashrom_internal_device_write_firmware(FuDevice *device,
 	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 90);
 	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_VERIFY, 10);
 
-	/* Check if CMOS needs a reset */
-	if (fu_device_has_private_flag(device, FU_FLASHROM_DEVICE_FLAG_RESET_CMOS)) {
-		g_debug("Attempting CMOS Reset");
-		if (!fu_flashrom_cmos_reset(error)) {
-			g_prefix_error(error, "failed CMOS reset: ");
-			return FALSE;
-		}
-	}
-
 	buf = g_bytes_get_data(blob_fw, &sz);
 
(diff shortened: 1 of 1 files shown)


plugins/flashrom/fu-flashrom-internal-device.c
```diff
@@ -126,7 +126,9 @@ fu_flashrom_internal_device_write_firmware(FuDevice *device,
 	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_VERIFY, 10);
 
 	/* Check if CMOS needs a reset */
-	if (fu_device_has_private_flag(device, FU_FLASHROM_DEVICE_FLAG_RESET_CMOS)) {
+	dmi_vendor = fu_context_get_hwid_value(ctx, FU_HWIDS_KEY_BIOS_VENDOR);
+	if ((fu_device_has_private_flag(device, FU_FLASHROM_DEVICE_FLAG_RESET_CMOS)) &&
+	    (dmi_vendor == "coreboot")) {
```
hughsie, line 131: you need it to be `g_strcmp0(dmi_vendor, "coreboot") == 0`

Sean-StarLabs: Can't type fast enough to keep up with you :)

I think this way would need `&& update == coreboot` but I'm not sure if fwupd knows that? So I was going to ask, do it like this, or `!swtiching-branch`?

hughsie: I guess you could check `FWUPD_INSTALL_FLAG_ALLOW_BRANCH_SWITCH` in install_flag is FALSE.

Sean-StarLabs: Which, I think, should make it the below, but that doesn't work:

```
if ((fu_device_has_private_flag(device, FU_FLASHROM_DEVICE_FLAG_RESET_CMOS)) &&
    (flags & FWUPD_INSTALL_FLAG_ALLOW_BRANCH_SWITCH) == 0)) {
```
Could you point me in the right direction, please?

hughsie (approved):
Maybe I learned a new word? "aftering" :)

PR #3815, opened 2021-09-29
plugins/uefi-capsule: Disable ux-capsule for all StarLabs laptops

Sean-StarLabs:


hughsie:
Before we do the nuclear option, lets debug this a bit over email. R.

hughsie (approved):

PR #3816, opened 2021-09-29
Write the BMP image upside down to avoid using a negative bitmap height

hughsie:
It would seem that some EFI BMP loaders are unable to flip the image
and hang when trying to load the UX capsule.

This was a real PITA to debug.

Many thanks to Sean Rhodes <sean@starlabs.systems> for all the help.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/make-images.py
+++ b/plugins/uefi-capsule/make-images.py
@@ -74,7 +74,7 @@ def _cairo_surface_write_to_bmp(img: cairo.ImageSurface) -> bytes:
             54,  # pixel array offset
             40,  # DIB header
             img.get_width(),  # width
-            -img.get_height(),  # height (top down)
+            img.get_height(),  # height
             1,  # planes
             32,  # BPP
             0,  # no compression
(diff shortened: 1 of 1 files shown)


superm1 (approved):
I do worry how do you avoid this on a bunch of really old fwupd versions though. It needs to backport to older ones and firmware metadata will need to set a minimum version to avoid this bug.

Sean-StarLabs:
> I do worry how do you avoid this on a bunch of really old fwupd versions though. It needs to backport to older ones and firmware metadata will need to set a minimum version to avoid this bug.

IRC, we started using fwupd at version 1.2.2 and didn't have any issue, so you'd only need to account for 1.6.2 -> 1.7.0.

@hughsie I believe the technical term is **boom!**. I'll test it on all of ours tomorrow, and not via ssh to be safe!

Sean-StarLabs (approved):
It works on all of ours (including the Lite's which never worked with the UX capsule).

However, we don't see the "Installing firmware update" text you mentioned, should that come from the firmware or does fwupd generate that?

hughsie:
> It works on all of ours

Great! Thanks for all your help (and persistence!) yesterday, this was a doozy to debug.

> should that come from the firmware

Yup, fwupd just says to the firmware "hey, there's your BMP" and lets the firmware do the rest. Maybe an AMI question.

hughsie:
Pushed to 1_6_X too.

PR #3823, opened 2021-10-01
plugins/uefi-capsule: Force ux-capsule over full size bgrt

Sean-StarLabs:
If bgrt is the same size as the screen, show ux-capsule on top of bgrt.

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -203,8 +203,12 @@ fu_plugin_uefi_capsule_write_splash_data(FuPlugin *plugin,
 	header.image_type = 0;
 	header.reserved = 0;
 	header.x_offset = (screen_x / 2) - (width / 2);
-	header.y_offset =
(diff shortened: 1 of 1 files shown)


plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -203,8 +203,12 @@ fu_plugin_uefi_capsule_write_splash_data(FuPlugin *plugin,
 	header.image_type = 0;
 	header.reserved = 0;
 	header.x_offset = (screen_x / 2) - (width / 2);
-	header.y_offset =
-	    fu_uefi_bgrt_get_yoffset(data->bgrt) + fu_uefi_bgrt_get_height(data->bgrt);
+	if (screen_y == fu_uefi_bgrt_get_height(data->bgrt)) {
+		header.y_offset = (screen_y * 0.8);
```
hughsie, line 207: I think you want `header.y_offset = (gdouble) screen_y * 0.8f;`

hughsie (approved):

PR #3824, opened 2021-10-01
plugins/uefi-capsule: Enable ux-capsule for all starlabs laptops

Sean-StarLabs:
#3823 makes it work, so enable for all.


plugins/uefi-capsule/uefi-capsule.quirk
```diff
@@ -1,23 +1,3 @@
-# StarLite Mk II
-[797f8bae-0ea2-4c0f-8a30-7d10ccfacbc0]
-Flags = no-ux-capsule,use-shim-unique
```
hughsie, line 3: we can loose use-shim-unique too?

Sean-StarLabs: Yup

hughsie (approved):

PR #3836, opened 2021-10-06
trivial: Match up the translator comments with the source strings

hughsie:
Doing clang-format on the codebase broke a few of these.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
how do we avoid this from getting reformatted again?

Issue #3829, opened 2021-10-04
Bitlocker asks for recovery key after firmware updates


pgera:
**Describe the bug**
On a dual-boot system with windows, updating firmware with fwupd on linux makes bitlocker on windows ask for recovery on the subsequent boot.

**Steps to Reproduce**
Update system firmware or bios with fwupd

**Expected behavior**
Updating firmware with fwupd should not disrupt bitlocker

**fwupd version information**
Please provide the version of the daemon and client.

```shell
fwupdmgr --version

client version:	1.5.11
compile-time dependency versions
	gusb:	0.3.4

daemon version:	1.5.11
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): apt

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
fwupdmgr get-devices --show-all-devices

20U9005LUS
│
├─Thunderbolt host controller:
│     Device ID:          3980395e9227783cb5ddbb846c934c9800619da4
│     Summary:            Unmatched performance for high-speed I/O
│     Current version:    46.00
│     Vendor:             Lenovo (THUNDERBOLT:0x0109, TBT:0x0109)
│     GUIDs:              32e299b7-9585-594c-833c-63ae28b14809
│                         13a933fd-ab08-5b89-b95f-f0fbd1e736af
│                         4117c99d-0660-5db6-bbfe-65ed036debe5
│                         bb3a7757-29b3-5950-9dba-71cc18c6fb86
│     Device Flags:       • Internal device
│                         • Updatable
│                         • System requires external power source
│                         • Device stages updates
│   
├─Embedded Controller:
│     Device ID:          0b3db288d2c013fcac57525c52aa6dc2c9934157
│     Current version:    0.1.13
│     Minimum Version:    0.1.13
│     Vendor:             DMI:LENOVO
│     GUIDs:              1ebe8ad3-3d40-4178-92ce-bbb3d0906332
│                         6dd2bd94-436c-5bb1-96a1-c8d0a87e4f09
│     Device Flags:       • Internal device
│                         • Updatable
│                         • System requires external power source
│                         • Supported on remote server
│                         • Needs a reboot after installation
│                         • Device is usable for the duration of the update
│   
├─Intel Management Engine:
│     Device ID:          cda6d9cd65fff6c200fd914ff50376263efddb1e
│     Current version:    225.53.1649
│     Minimum Version:    0.0.1
│     Vendor:             DMI:LENOVO
[...]

PR #3839, opened 2021-10-07
Show the user a warning if updating may affect full-disk-encryption

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/3829

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
> I'm also worried that GUI clients won't tell you this

Ahh, this is a perfect example of where we can use the client features. If a client doesn't have support for showing that warning then we can roll it into the update description perhaps.

PR #3844, opened 2021-10-08
plugins/superio: Don't check 0x14 in ITE signature

Sean-StarLabs:


Sean-StarLabs:
Yes, that would be awesome - thank you

PR #3862, opened 2021-10-15
Disallow unsigned firmware by default

superm1:
Currently if policy kit is enabled it will be used to verify the signature
on firmware archives.  If no signature is present in the archive a PK prompt
will come up to install it.

However if a user runs `sudo fwupdmgr install foo.cab` the PK prompt never comes up and
they might not be aware they have unsigned firmware they are installing.

During development unsigned firmware can still be installed using `fwupdtool`

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
This needs https://github.com/fwupd/fwupd/pull/3864 first; and we also need to think about an escape hatch for firmware engineers.

hughsie:
I think we decided against this; pls reopen if untrue.

PR #3865, opened 2021-10-15
lenovo-dock: Parse the firmware file - revised

victor-cheng:
Type of pull request:

lenovo-dock: Parse the firmware file - revised

parses two images for different DMC names (40B1/40B0)


hughsie:
also, you want to run ./contrib/setup before pushing -- it'll make your code more beautiful.

PR #3869, opened 2021-10-15
Share the SPI quirks between plugins

hughsie:
Now two plugins are using hardcoded SPI constants for various SPI chips,
and it makes sense to have some common quirk data that can be used by
both.

Add a FuSpiChip helper object that can be used by FuDevice subclasses
to get the specific SPI commands to use for each flash ID.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
> That's a matter of style, in this case I'd use a guint64 myself.

I worked out we can use the existing FirmwareSizeMax quirk key and fu_device_get_firmware_size_max() with no extra code.

PR #3878, opened 2021-10-18
steelseries: add Stratus gamepads support

d4s:
Added support for Stratus-series gamepads and dongles.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/steelseries/README.md
+++ b/plugins/steelseries/README.md
@@ -2,8 +2,14 @@
 
 ## Introduction
 
-This plugin is used to get the correct version number on SteelSeries gaming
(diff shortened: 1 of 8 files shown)


hughsie:
Also precommit is failing :( `contrib/reformat-code.py` should fix things.

hughsie:
Use the quirk file if you think devices with different vid/pids in the future are going to be different. It sounds like the code might be best here. Also, there's a lgtm issue it spotted. Can you make the loop guint and check for greater than MAX GUINT16 chucks pls.

PR #3876, opened 2021-10-18
dell dock: Support atomic and code fix

bennyLucky:
Type of pull request:

- [ ] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [* ] Code fix
- [*] Feature
- [ ] Documentation

PR #3890, opened 2021-10-20
Check for real EC before probing hubs

superm1:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

This should conceptually help with #3876.
It's untested however.


superm1:
> I'd feel a lot more warm-and-fuzzy with some real-world testing.

Yes I don't want this merged in unless crag tests it.

superm1:
@CragW I see, while calling open called probe which added the instance ID.  I just pushed another modification to prevent/delay that.  Can you have another look?

PR #3891, opened 2021-10-20
thunderbolt retimer nvm update in NDA case WIP: thunderbolt: set ports offline on host controller

krankun:
This change offline ports in the host controller for 5 seconds to allow
for retimers to enumerate in the NDA case.

BUG=b:187506425
TEST=emerge-volteer fwupd

Change-Id: I1fde5675f27e91e18951b72f24fbec4113410807

Type of pull request:
- thunderbolt retimer nvm update in NDA case

### DIFF
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -201,7 +201,7 @@ fwupd_device_flag_to_string(FwupdDeviceFlags device_flag)
 		return "unreachable";
 	if (device_flag == FWUPD_DEVICE_FLAG_AFFECTS_FDE)
 		return "affects-fde";
-	if (device_flag == FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP)
+	if (device_flag == FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)
 		return "requires-wakeup";
 	if (device_flag == FWUPD_DEVICE_FLAG_UNKNOWN)
 		return "unknown";
@@ -315,7 +315,7 @@ fwupd_device_flag_from_string(const gchar *device_flag)
 	if (g_strcmp0(device_flag, "affects-fde") == 0)
 		return FWUPD_DEVICE_FLAG_AFFECTS_FDE;
 	if (g_strcmp0(device_flag, "requires-wakeup") == 0)
-		return FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP;
+		return FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE;
 	return FWUPD_DEVICE_FLAG_UNKNOWN;
 }
 
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -202,7 +202,7 @@ fwupd_device_flag_to_string(FwupdDeviceFlags device_flag)
 	if (device_flag == FWUPD_DEVICE_FLAG_AFFECTS_FDE)
 		return "affects-fde";
 	if (device_flag == FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)
-		return "requires-wakeup";
+		return "no-auto-remove";
 	if (device_flag == FWUPD_DEVICE_FLAG_UNKNOWN)
 		return "unknown";
 	return NULL;
@@ -314,7 +314,7 @@ fwupd_device_flag_from_string(const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_UNREACHABLE;
 	if (g_strcmp0(device_flag, "affects-fde") == 0)
 		return FWUPD_DEVICE_FLAG_AFFECTS_FDE;
-	if (g_strcmp0(device_flag, "requires-wakeup") == 0)
+	if (g_strcmp0(device_flag, "no-auto-remove") == 0)
 		return FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE;
 	return FWUPD_DEVICE_FLAG_UNKNOWN;
 }
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -201,6 +201,8 @@ fwupd_device_flag_to_string(FwupdDeviceFlags device_flag)
 		return "unreachable";
 	if (device_flag == FWUPD_DEVICE_FLAG_AFFECTS_FDE)
 		return "affects-fde";
+	if (device_flag == FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP)
+		return "requires-wakeup";
 	if (device_flag == FWUPD_DEVICE_FLAG_UNKNOWN)
 		return "unknown";
 	return NULL;
@@ -312,6 +314,8 @@ fwupd_device_flag_from_string(const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_UNREACHABLE;
 	if (g_strcmp0(device_flag, "affects-fde") == 0)
 		return FWUPD_DEVICE_FLAG_AFFECTS_FDE;
+	if (g_strcmp0(device_flag, "requires-wakeup") == 0)
+		return FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP;
 	return FWUPD_DEVICE_FLAG_UNKNOWN;
 }
 
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -472,6 +472,7 @@ typedef enum {
  */
 #define FWUPD_DEVICE_FLAG_ONLY_VERSION_UPGRADE (1llu << 43)
 /**
+<<<<<<< HEAD
  * FWUPD_DEVICE_FLAG_UNREACHABLE:
  *
  * The device is currently unreachable, perhaps because it is in a lower power state or is out of
@@ -495,12 +496,16 @@ typedef enum {
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
 /**
  *
- * FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP :
  *
  * The device requires to be woken up to be re-enumerated.
- * Since 1.7.1
+ *
+ * FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE :
+ *
+ * The device requires to be woken up to be re-enumerated.
+ * Since 1.7.2
  */
-#define FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP (1llu << 46)
+
+#define FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE (1llu << 46)
 /**
  * FWUPD_DEVICE_FLAG_UNKNOWN:
  *
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -472,7 +472,6 @@ typedef enum {
  */
 #define FWUPD_DEVICE_FLAG_ONLY_VERSION_UPGRADE (1llu << 43)
 /**
-<<<<<<< HEAD
  * FWUPD_DEVICE_FLAG_UNREACHABLE:
  *
  * The device is currently unreachable, perhaps because it is in a lower power state or is out of
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -494,17 +494,14 @@ typedef enum {
  */
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
 /**
- *
- *
- * The device requires to be woken up to be re-enumerated.
- *
- * FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE :
- *
- * The device requires to be woken up to be re-enumerated.
- * Since 1.7.2
- */
-
+*
+* FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE :
+*
+* The device is not auto removed.
+* Since 1.7.2
+*/
 #define FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE (1llu << 46)
+
 /**
  * FWUPD_DEVICE_FLAG_UNKNOWN:
  *
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -493,6 +493,9 @@ typedef enum {
  * Since: 1.7.1
  */
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
+*FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP : **The device requires to be woken up to be re -
+    enumerated.**Since 1.6.5 * /
+#define FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP (1llu << 44)
 /**
  * FWUPD_DEVICE_FLAG_UNKNOWN:
  *
@@ -502,12 +505,12 @@ typedef enum {
  * Since 0.7.3
  */
 #define FWUPD_DEVICE_FLAG_UNKNOWN G_MAXUINT64
-/**
- * FwupdDeviceFlags:
- *
- * Flags used to represent device attributes
- */
-typedef guint64 FwupdDeviceFlags;
+	/**
+	 * FwupdDeviceFlags:
+	 *
+	 * Flags used to represent device attributes
+	 */
+	typedef guint64 FwupdDeviceFlags;
 
 /**
  * FWUPD_RELEASE_FLAG_NONE:
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -493,9 +493,14 @@ typedef enum {
  * Since: 1.7.1
  */
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
-*FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP : **The device requires to be woken up to be re -
-    enumerated.**Since 1.6.5 * /
-#define FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP (1llu << 44)
+/**
+ *
+ * FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP :
+ *
+ * The device requires to be woken up to be re-enumerated.
+ * Since 1.7.1
+ */
+#define FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP (1llu << 46)
 /**
  * FWUPD_DEVICE_FLAG_UNKNOWN:
  *
@@ -505,12 +510,12 @@ typedef enum {
  * Since 0.7.3
  */
 #define FWUPD_DEVICE_FLAG_UNKNOWN G_MAXUINT64
-	/**
-	 * FwupdDeviceFlags:
-	 *
-	 * Flags used to represent device attributes
-	 */
-	typedef guint64 FwupdDeviceFlags;
+/**
+ * FwupdDeviceFlags:
+ *
+ * Flags used to represent device attributes
+ */
+typedef guint64 FwupdDeviceFlags;
 
 /**
  * FWUPD_RELEASE_FLAG_NONE:
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -67,3 +67,23 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 {
 	return fu_plugin_thunderbolt_safe_kernel(plugin, error);
 }
+
+gboolean
+fu_plugin_update_prepare (FuPlugin *plugin,
+			  FwupdInstallFlags flags,
+			  FuDevice *device,
+			  GError **error)
+{
+	  g_warning("debug from fu_plugin_update_prepare");
+	  return TRUE;
+}
+
+gboolean
+fu_plugin_update_cleanup (FuPlugin *plugin,
+			  FwupdInstallFlags flags,
+			  FuDevice *device,
+			  GError **error)
+{
+	g_warning("debug from fu_plugin_update_cleanup");
+	return TRUE;
+}
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -75,7 +75,9 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 			  GError **error)
 {
 	  g_warning("debug from fu_plugin_update_prepare");
-	  return TRUE;
+	  if (device != NULL && fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
+		  return fu_thunderbolt_device_open(device,error);
+	  }
 }
 
 gboolean
@@ -85,5 +87,8 @@ fu_plugin_update_cleanup (FuPlugin *plugin,
 			  GError **error)
 {
 	g_warning("debug from fu_plugin_update_cleanup");
+	  if (device != NULL && fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
+		  return fu_thunderbolt_device_close(device, error);
+	  }
 	return TRUE;
 }
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -74,10 +74,10 @@ fu_plugin_update_prepare (FuPlugin *plugin,
 			  FuDevice *device,
 			  GError **error)
 {
-	  g_warning("debug from fu_plugin_update_prepare");
-	  if (device != NULL && fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
+	  if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
 		  return fu_thunderbolt_device_open(device,error);
 	  }
+	  return TRUE;
 }
 
 gboolean
@@ -86,8 +86,7 @@ fu_plugin_update_cleanup (FuPlugin *plugin,
 			  FuDevice *device,
 			  GError **error)
 {
-	g_warning("debug from fu_plugin_update_cleanup");
-	  if (device != NULL && fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
+	  if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
 		  return fu_thunderbolt_device_close(device, error);
 	  }
 	return TRUE;
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -69,25 +69,30 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
 }
 
 gboolean
-fu_plugin_update_prepare (FuPlugin *plugin,
-			  FwupdInstallFlags flags,
-			  FuDevice *device,
-			  GError **error)
+fu_plugin_composite_prepare (FuPlugin *plugin,
+			     GPtrArray *devices,
+			     GError **error)
 {
-	  if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
-		  return fu_thunderbolt_device_open(device,error);
-	  }
-	  return TRUE;
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *device = g_ptr_array_index (devices, i);
+		if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
+			return fu_thunderbolt_device_open(device,error);
+		}
+	}
+	return TRUE;
 }
 
+
 gboolean
-fu_plugin_update_cleanup (FuPlugin *plugin,
-			  FwupdInstallFlags flags,
-			  FuDevice *device,
-			  GError **error)
+fu_plugin_composite_cleanup	(FuPlugin	*plugin,
+							 GPtrArray	*devices,
+							 GError		**error)
 {
-	  if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
-		  return fu_thunderbolt_device_close(device, error);
-	  }
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *device = g_ptr_array_index (devices, i);
+		if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
+			return fu_thunderbolt_device_close(device, error);
+		}
+	}
 	return TRUE;
 }
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -320,6 +320,71 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 	return (guint16)val;
 }
 
+static void
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
+{
+	g_autoptr(GError) error_local = NULL;
+
+	if (!fu_udev_device_write_sysfs(device, "usb4_port3/offline", "1", &error_local)) {
+		g_warning("Setting port offline failed: %s", error_local->message);
+	}
+	error_local = NULL;
+	if (!fu_udev_device_write_sysfs(device, "usb4_port3/rescan", "1", &error_local)) {
+		g_warning("Rescan on port failed: %s", error_local->message);
+	}
+}
+
+static void
+fu_thunderbolt_device_set_port_online(FuDevice *device)
+{
+	FuUdevDevice *udev = FU_UDEV_DEVICE(device);
+	g_autoptr(GError) error_local = NULL;
+
+	if (!fu_udev_device_write_sysfs(udev, "usb4_port3/offline", "0", &error_local)) {
+		g_warning("Setting port online failed: %s", error_local->message);
+	}
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_online_cb(FuUdevDevice *device)
+{
+	fu_thunderbolt_device_set_port_online(device);
+	return G_SOURCE_REMOVE;
+}
+
+static gboolean
+fu_thunderbolt_device_open(FuDevice *device, GError **error)
+{
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER ||
+	    fu_thunderbolt_device_get_version(self, NULL))
+		return TRUE;
+
+	GUdevDevice *udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+	g_autoptr(GUdevDevice) udev_parent = g_udev_device_get_parent(udev_device);
+	udev_parent = g_udev_device_get_parent(udev_parent);
+	g_autoptr(FuUdevDevice) parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
+	fu_thunderbolt_device_set_port_offline(parent);
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_close(FuDevice *device, GError **error)
+{
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER)
+		return TRUE;
+
+	GUdevDevice *udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+	g_autoptr(GUdevDevice) udev_parent = g_udev_device_get_parent(udev_device);
+	udev_parent = g_udev_device_get_parent(udev_parent);
+	g_autoptr(FuUdevDevice) parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
+	fu_thunderbolt_device_set_port_online(parent);
+	return TRUE;
+}
+
 static gboolean
 fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 {
@@ -421,6 +486,13 @@ fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 		fu_device_add_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	}
 
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER) {
+		fu_thunderbolt_device_set_port_offline(FU_UDEV_DEVICE(device));
+		g_timeout_add_seconds(5,
+				      G_SOURCE_FUNC(fu_thunderbolt_device_set_port_online_cb),
+				      FU_UDEV_DEVICE(device));
+	}
+
 	return TRUE;
 }
 
@@ -441,6 +513,7 @@ fu_thunderbolt_device_setup_retimer(FuDevice *device, GError **error)
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_DUAL_IMAGE);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_INTERNAL);
+	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP);
 	vid = fu_udev_device_get_vendor(FU_UDEV_DEVICE(self));
 	if (vid == 0x0) {
 		g_set_error_literal(error,
@@ -809,6 +882,8 @@ fu_thunderbolt_device_class_init(FuThunderboltDeviceClass *klass)
 	klass_device->activate = fu_thunderbolt_device_activate;
 	klass_device->to_string = fu_thunderbolt_device_to_string;
 	klass_device->setup = fu_thunderbolt_device_setup;
+	klass_device->open = fu_thunderbolt_device_open;
+	klass_device->close = fu_thunderbolt_device_close;
 	klass_device->prepare_firmware = fu_thunderbolt_device_prepare_firmware;
 	klass_device->write_firmware = fu_thunderbolt_device_write_firmware;
 	klass_device->attach = fu_thunderbolt_device_attach;
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -515,7 +515,9 @@ fu_thunderbolt_device_setup_retimer(FuDevice *device, GError **error)
 	g_autofree gchar *instance = NULL;
 
 	/* as defined in PCIe 4.0 spec */
-	fu_device_set_summary (device, "A physical layer protocol-aware, software-transparent extension device "
+	fu_device_set_summary (
+			device,
+			"A physical layer protocol-aware, software-transparent extension device "
 				        "that forms two separate electrical link segments");
 	fu_device_set_name (device, fu_thunderbolt_device_type_to_string (self));
 	fu_device_add_flag (device, FWUPD_DEVICE_FLAG_UPDATABLE);
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -354,16 +354,10 @@ fu_thunderbolt_device_set_port_online(FuDevice *device)
 {
 	FuUdevDevice *udev = FU_UDEV_DEVICE(device);
 	g_autoptr(GError) error_local = NULL;
-	g_warning("debug fu_thunderbolt_device_set_port_online");
 
-	if (!fu_udev_device_write_sysfs(udev, "usb4_port1/offline", "0", &error_local)) {
+	if (!fu_udev_device_write_sysfs(udev, "../offline", "0", &error_local)) {
 		g_warning("Setting port online failed: %s", error_local->message);
 	}
-
-	if (!fu_udev_device_write_sysfs(udev, "usb4_port3/offline", "0", &error_local)) {
-		g_warning("Setting port online failed: %s", error_local->message);
-	}
-
 }
 
 gboolean
@@ -373,7 +367,6 @@ fu_thunderbolt_device_open(FuDevice *device, GError **error)
 	GUdevDevice *udev_device = NULL;
 	g_autoptr(GUdevDevice) udev_parent = NULL;
 	g_autoptr(FuUdevDevice) parent = NULL;
-	g_warning("debug from fu_thunderbolt_device_open");
 	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER ||
 	    fu_thunderbolt_device_get_version(self, NULL))
 		return TRUE;
@@ -382,7 +375,6 @@ fu_thunderbolt_device_open(FuDevice *device, GError **error)
 	udev_parent = g_udev_device_get_parent(udev_device);
 	udev_parent = g_udev_device_get_parent(udev_parent);
 	parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
-	g_warning("set offline from fu_thunderbolt_device_open");
 	return fu_thunderbolt_device_set_port_offline(parent);
 }
 
@@ -397,10 +389,9 @@ fu_thunderbolt_device_close(FuDevice *device, GError **error)
 		return TRUE;
 
 	udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
-	udev_parent = g_udev_device_get_parent(udev_device);
-	udev_parent = g_udev_device_get_parent(udev_parent);
-	parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
-	g_warning("set online from fu_thunderbolt_device_close");
+	//udev_parent = g_udev_device_get_parent(udev_device);
+	//udev_parent = g_udev_device_get_parent(udev_parent);
+	parent = fu_udev_device_new(g_steal_pointer(&udev_device));
 	fu_thunderbolt_device_set_port_online(parent);
 	return TRUE;
 }
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -323,14 +323,14 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 static void
 fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
 {
-	g_autoptr(GError) error_local = NULL;
+	g_autoptr(GError) error_local_offline = NULL;
+	g_autoptr(GError) error_local_rescan = NULL;
 
-	if (!fu_udev_device_write_sysfs(device, "usb4_port3/offline", "1", &error_local)) {
-		g_warning("Setting port offline failed: %s", error_local->message);
+	if (!fu_udev_device_write_sysfs(device, "usb4_port1/offline", "1", &error_local_offline)) {
+		g_warning("Setting port offline failed: %s", error_local_offline->message);
 	}
-	error_local = NULL;
-	if (!fu_udev_device_write_sysfs(device, "usb4_port3/rescan", "1", &error_local)) {
-		g_warning("Rescan on port failed: %s", error_local->message);
+	if (!fu_udev_device_write_sysfs(device, "usb4_port1/rescan", "1", &error_local_rescan)) {
+		g_warning("Rescan on port failed: %s", error_local_rescan->message);
 	}
 }
 
@@ -340,7 +340,7 @@ fu_thunderbolt_device_set_port_online(FuDevice *device)
 	FuUdevDevice *udev = FU_UDEV_DEVICE(device);
 	g_autoptr(GError) error_local = NULL;
 
-	if (!fu_udev_device_write_sysfs(udev, "usb4_port3/offline", "0", &error_local)) {
+	if (!fu_udev_device_write_sysfs(udev, "usb4_port1/offline", "0", &error_local)) {
 		g_warning("Setting port online failed: %s", error_local->message);
 	}
 }
@@ -356,15 +356,18 @@ static gboolean
 fu_thunderbolt_device_open(FuDevice *device, GError **error)
 {
 	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+	GUdevDevice *udev_device = NULL;
+	g_autoptr(GUdevDevice) udev_parent;
+	g_autoptr(FuUdevDevice) parent;
 
 	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER ||
 	    fu_thunderbolt_device_get_version(self, NULL))
 		return TRUE;
 
-	GUdevDevice *udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
-	g_autoptr(GUdevDevice) udev_parent = g_udev_device_get_parent(udev_device);
+	udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+	udev_parent = g_udev_device_get_parent(udev_device);
 	udev_parent = g_udev_device_get_parent(udev_parent);
-	g_autoptr(FuUdevDevice) parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
+	parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
 	fu_thunderbolt_device_set_port_offline(parent);
 	return TRUE;
 }
@@ -373,12 +376,14 @@ static gboolean
 fu_thunderbolt_device_close(FuDevice *device, GError **error)
 {
 	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+	GUdevDevice *udev_device;
+	g_autoptr(GUdevDevice) udev_parent;
 
 	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER)
 		return TRUE;
 
-	GUdevDevice *udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
-	g_autoptr(GUdevDevice) udev_parent = g_udev_device_get_parent(udev_device);
+	udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+	udev_parent = g_udev_device_get_parent(udev_device);
 	udev_parent = g_udev_device_get_parent(udev_parent);
 	g_autoptr(FuUdevDevice) parent = fu_udev_device_new(g_steal_pointer(&udev_parent));
 	fu_thunderbolt_device_set_port_online(parent);
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -350,12 +350,16 @@ fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
 }
 
 static void
-fu_thunderbolt_device_set_port_online(FuDevice *device)
+fu_thunderbolt_device_set_port_online(FuUdevDevice *device)
 {
 	FuUdevDevice *udev = FU_UDEV_DEVICE(device);
 	g_autoptr(GError) error_local = NULL;
 
-	if (!fu_udev_device_write_sysfs(udev, "../offline", "0", &error_local)) {
+	if (!fu_udev_device_write_sysfs(udev, "usb4_port1/offline", "0", &error_local)) {
+		g_warning("Setting port online failed: %s", error_local->message);
+	}
+
+	if (!fu_udev_device_write_sysfs(udev, "usb4_port3/offline", "0", &error_local)) {
 		g_warning("Setting port online failed: %s", error_local->message);
 	}
 }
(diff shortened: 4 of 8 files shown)


krankun:
> > > you might compare whether the daemon or tool launched from upstart vs from command line cause a difference. I believe upstart serializes all the events which might be causing some unexpected behaviors?
> > 
> > 
> > am launching daemon with below command as suggested by Daniel earlier : minijail0 --uts -l -p -N -v -P /mnt/empty -b / -b /proc -t -r -b /dev,,1 -b /sys,,1 -k /var,/var,tmpfs -b /var/cache/fwupd,,1 -b /var/lib/fwupd,,1 -k run,/run,tmpfs -b /run/dbus -b /run/lock,,1 -b /run/shill -b /run/udev/data,,1 -u fwupd -g fwupd -G -c cap_sys_rawio+e -- /usr/libexec/fwupd/fwupd -v
> > how do I launch daemon from upstart ? Thank you!
> 
> you'll have to talk to daniel about it, and to suggest if that would actually matter to do it from upstart or not.

latest status with the patch: 

localhost ~ # fwupdtool install /home/root/retimer-2.21.cab --allow-reinstall --plugins=thunderbolt --force
No supported devices found

at this point I can see the retimer in sysfs , that implies usb4_port/offline and usb4_port/rescan attributes have been written with fwupd 
cat /sys/bus/thunderbolt/devices/0-0\:1.1/nvm_version
19.0

I have root caused this issue to be coming from 
https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/third_party/fwupd/src/fu-install-task.c;l=235 
"if (!matches_guid) "

hughsie:
> I have root caused this issue

Root caused what? That looks entirely correct.

krankun:
> > I have root caused this issue
> 
> Root caused what? That looks entirely correct.

fwupdtool install /home/root/retimer-2.21.cab --allow-reinstall --plugins=thunderbolt --force 
the above should update nvm to 2_21 currently that doesn't happen, resulting in a "Not supported error"

I see following GUID's for retimer
      GUIDs:              f00257c8-0e09-5a32-bbb3-f49748ae6cc5 ? THUNDERBOLT\VEN_8087&DEV_15EE&REV_00
                          e6f2e4ac-f76a-5fc4-b36a-6f9cb548689e ? THUNDERBOLT\VEN_8087&DEV_15EE
                          6d71ac25-b368-5e7b-ac62-32df365fa640 ? TBT-808715ee-retimer0-0:1.1

currently code debug at the time of debug prints : e6f2e4ac-f76a-5fc4-b36a-6f9cb548689e , resulting in a non matching guid , am currently checking how to fix this guid for retimer so that I can continuw with update process from fwupdtool.

krankun:
> > > I have root caused this issue
> > 
> > 
> > Root caused what? That looks entirely correct.
> 
> fwupdtool install /home/root/retimer-2.21.cab --allow-reinstall --plugins=thunderbolt --force the above should update nvm to 2_21 currently that doesn't happen, resulting in a "Not supported error"
> 
> I see following GUID's for retimer GUIDs: f00257c8-0e09-5a32-bbb3-f49748ae6cc5 ? THUNDERBOLT\VEN_8087&DEV_15EE&REV_00 e6f2e4ac-f76a-5fc4-b36a-6f9cb548689e ? THUNDERBOLT\VEN_8087&DEV_15EE 6d71ac25-b368-5e7b-ac62-32df365fa640 ? TBT-808715ee-retimer0-0:1.1
> 
> currently code debug at the time of debug prints : e6f2e4ac-f76a-5fc4-b36a-6f9cb548689e , resulting in a non matching guid , am currently checking how to fix this guid for retimer so that I can continuw with update process from fwupdtool.

to be specific the code fails here https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/third_party/fwupd/libfwupd/fwupd-device.c;l=552

guild_tmp and guid don't match up...

guid_tmp : e72e778e-94f7-5ed2-b560-1c1262ee217c
guid : e6f2e4ac-f76a-5fc4-b36a-6f9cb548689e

superm1:
My suggestion is to fixup the things that we suggested, force push and share the output of `fwupdtool get-devices --verbose`.    That will show you the devices as they enumerate and what GUIDs they match against.  If that is all working, then share the output of `fwupdtool install foo.cab --verbose`.

krankun:
> My suggestion is to fixup the things that we suggested, force push and share the output of `fwupdtool get-devices --verbose`. That will show you the devices as they enumerate and what GUIDs they match against. If that is all working, then share the output of `fwupdtool install foo.cab --verbose`.
[fwupd_install.txt](https://github.com/fwupd/fwupd/files/7471342/fwupd_install.txt)
[get-devices.txt](https://github.com/fwupd/fwupd/files/7471343/get-devices.txt)

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,76 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
 }
 
+static void
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
+{
+	g_autoptr(GError) error_local_offline = NULL;
+	g_autoptr(GError) error_local_rescan = NULL;
+
+	if (!fu_udev_device_write_sysfs(device, "usb4_port1/offline", "1", &error_local_offline)) {
+		g_warning("Setting port offline failed: %s", error_local_offline->message);
```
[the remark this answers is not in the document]
krankun: fixed this.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -421,6 +491,13 @@ fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 		fu_device_add_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	}
 
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER) {
+		fu_thunderbolt_device_set_port_offline(FU_UDEV_DEVICE(device));
+		g_timeout_add_seconds(5,
```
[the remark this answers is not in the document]
krankun: fixed it

superm1:
> get-devices.txt

I only see one emit from thunderbolt plugin.  The retimer was never detected this means.

superm1:
So I'm wondering if maybe this is caused by the fact that the events are supposed to come later as a result of that writing the file.  So the "coldplug" doesn't show them, but if the daemon was running they would show up.
This can be proved by using the daemon instead.

In one tab run the daemon (IOW `/usr/libexec/fwupd/fwupd--verbose`)
In another tab run `fwupdmgr get-devices`

If both devices show up that is the root cause.

krankun:
> So I'm wondering if maybe this is caused by the fact that the events are supposed to come later as a result of that writing the file. So the "coldplug" doesn't show them, but if the daemon was running they would show up. This can be proved by using the daemon instead.
> 
> In one tab run the daemon (IOW `/usr/libexec/fwupd/fwupd--verbose`) In another tab run `fwupdmgr get-devices`
> 
> If both devices show up that is the root cause.
[for_mario_debug_output.txt](https://github.com/fwupd/fwupd/files/7487278/for_mario_debug_output.txt)
attached the debug out put , Yes I see two devices USB4 host controller and retimer

superm1:
> 15:38:58:0676 FuMain               Daemon ready for requests (locale (NULL))
15:38:58:0679 FuPlugin             fu_plugin_device_created(thunderbolt)
15:38:58:0679 FuIdle               inhibiting: thunderbolt requires device wakeup
15:38:58:0679 FuIdle               inhibiting: thunderbolt requires device wakeup
15:38:58:0679 FuIdle               inhibiting: thunderbolt requires device wakeup
15:38:58:0679 FuPluginThunderbolt  Kranthi fu_thunderbolt_device_probe begin 
15:38:58:0680 FuPluginThunderbolt  Kranthi fu_thunderbolt_device_probe thunderbolt_retimer
15:38:58:0680 FuDevice             using 8f92f06a1ff9413dbdd56b2337683e333ab56574 for 0-0:3.1
15:38:58:0680 FuPluginThunderbolt  fu_thunderbolt_device_open begin
15:38:58:0680 FuPluginThunderbolt  fu_thunderbolt_device_open NOT FU_THUNDERBOLT_DEVICE_TYPE_RETIMER
15:38:58:0680 FuPluginThunderbolt  retimer is set up
15:38:58:0681 FuPlugin             emit added from thunderbolt: 8f92f06a1ff9413dbdd56b2337683e333ab56574

Yup, basically the daemon finishes coldplug and then the events come through.  So that explains why `fwupdtool` doesn't work.  This approach will only work with `fwupdmgr`.  It's because you do it with `g_timeout_add_seconds`

Looking at the code change - Don't you have the calls to `fu_thunderbolt_device_set_port_offline` and `fu_thunderbolt_device_set_port_online` backwards in open/close?

krankun:
> > 15:38:58:0676 FuMain               Daemon ready for requests (locale (NULL))
> > 15:38:58:0679 FuPlugin             fu_plugin_device_created(thunderbolt)
> > 15:38:58:0679 FuIdle               inhibiting: thunderbolt requires device wakeup
> > 15:38:58:0679 FuIdle               inhibiting: thunderbolt requires device wakeup
> > 15:38:58:0679 FuIdle               inhibiting: thunderbolt requires device wakeup
> > 15:38:58:0679 FuPluginThunderbolt  Kranthi fu_thunderbolt_device_probe begin
> > 15:38:58:0680 FuPluginThunderbolt  Kranthi fu_thunderbolt_device_probe thunderbolt_retimer
> > 15:38:58:0680 FuDevice             using 8f92f06a1ff9413dbdd56b2337683e333ab56574 for 0-0:3.1
> > 15:38:58:0680 FuPluginThunderbolt  fu_thunderbolt_device_open begin
> > 15:38:58:0680 FuPluginThunderbolt  fu_thunderbolt_device_open NOT FU_THUNDERBOLT_DEVICE_TYPE_RETIMER
> > 15:38:58:0680 FuPluginThunderbolt  retimer is set up
> > 15:38:58:0681 FuPlugin             emit added from thunderbolt: 8f92f06a1ff9413dbdd56b2337683e333ab56574
> 
> Yup, basically the daemon finishes coldplug and then the events come through. So that explains why `fwupdtool` doesn't work. This approach will only work with `fwupdmgr`. It's because you do it with `g_timeout_add_seconds`
> 
> Looking at the code change - Don't you have the calls to `fu_thunderbolt_device_set_port_offline` and `fu_thunderbolt_device_set_port_online` backwards in open/close?

the place where the ports offline is enabled from is here: https://chromium-review.googlesource.com/c/chromiumos/third_party/fwupd/+/3172002/7/plugins/thunderbolt/fu-thunderbolt-device.c#504 { the same code change as this PR)

in fu_thunderbolt_device_open we have the ports offline set but that code is not executed instead I return from **below**:
	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER ||
	    fu_thunderbolt_device_get_version(self, NULL))
	{
		**return TRUE;**
	}

superm1:
But when you call install on them don't they need to un-offline on open?

krankun:
> But when you call install on them don't they need to un-offline on open?

the problem is when we open we want to set the device type to retimer so that we by pass https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L366 and set the ports offline 

currently that wont bypass and we exit the open function without setting the ports offline in open func

krankun:
> But when you call install on them don't they need to un-offline on open?

the flow is we enable offline , so that retimer enumerate , install nvm and disable offline so that retimer de-enumerates

superm1:
> the flow is we enable offline , so that retimer enumerate , install nvm and disable offline so that retimer de-enumerates

Maybe this needs to be done in update_prepare and update_cleanup then?

krankun:
> > the flow is we enable offline , so that retimer enumerate , install nvm and disable offline so that retimer de-enumerates
> 
> Maybe this needs to be done in update_prepare and update_cleanup then?

am not very familiar with the fwupd code yet , let me look at this , if needed check with Daniel and come back on this.

superm1:
> the problem is when we open we want to set the device type to retimer so that we by pass https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L366 and set the ports offline

I think the term offline and online means the opposite of my inclination is why I though they were backwards.  They seem correct.

>am not very familiar with the fwupd code yet , let me look at this , if needed check with Daniel and come back on this.

The names are `fu_plugin_prepare` and `fu_plugin_cleanup`.  basically implement a prepare method that checks for that flag you set on the device and whether the device node is there or not, and if it's flag is there but no device then call the function to do the wakeup process.  On close do the opposite.

krankun:
> > the problem is when we open we want to set the device type to retimer so that we by pass https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L366 and set the ports offline
> 
> I think the term offline and online means the opposite of my inclination is why I though they were backwards. They seem correct.
> 
> > am not very familiar with the fwupd code yet , let me look at this , if needed check with Daniel and come back on this.
> 
> The names are `fu_plugin_prepare` and `fu_plugin_cleanup`. basically implement a prepare method that checks for that flag you set on the device and whether the device node is there or not, and if it's flag is there but no device then call the function to do the wakeup process. On close do the opposite.

you mean fu_plugin_update_prepare and fu_plugin_update_cleanup ?

superm1:
> > > the problem is when we open we want to set the device type to retimer so that we by pass https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L366 and set the ports offline
> > 
> > 
> > I think the term offline and online means the opposite of my inclination is why I though they were backwards. They seem correct.
> > > am not very familiar with the fwupd code yet , let me look at this , if needed check with Daniel and come back on this.
> > 
> > 
> > The names are `fu_plugin_prepare` and `fu_plugin_cleanup`. basically implement a prepare method that checks for that flag you set on the device and whether the device node is there or not, and if it's flag is there but no device then call the function to do the wakeup process. On close do the opposite.
> 
> you mean fu_plugin_update_prepare and fu_plugin_update_cleanup ?

In less than 1.7.0 yes, it got renamed.
https://github.com/fwupd/fwupd/commit/a5a978b3f6050d9a013290cd5109cfeb103524ed

krankun:
> > the flow is we enable offline , so that retimer enumerate , install nvm and disable offline so that retimer de-enumerates
> 
> Maybe this needs to be done in update_prepare and update_cleanup then?
Hi am not able to get the debug print from https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-plugin-thunderbolt.c#L77  , fu_plugin_update_prepare is something I added now to check the flow , thanks!

superm1:
> > > the flow is we enable offline , so that retimer enumerate , install nvm and disable offline so that retimer de-enumerates
> > 
> > 
> > Maybe this needs to be done in update_prepare and update_cleanup then?
> > Hi am not able to get the debug print from https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-plugin-thunderbolt.c#L77  , fu_plugin_update_prepare is something I added now to check the flow , thanks!

right because this is based on main.  you should be using the symbols in main.
`fu_plugin_prepare` and `fu_plugin_cleanup`

krankun:
> > > > the flow is we enable offline , so that retimer enumerate , install nvm and disable offline so that retimer de-enumerates
> > > 
> > > 
> > > Maybe this needs to be done in update_prepare and update_cleanup then?
> > > Hi am not able to get the debug print from https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-plugin-thunderbolt.c#L77  , fu_plugin_update_prepare is something I added now to check the flow , thanks!
> 
> right because this is based on main. you should be using the symbols in main. `fu_plugin_prepare` and `fu_plugin_cleanup`

same cannot get debug prints if I rename to fu_plugin_prepare and fu_plugin_cleanup , these funcs should be executed if we run fwupdtool --install <.cab> ?

superm1:
> same cannot get debug prints if I rename to fu_plugin_prepare and fu_plugin_cleanup , these funcs should be executed if we run fwupdtool --install <.cab> ?

No - remember we hit that issue above because of how you're waiting 5 seconds for the second device to show up.  If you start fwupd daemon and use `fwupdmgr install <.cab>` it will probably work.

krankun:
> > same cannot get debug prints if I rename to fu_plugin_prepare and fu_plugin_cleanup , these funcs should be executed if we run fwupdtool --install <.cab> ?
> 
> No - remember we hit that issue above because of how you're waiting 5 seconds for the second device to show up. If you start fwupd daemon and use `fwupdmgr install <.cab>` it will probably work.

will check it , we have one more issue, 
[  472.370763] udevd[5061]: Process '/bin/chown -R fwupd:fwupd /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' failed with exit code 1.

path /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' should be changed to /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/0-0:3.1/nvm_non_active0'

> > same cannot get debug prints if I rename to fu_plugin_prepare and fu_plugin_cleanup , these funcs should be executed if we run fwupdtool --install <.cab> ?
> 
> No - remember we hit that issue above because of how you're waiting 5 seconds for the second device to show up. If you start fwupd daemon and use `fwupdmgr install <.cab>` it will probably work.



> > the problem is when we open we want to set the device type to retimer so that we by pass https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L366 and set the ports offline
> 
> I think the term offline and online means the opposite of my inclination is why I though they were backwards. They seem correct.
> 
> > am not very familiar with the fwupd code yet , let me look at this , if needed check with Daniel and come back on this.
> 
> The names are `fu_plugin_prepare` and `fu_plugin_cleanup`. basically implement a prepare method that checks for that flag you set on the device and whether the device node is there or not, and if it's flag is there but no device then call the function to do the wakeup process. On close do the opposite.

now developed to a stage in fu_plugin_prepare where device has FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP but device node is not present , /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/usb4_port3/offline: No such file or directory 

could you please point me to a func which does wake up process ?

superm1:
> [ 472.370763] udevd[5061]: Process '/bin/chown -R fwupd:fwupd /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' failed with exit code 1.

> path /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' should be changed to /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/0-0:3.1/nvm_non_active0'

fwupd doesn't call chown.  This must be coming from somewhere else.

>now developed to a stage in fu_plugin_prepare where device has FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP but device node is not present , /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/usb4_port3/offline: No such file or directory

Right - so in prepare you need to call your method to cause the wakeup.  You can check if the node is present (meaning something plugged in) and if it's not then call wakeup.

krankun:
> > [ 472.370763] udevd[5061]: Process '/bin/chown -R fwupd:fwupd /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' failed with exit code 1.
> 
> > path /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' should be changed to /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/0-0:3.1/nvm_non_active0'
> 
> fwupd doesn't call chown. This must be coming from somewhere else.
> 
> > now developed to a stage in fu_plugin_prepare where device has FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP but device node is not present , /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/usb4_port3/offline: No such file or directory
> 
> Right - so in prepare you need to call your method to cause the wakeup. You can check if the node is present (meaning something plugged in) and if it's not then call wakeup.

"call your method" am sorry i did not follow you mean fu_thunderbolt_device_open ?

superm1:
> > > [ 472.370763] udevd[5061]: Process '/bin/chown -R fwupd:fwupd /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' failed with exit code 1.
> > 
> > 
> > > path /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' should be changed to /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/0-0:3.1/nvm_non_active0'
> > 
> > 
> > fwupd doesn't call chown. This must be coming from somewhere else.
> > > now developed to a stage in fu_plugin_prepare where device has FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP but device node is not present , /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/usb4_port3/offline: No such file or directory
> > 
> > 
> > Right - so in prepare you need to call your method to cause the wakeup. You can check if the node is present (meaning something plugged in) and if it's not then call wakeup.
> 
> "call your method" am sorry i did not follow you mean fu_thunderbolt_device_open ?

I was thinking `fu_thunderbolt_device_set_port_offline` specifically because if you open it then calling close will close it again.  You ideally want it to be in a full ready to write state when you get into the write vfunc.

krankun:
> > > > [ 472.370763] udevd[5061]: Process '/bin/chown -R fwupd:fwupd /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' failed with exit code 1.
> > > 
> > > 
> > > > path /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_non_active0' should be changed to /sys//devices/pci0000:00/0000:00:0d.2/domain0/0-0/0-0:3.1/nvm_non_active0'
> > > 
> > > 
> > > fwupd doesn't call chown. This must be coming from somewhere else.
> > > > now developed to a stage in fu_plugin_prepare where device has FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP but device node is not present , /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/usb4_port3/offline: No such file or directory
> > > 
> > > 
> > > Right - so in prepare you need to call your method to cause the wakeup. You can check if the node is present (meaning something plugged in) and if it's not then call wakeup.
> > 
> > 
> > "call your method" am sorry i did not follow you mean fu_thunderbolt_device_open ?
> 
> I was thinking `fu_thunderbolt_device_set_port_offline` specifically because if you open it then calling close will close it again. You ideally want it to be in a full ready to write state when you get into the write vfunc.

I managed to write some code in fu_plugin_update_prepare  func to write offline and rescan values, paths seem correct now, however we are hitting below issue:

20:49:42:0541 FuPluginThunderbolt  Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error
libminijail[14272]: child process 14273 exited with status 254

But from command line doing the same was successful , do we need some permissions for the fwupdmgr to write values to sysfs ?

localhost ~ # cat /sys/devices/pci0000\:00/0000\:00\:0d.2/domain0/0-0/usb4_port3/offline 
0
localhost ~ # echo 1 > /sys/devices/pci0000\:00/0000\:00\:0d.2/domain0/0-0/usb4_port3/offline 
localhost ~ # echo 1 > /sys/devices/pci0000\:00/0000\:00\:0d.2/domain0/0-0/usb4_port3/rescan  
localhost ~ # ls -l /sys/bus/thunderbolt/devices/
0-0/     0-0:3.1/ domain0/

superm1:
I would suggest to check with Daniel about your minijail command. Maybe it needs expansion.

krankun:
> I would suggest to check with Daniel about your minijail command. Maybe it needs expansion.

the one am using is what I got from @djcampello  
minijail0     --uts -l -p -N     -v -P /mnt/empty -b / -b /proc -t -r -b /dev,,1 -b /sys,,1     -k /var,/var,tmpfs -b /var/cache/fwupd,,1 -b /var/lib/fwupd,,1     -k run,/run,tmpfs -b /run/dbus -b /run/lock,,1 -b /run/shill     -b /run/udev/data,,1 -u fwupd -g fwupd -G -c cap_sys_rawio+e     -- /usr/libexec/fwupd/fwupd -v

superm1:
Can you run your same commands by hand in the minijail and see they work?

krankun:
> Can you run your same commands by hand in the minijail and see they work?

I will check with Daniel how to run these commands with mijail env, 

after a reboot of DUT am also facing
21:44:27:0657 FuMain               first pass requirement on 8f92f06a1ff9413dbdd56b2337683e333ab56574:com.retimer.firmware.test failed: Device USB4 Retimer [8f92f06a1ff9413dbdd56b2337683e333ab56574] does not currently allow updates: Cannot install update when not on AC power

we had same issue with fwupdtool which was work arounded by  providing  --plugins=thunderbolt but that WA doesnt work with fwupdmgr

superm1:
> I will check with Daniel how to run these commands with mijail env,

I would think you can just replace the path to the fwupd binary with the path to a shell interpreter and then run them, but I might be wrong.

> 21:44:27:0657 FuMain first pass requirement on 8f92f06a1ff9413dbdd56b2337683e333ab56574:com.retimer.firmware.test failed: Device USB4 Retimer [8f92f06a1ff9413dbdd56b2337683e333ab56574] does not currently allow updates: Cannot install update when not on AC power

There is a config file option to ignore this:
https://github.com/fwupd/fwupd/blob/main/data/daemon.conf#L47

libfwupd/fwupd-enums.h
```diff
@@ -493,6 +493,14 @@ typedef enum {
  * Since: 1.7.1
  */
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
+/**
+ *
+ * FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP :
+ *
+ * The device requires to be woken up to be re-enumerated.
+ * Since 1.7.1
```
superm1, line 501: Needs to bump to 1.7.2 since 1.7.1 is out now

krankun:
> > I will check with Daniel how to run these commands with mijail env,
> 
> I would think you can just replace the path to the fwupd binary with the path to a shell interpreter and then run them, but I might be wrong.
> 
> > 21:44:27:0657 FuMain first pass requirement on 8f92f06a1ff9413dbdd56b2337683e333ab56574:com.retimer.firmware.test failed: Device USB4 Retimer [8f92f06a1ff9413dbdd56b2337683e333ab56574] does not currently allow updates: Cannot install update when not on AC power
> 
> There is a config file option to ignore this: https://github.com/fwupd/fwupd/blob/main/data/daemon.conf#L47

i made a ugly hack in my local for now to by pass will check this too

krankun:
I developed further some good signs , nvm update was sucessful but device restarts or some needs re-enumerate flag is set which doesnt clean up device properly after update 

localhost ~ # fwupdmgr install /home/root/retimer-2.21.cab --allow-reinstall --force
Authenticating?          [ -                                     ]
Installing on USB4 Retimer?-                                     ]
Restarting device?       [***************************************]
failed to get device after update: failed to wait for detach replug: device 8f92f06a1ff9413dbdd56b2337683e333ab56574 did not come back

keeping the demon up and running the fwupdmgr the nvm update is sucessful with out any errors, also clean up func is called too

localhost ~ # fwupdmgr install /home/root/retimer-2.21.cab --allow-reinstall --force
Authenticating?          [ -                                     ]
Installing on USB4 Retimer? -                                    ]
Restarting device?       [***************************************]
[for_mario_debug_fwupdmgr_update_sucess_todo_cleanup.txt](https://github.com/fwupd/fwupd/files/7515244/for_mario_debug_fwupdmgr_update_sucess_todo_cleanup.txt)


attached full log for sequence of steps , if we can avoid restarting the device and enable proper clean up path I see some path forward to get this fully functional

libfwupd/fwupd-enums.h
```diff
@@ -472,6 +472,7 @@ typedef enum {
  */
 #define FWUPD_DEVICE_FLAG_ONLY_VERSION_UPGRADE (1llu << 43)
 /**
+<<<<<<< HEAD
```
krankun, line 475: will remove

src/fu-engine.c
```diff
@@ -291,7 +291,6 @@ fu_engine_ensure_device_battery_inhibit(FuEngine *self, FuDevice *device)
 				  "battery-system",
 				  "Cannot install update when not on AC power");
 		return;
-	}
```
krankun: restored in next commit

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -512,16 +515,14 @@ fu_thunderbolt_device_setup_retimer(FuDevice *device, GError **error)
 	g_autofree gchar *instance = NULL;
 
 	/* as defined in PCIe 4.0 spec */
-	fu_device_set_summary(
-	    device,
```
krankun, line 437: restored in next commit

krankun (commented):
> fwupd doesn't call chown. This must be coming from somewhere else.
> 
> >
this looked like some udev event which calls the open/close continuously and setting the ports offline/online too quick and fast causing the port to be in suspended state for ever and resulting in the I/O error , update_prepare and up_cleanup path seemed better to bring ports offline/online and I have fixed this issue in latest commit https://github.com/fwupd/fwupd/pull/3891/commits/5651d48ccec1df0db8e3cf5f6cad8f7cb002913d

krankun:
> > fwupd doesn't call chown. This must be coming from somewhere else.
> > >
> 
> this looked like some udev event which calls the open/close continuously and setting the ports offline/online too quick and fast causing the port to be in suspended state for ever and resulting in the I/O error , update_prepare and up_cleanup path seemed better to bring ports offline/online and I have fixed this issue in latest commit [5651d48](https://github.com/fwupd/fwupd/commit/5651d48ccec1df0db8e3cf5f6cad8f7cb002913d)

cc: @djcampello

djcampello:
> > > fwupd doesn't call chown. This must be coming from somewhere else.
> > > >
> > 
> > 
> > this looked like some udev event which calls the open/close continuously and setting the ports offline/online too quick and fast causing the port to be in suspended state for ever and resulting in the I/O error , update_prepare and up_cleanup path seemed better to bring ports offline/online and I have fixed this issue in latest commit [5651d48](https://github.com/fwupd/fwupd/commit/5651d48ccec1df0db8e3cf5f6cad8f7cb002913d)
> 
> cc: @djcampello

I pointed this udev rule back in September to you on crrev.com/c/3188635 Inside the minijail the daemon runs a the fwupd user and this guarantees that the interfaces are accessible by that user.

superm1:
> crrev.com/c/3188635

is it possible that the daemon races with it?

djcampello:
Certainly possible

libfwupd/fwupd-enums.c
```diff
@@ -201,6 +201,8 @@ fwupd_device_flag_to_string(FwupdDeviceFlags device_flag)
 		return "unreachable";
 	if (device_flag == FWUPD_DEVICE_FLAG_AFFECTS_FDE)
 		return "affects-fde";
+	if (device_flag == FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)
+		return "requires-wakeup";
```
superm1, line 205: need to update this string too

libfwupd/fwupd-enums.c
```diff
@@ -312,6 +314,8 @@ fwupd_device_flag_from_string(const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_UNREACHABLE;
 	if (g_strcmp0(device_flag, "affects-fde") == 0)
 		return FWUPD_DEVICE_FLAG_AFFECTS_FDE;
+	if (g_strcmp0(device_flag, "requires-wakeup") == 0)
```
superm1, line 317: need to update this string too

libfwupd/fwupd-enums.h
```diff
@@ -493,6 +493,18 @@ typedef enum {
  * Since: 1.7.1
  */
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
+/**
+ *
+ *
+ * The device requires to be woken up to be re-enumerated.
```
superm1, line 499: don't include comment before the declaration

libfwupd/fwupd-enums.h
```diff
@@ -493,6 +493,18 @@ typedef enum {
...
+ *
+ * FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE :
+ *
+ * The device requires to be woken up to be re-enumerated.
```
superm1, line 503: the description is wrong now

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -67,3 +67,28 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
...
 	return fu_plugin_thunderbolt_safe_kernel(plugin, error);
 }
+
+gboolean
+fu_plugin_update_prepare (FuPlugin *plugin,
+			  FwupdInstallFlags flags,
+			  FuDevice *device,
+			  GError **error)
+{
+	  g_warning("debug from fu_plugin_update_prepare");
```
superm1, line 77: remove this debugging statement

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -67,3 +67,28 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
...
+	  if (device != NULL && fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
```
superm1, line 78: device will never be null coming into this function

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -67,3 +67,28 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
...
+	  }
+}
+
+gboolean
+fu_plugin_update_cleanup (FuPlugin *plugin,
+			  FwupdInstallFlags flags,
+			  FuDevice *device,
+			  GError **error)
+{
+	g_warning("debug from fu_plugin_update_cleanup");
```
superm1, line 89: remove this statement

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -67,3 +67,28 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
...
+	  if (device != NULL && fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
```
superm1, line 90: device will never be null coming into this function

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -421,6 +495,14 @@ fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 		fu_device_add_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	}
 
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER) {
+		if(fu_thunderbolt_device_set_port_offline(FU_UDEV_DEVICE(device)))
+			/*g_timeout_add_seconds(5,
+				      G_SOURCE_FUNC(fu_thunderbolt_device_set_port_online_cb),
+				      FU_UDEV_DEVICE(device));*/
```
superm1, line 502: if not using the callback remove the dead comments

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,80 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+		g_warning("Setting port online failed: %s", error_local->message);
+	}
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_online_cb(FuUdevDevice *device)
+{
+	fu_thunderbolt_device_set_port_online(device);
+	return G_SOURCE_REMOVE;
+}
```
superm1, line 356: I guess you don't need this anymore

src/fu-util-common.c
```diff
@@ -1216,6 +1216,13 @@ fu_util_device_flag_to_string(guint64 device_flag)
 	if (device_flag == FWUPD_DEVICE_FLAG_AFFECTS_FDE) {
 		/* TRANSLATORS: we might ask the user the recovery key when next booting Windows */
 		return _("Full disk encryption secrets may be invalidated when updating");
+		if (device_flag == FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP) {
+			/* TRANSLATORS: device needs to be re-enumerated to operate upon */
+			return _("Device needs to be re-enumerated");
+		}
+	if (device_flag == FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE) {
+		/* TRANSLATORS: device needs to be re-enumerated to operate upon */
+		return _("Device needs to be re-enumerated");
```
superm1, line 1225: whitespace is wrong, you should reformat using `./contrib/reformat-code.py`

src/fu-util-common.c
```diff
@@ -1216,6 +1216,13 @@ fu_util_device_flag_to_string(guint64 device_flag)
 	if (device_flag == FWUPD_DEVICE_FLAG_AFFECTS_FDE) {
 		/* TRANSLATORS: we might ask the user the recovery key when next booting Windows */
 		return _("Full disk encryption secrets may be invalidated when updating");
+		if (device_flag == FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP) {
```
superm1, line 1219: this tag should go away

superm1:
> Certainly possible

any way to avoid that?

djcampello:
> > Certainly possible
> 
> any way to avoid that?

Not that I know of. My understanding is that udev rules are applied following the priority given by the file numbering (via their filename)... Now I am not familiar how does that work with client applications like fwupd that interact with udev presumably via libudev (or whatever library is used). Since the relevant attributes in sysfs are only present when the device is enumerated I do not know of any other way to do the chgrp/chown in a timely manner.

superm1:
> > > Certainly possible
> > 
> > 
> > any way to avoid that?
> 
> Not that I know of. My understanding is that udev rules are applied following the priority given by the file numbering (via their filename)... Now I am not familiar how does that work with client applications like fwupd that interact with udev presumably via libudev (or whatever library is used). Since the relevant attributes in sysfs are only present when the device is enumerated I do not know of any other way to do the chgrp/chown in a timely manner.

Is it viable to have the kernel make the nodes at the default owner/group so no need to chwon/chgrp?  This seems like it should be a problem outside of retimer too.  Like if you connect a USB4 dock you'll possibly be racing to make it's nodes.

krankun:
> I pointed this udev rule back in September to you on crrev.com/c/3188635 Inside the minijail the daemon runs a the fwupd user and this guarantees that the interfaces are accessible by that user.

@djcampello  @superm1  I have always picked up the crrev.com/c/3188635 while doing development from the beginning and also now , somehow the path did not work and there was a continuous loop resulting in i/o error's , anyway I will clean up the patch but still need to fix the clean up path which am working on currently.
localhost ~ # fwupdmgr install /home/root/retimer-2.21.cab --allow-reinstall --force
Authenticating?          [ -                                     ]
Installing on USB4 Retimer? -                                    ]
**Restarting device?       [***************************************]
failed to get device after update: failed to wait for detach replug: device 8f92f06a1ff9413dbdd56b2337683e333ab56574 did not come back**

superm1:
> > I pointed this udev rule back in September to you on crrev.com/c/3188635 Inside the minijail the daemon runs a the fwupd user and this guarantees that the interfaces are accessible by that user.
> 
> @djcampello @superm1 I have always picked up the crrev.com/c/3188635 while doing development from the beginning and also now , somehow the path did not work and there was a continuous loop resulting in i/o error's , anyway I will clean up the patch but still need to fix the clean up path which am working on currently. localhost ~ # fwupdmgr install /home/root/retimer-2.21.cab --allow-reinstall --force Decompressing? [**_************************************] Authenticating? [ - ] Installing on USB4 Retimer? - ] **Restarting device? [**************************************_] failed to get device after update: failed to wait for detach replug: device 8f92f06a1ff9413dbdd56b2337683e333ab56574 did not come back**

Try preventing `FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG` from getting applied to your device.

krankun:
> > > I pointed this udev rule back in September to you on crrev.com/c/3188635 Inside the minijail the daemon runs a the fwupd user and this guarantees that the interfaces are accessible by that user.
> > 
> > 
> > @djcampello @superm1 I have always picked up the crrev.com/c/3188635 while doing development from the beginning and also now , somehow the path did not work and there was a continuous loop resulting in i/o error's , anyway I will clean up the patch but still need to fix the clean up path which am working on currently. localhost ~ # fwupdmgr install /home/root/retimer-2.21.cab --allow-reinstall --force Decompressing? [**_************************************] Authenticating? [ - ] Installing on USB4 Retimer? - ] **Restarting device? [**************************************_] failed to get device after update: failed to wait for detach replug: device 8f92f06a1ff9413dbdd56b2337683e333ab56574 did not come back**
> 
> Try preventing `FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG` from getting applied to your device.

fixed it by avoiding the flag for retimer device

superm1:
So what happens afterwards then? Everything good now?

If so please rebase on latest main and fix CI problems.

krankun:
> So what happens afterwards then? Everything good now?
> 
> If so please rebase on latest main and fix CI problems.

there is one last thing I want to fix , the nvm update is sucessful but if we update the nvm again for same device { keeping the demon still up) am facing 

04:22:46:0828 FuPluginThunderbolt  Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error

for the other case where I want to update nvm for all on board retimers at the same time {usb4_port1 and usb4_port3 in this case) only the 1st port retimer nvm update is sucessfull the other port setting offline fails with I/o error , 

from command line this doesnt happen so I assume something is not clean up properly ever after taking port to online and device node disappeared as expected after the 1st port retimer update 

[for_mario_2nd_attempt_fail.txt](https://github.com/fwupd/fwupd/files/7517757/for_mario_2nd_attempt_fail.txt)

krankun:
> > So what happens afterwards then? Everything good now?
> > If so please rebase on latest main and fix CI problems.
> 
> there is one last thing I want to fix , the nvm update is sucessful but if we update the nvm again for same device { keeping the demon still up) am facing
> 
> 04:22:46:0828 FuPluginThunderbolt Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error
> 
> for the other case where I want to update nvm for all on board retimers at the same time {usb4_port1 and usb4_port3 in this case) only the 1st port retimer nvm update is sucessfull the other port setting offline fails with I/o error ,
> 
> from command line this doesnt happen so I assume something is not clean up properly ever after taking port to online and device node disappeared as expected after the 1st port retimer update
> 
> [for_mario_2nd_attempt_fail.txt](https://github.com/fwupd/fwupd/files/7517757/for_mario_2nd_attempt_fail.txt)

krankun:
> > > So what happens afterwards then? Everything good now?
> > > If so please rebase on latest main and fix CI problems.
> > 
> > 
> > there is one last thing I want to fix , the nvm update is sucessful but if we update the nvm again for same device { keeping the demon still up) am facing
> > 04:22:46:0828 FuPluginThunderbolt Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error
> > for the other case where I want to update nvm for all on board retimers at the same time {usb4_port1 and usb4_port3 in this case) only the 1st port retimer nvm update is sucessfull the other port setting offline fails with I/o error ,
> > from command line this doesnt happen so I assume something is not clean up properly ever after taking port to online and device node disappeared as expected after the 1st port retimer update
> > [for_mario_2nd_attempt_fail.txt](https://github.com/fwupd/fwupd/files/7517757/for_mario_2nd_attempt_fail.txt)

superm1:
> 04:22:46:0828 FuPluginThunderbolt Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error

Is it timing?  IOW it's happening too fast by fwupd?  Does the node disappear and come back?

>From command line this doesnt happen so I assume something is not clean up properly ever after taking port to online and device node disappeared as expected after the 1st port retimer update

Maybe from command line if you did it in a tight script same thing can happen?

krankun:
> > 04:22:46:0828 FuPluginThunderbolt Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error
> 
> Is it timing? IOW it's happening too fast by fwupd? Does the node disappear and come back?
> 
> > From command line this doesnt happen so I assume something is not clean up properly ever after taking port to online and device node disappeared as expected after the 1st port retimer update
> 
> Maybe from command line if you did it in a tight script same thing can happen?

What I see is :
when we call fu_plugin_update_prepare  am setting both usb4_port1/offline and usb4_port3/offline , this will enumerate two retimers for each port 
localhost ~ # ls -l /sys/bus/thunderbolt/devices/
0-0/     0-0:1.1/ 0-0:3.1/ domain0/ 

the actual nvm update is happening there after for one of the port , lets say for usb4_port1

then we close both ports and immediately for other port the  fu_plugin_update_prepare is called too quick after the fu_plugin_update_cleanup is invoked resulting in an I/O error , 

how do we properly control like to set offline the port ,  nvm update , online the port and do same for other port 

alternatively , offline both ports , update nvm for both ports and online both ports again { this might not possible as I see fu_plugin_update_prepare is called for each retimer device }

superm1:
> > > 04:22:46:0828 FuPluginThunderbolt Setting port offline failed: could not write to /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/offline: Input/output error
> > 
> > Is it timing? IOW it's happening too fast by fwupd? Does the node disappear and come back?
> > 
> > > From command line this doesnt happen so I assume something is not clean up properly ever after taking port to online and device node disappeared as expected after the 1st port retimer update
> > 
> > Maybe from command line if you did it in a tight script same thing can happen?
> 
> What I see is :
> when we call fu_plugin_update_prepare  am setting both usb4_port1/offline and usb4_port3/offline , this will enumerate two retimers for each port 
> localhost ~ # ls -l /sys/bus/thunderbolt/devices/
> 0-0/     0-0:1.1/ 0-0:3.1/ domain0/ 
> 
> the actual nvm update is happening there after for one of the port , lets say for usb4_port1
> 
> then we close both ports and immediately for other port the  fu_plugin_update_prepare is called too quick after the fu_plugin_update_cleanup is invoked resulting in an I/O error , 
> 
> how do we properly control like to set offline the port ,  nvm update , online the port and do same for other port 
> 
> alternatively , offline both ports , update nvm for both ports and online both ports again { this might not possible as I see fu_plugin_update_prepare is called for each retimer device }
> 

We need some sort of relationship between these devices it sounds like. @hughsie thoughts?

superm1:
> when we call fu_plugin_update_prepare am setting both usb4_port1/offline and usb4_port3/offline , this will enumerate two retimers for each port

Why are you calling it for both if they're handled by separate payloads/devices?  If they're handled by separate payloads you should only set the one that you offlined.

krankun:
> > when we call fu_plugin_update_prepare am setting both usb4_port1/offline and usb4_port3/offline , this will enumerate two retimers for each port
> 
> Why are you calling it for both if they're handled by separate payloads/devices? If they're handled by separate payloads you should only set the one that you offlined.

we wont know before hand how many usb4_port-x are available under host router , my idea is to grep for all available and set the ports offline update retimer and bring the ports online.

krankun:
> > when we call fu_plugin_update_prepare am setting both usb4_port1/offline and usb4_port3/offline , this will enumerate two retimers for each port
> 
> Why are you calling it for both if they're handled by separate payloads/devices? If they're handled by separate payloads you should only set the one that you offlined.

[multiport_offline_online.txt](https://github.com/fwupd/fwupd/files/7524170/multiport_offline_online.txt)

krankun:
with the latest change https://github.com/fwupd/fwupd/pull/3891/commits/c068e56be6b73d32da78774fb4370235487bbba9 
Am bringing back the port online for each retimer device as opposed to bringing all ports online at the same time.

since open/close - online/offline methods are invoked per each device example in this case host router, retimer for usbport1 and retimer for usb4port3 closing all ports and opening same port after few milli seconds causes ec to go suspend state for ever.

the new flow:
the demon opens all ports{takes offline} under host router , this will expose retimers under sysfs
the clients identify the nvm update for each retimer {fwupdmgr}
the clients can call offline again for each device no harm here.
after nvm update for each retimer device call close method { online} only for that particular retimer device and not parent .

I think since am closing the retimer device under host router am able to update nvm but at the end I get "Error reading from file: Resource temporarily unavailable"
 
fwupdmgr install /tmp/retimer-2.21.cab --allow-reinstall --force
Authenticating?          [ -                                     ]
Installing on USB4 Retimer? -                                    ]
Writing?                 [***************************************]
Installing on USB4 Retimer?**************************************]
Error reading from file: Resource temporarily unavailable

krankun:
> with the latest change [c068e56](https://github.com/fwupd/fwupd/commit/c068e56be6b73d32da78774fb4370235487bbba9) Am bringing back the port online for each retimer device as opposed to bringing all ports online at the same time.
> 
> since open/close - online/offline methods are invoked per each device example in this case host router, retimer for usbport1 and retimer for usb4port3 closing all ports and opening same port after few milli seconds causes ec to go suspend state for ever.
> 
> the new flow: the demon opens all ports{takes offline} under host router , this will expose retimers under sysfs the clients identify the nvm update for each retimer {fwupdmgr} the clients can call offline again for each device no harm here. after nvm update for each retimer device call close method { online} only for that particular retimer device and not parent .
> 
> I think since am closing the retimer device under host router am able to update nvm but at the end I get "Error reading from file: Resource temporarily unavailable"
> 
> fwupdmgr install /tmp/retimer-2.21.cab --allow-reinstall --force Decompressing? [_**************************************] Authenticating? [ - ] Installing on USB4 Retimer? - ] Writing? [**************************************_] Installing on USB4 Retimer?**************************************] Decompressing? [***************************************] Error reading from file: Resource temporarily unavailable

"Error reading from file: Resource temporarily unavailable" is seen if I open two ports , if I open just one port the operation is successful.

superm1:
> "Error reading from file: Resource temporarily unavailable" is seen if I open two ports , if I open just one port the operation is successful.

So both resources need to stay awake until the end?  Is this kernel bug?   I guess as a workaround you might be able to push it out to a composite update and do the close for both devices together in the composite update then instead.

krankun:
> > "Error reading from file: Resource temporarily unavailable" is seen if I open two ports , if I open just one port the operation is successful.
> 
> So both resources need to stay awake until the end? Is this kernel bug? I guess as a workaround you might be able to push it out to a composite update and do the close for both devices together in the composite update then instead.
[skip_online_resource_unavailable.txt](https://github.com/fwupd/fwupd/files/7550672/skip_online_resource_unavailable.txt)


As a experiment I skipped port online , I dont think is a kernel bug , 

ever after skipping ports online I see that "Error reading from file: Resource temporarily unavailable", however when we open single port we dont see this issue

superm1:
Is it timing? If you poll and retry until error goes away maybe?

krankun:
> Is it timing? If you poll and retry until error goes away maybe?

not sure , just to clarify offline will call open and enumerate retimers , online will call close and de enumerate retimers 

skipping online means retimer enumeration stays for ever and after nvm update we see the issue.
localhost ~ # fwupdmgr install /tmp/retimer-2.21.cab --allow-reinstall --force 
Authenticating?          [ -                                     ]
Installing on USB4 Retimer?-                                     ]
Writing?                 [***************************************]
Installing on USB4 Retimer?**************************************]
Error reading from file: Resource temporarily unavailable

superm1:
So how about clean up in composite cleanup instead? This is once per transaction then

krankun:
> So how about clean up in composite cleanup instead? This is once per transaction then

composite clean up looks good in that way we dont have to track the devices to close and we can close at the end of the operation ,

however that doesnt solve the issue, i think clean up is not an issue here , the multi device nvm update is the issue.

if we update nvm for port1 only >> success
if we update nvm for port3 only >> sucess

if we update together for both ports only 1st one is successful and for second results in "Error reading from file: Resource temporarily unavailable"

krankun:
> > So how about clean up in composite cleanup instead? This is once per transaction then
> 
> composite clean up looks good in that way we dont have to track the devices to close and we can close at the end of the operation ,
> 
> however that doesnt solve the issue, i think clean up is not an issue here , the multi device nvm update is the issue.
> 
> if we update nvm for port1 only >> success if we update nvm for port3 only >> sucess
> 
> if we update together for both ports only 1st one is successful and for second results in "Error reading from file: Resource temporarily unavailable"

these both ports nvm share same payload  file, so my guess the first one after update cleans up the payload file , let me check more on this.

krankun:
> > > So how about clean up in composite cleanup instead? This is once per transaction then
> > 
> > 
> > composite clean up looks good in that way we dont have to track the devices to close and we can close at the end of the operation ,
> > however that doesnt solve the issue, i think clean up is not an issue here , the multi device nvm update is the issue.
> > if we update nvm for port1 only >> success if we update nvm for port3 only >> sucess
> > if we update together for both ports only 1st one is successful and for second results in "Error reading from file: Resource temporarily unavailable"
> 
> these both ports nvm share same payload file, so my guess the first one after update cleans up the payload file , let me check more on this.

Yep the second device returns from https://github.com/krankun/fwupd/blob/main/libfwupdplugin/fu-device.c#L3572

krankun:
> > > > So how about clean up in composite cleanup instead? This is once per transaction then
> > > 
> > > 
> > > composite clean up looks good in that way we dont have to track the devices to close and we can close at the end of the operation ,
> > > however that doesnt solve the issue, i think clean up is not an issue here , the multi device nvm update is the issue.
> > > if we update nvm for port1 only >> success if we update nvm for port3 only >> sucess
> > > if we update together for both ports only 1st one is successful and for second results in "Error reading from file: Resource temporarily unavailable"
> > 
> > 
> > these both ports nvm share same payload file, so my guess the first one after update cleans up the payload file , let me check more on this.
> 
> Yep the second device returns from https://github.com/krankun/fwupd/blob/main/libfwupdplugin/fu-device.c#L3572

further debugging revealed we are failing at https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L738 in fu_thunderbolt_device_prepare_firmware , the path seems to be correct thou for usb4_port3 retimer /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_active1/nvmem

krankun:
> > > > > So how about clean up in composite cleanup instead? This is once per transaction then
> > > > 
> > > > 
> > > > composite clean up looks good in that way we dont have to track the devices to close and we can close at the end of the operation ,
> > > > however that doesnt solve the issue, i think clean up is not an issue here , the multi device nvm update is the issue.
> > > > if we update nvm for port1 only >> success if we update nvm for port3 only >> sucess
> > > > if we update together for both ports only 1st one is successful and for second results in "Error reading from file: Resource temporarily unavailable"
> > > 
> > > 
> > > these both ports nvm share same payload file, so my guess the first one after update cleans up the payload file , let me check more on this.
> > 
> > 
> > Yep the second device returns from https://github.com/krankun/fwupd/blob/main/libfwupdplugin/fu-device.c#L3572
> 
> further debugging revealed we are failing at https://github.com/krankun/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L738 in fu_thunderbolt_device_prepare_firmware , the path seems to be correct thou for usb4_port3 retimer /sys/devices/pci0000:00/0000:00:0d.2/domain0/0-0/usb4_port3/0-0:3.1/nvm_active1/nvmem

we decided that since both retimers share same payload updating just one retimer nvm is OK, I will rebase tha patch and go thru all review comments , for now we can settle this multiport retimer nvm update , Thank you! CC @superm1 @djcampello

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -67,3 +67,32 @@ fu_plugin_startup(FuPlugin *plugin, GError **error)
...
+		}
+	}
+	return TRUE;
+}
+
+
+gboolean
+fu_plugin_composite_cleanup	(FuPlugin	*plugin,
+							 GPtrArray	*devices,
+							 GError		**error)
```
hughsie, line 89: the spacing is funky here -- are you not using the precommit and clang-format rules?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,73 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+gboolean
+fu_thunderbolt_device_open(FuDevice *device, GError **error)
+{
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+	GUdevDevice *udev_device = NULL;
+	g_autoptr(GUdevDevice) udev_parent = NULL;
+	g_autoptr(FuUdevDevice) parent = NULL;
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER ||
+	    fu_thunderbolt_device_get_version(self, NULL))
+		return TRUE;
```
hughsie, line 363: {} here please

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -764,10 +838,13 @@ fu_thunderbolt_device_write_firmware(FuDevice *device,
...
 	/* whether to wait for a device replug or not */
-	if (!fu_device_has_flag(device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
-		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_RESTART);
-		fu_device_set_remove_delay(device, FU_PLUGIN_THUNDERBOLT_UPDATE_TIMEOUT);
-		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+	if (fu_device_has_flag (device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
+		fu_device_set_remove_delay (device, FU_PLUGIN_THUNDERBOLT_UPDATE_TIMEOUT);
+		fu_device_set_status (device, FWUPD_STATUS_DEVICE_RESTART);
+		if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER)
+		{
```
hughsie, line 845: clang format isn't being run for some reason

plugins/thunderbolt/fu-plugin-thunderbolt.c
krankun: Thanks @hughsie  am using eclipse , I was developing on chrome till now but will need to format code compile code on native ubuntu fix all warnings etc , will post the updates soon here

superm1: Run contrib/setup

krankun: ./contrib/ci/generate_dependencies.py -o ubuntu | sudo xargs apt install -y
./contrib/setup
meson build -Ddocs=none
after above steps , did you face below any work around please ? 
plugins/tpm/meson.build:46:2: ERROR: Bad source of type 'SharedModule' in target 'tpm-self-test'.

superm1: I think you need to pull/rebase on latest main.  That's fixed in https://github.com/fwupd/fwupd/commit/3e2e0aa61600f73433b149a183befddf5ab1bcef

krankun: Thats fixed, "meson build -Ddocs=none" configures the build right , how do we compile code in this env ?

superm1: This will build and run test suite:

```
ninja -C build test
```

krankun: from https://github.com/fwupd/fwupd/wiki/Compilation using "ninja -C build"

src/fu-udev-backend.c
```diff
@@ -41,7 +41,8 @@ fu_udev_backend_device_remove(FuUdevBackend *self, GUdevDevice *udev_device)
 	/* find the device we enumerated */
 	device_tmp =
 	    fu_backend_lookup_by_id(FU_BACKEND(self), g_udev_device_get_sysfs_path(udev_device));
-	if (device_tmp != NULL) {
+	if (device_tmp != NULL &&
+	    fu_device_has_flag(device_tmp, FWUPD_DEVICE_FLAG_REQUIRES_WAKEUP)) {
```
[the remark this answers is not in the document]
krankun: fixed in a new PR https://github.com/fwupd/fwupd/pull/4044

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,71 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 	return (guint16)val;
 }
 
+static void
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
+{
+	g_autoptr(GError) error_local = NULL;
+
+	if (!fu_udev_device_write_sysfs(device, "usb4_port3/offline", "1", &error_local)) {
```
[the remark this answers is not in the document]
krankun: for now we are doing only for usb4_port1 as usb4_port3 share same payload fixed in new PR https://github.com/fwupd/fwupd/pull/4044

plugins/thunderbolt/fu-plugin-thunderbolt.c
krankun: I have fixed a failing test case since usb4_port1 dir was not available in https://github.com/fwupd/fwupd/pull/4044/commits/2442c83a5f2b5e59bb9069c4ef46571a5cf7960e

krankun: new PR is here https://github.com/fwupd/fwupd/pull/4044

krankun: not sure how to update this branch as it needs fast forward now, hence created a new https://github.com/fwupd/fwupd/pull/4044

superm1:
superceded by #4044

Issue #3901, opened 2021-10-22
Empty PCR check for HSI


kerneis-anssi:
**Is your feature request related to a problem? Please describe.**

Google security reported that some firmwares fail to measure anything in the PCRs, especially when they have two banks of PCRs (sha1 and sha256).
 
**Describe the solution you'd like**

A plugin in fwupd would report a HSI-1 failure if any PCR0-7 in any bank is empty.

**Describe alternatives you've considered**

- Reporting only if *all* PCRs in a given bank are empty. That would match more closely the reported issue, but I think any of those core PCRs being empty is a strong enough signal.
- Reporting only if a PCR is empty in one bank and filled in the other bank. Likewise, seems needlessly restrictive.
- 
**Additional context**

https://github.com/google/security-research/blob/master/pocs/bios/tpm-carte-blanche/writeup.md

kerneis-anssi:
Wouldn't it make more sense to refactor this code to a device implemented in the `tpm` plugin, and have both `eventlog` and `uefi-capsule` use that? The fact that it currently lives in uefi-capsule is a historical accident, and it doesn't really belong to tpm-eventlog either.

hughsie:
@kerneis-anssi does that make things better? It should be easy to implement now, correct?

superm1:
`ninja -C build test` will run the test suite.

Some unit tests do load the actual plugin file, others just instantiate pieces of code (as you described).  If you want to see the test working with the plugin loading you may need to change it slightly.

hughsie:
> I have what I think is working code.

Yay! Post a PR, mark it draft and we can review it gently.

> Do you have an example of a plugin that tests if a security attribute is added

The easiest way to test this is something like:

    ./src/fwupdtool security --force --plugins tpm --verbose

kerneis-anssi:
> Yay! Post a PR, mark it draft and we can review it gently.

I'll do this as soon as some paperwork has made its way through the maze of French bureaucracy ;-)

kerneis-anssi:
Thanks for your changes in response to my comments. It helped a lot setting things up on my side. I have a complete implementation together with unit test. I'll push it for review as soon as things are cleared on my side (it's taking an unusually long time so fingers crossed it'll be unblocked soon).

PR #3923, opened 2021-10-30
Move the TPM handling into the TPM place

hughsie:
The fact that it currently lives in uefi-capsule is a historical
accident, and it doesn't really belong to tpm-eventlog either.

Fixes some of https://github.com/fwupd/fwupd/issues/3901

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
LGTM, but @CragW suggest you to test this on Dell system to make sure that it still works since TPM handling is partially done by "dell" plugin.

hughsie:
I'll do this after the 1.7.1 release as it doesn't seem like a super-hot idea to do this on release day.

PR #3924, opened 2021-10-30
trivial: Add some more details to the HSI specification

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation


superm1 (commented):
I would suggest a minor change.  Rather than:
```For HSI-1 the test result must be....```
change to
```To meet HSI-1 on systems that this test, the result must be....```

hughsie:
> To meet HSI-1 on systems that this test, the result must be

Do you mean something like `To meet HSI-1 on systems that run this test, the result must be` -- I think you missed out a word somewhere.

PR #3925, opened 2021-11-01
trivial: system firmware may primarily use sha256 bank instead sha1

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

cherry-pick from b81604f6c17a0829e25fb4b0709889461936ce96


superm1:
Ah and this is also in 60ea02201 in 1.6.x.

superm1 (approved):
This LGTM.  The CI failure on 1.5.x is not caused by it.  Can you please look into fixing it though separately.

CragW:
CI failure looks like GLib suggesting to use new func wrapper `g_spawn_check_wait_status `as `g_spawn_check_exit_status `is trending deprecated.  We will need #3735 to pass CI tests.

```
../plugins/cpu/fu-plugin-cpu.c:92:2: error: ‘g_spawn_check_exit_status’ is deprecated: Use 'g_spawn_check_wait_status' instead [-Werror=deprecated-declarations]
   92 |  if (!g_spawn_check_exit_status (exit_status, &error_local)) {
      |  ^~

```

PR #3931, opened 2021-11-01
Merge tpm-eventlog into the tpm plugin

hughsie:
This is so much less complicated.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 if you think this is a good idea I'll merge the data files in fwupd-test-firmware and re-run CI.

superm1:
I mean I think it makes sense - but we need to keep test firmware repo compatible across different releases.

hughsie:
> but we need to keep test firmware repo compatible across different releases

Symlinks to the rescue. Done.

PR #3937, opened 2021-11-02
Version lock plugins to the daemon in the example spec file

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -164,6 +164,7 @@ Files for development with %{name}.
 
 %package tests
 Summary: Data files for installed tests
+Requires: %{name}%{?_isa} = %{version}-%{release}
 
 %description tests
 Data files for installed tests.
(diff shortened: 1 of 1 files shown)


hughsie:
..otherwise we get a "tainted" fwupd if we just do `dnf update fwupd`

superm1:
Is that the root of the problem really though?  Feels like a bandaid.  Maybe instead should we be examining `FU_BUILD_HASH` from the plugin and refusing to load it entirely?

hughsie:
> Is that the root of the problem really though

Well, it's the root of the problem in that it forces dnf to update all the fwupd things in one transaction. I don't think we want to be any more aggressive with the tainting (e.g. not loading plugins with the wrong hash) as it makes development and getting vendors onboard much harder.

superm1 (approved):

PR #3940, opened 2021-11-02
Speed up daemon startup by 15%

hughsie:
During my fwupd startup fu_plugin_has_custom_flag gets called 21 times
which causes all HWIDs to be enumerated with 346 calls to the quite
expensive fu_context_lookup_quirk_by_id() function.

Move the flag to a private hashset and enumerate the HWIDs only during
startup. There's nothing plugin specific about them anyway...

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -611,30 +611,12 @@ gboolean
 fu_plugin_has_custom_flag(FuPlugin *self, const gchar *flag)
 {
 	FuPluginPrivate *priv = GET_PRIVATE(self);
-	GPtrArray *guids;
 
 	g_return_val_if_fail(FU_IS_PLUGIN(self), FALSE);
 	g_return_val_if_fail(flag != NULL, FALSE);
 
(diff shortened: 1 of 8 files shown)


superm1 (approved):

libfwupdplugin/fu-plugin.c
```diff
@@ -611,30 +611,12 @@ gboolean
 fu_plugin_has_custom_flag(FuPlugin *self, const gchar *flag)
 {
 	FuPluginPrivate *priv = GET_PRIVATE(self);
-	GPtrArray *guids;
 
 	g_return_val_if_fail(FU_IS_PLUGIN(self), FALSE);
 	g_return_val_if_fail(flag != NULL, FALSE);
 
 	/* never set up, e.g. in tests */
```
YehezkelShB, line 618: I know I'm late to this party, but the comments seems irrelevant for the code now (and raises the question if tests should've been handled differently in the new change too)

PR #3944, opened 2021-11-03
Build logind plugin for elogind-enabled systems

ixti:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logind/meson.build
+++ b/plugins/logind/meson.build
@@ -1,4 +1,4 @@
-if get_option('systemd')
+if get_option('systemd') or get_option('elogind')
 if host_machine.system() != 'linux'
   error('linux is required for systemd')
 endif


hughsie:
@superm1 the Greetings CI target is doing something strange.

superm1:
> @superm1 the Greetings CI target is doing something strange.

What happened?

hughsie:
It failed; was that expected?

superm1:
> It failed; was that expected?

Oh... It's this: https://github.com/actions/first-interaction/issues/10
There is a workaround posted there too to try.  We should try that and if it doesn't work abandon the greeting.

superm1:
> > It failed; was that expected?
> 
> Oh... It's this: [actions/first-interaction#10](https://github.com/actions/first-interaction/issues/10) There is a workaround posted there too to try. We should try that and if it doesn't work abandon the greeting.

I applied it to main, we'll see if it helps.

PR #3947, opened 2021-11-04
Wip/hughsie/go faster stripes4

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5810,6 +5810,12 @@ fu_engine_add_device(FuEngine *self, FuDevice *device)
 		}
 	}
 
+	/* set or clear the SUPPORTED flag */
+	fu_engine_ensure_device_supported(self, device);
+
+	/* fixup the name and format as needed */
+	fu_engine_md_refresh_device_from_component(self, device, component);
+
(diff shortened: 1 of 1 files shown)


src/fu-engine.c
```diff
@@ -990,23 +991,53 @@ fu_engine_verify_update(FuEngine *self, const gchar *device_id, GError **error)
...
 	for (guint i = 0; i < guids->len; i++) {
 		const gchar *guid = g_ptr_array_index(guids, i);
-		xb_string_append_union(xpath,
-				       "components/component[@type='firmware']/"
-				       "provides/firmware[@type='flashed'][text()='%s']/"
-				       "../..",
-				       guid);
+		XbNode *component = fu_engine_get_component_by_guid(self, guid);
+		if (component != NULL)
+			return component;
```
superm1, line 1039: How about `break` here instead and make the final function return `component`?

superm1: (and move that component declaration up out of the for loop)

hughsie: Yes, I guess that's clearer; fixed.

superm1 (approved):

PR #3948, opened 2021-11-04
Wip/hughsie/mtd

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-mtd-device.c
```diff
@@ -0,0 +1,309 @@
...
+							  "erasesize",
+							  &priv->erasesize,
+							  NULL))
+			return FALSE;
+	}
+	if (flags & MTD_WRITEABLE)
+		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
+	if (flags & MTD_POWERUP_LOCK)
+		priv->powerup_lock = TRUE;
+#endif
```
superm1, line 106: I think you should fail probe without that `HAVE_MTD_USER_H` no?  It seems pointless to me to show the device as around but not updatable.

libfwupdplugin/fu-mtd-device.c
```diff
@@ -0,0 +1,309 @@
...
+		g_strdelimit(name_safe, " /\\\"", '-');
+		devid = g_strdup_printf("MTD\\NAME_%s", name_safe);
+		fu_device_add_instance_id(FU_DEVICE(self), devid);
+	}
+
+	/* get properties about the device */
+	if (!fu_udev_device_get_sysfs_attr_uint64(FU_UDEV_DEVICE(device), "size", &size, NULL))
+		return FALSE;
+	fu_device_set_firmware_size_max(device, size);
+	if (!fu_udev_device_get_sysfs_attr_uint64(FU_UDEV_DEVICE(device), "flags", &flags, NULL))
```
superm1, line 92: you forgot to pass in error here

meson.build
```diff
@@ -314,6 +314,9 @@ endif
 if cc.has_header('linux/ethtool.h')
   conf.set('HAVE_ETHTOOL_H', '1')
 endif
+if cc.has_header('mtd/mtd-user.h')
+  conf.set('HAVE_MTD_USER_H', '1')
+endif
```
superm1, line 319: I wonder if you should only compile this subsystem and device type when this header is around.  It's a lot of permanently dead code for BSD and Windows otherwise.

hughsie: It's part of the ABI if it's in libfwupdplugin (and I think it needs to be)

superm1: Yeah I see your point.  I think then it should fail probe without that header around at least for now.

libfwupdplugin/fu-mtd-device.c
hughsie: I think it's best to disable the `mtd` plugin (not yet pushed) in this case; this means we don't do the udev probing stuff at all and it seems cleaner.

plugins/mtd/meson.build
```diff
@@ -0,0 +1,29 @@
+cargs = ['-DG_LOG_DOMAIN="FuPluginMtd"']
```
superm1, line 1: I think we don't want to bother building this for other OSes besides Linux - shouldn't it have a guard?

plugins/mtd/fu-plugin-mtd.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+gboolean
+fu_plugin_device_created(FuPlugin *plugin, FuDevice *dev, GError **error)
+{
+	fu_device_set_summary(dev, "Memory Technology Device");
+	fu_device_add_protocol(dev, "org.infradead.mtd");
+	fu_device_add_flag(dev, FWUPD_DEVICE_FLAG_INTERNAL);
+	fu_device_add_flag(dev, FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
+	fu_device_add_vendor_id(dev, "XXX");
```
superm1, line 39: TODO I guess

plugins/mtd/fu-plugin-mtd.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+gboolean
+fu_plugin_startup(FuPlugin *plugin, GError **error)
+{
+#ifndef HAVE_MTD_USER_H
+	g_set_error_literal(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "Not supported as mtd-user.h is unavailable");
```
superm1, line 26: as a runtime error to a user this is confusing - it should be `Not compiled with mtd support`

hughsie: Agree, fixed, thanks.

plugins/mtd/fu-plugin-mtd.c
hughsie: I've changed this to FU_HWIDS_KEY_MANUFACTURER

meson.build
hughsie: > and I think it needs to be

Actually it doesn't need to be. I've moved it out to the plugin now, I think that's a lot cleaner.

superm1 (approved):

PR #3955, opened 2021-11-05
Allow setting the FuUdevDevice bind-id at runtime

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -41,6 +41,7 @@ typedef struct {
 	guint32 subsystem_model;
 	guint8 revision;
 	gchar *subsystem;
+	gchar *bind_id;
 	gchar *driver;
 	gchar *device_file;
 	gint fd;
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -163,6 +174,42 @@ fu_udev_device_to_string(FuDevice *device, guint idt, GString *str)
 #endif
 }
 
+static gboolean
+fu_udev_device_ensure_bind_id(FuUdevDevice *self, GError **error)
+{
+	FuUdevDevicePrivate *priv = GET_PRIVATE(self);
+
```
superm1, line 181: maybe a return val if fail if subsystem is null?

superm1 (approved):
I'm interested to hear what example doesn't let you derive it from heuristics, but this is fine.

libfwupdplugin/fu-udev-device.c
hughsie: I think my logic was that it would get a proper GError if NULL at the end of the function. e.g. `cannot derive bind-id from subsystem NULL`

hughsie:
> I'm interested to hear what example doesn't let you derive it from heuristics

At the moment it's for a device I've not got in my hands. It seems for some ARM hardware at least the bind-id might actually be available from DT, but not predictable from the subsystem.

PR #3956, opened 2021-11-05
Use sh instead of bash

rossburton:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

It's trivial to use sh instead of bash, and some people may not want bash pulled in.


hughsie:
@superm1 I bow to your better UNIX-foo.

superm1 (approved):

PR #3958, opened 2021-11-05
trivial: Add some Impact sections to HSI attributes

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/docs/hsi.md
+++ b/docs/hsi.md
@@ -144,7 +144,9 @@ UEFI Secure boot is a verification mechanism for ensuring that code launched by 
 
 Secure Boot requires that each binary loaded at boot is validated against trusted certificates.
 
-Possible results:
+**Impact:** When Secure Boot is not enabled any EFI binary can be run at startup, which gives the attacker full access to your hardware.
+
+**Possible results:**
 
 - `not-found`: support has not been detected
 - `not-enabled`: detected, but has been turned off
@@ -152,7 +154,11 @@ Possible results:
 
 To meet HSI-1 on UEFI systems that run this test, the result must be `enabled`. *[v1.5.0]*
 
-References:
+**Resolution:**
+
+Turn off CSM boot and enable Secure Boot in the BIOS setup.
+
+**References:**
 
 - [Ubuntu SecureBoot Wiki Page](https://wiki.ubuntu.com/UEFI/SecureBoot)
 
@@ -163,14 +169,17 @@ References:
 UEFI defines a platform key for the system.
 This should not be a test key, e.g. `DO NOT TRUST - AMI Test PK`
 
-Possible results:
+**Impact:** It is possible to sign an EFI binary with the test platform key, which invalidates the Secure Boot trust chain.
+It effectively gives the local attacker full access to your hardware.
+
+**Possible results:**
 
 - `valid`: valid key
 - `not-valid`: an invalid key has been enrolled
 
 To meet HSI-1 on UEFI systems that run this test, the result must be `valid`. *[v1.5.0]*
 
-References:
+**References:**
 
 - [Ubuntu SecureBoot Wiki Page](https://wiki.ubuntu.com/UEFI/SecureBoot/Testing)
 
@@ -181,7 +190,10 @@ References:
 Intel hardware provides this mechanism to protect the SPI ROM chip located on the motherboard from being overwritten by the operating
 system. The `BIOSWE` bit must be unset otherwise userspace can write to the SPI chip.
 
-Possible results:
+**Impact:** The system firmware can be written from userspace.
+This gives any attacker with root access a method to write persistent executable code to the firmware, which survives even a full disk wipe and OS reinstall.
+
+**Possible results:**
 
 - `not-found`: the SPI device was not found
 - `not-enabled`: write enable is disabled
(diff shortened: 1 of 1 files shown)


hughsie:
Screenshot 2021-11-05 at 14-58-55 FwupdPlugin Host Security ID Specification

hughsie:
@superm1 this is going to need proper review; both for my crumby English but also in that my impact statements are proportionate. @skontar happy for comments too.

docs/hsi.md
```diff
@@ -144,15 +144,21 @@ UEFI Secure boot is a verification mechanism for ensuring that code launched by
 
 Secure Boot requires that each binary loaded at boot is validated against trusted certificates.
 
-Possible results:
+**Impact:** When Secure Boot is not enabled any EFI binary can be run at startup, which gives the local attacker full access to your hardware.
```
superm1, line 147: Why they gotta be local? UEFI has network stacks these days...

docs/hsi.md
```diff
@@ -307,14 +337,16 @@ Converged Security and Manageability Engine is a standalone management module th
 The CSME lives in the PCH and can only be updated by the OEM vendor.
 The version of the CSME module can be checked to detect the most common and serious vulnerabilities: CVE-2017-5705, CVE-2017-5708, CVE-2017-5711, CVE-2017-5712, CVE-2017-5711, CVE-2017-5712, CVE-2017-5706, CVE-2017-5709, CVE-2017-5707 or CVE-2017-5710.
 
-Possible results:
+**Impact:** Using any one of the critical vulnerabilities, a remote hacker can take full control of the system and all connected devices, even when the system is powered off.
```
skontar, line 340: I recommend using word "attacker" instead of "hacker".

docs/hsi.md
```diff
@@ -512,14 +574,16 @@ References:
 
 Control enforcement technology is available on new Intel platforms and prevents exploits from hijacking the control-flow transfer instructions for both forward-edge (indirect call/jmp) and back-edge transfer (ret).
 
-Possible results:
+**Impact:** A local or physical attacker with an existing unrelated vulnerability can use a ROP gadget to run arbitrary code.
```
superm1, line 577: This is right but layman won't know what it means.

superm1 (dismissed):

docs/hsi.md
```diff
@@ -402,14 +446,17 @@ References:
 BootGuard is a processor feature that prevents the machine from running firmware images not released by the system manufacturer.
 It forms a root-of-trust by fusing in cryptographic keys into the processor itself that are used to verify the Authenticated Code Modules found in the SPI flash.
 
-Possible results:
+**Impact:** When BootGuard is not set up correctly then the chain-of-trust between the CPU and the bootloader can not be verified.
+This would allow subverting the Secure Boot protection.
```
skontar, line 450: I believe that any time you say "allow subverting the Secure Boot protection" you could go one step further. What can the attacker do when "Secure Boot protection is subverted"?

docs/hsi.md
hughsie: How about adding `If UEFI networking is enabled then this widens the attack surface to remote attackers too.` -- or I can just remove the `local` word...

docs/hsi.md
hughsie: `A local or physical attacker with an existing unrelated vulnerability can use a reliable and well-known method to run arbitrary code.` ?

docs/hsi.md
```diff
@@ -549,14 +615,16 @@ References:
 Without Supervisor Mode Access Prevention, the supervisor code usually has full read and write access to user-space memory mappings.
 This can make exploits easier to write, as it allows the kernel to access user-space memory when it did not intend to.
 
-Possible results:
+**Impact:** A local or remote hacker can use a simple exploit to modify the contents of kernel memory which can lead to privilege escalation.
```
skontar, line 618: "hacker" -> "attacker"

docs/hsi.md
superm1: Some bios vendors might not let you disable it.
I say remove local.

docs/hsi.md
superm1: That's fine

skontar (changes requested):

docs/hsi.md
hughsie: Done, thanks.

docs/hsi.md
hughsie: Fixed.

docs/hsi.md
hughsie: Fixed both places, thanks.

docs/hsi.md
hughsie: `This would allow subverting the Secure Boot protection which gives the attacker full access to your hardware.`?

docs/hsi.md
hughsie: Fixed, thanks.

docs/hsi.md
skontar: Sounds good.

hughsie: Fixed, thanks.

PR #3962, opened 2021-11-07
Wip/hughsie/vendor names

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -2145,7 +2145,9 @@ fu_device_fixup_vendor_name(FuDevice *self)
 	const gchar *name = fu_device_get_name(self);
 	const gchar *vendor = fu_device_get_vendor(self);
 	if (name != NULL && vendor != NULL) {
-		if (g_str_has_prefix(name, vendor)) {
+		g_autofree gchar *name_up = g_utf8_strup(name, -1);
+		g_autofree gchar *vendor_up = g_utf8_strup(vendor, -1);
+		if (g_str_has_prefix(name_up, vendor_up)) {
 			gsize vendor_len = strlen(vendor);
 			g_autofree gchar *name1 = g_strdup(name + vendor_len);
 			g_autofree gchar *name2 = fu_common_strstrip(name1);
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -2155,6 +2157,54 @@ fu_device_fixup_vendor_name(FuDevice *self)
...
+	struct {
+		const gchar *old;
+		const gchar *new;
+	} vendor_names[] = {{"ADATA Technology Co., Ltd.", "ADATA"},
+			    {"Chicony Electronics Co.,Ltd.", "Chicony"},
+			    {"LENOVO", "Lenovo"},
+			    {"Micron Technology Inc", "Micron"},
+			    {"Samsung Electronics Co Ltd", "Samsung"},
+			    {"VIA Labs, Inc.", "VIA"},
+			    {NULL, NULL}};
```
superm1, line 2186: Which vendor prompted this?  I wonder if some vendors might not be comfortable with their name being modified when it's displayed.

superm1:
The logitech commit is fine, but I think we shouldn't be messing with "all" vendor strings unless the vendors OK it.

libfwupdplugin/fu-device.c
hughsie: The root problem is that vendors are using the legal name in the NVMe or OUI and using the "friendly name" in the product string. So we get something like `Samsung Electronics Co Ltd SAMSUNG FOOBARBAZ` which is just terrible.

superm1: Should this fixup only be for NVME/ATA then perhaps?

hughsie: The other two is Chicony, which is the pixart plugin, and Lenovo which is the uefi_capsule plugin.

superm1: Hmm. I still don't think it's the right precedent to do for all plugins.
A few local fixups to plugins are better in my opinion.
 I don't want legal troubles in fwupd because it shows Foo instead of the legal name Foo Limited.

hughsie: How about we just do the `"Samsung Electronics Co Ltd" -> "Samsung"` one in the NVME plugin as it's the worst offender in the `Samsung Electronics Co Ltd SAMSUNG FOOBARBAZ` case.

superm1: sounds good

PR #3965, opened 2021-11-08
Wip/hughsie/go faster stripes5

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-plugin.h
```diff
@@ -61,6 +72,291 @@ typedef enum {
...
+typedef struct {
+	/**
+	 * init:
+	 * @self: A #FuPlugin
+	 *
+	 * Initializes the plugin.
+	 * Sets up any static data structures for the plugin.
+	 * Most plugins should call fu_plugin_set_build_hash in here.
+	 *
+	 * Since: 0.8.0
```
superm1, line 85: Uhh, no this didn't come from 0.8.0 with this change. Isn't this an ABI break?

libfwupdplugin/fu-plugin.h
```diff
@@ -61,6 +72,291 @@ typedef enum {
 	FU_PLUGIN_VERIFY_FLAG_LAST
 } FuPluginVerifyFlags;
 
+/* one day this will be part of FuPluginClass */
```
superm1, line 75: If you're breaking ABI already anyway with this change and need to bump up the soname, do this now.

libfwupdplugin/fu-plugin.h
superm1: I think yeah bump plugin version.

libfwupdplugin/fu-plugin.c
```diff
@@ -357,17 +352,30 @@ fu_plugin_open(FuPlugin *self, const gchar *filename, GError **error)
 		return FALSE;
 	}
 
+	/* call the new-style fu_plugin_init_vfuncs() vfunc setup */
```
superm1, line 355: no need to call it new style

libfwupdplugin/fu-plugin.c
```diff
@@ -1169,38 +1094,6 @@ fu_plugin_runner_coldplug_prepare(FuPlugin *self, GError **error)
 gboolean
 fu_plugin_runner_coldplug_cleanup(FuPlugin *self, GError **error)
```
superm1, line 1095: If bumping ABI anyways - drop this now

libfwupdplugin/fu-plugin.c
```diff
@@ -1120,38 +1077,6 @@ fu_plugin_runner_coldplug(FuPlugin *self, GError **error)
 gboolean
 fu_plugin_runner_coldplug_prepare(FuPlugin *self, GError **error)
```
superm1, line 1078: If bumping ABI anyways - drop this now

libfwupdplugin/fu-plugin.c
```diff
@@ -357,17 +352,30 @@ fu_plugin_open(FuPlugin *self, const gchar *filename, GError **error)
...
 	/* set automatically */
 	if (fu_plugin_get_name(self) == NULL) {
 		g_autofree gchar *str = fu_plugin_guess_name_from_fn(filename);
 		fu_plugin_set_name(self, str);
 	}
 
 	/* optional */
-	g_module_symbol(priv->module, "fu_plugin_init", (gpointer *)&func);
-	if (func != NULL) {
+	if (vfuncs->init != NULL) {
```
superm1, line 376: this should be compulsory no?

libfwupdplugin/fu-plugin.h
```diff
@@ -61,6 +72,291 @@ typedef enum {
...
+	 * @error: (nullable): optional return location for an error
+	 *
+	 * Function run after updating group of composite devices.
+	 *
+	 * Since: 1.0.9
+	 **/
+	gboolean (*composite_cleanup)(FuPlugin *self, GPtrArray *devices, GError **error);
+	/*< private >*/
+	gpointer padding[9];
+} FuPluginVfuncs;
```
superm1, line 358: hey I got an idea here.  How about you include a member here for `build_hash`?  Then you can stop having to have every plugin call set_build_hash and it can be part of the vtable.

hughsie: ack, good idea. will try.

libfwupdplugin/fu-plugin.h
```diff
@@ -61,6 +72,291 @@ typedef enum {
...
+	 * @devices: (element-type FuDevice): array of devices
+	 * @error: (nullable): optional return location for an error
+	 *
+	 * Function run after updating group of composite devices.
+	 *
+	 * Since: 1.0.9
+	 **/
+	gboolean (*composite_cleanup)(FuPlugin *self, GPtrArray *devices, GError **error);
+	/*< private >*/
+	gpointer padding[9];
```
superm1, line 357: any logic between this otherwise arbitrary number?

hughsie: 32 in total seemed a nice round number. A bigger number == more times we can add plugin vfunc without bumping plugin ABI, but plugins are slightly bigger. Maybe if it's const ro then it's not a concern at all.

libfwupdplugin/fu-plugin.c
hughsie: Agree, fixed, thanks.

libfwupdplugin/fu-plugin.c
hughsie: With the super new `vfuncs->build_hash`, no.

libfwupdplugin/fu-plugin.h
hughsie: It's not const ro.

libfwupdplugin/fu-plugin.h
hughsie: That works really well. Done.

libfwupdplugin/fu-plugin.h
superm1: Done? or not?

hughsie: Ohh, sorry, replied to the wrong one. I didn't do this because we can't -- the FuPluginClass is per-type, and not per-instance. I don't think it makes sense to have a FuDfuPlugin GType with vfuncs, at least not at this point, as we'd then need a dlsym entrypoint like now to return the custom GType. It's a whole ton of boilerplate without much benefit IMHO. Maybe we just remote the comment?

superm1: Yeah remove comment then

hughsie: Done

PR #3969, opened 2021-11-09
Allow overriding HwId data from a config file

hughsie:
To use this create a file in /var/lib/fwupd/hwids.conf

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/daemon.conf
+++ b/data/daemon.conf
@@ -48,3 +48,12 @@ IgnorePower=false
 
 # Only support installing firmware signed with a trusted key
 OnlyTrusted=true
+
+# these are only required when the SMBIOS or Device Tree data is invalid or missing
+#Manufacturer=
+#ProductName=
+#ProductSku=
+#Family=
+#EnclosureKind=
+#BaseboardProduct=
+#BaseboardManufacturer=
--- a/libfwupdplugin/fu-hwids.c
+++ b/libfwupdplugin/fu-hwids.c
@@ -351,6 +351,41 @@ fu_hwids_convert_integer_cb(FuSmbios *smbios, guint8 type, guint8 offset, GError
 	return g_strdup_printf("%x", tmp);
 }
 
+static gboolean
+fu_hwids_setup_overrides(FuHwids *self, GError **error)
+{
+	g_autofree gchar *localstatedir = fu_common_get_path(FU_PATH_KIND_LOCALSTATEDIR_PKG);
+	g_autofree gchar *sysconfigdir = fu_common_get_path(FU_PATH_KIND_SYSCONFDIR_PKG);
+	g_autoptr(GKeyFile) kf = g_key_file_new();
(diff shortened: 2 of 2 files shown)


superm1 (commented):
I guess this is an easier to use version of `FWUPD_SYSFSFWDIR`.  I think it makes sense, but shouldn't it be in `/etc`?

libfwupdplugin/tests/hwids.conf
```diff
@@ -0,0 +1,6 @@
+[fwupd]
```
superm1, line 1: does this need to be distributed in `fwupd-test` binary packages?

hughsie:
>  I think it makes sense, but shouldn't it be in /etc

I did ponder adding this to `daemon.conf` and then also supporting loading from `/usr/lib/fwupd/daemon.conf` as an override.

superm1:
> >  I think it makes sense, but shouldn't it be in /etc
> 
> I did ponder adding this to `daemon.conf` and then also supporting loading from `/usr/lib/fwupd/daemon.conf` as an override.

That feels more correct to me.

hughsie:
> That feels more correct to me.

Agreed; PR updated.

data/daemon.conf
```diff
@@ -48,3 +48,7 @@ IgnorePower=false
 
 # Only support installing firmware signed with a trusted key
 OnlyTrusted=true
+
+# these are only required when the SMBIOS or Device Tree data is invalid or missing;
+# the full list can be found here: https://fwupd.github.io/libfwupdplugin/
```
superm1, line 53: why not just list them all?  This is (relatively) static and there are only 13 of them.

hughsie: I think it's actually better than that; we only want to override the non-version ones. Patch updated.

superm1 (commented):
Just one idea, otherwise LGTM now

libfwupdplugin/fu-hwids.c
```diff
@@ -351,6 +351,50 @@ fu_hwids_convert_integer_cb(FuSmbios *smbios, guint8 type, guint8 offset, GError
...
+	/* optional overrides */
+	localstatedir = fu_common_get_path(FU_PATH_KIND_LOCALSTATEDIR_PKG);
+	fn_var = g_build_filename(localstatedir, "daemon.conf", NULL);
+	if (g_file_test(fn_var, G_FILE_TEST_EXISTS)) {
+		g_debug("loading HwId overrides from %s", fn_var);
+		if (!g_key_file_load_from_file(kf, fn_var, G_KEY_FILE_NONE, error))
+			return FALSE;
+	} else {
+		g_debug("not loading HwId overrides from %s", fn_var);
+	}
```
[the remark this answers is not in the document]
hughsie: Good idea, something like that^^?

superm1 (approved):
Yup, looks great!

PR #3976, opened 2021-11-11
Wip/hughsie/redfish openbmc

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/redfish/fu-plugin-redfish.c
+++ b/plugins/redfish/fu-plugin-redfish.c
@@ -304,30 +304,33 @@ fu_plugin_redfish_startup(FuPlugin *plugin, GError **error)
 	if (redfish_uri != NULL) {
 		const gchar *ip_str = NULL;
 		g_auto(GStrv) split = NULL;
-		guint64 port;
+		guint64 port = 0;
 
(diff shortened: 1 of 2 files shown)


plugins/redfish/fu-plugin-redfish.c
```diff
@@ -304,25 +304,28 @@ fu_plugin_redfish_startup(FuPlugin *plugin, GError **error)
...
+					    "invalid scheme");
 			return FALSE;
 		}
 
 		split = g_strsplit(ip_str, ":", 2);
 		fu_redfish_backend_set_hostname(data->backend, split[0]);
-		port = g_ascii_strtoull(split[1], NULL, 10);
+		if (g_strv_length(split) > 1)
+			port = g_ascii_strtoull(split[1], NULL, 10);
 		if (port == 0) {
```
superm1, line 329: you should also be checking for `G_MAXUINT64` and if so look at errno.

hughsie: I've added the G_MAXUINT64 check, although I'm not sure we want to care about overflow. The message is just `no valid port specified` now.

superm1 (approved):

PR #3981, opened 2021-11-12
Wip/hughsie/size

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -91,13 +91,15 @@ if get_option('systemd') and build_daemon
   con2.set('motd_dir', motd_dir)
 
   # replace @bindir@
-  configure_file(
-    input : 'fwupd-offline-update.service.in',
-    output : 'fwupd-offline-update.service',
-    configuration : con2,
-    install: true,
-    install_dir: systemdunitdir,
-  )
+  if get_option('offline')
+    configure_file(
+      input : 'fwupd-offline-update.service.in',
+      output : 'fwupd-offline-update.service',
+      configuration : con2,
+      install: true,
+      install_dir: systemdunitdir,
+    )
+  endif
 
   # replace @dynamic_options@
   configure_file(
--- a/data/meson.build
+++ b/data/meson.build
@@ -26,13 +26,14 @@ if build_standalone
     install_dir: join_paths(datadir, 'fwupd', 'quirks.d'))
 endif
 
-install_data(['org.freedesktop.fwupd.metainfo.xml'],
-  install_dir: join_paths(datadir, 'metainfo')
-)
-
-install_data(['org.freedesktop.fwupd.svg'],
-  install_dir : join_paths(datadir, 'icons', 'hicolor', 'scalable', 'apps')
-)
+if get_option('metainfo')
+  install_data(['org.freedesktop.fwupd.metainfo.xml'],
+    install_dir: join_paths(datadir, 'metainfo')
+  )
+  install_data(['org.freedesktop.fwupd.svg'],
+    install_dir : join_paths(datadir, 'icons', 'hicolor', 'scalable', 'apps')
+  )
+endif
 
 if build_daemon
   install_data(['org.freedesktop.fwupd.conf'],
--- a/data/meson.build
+++ b/data/meson.build
@@ -1,8 +1,14 @@
 subdir('builder')
 subdir('pki')
 subdir('remotes.d')
-subdir('bash-completion')
-subdir('fish-completion')
+
+if get_option('bash_completion')
+  subdir('bash-completion')
+endif
+
+if get_option('fish_completion')
+  subdir('fish-completion')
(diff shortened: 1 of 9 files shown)


hughsie:
@rossburton FYI

data/meson.build
```diff
@@ -2,7 +2,10 @@ subdir('builder')
 subdir('pki')
 subdir('remotes.d')
 subdir('bash-completion')
+
+if get_option('fish_completion')
```
superm1, line 6: I think this should group with bash completion. You don't want any completions for an embedded target probably.

meson_options.txt
```diff
@@ -47,3 +47,6 @@ option('curl', type : 'boolean', value : true, description : 'enable libcurl sup
 option('udevdir', type: 'string', value: '', description: 'Directory for udev rules')
 option('efi_os_dir', type: 'string', description : 'the hardcoded name of OS directory in ESP, e.g. fedora')
 option('efi_binary', type: 'boolean', value : true, description : 'generate uefi binary if missing')
+option('metainfo', type: 'boolean', value : true, description : 'install the project metainfo.xml information')
+option('fish_completion', type: 'boolean', value : true, description : 'install the fish integration files')
+option('offline', type: 'boolean', value : true, description : 'enable installing offline firmware')
```
superm1, line 52: This wording will be very confusing. Offline doesn't have the obvious meaning here

src/fu-engine.c
```diff
@@ -2423,6 +2423,15 @@ fu_engine_schedule_update(FuEngine *self,
 	g_autoptr(FuHistory) history = NULL;
 	g_autoptr(GFile) file = NULL;
 
+#ifndef HAVE_FWUPDOFFLINE
```
superm1, line 2426: Maybe we should change the default in packaging we have in tree for non-x86 too. Offline is only used for EC on an x86 design right?

data/meson.build
hughsie: maybe if `get_option('completion')` instead? Or add a `bash_completion` one too?

meson_options.txt
hughsie: `enable installing firmware using a pre-boot systemd target` perhaps?

data/meson.build
superm1: I say the former

meson_options.txt
superm1: Yes

src/fu-engine.c
hughsie: Kinda, it's not exactly designed for x86 -- but we know on some platforms (that happen to be non-x86) that we're not going to use it.

src/meson.build
```diff
@@ -107,7 +107,7 @@ meson.add_install_script(
       join_paths(get_option('prefix'), get_option('bindir'), 'fwupdagent')))
 endif
 
-if get_option('systemd')
+if get_option('systemd') and get_option('offline')
```
superm1, line 110: I think there's a new matrix here that you need to check where offline is enabled but systemd is disabled. That shouldn't be a valid configuration.

data/meson.build
hughsie: Although, there seems to be some precedent for `-Dbash_completion=false` looking at the meta-openembedded tree.

src/meson.build
hughsie: So if offline is true and systemd is false do a error() in the toplevel meson.build perhaps?

superm1: Yeah that's what I was thinking

meson_options.txt
hughsie: Fixed.

src/meson.build
hughsie: Done.

data/meson.build
```diff
@@ -26,13 +29,14 @@ if build_standalone
...
 endif
 
-install_data(['org.freedesktop.fwupd.metainfo.xml'],
-  install_dir: join_paths(datadir, 'metainfo')
-)
-
-install_data(['org.freedesktop.fwupd.svg'],
-  install_dir : join_paths(datadir, 'icons', 'hicolor', 'scalable', 'apps')
-)
+if get_option('metainfo')
```
superm1, line 32: we can probably exclude this from windows build too now that this option exists

hughsie: Yes, and FreeBSD perhaps too?

hughsie: No, FreeBSD makes sense to keep. Ignore me.

hughsie: Done

data/meson.build
hughsie: I've added the latter to match the others.

superm1 (dismissed):

PR #3982, opened 2021-11-13
Wip/hughsie/size2

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@djcampello this fixes your CI, right?

meson_options.txt
```diff
@@ -51,3 +51,4 @@ option('metainfo', type: 'boolean', value : true, description : 'install the pro
 option('bash_completion', type: 'boolean', value : true, description : 'enable bash completion')
 option('fish_completion', type: 'boolean', value : true, description : 'enable fish completion')
 option('offline', type: 'boolean', value : true, description : 'enable installing firmware using a pre-boot systemd target')
+option('compat_cli', type: 'boolean', value : true, description : 'enable command line programs provided for compatibility')
```
superm1, line 54: I think this description needs more details

meson.build
```diff
@@ -291,7 +291,7 @@ if valgrind.found()
 endif
 if get_option('offline')
   if not get_option('systemd')
-    error('-Doffline=true requires -Dsystemd=true')
+    warning('-Doffline=true requires -Dsystemd=true')
```
superm1, line 294: Although this fixes chrome CI isn't it more correct to have CI do the adjustment?

hughsie: Ya, it was just a bit antisocial to do this without some kind of a heads up. We could revert this after 1.7.2 drops.

meson_options.txt
hughsie: God any ideas? that was me being verbose :)

superm1: List out the programs it creates or why they're still there

hughsie: What about `enable command line programs provided for legacy compatibility`? I'm struggling a bit here.

superm1: Build legacy applications: $list ?

hughsie: Okay, fixed. Question: is `fwupdtpmevlog` a `compat_cli` program? I've included it here, but it doesn't feel quite right. Maybe we could rename the option to be `extra_binaries` or something? The alternative is that we actually stop installing fwupdtpmevlog and just build it in-src -- as I understood the only real purpose is for debugging the TPM eventlog.

superm1: At some point we should drop the compat binaries. You're right that tool is debugging only the tpm event log.vi think making it noinst is fine.

hughsie: > At some point we should drop the compat binaries

You have no idea the amount of "paperwork" it would be to rebase RHEL and _not_ provide fwupdate... :)

> tool is debugging only

Done.

superm1 (approved):

PR #3993, opened 2021-11-15
altos: Remove support for the ChaosKey

hughsie:
The hardware is discontinued, and no further update from 1.6.7 was ever supplied.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1 (changes requested):
I think you need to change contrib/ci/dependencies.xml for this change too

hughsie:
> I think you need to change contrib/ci/dependencies.xml

I did that, but then the ABI CI job failed. I think we can put it on the list to do after the 1.7.2 release.

superm1:
> > I think you need to change contrib/ci/dependencies.xml
> 
> I did that, but then the ABI CI job failed. I think we can put it on the list to do after the 1.7.2 release.

Ah OK.  I guess as long as it's on the list and not totally forgotten about - maybe a comment in dependencies.xml to drop later?

hughsie:
I've added it to my notes here. Thanks.

PR #3998, opened 2021-11-16
Add HSI check that PCR registers 0-7 are not empty

### LINKED ISSUE #3901: Empty PCR check for HSI

kerneis-anssi:
Some BIOSes forget add measurements to PCR registers, which results in
all-zero checksums and breaks measured boot guarantees.

Fixes #3901

Type of pull request:

- [ ] New plugin
- [ ] Code fix
- [x] Feature
- [ ] Documentation


docs/hsi.md
```diff
@@ -392,6 +392,28 @@ To meet HSI-2 on systems that run this test, the result must be `locked`. *[v1.5
...
+**Impact:** A local attacker could measure fake values into the empty PCR, corresponding to a firmware and OS that do not match the ones actually loaded.
+This allows hiding a compromised boot chain or fooling a remote-attestation server into believing that a different kernel is running.
+
+**Possible results:**
+
+- `valid`: all correct
+- `not-valid`: at least one empty checksum has been found
+- `not-found`: no TPM hardware could be found
+
+To meet HSI-1 on systems that run this test, all PCRs from 0 to 7 in all banks must have non-empty measurements *[v1.8.0]*
```
superm1, line 411: 1.7.2, I don't think we're branching for 1.8.x yet

libfwupd/fwupd-security-attr-private.h
```diff
@@ -221,6 +221,14 @@ G_BEGIN_DECLS
  * Since: 1.5.0
  **/
 #define FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_RAM "org.fwupd.hsi.SuspendToRam"
+/**
+ * FWUPD_SECURITY_ATTR_ID_TPM_EMPTY_PCR:
+ *
+ * Host Security ID attribute for empty PCR
+ *
+ * Since: 1.8.0
```
superm1, line 229: 1.7.2

plugins/tpm/fu-plugin-tpm.c
```diff
@@ -197,11 +197,63 @@ fu_plugin_tpm_add_security_attr_eventlog(FuPlugin *plugin, FuSecurityAttrs *attr
...
+	attr = fwupd_security_attr_new(FWUPD_SECURITY_ATTR_ID_TPM_EMPTY_PCR);
+	fwupd_security_attr_set_plugin(attr, fu_plugin_get_name(plugin));
+	fwupd_security_attr_set_level(attr, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL);
+	fwupd_security_attr_add_guids(attr, fu_device_get_guids(data->tpm_device));
+	fu_security_attrs_append(attrs, attr);
+
+	if (fu_plugin_has_flag(plugin, FWUPD_PLUGIN_FLAG_DISABLED)) {
+		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+		return;
+	}
```
superm1, line 222: Should this be the first check before checking for no TPM device?

plugins/tpm/fu-self-test.c
```diff
@@ -84,15 +88,15 @@ fu_tpm_device_2_0_func(void)
...
+		goto done;
 	}
 #endif
 
 	if (!fu_device_setup(FU_DEVICE(device), &error)) {
 		if (tpm_server_running == NULL &&
 		    g_error_matches(error, FWUPD_ERROR, FWUPD_ERROR_NOT_FOUND)) {
 			g_test_skip("no physical or simulated TPM 2.0 device available");
-			return;
+			goto done;
```
superm1, line 99: we don't use goto in the fwupd project, so just manually add g_unsetenv("FWUPD_FORCE_TPM2"); here.

plugins/tpm/fu-self-test.c
```diff
@@ -102,6 +106,7 @@ fu_tpm_device_2_0_func(void)
 	pcrXs = fu_tpm_device_get_checksums(device, 999);
 	g_assert_nonnull(pcrXs);
 	g_assert_cmpint(pcrXs->len, ==, 0);
+done:
```
superm1, line 109: remove this label

docs/hsi.md
hughsie: Agree

docs/hsi.md
```diff
@@ -392,6 +392,28 @@ To meet HSI-2 on systems that run this test, the result must be `locked`. *[v1.5
...
 - [Intel Direct Connect Interface](https://www.intel.co.uk/content/www/uk/en/support/articles/000029393/processors.html)
 
+<a id="org.fwupd.hsi.Tpm.EmptyPcr"></a>
+
+### [Empty PCR in TPM](#org.fwupd.hsi.Tpm.EmptyPcr)
+
+The system firmware is responsible for measuring values about its boot stage in PCRs 0 through 7.
+Some firmwares have bugs that prevent them from measuring some of those values, breaking the fundamental assumption of the Measured Boot chain-of-trust.
+
+**Impact:** A local attacker could measure fake values into the empty PCR, corresponding to a firmware and OS that do not match the ones actually loaded.
```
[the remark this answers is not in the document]
kerneis-anssi: Not necessarily: in the case of this bug, the TPM event log (for that register) would be empty as well. The bug is the firmware doesn't measure anything when it should. There's no reason to think it would correctly set the eventlog when it's not measuring. (I didn't check with the impacted firmwares though.)

plugins/tpm/fu-plugin-tpm.c
kerneis-anssi: It would make sense to me, but I'm pretty sure I copy-pasted this from existing code by @hughsie.

docs/hsi.md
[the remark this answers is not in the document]
superm1: Got it thanks.

plugins/tpm/fu-self-test.c
```diff
@@ -84,15 +88,15 @@ fu_tpm_device_2_0_func(void)
 	if (tpm_server_running == NULL && (getuid() != 0 || geteuid() != 0)) {
 		g_test_skip("TPM2.0 tests require simulated TPM2.0 running or need root access "
 			    "with physical TPM");
-		return;
+		goto done;
```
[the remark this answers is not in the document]
kerneis-anssi: Will do, I was trying to follow the [single-path cleanup](https://developer.gnome.org/GSettings/documentation/guidelines/programming/memory-management.html#single-path-cleanup) pattern to avoid precisely the kind of bug that occurred here — unfortunately, GLib doesn't seem to have an RAII-like cleanup helper.

docs/hsi.md
kerneis-anssi: Done.

libfwupd/fwupd-security-attr-private.h
kerneis-anssi: Done.

plugins/tpm/fu-plugin-tpm.c
kerneis-anssi: Done (and moved in existing code as well).

plugins/tpm/fu-self-test.c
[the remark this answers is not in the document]
kerneis-anssi: Done.

plugins/tpm/fu-plugin-tpm.c
```diff
@@ -141,6 +141,13 @@ fu_plugin_tpm_add_security_attr_eventlog(FuPlugin *plugin, FuSecurityAttrs *attr
 	g_autoptr(GPtrArray) pcr0s_calc = NULL;
 	g_autoptr(GPtrArray) pcr0s_real = NULL;
 
+	/* check reconstructed to PCR0 */
+	if (fu_plugin_has_flag(plugin, FWUPD_PLUGIN_FLAG_DISABLED) || data->bios_device == NULL ||
+	    data->ev_items == NULL) {
+		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
```
superm1, line 147: it needs to be a little more subtle how this is done - the attribute needs to be made first and then the guids only added if `data->tpm_device` is not `NULL`

kerneis-anssi: I should add a pre-push hook to run the tests… Now fixed.

superm1: > I should add a pre-push hook to run the tests… Now fixed.

That's a good idea.  If you want to add something to `contrib/setup` it would be a good thing to set everyone up for success and higher quality code.

superm1 (approved):
LGTM, @hughsie ?

hughsie (dismissed):
I've fixed this up a little, I hope that's okay.

kerneis-anssi:
> I've fixed this up a little, I hope that's okay.

Thanks @hughsie. For my enlightenment, where is fu_util_security_event_to_string used in the UI? I didn't notice it because things looked correct in my tests.

hughsie:
> where is fu_util_security_event_to_string used in the UI

If you do `fwupdtool security --force` you can see the event log.

PR #4003, opened 2021-11-17
Rename fuzzer source file so that pre-commit picks it up

superm1:
By being .in.c it's getting missed.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-fuzzer-firmware.c
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#include "config.h"
+
+#include "@INCLUDE@"
+
+int
+LLVMFuzzerTestOneInput(const guint8 *data, gsize size)
+{
+	g_autoptr(FuFirmware) firmware = FU_FIRMWARE(@FIRMWARENEW @());
```
superm1, line 14: gah, it looks like this breaks though from style with the space :(

superm1: I guess the right answer is to just fixup the file manually and leave it out of pre-commit then ok.

hughsie: I pushed just that one fixed-up hunk to main. Thanks.

PR #4007, opened 2021-11-18
Run tests before pushing

kerneis-anssi:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

Was suggested by @superm1 when I mentioned it during a code review, but now that I've done it I'm not sure because it adds some significant delay before a push can complete.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+default_stages: [commit]
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.0.1
@@ -53,6 +54,11 @@ repos:
       language: system
       entry: shellcheck --severity=warning -e SC2068
       types: [shell]
+    - id: run-tests
+      name: run tests before pushing
+      language: script
+      entry: ./contrib/run-tests.sh
+      stages: [push]
     - id: clang-format
       name: clang-format
       language: script
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
           ./contrib/setup
           source venv/bin/activate
           sed -i "/no-commit-to-branch/,+1d" .pre-commit-config.yaml
-          pre-commit run --all-files
+          pre-commit run --hook-stage commit --all-files
   abi:
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -53,6 +54,12 @@ repos:
       language: system
       entry: shellcheck --severity=warning -e SC2068
       types: [shell]
+    - id: run-tests
+      name: run tests before pushing
+      language: system
+      entry: ninja -C build test
```
superm1, line 60: I think this also means needing to run "meson build" first no?

I also worry about the default configuration not working for some environment that might be pushing.

superm1 (commented):
Haven't tried yet but how much time does this seem to add?

It's possible to skip checks from pre-commit, so small obvious things we could be setting that environment variable.

.pre-commit-config.yaml
kerneis-anssi: > Haven't tried yet but how much time does this seem to add?

Depends of your state. From a clean repo, on my machine, meson takes 4 seconds and ninja 53 seconds. On an already built repo where only the tests need to execute, 9 seconds.

> It's possible to skip checks from pre-commit, so small obvious things we could be setting that environment variable.

My main issue is that there's no visual feedback while the tests are running: there is the "pre-commit: run tests" line, but no sense of progress of the build+test process itself.

kerneis-anssi: > I think this also means needing to run "meson build" first no?

Done.

> I also worry about the default configuration not working for some environment that might be pushing.

Done.

superm1: > My main issue is that there's no visual feedback while the tests are running: there is the "pre-commit: run tests" line, but no sense of progress of the build+test process itself.

The lack of visual feedback is "generally" too problematic IMO.  I think if we add this is should be a "default" N in `contrib/setup` and should clearly disclaim that it will slow down pushes.

kerneis-anssi: The latest version defaults to N, says so explicitly, has a detailed explanation of consequences, and shouldn't impact the CI at all (unless I missed something). I'm also happy to drop this PR if you don't think it's necessary.

hughsie:
@superm1 I think asking as part of setup is a good idea. Fine for me if it works for you.

superm1 (approved):

Issue #4014, opened 2021-11-22
`fwupd-self-test` failed if no `fwupd` has been previously installed


piyueh:
**Describe the bug**

When running the tests, `fwupd-self-test` failed if no `fwupd` has been previously installed, i.e., it failed due to not being able to find running `fwupd` service. I believe this is the same as #64. The error message is:

```
Fwupd:ERROR:../fwupd-git/libfwupd/fwupd-self-test.c:597:fwupd_client_remotes_func: assertion failed (error == NULL): The name org.freedesktop.fwupd was not provided by any .service files (FwupdError, 10)
Bail out! Fwupd:ERROR:../fwupd-git/libfwupd/fwupd-self-test.c:597:fwupd_client_remotes_func: assertion failed (error == NULL): The name org.freedesktop.fwupd was not provided by any .service files (FwupdError, 10)
```

I tried to add the following code snippet to before [line 543](https://github.com/fwupd/fwupd/blob/c92f352393f534b34d0632290dcf65df8ac7bfa8/libfwupd/fwupd-self-test.c#L543) and [line 597](https://github.com/fwupd/fwupd/blob/c92f352393f534b34d0632290dcf65df8ac7bfa8/libfwupd/fwupd-self-test.c#L597), and the error was gone.
```c
if (array == NULL && g_error_matches(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED)) {
    g_test_skip("no available fwupd daemon");
    return;
}
```
But I'm not sure my workaround was correct or not.

**Steps to Reproduce**

- step 1: configuration
  ```
  meson setup ${source_dir} ${build_dir} \
      --prefix /usr \
      --libexecdir lib \
      --sbindir bin \
      --buildtype plain \
      --auto-features enabled \
      --wrap-mode nodownload \
      -D b_pie=true \
      -D plugin_flashrom=true \
      -D plugin_modem_manager=true \
      -D plugin_intel_spi=true \
      -D b_lto=false \
      -D lzma=true \
      -D docs=gtkdoc \
      -D supported_build=true \
      -D efi_binary=false
  ```

- step 2: build
  ```
  meson compile -C ${build_dir} -j 0
  ```

- step 3: test
  ```
  CACHE_DIRECTORY=/tmp meson test -C ${build_dir}
  ```

**Expected behavior**

All tests pass.

**fwupd version information**

[...]

PR #4015, opened 2021-11-22
Fix running the tests when fwupd is not installed

hughsie:
This probably regressed in 70f9124545e35c1e55915594722f5aa8c7b7edcc as
the callers actually expect an error in the GDBusError domain, rather
than one fixed up in the FwupdError domain.

Fixes https://github.com/fwupd/fwupd/issues/4014

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Isnt this supposed to be || not |?

hughsie:
Good catch!

superm1 (approved):

PR #4018, opened 2021-11-22
trivial: Only install files that are going to be used

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
This doesn't seem sensible to me - shouldn't compilation not use GPG or pkcs7 if they're hot to be used?

PR #4032, opened 2021-11-25
Add fu_common_sum8() common functionality

hughsie:
We now have 9 different plugins all using this functionality, and we're
about to add one more. Move this into common code so that all the
plugins are using the same endian and bufsz-safe versions.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


plugins/ccgx/fu-ccgx-firmware.c
```diff
@@ -238,7 +227,7 @@ fu_ccgx_firmware_parse_md_block(FuCcgxFirmware *self, FwupdInstallFlags flags, G
 	}
 	for (guint i = 0; i < self->records->len - 1; i++) {
 		rcd = g_ptr_array_index(self->records, i);
-		checksum_calc += fu_ccgx_firmware_record_calc_checksum(rcd);
+		checksum_calc += fu_common_sum16_bytes(rcd->data);
```
d4s, line 230: `fu_ccgx_firmware_record_calc_checksum()` returns **guint8**

plugins/wacom-usb/fu-wac-device.c
```diff
@@ -583,7 +583,7 @@ fu_wac_device_write_firmware(FuDevice *device,
 		}
 
 		/* calculate expected checksum and save to device RAM */
-		csum_local[i] = fu_wac_calculate_checksum32le_bytes(blob_block);
+		csum_local[i] = fu_common_sum32w_bytes(blob_block, G_LITTLE_ENDIAN);
```
d4s, line 586: `fu_wac_calculate_checksum32le_bytes()` returns guint32 as G_LITTLE_ENDIAN while new API is a host-based.

d4s (changes requested):

d4s:
Tested with SteelSeries device only -- works well.

plugins/ccgx/fu-ccgx-firmware.c
hughsie: Fixed, thanks; Good catch!

plugins/wacom-usb/fu-wac-device.c
hughsie: Yes, agreed!

    - 		csum_local[i] = fu_common_sum32w_bytes(blob_block, G_LITTLE_ENDIAN); 
    +		csum_local[i] = GUINT32_TO_LE(fu_common_sum32w_bytes(blob_block, G_LITTLE_ENDIAN));

right?

d4s (approved):

PR #4033, opened 2021-11-25
Wip/hardboprobot/nordic

d4s:
Type of pull request:

- [x ] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

- [ x] Fill out `README.md` with update protocol
- [ x] N/A: ~~Fill out `README.md` with any custom quirks and flags~~ 
- [ x] Fill out `README.md` with the vendor ID security value
- [ x] N/A: ~~Implement `FuFirmware->write()` and include at least one fuzzer testcase in `src/fuzzing/firmware` for any custom `FuFirmware` subclass~~


plugins/nordic-hid/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+* "Nordic HID Config Channel"
+
+## GUID Generation
+
+For GUID generation the standard HIDRAW DeviceInstanceId values are used
+with the flash bank ID addition:
+
+* `VEN_1915&DEV_52DE&BANK_0`
+* `VEN_1915&DEV_52DE&BANK_1`
```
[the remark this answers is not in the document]
d4s: Banks for A/B, however the format of FW has been changed and nowadays the bank is not the part of GUID.

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -0,0 +1,1067 @@
...
+fu_nordic_hid_cfg_channel_receive_cb(FuDevice *device, gpointer user_data, GError **error)
+{
+	FuNordicCfgChannelRcvHelper *args = (FuNordicCfgChannelRcvHelper *)user_data;
+	FuNordicDeviceCfgChannel *self = FU_NORDIC_HID_CFG_CHANNEL(device);
+	FuNordicCfgChannelMsg *recv_msg = NULL;
+
+	if (!fu_nordic_hid_cfg_channel_receive(self, args->buf, args->bufsz, error))
+		return FALSE;
+	recv_msg = (FuNordicCfgChannelMsg *)args->buf;
+	if (args->check_status == TRUE && recv_msg->status != args->status) {
```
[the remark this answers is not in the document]
d4s: Ha! good catch... there is no any call where `check_status` is set to `FALSE`, so unneeded at all.
Thanks for noticing that!

plugins/nordic-hid/fu-nordic-hid-archive.c
```diff
@@ -0,0 +1,199 @@
...
+	BL_B0,
+	BL_MCUBOOT,
+} FuNordicHidBootloader;
+
+typedef struct {
+	FuNordicHidBootloader bootloader;
+	gchar *target_board;
+	gchar *img_name;
+	guint img_size;
+} FuNordicHidArchiveImage;
```
hughsie, line 26: Unused, no?

plugins/nordic-hid/fu-nordic-hid-archive.c
```diff
@@ -0,0 +1,199 @@
...
+
+struct _FuNordicHidArchive {
+	FuFirmwareClass parent_instance;
+	GPtrArray *images;
```
hughsie, line 30: unused?

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -0,0 +1,1054 @@
...
+	/* select correct firmware per target board, bootloader and bank */
+	image_id =
+	    g_strdup_printf("%s_%s_bank%01u", self->board_name, self->bl_name, self->flash_area_id);
+	firmware = fu_firmware_get_image_by_id(firmware, image_id, error);
+	if (firmware == NULL)
+		return FALSE;
+
+	checksum = g_ascii_strtoull(fu_firmware_get_checksum(firmware, 0, error), NULL, 16);
+	if (checksum == 0 && error != NULL)
+		return FALSE;
```
hughsie, line 979: This looks fishy. Maybe get the checksum as a `const gchar *csum` and then return FALSE if that fails. Then check for checksum=0 and do a `g_set_error()` in that case.

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -0,0 +1,1054 @@
...
+			return FALSE;
+		}
+		if (res->data[0] == expecting_state)
+			break;
+
+		if (tries-- == 0) {
+			g_set_error_literal(error,
+					    G_IO_ERROR,
+					    G_IO_ERROR_NOT_SUPPORTED,
+					    "too lot of sync attempts");
```
hughsie, line 697: `too many sync attempts` perhaps -- and I do wonder if fu_device_retry_full might be better to use here.

plugins/nordic-hid/fu-nordic-hid-archive.c
d4s: Thanks!!!

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
d4s: ok, reworked a bit. However crc32 might be zero, afair, so have no sense to check it against 0.
Instead I set the '-1' for requested checksum kind, hence checksum will work only for the custom calculation function and added error handling to firmware's implementation.

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -0,0 +1,1067 @@
...
+				  gsize bufsz,
+				  GError **error)
+{
+#ifdef HAVE_HIDRAW_H
+	if (!fu_udev_device_ioctl(FU_UDEV_DEVICE(self), HIDIOCGFEATURE(bufsz), buf, NULL, error))
+		return FALSE;
+	if (g_getenv(FWUPD_NORDIC_HID_VERBOSE) != NULL)
+		fu_common_dump_raw(G_LOG_DOMAIN, "Received", buf, bufsz);
+	/*
+	 * [TODO]: Possibly add the report-id fix for Bluez versions < 5.56:
```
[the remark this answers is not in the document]
d4s: @hughsie do you think we need to support Bluez versions <5.56?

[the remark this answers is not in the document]
hughsie: Upstream maintainer hat on, "no", but it depends on whether nordic wants this backported to earlier versions. Personally, I'd KISS.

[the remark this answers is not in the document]
d4s: Ok, I'll leave it as is for now then. In case if we need to support old version -- we know the place to fix.

d4s:
This is the initial version supporting/tested with Nordic nrf52840dk board.
- configuration channel is used for updates
- B0 bootloader type is tested

Will squash before merge

hughsie:
I think rebase against master and squash into one commit. Then I'll do one more round.

hughsie:
@superm1 this LGTM, but can you do a quick pass with fresh eyes please?

hughsie (approved):

superm1 (approved):

PR #4038, opened 2021-11-27
Make the HSI tests optional for embedded targets

hughsie:
70% of the plugins installed on OpenBMC are not being used...

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


contrib/build-openbmc.sh
```diff
@@ -0,0 +1,48 @@
...
+    -Dplugin_tpm=false \
+    -Dplugin_uefi_capsule=false \
+    -Dplugin_upower=false \
+    -Dplugin_upower=false \
+    -Dpolkit=false \
+    -Dsqlite=false \
+    -Dtests=false \
+    -Dudevdir=/tmp \
+    -Dsystemd_root_prefix=/tmp \
+    --prefix=/home/hughsie/.root
```
superm1, line 48: Probably not best to hardcode this

plugins/platform-integrity/meson.build
```diff
@@ -1,4 +1,4 @@
-if get_option('plugin_platform_integrity')
```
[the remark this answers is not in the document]
hughsie: Ahh yes, but not for that reason; platform_integrity got NAKd by LKML, so that still needs to be disabled by default.

contrib/build-openbmc.sh
hughsie: Oop, agree. I've switched it to use $@ like the other scripts.

PR #4042, opened 2021-11-29
usi-dock: Add support for various TBT4 docks for Lenovo and HP

hughsie:
With much help from Victor Cheng <victor_cheng@usiglobal.com>

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


plugins/usi-dock/README.md
```diff
@@ -0,0 +1,37 @@
+# USI Dock
+
+## Introduction
+
+This plugin uses the MCU to write all the dock firmware components. Interestingly, the MCU version
```
superm1, line 5: Drop interestingly

plugins/usi-dock/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+These devices use the standard USB DeviceInstanceId values, e.g.
+
+* `USB\VID_17EF&PID_7226&REV_0001`
+* `USB\VID_17EF&PID_7226`
+* `USB\VID_17EF`
+
+Additionally, some extra "component ID" instance IDs are added.
+
+* `USB\VID_17EF&PID_7226&CID_TBT4`
```
superm1, line 22: I think it should be providing exact same component ID for TBT NVM so that thunderbolt plugin or this one can handle it (but set priorities accordingly for preferred one)

plugins/usi-dock/fu-usi-dock-dmc-device.c
```diff
@@ -0,0 +1,55 @@
...
+	FuUsbDevice parent_instance;
+};
+
+G_DEFINE_TYPE(FuUsiDockDmcDevice, fu_usi_dock_dmc_device, FU_TYPE_USB_DEVICE)
+
+static void
+fu_usi_dock_dmc_device_parent_notify_cb(FuDevice *device, GParamSpec *pspec, gpointer user_data)
+{
+	FuDevice *parent = fu_device_get_parent(device);
+	if (parent != NULL) {
```
superm1, line 21: Seems like this should be a critical failure no?

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -0,0 +1,777 @@
...
+}
+
+static void
+fu_usi_dock_mcu_device_init(FuUsiDockMcuDevice *self)
+{
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_DUAL_IMAGE);
+
+	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_NO_SERIAL_NUMBER);
```
superm1, line 753: really?  is there no way to get this from the dock?

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -0,0 +1,777 @@
...
+	switch (cmd) {
+	case USBUID_ISP_DEVICE_CMD_MCU_JUMP2BOOT:
+		g_debug("got correct jump");
+		break;
+	case USBUID_ISP_INTERNAL_FW_CMD_TARGET_CHECKSUM:
+		if (buf[6] == FIRMWARE_IDX_AUDIO) {
+			g_debug("got a quick jump at audio updates");
+		} else {
+			g_debug("got a ignored report");
+			return FALSE;
```
superm1, line 48: are these pretty common?  If so, then maybe this should go down into plugin specific debug guard.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -0,0 +1,777 @@
...
+}
+
+static gboolean
+fu_usi_dock_mcu_device_tx(FuUsiDockMcuDevice *self,
+			  guint8 tag2,
+			  const guint8 *inbuf,
+			  gsize inbufsz,
+			  GError **error)
+{
+	guint8 buf[64] = {USB_HID_REPORT_ID2,
```
superm1, line 65: I wonder if this buffer should be a struct that you define?  It seems cmd is always the first, and then it's a certain number of args.  Not sure if those are variable behaviors per command or same for all.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -0,0 +1,777 @@
...
+{
+	if (!fu_usi_dock_mcu_device_tx(self, tag2, inbuf, inbufsz, error))
+		return FALSE;
+	return fu_usi_dock_mcu_device_rx(self, USBUID_ISP_CMD_ALL, outbuf, outbufsz, error);
+}
+
+static gboolean
+fu_usi_dock_mcu_device_get_status(FuUsiDockMcuDevice *self, GError **error)
+{
+	guint8 buf[] = {USBUID_ISP_DEVICE_CMD_MCU_STATUS};
```
superm1, line 175: buf is normally 64 bytes, why isn't this one?  Should it be set that way for safety?

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -0,0 +1,777 @@
...
+			fu_device_set_version(child, version);
+			fu_device_add_icon(child, "thunderbolt");
+			fu_device_set_name(child, "Thunderbolt 4 Controller");
+		} else if (g_strcmp0(components[i].name, "DP5x") == 0) {
+			if ((val[2] == 0x00 && val[3] == 0x00 && val[4] == 0x00) ||
+			    (val[2] == 0xFF && val[3] == 0xFF && val[4] == 0xFF)) {
+				g_debug("ignoring %s", components[i].name);
+				continue;
+			}
+			version = g_strdup_printf("%d.%02d.%03d", val[2], val[3], val[4]);
```
superm1, line 310: this device doesn't have a HP version format like the rest?

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -0,0 +1,777 @@
...
+			fu_device_set_version(child, version);
+			fu_device_set_name(child, "Power Delivery");
+		} else if (g_strcmp0(components[i].name, "TBT4") == 0) {
+			if ((val[1] == 0x00 && val[2] == 0x00 && val[3] == 0x00) ||
+			    (val[1] == 0xFF && val[2] == 0xFF && val[3] == 0xFF)) {
+				g_debug("ignoring %s", components[i].name);
+				continue;
+			}
+			version = g_strdup_printf("%02x.%02x.%02x", val[1], val[2], val[3]);
+			fu_device_set_version_format(child, FWUPD_VERSION_FORMAT_TRIPLET);
```
superm1, line 300: really?  This isn't a pair?  I've never seen a thunderbolt device with a triplet.

plugins/usi-dock/fu-usi-dock-mcu-device.c
victor-cheng: #4062

plugins/usi-dock/README.md
victor-cheng: This plugin handles MCU instead of TBT devices. The component ID is used for display only.

plugins/usi-dock/fu-usi-dock-mcu-device.c
victor-cheng: Only a few components are needed on the HP G2 project so far.

plugins/usi-dock/fu-usi-dock-mcu-device.c
victor-cheng: Internal defined for display only.

plugins/usi-dock/README.md
superm1: If this plugin is supposed to entirely handle TBT - my opinion is the TBT plugin needs to be disabled for any devices that are assumed to be controlled here when they're registered.  Perhaps model it off of this:
https://github.com/fwupd/fwupd/blob/main/plugins/dell-dock/fu-plugin-dell-dock.c#L221

plugins/usi-dock/fu-usi-dock-dmc-device.c
victor-cheng: @hughsie could you help to explain this?

hughsie: Ohh I don't think it has to be critical; in the future we might have a parent-less DMC device.

plugins/usi-dock/fu-plugin-usi-dock.c
```diff
@@ -0,0 +1,44 @@
...
+
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include "fu-usi-dock-dmc-device.h"
+#include "fu-usi-dock-firmware.h"
+#include "fu-usi-dock-mcu-device.h"
+
+#define USI_DOCK_TBT_INSTANCE_ID "THUNDERBOLT\\VEN_0108&DEV_2031"
```
superm1, line 15: just to confirm since this used plural of docks - is this the only TBT ID handled by the plugin internally for now?  Need to make sure that these stay in sync as more docks are released.

victor-cheng: for the present, yes.

victor-cheng (dismissed):

PR #4046, opened 2021-12-01
thunderbolt: set ports offline on host controller

krankun:
This change offline ports in the host controller
for retimers to enumerate in the NDA case.

-offline and rescan usb4 ports
-this enable enumeration of the retimers
-updates nvm to the enumerated retimers
-online usb4 port this will de-enumerate retimers

BUG=b:187506425
TEST=emerge-volteer fwupd

Change-Id: I1fde5675f27e91e18951b72f24fbec4113410807

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -230,6 +230,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "use-parent-for-battery";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK)
 		return "use-proxy-fallback";
+	if (flag == FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)
+		return "no-auto-remove";
 	return NULL;
 }
 
@@ -284,6 +286,8 @@ fu_device_internal_flag_from_string(const gchar *flag)
 		return FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_BATTERY;
 	if (g_strcmp0(flag, "use-proxy-fallback") == 0)
 		return FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK;
+	if (g_strcmp0(flag, "no-auto-remove") == 0)
+		return FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE;
 	return FU_DEVICE_INTERNAL_FLAG_UNKNOWN;
 }
 
--- a/libfwupdplugin/fu-device.h
+++ b/libfwupdplugin/fu-device.h
@@ -400,6 +400,14 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK (1ull << 18)
 
+/**
+ * FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE :
+ *
+ * The device is not auto removed.
+ * Since 1.7.3
+ */
+#define FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE (1llu << 19)
+
 /* accessors */
 gchar *
 fu_device_to_string(FuDevice *self);
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -67,6 +67,32 @@ fu_plugin_thunderbolt_startup(FuPlugin *plugin, GError **error)
 	return fu_plugin_thunderbolt_safe_kernel(plugin, error);
 }
 
+static gboolean
+fu_plugin_thunderbolt_composite_prepare(FuPlugin *plugin, GPtrArray *devices, GError **error)
+{
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *dev = g_ptr_array_index(devices, i);
+		if ((g_strcmp0(fu_device_get_plugin(dev), "thunderbolt") == 0) &&
+		    fu_device_has_flag(dev, FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)) {
+			return fu_thunderbolt_probe_retimer(dev, error);
+		}
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_plugin_thunderbolt_composite_cleanup(FuPlugin *plugin, GPtrArray *devices, GError **error)
+{
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *dev = g_ptr_array_index(devices, i);
+		if ((g_strcmp0(fu_device_get_plugin(dev), "thunderbolt") == 0) &&
+		    fu_device_has_flag(dev, FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)) {
+			return fu_thunderbolt_device_close(dev, error);
+		}
+	}
+	return TRUE;
+}
+
 void
 fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
 {
@@ -75,4 +101,6 @@ fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
 	vfuncs->startup = fu_plugin_thunderbolt_startup;
 	vfuncs->device_registered = fu_plugin_thunderbolt_device_registered;
 	vfuncs->device_created = fu_plugin_thunderbolt_device_created;
+	vfuncs->composite_prepare = fu_plugin_thunderbolt_composite_prepare;
+	vfuncs->composite_cleanup = fu_plugin_thunderbolt_composite_cleanup;
 }
--- a/plugins/thunderbolt/fu-self-test.c
+++ b/plugins/thunderbolt/fu-self-test.c
@@ -61,6 +61,25 @@ udev_mock_add_nvmem(UMockdevTestbed *bed, gboolean active, const char *parent, i
 	return path;
 }
 
+static gchar *
+udev_mock_add_usb4_port(UMockdevTestbed *bed, int id)
+{
+	g_autofree gchar *name = g_strdup_printf("usb4_port%d", id);
+	gchar *path = umockdev_testbed_add_device(bed,
+						  "thunderbolt",
+						  name,
+						  NULL,
+						  "security",
+						  "secure",
+						  NULL,
+						  "DEVTYPE",
+						  "thunderbolt_usb4_port",
+						  NULL);
+
+	g_assert_nonnull(path);
+	return path;
+}
+
 typedef struct MockDevice MockDevice;
 
 struct MockDevice {
@@ -561,6 +580,7 @@ mock_tree_attach(MockTree *root, UMockdevTestbed *bed, FuPlugin *plugin)
 {
 	root->bed = g_object_ref(bed);
 	root->sysfs_parent = udev_mock_add_domain(bed, root->device->domain_id);
+	root->sysfs_parent = udev_mock_add_usb4_port(bed, 1);
 	g_assert_nonnull(root->sysfs_parent);
 
 	g_timeout_add(root->device->delay_ms, mock_tree_attach_device, root);
--- a/plugins/thunderbolt/fu-thunderbolt-device.c
+++ b/plugins/thunderbolt/fu-thunderbolt-device.c
@@ -319,6 +319,138 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 	}
 	return (guint16)val;
 }
+static gboolean
+fu_thunderbolt_check_usb4_port_path(FuUdevDevice *device, const gchar *attribute, GError **err)
+{
+	g_autofree const gchar *path =
+	    g_build_filename(fu_udev_device_get_sysfs_path(device), attribute, NULL);
+	g_autofree gchar *fn = g_strdup_printf("%s", path);
+	g_autoptr(GFile) file = g_file_new_for_path(fn);
+	if (!g_file_query_exists(file, NULL)) {
+		g_set_error(err, G_IO_ERROR, G_IO_ERROR_NOT_FOUND, "failed to find %s", fn);
+		return FALSE;
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device, GError **error)
+{
+	const gchar *offline = "usb4_port1/offline";
+	const gchar *rescan = "usb4_port1/rescan";
+	g_autoptr(GError) error_offline = NULL;
+	g_autoptr(GError) error_rescan = NULL;
+	if (fu_thunderbolt_check_usb4_port_path(device, offline, &error_offline)) {
+		if (!fu_udev_device_write_sysfs(device, offline, "1", error)) {
+			g_set_error_literal(error,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "Setting usb4 port offline failed");
+			return FALSE;
+		}
+	} else {
+		g_debug("%s", error_offline->message);
+		return TRUE;
+	}
+	if (fu_thunderbolt_check_usb4_port_path(device, rescan, &error_rescan)) {
+		if (!fu_udev_device_write_sysfs(device, rescan, "1", error)) {
+			g_set_error_literal(error,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "Rescan on port failed");
+			return FALSE;
+		}
+	} else {
+		g_debug("%s", error_rescan->message);
+		return TRUE;
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_online(FuUdevDevice *device, GError **err)
+{
+	FuUdevDevice *udev = FU_UDEV_DEVICE(device);
+	const gchar *offline = "usb4_port1/offline";
+	g_autoptr(GError) error_offline = NULL;
+
+	if (fu_thunderbolt_check_usb4_port_path(device, offline, &error_offline)) {
+		if (!fu_udev_device_write_sysfs(udev, offline, "0", err)) {
+			g_set_error_literal(err,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "Setting port online failed");
+			return FALSE;
+		}
+	}
+	return TRUE;
+}
+
+static FuUdevDevice *
+fu_thunderbolt_get_udev_parent_device(FuDevice *device, GError **error)
+{
+	g_autoptr(GUdevDevice) udev_parent1 = NULL;
+	g_autoptr(GUdevDevice) udev_parent2 = NULL;
+	GUdevDevice *udev_device = NULL;
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "not a retimer device");
+		return NULL;
+	}
+	udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+	if (udev_device == NULL) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "failed to get udev device for retimer");
+		return NULL;
+	}
+	udev_parent1 = g_udev_device_get_parent(udev_device);
+	if (udev_parent1 == NULL) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "failed to get parent device for retimer");
+		return NULL;
+	}
+	udev_parent2 = g_udev_device_get_parent(udev_parent1);
+	if (udev_parent2 == NULL) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "failed to get host router device for retimer");
+		return NULL;
+	}
+	return fu_udev_device_new_with_context(fu_device_get_context(FU_DEVICE(self)),
+					       g_steal_pointer(&udev_parent2));
+}
+gboolean
+fu_thunderbolt_probe_retimer(FuDevice *device, GError **error)
+{
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+	g_autoptr(FuUdevDevice) parent = NULL;
+	if (device == NULL || self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "fail to probe retimer device");
+		return FALSE;
+	}
+	parent = fu_thunderbolt_get_udev_parent_device(device, error);
+	return fu_thunderbolt_device_set_port_offline(parent, error);
+}
+
+gboolean
+fu_thunderbolt_device_close(FuDevice *device, GError **error)
+{
+	g_autoptr(FuUdevDevice) parent = fu_thunderbolt_get_udev_parent_device(device, error);
+	if (parent == NULL)
+		return FALSE;
+	return fu_thunderbolt_device_set_port_online(parent, error);
+}
 
 static gboolean
 fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
@@ -422,6 +554,11 @@ fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 		fu_device_add_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	}
 
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER) {
+		if (!fu_thunderbolt_device_set_port_offline(FU_UDEV_DEVICE(device), error))
+			return FALSE;
+	}
+
 	return TRUE;
 }
 
@@ -442,6 +579,8 @@ fu_thunderbolt_device_setup_retimer(FuDevice *device, GError **error)
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_DUAL_IMAGE);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_INTERNAL);
+	fu_device_add_flag(device, FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE);
+
 	vid = fu_udev_device_get_vendor(FU_UDEV_DEVICE(self));
 	if (vid == 0x0) {
 		g_set_error_literal(error,
@@ -763,10 +902,11 @@ fu_thunderbolt_device_write_firmware(FuDevice *device,
 	}
 
 	/* whether to wait for a device replug or not */
-	if (!fu_device_has_flag(device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
-		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_RESTART);
+	if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
 		fu_device_set_remove_delay(device, FU_PLUGIN_THUNDERBOLT_UPDATE_TIMEOUT);
-		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_RESTART);
+		if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER)
+			fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
 	}
 
 	return TRUE;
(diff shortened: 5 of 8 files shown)


libfwupd/fwupd-enums.h
```diff
@@ -493,6 +493,15 @@ typedef enum {
...
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
+/**
+ *
+ * FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE :
+ *
+ * The device is not auto removed.
+ * Since 1.7.3
+ */
+#define FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE (1llu << 46)
+
```
superm1, line 504: @hughsie should these be internal flags perhaps?  I'm thinking yes.
@krankun IOW define them in `libfwupdplugin/fu-device.h`

libfwupd/fwupd-enums.c
```diff
@@ -312,6 +314,8 @@ fwupd_device_flag_from_string(const gchar *device_flag)
 		return FWUPD_DEVICE_FLAG_UNREACHABLE;
 	if (g_strcmp0(device_flag, "affects-fde") == 0)
 		return FWUPD_DEVICE_FLAG_AFFECTS_FDE;
+	if (g_strcmp0(device_flag, "no-auto-remove") == 0)
+		return FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE;
```
superm1, line 318: if these end up being internal flags, then this should move to `libfwupdplugin/fu-device.c`

libfwupd/fwupd-enums.h
hughsie: Yes, I think internal too -- they're of no use to the end user.

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -76,3 +76,32 @@ fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
...
+		}
+	}
+	return TRUE;
+}
+
+
+gboolean
+fu_plugin_composite_cleanup	(FuPlugin	*plugin,
+							 GPtrArray	*devices,
+							 GError		**error)
```
superm1, line 98: whitespace seems funny here, did clang do this?  It should be more like the composite_prepare function.

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -76,3 +76,32 @@ fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
...
+{
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *device = g_ptr_array_index (devices, i);
+		if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
```
superm1, line 102: I think you also need to check that the device comes from the thunderbolt plugin.  Composite prepare/composite cleanup could potentially have devices from other plugins.

plugins/thunderbolt/fu-self-test.c
```diff
@@ -61,6 +61,28 @@ udev_mock_add_nvmem(UMockdevTestbed *bed, gboolean active, const char *parent, i
 	return path;
 }
 
+static gchar *
+udev_mock_add_usb4_port(UMockdevTestbed *bed, int id)
+{
+	gchar *path;
+	g_autofree gchar *name = NULL;
```
superm1, line 68: reverse xmas tree

plugins/thunderbolt/fu-self-test.c
```diff
@@ -61,6 +61,28 @@ udev_mock_add_nvmem(UMockdevTestbed *bed, gboolean active, const char *parent, i
...
+
+	name = g_strdup_printf("usb4_port%d", id);
```
superm1, line 70: nit: just combine these to one line

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 	return (guint16)val;
 }
 
+static gboolean
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
```
superm1, line 324: you should be passing in the GError **error as an argument here and then using that in this function rather than local errors.  The idea is supposed to be that if you set FALSE, then you need to return something in error.

The way this function is working right now, you're setting all those errors, but you're discarding them as they're local to the function.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+{
+	GError **error_local_offline = NULL;
+	GError **error_local_rescan = NULL;
+
```
superm1, line 328: drop these, and just pass a single `**error` into the function.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+				    FWUPD_ERROR_INTERNAL,
+				    "Rescan on port failed");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static void
+fu_thunderbolt_device_set_port_online(FuUdevDevice *device)
```
superm1, line 348: pass an error into the function like the  `set_port_offline` one.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	if (udev_parent1 == NULL)
+		return NULL;
+	udev_parent2 = g_udev_device_get_parent(udev_parent1);
+	return fu_udev_device_new_with_context(fu_device_get_context(FU_DEVICE(self)),
+					       g_steal_pointer(&udev_parent2));
+}
+gboolean
+fu_thunderbolt_device_open(FuDevice *device, GError **error)
+{
+	g_autoptr(FuUdevDevice) parent = get_udev_parent_device(device);
```
superm1, line 383: you need to be guarding for the NULL here and return FALSE in that case

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+
+static FuUdevDevice *
+get_udev_parent_device(FuDevice *device)
+{
+	g_autoptr(GUdevDevice) udev_parent1 = NULL;
+	g_autoptr(GUdevDevice) udev_parent2 = NULL;
+	GUdevDevice *udev_device = NULL;
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
+	if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER)
+		return NULL;
```
superm1, line 369: error should be an argument to the futnction function and you need to set it if this was false

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	udev_parent1 = g_udev_device_get_parent(udev_device);
+	if (udev_parent1 == NULL)
+		return NULL;
+	udev_parent2 = g_udev_device_get_parent(udev_parent1);
+	return fu_udev_device_new_with_context(fu_device_get_context(FU_DEVICE(self)),
+					       g_steal_pointer(&udev_parent2));
+}
+gboolean
+fu_thunderbolt_device_open(FuDevice *device, GError **error)
```
superm1, line 381: I'd suggest to rename it to `fu_thunderbolt_retimer_open` as this is only used in the retimer case

plugins/thunderbolt/fu-thunderbolt-device.c
superm1, line 383: I think maybe an assertion makes sense here if this is not a retimer device then there was a programmer error

libfwupd/fwupd-enums.h
superm1: Oh but they're used in `src/fu-udev-backend.c`, so maybe not.

src/fu-udev-backend.c
```diff
@@ -41,9 +41,11 @@ fu_udev_backend_device_remove(FuUdevBackend *self, GUdevDevice *udev_device)
 	/* find the device we enumerated */
 	device_tmp =
 	    fu_backend_lookup_by_id(FU_BACKEND(self), g_udev_device_get_sysfs_path(udev_device));
-	if (device_tmp != NULL) {
-		if (g_getenv("FWUPD_PROBE_VERBOSE") != NULL) {
-			g_debug("UDEV %s removed", g_udev_device_get_sysfs_path(udev_device));
+	if (device_tmp != NULL &&
+	    fu_device_has_flag(device_tmp, FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)) {
```
superm1, line 45: shouldn't this be `!fu_device_has_flag(..)`?

superm1 (changes requested):
pre-commit continues to fail and various nested comments.
No need to open a new PR each time.  You can just `git push --force`

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -76,3 +76,32 @@ fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
 	vfuncs->device_registered = fu_plugin_thunderbolt_device_registered;
 	vfuncs->device_created = fu_plugin_thunderbolt_device_created;
 }
+
+gboolean
+fu_plugin_composite_prepare (FuPlugin *plugin,
```
superm1, line 81: On main the way this works, you need to declare the vfuncs for composite prepare and cleanup as `static` and then declare what they are in `fu_plugin_init_vfuncs`.  You can look at `plugins/dell-dock/fu-plugin-dell-dock.c` for an example of how this is supposed to be done.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,78 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 	return (guint16)val;
 }
 
+static gboolean
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device)
+{
+	GError **error_local_offline = NULL;
+	GError **error_local_rescan = NULL;
+
+	if (!fu_udev_device_write_sysfs(device, "usb4_port1/offline", "1", error_local_offline)) {
```
djcampello, line 329: Are we hard coding usb4_port1?

krankun: @djcampello  Yes currently there is just one port to update nvm and the other port shares the same so no need to update it , in later platforms EC needs to populate only the retimers which don't share the same NVM in that way we would grep the available usb4_portx and update nvm for all of them

src/fu-udev-backend.c
krankun: I have corrected it and got it functional with the new change but will keep this open until I debug and understand the implications if any

krankun:
> pre-commit continues to fail and various nested comments. No need to open a new PR each time. You can just `git push --force`

lets see if the new force push fixes it

libfwupd/fwupd-enums.c
krankun: let me look at these , still open

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -75,4 +101,6 @@ fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
 	vfuncs->startup = fu_plugin_thunderbolt_startup;
 	vfuncs->device_registered = fu_plugin_thunderbolt_device_registered;
 	vfuncs->device_created = fu_plugin_thunderbolt_device_created;
+	vfuncs->composite_prepare = fu_plugin_composite_prepare;
+	vfuncs->composite_cleanup = fu_plugin_composite_cleanup;
```
superm1, line 105: you should include the word `thunderbolt` in these like other vfuncs to avoid potential symbol collisions when loaded into the daemon.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,105 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "Rescan on port failed");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static void
```
superm1, line 344: this should be type gboolean (and return TRUE/FALSE)

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,105 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	}
+	parent = get_udev_parent_device(device, error);
+	return fu_thunderbolt_device_set_port_offline(parent, error);
+}
+
+gboolean
+fu_thunderbolt_device_close(FuDevice *device, GError **error)
+{
+	g_autoptr(FuUdevDevice) parent = get_udev_parent_device(device, error);
+	fu_thunderbolt_device_set_port_online(parent, error);
```
superm1, line 418: just return fu_thunderbolt_device_set_port_online(parent, error)

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,105 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+		return FALSE;
+	}
+	parent = get_udev_parent_device(device, error);
+	return fu_thunderbolt_device_set_port_offline(parent, error);
+}
+
+gboolean
+fu_thunderbolt_device_close(FuDevice *device, GError **error)
+{
+	g_autoptr(FuUdevDevice) parent = get_udev_parent_device(device, error);
```
superm1, line 417: you need to test if `get_udev_parent_device` was NULL before passing it on

krankun:
> pre-commit continues to fail and various nested comments. No need to open a new PR each time. You can just `git push --force`

I hope pre commit should pass now as there was a clang format issue with fu-thunderbolt-device.h

krankun:
> > pre-commit continues to fail and various nested comments. No need to open a new PR each time. You can just `git push --force`
> 
> I hope pre commit should pass now as there was a clang format issue with fu-thunderbolt-device.h

precommit passed now with https://github.com/fwupd/fwupd/commit/b741625dcee3fdc6df5fe3d5c45e6d9cb24fc7c2

plugins/thunderbolt/fu-thunderbolt-device.c
krankun: Closing this , Thank you!

src/fu-udev-backend.c
krankun: this is incorporated in latest

libfwupd/fwupd-enums.c
krankun: should I move this flags to libfwupdplugin/fu-device.c ?

superm1: Yes, move it to fu-device.c and the matching one from fwupd-enums.h to fu-device.h

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -320,6 +320,108 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
 	return (guint16)val;
 }
 
+static gboolean
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device, GError **err)
+{
+	if (!fu_udev_device_write_sysfs(device, "usb4_port1/offline", "1", err)) {
```
superm1, line 326: I think you might need an extra check that this file exists first.  Think about the case that this file isn't existing, this function should return true immediately right?

src/fu-util-common.c
```diff
@@ -1217,6 +1217,10 @@ fu_util_device_flag_to_string(guint64 device_flag)
 		/* TRANSLATORS: we might ask the user the recovery key when next booting Windows */
 		return _("Full disk encryption secrets may be invalidated when updating");
 	}
+	if (device_flag == FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE) {
```
superm1, line 1220: once you move the flag to an internal flag the changes for this file should go away

superm1 (commented):
getting much closer, just move some stuff around and a minor nit left

plugins/thunderbolt/fu-thunderbolt-device.c
krankun: as long as the usb4 port exists and kernel driver is loaded this file should exist, I can do a pre check if needed but I cant find any case where this file will be missing and we trigger retimer nvm update.

libfwupd/fwupd-enums.c
krankun: done in https://github.com/fwupd/fwupd/pull/4046/commits/8d0d551cc1a62fbd54d246ec7a5ddb513bac95bd

libfwupd/fwupd-enums.h
krankun: done in https://github.com/fwupd/fwupd/pull/4046/commits/8d0d551cc1a62fbd54d246ec7a5ddb513bac95bd

plugins/thunderbolt/fu-thunderbolt-device.c
krankun: fixed in https://github.com/fwupd/fwupd/commit/98336139be858dd20f4089bfea421f0fda9031b1

libfwupdplugin/fu-device.c
```diff
@@ -230,6 +230,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "use-parent-for-battery";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK)
 		return "use-proxy-fallback";
+	if (flag == FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE)
```
superm1, line 233: Since moved to this file needs "internal" in flag name

libfwupdplugin/fu-device.h
```diff
@@ -400,6 +400,14 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_USE_PROXY_FALLBACK (1ull << 18)
 
+/**
+ * FWUPD_DEVICE_FLAG_NO_AUTO_REMOVE :
```
superm1, line 404: Since moved here needs internal in flag name

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,133 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	if (fu_thunderbolt_check_usb4_port_path(device, offline, err)) {
+		if (!fu_udev_device_write_sysfs(device, offline, "1", err)) {
+			g_set_error_literal(err,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "Setting usb4 port offline failed");
+			return FALSE;
+		}
+	} else {
+		return FALSE;
```
superm1, line 350: This particular one needs special handling - you need to return TRUE here

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,133 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device, GError **err)
+{
+	const gchar *offline = "usb4_port1/offline";
+	const gchar *rescan = "usb4_port1/rescan";
+	if (fu_thunderbolt_check_usb4_port_path(device, offline, err)) {
```
superm1, line 341: This particular one I wouldn't bother having an error argument or you need to pass a local error since you're going to return true if the file doesn't exist.

libfwupd/fwupd-enums.h
```diff
@@ -493,6 +493,7 @@ typedef enum {
  * Since: 1.7.1
  */
 #define FWUPD_DEVICE_FLAG_AFFECTS_FDE (1llu << 45)
+
```
superm1, line 496: Don't change whitespace

libfwupdplugin/fu-device.c
krankun: my bad fixed it

plugins/thunderbolt/fu-thunderbolt-device.c
krankun: fu_thunderbolt_check_usb4_port_path has err argument already !

superm1: You need to pass a local error so you don't set error on true condition

krankun: got it ok as done in https://github.com/fwupd/fwupd/commit/d4cca8d8eb125b4e86abc8b7cff2e7192bb664d3 ?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,136 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	if (fu_thunderbolt_check_usb4_port_path(device, offline, err_usb4_port_offline_path)) {
+		if (!fu_udev_device_write_sysfs(device, offline, "1", err)) {
+			g_set_error_literal(err,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "Setting usb4 port offline failed");
+			return FALSE;
+		}
+	} else {
+		return TRUE;
```
superm1, line 352: just a nit, you may want `g_debug("%s", err_usb4_port_offline_path->message);` here.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,136 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	if (fu_thunderbolt_check_usb4_port_path(device, rescan, err_usb4_port_rescan_path)) {
+		if (!fu_udev_device_write_sysfs(device, rescan, "1", err)) {
+			g_set_error_literal(err,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "Rescan on port failed");
+			return FALSE;
+		}
+	} else {
+		return TRUE;
```
superm1, line 363: just a nit, you may want `g_debug("%s", err_usb4_port_rescan_path->message);` here.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,136 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+					    FWUPD_ERROR_INTERNAL,
+					    "Setting port online failed");
+			return FALSE;
+		}
+	}
+	return TRUE;
+}
+
+static FuUdevDevice *
+get_udev_parent_device(FuDevice *device, GError **error)
```
superm1, line 388: nit: prefix this with `fu_thunderbolt`

superm1 (commented):
just some nits left to me.  @hughsie can you have a look again?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,138 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	g_autoptr(GFile) file = g_file_new_for_path(fn);
+	if (!g_file_query_exists(file, NULL)) {
+		g_set_error(err, G_IO_ERROR, G_IO_ERROR_NOT_FOUND, "failed to find %s", fn);
+		return FALSE;
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_offline(FuUdevDevice *device, GError **err)
```
hughsie, line 337: s/err/error please

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,138 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+{
+	const gchar *offline = "usb4_port1/offline";
+	const gchar *rescan = "usb4_port1/rescan";
+	GError *err_usb4_port_offline_path = NULL;
+	GError *err_usb4_port_rescan_path = NULL;
```
hughsie, line 342: /s/err_usb4_port_offline_path/error_offline please

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,138 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+	}
+	return TRUE;
+}
+
+static gboolean
+fu_thunderbolt_device_set_port_online(FuUdevDevice *device, GError **err)
+{
+	FuUdevDevice *udev = FU_UDEV_DEVICE(device);
+	const gchar *offline = "usb4_port1/offline";
+	GError **err_usb4_port_offline_path = NULL;
```
hughsie, line 375: same; and I think this needs to be g_autoptr

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -319,6 +319,138 @@ fu_thunderbolt_device_get_attr_uint16(FuThunderboltDevice *self, const gchar *na
...
+		return NULL;
+	}
+	return fu_udev_device_new_with_context(fu_device_get_context(FU_DEVICE(self)),
+					       g_steal_pointer(&udev_parent2));
+}
+gboolean
+fu_thunderbolt_probe_retimer(FuDevice *device, GError **error)
+{
+	g_autoptr(FuUdevDevice) parent = NULL;
+	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
```
hughsie, line 434: can you flip these the other way around please? just by convention self is first.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -422,6 +554,11 @@ fu_thunderbolt_device_setup_controller(FuDevice *device, GError **error)
 		fu_device_add_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	}
 
+	if (self->device_type == FU_THUNDERBOLT_DEVICE_TYPE_HOST_CONTROLLER) {
+		if (fu_thunderbolt_device_set_port_offline(FU_UDEV_DEVICE(device), error))
+			return TRUE;
```
hughsie, line 559: what happens if it fails? surely you want:

	if (!fu_thunderbolt_device_set_port_offline(FU_UDEV_DEVICE(device), error))
			return FALSE;

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -763,10 +902,12 @@ fu_thunderbolt_device_write_firmware(FuDevice *device,
...
 	/* whether to wait for a device replug or not */
-	if (!fu_device_has_flag(device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
-		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_RESTART);
+	if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_USABLE_DURING_UPDATE)) {
 		fu_device_set_remove_delay(device, FU_PLUGIN_THUNDERBOLT_UPDATE_TIMEOUT);
-		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_RESTART);
+		if (self->device_type != FU_THUNDERBOLT_DEVICE_TYPE_RETIMER) {
+			fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+		}
```
hughsie, line 910: no real need for braces

src/fu-udev-backend.c
```diff
@@ -41,7 +41,8 @@ fu_udev_backend_device_remove(FuUdevBackend *self, GUdevDevice *udev_device)
 	/* find the device we enumerated */
 	device_tmp =
 	    fu_backend_lookup_by_id(FU_BACKEND(self), g_udev_device_get_sysfs_path(udev_device));
-	if (device_tmp != NULL) {
+	if (device_tmp != NULL &&
+	    !fu_device_has_flag(device_tmp, FWUPD_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)) {
```
hughsie, line 45: i think you want the same thing in fu-**usb**-backend.c too.

krankun: Sure added for  fu-usb-backend.c too  , what are the the benefits to add it to usb backend ?
for later platforms we want to extend this to multiport where both ports may or may not share same nvm payload, is there a place to keep a check on the extracted nvm files not to be cleaned up until all the ports nvm is updated ? currently we dont have this issue but just checking to extend the functionality for future...

hughsie: > what are the the benefits to add it to usb backend

To make the codebase consistent. _NO_AUTO_REMOVE is the kind of thing I could see being used for USB devices too.

> , is there a place to keep a check on the extracted nvm files not to be cleaned up until all the ports nvm is updated

Back a bit; how do you mean? Surely each "port" would have a unique FuDevice and associated GUID?

hughsie:
Thanks @superm1, much appreciated for all the reviews, and most of all for @krankun for all the fixes.

krankun:
> Thanks @superm1, much appreciated for all the reviews, and most of all for @krankun for all the fixes.

Thanks for the reviews and support @superm1 @hughsie @djcampello @RajmohanMani
while this change serves the needs of current platforms we are good to go , in future we would need to keep supporting the few things below 
- Implement a generic code for  below three cases
1) two ports with two retimers share same nvm pay load { just update for one of them which is current case }
2) two ports with two retimers with each have their own nvm pay load { update nvm for both of them }
3) one port can have two retimers { if they share same pay load just update for one of them otherwise for both}

we will keep enhancing the tool for all the above three cases , so there is still some work ahead :)

Issue #4047, opened 2021-12-01
MOTD doesn't show on Fedora


hughsie:
I'm assuming Fedora isn't looking at /run/motd.d/fwupd/85-fwupd -- there's no update-motd like on Ubuntu. @superm1 perhaps we need to install the 85-fwupd.motd to /etc/motd.d/ in this instance?

superm1:
Is there pam_motd` by default?  Or anything else that can be used for this purpose?

hughsie:
@mezcalero got any insight here? What are we supposed to do?

superm1:
I see I left a comment linking back to https://bugs.launchpad.net/ubuntu/+source/pam/+bug/399071 and also mentioning that we don't install 85-fwupd by default outside of Debian/Ubuntu.  
There we install it to `/etc/update-motd.d`.  It essentially just sources `/run/motd.d/fwupd/85-fwupd` which is generated by the daemon.

I would _expect_ Fedora to just get stuff from `/run/motd.d` using `pam_motd`

PR #4050, opened 2021-12-01
Enable the MOTD integration when using pam_motd

hughsie:
This has to be a file, not a file-in-a-directory.

Fixes https://github.com/fwupd/fwupd/issues/4047

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/motd/meson.build
+++ b/data/motd/meson.build
@@ -4,7 +4,7 @@ if get_option('systemd')
     install_dir: systemdunitdir)
   install_data(['fwupd-refresh.preset'],
     install_dir: systemdsystempresetdir)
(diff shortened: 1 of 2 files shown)


hughsie:
@superm1 can you check I've not broken Ubuntu with this change please.

superm1 (approved):

superm1:
> @superm1 can you check I've not broken Ubuntu with this change please.

That looks correct to me.  Given how the Ubuntu one works (it just runs `cat` on the file), I don't expect any breakage.
(I won't be able to easily check on a real system any time soon though)

Issue #3790, opened 2021-09-24
1.6.4: test suite is failing


kloczek:
**Describe the bug**
Test suite is failing in two units.

**Steps to Reproduce**
I'm building rom package so normal steps:
- meson with below setup:
```spec
%meson \
        -D agent=true \
        -D build=all \
        -D consolekit=true \
        -D firmware-packager=true \
        -D gpg=true \
        -D gtkdoc=true \
        -D gudev=true \
        -D introspection=true \
        -D lvfs=true \
        -D man=true \
        -D pkcs7=true \
        -D plugin_altos=true \
        -D plugin_amt=true \
        -D plugin_coreboot=true \
        -D plugin_dell=true \
        -D plugin_dummy=true \
        -D plugin_emmc=true \
        -D plugin_flashrom=true \
        -D plugin_modem_manager=true \
        -D plugin_msr=true \
        -D plugin_nvme=false \
        -D plugin_nvme=true \
        -D plugin_redfish=true \
        -D plugin_synaptics=true \
        -D plugin_thunderbolt=true \
        -D plugin_uefi_capsule=true \
        -D plugin_uefi_pk=true \
        -D supported_build=true \
        -D systemd=true \
        -D tests=true \
        -D tpm=true \
        %{nil}
```
- meson build
- meson install
- meson test

**Expected behavior**
test suite should not fail

**fwupd version information**
1.6.4

**Additional questions**

- Operating system and version:
Linux/x86_64
- Have you tried rebooting?
N/A
- Is this a regression?
In the same env 1.6.3 test suyite passes without issues.

Test suite output:
```console
+ cd fwupd-1.6.4
+ /usr/bin/meson test -C x86_64-redhat-linux-gnu --num-processes 48 --print-errorlogs
ninja: Entering directory `/home/tkloczko/rpmbuild/BUILD/fwupd-1.6.4/x86_64-redhat-linux-gnu'
ninja: no work to do.
 1/32 fwupd-exported-api                     OK               0.05s
 2/32 fwupdplugin-exported-api               OK               0.02s
 3/32 fwupd-context-test                     OK               2.92s
 4/32 fwupd-thread-test                      OK               2.99s
 5/32 fwupd-self-test                        OK               3.00s
 6/32 fwupdplugin-self-test                  OK               2.34s
 7/32 fu-self-test                           FAIL             1.62s   killed by signal 6 SIGABRT
[...]

PR #4055, opened 2021-12-06
Ensure max_user_instances is set to a sane value at startup

hughsie:
Fixes the hard-to-debug https://github.com/fwupd/fwupd/issues/3790

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -89,6 +89,7 @@
 #endif
 
 #define MINIMUM_BATTERY_PERCENTAGE_FALLBACK 10
+#define INOTIFY_MIN_USER_INSTANCES	    64
 
 static void
 fu_engine_finalize(GObject *obj);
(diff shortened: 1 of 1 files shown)


hughsie:
On the logic, it's better to "_just work_" rather than fail in a "_good way_".

superm1 (commented):
I don't like a daemon going and changing this just because it has the permissions to do so.  What if an admin intentionally lowered it and is closely tracking what apps are using it and pushing the threshold?

hughsie:
> I don't like a daemon going and changing this just because it has the permissions to do so

Agreed, it didn't sit 100% right with me either. I've switched it to just dump a warning in the logs instead.

travier:
We had a "similar" issue (https://github.com/coreos/fedora-coreos-tracker/issues/637) and asked for it to be more dynamic (https://bugzilla.redhat.com/show_bug.cgi?id=1888617) and that landed in 5.11.

src/fu-engine.c
```diff
@@ -89,6 +89,7 @@
 #endif
 
 #define MINIMUM_BATTERY_PERCENTAGE_FALLBACK 10
+#define INOTIFY_MIN_USER_INSTANCES	    64
```
superm1, line 92: Is this really going to find anything? 64 is veryyyyy low. Given the commit linked I have an idea.

How about if you detect it's less than 1% of system ram just leave a warning in the logs.

hughsie: >64 is veryyyyy low

Any higher than that and my fwupd started up just fine. I'm hesitant to make this a percentage of the RAM as that gets into all kind of edge cases. How about we make it hardcoded to 8192/2, i.e. 4096?

superm1: >Any higher than that and my fwupd started up just fine

Right - this should only be an issue while system is under pressure.  Like if you started up vscode with the C++ extension and opened a kernel directory and then started fwupd.

>How about we make it hardcoded to 8192/2, i.e. 4096

I'm not sure I agree with this either though.  It's not going to be a deterministic failure.  What about instead moving it back to the failure path like my original idea?

hughsie: Like https://github.com/fwupd/fwupd/pull/4060 ?

superm1: Yup, I think that's a lot more useful.

PR #4059, opened 2021-12-07
Add the CFI JEDEC instance ID if using vendor-extended version

hughsie:
Setting a flash ID of `C84016` would then add `CFI\FLASHID_C84016` and
also `CFI\FLASHID_C840`. This makes it easier to write quirks.

Type of pull request:

- [X] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/tests/quirks.d/tests.quirk
+++ b/libfwupdplugin/tests/quirks.d/tests.quirk
@@ -20,4 +20,6 @@ Flags = updatable,internal
 Name = A25Lxxx
 CfiDeviceCmdChipErase = 0xc7
 CfiDeviceCmdSectorErase = 0x20
+CfiDevicePageSize = 0x200
+CfiDeviceSectorSize = 0x2000
 FirmwareSizeMax = 0x10000
--- a/data/cfi.quirk
+++ b/data/cfi.quirk
@@ -5,7 +5,7 @@ CfiDeviceCmdSectorErase = 0x20
 
 [CFI\FLASHID_1F65]
 Name = AT25F512A/B
-# CfiDeviceCmdReadId = 0x15
+CfiDeviceCmdReadId = 0x15
 CfiDeviceCmdChipErase = 0x62
 CfiDeviceCmdSectorErase = 0x00
 FirmwareSizeMax = 0x10000
@@ -37,19 +37,19 @@ CfiDeviceCmdSectorErase = 0x20
 
 [CFI\FLASHID_00BF]
 Name = PCT/SST25VFxxx/xxxA
-# CfiDeviceCmdReadId = 0x90
+CfiDeviceCmdReadId = 0x90
 CfiDeviceCmdChipErase = 0x60
 CfiDeviceCmdSectorErase = 0x20
 
 [CFI\FLASHID_009D]
 Name = PM25LDxxx
-# CfiDeviceCmdReadId = 0x90
+CfiDeviceCmdReadId = 0x90
 CfiDeviceCmdChipErase = 0xC7
 CfiDeviceCmdSectorErase = 0xD7
 
 [CFI\FLASHID_009D]
 Name = PM25LVxxx
-# CfiDeviceCmdReadId = 0xAB
+CfiDeviceCmdReadId = 0xAB
 CfiDeviceCmdChipErase = 0xC7
 CfiDeviceCmdSectorErase = 0xD7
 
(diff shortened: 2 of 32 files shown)


gportay:
I have tested this so far, and it works!

if I query `C84012` and this quirk does not exist, it falls back to the `C840` which exists.

also, I wonder if it can be ~useful~ relevent to add the `CfiDeviceBlockSize` and `CfiDeviceSectoreSize`.

hughsie:
> to add the CfiDeviceBlockSize and CfiDeviceSectoreSize

Yes, I think that makes sense. It'd have to be something like fu_cfi_device_get_block_size() rather than FU_CFI_DEVICE_CMD_XXX as it's not a command, and it won't fit in a uint8_t. Do you want me to add those today?

gportay:
> Yes, I think that makes sense. It'd have to be something like fu_cfi_device_get_block_size() rather than FU_CFI_DEVICE_CMD_XXX as it's not a command, and it won't fit in a uint8_t. Do you want me to add those today?

Whenever you want, I am fine with hardcoded values for now.

hughsie:
>  I am fine with hardcoded values for now.

Out of

hughsie:
@gportay like that^^?

gportay:
> @gportay like that^^?

ho, pretty fast... sounds good.

hughsie:
@gportay test it out in your new plugin, and if it works, can you mark as review approval pls.

libfwupdplugin/tests/quirks.d/tests.quirk
```diff
@@ -20,4 +20,6 @@ Flags = updatable,internal
 Name = A25Lxxx
 CfiDeviceCmdChipErase = 0xc7
 CfiDeviceCmdSectorErase = 0x20
+CfiDeviceBlockSize = 0x200
```
gportay, line 23: Well, I am a bit confused, does a block size should be bigger than sector size?

gportay (approved):
I have tested this and it works.

also, typo in commit title: ~~FuCfuDevice~~ FuCfiDevice.

libfwupdplugin/tests/quirks.d/tests.quirk
gportay: also, does a page == sector?

my understanding is that the flash is divided into several blocks that are divided into several pages (i.e. sector?).

hughsie: > does a block size should be bigger than sector size

Ahh, no, normally a block is much smaller. i.e. lots of blocks in each sector.

gportay: oh, so when splitting into chunks, the code below is wrong:

```
        chunks = fu_chunk_array_new(NULL,
                                    len,
                                    start_addr,
                                    self->flash_block_size,
                                    self->flash_sector_size);
```
except if `block_size` is bigger than `sector_size`; by that, I mean the last argument should be smaller than the one before; or maybe that is not important.

hughsie: > normally a block is much smaller

I don't think this has to be true; I think the other way words as well, but it would be unusual indeed.

hughsie: For SPI devices, normally the block size is something like 256 bytes, and the sector size is much larger, e.g. 4Kb. You can only erase in whole sector amounts, but you can normally write in either whole blocks, or even one byte at a time.

hughsie: maybe s/fu_cfi_device_get_block_size/fu_cfi_device_get_page_size might be more obvious?

hughsie: I've renamed this.

gportay: I do not have a strong opinion; as long it is consistent with the name of the commands :/

gportay: Well, I am still confused when reading the datasheets.

Look at section [3 MEMORY ORGANIZATION](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwiSkOuT4Nb0AhVLXhoKHZLOB2AQFnoECAYQAQ&url=https%3A%2F%2Fwww.gigadevice.com%2Fdatasheet%2Fgd25d40c%2F&usg=AOvVaw0I5W8LUwzN7g-AqxCslZFk)

Memory is organized into 8 blocks of 16 sectors.

gportay: And a sector is composed of pages.

gportay: so I guess, we may need to have three getters/setters in the end.

gportay: re-reading, I will need also the block size.

I need the values 0x1.0000 and 0x1000 which are the block-size and sector-size for the C84016.

PR #4060, opened 2021-12-07
Fixup the GLib error message when inotify/max_user_instances is too low

hughsie:
This can be reproduced using sysctl fs.inotify.max_user_instances=25
and then running the daemon.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
You perfectly transcribed the idea from my mind

PR #4061, opened 2021-12-07
trivial: synaptics-prometheus: add another product ID

superm1:
This is used in some upcoming systems

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/synaptics-prometheus/synaptics-prometheus.quirk
+++ b/plugins/synaptics-prometheus/synaptics-prometheus.quirk
@@ -38,3 +38,14 @@ InstallDuration = 2
 Plugin = synaptics_prometheus
 InstallDuration = 2
 
+[USB\VID_06CB&PID_0123]
+Plugin = synaptics_prometheus
+InstallDuration = 2
+
+[USB\VID_06CB&PID_0126]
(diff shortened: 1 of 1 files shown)


superm1:
On my side I did confirm a 0123 device at least reads version correctly.

plugins/synaptics-prometheus/synaptics-prometheus.quirk
```diff
@@ -38,3 +38,14 @@ InstallDuration = 2
 Plugin = synaptics_prometheus
 InstallDuration = 2
 
+[USB\VID_06CB&PID_0123]
```
hughsie, line 41: 0123 sure looks suspicious...

superm1: I thought the same thing, but I reached out to Vincent at Synaptics and then noticed the exact same ID in fprint
https://gitlab.freedesktop.org/libfprint/libfprint/-/commit/128d809227bbef1d3bd8e7c7e1f9b68a335db252

hughsie (approved):

PR #4070, opened 2021-12-10
Wip/hughsie/tpm

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm/fu-self-test.c
+++ b/plugins/tpm/fu-self-test.c
@@ -26,7 +26,8 @@ fu_tpm_device_1_2_func(void)
 	g_autoptr(FuContext) ctx = fu_context_new();
 	g_autoptr(FuPlugin) plugin = fu_plugin_new(ctx);
 	g_autoptr(FuSecurityAttrs) attrs = fu_security_attrs_new();
-	g_autoptr(FwupdSecurityAttr) attr = NULL;
+	g_autoptr(FwupdSecurityAttr) attr0 = NULL;
+	g_autoptr(FwupdSecurityAttr) attr1 = NULL;
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GPtrArray) pcr0s = NULL;
 	g_autoptr(GPtrArray) pcrXs = NULL;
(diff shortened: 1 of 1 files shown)


superm1 (approved):

plugins/tpm/fu-self-test.c
```diff
@@ -26,7 +26,8 @@ fu_tpm_device_1_2_func(void)
 	g_autoptr(FuContext) ctx = fu_context_new();
 	g_autoptr(FuPlugin) plugin = fu_plugin_new(ctx);
 	g_autoptr(FuSecurityAttrs) attrs = fu_security_attrs_new();
-	g_autoptr(FwupdSecurityAttr) attr = NULL;
+	g_autoptr(FwupdSecurityAttr) attr0 = NULL;
```
kerneis-anssi, line 29: I'm not sure if I introduced this or not, but I definitely could have since I'm used to C++'s `unique_ptr` which does the right thing when re-assigning. In hindsight, it's obvious why the same is not possible in C.

Is there some systematic way to catch this, eg. by running the tests under LeakSanitizer or Valgrind in CI?

plugins/tpm/fu-self-test.c
```diff
@@ -189,9 +189,9 @@ static void
 fu_tpm_empty_pcr_func(void)
 {
 	gboolean ret;
-	g_autofree gchar **environ = NULL;
 	g_autofree gchar *pluginfn = NULL;
 	g_autofree gchar *testdatadir = NULL;
+	g_auto(GStrv) environ_old = NULL;
```
kerneis-anssi, line 197: I'm fine with the change but I'm curious why clang thinks there is a clash: is the global `environ` from libc?

Also is there a typo in the commit message? I think you mean it's *not* a `gchar**`, is it?

kerneis-anssi (approved):

plugins/tpm/fu-self-test.c
hughsie: > Is there some systematic way to catch this

I do runs with Valgrind, Coverity and `-Db_sanitize=none, address, thread, undefined, memory, address,undefined` before each release but there are too many false positives to do this in CI. Valgrind gets closest, but it's really really slow.

PR #4074, opened 2021-12-11
Add support for UF2 devices

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
We should probably tag a release before we review+merge this.

plugins/uf2/uf2.quirk
```diff
@@ -0,0 +1,136 @@
+[BLOCK\UUID_E478-FA50]
+Plugin = uf2
+Name = ColorHug2
```
superm1, line 3: huh?

hughsie: Ohh it's a test device I used for testing. We probably don't want this in the proper PR, and by then some of the proper hardware I've ordered will have turned up :)

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+	FuBlockDevicePrivate *priv = GET_PRIVATE(self);
+
+	if (priv->uuid != NULL) {
+		g_autofree gchar *instance_id0 = NULL;
+		instance_id0 = g_strdup_printf("BLOCK\\UUID_%s", priv->uuid);
+		fu_device_add_instance_id(FU_DEVICE(self), instance_id0);
+	}
+	if (priv->label != NULL) {
+		g_autofree gchar *instance_id0 = NULL;
+		instance_id0 = g_strdup_printf("BLOCK\\LABEL_%s", priv->label);
```
superm1, line 61: I'm a bit worried about instances that are entirely built on a label or UUID.  need more factors in these UUIDs and/or drop them.

* UUID: Too specific, if it's reformatted you can no longer target it.
* Label: Too generic.  It's just asking for me to be installing updates for my camera to my OS recovery partition if they're both called `RECOVERY`

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+
+static void
+fu_block_device_to_string(FuDevice *device, guint idt, GString *str)
+{
+	FuBlockDevice *self = FU_BLOCK_DEVICE(device);
+	FuBlockDevicePrivate *priv = GET_PRIVATE(self);
+	fu_common_string_append_kv(str, idt, "Uuid", priv->uuid);
+	fu_common_string_append_kv(str, idt, "Label", priv->label);
+	fu_common_string_append_kv(str, idt, "Filename", priv->filename);
+}
```
superm1, line 46: I think it would be good to include details about the filesystem.  This might be a good way to narrow down devices later if we need to.

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * fu_block_device_set_uuid:
+ * @self: a #FuBlockDevice
+ * @uuid: the uuid, e.g. `E478-FA50`
+ *
+ * Sets the UUID.
+ *
+ * Since: 1.7.3
```
superm1, line 76: 4

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * fu_block_device_get_uuid:
+ * @self: a #FuBlockDevice
+ *
+ * Gets the UUID.
+ *
+ * Returns: string, or %NULL for unset
+ *
+ * Since: 1.7.3
```
superm1, line 102: 4

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/**
+ * fu_block_device_set_label:
+ * @self: a #FuBlockDevice
+ * @label: the label, e.g. `FWUPDATE`
+ *
+ * Sets the block device label.
+ *
+ * Since: 1.7.3
```
superm1, line 119: 4

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * fu_block_device_get_label:
+ * @self: a #FuBlockDevice
+ *
+ * Gets the label.
+ *
+ * Returns: string, or %NULL for unset
+ *
+ * Since: 1.7.3
```
superm1, line 145: 4

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/**
+ * fu_block_device_set_filename:
+ * @self: a #FuBlockDevice
+ * @filename: the filename, e.g. `FWUPDATE`
+ *
+ * Sets the filename to write to the volume.
+ *
+ * Since: 1.7.3
```
superm1, line 162: 4

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * fu_block_device_get_filename:
+ * @self: a #FuBlockDevice
+ *
+ * Gets the filename to write to the volume.
+ *
+ * Returns: string, or %NULL for unset
+ *
+ * Since: 1.7.3
```
superm1, line 188: 4

libfwupdplugin/fu-block-device.c
```diff
@@ -0,0 +1,395 @@
...
+	/* get blob */
+	fw = fu_firmware_get_bytes(firmware, error);
+	if (fw == NULL)
+		return FALSE;
+
+	/* sanity check */
+	if (fu_device_get_logical_id(device) == NULL || priv->filename == NULL) {
+		g_set_error_literal(error, G_IO_ERROR, G_IO_ERROR_FAILED, "no valid path");
+		return FALSE;
+	}
```
superm1, line 249: same sanity check in `fu_block_device_io_channel_new`, drop it here

libfwupdplugin/fu-block-device.c
hughsie: What kind of thing did you have in mind?

libfwupdplugin/fu-block-device.c
hughsie: > UUID: Too specific, if it's reformatted you can no longer target it.

I actually think UUID is okay; it's not possible to "format" a UF2 drive, it's just ignored. It's not really a filesystem, it's just a driver that reads the 512 byte blocks and writes them to each sector. If you formatted it as ext4 for instance it'd come back as a FAT32 drive with the hardcoded UUID.

Label is tricky, I agree. How about `BLOCK\\UUID_%s` and `BLOCK\\UUID_%s&LABEL_%s"?

libfwupdplugin/fu-block-device.c
superm1: Well I didn't think U2F spec required FAT32 but you can get this info from udisks.  So just adding it as a string might be useful for a future day we ask for `--verbose` output for a problem.

superm1: The spec just seems to say FAT, so I don't know if that can also mean exFAT or FAT16, etc.

libfwupdplugin/fu-block-device.c
superm1: `INFO_UF2.TXT` should have a bunch of information - that is what should be going into the instance IDs, not this stuff in my mind.

libfwupdplugin/fu-block-device.c
hughsie: God catch, fixed.

libfwupdplugin/fu-block-device.c
hughsie: > that is what should be going into the instance IDs

Right, but that's one layer "up" -- this is the `FuBlockDevice` that lives in libfwupdplugin (that we match on in the quirk file) and then the `FuUf2Device` that superclasses the `FuBlockDevice` that adds the `INFO_UF2.TXT`-derived information for matching the firmware stream.

superm1 (changes requested):
I think change to use `INFO_UF2.TXT`

libfwupdplugin/fu-block-device.c
hughsie: Well, FuUdisksBackend will only create devices if the type is `vfat` -- maybe that's too draconian. How about we add the id_type *if* we allow non-vfat device types?

libfwupdplugin/fu-block-device.c
superm1: > Right, but that's one layer "up
Oh right.

>and then the FuUf2Device that superclasses the FuBlockDevice that adds the INFO_UF2.TXT-derived information for matching the firmware stream

Even being the wrong layer - I don't think we want these UUIDs to be anything that can be used to target a device for the same generic or too specific reasons I mentioned.  I think that step should just be skipped.

hughsie: > I think that step should just be skipped.

Well, we need to match the device by UUID or LABEL so we know if we should create the FuBlockDevice and offer it to plugins -- I really don't think we want to be reading from *all* random VFAT drives as soon as they're plugged in -- that seems like the kind of change that would make people upset. I think fwupd looking for a `INFO_UF2.TXT` on a drive called `FWUPDATE` or where the UUID is allow-listed by a vendor is probably acceptable. I suppose one compromise would be to add the `BLOCK\\UUID_%s` instance IDs as `FU_DEVICE_INSTANCE_FLAG_ONLY_QUIRKS` so you can't use them to match to a firmware stream.

plugins/uf2/uf2.quirk
hughsie: I've tested this with a `Raspberry Pi RP2` -- unfortunately this device seems to have a random UUID :(

superm1: Maybe it's better to go off something else like SMBIOS / DT for devices like rpi.

libfwupdplugin/fu-block-device.c
superm1: sounds good

plugins/uf2/uf2.quirk
hughsie: Ohh, it's not running *on* the rpi, it's when the rpi is in "peripheral" mode, i.e. it's the device you're attaching to your Linux device, rather than the other way around.

superm1: is there anything else in the USB descriptor you could go off maybe then?

hughsie: That's a good question. Will poke.

hughsie: I'm struggling a bit going from dev path e.g. /dev/sda1 to the USB device. Any ideas? There's /sys/class/block/sda/device/model but that seems pretty raw.

superm1: Presumably you can't go from udev parents? Did you look at the udev database for any other links?

hughsie: So this lead down a rabbithole, which then involved rewriting the plugin, possibly for the better. New idea here: https://github.com/fwupd/fwupd/pull/4096

PR #4077, opened 2021-12-13
fu-device: copy private flags from donor devices

superm1:
This may have an unintended side effect that devices from different
plugins have undefined flags in their new plugin, watch out for dragons!

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
> watch out for dragons

Whoa, more than just dragons; this would cause carnage no? I was going to suggest a patch that just used `fu_device_set_custom_flags()` during incorporate...

PR #4078, opened 2021-12-13
Copy private flags from donor devices

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -74,6 +74,7 @@ typedef struct {
 	FuDeviceInternalFlags internal_flags;
 	guint64 private_flags;
 	GPtrArray *private_flag_items; /* (nullable) */
+	gchar *custom_flags;
 	gulong notify_flags_handler_id;
 } FuDevicePrivate;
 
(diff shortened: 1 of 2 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -4516,6 +4523,8 @@ fu_device_incorporate(FuDevice *self, FuDevice *donor)
 		fu_device_set_proxy(self, priv_donor->proxy);
 	if (priv->proxy_guid == NULL && priv_donor->proxy_guid != NULL)
 		fu_device_set_proxy_guid(self, priv_donor->proxy_guid);
+	if (priv->custom_flags == NULL && priv_donor->custom_flags != NULL)
+		fu_device_set_custom_flags(self, priv_donor->custom_flags);
```
superm1, line 4527: This has the same risk as my original one doesn't it?  I think we need a warning if fu_device_get_plugin varies between them perhaps.

hughsie: I don't think so; it's not registering them, it's just doing fu_device_private_flag_item_find_by_str() on the *new* gtype with the cached `Flags` string.

superm1 (approved):

PR #4076, opened 2021-12-12
modem-manager: add Quectel EG25-G modem support

DylanVanAssche:
Type of pull request:

- [ ] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [x] Code fix
- [ ] Feature
- [ ] Documentation

# Description

The Quectel EG25-G modem is used in the [Pine64 PinePhone](https://pine64.org/pinephone) and [PinePhone Pro](https://pine64.org/pinephonepro) .
By default it ships with an old firmware version which can be upgraded to newer versions from Quectel or a more open firmware developed at https://github.com/Biktorgj/pinephone_modem_sdk.
The current `fwupd` version sees the Quectel EG25-G through the `modem-manager` plugin as upgradable using the `fastboot` plugin. However, the Quectel EG25-G fastboot bootloader is picky when you interact with it:

- Blocksize must be `16384` bytes, not a byte more or less. If you still try another size, the next chunk the `fastboot` plugin writes to the modem fails which reboots the bootloader. I discovered this by analyzing the USB traffic of the same payload (once with `fwupd` and once with the original fastboot binary) with Wireshark.
- Writing and reading data cannot happen at `fwupd`'s rate otherwise the fastboot bootloader crashes and reboot, the original fastboot binary seems to have some delay between writes/reads. If you enable the debug prints of the `fastboot` plugin, this is also 'fixed', but I properly fixed it by introducing a new quirk.
- The modem shows up under an entirely different USB vendor and device ID which causes `fwupd` to not see it. Adding the necessary quirks in `modem-manager` fixes it (like done for all other modems there).

# Changes

- Add GUIDs for the Quectel EG25-G modem in fastboot to `modem-manager` plugin's quirks
- Introduce a new quirk `ModemManagerAtHasResponse` to ignore the responses when entering fastboot mode, default: expects a response.
- Add this no response quirk for the Quectel EG25-G modem
- Introduce a delay operations quirk in the `fastboot` plugin to sleep after a read or write operation, default: 0 us.
- Add this delay operations quirk tor the Quectel EG25-G modem

# Test environment

- Pine64 PinePhone with a Quectel EG25-G modem
- Alpine Linux / postmarketOS edge
- musl

# Disclaimer for distros

*These patches are not ready to be shipped to end-users, I kindly ask you to not ship them to any users, even when they are bleeding edge channels of your distro. At all times, I'm not responsible for complains about bricked modems!*

# Related issues

Fixes #3968

PR #4086, opened 2021-12-15
Quectel EG25-G: improve upload stability and require AC power

DylanVanAssche:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

With #4076 merged, I have a following up MR:

# Changes

## Improve upload stability

I did more tests with bigger firmware packages and in some rare cases, 100ms between every read/write operation is too low for a stable upgrade process. Increasing this to 250ms improves the upload stability a lot.
Without it, it might be that the upgrade process stalls after ~5-10 minutes as the bootloader doesn't respond anymore.

## Require AC power to upgrade

To improve the safety of upgrades, I added a flag to the `modem-manager` to require AC power before starting the upgrade.
This is more safe as you don't want to brick the modem due to an empty battery on a laptop or in my case a phone.

# Test environment

- Pine64 PinePhone
- Quectel EG25-G modem
- postmarketOS edge
- fwupd 90baeaabbc94589512642fe970eecbd852ae3532

CC: @aleksander0m @ivanmikh

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -1520,6 +1520,7 @@ fu_mm_device_init(FuMmDevice *self)
 {
 	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_USE_RUNTIME_VERSION);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
 	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_PLAIN);
 	fu_device_set_summary(FU_DEVICE(self), "Mobile broadband device");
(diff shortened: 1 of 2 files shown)


plugins/modem-manager/fu-mm-device.c
```diff
@@ -1520,6 +1520,7 @@ fu_mm_device_init(FuMmDevice *self)
 {
 	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
 	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_USE_RUNTIME_VERSION);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_REQUIRE_AC);
```
hughsie, line 1523: is this true?

DylanVanAssche: In theory you can upgrade it without external power, but the update takes a lot of time (~20 minutes) and if the battery runs out the modem can start to act weird. I don't know what the consequences would be if you lose power right during the flashing.

If the modem is bricked, you need to enter EDL mode to fix it which is hard:
- You have to connect 2 GPIO pins on the modem manually with some wire
- EDL mode is not really documented somewhere on how to fix it, so that's hard for people who don't know the internals of the modem.

That's why I purpose this patch, to avoid this situation as much as possible since this runs on phones now as well.
These devices have a battery that is way smaller than laptops and IMO we don't want people to upgrade modem firmwares when they are on the go without access to an external power source.

hughsie: @aleksander0m are you okay with adding this attribute for *all* MM devices? Or would you rather make it per-device with a `Flags = require-ac` quirk?

aleksander0m: I think it's a good idea to apply that to all modems, as they usually take quite a lot of time to upgrade.

hughsie (approved):

PR #4087, opened 2021-12-15
Add fu_memmem_safe() with fallback code for win32

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-fmap-firmware.c
+++ b/libfwupdplugin/fu-fmap-firmware.c
@@ -6,8 +6,6 @@
 
 #include "config.h"
 
-#include <string.h>
-
 #include "fu-common.h"
 #include "fu-fmap-firmware.h"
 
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-fmap-firmware.c
```diff
@@ -75,8 +47,9 @@ fu_fmap_firmware_parse(FuFirmware *firmware,
 
 	/* only search for the fmap signature if not fuzzing */
 	if ((flags & FWUPD_INSTALL_FLAG_NO_SEARCH) == 0) {
```
superm1, line 49: This guard still needed? Wasn't it because of CI maybe not having memmem?

hughsie: no, it was the fuzzer would spend hours in stuck in the memmem call trying every possible offset.

superm1 (approved):

PR #4092, opened 2021-12-16

Sean-StarLabs:
Remove "number" version format to allow different versioning


hughsie:
Woah, what's all the launchpad stuff about?

PR #4093, opened 2021-12-16

Sean-StarLabs:
Remove "number" version format to allow different versioning


Sean-StarLabs:
@hughsie Sorry - wrong branch.

Launchpad stuff is for building debs

superm1 (commented):
Is this going to cause problems for firmware already on LVFS with older tools and/or newer firmware if this change was accepted?

PR #4096, opened 2021-12-16
Add support for UF2 devices

hughsie:
A UF2 device exposes a VFAT block device (sometimes called a Mass
Storage Device) which has a virtual file called `INFO_UF2.TXT` where
metadata can be read. It may also have a the current firmware exported
as a file called `CURRENT.UF2` which is in a 512 byte-block UF2 format.

Writing any file to the MSD will cause the firmware to be written.
Sometimes the device will restart and the volume will be unmounted
and then mounted again. In some cases the volume may not “come back”
until the user manually puts the device back in programming mode.

Match the block devices using the VID*PID, UUID or label, and then
create a UF2 device which can be used to flash firmware.

Note: We only read metadata from allow-listed IDs to avoid causing
regressions on non-UF2 volumes. To get the UUID and label you can
use commands like:

    udisksctl info -b /dev/sda1

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


meson_options.txt
```diff
@@ -43,6 +43,7 @@ option('plugin_mtd', type : 'boolean', value : true, description : 'enable MTD s
 option('plugin_flashrom', type : 'boolean', value : false, description : 'enable libflashrom support')
 option('plugin_platform_integrity', type : 'boolean', value : false, description : 'enable platform integrity support')
 option('plugin_intel_spi', type : 'boolean', value : false, description : 'enable Intel SPI support')
+option('plugin_uf2', type : 'boolean', value : true, description : 'enable support for U2F')
```
superm1, line 46: s/U2F/UF2

plugins/uf2/README.md
```diff
@@ -0,0 +1,43 @@
...
+This plugin supports the following protocol ID:
+
+* com.microsoft.uf2
+
+## GUID Generation
+
+These devices use standard block DeviceInstanceId values, e.g.
+
+* `BLOCK\UUID_E478-FA50`
+* `BLOCK\LABEL_FWUPDATE`
```
superm1, line 24: since you have UF2 ones now, how about drop these?

plugins/uf2/fu-uf2-device.c
```diff
@@ -0,0 +1,390 @@
...
+	if (!fu_firmware_parse(firmware, fw, flags, error))
+		return NULL;
+
+	/* check the family_id matches if we can read the old firmware */
+	if (self->family_id > 0 && fu_firmware_get_idx(firmware) > 0 &&
+	    self->family_id != fu_firmware_get_idx(firmware)) {
+		g_set_error(error,
+			    G_IO_ERROR,
+			    G_IO_ERROR_NOT_SUPPORTED,
+			    "family ID was different, expected 0x%08x bytes and got 0x%08x",
```
superm1, line 39: family id doesn't have bytes, isn't it just a number?

plugins/uf2/fu-uf2-device.c
```diff
@@ -0,0 +1,390 @@
...
+			       GError **error)
+{
+	FuUf2Device *self = FU_UF2_DEVICE(device);
+	g_autoptr(FuFirmware) firmware = fu_uf2_firmware_new();
+
+	if (!fu_firmware_parse(firmware, fw, flags, error))
+		return NULL;
+
+	/* check the family_id matches if we can read the old firmware */
+	if (self->family_id > 0 && fu_firmware_get_idx(firmware) > 0 &&
```
superm1, line 34: what's wrong with comparing 0?

plugins/uf2/fu-uf2-device.c
```diff
@@ -0,0 +1,390 @@
...
+	/* parse to get version */
+	if (!fu_firmware_parse(firmware, fw, FWUPD_INSTALL_FLAG_NONE, error))
+		return FALSE;
+	if (fu_firmware_get_version(firmware) != NULL)
+		fu_device_set_version(device, fu_firmware_get_version(firmware));
+
+	/* add instance ID for quirks */
+	if (fu_firmware_get_idx(firmware) != 0x0) {
+		g_autofree gchar *id0 = NULL;
+		id0 = g_strdup_printf("UF2\\FAMILY_%08X", (guint32)fu_firmware_get_idx(firmware));
```
superm1, line 65: since family is just an integer - how would you feel about adding a VENDOR_%s as well into the string?

plugins/uf2/README.md
hughsie: Do you think we can rely on a valid USB VID/PID? I'm worried about a devboard having the same PID and a different UUID. Maybe we want `VID_1234&PID_5678&UUID_E478-FA50`` as an additional instance ID?

plugins/uf2/fu-uf2-device.c
hughsie: I was using 0 for "unset"

plugins/uf2/fu-uf2-device.c
```diff
@@ -0,0 +1,390 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_uf2_device_open(FuDevice *device, GError **error)
+{
+	/* wait for the user session to auto-mount the volume --  we do not want to use
+	 * fu_volume_mount() here and make the volume only accessible by the fwupd user */
```
superm1, line 217: if it's not mounted, why not?  Just take note that you mounted it, let fwupd do it's thing and unmount it.

meson_options.txt
hughsie: Ta, fixed.

plugins/uf2/fu-uf2-device.c
hughsie: In theory, the family ID is like a registry -- so the vendor is not required.

hughsie: You also don't get the vendor ID from the firmware, but you do get the family ID.

plugins/uf2/fu-uf2-device.c
hughsie: Oops, thanks.

plugins/uf2/README.md
hughsie: Okay, I've dropped LABEL completely, and done the `VID_1234&PID_5678&UUID_E478-FA50` thing.

plugins/uf2/fu-uf2-device.c
hughsie: Okay, I've added something like that; 100% untested but feedback welcome.

plugins/uf2/fu-uf2-device.c
superm1: what is to stop cheap vendor XYZ from using reference design and forgetting to change it?

plugins/uf2/README.md
superm1: the UUID thing I do feel like is so specific it only works for a single devboard.  Maybe it needs a description in README not to rely upon it for anything other than prototyping.

hughsie: So maybe we nuke the UUID thing completely and then re-add it if we find a board where the VID/PID is insufficient?

plugins/uf2/fu-uf2-device.c
hughsie: the family ID is specified in the firmware, and it's the only thing we've got to match up the custom firmware with the generic devboard.

plugins/uf2/fu-uf2-device.c
hughsie: Seems to work.

plugins/uf2/fu-uf2-device.c
hughsie: I've now tested this on the *Adafruit Trinket M0* and *Adafruit Trinket M0* and it seems to work well.

plugins/uf2/README.md
```diff
@@ -0,0 +1,61 @@
...
+`INFO_UF2.TXT` where metadata can be read from. It may also have a the
+current firmware exported as a file `CURRENT.UF2` which is in a 512
+byte-block UF2 format.
+
+Writing any file to the MSD will cause the firmware to be written.
+Sometimes the device will restart and the volume will be unmounted and then
+mounted again. In some cases the volume may not “come back” until the user
+manually puts the device back in programming mode.
+
+Match the block devices using the VID*PID, UUID or label, and then create a
```
superm1, line 18: uuid and label are gone now

plugins/uf2/README.md
```diff
@@ -0,0 +1,61 @@
...
+UF2 device which can be used to flash firmware.
+
+Note: We only read metadata from allow-listed IDs to avoid causing regressions
+on non-UF2 volumes. To get the UUID and label you can use commands like:
+
+    udisksctl info -b /dev/sda1
+
+The UF2 format is specified here: <https://github.com/Microsoft/uf2>
```
superm1, line 26: why not use markdown syntax instead?

plugins/uf2/uf2.quirk
```diff
@@ -0,0 +1,156 @@
...
+
+# Adafruit Feather M0 Express
+[USB\VID_239A&PID_001B]
+Flags = updatable
+
+# nRF52840 MDK
+[USB\VID_239A&PID_0029]
+Flags = updatable
+
+[UF2\FAMILY_16573617]
```
superm1, line 26: where did all the family values come from?  Maybe need a registry import helper script that is part of the release process.

plugins/uf2/README.md
hughsie: I kept UUID as a postfix to one of the instance IDs. I've removed label.

plugins/uf2/README.md
hughsie: Fixed.

plugins/uf2/uf2.quirk
hughsie: From here: https://github.com/microsoft/uf2/blob/master/utils/uf2families.json -- it's not often updated, and it's only really NTH. I've also fixed-up some of the vendor/names as we have two properties rather than just a string. I'll add that as a comment, but I'm not sure it's worth the pain to make it auto-update.

PR #4097, opened 2021-12-16
uefi-capsule: Check for OsIndications before enabling CapsuleOnDisk

hughsie:
It shouldn't be possible to have `FILE_CAPSULE_DELIVERY_SUPPORTED` set
in `OsIndicationsSupported` without a valid `OsIndications`, but here
we are. I suspect Windows 10 is checking in the same way.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -678,6 +678,16 @@ fu_plugin_uefi_capsule_check_cod_support(GError **error)
 	guint64 value = 0;
 	g_autofree guint8 *buf = NULL;
 
+	/* we don't care what the data is, but we need it to exist */
+	if (!fu_efivar_get_data(FU_EFIVAR_GUID_EFI_GLOBAL,
+				"OsIndications",
+				NULL,
+				NULL,
+				NULL,
+				error)) {
+		g_prefix_error(error, "failed to read EFI variable: ");
+		return FALSE;
+	}
 	if (!fu_efivar_get_data(FU_EFIVAR_GUID_EFI_GLOBAL,
 				"OsIndicationsSupported",
 				&buf,


hughsie:
I'd appreciate the comment from @apalos --  I know some ARM machines should write to OsIndications, but I want to make sure it still exists for reading.

apalos:
> I'd appreciate the comment from @apalos -- I know some ARM machines should not write to OsIndications, but I want to make sure it still exists for reading.

Hi @hughsie, I don't this is needed.  OsIndicationSupported is set by the firmware (on every boot and is RO) and indicates *to* the OS what the firmware supports.

OTOH OsIndications is set by the OS request specific features.

xypron:
> I'd appreciate the comment from @apalos -- I know some ARM machines should not write to OsIndications, but I want to make sure it still exists for reading.

U-Boot does not support SetVariable() at runtime. This is not specific to ARM but also valid for RISC-V and amd64. OsIndicationsSupported is supplied by U-Boot.

hughsie:
@xypron I know it's not writeable, but does it *exist*?

xypron:
> @xypron I know it's not writeable, but does it _exist_?

If the user set it manually, yes. Otherwise no. It is not owned by the firmware.

ardbiesheuvel:
I'm not sure I follow the claim in the commit log: why would it depend on OsIndications whether or not some bit may be set in OsIndicationsSupported? Doesn't the latter simply define which bits the OS is permitted to set in the former? Or is the sentence just the wrong way around?

hughsie:
Hmm, so not only is this patch probably not needed, it probably also breaks the ARM CoD functionality. Maybe I just need to go shout at the OEM who's enabling FILE_CAPSULE_DELIVERY_SUPPORTED on a device that doesn't support CoD.

hughsie:
Also, my boss mentioned that Windows 10 doesn't support CoD, which is why the OEM hasn't cared about this until now. Is that plausible?

apalos:
> Hmm, so not only is this patch probably not needed, it probably also breaks the ARM CoD functionality. Maybe I just need to go shout at the OEM who's enabling FILE_CAPSULE_DELIVERY_SUPPORTED on a device that doesn't support CoD.

Right when we committed this we reasoned that the firmware is sane.  So we made capsule update on-disk the default, since we can check the OsIndicationsSupported and decide.  

The firmware vendor should most definitely disable this if  he doesn't support it.

hughsie:
> The firmware vendor should most definitely disable this if he doesn't support it.

Okay, indeed. Thanks all!

Issue #4099, opened 2021-12-16
json-glib version too old in snap package of 1.7.3


PhilippWendler:
**Describe the bug**
`fwupdmgr security --force` from snap outputs the following message instead of the desired output:
```
WARNING: This package has not been validated, it may not work properly.
json-glib version too old
```

**Steps to Reproduce**
```
sudo fwupd.fwupdmgr security --force
```

**Expected behavior**
The snap package should bundle the library in the required version such that all features of fwupd work.

**fwupd version information**
1.7.3 (snap package `fwupd` from `richard-hughes`)
Same problem with the `1.7.3-30-gdb94a995` version from the `--edge` snap.

**fwupd device information**
I don't think this is relevant here, if you still require it I will post it.

**Additional questions**

- Operating system and version: *Ubuntu 21.10*
- Have you tried rebooting? *no*
- Is this a regression? *Yes, when installing 1.6.2 with the `--candidate` snap it works.*

superm1:
The snap is built from core18 (Ubuntu 18.04 archives), so it grabs json-glib 1.4.2, but as of https://github.com/fwupd/fwupd/commit/dbdd3740a34f8de07ec3a988308ce8ea9c4f3508 it's not a 1.6.0 dependency.

Even bumping up to core20 that won't help the situation as it's still 1.4.4.  So I think the only way to get around this would be to pull a newer json-glib as part of the snap build process instead of the one from the Ubuntu archives.

PR #4100, opened 2021-12-16
snap: use newer json-glib (Fixes: #4099)

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/redfish/fu-plugin-redfish.c
+++ b/plugins/redfish/fu-plugin-redfish.c
@@ -272,10 +272,11 @@ fu_redfish_plugin_ipmi_create_user(FuPlugin *plugin, GError **error)
 	fu_redfish_backend_set_password(data->backend, password_new);
 
 	/* success */
-	if (!fu_plugin_set_config_value(plugin, "Username", username_fwupd, error))
+	if (!fu_plugin_set_secure_config_value(plugin, "Username", username_fwupd, error))
 		return FALSE;
-	if (!fu_plugin_set_config_value(plugin, "Password", password_new, error))
+	if (!fu_plugin_set_secure_config_value(plugin, "Password", password_new, error))
 		return FALSE;
+
 	return TRUE;
 }
 #endif
--- a/contrib/snap/fwupd-command
+++ b/contrib/snap/fwupd-command
@@ -17,8 +17,8 @@ else
 fi
 
 # don't update between versions, we want to preserve previous data
-[ ! -d "$SNAP_USER_DATA/etc" ] && cp -R "$SNAP/etc" "$SNAP_USER_DATA"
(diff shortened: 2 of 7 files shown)


hughsie:
Concept yes please, but CI unhappy: 

    meson.build:1:0: ERROR: Meson version is 0.54.0 but project requires >= 0.55.3

I wonder what we're doing that needs the new json-glib...

hughsie:
> I wonder what we're doing that needs the new json-glib...

I think https://github.com/fwupd/fwupd/pull/4101 "fixes" the bug, but I also think we need this PR too.

hughsie (dismissed):
If CI likes it, so do I. Thanks!

superm1:
> Ohh we need this; the file has usernames and passwords and can't be readable by users.

Hmm, not the default one though in the snap, that's only after it's been modified by the users.  So will need to think about how to solve this considering that.

hughsie:
> Hmm, not the default one though in the snap

Ohh? Maybe we want to do a manual `g_chmod()` after writing the password data? I'm not sure if that's going to make rpm upset, but I think it's okay if it's marked confignoreplace.

superm1:
> > Hmm, not the default one though in the snap
> 
> Ohh? Maybe we want to do a manual `g_chmod()` after writing the password data? I'm not sure if that's going to make rpm upset, but I think it's okay if it's marked confignoreplace.

I think the problem ends up being that someone can modify it by hand.  I suspect doing a fixup in the snap build scripts is better.

superm1:
> > > Hmm, not the default one though in the snap
> > 
> > 
> > Ohh? Maybe we want to do a manual `g_chmod()` after writing the password data? I'm not sure if that's going to make rpm upset, but I think it's okay if it's marked confignoreplace.
> 
> I think the problem ends up being that someone can modify it by hand. I suspect doing a fixup in the snap build scripts is better.

oh but the application programmaticaly sets it.  Let me do the `g_chmod` thing.

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -276,6 +276,9 @@ fu_redfish_plugin_ipmi_create_user(FuPlugin *plugin, GError **error)
 		return FALSE;
 	if (!fu_plugin_set_config_value(plugin, "Password", password_new, error))
 		return FALSE;
+	if (!fu_plugin_set_secure_config_file(plugin, error))
```
hughsie, line 279: Maybe we want a `fu_plugin_set_config_value_secure()` that does this atomically? It seems kinda risky to have a window of time where the file is readable by non-root with a password in.

superm1: That's true.  Will swap around.

hughsie (approved):

PR #4104, opened 2021-12-17

Sean-StarLabs:
Remove "number" version format to allow different versioning (i.e. 7.4 and 8)


hughsie:
Are you sure you want plain? that'll do a strcmp, e.g. compare the chars in `7.4` and `8` I think that works, but I wanted to be  sure.

superm1:
> Are you sure you want plain? that'll do a strcmp, e.g. compare the chars in `7.4` and `8` I think that works, but I wanted to be sure.

also, please consider the comment from before about what happens to older tools + newer firmware or newer tools + older firmware.  You might be causing a breakage if you change this.

Sean-StarLabs:
Plain works perfectly for what I need to do, and yes, it will break older coreboot firmwares - that's not an issue :)

PR #4117, opened 2021-12-21
logitech-hidpp: Always hardcode the vendor for FuLogitechHidPpDevice

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
I don't see a problem with this generally BUT I suggest you confirm Logitech doesn't OEM certain products for other companies. I don't know.

hughsie:
@mviredaz any insight?

PR #4118, opened 2021-12-21
dfu: Always set a summary line for DFU peripherals

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


plugins/dfu/fu-dfu-device.c
```diff
@@ -1908,6 +1908,7 @@ fu_dfu_device_init(FuDfuDevice *self)
 	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_ADD_COUNTERPART_GUIDS);
 	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
 	fu_device_set_remove_delay(FU_DEVICE(self), FU_DEVICE_REMOVE_DELAY_RE_ENUMERATE);
+	fu_device_set_summary(FU_DEVICE(self), "DFU peripheral");
```
superm1, line 1911: I see webcams that are internal with DFU endpoints. Do you consider that a peripheral?  I don't.

hughsie: True. Maybe just DFU device then? Or is that completely redundant? "Device firmware update device" is correct but weird.

superm1: I'd say just don't bother with a summary if you can't add anything to what the title does.

PR #4122, opened 2021-12-21
[thunderbolt] set install flags for thunderbolt retimer

krankun:
-PD gets suspended when ports go offline for retimer
-hence removing FWUPD_DEVICE_FLAG_REQUIRE_AC
-after install de-enumerate of retimers is issued after port goes back to online
-hence added FWUPD_DEVICE_FLAG_SKIPS_RESTART to avoid auto remove

TEST=emerge-volteer fwupd

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


krankun:
@superm1  @hughsie @djcampello @RajmohanMani here is the fix for the issues encountered in further validation,
the usb4_port1/offline suspends PD so the AC wont be available at the time of install , 

Also we found that at the start up the demon takes port offline this causes PD suspend too , the re timers are probed from https://github.com/fwupd/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L558 at the start up, we would like to avoid it and possibly invoke fu_thunderbolt_probe_retimer at https://github.com/fwupd/fwupd/blob/main/plugins/thunderbolt/fu-thunderbolt-device.c#L431 at the time of install

superm1 (dismissed):

krankun:
@superm1  Thank you , could you please merge too, 

I have an offline question on how the demon and auto update works in real case on chromium , will the demon run all the time a chromeos is booted ? CC: @djcampello

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -580,6 +580,8 @@ fu_thunderbolt_device_setup_retimer(FuDevice *device, GError **error)
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_DUAL_IMAGE);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_INTERNAL);
 	fu_device_add_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE);
+	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_SKIPS_RESTART);
```
hughsie, line 583: is this true of all tbt devices, or just retimers?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -580,6 +580,8 @@ fu_thunderbolt_device_setup_retimer(FuDevice *device, GError **error)
...
+	fu_device_remove_flag(device, FWUPD_DEVICE_FLAG_REQUIRE_AC);
```
hughsie, line 584: what added it in the first place? I'd rather not add it in the first place rather than add it and take it away again.

plugins/thunderbolt/fu-thunderbolt-device.c
krankun: this is needed just for re timers

plugins/thunderbolt/fu-thunderbolt-device.c
krankun: @hughsie this flags are inherited for all TBT devices , I adjusted this flags in fu_thunderbolt_device_setup_retimer 
added FWUPD_DEVICE_FLAG_SKIPS_RESTART as the restart is causing the retimers to de-enumerate from sysfs after nvm update { this leaves ports offline for ever} and wont properly exit.

I have removed FWUPD_DEVICE_FLAG_REQUIRE_AC since when ports go offline the PD gets suspended hence AC wont be available.

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: But this is going to be set for all thunderbolt devices, no? i.e. not just retimers.

hughsie: Ohh, hmm. This is because FuThunderboltDevice is being used for both rather than splitting them up into two different GTypes. I think given the functionality is diverging we probably want to do that sooner rather than later.

hughsie: I'll work on a WIP to do the split.

hughsie:
Alternate proposal in https://github.com/fwupd/fwupd/pull/4135 -- 100% untested for the retimer hardware

krankun:
> can you do this on top of #4135 please -- I think that's what we want to work towards

sure we have two changes now in this PR , we will work on top of #4135 ,

krankun:
> > can you do this on top of #4135 please -- I think that's what we want to work towards
> 
> sure we have two changes now in this PR , we will work on top of #4135 , cc: @djcampello

since now the CL has been merged [e03b8c46ed96](https://github.com/fwupd/fwupd/commit/e03b8c46ed963ff5e952697ab2a9c5b73b4087c9)

we wont need this PR. however the nvm updates with fwupd tool have an issue here is the complete log for debug 
[debug_retimer_nvm_update.txt](https://github.com/fwupd/fwupd/files/7850780/debug_retimer_nvm_update.txt)

fwupdmgr install /home/root/retimer-2.21.cab --allow-reinstall --force --verbose
(fwupdmgr:4275): GLib-DEBUG: 16:50:10.110: setenv()/putenv() are not thread-safe and should not be used after threads are created
(fwupdmgr:4275): GLib-GIO-DEBUG: 16:50:10.110: _g_io_module_get_default: Found default implementation dummy (GDummyProxyResolver) for ?gio-proxy-resolver?
(fwupdmgr:4275): Fwupd-DEBUG: 16:50:10.115: Emitting ::status-changed() [idle]
(fwupdmgr:4275): Fwupd-DEBUG: 16:50:10.120: Emitting ::status-changed() [decompressing]
Decompressing?           [ -                                     ](fwupdmgr:4275): Fwupd-DEBUG: 16:50:10.134: Emitting ::status-changed() [idle]
No supported devices found

@hughsie  @superm1

PR #4125, opened 2021-12-24
nvme: ignore NVMe-over-Fabrics devices

ematsumiya:
Ignore NVMe-over-Fabrics devices when probing. Otherwise, logs might be
flooded with error messages:

03:10:53:0251 FuEngine             failed to add device /sys/devices/virtual/nvme-fabrics/ctl/nvme1: failed to find device with subsystems pci, only got nvme,nvme-fabrics

These devices are not supported as they are not on the PCI subsystem.

Signed-off-by: Enzo Matsumiya <ematsumiya@suse.de>

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/nvme/fu-nvme-device.c
+++ b/plugins/nvme/fu-nvme-device.c
@@ -267,6 +267,31 @@ fu_nvme_device_dump(const gchar *title, const guint8 *buf, gsize sz)
 	g_print("\n");
 }
 
+/*
+ * Returns:
+ * %TRUE: device is in PCI subsystem
+ * %FALSE: device is, probably, NVMe-over-Fabrics
+ */
+static gboolean
+fu_nvme_device_is_pci(FuDevice *device, GError **error)
+{
+	g_autoptr(GUdevDevice) device_tmp = NULL;
+	GUdevDevice *gdev;
+
+	gdev = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+
+	device_tmp = g_udev_device_get_parent_with_subsystem(gdev, "pci", NULL);
+	if (device_tmp == NULL) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "device is not on PCI subsystem");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
 static gboolean
 fu_nvme_device_probe(FuDevice *device, GError **error)
 {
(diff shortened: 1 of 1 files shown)


plugins/nvme/fu-nvme-device.c
```diff
@@ -267,6 +267,29 @@ fu_nvme_device_dump(const gchar *title, const guint8 *buf, gsize sz)
...
+
+	gdev = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+
+	device_tmp = g_udev_device_get_parent_with_subsystem(gdev, "nvme-fabrics", NULL);
+	if (device_tmp != NULL) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "device is NVMe-over-Fabrics");
+		return TRUE;
```
superm1, line 287: It's unusual for return true on the failure scenario in this project. Could you instead swap the logic around?

Maybe just match for pci in this new function and if it's anything else fail?

plugins/nvme/fu-nvme-device.c
```diff
@@ -280,6 +303,10 @@ fu_nvme_device_probe(FuDevice *device, GError **error)
 	if (g_strcmp0(fu_device_get_vendor(FU_DEVICE(device)), "Samsung Electronics Co Ltd") == 0)
 		fu_device_set_vendor(FU_DEVICE(device), "Samsung");
 
+	/* ignore NVMe-over-Fabrics devices */
+	if (fu_nvme_device_is_fabrics(device, error))
```
superm1, line 307: Usually we match the fail here in this project.

superm1 (changes requested):

plugins/nvme/fu-nvme-device.c
ematsumiya: Thanks. I force pushed an amend with those changes. Please review.

plugins/nvme/fu-nvme-device.c
```diff
@@ -267,6 +267,33 @@ fu_nvme_device_dump(const gchar *title, const guint8 *buf, gsize sz)
...
+ * %TRUE: device is in PCI subsystem
+ * %FALSE: device is, probably, NVMe-over-Fabrics
+ */
+static gboolean
+fu_nvme_device_is_pci(FuDevice *device, GError **error)
+{
+	g_autoptr(GUdevDevice) device_tmp = NULL;
+	GUdevDevice *gdev;
+
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
```
superm1, line 281: is this strictly needed?  _Sometimes_ we do like to send `NULL` for the error argument to functions if we really don't care about the details of the outcome.

hughsie: In general we don't do these error checks for static functions. I'd say just remove the line then LGTM.

ematsumiya: Not strictly needed. While testing my patches, I was getting a segfault on error being dereferenced and I was checking whether it was me not setting error or if it was being derefereced without NULL checks. Turns out it was both. I should send a new patch for the second case though.
For now, I'll remove this line so we can go on.
Thanks.

hughsie (approved):

superm1 (approved):

PR #4126, opened 2021-12-24
trivial: incorporate version format if metadata tells

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -4299,6 +4299,7 @@ fu_engine_get_result_from_component(FuEngine *self,
 		if (guid == NULL)
 			continue;
 		device = fu_device_list_get_by_guid(self->device_list, guid, NULL);
+		fu_device_set_internal_flags(dev, FU_DEVICE_INTERNAL_FLAG_MD_SET_VERFMT);
 		if (device != NULL) {
 			fu_device_set_name(dev, fu_device_get_name(device));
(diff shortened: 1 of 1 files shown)


CragW:
Presently internal flag `FU_DEVICE_INTERNAL_FLAG_MD_SET_VERFMT` is added only [if the deice is found](https://github.com/fwupd/fwupd/blob/e0f7087bfb288a69687dc1f5255cae92cc55af90/src/fu-engine.c#L4305) and normally [is a uefi device](https://github.com/fwupd/fwupd/blob/e0f7087bfb288a69687dc1f5255cae92cc55af90/plugins/uefi-capsule/fu-uefi-device.c#L644).  This patch adds the internal flag to support runtime command `get-details foo.cab` reading metadata's version format so as to print the version properly.

hughsie:
I'm not 100% sure this is *safe* -- we have to check the cab archive is trusted before we can copy properties from it.

CragW:
> I'm not 100% sure this is _safe_ -- we have to check the cab archive is trusted before we can copy properties from it.

Integrity check before applying all the properties described in foo.cab or just before the `version format` exclusively?

I thought this change has limited impact (I might be wrong here) as only `get-details` uses it, purposed to print it more precisely if device absence. On the contrary, if a device is found on the system (or connected peripheral), the `version format` for the device and silo's release will be overridden by the physical device right away.

PR #4134, opened 2021-12-29
Do not allow UEFI updates when the lid is closed

hughsie:
Most vendors do not mirror the firmware update to an external display,
and some don't behave correctly when the lid is shut and the machine is
docked.

Fixes https://github.com/fwupd/firmware-lenovo/issues/181

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/upower/fu-plugin-upower.c
+++ b/plugins/upower/fu-plugin-upower.c
@@ -10,6 +10,7 @@
 
 struct FuPluginData {
 	GDBusProxy *proxy; /* nullable */
+	GDBusProxy *proxy_manager; /* nullable */
 };
 
 static void
@@ -24,10 +25,12 @@ fu_plugin_upower_destroy(FuPlugin *plugin)
 	FuPluginData *data = fu_plugin_get_data(plugin);
 	if (data->proxy != NULL)
 		g_object_unref(data->proxy);
+	if (data->proxy_manager != NULL)
+		g_object_unref(data->proxy_manager);
 }
 
 static void
-fu_plugin_upower_rescan(FuPlugin *plugin)
+fu_plugin_upower_rescan_devices(FuPlugin *plugin)
 {
 	FuContext *ctx = fu_plugin_get_context(plugin);
 	FuPluginData *data = fu_plugin_get_data(plugin);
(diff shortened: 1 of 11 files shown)


plugins/upower/fu-plugin-upower.c
```diff
@@ -62,13 +65,41 @@ fu_plugin_upower_rescan(FuPlugin *plugin)
...
 
+static void
+fu_plugin_upower_rescan_manager(FuPlugin *plugin)
+{
+	FuContext *ctx = fu_plugin_get_context(plugin);
+	FuPluginData *data = fu_plugin_get_data(plugin);
+	g_autoptr(GVariant) lid_is_closed = NULL;
+	g_autoptr(GVariant) lid_is_present = NULL;
+
+	/* check that we "have" a battery */
```
superm1, line 76: Lid

superm1 (commented):
To be honest - this feels like a bandaid for a problem that OEMs should fix. It effectively means that for anything quirked to use this behavior that firmware updates docked won't work for some people.

Like think of a situation that you get Ethernet from your dock but you always dock closed..

plugins/uefi-capsule/fu-uefi-device.c
```diff
@@ -739,6 +739,7 @@ fu_uefi_device_init(FuUefiDevice *self)
 {
 	fu_device_set_summary(FU_DEVICE(self), "UEFI ESRT device");
 	fu_device_add_protocol(FU_DEVICE(self), "org.uefi.capsule");
+	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_NO_LID_CLOSED);
```
superm1, line 742: I think you need to check with all the OEMs on LVFS before applying this to everyone. It may only make sense for certain ones.

plugins/upower/fu-plugin-upower.c
hughsie: Fixed.

plugins/uefi-capsule/fu-uefi-device.c
hughsie: I've restricted this to Lenovo as it's this that does entirely the wrong thing when the lid is closed. I'll check with Crag and the HP guys.

superm1 (approved):

PR #4143, opened 2022-01-04
Wip/hughsie/fu firmware add patch

hughsie:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-firmware.c
+++ b/libfwupdplugin/fu-firmware.c
@@ -34,6 +34,7 @@ typedef struct {
 	guint64 offset;
 	gsize size;
 	GPtrArray *chunks; /* nullable, element-type FuChunk */
+	GPtrArray *patches; /* nullable, element-type FuFirmwarePatch */
 } FuFirmwarePrivate;
 
 G_DEFINE_TYPE_WITH_PRIVATE(FuFirmware, fu_firmware, G_TYPE_OBJECT)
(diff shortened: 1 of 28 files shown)


libfwupdplugin/fu-firmware.c
```diff
@@ -1038,13 +1099,52 @@ fu_firmware_write(FuFirmware *self, GError **error)
...
+
+	/* ensure exists */
+	if (priv->patches == NULL)
+		priv->patches =
+		    g_ptr_array_new_with_free_func((GDestroyNotify)fu_firmware_patch_free);
+
+	/* find existing */
+	for (guint i = 0; i < priv->patches->len; i++) {
+		ptch = g_ptr_array_index(priv->patches, i);
+		if (ptch->offset == offset) {
```
superm1, line 1136: You need to make sure the new patch is the same length or you might be throwing away part of the old one.

hughsie: Absolutely correct, fixed. Do you think we need a `fu_firmware_remove_patch` (just taking the offset) too?

superm1: I don't know how important that is - users can just re-run to redo patches right?

hughsie: Right, just from a programmatic API point of view.

superm1: I don't see harm in it if you think something might use it some day.

The only use case I can think of is some vendor wants to ship patches inside a cab and the patches are applied by the software. If that's not a good idea then I would say no need.

hughsie: > The only use case I can think of

Ya, I couldn't think of a use case either; lets add it if/when we need it. Lets see if this works for the reporter.

superm1 (dismissed):
LGTM if this meets the needs from OP

PR #4145, opened 2022-01-04
Pull in some packaging changes from debian

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


contrib/debian/rules
```diff
@@ -70,6 +70,11 @@ override_dh_install:
 	rm -f debian/fwupd/usr/lib/*/fwupd-plugins-*/libfu_plugin_test_ble.so
 	rm -f debian/fwupd/usr/lib/*/fwupd-plugins-*/libfu_plugin_invalid.so
 	rm -f debian/fwupd/etc/fwupd/remotes.d/fwupd-tests.conf
+	mv debian/fwupd/usr/share/man/man1 debian/fwupd/usr/share/man/man8
+	for file in debian/fwupd/usr/share/man/man8/*.1; do \
+		file8=$${file%%.1}.8; \
+		mv -v $$file $$file8; \
```
hughsie, line 76: is this something we should do upstream?

superm1, line 76: the logic behind this is https://lintian.debian.org/tags/manual-page-for-system-command?version=2.104.244

superm1: I was wondering the same thing.  At least wanted to sync up the packaging for now, and can revert it if we change it upstream too.

hughsie: Hmm, not convinced e.g. `fwupdmgr` is designed to be run by end users, rather than sysadmins -- you don't even need root.

superm1: I guess that's true, I'm not sure why this lintian warning came up.

hughsie (dismissed):
Maybe squash down?

superm1:
> Maybe squash down?

I'll add trivial tags to them, but I want to keep them as individual commits to match where they came from in Debian for future merges.

hughsie (approved):

PR #4149, opened 2022-01-05
Wip/hughsie/no root

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -68,6 +68,7 @@ struct FuUtilPrivate {
 	gboolean prepare_blob;
 	gboolean cleanup_blob;
 	gboolean enable_json_state;
+	gboolean interactive;
 	FwupdInstallFlags flags;
 	gboolean show_all;
 	gboolean disable_ssl_strict;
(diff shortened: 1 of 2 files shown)


src/fu-tool.c
```diff
@@ -236,6 +237,15 @@ fu_util_start_engine(FuUtilPrivate *priv, FuEngineLoadFlags flags, GError **erro
 #ifdef HAVE_SYSTEMD
 	g_autoptr(GError) error_local = NULL;
 #endif
+
+#ifdef HAVE_GETUID
+	/* ensure root user */
+	if (priv->interactive && (getuid() != 0 || geteuid() != 0)) {
+		/* TRANSLATORS: we're poking around as a power user */
+		g_printerr("%s\n", _("This program may only work correctly as root"));
```
superm1, line 245: Just a thought - what if we move this to error path instead?

superm1 (approved):

src/fu-tool.c
hughsie: I thought about this a bit, but I think we want to warn *before* we throw up warnings on the console.

PR #4153, opened 2022-01-06
cros-ec: Add progress completion when writing blocks

hughsie:
Also, simplify the helper that's passed to the retry function.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/cros-ec/fu-cros-ec-firmware.c
+++ b/plugins/cros-ec/fu-cros-ec-firmware.c
@@ -51,9 +51,26 @@ fu_cros_ec_firmware_pick_sections(FuCrosEcFirmware *self, guint32 writeable_offs
 }
 
 GPtrArray *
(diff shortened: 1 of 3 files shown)


hughsie:
This now has a progress bar that goes from 0..100 in nice linear steps.

hughsie:
@bleungatchromium would it be okay to include public links to the Google Servo Micro Device Update v2.4.17 and v2.4.0 firmwares so I can also add a device test?

bleungatchromium:
> @bleungatchromium would it be okay to include public links to the Google Servo Micro Device Update v2.4.17 and v2.4.0 firmwares so I can also add a device test?

Of course! Do you want me to flip those to public in LVFS by the way, or we can just link to them as is?

hughsie:
Public would be best, but I don't know if your vendor group has the legal acl yet. I can just use the embargo ones if not.

superm1 (approved):

PR #4154, opened 2022-01-06
Increase the chances of success of WAIT_FOR_REPLUG on WIN32

hughsie:
Decrease the GUsb polling interval when any of the devices is in
`WAIT_FOR_REPLUG` on when running on Windows. Any device that can
re-enumerate much faster than the default 1000ms may be missed and the
detach may fail.

Linux doesn't have this problem as it has `LIBUSB_CAP_HAS_HOTPLUG`.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-usb-backend.c
+++ b/src/fu-usb-backend.c
@@ -20,6 +20,33 @@ struct _FuUsbBackend {
 
 G_DEFINE_TYPE(FuUsbBackend, fu_usb_backend, FU_TYPE_BACKEND)
 
+#define FU_USB_BACKEND_POLL_INTERVAL_DEFAULT	 1000 /* ms */
+#define FU_USB_BACKEND_POLL_INTERVAL_WAIT_REPLUG 5    /* ms */
+
+#ifdef _WIN32
+static void
+fu_usb_backend_device_notify_flags_cb(FuDevice *device, GParamSpec *pspec, FuBackend *backend)
+{
+#if G_USB_CHECK_VERSION(0, 3, 10)
+	FuUsbBackend *self = FU_USB_BACKEND(backend);
+
+	/* if waiting for a disconnect, set win32 to poll insanely fast -- and set it
+	 * back to the default when the device removal was detected */
+	if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG)) {
+		g_debug("setting USB poll interval to %ums to detect replug",
+			(guint)FU_USB_BACKEND_POLL_INTERVAL_WAIT_REPLUG);
+		g_usb_context_set_hotplug_poll_interval(self->usb_ctx,
+							FU_USB_BACKEND_POLL_INTERVAL_WAIT_REPLUG);
+	} else {
+		g_usb_context_set_hotplug_poll_interval(self->usb_ctx,
+							FU_USB_BACKEND_POLL_INTERVAL_DEFAULT);
+	}
+#else
+	g_warning("GUsb >= 0.3.10 may be needed to notice device enumeration");
+#endif
+}
+#endif
+
 static void
 fu_usb_backend_device_added_cb(GUsbContext *ctx, GUsbDevice *usb_device, FuBackend *backend)
 {
@@ -82,6 +109,22 @@ fu_usb_backend_coldplug(FuBackend *backend, GError **error)
 }
 
 static void
+fu_usb_backend_registered(FuBackend *backend, FuDevice *device)
+{
+#ifdef _WIN32
+	/* not required */
+	if (!FU_IS_USB_DEVICE(device))
+		return;
+
+	/* on win32 we need to poll the context faster */
+	g_signal_connect(FU_DEVICE(device),
+			 "notify::flags",
+			 G_CALLBACK(fu_usb_backend_device_notify_flags_cb),
+			 backend);
+#endif
+}
+
+static void
 fu_usb_backend_finalize(GObject *object)
 {
 	FuUsbBackend *self = FU_USB_BACKEND(object);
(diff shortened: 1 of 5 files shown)


hughsie:
Depends on the as-yet-unmerged https://github.com/hughsie/libgusb/pull/56

src/fu-usb-backend.c
```diff
@@ -20,6 +20,29 @@ struct _FuUsbBackend {
 
 G_DEFINE_TYPE(FuUsbBackend, fu_usb_backend, FU_TYPE_BACKEND)
 
+#define FU_USB_BACKEND_POLL_INTERVAL_DEFAULT	 1000 /* ms */
+#define FU_USB_BACKEND_POLL_INTERVAL_WAIT_REPLUG 5 /* ms */
+
+static void
+fu_usb_backend_device_notify_flags_cb(FuDevice *device, GParamSpec *pspec, FuBackend *backend)
+{
+#if G_USB_CHECK_VERSION(0, 3, 10)
```
superm1, line 29: I think the subproject needs to be updated toom

hughsie: right, if the libgusb PR goes in i'll tag a new release for simplicity

src/fu-usb-backend.c
```diff
@@ -20,6 +20,29 @@ struct _FuUsbBackend {
...
+	FuUsbBackend *self = FU_USB_BACKEND(backend);
+
+	/* if waiting for a disconnect, set win32 to poll insanely fast -- and set it
+	 * back to the default when the device removal was detected */
+	if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG)) {
+		g_debug("setting USB poll interval to %ums to detect replug",
+			(guint)FU_USB_BACKEND_POLL_INTERVAL_WAIT_REPLUG);
+		g_usb_context_set_hotplug_poll_interval(self->usb_ctx,
```
superm1, line 37: Shouldn't this only run on win32?

hughsie: it's a NOP on !WIN32, but maybe there's a valid argument to not do the `g_signal_connect` on !WIN32

superm1: Yeah sounds good.

src/fu-tool.c
```diff
@@ -3067,6 +3073,48 @@ fu_util_switch_branch(FuUtilPrivate *priv, gchar **values, GError **error)
...
+	}
+	dwMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
+	if (!SetConsoleMode(hOut, dwMode)) {
+		g_set_error(error,
+			    G_IO_ERROR,
+			    G_IO_ERROR_NOT_SUPPORTED,
+			    "failed to set mode [%u]",
+			    (guint)GetLastError());
+		return FALSE;
+	}
```
superm1, line 3112: This make Windows pretty?

hughsie: Gahh, ETOOMANYBRANCHES. Maybe. Will try in wip/hughsie/cros-ec-progress

src/fu-tool.c
```diff
@@ -3067,6 +3073,48 @@ fu_util_switch_branch(FuUtilPrivate *priv, gchar **values, GError **error)
 	return fu_util_prompt_complete(priv->completion_flags, TRUE, error);
 }
 
+static gboolean
+fu_util_setup_console(GError **error)
```
superm1, line 3077: What do you think about putting this in common?  It can better prepare if some day we get daemon working on Windows too (dbus for Windows is a thing after all).

src/fu-tool.c
```diff
@@ -3240,11 +3288,12 @@ main(int argc, char *argv[])
 	     NULL},
 	    {NULL}};
 
-#ifdef _WIN32
-	/* workaround Windows setting the codepage to 1252 */
-	g_setenv("LANG", "C.UTF-8", FALSE);
-#endif
-
+	/* do early */
+	if (!fu_util_setup_console(&error)) {
```
superm1, line 3292: Maybe merge this section in:
```
priv->interactive = isatty(fileno(stdout)) != 0;
	if (!priv->interactive) {
		priv->no_reboot_check = TRUE;
		priv->no_safety_check = TRUE;
	} else {
		/* set our implemented feature set */
		fu_engine_request_set_feature_flags(
		    priv->request,
		    FWUPD_FEATURE_FLAG_DETACH_ACTION | FWUPD_FEATURE_FLAG_SWITCH_BRANCH |
			FWUPD_FEATURE_FLAG_FDE_WARNING | FWUPD_FEATURE_FLAG_UPDATE_ACTION);
	}
```

superm1: but if that function goes to common then no.

src/fu-usb-backend.c
hughsie: Done.

src/fu-usb-backend.c
```diff
@@ -20,6 +20,33 @@ struct _FuUsbBackend {
...
+							FU_USB_BACKEND_POLL_INTERVAL_WAIT_REPLUG);
+	} else {
+		g_usb_context_set_hotplug_poll_interval(self->usb_ctx,
+							FU_USB_BACKEND_POLL_INTERVAL_DEFAULT);
+	}
+#else
+	g_warning("GUsb >= 0.3.10 may be needed to notice device enumeration");
```
superm1, line 45: this is going to be needlessly noisy for non-Win32.

superm1: Oh nvm- nested `#ifdef`, I missed it.

src/fu-tool.c
hughsie: Well, it makes the text multicoloured now. There's still some UTF-8 issue.

superm1 (dismissed):

superm1:
Make pre-commit happy and then LGTM.

src/fu-usb-backend.c
hughsie: Ohh the entire function is in `#ifdef _WIN32`

PR #4156, opened 2022-01-06
Wip/hughsie/win32 console

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -3080,6 +3080,7 @@ main(int argc, char *argv[])
 	gboolean ignore_vid_pid = FALSE;
 	g_auto(GStrv) plugin_glob = NULL;
 	g_autoptr(FuUtilPrivate) priv = g_new0(FuUtilPrivate, 1);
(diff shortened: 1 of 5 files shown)


hughsie:
Untitled

superm1 (dismissed):

src/fu-tool.c
```diff
@@ -3545,11 +3546,12 @@ main(int argc, char *argv[])
 	fu_util_cmd_array_sort(cmd_array);
 
 	/* non-TTY consoles cannot answer questions */
-	priv->interactive = isatty(fileno(stdout)) != 0;
```
superm1, line 3548: This also goes on fu-util right?

hughsie: Not sure I understand the question, sorry.

superm1: Don't we also have an interactive check there too?

hughsie: Ahh! Indeed we do. Fixed.

PR #4158, opened 2022-01-07
fu-tool: put lockfile in user's cache directory if no permissions for…

superm1:
… /run

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -186,12 +186,21 @@ fu_util_lock(FuUtilPrivate *priv, GError **error)
 	    .l_type = F_WRLCK,
 	    .l_whence = SEEK_SET,
 	};
-	g_autofree gchar *lockdir = NULL;
 	g_autofree gchar *lockfn = NULL;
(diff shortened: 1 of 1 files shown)


hughsie:
Do we want a lock file ever if we're running as non-root? It seems redundant, no?

superm1:
> Do we want a lock file ever if we're running as non-root? It seems redundant, no?

I don't think so - what if you have two udev hotplug rules that fired at the same time as a limited permissions user that only has access to certain devfs or sysfs files?

hughsie:
> what if you have two udev hotplug rules

Hm, maybe. I don't think that's totally valid as we'll have two processes, one per-user and one per-system trying to do the same thing with different locks held. I also think we should only use `fu_util_get_user_cache_path` if the UID is not zero, i.e. not fallback, but explicitly chose the _correct_ path.

hughsie (approved):

PR #4163, opened 2022-01-10
trivial: Show quirk matches when running verbose

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -51,6 +51,7 @@ typedef struct {
 	GHashTable *metadata; /* (nullable) */
 	GRWLock metadata_mutex;
 	GPtrArray *parent_guids;
+	GPtrArray *instance_ids_only_quirks; /* (nullable) */
 	GRWLock parent_guids_mutex;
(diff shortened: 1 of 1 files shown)


superm1:
So the idea is you set the flag on a device and set verbose and you get quirk matching stuff for the device? Why not just do it for all of them when in verbose or add a new flag for just that?

hughsie:
Ohh sorry, i was unclear. This is just for dumping to the console all the `ONLY_QUIRK` matches. At the moment when we do --verbose there's no way of seeing what quirks were applied and so it makes it tricky to debug. With this PR we set the env variable and then any instance IDs we add with `FU_DEVICE_INSTANCE_FLAG_ONLY_QUIRKS` get added to a special list and then get printed when we do the device `to_string()` fvunc. There should be no behaviour change.

superm1:
Right - but wouldn't it be useful to be able to see what happens with all quirks, not only the ONLY_QUIRKS ones?

hughsie:
Maybe https://github.com/fwupd/fwupd/pull/4165 is a better way to solve this?

PR #4165, opened 2022-01-11
Wip/hughsie/quirk quids

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-quirks.c
```diff
@@ -522,6 +523,7 @@ fu_quirks_load(FuQuirks *self, FuQuirksLoadFlags load_flags, GError **error)
 	g_return_val_if_fail(FU_IS_QUIRKS(self), FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 	self->load_flags = load_flags;
+	self->verbose = g_getenv("FWUPD_XMLB_VERBOSE") != NULL;
```
superm1, line 526: need to update documentation for the debugging variables to indicate this does more useful things now

superm1 (dismissed):

libfwupdplugin/fu-quirks.c
hughsie: Agree, fixed, thanks.

hughsie:
In case you were wondering, this is the counterpart commit that makes the libxmlb output actually useful: https://github.com/hughsie/libxmlb/pull/114

PR #4168, opened 2022-01-12
Show why devices are not marked as updatable

hughsie:
At the moment a lot of the failures are only visible when running the
daemon in verbose mode, and the inhibit functionalit provides us a way
to unset FWUPD_DEVICE_FLAG_UPDATABLE from multiple places, as well as
setting the update error for the user to see why.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-dell-dock-common.c
+++ b/plugins/dell-dock/fu-dell-dock-common.c
@@ -59,20 +59,3 @@ fu_dell_dock_will_replug(FuDevice *device)
 	fu_device_set_remove_delay(device, timeout * 1000);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
 }
-
-void
-fu_dell_dock_clone_updatable(FuDevice *device)
-{
-	FuDevice *parent;
-	parent = fu_device_get_parent(device);
-	if (parent == NULL)
-		return;
-	if (fu_device_has_flag(parent, FWUPD_DEVICE_FLAG_UPDATABLE)) {
-		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
-	} else {
-		const gchar *message = fu_device_get_update_error(parent);
-		if (message != NULL)
-			fu_device_set_update_error(device, message);
-		fu_device_remove_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
-	}
-}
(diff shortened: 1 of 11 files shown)


plugins/dell-dock/fu-dell-dock-common.c
```diff
@@ -71,8 +71,8 @@ fu_dell_dock_clone_updatable(FuDevice *device)
 		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
 	} else {
 		const gchar *message = fu_device_get_update_error(parent);
-		if (message != NULL)
-			fu_device_set_update_error(device, message);
-		fu_device_remove_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
+		if (message == NULL)
+			message = "Parent is not updatable";
+		fu_device_inhibit(device, "parent-not-updatable", message);
```
superm1, line 76: @CragW needs to review this more closely.

superm1 (approved):
LGTM, but want @CragW's blessing on dell-dock since it does some magic with cloning updatable that I want to make sure we didn't miss any nuance.

plugins/dell-dock/fu-dell-dock-common.c
hughsie: Hang on one minute; we have a flag to do all this automatically! `FU_DEVICE_INTERNAL_FLAG_INHIBIT_CHILDREN`

superm1: It's not just inhibiting children that matters but also setting up all the right ones to gain the updatable flag. If it's the inhibition route then I think that flag needs to be added by default to everything.

hughsie: Ahh, of course, because it's adding it too. Will see how it looks.

hughsie: It looks much better IMO. @CragW if you could test this PR on actual hardware I think we'd all feel more warm and fuzzy. Thanks!

superm1 (dismissed):

plugins/dell-dock/fu-dell-dock-common.c
hughsie: I've merged the uncontroversial parts to main, and the others we can test and merge post release. There are a few vendors who are nagging me for a tarball now.

CragW (approved):

PR #4170, opened 2022-01-12
Move root user ID check message into the error path for commands

superm1:
This should make it so that the message doesn't pop up for someone
who has used udev rules that match permissions to do

```
fwupdtool install-blob --plugins=dfu foo.bin
```

Unless there really was a failure.

@hughsie, I know you considered this before, but I wanted to give it another thought and see if we can overcome anything negative by doing it this way.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I think you're probably right fwiw.

hughsie (approved):

PR #4172, opened 2022-01-14
Fix the logic for removable fallback path in uefi-capsule

superm1:
The logic was built on the presumption of Ubuntu core which doesn't
have /etc/os-release.  Since other distributions do have /etc/os-release
adjust the logic to instead build the path explicitly.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-bootmgr.c
+++ b/plugins/uefi-capsule/fu-uefi-bootmgr.c
@@ -304,7 +304,7 @@ fu_uefi_bootmgr_bootnext(FuDevice *device,
 				device,
 				FU_UEFI_DEVICE_FLAG_FALLBACK_TO_REMOVABLE_PATH)) {
 				shim_app =
-				    fu_uefi_get_esp_app_path(device, esp_path, "boot", error);
+				    fu_uefi_get_fallback_app_path(device, esp_path, "boot", error);
 				if (shim_app == NULL)
 					return FALSE;
 			}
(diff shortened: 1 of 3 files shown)


plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -327,6 +327,7 @@ fu_uefi_bootmgr_bootnext(FuDevice *device,
 			}
 			use_fwup_path = FALSE;
 		} else if ((flags & FU_UEFI_BOOTMGR_FLAG_USE_SHIM_FOR_SB) > 0) {
+			g_debug("couldn't find shim binary at %s\n", shim_app);
```
hughsie, line 330: can this go into an inhibit perhaps? then we can see the message as a user.

superm1: I thought about that, but then we need an inotify on the directory to be able to restore it.  Since the partition is mounted and unmounted dynamically in some circumstances that is not a good idea

superm1: Could potentially just add it into the error message I suppose.

hughsie: > Could potentially just add it into the error message I suppose

Yup, this I think .

superm1: fixed

hughsie (approved):

PR #4179, opened 2022-01-15
trivial: show uefi_recovery message in get-plugins output

superm1:
By default it shows just a blank line right now.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1490,8 +1490,10 @@ fu_util_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		return NULL;
 	if (plugin_flag == FWUPD_PLUGIN_FLAG_USER_WARNING)
 		return NULL;
-	if (plugin_flag == FWUPD_PLUGIN_FLAG_REQUIRE_HWID)
-		return NULL;
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_REQUIRE_HWID) {
+		/* TRANSLATORS: Plugin is active only if hardware is found */
+		return _("Enabled if hardware matches");
+	}
 	if (plugin_flag == FWUPD_PLUGIN_FLAG_NONE) {
 		/* TRANSLATORS: Plugin is active and in use */
 		return _("Enabled");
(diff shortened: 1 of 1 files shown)


src/fu-util-common.c
```diff
@@ -1490,8 +1490,10 @@ fu_util_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		return NULL;
 	if (plugin_flag == FWUPD_PLUGIN_FLAG_USER_WARNING)
 		return NULL;
-	if (plugin_flag == FWUPD_PLUGIN_FLAG_REQUIRE_HWID)
-		return NULL;
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_REQUIRE_HWID) {
+		/* TRANSLATORS: Plugin is active only if hardware is found */
+		return _("Enabled if hardware is found");
```
hughsie, line 1495: "hardware matches" perhaps?

superm1: that or just make it "Enabled".  This detail might not be needed for the user.

src/fu-util-common.c
```diff
@@ -1490,9 +1490,8 @@ fu_util_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		return NULL;
 	if (plugin_flag == FWUPD_PLUGIN_FLAG_USER_WARNING)
 		return NULL;
-	if (plugin_flag == FWUPD_PLUGIN_FLAG_REQUIRE_HWID)
-		return NULL;
-	if (plugin_flag == FWUPD_PLUGIN_FLAG_NONE) {
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_NONE ||
+	    plugin_flag == FWUPD_PLUGIN_FLAG_REQUIRE_HWID) {
```
hughsie, line 1494: but that's not right, surely:

		g_debug("enabling %s due to HwId %s", plugins[i], hwid);
		fu_plugin_remove_flag(plugin, FWUPD_PLUGIN_FLAG_REQUIRE_HWID);

i.e. the engine removes the REQUIRE_HWID flag if the HWID matches, otherwise the plugin remains disabled.

superm1: Then - maybe the right answer is to let the engine run and then show enabled plugins.

hughsie: > let the engine run

I'm pretty sure that won't be useful; that'll take a few *seconds* and we want results in ms for user interactions...

src/fu-util-common.c
```diff
@@ -1548,9 +1550,8 @@ fu_util_plugin_flag_to_cli_text(FwupdPluginFlags plugin_flag)
 	case FWUPD_PLUGIN_FLAG_UNKNOWN:
 	case FWUPD_PLUGIN_FLAG_CLEAR_UPDATABLE:
 	case FWUPD_PLUGIN_FLAG_USER_WARNING:
-	case FWUPD_PLUGIN_FLAG_REQUIRE_HWID:
-		return NULL;
 	case FWUPD_PLUGIN_FLAG_NONE:
+	case FWUPD_PLUGIN_FLAG_REQUIRE_HWID:
```
hughsie, line 1554: that's changing the "falling through" behaviour, no?

hughsie (approved):

Issue #3652, opened 2021-08-19
Failed to update metadata for lvfs on Kubuntu 21.04


shmu26:
**Describe the bug**
When I run
`fwupdmgr refresh`
It fails with the message
`Updating lvfs
Downloading…             [ -                        Downloading…             [ -                        Downloading…             [***************************************]
Downloading…             [**************************Downloading…             [***************************************]
Failed to update metadata for lvfs: checksum failure: failed to verify data, expected 98261db7124c8026fb88f77768787efc2f8355c3
`

**Steps to Reproduce**
Run the above-mentioned command, or open "Discovery" in Kubuntu or the Gnome-software in Ubuntu.

**Expected behavior**
The metadata should be updated

**fwupd version information**
Please provide the version of the daemon and client.

```shell
client version: 1.5.11
compile-time dependency versions
        gusb:   0.3.5

daemon version: 1.5.11

```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
It was bundled with the distro

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
fwupdmgr get-devices --show-all-devices
System Product Name
│
├─Intel(R) Core™ i7-9700 CPU @ 3.00GHz:
│     Device ID:          4bde70ba4e39b28f9eab1628f9dd6e6244c03027
│     Current version:    0x000000ea
│     Vendor:             Intel
│     GUIDs:              b9a2dd81-159e-5537-a7db-e7101d164d3f
│                         30249f37-d140-5d3e-9319-186b1bd5cac3
│                         809a0b93-8a12-5338-a571-ad5583acf896
│                         50a811ae-a8fd-5cd0-90f4-33583974b789
│     Device Flags:       • Internal device
│   
├─KINGSTON SA400S37480G:
│     Device ID:          5dbeb140337f610d54913c0f43d15fdaa0eafb51
│     Summary:            ATA Drive
│     Current version:    SBFKK1B3
│     Vendor:             Kingston (ATA:0x2646, OUI:0026b7)
│     GUIDs:              d8e02563-67f0-5da4-b388-e3b18a5f0dd3
│                         1ab2d344-65ac-5bc5-adbd-9aab1dfc2be3
│                         b4b189f9-ed32-5438-a5e4-d3313483647a
│     Device Flags:       • Internal device
│                         • Updatable
│                         • System requires external power source
│                         • Needs a reboot after installation
│                         • Needs shutdown after installation
[...]

PR #4182, opened 2022-01-17
trivial: given https://github.com/fwupd/fwupd/issues/3652, it's seems…

ycheng:
trivial: put a condition on libjcat version to make the story complete.


superm1 (changes requested):
We don't have a true runtime dependency on the newer version, it's just to avoid a particular bug.

I don't think this is a good idea to do in the runtime fields for packaging TBH.  It means that if you happen to have an older version you compile against you can't install the packages generated from that source package.

So if we're deciding to make this change in the upstream packaging the right solution here is that the **build** dependencies needs this change too.

superm1:
BTW don't worry about the CI failure compiling against uefi capsule plugin - that isn't your fault.  https://github.com/fwupd/fwupd/pull/4184 is tracking the fix for it.

superm1:
> We don't have a true runtime dependency on the newer version, it's just to avoid a particular bug.
> 
> I don't think this is a good idea to do in the runtime fields for packaging TBH. It means that if you happen to have an older version you compile against you can't install the packages generated from that source package.
> 
> So if we're deciding to make this change in the upstream packaging the right solution here is that the **build** dependencies needs this change too.

BTW `contrib/ci/dependencies.xml`.  Add blocks like this:

```
      <control>
        <version>(>= 0.1.4)</version>
      </control>
```

hughsie:
> the build dependencies needs this change too

I'm not against this if all the CI targets have a new enough version. It sure makes the QA matrix smaller.

superm1:
> > the build dependencies needs this change too
> 
> I'm not against this if all the CI targets have a new enough version. It sure makes the QA matrix smaller.

It looks like it won't happen until at least 5 days from now.

```
∙ ∙ Too young, only 2 of 7 days old
```

ycheng:
IMHO, If we close the mentioned bug because we think libjcat has a bug, then we should just stop people from using it, or there will be an increase in the service loading.

It's not an API level runtime dependency, however, the bug in libjcat causes trouble, so we just recommend using a new version.

People can choose to ignore our recommendation, but they need to take their own risk.

superm1:
I don't see a problem to make the packaging have this check but to bump the dependency we need CI to pass. So it needs a few days to promote new libjcat to debian testing from debian unstable so CI has it.

ycheng:
btw, https://bugs.launchpad.net/ubuntu/+source/fwupd/+bug/1943833/comments/17

hughsie:
> btw, https://bugs.launchpad.net/ubuntu/+source/fwupd/+bug/1943833/comments/17

I think you'll need to rebuild fwupd against that new libcat version if you want to see the fix.

superm1:
The thing is they build against 0.1.3 which is what the macro  in fwupd checks.

Why would 0.1.4 building matter?

hughsie:
> The thing is they build against 0.1.3 which is what the macro in fwupd checks

Ahh then I'm wrong, I though the LIBJCAT_CHECK_VERSION was 0.1.4.

superm1:
Okay so we need to open back up that other issue then, there appears to likely still be a bug happening.

Issue #4180, opened 2022-01-16
Vaydeer Multimedia Console doesn't work when fwupd.service is running


rjahanbakhshi:
**Describe the bug**
If I connect my Vaydeer Multimedia Console when fwupd.service is running, the device driver is correctly bound and immediately removed. If fwupd.service is not running, the driver stays bound and the device works correctly. 
Here's the systemd journal output when the problem occurs:
```Jan 16 13:03:54 rj-pc kernel: usb 1-6: new full-speed USB device number 11 using xhci_hcd
Jan 16 13:03:55 rj-pc kernel: usb 1-6: New USB device found, idVendor=0483, idProduct=5750, bcdDevice= 2.00
Jan 16 13:03:55 rj-pc kernel: usb 1-6: New USB device strings: Mfr=1, Product=2, SerialNumber=3
Jan 16 13:03:55 rj-pc kernel: usb 1-6: Product: Vaydeer Multimedia Console
Jan 16 13:03:55 rj-pc kernel: usb 1-6: Manufacturer: Vaydeer
Jan 16 13:03:55 rj-pc kernel: usb 1-6: SerialNumber: 4EA2335F5000
Jan 16 13:03:55 rj-pc kernel: input: Vaydeer Vaydeer Multimedia Console Keyboard as /devices/pci0000:00/0000:00:01.2/0000:20:00.0/0000:21:08.0/0000:2a:00.1/usb1/1-6/1-6:1.0/0003:0483:5750.001D/input/input50
Jan 16 13:03:55 rj-pc mtp-probe[456400]: checking bus 1, device 11: "/sys/devices/pci0000:00/0000:00:01.2/0000:20:00.0/0000:21:08.0/0000:2a:00.1/usb1/1-6"
Jan 16 13:03:55 rj-pc kernel: hid-generic 0003:0483:5750.001D: input,hidraw19: USB HID v1.11 Keyboard [Vaydeer Vaydeer Multimedia Console] on usb-0000:2a:00.1-6/input0
Jan 16 13:03:55 rj-pc mtp-probe[456400]: bus: 1, device: 11 was not an MTP device
Jan 16 13:03:55 rj-pc fwupd[456230]: 12:03:55:0481 FuEngine             failed to add device usb:01:00:06: failed to send to device on ep 0x01: endpoint stalled or request not supported
Jan 16 13:03:55 rj-pc systemd-logind[981]: Failed to open /dev/input/event263: No such file or directory
[...]

PR #4183, opened 2022-01-17
trivial: Do not match the legacy bootloader VID/PID

hughsie:
This is the ID assigned to STMicroelectronics... more specifically for
a mini LED display. It also seems to be used by other vendors who also
did not change the default from the devkit, e.g. the Vaydeer Multimedia
Console.

Fixes https://github.com/fwupd/fwupd/issues/4180

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Can we maybe move it into a noinst quirk file with some comments pointing back to these links? 
 Then it gives anyone stuck in that connundrum some more information about what is going on and they can manually grab it.

hughsie:
@superm1 something like that^^?

superm1 (approved):

hughsie:
Backported to 1_4_X, 1_5_X and 1_6_X

Issue #4181, opened 2022-01-16
plugins/uefi-capsule with efivars-38: fwupd-1.7.4/plugins/uefi-capsule/fu-uefi-common.h:32:3: error: conflicting types for ‘efi_time_t’; have ‘struct <anonymous>’


Polynomial-C:
This has been reported downstream at Gentoo: https://bugs.gentoo.org/831327

```
In file included from ../fwupd-1.7.4/plugins/uefi-capsule/fu-plugin-uefi-capsule.c:19:
../fwupd-1.7.4/plugins/uefi-capsule/fu-uefi-common.h:32:3: error: conflicting types for ‘efi_time_t’; have ‘struct <anonymous>’
   32 | } efi_time_t;
      |   ^~~~~~~~~~
In file included from /usr/include/efivar/efivar.h:21,
                 from ../fwupd-1.7.4/plugins/uefi-capsule/fu-uefi-bootmgr.h:12,
                 from ../fwupd-1.7.4/plugins/uefi-capsule/fu-plugin-uefi-capsule.c:17:
/usr/include/efivar/efivar-types.h:81:3: note: previous declaration of ‘efi_time_t’ with type ‘efi_time_t’
   81 | } efi_time_t __attribute__((__aligned__(1)));
      |   ^~~~~~~~~~
```
Offending efivars change: https://github.com/rhboot/efivar/commit/011f3e5212d4bde5f651795435ccf0f77bd2633d

PR #4184, opened 2022-01-17
Fix compiling with new versions of efivar

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4181

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-common.h
+++ b/plugins/uefi-capsule/fu-uefi-common.h
@@ -9,6 +9,7 @@
 
 #include <fwupdplugin.h>
 
+#include <efivar/efivar.h>
 #include <glib.h>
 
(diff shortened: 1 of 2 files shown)


plugins/uefi-capsule/fu-uefi-common.h
```diff
@@ -9,6 +9,7 @@
 
 #include <fwupdplugin.h>
 
+#include <efivar/efivar-types.h>
```
superm1, line 12: seems you can't include this header in all cases, it should be part of the ifdef?

hughsie: I should have used the main header; fixed.

superm1 (approved):

superm1:
I think backport to various older branches too after merging.

hughsie:
Backported to 1_4_X, 1_5_X and 1_6_X

PR #4188, opened 2022-01-18
trivial: get-details finishes all the components, don't error out immediately

CragW:
trivial: get-details finishes all the components, don't error out immediately

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Before:
```
Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.13
```

After:
```
09:19:46:0241 FuEngine             ✘ Dock Package: Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.13
09:19:46:0241 FuEngine             ✘ EC: Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.13
09:19:46:0242 FuEngine             ✘ VMM6221 MST: Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.13
09:19:46:0242 FuEngine             ✘ USB 3.1 Generation 1: Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.13
09:19:46:0243 FuEngine             ✘ USB 3.1 Generation 2: Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.13
09:19:46:0243 FuEngine             ✘ USB4: Not compatible with org.freedesktop.fwupd version 1.7.5, requires >= 1.9.11
```

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -4291,6 +4291,7 @@ fu_engine_get_result_from_component(FuEngine *self,
 	g_autoptr(FuDevice) dev = NULL;
 	g_autoptr(FwupdRelease) rel = NULL;
 	g_autoptr(GError) error_local = NULL;
+	g_autoptr(GError) error_reqs = NULL;
 	g_autoptr(GPtrArray) provides = NULL;
 	g_autoptr(GPtrArray) tags = NULL;
 	g_autoptr(XbNode) description = NULL;
@@ -4327,6 +4328,8 @@ fu_engine_get_result_from_component(FuEngine *self,
 			fu_device_set_version_raw(dev, fu_device_get_version_raw(device));
 			fu_device_set_version_format(dev, fu_device_get_version_format(device));
 			fu_device_set_version(dev, fu_device_get_version(device));
+		} else {
+			fu_device_inhibit(dev, "not-found", "Device was not found");
 		}
 
 		/* add GUID */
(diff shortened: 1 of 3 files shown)


src/fu-engine.c
```diff
@@ -4490,9 +4490,15 @@ fu_engine_get_details(FuEngine *self, FuEngineRequest *request, gint fd, GError
 	for (guint i = 0; i < components->len; i++) {
 		XbNode *component = g_ptr_array_index(components, i);
 		FuDevice *dev;
-		dev = fu_engine_get_result_from_component(self, request, component, error);
-		if (dev == NULL)
-			return NULL;
+		error_local = NULL;
+
+		dev = fu_engine_get_result_from_component(self, request, component, &error_local);
```
hughsie, line 4495: no, I think we should do this one layer further in, i.e. if `fu_engine_check_requirements()` in `fu_engine_get_result_from_component` fails then use an error_local and do `fu_device_inhibit()` with the error message. That way we still return a device, and the user can see why the update can't be installed.

CragW: I've noticed that sometimes we want `inhibit` a device and add it some error message while it's not `updatable` at that moment. In this case, I've removed the `flag` check in `libfwupdplugin`.

hughsie: I've fixed it up (and added some tests) -- can you check this new version still works for you?

CragW: LGTM, it works nicely.

Issue #4190, opened 2022-01-18
(Plugin Modem Manager) `fu_io_channel_new_file: assertion 'filename != NULL' failed` (SIGSEGV)


j5k:
Can't run `fwupd` because of **SIGSEGV**:

**Service status:**
``` shell
× fwupd.service - Firmware update daemon
     Loaded: loaded (/usr/lib/systemd/system/fwupd.service; static)
     Active: failed (Result: core-dump) since Tue 2022-01-18 21:33:39 CET; 5min ago
       Docs: https://fwupd.org/
    Process: 6127 ExecStart=/usr/lib/fwupd/fwupd (code=dumped, signal=SEGV)
   Main PID: 6127 (code=dumped, signal=SEGV)
        CPU: 93ms

Jan 18 21:33:39 xxx systemd[1]: Starting Firmware update daemon...
Jan 18 21:33:39 xxx fwupd[6127]: 20:33:39:0822 FuIOChannel          fu_io_channel_new_file: assertion 'filename != NULL' failed
Jan 18 21:33:39 xxx systemd[1]: fwupd.service: Main process exited, code=dumped, status=11/SEGV
Jan 18 21:33:39 xxx systemd[1]: fwupd.service: Failed with result 'core-dump'.
Jan 18 21:33:39 xxx systemd[1]: Failed to start Firmware update daemon.
```
**gdb:**
``` shell
gdb --args /usr/lib/fwupd/fwupd --verbose
...
(gdb) r
...
20:41:25:0451 FuDeviceList         ::added 4bde70ba4e39b28f9eab1628f9dd6e6244c03027
20:41:25:0451 FuPlugin             coldplug(modem_manager)
20:41:25:0452 FuPluginMm           ModemManager 1.18.4-1 is available
20:41:25:0452 FuPluginMm           added modem: /org/freedesktop/ModemManager1/Modem/0
20:41:25:0452 FuPluginMm           valid firmware update combination: mbim-qdu
20:41:25:0454 FuIOChannel          fu_io_channel_new_file: assertion 'filename != NULL' failed

Thread 1 "fwupd" received signal SIGSEGV, Segmentation fault.
0x00007fffefdd2d1b in ?? () from /usr/lib/fwupd-plugins-5/libfu_plugin_modem_manager.so
(gdb) bt
#0  0x00007fffefdd2d1b in  () at /usr/lib/fwupd-plugins-5/libfu_plugin_modem_manager.so
[...]

PR #4193, opened 2022-01-19
modem-manager: Fix a segfault on startup for some MBIM-QDU devices

hughsie:
Works around https://github.com/fwupd/fwupd/issues/4190 although we
still need to find the root cause.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Merging because it's fixing a fairly common crash people are seeing. @aleksander0m we can continue to discuss on the issue for the root cause.

PR #4200, opened 2022-01-20
msr: make tsme security attribute match kernel doc more closely

superm1:
The kernel documentation for AMD memory encryption includes information
on both SME and SEV, but they are separate attributes. Read the attributes
separately.

Link: https://www.kernel.org/doc/html/v5.16/x86/amd-memory-encryption.html

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
Ack, thanks for taking the time to make all this sane.

PR #4202, opened 2022-01-20
Use GCC static analysis when available

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -108,6 +108,9 @@ warning_flags = [
   '-Wvla',
   '-Wwrite-strings'
 ]
+if host_machine.system() != 'windows'
+  warning_flags += ['-fanalyzer', '-Wno-analyzer-null-dereference']
+endif
 cc = meson.get_compiler('c')
 add_project_arguments(cc.get_supported_arguments(warning_flags), language : 'c')
 


hughsie:
> `x86_64-w64-mingw32-gcc: fatal error: Killed signal terminated program cc1`

/me goes to file a GCC bug...

meson.build
```diff
@@ -108,6 +108,9 @@ warning_flags = [
   '-Wvla',
   '-Wwrite-strings'
 ]
+if not meson.is_cross_build()
```
superm1, line 111: Can we detect mingw specifically instead?

hughsie: I don't think so, as this happens before `cc = meson.get_compiler('c')`

superm1: How about using
```
if host_machine.system() == 'windows'
```

hughsie: Lets try that; fixed

superm1 (approved):

meson.build
superm1: Maybe leave a comment with your GCC bug so future selves can find it when it's fixed to take this out.

hughsie: It might just be the mingw toolchain; investigating after I've done all my meetings for the day.

PR #4203, opened 2022-01-20
Allow overriding daemon parameters using /var/etc/fwupd/daemon.conf

hughsie:
This is super useful for debugging immutable systems like ChromeOS.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -1448,6 +1448,13 @@ fu_common_get_path(FuPathKind path_kind)
 			return g_build_filename(tmp, NULL);
 		basedir = fu_common_get_path(FU_PATH_KIND_LOCALSTATEDIR);
 		return g_build_filename(basedir, "cache", PACKAGE_NAME, NULL);
+	/* /var/etc/fwupd */
+	case FU_PATH_KIND_LOCALCONFDIR_PKG:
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -1448,6 +1448,13 @@ fu_common_get_path(FuPathKind path_kind)
 			return g_build_filename(tmp, NULL);
 		basedir = fu_common_get_path(FU_PATH_KIND_LOCALSTATEDIR);
 		return g_build_filename(basedir, "cache", PACKAGE_NAME, NULL);
+	/* /var/etc/fwupd */
+	case FU_PATH_KIND_LOCALCONFDIR_PKG:
+		tmp = g_getenv("LOCALCONF_DIRECTORY");
+		if (tmp != NULL && g_file_test(tmp, G_FILE_TEST_EXISTS))
```
superm1, line 1454: don't you want to test it's a directory not that it exists?

superm1 (approved):

libfwupdplugin/fu-common.c
hughsie: I think we have bigger problems if it's not a directory.

Issue #4040, opened 2021-11-29
Managing Best Known Configuration downstream


ErwanAliasr1:
**Is your feature request related to a problem? Please describe.**
By reading https://blogs.gnome.org/hughsie/2021/11/29/firmware-best-known-configuration-in-fwupd/, I was wondering if we could have some override downstream.
With the existing feature, Vendors can define the set of stable firmware to be used, therefore on some infra, users would love to get another set of stable firmware for the following reasons : 
- the fw didn't got tested yet
- the vendor consider the fw as stable but the end users have issues and need to rollback on other releases before a fix is distributed
- the newest fw breaks some configuration/features, users needs to stay on a different version

**Describe the solution you'd like**
Having a way to define overrides downstream to set our list of stable firmware for a given fleet of servers.

**Additional context**
By running 10000s of servers, you know fw are broken on a daily basis and you need to maintain your own list proved to be perfectly functional.

Thx !

superm1:
I think some more explanation of your situation needs to be described to come up with the right solution.
* Specifically I would wonder - are you directly taking LVFS as a remote or will you have a local remote that has the firmware?

PR #4205, opened 2022-01-20
Allow specifying the BKC locally

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4040

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -123,6 +123,7 @@ struct _FuEngine {
 	gboolean loaded;
 	gchar *host_security_id;
 	FuSecurityAttrs *host_security_attrs;
+	GPtrArray *local_monitors; /* (element-type GFileMonitor) */
 };
 
 enum {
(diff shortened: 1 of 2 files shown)


src/fu-engine.c
```diff
@@ -3809,6 +3885,38 @@ fu_engine_md_refresh_devices(FuEngine *self)
 	}
 }
 
+static gboolean
+fu_engine_load_metadata_store_local(FuEngine *self, XbBuilder *builder, GError **error)
+{
+	g_autofree gchar *localstatedir_pkg = fu_common_get_path(FU_PATH_KIND_LOCALSTATEDIR_PKG);
```
hughsie, line 3891: Maybe we want to load the immutable ones in /usr/share/fwupd/metadata.d too?

docs/local-metadata.md
```diff
@@ -0,0 +1,54 @@
+Local Metadata
+==============
+
+Local metadata can be saved in `/var/lib/fwupd/metadata.d` or `/usr/share/fwupd/metadata.d` which is
+scanned at daemon startup. This can be used to add site-specific BKC tags to existing metadata
+stores. For instance:
```
superm1, line 6: Please tell me this isn't going to turn into Acme Corp distributing debs or RPMs with this information...

superm1 (dismissed):

docs/local-metadata.md
hughsie: I think the plan is for the Google management agent thing to write files into /var/lib/fwupd/metadata.d as required.

PR #4211, opened 2022-01-21
Freebsd CI fix

Asiderr:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Fixes the problems with FreeBSD CI. There was a regression related to the following commit: https://github.com/fwupd/fwupd/commit/e74d38bfd3097471fe60dbe843a68c16516a78da
`<efivar/efivar.h>` is wrong path for the FreeBSD efivar library. Also it is not needed in the common header, when there is no  `efi_time_t`.

The second problem was connected to the libxmlb FreeBSD recipe. I fixed it locally at https://github.com/3mdeb/freebsd-ports/tree/fwupd, but we need to refresh this [topic](https://reviews.freebsd.org/D29332). It needs to be merged to make CI maintenance easier.

### DIFF
--- a/contrib/freebsd/Makefile
+++ b/contrib/freebsd/Makefile
@@ -52,6 +52,8 @@ MESON_ARGS=	-Dgudev=false \
 		-Dplugin_realtek_mst=false \
 		-Dplugin_thunderbolt=false \
 		-Dplugin_tpm=false \
+		-Dplugin_bcm57xx=false \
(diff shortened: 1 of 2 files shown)


Asiderr:
CI job proof: https://github.com/3mdeb/fwupd/runs/4898339374?check_suite_focus=true

hughsie (approved):

PR #4217, opened 2022-01-24
Translate the firmware flags when using fwupdmgr

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1647,9 +1647,49 @@ fu_util_release_urgency_to_string(FwupdReleaseUrgency release_urgency)
 	return _("Unknown");
 }
 
+static const gchar *
+fu_util_release_flag_to_string(FwupdReleaseFlags release_flag)
+{
+	if (release_flag == FWUPD_RELEASE_FLAG_NONE)
+		return NULL;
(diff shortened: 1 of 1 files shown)


hughsie:
Screenshot from 2022-01-24 17-03-28

src/fu-util-common.c
```diff
@@ -1761,18 +1803,26 @@ fu_util_release_to_string(FwupdRelease *rel, guint idt)
...
+		if (title != NULL) {
+			g_autofree gchar *bullet = NULL;
+			bullet = g_strdup_printf("• %s", tmp2);
+			fu_common_string_append_kv(str, idt + 1, title, bullet);
+			title = NULL;
+		} else {
+			g_autofree gchar *bullet = NULL;
+			bullet = g_strdup_printf("• %s", tmp2);
+			fu_common_string_append_kv(str, idt + 1, "", bullet);
+		}
```
superm1, line 1823: How about `title != NULL ? title : ""`?

superm1 (approved):
just one nit to remove some code lines, otherwise LGTM

src/fu-util-common.c
hughsie: Yes, and we can make it one line simpler too. Thanks.

PR #4218, opened 2022-01-26
Master

LuckyCandice:
Type of pull request:

- [ ] Code fix
- [.] Feature
- [ ] Documentation


plugins/analogix/analogix.quirk
```diff
@@ -1,3 +1,15 @@
 # Phoenix-Lite
 [USB\VID_1F29&PID_7518]
 Plugin = analogix
+
+//Belkin
```
hughsie, line 5: can you use `#` for comments please

plugins/analogix/fu-analogix-device.c
```diff
@@ -241,7 +241,24 @@ fu_analogix_device_probe (FuDevice *device, GError **error)
...
+	}
+	else if (vid == 0x047D)
+	{
+		fu_device_set_vendor (device, "Kensington");
+	}
+	else if (vid == 0x0502)
+	{
+		fu_device_set_vendor (device, "Acer");
+	}
+	else{}
```
hughsie, line 261: Do you need an of these? The vendor should be bieng set automatically using the udev vendor database.

LuckyCandice: Do you mean I don't need call fu_device_set_verdor to set vendor name? But if I didn't call this API, I found verdor is displayed as "Analogix".

hughsie: Try removing the manual:

    > plugins/analogix/fu-analogix-device.c:  fu_device_set_vendor(FU_DEVICE(self), "Analogix Semiconductor Inc.");

hughsie:
I don't think you wanted that last commit

plugins/analogix/fu-analogix-device.c
LuckyCandice: I removed this line codes, but the vendor displays "Analogix".

hughsie: What does the USB descriptor say? i.e. the string pointed by iManufacturer?

LuckyCandice: I see, let me check this, I will pull a new request, thank you!

PR #4222, opened 2022-01-26
Allow fwupd to operate without a D-Bus daemon

hughsie:
This adds support for optionally using a UNIX domain socket where a
D-Bus daemon may not be running.

To use this, launch the daemon and clients with something like
`FWUPD_DBUS_SOCKET=/var/run/fwupd.sock fwupdmgr get-devices`

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/installed-tests/fwupdmgr-p2p.sh
+++ b/data/installed-tests/fwupdmgr-p2p.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+# only run as root, possibly only in CI
+if [ "$(id -u)" -ne 0 ]; then exit 0; fi
+
+# ---
+echo "Starting P2P daemon..."
+export FWUPD_DBUS_SOCKET="/var/run/fwupd.sock"
+rm -rf ${FWUPD_DBUS_SOCKET}
+/usr/libexec/fwupd/fwupd --verbose --timed-exit --no-timestamp &
+while [ ! -e ${FWUPD_DBUS_SOCKET} ]; do sleep 1; done
+
+# ---
+echo "Starting P2P client..."
+fwupdmgr get-devices --json
+rc=$?; if [ $rc != 0 ]; then exit $rc; fi
+
+# ---
+echo "Shutting down P2P daemon..."
+gdbus call --system --dest org.freedesktop.fwupd --object-path / --method org.freedesktop.fwupd.Quit
+
+# success!
+exit 0
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -12,6 +12,7 @@
 #include <gio/gunixfdlist.h>
 #include <glib-unix.h>
 #include <glib/gi18n.h>
+#include <glib/gstdio.h>
 #include <locale.h>
 #include <xmlb.h>
 #ifdef HAVE_MALLOC_H
@@ -239,6 +240,12 @@ fu_main_create_request(FuMainPrivate *priv, const gchar *sender, GError **error)
 	uid_t calling_uid = 0;
 	g_autoptr(FuEngineRequest) request = fu_engine_request_new(FU_ENGINE_REQUEST_KIND_ACTIVE);
 	g_autoptr(GVariant) value = NULL;
+
+	/* if using FWUPD_DBUS_SOCKET... */
+	if (sender == NULL) {
+		fu_engine_request_set_device_flags(request, FWUPD_DEVICE_FLAG_TRUSTED);
+		return g_steal_pointer(&request);
+	}
 
 	g_return_val_if_fail(sender != NULL, NULL);
 
@@ -962,7 +969,11 @@ fu_main_install_with_helper(FuMainAuthHelper *helper_ref, GError **error)
 static FuSenderItem *
 fu_main_ensure_sender_item(FuMainPrivate *priv, const gchar *sender)
 {
-	FuSenderItem *sender_item;
+	FuSenderItem *sender_item = NULL;
+
+	/* operating in point-to-point mode */
+	if (sender == NULL)
+		sender = "";
 	sender_item = g_hash_table_lookup(priv->sender_items, sender);
 	if (sender_item == NULL) {
 		sender_item = g_new0(FuSenderItem, 1);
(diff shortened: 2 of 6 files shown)


superm1 (commented):
You know this is begging to have support for using unix sockets on Windows and building / using a daemon there....
https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/#:~:text=Unix%20sockets%20allow%20inter-process%20communication%20%28IPC%29%20between%20processes,alternatives%20for%20local%20IPC%2C%20such%20as%20named%20pipes.

hughsie:
> You know this is begging to have support for using unix sockets on Windows

I never thought about that; I guess it gets us one step further. I'm not sure Jared would be super happy me spending even more time on the windows port :)

superm1 (changes requested):
This stuff is unique enough that I think this needs some explicit unit tests in CI so it doesn't bit-rot.

src/fu-main.c
```diff
@@ -2160,15 +2200,41 @@ main(int argc, char *argv[])
...
-					fu_main_on_name_lost_cb,
-					priv,
-					NULL);
+	if (socket_filename != NULL) {
+		g_autofree gchar *address = g_strdup_printf("unix:path=%s", socket_filename);
+		g_autofree gchar *guid = g_dbus_generate_guid();
+		g_autoptr(GDBusServer) server = NULL;
+
+		/* this must be owned by root */
+		g_unlink(socket_filename);
```
superm1, line 2209: I would think you want some safety checks to make sure that the file being unlinked is a socket or so no?

hughsie: I'm wondering if we want to do the unlink at all; it seems to me that writing a systemd service that manages the socket outside of the daemon seems to be the 2022 way to do it.

superm1: Thinking about flexibility how about check if compiled with systemd and do that, and if not do the unlink?

data/installed-tests/fwupdmgr-p2p.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# ---
+echo "Starting P2P daemon..."
+FWUPD_DBUS_SOCKET=/var/run/fwupd.sock /usr/libexec/fwupd/fwupd --verbose --timed-exit --no-timestamp &
```
superm1, line 5: for the installed test, don't you want to use a systemd unit with the socket defined and managed to start it?

data/installed-tests/fwupdmgr-p2p.sh
```diff
@@ -0,0 +1,14 @@
...
+sleep 5
+
+# ---
+echo "Starting P2P client..."
+FWUPD_DBUS_SOCKET=/var/run/fwupd.sock /usr/bin/fwupdmgr get-devices --json
```
superm1, line 10: so I was kinda just thinking that you could do something like:
FWUPD_DBUG_SOCKET=foo `/usr/share/installed-tests/fwupdmgr`

Then it proves all the existing installed tests work in socket mode or in real dbus mode.

data/installed-tests/fwupdmgr-p2p.sh
hughsie: Is systemd running in the CI?

superm1: At least for Debian it seems to be:

```
        service dbus restart
        gnome-desktop-testing-runner fwupd
        apt purge -y fwupd-tests
```

superm1: Doesn't systemd allow environment override files? My idea was create one that sets dbus variable.

Although for installed tests on real systems this falls apart.
So scratch this idea.

However I think you want to make sure the socket exists rather than just sleep to make this more foolproof.

data/installed-tests/fwupdmgr-p2p.sh
```diff
@@ -0,0 +1,19 @@
...
+rm -rf ${FWUPD_DBUS_SOCKET}
+/usr/libexec/fwupd/fwupd --verbose --timed-exit --no-timestamp &
+while [ ! -e ${FWUPD_DBUS_SOCKET} ]; do sleep 1; done
+
+# ---
+echo "Starting P2P client..."
+/usr/bin/fwupdmgr get-devices --json
+rc=$?; if [ $rc != 0 ]; then exit $rc; fi
+
+# success!
```
superm1, line 18: I think you want to get the PID and explicitly send a kill signal to the backend here right?

superm1, line 18: I think you want to get the PID and explicitly send a kill signal to the backend here right?

hughsie: I just used the `quit` action I proposed; I think that's a lot nicer.

data/installed-tests/fwupdmgr-p2p.sh
hughsie: I don't think we need to test everything; if the p2p get-devices works then that's already ~6 DBus messages and signals. I think having two tests (one of which only runs as root) means it's a lot easier to debug failures.

data/installed-tests/fwupdmgr-p2p.sh
```diff
@@ -0,0 +1,23 @@
...
+while [ ! -e ${FWUPD_DBUS_SOCKET} ]; do sleep 1; done
+
+# ---
+echo "Starting P2P client..."
+/usr/bin/fwupdmgr get-devices --json
+rc=$?; if [ $rc != 0 ]; then exit $rc; fi
+
+# ---
+echo "Shutting down P2P daemon..."
+sudo gdbus call --system --dest org.freedesktop.fwupd --object-path / --method org.freedesktop.fwupd.Quit
```
superm1, line 20: No need for sudo

data/installed-tests/fwupdmgr-p2p.sh
```diff
@@ -0,0 +1,23 @@
...
+# ---
+echo "Starting P2P daemon..."
+export FWUPD_DBUS_SOCKET="/var/run/fwupd.sock"
+rm -rf ${FWUPD_DBUS_SOCKET}
+/usr/libexec/fwupd/fwupd --verbose --timed-exit --no-timestamp &
+while [ ! -e ${FWUPD_DBUS_SOCKET} ]; do sleep 1; done
+
+# ---
+echo "Starting P2P client..."
+/usr/bin/fwupdmgr get-devices --json
```
superm1, line 15: Does the path need to be hardcoded? I would think as assumption that you're executing from PATH is fine

data/installed-tests/fwupdmgr-p2p.sh
hughsie: Indeed, thanks!

data/installed-tests/fwupdmgr-p2p.sh
hughsie: Agree, fixed.

PR #4229, opened 2022-01-28
Allow the root user to ask the daemon to quit

hughsie:
This makes it easier to test the installed fwupd for the installed tests.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine-request.c
+++ b/src/fu-engine-request.c
@@ -80,6 +80,13 @@ fu_engine_request_set_device_flags(FuEngineRequest *self, FwupdDeviceFlags devic
 	self->device_flags = device_flags;
 }
 
+gboolean
+fu_engine_request_has_device_flag(FuEngineRequest *self, FwupdDeviceFlags device_flag)
+{
+	g_return_val_if_fail(FU_IS_ENGINE_REQUEST(self), FALSE);
+	return (self->device_flags & device_flag) > 0;
+}
+
 static void
 fu_engine_request_init(FuEngineRequest *self)
 {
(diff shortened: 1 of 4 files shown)


superm1 (commented):
TBH this feels pretty heavy handed. 99% of people won't quit this way.

Why not just support a unix signal sent as root to let the daemon signal handler intercept and cleanly shutdown?

hughsie:
> TBH this feels pretty heavy handed. 99% of people won't quit this way.

Sure, I don't think it makes sense for users to know this subcommand, I even thought about hiding it so that it doesn't show in the --help output.

> Why not just support a unix signal sent as root to let the daemon signal handler intercept and cleanly shutdown?

I thought we wanted this to work with Win32, no? I don't think we can handle signals there. Also, assuming we don't need this on win32, we'd also need a way (a property?) to get the PID of the fwupd process so we knew what to kill.

superm1:
Oh yeah great arguments.

How about moving the command to fwupdtool instead?

hughsie:
> How about moving the command to fwupdtool instead?

I guess we could, but I'm not sure it makes sense. In my head `fwupdmgr` is *tell the daemon to do something* and `fwupdtool` is *JFDI, right now*. I don't think it makes sense to have fwupdtool as an overflow of "half useful stuff".

superm1:
As this is limited purpose - how about just dbus send or call the CI app in python? I just don't think this needs to be in fwupdmgr.

hughsie:
> how about just dbus send

Yup, that works, good idea.

PR #4243, opened 2022-02-03
Add support for outputting versions in JSON format

hughsie:
To use this, do: `fwupdmgr --version --json`

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -3676,6 +3676,25 @@ fu_util_show_plugin_warnings(FuUtilPrivate *priv)
 	}
 }
 
+static gboolean
+fu_util_version(FuUtilPrivate *priv, GError **error)
+{
+	g_autoptr(GHashTable) metadata = NULL;
+	g_autofree gchar *str = NULL;
+
+	/* get metadata */
+	metadata = fwupd_client_get_report_metadata(priv->client, priv->cancellable, error);
+	if (metadata == NULL)
+		return FALSE;
+
+	/* dump to the screen in the most appropriate format */
+	if (priv->as_json)
+		return fu_util_project_versions_as_json(metadata, error);
+	str = fu_util_project_versions_to_string(metadata);
+	g_print("%s", str);
+	return TRUE;
+}
+
 int
 main(int argc, char *argv[])
 {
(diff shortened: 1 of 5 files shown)


src/fu-util.c
```diff
@@ -3676,6 +3676,119 @@ fu_util_show_plugin_warnings(FuUtilPrivate *priv)
...
+	if (g_str_has_prefix(str, "CompileVersion(")) {
+		if (dependency_kind != NULL)
+			*dependency_kind = FU_UTIL_DEPENDENCY_KIND_COMPILE;
+		return g_strndup(str + 15, strsz - 16);
+	}
+	return g_strdup(str);
+}
+
+static gboolean
+fu_util_version_as_json(FuUtilPrivate *priv, GHashTableIter *iter, GError **error)
```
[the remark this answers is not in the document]
hughsie: > I think we want json support in both tools

Yes, I suppose so. Will respin.

src/fu-util.c
```diff
@@ -3676,6 +3676,119 @@ fu_util_show_plugin_warnings(FuUtilPrivate *priv)
...
+		return "runtime";
+	if (dependency_kind == FU_UTIL_DEPENDENCY_KIND_COMPILE)
+		return "compile";
+	return NULL;
+}
+
+static gchar *
+fu_util_parse_project_dependency(const gchar *str, FuUtilDependencyKind *dependency_kind)
+{
+	gsize strsz = strlen(str);
```
superm1, line 3698: I think you want a g_return_val_if_fail  check that dependency_kind is not a NULL pointer.

src/fu-util.c
```diff
@@ -3676,6 +3676,119 @@ fu_util_show_plugin_warnings(FuUtilPrivate *priv)
...
+		g_string_append_printf(str,
+				       "%-10s%-30s%s\n",
+				       fu_util_dependency_kind_to_string(dependency_kind),
+				       project,
+				       value);
+	}
+
+	/* add this as a sanity check */
+	if (fwupd_client_get_host_bkc(priv->client) != NULL) {
+		g_string_append_printf(
```
superm1, line 3781: isn't this missed in the json?

hughsie: Yup, I just noticed that too :) I actually think adding it to the report metadata is a good idea too, so we can understand "odd" upgrade paths.

src/fu-util.c
hughsie: fixed

src/fu-util.c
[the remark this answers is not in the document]
hughsie: Done

superm1 (approved):

Issue #4244, opened 2022-02-03
Tests use quirk files installed in system instead of local versions


marv:
When updating fwupd from 1.6.3 to 1.7.4 two tests failed for me with the following error:
```
>>> MALLOC_PERTURB_=195 G_TEST_BUILDDIR=/var/tmp/paludis/build/sys-apps-fwupd-1.7.4/work/_build/plugins/ata G_TEST_SRCDIR=/var/tmp/paludis/build/sys-apps-fwupd-1.7.4/work/fwupd-1.7.4/plugins/ata /var/tmp/paludis/build/sys-apps-fwupd-1.7.4/work/_build/plugins/ata/ata-self-test
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ✀  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
stdout:
# random seed: R02S47214c8f88549119f35d57d0fbb07206
1..2
# Start of fwupd tests
# Start of ata tests
# GLib-GIO-DEBUG: _g_io_module_get_default: Found default implementation gvfs (GDaemonVfs) for ?gio-vfs?
# XbSilo-DEBUG: attempting to load /var/tmp/paludis/build/sys-apps-fwupd-1.7.4/temp/.E38YG1
# XbSilo-DEBUG: failed to load silo: blob too small
Bail out! FuQuirks-FATAL-WARNING: [INTEL_SPI_CHIPSET\APL] Flags = PCH is invalid: P is not lowercase
stderr:

(/var/tmp/paludis/build/sys-apps-fwupd-1.7.4/work/_build/plugins/ata/ata-self-test:3250226): FuQuirks-WARNING **: 08:50:02.104: [INTEL_SPI_CHIPSET\APL] Flags = PCH is in
valid: P is not lowercase
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
```

Running the test with verbose output showed that it's using the quirk file already installed in the system (`/usr/share/fwupd/quirks.d/intel-spi.quirk`)
```
# XbSilo-DEBUG: compiling /usr/share/fwupd/quirks.d/intel-spi.quirk:ctime=1633365579.446895?
Bail out! FuQuirks-FATAL-WARNING: [INTEL_SPI_CHIPSET\APL] Flags = PCH is invalid: P is not lowercase

[...]

PR #4245, opened 2022-02-03
trivial: use src directory for quirks for ata and nvme plugins

superm1:
Fixes: #4244

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/ata/fu-self-test.c
```diff
@@ -87,6 +87,7 @@ fu_ata_oui_func(void)
 int
 main(int argc, char **argv)
 {
+	g_setenv("FWUPD_DATADIR_QUIRKS", PLUGINSOURCEDIR, TRUE);
```
hughsie, line 90: Shouldn't we be setting this directly in the self test env={} dictionary in the meson file?

hughsie (approved):

PR #4248, opened 2022-02-04
Allow marking a device as EOL

hughsie:
This probably means it is unlikely to get any new security updates.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -756,6 +756,8 @@ fu_engine_set_release_from_appstream(FuEngine *self,
 			fwupd_release_set_update_image(rel, tmp);
 		}
 	}
+	if (xb_node_get_attr(release, "date_eol") != NULL)
+		fu_device_add_flag(dev, FWUPD_DEVICE_FLAG_END_OF_LIFE);
 
 	/* sort the locations by scheme */
 	g_ptr_array_sort_with_data(fwupd_release_get_locations(rel),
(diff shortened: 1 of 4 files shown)


src/fu-engine.c
```diff
@@ -756,6 +756,8 @@ fu_engine_set_release_from_appstream(FuEngine *self,
 			fwupd_release_set_update_image(rel, tmp);
 		}
 	}
+	if (xb_node_get_attr(release, "date_eol") != NULL)
```
superm1, line 759: isn't it `is_eol`?

hughsie: Nah, it's a ISO date in https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

superm1: Oh got it.  In that case is it worth changing this all around a bit and letting OEMs set a "future" EOL date that can be reflected in the GUI/tool?

I think the supported flag can become either:
* Supported (when no EOL date is set)
* Supported until $DATE (when one is set in future)
* End of Life since $DATE (when one was set in past)

hughsie: > and letting OEMs set a "future" EOL date

I think I'm going to struggle getting them to click a big red button, to be honest. I put the LVFS mtime as a DateTime deliberately (rather than a bool is_eol) so we can support this in the future.

hughsie: Maybe we also want a `fwupd_device_get_end_of_life_date()` or something like that? Not sure about the name tho :/

PR #4249, opened 2022-02-04
Drop the AMD SME check for security attributes (Fixes: #4176)

superm1:
This method does not indicate whether the kernel has enabled encryption.
That information is only visible from the kernel log at this time.

Continue to read the MSRs though, because if/when we get a method for the
kernel to export it we may use these as a basis of "support" but not a
indication of it being enabled.

encrypted:
```FuPluginMsr          SME/SEV check MSR: eax 01000f, ebx 016f
FuPluginMsr          PCI_MSR_AMD64_SYSCFG: 0f40000, sme_is_enabled=1
FuPluginMsr          PCI_MSR_AMD64_SEV: 00, sev_is_enabled=0
```

not_encrypted:
```
FuPluginMsr          SME/SEV check MSR: eax 01000f, ebx 016f
FuPluginMsr          PCI_MSR_AMD64_SYSCFG: 0f40000, sme_is_enabled=1
FuPluginMsr          PCI_MSR_AMD64_SEV: 00, sev_is_enabled=0
```

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/msr/fu-plugin-msr.c
+++ b/plugins/msr/fu-plugin-msr.c
@@ -291,49 +291,10 @@ fu_plugin_add_security_attr_dci_locked(FuPlugin *plugin, FuSecurityAttrs *attrs)
 }
 
 static void
-fu_plugin_add_security_attr_amd_sme_enabled(FuPlugin *plugin, FuSecurityAttrs *attrs)
-{
-	FuPluginData *priv = fu_plugin_get_data(plugin);
-	FuDevice *device = fu_plugin_cache_lookup(plugin, "cpu");
-	g_autoptr(FwupdSecurityAttr) attr = NULL;
-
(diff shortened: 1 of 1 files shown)


hughsie (approved):
Kinda sad we don't have that data yet. Fingers crossed for the future.

superm1:
> Kinda sad we don't have that data yet. Fingers crossed for the future.

I think you're on CC on the thread about it.   There seems to be a lot of resistance to a new sysfs file for it.  The unfortunate thing is it listerally stored in a global variable in the kernel `sme_mask`.

hughsie:
>  There seems to be a lot of resistance to a new sysfs file for it

Yes, the cynic in me would say that's deliberate, but we certainly need a way to read it back without parsing the entire journal log :(

PR #4257, opened 2022-02-07
Get the host machine ID when running on Windows

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -19,6 +19,12 @@
 #endif
 #include <errno.h>
 
+#ifdef _WIN32
+#include <sysinfoapi.h>
+#include <winerror.h>
+#include <winreg.h>
+#endif
+
 #include "fwupd-common-private.h"
 #include "fwupd-enums-private.h"
 #include "fwupd-error.h"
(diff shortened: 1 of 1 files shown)


src/fu-engine.c
```diff
@@ -6980,6 +6986,31 @@ fu_engine_load_local_metadata_watches(FuEngine *self, GError **error)
...
+	LSTATUS rc;
+
+	rc = RegGetValue(hkey, subkey, value, RRF_RT_REG_SZ, NULL, (PVOID)&buf, &bufsz);
+	if (rc != ERROR_SUCCESS) {
+		g_set_error(error,
+			    G_IO_ERROR,
+			    G_IO_ERROR_INVAL,
+			    "Failed to get registry string %s [0x%lX]",
+			    subkey,
+			    (unsigned long)rc);
```
superm1, line 7007: Can you use GetLastError?

superm1 (commented):
Have you tried on real Windows? I'm curious if this works without admin rights. Everything else so far does.

src/fu-engine.c
hughsie: I don't think it's the same thing, no?

hughsie:
> Have you tried on real Windows

Nope, will do now. I've been testing in WINE.

hughsie:
>  I'm curious if this works without admin rights

Seems to work in QEMU with win10.

superm1 (approved):

PR #4275, opened 2022-02-10
trivial: Add new category type for video display

gportay:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -3786,6 +3786,8 @@ fu_common_device_category_to_name(const gchar *cat)
 		return "Mouse";
 	if (g_strcmp0(cat, "X-Keyboard") == 0)
 		return "Keyboard";
+	if (g_strcmp0(cat, "X-VideoDisplay") == 0)
+		return "Display";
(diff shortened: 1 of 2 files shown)


src/fu-engine.c
```diff
@@ -3786,6 +3786,8 @@ fu_common_device_category_to_name(const gchar *cat)
 		return "Mouse";
 	if (g_strcmp0(cat, "X-Keyboard") == 0)
 		return "Keyboard";
+	if (g_strcmp0(cat, "X-VideoDisplay") == 0)
+		return "Video Display";
```
hughsie, line 3790: Maybe just Display? at least to my ears `Video Display` seems a bit odd to hear in 2022.

gportay: It was to match the icon name.

okay for the string.

gportay: done.

hughsie (approved):

hughsie:
I've also added the category to the LVFS too.

PR #4278, opened 2022-02-11
flashrom: Try harder to convert to a correct verfmt semver

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common-version.c
+++ b/libfwupdplugin/fu-common-version.c
@@ -235,6 +235,73 @@ _g_ascii_is_digits(const gchar *str)
 	return TRUE;
 }
 
+static guint
+fwupd_version_format_number_sections(FwupdVersionFormat fmt)
+{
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-common-version.c
```diff
@@ -235,6 +235,73 @@ _g_ascii_is_digits(const gchar *str)
 	return TRUE;
 }
 
+static guint
+_fwupd_version_format_sections(FwupdVersionFormat fmt)
```
superm1, line 239: nit; it's not really format sections - it's count the "number" of sections

superm1 (dismissed):
one nit and the commit message should be fixed up - this isn't flashrom (but it might be _for_ that)
otherwise LGTM

libfwupdplugin/fu-common-version.c
hughsie: _fwupd_version_format_number_sections perhaps?

superm1: thats good

PR #4279, opened 2022-02-13
Add a plugin to set GPIOs before and after an update

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/gpio/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+The GPIO bits to set during the update prepare phase, e.g. `5:high,6:low`.
+
+Since: 1.7.6
+
+### GpioCleanup
+
+The GPIO bits to set during the update cleanup phase, e.g. `5:high,6:low`.
+
+Since: 1.7.6
```
superm1, line 22: missing information about what kernel interfaces are used

plugins/gpio/fu-plugin-gpio.c
```diff
@@ -0,0 +1,138 @@
...
+{
+	return fu_plugin_gpio_process_quirks(self, device, "GpioCleanup", error);
+}
+
+void
+fu_plugin_init_vfuncs(FuPluginVfuncs *vfuncs)
+{
+	vfuncs->build_hash = FU_BUILD_HASH;
+	vfuncs->init = fu_plugin_gpio_init;
+	vfuncs->prepare = fu_plugin_gpio_backend_device_prepare_cb;
```
superm1, line 136: presumably some plugins may need this for probing too?  Need to think about that at all?

plugins/gpio/README.md
```diff
@@ -0,0 +1,22 @@
...
+This plugin sets GPIO outputs either high or low before and/or after an
+update has been deployed.
+
+## Quirk Use
+
+This plugin uses the following plugin-specific quirks:
+
+### GpioPrepare
+
+The GPIO bits to set during the update prepare phase, e.g. `5:high,6:low`.
```
superm1, line 14: There will need to be a lot of caution here about when quirking GPIOs since they are very unlikely to be the same across manufacturers or in many cases even models.  I hope there this plugin ends up not being used very frequently, it's a lot better for manufacturers to set up GPIOs in the correct direction using ACPI tables to start with.

plugins/gpio/fu-plugin-gpio.c
```diff
@@ -0,0 +1,138 @@
...
+static gboolean
+fu_plugin_gpio_set_direction(FuPlugin *self, guint gpio_num, const gchar *direction, GError **error)
+{
+	g_autofree gchar *fn = NULL;
+	g_autofree gchar *devname = g_strdup_printf("gpio%u", gpio_num);
+	g_autoptr(FuIOChannel) io_channel = NULL;
+
+	g_return_val_if_fail(direction != NULL, FALSE);
+
+	fn = g_build_filename("/sys/class/gpio", devname, "direction", NULL);
```
superm1, line 53: for future debugging and evaluating true need, I think it might be useful to read the current value and direction of the GPIO before changing it.

plugins/gpio/fu-plugin-gpio.c
```diff
@@ -0,0 +1,138 @@
...
+	FuContext *ctx = fu_plugin_get_context(plugin);
+	fu_context_add_quirk_key(ctx, "GpioPrepare");
+	fu_context_add_quirk_key(ctx, "GpioCleanup");
+}
+
+static gboolean
+fu_plugin_gpio_create_node(FuPlugin *self, guint gpio_num, GError **error)
+{
+	g_autofree gchar *devname = g_strdup_printf("gpio%u", gpio_num);
+	g_autofree gchar *fn = g_build_filename("/sys/class/gpio", devname, NULL);
```
superm1, line 23: why is this using the sysfs interface instead of the character device?  The character device should be a lot safer in a multi process system.

plugins/gpio/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Since: 1.7.6
+
+### GpioCleanup
+
+The GPIO bits to set during the update cleanup phase, e.g. `5:high,6:low`.
```
superm1, line 20: I think this is most likely too simplistic, no?  What about edge vs level triggers?  What about debounce?  Active high vs active low?

superm1 (commented):
I think you should look into using and building on top of `libgpiod` instead for gpio manipulation.

plugins/gpio/fu-plugin-gpio.c
hughsie: Hmm, good question. I suppose it's almost free to add, but my primary usecase was something like turning on a LED, clearing a WR-EN signal or changing the MUX on a host<->BMC.

plugins/gpio/README.md
hughsie: Yes, my use case is quirking the MTD device for a specific board. It'll use the HwIDS for sure.

plugins/gpio/fu-plugin-gpio.c
hughsie: ....because I didn't know the sysfs api was deprecated until about 10 minutes ago :) Will fix.

plugins/gpio/README.md
hughsie: This is really for *output* signals only; I don't think we want to get into the game of listening for GPIO events.

hughsie:
> I think you should look into using and building on top of libgpiod instead for gpio manipulation.

I think there would be a huge amount of pushback if I added that as a dep from either an embedded point of view, or from a RHEL point of view. However, I think we can re-implement the tiny bits of libgpiod we need like we did for things like efivar and smbios.

plugins/gpio/README.md
hughsie: Fixed.

plugins/gpio/fu-plugin-gpio.c
hughsie: Lets leave this just for prepare for the moment, and we can add another key as-and-when required. It'll be all the same plumbing.

plugins/gpio/fu-plugin-gpio.c
hughsie: I nuked this code moving to the chardev mode.

plugins/gpio/fu-plugin-gpio.c
hughsie: All fixed. It's a weird ioctl interface!

meson.build
```diff
@@ -383,6 +383,12 @@ if cc.has_header('linux/ipmi.h')
 else
   have_linux_ipmi = false
 endif
+if cc.has_header('linux/gpio.h')
```
superm1: Should only be when plugin _gpio is set right?

hughsie: That's only setting the `have_linux_gpio` variable, the logic we moved into the plugins themselves some time ago.

plugins/gpio/fu-plugin-gpio.c
```diff
@@ -0,0 +1,163 @@
...
+
+	/* add to cache, which maintains a ref */
+	fu_plugin_cache_add(self, fu_device_get_logical_id(device), device);
+
+	/* we do *not* want to export this device over D-Bus */
+	g_set_error_literal(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "not exporting GPIO device");
+	return FALSE;
```
superm1, line 151: this feels pretty hacky.  Why can't it export and just not be an updatable device?  Like the CPU is.

hughsie: Hmm, I had it that way before. Maybe it makes it better to export.

plugins/gpio/fu-gpio-device.c
```diff
@@ -0,0 +1,206 @@
...
+fu_gpio_device_init(FuGpioDevice *self)
+{
+}
+
+static void
+fu_gpio_device_finalize(GObject *object)
+{
+	FuGpioDevice *self = FU_GPIO_DEVICE(object);
+	if (self->fd > 0)
+		g_close(self->fd, NULL);
```
superm1, line 193: Just to double check we're not digging a blocker hole - can you make sure that gpio-info still works while fwupd holds the `fd` for the duration of the daemon being opened?

plugins/gpio/fu-plugin-gpio.c
```diff
@@ -0,0 +1,148 @@
...
+ */
+
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include "fu-gpio-device.h"
+
+struct FuPluginData {
+	FuGpioDevice *gpio_device;
```
superm1, line 14: Is this a valid assumption that there is only ever 1 GPIO device on a system?

plugins/gpio/fu-plugin-gpio.c
```diff
@@ -0,0 +1,148 @@
...
+fu_plugin_gpio_process_quirk(FuPlugin *self, const gchar *str, GError **error)
+{
+	FuPluginData *data = fu_plugin_get_data(self);
+	gboolean value = FALSE;
+	FuDevice *device;
+	g_auto(GStrv) split = g_strsplit(str, ",", -1);
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	/* already set by previous GUID */
+	if (data->gpio_device != NULL)
```
superm1, line 64: couldn't this also happen if multiple GPIO devices came by?  In which case you have a probe order race condition.

superm1 (dismissed):
Just two scalability questions.  If they're already considered by you and not really a worry everything else is fine to me.

plugins/gpio/fu-plugin-gpio.c
hughsie: That's an excellent question. There can be multiple GPIO devices, but we can only control one pin of one of them for the prepare->cleanup action. I think renaming the variable to `gpio_device_active` might make that more obvious. A better idea still might just be to look up the device from the plugin cache from the logical_id, which means less ref's everywhere, and also means if the gpio device someone weirdly disappears during the update then the right thing happens -- and we also solve the race issue you point out.

plugins/gpio/fu-gpio-device.c
hughsie: yup, it works fine, and you get the "consumer" shown as "fwupd" for the lifetime of the update.

plugins/gpio/fu-plugin-gpio.c
hughsie: Those changes make it a lot clearer what is going on.

PR #4280, opened 2022-02-14
fall back to signed EFI app as secure boot is off.

ycheng:
Some distribution hopes to not install the not-signed EFI app by default.
However, for the corner case, the user might still want to use not-signed
one (maybe to work around bugs or do testing). This implements
a logic that uses the not-signed EFI app as the secure boot is off, and
if it does not exist, then we fall back to using the signed one.

I have not tested this patch. I think that change looks trivial. I'll try to
test it as soon as I can and update the result here.

I created this PR on a release sooner and earlier basis.

Please kindly share if you think the direction of this change is fine
with you or not.

Type of pull request:

- [V] Feature

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-common.c
+++ b/plugins/uefi-capsule/fu-uefi-common.c
@@ -88,26 +88,48 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
 gchar *
 fu_uefi_get_built_app_path(GError **error)
 {
-	const gchar *extension = "";
 	const gchar *suffix;
-	g_autofree gchar *source_path = NULL;
 	g_autofree gchar *prefix = NULL;
-	if (fu_efivar_secure_boot_enabled())
-		extension = ".signed";
+	g_autofree gchar *source_path = NULL;
+	g_autofree gchar *source_path_signed = NULL;
+	gboolean source_path_exists = FALSE;
+	gboolean source_path_signed_exists = FALSE;
+
 	suffix = fu_uefi_bootmgr_get_suffix(error);
 	if (suffix == NULL)
 		return NULL;
 	prefix = fu_common_get_path(FU_PATH_KIND_EFIAPPDIR);
-	source_path = g_strdup_printf("%s/fwupd%s.efi%s", prefix, suffix, extension);
-	if (!g_file_test(source_path, G_FILE_TEST_EXISTS)) {
-		g_set_error(error,
-			    G_IO_ERROR,
-			    G_IO_ERROR_NOT_FOUND,
-			    "%s cannot be found",
-			    source_path);
-		return NULL;
+
+	source_path = g_strdup_printf("%s/fwupd%s.efi", prefix, suffix);
+	source_path_signed = g_strdup_printf("%s.signed", source_path);
+
+	source_path_exists = g_file_test(source_path, G_FILE_TEST_EXISTS);
+	source_path_signed_exists = g_file_test(source_path_signed, G_FILE_TEST_EXISTS);
+
+	if (fu_efivar_secure_boot_enabled()) {
+		if (!source_path_signed_exists) {
+			g_set_error(error,
+				    G_IO_ERROR,
+				    G_IO_ERROR_NOT_FOUND,
+				    "%s cannot be found",
+				    source_path_signed);
+			return NULL;
+		}
+		return g_steal_pointer(&source_path_signed);
 	}
-	return g_steal_pointer(&source_path);
+
+	if (source_path_exists)
+		return g_steal_pointer(&source_path);
+	if (source_path_signed_exists)
+		return g_steal_pointer(&source_path_signed);
+
+	g_set_error(error,
+		    G_IO_ERROR,
+		    G_IO_ERROR_NOT_FOUND,
+		    "%s and %s cannot be found",
+		    source_path,
+		    source_path_signed);
+	return NULL;
 }
 
 gboolean
--- a/contrib/debian/control.in
+++ b/contrib/debian/control.in
@@ -53,7 +53,6 @@ Recommends: python3,
 	    dbus,
 	    secureboot-db,
 	    udisks2,
-	    fwupd-unsigned,
 	    fwupd-signed,
 	    jq
 Suggests: gir1.2-fwupd-2.0


hughsie:
> However, for the corner case, the user might still want to use not-signed one

I'm confused why you'd want that, *and* and want to use fwupd to do the upgrade.

ycheng:
> I'm confused why you'd want that, _and_ and want to use fwupd to do the upgrade.

Do you mean that: we just use the signed one in both secure boot on and off?

hughsie:
> Do you mean that: we just use the signed one in both secure boot on and off?

Yes.

ycheng:
I think giving that EFI app developer that does not have access to the signing key a convenient way to test the EFI app will be good.

That probably does not happen a lot, since the EFI app is quite stable in general, but will be handy for that case, isn't it?

As the distro design the packing that the signed one will be always installed and the not-signed is optional, the new code will just use the signed one. The not-signed one will only be used if the user is requested to test something from a private build package before that's accepted by the distro. I think this use case happens a lot, right?

superm1:
The way this is done it still preserves the path to use shim in secure boot and to use directly loading the binary in unsecure boot.


I don't think that's inherently a bad thing but I want to point out the implications.
So this means for testing purposes you still do have different code paths depending on secure boot on/off. If the aim is less testing for all cases in distro this won't change anything. However because that bug existed previously with shim chain loading broken for a very long time I think it's good for having a non shim code path for testing or using in case it breaks again.

plugins/uefi-capsule/fu-uefi-common.c
```diff
@@ -88,26 +88,51 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
...
-	if (!g_file_test(source_path, G_FILE_TEST_EXISTS)) {
-		g_set_error(error,
-			    G_IO_ERROR,
-			    G_IO_ERROR_NOT_FOUND,
-			    "%s cannot be found",
-			    source_path);
-		return NULL;
+
+	source_path = g_strdup_printf("%s/fwupd%s.efi%s", prefix, suffix, "");
+	source_path_signed = g_strdup_printf("%s/fwupd%s.efi%s", prefix, suffix, ".signed");
```
superm1, line 104: Just a nit, I would think this can just be

`g_strdup_printf("%s.signed", source_path)`

plugins/uefi-capsule/fu-uefi-common.c
```diff
@@ -88,26 +88,51 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
...
+				    G_IO_ERROR_NOT_FOUND,
+				    "%s cannot be found",
+				    source_path_signed);
+			return NULL;
+		}
+	} else {
+		if (source_path_exists) {
+			return g_steal_pointer(&source_path);
+		}
+		if (source_path_signed_exists) {
```
superm1, line 124: Nit, no need for curly braces on one line statements.

superm1 (commented):
Directionally it would be good to understand if the aim is less testing, less space on the target image, or what.

I believe that contrib/debian and contrib/fedora.spec.in might need some packaging changes too at the same time as this as well.

plugins/uefi-capsule/fu-uefi-common.c
ycheng: refined.

plugins/uefi-capsule/fu-uefi-common.c
ycheng: refined.

plugins/uefi-capsule/fu-uefi-common.c
```diff
@@ -100,8 +100,8 @@ fu_uefi_get_built_app_path(GError **error)
 		return NULL;
 	prefix = fu_common_get_path(FU_PATH_KIND_EFIAPPDIR);
 
-	source_path = g_strdup_printf("%s/fwupd%s.efi%s", prefix, suffix, "");
-	source_path_signed = g_strdup_printf("%s/fwupd%s.efi%s", prefix, suffix, ".signed");
+	source_path = g_strdup_printf("%s/fwupd%s.efi", prefix, suffix);
+	source_path_signed = g_strdup_printf("%s/fwupd%s.efi.signed", prefix, suffix);
```
superm1, line 104: I think you got this wrong in the fixup

superm1: (Just use the string you just built was my point)

ycheng: aha, you mean the 2nd one, right!

superm1:
> > So this means for testing purposes you still do have different code paths depending on secure boot on/off.
> 
> for deb, I think maybe we can add "Provides: fwupd-signed" in fwupd-unsigned, so that we don't necessarily need to install fwupd-signed.

No, I think that is wrong and confusing.  My suggestion is to just drop this line if this PR direction is the way we go: https://github.com/fwupd/fwupd/blob/main/contrib/debian/control.in#L56

ycheng:
> No, I think that is wrong and confusing. My suggestion is to just drop this line if this PR direction is the way we go: https://github.com/fwupd/fwupd/blob/main/contrib/debian/control.in#L56

Ok. Do you want me to drop that in this PR?

plugins/uefi-capsule/fu-uefi-common.c
ycheng: resolved

superm1:
> > No, I think that is wrong and confusing. My suggestion is to just drop this line if this PR direction is the way we go: https://github.com/fwupd/fwupd/blob/main/contrib/debian/control.in#L56
> 
> Ok. Do you want me to drop that in this PR?

Yeah. Please squash to one commit and force push to the branch.

superm1 (dismissed):
I think the code change looks good if we're approving this direction upstream.

@hughsie what do you think about it?

plugins/uefi-capsule/fu-uefi-common.c
```diff
@@ -88,26 +88,49 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
...
+		if (source_path_signed_exists) {
+			return g_steal_pointer(&source_path_signed);
+		} else {
+			g_set_error(error,
+				    G_IO_ERROR,
+				    G_IO_ERROR_NOT_FOUND,
+				    "%s cannot be found",
+				    source_path_signed);
+			return NULL;
+		}
```
hughsie, line 119: this is weird. Just do

 		if (!source_path_signed_exists) {
			g_set_error(error,
				    G_IO_ERROR,
				    G_IO_ERROR_NOT_FOUND,
				    "%s cannot be found",
				    source_path_signed);
			return NULL;
		}
		return g_steal_pointer(&source_path_signed);

there's no need for `else` if the other conditional returns.

hughsie:
> @hughsie what do you think about it?

I'm not sure I get all the nuances. From a KISS point of view I can see the benefit in just requiring the signed version, but if this is problematic maybe we should be pragmatic.

plugins/uefi-capsule/fu-uefi-common.c
ycheng: refined and force pushed.

ycheng:
> > @hughsie what do you think about it?
> 
> I'm not sure I get all the nuances. From a KISS point of view I can see the benefit in just requiring the signed version, but if this is problematic maybe we should be pragmatic.

For one who wants to do a EFI development and send the output for testing, he will need to send a file and ask the QA to override the signed efi binary with a not-signed EFI. That also sounds very bad, right?

plugins/uefi-capsule/fu-uefi-common.c
```diff
@@ -88,26 +88,48 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
...
+		return g_steal_pointer(&source_path_signed);
+	} else {
+		if (source_path_exists)
+			return g_steal_pointer(&source_path);
+		if (source_path_signed_exists)
+			return g_steal_pointer(&source_path_signed);
```
hughsie, line 123: do we need this else block?

superm1: Yes, if it's not there then you could only test with secure boot enabled

ycheng: Done

Either you have compiler optimizer build-in in your mind, or I write too many python codes recently. ;D

hughsie (approved):

ycheng:
Thank you!

Issue #4281, opened 2022-02-14
No updates (or warnings) with lid is closed


iyanmv:
**Describe the bug**
Some updates cannot be installed when the lid is closed but no warnings or messages are shown to the user. I learned this the hard way, when I was about to open a bug report at the [lenovo firmware repo](https://github.com/fwupd/firmware-lenovo).

Most of the times I have my laptop with the lid closed and connected to a dock station, so running
```fwupdmgr refresh && fwupdmgr get-updates```
was not showing me anything.

Only when I did a `fwupdmgr get-devices` I saw a few items with the message `Update Error:       Cannot install update when the lid is closed`

**Steps to Reproduce**
With a Lenovo Thinkpad X1 Carbon Gen 9 / Yoga Gen 6 run `fwupdmgr refresh && fwupdmgr get-updates` with the lid closed

**Expected behavior**
Running `fwupdmgr refresh && fwupdmgr get-updates` should so all available updates, no matter if they cannot be installed at that time for whatever reason (no AC cord, lid closed, etc.) and get a `No updates available for remaining devices`. Open the lid, run the same command and get available updates.

**fwupd version information**
```
runtime   org.freedesktop.fwupd         1.7.5
runtime   com.dell.libsmbios            2.4
compile   org.freedesktop.gusb          0.3.10
runtime   org.kernel                    5.16.8-arch1-1
compile   org.freedesktop.fwupd         1.7.5
runtime   org.freedesktop.gusb          0.3.10
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): `pacman`

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```
20XYCTO1WW
│
├─ELAN0672:00 04F3:3187:
│     Device ID:          0867c07f040e810fa5ea153981e63e324c29eb1f
│     Summary:            Touchpad
│     Current version:    0x0003
│     Bootloader Version: 0x0002
│     Vendor:             HIDRAW:0x04F3
│     GUIDs:              940805ff-ca7a-56b2-af86-32e995004af9
│                         ab16bce2-e219-52f5-89a6-fd6b1c4244c0
│                         e4c6161e-2b1c-59cb-ad07-c8fb7feb155e
│                         4b808d27-c833-50cc-8177-9a401f5c08fe
│                         5b73c06e-3dfd-5e6c-b06a-24e3489206c2
│     Device Flags:       • Internal device
│                         • Updatable
│   
├─WD19TB:
[...]

PR #4283, opened 2022-02-14
Show the user when devices are updatable by inhibited

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4281

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -584,8 +584,9 @@ fu_util_get_updates(FuUtilPrivate *priv, gchar **values, GError **error)
 	g_autoptr(GPtrArray) devices = NULL;
 	g_autoptr(GNode) root = g_node_new(NULL);
 	g_autofree gchar *title = NULL;
-	gboolean no_updates_header = FALSE;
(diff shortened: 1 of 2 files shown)


hughsie:
Screenshot from 2022-02-14 09-17-33

superm1 (commented):
Presumably also should do same for fu-tool?

hughsie:
> Presumably also should do same for fu-tool?

Ya, I just wanted broad consensus really. Will copy it over.

superm1:
> > Presumably also should do same for fu-tool?
> 
> Ya, I just wanted broad consensus really. Will copy it over.

Yeah makes sense.  I do think you'll want to do something similar in gnome-software too, this is quite useful IMO.

hughsie:
>  I do think you'll want to do something similar in gnome-software too

That's harder as it'll need UX review, and GNOME is in UI freeze. Will put on the list for when main opens up again. Ive copied the behaviour in fwupdtool now.

superm1 (approved):

PR #4286, opened 2022-02-15
[thunderbolt] Add 5 secs delay after nvm update

krankun:
- this is needed for the retimer to be accessible after update
- and to bring the port back to online

Change-Id: I2234ebe23a55fc03f9ef30c1ea10febfe46a2003

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/thunderbolt/fu-plugin-thunderbolt.c
+++ b/plugins/thunderbolt/fu-plugin-thunderbolt.c
@@ -13,6 +13,10 @@
 #include "fu-thunderbolt-firmware.h"
 #include "fu-thunderbolt-retimer.h"
 
+/*5 seconds sleep until retimer is available                                       \
+				     after nvm update*/
+#define FU_THUNDERBOLT_RETIMER_CLEANUP_DELAY 5000000
+
 static gboolean
 fu_plugin_thunderbolt_safe_kernel(FuPlugin *plugin, GError **error)
 {
@@ -89,6 +93,7 @@ fu_plugin_thunderbolt_composite_cleanup(FuPlugin *plugin, GPtrArray *devices, GE
 		FuDevice *dev = g_ptr_array_index(devices, i);
 		if ((g_strcmp0(fu_device_get_plugin(dev), "thunderbolt") == 0) &&
 		    fu_device_has_internal_flag(dev, FU_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)) {
+			g_usleep(FU_THUNDERBOLT_RETIMER_CLEANUP_DELAY);
 			return fu_thunderbolt_retimer_set_parent_port_online(dev, error);
 		}
 	}


plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -89,6 +89,9 @@ fu_plugin_thunderbolt_composite_cleanup(FuPlugin *plugin, GPtrArray *devices, GE
 		FuDevice *dev = g_ptr_array_index(devices, i);
 		if ((g_strcmp0(fu_device_get_plugin(dev), "thunderbolt") == 0) &&
 		    fu_device_has_internal_flag(dev, FU_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)) {
+			g_usleep(
```
superm1, line 92: Since this is a value that you may tune later - maybe put it in `#define` at the top and include your comment there.
then this can just be something like: 
`g_usleep(RETIMER_REPLUG_DELAY);`

krankun: done!

djcampello:
This is a workaround to a lower level bug.. This should not go in.... If the device needs some time to settle before accepting this request it should be at least handled by the kernel interface or some lower level interface and not managed by user space by adding some random time delay.

superm1:
Great, thanks Daniel.

westeri:
It is not a workaround to a lower level bug at all. And certainly not random delay. It is all documented in the USB4 re-timer spec and also in the kernel documentation:

https://docs.kernel.org/admin-guide/thunderbolt.html#upgrading-on-board-retimer-nvm-when-there-is-no-cable-connected

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -89,6 +91,8 @@ fu_plugin_thunderbolt_composite_cleanup(FuPlugin *plugin, GPtrArray *devices, GE
 		FuDevice *dev = g_ptr_array_index(devices, i);
 		if ((g_strcmp0(fu_device_get_plugin(dev), "thunderbolt") == 0) &&
 		    fu_device_has_internal_flag(dev, FU_DEVICE_INTERNAL_FLAG_NO_AUTO_REMOVE)) {
+			g_usleep(RETIMER_REPLUG_DELAY); /*5 seconds sleep until retimer is available
```
superm1, line 94: I think move this comment up above to right before the `#define`

superm1:
> It is not a workaround to a lower level bug at all. And certainly not random delay. It is all documented in the USB4 re-timer spec and also in the kernel documentation:
> 
> https://docs.kernel.org/admin-guide/thunderbolt.html#upgrading-on-board-retimer-nvm-when-there-is-no-cable-connected

Thanks Mika.  In that case, I've opened the PR back up.  @krankun please just fix the nit and we can merge this.

krankun:
Thank you all for the great points bought up in review and clarification , The CL is ready now addressing the comments !

plugins/thunderbolt/fu-plugin-thunderbolt.c
```diff
@@ -13,6 +13,10 @@
 #include "fu-thunderbolt-firmware.h"
 #include "fu-thunderbolt-retimer.h"
 
+#define RETIMER_REPLUG_DELAY                                                                       \
+	5000000 /*5 seconds sleep until retimer is available                                       \
+				      after nvm update*/
```
hughsie, line 18: can you just put the comment on the line before please :)

hughsie: Also, maybe s/RETIMER_REPLUG_DELAY/FU_THUNDERBOLT_RETIMER_CLEANUP_DELAY?

krankun: FU_THUNDERBOLT_RETIMER_CLEANUP_DELAY sounds good

djcampello:
> It is not a workaround to a lower level bug at all. And certainly not random delay. It is all documented in the USB4 re-timer spec and also in the kernel documentation:
> 
> https://docs.kernel.org/admin-guide/thunderbolt.html#upgrading-on-board-retimer-nvm-when-there-is-no-cable-connected

How come nvm_authenticate does not block for the 5 seconds itself? I see that it is documented, not sure that it makes it less of a bad implementation.

djcampello:
Also it seems that the documentation asks for a second call to rescan. This is currently not been done in the thunderbolt plugin. Would this explain why the nvm_version does not seem to be updated on the retimer FuDevice after the write? Should we be doing this rescan before issuing a `0 > offline` ?

djcampello:
> > It is not a workaround to a lower level bug at all. And certainly not random delay. It is all documented in the USB4 re-timer spec and also in the kernel documentation:
> > https://docs.kernel.org/admin-guide/thunderbolt.html#upgrading-on-board-retimer-nvm-when-there-is-no-cable-connected
> 
> How come nvm_authenticate does not block for the 5 seconds itself? I see that it is documented, not sure that it makes it less of a bad implementation.

Kernel is allowing for an attacker to simply bring down the expected capability by simply issuing fast requests. The system enters in an state that cannot be recovered from, even with a warm reboot. I believe this was an oversight when reviewed in the kernel. This should be addressed at a lower layer and not in user space.

keyonjie:
In that case, instead of blocking the nvm_authenticate itself for as long as 5 seconds, can we just e.g. put the 'rescan' to be unusable while that is expected, we can e.g. schedule a 5 seconds' delay work which will help to change the 'rescan' back to the usable mode.

westeri:
> > > It is not a workaround to a lower level bug at all. And certainly not random delay. It is all documented in the USB4 re-timer spec and also in the kernel documentation:
> > > https://docs.kernel.org/admin-guide/thunderbolt.html#upgrading-on-board-retimer-nvm-when-there-is-no-cable-connected
> > 
> > 
> > How come nvm_authenticate does not block for the 5 seconds itself? I see that it is documented, not sure that it makes it less of a bad implementation.
> 
> Kernel is allowing for an attacker to simply bring down the expected capability by simply issuing fast requests. The system enters in an state that cannot be recovered from, even with a warm reboot. I believe this was an oversight when reviewed in the kernel. This should be addressed at a lower layer and not in user space.

Not sure I follow. If the "attacker" can access the NVM or these parameters it means it has root privileges and can do much worse than messing with the NVM.

This whole "offline" re-timer update was designed exactly like this that the userspace component knows exactly what it is doing, like bringing the USB4 port into offline mode and waiting for the NVM update to finish e.g following what the USB4 spec tells you to do.

westeri:
> > It is not a workaround to a lower level bug at all. And certainly not random delay. It is all documented in the USB4 re-timer spec and also in the kernel documentation:
> > https://docs.kernel.org/admin-guide/thunderbolt.html#upgrading-on-board-retimer-nvm-when-there-is-no-cable-connected
> 
> How come nvm_authenticate does not block for the 5 seconds itself? I see that it is documented, not sure that it makes it less of a bad implementation.

We don't want to add  long delays like this to kernel drivers.

djcampello:
> Not sure I follow. If the "attacker" can access the NVM or these parameters it means it has root privileges and can do much worse than messing with the NVM.

fwupd is not run as root in Chrome OS precisely to harden against attack vectors. If by any chance an attacker gets access to fwupd user and privileges, it will be able to cause this damage without root privileges (i.e., sysfs entries are owned by fwupd user).

> This whole "offline" re-timer update was designed exactly like this that the userspace component knows exactly what it is doing, like bringing the USB4 port into offline mode and waiting for the NVM update to finish e.g following what the USB4 spec tells you to do.

The specification does not say that if not enough time is giving the whole interface is bricked and not even a reboot of the kernel would bring it back. I could imagine that there would have been complains if that was the case.

> > How come nvm_authenticate does not block for the 5 seconds itself? I see that it is documented, not sure that it makes it less of a bad implementation.
> 
> We don't want to add long delays like this to kernel drivers.

I would normally agree with this statement, but when dealing with an upgrade process that has a `nvm_authenticate` step that blocks O(minutes) flushing in memory buffers to the device, adding 5 seconds for trading off stability in the system sounds like a good trade off to me.

Even if we don't want to go the route of blocking at the kernel level, which I can understand: (a) -EIO does not the seem like the right error to surface in this case, -EBUSY sounds much more appropriate and (b) (and my main concern of addressing this in user space) hitting the error should not leave the system in a unrecoverable state, the -EBUSY error should be retryable (once the expected delay by specification has happened).

I would suggest continue the discussion in the issue partner tracker where more details are embedded.

krankun:
> I would suggest continue the discussion in the issue partner tracker where more details are embedded.

Mika doesn't have access to partner tracker.

westeri:
Please keep in mind that fwupd and Linux kernel are used outside of ChromeOS too.

The USB4 spec gives the steps and the kernel interface makes them available to the userspace. If there is something breaking in the depths of the EC/whatever mux thing the ChromeOS has then of course that needs to be fixed but that definitely does not warrant changing the userspace ABI.

I have no idea where that -EIO is even coming from unless it is the ACPI _DSM that fails to power on the sideband but again that is an implementation detail that can be fixed where it belongs (Coreboot in this case is the one providing ACPI  tables).

djcampello:
> I have no idea where that -EIO is even coming from unless it is the ACPI _DSM that fails to power on the sideband but again that is an implementation detail that can be fixed where it belongs (Coreboot in this case is the one providing ACPI tables).

We agree on this. As it is this PR is trying to address some underlying bug (similar to what you describe here) in user space. I CCed you to the partner issue tracker, hopefully that would grant you access.

djcampello:
> does not warrant changing the userspace ABI.

Is there documentation of what exactly is the ABI in case of failures?

RajmohanMani:
RE: I have no idea where that -EIO is even coming from unless it is the ACPI _DSM that fails to power on the sideband but again that is an implementation detail that can be fixed where it belongs (Coreboot in this case is the one providing ACPI tables).

Hi Mika,
As you guessed correctly, this is coming from coreboot, when it talks to EC to set the required mux (as part of turning the retimer into online mode).

Hi Mika, Daniel,

I see a potential solution with the following approach that prevents the system from reaching the unrecoverable state, while allowing the user space to do a retry, when a retimer online/offline request is made within 5 seconds of retimer nvm authentication.

1) Kernel keeps track of the time when the retimer nvm was last authenticated
2) Kernel returns -EAGAIN if there is a retimer online/offline request within 5 seconds of retimer nvm authentication
3) The user space code upon receiving -EAGAIN, waits for 5 seconds (in one full shot or in units of 1 second) and retries the retimer online/offline command.

This way, the system is not allowed to enter an unrecoverable state and kernel does not have to implement the long delay.
Let me know what you think.

westeri:
> I see a potential solution with the following approach that prevents the system from reaching the unrecoverable state, while allowing the user space to do a retry, when a retimer online/offline request is made within 5 seconds of retimer nvm authentication.
> 
>     1. Kernel keeps track of the time when the retimer nvm was last authenticated

No. What needs to be fixed is the EC/Coreboot to actually handle this properly instead of failing. We don't want to add workarounds anywhere unless we have to. So both kernel and userspace can keep following the USB4 spec steps as intended.

hughsie:
So.... should I merge this or not?

krankun:
> So.... should I merge this or not?

we have discussed with all the component owners , EC/CB  & Kernel and based on the discussions so far , its mostly trending that fwupd is the right place to add this 5 secs delay.  ( EC/CB cannot fix this issue based on conversation above  and Kernel expects user space should wait)

hughsie:
Okay; lets do this here; worse case we wait 2x5s rather than just 5s. I think it's more important it works at this stage.

PR #4297, opened 2022-02-16
mtd: Add two more instance IDs to the MTD devices

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/mtd/fu-mtd-device.c
+++ b/plugins/mtd/fu-mtd-device.c
@@ -28,6 +28,42 @@ fu_mtd_device_to_string(FuDevice *device, guint idt, GString *str)
 }
 
 static gboolean
+fu_mtd_device_setup(FuDevice *device, GError **error)
+{
+	GType firmware_gtype = fu_device_get_firmware_gtype(device);
+	const gchar *fn;
+	g_autoptr(FuFirmware) firmware = NULL;
+	g_autoptr(GFile) file = NULL;
+
+	/* nothing to do */
+	if (firmware_gtype == G_TYPE_INVALID)
+		return TRUE;
+
+	/* read entire contents */
+	fn = fu_udev_device_get_device_file(FU_UDEV_DEVICE(device));
+	if (fn == NULL) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "Not supported as no device file");
+		return FALSE;
+	}
+	file = g_file_new_for_path(fn);
+	firmware = g_object_new(firmware_gtype, NULL);
+	if (!fu_firmware_parse_file(firmware, file, FWUPD_INSTALL_FLAG_NONE, error))
+		return FALSE;
+
+	/* copy over the version */
+	if (fu_firmware_get_version(firmware) != NULL)
+		fu_device_set_version(device, fu_firmware_get_version(firmware));
+	if (fu_firmware_get_version_raw(firmware) != G_MAXUINT64)
+		fu_device_set_version_raw(device, fu_firmware_get_version_raw(firmware));
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
 fu_mtd_device_probe(FuDevice *device, GError **error)
 {
 	FuContext *ctx = fu_device_get_context(device);
@@ -309,6 +345,7 @@ fu_mtd_device_class_init(FuMtdDeviceClass *klass)
 {
 	FuDeviceClass *klass_device = FU_DEVICE_CLASS(klass);
 	klass_device->probe = fu_mtd_device_probe;
+	klass_device->setup = fu_mtd_device_setup;
 	klass_device->to_string = fu_mtd_device_to_string;
 	klass_device->write_firmware = fu_mtd_device_write_firmware;
 }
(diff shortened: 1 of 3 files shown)


hughsie:
@DylanVanAssche

DylanVanAssche:
Hi!

I will test this out soon :)

I gave it some thought about the firmware version reporting and I think I just need to write a plugin called 'platform-firmware' or 'ebbr' (https://arm-software.github.io/ebbr/) which:

- Finds the location of the platform firmware:
  - SPI MTD device
  - eMMC boot partition
  - shared storage (just a protective GPT partition on an eMMC/SD card)
- Fetches the version
- Delegates the flashing to the MTD plugin for MTD devices or a different plugin that does something like `dd`

I think this is more modular and allows us to support not just the Pine64 PinePhone Pro & PineBook Pro (both have SPI MTD), but also Pine64 PinePhone and other SBCs (only eMMC boot or shared storage available). Once we have this, other SBC devices, the Purism Librem 5, etc. are 'automatically' supported.
I won't mind to write it myself ;) I was planning to do it anyway, but wanted some advice first :)

plugins/mtd/fu-uboot-firmware.c
```diff
@@ -0,0 +1,98 @@
...
+	/* set version */
+	version = g_strndup((const gchar *)g_bytes_get_data(fw, NULL) + offset1, offset2);
+	fu_firmware_set_version(firmware, version);
+
+	/* success */
+	return TRUE;
+}
+
+static GBytes *
+fu_uboot_firmware_write(FuFirmware *firmware, GError **error)
```
DylanVanAssche, line 59: I cannot follow here, are we flashing something different than the binary here?

plugins/mtd/fu-mtd-device.c
```diff
@@ -258,6 +327,12 @@ fu_mtd_device_write_firmware(FuDevice *device,
 static void
 fu_mtd_device_init(FuMtdDevice *self)
 {
+	fu_device_set_summary(FU_DEVICE(self), "Memory Technology Device");
+	fu_device_add_protocol(FU_DEVICE(self), "org.infradead.mtd");
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_INTERNAL);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
+	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_ENSURE_SEMVER);
```
DylanVanAssche, line 334: Are we sure we have semantic version everywhere?

plugins/mtd/fu-uboot-firmware.c
hughsie: Ya, this is for hypothetical fuzzing support which I didn't add yet. This allows us to say "write me a UBOOT firmware that looks plausible with this fake data" to allow the fuzzer to do it's thing. If you've got any more info on what the uboot image actually looks like (e.g. offsets and stuff) it makes sense to properly parse the image rather than doing a dumb `memmem()` on a specific string.

plugins/mtd/fu-mtd-device.c
hughsie: That's a good question. I'm pretty sure we don't want to support UBoot versions of things like `1.2.3-cafebeef` as then we'll be doing `strcmp`-style version comparisons and that's not really going to work when `1.12.3-beefcafe` comes along. At least with this we can say for a specific device *we're expecting number.number* and do some sensible version comparisons.

plugins/mtd/fu-uboot-firmware.c
hughsie: Actually; on this; rather than doing the memmem thing; what are we looking at here? It's not the FIT as that's later in the image I think. Is this the SPL that contains the U-Boot version string? i.e. what blob am I seeing at 0x15000 from the SPI dump? Is there a header that specifies the jump to 0x15000 to avoid us reading in a ton of stuff we won't need?

hughsie: I've moved this out to another PR, we can discuss there. I'd still like to know what the layout is. I'm also erring on looking for the *pinephone* firmware version (e.g. a custom firmware header, embedded in the firmware somewhere) rather than a *uboot* version.

plugins/mtd/fu-mtd-device.c
hughsie: I agree with you now; we can add this as a `Flags = ensure-semver` in the same place we set the firmware format.

plugins/mtd/fu-uboot-firmware.c
DylanVanAssche: @hughsie I will be able to dig up more soon, currently swamped with $DAYJOB deadlines. I opened the issue to discuss a bit, didn't expect that things would be moving so fast :joy: I will invite the Tow-Boot maintainer to have a look, he knows more.

hughsie: > didn't expect that things would be moving so fast

Ha! It's **me**, what did you expect :)

DylanVanAssche: Yeah true, you're so fast in fixing things, awesome!

PR #4300, opened 2022-02-17
systemd: don't run fwupd/fwupd-refresh in containers

simondeziel:
Containers typically don't need to run fwupd as they don't have access to the real hardware.


hughsie:
Seems sane to me, @superm1 any comments?

superm1 (approved):

PR #4303, opened 2022-02-18
Try to fix autopkgtests

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -68,7 +68,12 @@ fu_util_using_correct_daemon(GError **error)
 #ifdef HAVE_SYSTEMD
 	g_autofree gchar *default_target = NULL;
 	g_autoptr(GError) error_local = NULL;
-	const gchar *target = fu_util_get_systemd_unit();
+	const gchar *target;
+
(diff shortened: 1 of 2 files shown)


src/fu-util-common.c
```diff
@@ -65,10 +65,14 @@ fu_util_get_expected_command(const gchar *target)
 gboolean
 fu_util_using_correct_daemon(GError **error)
 {
-#ifdef HAVE_SYSTEMD
 	g_autofree gchar *default_target = NULL;
 	g_autoptr(GError) error_local = NULL;
-	const gchar *target = fu_util_get_systemd_unit();
+	const gchar *target;
```
hughsie, line 70: I think you've got to be a bit more careful here e.g. error_local, default_target are unused in the non HAVE_SYSTEMD case.

superm1: fixed.  I forgot we just return TRUE in the non-systemd case

hughsie (dismissed):
If it works, LGTM, thanks.

superm1:
This seems to work for me now if that looks good.

Issue #4299, opened 2022-02-17
autopkgtests are failing in Ubuntu and Debian for 1.7.5


superm1:
It seems that the P2P client is failing.
```
Starting P2P client...
FuPlugin             add_security_attrs(acpi_dmar)
FuPlugin             add_security_attrs(acpi_facp)
FuPluginAcpiFacp     failed to load /sys/firmware/acpi/tables/FACP: Failed to open file "/sys/firmware/acpi/tables/FACP": No such file or directory
FuPlugin             add_security_attrs(iommu)
FuPlugin             add_security_attrs(linux_lockdown)
FuPlugin             add_security_attrs(linux_sleep)
FuPluginLinuxSleep   could not open /sys/power/mem_sleep: Error opening file /sys/power/mem_sleep: No such file or directory
FuPlugin             add_security_attrs(linux_swap)
FuPlugin             add_security_attrs(linux_tainted)
FuPlugin             add_security_attrs(pci_bcr)
FuPlugin             add_security_attrs(pci_mei)
FuPlugin             add_security_attrs(tpm)
FuPlugin             add_security_attrs(uefi_pk)
FAIL: fwupd/fwupdmgr-p2p.test (Child process exited with code 1)
FuMain               Called SetHints()
FuMain               got hint locale=C.UTF-8
[31m[1mWARNING:[0m The daemon has loaded 3rd party code and is no longer supported by the upstream developers!
FuMain               Called GetPlugins()
Mismatched daemon and client, use fwupdmgr instead
FuMain               client connection closed: Underlying GIOStream returned 0 bytes on an async read
```

Here is the full log.  https://ci.debian.net/data/autopkgtest/testing/amd64/f/fwupd/19232275/log.gz

superm1:
In the failure I specifically noticed:
```
Mismatched daemon and client, use fwupdmgr instead
```

Which makes me wonder if we just need to be using `--force` in the p2p test or detect `FWUPD_DBUS_SOCKET` and skip that test (someone would really have have to go out of their way to launch that way and probably doesn't need that infro).

superm1:
I also am surprised we aren't catching this in CI.

hughsie:
Not only that, but this:

> The daemon has loaded 3rd party code and is no longer supported by the upstream developers!

You able to get the daemon logs on a failing system?

superm1:
[...]

PR #4304, opened 2022-02-18
trivial: add allow-stderr to ci autopkgtest

ycheng:
This commit + [87470ea](https://github.com/fwupd/fwupd/commit/87470ea3e010c58d927e0a57be7162c9fc740e73) fix https://github.com/fwupd/fwupd/issues/4299 per my test (against fwupd 1.7.5)

Type of pull request:

- [V] Code fix

### DIFF
--- a/contrib/debian/tests/control
+++ b/contrib/debian/tests/control
@@ -1,5 +1,5 @@
 Tests: ci
-Restrictions: needs-root
+Restrictions: needs-root, allow-stderr
 
 Tests: libfwupd-dev
 Depends: build-essential, libfwupd-dev, pkg-config


superm1:
Do to know what is actually emitted to stderr? I'm a bit surprised this is needed by default and would think it means a failure is hidden.

ycheng:
Hi Mario, If you check https://ci.debian.net/data/autopkgtest/testing/amd64/f/fwupd/19232275/log.gz
Search for "test ci:  - - - - - - - - - - stderr - - - - - - - - - -" you will see lots of stderr output.

AFAIK, for autopkgtest, the default is: if the test job output anything to stderr, then the test failed.
So we need to add this, as --verbose is added to /usr/libexec/fwupd/fwupd in the test script.

Ref: https://github.com/fwupd/fwupd/blob/cda5879974d79e04bce3aa942636c0137c42b242/data/installed-tests/fwupdmgr-p2p.sh#L10

superm1:
#4303 is merged, hopefully that fixes things.  Thanks a lot for the pointer about stderr!

ycheng:
Hard to say if we should remove "--verbose". If a test fails, debugging is very struggling just like this time.
The test setup / env might not be trivial to everyone.

I think provide as log as much as possible shold be the right way to go.

IMO, maybe only 2>&1 only for the command with --verbose.

PR #4306, opened 2022-02-18
contrib/ci: Add libpgiod to dependencies

Sean-StarLabs:

### DIFF
--- a/contrib/ci/dependencies.xml
+++ b/contrib/ci/dependencies.xml
@@ -355,6 +355,27 @@
       <package variant="x86_64" />
     </distro>
   </dependency>
+  <dependency type="build" id="libgpiod">
+    <distro id="centos">
+      <package />
+    </distro>
+    <distro id="fedora">
+      <package />
+    </distro>
+    <distro id="void">
+      <package />
+    </distro>
+    <distro id="debian">
+      <control />
+      <package variant="x86_64" />
+      <package variant="s390x" />
+      <package variant="i386" />
+    </distro>
+    <distro id="ubuntu">
+      <control />
+      <package variant="x86_64" />
+    </distro>
+  </dependency>
   <dependency type="build" id="gcab">
     <distro id="centos">
       <package />


superm1:
Why is this needed?  I thought `linux/gpio.h` comes from the linux headers.  Does this fix anything?

Sean-StarLabs:
Not sure on the first - might be an Ubuntu special as I didn't try building for others.
Fixes building:
```
plugins/gpio/meson.build:6:2: ERROR: Problem encountered: linux/gpio.h (with API v2) is required for plugin_gpio
```

Sean-StarLabs:
Might be more helpful - https://launchpadlibrarian.net/586733816/buildlog_ubuntu-focal-amd64.fwupd_1.7.6+f_BUILDING.txt.gz

hughsie:
Nah, linux/gpio.h is a linux header not a userspace thing. Maybe it makes sense to ship a copy in-tree?

hughsie:
> But so they could just -Dplugin_gpio=true or -Dplugin_gpio=false couldn't they?

I think the idea is that we should define a default and not make the deliverable different depending on what headers are installed on the specific system.

hughsie:
>  I do think we should take a copy of the GPIO header

Isn't that going to be GPLv2 rather than the 99% LGPLv+ and 1% GPLv2+ stuff we have now? Maybe we just fix the message to say "kernel too old"?

superm1:
I didn't think about the licensing implication - good point.  How about we just set the plugin to disabled by default for now then.  The cases that we know build kernel is new enough and it's needed we turn it on.

eli-schwartz:
> I think then we get shouted at by the automagic-hating people: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies

The good news is that Meson's builtin tristate option, "feature" options, has a core builtin option just for automagic-hating people.

```
--auto-features=disabled
```

So that when building projects which have auto detection of dependencies, they can be globally forced on or off.

Automagic dependencies specifically refers to cases where you cannot pass an option to disable it, you need to patch the build files. Manually coding yes/no/check as a tristate in configure.ac is the classic recommended route, and even defaulting to "check" is considered fine. Meson's auto-features takes that a step further.

hughsie:
I also switched around the gpio default. Lets close this one and concentrate on the feature flags.

PR #4308, opened 2022-02-18
Add a flag to indicate the device is signed or unsigned

hughsie:
Devices without either flag are unknown.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 I've asked *dozens* of vendors about the signing status of their devices are. I'll update this branch as the answers come in, but I wanted a sanity check on the "two flags" approach.

superm1:
> A nasty g_warning(), or should we just auto remove the unsigned if the signed flag gets added? I also pondered adding a g_warning() if a device gets added that;s neither signed or unsigned, although that seems a little obnoxious at this stage.

Too soon for warning for things not signed or unsigned.  The only reason that I think two flags makes sense is that exact reason.  It will take time to figure out what needs what.

> Are all UEFI Capsule updates signed? Do they have to be?

Nope.  They don't have to be.  It's up to the vendor what they want to do.

> Are all Dell UEFI capsules signed for all things registered in the ESRT? I know for Lenovo there are things in the ESRT that I suspect are unsigned, e.g. USB webcams and things.

Yes, @CragW can double check nothing changed since I left though.

> What "cryptographic algorithm" is used for UEFI capsules? Is there any way to know?

Impossible to know without digging into the code or asking the OEM.  Nothing in the ESRT advertises this.

superm1:
If we can avoid it I really don't want to be in the habit of having to figure out what cryptographic verification is used.  What happens when an IHV decides to change it?  The code/comment will age and maybe it makes someone look bad for something they don't keep track of and is normally invisible.

If this is important information (at least for the UEFI case), I think going to the USWG and proposing a new ESRT version with all the stuff that is missing is the way to go.  IMO there's a pile of other hacks we have in fwupd (like version format, install order, dependencies etc).  Maybe it's time for ESRT v2 in that case.

hughsie:
> If we can avoid it I really don't want to be in the habit of having to figure out what cryptographic verification is used. 

I guessed the opposite. What if the vendor is using 3DES right now? i.e. it's probably just about good enough in 2022, but if there's a huge hole found in 2025 is it still good enough then?

> The code/comment will age and maybe it makes someone look bad for something they don't keep track of

That's true I suppose. Maybe it's an important question to ask the vendors (to ensure they don't think CRC32 is _signing_) but not to record in the code.

> I think going to the USWG and proposing a new ESRT version

Yes, this is increasingly becoming more and more required. I'll talk to Peter next week and see if there's any appetite for change there.

libfwupdplugin/fu-device.c
```diff
@@ -2968,6 +2972,12 @@ fu_device_add_flag(FuDevice *self, FwupdDeviceFlags flag)
 	if (flag & FWUPD_DEVICE_FLAG_IS_BOOTLOADER)
 		fu_device_remove_flag(self, FWUPD_DEVICE_FLAG_NEEDS_BOOTLOADER);
 
+	/* being both a signed and unsigned is invalid */
+	if (flag & FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD)
+		fu_device_remove_flag(self, FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);
+	if (flag & FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD)
+		fu_device_remove_flag(self, FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD);
```
superm1, line 2979: I don't like this - it's a deterministic failure order and may mask a problem.  Can you also a warning?

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -444,6 +444,28 @@ fu_thunderbolt_device_write_firmware(FuDevice *device,
...
+	/* FuUdevDevice->probe */
+	if (!FU_DEVICE_CLASS(fu_thunderbolt_device_parent_class)->probe(device, error))
+		return FALSE;
+
+	/* if the PCI ID is Intel then it's signed, no idea otherwise */
+	udev_parent = fu_udev_device_get_parent_with_subsystem(FU_UDEV_DEVICE(device), "pci");
+	if (udev_parent != NULL) {
+		if (!fu_device_probe(FU_DEVICE(udev_parent), error))
+			return FALSE;
+		if (fu_udev_device_get_vendor(udev_parent) == 0x8086)
```
superm1, line 461: 8087 too, no?

plugins/vli/fu-vli-device.c
```diff
@@ -426,6 +426,18 @@ fu_vli_device_set_kind(FuVliDevice *self, FuVliDeviceKind device_kind)
...
 	}
 
+	/* newer chips use SHA-256 and ECDSA-256 */
+	switch (device_kind) {
+	case FU_VLI_DEVICE_KIND_VL107:
+	case FU_VLI_DEVICE_KIND_VL650:
+	case FU_VLI_DEVICE_KIND_VL830:
+		fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD);
+		break;
+	default:
```
superm1, line 436: rather than default, can you match older?  Otherwise a newer chip might get this flagged wrong if we forget to populate this block.  I'd rather flag an "unknown" with a warning in the daemon than misattribute it.

superm1 (commented):
I think you're missing dell-dock here.  Everything in that is signed.

libfwupdplugin/fu-device.c
hughsie: Yes, I was worried about that too; but I think it's being a bit pragmatic. For instance the hundreds of PixArt devices are all unsigned, with the exception of one vendor design. In this case I think it makes sense to set unsigned in the .c code and then just "override" it for that one VID/PID in the quirk file.

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: I've not seen that for tbt controllers, but I can certainly add it if you think it's likely.

plugins/vli/fu-vli-device.c
hughsie: There's not any device API to get this metadata yet; I've asked for some to be added. What we could do is all the other ones as another case section and then do the g_warning in the default case.

hughsie:
> I think you're missing dell-dock here

I requested that data from @CragW -- I think he's probably checking that's still true with all the latest stuff. There's still quite a bit of data dribbling in; I'll add you to the spreadsheet of doom.

plugins/vli/fu-vli-device.c
hughsie: > do the g_warning in the default case

I did that.

plugins/thunderbolt/fu-thunderbolt-device.c
hughsie: 8087 is for USB devices only apparently: https://pci-ids.ucw.cz/read/PC/8087

libfwupd/fwupd-enums.c
```diff
@@ -203,6 +203,10 @@ fwupd_device_flag_to_string(FwupdDeviceFlags device_flag)
 		return "affects-fde";
 	if (device_flag == FWUPD_DEVICE_FLAG_END_OF_LIFE)
 		return "end-of-life";
+	if (device_flag == FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD)
+		return "signed-payload";
+	if (device_flag == FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD)
+		return "unsigned-payload";
```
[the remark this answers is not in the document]
hughsie: Or `FWUPD_DEVICE_FLAG_UNVERIFIED_FIRMWARE` and `FWUPD_DEVICE_FLAG_VERIFIED_FIRMWARE` for instance.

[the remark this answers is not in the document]
superm1: `FIRMWARE_IMAGE` maybe?  or `FIRMWARE_PAYLOAD` or `REQUIRES_SIGNED_IMAGE`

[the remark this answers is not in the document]
hughsie: Thinking some more, payload does kinda work, as it's not the firmware as a whole that's being verified. Image we're using as a subsection of the firmware. Maybe we stick to the original proposal?

hughsie:
@superm1 I think my proposal would be to merge this soonish, and then to add the other data as-and-when the replies come back. There's a ton of super useful data already.

superm1:
> @superm1 I think my proposal would be to merge this soonish, and then to add the other data as-and-when the replies come back. There's a ton of super useful data already.

@hughsie something that comes to mind to me is that some of these devices might have a storage area you can write unsigned data to but that it's parsed (not OP codes executed by the firmware).  How do you think that should factor into this?

@CragW at least as it pertains to dell-dock - there are a variety of protocols supported that already made assertions on some of them that can be populated no?
```
fu-dell-dock-hub.c:     fu_device_add_protocol(device, "com.dell.dock");
fu-dell-dock-i2c-ec.c:  fu_device_add_protocol(FU_DEVICE(self), "com.dell.dock");
fu-dell-dock-i2c-mst.c: fu_device_add_protocol(FU_DEVICE(self), "com.synaptics.mst");
fu-dell-dock-i2c-tbt.c: fu_device_add_protocol(FU_DEVICE(self), "com.intel.thunderbolt");
fu-dell-dock-status.c:  fu_device_add_protocol(FU_DEVICE(self), "com.dell.dock");
fu-dell-dock-usb-usb4.c:        fu_device_add_protocol(FU_DEVICE(self), "com.intel.thunderbolt");
```

libfwupd/fwupd-enums.c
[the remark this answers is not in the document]
superm1: fine with me

hughsie:
> some of these devices might have a storage area you can write unsigned data to

I think that's totally fine. At the moment 90% of all the devices are completely unsigned and just "protected" with a CRC32, so it's really low hanging fruit. Dell dock will do remarkably well here.

> fu-dell-dock-i2c-mst.c: fu_device_add_protocol(FU_DEVICE(self), "com.synaptics.mst");

This depends on which family is being used, e.g. Tesla, Leaf etc.

> fu-dell-dock-i2c-tbt.c: fu_device_add_protocol(FU_DEVICE(self), "com.intel.thunderbolt");

I'm assuming this is only ever going to be Intel silicon?

> fu-dell-dock-usb-usb4.c:        fu_device_add_protocol(FU_DEVICE(self), "com.intel.thunderbolt");

And Intel also?

superm1:
> > some of these devices might have a storage area you can write unsigned data to
> 
> I think that's totally fine. At the moment 90% of all the devices are completely unsigned and just "protected" with a CRC32, so it's really low hanging fruit. Dell dock will do remarkably well here.
> 
> > fu-dell-dock-i2c-mst.c: fu_device_add_protocol(FU_DEVICE(self), "com.synaptics.mst");
> 
> This depends on which family is being used, e.g. Tesla, Leaf etc.

Right - agree.  I remember now, this was the only unsigned component in WD19, but Atomic moved to the signed version.
`fu_dell_dock_mst_check_type` can flag it (among other places).

The EC is signed, the hubs are signed.

> 
> > fu-dell-dock-i2c-tbt.c: fu_device_add_protocol(FU_DEVICE(self), "com.intel.thunderbolt");
> 
> I'm assuming this is only ever going to be Intel silicon?

Yes.  It's specific to the generation of Intel TBT controller.

> 
> > fu-dell-dock-usb-usb4.c:        fu_device_add_protocol(FU_DEVICE(self), "com.intel.thunderbolt");
> 
> And Intel also?

At least for now - yeah.  I think if that changes it will be new code anyway.

hughsie:
>  the only unsigned component in WD19, but Atomic moved to the signed version.
> fu_dell_dock_mst_check_type can flag it (among other places).

Super useful, thanks. I've copied the logic from the synaptics MST plugin.

> The EC is signed, the USB hubs are signed.

Got it, thanks. All updated.

superm1 (approved):

hughsie:
@superm1 what do you think of the lvfs-announce text? https://gitlab.com/fwupd/lvfs-website/-/merge_requests/1233

PR #4311, opened 2022-02-19
Update build system to use meson tristate features

superm1:
tristate features will automatically disable if dependencies marked
as required are missing.

Packagers can manually override using `auto_features`.

Link: https://mesonbuild.com/Build-options.html#features

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/gpio/meson.build
```diff
@@ -1,10 +1,7 @@
-if get_option('plugin_gpio')
+if get_option('plugin_gpio').enabled()
```
hughsie, line 1: Is this how you're supposed to use autofeatures? I think when it's set to auto then `enabled()` is going to false, no? I thought the way you were supposed to use it by setting a variable inside the cc.has_header_symbol conditional block? On that, maybe the gpio.h check should move into the gpio plugin directory?

superm1: Oh I see what you mean, it should be `.auto()`.

Yeah I think since this is only plugin using it moving the header check here is reasonable.

superm1:
Fixed that, and also converted 2 others.

contrib/build-openbmc.sh
```diff
@@ -25,13 +25,13 @@ meson ../ \
     -Dplugin_ep963x=false \
     -Dplugin_fastboot=false \
     -Dplugin_logitech_bulkcontroller=false \
-    -Dplugin_nitrokey=false \
-    -Dplugin_nvme=false \
+    -Dplugin_nitrokey=disabled \
+    -Dplugin_nvme=disabled \
```
hughsie, line 29: it does make me a bit uncomfortable having `disabled` and `false` but i see you're busy converting some others too. Keep going and we'll see what it looks like later.

superm1: Before I go too much further - can you let me know what you're thinking of this and what it does?

plugins/flashrom/meson.build
```diff
@@ -1,4 +1,5 @@
-if get_option('plugin_flashrom')
+if get_option('plugin_flashrom').enabled() or \
+   (get_option('plugin_flashrom').auto() and libflashrom.found())
```
hughsie, line 2: I'm not convinced this is the way we're supposed to use the meson feature thing. I think we're supposed to just use `if libflashrom.found()` -- no?

superm1: I don't think so - because then you can end up with a case that flashrom is there but you set plugin_flashrom explicitly to disabled

hughsie: But the docs say `disabled -> do not look for the dependency and always return 'not-found'.` e.g. https://mesonbuild.com/Build-options.html

superm1: oh!  that changes things.  the problem however is going to be when two things depend on it.  I can't see how you set required for two things.

hughsie: agree; i assume you can just look for the dep twice, the second being cached?

superm1: I guess I'll need to experiment a little; it's ambiguous to me what happens.  Like what if you set redfish to enabled and curl to disabled but you happen to have the curl library around - does the daemon get curl support?  Or does it matter which call happened first?

hughsie: > does the daemon get curl support

I'd guess not, but I agree it's a bit underspecified.

superm1: I think this is what is needed; https://mesonbuild.com/Reference-manual_returned_feature.html#featurerequire

hughsie: Be careful tho; that's `since 0.59.0` and we depend on 0.50.0 -- which might not be a problem, but might also be a can of worms.

plugins/dell/meson.build
```diff
@@ -1,4 +1,11 @@
-if get_option('plugin_dell')
+tpm2tss_dell = dependency('tss2-esys', version : '>= 2.0', required: get_option('plugin_dell'))
+libsmbios_c = dependency('libsmbios_c', version : '>= 2.4.0', required: get_option('plugin_dell'))
+
+if tpm2tss_dell.found() and libsmbios_c.found()
+  if not get_option('plugin_uefi_capsule')
+    error('plugin_dell also needs plugin_uefi_capsule to work')
+  endif
```
eli-schwartz, line 7: This check seems a bit reversed. If plugin_dell=auto, then the feature is optional. If the optional feature's *dependencies* are autodetected, but plugin_uefi_capsule which it needs is not auto-selected, you want to disable the optional feature.

If plugin_dell=enabled then you have *three* requirements:
- find tss2-esys
- find libsmbios_c
- select the `plugin_uefi_capsule` option

So what you really want here is:

```meson
if get_option('plugin_dell').enabled() and not get_option('plugin_uefi_capsule')
    error('plugin_dell is force-enabled but plugin_uefi_capsule is false, these options are incompatible')
elif tpm2tss_dell.found() and libsmbios_c.found()
    # optional, requirements met. Continue processing this file and build relevant targets
```

plugins/flashrom/meson.build
eli-schwartz: >  Like what if you set redfish to enabled and curl to disabled but you happen to have the curl library around - does the daemon get curl support? Or does it matter which call happened first?

If curl is disabled, and redfish is enabled, then the libcurl `lookup()` for curl is status "skipped" and does not get cached.

The libcurl lookup for redfish is status "look for it" (and also status "error if not found") so it progresses past the "skipped" early `return self._notfound_dependency()`, gets to the caching code, doesn't find a cached value, and does a full check.

If you have a third libcurl lookup that is auto (look for it, do not fail if not found) then it will do full lookup, cache the result, and persist that result for redfish too. But in that case, if it wasn't found then looking it up again was anyways going to fail.

tl;dr the cache is, or should be based on my reading of the implementation, safe. (If it isn't safe, that's a Meson bug. It is not *allowed* to have any effects other than speeding up successive lookups, so if it is doing something that is not allowed, and breaking the build as a result, that's a definite bug.)

plugins/gpio/meson.build
```diff
@@ -1,9 +1,8 @@
-if get_option('plugin_gpio')
+gpio_header = cc.has_header_symbol('linux/gpio.h', 'GPIO_V2_LINE_FLAG_OUTPUT', required: get_option('plugin_gpio'))
+if get_option('plugin_gpio').enabled() or \
+   (get_option('plugin_gpio').auto() and gpio_header)
```
eli-schwartz, line 3: Isn't the point of `required: get_option('plugin_gpio')` to make this *not* need manual checking of tristate values?

- if enabled: then has_header_symbol must have returned true, or the build would error
- if auto, then has_header_symbol will return true if found, and false if not found
- if disabled, then has_header_symbol immediately returns false without checking

So the only thing that matters is whether `if gpio_header`.

plugins/nitrokey/meson.build
```diff
@@ -1,4 +1,5 @@
-if get_option('plugin_nitrokey')
+if get_option('plugin_nitrokey').auto() or \
+   get_option('plugin_nitrokey').enabled()
```
eli-schwartz, line 2: This too seems like a situation where you want a boolean option (yes/no to build it) rather than a tristate (in the auto case, only build it when dependencies are autodetected).


What is the purpose of auto, here? Is it literally identical to enabled?

plugins/redfish/meson.build
```diff
@@ -1,7 +1,6 @@
-if get_option('plugin_redfish')
-if not get_option('curl')
-  error('curl is required for plugin_redfish')
-endif
+libcurl_redfish = dependency('libcurl', version : '>= 7.56.0', required: get_option('plugin_redfish'))
```
eli-schwartz, line 1: This slightly changes semantics, because before you considered it an error if regular curl support was disabled, but now they are mutually independent options.

That is probably fine assuming you're okay with it. :D The fact that those options used to be bound together can be considered an unfortunate implementation detail, I guess.

I see in a couple other places, you also disentangled options that used to be cascading together but now allow mutually independent behavior.

src/meson.build
```diff
@@ -29,18 +29,18 @@ if get_option('gusb')
...
 
-if get_option('libarchive')
+if libarchive.found()
   daemon_dep += libarchive
 endif
-if get_option('sqlite')
+if sqlite.found()
   daemon_dep += sqlite
   client_dep += sqlite
 endif
```
eli-schwartz, line 38: These can be simplified to unconditionally add them as dependencies, as long as they are defined in the global scope there is no downside to adding a not-found-dependency to the list of deps.

eli-schwartz (changes requested):

plugins/nitrokey/meson.build
superm1: Yeah, I understand much better now, thanks for explaining it.  That was my initial inclination that it should have stayed boolean.  I'll drop the commits for the things that should really stay boolean.

src/meson.build
superm1: Awesome, will adjust it.

meson.build
```diff
@@ -204,44 +204,44 @@ if get_option('gudev')
...
   gudev = dependency('', required : false)
 endif
-if get_option('bluez')
+if not get_option('bluez').disabled()
   conf.set('HAVE_BLUEZ', '1')
 endif
-if get_option('hsi')
+if not get_option('hsi').disabled()
   conf.set('HAVE_HSI', '1')
 endif
```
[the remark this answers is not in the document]
superm1: Yeah, agree.  Will drop those commits and change it back to boolean.

libfwupdplugin/meson.build
```diff
@@ -193,12 +193,12 @@ library_deps = [
...
 
-if get_option('libarchive')
+if libarchive.found()
   library_deps += libarchive
 endif
 
-if get_option('lzma')
-  library_deps += lzma
+if lzma_daemon.found()
+  library_deps += lzma_daemon
```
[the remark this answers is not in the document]
superm1: Awesome, will adjust it.

libfwupdplugin/meson.build
```diff
@@ -193,12 +193,12 @@ library_deps = [
   platform_deps,
 ]
 
-if get_option('libarchive')
+if libarchive.found()
   library_deps += libarchive
```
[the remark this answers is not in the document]
superm1: Awesome, will adjust it.

libfwupd/meson.build
```diff
@@ -92,8 +92,8 @@ if get_option('introspection')
   fwupd_gir_deps = [
       giounix,
   ]
-  if get_option('curl')
-    fwupd_gir_deps += libcurl
+  if libcurl_daemon.found()
+    fwupd_gir_deps += libcurl_daemon
```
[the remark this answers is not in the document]
superm1: Awesome, will adjust it.

libfwupd/meson.build
```diff
@@ -34,8 +34,8 @@ libfwupd_deps = [
     libjsonglib,
 ]
 
-if get_option('curl')
-  libfwupd_deps += libcurl
+if libcurl_daemon.found()
+  libfwupd_deps += libcurl_daemon
```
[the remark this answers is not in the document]
superm1: Awesome, will adjust it.

meson.build
```diff
@@ -352,12 +336,9 @@ if cc.has_header('malloc.h')
 	 conf.set('HAVE_MALLOC_TRIM', '1')
   endif
 endif
-if cc.has_header('cpuid.h') and cc.has_header_symbol('cpuid.h', '__get_cpuid_count') and (host_cpu == 'x86' or host_cpu == 'x86_64')
+if cc.has_header('cpuid.h', required: get_option('plugin_msr')) and \
+   cc.has_header_symbol('cpuid.h', '__get_cpuid_count', required: get_option('plugin_msr'))
```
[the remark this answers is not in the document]
superm1: Yeah; agree, thanks!

meson.build
```diff
@@ -204,44 +204,44 @@ if get_option('gudev')
...
-  if libcurl.version().version_compare('>= 7.62.0')
+  if libcurl_daemon.version().version_compare('>= 7.62.0')
     conf.set('HAVE_LIBCURL_7_62_0', '1')
   endif
 endif
+polkit = dependency('polkit-gobject-1', version : '>= 0.103', required: get_option('polkit'))
 if build_daemon
-  if get_option('polkit')
-    polkit = dependency('polkit-gobject-1', version : '>= 0.103')
+  if polkit.found()
```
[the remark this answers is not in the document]
superm1: > You can simplify checks for polkit.found() elsewhere, using this backward-compat hack:

Thanks for these suggestions.

>The best approach would be Meson 0.59's feature.required() method, and you should leave a TODO comment to that effect so you don't forget. :)

>The second-best option would be, I think, to still only check for it in if build_daemon, but you can still simplify and remove the if get_option('polkit') stuff.

`build_daemon`/`build_standalone`/`build_library` have been built around some coupling and assumptions before we had tristate features.   I'm going to need to look more closely where everything is used and rethink it.

plugins/gpio/meson.build
superm1: Yeah, this was heavy handed.  I fixed it up, but I still need to detangle gudev still.

meson.build
```diff
@@ -352,12 +336,9 @@ if cc.has_header('malloc.h')
 	 conf.set('HAVE_MALLOC_TRIM', '1')
   endif
 endif
-if cc.has_header('cpuid.h') and cc.has_header_symbol('cpuid.h', '__get_cpuid_count') and (host_cpu == 'x86' or host_cpu == 'x86_64')
+has_cpuid = cc.has_header_symbol('cpuid.h', '__get_cpuid_count', required: get_option('plugin_msr'))
+if has_cpuid
```
eli-schwartz, line 340:
```suggestion
if  cc.has_header_symbol('cpuid.h', '__get_cpuid_count', required: get_option('plugin_msr'))
```

I don't think this really needs a temporary variable that is only used once.

meson.build
```diff
@@ -211,40 +211,40 @@ if get_option('hsi')
...
-    warning('Polkit is disabled, the daemon will allow ALL client actions')
+polkit = dependency('polkit-gobject-1', version : '>= 0.103', required: get_option('polkit'))
+if polkit.found()
+  conf.set('HAVE_POLKIT', '1')
+  if polkit.version().version_compare('>= 0.114')
+    conf.set('HAVE_POLKIT_0_114', '1')
   endif
+  conf.set_quoted ('POLKIT_ACTIONDIR', polkit.get_pkgconfig_variable('actiondir'))
+else
+  warning('Polkit is disabled, the daemon will allow ALL client actions')
```
eli-schwartz, line 245: Note that this warning will now be printed even when the daemon doesn't allow anything, because it is not built.

When running meson with `--fatal-meson-warnings` that means you cannot disable the daemon?

superm1: Maybe just the warning needs to be the only thing that moves into `build_daemon` block then like
```
if build_daemon and not polkit.found()
    warning('...)
```

plugins/redfish/meson.build
superm1: > That is probably fine assuming you're okay with it. :D The fact that those options used to be bound together can be considered an unfortunate implementation detail, I guess.

Yup, this was a semantics change I noticed when putting this together too.  I am fine with it.  It's very unlikely that someone would want to disable the daemon curl support but enable the plugin curl support, but yeah now that will work.

> I see in a couple other places, you also disentangled options that used to be cascading together but now allow mutually independent behavior.

Yeah the other really big one I need to fix is gudev.  A bunch of plugins don't "truly" have a compile dependency on it, but if the daemon doesn't have it they won't work.  Probably the last thing I will attack in this PR :)

contrib/ci/oss-fuzz.py
```diff
@@ -26,7 +26,15 @@ def __init__(self) -> None:
...
         self.srcdir = self._ensure_environ("SRC", os.path.realpath(".."))
-        self.ldflags = ["-lpthread", "-lresolv", "-ldl", "-lffi", "-lz", "-llzma"]
+        self.ldflags = [
+            "-lpthread",
+            "-lresolv",
+            "-ldl",
+            "-lffi",
+            "-lz",
+            "-llzma",
+            "-lgnutls",
```
hughsie, line 36: ohh? I thought we wanted to keep this out the fuzzer static binaries to decrease the size?

plugins/dell/meson.build
```diff
@@ -1,4 +1,10 @@
-if get_option('plugin_dell')
+tpm2tss_dell = dependency('tss2-esys', version : '>= 2.0', required: get_option('plugin_dell'))
+libsmbios_c = dependency('libsmbios_c', version : '>= 2.4.0', required: get_option('plugin_dell'))
+
+if get_option('plugin_dell').enabled() and get_option('plugin_uefi_capsule').disabled()
+    error('plugin_dell is force-enabled but plugin_uefi_capsule is disabled, these options are incompatible')
```
hughsie, line 5: there's no cute way of doing this? e.g. `required: get_option('plugin_dell') and get_option('plugin_uefi_capsule'))`

plugins/intel-spi/meson.build
```diff
@@ -1,4 +1,8 @@
-if get_option('plugin_intel_spi')
+lzma_spi = dependency('liblzma', required: get_option('plugin_intel_spi'))
```
hughsie, line 1: a/lzma_spi/lzma_intel_spi perhaps?

plugins/meson.build
```diff
@@ -1,3 +1,15 @@
+plugin_deps = [
+  gio,
+  giounix,
+  gmodule,
+  gudev,
+  gusb,
+  libarchive,
+  libjsonglib,
+  libxmlb,
+  protobufc,
```
hughsie, line 10: maybe add a comment to give us a hint it's not actually a hard dep?

plugins/msr/meson.build
```diff
@@ -1,11 +1,11 @@
-if get_option('hsi') and get_option('plugin_msr')
+if get_option('hsi') and has_cpuid
```
hughsie, line 1: Ohh? No plugin_msr still?

plugins/pixart-rf/meson.build
```diff
@@ -1,8 +1,4 @@
 if get_option('plugin_pixart_rf') and host_machine.system() == 'linux'
-if not get_option('gudev')
-  error('gudev is required for plugin_pixart_rf')
-endif
```
hughsie, line 4: ohh, so we compile it even without gudev. I guess that makes sense.

plugins/synaptics-rmi/fu-synaptics-rmi-common.c
```diff
@@ -15,10 +15,8 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
-#ifdef HAVE_GNUTLS
 #include <gnutls/abstract.h>
 #include <gnutls/crypto.h>
-#endif
```
hughsie, line 21: this was for the fuzzing; fuzzing with gnutls took a really long time.

plugins/dell/meson.build
eli-schwartz: The cute way of doing it is Meson 0.59.0 and the `.require()` method. Using `and` only works with two booleans to yield a boolean, it will involve dropping the advantage of a feature option.

hughsie: Gotcha, thanks.

contrib/ci/oss-fuzz.py
superm1: Oh, I didn't see a comment for that anywhere and so I didn't remember this.  I'll fix it.

plugins/intel-spi/meson.build
superm1: ok

plugins/meson.build
superm1: k

plugins/msr/meson.build
superm1: Nope.
`has_cpuid = cc.has_header_symbol('cpuid.h', '__get_cpuid_count', required: get_option('plugin_msr'))`

plugins/pixart-rf/meson.build
superm1: yeah when we go up to meson 0.59.0, there is some fancy stuff to allow us to drop this and auto-disable the plugin, but this is the best for now

plugins/synaptics-rmi/fu-synaptics-rmi-common.c
superm1: OK - will revert that

superm1:
@hughsie any ideas when we can move to 0.59.0?  I think the big blocker is RHEL backport right?    A lot of this can be cleaner when we can.  But for now I think this is the best that I can come up with.  Assuming CI passes I think it's done now.

hughsie:
This might help: https://rpms.remirepo.net/rpmphp/zoom.php?rpm=meson -- the one I care about is crb-9-stream as if a new fwupd bumps the BR for meson I'm the person that is responsible for bumping the meson package and fixing all the (possibly unrelated) problems in all the other packages that depend on meson. It's the kind of thing that takes weeks of time. :/

hughsie:
@superm1 plugins/synaptics-rmi/meson.build:3:2: ERROR: Problem encountered: gnutls is needed for plugin_synaptics_rmi on the openbmc one.

superm1:
> This might help: https://rpms.remirepo.net/rpmphp/zoom.php?rpm=meson -- the one I care about is crb-9-stream as if a new fwupd bumps the BR for meson I'm the person that is responsible for bumping the meson package and fixing all the (possibly unrelated) problems in all the other packages that depend on meson. It's the kind of thing that takes weeks of time. :/

But so it's at 0.58.2, so not toooo far off.  Is it possible to do a check with all the other packages to see that 0.59 breaks them in any way or not?

hughsie:
>  Is it possible to do a check with all the other packages to see that 0.59 breaks them in any way or not?

Will ponder.

superm1:
> Will ponder.

The other big thing that doing this at the same time as a 0.59 bump buys is the ability to set `deprecated` options.  Then we can let false mean the same as disabled for the newly changed ones.  It will break folks like chrome CI less hard.

eli-schwartz:
The other question to ask is which versions of Debian and Ubuntu, if any, you want to keep support for.

superm1:
> The other question to ask is which versions of Debian and Ubuntu, if any, you want to keep support for.

Debian testing/unstable is 0.61, and Ubuntu 22.04 is the same.  So i'm not worried about Debian.
This means an (implicit) drop of support for Ubuntu 20.04 with the inbox meson, you would need to upgrade via pip first.  If they need to support packaged backports they could revert this (series of) commit(s) too though.

hughsie:
> Debian testing/unstable is 0.61

What about debian stable tho?

superm1:
> > Debian testing/unstable is 0.61
> 
> What about debian stable tho?

0.56.  But fixating on stable isn't worthwhile - it moves glacially.
I would be surprised though if all our other deps allows us to build on debian/stable.

eli-schwartz:
Debian stable isn't going to package a new feature release of some software (and a build system requirements update is not something that happens in a bugfix release). The question is probably more whether users should be able to do so using the distro-available tooling such as meson.

superm1:
> Debian stable isn't going to package a new feature release of some software (and a build system requirements update is not something that happens in a bugfix release). The question is probably more whether users should be able to do so using the distro-available tooling such as meson.

I had a try and it does build with the current stuff.  So yes, if we bump meson version then people will need to upgrade meson to build on debian stable.  But the other thing we can do is add to `contrib/setup` to do that for them with `pip` if the meson version is too old at setup time.  It can be invisible.

superm1:
> > Debian stable isn't going to package a new feature release of some software (and a build system requirements update is not something that happens in a bugfix release). The question is probably more whether users should be able to do so using the distro-available tooling such as meson.
> 
> I had a try and it does build with the current stuff. So yes, if we bump meson version then people will need to upgrade meson to build on debian stable. But the other thing we can do is add to `contrib/setup` to do that for them with `pip` if the meson version is too old at setup time. It can be invisible.

Oh but there's a stable backport for meson in Debian stable.
https://packages.debian.org/source/stable-backports/meson

So yeah I think that's irrelevant to the decision.

hughsie:
Okay, bump the meson version and I'll deal with the RHEL politics. If we want to use the feature stuff we may as well do it properly. Perhaps I should do a new tarball before we land this big change?

superm1:
> Okay, bump the meson version and I'll deal with the RHEL politics. If we want to use the feature stuff we may as well do it properly. Perhaps I should do a new tarball before we land this big change?

OK.  I'll land all the stuff with the improved API then.

I actually don't think we should do a new tarball before, unless you ALSO change the default `plugin_gpio` to false.  That was what prompted all of this anyways :)

superm1:
@eli-schwartz can you hypothesize on this remaining CI failure?  I'm baffled.

eli-schwartz:
Due to the use of `--auto-features=enabled`, libelogind's default "auto" value has been toggled to enabled. And nevertheless cannot be found on systems where systemd is present instead.

Since Meson 0.60, you can do `dependency('foo', 'bar', required: get_option('foo-or-bar'))` and find either one. This is better UX for the end user than having an option which should only be meaningful on a handful of systems.

You could also only check for elogind if systemd did not get found.

meson.build
```diff
@@ -529,7 +535,7 @@ elif get_option('docs') == 'docgen'
     native: true,
     fallback: ['gi-docgen', 'dummy_dep'],
   )
-  gidocgen = find_program('gi-docgen', required: get_option('introspection'))
+  gidocgen = find_program('gi-docgen', required: introspection.allowed())
```
eli-schwartz, line 538: `introspection is a feature option, why not use it directly there?`

superm1:
> Due to the use of `--auto-features=enabled`, libelogind's default "auto" value has been toggled to enabled. And nevertheless cannot be found on systems where systemd is present instead.

But how did this get set?  I didn't set it anywhere I thought.  I guess this makes `.disable_auto_if` more important to get right at least for Fedora and Arch (it failed there too).

> 
> Since Meson 0.60, you can do `dependency('foo', 'bar', required: get_option('foo-or-bar'))` and find either one. This is better UX for the end user than having an option which should only be meaningful on a handful of systems.
> 
> You could also only check for elogind if systemd did not get found.

OK, so combine systemd and elogind to one.  I'll try.

meson.build
superm1: That my inclination but I wasn't sure if the `.disable_auto_if` gets saved into the feature object or it was stateful to the only time I ran it.

eli-schwartz: `get_option('introspection')` will always return the original unmodified feature object. Use of `.disable_auto_if()` creates a cloned copy, and saves that cloned copy for later use any time `introspection` is referenced.

Either way the .allowed() would affect the same thing that find_program(..., required: XXX) would look at.

superm1:
> OK, so combine systemd and elogind to one. I'll try.

based on the decisions made in the code, I think I'd prefer to keep them decoupled.

eli-schwartz:
What you'd probably want to do is check whether logind integration is enabled, not actually caring whether that is implemented via elogind or systemd.

So, `logind_dep = dependency('systemd', 'libelogind', required: get_option('logind'))` and if it is found then you know you have some form of logind support regardless of system.

Yes, you previously looked up systemd, but you can look it up again. ;)

meson.build
superm1: Ah got it, thanks!

superm1:
@hughsie this should be complete now if you want to have another look.

superm1:
@hughsie do you want me to squash this down to a single commit that changes a whole bunch of stuff for tristate options and a few for the deprecated meson stuff?  Or should I keep it this way?

hughsie:
Hmm, I think broken out is easier to understand, but no strong feeling.

superm1:
> Hmm, I think broken out is easier to understand, but no strong feeling.

OK I'm going to squash then, they are all so close that changing an early one means a rebase and changing 15 more.

superm1:
superseded by https://github.com/fwupd/fwupd/pull/4336

PR #4312, opened 2022-02-21
fu-util: prompt devices to stderr

gportay:
Prompting the devices to stdout breaks the JSON output. Prompts to
stderr instead.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

This intends to fix the command `fwupdmgr get-updates --json b585990a-003e-5270-89d5-3705a17f9a43`.

```
[gportay@archlinux fwupd]$ fwupdmgr get-updates --json b585990a-003e-5270-89d5-3705a17f9a43 | jq .
WARNING: The daemon has loaded 3rd party code and is no longer supported by the upstream developers!
parse error: Invalid numeric literal at line 1, column 9
```

```
[gportay@archlinux fwupd]$ make fixme 
build/src/fwupdmgr get-updates --json b585990a-003e-5270-89d5-3705a17f9a43 | jq .
WARNING: The daemon has loaded 3rd party code and is no longer supported by the upstream developers!
Selected device: Integrated Webcam™
{
  "Devices": []
}
```

It is limited to prompting devices only for now. however, it can be generalized to every other commands.

We should probably discuss if prompting should go to stderr. or make sure there is no path that prompt with `--json` (after `if (priv->as_json) { ... }`

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -207,14 +207,14 @@ fu_util_prompt_for_device(FuUtilPrivate *priv, GPtrArray *devices, GError **erro
 	if (devices_filtered->len == 1) {
 		dev = g_ptr_array_index(devices_filtered, 0);
 		/* TRANSLATORS: Device has been chosen by the daemon for the user */
-		g_print("%s: %s\n", _("Selected device"), fwupd_device_get_name(dev));
+		g_printerr("%s: %s\n", _("Selected device"), fwupd_device_get_name(dev));
 		return g_object_ref(dev);
 	}
 
 	/* TRANSLATORS: get interactive prompt */
-	g_print("%s\n", _("Choose a device:"));
+	g_printerr("%s\n", _("Choose a device:"));
 	/* TRANSLATORS: this is to abort the interactive prompt */
-	g_print("0.\t%s\n", _("Cancel"));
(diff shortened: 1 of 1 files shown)


superm1:
IMO I think in json output mode it's fully expected to be run by scripts (IE non-interactive).  So I would suggest fixate on anything needed to change it to work that way for your case.  Maybe set `is_interactive` to false when json output is on?

superm1:
Also I don't think we want to emit those warnings in json mode. Really they should either be a json field or not present

gportay:
Makes sense, I will study the path that leads me to that point.

hughsie:
>  Maybe set is_interactive to false when json output is on

Yes, I think this makes sense.

> Also I don't think we want to emit those warnings in json mode

I think they're okay on stderr, just never on stdout.

gportay:
to sum up: prompting to stderr is fine, so I leave this as is; and I should also check for `is_interactive` as well, right?

superm1:
> to sum up: prompting to stderr is fine, so I leave this as is; and I should also check for `is_interactive` as well, right?

Output to stderr is fine, but I don't think we should prompt from it.

The root of your problem sounds like interactivity with is_json, so that is probably the best way to solve it.

PR #4323, opened 2022-02-22
Add the JCat version to the built and runtime versions

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -7524,6 +7524,9 @@ fu_engine_init(FuEngine *self)
 #if G_USB_CHECK_VERSION(0, 3, 1)
 	fu_engine_add_runtime_version(self, "org.freedesktop.gusb", g_usb_version_string());
 #endif
+#if LIBJCAT_CHECK_VERSION(0, 1, 11)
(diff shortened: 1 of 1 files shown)


src/fu-engine.c
```diff
@@ -7524,6 +7524,9 @@ fu_engine_init(FuEngine *self)
 #if G_USB_CHECK_VERSION(0, 3, 1)
 	fu_engine_add_runtime_version(self, "org.freedesktop.gusb", g_usb_version_string());
 #endif
+#if LIBJCAT_CHECK_VERSION(0, 1, 11)
+	fu_engine_add_runtime_version(self, "com.hughsie.libjcat", jcat_version_string());
+#endif
```
superm1, line 7529: how about for the `#else` case, set it to a string like "less than 0.1.11"?

hughsie: well, we'd still get the build-time version -- and I'd prefer we kept these semvers from a LVFS report point of view.

superm1 (approved):

PR #4324, opened 2022-02-22
vli: Skip probing the Dell DA300 device

hughsie:
Although 2109:2820 and 2109:0820, the firmware revision of 3003 is
specific to that hardware.

Fixes https://github.com/fwupd/fwupd/issues/4305

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -237,6 +237,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "md-set-vendor";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_NO_LID_CLOSED)
 		return "no-lid-closed";
+	if (flag == FU_DEVICE_INTERNAL_FLAG_NO_PROBE)
+		return "no-probe";
 	return NULL;
 }
 
@@ -297,6 +299,8 @@ fu_device_internal_flag_from_string(const gchar *flag)
 		return FU_DEVICE_INTERNAL_FLAG_MD_SET_VENDOR;
 	if (g_strcmp0(flag, "no-lid-closed") == 0)
 		return FU_DEVICE_INTERNAL_FLAG_NO_LID_CLOSED;
+	if (g_strcmp0(flag, "no-probe") == 0)
+		return FU_DEVICE_INTERNAL_FLAG_NO_PROBE;
 	return FU_DEVICE_INTERNAL_FLAG_UNKNOWN;
 }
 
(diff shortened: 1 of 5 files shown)


superm1 (dismissed):

hughsie:
@superm1 have you tested this, or should we wait for @CragW to test on actual hardware?

superm1:
Wait for actual hardware test. My review is code only

superm1:
Wait for actual hardware test. My review is code only

CragW (approved):

hughsie:
@CragW it works?! In which case good job everybody.

CragW:
> @CragW it works?! In which case good job everybody.

Yes, that works on Ubuntu and Chrome. Great job!

libfwupdplugin/fu-device.c
```diff
@@ -237,6 +237,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "md-set-vendor";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_NO_LID_CLOSED)
 		return "no-lid-closed";
+	if (flag == FU_DEVICE_INTERNAL_FLAG_DISABLED)
+		return "disabled";
```
hughsie, line 241: @superm1 I wonder if `FU_DEVICE_INTERNAL_FLAG_IGNORE` might be more appropriate here?

hughsie: Ya, lets do FU_DEVICE_INTERNAL_FLAG_IGNORE -- it's saying that the device _should be_ ignored, not that it's _in_ a disabled state.

hughsie:
I think FWUPD_ERROR_NOT_SUPPORTED is already a g_debug, no?

CragW:
> I think FWUPD_ERROR_NOT_SUPPORTED is already a g_debug, no?

Maybe it's the engine failed probing a device thus it's a warning here.
Screenshot from 2022-02-22 22-44-25

CragW (dismissed):

superm1:
> > I think FWUPD_ERROR_NOT_SUPPORTED is already a g_debug, no?
> 
> Maybe it's the engine failed probing a device thus it's a warning here.
> Screenshot from 2022-02-22 22-44-25
> 
> 

Do you have the verbose debugging messages turned on or anything? I agree that shouldn't be a warning... It's intentional.

CragW:
> Do you have the verbose debugging messages turned on or anything? I agree that shouldn't be a warning... It's intentional.

It was a normal `get-devices` without verbose output.
```
$ sudo ./dist/fwupdtool.sh get-devices
[sudo] password for crabx: 
Loading…                 [-                                      ]16:09:21:0116 FuEngine             failed to probe device usb:03:00:08: disabled
16:09:22:0250 FuEngine             failed to probe device usb:02:00:01: disabled
XPS 13 9310
│
```

superm1:
I pushed another patch to this branch to lower the verbosity for that case if you can check it @hughsie .  Otherwise PR LGTM.

PR #4327, opened 2022-02-22
Do not ask for device when using --json

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -534,6 +534,15 @@ fu_util_prompt_for_device(FuUtilPrivate *priv, GPtrArray *devices_opt, GError **
 		return g_object_ref(dev);
 	}
 
+	/* ignore */
+	if (priv->as_json) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_FOUND,
+				    "Cannot be interactive with --json");
+		return NULL;
+	}
+
 	/* TRANSLATORS: get interactive prompt */
 	g_print("%s\n", _("Choose a device:"));
 	/* TRANSLATORS: this is to abort the interactive prompt */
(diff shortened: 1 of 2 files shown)


superm1 (commented):
I don't agree with this approach.  It means that you get different errors based on if it's launched by a script or by you in shell.

I think `as_json` should factor into the `is_interactive` check, and be a way that forces it not to be interactive anymore.

hughsie:
> I don't agree with this approach. It means that you get different errors based on if it's launched by a script or by you in shell.

True, but this is for when the device is ambiguous. e.g. you typed `fwupdmgr update GUID` and you have two devices with the same GUID.

> I think as_json should factor into the is_interactive check, and be a way that forces it not to be interactive anymore.

I don't get it; how do we guess the device the script actually meant?

superm1:
> True, but this is for when the device is ambiguous. e.g. you typed fwupdmgr update GUID and you have two devices with the same GUID.

Then you should be using the device ID instead.  A script running this in JSON mode can get the devices, figure out the information that indicates it's the right one and then call the command with the right one as an argument.

> I don't get it; how do we guess the device the script actually meant?

You don't.  In my mind it means that if you're operating in json mode you're not allowed to make guesses.
Think of it this way - let's say you're an organization and you run some script that is supposed to check for updates using `fwupdmgr --json` on bootup.  You don't want prompting of course.
Now lets say there is some problem and you want to debug it.  Then you run that same script interactively but things happen to work just fine because it's interactive and it guesses the first device.

You want deterministic behavior when it's in JSON mode.

hughsie:
>  Then you run that same script interactively but things happen to work just fine because it's interactive and it guesses the first device

Ohh it's not doing that; it's returning with an error if we don't just have one matching device; as in return with failure if there is anything more than one "correct" answer.

superm1:
Counter-proposal in #4328

hughsie:
Closing in favour of https://github.com/fwupd/fwupd/pull/4328

PR #4336, opened 2022-02-25
Convert build system to meson tristate features

superm1:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


superm1:
@hughsie that should have the fixed up platform-integrity and intel spi changes to not automatically enable it.  I've also re-coupled redfish to the single curl library declaration as it makes it work more like how gudev/gusb based plugins work now in meson 0.60.

superm1:
I would also like to consider branching 1_7_x before merging this and make this the first of 1.8.x.  That will let 1.7.x family be more easily backported in stable distros.

hughsie (approved):
Merge at will, we can branch on a commit before this afterwards if required. Thanks!

PR #4343, opened 2022-02-28
Draft: modem-manager: detect presence of modem-power driver

DylanVanAssche:
Quectel EG25-G modem in the Pine64 PinePhone and PinePhone Pro
can be booted with the eg25-manager (user space) or modem-power (kernel space).
We only support the eg25-manager for upgrading these modems.

CC: @superm1 

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -38,6 +38,9 @@
 /* Amount of time for the modem to get firmware version */
 #define MAX_WAIT_TIME_SECS 150 /* s */
 
+/* Out-of-tree modem-power driver is unsupported */
+#define MODEM_POWER_SYSFS_PATH "/sys/class/modem-power"
+
 /**
  * FU_MM_DEVICE_FLAG_DETACH_AT_FASTBOOT_HAS_NO_RESPONSE
  *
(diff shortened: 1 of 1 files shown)


superm1 (dismissed):
Looks great, thanks.

DylanVanAssche:
I don't have a kernel available with the `modem-power` driver so testing is a bit hard.
If the driver is present in the kernel, it provides a sysfs entry in `/sys/class/modem-power` like:

- `/sys/class/modem-power/modem-power/is_busy`
- `/sys/class/modem-power/modem-power/powered`
- `/sys/class/modem-power/modem-power/killswitched`
- ...

superm1:
> I don't have a kernel available with the `modem-power` driver so testing is a bit hard. If the driver is present in the kernel, it provides a sysfs entry in `/sys/class/modem-power` like:
> 
> * ` /sys/class/modem-power/modem-power/device/is_busy`
> * `/sys/class/modem-power/modem-power/device/powered`
> * `/sys/class/modem-power/modem-power/device/killswitched`
> * ...
> 
> Source: https://github.com/megous/linux/blob/modem-5.17/drivers/misc/modem-power.c

yeah I think that's sufficient.

Fixup the CI failures and then we can merge this.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1171,6 +1174,24 @@ fu_mm_device_get_firmware_version_mbim(FuDevice *device, GError **error)
...
 }
 
+static FuFirmware *
+fu_mm_device_prepare_firmware(FuDevice *device,
+		              GBytes *fw,
+		              FwupdInstallFlags flags,
+		              GError **error)
+{
+	/* detect presence of unsupported modem-power driver */
+	if (g_file_test (MODEM_POWER_SYSFS_PATH, G_FILE_TEST_EXISTS) {
```
superm1, line 1184: missing parenthesis

DylanVanAssche:
My clang-formatter was acting up, fixed :)

superm1 (approved):
OK, as long as CI passes now LGTM.

hughsie (changes requested):
I think these need to be device inhibits. E.g added in plugin.c when the device is added. I can do an example when back home. Just before the update is deployed is too late and will scare the user.

superm1:
> I think these need to be device inhibits. E.g added in plugin.c when the device is added. I can do an example when back home. Just before the update is deployed is too late and will scare the user.

The thing about making it an inhibit is that a kernel module can (potentially) be loaded after the daemon starts or before it quits.  So unless you do an `inotify` on `/sys/class` how will you catch that?

hughsie:
> So unless you do an inotify on /sys/class how will you catch that?

I think that's exactly what we should do. Alternate proposal in https://github.com/fwupd/fwupd/pull/4353

PR #4351, opened 2022-03-01
Allow devices to build instance IDs more easily

hughsie:
Provide a device instance builder that allows plugins to easily
create multiple instance IDs based on parent attributes.

Also fix a lot of the instance ID orders, so that we add more generic
IDs first, and more specific IDs after.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-bluez-device.c
+++ b/libfwupdplugin/fu-bluez-device.c
@@ -171,24 +171,19 @@ fu_bluez_device_set_modalias(FuBluezDevice *self, const gchar *modalias)
 		fu_firmware_strparse_uint16_safe(modalias, modaliaslen, 21, &rev, NULL);
 	}
 
-	if (vid != 0x0 && pid != 0x0 && rev != 0x0) {
-		g_autofree gchar *devid = NULL;
-		devid = g_strdup_printf("BLUETOOTH\\VID_%04X&PID_%04X&REV_%04X", vid, pid, rev);
-		fu_device_add_instance_id(FU_DEVICE(self), devid);
-	}
-	if (vid != 0x0 && pid != 0x0) {
-		g_autofree gchar *devid = NULL;
-		devid = g_strdup_printf("BLUETOOTH\\VID_%04X&PID_%04X", vid, pid);
-		fu_device_add_instance_id(FU_DEVICE(self), devid);
-	}
+	/* add generated IDs */
+	if (vid != 0x0)
+		fu_device_add_instance_u16(FU_DEVICE(self), "VID", vid);
+	if (pid != 0x0)
+		fu_device_add_instance_u16(FU_DEVICE(self), "PID", pid);
+	fu_device_add_instance_u16(FU_DEVICE(self), "REV", rev);
+	fu_device_build_instance_id_quirk(FU_DEVICE(self), NULL, "BLUETOOTH", "VID", NULL);
+	fu_device_build_instance_id(FU_DEVICE(self), NULL, "BLUETOOTH", "VID", "PID", NULL);
+	fu_device_build_instance_id(FU_DEVICE(self), NULL, "BLUETOOTH", "VID", "PID", "REV", NULL);
+
+	/* set vendor ID */
 	if (vid != 0x0) {
-		g_autofree gchar *devid = NULL;
-		g_autofree gchar *vendor_id = NULL;
-		devid = g_strdup_printf("BLUETOOTH\\VID_%04X", vid);
-		fu_device_add_instance_id_full(FU_DEVICE(self),
-					       devid,
-					       FU_DEVICE_INSTANCE_FLAG_ONLY_QUIRKS);
-		vendor_id = g_strdup_printf("BLUETOOTH:%04X", vid);
+		g_autofree gchar *vendor_id = g_strdup_printf("BLUETOOTH:%04X", vid);
 		fu_device_add_vendor_id(FU_DEVICE(self), vendor_id);
 	}
 }
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -76,6 +76,7 @@ typedef struct {
 	GPtrArray *private_flag_items; /* (nullable) */
 	gchar *custom_flags;
 	gulong notify_flags_handler_id;
+	GHashTable *instance_hash;
 } FuDevicePrivate;
 
 typedef struct {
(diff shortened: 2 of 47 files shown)


libfwupdplugin/fu-bluez-device.c
```diff
@@ -171,24 +171,19 @@ fu_bluez_device_set_modalias(FuBluezDevice *self, const gchar *modalias)
...
-		fu_device_add_instance_id(FU_DEVICE(self), devid);
-	}
-	if (vid != 0x0 && pid != 0x0) {
-		g_autofree gchar *devid = NULL;
-		devid = g_strdup_printf("BLUETOOTH\\VID_%04X&PID_%04X", vid, pid);
-		fu_device_add_instance_id(FU_DEVICE(self), devid);
-	}
+	/* add generated IDs */
+	if (vid != 0x0)
+		fu_device_instance_assign_uint16(FU_DEVICE(self), "VID", vid);
```
hughsie, line 176: @superm1 I did ponder calling this fu_device_instance_assign_u16 to make it shorter, but we use uint16 in other parts of the code.

libfwupdplugin/fu-device.c
```diff
@@ -4735,6 +4739,246 @@ fu_device_flags_notify_cb(FuDevice *self, GParamSpec *pspec, gpointer user_data)
...
+
+	va_start(args, subsystem);
+	for (guint i = 0;; i++) {
+		const gchar *key = va_arg(args, const gchar *);
+		const gchar *value;
+		if (key == NULL)
+			break;
+		value = fu_device_instance_lookup(self, key);
+		if (value == NULL && parent != NULL)
+			value = fu_device_instance_lookup(parent, key);
```
hughsie, line 4930: we could also check the proxy here I suppose.

libfwupdplugin/fu-bluez-device.c
superm1: Maybe it makes sense to go the shorter approach and drop a comment for the next time we break ABI to rename the others?

hughsie: Makes sense to me. The other thing I pondered was to rename `fu_device_instance_assign_safe` to `fu_device_instance_assign` and  `fu_device_instance_assign` to  `fu_device_instance_assign_unsafe` just so it's obvious we're using the unsafe one. Or maybe the latter should be  `fu_device_instance_assign_raw` to show we're not going to be substituting spaces and such.

hughsie: Also, maybe `fu_device_instance_assign` to `fu_device_instance_add` might be better as well and more in keeping? Ideas welcome.

libfwupdplugin/fu-udev-device.c
```diff
@@ -459,71 +455,37 @@ fu_udev_device_probe(FuDevice *device, GError **error)
...
+		    g_strdup_printf("%04X%04X", priv->subsystem_vendor, priv->subsystem_model);
+		fu_device_instance_assign(device, "SUBSYS", subsys);
+	}
+	fu_device_instance_assign_uint8(device, "REV", priv->revision);
+
+	fu_device_instance_build_quirk(device, subsystem, "VEN", NULL);
+	fu_device_instance_build(device, subsystem, "VEN", "DEV", NULL);
+	fu_device_instance_build(device, subsystem, "VEN", "DEV", "REV", NULL);
+	fu_device_instance_build(device, subsystem, "VEN", "DEV", "SUBSYS", NULL);
+	fu_device_instance_build(device, subsystem, "VEN", "DEV", "SUBSYS", "REV", NULL);
```
superm1, line 473: there's something subtly different here - the old code wouldn't make any IDs if you didn't have a vendor.  This code will now try to make an ID even with vendor not set.

hughsie: Nope, if `priv->vendor == 0x0000` then we don't call `fu_device_instance_assign_uint16(device, "VEN", priv->vendor);` and if the key isn't set then we return from `fu_device_instance_build` without adding anything. Right?

libfwupdplugin/fu-device.c
hughsie: I added this, it cleaned up CFU nicely.

libfwupdplugin/fu-bluez-device.c
superm1: function names entirely up to you here, I don't feel strongly.

libfwupdplugin/fu-udev-device.c
superm1: Oh I guess that's true, but maybe the thought process that lead me here means that `fu_device_instance_build` should not be `void`.  If no strings are going to be made, I would expect some sort of failure that the caller would know about rather than silently doing nothing.

hughsie: Returning just True/False, or setting an optional GError too? I guess it makes sense, as some plugins *need* specific IDs to make sense.

superm1: GError too make, it will make our lives easier in the future maybe

hughsie: Agree, this works well. Fixed.

libfwupdplugin/fu-bluez-device.c
```diff
@@ -171,24 +171,19 @@ fu_bluez_device_set_modalias(FuBluezDevice *self, const gchar *modalias)
...
+		fu_device_add_instance_u16(FU_DEVICE(self), "VID", vid);
+	if (pid != 0x0)
+		fu_device_add_instance_u16(FU_DEVICE(self), "PID", pid);
+	fu_device_add_instance_u16(FU_DEVICE(self), "REV", rev);
+	fu_device_build_instance_id_quirk(FU_DEVICE(self), NULL, "BLUETOOTH", "VID", NULL);
+	fu_device_build_instance_id(FU_DEVICE(self), NULL, "BLUETOOTH", "VID", "PID", NULL);
+	fu_device_build_instance_id(FU_DEVICE(self), NULL, "BLUETOOTH", "VID", "PID", "REV", NULL);
+
+	/* set vendor ID */
 	if (vid != 0x0) {
```
superm1, line 185: maybe this makes sense to just combine with the above `vid != 0)`

superm1 (approved):

libfwupdplugin/fu-bluez-device.c
hughsie: I wanted to make it optional in case it was missing here, but we inherited the VID from a parent or proxy.

PR #4353, opened 2022-03-02
modem-manager: Detect presence of modem-power driver

hughsie:
Quectel EG25-G modem in the Pine64 PinePhone and PinePhone Pro can be
booted with the eg25-manager (user space) or modem-power (kernel space).

We only support the eg25-manager for upgrading these modems.

Based on a patch by Dylan Van Assche <me@dylanvanassche.be>, thanks.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-plugin-modem-manager.c
+++ b/plugins/modem-manager/fu-plugin-modem-manager.c
@@ -18,10 +18,14 @@
 /* amount of time to wait for ports of the same device	being exposed by kernel */
 #define FU_MM_UDEV_DEVICE_PORTS_TIMEOUT 3 /* s */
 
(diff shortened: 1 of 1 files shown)


superm1 (approved):
It's heavier handed than I think we should do for something out of tree, but LGTM.

DylanVanAssche:
@superm1 Yeah indeed, but LGTM as well!

DylanVanAssche (approved):

hughsie:
> It's heavier handed than I think we should do for something out of tree

Ya, but I figured half the deployments of that hardware that are going to see a new enough fwupd are going to be affected, so it was worth the pain.

Issue #4352, opened 2022-03-01
vli plugin probing prevents Targus USB3.0 SV Docking Station enumeration


djcampello:
The following entry in `plugins/vli/vli.quirk` prevents the enumeration of the `Targus USB3.0 SV Docking Station`:

```
GType = FuVliUsbhubDevice
[USB\VID_2109&PID_0812]
Plugin = vli
```

With the entry:

```
??USB3.0 Hub:
?     Device ID:          7c5b6dadf8126e3820c86cb56d9e1777084a98a8
?     Summary:            USB 2.x hub
?     Current version:    6.94
?     Vendor:             TARGUS (USB:0x2109)
?     Install Duration:   10 seconds
?     GUIDs:              039ee6a4-d140-5714-a480-a96b2e7cd638 ? USB\VID_2109&PID_0812&REV_0694
?                         7aaf809e-a3f1-5beb-8a28-09e44471534b ? USB\VID_2109&PID_0812
?                         08d0f0c8-f42d-55db-a971-b657bb7636ed ? USB\VID_2109&PID_0812&HUB_0002
?                         3e777b40-15fb-51cc-a3fe-ec2fcc5245f0 ? USB\VID_2109&PID_0812&SPI_C220&REV_0694
?                         cd3a7a5a-f5cf-546e-8f18-0a758327aa3e ? USB\VID_2109&PID_0812&SPI_C220
?     Device Flags:       ? Updatable
?                         ? Cryptographic hash verification is available
```

Without the problematic entry:

```
??Targus USB3.0 SV Docking Station:
?     Device ID:          0e399126586fa6c849149e941cb6c86b6f75377f
?     Current version:    30.39
?     Vendor:             DisplayLink (USB:0x17E9)
?     Serial Number:      10002159
?     GUIDs:              6a577e80-133e-54c8-bac9-771463481f5c ? USB\VID_17E9&PID_4306&REV_3039
?                         1d7be0f1-6d7e-5a29-a7d5-4290f6c4dc48 ? USB\VID_17E9&PID_4306
?     Device Flags:       ? Updatable
```

Expected kernel log entries (vli disabled):

```
2022-02-25T08:30:24.303143Z INFO kernel: [ 1569.605828] usb 2-3: new SuperSpeed USB device number 9 using xhci_hcd
[...]

PR #4355, opened 2022-03-02
vli: Do not probe devices with the firmware reference VID and PID

hughsie:
But, leave the products with the VIA VID, but a non-default PID.

Fixes https://github.com/fwupd/fwupd/issues/4352

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


djcampello:
I do not have hardware at hand to test, but looks good to me!

djcampello (approved):

djcampello:
Should we merge this already?

hughsie:
I forgot to, in all honesty. In the future feel free to merge stuff like this if you're asked for approval, and you approve it. I'm not a big scary maintainer, and there's always git revert 🙂

djcampello:
Sounds good! Thank you!

memily:
Why howdy,

> I do not have hardware at hand to test, but looks good to me!

I tested the Dell DA300 device with both firmware versions x09F3/x3003 today. Result being PASS. Neither appeared on `./build/src/fwupdtool get-devices` or `./build/src/fwupdmgr get-devices` (using fwupd github version). We will continue to test other devices into next week to verify that devices with VID 0x2109 will not be seen or allowed to update (devices outside whitelist exception). Please stand by! Thanks.

Issue #4360, opened 2022-03-06
fwupdmgr-p2p.sh should use `@libexecdir@/fwupd/fwupd`


tomfitzhenry:
While packaging fwupd 1.7.6 for NixOS, a review suggested that https://github.com/fwupd/fwupd/blob/777103f93a641a932277158fe86139f0e3916e8c/data/installed-tests/fwupdmgr-p2p.sh#L13 should use `@libexecdir@/fwupd/fwupd` rather than `/usr/libexec/fwupd/fwupd`, along with an appropriate meson directive.

This is useful to allow distributes to run this test during its builds, while allowing for a configuration libexec directory.

hughsie:
Yup, makes sense. Do you want to have a go at this and do a PR? Thanks.

PR #4361, opened 2022-03-07
Do not hardcode the libexecdir to /usr/libexec

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4360

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/installed-tests/fwupdmgr-p2p.sh
+++ b/data/installed-tests/fwupdmgr-p2p.sh
@@ -8,7 +8,7 @@ if [ "$(id -u)" -ne 0 ]; then exit 0; fi
 
 # ---
 echo "Starting P2P daemon..."
-export FWUPD_DBUS_SOCKET="/var/run/fwupd.sock"
+export FWUPD_DBUS_SOCKET="/run/fwupd.sock"
 rm -rf ${FWUPD_DBUS_SOCKET}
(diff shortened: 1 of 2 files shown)


data/installed-tests/fwupdmgr-p2p.sh
```diff
@@ -10,7 +10,7 @@ if [ "$(id -u)" -ne 0 ]; then exit 0; fi
 echo "Starting P2P daemon..."
 export FWUPD_DBUS_SOCKET="/var/run/fwupd.sock"
```
superm1, line 11: If you're processing the file anyway shouldn't this be swapped out by configuration variable too?

hughsie: ohh hmm. Do we have any config for the socket location now? In the common code we have:

	/* /run/lock */
	case FU_PATH_KIND_LOCKDIR:
		return g_strdup("/run/lock");

superm1: I think specifically systems directory to use for configuration (/run) can it be discovered using systemd pkgconfig?

hughsie: The twitterati seems to think that we should just hardcode /run now. I've made that change as a followup.

superm1 (approved):

PR #4362, opened 2022-03-07
plugin/flashrom: Add MaxSize quirk for Lite II and III

Sean-StarLabs:
Add FirmwareSizeMax = 0x800000 for StarLite Mk II and Mk III to allow
switching from AMI to coreboot firmware.


hughsie (approved):
If it works, LGTM, thanks!

PR #4363, opened 2022-03-08
trivial: temporarily disable lvfs signing to proceed BIOS.exe installation

CragW:
trivial: temporarily disable lvfs signing to proceed BIOS.exe installation

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/firmware_packager/install_dell_bios_exe.py
+++ b/contrib/firmware_packager/install_dell_bios_exe.py
@@ -16,7 +16,7 @@ except ValueError:
     print("Missing gobject-introspection packages.  Try to install gir1.2-fwupd-2.0.")
     sys.exit(1)
 from gi.repository import Fwupd  # pylint: disable=wrong-import-position
-from simple_client import install, check_exists
+from simple_client import get_daemon_property, install, check_exists, modify_config
 from add_capsule_header import add_header
 from firmware_packager import make_firmware_metainfo, create_firmware_cab
 
@@ -81,6 +81,19 @@ def find_uefi_device(client, deviceid):
     sys.exit(1)
 
 
+def set_conf_only_trusted(client, setval):
+    prop = "OnlyTrusted"
+    current_val = get_daemon_property(prop)
+    if current_val:
+        pass
+    elif setval:
+        pass
+    else:
+        return False
+    modify_config(client, prop, str(setval).lower())
+    return get_daemon_property(prop) == setval
+
+
 def prompt_reboot():
     print("An update requires a reboot to complete")
     while True:
@@ -101,8 +114,15 @@ if __name__ == "__main__":
     ARGS = parse_args()
     CLIENT = Fwupd.Client()
     check_exists(ARGS.exe)
-    directory = tempfile.mkdtemp()
-    guid, deviceid, version = find_uefi_device(CLIENT, ARGS.deviceid)
-    cab = generate_cab(ARGS.exe, directory, guid, version)
-    install(CLIENT, cab, deviceid, True, True)
+    try:
+        is_restore_required = set_conf_only_trusted(CLIENT, False)
+        directory = tempfile.mkdtemp()
+        guid, deviceid, version = find_uefi_device(CLIENT, ARGS.deviceid)
+        cab = generate_cab(ARGS.exe, directory, guid, version)
+        install(CLIENT, cab, deviceid, True, True)
+    except Exception as e:
+        print(e)
+
+    if is_restore_required:
+        set_conf_only_trusted(CLIENT, True)
     prompt_reboot()
--- a/contrib/firmware_packager/simple_client.py
+++ b/contrib/firmware_packager/simple_client.py
@@ -3,6 +3,7 @@
 """A simple fwupd frontend"""
 import sys
 import os
+import dbus
 import gi
 from gi.repository import GLib
 
@@ -110,6 +111,16 @@ def device_changed(client, device, progress):  # pylint: disable=unused-argument
     progress.device_changed(device.get_name())
 
 
+def modify_config(client, key, value):
+    """Use fwupd client to modify daemon configuration value"""
+    try:
+        print("setting configuration key %s to %s" % (key, value))
+        client.modify_config(key, value, None)
+    except Exception as e:
+        print("%s" % str(e))
+        sys.exit(1)
+
+
 def install(client, cab, target, older, reinstall):
     """Use fwupd client to install CAB file to applicable devices"""
     # FWUPD_DEVICE_ID_ANY
@@ -132,6 +143,25 @@ def install(client, cab, target, older, reinstall):
         print("%s" % glib_err)
         sys.exit(1)
     print("\n")
+
+
+def get_daemon_property(key: str):
+    try:
+        bus = dbus.SystemBus()
+        proxy = bus.get_object(bus_name="org.freedesktop.fwupd", object_path="/")
+        iface = dbus.Interface(proxy, "org.freedesktop.DBus.Properties")
+        val = iface.Get("org.freedesktop.fwupd", key)
+        if isinstance(val, dbus.Boolean):
+            print(
+                "org.freedesktop.fwupd property %s, current value is %s"
+                % (key, bool(val))
+            )
+        else:
+            print("org.freedesktop.fwupd property %s, current value is %s" % (key, val))
+        return val
+    except dbus.DBusException as e:
+        print(e)
+    return None
 
 
 def check_exists(cab):


CragW:
Do we have dbus method call something like `Get-Config` to read the daemon configuration? I was thinking to backup the setting and restore it when finished.

hughsie:
Hmm. I do see the need for this, but it does make me a bit uncomfortable...

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -40,9 +40,19 @@ def parse_args():
...
     parser.add_argument("exe", nargs="?", help="exe file")
     parser.add_argument("deviceid", nargs="?", help="DeviceID to operate on(optional)")
+    parser.add_argument(
+        "-o",
+        "--off-signing",
+        dest="off_signing",
+        default=False,
+        action="store_true",
+        help="disable lvfs signing after the installation, enable if unspecified",
+    )
```
superm1, line 55: I don't like off-signing.  I don't know what it means, even after reading the help string.  Maybe:

* `disable-fwupd-signing-requirement`
* default: `True`
* help: `disable daemon enforcement of signed CAB files`

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -101,8 +111,11 @@ def prompt_reboot():
     ARGS = parse_args()
     CLIENT = Fwupd.Client()
     check_exists(ARGS.exe)
+    modify_config(CLIENT, "OnlyTrusted", "false")
     directory = tempfile.mkdtemp()
     guid, deviceid, version = find_uefi_device(CLIENT, ARGS.deviceid)
     cab = generate_cab(ARGS.exe, directory, guid, version)
     install(CLIENT, cab, deviceid, True, True)
+    if not ARGS.off_signing:
+        modify_config(CLIENT, "OnlyTrusted", "true")
```
superm1, line 120: Maybe only restore it if it got changed

superm1 (commented):
I think you need to do a little bit more with making sure it's only changed when needed

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -101,8 +111,11 @@ def prompt_reboot():
     ARGS = parse_args()
     CLIENT = Fwupd.Client()
     check_exists(ARGS.exe)
+    modify_config(CLIENT, "OnlyTrusted", "false")
```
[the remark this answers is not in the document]
CragW: It looks like we don't have `Get-Config` to read the current value, I'll see if i can make up one.

contrib/firmware_packager/install_dell_bios_exe.py
CragW: Hmm.. disabling the signing requirement is likely enforced otherwise can't move forward.  My original plan of this `arg` is to determine what value should be set after an installation.

superm1: If you can read old value and just restore it you can drop this all together then

hughsie: I think you can rebase this off main again, right?

contrib/firmware_packager/simple_client.py
```diff
@@ -110,6 +111,16 @@ def device_changed(client, device, progress):  # pylint: disable=unused-argument
     progress.device_changed(device.get_name())
 
 
+def modify_config(client, key, value):
+    """Use fwupd client to modify daemon configuration value"""
+    try:
+        client.modify_config(key, value, None)
+    except GLib.Error as glib_err:
+        print("%s" % glib_err)
+        sys.exit(1)
```
hughsie, line 120: I think the convention is to call the exception `e` and then to do something like `print(str(e))`

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -101,6 +108,7 @@ def prompt_reboot():
     ARGS = parse_args()
     CLIENT = Fwupd.Client()
     check_exists(ARGS.exe)
+    early_exit_if_trusted_only()
```
superm1, line 111: Weren't you going to try to do this automatically and then only exit if you couldn't?

CragW: > Weren't you going to try to do this automatically and then only exit if you couldn't?

Yeah honestly if the `prop` is known, I was thinking to turn it off silently and restore it when finished. On the contrary what if it is unknown, between the listed options which is better? Ideas welcome.

1. let it be ultimately users will see the update failure
2. yes/no question to turn it off

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -81,6 +81,17 @@ def find_uefi_device(client, deviceid):
     sys.exit(1)
 
 
+def set_conf_only_trusted(client, setval):
+    prop = "OnlyTrusted"
+    curval = get_daemon_property(prop)
+    if curval is None:
+        print("Can't determine %s via D-Bus, please turn it off manually" % prop)
+    elif curval or setval:
+        modify_config(client, prop, str(setval).lower())
```
hughsie, line 90: client.modify_config( surely?

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -81,6 +81,17 @@ def find_uefi_device(client, deviceid):
...
+        return get_daemon_property(prop) == setval
```
superm1, line 91: I think the daemon needs to be restarted after making this change doesn't it?  So it shouldn't just read that new value immediately.

CragW: With some tests here around `./dist/fwupd.sh` and `./contrib/firmware_packager/install_dell_bios_exe.py`, the property change gets emitted and notified upon writing to `./dist/etc/fwupd/daemon.conf` completed, it seems the daemon stats the configuration file therefore changes will takes effect immediately.

superm1:
Are you planning to rebase this to latest?

CragW:
> Are you planning to rebase this to latest?

Done, thanks.

contrib/firmware_packager/install_dell_bios_exe.py
superm1: @CragW can you please fix this?

contrib/firmware_packager/install_dell_bios_exe.py
```diff
@@ -101,8 +112,11 @@ def prompt_reboot():
     ARGS = parse_args()
     CLIENT = Fwupd.Client()
     check_exists(ARGS.exe)
+    is_restore_required = set_conf_only_trusted(CLIENT, False)
     directory = tempfile.mkdtemp()
     guid, deviceid, version = find_uefi_device(CLIENT, ARGS.deviceid)
     cab = generate_cab(ARGS.exe, directory, guid, version)
     install(CLIENT, cab, deviceid, True, True)
+    if is_restore_required:
+        set_conf_only_trusted(CLIENT, True)
```
superm1, line 121: I think it would be good to do a try/except/finally here to ensure that even if the install failed for any reason, that you still restore the configuration back to normal.

contrib/firmware_packager/simple_client.py
```diff
@@ -110,6 +111,16 @@ def device_changed(client, device, progress):  # pylint: disable=unused-argument
     progress.device_changed(device.get_name())
 
 
+def modify_config(client, key, value):
+    """Use fwupd client to modify daemon configuration value"""
+    try:
+        client.modify_config(key, value, None)
```
superm1, line 117: I think you should emit something to stdout that this happened.  It shouldn't be FULLY invisible to the user that they're waiving security to do this.

CragW:
Hereunder is the user interaction in stdout, added some prints to explicit what the script is doing. Also, root privilege is required to confirm the change introduced by `modify_config`.

```
$ ./install_dell_bios_exe.py ~/Downloads/XPS_13_9310_3.4.0.exe 
org.freedesktop.fwupd property OnlyTrusted, current value is True
setting configuration key OnlyTrusted to false
org.freedesktop.fwupd property OnlyTrusted, current value is False
Installing to System Firmware
Wrote capsule /tmp/tmp4ak9plns/firmware.bin
GUID:      cfb08d7c-8d3d-41b1-846b-48f634dfac9d
HdrSz:     0x1000
Flags:     0x70000
PayloadSz: 0x31049a8
Generated CAB file /tmp/tmp4ak9plns/firmware.cab


org.freedesktop.fwupd property OnlyTrusted, current value is False
setting configuration key OnlyTrusted to true
org.freedesktop.fwupd property OnlyTrusted, current value is True
An update requires a reboot to complete
Restart now? (Y/N) 
```

superm1 (approved):

PR #4371, opened 2022-03-09
trivial: Add daemon configuration key 'OnlyTrusted' to dbus Properties

CragW:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -66,6 +66,7 @@ typedef struct {
 	gchar *host_product;
 	gchar *host_machine_id;
 	gchar *host_security_id;
+	gboolean only_trusted;
 	GMutex proxy_mutex; /* for @proxy */
 	GDBusProxy *proxy;
 	GProxyResolver *proxy_resolver;
@@ -108,6 +109,7 @@ enum {
 	PROP_HOST_SECURITY_ID,
 	PROP_HOST_BKC,
 	PROP_INTERACTIVE,
+	PROP_ONLY_TRUSTED,
 	PROP_LAST
 };
 
@@ -421,6 +423,14 @@ fwupd_client_properties_changed_cb(GDBusProxy *proxy,
 		if (val != NULL)
 			fwupd_client_set_host_security_id(self, g_variant_get_string(val, NULL));
 	}
+	if (g_variant_dict_contains(dict, "OnlyTrusted")) {
+		g_autoptr(GVariant) val = NULL;
+		val = g_dbus_proxy_get_cached_property(proxy, "OnlyTrusted");
+		if (val != NULL) {
+			priv->only_trusted = g_variant_get_boolean(val);
+			fwupd_client_object_notify(self, "only-trusted");
+		}
+	}
 }
 
 static void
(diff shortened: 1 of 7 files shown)


libfwupd/fwupd-client.c
```diff
@@ -3280,6 +3294,24 @@ fwupd_client_get_tainted(FwupdClient *self)
 	return priv->tainted;
 }
 
+/**
+ * fwupd_client_get_only_trusted:
+ * @self: a #FwupdClient
+ *
+ * Gets if the daemon is verifying lvfs sign for firmware files.
```
superm1, line 3301: Verifying signatures from a trusted authority. LVFS isn't the only authority

libfwupd/fwupd-client.c
```diff
@@ -3280,6 +3294,24 @@ fwupd_client_get_tainted(FwupdClient *self)
...
+ *
+ * Returns: %TRUE if the daemon is checking lvfs sign
```
superm1, line 3303: Checking signatures.

src/org.freedesktop.fwupd.xml
```diff
@@ -111,6 +111,17 @@
       </doc:doc>
     </property>
 
+    <!--***********************************************************-->
+    <property name='OnlyTrusted' type='b' access='read'>
+      <doc:doc>
+        <doc:description>
+          <doc:para>
+            If the daemon is verifying lvfs signed firmware.
```
superm1, line 119: Daemon requires trusted payloads

superm1 (commented):
Overall looks good, just some nits.

superm1 (approved):
This lgtm. I think a logical use of this would be to show a warning on the client that the Daemon is requiring verifying signatures at the time too. @hughsie thoughts on that?

libfwupd/fwupd-client.c
```diff
@@ -421,6 +423,14 @@ fwupd_client_properties_changed_cb(GDBusProxy *proxy,
 		if (val != NULL)
 			fwupd_client_set_host_security_id(self, g_variant_get_string(val, NULL));
 	}
+	if (g_variant_dict_contains(dict, "OnlyTrusted")) {
+		g_autoptr(GVariant) val = NULL;
+		val = g_dbus_proxy_get_cached_property(proxy, "OnlyTrusted");
+		if (val != NULL) {
+			priv->only_trusted = g_variant_get_boolean(val);
+			fwupd_client_object_notify(self, "onlytrusted");
```
hughsie, line 431: this needs to be `only-trusted` -- I've fixed it up for you.

libfwupd/fwupd-client.c
```diff
@@ -5414,6 +5449,20 @@ fwupd_client_class_init(FwupdClientClass *klass)
...
 	g_object_class_install_property(object_class, PROP_HOST_SECURITY_ID, pspec);
+
+	/**
+	 * FwupdClient:only_trusted:
+	 *
+	 * If the daemon is verifying lvfs signed firmware.
+	 *
+	 * Since: 1.8.0
+	 */
+	pspec = g_param_spec_boolean("onlytrusted",
```
hughsie, line 5460: same, with a `-` -- fixed.

src/fu-engine.c
```diff
@@ -6274,6 +6275,13 @@ fu_engine_get_tainted(FuEngine *self)
 	return self->tainted;
 }
 
+gboolean
+fu_engine_get_only_trusted(FuEngine *self)
+{
+	g_return_val_if_fail(FU_IS_ENGINE(self), TRUE);
```
hughsie, line 6281: we always return FALSE or NULL from these, fixed for you.

hughsie:
> show a warning on the client that the daemon is not requiring verifying signatures

I think that's a really good idea, but we can follow this idea up in a new PR.

hughsie (approved):

PR #4372, opened 2022-03-09
fwupdplugin-self-test: fix spawn tests for qemu

djcampello:
Wrap the spawned script with an explicit call to the /bin/bash
interpreter needed when running tests in a cross compiled enviroment.

Also increase the timeout for the spawned script to execute to give
enough time for the print to happen in the much slower virtualized
environment.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -1017,7 +1017,7 @@ fu_common_spawn_func(void)
 	guint lines = 0;
 	g_autoptr(GError) error = NULL;
 	g_autofree gchar *fn = NULL;
-	const gchar *argv[3] = {"replace", "test", NULL};
+	const gchar *argv[4] = {"/bin/sh", "replace", "test", NULL};
 
 #ifdef _WIN32
 	g_test_skip("Known failures on Windows right now, skipping spawn func test");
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-self-test.c
```diff
@@ -1017,15 +1017,15 @@ fu_common_spawn_func(void)
 	guint lines = 0;
 	g_autoptr(GError) error = NULL;
 	g_autofree gchar *fn = NULL;
-	const gchar *argv[3] = {"replace", "test", NULL};
+	const gchar *argv[4] = {"/bin/bash", "replace", "test", NULL};
```
superm1, line 1020: Does /bin/sh work as well? If so I think that's preferable - it's a faster launch on a lot of machines.

djcampello: It does, done!

hughsie (approved):

superm1 (approved):

djcampello:
It seems that the increase in timeout was not sufficient :(

hughsie:
> It seems that the increase in timeout was not sufficient :(

An order of magnitude more perhaps? Or can we just disable that one test if we're running in qemu perhaps (if we can detect such a thing... cpuid perhaps?)

djcampello:
It seems that I was mistaken. This change was probably (I have not confirmed it yet) not needed and it was addressing a miss match between my local environment and the CrOS builders (I was missing the /bin/sh binary on my sysroot). The timeout increase was also probably only needed when I switched the interpreter to /bin/bash in a previous version of the change and not needed either once it went back to sh)

The real issue the builders complain about is that the test - fu_common_spawn_timeout_func - is leaking children processes and our scripts do not like that. We should address this by waiting for the spawned script or sending a kill signal before the test finishes.

PR #4379, opened 2022-03-14
Add coSWID and uSWID parsers to libfwupdplugin

hughsie:
These parse the structures as defined in:

 * https://datatracker.ietf.org/doc/draft-ietf-sacm-coswid/
 * https://github.com/hughsie/python-uswid

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


meson_options.txt
```diff
@@ -14,6 +14,7 @@ option('polkit', type: 'feature', description : 'PolKit support in daemon', depr
 option('gnutls', type: 'feature', description : 'GnuTLS support', deprecated: {'true': 'enabled', 'false': 'disabled'})
 option('sqlite', type: 'feature', description : 'sqlite support', deprecated: {'true': 'enabled', 'false': 'disabled'})
 option('lzma', type: 'feature', description : 'LZMA support', deprecated: {'true': 'enabled', 'false': 'disabled'})
+option('cbor', type: 'feature', description : 'CBOR support')
```
superm1, line 17: Maybe expand this description some.  I don't think CBOR is a generally well known acronym like LZMA is.

meson.build
```diff
@@ -278,6 +278,11 @@ if lzma.found()
   conf.set('HAVE_LZMA', '1')
 endif
 
+cbor = dependency('libcbor', required: get_option('cbor'))
```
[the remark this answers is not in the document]
hughsie: On my machines it's dragged in by `libfido2` which is in turn required by `openssh-clients` and so it seems installed almost everywhere.

meson_options.txt
hughsie: something like `CBOR (Concise Binary Object Representation) support` or something like `CBOR support (for coSWID and uSWID)`

meson.build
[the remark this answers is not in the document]
superm1: ah yeah I see it for same and also `glewlwyd`.  so does seem pretty openly available.

meson_options.txt
superm1: Yeah either of those

meson.build
superm1, line 281: what version are you starting with on your side?  0.8?  I'd rather anchor it somewhere if we can.

hughsie: 0.7.0 here, good idea.

libfwupdplugin/fu-coswid-firmware.c
```diff
@@ -0,0 +1,233 @@
...
+#define COSWID_GLOBAL_MAP_REVISION		    54
+#define COSWID_GLOBAL_MAP_SUMMARY		    55
+#define COSWID_GLOBAL_MAP_UNSPSC_CODE		    56
+#define COSWID_GLOBAL_MAP_UNSPSC_VERSION	    57
+
+#if HAVE_CBOR
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(cbor_item_t, cbor_intermediate_decref)
+#endif
+
+#if HAVE_CBOR
```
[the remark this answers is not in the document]
hughsie: Agree, fixed.

meson.build
hughsie: Fixed.

superm1 (approved):

hughsie:
Let's wait for the fuzzer image to re-gen and then I can check if it's going to explode in the real world: https://github.com/google/oss-fuzz/pull/7381

Issue #4382, opened 2022-03-15
Lenovo Gen2 Dock Firmware Update Fail


chlee75:
1. fwupd version: main line on March 15th 2020
2. Plugin: ccgx
3. Target Device: Lenovo ThinkPad USB-C Dock Gen2
4. Reproduction
    Execute sudo fwupd with -v 
    Execute fwupdmgr install --allow-reinstall Lenovo-USBC-Gen2_Dock-V0.0.34.cab
6. Issue 
   Firmware update failed with error message in fwupd below.
     01:31:11:0635 FuEngine             Emitting PropertyChanged('Status'='decompressing')
     01:31:11:0635 FuMain               Emitting PropertyChanged('Status'='decompressing')
     01:31:11:0636 FuCabinet            failed to verify cypress_fw2.metainfo.xml: no JcatItem
     01:31:11:0636 FuCabinet            processing file: cypress_fw2.metainfo.xml
     01:31:11:0637 FuCabinet            processing release: 0.0.34
     01:31:11:0637 FuEngine             Emitting PropertyChanged('Status'='idle')
     01:31:11:0637 FuMain               Emitting PropertyChanged('Status'='idle')

7. Comments 
    Firmware update succeeded in fwupd version 1.7.6
    Firmware failed by wrong instance GUID added in fwupd main line. 
    See fu_ccgx_hpi_device_setup().

PR #4384, opened 2022-03-15
Only delay the device removal when waiting for a replug

hughsie:
This stops the device showing up in 'fwupdmgr get-devices' with an
inhibit when removed normally, i.e. not during or after a fw update.

Fixes https://github.com/fwupd/fwupd/issues/4382

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
This has the potential to cause regressions with complicated devices like docks and needs some careful testing.

PR #4386, opened 2022-03-15
Do not show unconnected or unreachable devices in the client tools

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -2653,6 +2653,30 @@ fu_device_inhibit(FuDevice *self, const gchar *inhibit_id, const gchar *reason)
 }
 
 /**
+ * fu_device_has_inhibit:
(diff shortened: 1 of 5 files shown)


hardboprobot (commented):
I don't have a preference about this, but ooc, why do you want them eliminated from the device list instead of shown with a "not-updatable" flag?

hughsie:
I guess my logic is that it's confusing to see "unconnected" devices in the get-devices output. Maybe including unreachable is too much, and we should just do the former?

superm1 (commented):
I feel like the client should be filtering this if anything not the daemon.  What if a GUI client wanted to render a section of unreachable devices?

hardboprobot (approved):
Fair enough, LGTM

PR #4387, opened 2022-03-15
Fix build for MacOS

hughsie:
This is more for curiosity than anything else.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I've left out the CI (and left that PR open) but wanted to get these obvious bits in so they didn't bitrot.

PR #4390, opened 2022-03-15
Set the context in the right place more often

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
LGTM, and the criticals should certainly help catch fallout.  I like the idea of `#ifndef SUPPORTED_BUILD` for stuff like that.

PR #4393, opened 2022-03-16
Fix the 'Device has been removed' warning incorrectly seen on devices

hughsie:
Uninhibit the returning device if no firmware update has been performed.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


chlee75:
@hughsie 

It works fine. Update Error is not displayed after reconnect.

I thought before that device wait during RemoveDelay only when it has FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG flag.
RemoveDelay seems to be waiting time from replug.
Dell dock also set device remove delay when it need to replug.

How about your opinion?

PR #4399, opened 2022-03-17
vli:fixing vid/pids

memily:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Cleaning up and fixing vids/pids. Did not remove vli.quirk in first commit.


plugins/vli/vli-noinst.quirk
```diff
@@ -113,3 +107,58 @@ VliDeviceKind = VL106
...
+
+# VL650, VT3555
+[USB\VID_2109&PID_8885]
+Plugin = vli
+GType = FuVliPdDevice
+
+# MStar, VT3518
+[USB\VID_2109&PID_8886]
+Plugin = vli
+GType = FuVliPdDevice
```
hughsie, line 147: @memily are we sure none of these are on the LVFS?

memily: Yes, these are vids/pids created for designs that never made it to production --ergo that one vid/pid in the Lenovo file...

Issue #4404, opened 2022-03-18
Dependency check - need to check the version of primary BMC, not the Backup version


tcap26:
**Describe the bug**
There two version of BMC in the server: primary and backup. Primary version of BMC is the one that is working. So when fwupd check the firmware dependency based on "required=" declaration in metadata, it should compare with primary version of BMC, not the backup version.

**Steps to Reproduce**
Define firmware dependency in meta data:

  <requires>
    <id compare="ge" version="1.6.2">org.freedesktop.fwupd</id>
    <firmware compare="ge" version="1.50">[GUID of BMC]</firmware>
  </requires>

**Expected behavior**
Expected to compare with primary version of BMC, not the backup one.

**fwupd version information**
fwupdmgr --version
client version: 1.7.5

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
fwupdmgr get-devices --show-all-devices
```

**Additional questions**

- Operating system and version:
- Have you tried rebooting?
- Is this a regression?

PR #4405, opened 2022-03-18
redfish: Don't add the backup BMC device as it shares the same GUIDs

hughsie:
This fixes the problem when the UEFI update depends on a specific BMC
version -- including the backup BMC device means we checking that both
the primary and the backup were above a specific version.

I don't think it's ever useful to show the backup BMC device, so just
don't include it as an enumerated device.

Fixes https://github.com/fwupd/fwupd/issues/4404

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


tcap26:
I think it should be safe for now to ignore the backup version of BMC in normal use case

tcap26 (dismissed):
It's should be safe to ignore backup BMC for now.

PR #4411, opened 2022-03-18
[thunderbolt] disable retimer probing

krankun:
TEST=none

Change-Id: Idd59ab28079a769bdc986362dd4a43d716adf126

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-bluez-backend.c
+++ b/src/fu-bluez-backend.c
@@ -81,7 +81,6 @@ static void
 fu_bluez_backend_object_added(FuBluezBackend *self, GDBusObject *object)
 {
 	g_autoptr(GDBusInterface) iface = NULL;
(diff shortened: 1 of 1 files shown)


djcampello (commented):
I reviewed this on crrev.com/c/3533215 and suggested to use a configuration entry instead. Having these `return TRUE` before blocks of code is bad practice.

See https://github.com/fwupd/fwupd/pull/4412 for alternative approach

krankun:
> I reviewed this on crrev.com/c/3533215 and suggested to use a configuration entry instead. Having these `return TRUE` before blocks of code is bad practice.
> 
> See #4412 for alternative approach

Flag related approach looks good I was also thinking the same ! will follow on there for disabling the probe!
closing this PR

PR #4413, opened 2022-03-18
ccgx: Add alternate protocol names

hughsie:
Cypress was acquired by Infineon.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/ccgx/README.md
+++ b/plugins/ccgx/README.md
@@ -11,6 +11,8 @@ This plugin supports the following protocol IDs:
 
 * com.cypress.ccgx
 * com.cypress.ccgx.dmc
+* com.infineon.ccgx
+* com.infineon.ccgx.dmc
 
 ## Device Flash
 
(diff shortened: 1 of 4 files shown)


superm1 (commented):
How do you plan to handle this on LVFS though? It will break old versions to change any fw protocol and using the new name will mean only new sw can use it for artificial reasons.

It's almost like you need a deprecated meaning for protocol and use both until a long time from now.

hughsie:
> How do you plan to handle this on LVFS though

I was worried about this too; but it seems the daemon doesn't actually check the protocol matches -- so we can just rename it on the LVFS and push new firmware. I actually made a branch that essentially did `s/fwupd_release_get_protocol/fwupd_release_get_protocols` in a backwards compat way, but there was nothing in the daemon that actually verified that the device protocol matched the release protocol. Maybe that's a valid bug?

I wasn't also sure whether it made sense for a release to *have* multiple protocols -- thinking bigger it seems a release should only be targetted at one, but the device can support multiple. I think we probably ought to have an LVFS check on the new fwupd version for renamed protocol IDs if we start doing the check.

hughsie:
> Maybe that's a valid bug?

e.g. https://github.com/fwupd/fwupd/pull/4415

chlee75:
@hughsie 

Uploading cab file that has com.infineon.ccgx.dmc is succeful.

But I have error in LVFS when I update cab file that has com.cypress.ccgx.dmc

2ac61e6ebd9c08ad367729e64a3b2384801b56bbeadb1ae30e27c350510ddef7-HP-updatemsg-USBC_DOCK_G5-V1.0.8.0.cab
File is not supported: Metadata is invalid: No valid UpdateProtocol com.cypress.ccgx.dmc found 

Can you make support both com.cypress.ccgx.dmc and com.infineon.ccgx.dmc?

hughsie:
> Can you make support both com.cypress.ccgx.dmc and com.infineon.ccgx.dmc?

Okay, I've added both to unstick things.

PR #4415, opened 2022-03-19
Check the update protocol exists when checking requirements

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
This will presumably make migrations harder in the future. How about making it a warning instead?

superm1 (approved):

PR #4416, opened 2022-03-20
genesys: scaler: Get real software version

gportay:
The plugin gets the panel-type (i.e. hardware revision) as the firmware
version.

This gets the real firmware version, and appends the panel-type to the
instance-id string as the blobs are specific to panel-types.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


gportay:
@adamgene, could you confirm and test this is correct for the HP monitors?

plugins/genesys/fu-genesys-scaler-device.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Gaël PORTAY <gael.portay@collabora.com>
+ * Copyright (C) 2021-2022 Gaël PORTAY <gael.portay@collabora.com>
```
hughsie, line 2: in general we only list the start year in source files

plugins/genesys/fu-genesys-scaler-device.c
```diff
@@ -1417,17 +1551,28 @@ fu_genesys_scaler_device_probe(FuDevice *device, GError **error)
...
 	fu_device_set_version(device, version);
 	fu_device_set_version_format(device, FWUPD_VERSION_FORMAT_PLAIN);
 	fu_device_set_logical_id(device, "scaler");
 
 	/* add instance ID */
 	fu_device_add_instance_str(device, "MSTAR", "TSUM_G");
 	fu_device_add_instance_strup(device, "PUBKEY", guid);
-	fu_device_build_instance_id(device, NULL, "GENESYS_SCALER", "MSTAR", "PUBKEY", NULL);
+	fu_device_add_instance_strup(device, "PANELTYPE", panel_type);
+	fu_device_build_instance_id(device,
```
hughsie, line 1569: I think we probably want *both* instance IDs, no?

plugins/genesys/fu-genesys-scaler-device.c
```diff
@@ -1417,17 +1551,28 @@ fu_genesys_scaler_device_probe(FuDevice *device, GError **error)
...
+	version = g_strdup_printf("%d.%d.%d.%d", ver.stage, ver.model, ver.major, ver.minor);
 	fu_device_set_version(device, version);
 	fu_device_set_version_format(device, FWUPD_VERSION_FORMAT_PLAIN);
 	fu_device_set_logical_id(device, "scaler");
 
 	/* add instance ID */
 	fu_device_add_instance_str(device, "MSTAR", "TSUM_G");
 	fu_device_add_instance_strup(device, "PUBKEY", guid);
-	fu_device_build_instance_id(device, NULL, "GENESYS_SCALER", "MSTAR", "PUBKEY", NULL);
+	fu_device_add_instance_strup(device, "PANELTYPE", panel_type);
```
hughsie, line 1568: maybe `PANEL`?

plugins/genesys/fu-genesys-scaler-device.c
gportay: Indeed, it makes sens to have both.

plugins/genesys/fu-genesys-scaler-device.c
gportay: done.

plugins/genesys/fu-genesys-scaler-device.c
gportay: This field refers to the attribute of `panel_type` in structure `FuGenesysMtkRsaHeader`; I like they have the same name if that possible. However, I like `PANEL`.

I would prefer `REV` or `HWREV`; but it is not clear to me if the that attribute always refers to a panel or not :/.

@adamgene, any opinion?

adamgene: >I would prefer `REV` or `HWREV`; but it is not clear to me if the that attribute always refers to a panel or not :/.
>

I researched the document for the `FuGenesysMtkRsaHeader`, it says `HW FW version`. But I found the string in HP definition is `REV`. The detail format can ask HP.

IMO, It is means panel revision for MSTAR scaler chip.

hughsie: REV won't work as we use that for the USB revision already.

hughsie (dismissed):

hughsie:
@gportay ohh i see the CI failure now:

      1501 |                                     "error dddci length too large, got 0x%lx, expected <= 0x%zx

just use 0x%x and then cast the parameter to (guint) manually.

plugins/genesys/fu-genesys-scaler-device.c
adamgene: How about PANEL_REV?

hughsie: I like PANELREV more, but that works for me.

adamgene: `PANELREV` is works for me too.

gportay: so let's go for `PANELREV`.

gportay:
> just use 0x%x and then cast the parameter to (guint) manually.

done

Issue #4418, opened 2022-03-21
Don't strip event payloads from log


flanfly:
Hi.

**Is your feature request related to a problem? Please describe.**
While fwupd does export the TPM 2.0 event log, it strips most of the event's payload. While this still allows for verifying the log against the PCRs, we don't know what has been logged.

**Describe the solution you'd like**
Instead of replacing all non-printable characters in the (likely binary) payload with periods why not base64 encode it?

**Describe alternatives you've considered**
None

**Additional context**
https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf

PR #4419, opened 2022-03-22
Do not throw away the TPM eventlog when uploading to the LVFS

hughsie:
To debug the eventlog reconstruction issues we need the raw blob,
rather than just the printable text.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4418

superm1 (approved):

PR #4424, opened 2022-03-23
genesys: trivial rework some usbhub errors

gportay:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

The idea behind those changes is to prepare the genesys plugin for upgrading the genuine Genesys Logic Hubs products.

These products uses `USB\VID_05E3&PID_0610` which is not set in the quirk file yet; will be in a dedicated PR, once the support is fully functional.

The support for genuine Genesys Logic Hubs is splited into several PR. This one is the second of the serie.

In short, this PR:
 - Print the Genesys IC if it is not yet supported by the plugin.
 - Fails silently if the IC is not yet supported.

### DIFF
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -747,10 +747,14 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		} else if (memcmp(self->static_ts.mask_project_ic_type, "3590", 4) == 0) {
 			self->isp_model = ISP_MODEL_HUB_GL3590;
 		} else {
-			g_set_error_literal(error,
-					    FWUPD_ERROR,
-					    FWUPD_ERROR_INTERNAL,
-					    "Unknown ISP model");
+			g_autofree gchar *ic_type =
+			    fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
+					      sizeof(self->static_ts.mask_project_ic_type));
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "IC type %s not supported",
+				    ic_type);
 			return FALSE;
 		}
 		memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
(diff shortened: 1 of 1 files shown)


gportay:
Tested with a hub found in my stuff that is based on the `GL33521-22`.

gportay:
@adamgene, do you have any thing else to add?

adamgene (changes requested):
```
	if (memcmp(&self->static_ts.tool_string_version, "USB", 3) == 0) {
		g_set_error(error,
			    FWUPD_ERROR,
			    FWUPD_ERROR_INTERNAL,
			    "Unsupported IC: tool string unsupported");
		return FALSE;
	}
```
Can you add these code @741? I found out old GL3520 unsupported tool string.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -746,10 +748,19 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
-					    FWUPD_ERROR,
-					    FWUPD_ERROR_INTERNAL,
-					    "Unknown ISP model");
+			g_autofree gchar *ic_type =
+			    fu_common_strsafe((const gchar *)self->static_ts.mask_project_ic_type,
+					      6);
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "Unsupported IC GL%c%c%c%c-%c%c",
```
hughsie, line 757: I don't think you know that strlen(ic_type) >= 6 -- I think you either want to print this as %s (probably fine given it's in the error path) or make a `convert_mask_project_ic_type_to_string` function that adds the dash in the right place.

gportay: make sense.
done

gportay:
> ```
> 	if (memcmp(&self->static_ts.tool_string_version, "USB", 3) == 0) {
> 		g_set_error(error,
> 			    FWUPD_ERROR,
> 			    FWUPD_ERROR_INTERNAL,
> 			    "Unsupported IC: tool string unsupported");
> 		return FALSE;
> 	}
> ```
> 
> Can you add these code @741? I found out old GL3520 unsupported tool string.

I will do it in another PR, as it does not appear to be directly linked to that one.

gportay:
@hughsie How can I re-trigger the CI? I did not understand what happened.

Ready to get merged otherwise.

hughsie (approved):

PR #4426, opened 2022-03-23
genesys: usbhub: support for gl3590 and other generic case with customized tool string info

adamgene:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation
---
- There are many bonding types of chip, and a firmware has built for customize request with specific chip. Therefore, we have to create a unique instance-id to avoid updating wrong firmware.
- There is a `bonding` byte in `FuGenesysDynamicToolString`, the byte also indicate current running bank.
- There has a unique public key for HP products; other products do not have a unique key to indicate products.
  - When running mask code bank, the characteristics only
    1.  **chip**, the value in `FuGenesysStaticToolString`, e.g. "352310", "359010".
    2.  **bonding**, the value in `FuGenesysDynamicToolString`, each bonding already defined supporting DFP number.
  - When running bank 1st/2nd, the firmware already customized, obviously characteristics are
    1. `VID`  / `PID` / `VENDOR` from USB descriptor. 'PRODUCT' already replaced by 'NAME', so it cannot be reference I guess.
    2.  **port number** for HS and SS, the values in `FuGenesysDynamicToolString`. Firmware can turn off hub's DFPs, so two products have the same bonding of chip may enable different number of DFPs.
    3.  **vendor supporting** for special function, the value in `FuGenesysVendorSupportToolString`.
    4. Other characteristics are in the firmware configuration, but they may changed due to requests or bug fix.

### DIFF
--- a/plugins/genesys/fu-genesys-common.h
+++ b/plugins/genesys/fu-genesys-common.h
@@ -32,7 +32,7 @@ typedef struct {
 } FuGenesysChip;
 
 typedef struct __attribute__((packed)) {
-	guint8 tool_string_version; /* 0xff = not supported */
+	guint8 tool_string_version;
 
 	/* byte arrays are ASCII encoded and not NUL terminated */
 	guint8 mask_project_code[4];
--- a/plugins/genesys/fu-genesys-common.h
+++ b/plugins/genesys/fu-genesys-common.h
@@ -11,6 +11,26 @@ typedef struct {
 	guint8 expected_val;
 } FuGenesysWaitFlashRegisterHelper;
 
+typedef enum {
+	ISP_MODEL_UNKNOWN,
+
+	/* hub */
+	ISP_MODEL_HUB_GL3510,
+	ISP_MODEL_HUB_GL3521,
+	ISP_MODEL_HUB_GL3523,
+	ISP_MODEL_HUB_GL3590,
+	ISP_MODEL_HUB_GL7000,
+	ISP_MODEL_HUB_GL3525,
+
+	/* pd */
+	ISP_MODEL_PD_GL9510,
+} FuGenesysModel;
+
+typedef struct {
+	FuGenesysModel model;
+	gint32 revision;
+} FuGenesysChip;
+
 typedef struct __attribute__((packed)) {
 	guint8 tool_string_version; /* 0xff = not supported */
 
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -534,6 +534,15 @@ fu_genesys_usbhub_device_get_descriptor_data(GBytes *desc_bytes,
 	for (gsize i = 0, j = 0; i < bufsz && j < dst_size; i += 2, j++)
 		dst[j] = buf[i];
 
+	/* legacy hub replies "USB2.0 Hub" or "USB3.0 Hub" */
+	if (memcmp(dst, "USB", 3) == 0) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "tool string unsupported");
+		return FALSE;
+	}
+
 	/* success */
 	return TRUE;
 }
@@ -597,6 +606,16 @@ fu_genesys_usbhub_device_get_fw_size(FuGenesysUsbhubDevice *self, int bank_num, 
 
 	/* success */
 	return TRUE;
+}
+
+static gint
+fu_genesys_tsdigit_value(gchar c)
+{
+	if (c >= 'A' && c <= 'Z')
+		return c - 'A' + 10;
+	if (c >= 'a' && c <= 'z')
+		return c - 'a' + 10;
+	return g_ascii_digit_value(c);
 }
 #endif
 
@@ -678,6 +697,8 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 	guint16 version_raw;
 	guint8 static_idx = 0;
 	guint8 dynamic_idx = 0;
+	gchar rev[3] = {0};
+	gint tool_string_version = 0;
 	g_autoptr(FuFirmware) firmware = NULL;
 	g_autoptr(GBytes) static_buf = NULL;
 	g_autoptr(GBytes) dynamic_buf = NULL;
@@ -732,27 +753,25 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				   (guint8 *)&self->static_ts,
 				   sizeof(FuGenesysStaticToolString));
 	}
-	if (self->static_ts.tool_string_version != 0xff) {
-		gchar rev[3] = {0};
 
-		if (memcmp(self->static_ts.mask_project_ic_type, "3523", 4) == 0) {
-			self->chip.model = ISP_MODEL_HUB_GL3523;
-		} else if (memcmp(self->static_ts.mask_project_ic_type, "3590", 4) == 0) {
-			self->chip.model = ISP_MODEL_HUB_GL3590;
-		} else {
-			g_autofree gchar *ic_type =
-			    fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
-					      sizeof(self->static_ts.mask_project_ic_type));
-			g_set_error(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "IC type %s not supported",
-				    ic_type);
-			return FALSE;
-		}
-		memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
-		self->chip.revision = fu_common_strtoull(rev);
+	if (memcmp(self->static_ts.mask_project_ic_type, "3523", 4) == 0) {
+		self->chip.model = ISP_MODEL_HUB_GL3523;
+	} else if (memcmp(self->static_ts.mask_project_ic_type, "3590", 4) == 0) {
+		self->chip.model = ISP_MODEL_HUB_GL3590;
+	} else {
+		g_autofree gchar *ic_type =
+		    fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
+				      sizeof(self->static_ts.mask_project_ic_type));
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "IC type %s not supported",
+			    ic_type);
+		return FALSE;
 	}
+	memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
+	self->chip.revision = fu_common_strtoull(rev);
+	tool_string_version = fu_genesys_tsdigit_value(self->static_ts.tool_string_version);
 
 	dynamic_buf =
 	    g_usb_device_get_string_descriptor_bytes_full(usb_device,
@@ -789,7 +808,7 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				   sizeof(FuGenesysFirmwareInfoToolString));
 	}
 
-	if (self->static_ts.tool_string_version >= TOOL_STRING_VERSION_VENDOR_SUPPORT) {
+	if (tool_string_version >= TOOL_STRING_VERSION_VENDOR_SUPPORT) {
 		g_autoptr(GBytes) vendor_buf = g_usb_device_get_string_descriptor_bytes_full(
 		    usb_device,
 		    GENESYS_USBHUB_VENDOR_SUPPORT_DESC_IDX,
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -132,6 +132,11 @@ typedef struct __attribute__((packed)) {
 	guint8 update_fw_time[12]; /* YYYYMMDDhhmm */
 } FuGenesysFirmwareInfoToolString;
 
+typedef struct __attribute__((packed)) {
+	guint8 version[2];
+	guint8 supports[29];
+} FuGenesysVendorSupportToolString;
+
 typedef struct {
 	guint8 req_switch;
 	guint8 req_read;
@@ -143,6 +148,7 @@ struct _FuGenesysUsbhubDevice {
 	FuGenesysStaticToolString static_ts;
 	FuGenesysDynamicToolString dynamic_ts;
 	FuGenesysFirmwareInfoToolString fwinfo_ts;
+	FuGenesysVendorSupportToolString vs_ts;
 	FuGenesysVendorCommandSetting vcs;
 	FuGenesysChip chip;
 	guint32 running_bank;
@@ -763,6 +769,7 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 	g_autoptr(GError) error_local = NULL;
 	g_autoptr(GBytes) blob = NULL;
 	g_autofree guint8 *buf = NULL;
+	g_autofree gchar *ic_type = NULL;
 
 	/* FuUsbDevice->setup */
 	if (!FU_DEVICE_CLASS(fu_genesys_usbhub_device_parent_class)->setup(device, error)) {
@@ -816,9 +823,8 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 	} else if (memcmp(self->static_ts.mask_project_ic_type, "3590", 4) == 0) {
 		self->chip.model = ISP_MODEL_HUB_GL3590;
 	} else {
-		g_autofree gchar *ic_type =
-		    fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
-				      sizeof(self->static_ts.mask_project_ic_type));
+		ic_type = fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
+					    sizeof(self->static_ts.mask_project_ic_type));
 		g_set_error(error,
 			    FWUPD_ERROR,
 			    FWUPD_ERROR_NOT_SUPPORTED,
@@ -888,6 +894,20 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		if (vendor_buf == NULL) {
 			g_prefix_error(error, "failed to get vendor support info from device: ");
 			return FALSE;
+		}
+		if (!fu_genesys_usbhub_device_get_descriptor_data(
+			vendor_buf,
+			(guint8 *)&self->vs_ts,
+			sizeof(FuGenesysVendorSupportToolString),
+			error)) {
+			g_prefix_error(error, "failed to get vendor support info from device: ");
+			return FALSE;
+		}
+		if (g_getenv("FWUPD_GENESYS_USBHUB_VERBOSE") != NULL) {
+			fu_common_dump_raw(G_LOG_DOMAIN,
+					   "Vendor support",
+					   (guint8 *)&self->vs_ts,
+					   sizeof(FuGenesysVendorSupportToolString));
 		}
 	}
 
@@ -1056,6 +1076,37 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		fu_device_add_instance_strup(device, "PUBKEY", guid);
 	}
 
+	/* add specific product info */
+	ic_type = fu_common_strsafe((const gchar *)self->static_ts.mask_project_ic_type,
+				    sizeof(self->static_ts.mask_project_ic_type));
+	fu_device_add_instance_str(device, "IC", ic_type);
+	fu_device_add_instance_u8(device, "BONDING", self->bonding);
+
+	if (self->running_bank != BANK_MASK_CODE) {
+		const gchar *vendor = fwupd_device_get_vendor(FWUPD_DEVICE(device));
+		guint16 port_num =
+		    (self->dynamic_ts.ss_port_number << 8) + self->dynamic_ts.hs_port_number;
+		g_autofree gchar *guid = NULL;
+		guid = fwupd_guid_hash_data((const guint8 *)&self->vs_ts,
+					    sizeof(self->vs_ts),
+					    FWUPD_GUID_FLAG_NONE);
+		fu_device_add_instance_strup(device, "VENDOR", vendor);
+		fu_device_add_instance_u16(device, "PORTNUM", port_num);
+		fu_device_add_instance_strup(device, "VENDORSUP", guid);
+	}
+
+	fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "IC", "BONDING", NULL);
+	fu_device_build_instance_id(device,
+				    NULL,
+				    "USB",
+				    "VID",
+				    "PID",
+				    "VENDOR",
+				    "IC",
+				    "BONDING",
+				    "PORTNUM",
+				    "VENDORSUP",
+				    NULL);
 	fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "PUBKEY", NULL);
 
 	/* have MStar scaler */
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -66,6 +66,60 @@ typedef enum {
 	TOOL_STRING_VERSION_13BYTE_DYNAMIC,
 } FuGenesysToolStringVersion;
 
+typedef struct __attribute__((packed)) {
+	guint8 running_mode; /* 'M' for mask code, the others for bank code */
+
+	guint8 ss_port_number; /* super-speed port number */
+	guint8 hs_port_number; /* high-speed port number */
+
+	guint8 ss_connection_status; /* bit field. ON = DFP is a super-speed device */
+	guint8 hs_connection_status; /* bit field. ON = DFP is a high-speed device */
+	guint8 fs_connection_status; /* bit field. ON = DFP is a full-speed device */
+	guint8 ls_connection_status; /* bit field. ON = DFP is a low-speed device */
+
+	guint8 charging;		  /* bit field. ON = DFP is a charging port */
+	guint8 non_removable_port_status; /* bit field. ON = DFP is a non-removable port */
+
+	/*
+	 * Bonding reports Hardware register status for GL3523:
+	 *   2 / 4 ports         : 1 means 4 ports, 0 means 2 ports
+	 *   MTT / STT           : 1 means Multi Token Transfer, 0 means Single TT
+	 *   Type - C            : 1 means disable, 0 means enable
+	 *   QC                  : 1 means disable, 0 means enable
+	 *   Flash dump location : 1 means 32KB offset bank 1, 0 means 0 offset bank 0.
+	 *
+	 * Tool string Version 1:
+	 *   Bit3 : Flash dump location
+	 *   BIT2 : Type - C
+	 *   BIT1 : MTT / STT
+	 *   BIT0 : 2 / 4 ports
+	 *
+	 * Tool string Version 2 or newer :
+	 *   Bit4 : Flash dump location
+	 *   BIT3 : Type - C
+	 *   BIT2 : MTT / STT
+	 *   BIT1 : 2 / 4 ports
+	 *   BIT0 : QC
+	 *
+	 * Default use '0'~'F', plus Bit4 may over value, should extract that.
+	 *
+	 * Bonding for GL3590:
+	 *   Bit7 : Flash dump location, 0 means bank 0, 1 means bank 1.
+	 */
+	guint8 bonding;
+
+	guint8 reserved[22];
+} FuGenesysDynamicToolString;
+
+typedef enum {
+	BANK_MASK_CODE,
+	BANK_FIRST,
+	BANK_SECOND,
+} FuGenesysRunningBank;
+
+#define GL3523_BONDING_VALID_BIT 0x0F
+#define GL3590_BONDING_VALID_BIT 0x7F
+
 typedef enum {
 	ISP_EXIT,
 	ISP_ENTER,
@@ -87,9 +141,12 @@ typedef struct {
 struct _FuGenesysUsbhubDevice {
 	FuUsbDevice parent_instance;
 	FuGenesysStaticToolString static_ts;
+	FuGenesysDynamicToolString dynamic_ts;
 	FuGenesysFirmwareInfoToolString fwinfo_ts;
 	FuGenesysVendorCommandSetting vcs;
 	FuGenesysChip chip;
+	guint32 running_bank;
+	guint8 bonding;
 	guint32 flash_erase_delay;
 	guint32 flash_write_delay;
 	guint32 flash_block_size;
@@ -783,6 +840,19 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		g_prefix_error(error, "failed to get dynamic tool info from device: ");
 		return FALSE;
 	}
+	if (!fu_genesys_usbhub_device_get_descriptor_data(dynamic_buf,
+							  (guint8 *)&self->dynamic_ts,
+							  sizeof(FuGenesysDynamicToolString),
+							  error)) {
+		g_prefix_error(error, "failed to get dynamic tool info from device: ");
+		return FALSE;
+	}
+	if (g_getenv("FWUPD_GENESYS_USBHUB_VERBOSE") != NULL) {
+		fu_common_dump_raw(G_LOG_DOMAIN,
+				   "Dynamic info",
+				   (guint8 *)&self->dynamic_ts,
+				   sizeof(FuGenesysDynamicToolString));
+	}
 
 	fw_buf =
 	    g_usb_device_get_string_descriptor_bytes_full(usb_device,
@@ -827,6 +897,7 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 	}
 	if (!fu_genesys_usbhub_device_set_isp_mode(self, ISP_ENTER, error))
 		return FALSE;
+	/* setup cfi device */
 	self->cfi_device = fu_genesys_usbhub_device_cfi_setup(self, error);
 	if (self->cfi_device == NULL)
 		return FALSE;
@@ -839,10 +910,12 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 
 	/* setup firmware parameters */
 	switch (self->chip.model) {
-	case ISP_MODEL_HUB_GL3523:
+	case ISP_MODEL_HUB_GL3523: {
+		gint bonding = 0;
 		self->fw_bank_addr[0] = 0x0000;
 		self->fw_bank_addr[1] = 0x8000;
 		self->extend_size = GL3523_PUBLIC_KEY_LEN + GL3523_SIG_LEN;
+
 		if (self->chip.revision == 50) {
 			self->fw_data_total_count = 0x8000;
 			if (!fu_genesys_usbhub_device_get_fw_size(self, 0, error))
@@ -851,13 +924,44 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 			self->fw_data_total_count = 0x6000;
 			self->code_size = self->fw_data_total_count;
 		}
+
+		bonding = fu_genesys_tsdigit_value((gchar)self->dynamic_ts.bonding);
+		if (bonding == -1) {
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "GL3523 bonding value(0x%02x) is wrong",
+				    self->dynamic_ts.bonding);
+			return FALSE;
+		}
+		if (self->static_ts.tool_string_version < TOOL_STRING_VERSION_BONDING_QC)
+			bonding <<= 1;
+		self->bonding = (bonding & GL3523_BONDING_VALID_BIT);
+
+		if (self->dynamic_ts.running_mode == 'M') {
+			self->running_bank = BANK_MASK_CODE;
+		} else if ((bonding & 0x10) > 0) {
+			self->running_bank = BANK_SECOND;
+		} else {
+			self->running_bank = BANK_FIRST;
+		}
 		break;
+	}
 	case ISP_MODEL_HUB_GL3590:
 		if (!fu_genesys_usbhub_device_get_fw_size(self, 0, error))
 			return FALSE;
 		self->fw_bank_addr[0] = 0x0000;
 		self->fw_bank_addr[1] = 0x10000;
 		self->fw_data_total_count = 0x10000;
+		self->bonding = (self->dynamic_ts.bonding & GL3590_BONDING_VALID_BIT);
+
+		if (self->dynamic_ts.running_mode == 'M') {
+			self->running_bank = BANK_MASK_CODE;
+		} else if ((self->dynamic_ts.bonding & 0x80) > 0) {
+			self->running_bank = BANK_SECOND;
+		} else {
+			self->running_bank = BANK_FIRST;
+		}
 		break;
 	default:
 		break;
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -71,21 +71,6 @@ typedef enum {
 	ISP_ENTER,
 } FuGenesysIspMode;
 
-typedef enum {
-	ISP_MODEL_UNKNOWN,
-
-	/* hub */
-	ISP_MODEL_HUB_GL3510,
-	ISP_MODEL_HUB_GL3521,
-	ISP_MODEL_HUB_GL3523,
-	ISP_MODEL_HUB_GL3590,
-	ISP_MODEL_HUB_GL7000,
-	ISP_MODEL_HUB_GL3525,
-
-	/* pd */
-	ISP_MODEL_PD_GL9510,
-} FuGenesysModel;
-
 typedef struct __attribute__((packed)) {
 	guint8 tool_version[6]; /* ISP tool defined by itself */
 	guint8 address_mode;
@@ -104,8 +89,7 @@ struct _FuGenesysUsbhubDevice {
 	FuGenesysStaticToolString static_ts;
 	FuGenesysFirmwareInfoToolString fwinfo_ts;
 	FuGenesysVendorCommandSetting vcs;
-	guint32 isp_model;
-	guint8 isp_revision;
+	FuGenesysChip chip;
 	guint32 flash_erase_delay;
 	guint32 flash_write_delay;
 	guint32 flash_block_size;
@@ -752,9 +736,9 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		gchar rev[3] = {0};
 
 		if (memcmp(self->static_ts.mask_project_ic_type, "3523", 4) == 0) {
-			self->isp_model = ISP_MODEL_HUB_GL3523;
+			self->chip.model = ISP_MODEL_HUB_GL3523;
 		} else if (memcmp(self->static_ts.mask_project_ic_type, "3590", 4) == 0) {
-			self->isp_model = ISP_MODEL_HUB_GL3590;
+			self->chip.model = ISP_MODEL_HUB_GL3590;
 		} else {
 			g_autofree gchar *ic_type =
 			    fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
@@ -767,7 +751,7 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 			return FALSE;
 		}
 		memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
-		self->isp_revision = fu_common_strtoull(rev);
+		self->chip.revision = fu_common_strtoull(rev);
 	}
 
 	dynamic_buf =
@@ -835,12 +819,12 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		self->flash_sector_size = sector_size;
 
 	/* setup firmware parameters */
-	switch (self->isp_model) {
+	switch (self->chip.model) {
 	case ISP_MODEL_HUB_GL3523:
 		self->fw_bank_addr[0] = 0x0000;
 		self->fw_bank_addr[1] = 0x8000;
 		self->extend_size = GL3523_PUBLIC_KEY_LEN + GL3523_SIG_LEN;
-		if (self->isp_revision == 50) {
+		if (self->chip.revision == 50) {
 			self->fw_data_total_count = 0x8000;
 			if (!fu_genesys_usbhub_device_get_fw_size(self, 0, error))
 				return FALSE;
@@ -927,7 +911,7 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		}
 
 		self->read_first_bank =
-		    (self->isp_model == ISP_MODEL_HUB_GL3523) && self->fw_bank_vers[0] != 0;
+		    (self->chip.model == ISP_MODEL_HUB_GL3523) && self->fw_bank_vers[0] != 0;
 		self->write_recovery_bank = address == self->fw_bank_addr[1];
 	}
 
--- a/plugins/genesys/fu-genesys-usbhub-firmware.c
+++ b/plugins/genesys/fu-genesys-usbhub-firmware.c
@@ -166,15 +166,6 @@ fu_genesys_usbhub_firmware_parse(FuFirmware *firmware,
 			    error))
 		return FALSE;
 
-	/* unsupported static tool string */
-	if (self->static_ts.tool_string_version == 0xff) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "Static Tool String not supported");
-		return FALSE;
-	}
-
 	/* deduce code size */
 	switch (self->chip.model) {
 	case ISP_MODEL_HUB_GL3523: {
--- a/plugins/genesys/fu-genesys-usbhub-firmware.c
+++ b/plugins/genesys/fu-genesys-usbhub-firmware.c
@@ -14,9 +14,54 @@
 struct _FuGenesysUsbhubFirmware {
 	FuFirmwareClass parent_instance;
 	FuGenesysStaticToolString static_ts;
+	FuGenesysChip chip;
 };
 
 G_DEFINE_TYPE(FuGenesysUsbhubFirmware, fu_genesys_usbhub_firmware, FU_TYPE_FIRMWARE)
+
+static gboolean
+fu_genesys_usbhub_firmware_get_chip(FuGenesysUsbhubFirmware *self,
+				    const guint8 *buf,
+				    gsize bufsz,
+				    GError **error)
+{
+	guint8 project_ic_type[6];
+
+	if (!fu_memcpy_safe(project_ic_type,
+			    sizeof(project_ic_type),
+			    0, /* dst */
+			    buf,
+			    bufsz,
+			    GENESYS_USBHUB_STATIC_TOOL_STRING_OFFSET_GL3523 + 8, /* src */
+			    sizeof(project_ic_type),
+			    error))
+		return FALSE;
+
+	if (memcmp(project_ic_type, "3523", 4) == 0) {
+		self->chip.model = ISP_MODEL_HUB_GL3523;
+		self->chip.revision = 10 * (project_ic_type[4] - '0') + (project_ic_type[5] - '0');
+		return TRUE;
+	}
+
+	if (!fu_memcpy_safe(project_ic_type,
+			    sizeof(project_ic_type),
+			    0, /* dst */
+			    buf,
+			    bufsz,
+			    GENESYS_USBHUB_STATIC_TOOL_STRING_OFFSET_GL3590 + 8, /* src */
+			    sizeof(project_ic_type),
+			    error))
+		return FALSE;
+
+	if (memcmp(project_ic_type, "3590", 4) == 0) {
+		self->chip.model = ISP_MODEL_HUB_GL3590;
+		self->chip.revision = 10 * (project_ic_type[4] - '0') + (project_ic_type[5] - '0');
+		return TRUE;
+	}
+
+	g_set_error_literal(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED, "Unsupported IC");
+	return FALSE;
+}
 
 static gboolean
 fu_genesys_usbhub_firmware_verify(const guint8 *buf, gsize bufsz, guint16 code_size, GError **error)
@@ -69,9 +114,10 @@ fu_genesys_usbhub_firmware_parse(FuFirmware *firmware,
 	gsize bufsz = 0;
 	const guint8 *buf = g_bytes_get_data(fw, &bufsz);
 	guint8 sign[4];
-	guint32 code_size = 0x6000;
+	guint32 code_size = 0;
+	guint32 static_ts_offset = 0;
 	guint16 version_raw = 0;
-	gboolean is3590 = FALSE;
+	g_autoptr(GBytes) fw_payload = NULL;
 	g_autofree gchar *version = NULL;
 
 	/* check signature */
@@ -94,44 +140,31 @@ fu_genesys_usbhub_firmware_parse(FuFirmware *firmware,
 		return FALSE;
 	}
 
+	/* get chip */
+	if (!fu_genesys_usbhub_firmware_get_chip(self, buf, bufsz, error))
+		return FALSE;
+
+	switch (self->chip.model) {
+	case ISP_MODEL_HUB_GL3523:
+		static_ts_offset = GENESYS_USBHUB_STATIC_TOOL_STRING_OFFSET_GL3523;
+		break;
+	case ISP_MODEL_HUB_GL3590:
+		static_ts_offset = GENESYS_USBHUB_STATIC_TOOL_STRING_OFFSET_GL3590;
+		break;
+	default:
+		break;
+	}
+
 	/* get static tool string */
 	if (!fu_memcpy_safe((guint8 *)&self->static_ts,
 			    sizeof(self->static_ts),
 			    0, /* dst */
 			    buf,
 			    bufsz,
-			    GENESYS_USBHUB_STATIC_TOOL_STRING_OFFSET_GL3523, /* src */
+			    static_ts_offset, /* src */
 			    sizeof(self->static_ts),
 			    error))
 		return FALSE;
-
-	/* not a GL3523, is GL3590? */
-	if (memcmp(self->static_ts.mask_project_ic_type, "3523", 4) != 0) {
-		if (!fu_memcpy_safe((guint8 *)&self->static_ts,
-				    sizeof(self->static_ts),
-				    0, /* dst */
-				    buf,
-				    bufsz,
-				    GENESYS_USBHUB_STATIC_TOOL_STRING_OFFSET_GL3590, /* src */
-				    sizeof(self->static_ts),
-				    error))
-			return FALSE;
-
-		/* not a GL3590 either */
-		if (memcmp(self->static_ts.mask_project_ic_type, "3590", 4) != 0) {
-			g_autofree gchar *tmp =
-			    fu_common_strsafe((const gchar *)&self->static_ts.mask_project_ic_type,
-					      sizeof(self->static_ts.mask_project_ic_type));
-			g_set_error(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "IC type %s not supported",
-				    tmp);
-			return FALSE;
-		}
-
-		is3590 = TRUE;
-	}
 
 	/* unsupported static tool string */
 	if (self->static_ts.tool_string_version == 0xff) {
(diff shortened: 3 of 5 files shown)


adamgene:
@gportay Can You help me to identify the check failure

>Analysis failed (could not build the base commit (https://github.com/fwupd/fwupd/commit/dcfda2086764c0571300da4cac7323ed86e852bd))
The analysis for the base commit (dcfda20) was not successful. For more information, see the [build and analysis logs](https://lgtm.com/projects/g/fwupd/fwupd/logs/rev/pr-3fe8282e6dc96af705161b9989101aa4cd20107a/lang:cpp) on LGTM.com

gportay:
> @gportay Can You help me to identify the check failure
> 
> > Analysis failed (could not build the base commit ([dcfda20](https://github.com/fwupd/fwupd/commit/dcfda2086764c0571300da4cac7323ed86e852bd)))
> > The analysis for the base commit ([dcfda20](https://github.com/fwupd/fwupd/commit/dcfda2086764c0571300da4cac7323ed86e852bd)) was not successful. For more information, see the [build and analysis logs](https://lgtm.com/projects/g/fwupd/fwupd/logs/rev/pr-3fe8282e6dc96af705161b9989101aa4cd20107a/lang:cpp) on LGTM.com

Hum, I do not know yet. I need to look deeper of what happened. I cannot reproduce it on my machine.

gportay:
@adamgene I have marked the PR as draft, in my opinion, the changes are not ready to get merged.

We probably should have synced up by making PR on our own user repositories before making that PR upstream. Because it is not common to make a PR on a branch other that main.

However, we would make the PR against the main branch once my commits get merged.

gportay:
The commit look like atomic which is a very good point.

We can do a bit of rewording on the commit stack.

I suggest to prefix the commit titles (aka first line) by "genesys: usbhub: " so the git history looks like consistent with the changes made in the genesys plugin.

Also, could you detail on the commit messages why we need to add extra element in the instance-id string. Giving example of products that need this change will be great; and we make sure this is truly needed.

hughsie:
> Because it is not common to make a PR on a branch other that main

It's OK for me -- it might explain the LGTM failure but for WIP stuff like this it's completely fine.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -605,6 +613,61 @@ fu_genesys_usbhub_device_get_fw_size(FuGenesysUsbhubDevice *self, int bank_num,
...
+
+	version_raw = fu_firmware_get_version_raw(firmware);
+	if (version_raw != 0xffff)
+		self->fw_bank_vers[bank_num] = version_raw;
+
+	return TRUE;
+}
+
+static guint8
+_fu_genesys_revert_hex_char_to_int(guint8 var)
```
gportay, line 659: I think you are re-implementing `g_ascii_xdigit_value()`.

If so, I prefer you reuse it.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -829,6 +858,10 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				   (guint8 *)&self->dynamic_ts,
 				   sizeof(FuGenesysDynamicToolString));
 	}
+	self->dynamic_ts.ss_port_number =
+	    _fu_genesys_revert_hex_char_to_int(self->dynamic_ts.ss_port_number);
+	self->dynamic_ts.hs_port_number =
+	    _fu_genesys_revert_hex_char_to_int(self->dynamic_ts.hs_port_number);
```
gportay, line 864: ditto

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -898,13 +946,18 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 			self->fw_key_addr[1] = 0xE000;
 			self->code_size[0] = self->code_size[1] = self->fw_data_total_count;
 		}
+		self->dynamic_ts.bonding =
+		    _fu_genesys_revert_hex_char_to_int(self->dynamic_ts.bonding);
```
gportay, line 950: can you create a local variable bonding and convert it to int with `bonding = g_ascii_xdigit_value(self->dynamic_ts.bonding));`

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -898,13 +946,18 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		if (self->static_ts.tool_string_version < TOOL_STRING_VERSION_BONDING_QC)
```
gportay, line 951:
```
ts_version = g_ascii_xdigit_value(self->static_ts.tool_string_version);
if (ts_version < ...)
```

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -898,13 +946,18 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+			self->dynamic_ts.bonding <<= 1;
```
gportay, line 952: `bonding <<= 1;`

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -898,13 +946,18 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
 		if (self->dynamic_ts.running_mode == 'M') {
 			self->running_bank = BANK_MASK_CODE;
 		} else if ((self->dynamic_ts.bonding & 0x10) > 0) {
 			self->running_bank = BANK_SECOND;
 		} else {
 			self->running_bank = BANK_FIRST;
 		}
+		fu_device_add_instance_u8(device, "BONDING", (self->dynamic_ts.bonding & 0x0F));
```
gportay, line 960: use the local variable

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -919,6 +972,9 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		} else {
 			self->running_bank = BANK_FIRST;
 		}
+		fu_device_add_instance_u8(device,
+					  "BONDING",
+					  self->dynamic_ts.bonding & GL3590_BONDING_VALID_BIT);
```
gportay, line 977: ditto

plugins/genesys/fu-genesys-usbhub-device.c
gportay: Is that possible to have a macro for the 0x0F value, as for GL3590?

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -971,6 +1027,51 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+			if (!fu_device_build_instance_id(device,
+							 NULL,
+							 "USB",
+							 "VID",
+							 "PID",
+							 "IC",
+							 "BONDING",
+							 NULL))
+				return FALSE;
+
```
gportay, line 1046: remove this empty line

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -971,6 +1027,51 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		} else {
+			const gchar *vendor = fwupd_device_get_vendor(FWUPD_DEVICE(device));
+			g_autofree gchar *ic = NULL;
+			g_autofree gchar *guid_vs_ts = NULL;
+			guint16 port_num = (self->dynamic_ts.ss_port_number << 8) +
```
gportay, line 1051: move the `port_num` before the `g_autofree` locals. to-be-freed variable should be listed at the end.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -971,6 +1027,51 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+							 "PID",
+							 "IC",
+							 "BONDING",
+							 NULL))
+				return FALSE;
+
+		} else {
+			const gchar *vendor = fwupd_device_get_vendor(FWUPD_DEVICE(device));
+			g_autofree gchar *ic = NULL;
+			g_autofree gchar *guid_vs_ts = NULL;
```
gportay, line 1050: `guid` is fine a long as you have a single guid local.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -971,6 +1027,51 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+			guint16 port_num = (self->dynamic_ts.ss_port_number << 8) +
+					   self->dynamic_ts.hs_port_number;
+			ic = fu_common_strsafe((const gchar *)self->static_ts.mask_project_ic_type,
+					       sizeof(self->static_ts.mask_project_ic_type));
+			guid_vs_ts = fwupd_guid_hash_data((const guint8 *)&self->vs_ts,
+							  sizeof(self->vs_ts),
+							  FWUPD_GUID_FLAG_NONE);
+			fu_device_add_instance_strup(device, "VENDOR", vendor);
```
gportay, line 1058: I wonder if the `VENDOR` exists already...

hughsie: VID yes, VENDOR no

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: `_fu_genesys_revert_hex_char_to_int()` converts extend char 'g'~'z', because the string descriptor replay as UTF-16 string format as default. But we use the tool string to contain some info.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -801,9 +829,10 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				    ic_type[5]);
 			return FALSE;
 		}
-		memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
-		tmp = fu_common_strtoull(rev);
-		self->chip.revision = tmp;
+		self->chip.revision = 10 * (self->static_ts.mask_project_ic_type[4] - '0') +
+				      (self->static_ts.mask_project_ic_type[5] - '0');
+		self->static_ts.tool_string_version =
+		    _fu_genesys_revert_hex_char_to_int(self->static_ts.tool_string_version);
```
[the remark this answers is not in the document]
adamgene: OK, I will change the definition in enum `FuGenesysToolStringVersion` to fulfill the usage. Keep the attribute of the structure is make sense.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: I will change the definition in enum FuGenesysToolStringVersion to fulfill the usage with the origin value.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Add GL3523_BONDING_VALID_BIT for that.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: I think these value only use to generate guid, I will keep it as a hex value.

gportay:
@adamgene I have update the branch [wip/gportay/genesys-fix-support-for-gl3590](https://github.com/fwupd/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590); so you should rebase your work on top of it; once it is done, we would set the PR to push directly to the fwupd master branch (if that possible) and mark the PR as "Ready for review".

adamgene:
> @adamgene I have update the branch [wip/gportay/genesys-fix-support-for-gl3590](https://github.com/fwupd/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590); so you should rebase your work on top of it; once it is done, we would set the PR to push directly to the fwupd master branch (if that possible) and mark the PR as "Ready for review".

@gportay  did you update into [wip/gportay/genesys-fix-gl3590-firmware-size](https://github.com/fwupd/fwupd/tree/wip/gportay/genesys-fix-gl3590-firmware-size)? I found genesys-fix-support-for-gl3590 does not changed.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: Oh I missed to till 'Z' range.

So I guess we can reuse the glib code and hack it a bit for the z range.

I would suggest the following:

```
static gint
fu_genesys_tsdigit_value (gchar c)
{
	if (c >= 'A' && c <= 'Z')
		return c - 'A' + 10;
	if (c >= 'a' && c <= 'f')
		return c - 'a' + 10;
	return g_ascii_digit_value (c);                                               
}
```

This way, the function returns -1 if the character is not a toolstring digit; and to some paranoïd testing accross the _setup() vfunc.

What do you think?

gportay:
> @gportay did you update into [wip/gportay/genesys-fix-gl3590-firmware-size](https://github.com/fwupd/fwupd/tree/wip/gportay/genesys-fix-gl3590-firmware-size)? I found genesys-fix-support-for-gl3590 does not changed.

Did you update your remotes? something like `git remote update` ?

adamgene:
> Did you update your remotes? something like `git remote update` ?

image

I saw as the image. Do I miss anything?

gportay:
> I saw as the image. Do I miss anything?

I was the culprit. I pushed to wip/gportay/genesys-fix-gl3590-firmware-size, my bad.

However, it is fixed! sorry for that inconvenience :facepalm:

gportay:
However, the changes look good in general; a little bit of rewording the stack of commits, a second run of review with another round of nitpicks and I guess we are there.

gportay:
@adamgene: it looks your branch needs to be cleaned.

Should should not have merge commit. Instead you should rebase your work.

IMO, the easier way is to use the command line and enter `git rebase --interactive origin/wip/gportay/genesys-fix-support-for-gl3590`, then your editor opens with the list of the commit you want to pick on top of  origin/wip/gportay/genesys-fix-support-for-gl3590. And you will pick only your commits:

```
# remove al line before that the following vvv
pick 2705d957 To skip legacy device that unsupported tool string.
pick ba2bc630 add _fu_genesys_revert_hex_char_to_int() to convert tool string byte. Convert list: 1.static_ts.tool_string_version 2.dynamic_ts.ss_port_number 3.dynamic_ts.hs_port_number 4.dynamic_ts.bonding for gl3523.
pick afe1bc67 add hub instance "BONDING"
pick 4b4188ad add FuGenesysVendorSupportToolString to contain extra tool string info.
pick 7dcdf154 add specific product instance id with: 1."IC" 2."BONDING" 3."VENDOR" - mask code unsupported 4."PORTNUM"  - mask code unsupported 5."VENDORSUP"  - mask code unsupported
pick 41143f0f fix compile and link error for gusb less 0.3.8

# and you don't care about the commented lines, they are discarded.
# Rebase a4d7d8e8..f37b825f onto a4d7d8e8 (32 commands)
# ...
```

git will apply your changes and you will have to resolved the conflicts manually.

```
<<<<<<< HEAD
# original code between <<< and ===
=======
# your changes between === and >>>
>>>>>>> YOUR-COMMIT-ID (YOUR-COMMIT-TITLE)
```

You will have to decide which part you want to keep, according to what the (single) commit changes.

You can see the initial changes of your commit by running `git rebase --show-current-patch`.

Take the time to (re)edit the sources according to the review we made together.

Once, it is done; check the commit compiles.

Once it compiles, you continue the rebase with `git rebase --continue`; the editor opens again and you will reword the commit by prefixing the commit title with "genesys: usbhub: " and adds some useful information in there.

You should do this for every commit that have conflicts, the 6 commits you made in the worse case.

adamgene:
Sorry for that, I do some wrong works during rebase. the final commit will be done.

gportay:
> Sorry for that, I do some wrong works during rebase. the final commit will be done.

No worries, it happens frequently.

I gave a try, the conflicts are very easy to solve ;)

plugins/genesys/fu-genesys-usbhub-device.c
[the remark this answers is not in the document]
adamgene: Please check [ac1f96a](https://github.com/fwupd/fwupd/pull/4426/commits/ac1f96aedb70f22369f90edadb7caf8b45c6f07e).

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -801,9 +829,10 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				    ic_type[5]);
 			return FALSE;
 		}
-		memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
-		tmp = fu_common_strtoull(rev);
-		self->chip.revision = tmp;
```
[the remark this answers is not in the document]
adamgene: I reverted it [ac1f96a](https://github.com/fwupd/fwupd/pull/4426/commits/ac1f96aedb70f22369f90edadb7caf8b45c6f07e).

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: done.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Done.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Please check [ac1f96a](https://github.com/fwupd/fwupd/pull/4426/commits/ac1f96aedb70f22369f90edadb7caf8b45c6f07e).

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Done.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Done.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Done.

adamgene:
I try to rebase main, but it is difficult to me. I will check how to easy do rebase on my IDE.

hughsie:
> I try to rebase main

If it helps, this is what I do:

    git checkout main
    git pull
    git checkout your-branch-name-here
    git rebase main -i

...then squash up commits as required, then exit the exitor

..then fix up rebase failures, then `git rebase --continue` for each one

then

    git push --force

...to update the GitHub PR.

gportay:
@adamgene: I have changed the target branch for the pull request.

Your next step is to rebase your [branch](/adamgene/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590) on top of [main](https://github.com/fwupd/fwupd).

The goal of rebasing is to apply the stack of commits from the *old* base to a *new* base.

In your case you are rebasing your branch [genesys-fix-support-for-gl3590](/adamgene/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590) from the *old* base [genesys-fix-support-for-gl3590](/fwupd/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590) to the new *base* [main](/fwupd/fwupd). *Having identical branch name looks confusing :)*

It looks complicated because the stack has grown too much because of multiple rebase attempt errors, but it is simple actually!

Your true goal beside rebasing your work is to apply **only** the commits you made; and you have only few.

I do not like (trust?) the GUI because I do not know what they doing. So we will do it step by step with the CLI.

0 - I supposed your are already on your branch; if not `git checkout -b rebase adamgene/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590` where `adamgene` is the name of your remote.
 
1 - The first step is to update the fwupd remote: I usually run `git remote update` to get the latest changes of all remotes.

2 - The second step is to start an interactive rebase on top of  [main](/fwupd/fwupd): `git rebase --interactive origin/main` where `origin` is the name of the fwupd upstream remote.

3 - The third step is to identify the commit you want to apply (`pick`) on [main](/fwupd/fwupd). Once you listed the commit you have done the half way!

Your favorite editor was opened with the list of commit you made since the last old base branch (39 commits to be precise); you made 11 commits yourself (including fixes) + the 3 commits that were pending on my [branch](/fwupd/fwupd/tree/wip/gportay/genesys-fix-support-for-gl3590). Normally, you should `pick` only 14 commits for your rebase.

Since a part of your commit appear twice (because of bad rebase attempts); you have to decide either to you pick the first commits or to pick the last ones; I do not recommend to pick up both (much more works). however that is you to decide. I would go with the first.

So, you read line by line the stack of commit in your editor and you remove the line of the commit you are not interested in.

I would suggest the following, but do what you want.

```
pick 9c10e13a genesys: usbhub: Read firmware size from blob
pick ade872fa genesys: usbhub: Get code size from device
pick fa688591 To skip legacy device that unsupported tool string.
pick c4cbd742 add _fu_genesys_revert_hex_char_to_int() to convert tool string byte. Convert list: 1.static_ts.tool_string_version 2.dynamic_ts.ss_port_number 3.dynamic_ts.hs_port_number 4.dynamic_ts.bonding for gl3523.
pick bb373e46 add hub instance "BONDING"
pick 6cfaaf09 add FuGenesysVendorSupportToolString to contain extra tool string info.
pick 4d330247 add specific product instance id with: 1."IC" 2."BONDING" 3."VENDOR" - mask code unsupported 4."PORTNUM"  - mask code unsupported 5."VENDORSUP"  - mask code unsupported
pick 509f0ee4 genesys: usbhub: Add genuine Genesys Logic USB Hub
pick 41143f0f fix compile and link error for gusb less 0.3.8
pick 72676e1d Fix rebase error.
pick a475a3f3 1. delete empty line. 2. rename guid_vs_ts ti guid. 3. move port_num forward.
pick ac1f96ae 1. modify enum FuGenesysToolStringVersion base value to fulfill static_ts.tool_string_version value. 2. remove check static_ts.tool_string_version, because it for unsupported case and it has process before. 3. revert to set chip.revision.
pick 4590fd09 do not convert dynamic_ts.ss_port_number/dynamic_ts.hs_port_number, just use hex value to generate guid.
pick fd0eb6a3 1. refactoring _fu_genesys_revert_hex_char_to_int() to fu_genesys_tsdigit_value(), to unify use. 2. modify to use local value to keep  dynamic_ts.bonding original.
```

4 - There starts the second half of the job: fixing the conflicts! The rebase will apply the commit one by one and stops once a conflicts has to be resolved. Resolved it manually and run `git rebase --continue` to get to the next conflicts or to the end of the rebase.

I had to resolved three conflicts only.

Once you get there, we can do second review and after that we will have to make clean commits by doing some rebase again :fearful:!

Hope this long story is helping!

gportay:
The stack of commits is nicer, good job!

I will review it shortly.

plugins/genesys/fu-genesys-usbhub-device.c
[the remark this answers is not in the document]
gportay: Next step is to make the change directly to the correct commit, and get rid of an additional "fixup" commit.

plugins/genesys/fu-genesys-usbhub-device.c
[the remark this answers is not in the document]
gportay: Next step is to make the change directly to the correct commit, and get rid of an additional "fixup" commit.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: ditto.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: ditto.

gportay:
General remark: A commit is composed of:
- a commit title
- an empty line, and
- the commit message.

The commit does not look good if the empty line is omitted.

See the three dot in the history above.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -748,27 +800,32 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				   (guint8 *)&self->static_ts,
 				   sizeof(FuGenesysStaticToolString));
 	}
-	if (self->static_ts.tool_string_version != 0xff) {
-		gchar rev[3] = {0};
+	if (memcmp(&self->static_ts.tool_string_version, "USB", 3) == 0) {
```
gportay, line 803: `if (memcmp(&self->static_ts, "USB", 3) == 0) {`makes more sense to me as `tool_string_version` is one byte long.

OOI:
- Also, does the `USB` string is a prefix or the full string?
- Does the string `USB`... is reported by the firmware itself or by some USB stack (linux)?
- I wonder if the check should be done in the function `fu_genesys_usbhub_device_get_descriptor_data()` so it would be available for other Tool String data. What do you think?

gportay: My opinion is it makes sense to have two checks "USB" first, and tool string version then `0xff`

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -748,27 +800,32 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INTERNAL,
+			    "Unsupported IC: tool string unsupported");
```
gportay, line 807: I would prefer `"Static Tool String not supported"` or `"Static Tool String %s not supported", (const gchar *)self->static_ts` to stay consistent with other error message.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -748,27 +800,32 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				   (guint8 *)&self->static_ts,
 				   sizeof(FuGenesysStaticToolString));
 	}
-	if (self->static_ts.tool_string_version != 0xff) {
-		gchar rev[3] = {0};
+	if (memcmp(&self->static_ts.tool_string_version, "USB", 3) == 0) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INTERNAL,
```
gportay, line 806: `FWUPD_ERROR_NOT_SUPPORTED`?

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -748,27 +800,32 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "IC type %s not supported",
+			    ic_type);
+		return FALSE;
 	}
+	memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
+	tmp = fu_common_strtoull(rev);
+	self->chip.revision = tmp;
```
gportay, line 828: You can get rid of the `tmp` variable with `self->chip.revision = fu_common_strtoull(rev);` as it was changed by the commit 79228d8b7d69dd76277a85e444835e38aaf6b954.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -816,8 +886,23 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		}
+		if (g_getenv("FWUPD_GENESYS_USBHUB_VERBOSE") != NULL) {
+			fu_common_dump_raw(G_LOG_DOMAIN,
+					   "Vendor support",
+					   (guint8 *)&self->vs_ts,
+					   sizeof(FuGenesysVendorSupportToolString));
+		}
 	}
 
+	/* setup cfi device */
```
gportay, line 905: Make sense to space the code.

That comment would go before `self->cfi_device = fu_genesys_usbhub_device_cfi_setup(self, error);` with an empty line right before to space the blocks:

The following would make me much more happy :wink: 
```
if (!fu_genesys_usbhub_device_set_isp_mode(self, ISP_ENTER, error))     
        return FALSE;

/* setup cfi device */
self->cfi_device = fu_genesys_usbhub_device_cfi_setup(self, error);
```

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -835,84 +920,83 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 		self->flash_sector_size = sector_size;
 
 	/* setup firmware parameters */
-	switch (self->isp_model) {
-	case ISP_MODEL_HUB_GL3523:
+	switch (self->chip.model) {
+	case ISP_MODEL_HUB_GL3523: {
+		gint var = 0;
+		guint8 bonding = 0;
```
gportay, line 926: may be we can use a single variable, instead of two. keep `bonding` and remove `var`.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -835,84 +920,83 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
-		break;
-	case ISP_MODEL_HUB_GL3590:
-		if (!fu_genesys_usbhub_device_get_fw_size(self, 0, error))
+
+		var = fu_genesys_tsdigit_value((gchar)self->dynamic_ts.bonding);
+		if (var == -1) {
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "GL3523 bonding value(%hhX) is wrong",
```
gportay, line 947: `"wrong bonding value %02x"`,

Note: I have not seen `%hh` before.

hughsie: `0x%02x` please -- then we know it's hex at a glance

gportay: yes forgot the `0x` prefix

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > - Does the string USB... is reported by the firmware itself or by some USB stack (linux)?

`USB String Descriptor` reported by the firmware.

---
> - Also, does the USB string is a prefix or the full string? 

It is a prefix, returned values are like
- `09:33:17:0613 FuPluginGenesys      Static info:    55 53 42 32 2e 30 20 48 75 62 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00`, the string is "USB2.0 Hub"
- `09:33:17:0358 FuPluginGenesys      Static info:    55 53 42 33 2e 30 20 48 75 62 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00`, the string is "USB3.0 Hub";
---
> - I wonder if the check should be done in the function fu_genesys_usbhub_device_get_descriptor_data() so it would be available for other Tool String data. What do you think?

A hub unsupported tool string still can update firmware by this plugin, but we don't need to do that at this phase :). So check the `static_ts` at first is fine.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: Well, I am confused about the structure `FuGenesysToolStringVersion`.

``` c
typedef enum {
	TOOL_STRING_VERSION_9BYTE_DYNAMIC,
	TOOL_STRING_VERSION_BONDING,
	TOOL_STRING_VERSION_BONDING_QC,
	TOOL_STRING_VERSION_VENDOR_SUPPORT,
	TOOL_STRING_VERSION_MULTI_TOKEN,
	TOOL_STRING_VERSION_2ND_DYNAMIC,
	TOOL_STRING_VERSION_RESERVED,
	TOOL_STRING_VERSION_13BYTE_DYNAMIC,
 } FuGenesysToolStringVersion;
 ```
 
The following change would break the line 808 `if (self->static_ts.tool_string_version >= TOOL_STRING_VERSION_VENDOR_SUPPORT) {`

 ``` diff
-	TOOL_STRING_VERSION_9BYTE_DYNAMIC,
+	TOOL_STRING_VERSION_9BYTE_DYNAMIC = 0x30, /* '0' */
```

All the enum values are shifted by `0x30`.

You may want to set all correct values or move them as `#define`.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -835,84 +920,83 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		} else if ((bonding & 0x10) > 0) {
+			self->running_bank = BANK_SECOND;
+		} else {
+			self->running_bank = BANK_FIRST;
+		}
+		fu_device_add_instance_u8(device, "BONDING", (bonding & GL3523_BONDING_VALID_BIT));
+		break;
+	}
+	case ISP_MODEL_HUB_GL3590: {
+		guint8 bonding = self->dynamic_ts.bonding;
```
gportay, line 966: This temporary variable does not look to be necessary.

BTW, is it normal the process to get the bonding value is different between GL3590 and GL3523?

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -927,30 +1011,72 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+							 self->public_key,
+							 sizeof(self->public_key),
+							 NULL,
+							 error))
 			return FALSE;
 		guid = fwupd_guid_hash_data(self->public_key,
 					    sizeof(self->public_key),
 					    FWUPD_GUID_FLAG_NONE);
 		fu_device_add_instance_strup(device, "PUBKEY", guid);
+		if (!fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "PUBKEY", NULL))
```
gportay, line 1032: we will group all the call to `fu_device_build_instance_id()` together after all this if/statement statement

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > My opinion is it makes sense to have two checks "USB" first, and tool string version then 0xff

I'm sorry to mislead the `tool_string_version` value in `0xff`, as I said in last comment, the tool string was replied by firmware. In-house usage alters the `tool_string_version` to `0xff` to indicate unsupported of tool string, but we don't alter here, therefore I remove the `0xff` checking.

adamgene: > if (memcmp(&self->static_ts, "USB", 3) == 0) {makes more sense to me as tool_string_version is one byte long.

OK, let me modify that.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: As the [comment](https://github.com/fwupd/fwupd/pull/4426/files#r835114733).
How about `String Descriptor of Tool not supported`?

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -927,30 +1011,72 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+			return FALSE;
+	} else {
```
gportay, line 1034: I guess it is fair to have the following instance id for device that support public-key; i.e. to put the following change not under the else statement.

``` diff
                fu_device_add_instance_strup(device, "PUBKEY", guid);
        }
+
+        /* add specific product info */
+        if (self->running_bank != BANK_MASK_CODE) {
+ ...
+        }

+        fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "IC", NULL);
+        fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "IC", "BONDING", NULL);
+        fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "VENDOR", IC", "BONDING", "PORTNUM", "VENDORSUP", NULL);
        fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "PUBKEY", NULL);
```

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -927,30 +1011,72 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		/* add specific product info */
+		if (self->running_bank == BANK_MASK_CODE) {
```
gportay, line 1036: you can move that outside the if statement. and have a `if (... != BANK_MASK_CODE)` with the process that was in the else statement.

This may share some code with the error IC type not supported code at line 759.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -816,8 +886,23 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
+		if (g_getenv("FWUPD_GENESYS_USBHUB_VERBOSE") != NULL) {
+			fu_common_dump_raw(G_LOG_DOMAIN,
+					   "Vendor support",
+					   (guint8 *)&self->vs_ts,
+					   sizeof(FuGenesysVendorSupportToolString));
+		}
 	}
 
+	/* setup cfi device */
 	if (fu_device_has_private_flag(device, FU_GENESYS_USBHUB_FLAG_HAS_PUBLIC_KEY)) {
```
adamgene, line 906: I find this block is like `fu_genesys_usbhub_device_detach()`, should we replace to call it?

plugins/genesys/fu-genesys-usbhub-firmware.c
```diff
@@ -14,10 +14,55 @@
...
+			    error))
+		return FALSE;
+
+	if (memcmp(project_ic_type, "3590", 4) == 0) {
+		self->chip.model = ISP_MODEL_HUB_GL3590;
+		self->chip.revision = 10 * (project_ic_type[4] - '0') + (project_ic_type[5] - '0');
+		return TRUE;
+	}
+
+	g_set_error_literal(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED, "Unsupported IC");
```
gportay, line 62: Can we add the following?

```
g_autofree gchar *ic_type =                             
fu_common_strsafe((const gchar *)mask_project_ic_type, sizeof(mask_project_ic_type));
g_set_error(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED, "IC type %s not supported", ic_type); 
```

gportay (commented):
@adamgene Even if I am requesting some changes (many?), the commits look very good to me.

Many thanks for what you did!

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -816,8 +886,23 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
...
 		if (!fu_genesys_usbhub_device_authenticate(self, error))
 			return FALSE;
```
adamgene, line 908: I find this block is like `fu_genesys_usbhub_device_detach()`, should we replace to call it?

plugins/genesys/fu-genesys-usbhub-firmware.c
adamgene: The field may not contain correct string in "IC type". There are different offsets of firmware to indicate IC type.

gportay: oh yes I remember why I have not add it :)

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: The values are enumerated as the same as the string descriptor reported.
If I want  a series of definition, should I use `#define` or `enum`?

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Sorry, I'm not sure your question. Do you mean the process to get right bonding value from GL3590 and GL3523 has different?

As FuGenesysDynamicToolString comments, the `bonding` byte contains many information for GL3590 and GL3523. For new ICs(e.g. GL3525) they have new dynamic tool string format.

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -609,11 +611,61 @@ fu_genesys_usbhub_device_get_fw_size(FuGenesysUsbhubDevice *self, int bank_num,
...
+	g_autoptr(FuFirmware) firmware = NULL;
+	g_autofree guint8 *buf = NULL;
+
+	g_return_val_if_fail(bank_num < 2, FALSE);
+
+	if (self->code_size[bank_num] == 0)
+		if (!fu_genesys_usbhub_device_get_fw_size(self, bank_num, error))
+			return FALSE;
+
+	if (!fu_genesys_usbhub_device_check_fw_signature(self, bank_num, error))
```
adamgene, line 636: I found `fu_genesys_usbhub_device_check_fw_signature()` already checked in `fu_genesys_usbhub_device_get_fw_size()`, can I add `else{}` to separate them?

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: `PUBKEY` cannot be set without `FU_GENESYS_USBHUB_FLAG_HAS_PUBLIC_KEY` set.
How about add `fu_genesys_usbhub_build_extend_instance_id()`?

gportay: You can just ignore the return value, as it was.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: > I'm sorry to mislead the tool_string_version value in 0xff, as I said in last comment, the tool string was replied by firmware. In-house usage alters the tool_string_version to 0xff to indicate unsupported of tool string, but we don't alter here, therefore I remove the 0xff checking.

Well, to make sure I get it right; none of the firmware (current and past) would reply a static tool string version value of `0xff`? which means the previous check was dead code, right?

And you replaced the check by `"USB"` because you found `3520` does not reply a static tool string structure, right?

Just for curiosity, I found two GL3521-22 (with genesys VID/PID) in my devices with firmware `92.??` (do not recall the version). Can I found firmware upgrade for these devices? Also, is that possible to buy GL3520 and GL3590 end-products somewhere?

plugins/genesys/fu-genesys-usbhub-device.c
gportay: >  How about String Descriptor of Tool not supported?

I wanted to specify that it is the *Static Tool String* (i.e. not Dynamic, not Firmware, not Vendor...).

I guess, your suggestion makes sense if the check is done in the  function `fu_genesys_usbhub_device_get_descriptor_data()`.

Therefore, if it is correct to check for the "USB" prefix for Static, Dynamic, Firmware, Vendor TS... go ahead and move the `if (memcmp(...))` to `fu_genesys_usbhub_device_get_descriptor_data()`, that looks better to my eyes. Does it make sense?

plugins/genesys/fu-genesys-usbhub-device.c
gportay: I mean that for GL3523, you do

``` c
var = fu_genesys_tsdigit_value(...);
var = (var & 0xff)
if (self->static_ts.tool_string_version < TOOL_STRING_VERSION_BONDING_QC)
    var <<= 1;
self->bonding = (var & GL3523_BONDING_VALID_BIT);  
```

while it is that short for GL3590

``` c
self->bonding = (self->dynamic_ts.bonding & GL3590_BONDING_VALID_BIT);
```

gportay: which means that, the GL3523 is using the Static TS, and the GL3590 is using the Dynamic TS.

I am just checking if that is correct.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: It makes sense to me.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: Well, I do not like that new function `fu_genesys_usbhub_device_get_fw_version()`.

I prefer you copy/paste `fu_genesys_usbhub_device_get_fw_size()`, create and use the new `GENESYS_USBHUB_VERSION_OFFSET` instead.

``` c
/* read the firmware vresion from the firmware stored in the device */
static gboolean
fu_genesys_usbhub_device_get_fw_version(FuGenesysUsbhubDevice *self, int bank_num, GError **error)
{
	guint16 version_raw = 0;

	g_return_val_if_fail(bank_num < 2, FALSE);
	g_return_val_if_fail(self->fw_bank_vers[bank_num] == 0, FALSE);

	if (!fu_genesys_usbhub_device_check_fw_signature(self, bank_num, error))
		return FALSE;

	/* get firmware version from device */
	if (!fu_genesys_usbhub_device_read_flash(self,
						 self->fw_bank_addr[bank_num] +
						     GENESYS_USBHUB_VERSION_OFFSET,
						 &kbs,
						 1,
						 NULL,
						 error)) {
		g_prefix_error(error, "error getting fw version from device: ");
		return FALSE;
	}
	if (version_raw != 0xffff)
		self->fw_bank_vers[bank_num] = version_raw;

	/* success */
	return TRUE;
}
```

But the whole thing is undoing the verifying of firmware integrity. Is that correct to not verify the firmware integrity?

plugins/genesys/fu-genesys-usbhub-device.c
gportay: Could you list all the values and there "names"?

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -889,10 +901,12 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 
 	/* setup firmware parameters */
 	switch (self->chip.model) {
-	case ISP_MODEL_HUB_GL3523:
+	case ISP_MODEL_HUB_GL3523: {
+		gint var = 0;
```
gportay, line 925: I tend to use `tmp` for temporary variables; it as more sense that just `var`.

However, `bonding` sounds better.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: The `README.md` should be updated accordingly with the new introduced instance id.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > Could you list all the values and there "names"?

The `FuGenesysToolStringVersion` is enough for current usage.
In-house firmware spec, they are separated in different files. The  `FuGenesysToolStringVersion` is re-translated by my opinion. I have to search and collect again to update them.

Do you have `GL Tool String v1.XX.pdf` on hand?

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: >Well, to make sure I get it right; none of the firmware (current and past) would reply a static tool string version value of 0xff? which means the previous check was dead code, right?

Yes, the `0xff` check is copied from windows code. For linux code base I gave to you, is supporting for GL3523 and GL3590 only. But for other product, there has extra check code.

---
> And you replaced the check by "USB" because you found 3520 does not reply a static tool string structure, right?

Yes, USB string descriptor replied by firmware. And our USB hub replies "Product String Descriptor" for unknown index of String Descriptor. BTW, "Product String Descriptor" default is "USB2.0 Hub" or "USB2.1 Hub" for our USB3.x hub(pure 2.0 hub does not support).

---
>Just for curiosity, I found two GL3521-22 (with genesys VID/PID) in my devices with firmware 92.?? (do not recall the version). Can I found firmware upgrade for these devices?

You can do these,
1. Dump firmware from the flash.
2. Erase flash to run mask code. 
3. Update the dumped firmware.

---
> Also, is that possible to buy GL3520 and GL3590 end-products somewhere?

I don't know how many products are there, you can ask for our PMs and sales to query more info.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > I am just checking if that is correct.

The spec history says the improvement in newer tool string version; the bonding byte has shift to indicate QC supporting for GL3523.

How about add `fu_genesys_usbhub_device_get_bonding(FuGenesysUsbhubDevice *self, GError **error)` to do that?

plugins/genesys/fu-genesys-usbhub-device.c
gportay: > Do you have GL Tool String v1.XX.pdf on hand?

Sorry, I don't; I have just understood why you shift if by 0x30, which is ascii for '0' as suggested by your comment. Maybe we should make the comment more explicit.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > But the whole thing is undoing the verifying of firmware integrity. Is that correct to not verify the firmware integrity?

This function's goal is to get the firmware version of the valid firmware bank. So verify the firmware integrity before get the version.

The verify flow should likes:
1. check signature(XROM for hub).
2. get firmware code size if it is not fixed.
3. read entire firmware.
4. calculate and match checksum.
--- 
Therefore,  I think we can modify below.
1. remove calling `fu_genesys_usbhub_device_check_fw_signature()` in fu_genesys_usbhub_device_get_fw_size().
2. modify in `fu_genesys_usbhub_device_get_fw_version()`
```
	if (!fu_genesys_usbhub_device_check_fw_signature(self, bank_num, error))
		return FALSE;

	if (self->code_size[bank_num] == 0) {
		if (!fu_genesys_usbhub_device_get_fw_size(self, bank_num, error))
			return FALSE;
	}
```
3. rename `fu_genesys_usbhub_device_get_fw_version()`, I think it is confuse for other. Which one is better, get_bank_version or query_bank_fw_version? Do you have any idea?

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Ok, that's do it.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > You can just ignore the return value, as it was.

I saw that :). It's interesting to check, in somewhere this "must" to do. I will modify on next.

---
BTW, @hughsie I found an error in `fu_usb_device_query_hub()` https://github.com/fwupd/fwupd/blob/d2adec18ca6838bfc0312fbe6b02983af680ac00/libfwupdplugin/fu-usb-device.c#L200-L207

The [USB 2.0 Specification](https://www.usb.org/document-library/usb-20-specification) `11.23.2.1 Hub Descriptor` described the `bDescLength` should fulfill `bNbrPorts`.

When I test our 8 ports GL3590, it was rejected in `fu_usb_device_setup()` https://github.com/fwupd/fwupd/blob/d2adec18ca6838bfc0312fbe6b02983af680ac00/libfwupdplugin/fu-usb-device.c#L324-L328

hughsie: @adamgene would this be more accurate?

	if (sz >= 0x0C) {
		g_autofree gchar *hub = g_strdup_printf("%02X%02X", data[0x0B], data[0x0A]);
		fu_device_add_instance_str(FU_DEVICE(self), "HUB", hub);
	} else if (sz >= 0x08) {
		fu_device_add_instance_u8(FU_DEVICE(self), "HUB", data[0x07]);
	}
	fu_device_build_instance_id(FU_DEVICE(self), NULL, "VID", "PID", "HUB", NULL)

adamgene: In USB 3 hub, the sz is fixed 0x0C(because max supporting ports are 15).
In USB 2 hub, the sz is in 0x09~(because max supporting ports are 255).

My 8 ports GL3590 replied as below. Hope this is a good info for you.

- Replied from USB 2
```
        ------------------- USB Hub Descriptor -----------------
bDescriptorLength        : 0x0A (10 bytes)
bDescriptorType          : 0x29 (Hub Descriptor)
bNumberOfPorts           : 0x08 (8 Ports)
wHubCharacteristics      : 0x00E9
 Power switching         : Individual
 Compound device         : No
 Over-current protection : Individual
 TT Think Time           : At most 32 FS bit times
 Port Indicators         : Supported
bPowerOnToPowerGood      : 0x3C (120 ms)
bHubControlCurrent       : 0x64 (100 mA)
DeviceRemovable          : 0x00, 0x00
 Bit 0                   : 0 (reserved, any value)
 Bit 1                   : 0 (Device at Port 1 is removable)
 Bit 2                   : 0 (Device at Port 2 is removable)
 Bit 3                   : 0 (Device at Port 3 is removable)
 Bit 4                   : 0 (Device at Port 4 is removable)
 Bit 5                   : 0 (Device at Port 5 is removable)
 Bit 6                   : 0 (Device at Port 6 is removable)
 Bit 7                   : 0 (Device at Port 7 is removable)
 Bit 8                   : 0 (Device at Port 8 is removable)
 Bit 9                   : 0 (unused, must be 0)
 Bit10                   : 0 (unused, must be 0)
 Bit11                   : 0 (unused, must be 0)
 Bit12                   : 0 (unused, must be 0)
 Bit13                   : 0 (unused, must be 0)
 Bit14                   : 0 (unused, must be 0)
 Bit15                   : 0 (unused, must be 0)
PowerControlMask         : 0xFF, 0x00
Data (HexDump)           : 0A 29 08 E9 00 3C 64 00 00 FF
```

- Replied from USB 3
```
        ------------------- USB Hub Descriptor -----------------
bDescriptorLength        : 0x09 (9 bytes)
bDescriptorType          : 0x29 (Hub Descriptor)
bNumberOfPorts           : 0x04 (4 Ports)
wHubCharacteristics      : 0x0009
 Power switching         : Individual
 Compound device         : No
 Over-current protection : Individual
 TT Think Time           : At most 8 FS bit times
 Port Indicators         : Not supported
bPowerOnToPowerGood      : 0x3C (120 ms)
bHubControlCurrent       : 0x90 (144 mA)
DeviceRemovable          : 0x00
 Bit 0                   : 0 (reserved, any value)
 Bit 1                   : 0 (Device at Port 1 is removable)
 Bit 2                   : 0 (Device at Port 2 is removable)
 Bit 3                   : 0 (Device at Port 3 is removable)
 Bit 4                   : 0 (Device at Port 4 is removable)
 Bit 5                   : 0 (unused, must be 0)
 Bit 6                   : 0 (unused, must be 0)
 Bit 7                   : 0 (unused, must be 0)
Data (HexDump)           : 09 29 04 09 00 3C 90 00 FF

        ---------------- USB 3.0 Hub Descriptor ---------------
bLength                  : 0x0C (12 bytes)
bDescriptorType          : 0x2A
bNumberOfPorts           : 0x04 (4 Ports)
wHubCharacteristics      : 0x0009
 Power switching         : Individual
 Compound device         : No
 Over-current protection : Individual
 TT Think Time           : At most 8 FS bit times
 Port Indicators         : Not supported
bPowerOnToPowerGood      : 0x3C (120 ms)
bHubControlCurrent       : 0x90 (576 mA)
bHubHdrDecLat            : 0x04 (0.4 µs)
wHubDelay                : 0x190 (400 ns)
DeviceRemovable          : 0x00
 Bit 0                   : 0 (reserved, any value)
 Bit 1                   : 0 (Device at Port 1 is removable)
 Bit 2                   : 0 (Device at Port 2 is removable)
 Bit 3                   : 0 (Device at Port 3 is removable)
 Bit 4                   : 0 (Device at Port 4 is removable)
 Bit 5                   : 0 (unused, must be 0)
 Bit 6                   : 0 (unused, must be 0)
 Bit 7                   : 0 (unused, must be 0)
Data (HexDump)           : 0C 2A 04 09 00 3C 90 04 90 01 00 00
```

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > Maybe we should make the comment more explicit.

Current comment is `/* start at 0x30('0'), align to replied values. */`
How about change to `/* start at 0x30('0'), align to tool string replied values. */`?

plugins/genesys/fu-genesys-usbhub-device.c
hughsie: TIL: https://github.com/fwupd/fwupd/pull/4446 -- does that work?

adamgene (commented):
https://github.com/fwupd/fwupd/pull/4426#discussion_r835956076

This issue is big for us currently. If the first tool string is replied "USB XXX", I don't want to get the next ts and return unsupported. But in the future, we may be requested supporting old GL3520.

---
In this moment, we can judge "USB" in `fu_genesys_usbhub_device_get_descriptor_data()` as well.

plugins/genesys/fu-genesys-common.h
```diff
@@ -28,6 +48,51 @@ typedef struct __attribute__((packed)) {
 	guint8 firmware_version[4]; /* MMmm=MM.mm (ASCII string) */
 } FuGenesysStaticToolString;
 
+typedef struct __attribute__((packed)) {
+	guint8 running_mode; /* 'M' or 'C' */
```
gportay, line 52: I forgot to ask. could you tell in one word what means 'M' and 'C'?

For example: `*/ 'M' for mask or 'C' for code*/`; any other values?

plugins/genesys/fu-genesys-usbhub-device.c
gportay: OOI, '0' is for `0`; after the nineth version, '9' what will be the next TS value? 'A` or 'a'?

The enum would be broken. 0x30 -> 0x41 or 0x61.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: I was just asking.

If you think this makes the code nicer go ahead. I have no strong opinion for now.

plugins/genesys/fu-genesys-usbhub-device.c
gportay: For me `"Static Tool String not supported"` makes more sense; because it checks for the static.

But if you prefer `"Tool String not supported"` or `"String Descriptor of Tool not supported"` go ahead.

I tend to stay consistent when I can; ~~Unsupported IC~~ and ~~unsupported~~ -> not supported;

plugins/genesys/fu-genesys-usbhub-device.c
gportay: What was wrong with the previous code already? the one in the `_setup()` vfunc?
  
> 1. check signature(XROM for hub).
> 1. get firmware code size if it is not fixed.
> 1. read entire firmware.
> 1. calculate and match checksum.

I guess the `_setup()` should handle the check point 2.

Can we query the running bank to the firmware? If so, we can get rid of points 3. and 4.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: The values defined are not over '9'. I will discuss your question in-house.

gportay: > The values defined are not over '9'. I will discuss your question in-house.

At least not yet :D

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: > Let's step back a bit; what was wrong with the previous code already?

The fu_genesys_usbhub_device_check_fw_signature() already called in fu_genesys_usbhub_device_get_fw_size(), it may check twice when `if (self->code_size[bank_num] == 0)`. 
And it becomes wrong if I remove calling fu_genesys_usbhub_device_check_fw_signature() in fu_genesys_usbhub_device_get_fw_size(). Therefore I have to change the calling serial.

> Can we query the running bank to the firmware? If so, we can get rid of points 3. and 4.

Yes, we already set into `_FuGenesysUsbhubDevice::running_bank`. We can check the not running bank only.

gportay: > The fu_genesys_usbhub_device_check_fw_signature() already called in fu_genesys_usbhub_device_get_fw_size(), it may check twice when if (self->code_size[bank_num] == 0).
And it becomes wrong if I remove calling fu_genesys_usbhub_device_check_fw_signature() in fu_genesys_usbhub_device_get_fw_size(). Therefore I have to change the calling serial.

... I was thinking even further. What was wrong in the code that was in the `_setup()` vfunc (before your changes)? because, if you do not need those current changes, we can leave it as it was, merge the PR and clean that code afterward.

> Yes, we already set into _FuGenesysUsbhubDevice::running_bank. We can check the not running bank only.

Nice. This would be nice to update the `_dump_firmware()` to dump the running bank instead of the first one; we can do it later.

gportay: Perfect. We will do the change after the PR is merged.

plugins/genesys/fu-genesys-common.h
gportay: what does mask mean exactly out of curiosity?

plugins/genesys/fu-genesys-usbhub-device.c
gportay: I would prefer to stick to the original enum (without the `= 0x30`) and use `_tsdigit()` when necessary.

plugins/genesys/fu-genesys-common.h
adamgene: > what does mask mean exactly out of curiosity?

An easy way to know is that the mask code was designed in a IC(fixed in hardware). Our IC can load a firmware from flash(bank code or ram code).

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: OK, I will do that.

adamgene:
@gportay I have minimize the commit, please check that.

gportay:
@adamgene looks good to me.

@hughsie do you want to make a pass with your eagle eyes?

hughsie (approved):

adamgene:
@hughsie I changed the draft to open state, and I understood it may not need to rebase at last.

PR #4431, opened 2022-03-23
tpm: Fix eventlog replay for Intel TXT machines

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm/fu-tpm-eventlog-common.c
+++ b/plugins/tpm/fu-tpm-eventlog-common.c
@@ -149,6 +149,11 @@ fu_tpm_eventlog_blobstr(GBytes *blob)
 			       g_bytes_get_size(blob));
 }
 
+typedef struct __attribute__((packed)) {
+	guint8 Signature[16];
+	guint8 StartupLocality;
+} TCG_EfiStartupLocalityEvent;
+
 GPtrArray *
 fu_tpm_eventlog_calc_checksums(GPtrArray *items, guint8 pcr, GError **error)
 {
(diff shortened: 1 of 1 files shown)


plugins/tpm/fu-tpm-eventlog-common.c
```diff
@@ -175,6 +175,22 @@ fu_tpm_eventlog_calc_checksums(GPtrArray *items, guint8 pcr, GError **error)
 		FuTpmEventlogItem *item = g_ptr_array_index(items, i);
 		if (item->pcr != pcr)
 			continue;
+
+		/* if TXT is enabled then the first event for PCR0 should be a StartupLocality */
+		if (item->kind == EV_NO_ACTION && item->pcr == 0 && item->blob != NULL && i == 0) {
+			gsize bufsz;
+			const guint8 *buf = g_bytes_get_data(item->blob, &bufsz);
+			if (bufsz == 17 && buf[bufsz - 2] == 0) {
```
superm1, line 183: still want to decode the magic buffer size here so we better understand them when the size changes.

hughsie: @flanfly is this encoded into the event data or hardcoded? Thanks.

hughsie: @superm1 how about that? It's pretty self documenting now...

superm1 (approved):

plugins/tpm/fu-tpm-eventlog-common.c
flanfly: The event data is always "StartupLocality\x00\xXX" whereas \xXX is the TPM 2.0 locality the Startup command was issued from (should be either 0 or 3).

PR #4434, opened 2022-03-23
WIP: fu-tool: Error if too many argument given

gportay:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

I found an something inconsistent between the help message and the number of argument; but I do not recall what... However, this is a draft and I will update the message latter.

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -447,6 +447,15 @@ fu_util_get_plugins(FuUtilPrivate *priv, gchar **values, GError **error)
 {
 	GPtrArray *plugins;
 
+	/* parse arguments */
+	if (g_strv_length(values) != 0) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_ARGS,
+				    "Invalid arguments");
+		return FALSE;
+	}
+
 	/* load engine */
 	if (!fu_util_start_engine(priv, FU_ENGINE_LOAD_FLAG_HWINFO, error))
 		return FALSE;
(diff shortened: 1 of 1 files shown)


hughsie:
@superm1 I've got a neuron that's buzzing; wasn't there a reason we ignore the extra args?

superm1:
> @superm1 I've got a neuron that's buzzing; wasn't there a reason we ignore the extra args?

Yes!  I think it's bash-completion breaks for autocompleting devices.

gportay:
> Yes! I think it's bash-completion breaks for autocompleting devices.

What is wrong with bash autocompleting?

Does it make sense to fix the the argument parsing with a "too many argument given" error?

superm1:
> > Yes! I think it's bash-completion breaks for autocompleting devices.
> 
> What is wrong with bash autocompleting?
> 
> Does it make sense to fix the the argument parsing with a "too many argument given" error?

Bash completion at parses get-devices output but you need to have device id and also a string describing it to know which one you are using.

We pass the combined string in for bash completion and the tools throw away the non GUID/device id output.

If you want to do this pull request, we need a better way to handle that.

superm1:
> > We pass the combined string in for bash completion
> 
> Is there any env var we could use to know that we're being called from bash-completion and disable the (IMHO useful) checks here?

I would think we can add one

superm1:
Mmm actually I don't know that will work. The problem isn't the bash completion - it's the use of the output for bash completion.

What about string detection from the name instead? Or instead putting everything in a quote so that it's "one" argument.

hughsie:
> it's the use of the output for bash completion

Could we use the json output? I'm not sure I understand quite how cnf actually works....

gportay:
It is in my TODO list, will do it after the new supported device.

PR #4435, opened 2022-03-24
ccgx: add new dock, add CcgxDmcTrigCode quirk key

chlee75:
Type of pull request : Feature

It is ccgx plugin candidate that support Lenovo new Dock.
Please review this code.

Changes
 1. Add Lenovo ThinkPad USB-C Dock in quirk file
   - It support update on disconnect that user need to disconnect a dock from a computer after firmware update finish.
 2. Add CcgxDmcTrigCode quirk key instead of DmcTriggerCode enum.
   - Some DMC dock need trigger code to update firmware while other DMC dock do not need it.
   - New Trigger code can be added according to customer request
 3. Add signed payload check in fu_ccgx_dmc_device_setup()


plugins/ccgx/fu-ccgx-dmc-device.c
```diff
@@ -318,6 +319,7 @@ fu_ccgx_dmc_device_to_string(FuDevice *device, guint idt, GString *str)
 				   fu_ccgx_fw_image_type_to_string(self->fw_image_type));
 	fu_common_string_append_kx(str, idt, "EpBulkOut", self->ep_bulk_out);
 	fu_common_string_append_kx(str, idt, "EpIntrIn", self->ep_intr_in);
+	fu_common_string_append_kx(str, idt, "Trigger Code", self->trig_code);
```
hughsie, line 322: `Trigger`Code` here please

plugins/ccgx/fu-ccgx-dmc-device.c
```diff
@@ -684,6 +682,9 @@ fu_ccgx_dmc_device_setup(FuDevice *device, GError **error)
 	fu_device_set_version(FU_DEVICE(self), version);
 	fu_device_set_version_raw(FU_DEVICE(self), version_raw);
 	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+
+	if (dock_id.custom_meta_data_flag > 0)
+		fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD);
```
hughsie, line 687: does that mean we can add `FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD` if the opposite is true?

plugins/ccgx/fu-plugin-ccgx.c
```diff
@@ -26,6 +26,7 @@ fu_plugin_ccgx_init(FuPlugin *plugin)
 	fu_context_add_quirk_key(ctx, "CcgxFlashRowSize");
 	fu_context_add_quirk_key(ctx, "CcgxFlashSize");
 	fu_context_add_quirk_key(ctx, "CcgxImageKind");
+	fu_context_add_quirk_key(ctx, "CcgxDmcTrigCode");
```
hughsie, line 29: I think this should probably be `CcgxDmcTriggerCode` and also documented in README.md

mrhpearson:
Hate to be the bearer of bad tidings - but support for this dock has not been approved internally at Lenovo and there are concerns at adding LVFS support until that is confirmed. The docking team don't want to set incorrect expectations for customers.
Also - please don't use internal codenames publicly.

Can we relabel this so it's for supporting the Infineon device and not neccessarily linked to the Lenovo dock to avoid any misunderstandings? Either that or I'm afraid it should be mothballed until formal support is agreed.

Thanks
Mark

superm1:
> Hate to be the bearer of bad tidings - but support for this dock has not been approved internally at Lenovo and there are concerns at adding LVFS support until that is confirmed. The docking team don't want to set incorrect expectations for customers. Also - please don't use internal codenames publicly.
> 
> Can we relabel this so it's for supporting the Infineon device and not neccessarily linked to the Lenovo dock to avoid any misunderstandings? Either that or I'm afraid it should be mothballed until formal support is agreed.
> 
> Thanks Mark

Just to make sure I follow - the ask is just to drop/rename the comment in the quirk file right?  But still "do the change".  I think that's the "right" answer at least to keep the change in place.  Then if/when it's decided to make it POR then add it to LVFS website.
This will cause all clients to not show the "supported" flag when they probe it until there is something on LVFS.  It will also let this percolate into more distributions so that by the time things are decided to be supported there isn't also extra work at at that time to do too.

superm1:
### Thanks for your commit

**Clang Format**
 
**Pre-commit overview**

plugins/ccgx/fu-ccgx-dmc-device.c
```diff
@@ -616,33 +621,26 @@ fu_ccgx_dmc_device_attach(FuDevice *device, FuProgress *progress, GError **error
...
-		if (!fu_ccgx_dmc_device_send_download_trigger(self, trigger_code, error)) {
-			g_prefix_error(error, "download trigger error: ");
-			return FALSE;
-		}
+		if (self->trig_code > 0) {
+			if (!fu_ccgx_dmc_device_send_download_trigger(self, self->trig_code, error)) {
+				g_prefix_error(error, "download trigger error: ");
+				return FALSE;
+			}
+			if (self->trig_code == 2) /* trigger code for update on disconnect */
```
superm1, line 629: so I think you want to probably do something to artificially stop updates from being staged again until it's been unplugged.  I suggest adding an inhibit.

mrhpearson:
> Just to make sure I follow - the ask is just to drop/rename the comment in the quirk file right? But still "do the change". I think that's the "right" answer at least to keep the change in place. Then if/when it's decided to make it POR then add it to LVFS website. This will cause all clients to not show the "supported" flag when they probe it until there is something on LVFS. It will also let this percolate into more distributions so that by the time things are decided to be supported there isn't also extra work at at that time to do too.

I _think so_ - that makes sense to me.
The quirk shouldn't mention Lenovo (or particularly Viking) - Lenovo is not supporting this dock on Linux yet, and don't want to set the expectation that we are or that it should be expected (I personally want it to happen...but that's by the by).

plugins/ccgx/fu-ccgx-dmc-device.c
hughsie, line 629: Isn't this just DMC_TRIGGER_CODE_UPDATE_ON_DISCONNECT -- which was removed in this commit?

plugins/ccgx/ccgx.quirk
```diff
@@ -51,13 +51,25 @@ Summary = CCGx Power Delivery Device (Symmetric FW1)
 [USB\VID_04B4&PID_5218&SID_1F00&APP_6432&MODE_FW2]
 Summary = CCGx Power Delivery Device (Symmetric FW2)
 
+# Lenovo ThinkPad USB-C Dock Viking
+[USB\VID_17EF&PID_30A9]
+Plugin = ccgx
+GType = FuCcgxDmcDevice
+Summary = Dock Management Controller Device
+ParentGuid = USB\VID_17EF&PID_30AF
+Name = Lenovo ThinkPad USB-C Dock Viking
```
superm1, line 60: I guess based on Mark's comments this should just be "USB-C Dock".  If later this is made POR it can be renamed to include some branding.

plugins/ccgx/fu-ccgx-dmc-device.c
chlee75: > does that mean we can add FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD if the opposite is true?

Yes. your are right. I will add FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD if the opposite is true?

plugins/ccgx/fu-ccgx-dmc-device.c
chlee75: I will change it to TriggerCode

plugins/ccgx/fu-ccgx-dmc-device.c
chlee75: > so I think you want to probably do something to artificially stop updates from being staged again until it's been unplugged. I suggest adding an inhibit.

Device start update firmware internally when user disconnect device from a computer.

Do I need to add an inhibit in this case?

If yes, could you let me know how to add inhibit for this case?

superm1: Yes you should inhibit it so that a user doesn't try to run the update twice in a row before they unplug.
Here's an example: https://github.com/fwupd/fwupd/blob/da37d93eb03822e67dfb13377f14d95b27426291/plugins/dell-dock/fu-dell-dock-i2c-ec.c#L585

chlee75: > Isn't this just DMC_TRIGGER_CODE_UPDATE_ON_DISCONNECT -- which was removed in this commit?
 
I will remove this code and add has-manual-replug to the quirk file for the device that support update on disconnect

plugins/ccgx/ccgx.quirk
chlee75: > I guess based on Mark's comments this should just be "USB-C Dock". If later this is made POR it can be renamed to include some branding.

I will change Name as USB-C Dock.

@mrhpearson 
Wistron asked me to develop LVFS for Lenovo Viking dock three weeks ago.
So I requested code review as a prototype. There will be many things to be defined.
Let me know your email in Lenovo so that l make email for this issue.

chlee75: Please check [53c653c](https://github.com/fwupd/fwupd/pull/4435/commits/53c653ca9db90a142b95992e167f06f819396a0f)

plugins/ccgx/fu-ccgx-dmc-device.c
chlee75: Please check [53c653c](https://github.com/fwupd/fwupd/pull/4435/commits/53c653ca9db90a142b95992e167f06f819396a0f)

plugins/ccgx/fu-plugin-ccgx.c
chlee75: Please check [53c653c](https://github.com/fwupd/fwupd/pull/4435/commits/53c653ca9db90a142b95992e167f06f819396a0f)

plugins/ccgx/fu-ccgx-dmc-device.c
chlee75: v

plugins/ccgx/fu-ccgx-dmc-device.c
chlee75: Please check [53c653c](https://github.com/fwupd/fwupd/pull/4435/commits/53c653ca9db90a142b95992e167f06f819396a0f)

chlee75:
It is just a prototype for a new dock and a future dock.
Please review the changes and no need to merge this branch to main.

After changes are OK, then I will make a new pull request later when I finish fixing details such as dock name, remove delay etc.

chlee75:
@superm1 

> Run ./contrib/setup to set up [pre-commit](https://pre-commit.com/) locally. This will help ensure that you have all the build > dependencies necessary and automatically reformat code in the future.
> Run ./contrib/reformat-code.py main to reformat all code that has changed in your local branch relative to main
> Run git commit add .; git commit --amend or use git rebase -i to amend your commit(s).

I run ./contrib/setup in other fwupd source code before.
Do I need to run ./contrib/setup in fwupd that I am working before I commit?

Should I need to run ./contrib/setup whenever I commit?

plugins/ccgx/fu-ccgx-dmc-device.c
```diff
@@ -616,33 +620,32 @@ fu_ccgx_dmc_device_attach(FuDevice *device, FuProgress *progress, GError **error
...
-			    FWUPD_ERROR,
-			    FWUPD_ERROR_NOT_SUPPORTED,
-			    "invalid update model = %u",
-			    self->update_model);
-		return FALSE;
 	}
 
-	if (manual_replug)
+	if (manual_replug) {
+		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_NEEDS_ACTIVATION);
```
superm1, line 639: Is there a way to actually trigger update without disconnect too after staged?

If so this flag makes sense but you need to add another vfunc to handle sending that separately if requested by the client. I can show you an example from another plugin if so.

If it's not possible (or desirable) then please remove this flag and only use the inhibition.

chlee75:
Close this pull request due to confidential code name.

chlee75:
I will reset commit and PR.

PR #4437, opened 2022-03-24
modem-manager: Add signed-payload metadata for some Quectel devices

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -196,6 +196,41 @@ validate_firmware_update_method(MMModemFirmwareUpdateMethod methods, GError **er
 	return FALSE;
 }
 
+static void
+fu_mm_device_probe_quectel_flags(FuMmDevice *self)
+{
+	const gchar *version = fu_device_get_version(FU_DEVICE(self));
+	g_autofree gchar *name = NULL;
+	struct {
+		const gchar *name;
+		const gchar *version;
+	} secboot[] = {{"EM05GF", "EM05GFAR07A07M1G_01.005.01.005"},
+		       {"EM05CE", "EM05CEFCR08A16M1G_LNV"},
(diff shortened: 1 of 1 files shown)


plugins/modem-manager/fu-mm-device.c
```diff
@@ -196,6 +196,41 @@ validate_firmware_update_method(MMModemFirmwareUpdateMethod methods, GError **er
...
 
+static void
+fu_mm_device_probe_quectel_flags(FuMmDevice *self)
+{
+	const gchar *version = fu_device_get_version(FU_DEVICE(self));
+	g_autofree gchar *name = NULL;
+	struct {
+		const gchar *name;
+		const gchar *version;
+	} secboot[] = {{"EM05GF", "EM05GFAR07A07M1G_01.005.01.005"},
```
aleksander0m, line 207: I'm sure you're not happy with having a hardcoded list; I guess there's no better way to handle this, right?

hughsie: Ya I'm not 100% happy, but apparently there's no way of asking the firmware "_hey are you signed_" -- at least not yet. Even some heuristic on the version string would be good, but alas.

superm1: Are those earlier versions preproduction presumably? Maybe it's TOO much detail?
Especially if downgrades aren't supported.

superm1: This has me questioning the efficacy of such an effort. If the boot rom isn't signed and the first payload you had in there wasn't signed how can you guarantee the safety of a payload that introduced signing?

Maybe that's too deep a question in root of trust and we should just be happy it's signed now and move on.

hughsie: The new firmware blows an eFuse!

Issue #4080, opened 2021-12-14
Contribution setup breaks Fedora 35 boot


DylanVanAssche:
**Describe the bug**

Last night I installed the contribution tools as mentioned in #4076 to fix some indentation in my patches on Fedora 35.
I didn't noticed anything until I powered on my desktop again this morning, Fedora couldn't get into the GNOME desktop environment anymore.

After some emergency mode and dnf history voodoo I figured out the reason: fwupd's contribution setup was causing this.
Steps I did to fix my desktop:

1. I removed all the installed packages with a DNF rollback. Didn't help as expected.
2. I listed the failed systemd units in emergency mode and the Kernel Load Modules one was failing so I had no kernel modules loaded. I removed the modules from the load list that where listed there as failed. No luck either.
3. I digged into the setup script and executed this part manually: https://github.com/fwupd/fwupd/blob/main/contrib/prepare-system#L9-L12 Since then I have a working Fedora 35 Workstation environment again.

**Steps to Reproduce**
1. Run Fedora 35 Workstation
2. Install the contribution scripts
3. Reboot
4. Party!

**Expected behavior**

Install normally without killing the desktop's boot process.

**fwupd version information**

1.7.3

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): git (as I was developing for this)

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

Not Applicable, it is about the fwupd contrib scripts

**Additional questions**

- Operating system and version: Fedora 35 Workstation
- Have you tried rebooting? Yes
- Is this a regression? No idea

hughsie:
I can't reproduce this here; could you try using `setup` again and deleting the files referenced in cleanup one at a time? Then we can work out which bit is causing the issues. Thanks.

hughsie:
Ohh, and in the general case you don't need to install those files -- I install in a prefix and just use that to avoid "polluting" my system versions of stuff.

DylanVanAssche:
This seems to be the issue:

```
● dbus-broker.service - D-Bus System Message Bus
     Loaded: loaded (/usr/lib/systemd/system/dbus-broker.service; enabled; vendor preset: enabled)
     Active: active (running) since Tue 2021-12-14 18:05:54 CET; 5min ago
TriggeredBy: ● dbus.socket
       Docs: man:dbus-broker-launch(1)
[...]

PR #4440, opened 2022-03-24
Detect SELinux before offering to install into /usr/local

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4080

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/setup
+++ b/contrib/setup
@@ -22,6 +22,16 @@ rename_branch()
     fi
 }
 
+detect_selinux()
+{
+    SELINUX="0"
+    if which getenforce &> /dev/null; then
+        if getenforce | grep -i "enforcing" >/dev/null; then
+            SELINUX="1"
+        fi
+    fi
+}
+
 setup_deps()
 {
     read -p "Install build dependencies? (y/N) " question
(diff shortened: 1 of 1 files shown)


contrib/setup
```diff
@@ -32,6 +44,9 @@ setup_deps()
 
 setup_run_dev()
 {
+    if [ "$SELINUX" -eq "1" ]; then
+        return
```
superm1, line 48: I would suggest still emitting a message here to let people know /usr/local isn't going to fly BECAUSE of selinux though

contrib/setup
```diff
@@ -22,6 +22,18 @@ rename_branch()
     fi
 }
 
+detect_selinux()
+{
+    SELINUX="0"
+    which getenforce
+    if [ "$?" -eq 0 ]; then
+        getenforce | grep -i "enforcing"
+        if [ "$?" -eq 0 ]; then
```
superm1, line 31: could this be simpler?

```
if getenforce | grep -i "enforcing" 2> /dev/null 1>&2; then
   SELINUX=1
fi
```

I think that if it doesn't exist that fails, and if it's not enforcing that fails, and if it is it passes

superm1 (dismissed):
as long as this works, LGTM now.

PR #4443, opened 2022-03-25
Modernize and refactor the requirement checking code

hughsie:
Before we had FuInstallTask which represented the specific install
action (install foo on bar), FuEngineRequest which described why the
install task was created, and FuEngine. The latter being both reposible
for calling *into* FuInstallTask and also *from* FuInstallTask as well.

Depending on the code, we had XbNodes of component and release, *and*
FwupdReleases as releases and lots of duplicated code to marshall the
former into various forms of the latter.

Create a FuRelease wrapper around FwupdRelease to replace FuInstallTask
and to simplify the code. Make the FuRelease builder do the device
requirement checks, and then use the engine to do checks requiring
engine state and context.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 this fixes a few problems were we suggesting releases which had a version number less than the `version_min` -- the code was all so complicated I ended up refactoring it into this. The original code has been hacked around so much I couldn't even follow the logic of what it was doing; it's much simpler now. Potential for regressions is large, and I'll go mad testing this on all the hardware I have next week -- but some early code review would be great. No panic.

hughsie:
Lets push this into main now (as it fixes a bug I'm sitting on) and leave 1_7_X with the old code.

Issue #4394, opened 2022-03-16
modem-manager: device removed or did not come back


DylanVanAssche:
**Describe the bug**

It seems that in some rare cases ModemManager --> fastboot hand-over seems to fail as seen by LVFS reports.
I have a hard time to reproduce this personally on my setups, but in the past I have seem some similar things and the modem appears though with `fastboot devices`. Error messages:

- `failed to get device before update: failed to wait for detach replug: device 976c4a39e87f61e6940ea6a8d39c583cfa99615f did not come back`
- `Device has been removed`

AFAIK, retrying the upgrade 'fixes' it.
Most reports come from Debian but I saw a few as well from postmarketOS.
I hope somebody else has an idea here what may go wrong in this case.

**Steps to Reproduce**
1. Upgrade a modem with fwupd through ModemManager + fastboot
2. Upgrade fails because the modem seems to be lost when handing over, the modem appears though with `fastboot devices`

**Expected behavior**
Upgrade just works

**fwupd version information**
Please provide the version of the daemon and client.

```shell
pine64-pinephone:/home/dylan# fwupdmgr --version
runtime   org.freedesktop.fwupd         1.7.6
compile   org.freedesktop.gusb          0.3.10
runtime   org.kernel                    5.16.4
compile   com.hughsie.libjcat           0.1.10
compile   org.freedesktop.fwupd         1.7.6
runtime   org.freedesktop.gusb          0.3.10
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
Pine64 PinePhone (1.2)
│
└─QUECTEL Mobile Broadband Module:
      Device ID:          976c4a39e87f61e6940ea6a8d39c583cfa99615f
      Summary:            Quectel EG25-G modem
      Current version:    0.5.7
      Vendor:             QUALCOMM INCORPORATED (USB:0x2C7C, USB:0x18D1)
      Release Branch:     community
      GUIDs:              db379a33-254f-5140-b37e-d36ae7e5c039
                          d18f31f1-a3fa-55a2-b4ed-decfbc1e004d ← USB\VID_2C7C&PID_0125&REV_0318&NAME_EG25GGB
                          1a2996cb-f86e-5583-a464-e1b96e1c6ae9 ← USB\VID_2C7C&PID_0125&REV_0318
[...]

Issue #4128, opened 2021-12-26
modem-manager: fwupd crashes when modem is reprobed by ModemManager


DylanVanAssche:
**Describe the bug**

In some rare cases, modems may disconnect from the system causing ModemManager to reprobe the modem and expose it under a new DBus object. When this happens when fwupd tries to detach the modem using the modem-manager plugin, fwupd crashes.
Fwupd never asks ModemManager to inhibit the modem, instead it crashes and is restarted by the supervisor or DBus.

I truncated the logs right before starting an upgrade so I don't have the full logs, the logs became too big otherwise to find this needle in the haystack.

**Logs for a crash**

*fwupd*

fwupd doesn't seem to noticed that the modem ID is here incremented to `1` as a port hangup happened.

```
10:36:50:0909 FuPlugin             detach(modem_manager)
10:36:50:0983 FuPluginMm           No response expected for AT command: 'AT+QFASTBOOT', assuming succeed
10:36:50:0985 FuMain               Emitting PropertyChanged('Status'='device-write')
10:36:50:0985 FuMain               Emitting PropertyChanged('Percentage'='2%')
10:36:50:0986 FuDeviceList         waiting 20000ms for replug
--------- fwupd segfaults and restarts ----------
10:36:51:0218 FuDebug              Verbose debugging enabled (on console 0)
10:36:51:0220 FuConfig             loading config values from /etc/fwupd/daemon.conf
10:36:51:0220 FuConfig             using autodetected max archive size 18.4 EB
10:36:51:0225 FuRemoteList         loading /usr/share/fwupd/metainfo
```

*ModemManager*

MM reprobed the modem when sending 'AT' and after response a port hangup happened.

```
[25098]: <debug> [1640169410.989355] [modem1/ttyUSB2/at] <-- '<CR><LF>OK<CR><LF>'
[25098]: <debug> [1640169411.072749] [modem1/ttyUSB2/at] unexpected port hangup!
[25098]: <debug> [1640169411.073094] [modem1/ttyUSB2/at] forced to close port
[25098]: <debug> [1640169411.073172] [modem1/ttyUSB2/at] device open count is 0 (close)
[25098]: <debug> [1640169411.073283] [modem1/ttyUSB2/at] closing serial port...
[25098]: <debug> [1640169411.073479] [modem1/ttyUSB2/at] serial port closed
[...]

PR #4453, opened 2022-03-28
modem-manager: Remove the cache item and use a shadow device instead

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4394

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -1777,6 +1777,20 @@ fu_mm_device_setup(FuDevice *device, GError **error)
 }
 
 static void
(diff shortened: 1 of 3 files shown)


DylanVanAssche:
Alright, it looks like this works. Tried a few times and it doesn't show up anymore.
UPower detection also properly works now because of the UPower type fixes in 1.8.0.

2 things, unrelated:
- I could reproduce #4128, with the same verbose logs, but without a backtrace :/ Each time I try to run it in gdb it doesn't happen of course...
- However, the modem does not come back because this commit was reverted: https://github.com/fwupd/fwupd/commit/7f6605cf8d9090b052be2f6d7d84fa4633e9087d (100% reproducibility)

PR #4464, opened 2022-03-31
Add the CFI quirks in setup() so the device can be open

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@adamgene can you just verify I've not broken anything for you please.

adamgene (approved):
It's normal as before.

PR #4465, opened 2022-03-31
Use the CFI manufacturer ID to set the vendor

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
100% untested, I'm on the wrong desk today.

adamgene (commented):
There is no error, and nothing changed.
I tested `[CFI\FLASHID_EF]` and it still reported `no CFI device found`, even I add `Name`.

PR #4467, opened 2022-03-31
Add a plugin for the CH341A device

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


gportay:
I will receive a new ch341a later today, and I will a try once I receive it!

Have you planned to merge it before the next release?

hughsie:
> Have you planned to merge it before the next release?

Maybe?! It's pretty low risk from a release point of view.

gportay:
Oh, the 3.3V regulator of my CH341A has dropped to 1.6~1.8V... it means I may probably reused it for 1.8V chips!

gportay:
Oh!

```
XPS 13 9370
│
├─CH341A:
│ │   Device ID:          e8055e4e2470749d3ebc7735868bf869c94ebd8f
│ │   Current version:    3.4
│ │   Vendor:             WinChipHead (USB:0x1A86)
│ │   GUIDs:              e06cabdb-2908-5d74-a63f-0daa8a1be58a ← USB\VID_1A86&PID_5512
│ │                       c78ab180-b9f0-5ec4-8ef4-44c99db4ffda ← USB\VID_1A86&PID_5512&REV_0304
│ │ 
│ └─MX25V2033F:
│       Device ID:        ad7678ada323cf0a1e6bbf03af92da8d9721556f
│       Summary:          CFI flash chip
│       Vendor:           Macronix (SPI:*)
│       GUID:             fdae557c-1f42-5e24-925f-4b6495c8b35c ← SPI
│       Device Flags:     • Updatable
│                         • Unsigned Payload
│     
```

gportay:
Got an unknow device for the M24fd.

```
│ └─Unknown Device:
│       Device ID:        ad7678ada323cf0a1e6bbf03af92da8d9721556f
│       Summary:          CFI flash chip
│       Vendor:           WinChipHead (SPI:*)
│       GUID:             fdae557c-1f42-5e24-925f-4b6495c8b35c ← SPI
│       Device Flags:     • Updatable
│                         • Unsigned Payload
```

Can we display the chip id that was read?

hughsie:
Yup, --verbose should show it.

gportay:
And I got garbage chip-id for the [GD25Q32C](https://www.gigadevice.com/flash-memory/gd25q32c/); and after a while (couple of seconds) I read ffffff as chip-id at every attempts.

I will have to find a solution to try to power only the flash chip (it that a matter of power supply).

However, I can still review the code :)

libfwupdplugin/fu-cfi-device.c
```diff
@@ -467,6 +467,64 @@ fu_cfi_device_to_string(FuDevice *device, guint idt, GString *str)
...
+ * @error: (nullable): optional return location for an error
+ *
+ * Sets the chip select value.
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 1.8.0
+ **/
+gboolean
+fu_cfi_device_chip_select(FuCfiDevice *self, gboolean value, GError **error)
```
gportay, line 483: Can we imagine that a programmer support multiple cs pins?

plugins/ch341a/fu-ch341a-cfi-device.c
```diff
@@ -0,0 +1,431 @@
...
+	cslocker = fu_cfi_device_chip_select_locker_new(FU_CFI_DEVICE(self), error);
+	if (cslocker == NULL)
+		return FALSE;
+
+	/* read JEDEC ID */
+	if (!fu_ch341a_device_spi_transfer(proxy, buf, sizeof(buf), error)) {
+		g_prefix_error(error, "failed to request JEDEC ID: ");
+		return FALSE;
+	}
+	if (buf[1] == 0xFF && buf[2] == 0xFF && buf[3] == 0xFF) {
```
gportay, line 105: I get 00:00:00, does it make sense to consider that value as a error?

plugins/ch341a/fu-ch341a-cfi-device.c
```diff
@@ -0,0 +1,431 @@
...
+		return FALSE;
+	}
+	if (!fu_ch341a_cfi_device_chip_erase(self, error)) {
+		g_prefix_error(error, "failed to erase: ");
+		return FALSE;
+	}
+	fu_progress_step_done(progress);
+
+	/* write each block */
+	pages = fu_chunk_array_new_from_bytes(fw, 0x0, 0x0, 0x100);
```
gportay, line 350: The `0x100` value is the page size? A transfert size value specific to ch341a?

gportay: Or to the CFI device?

plugins/ch341a/fu-ch341a-cfi-device.c
```diff
@@ -0,0 +1,431 @@
...
+						       error);
+	if (fw_verify == NULL) {
+		g_prefix_error(error, "failed to verify blocks: ");
+		return FALSE;
+	}
+	if (!fu_common_bytes_compare(fw, fw_verify, error))
+		return FALSE;
+	fu_progress_step_done(progress);
+
+	/* success! */
```
gportay, line 373: Ah ah I like the `!` ponctuation, because we are very happy to have succeed!

plugins/ch341a/fu-ch341a-cfi-device.c
```diff
@@ -0,0 +1,431 @@
...
+	gsize bufsz = fu_device_get_firmware_size_max(device);
+	g_autoptr(FuDeviceLocker) locker = NULL;
+
+	/* open programmer */
+	locker = fu_device_locker_new(proxy, error);
+	if (locker == NULL)
+		return NULL;
+
+	/* sanity check */
+	if (bufsz == 0x0) {
```
gportay, line 391: This is out of scope, but I find huge value in that plugin; especially for debugging from very special situations.

However, it is kind of limited because of the "API"; I really like to be able to read and write x bytes at address y.

But, this implies passing those extra parameters, via command line or environment variables and update/add vfunc.

gportay: For example: I would like to write a part of the blob somewhere in the flash without rewriting the whole flash. but this brings lots of complication. May the thing I would like is just to be able to set the start address.

libfwupdplugin/fu-cfi-device.c
hughsie: Yes, but in that case they'd be multiple FuCfiDevices.

plugins/ch341a/fu-ch341a-cfi-device.c
hughsie: Agreed, fixed.

plugins/ch341a/fu-ch341a-cfi-device.c
hughsie: Agreed, I need to use fu_cfi_device_get_page_size() -- fixed!

plugins/ch341a/fu-ch341a-cfi-device.c
hughsie: Reading at random offsets is easily possible (and we could certainly change the vfunc if required) but writing is much harder as you typically need to back up any of the affected sector (and possibly block...) then erase, write and verify each sector.

gportay: Exact; the user has to know what he is doing.

gportay:
> Since I fail to understand, is this a plugin for using the CH341a for reading/writing other devices, or for accessing the CH341a itself (as odd as that sounds)?

The first. This is a plugin to read/write blobs from a memory chip via connected to a CH341a device.

moriel5:
@gportay Thanks, that certainly clears things up.
That could also potentially make things easier to people who aren't well-versed in the commandline.

By the way, did you find a solution to the garbage data? Since I am having similar issues with my CH341a (mine doesn't even detect the chips anymore).

gportay:
Nop, I have no solution yet except to unsolder the chip.

moriel5:
Yikes, that certainly is a shame.
And since I am at this point dragging things to becoming unrelated, I'll just finish with saying that if I think of a way that can see this (and have tested it out), I'll post it.

PR #4470, opened 2022-03-31
fu-engine.c: do not inhibit on battery full but discharging

djcampello:
Change-Id: I45f71dbfc70a90aea2e33ec98fab1b247da3c714

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -290,8 +290,9 @@ fu_engine_ensure_device_battery_inhibit(FuEngine *self, FuDevice *device)
 		return;
 
 	if (fu_device_has_flag(device, FWUPD_DEVICE_FLAG_REQUIRE_AC) &&
-	    (fu_context_get_battery_state(self->ctx) == FU_BATTERY_STATE_DISCHARGING ||
-	     fu_context_get_battery_state(self->ctx) == FU_BATTERY_STATE_EMPTY)) {
+	    (fu_context_get_battery_state(self->ctx) == FU_BATTERY_STATE_EMPTY ||
+	     (fu_context_get_battery_state(self->ctx) == FU_BATTERY_STATE_DISCHARGING &&
+	      fu_context_get_battery_level(self->ctx) < 100))) {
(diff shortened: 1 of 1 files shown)


superm1:
> Strip out gerrit Change-IDs from commit message before upstream submission.

I don't have a strong feeling on this.  If it makes your backport / tagging process easier to get stuff you can leave them.  If you think they shouldn't be there, then perhaps we should have a pre-commit hook catch them in commit messages?

hughsie:
I'm not sure I care, if it's helpful for you it's fine for me. I'm not sure about the logic tho; this feels like something that belongs in the powerd plugin, no?

djcampello:
> > Strip out gerrit Change-IDs from commit message before upstream submission.
> 
> I don't have a strong feeling on this. If it makes your backport / tagging process easier to get stuff you can leave them. If you think they shouldn't be there, then perhaps we should have a pre-commit hook catch them in commit messages?

It makes life easier :)

> I'm not sure I care, if it's helpful for you it's fine for me. I'm not sure about the logic tho; this feels like something that belongs in the powerd plugin, no?

I think upower suffers of the same but I cannot confirm it. It seems that in some platforms when the battery is full, acpi? reports the state as discharging

djcampello:
I need to think about this a bit more.... I will follow up here...

djcampello:
If needed (we want to apply only to powerd) I can make the plugin set the battery state as FULL whenever it sees 100% battery regardless of the reported state. Let me know what you think

hughsie:
> make the plugin set the battery state as FULL whenever it sees 100% battery

This does make me uncomfortable. If it's a hardware bug it should be worked around in the kernel, and if it's a powerd bug then it should be fixed in powerd -- otherwise everything that consumes that property will need to be fixed -- possibly in multiple projects. i.e. it seems like a workaround that needs to be pushed down.

djcampello:
Agreed, closed this now and will look into a fix at a lower layer! Thanks for the review!

PR #4471, opened 2022-03-31
Add new plugin for AMD PSP

superm1:
This is the fwupd support to match the series posted at https://www.spinics.net/lists/linux-crypto/msg62965.html


src/fu-security-attr.c
```diff
@@ -125,14 +125,14 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MANUFACTURING_MODE) == 0) {
 		const gchar *kind = fwupd_security_attr_get_metadata(attr, "kind");
 		if (kind != NULL) {
 			/* TRANSLATORS: Title: %s is ME kind, e.g. CSME/TXT */
 			return g_strdup_printf(_("%s manufacturing mode"), kind);
 		}
-		/* TRANSLATORS: Title: MEI = Intel Management Engine */
-		return g_strdup(_("MEI manufacturing mode"));
+		/* TRANSLATORS: Title: A SOC mode with less permissive security */
+		return g_strdup(_("Manufacturing/Debug mode"));
```
hughsie, line 135: I'm not sure about this. I don't think we should combine things that are not 100% the same. I actually think it makes sense to have a `MEI manufacturing mode` on Intel that points to documentation about Intel AMT and CSME, and a `PSP Debug Mode` item on AMD that points to information about AMD PSP. I think it's fine to combine items like IOMMU where the implementation is basically the same, and the security problems with having it disabled are exactly the same between vendors - but I don't think we should collapse things that are actually quite different into one property.

plugins/pci-psp/fu-plugin-pci-psp.c
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	return TRUE;
+}
+
+static gboolean
+fu_plugin_pci_psp_get_attr(FuPlugin *plugin,
+			   FwupdSecurityAttr *attr,
+			   const gchar *file,
+			   gchar *out,
```
hughsie, line 59: it feels a bit messy to pass back a char of an int. Maybe do `fu_common_strtoull()`?

src/fu-util-common.c
```diff
@@ -2350,12 +2351,15 @@ fu_util_security_attrs_to_string(GPtrArray *attrs, FuSecurityAttrToStringFlags s
...
 				low_help = TRUE;
 
-			/* check for PCR0 not matching */
-			if (g_strcmp0(fwupd_security_attr_get_appstream_id(attr),
-				      FWUPD_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0) == 0 &&
-			    fwupd_security_attr_get_result(attr) ==
-				FWUPD_SECURITY_ATTR_RESULT_NOT_VALID)
-				pcr0_help = TRUE;
+			if (fwupd_security_attr_get_result(attr) ==
+			    FWUPD_SECURITY_ATTR_RESULT_NOT_VALID) {
```
hughsie, line 2355: it almost feels like we need a result of `MISSING_DATA` that plugins can emit -- that results in the HSI result being invalid because some required data is not present.

plugins/msr/fu-plugin-msr.c
```diff
@@ -370,6 +370,7 @@ fu_plugin_add_security_attr_amd_sme_enabled(FuPlugin *plugin, FuSecurityAttrs *a
 	/* success */
 	fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
 	fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_ENCRYPTED);
+	fwupd_security_attr_add_obsolete(attr, "pci_psp");
```
kerneis-anssi, line 373: I'm not familiar with fwupd_security_attr_add_obsolete so this is merely a request for clarification:
- what happens in practice, is there a race depending on which module is loaded first since they obsolete each other?
- since you mention in the commit message it's not intuitive, worth adding a comment here (and in the other file) to explain what is going on?

kerneis-anssi (commented):
github doesn't allow comments on commit messages, so here we go:

- typo: exporing -> exporting in the last commit

src/fu-security-attr.c
superm1: My goal is to avoid too many attributes that are manufacturer specific concepts.

Technically the PSP implementation is tying two things in:

1. part is fused
2. part is locked from debug

I am tempted to still combine the first with a generic "manufacturing mode" and make a new option for debug locked.

plugins/pci-psp/fu-plugin-pci-psp.c
superm1: the kernel only exports 0/1.  Maybe just pass back a bool.

src/fu-util-common.c
superm1: If we go with the `MISSING_DATA` idea we can also add in a plugin for other unsupported architectures to just emit missing data too.
This makes me tempted then to make `fwupdmgr security` not require --force unless no `MISSING_DATA`.

docs/hsi.md
```diff
@@ -657,6 +657,25 @@ To meet HSI-1 on systems that run this test, the result must be `enabled`. *[v1.
 
 <a id="org.fwupd.hsi.RollbackProtection"></a>
 
+### [SPI Rom protections](#org.fwupd.hsi.SpiProtection)
+
+SOCs may enforce control of the SPI bus to prevent writes other than by verified entities. Also SOCs may include support for replay protected monotonic counters to prevent replay attacks.
+
+**Impact:** SOCs without this feature may be attacked by an attacker modifying the SPI during unvalidated environments.
```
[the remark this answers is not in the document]
superm1: Unvalidated environments is referring to the fact that at least some of the SPI protections need to be inactive at some time to be able to flash an update.

plugins/msr/fu-plugin-msr.c
superm1: Just to preface - I still don't like how this works right now, but don't have a better solution.  Hoping we can come up with one in this PR.  You guys don't have hardware that can see this yet, so let me show you what it looks like.

If TSME and SME are disabled you'll see this:
```
HSI-4
✘ Encrypted RAM:                 Invalid
✘ Encrypted RAM:                 Failed
```

With obsoletes If TSME or SME is enabled you see:
```
HSI-4
✔  Encrypted RAM:                 Encrypted
```

If you don't have obsoletes and only one of those is set (what I think is likely to be the more common case for both Ryzen Pro and Epyc) you'll see something like this:
```
HSI-4
✔  Encrypted RAM:                 Encrypted
✘ Encrypted RAM:                 Failed
```

So what happens is if one is set as success then the other doesn't show up.  If both are fail (for whatever reason) then both show up.  We could potentially approach it as dedicated attributes for SME and TSME, but I'm not sure that makes sense either.

docs/hsi.md
```diff
@@ -657,6 +657,25 @@ To meet HSI-1 on systems that run this test, the result must be `enabled`. *[v1.
 
 <a id="org.fwupd.hsi.RollbackProtection"></a>
 
+### [SPI Rom protections](#org.fwupd.hsi.SpiProtection)
+
+SOCs may enforce control of the SPI bus to prevent writes other than by verified entities. Also SOCs may include support for replay protected monotonic counters to prevent replay attacks.
```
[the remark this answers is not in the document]
superm1: Yeah; it's a SPI controller and SPI flash hardware implementation of a rollback protection.  It's different than PSP rollback protection which compares versions in the signed payloads and doesn't let you downgrade..

Maybe what we want is two classes of SPI protections then actually.  My proposal would be
"SPI rollback protection" and "SPI write protection".

src/fu-security-attr.c
hughsie: > avoid too many attributes that are manufacturer specific

I think I'd rather have dozens of vendor specific attributes rather than trying to squash different things together and trying to find a suitable HSI level integer for the superset.

plugins/pci-psp/fu-plugin-pci-psp.c
hughsie: WFM.

src/fu-util-common.c
hughsie: Can you split this bit out as a new PR and branch and I'll have a play next week.

src/fu-security-attr.c
superm1: Adjusted

docs/hsi.md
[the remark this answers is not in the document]
superm1: adjusted

src/fu-util-common.c
superm1: https://github.com/fwupd/fwupd/pull/4481

docs/hsi.md
```diff
@@ -372,21 +372,39 @@ To meet HSI-1 on systems that run this test, the result must be `not-enabled`. *
 - [Chipsec 4xxlp register definitions](https://github.com/chipsec/chipsec/blob/master/chipsec/cfg/8086/pch_4xxlp.xml#L270)
 - [RISC-V EDK PCH register definitions](https://github.com/riscv/riscv-edk2-platforms/blob/85a50de1b459d1d6644a402081120770aa6dd8c7/Silicon/Intel/CoffeelakeSiliconPkg/Pch/Include/Register/PchRegsDci.h)
 
+<a id="org.fwupd.hsi.Fused"></a>
+
+### [Part is fused](#org.fwupd.hsi.Fused)
```
hughsie, line 377: I'm wondering if this should be org.fwupd.hsi.CpuFused instead

plugins/pci-psp/fu-plugin-pci-psp.c
```diff
@@ -0,0 +1,224 @@
...
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_MISSING_DATA);
+		return FALSE;
+	}
+	*out = fu_common_strtoull(buf) ? TRUE : FALSE;
+
+	return TRUE;
+}
+#define simple_security_attr(attr_name, level, name, sysfs, fail, success, OBS)                    \
+	static void fu_plugin_pci_psp_security_attr_##name(FuPlugin *plugin,                       \
+							   FuSecurityAttrs *attrs)                 \
```
hughsie, line 68: not really a fan of mega macros like this

docs/hsi.md
superm1: It's not a property of individual CPU cores or even chiplets. How about FusedSoc? Or Fused Part?

plugins/pci-psp/fu-plugin-pci-psp.c
superm1: Yeah I left it split out because I wasn't sure either on it. I'll take it out.

docs/hsi.md
hughsie: What about org.fwupd.hsi.PlatformFused?

superm1: That's good, will swap.

Also what do you think about me splitting this PR into 2 - all the new attributes and the new plug-in?

I just thought it made sense to see them in context is why it's all in one.

hughsie: Yup, definitely a split; leave the new plugin here and split out the other stuff we want in tree ASAP.

plugins/pci-psp/fu-plugin-pci-psp.c
```diff
@@ -0,0 +1,298 @@
...
+{
+	FuPluginData *priv = fu_plugin_get_data(plugin);
+
+	/* interesting device? */
+	if (!FU_IS_UDEV_DEVICE(device))
+		return TRUE;
+	if (g_strcmp0(fu_udev_device_get_subsystem(FU_UDEV_DEVICE(device)), "pci") != 0)
+		return TRUE;
+
+	priv->sysfs_path = g_strdup(fu_udev_device_get_sysfs_path(FU_UDEV_DEVICE(device)));
```
hughsie, line 40: I wonder if you'd rather store a ref to the FuDevice instead of just the sysfs_path?

superm1:
> I wonder if you'd rather store a ref to the FuDevice instead of just the sysfs_path? 

Sure; adjusted to use a cached device instesad for the attribute.

hughsie (approved):

PR #4472, opened 2022-04-01
trivial: fix behavior when running on old json-glib

superm1:
older json-glib versions can't work with saving or retrieving
HSI security events.  Rather than showing non-sensical errors
about too old of a json glib version that are out of context,
just compile this stuff out on the older versions.

This fixes HSI security events showing up too as a result.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


src/fu-engine.c
```diff
@@ -6054,6 +6056,7 @@ fu_engine_get_host_security_events(FuEngine *self, guint limit, GError **error)
 
 	g_return_val_if_fail(FU_IS_ENGINE(self), NULL);
 
+#if JSON_CHECK_VERSION(1, 6, 0)
 	attrs_array = fu_history_get_security_attrs(self->history, limit, error);
```
hughsie, line 6060: I think you need to ifdef the

	g_autoptr(GPtrArray) attrs_array = NULL;

otherwise you'll get a clang -Wunused warning. Then LGTM, thanks.

hughsie (approved):

PR #4481, opened 2022-04-01
Add support for mandatory attributes

superm1:
This will change HSI to be supported on Intel and AMD by default, but only if enough plugins advertise data.
When plugins for other architectures are they can be removed from the dummy attribute in the cpu plugin.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


libfwupd/fwupd-security-attr.h
```diff
@@ -109,6 +110,7 @@ typedef enum {
 	FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND,	  /* Since: 1.5.0 */
 	FWUPD_SECURITY_ATTR_RESULT_SUPPORTED,	  /* Since: 1.5.0 */
 	FWUPD_SECURITY_ATTR_RESULT_NOT_SUPPORTED, /* Since: 1.5.0 */
+	FWUPD_SECURITY_ATTR_RESULT_MISSING_DATA,  /* Since: 1.8.0 */
```
hughsie, line 113: Maybe this is a *flag* more than a result; it's not saying "this is the answer" it's saying "not enough data to get an answer"

superm1: updated

src/fu-security-attr.c
```diff
@@ -179,6 +179,10 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
 		 * debugging of Intel processors using the USB3 port */
 		return g_strdup(_("Intel DCI debugger"));
 	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUPPORTED_SOC) == 0) {
```
hughsie, line 182: _CPU no?

src/fu-security-attr.c
```diff
@@ -179,6 +179,10 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		/* TRANSLATORS: Title: if fwupd supports HSI on this chip */
+		return g_strdup(_("Supported SoC"));
```
hughsie, line 184: I think most people know what a CPU is, and soc is probably much less understood.

src/fu-tool.c
```diff
@@ -2815,43 +2815,50 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
...
+
+		if (!fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_MISSING_DATA))
+			continue;
+		if (priv->flags & FWUPD_INSTALL_FLAG_FORCE)
+			continue;
+		g_set_error_literal(
+		    error,
+		    FWUPD_ERROR,
+		    FWUPD_ERROR_NOT_SUPPORTED,
+		    "Not enough data was provided by plugins to make an HSI calculation. "
```
hughsie, line 2843: Maybe remove "by plugins"

src/fu-security-attr.c
superm1: Hmm, yeah I guess this is the CPU plugin, so it should be _CPU, you're right.

src/fu-tool.c

hughsie (approved):

PR #4483, opened 2022-04-01
genesys: usbhub: Fix test for static_tool_string value

gportay:
Additionally, this regroups the call of a function and the test for the
returned value.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

This fixes a minor issue that has escaped from my review :facepalm:.

### DIFF
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -832,7 +832,6 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 	}
 	memcpy(rev, &self->static_ts.mask_project_ic_type[4], 2);
 	self->chip.revision = fu_common_strtoull(rev);
-	tool_string_version = fu_genesys_tsdigit_value(self->static_ts.tool_string_version);
 
 	dynamic_buf =
 	    g_usb_device_get_string_descriptor_bytes_full(usb_device,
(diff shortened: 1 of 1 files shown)


gportay:
@adamgene could you confirm this is a fix?

hughsie (approved):

plugins/genesys/fu-genesys-usbhub-device.c
```diff
@@ -954,7 +954,7 @@ fu_genesys_usbhub_device_setup(FuDevice *device, GError **error)
 				    self->dynamic_ts.bonding);
 			return FALSE;
 		}
-		if (self->static_ts.tool_string_version < TOOL_STRING_VERSION_BONDING_QC)
```
gportay, line 957: That condition is never met as `self->static_ts.tool_string_version` is at least `0x30` (aka `'0'`) and `TOOL_STRING_VERSION_BONDING_QC` is `2`.

gportay:
@hughsie this can be merged, I have tested it and got the value I expect with this change.

plugins/genesys/fu-genesys-usbhub-device.c
adamgene: Yes, you are right.

gportay: cool, thanks.

PR #4486, opened 2022-04-02
Add support for ACPI IVRS table for HSI security attributes

superm1:
Type of pull request:

- [x] New plugin

### DIFF
--- a/docs/hsi.md
+++ b/docs/hsi.md
@@ -437,14 +437,16 @@ To meet HSI-2 on systems that run this test, the result must be `valid`. *[v1.5.
 - [Linux Kernel TPM Documentation](https://www.kernel.org/doc/html/latest/security/tpm/tpm_event_log.html)
 
 <a id="org.fwupd.hsi.AcpiDmar"></a>
+<a id="org.fwupd.hsi.PrebootDma"></a>
 
-### [Pre-boot DMA protection](#org.fwupd.hsi.AcpiDmar)
+### [Pre-boot DMA protection](#org.fwupd.hsi.PrebootDma)
 
 The IOMMU on modern systems is used to mitigate against DMA attacks.
 All I/O for devices capable of DMA is mapped into a private virtual memory region.
-The ACPI DMAR table is used to set up pre-boot DMA protection which eliminates some firmware attacks.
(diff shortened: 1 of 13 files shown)


docs/hsi.md
```diff
@@ -436,15 +436,16 @@ To meet HSI-2 on systems that run this test, the result must be `valid`. *[v1.5.
 
 - [Linux Kernel TPM Documentation](https://www.kernel.org/doc/html/latest/security/tpm/tpm_event_log.html)
 
-<a id="org.fwupd.hsi.AcpiDmar"></a>
+<a id="org.fwupd.hsi.PrebootDma"></a>
```
hughsie, line 439: I think that we need to have both, i.e. two <a> sections with the old name too.

docs/hsi.md
```diff
@@ -454,11 +455,12 @@ This would allow subverting the Secure Boot protection, and also invalidate any
 - `not-valid`: could not determine state
 - `not-enabled`: was not enabled
 
-To meet HSI-3 on systems that run this test, the result must be `enabled`. *[v1.5.0]*
+To meet HSI-3 on systems that run this test, the result must be `enabled`. *[v1.8.0]*
```
hughsie, line 458: I think we need to capture that the attribute existed in 1.5.0 for Intel and 1.8.0 for AMD.

libfwupd/fwupd-security-attr-private.h
```diff
@@ -14,13 +14,13 @@
 G_BEGIN_DECLS
 
 /**
- * FWUPD_SECURITY_ATTR_ID_ACPI_DMAR:
+ * FWUPD_SECURITY_ATTR_ID_PREBOOT_DMA_PROTECTION:
  *
  * Host Security ID attribute for ACPI DMAR table
  *
```
hughsie, line 20: The old name for this was org.fwupd.hsi.AcpiDmar etc.

superm1: fixed

docs/hsi.md
superm1: fixed

docs/hsi.md
superm1: fixed

hughsie (approved):

PR #4487, opened 2022-04-03
Add new security attributes

superm1:
These are to be used by https://github.com/fwupd/fwupd/pull/4471

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/docs/hsi.md
+++ b/docs/hsi.md
@@ -351,8 +351,9 @@ To meet HSI-1 on systems that run this test, the result must be `valid`. *[v1.5.
 - [Intel CSME Security Review Cumulative Update](https://www.intel.com/content/www/us/en/security-center/advisory/intel-sa-00086.html)
 
 <a id="org.fwupd.hsi.IntelDci.Enabled"></a>
+<a id="org.fwupd.hsi.PlatformDebugEnabled"></a>
 
-### [Intel DCI](#org.fwupd.hsi.IntelDci.Enabled)
+### [Intel DCI](#org.fwupd.hsi.PlatformDebugEnabled)
 
 Newer Intel CPUs support debugging over USB3 via a proprietary Direct Connection Interface (DCI) with the use of off-the-shelf hardware.
 
@@ -366,6 +367,8 @@ This makes privilege escalation from user to root possible, and also modifying S
 
 To meet HSI-1 on systems that run this test, the result must be `not-enabled`. *[v1.5.0]*
 
+Note: this attribute was previously known as org.fwupd.hsi.IntelDci.Enabled in 1.5.0, but was renamed in 1.8.0 to support other vendors.
+
 **References:**
 
 - [Intel Direct Connect Interface](https://www.intel.co.uk/content/www/uk/en/support/articles/000029393/processors.html)
@@ -373,20 +376,23 @@ To meet HSI-1 on systems that run this test, the result must be `not-enabled`. *
 - [RISC-V EDK PCH register definitions](https://github.com/riscv/riscv-edk2-platforms/blob/85a50de1b459d1d6644a402081120770aa6dd8c7/Silicon/Intel/CoffeelakeSiliconPkg/Pch/Include/Register/PchRegsDci.h)
 
 <a id="org.fwupd.hsi.IntelDci.Locked"></a>
+<a id="org.fwupd.hsi.PlatformDebugLocked"></a>
 
-### [Intel DCI](#org.fwupd.hsi.IntelDci.Locked)
+### [Part is debug locked](#org.fwupd.hsi.PlatformDebugLocked)
 
-Newer Intel CPUs support debugging over USB3 via a proprietary Direct Connection Interface (DCI) with the use of off-the-shelf hardware.
+Some devices support a concept of whether a part has been unlocked for debugging using proprietary hardware.  Such parts allow access to registers that are typically restricted when parts are fused.
+On Intel systems access to this interface is done via a proprietary Direct Connection Interface (DCI).
 
-**Impact:** A local attacker with root access would be able to enable DCI. This would allow them full access to all registers and memory in the system, and is able to make changes.
-This allows using SMM to write to system firmware for a persistent backdoor.
+**Impact:** If using a debug unlocked part, the platform's overall security will be decreased as an attacker may have elevated access to registers and memory within the system and can potentially enable persistent backdoors.
 
 **Possible results:**
 
-- `locked`: CPU debugging has been disabled
-- `not-locked`: is is still possible to enable CPU debugging
+- `locked`: device is locked
+- `not-locked`: device is not not locked
 
-To meet HSI-2 on systems that run this test, the result must be `locked`. *[v1.5.0]*
+To meet HSI-2 on systems that run this test, the result must be `locked`. *[v1.8.0]*
+
+Note: this attribute was previously known as org.fwupd.hsi.IntelDci.Locked in 1.5.0, but was renamed in 1.8.0 to support other vendors.
 
 **References:**
 
--- a/docs/hsi.md
+++ b/docs/hsi.md
@@ -661,6 +661,36 @@ To meet HSI-1 on AMD systems that run this test, the result must be `enabled`. *
 
 <a id="org.fwupd.hsi.IntelSmap"></a>
 
+### [AMD SPI Write protections](#org.fwupd.hsi.Amd.SpiWriteProtection)
+
+SOCs may enforce control of the SPI bus to prevent writes other than by verified entities.
+
+**Impact:** SOCs without this feature may be attacked by an attacker modifying the SPI.
+
+**Possible results:**
+
+- `enabled`: spi protections enabled
+- `not-enabled`: spi protections disabled
+
+To meet HSI-2 on systems that run this test, the result must be `enabled`. *[v1.8.0]*
+
+<a id="org.fwupd.hsi.Amd.SpiWriteProtection"></a>
+
+### [AMD SPI Replay protections](#org.fwupd.hsi.Amd.SpiReplayProtection)
+
+SOCs may include support for replay-protected monotonic counters to prevent replay attacks.
+
+**Impact:** SOCs without this feature may be attacked by an attacker modifying the SPI.
+
+**Possible results:**
+
+- `enabled`: spi protections enabled
+- `not-enabled`: spi protections disabled
+
+To meet HSI-3 on systems that run this test, the result must be `enabled`. *[v1.8.0]*
+
+<a id="org.fwupd.hsi.Amd.SpiReplayProtection"></a>
+
 ### [Supervisor Mode Access Prevention](#org.fwupd.hsi.IntelSmap)
 
 Without Supervisor Mode Access Prevention, the supervisor code usually has full read and write access to user-space memory mappings.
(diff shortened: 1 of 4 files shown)


libfwupd/fwupd-security-attr-private.h
```diff
@@ -265,13 +265,25 @@ G_BEGIN_DECLS
  **/
 #define FWUPD_SECURITY_ATTR_ID_INTEL_DCI_ENABLED "org.fwupd.hsi.IntelDci.Enabled"
 /**
- * FWUPD_SECURITY_ATTR_ID_INTEL_DCI_LOCKED:
+ * FWUPD_SECURITY_ATTR_ID_FUSED:
```
hughsie, line 268: Platform_fused, no?

docs/hsi.md
```diff
@@ -639,8 +657,58 @@ To meet HSI-4 on systems that run this test, the result must be `enabled`. *[v1.
 - [Intel TME Press Release](https://software.intel.com/content/www/us/en/develop/blogs/intel-releases-new-technology-specification-for-memory-encryption.html)
 - [WikiChip SME Overview](https://en.wikichip.org/wiki/x86/sme)
 
+<a id="org.fwupd.hsi.RollbackProtection"></a>
+
+### [Rollback protection](#org.fwupd.hsi.RollbackProtection)
```
hughsie, line 662: Roll back of *which* firmware? Maybe it needs a "Platform" in there somewhere.

src/fu-security-attr.c
```diff
@@ -174,15 +174,30 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
 	}
 	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_DCI_ENABLED) == 0 ||
-	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_DCI_LOCKED) == 0) {
-		/* TRANSLATORS: Title: Direct Connect Interface (DCI) allows
-		 * debugging of Intel processors using the USB3 port */
-		return g_strdup(_("Intel DCI debugger"));
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_DEBUG_LOCKED) == 0) {
+		/* TRANSLATORS: Title: Allows debugging of parts using proprietary hardware */
+		return g_strdup(_("Debugging"));
 	}
```
hughsie, line 180: Platform Debugging? Or CPU Debugging? The sole word is a bit generic...

superm1: It's platform. Very rarely would you be debugging CPU cores directly. Will update.

hughsie: Isn't DCI exactly that? i.e. CPU-level debugging?

superm1: IMO CPU debugging is a side effect; aren't you getting access to platform registers from it and that's usually where the focus is?

superm1: Taking the Intel view - DCI is CPU debugging **and** PCH debugging.  So I think "Platform" debugging is appropriate for that as well as AMD.  I feel it could in the future cover things like JTAG on ARM too.

hughsie: I think it makes sense if DCI, the AMD on and JTAG are all the same HSI level -- which maybe they are?

docs/hsi.md
```diff
@@ -372,21 +372,39 @@ To meet HSI-1 on systems that run this test, the result must be `not-enabled`. *
...
+To meet HSI-1 on systems that run this test, the result must be `locked`. *[v1.8.0]*
+
 <a id="org.fwupd.hsi.IntelDci.Locked"></a>
+<a id="org.fwupd.hsi.PlatformDebugLocked"></a>
 
-### [Intel DCI](#org.fwupd.hsi.IntelDci.Locked)
+### [Part is debug locked](#org.fwupd.hsi.PlatformDebugLocked)
 
-Newer Intel CPUs support debugging over USB3 via a proprietary Direct Connection Interface (DCI) with the use of off-the-shelf hardware.
+Some devices support a concept of whether a part has been unlocked for debugging using proprietary hardware.  Such parts allow access to registers that are typically restricted when parts are fused.
```
hughsie, line 396: I do think we need to refer to things like DCI in these descriptions; if the ODM gets a `PlatformDebugLocked` failure they need to be able to correlate this back to something like `DCI` so they know what needs fixing. I think you can also replace *ODM* with *Security Researcher` too.

libfwupd/fwupd-security-attr-private.h
```diff
@@ -265,13 +264,24 @@ G_BEGIN_DECLS
...
+ * Since: 1.8.0
+ **/
+#define FWUPD_SECURITY_ATTR_ID_PLATFORM_FUSED "org.fwupd.hsi.PlatformFused"
+/**
+ * FWUPD_SECURITY_ATTR_ID_PLATFORM_DEBUG_LOCKED:
+ *
+ * Host Security ID attribute for parts locked from debugging
+ *
+ * This was previously known as org.fwupd.hsi.IntelDci.Locked for Intel 1.5.0+
+ * It was renamed for all vendor support in 1.8.0.
```
hughsie, line 280: We've also got org.fwupd.hsi.IntelDci.Enabled as well as Locked; I'm really questioning whether we just want an org.fwupd.hsi.AmdPlatformDebug.Enabled -- I think it would be a lot more descriptive if we can link to per-vendor terms like DCI and JTAG.

superm1: I think in this case the abstraction of details is better in documentation than in strings a user sees.

If I was an end user do I care if it's AMD Platform Debugging locked or Intel DCI locked? Probably not. I just care it's not locked down.

src/fu-security-attr.c
superm1: yeah they are IMO.  My take is we should combine this attribute for Intel and AMD and then if ARM ends up being different we cross that bridge at that time.

docs/hsi.md
superm1: I've modified it, can you take another look?

libfwupd/fwupd-security-attr-private.h
hughsie: I think it matters what is actually debug locked. I think it makes sense to rename:

 * org.fwupd.hsi.IntelDci.Enabled -> org.fwupd.hsi.CpuDebug.Enabled 
 * org.fwupd.hsi.IntelDci.Locked -> org.fwupd.hsi.CpuDebug.Locked

And then define new attributes:

 * org.fwupd.hsi.PlatformDebug.Enabled 
 * org.fwupd.hsi.PlatformDebug.Locked

I think trying to squash both into the same attribute is a bit of a pain; in essence they're the same, but it makes it really complicated to explain.

libfwupd/fwupd-security-attr-private.h
```diff
@@ -287,6 +298,31 @@ G_BEGIN_DECLS
  * Since: 1.8.0
  **/
 #define FWUPD_SECURITY_ATTR_ID_SUPPORTED_CPU "org.fwupd.hsi.SupportedCpu"
+/**
+ * FWUPD_SECURITY_ATTR_ID_PLATFORM_ROLLBACK_PROTECTION
+ *
+ * Host Security ID attribute for Rollback protection of platform firmware
```
hughsie, line 304: I think we need to define what rollback is. i.e. downgrading. Maybe we want FWUPD_SECURITY_ATTR_ID_SPI_DOWNGRADE_PROTECTION instead?

src/fu-security-attr.c
hughsie: > DCI is CPU debugging and PCH debugging

Really? I've not seen people debug the PCH using DCI, although I've not really played with it myself.

superm1: https://conference.hitb.org/hitbsecconf2017ams/materials/D2T4%20-%20Maxim%20Goryachy%20and%20Mark%20Ermalov%20-%20Intel%20DCI%20Secrets.pdf shows that you can get at everything.  It's not just CPUs.

libfwupd/fwupd-security-attr-private.h
superm1: I'm not convinced.  See the link from my other comment.

libfwupd/fwupd-security-attr-private.h
superm1: That seems _too_ specific to me for a generic security attribute.  What if the rollback protection is at a layer that is above the payload going to the SPI chip?  Or what if it's not?  I think the attribute should be generic enough to say "this vendor has a feature for rollback back protection that would prevent installing older updates" and leave implementation details to the vendor how they do it.

src/fu-security-attr.c
hughsie: Okay, agreed, thanks for the link.

libfwupd/fwupd-security-attr-private.h
hughsie: Okay, I think we need to do org.fwupd.hsi.IntelDci.Enabled -> org.fwupd.hsi.Platform.Enabled still tho.

src/fu-security-attr.c
superm1: OK, in that case I'm going to rename the other DCI attribute as well to match and I'll drop a link to this in the description too.

libfwupd/fwupd-security-attr-private.h
hughsie: Then I think we need to clarify what the difference between downgrade and rollback is, as I'm not exactly sure myself.

superm1: I can't add more details about the implementation used.  If having a generic attribute for rollback protection is insufficient, then I would rather just not populate it.

hughsie: > I can't add more details about the implementation used

I don't think it makes sense to add it to the spec if we can't describe it in a fair amount of detail. If it's a generic attribute then it needs enough detail to be able to say "does RISC-V do the same thing".

superm1: OK - would you rather an AMD specific attribute then or no attribute at all?

hughsie: The former please; we can always make it generic in the future like we did with DCI, but I don't think we can ever do the opposite.

superm1: OK; everything updated up for these comments.

docs/hsi.md
```diff
@@ -639,8 +661,58 @@ To meet HSI-4 on systems that run this test, the result must be `enabled`. *[v1.
 - [Intel TME Press Release](https://software.intel.com/content/www/us/en/develop/blogs/intel-releases-new-technology-specification-for-memory-encryption.html)
 - [WikiChip SME Overview](https://en.wikichip.org/wiki/x86/sme)
 
+<a id="org.fwupd.hsi.PlatformRollbackProtection"></a>
```
hughsie, line 664: I think you need an `Amd.` prefix here, no?

docs/hsi.md
```diff
@@ -639,8 +661,58 @@ To meet HSI-4 on systems that run this test, the result must be `enabled`. *[v1.
...
+
+To meet HSI-1 on AMD systems that run this test, the result must be `enabled`. *[v1.8.0]*
+
+**References:**
+
+- [Rollback protection](https://www.psacertified.org/blog/anti-rollback-explained/)
+
 <a id="org.fwupd.hsi.IntelSmap"></a>
 
+### [SPI Write protections](#org.fwupd.hsi.SpiWriteProtection)
```
hughsie, line 686: Is this different to org.fwupd.hsi.Spi.Bioswe from 40,000ft? If so, Maybe we could add the `Amd` prefix to this and SpiWriteProtection and then we can always merge them to be generic in the future. If you do that, the PR LGTM.

PR #4491, opened 2022-04-04
trivial: scsi: fix a segfault with USB SCSI device plugged in

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/scsi/fu-scsi-device.c
+++ b/plugins/scsi/fu-scsi-device.c
@@ -60,13 +60,12 @@ fu_scsi_device_probe(FuDevice *device, GError **error)
 	fu_device_add_instance_strsafe(device, "VEN", vendor);
 	fu_device_add_instance_strsafe(device, "DEV", name);
 	fu_device_add_instance_strsafe(device, "REV", version);
-	if (!fu_device_build_instance_id_quirk(device, NULL, "SCSI", "VEN", NULL))
+	if (!fu_device_build_instance_id_quirk(device, error, "SCSI", "VEN", NULL))
 		return FALSE;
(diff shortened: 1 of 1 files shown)


plugins/scsi/fu-scsi-device.c
```diff
@@ -60,13 +60,27 @@ fu_scsi_device_probe(FuDevice *device, GError **error)
 	fu_device_add_instance_strsafe(device, "VEN", vendor);
 	fu_device_add_instance_strsafe(device, "DEV", name);
 	fu_device_add_instance_strsafe(device, "REV", version);
-	if (!fu_device_build_instance_id_quirk(device, NULL, "SCSI", "VEN", NULL))
+	if (!fu_device_build_instance_id_quirk(device, NULL, "SCSI", "VEN", NULL)) {
```
hughsie, line 63: Can we just do `fu_device_build_instance_id_quirk(device, error, "SCSI", "VEN", NULL)`?

hughsie (approved):

plugins/scsi/fu-scsi-device.c
superm1: It's noisier, but yeah that works.

superm1:
This shows up in red:
> 18:45:19:0731 FuEngine             failed to add device /sys/devices/pci0000:00/0000:00:08.3/0000:64:00.4/usb9/9-1/9-1.3/9-1.3:1.0/host0/target0:0:0/0:0:0:0: no value for DEV

But it doesn't crash, so better to me.

PR #4493, opened 2022-04-05
Fix bash completion scripts

gportay:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Very trivial changes on bash-completion.

### DIFF
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -104,11 +104,12 @@ _show_modifiers()
 
 _fwupdtool()
 {
-	local cur prev command
+	local cur prev command args
 	COMPREPLY=()
 	cur=`_get_cword`
 	prev=${COMP_WORDS[COMP_CWORD-1]}
 	command=${COMP_WORDS[1]}
+	_count_args
 
 	case $prev in
 	--plugins)
@@ -124,83 +125,68 @@ _fwupdtool()
 	case $command in
 	get-details|install|install-blob|firmware-dump)
 		#find files
-		if [[ "$prev" = "$command" ]]; then
+		if [[ "$args" = "2" ]]; then
 			_filedir
 		#device ID
-		elif [[ "$prev" = "${COMP_WORDS[2]}" ]]; then
+		elif [[ "$args" = "3" ]]; then
 			_show_device_ids
-		#modifiers
-		else
-			_show_modifiers
 		fi
 		;;
 	attach|detach|activate|verify-update|reinstall|get-updates)
 		#device ID
-		if [[ "$prev" = "$command" ]]; then
+		if [[ "$args" = "2" ]]; then
 			_show_device_ids
-		#modifiers
-		else
-			_show_modifiers
 		fi
 		;;
 	build-firmware)
 		#file in
-		if [[ "$prev" = "$command" ]]; then
+		if [[ "$args" = "2" ]]; then
 			_filedir
 		#file out
-		elif [[ "$prev" = "${COMP_WORDS[2]}" ]]; then
+		elif [[ "$args" = "3" ]]; then
 			_filedir
 		#script
-		elif [[ "$prev" = "${COMP_WORDS[3]}" ]]; then
+		elif [[ "$args" = "4" ]]; then
 			_filedir
 		#output
-		elif [[ "$prev" = "${COMP_WORDS[4]}" ]]; then
+		elif [[ "$args" = "5" ]]; then
 			_filedir
-		#modifiers
-		else
-			_show_modifiers
 		fi
 		;;
 	firmware-parse|firmware-patch)
 		#find files
-		if [[ "$prev" = "$command" ]]; then
+		if [[ "$args" = "2" ]]; then
 			_filedir
 		#firmware_type
-		elif [[ "$prev" = "${COMP_WORDS[2]}" ]]; then
+		elif [[ "$args" = "3" ]]; then
 			_show_firmware_types
-		#modifiers
-		else
-			_show_modifiers
 		fi
 		;;
 	firmware-convert)
 		#file in
-		if [[ "$prev" = "$command" ]]; then
+		if [[ "$args" = "2" ]]; then
 			_filedir
 		#file out
-		elif [[ "$prev" = "${COMP_WORDS[2]}" ]]; then
+		elif [[ "$args" = "3" ]]; then
 			_filedir
 		#firmware_type in
-		elif [[ "$prev" = "${COMP_WORDS[3]}" ]]; then
+		elif [[ "$args" = "4" ]]; then
 			_show_firmware_types
 		#firmware_type out
-		elif [[ "$prev" = "${COMP_WORDS[4]}" ]]; then
+		elif [[ "$args" = "5" ]]; then
 			_show_firmware_types
-		#modifiers
-		else
-			_show_modifiers
 		fi
 		;;
 	*)
 		#find first command
-		if [[ ${COMP_CWORD} = 1 ]]; then
+		if [[ "$args" = "1" ]]; then
 			COMPREPLY=( $(compgen -W '${_fwupdtool_cmd_list[@]}' -- "$cur") )
-		#modifiers for all commands
-		else
-			_show_modifiers
 		fi
 		;;
 	esac
+
+	#modifiers
+	_show_modifiers
 
 	return 0
 }
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -126,6 +126,9 @@ _fwupdtool()
 		#find files
 		if [[ "$prev" = "$command" ]]; then
 			_filedir
+		#device ID
+		elif [[ "$prev" = "${COMP_WORDS[2]}" ]]; then
+			_show_device_ids
 		#modifiers
 		else
 			_show_modifiers
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -135,6 +135,7 @@ _fwupdtool()
 		fi
 		;;
 	attach|detach|activate|verify-update|reinstall|get-updates)
+		#device ID
 		if [[ "$prev" = "$command" ]]; then
 			_show_device_ids
 		#modifiers
@@ -155,6 +156,7 @@ _fwupdtool()
 		#output
 		elif [[ "$prev" = "${COMP_WORDS[4]}" ]]; then
 			_filedir
+		#modifiers
 		else
 			_show_modifiers
 		fi
@@ -166,6 +168,7 @@ _fwupdtool()
 		#firmware_type
 		elif [[ "$prev" = "${COMP_WORDS[2]}" ]]; then
 			_show_firmware_types
+		#modifiers
 		else
 			_show_modifiers
 		fi
@@ -183,6 +186,7 @@ _fwupdtool()
 		#firmware_type out
 		elif [[ "$prev" = "${COMP_WORDS[4]}" ]]; then
 			_show_firmware_types
+		#modifiers
 		else
 			_show_modifiers
 		fi
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -77,6 +77,15 @@ _show_firmware_types()
 	COMPREPLY+=( $(compgen -W "${firmware_types}" -- "$cur") )
 }
 
+_show_device_ids()
+{
+	if ! command -v jq &> /dev/null; then
+		return 0
+	fi
+	local description
+	description="$(command fwupdmgr get-devices --json 2>/dev/null | jq '.Devices | .[] | .DeviceId')"
+	COMPREPLY+=( $(compgen -W "${description}" -- "$cur") )
+}
 
 _show_plugins()
 {
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -104,11 +104,10 @@ _show_modifiers()
 
 _fwupdtool()
 {
-	local cur prev command args
+	local cur prev command arg args
 	COMPREPLY=()
-	cur=`_get_cword`
-	prev=${COMP_WORDS[COMP_CWORD-1]}
-	command=${COMP_WORDS[1]}
+	_get_comp_words_by_ref cur prev
+	_get_first_arg
 	_count_args
 
 	case $prev in
@@ -122,7 +121,7 @@ _fwupdtool()
 		;;
 	esac
 
-	case $command in
+	case $arg in
 	get-details|install|install-blob|firmware-dump|firmware-read)
 		#find files
 		if [[ "$args" = "2" ]]; then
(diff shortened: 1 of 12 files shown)


data/bash-completion/fwupdtool
```diff
@@ -76,6 +76,15 @@ _show_firmware_types()
 	COMPREPLY+=( $(compgen -W "${firmware_types}" -- "$cur") )
 }
 
+_show_device_ids()
+{
+	if ! command -v jq &> /dev/null; then
+		return 0
+	fi
+	local description
+	description="$(command fwupdmgr get-devices --json 2>/dev/null | jq '.Devices | .[] | .DeviceId')"
```
superm1, line 85: I don't think fwupdtool can bash autocomplete with fwupdmgr can it?

superm1: In general I don't think we want to be calling one from the other if we can avoid it.

superm1: The reason I'm saying to avoid it is that fwupdtool will close fwupd daemon when it runs.  However fwupdtool is a lot slower to startup, particularly for bash completion.

Maybe we should make an exception on this.  @hughsie thoughts?

hughsie: I think two things fighting for the same lock is a terrible idea. Maybe fwupd could write a list of current device Ids somewhere in /var/lib/fwupd for just this purpose?

superm1: Yeah that's a good idea!  Don't we do that with fwupdtool?  I seem to remember there was an argument to make it happen in chrome.

gportay: > The reason I'm saying to avoid it is that fwupdtool will close fwupd daemon when it runs. However fwupdtool is a lot slower to startup, particularly for bash completion.
> 

Oh, that is not kind indeed... `fwupdmgr attach [TAB]` would start the daemon to list the devices-id, and kill it then (if is not up and running already).

I preferred used `fwupdmgr` because if is faster to get the device-ids cached by the daemon; `fwupdtool get-devices` may take a while and more to output the device-ids.

For now, it makes sense to make the change, with `fwupdtool` instead. However, `fwupdtool` would kill the daemon anyway. And the `fwupdmgr` bash completion script should not run `fwupdtool` for completion as it is like doing hara-kiri (killing the daemon to restart it just for completion).

I need time for reflexion...

superm1: > Yeah that's a good idea! Don't we do that with fwupdtool? I seem to remember there was an argument to make it happen in chrome.

So we have a `--enable-json-state` argument for `fwupdtool`.  I think this would make sense to use to store the devices.  
I would propose that we drop/deprecate that option and add a configuration file option for that behavior (on by default for `fwupdtool`).  Then any time that `fwupdtool` is launched it updates the state file and your PR uses that for bash completion invisibly.  If it's missing, then you return 0.  You launch get-devices once it updates.

> Btw, why do we have two tools already? I am laking a bit of history.

Ah history.  So originally we have only client / daemon architecture but then there are certain commands/arguments that don't make sense to expose to all unprivileged users.  For example `install-blob`.  These were put into a tool that wasn't in PATH for advanced use cases.  Later it was determined that we shouldn't put tools like that out of PATH, and it got moved to PATH.  
Later again there was a desire to adopt fwupd in Chrome OS but they weren't willing to take on a new privileged daemon on the system.  So it was overhauled to support many of the same features as the client does.

Later that changed and now ChromeOS takes client/daemon.

But the net outcome is that now we have two tools with lots of overlapping functionality.  `fwupdtool` is more useful for developers/developing plugins and `fwupdmgr` is more useful for end users who do things like update their firmware only.

Hopefully I remembered that all right :)

superm1: Makes sense to me. Probably /var/cache though.

gportay: @hughsie I am running out of time too. I would do it later.

gportay: When/where should I create the file? in the `fu_engine_update_motd()`?

hughsie: Maybe not that exact function, but yes, another new function that's called at the same time as that one.

gportay: I have pushed an initial sketch.

However, it does not update the file if a device is un-pluged; any idea?

_Unrelated_: I have noticed that removing the device causes the wait for the re-plug delay, and I guess this is not useful in case of none-upgrade unless I miss something.

```
08:44:54:0472 FuEngine             auto-removing backend device Rival 3 Wireless [b0a78eb71f4eeea7df8fb114522556ba8ce22074]
08:44:54:0472 FuDeviceList         waiting 40000ms for Rival 3 Wireless device removal
(...)
08:45:34:0485 FuDeviceList         doing delayed removal
08:45:34:0486 FuDeviceList         ::removed b0a78eb71f4eeea7df8fb114522556ba8ce22074
```

hughsie: Is fu_engine_emit_changed() on remove? It might only be fired after the 40000ms delayed removal.

> Unrelated: I have noticed that removing the device causes the wait for the re-plug delay

Out of interest; if you pull in the patch from wip/hughsie/ONLY_WFR_REMOVE_DELAY does that make things more obvious? You'll have to set FU_DEVICE_INTERNAL_FLAG_ONLY_WFR_REMOVE_DELAY on the device in _init().

gportay:
```
09:33:50:0350 FuEngine             auto-removing backend device Rival 3 Wireless [b0a78eb71f4eeea7df8fb114522556ba8ce22074]
09:33:50:0350 FuDeviceList         ::removed b0a78eb71f4eeea7df8fb114522556ba8ce22074
```
:tada:

gportay: And tested after `_write()`

```
09:37:04:0367 FuDeviceList         waiting 40000ms for replug
09:37:26:0430 FuDevice             already set GType to FuSteelseriesFizz, ignoring FuSteelseriesFizz
09:37:26:0431 FuDevice             using e8055e4e2470749d3ebc7735868bf869c94ebd8f for usb:03:00:02:03
09:37:26:0435 FuPlugin             emit added from steelseries: e8055e4e2470749d3ebc7735868bf869c94ebd8f
09:37:26:0435 FuDeviceList         ::added e8055e4e2470749d3ebc7735868bf869c94ebd8f
09:37:26:0436 FuEngine             writing motd target /home/gportay/src/GOO0034/fwupd/cache/motd.d/85-fwupd
09:37:26:0437 FuCommon             writing /home/gportay/src/GOO0034/fwupd/cache/devices.json with 1837 bytes
09:37:44:0368 FuHistory            modifying device Rival 3 Wireless [b0a78eb71f4eeea7df8fb114522556ba8ce22074]
09:37:44:0383 FuHistory            modifying device Rival 3 Wireless [b0a78eb71f4eeea7df8fb114522556ba8ce22074]
09:37:44:0391 FuIdle               uninhibiting: update
```

gportay: > Is fu_engine_emit_changed() on remove? It might only be fired after the 40000ms delayed removal.

I added `fu_engine_emit_changed()` on remove.

src/fu-engine-helper.c
```diff
@@ -133,3 +135,53 @@ fu_engine_update_motd(FuEngine *self, GError **error)
...
+	data = json_generator_to_data(generator, &len);
+	if (data == NULL) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "Failed to convert to JSON string");
+		return FALSE;
+	}
+
+	directory = fu_common_get_path(FU_PATH_KIND_CACHEDIR_PKG);
```
gportay, line 182: What about storing in the `/run` directory instead of the `/var/cache`?

hughsie: @superm1 ? this feels like the wong things to do from a FHS but it is appealing because it's cleared on reboot.

gportay:
Back to that PR; work is still in progress.

data/bash-completion/fwupdtool
hughsie: I think we need to jq that cached file now, right?

src/fu-engine-helper.c
hughsie: I think /var/cache seems most appropriate tbh.

gportay: Ok

data/bash-completion/fwupdtool
gportay: With redirecting `stderr` to `null` to hide error if file does not exist.

gportay:
In the current state, this works pretty smoothly (aka `jq` cache in the completion script).

data/bash-completion/fwupdtool
gportay: @hughsie I am not a master or `meson`; what is the best solution to `sed`-ish the bash completion script if the prefix is changed at configure time? because the path to the file is hardcoded for now.

However, this is pointless if I am parsing the file in `fwupdtool` if using a new option (`--use-cache`?) or reusing `--enable-json-state` mentioned by @superm1.

hughsie: That's done with configure_file() e.g. see https://github.com/fwupd/fwupd/blob/main/data/installed-tests/meson.build -- the filename doesn't have to have a .in suffix, that's a bit of a relic from the past we don't need to repeat.

data/bash-completion/fwupdtool
```diff
@@ -77,6 +77,15 @@ _show_firmware_types()
 	COMPREPLY+=( $(compgen -W "${firmware_types}" -- "$cur") )
 }
 
+_show_device_ids()
+{
+	if ! command -v jq &> /dev/null; then
+		return 0
+	fi
+	local description
+	description="$(command jq '.Devices | .[] | .DeviceId' /var/cache/fwupd/devices.json 2>/dev/null)"
```
superm1, line 86: I don't like hardcoding this path; can we perhaps instead fill in the `/var/cache/fwupd` from meson?

src/fu-engine.c
```diff
@@ -6366,6 +6372,7 @@ fu_engine_backend_device_removed_cb(FuBackend *backend, FuDevice *device, FuEngi
 				fu_device_get_name(device_tmp),
 				fu_device_get_id(device_tmp));
 			fu_device_list_remove(self->device_list, device_tmp);
+			fu_engine_emit_changed(self);
```
superm1, line 6375: I wonder if we should deprecate/drop `enable-json-state` now for fwupdtool.  This is effectively done by all updates this means from fwupd or fwupdtool.

data/bash-completion/fwupdtool
gportay: > I don't like hardcoding this path; can we perhaps instead fill in the `/var/cache/fwupd` from meson?

The other way around is to implement the load from json in `fwupdtool` directly (https://github.com/fwupd/fwupd/pull/4493#discussion_r896652875).

src/fu-engine.c
gportay: I do not mind deprecating that option; I wonder if people is using it.

superm1: > I do not mind deprecating that option; I wonder if people is using it.

Yeah let's please do that as part of this PR.

superm1:
> @superm1 do you think we need a config bool for this?

I don't think we need a config file, but the one thing we need to concern ourselves with is whether this is leaking privileged information.  Like should serial numbers or GUIDs end up in the JSON state file?  As long as we follow config file options for that information I'm fine.

data/bash-completion/fwupdtool
superm1: Either way is fine to me.  If you keep it this way just use `configure_file` to populate the path. If you switch to `fwupdtool` to parse it just make sure that it's fast path (don't load plugins and engine and stuff).

gportay: > Either way is fine to me. If you keep it this way just use `configure_file` to populate the path. If you switch to `fwupdtool` to parse it just make sure that it's fast path (don't load plugins and engine and stuff).

I will go for the `configure_file` and then switch back to `fwupdtool` in a later PR as it takes lots of changes.

src/fu-engine.c
gportay: what does deprecating the option imply?

Adds a warning in a first step, or drop all the code off directly?

superm1: > what does deprecating the option imply?
> 
> Adds a warning in a first step, or drop all the code off directly?

I think you can just drop all the code and just have it emit a message when that parameter is used can't you?  Since you're writing out that file by default now.

hughsie:
> whether this is leaking privileged information

Also my worry. We do have `fwupd_device_to_variant_full()` and maybe we should also have `fwupd_device_to_json_full()` that also takes the `FwupdDeviceFlags` and doesn't include the serial numbers and instance IDs when it's not trusted. e.g. like https://github.com/fwupd/fwupd/blob/main/libfwupd/fwupd-device.c#L2208

superm1:
> > whether this is leaking privileged information
> 
> Also my worry. We do have `fwupd_device_to_variant_full()` and maybe we should also have `fwupd_device_to_json_full()` that also takes the `FwupdDeviceFlags` and doesn't include the serial numbers and instance IDs when it's not trusted. e.g. like https://github.com/fwupd/fwupd/blob/main/libfwupd/fwupd-device.c#L2208

yeah that sounds good.

src/fu-engine.c
gportay: Done.

@superm1, could you check I have not missed something please?

gportay:
> > > whether this is leaking privileged information
> > 
> > 
> > Also my worry. We do have `fwupd_device_to_variant_full()` and maybe we should also have `fwupd_device_to_json_full()` that also takes the `FwupdDeviceFlags` and doesn't include the serial numbers and instance IDs when it's not trusted. e.g. like https://github.com/fwupd/fwupd/blob/main/libfwupd/fwupd-device.c#L2208
> 
> yeah that sounds good.

I have to understand all of this 😅

hughsie:
> I have to understand all of this sweat_smile

tl;dr: rename() fwupd_device_to_json to fwupd_device_to_json_full and add the extra arg to control the slightly secret bits like fwupd_device_to_variant_full does. Then add a compat fwupd_device_to_json() that calls fwupd_device_to_json_full() so we don't break API.

src/fu-tool.c
```diff
@@ -3473,14 +3408,6 @@ main(int argc, char *argv[])
...
 	     N_("Run the plugin composite cleanup routine when using install-blob"),
 	     NULL},
-	    {"enable-json-state",
-	     '\0',
-	     0,
-	     G_OPTION_ARG_NONE,
-	     &priv->enable_json_state,
-	     /* TRANSLATORS: command line option */
-	     N_("Save device state into a JSON file between executions"),
-	     NULL},
```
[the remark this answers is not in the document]
gportay: It makes sense to not drop it if it is used.

I prefer to wait for @hughsie opinion about this before clicking the merge button :)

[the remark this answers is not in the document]
hughsie: I think we can drop it; IIRC it was @superm1 who wanted it in the first place.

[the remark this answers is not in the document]
superm1: Okay thanks

PR #4499, opened 2022-04-06
Add the fwupd-efi version as a firmware requirement

hughsie:
If we rely on a fix in fwupd-efi, the firmware may want to depend on it.

Although we're adding a compile time version provide, normally the
fwupd-efi.pc file is actually shipped in the binary package.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
+++ b/plugins/uefi-capsule/fu-plugin-uefi-capsule.c
@@ -23,13 +23,107 @@ struct FuPluginData {
 	FuUefiBgrt *bgrt;
 	FuVolume *esp;
 	FuBackend *backend;
+	GFile *fwupd_efi_file;
+	GFileMonitor *fwupd_efi_monitor;
 };
+
+static gboolean
+fu_plugin_uefi_capsule_fwupd_efi_parse(FuPlugin *plugin, GError **error)
+{
+	FuPluginData *data = fu_plugin_get_data(plugin);
+	FuContext *ctx = fu_plugin_get_context(plugin);
+	const guint8 needle[] = "f\0w\0u\0p\0d\0-\0e\0f\0i\0 \0v\0e\0r\0s\0i\0o\0n\0 ";
+	gsize bufsz = 0;
+	gsize offset = 0;
+	guint16 version_tmp[16] = {0x0};
+	g_autofree gchar *buf = NULL;
+	g_autofree gchar *version = NULL;
+
+	/* find the UTF-16 version string */
+	if (!g_file_load_contents(data->fwupd_efi_file, NULL, &buf, &bufsz, NULL, error))
+		return FALSE;
+	if (!fu_memmem_safe((const guint8 *)buf, bufsz, needle, sizeof(needle), &offset, error)) {
+		g_autofree gchar *fn = g_file_get_path(data->fwupd_efi_file);
+		g_prefix_error(error, "searching %s: ", fn);
+		return FALSE;
+	}
+
+	/* align */
+	if (!fu_memcpy_safe((guint8 *)version_tmp,
+			    sizeof(version_tmp),
+			    0x0, /* dst */
+			    (const guint8 *)buf,
+			    bufsz,
+			    offset + sizeof(needle),
+			    sizeof(version_tmp) - sizeof(guint16),
+			    error))
+		return FALSE;
+
+	/* convert to UTF-8 */
+	version = g_utf16_to_utf8(version_tmp, -1, NULL, NULL, error);
+	if (version == NULL) {
+		g_autofree gchar *fn = g_file_get_path(data->fwupd_efi_file);
+		g_prefix_error(error, "converting %s: ", fn);
+		return FALSE;
+	}
+
+	/* success */
+	fu_context_add_runtime_version(ctx, "org.freedesktop.fwupd-efi", version);
+	return TRUE;
+}
+
+static void
+fu_plugin_uefi_capsule_fwupd_efi_changed_cb(GFileMonitor *monitor,
+					    GFile *file,
+					    GFile *other_file,
+					    GFileMonitorEvent event_type,
+					    gpointer user_data)
+{
+	FuPlugin *plugin = FU_PLUGIN(user_data);
+	FuContext *ctx = fu_plugin_get_context(plugin);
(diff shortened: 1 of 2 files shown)


superm1:
> Although we're adding a compile time version provide, normally the
fwupd-efi.pc file is actually shipped in the binary package.

The whole point of splitting the two was to avoid stuff like this and let the packages move independently.  This means that fwupd needs to be _rebuilt_ against a new fwupd-efi.  Any other way we can do this at runtime instead perhaps?  We could encode the version in the EFI binary header and then parse that EFI binary at runtime.  If it's not there assume it's older version.

hughsie:
>  Any other way we can do this at runtime instead perhaps

I think that's what we have to do. Will ponder.

hughsie:
> Will ponder.

Is this insane:

 * Generate a uSWID tag and embed it the .c source somewhere. e.g. `static const UINT8 _unused_version[] = "\x53\x42\x4f\x4d\xd\x6\xba\x2e\xac\x...";`
 
or:

* Build an actual .o file with a bona-fide .SBOM section with the FuUswidFirmware class and fwupdtool, with a bit of help from objcopy

..either of these would let us get the runtime version by doing `fu_uswid_firmware_new()` + `fu_firmware_parse()` + `fu_firmware_get_version()`

Not sure if there's an easier way to get a random string into a different architecture binary...

superm1:
I think the second idea is a lot cleaner, it sounds reasonable to me.  With both approaches you can also go and take this information to make the fwupd binary emit something like:
`fwupd EFI binary v2.5` at startup too since it's encoded in the binary.

An easier (but dirtier) solution would just be a sentinel surrounding a static string, and as long as that string is used the compiler wouldn't optimize it out you can look for the sentinel.  I guess that's conceptually similar to the first idea, but by actually using it for outputting a version you have a better chance compiler keeps it.

hughsie:
> emit something like fwupd EFI binary v2.5 at startup

Yes, this seems the pragmatic solution. Done in https://github.com/fwupd/fwupd-efi/pull/23 and I've updated this PR to include both build-time and runtime fwupd-efi versions.

superm1 (changes requested):
The check for this needs to be during firmware install not fwupd startup. It's possible for a user to replace the EFI binary on the filesystem without restarting fwupd daemon.

hughsie:
> The check for this needs to be during firmware install not fwupd startup

I'm not sure we can do that; we depsolve based on the fulfilled requirements. What we could do is set up a GFileMonitor on that binary and update the runtime version when the daemon is running.

hughsie:
> Probably need to monitor the directory not binary though

will check.

hughsie:
> There might be nothing installed yet when the daemon starts

I think that's super unlikely and would complicate the code even more than now. I think what we have now is probably complete enough IMHO.

src/fu-engine.c
```diff
@@ -7151,6 +7151,11 @@ fu_engine_init(FuEngine *self)
 	g_hash_table_insert(self->compile_versions,
 			    g_strdup("org.freedesktop.fwupd"),
 			    g_strdup(VERSION));
+#ifdef FWUPD_EFI_VERSION
```
superm1, line 7154: I don't see a lot of value to compile version. I think runtime is all we need.

hughsie: Ya, agree.  I've dropped this bit.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -23,13 +23,102 @@ struct FuPluginData {
...
+					    GFile *other_file,
+					    GFileMonitorEvent event_type,
+					    gpointer user_data)
+{
+	FuPlugin *plugin = FU_PLUGIN(user_data);
+	FuContext *ctx = fu_plugin_get_context(plugin);
+	g_autoptr(GError) error_local = NULL;
+
+	if (!fu_plugin_uefi_capsule_fwupd_efi_parse(plugin, &error_local)) {
+		fu_context_add_runtime_version(ctx, "org.freedesktop.fwupd-efi", NULL);
```
superm1, line 87: I think if you found the file but not the string you can make an assumption on the version can't you?

hughsie: I don't think so; or do you mean something like fwupd 1.3.whatever it was

superm1: > I don't think so; or do you mean something like fwupd 1.3.whatever it was

Yeah that's what I mean. Set it to last release of fwupd-efi we have before this fix goes in.
It's a reasonable assumption and the most "common" case right now.

superm1: > I think i'd rather have the number of firmwares depending on fwupd-efi to be as small as possible....

If that system is preproduction maybe it can be 0...

hughsie: > Set it to last release of fwupd-efi we have before this fix goes in

Surely, it should be the first release of fwupd-efi? we don't know it's not 1.1, 1.2. etc.

> If that system is preproduction

😬

superm1: > > Set it to last release of fwupd-efi we have before this fix goes in
> 
> Surely, it should be the first release of fwupd-efi? we don't know it's not 1.1, 1.2. etc.
> 

Yeah

hughsie: Done.

src/fu-tool.c
```diff
@@ -3132,6 +3132,10 @@ fu_util_version(FuUtilPrivate *priv, GError **error)
 	g_autoptr(GHashTable) metadata = NULL;
 	g_autofree gchar *str = NULL;
 
+	/* load engine */
```
superm1, line 3135: This code change no longer needed if not looking at compile version right?

superm1 (approved):
One nit then lgtm

src/fu-tool.c
hughsie: Nah, we do need to init the plugins to get the runtime version. Thanks for all the reviews btw.

PR #4500, opened 2022-04-06
uefi-capsule: Fix installing updates on a HP M60 Workstation

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-device.c
+++ b/plugins/uefi-capsule/fu-uefi-device.c
@@ -395,23 +395,37 @@ GBytes *
 fu_uefi_device_fixup_firmware(FuUefiDevice *self, GBytes *fw, GError **error)
 {
 	FuUefiDevicePrivate *priv = GET_PRIVATE(self);
-	gsize fw_length;
-	const guint8 *data = g_bytes_get_data(fw, &fw_length);
+	gsize bufsz;
+	const guint8 *buf = g_bytes_get_data(fw, &bufsz);
 	g_autofree gchar *guid_new = NULL;
 
 	priv->missing_header = FALSE;
 
 	/* GUID is the first 16 bytes */
-	if (fw_length < sizeof(fwupd_guid_t)) {
+	if (bufsz < sizeof(fwupd_guid_t)) {
 		g_set_error_literal(error,
 				    FWUPD_ERROR,
 				    FWUPD_ERROR_INVALID_FILE,
 				    "Invalid payload");
 		return NULL;
 	}
-	guid_new = fwupd_guid_to_string((fwupd_guid_t *)data, FWUPD_GUID_FLAG_MIXED_ENDIAN);
+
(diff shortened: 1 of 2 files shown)


plugins/uefi-capsule/fu-uefi-device.c
```diff
@@ -395,23 +395,37 @@ GBytes *
...
-	guid_new = fwupd_guid_to_string((fwupd_guid_t *)data, FWUPD_GUID_FLAG_MIXED_ENDIAN);
+
+	/* check for HP firmware that has some weird logo header prefix */
+	if (!fu_device_has_private_flag(FU_DEVICE(self),
+					FU_UEFI_DEVICE_FLAG_NO_CAPSULE_HEADER_FIXUP)) {
+		guint8 hdr[5] = {0x0};
+		if (!fu_memcpy_safe(hdr, sizeof(hdr), 0x0, buf, bufsz, 0x0, sizeof(hdr), error))
+			return NULL;
+		if (memcmp(hdr, "$LOGO", 5) == 0) {
+			g_warning("auto-enabling NO_CAPSULE_HEADER_FIXUP, this HP system probably "
```
superm1, line 420: don't assume it's an HP system in the warning

hughsie: I'm really hesitant to turn it on unconditionally too. Maybe we just do the warning and don't set `NO_CAPSULE_HEADER_FIXUP`?

superm1:
Something that never sat well with me is that how can we have different behavior for headers with Windows and Linux?  Do we have a core issue here that Windows always adds or always strips the header or something?  The special case of devices not in the ESRT for Dell excluded shouldn't we have the same behavior for everything else?

plugins/uefi-capsule/fu-uefi-device.c
hughsie: I've asked HP for more information about the $LOGO thing. It might not be HP specific etc.

superm1:
> > Windows always adds or always strips the header or something
> 
> I think, with hindsight, that Windows probably doesn't auto-add a header. The UEFI spec says that updates should have a header, but I know there are systems that ship updates on the LVFS without a header and that rely on fwupd adding a header so that they can actually work on the hardware, so I'm hesitant to flip the default now. Ideally hardware adheres to the specification, and vendors upload the correct thing, but we both know that's a reality we don't live in.

I don't understand how vendors can share the update with windows if it doesn't add the header though. I guess there is a possibility that they don't send updates to WU..

hughsie:
> I think, with hindsight, that Windows probably doesn't auto-add a header

I think the pragmatic thing to do here is just to tell HP to use the https://github.com/fwupd/fwupd/blob/main/contrib/firmware_packager/add_capsule_header.py script and reupload packages. It seems I'm making things more complicated than I need to.

PR #4507, opened 2022-04-08
dell-dock: set minimum version of EC

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-dell-dock-i2c-ec.c
+++ b/plugins/dell-dock/fu-dell-dock-i2c-ec.c
@@ -826,7 +826,6 @@ fu_dell_dock_ec_write_fw(FuDevice *device,
 
 	/* meet the minimum EC version */
 	if ((flags & FWUPD_INSTALL_FLAG_FORCE) == 0 &&
-	    (self->data->module_type != MODULE_TYPE_130_USB4) &&
 	    (fu_common_vercmp_full(dynamic_version,
 				   self->ec_minimum_version,
 				   FWUPD_VERSION_FORMAT_QUAD) < 0)) {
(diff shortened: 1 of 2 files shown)


CragW:
New WD dock is announced, the minimum requirement is therefore added to avoid accidental downgrade to an unsupported version given the WD docks are sharing the same device id.

plugins/dell-dock/fu-dell-dock-i2c-ec.c
```diff
@@ -826,7 +826,6 @@ fu_dell_dock_ec_write_fw(FuDevice *device,
 
 	/* meet the minimum EC version */
 	if ((flags & FWUPD_INSTALL_FLAG_FORCE) == 0 &&
-	    (self->data->module_type != MODULE_TYPE_130_USB4) &&
```
hughsie, line 829: This is intentional, right?

hughsie (approved):

hughsie:
> the WD docks are sharing the same device id

:face_with_spiral_eyes:

plugins/dell-dock/fu-dell-dock-i2c-ec.c
CragW: Yes, it was a quirk for the new dock during development phase.

superm1:
> > the WD docks are sharing the same device id
> 
> :face_with_spiral_eyes: 

They're hardware upgradable

superm1:
> > > the WD docks are sharing the same device id
> > 
> > :face_with_spiral_eyes: 
> 
> They're hardware upgradable 

Sorry didn't finish my thought. So you want to ensure that someone doesn't take an older release in case they switch modules.

PR #4510, opened 2022-04-08
dell-dock: set mst payload default unsigned

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -81,10 +81,12 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 		instance_id_mst = DELL_DOCK_VMM6210_INSTANCE_ID;
 	else
 		instance_id_mst = DELL_DOCK_VM5331_INSTANCE_ID;
+	fu_device_add_instance_id(FU_DEVICE(mst_device), instance_id_mst);
 	instance_guid_mst = fwupd_guid_hash_string(instance_id_mst);
 	fu_device_add_guid(FU_DEVICE(mst_device), instance_guid_mst);
+	if (!fu_device_probe(FU_DEVICE(mst_device), error))
+		return FALSE;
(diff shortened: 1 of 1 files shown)


plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -79,6 +79,7 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 	else
 		instance = DELL_DOCK_VM5331_INSTANCE_ID;
 	fu_device_add_guid(FU_DEVICE(mst_device), fwupd_guid_hash_string(instance));
+	fu_device_add_flag(FU_DEVICE(mst_device), FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);
```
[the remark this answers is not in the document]
CragW: Good idea, I think calling `probe()` would work, just out of curiosity why don't we `open` the device through `fu_plugin_dell_dock_create_node()`? I thought it walks through the flow in an orderly way thus may be needed, no?

plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -81,10 +81,12 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 		instance_id_mst = DELL_DOCK_VMM6210_INSTANCE_ID;
 	else
 		instance_id_mst = DELL_DOCK_VM5331_INSTANCE_ID;
+	fu_device_add_instance_id(FU_DEVICE(mst_device), instance_id_mst);
 	instance_guid_mst = fwupd_guid_hash_string(instance_id_mst);
 	fu_device_add_guid(FU_DEVICE(mst_device), instance_guid_mst);
```
hughsie, line 86: Do you still need these two lines? If so, LGTM.

CragW: Yes, they are needed.

superm1 (approved):

PR #4513, opened 2022-04-11
trivial: add device GUID to avoid warning message 'no GUIDs for device'

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

add-child-tbt-device-no-guids

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -61,7 +61,10 @@ fu_plugin_dell_dock_create_node(FuPlugin *plugin, FuDevice *device, GError **err
 static gboolean
 fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 {
-	const gchar *instance;
+	const gchar *instance_id_mst;
+	const gchar *instance_id_status;
+	g_autofree const gchar *instance_guid_mst = NULL;
+	g_autofree const gchar *instance_guid_status = NULL;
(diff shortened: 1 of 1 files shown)


plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -101,6 +101,9 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 	/* create TBT endpoint if Thunderbolt SKU and Thunderbolt link inactive */
 	if (fu_dell_dock_ec_needs_tbt(FU_DEVICE(ec_device))) {
 		g_autoptr(FuDellDockTbt) tbt_device = fu_dell_dock_tbt_new(proxy);
+		fu_device_set_context(FU_DEVICE(tbt_device), ctx);
+		fu_device_add_guid(FU_DEVICE(tbt_device),
+				   fwupd_guid_hash_string(DELL_DOCK_TBT_INSTANCE_ID));
```
hughsie, line 106: this leaks the string; I really think you should use `fu_device_add_instance_id(FU_DEVICE(tbt_device), DELL_DOCK_TBT_INSTANCE_ID)` instead -- although it might mean calling `fu_device_setup()` manually.

CragW: We need the `parent` device prepared prior to running `setup`, the flow would look like `add_guid` > `add_child` > `create_node` for `open` and `setup`. I can make a change to `autofree` the string which is introduced by hash.

hughsie (approved):

PR #4515, opened 2022-04-11
Feature: Clevo/Tuxedo me unlock

SergiiDmytruk:
This is follow up on [Enabling support for ME unlock](https://groups.google.com/g/fwupd/c/e6yE6xy5Hjg) discussion on ML.

What it's needed in the end:
 1. Reading lock status from HSFS and sharing it for two use cases:
    1. State verification before unlocking.
    2. State verification before flashing ME firmware.
 2. Ability to unlock ME.
 3. Ability to flash ME firmware.

How it's done in this PR in its WIP state:
 1. Sharing ME lock status via `FuContext`.  Value is set by `intel-spi` plugin.  Consumed in `superio` and `flashrom`.
 2. Via `unlock` command for IT55 `superio` device, because unlocking is done through EC.
 3. Happens in `flashrom` plugin.  Device for ME firmware is added if ME region is unlocked.

Things to consider:
 1. What's the best way of presenting ME flashing to the user?  I guess ideally unlocking and flashing should be done for the same device. I saw that plugins can monitor created devices, but I think they can't invoke code from one another, so not sure how to implement all functionality in one place.
 2. `intel-spi` is re-used to not duplicate neither code nor quirks that are there and also because some kind of sharing would be necessary anyway.

ME device is incomplete at the moment, despite that the changes should capture the end goal relatively well. Please take a look to see if such changes are sane and if there are better ways of going about them.

P.S. First two commits are trivial changes for things I've noticed when going through the code and can be cherry-picked right away.

### DIFF
--- a/plugins/flashrom/README.md
+++ b/plugins/flashrom/README.md
@@ -2,7 +2,10 @@
 
 ## Introduction
 
-This plugin uses `flashrom` to update the system firmware.
+This plugin uses `libflashrom` to update the system firmware.  It can be used
+to update BIOS or ME regions of the flash.  Device for ME region is created
+only if "Intel SPI" plugin indicates that such a region exists, which makes
+"Intel SPI" a dependency of this plugin for doing ME updates.
 
 ## Firmware Format
 
@@ -44,6 +47,16 @@ Internal device uses hardware ID values which are derived from SMBIOS.
 
 They should match the values provided by `fwupdtool hwids` or the
 `ComputerHardwareIds.exe` Windows utility.
+
+One more GUID has the following form:
+
+* `FLASHROM\VENDOR_{manufacturer}&PRODUCT_{product}&REGION_{ifd_region_name}`
+
+Its purpose is to target specific regions of the flash as defined by IFD (Intel
+SPI Flash Descriptor), examples:
+
+* `FLASHROM\VENDOR_Notebook&PRODUCT_NS50MU&REGION_BIOS`
+* `FLASHROM\VENDOR_Notebook&PRODUCT_NS50MU&REGION_ME`
 
 ## Update Behavior
 
--- a/plugins/flashrom/fu-flashrom-device.c
+++ b/plugins/flashrom/fu-flashrom-device.c
@@ -28,6 +28,7 @@ struct _FuFlashromDevice {
 	FuUdevDevice parent_instance;
 	FuIfdRegion region;
 	struct flashrom_flashctx *flashctx;
+	struct flashrom_layout *layout;
 };
 
 G_DEFINE_TYPE(FuFlashromDevice, fu_flashrom_device, FU_TYPE_UDEV_DEVICE)
@@ -80,6 +81,7 @@ static gboolean
 fu_flashrom_device_open(FuDevice *device, GError **error)
 {
 	FuFlashromDevice *self = FU_FLASHROM_DEVICE(device);
+	struct flashrom_layout *layout;
 
 	/* get the flash size from the device if not already been quirked */
 	if (fu_device_get_firmware_size_max(device) == 0) {
@@ -92,6 +94,44 @@ fu_flashrom_device_open(FuDevice *device, GError **error)
 			return FALSE;
 		}
 		fu_device_set_firmware_size_max(device, flash_size);
+	}
+
+	if (flashrom_layout_read_from_ifd(&layout, self->flashctx, NULL, 0)) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_READ,
+				    "failed to read layout from Intel ICH descriptor");
+		return FALSE;
+	}
+
+	/* update only one specific region of the flash and do not touch others */
+	if (flashrom_layout_include_region(layout, fu_ifd_region_to_string(self->region))) {
+		flashrom_layout_release(layout);
+
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "invalid region name");
+		return FALSE;
+	}
+
+	/* flashrom_layout_set() doesn't transfer ownership, so we must manage layout's lifetime */
+	self->layout = layout;
+	flashrom_layout_set(self->flashctx, self->layout);
+
+	return TRUE;
+}
+
+static gboolean
+fu_flashrom_device_close(FuDevice *device, GError **error)
+{
+	FuFlashromDevice *self = FU_FLASHROM_DEVICE(device);
+
+	if (self->layout != NULL) {
+		flashrom_layout_release(self->layout);
+		self->layout = NULL;
+
+		flashrom_layout_set(self->flashctx, NULL);
 	}
 
 	return TRUE;
@@ -115,30 +155,9 @@ fu_flashrom_device_prepare(FuDevice *device,
 	if (!fu_common_mkdir_parent(firmware_orig, error))
 		return FALSE;
 	if (!g_file_test(firmware_orig, G_FILE_TEST_EXISTS)) {
-		struct flashrom_layout *layout;
 		gsize flash_size = fu_device_get_firmware_size_max(device);
 		g_autofree guint8 *newcontents = g_malloc0(flash_size);
 		g_autoptr(GBytes) buf = NULL;
-
-		if (flashrom_layout_read_from_ifd(&layout, self->flashctx, NULL, 0)) {
-			g_set_error_literal(error,
-					    FWUPD_ERROR,
-					    FWUPD_ERROR_READ,
-					    "failed to read layout from Intel ICH descriptor");
-			return FALSE;
-		}
-
-		/* update only one specific region of the flash and do not touch others */
-		if (flashrom_layout_include_region(layout, fu_ifd_region_to_string(self->region))) {
-			g_set_error_literal(error,
-					    FWUPD_ERROR,
-					    FWUPD_ERROR_NOT_SUPPORTED,
-					    "invalid region name");
-			return FALSE;
-		}
-
-		/* read region */
-		flashrom_layout_set(self->flashctx, layout);
 
 		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_READ);
 		if (flashrom_image_read(self->flashctx, newcontents, flash_size)) {
@@ -164,7 +183,6 @@ fu_flashrom_device_write_firmware(FuDevice *device,
 				  GError **error)
 {
 	FuFlashromDevice *self = FU_FLASHROM_DEVICE(device);
-	struct flashrom_layout *layout;
 	gsize sz = 0;
 	gint rc;
 	const guint8 *buf;
@@ -182,25 +200,7 @@ fu_flashrom_device_write_firmware(FuDevice *device,
 		return FALSE;
 	buf = g_bytes_get_data(blob_fw, &sz);
 
-	if (flashrom_layout_read_from_ifd(&layout, self->flashctx, NULL, 0)) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_READ,
-				    "failed to read layout from Intel ICH descriptor");
-		return FALSE;
-	}
-
-	/* update only one specific region of the flash and do not touch others */
-	if (flashrom_layout_include_region(layout, fu_ifd_region_to_string(self->region))) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "invalid region name");
-		return FALSE;
-	}
-
 	/* write region */
-	flashrom_layout_set(self->flashctx, layout);
 	if (sz != fu_device_get_firmware_size_max(device)) {
 		g_set_error(error,
 			    FWUPD_ERROR,
@@ -226,7 +226,6 @@ fu_flashrom_device_write_firmware(FuDevice *device,
 		return FALSE;
 	}
 	fu_progress_step_done(progress);
-	flashrom_layout_release(layout);
 
 	/* Check if CMOS needs a reset */
 	if (fu_device_has_private_flag(device, FU_FLASHROM_DEVICE_FLAG_RESET_CMOS)) {
@@ -321,6 +320,10 @@ fu_flashrom_device_set_property(GObject *object,
 static void
 fu_flashrom_device_finalize(GObject *object)
 {
+	FuFlashromDevice *self = FU_FLASHROM_DEVICE(object);
+	if (self->layout != NULL)
+		flashrom_layout_release(self->layout);
+
 	G_OBJECT_CLASS(fu_flashrom_device_parent_class)->finalize(object);
 }
 
@@ -365,6 +368,7 @@ fu_flashrom_device_class_init(FuFlashromDeviceClass *klass)
 	klass_device->set_quirk_kv = fu_flashrom_device_set_quirk_kv;
 	klass_device->probe = fu_flashrom_device_probe;
 	klass_device->open = fu_flashrom_device_open;
+	klass_device->close = fu_flashrom_device_close;
 	klass_device->set_progress = fu_flashrom_device_set_progress;
 	klass_device->prepare = fu_flashrom_device_prepare;
 	klass_device->write_firmware = fu_flashrom_device_write_firmware;
--- a/plugins/flashrom/fu-flashrom-device.c
+++ b/plugins/flashrom/fu-flashrom-device.c
@@ -28,12 +28,11 @@ struct _FuFlashromDevice {
 	FuUdevDevice parent_instance;
 	FuIfdRegion region;
 	struct flashrom_flashctx *flashctx;
-	struct flashrom_programmer *flashprog;
 };
 
 G_DEFINE_TYPE(FuFlashromDevice, fu_flashrom_device, FU_TYPE_UDEV_DEVICE)
 
-enum { PROP_0, PROP_REGION, PROP_LAST };
+enum { PROP_0, PROP_FLASHCTX, PROP_REGION, PROP_LAST };
 
 static gboolean
 fu_flashrom_device_set_quirk_kv(FuDevice *device,
@@ -81,37 +80,6 @@ static gboolean
 fu_flashrom_device_open(FuDevice *device, GError **error)
 {
 	FuFlashromDevice *self = FU_FLASHROM_DEVICE(device);
-	gint rc;
-
-	if (flashrom_programmer_init(&self->flashprog, "internal", NULL)) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "programmer initialization failed");
-		return FALSE;
-	}
-	rc = flashrom_flash_probe(&self->flashctx, self->flashprog, NULL);
-	if (rc == 3) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "flash probe failed: multiple chips were found");
-		return FALSE;
-	}
-	if (rc == 2) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "flash probe failed: no chip was found");
-		return FALSE;
-	}
-	if (rc != 0) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "flash probe failed: unknown error");
-		return FALSE;
-	}
 
 	/* get the flash size from the device if not already been quirked */
 	if (fu_device_get_firmware_size_max(device) == 0) {
@@ -126,15 +94,6 @@ fu_flashrom_device_open(FuDevice *device, GError **error)
 		fu_device_set_firmware_size_max(device, flash_size);
 	}
 
-	return TRUE;
-}
-
-static gboolean
-fu_flashrom_device_close(FuDevice *device, GError **error)
-{
-	FuFlashromDevice *self = FU_FLASHROM_DEVICE(device);
-	flashrom_flash_release(self->flashctx);
-	flashrom_programmer_shutdown(self->flashprog);
 	return TRUE;
 }
 
@@ -326,6 +285,9 @@ fu_flashrom_device_get_property(GObject *object, guint prop_id, GValue *value, G
 {
 	FuFlashromDevice *self = FU_FLASHROM_DEVICE(object);
 	switch (prop_id) {
+	case PROP_FLASHCTX:
+		g_value_set_pointer(value, self->flashctx);
+		break;
 	case PROP_REGION:
 		g_value_set_uint(value, self->region);
 		break;
@@ -343,6 +305,9 @@ fu_flashrom_device_set_property(GObject *object,
 {
 	FuFlashromDevice *self = FU_FLASHROM_DEVICE(object);
 	switch (prop_id) {
+	case PROP_FLASHCTX:
+		self->flashctx = g_value_get_pointer(value);
+		break;
 	case PROP_REGION:
 		self->region = g_value_get_uint(value);
 		fu_device_set_logical_id(FU_DEVICE(self), fu_ifd_region_to_string(self->region));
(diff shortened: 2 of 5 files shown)


plugins/flashrom/fu-flashrom-device.c
```diff
@@ -309,6 +312,38 @@ fu_flashrom_device_constructed(GObject *obj)
...
+		self->region = g_value_dup_string(value);
+		break;
+	default:
+		G_OBJECT_WARN_INVALID_PROPERTY_ID(object, prop_id, pspec);
+		break;
+	}
+}
+
 static void
 fu_flashrom_device_finalize(GObject *object)
```
hughsie, line 348: don't you need a `g_free(self->region);` here too?

plugins/intel-spi/fu-intel-spi-device.c
```diff
@@ -358,6 +362,11 @@ fu_intel_spi_device_setup(FuDevice *device, GError **error)
 		fu_device_add_child(device, child);
 	}
 
+	/* publish HSFS value for anyone else to use */
+	me_is_locked = !(self->hsfs & (1 << HSFS_FDV_OFF))     /* assume locked if not valid */
+	             || (self->hsfs & (1 << HSFS_FDOPSS_OFF)); /* use status bit if valid */
+	fu_context_set_me_lock_status(fu_device_get_context(device), me_is_locked);
```
hughsie, line 368: rather than using the context, could we use something like https://github.com/fwupd/fwupd/blob/main/libfwupdplugin/fu-device-metadata.h#L10 and then get access to the device metadata using `vfuncs->device_registered` in the other plugins? It seems to me this is more of the state of a specific device rather than a whole system attribute, although I can certainty argue it the other way around too.

libfwupdplugin/fu-context.c
```diff
@@ -32,6 +32,7 @@ typedef struct {
 	FuLidState lid_state;
 	guint battery_level;
 	guint battery_threshold;
+	gboolean me_locked;
```
superm1, line 35: ME is only specific to Intel - this shouldn't be in the context for all of FuContextPrivate.

plugins/flashrom/fu-flashrom-device.c
SergiiDmytruk: Yes, thanks. And we need `g_free()` in  `fu_superio_device_finalize()` too. The lack of `g_free()` there confused me.

libfwupdplugin/fu-context.c
```diff
@@ -888,6 +889,20 @@ fu_context_class_init(FuContextClass *klass)
...
+	 * ME firmware update lock state.
+	 *
+	 * Since: 1.8.0
+	 */
+	pspec = g_param_spec_boolean("me-locked",
+				     NULL,
+				     NULL,
+				     FALSE,
+				     G_PARAM_READWRITE | G_PARAM_STATIC_NAME);
+	g_object_class_install_property(object_class, PROP_BATTERY_STATE, pspec);
```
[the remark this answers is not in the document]
superm1: OK, I think review is trending it shouldn't be a context property anyway but if it does end up to stay in the context drop it here unless we think a client will need it.

plugins/intel-spi/fu-intel-spi-device.c
superm1: You can look at what plugin the device comes from.  There should only be one device from the intel-spi plugin, right?

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -143,32 +145,65 @@ fu_plugin_flashrom_device_set_hwids(FuPlugin *plugin, FuDevice *device)
...
+			fu_device_add_vendor_id(FU_DEVICE(device), vendor_id);
+		}
+		fu_plugin_flashrom_device_set_version(plugin, device);
+		fu_plugin_flashrom_device_set_hwids(plugin, device);
+		fu_plugin_flashrom_device_set_bios_info(plugin, device);
+		if (!fu_device_setup(device, error))
+			return FALSE;
+	} else if (g_strcmp0(region, "me") == 0) {
+		// XXX: should there be a hierarchy of devices?
+		// XXX: always create device but mark as FWUPD_DEVICE_FLAG_LOCKED if ME is locked?
```
superm1, line 173: I think you need to use `FWUPD_DEVICE_FLAG_LOCKED` somwehere in this PR.  Probably on whatever device is to be locked/unlocked.

plugins/intel-spi/fu-intel-spi-device.c
SergiiDmytruk: > There should only be one device from the intel-spi plugin, right?

I think it's one device but with one or more children, one of which actually corresponds to ME entry in IFD if it's present.

superm1: Then my proposal was that you either add that attribute to the top level device and just look for the top level device from the plugin or you add it to a child and check for a device that has the metadata set.
If you end up needing a reference again in the plugin, you can always cache it locally in the plugin when you find it.

SergiiDmytruk: Resolved last hardware issue and can now test it properly. The child device even shows up in `get-devices --show-all-devices` as "Intel Management Engine", so I'll mark it with metadata. Thanks.

superm1: Another thing you can do is just use the locked device flag. Then no need to define a new metadata string.

plugins/superio/fu-superio-it55-device.c
```diff
@@ -47,6 +48,7 @@ struct _FuEcIt55Device {
 	FuSuperioDevice parent_instance;
 	gchar *prj_name;
 	AutoloadAction autoload_action;
+	gboolean me_locked;
```
hughsie, line 51: I think you need to print this in the to_string() vfunc -- e.g. https://github.com/fwupd/fwupd/blob/main/plugins/cfu/fu-cfu-device.c#L27

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -143,15 +151,21 @@ fu_plugin_flashrom_device_set_hwids(FuPlugin *plugin, FuDevice *device)
...
 static gboolean
-fu_plugin_flashrom_coldplug(FuPlugin *plugin, GError **error)
+fu_plugin_flashrom_add_device(FuPlugin *plugin, FuIfdRegion region, GError **error)
 {
+	g_autofree gchar *instance_id = NULL;
+
 	FuContext *ctx = fu_plugin_get_context(plugin);
 	const gchar *dmi_vendor;
-	g_autoptr(FuDevice) device = fu_flashrom_device_new(ctx);
+
```
hughsie, line 160: no blank line here please

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -143,15 +151,21 @@ fu_plugin_flashrom_device_set_hwids(FuPlugin *plugin, FuDevice *device)
...
+	g_autoptr(FuDevice) device = fu_flashrom_device_new(ctx, region);
 
 	fu_device_set_name(device, fu_context_get_hwid_value(ctx, FU_HWIDS_KEY_PRODUCT_NAME));
 	fu_device_set_vendor(device, fu_context_get_hwid_value(ctx, FU_HWIDS_KEY_MANUFACTURER));
 
+	instance_id = g_strdup_printf("IFD:%s", fu_ifd_region_to_string(region));
+	fu_device_add_instance_id(device, instance_id);
```
hughsie, line 167: This isn't really an instance ID. What's here's already (i.e. what's the output of `sudo fwupdtool get-devices --plugins flashrom`)? Could you do something like `fu_device_add_instance_str(device, "REGION", fu_ifd_region_to_string(region));` and then `fu_device_build_instance_id(device, NULL, "FLASHROM", "VENDOR", "PRODUCT", "REGION", NULL);` -- We might also have to define the VENDOR and PRODUCT with `fu_device_add_instance_str` the same way -- and then we need it defined in the plugins/flashrom/README.md file....

SergiiDmytruk: > This isn't really an instance ID.

Right, sorry, I'm confused about numerous IDs. Read relevant section of the tutorial and sources to improve my understanding.

> This isn't really an instance ID. What's here's already (i.e. what's the output of `sudo fwupdtool get-devices --plugins flashrom`)?

If I update `flashrom.quirk`, I get:
```
$ sudo fwupdtool get-devices --plugins flashrom
NS50MU
│
└─NS50MU:
      Device ID:          84bf22110f952e8493cc6ed99039b756d4129b5c
      Current version:    1.0
      Vendor:             Notebook (DMI:3mdeb Embedded Systems Consulting)
      GUIDs:              230c8b18-8d9b-53ec-838b-6cfc0383493a _ main-system-firmware
                          dc0de287-201b-5240-b8d2-0e5cead707ea _ Notebook&Not Applicable&NS50MU&Not Applicable&Notebook&NS50MU
                          8da9fe2b-90f4-5028-9bd9-c692b107d4ea _ Notebook&Not Applicable&NS50MU&Not Applicable
                          0039fec6-9ecf-5921-9961-905e148e052a _ Notebook&Not Applicable&NS50MU
                          6b26127f-384f-5ecc-9e94-02359da05f54 _ Notebook&Not Applicable&Notebook&NS50MU
                          58b02b86-65a5-56b6-bf24-fd9767043c58 _ Notebook&Not Applicable&NS50MU&Not Applicable&3mdeb Embedded Systems Consulting
      Device Flags:       _ Internal device
                          _ Updatable
                          _ System requires external power source
                          _ Needs shutdown after installation
```

This isn't stock BIOS, but coreboot ported for these laptops:

```
$ fwupdtool hwids
Computer Information
--------------------
BiosVendor: 3mdeb Embedded Systems Consulting
BiosVersion: Dasharo (coreboot+UEFI) v1.0.0
Manufacturer: Notebook
Family: Not Applicable
ProductName: NS50MU
ProductSku: Not Applicable
EnclosureKind: a
BaseboardManufacturer: Notebook
BaseboardProduct: NS50MU

Hardware IDs
------------
not available as 'BiosMajorRelease' unknown
not available as 'BiosMajorRelease' unknown
not available as 'BiosMajorRelease' unknown
{60f53465-e8fc-5122-b79b-f7b03f063037}   <- Manufacturer + Family + ProductName + ProductSku + BaseboardManufacturer + BaseboardProduct
{99a74816-5cd4-5166-b844-df3ef29e7300}   <- Manufacturer + Family + ProductName + ProductSku
{00bb8e01-c72a-5906-aa0e-2279c54ba369}   <- Manufacturer + Family + ProductName
{b05c067a-a96f-591b-9c69-7dda03324e00}   <- Manufacturer + ProductSku + BaseboardManufacturer + BaseboardProduct
{cd9d0da1-e88d-5d36-aae5-2ba60e8d0ed6}   <- Manufacturer + ProductSku
{f42b9241-6cdc-55cb-bb62-393a52752fc8}   <- Manufacturer + ProductName + BaseboardManufacturer + BaseboardProduct
{b7b416f9-55aa-5fd4-bff7-fbaa4143bd92}   <- Manufacturer + ProductName
{b05c067a-a96f-591b-9c69-7dda03324e00}   <- Manufacturer + Family + BaseboardManufacturer + BaseboardProduct
{cd9d0da1-e88d-5d36-aae5-2ba60e8d0ed6}   <- Manufacturer + Family
{a127c2d7-5537-5b72-8bc1-9ccd683f18e6}   <- Manufacturer + EnclosureKind
{774c2cef-39d8-5b40-9477-c220b4fc1f04}   <- Manufacturer + BaseboardManufacturer + BaseboardProduct
{56bb94fc-caf1-5863-9e85-1e449a0c5a01}   <- Manufacturer
```

> and then we need it defined in the plugins/flashrom/README.md file....

You mean describe structure of instance ID there?

SergiiDmytruk:
By the way, there is an alternative method for unlocking ME region of these laptops: holding <kbd>Fn + M</kbd> when pressing power on button.

So there is also an option of just giving a user instructions like that before updating ME firmware. This makes changes to SuperIO plugin unnecessary and gets rid of unlocking of EC device to get to ME region, which looks a bit weird.

Which way of unlocking would you prefer?

hughsie:
I'm not a fan of the manual action, but I do like the simplicity.

superm1:
From a purely security oriented view, should you really be able to allow to turn off ME from software in an operating system? I feel like at least requiring physical interaction is a feature.

SergiiDmytruk:
Changed to manual unlocking to see how it might look. It caused addition of some new code in flashrom plugin to make it possible, so not pure simplification. Changes in superio plugin are rather independent, so both ways of unlocking can even be provided at the same time.

> From a purely security oriented view, should you really be able to allow to turn off ME from software in an operating system? I feel like at least requiring physical interaction is a feature.

An argument, but If kernel isn't locked down, there is a long list of things that can be done to the system besides ME unlocking.

superm1:
>Changed to manual unlocking to see how it might look. It caused addition of some new code in flashrom plugin to make it possible, so not pure simplification

OK, will look through it.

> An argument, but If kernel isn't locked down, there is a long list of things that can be done to the system besides ME unlocking.

Sure; but the ME unlocking is a permanent problem.  So if userspace software can accomplish this, that means malware can accomplish it.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -183,11 +271,39 @@ fu_plugin_flashrom_startup(FuPlugin *plugin, GError **error)
...
+static gboolean
+fu_plugin_flashrom_unlock(FuPlugin *self, FuDevice *device, GError **error)
+{
+	FuFlashromDevice *flashrom_device = FU_FLASHROM_DEVICE(device);
+
+	if (fu_flashrom_device_get_region(flashrom_device) == FU_IFD_REGION_ME &&
+	    fu_flashrom_device_has_fn_m_me_unlock(flashrom_device)) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
```
superm1, line 283: maybe this should be NOTHING_TO_DO?

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -183,11 +271,39 @@ fu_plugin_flashrom_startup(FuPlugin *plugin, GError **error)
...
+				    "\n"
+				    "ME region on Tuxedo laptops should be unlocked manually:\n"
+				    " 1. Power off your laptop\n"
+				    " 2. Press and keep holding Fn + M during the next step\n"
+				    " 3. Press power on button");
```
superm1, line 288: I suspect @hughsie won't be a fan of hardcoded error strings, this means translators won't be able to help in translations.  But this is a pretty advanced use case.  Maybe it is reasonable.

If the strings stay, presumably you don't want to include the words "Tuxedo" or "laptop" to future proof the string.  That way other models may support it, and other form factors may support it.  You can move tuxdeo and laptop into a comment near by.

superm1 (dismissed):
Sort out the pre-commit problem, and then this approach LGTM overall.
I do think you should remove the unlock interface from the EC too for security of the system, but that is up to you.

plugins/flashrom/fu-plugin-flashrom.c
SergiiDmytruk: Right, thanks.

plugins/flashrom/fu-plugin-flashrom.c
SergiiDmytruk: I see only one use of translatable strings in plugins:
https://github.com/fwupd/fwupd/blob/7ec0f11e6925fd59f3a1a8f177b90de54f40a300/plugins/uefi-capsule/fu-plugin-uefi-capsule.c#L447
And it seems to not be used... I'll add `_(` and `TRANSLATORS:` comment as it seems to be supported in plugins.

Another thing is that verbose instructions are given via an error message, which is sort of a misuse, but I didn't see another way of doing it.

hughsie: Na, no translations in plugins please. Ideally it'd be an enumerated fwupdrequest type, but that's hard for this kind of thing.

SergiiDmytruk:
> Sort out the pre-commit problem, and then this approach LGTM overall.

That's just formatting. I'll run `clang-format` when marking PR as ready for review.

> I do think you should remove the unlock interface from the EC too for security of the system, but that is up to you.

I did. Just wanted to mention that it can be brought back if needed.

By the way, I noticed a weird difference between outputs of `fwupdtool get-devices --plugins flashrom` and `fwupdmgr get-devices --show-all-devices`. Output of the first command doesn't have these lines:
```
                    ? Device is locked
                    ? Cryptographic hash verification is available
```
Is this expected or a bug? At first glance, the same code is used in both cases.

superm1:
> That's just formatting. I'll run clang-format when marking PR as ready for review.

OK

>I did. Just wanted to mention that it can be brought back if needed.

Right I see it's gone here; but I think you should remove it from the EC codebase too.  That's of course out of my control though :)

>Is this expected or a bug? At first glance, the same code is used in both cases.

That doesn't sound expected to me.

SergiiDmytruk:
> but I think you should remove it from the EC codebase too

That's developed by Clevo, not Tuxedo or 3mdeb :) We're making it possible to update EC using open source tools and also [ported](https://docs.dasharo.com/variants/tuxedo_ibs15/releases/) coreboot to these laptops.

> That doesn't sound expected to me.

It's fine after all. `--plugins` is actually `--plugin-whitelist`, so no plugin interactions on `--plugins flashrom` and result is different.

superm1:
Okay no concerns left for me.

plugins/flashrom/fu-flashrom-device.h
```diff
@@ -13,4 +13,10 @@
...
 
 FuDevice *
-fu_flashrom_device_new(FuContext *ctx);
+fu_flashrom_device_new(FuContext *ctx, FuIfdRegion region);
+
+FuIfdRegion
+fu_flashrom_device_get_region(FuFlashromDevice *self);
+
+gboolean
+fu_flashrom_device_has_fn_m_me_unlock(FuFlashromDevice *self);
```
hughsie, line 22: I don't like this much, can you just do `fu_device_has_private_flag(device, FU_FLASHROM_DEVICE_FLAG_FN_M_ME_UNLOCK);` in the caller please. You can move the `FU_FLASHROM_DEVICE_FLAG_FN_M_ME_UNLOCK` to header if that helps.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -143,14 +143,41 @@ fu_plugin_flashrom_device_set_hwids(FuPlugin *plugin, FuDevice *device)
...
 
-	fu_device_set_name(device, fu_context_get_hwid_value(ctx, FU_HWIDS_KEY_PRODUCT_NAME));
-	fu_device_set_vendor(device, fu_context_get_hwid_value(ctx, FU_HWIDS_KEY_MANUFACTURER));
+	/* add this so we can attach all the other quirks */
+	fu_device_add_instance_str(FU_DEVICE(device), "GUID", guid);
+	if (!fu_device_build_instance_id(FU_DEVICE(device), error, "FLASHROM", "GUID", NULL)) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "flashrom failed to add GUID instance id for %s device", region_str);
```
hughsie, line 178: you don't need this, fu_device_build_instance_id will set the error for you.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -143,14 +143,31 @@ fu_plugin_flashrom_device_set_hwids(FuPlugin *plugin, FuDevice *device)
...
+
+	fu_device_add_instance_str(device, "REGION", region_str);
+	fu_device_add_instance_str(device, "VENDOR", vendor);
+	fu_device_add_instance_str(device, "PRODUCT", product);
+	if (!fu_device_build_instance_id(device, NULL, "FLASHROM", "VENDOR", "PRODUCT", "REGION",
+					 NULL)) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "flashrom failed to add instance id for %s device", region_str);
```
hughsie, line 168: just set the error for fu_device_build_instance_id please

plugins/flashrom/fu-plugin-flashrom.c
SergiiDmytruk: Sure, thanks, varargs confused me.

SergiiDmytruk:
Noticeable change apart from addressing comments: device for ME partition isn't created unless `intel_spi` creates its corresponding device. If depending on `intel_spi` in this way is undesirable, can check for the region via flashrom by creating a context, loading layout from IFD and checking if including `me` region fails (no "is_present" function). Which makes me wonder whether flashrom context should be managed by the plugin instead of device object.

hughsie (dismissed):

hughsie:
> whether flashrom context should be managed by the plugin

Yes, this sounds right.

SergiiDmytruk:
Moved context management and also fixed leak of layout. One thing to note: operations depend on the state of the context, so if it's possible to use two devices at the same time, this change shouldn't be accepted.

plugins/flashrom/fu-flashrom-device.c
```diff
@@ -23,12 +23,12 @@ struct _FuFlashromDevice {
 	FuUdevDevice parent_instance;
 	FuIfdRegion region;
 	struct flashrom_flashctx *flashctx;
-	struct flashrom_programmer *flashprog;
+	struct flashrom_layout *layout;
```
hughsie, line 26: Why do we need to store the flashrom_layout in the device? It doesn't seem to me we use it anywhere.

hughsie: Ohh i see, we use it during ->open() to set on the context; ignore me.

hughsie: Nope, we clear it during ->close() -- I think my original point holds.

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -142,15 +161,40 @@ fu_plugin_flashrom_device_set_hwids(FuPlugin *plugin, FuDevice *device)
...
+	fu_device_add_instance_str(device, "VENDOR", vendor);
+	fu_device_add_instance_str(device, "PRODUCT", product);
+	if (!fu_device_build_instance_id(device,
+					 error,
+					 "FLASHROM",
+					 "VENDOR",
+					 "PRODUCT",
+					 "REGION",
+					 NULL))
+		return NULL;
```
hughsie, line 192: I think this needs documenting in README.md

plugins/flashrom/fu-plugin-flashrom.c
```diff
@@ -205,15 +286,76 @@ fu_plugin_flashrom_startup(FuPlugin *plugin, GError **error)
...
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "flash probe failed: unknown error");
+		return FALSE;
+	}
+
 	return TRUE;
 }
 
+static gboolean
+fu_plugin_flashrom_unlock(FuPlugin *self, FuDevice *device, GError **error)
```
hughsie, line 325: out of interest; why here and not in fu-flashrom-device.c?

plugins/flashrom/fu-flashrom-device.c
SergiiDmytruk: I just needed to store it somewhere so it doesn't leak. `flashrom_layout_set()` doesn't transfer ownership:
```c
/**
 * @brief Set the active layout for a flash context.
 *
 * Note: This just sets a pointer. The caller must not release the layout
 *       as long as he uses it through the given flash context.
 *
 * @param flashctx Flash context whose layout will be set.
 * @param layout   Layout to bet set.
 */
void flashrom_layout_set(struct flashrom_flashctx *const flashctx, const struct flashrom_layout *const layout)
{
	flashctx->layout = layout;
}
```

And there is no `flashrom_layout_get()`, so you can't extract it from the context to free later.

plugins/flashrom/fu-plugin-flashrom.c
SergiiDmytruk: Unlock operation is defined only at the plugin level, at least I don't see corresponding function for a device. Or do you suggest to create `fu_flashrom_device_unlock()` and call it here?

plugins/flashrom/fu-flashrom-device.c
hughsie: > flashrom_layout_set() doesn't transfer ownership

Bleh. Okay, thanks for the explanation. Could you put that rationale in the source code please; one line is fine.

plugins/flashrom/fu-plugin-flashrom.c
hughsie: > to create fu_flashrom_device_unlock() and call it here

Yes, I think that's a good idea. Then if we do a FuDevice->unlock() vfunc in the future we can connect it up easily.

plugins/flashrom/README.md
```diff
@@ -45,6 +48,16 @@ Internal device uses hardware ID values which are derived from SMBIOS.
...
 
+One more GUID has the following form:
+
+* `FLASHROM\VENDOR_{manufacturer}&PRODUCT_{product}&REGION_{ifd_region_name}`
+
+Its purpose is to target specific regions of the flash as defined by IFD (Intel
+SPI Flash Descriptor), examples:
+
+* `FLASHROM\VENDOR_Notebook&PRODUCT_NS50MU&REGION_bios`
+* `FLASHROM\VENDOR_Notebook&PRODUCT_NS50MU&REGION_me`
```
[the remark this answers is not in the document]
hughsie: on the REGION only that is.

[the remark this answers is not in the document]
SergiiDmytruk: Probably a good idea to not depend on the case here. Updated.

PR #4518, opened 2022-04-12
dfu: modified VID/PID for atlas camera.

Jierger:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/dfu.quirk
+++ b/plugins/dfu/dfu.quirk
@@ -431,7 +431,7 @@ Flags = manifest-poll,allow-zero-polltimeout,unsigned-payload
 RemoveDelay = 9000
 
 # AVer ATLAS CAM
-[USB\VID_2574&PID_0A70]
+[USB\VID_34AD&PID_0006]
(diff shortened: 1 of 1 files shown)


plugins/dfu/dfu.quirk
```diff
@@ -431,7 +431,7 @@ Flags = manifest-poll,allow-zero-polltimeout,unsigned-payload
 RemoveDelay = 9000
 
 # AVer ATLAS CAM
-[USB\VID_2574&PID_0A70]
+[USB\VID_34AD&PID_0006]
```
superm1, line 434: Should both IDs be added or is this some circumstance that only the new one is needed?

superm1 (approved):

plugins/dfu/dfu.quirk
hughsie: If you want to add both that's fine; just do a new PR and we can have the old and new ones.

PR #4519, opened 2022-04-12
Automatically obsolete negative duplicate attributes

superm1:
To a user there is no point to explicitly showing two failures for
encrypted RAM if the system supports SME and TSME both.

To show them a single failure, de-duplicate the attributes by appstream
ID during depsolve.

If the plugin didn't explicitly set obsoletes and two attributes are both
failures and both share the same appstream ID obsolete one of them.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-security-attrs.c
+++ b/libfwupdplugin/fu-security-attrs.c
@@ -329,44 +329,60 @@ fu_security_attrs_sort_cb(gconstpointer item1, gconstpointer item2)
 void
 fu_security_attrs_depsolve(FuSecurityAttrs *self)
 {
-	g_autoptr(GHashTable) attrs_by_id = NULL;
-
 	g_return_if_fail(FU_IS_SECURITY_ATTRS(self));
-
-	/* make hash of ID -> object */
-	attrs_by_id = g_hash_table_new(g_str_hash, g_str_equal);
-	for (guint i = 0; i < self->attrs->len; i++) {
-		FwupdSecurityAttr *attr = g_ptr_array_index(self->attrs, i);
-		g_hash_table_insert(attrs_by_id,
-				    (gpointer)fwupd_security_attr_get_appstream_id(attr),
-				    (gpointer)attr);
-	}
 
 	/* set flat where required */
 	for (guint i = 0; i < self->attrs->len; i++) {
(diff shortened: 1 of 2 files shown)


hughsie:
Looks good, but can you add a couple of testcases to fu-self-tests.c please -- this is exactly the kind of thing I want to make sure works in both cases.

libfwupdplugin/fu-security-attrs.c
```diff
@@ -329,44 +329,67 @@ fu_security_attrs_sort_cb(gconstpointer item1, gconstpointer item2)
...
+
+			/* skip self */
+			if (g_strcmp0(fwupd_security_attr_get_plugin(attr),
+				      fwupd_security_attr_get_plugin(attr_tmp)) == 0 &&
+			    g_strcmp0(fwupd_security_attr_get_appstream_id(attr),
+				      fwupd_security_attr_get_appstream_id(attr_tmp)) == 0)
+				continue;
+
+			/* look for duplicate (negative) attributes when obsolete not explicitly set
+			 */
```
hughsie, line 350: `add duplicate (negative) attributes when obsolete not explicitly set` might be more accurate

hughsie (approved):
Two nits, then LGTM, thanks.

libfwupdplugin/fu-security-attrs.c
```diff
@@ -329,44 +329,67 @@ fu_security_attrs_sort_cb(gconstpointer item1, gconstpointer item2)
...
-			if (attr_tmp != NULL) {
-				g_debug("security attr %s obsoleted by %s",
-					obsolete,
-					fwupd_security_attr_get_appstream_id(attr_tmp));
-				fwupd_security_attr_add_flag(attr_tmp,
-							     FWUPD_SECURITY_ATTR_FLAG_OBSOLETED);
+
+		for (guint j = 0; j < self->attrs->len; j++) {
+			FwupdSecurityAttr *attr_tmp = g_ptr_array_index(self->attrs, j);
+
```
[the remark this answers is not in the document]
superm1: fixed

libfwupdplugin/fu-security-attrs.c
superm1: fixed

PR #4525, opened 2022-04-14
Add FuProgress to ->prepare and ->cleanup

hughsie:
Some plugins were creating local versions (which were not attached to
the daemon progress in any way) as a workaround as they needed to do
actions that took a long time to complete.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


meson.build
```diff
@@ -52,7 +52,7 @@ libfwupd_lt_current = '2'
 libfwupd_lt_revision = '0'
 libfwupd_lt_age = '0'
 libfwupd_lt_version = '@0@.@1@.@2@'.format(libfwupd_lt_current, libfwupd_lt_age, libfwupd_lt_revision)
-libfwupdplugin_lt_current = '5'
+libfwupdplugin_lt_current = '6'
```
superm1, line 55: need to bump contrib/debian/control.in too (s/libfwupdplugin5/libfwupdplugin6/).

hughsie: Fixed.

PR #4527, opened 2022-04-14
Wip/hughsie/fu backend invalidate

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-backend.c
+++ b/libfwupdplugin/fu-backend.c
@@ -23,13 +23,14 @@ typedef struct {
 	gchar *name;
 	gboolean enabled;
 	gboolean done_setup;
+	gboolean can_invalidate;
 	GHashTable *devices; /* device_id : * FuDevice */
 	GThread *thread_init;
 } FuBackendPrivate;
 
 enum { SIGNAL_ADDED, SIGNAL_REMOVED, SIGNAL_CHANGED, SIGNAL_LAST };
 
-enum { PROP_0, PROP_NAME, PROP_CONTEXT, PROP_LAST };
+enum { PROP_0, PROP_NAME, PROP_CAN_INVALIDATE, PROP_CONTEXT, PROP_LAST };
 
 static guint signals[SIGNAL_LAST] = {0};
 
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-backend.c
```diff
@@ -123,6 +123,35 @@ fu_backend_registered(FuBackend *self, FuDevice *device)
...
+fu_backend_invalidate(FuBackend *self)
+{
+	FuBackendClass *klass = FU_BACKEND_GET_CLASS(self);
+	FuBackendPrivate *priv = GET_PRIVATE(self);
+
+	g_return_if_fail(FU_IS_BACKEND(self));
+
+	priv->done_setup = FALSE;
+	if (klass->invalidate != NULL)
+		klass->invalidate(self);
```
superm1, line 152: I think you probably want to prohibit certain backend types from this.  Like the built in ones shouldn't be doing this.

hughsie: > the built in ones shouldn't be doing this

Why not?

superm1: What happens when a different plugin relies on it? If one USB plugin invalidates the USB backend and it takes down  the other USB devices too.

hughsie: Hmm, fair point. What about if fu_backend_device_added() saw more than one kind of plugin it would prevent invalidate (and set a GError?). Or maybe we just want a fu_backend_add_flag(backend, FU_BACKEND_FLAG_CAN_INVALIDATE) or something? Ideas welcome.

superm1: What's the use case? I don't think I follow the background.

superm1: > Or maybe we just want a fu_backend_add_flag(backend, FU_BACKEND_FLAG_CAN_INVALIDATE)

I think that's a reasonable solution.  Then if we find a need to modify other backends can think them through very closely.

hughsie: I'll just move it to the redfish plugin for now; we can do the complicated generic thing if any other backends need this functionality...

hughsie: Gah, I need the priv->done_setup = FALSE clearing, which needs to be in fu-backend.c. Back to plan A.

hughsie: Okay, there's now a construction-only property backends have to set for safety.

superm1 (approved):

Issue #4524, opened 2022-04-14
Add a list of trusted users


lancer1268:
**Is your feature request related to a problem? Please describe.**
When the `ShowDevicePrivate` is enabled, non-root users can see the `serials` and `InstanceIds`.
This is useful but the configuration is all-or-none. A finer control, like an allowed-list-based configuration, is desirable.

**Describe the solution you'd like**
Add an array item, e.g. `TrustedUids`, to `daemon.conf` so that we can specify the list of trusted users.
The "trusted" here means those who are granted the access to some device-private like serials and InstanceIds.

**Describe alternatives you've considered**
Use group id instead of user id. However, it seems that the `sender` in dbus cannot be converted to group id directly.
If the group id approach is accessible, using group id should be a better approach.

**Additional context**
In `fu_main_create_request`, the `FWUPD_DEVICE_FLAG_TRUSTED` flag is added if `calling_uid == 0`.
This means that only the root user is trusted. It would be helpful if that is configurable.

PR #4528, opened 2022-04-14
Add support for a new TrustedUids key

superm1:
This key is used to specify that a dedicated user runs the fwupd
client process and sensitive strings such as the serial number should
be shared with the calling process.

(Fixes: #4524)

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-config.c
+++ b/src/fu-config.c
@@ -30,6 +30,7 @@ struct _FuConfig {
 	GPtrArray *blocked_firmware;  /* (element-type utf-8) */
 	GPtrArray *uri_schemes;	      /* (element-type utf-8) */
 	GPtrArray *filenames;	      /* (element-type utf-8) */
+	GArray *trusted_uids;	      /* (elementy type guint64) */
 	guint64 archive_size_max;
 	guint idle_timeout;
 	gchar *host_bkc;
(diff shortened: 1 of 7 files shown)


src/fu-config.c
```diff
@@ -241,6 +243,19 @@ fu_config_reload(FuConfig *self, GError **error)
...
+	/* get trusted uids */
+	g_ptr_array_set_size(self->trusted_uids, 0);
+	uids = g_key_file_get_string_list(keyfile,
+					  "fwupd",
+					  "TrustedUids",
+					  NULL, /* length */
+					  NULL);
+	if (uids != NULL) {
+		for (guint i = 0; uids[i] != NULL; i++) {
+			g_ptr_array_add(self->trusted_uids, g_strdup(uids[i]));
```
hughsie, line 255: I think it makes sense to do the fu_common_strtoull() here and return a GArray of guint32 (or is guint64 more appropriate?) rather than doing all the string parsing again and again.

superm1: fixed

hughsie:
Windows gets upset:

    ../src/fu-engine.c:5774:42: error: unknown type name 'uid_t'
     5774 | fu_engine_is_uid_trusted(FuEngine *self, uid_t calling_uid)

maybe just use guint64 in anything not ifdef'd? Then LGTM.

PR #4531, opened 2022-04-15
redfish: Restart the BMC after installing BCM updates

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/redfish/fu-plugin-redfish.c
+++ b/plugins/redfish/fu-plugin-redfish.c
@@ -14,8 +14,12 @@
 
 #include "fu-redfish-backend.h"
 #include "fu-redfish-common.h"
+#include "fu-redfish-device.h"
 #include "fu-redfish-network.h"
 #include "fu-redfish-smbios.h"
+
+#define FU_REDFISH_PLUGIN_CLEANUP_RETRIES_COUNT 120
+#define FU_REDFISH_PLUGIN_CLEANUP_RETRIES_DELAY 5000 /* ms */
 
 struct FuPluginData {
 	FuRedfishBackend *backend;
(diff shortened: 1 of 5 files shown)


plugins/redfish/fu-plugin-redfish.c
```diff
@@ -414,6 +418,70 @@ fu_plugin_redfish_startup(FuPlugin *plugin, GError **error)
...
+				  fu_plugin_redfish_cleanup_cb,
+				  FU_REDFISH_PLUGIN_CLEANUP_RETRIES_COUNT,
+				  FU_REDFISH_PLUGIN_CLEANUP_RETRIES_DELAY,
+				  self,
+				  error)) {
+		g_prefix_error(error, "manager failed to come back: ");
+		return FALSE;
+	}
+
+	/* work around an XCC bug... */
```
superm1, line 480: Link to bug upstream?

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -414,6 +418,70 @@ fu_plugin_redfish_startup(FuPlugin *plugin, GError **error)
...
+	g_usleep(40 * G_USEC_PER_SEC);
+	return fu_plugin_redfish_coldplug(self, error);
```
superm1, line 482: Why shouldn't this be a detach ? Seems a long time for a plugin to hang.

hughsie: Attach is supposed to put one device into a specific mode, pushing it out of programmer mode or locking it again in some way. The ->detach() and ->attach() vfuncs are only called on the specific device we're updating, and there's no provision for calling it on another device. For instance it might be valid to want to restart the BMC after you've installed a power supply update for instance -- or to restart the mouse main firmware when installing a mouse radio update. For this I think the composite (all plugins) ->cleanup() is the only thing that makes sense. I do agree we need to use the now-passed-in FuProgress to actually set up the timeouts and provide some feedback to the user.

plugins/redfish/fu-plugin-redfish.c
hughsie: I've pinged Lenovo about some of the reasons for the big delay; there's some wonkyness in the BMC replies too.

superm1: yeah at least leave some comments on the outcome of this or it will be difficult to remember why +40s.  It might be best to add a plugin quirk for it too.

plugins/redfish/fu-plugin-redfish.c
superm1: I think the source of my confusion is this is run from **plugin** cleanup not **composite** cleanup.  Thanks for explaining.

plugins/redfish/fu-plugin-redfish.c
```diff
@@ -414,6 +418,70 @@ fu_plugin_redfish_startup(FuPlugin *plugin, GError **error)
...
+	/* remove all the devices */
+	devices = fu_backend_get_devices(FU_BACKEND(data->backend));
+	for (guint i = 0; i < devices->len; i++) {
+		FuDevice *device_tmp = g_ptr_array_index(devices, i);
+		fu_backend_device_removed(FU_BACKEND(data->backend), device_tmp);
+	}
+
+	/* wait for the BMC to come back */
+	fu_backend_invalidate(FU_BACKEND(data->backend));
+	g_usleep(10 * 1000);
```
superm1, line 469: `s/1000/G_USEC_PER_SEC/` perhaps?

hughsie: `fu_progress_sleep()` is even better. :)

plugins/redfish/fu-plugin-redfish.c
hughsie: Done and done.

superm1 (approved):

PR #4535, opened 2022-04-18
trivial: bios: Add an attribute to fail HSI in legacy BIOS mode

superm1:
Fixes: #4530

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/bios/fu-plugin-bios.c
+++ b/plugins/bios/fu-plugin-bios.c
@@ -53,10 +53,28 @@ fu_plugin_bios_coldplug(FuPlugin *plugin, GError **error)
 	return TRUE;
 }
 
+static void
+fu_plugin_bios_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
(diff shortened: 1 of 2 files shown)


plugins/bios/fu-plugin-bios.c
```diff
@@ -53,10 +53,30 @@ fu_plugin_bios_coldplug(FuPlugin *plugin, GError **error)
 	return TRUE;
 }
 
+static void
+fu_plugin_bios_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
+{
+#ifdef HAVE_HSI
```
hughsie, line 59: LGTM, thanks. Out of interest, why the #ifdef? If there's no HSI then the vfunc just won't ever be called -- it might be best to just leave it compiled in.

superm1: Good point, I'll drop it.

PR #4538, opened 2022-04-19
trivial: Make the language in the stale job more pathetic

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -11,7 +11,17 @@ staleLabel: wontfix
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. It will be closed if no further activity occurs.
+  Please note: We are just a few people who contribute to a shared project,
+  and it's impossible for us to fix every bug with such limited resources.
+  
+  If you want to investigate and try to help solve this yourself, we will
+  review all pull requests from new contributors.
+  
+  If this is issue is important to you for your business please talk with
+  your technical account manager about arranging resources to solve this issue.
+  You might even consider hiring someone to write the code if you're unable
+  to do so yourself, e.g. see: https://fwupd.org/lvfs/docs/consulting
+
 # Comment to post when closing a stale issue. Set to `false` to disable
 closeComment: false


.github/stale.yml
```diff
@@ -11,7 +11,14 @@ staleLabel: wontfix
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. It will be closed if no further activity occurs.
+  Please note: We are just a few people who contribute to a shared project,
+  it's impossible for us to fix every bug with such limited resources.
+  If this is important to you for your business then please talk with your
+  Red Hat, Canonical or SUSE technical account manager and file a bug in
```
superm1, line 18: I think rather than enumerate distros, bug trackers and priorities, it should be something like this.
"If this is important to you for your business please talk with your technical account manager about arranging resources to solve this issue".

.github/stale.yml
```diff
@@ -11,7 +11,14 @@ staleLabel: wontfix
...
+  bugzilla so it can be prioritized accordingly.
+  You might even consider hiring someone from Collabora to write the code
```
superm1, line 20: Collabora is great, but how about instead point people to https://fwupd.org/lvfs/docs/consulting as there are plenty of consulting services that can help?

.github/stale.yml
```diff
@@ -11,7 +11,14 @@ staleLabel: wontfix
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
-  recent activity. It will be closed if no further activity occurs. Thank you
-  for your contributions.
+  recent activity. It will be closed if no further activity occurs.
+  Please note: We are just a few people who contribute to a shared project,
+  it's impossible for us to fix every bug with such limited resources.
```
superm1, line 16: I think that in here there needs to be something about people trying to submit PRs.  Maybe
"If you want to investigate and try to help solve this yourself, we encourage and will review all pull requests from new contributors"

.github/stale.yml
hughsie: Ack, thanks.

.github/stale.yml
hughsie: Good idea.

.github/stale.yml
hughsie: Got it, how about that?

superm1 (approved):

Issue #4540, opened 2022-04-21
Progress message is not updated properly when a device is waiting for replug


chlee75:
**Describe the bug**
Issue 1. Progress bar message (e.g. Less than one minute remaining...) used in the writing process remains in the waiting process in the gnome-firmware or fwupmgr.

![image](https://user-images.githubusercontent.com/32348354/164357546-603f492e-4082-4d11-8d0a-831279d8e816.png)

Issue 2.  The progress bar message in the writing process is not consistent.
  Sometimes it ends with "Writing..., Less than one minute remaining..."  
  Sometimes it ends with "Writing..., 17 minutes remaining..."  
  Sometimes it ends with "Writing..., 1 minute remaining..."  

**Steps to Reproduce**
 1) Connect HP G5 or G4 dock to your computer.
 2) Open gnome-firmware
 3) Start firmware upgrade or reinstall or downgrade
 4) The Progress bar message is changed from "Writing..., Less than one minute remaining..." to "Waiting..., Less than one minute remaining..."  
    or 
   The progress bar message is changed from "Writing..., 17 minutes remaining..." to "Waiting..., 17 minutes remaining..." 
  
**Expected behavior**
For Issue 1. The Progress bar message (e.g. Less than one minute remaining...) used in the writing process should not be displayed in the waiting process.

For Issue 2. The Progress bar message in the writing process should be consistent.

**fwupd version information**
fwupd version is 1.8.0 and gnome-firmware version is 42.1

**fwupd device information**
plugin is ccgx and device is HP USB-C Dock G5 or HP Thunderbolt Dock G4.

chlee75:
@hughsie 

"Less than one minute remaining... " message also remained when I use fwupdmgr command.
Waiting…                 [***************************** ] Less than one minute remaining...

I think that fwupd seems to be modified.

And Let me know how to test the new gnome-firmware.

PR #4541, opened 2022-04-22
Only show the CLI time remaining for predictable status phases

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4540

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-progressbar.c
+++ b/src/fu-progressbar.c
@@ -104,6 +104,22 @@ fu_progressbar_erase_line(FuProgressbar *self)
 }
 
 static gboolean
+_fu_status_is_predictable(FwupdStatus status)
+{
+	if (status == FWUPD_STATUS_DEVICE_ERASE)
+		return TRUE;
+	if (status == FWUPD_STATUS_DEVICE_VERIFY)
+		return TRUE;
+	if (status == FWUPD_STATUS_DEVICE_READ)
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Isn't the time to wait also predictable though? You know how long replug will take.

hughsie:
> You know how long replug will take

We know the maximum time it will take, but it's typically a *lot* smaller. e.g. we allow 10 seconds for the OS to re-enumerate a USB device, but typically it's less than 500ms.

superm1:
> > You know how long replug will take
> 
> We know the maximum time it will take, but it's typically a *lot* smaller. e.g. we allow 10 seconds for the OS to re-enumerate a USB device, but typically it's less than 500ms.

How about then "waiting for up to XX seconds for device to return"

hughsie:
> How about then "waiting for up to XX seconds for device to return"

I'm not sure that's terribly useful for the end user. Maybe for the developer, e.g. in --verbose, but not in the main UI.

superm1:
> > How about then "waiting for up to XX seconds for device to return"
> 
> I'm not sure that's terribly useful for the end user. Maybe for the developer, e.g. in --verbose, but not in the main UI.

okay

PR #4542, opened 2022-04-22
Wip/hughsie/ci

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Wait, why was this merged when CI still failed?

hughsie:
I was trying to regenerate the daily containers so that we could fix the CI, but retriggering the container generation seemed to use the checkout from before, rather than the new contents. I'm a bit at a loss on how to force the containers to regenerate without waiting 12 hours. Ideas welcome!

PR #4543, opened 2022-04-22
Wip/hughsie/icons

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dfu/dfu.quirk
+++ b/plugins/dfu/dfu.quirk
@@ -18,6 +18,10 @@ CounterpartGuid = USB\VID_0BDA&PID_5800
 [USB\VID_0BDA&PID_5800]
 Flags = detach-for-attach
 
+# As found in Lenovo ThinkPad P1 Gen3
+[USB\VID_13D3&PID_5405]
+Icon = camera-web
+
 # Openmoko Freerunner / GTA02
 [USB\VID_1D50&PID_5119]
 Plugin = dfu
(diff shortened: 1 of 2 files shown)


plugins/dfu/dfu.quirk
```diff
@@ -18,6 +18,10 @@ CounterpartGuid = USB\VID_0BDA&PID_5800
 [USB\VID_0BDA&PID_5800]
 Flags = detach-for-attach
 
+# As found in Lenovo ThinkPad P1 Gen3
+[USB\VID_13D3&PID_5405]
+Icon = camera-web
```
superm1, line 23: There's other ones that are clearly USB Cameras too that should take this. (`# Realtek USB camera`).
As a thought - maybe you wan tto instead look at the configuration descriptor for the device and see if it has `bFucntionClass` of `Video`/`bFucntionSubClass` `Video Interface Collection`.

This heuristic will scale better then.

hughsie: Good idea. Maybe we also want fwupd_device_add_icon to actually prepend the icon, rather than append it, i.e. then a subclass adding an icon won't be overwritten.

hughsie: Something like https://github.com/fwupd/fwupd/pull/4546 ? that's probably the better way to do it.

PR #4545, opened 2022-04-22
Fix a CI failure for any distro using git >= 2.35.3

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/create_containers.yml
+++ b/.github/workflows/create_containers.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: "Generate Dockerfile"
         env:
           OS: ${{ matrix.os }}
--- a/contrib/ci/fedora.sh
+++ b/contrib/ci/fedora.sh
@@ -5,6 +5,9 @@ set -x
 #get any missing deps from the container
 ./contrib/ci/fwupd_setup_helpers.py install-dependencies --yes -o fedora
 
+# disable the safe directory feature
+git config --global safe.directory "*"
+
 #generate a tarball
 mkdir -p build && pushd build
 rm -rf *
--- a/plugins/dell/fu-dell-smi.c
+++ b/plugins/dell/fu-dell-smi.c
@@ -74,6 +74,8 @@ fu_dell_get_res(FuDellSmiObj *smi_obj, guint8 arg)
 	return dell_smi_obj_get_res(smi_obj->smi, arg);
 }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overread"
 gboolean
 fu_dell_execute_simple_smi(FuDellSmiObj *obj, guint16 class, guint16 select)
 {
@@ -87,6 +89,7 @@ fu_dell_execute_simple_smi(FuDellSmiObj *obj, guint16 class, guint16 select)
 	}
 	return TRUE;
 }
+#pragma GCC diagnostic pop
 
 gboolean
 fu_dell_detect_dock(FuDellSmiObj *smi_obj, guint32 *location)


superm1:
I saw you had this branch, which it seems like the right move to me if it fixes Fedora CI.

hughsie:
> which it seems like the right move to me if it fixes Fedora CI

I think **if** is the key word there. I had high hopes; but I regen'd the containers with this change and Fedora (or any distro with the newest git) still fails. I still need to test just doing the v2 to v3 change, but gah.

hughsie:
Also see this: https://github.com/actions/checkout/pull/770 -- it's all a bit of a mess.

hughsie:
I think merging this is probably the correct thing to do, although I'm still not sure what the root issue is. I suspect meson is calling git and getting the permission error caused by the new git feature, and swallowing it, and returning as if there's no git tree there.

hughsie:
@superm1 getting there! I'm now getting what I think is a false positive:

	In function ‘fu_dell_execute_simple_smi’,
	    inlined from ‘fu_dell_detect_dock’ at ../plugins/dell/fu-dell-smi.c:106:7:
	../plugins/dell/fu-dell-smi.c:84:13: error: ‘dell_simple_ci_smi’ reading 16 bytes from a region of size 4 [-Werror=stringop-overread]
	   84 |         if (dell_simple_ci_smi(class, select, obj->input, obj->output)) {
	      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	../plugins/dell/fu-dell-smi.c:84:13: note: referencing argument 3 of type ‘const u32[4]’
	../plugins/dell/fu-dell-smi.c:84:13: note: referencing argument 4 of type ‘u32[4]’
	In file included from ../plugins/dell/fu-dell-smi.h:12,
			 from ../plugins/dell/fu-dell-smi.c:9:
	/usr/include/smbios_c/smi.h: In function ‘fu_dell_detect_dock’:
	/usr/include/smbios_c/smi.h:40:26: note: in a call to function ‘dell_simple_ci_smi’
	   40 | LIBSMBIOS_C_DLL_SPEC int dell_simple_ci_smi(u16 smiClass, u16 select, const u32 args[4], u32 res[4]);
	      |                          ^~~~~~~~~~~~~~~~~~
	cc1: all warnings being treated as errors

Sanity check please?

superm1:
> @superm1 getting there! I'm now getting what I think is a false positive:
> 
> ```
> In function ‘fu_dell_execute_simple_smi’,
>     inlined from ‘fu_dell_detect_dock’ at ../plugins/dell/fu-dell-smi.c:106:7:
> ../plugins/dell/fu-dell-smi.c:84:13: error: ‘dell_simple_ci_smi’ reading 16 bytes from a region of size 4 [-Werror=stringop-overread]
>    84 |         if (dell_simple_ci_smi(class, select, obj->input, obj->output)) {
>       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ../plugins/dell/fu-dell-smi.c:84:13: note: referencing argument 3 of type ‘const u32[4]’
> ../plugins/dell/fu-dell-smi.c:84:13: note: referencing argument 4 of type ‘u32[4]’
> In file included from ../plugins/dell/fu-dell-smi.h:12,
> 		 from ../plugins/dell/fu-dell-smi.c:9:
> /usr/include/smbios_c/smi.h: In function ‘fu_dell_detect_dock’:
> /usr/include/smbios_c/smi.h:40:26: note: in a call to function ‘dell_simple_ci_smi’
>    40 | LIBSMBIOS_C_DLL_SPEC int dell_simple_ci_smi(u16 smiClass, u16 select, const u32 args[4], u32 res[4]);
>       |                          ^~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> ```
> 
> Sanity check please?

Yeah looks false positive to me too. I wonder if you can cast an array to make it happier. But if not then a pragma to ignore makes sense to me

Issue #4551, opened 2022-04-24
Install the D-Bus interfaces XML file even if using `-Dintrospection=false`


lancer1268:
**Is your feature request related to a problem? Please describe.**
I want to access the D-Bus interfaces XML file in `$(datadir)/dbus-1/interfaces/` with the `-Dintrospection=false` option.

As described in the [D-Bus API Design Guidelines](https://dbus.freedesktop.org/doc/dbus-api-design.html)
> Interface files for public API should be installed to `$(datadir)/dbus-1/interfaces` so that other services can load them.

**Describe the solution you'd like**
Install the D-Bus interfaces XML file even if using `-Dintrospection=false`. For example, revert [this commit](https://github.com/fwupd/fwupd/pull/4038/commits/e284b90cc102c642caaefcbec0cb08285243af1e).

**Describe alternatives you've considered**
Use another flag to enable the installation of the D-Bus XML file.

**Additional context**
I find the description of `introspection` in the `meson_options.txt`. It seems that the `introspection` here means the "GObject Introspection" instead of the "D-BUS Interface Introspection". Therefore, the D-Bus XML file should still be installed even if using `-Dintrospection=false`.
```
option('introspection', type : 'feature', description : 'generate GObject Introspection data', deprecated: {'true': 'enabled', 'false': 'disabled'})
```

superm1:
Would you mind doing a PR to change this?

hughsie:
Ack, pr please.

lancer1268:
Ok, I can create PRs for this.

The `meson.build` for the `main` branch and `1_7_X` branch are different. Should I create one PR for each of them?

Note: My project currently uses the `1_7_X` branch.

superm1:
> Ok, I can create PRs for this.
> 
> The `meson.build` for the `main` branch and `1_7_X` branch are different. Should I create one PR for each of them?
> 
> Note: My project currently uses the `1_7_X` branch.

It should be done first in main. Then it should be backported/redone for 1_7_X.

lancer1268:
I see. Thank you.

hughsie:
Done! Can you do the same for 1_7_X pls

lancer1268:
Sure, but how can I do that for `1_7_X`? The presubmit check fails when I try to commit code to the `1_7_X` branch.
```
Don't commit to branch...................................................Failed
- hook id: no-commit-to-branch
- exit code: 1
```

superm1:
[...]

PR #4553, opened 2022-04-25
Install D-Bus introspection data even if introspection is disabled

lancer1268:
According to the D-Bus API Design Guidelines, the D-Bus interface files
for public API should be installed so that other services can load them.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Related issue: https://github.com/fwupd/fwupd/issues/4551

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -2,7 +2,7 @@ if get_option('tests')
 subdir('tests')
 endif
 
-if build_daemon and get_option('introspection')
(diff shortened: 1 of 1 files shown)


superm1 (approved):
Don't worry about the Fedora build failure; that is a fix we need to backport from master.

lancer1268:
Thank you for the clarification.

PR #4557, opened 2022-04-25
Make the security policy reflect reality

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation


hughsie:
> Add something to RELEASE actions to drop older releases

I did think about that; I'd rather choose the branches on what I see on the LVFS, e.g. what's installed in RHEL 8 and Ubuntu.

PR #4558, opened 2022-04-25
Adjust handling of fwupd-refresh.service

superm1:
This PR does a few things:
* Disable `fwupd-refresh.timer` by default as well.  I think we have too many circumstances this can lead to a failure to make it a default to enabled everywhere.  It should be enabled by packaging until when we have this sorted out.
* Add new configuration option to choose to set a static user by packaging
* Modifications to Debian packaging to create a user by default.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/debian/fwupd.postinst
+++ b/contrib/debian/fwupd.postinst
@@ -48,3 +48,10 @@ for dir in /var/cache/fwupdate /var/lib/fwupdate; do
 	        rmdir --ignore-fail-on-non-empty $dir || true
 	fi
 done
+
+#create a user for fwupd-refresh.service/fwupd-refresh.timer
+adduser --quiet --system --group --no-create-home --home /run/systemd \
+    --gecos "fwupd-refresh user" fwupd-refresh
+if [ -d /run/systemd/system ]; then
+    systemctl reload dbus || true
+fi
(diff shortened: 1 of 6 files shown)


contrib/debian/fwupd.postinst
```diff
@@ -48,3 +48,10 @@ for dir in /var/cache/fwupdate /var/lib/fwupdate; do
 	        rmdir --ignore-fail-on-non-empty $dir || true
 	fi
 done
+
+#create a user for fwupd-refresh.service/fwupd-refresh.timer
+adduser --quiet --system --group --no-create-home --home /run/systemd \
+    --gecos "fwupd-refresh user" fwupd-refresh
+if [ -d /run/systemd/system ]; then
+    systemctl reload dbus || true
```
hughsie, line 56: at least on fedora, this is a really good way to make your system explode and need rebooting -- are you sure it's safe on debian?

hughsie: Ohh, reload -- not restart -- I guess?

hughsie (approved):
I guess this is pragmatic, although it does seem like it's working around the root issue.

superm1:
> 

I mean it definitely is.  But the reality is that it will take some time for systemd to fix things.  I think when systemd fixes are rolled out more widely we can swap things around.

PR #4570, opened 2022-04-28
Wip/hughsie/ufs ffu

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -332,6 +332,9 @@ endif
 if cc.has_header('linux/ethtool.h')
   conf.set('HAVE_ETHTOOL_H', '1')
 endif
+if cc.has_header('linux/bsg.h') and cc.has_header('scsi/scsi_bsg_ufs.h')
+  conf.set('HAVE_BSG_H', '1')
+endif
 if cc.has_header('mtd/mtd-user.h')
   conf.set('HAVE_MTD_USER_H', '1')
 endif
--- a/plugins/scsi/fu-scsi-device.c
+++ b/plugins/scsi/fu-scsi-device.c
@@ -308,6 +308,9 @@ fu_scsi_device_write_firmware(FuDevice *device,
 			return FALSE;
 		}
 
+		/* wait for the timeout */
+		fu_progress_sleep(fu_progress_get_child(progress), self->ffu_timeout * 1000);
+
 		/* chunk done */
 		fu_progress_step_done(progress);
 	}
--- a/plugins/scsi/fu-scsi-device.c
+++ b/plugins/scsi/fu-scsi-device.c
@@ -46,6 +46,7 @@ static gboolean
 fu_scsi_device_probe(FuDevice *device, GError **error)
 {
 	FuScsiDevice *self = FU_SCSI_DEVICE(device);
+	const gchar *device_file = fu_udev_device_get_device_file(FU_UDEV_DEVICE(device));
 	const gchar *name;
 	const gchar *vendor;
 	const gchar *version;
@@ -105,6 +106,13 @@ fu_scsi_device_probe(FuDevice *device, GError **error)
 			g_prefix_error(error, "no ffu timeout specified: ");
 			return FALSE;
 		}
+	}
+
+	/* which sg API do we use? */
+	if (g_str_has_prefix(device_file, "/dev/sd")) {
+		self->sg_version = 0x3;
+	} else if (g_str_has_prefix(device_file, "/dev/bsg/ufs-bsg")) {
+		self->sg_version = 0x4;
 	}
 
 	/* add GUIDs */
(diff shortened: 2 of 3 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -1411,7 +1411,13 @@ fu_udev_device_ioctl(FuUdevDevice *self, gulong request, guint8 *buf, gint *rc,
 		return FALSE;
 	}
 
+#ifdef HAVE_ERRNO_H
+	while (((rc_tmp = ioctl(priv->fd, request, buf)) < 0) &&
+	       (errno == EINTR || errno == EAGAIN))
+		;
+#else
 	rc_tmp = ioctl(priv->fd, request, buf);
+#endif
```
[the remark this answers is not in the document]
hughsie: yup, that works. I'm wondering whether we also need a fu_udev_device_ioctl_full that controls this behaviour? And if so, weather we want to limit the number of retries, or have some kind of ms timeout to avoid waiting here forever for a faulty device.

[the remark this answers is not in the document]
superm1: Yes I was wondering about the eventual timeout thing too. Good idea.

plugins/scsi/fu-scsi-device.c
```diff
@@ -6,20 +6,50 @@
 
 #include "config.h"
 
+#ifdef HAVE_BSG_H
```
superm1, line 9: maybe change the plugin_scsi meson.build to auto-enable when these headers are present instead?

meson.build
```diff
@@ -332,6 +332,9 @@ endif
 if cc.has_header('linux/ethtool.h')
   conf.set('HAVE_ETHTOOL_H', '1')
 endif
+if cc.has_header('linux/bsg.h') and cc.has_header('scsi/scsi_bsg_ufs.h')
```
superm1, line 335: I think technically these should be separate checks, particularly if you follow my other suggestion to drop `HAVE_BSG_H` and make them mandatory for plugin_scsi to enable.

plugins/scsi/fu-scsi-device.c
hughsie: Hmm, I think we still want the plugin even in read-only mode, e.g. to detect those HEP disks that self destruct.

superm1: > Hmm, I think we still want the plugin even in read-only mode, e.g. to detect those HEP disks that self destruct.

what is the goal for that info though?

hughsie: We show that info in the "fwupdmgr get-devices" IIRC.

superm1: If they're not updatable they're hidden by default though. So someone really has to be looking for such a device.

plugins/scsi/fu-scsi-device.c
```diff
@@ -83,12 +134,214 @@ fu_scsi_device_probe(FuDevice *device, GError **error)
...
 static void
 fu_scsi_device_init(FuScsiDevice *self)
 {
 	fu_device_add_icon(FU_DEVICE(self), "drive-harddisk");
 	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_PLAIN);
 	fu_device_set_summary(FU_DEVICE(self), "SCSI device");
+	fu_udev_device_set_flags(FU_UDEV_DEVICE(self),
+				 FU_UDEV_DEVICE_FLAG_OPEN_READ | FU_UDEV_DEVICE_FLAG_OPEN_SYNC |
+				     FU_UDEV_DEVICE_FLAG_IOCTL_RETRY);
+	self->sg_version = fu_common_check_kernel_version("5.13.0", NULL) ? 0x4 : 0x3;
```
superm1, line 344: Rhel 8 :(
But I guess if there is no other way to know better than nothing.

plugins/scsi/fu-scsi-device.c
```diff
@@ -56,6 +86,27 @@ fu_scsi_device_probe(FuDevice *device, GError **error)
...
+	/* is FFU capable */
+	if (fu_udev_device_get_sysfs_attr_uint64(FU_UDEV_DEVICE(device),
+						 "device_descriptior/ufs_features",
+						 &ufs_features,
+						 NULL)) {
+		fu_device_set_summary(device, "UFS device");
+		if (ufs_features & 0x1) {
+#ifdef HAVE_BSG_H
+			fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UPDATABLE);
+			fu_device_add_protocol(device, "org.jedec.ufs");
```
superm1, line 98: Make sure goes to plugin readme.

hughsie: Ack, thanks. Done.

PR #4582, opened 2022-05-03
Add plugin quirk keys earlier in the startup process

hughsie:
This allows creating the silo when starting the engine with custom
plugin keys such as WacomI2cFlashBaseAddr.

If we move the plugin initialization earlier then we don't get the
HwID matches, so we really do have to split this into a 4-stage startup,
e.g. ->class_init(), ->init(), ->startup() and ->coldplug().

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin-private.h
+++ b/libfwupdplugin/fu-plugin-private.h
@@ -34,6 +34,8 @@ GHashTable *
 fu_plugin_get_report_metadata(FuPlugin *self);
 gboolean
 fu_plugin_open(FuPlugin *self, const gchar *filename, GError **error) G_GNUC_WARN_UNUSED_RESULT;
+void
+fu_plugin_runner_init(FuPlugin *self);
 gboolean
 fu_plugin_runner_startup(FuPlugin *self, GError **error) G_GNUC_WARN_UNUSED_RESULT;
 gboolean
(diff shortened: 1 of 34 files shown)


hughsie:
@superm1 I'm not 100% on the `->class_init()` name -- other ideas welcome. We really do need to split up the init like this if we  want to support `--plugins` without that ugly silo build warning.

superm1:
> @superm1 I'm not 100% on the `->class_init()` name -- other ideas welcome. We really do need to split up the init like this if we  want to support `--plugins` without that ugly silo build warning.

How about splitting to a "load" and "init" phase with load being the new stuff and init being the old.

hughsie:
Maybe `->context_init()` is even more obvious? or maybe `->context_setup()` -- the former is actually what I had originally.

superm1:
> Maybe `->context_init()` is even more obvious? or maybe `->context_setup()` -- the former is actually what I had originally.

How about `context_seed_plugin`

hughsie:
Maybe the obvious one is `->open`? it's when the vfunc is run after-all!

hughsie:
> Maybe the obvious one is ->open

Seeing this in reality is less good. Open implies there's a close, which isn't for this. We also use `FuDevice->open()` to mean something entirely different. Current contenders are `module_open()` and also providing an un-yet unused `module_close(). Maybe `early_init` works too? or maybe just `context_init` is best afterall. Naming things is hard!

PR #4588, opened 2022-05-05
Allow the device to pause the polling when writing firmware

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-device.c
```diff
@@ -3684,6 +3723,13 @@ fu_device_write_firmware(FuDevice *self,
 		return FALSE;
 	}
 
+	/* pause the polling */
+	if (fu_device_has_internal_flag(self, FU_DEVICE_INTERNAL_AUTO_PAUSE_POLLING)) {
+		g_autoptr(FuDeviceLocker) poll_locker = fu_device_poll_locker_new(self, error);
```
gportay, line 3728: Hum, the locker will be created and released instantly.

The scope for the locker should be global to the function?

gportay: tested by rising the scope to the whole function and it works!

hughsie: doh! will fix

gportay:
Tested!

gportay:
I am testing using the daemon (I did my develpment with `fwupdmgr`) and I met

```
14:14:06:0664 FuPluginSteelSeries  ignoring error on getting battery state: failed to do control transfer: Device 1038:1830 has not been opened

hughsie:
> That means that I cannot do USB transfer in _poll().

Ohh you can, you just need to open it with a FuDeviceLocker if the device is not "kept open".

hughsie:
> maybe it should also includes the _attach() and _detach() vfunc while doing _write_firmware()

Fixed!

gportay:
It seems to run fine;

However, I wonder what is the context of execution for `_detach()`, `_write()`, `_attach()` and `_poll()`, and if the `_poll()` may runs between the locker is released after `_detach()` or `_write()` returns and before `_write()` or `attach()` starts. Is that possible?

hughsie:
> Is that possible

I don't think so; there's just one GMainContext and only the main thread.

PR #4593, opened 2022-05-05
trivial: set up win32 build not to take GPG by default

superm1:
Unfortunately you can't detect the meson option:
```
../data/pki/meson.build:4:4: ERROR: Having a colon in option name is forbidden, projects are not allowed to directly access options of other subprojects.
```
So hardcode in top level meson.build that Windows doesn't take GPG.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/pki/meson.build
+++ b/data/pki/meson.build
@@ -1,13 +1,13 @@
 # only install files that are going to be used
 if libjcat.type_name() != 'internal' and libjcat.version().version_compare('>= 0.1.9')
-  supported_gpg = libjcat.get_variable(pkgconfig : 'supported_gpg')
-  supported_pkcs7 = libjcat.get_variable(pkgconfig : 'supported_pkcs7')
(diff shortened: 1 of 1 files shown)


data/pki/meson.build
```diff
@@ -3,8 +3,8 @@ if libjcat.type_name() != 'internal' and libjcat.version().version_compare('>= 0
   supported_gpg = libjcat.get_variable(pkgconfig : 'supported_gpg')
   supported_pkcs7 = libjcat.get_variable(pkgconfig : 'supported_pkcs7')
 else
-  supported_gpg = '1'
-  supported_pkcs7 = '1'
+  supported_gpg = implied_gpg
+  supported_pkcs7 = implied_pkcs7
```
hughsie, line 7: would this be better:

  supported_gpg = host_machine.system() != 'windows'
  supported_pkcs7 = '1'

I don't think we need the global values anywhere else, and it seems sane to put the condition here.

hughsie (approved):

PR #4594, opened 2022-05-06
scsi: Rearchitect the plugin to match a block device

hughsie:
This makes it match the eMMC and SATA plugins, and means we can more easily
walk up to the parent SCSI controller, rather than trying to find the correct
block child.

No functional changes, other than now we also read the disk serial number, and
use one less udev subsystem match.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


daniil-lunev (approved):
LGTM, verified on my hw. One good side effect - device is now singular, not in 3 instances :D

PR #4597, opened 2022-05-06
Switch the Windows installer from NSIS to MSI

hughsie:
This means we get a few bugs fixed:

 * We appear in the Add/Remove Programs UI
 * We have an uninstaller that works on Windows 10
 * The installation is pretty, and scriptable
 * We actually install the translations

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


contrib/ci/build_windows.sh
```diff
@@ -27,6 +28,8 @@ meson .. \
     -Ddocs=none \
     -Dhsi=false \
     -Dman=false \
+    -Dfish_completion=false \
```
superm1, line 31: I'd actually argue we don't want bash completion either on Windows

contrib/ci/build_windows.sh
```diff
@@ -39,14 +42,23 @@ meson .. \
...
-ninja -v
-ninja -v install
+
+# run tests
+export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/;$build/libfwupd/;$build/libfwupdplugin/;$build/subprojects/libxmlb/src/;$build/subprojects/gcab/libgcab/;$build/subprojects/libjcat/libjcat/;$build/subprojects/gusb/gusb/"
+ninja -C $build -v
+ninja -C $build test
+
+# switch to release optimisations
+meson configure -Dtests=false -Dbuildtype=release
```
superm1, line 59: do any other subprojects need tests turned off too for this?

contrib/ci/build_windows.sh
```diff
@@ -57,25 +69,91 @@ echo $VERSION > $DESTDIR/VERSION
...
+	-x lib/pkgconfig/ \
+	--directory-ref INSTALLDIR \
+	--var "var.DESTDIR" \
+	--component-group "CG.fwupd-files" | \
+	tee $build/contrib/fwupd-files.wxs
 
-export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/;$build/libfwupd/;$build/libfwupdplugin/;$build/subprojects/libxmlb/src/;$build/subprojects/gcab/libgcab/;$build/subprojects/libjcat/libjcat/;$build/subprojects/gusb/gusb/"
-ninja -C $build test
+MSI_FILENAME="setup/fwupd-$VERSION-setup-x86_64.msi"
+mkdir -p setup
```
superm1, line 144: I think from this point onwards it would be good to only run the code if `CI=1`.  That way we don't need to make the MSI for "local" builds.

meson.build
```diff
@@ -520,6 +520,15 @@ motd_dir = 'motd.d'
 conf.set_quoted('MOTD_FILE', motd_file)
 conf.set_quoted('MOTD_DIR', motd_dir)
 
+if host_machine.system() == 'windows'
+  if host_machine.cpu() != 'x86_64'
+    wixl_arch = 'x86'
+  else
+    wixl_arch = 'x64'
```
superm1, line 527: <cough> arm </cough>

meson.build
```diff
@@ -564,13 +573,3 @@ subdir('docs')
 if build_daemon and libsystemd.found()
   meson.add_install_script('meson_post_install.sh', systemdunitdir)
 endif
-
-makensis = find_program('makensis', required : host_machine.system() == 'windows')
```
superm1, line 568: I think you want to add a `find_program('wixl', required: host_machine.system == 'windows')` here still

contrib/ci/build_windows.sh
```diff
@@ -57,25 +69,91 @@ echo $VERSION > $DESTDIR/VERSION
...
+	-o -name libusb-1.0.dll \
+	-o -name libwinpthread-1.dll \
+	-o -name libxml2-2.dll \
+	-o -name zlib1.dll \
+	| wixl-heat \
+	-p $MINGW32BINDIR/ \
+	--directory-ref BINDIR \
+	--var "var.MINGW32BINDIR" \
+	--component-group "CG.fwupd-deps" | \
+	tee $build/contrib/fwupd-deps.wxs
```
superm1, line 118: although I really like how programattic this list is, this change breaks running without installing.  As it stands today you can drop a `dist` directory onto a real Windows host (or "natively" launch it from WSL2) and it runs.  I want to maintain that if at all possible.

Can we maybe make symlinks for all the files and just not install the symlinks from the MSI?

contrib/ci/build_windows.sh
```diff
@@ -39,14 +42,23 @@ meson .. \
...
+
+# run tests
+export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/;$build/libfwupd/;$build/libfwupdplugin/;$build/subprojects/libxmlb/src/;$build/subprojects/gcab/libgcab/;$build/subprojects/libjcat/libjcat/;$build/subprojects/gusb/gusb/"
+ninja -C $build -v
+ninja -C $build test
+
+# switch to release optimisations
+meson configure -Dtests=false -Dbuildtype=release
+ninja -C $build -v install
+#find $DESTDIR -name "*.a" | xargs rm
```
superm1, line 61: any idea why these are still made?  or that's TODO?

contrib/ci/build_windows.sh
hughsie: I think we turn them off for the "main" configure step already.

meson.build
hughsie: I'm guessing it's `host_machine.cpu() of "aarch64" to "arm64"`?

meson.build
hughsie: I figured it was just a non-installed configure.in-style replacement -- your logic is just to avoid doing that on win32 if it's not going to be used?

contrib/ci/build_windows.sh
hughsie: would copying those files into `$DESTDIR/bin` also work? That does feel a bit of a hack. Can you describe a bit how you use WSL right now so I can make sure this doesn't break things?

contrib/ci/build_windows.sh
hughsie: deadcode, removed.

contrib/ci/build_windows.sh
hughsie: Agree, fixed.

contrib/ci/build_windows.sh
hughsie: I'm not sure much of contrib/ci/build_windows.sh makes sense outside of CI -- it unconditionally `rm -rf`'s your build-win32 for instance. Maybe we make the script quit right at the start if CI != 1?

superm1: If we want people to be able to make changes to Win32, we have to make building for Win32 hosts accessible though..

meson.build
superm1: My thought is that if the build script needs it, then it should be something that we check for in meson.build so we can message accordingly.

contrib/ci/build_windows.sh
hughsie: I don't disagree, but I think that's a different script to the CI one.

meson.build
hughsie: Well, the CI script needs it, but not generally for building win32...

contrib/ci/build_windows.sh
superm1: What i've been doing is:
1. Open up Fedora in WSL
2. `./contrib/ci/build_windows.sh`
3. `./dist/bin/fwupdtool.exe`

I don't mind changing that workflow a bit (add environment variable here, or there etc), but adding a step of installing an MSI is too much to me.

.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
                                   gcab
                                   ca-certificates
                                   python3-pip
-                                  mingw32-nsis
+                                  https://kojipkgs.fedoraproject.org//work/tasks/3617/86733617/msitools-0.101.32-4.fc36.x86_64.rpm
```
superm1, line 37: drop a comment on why this specific version is needed so we know when to switch away?

contrib/ci/build_windows.sh
superm1: Copying those files to `$DESTDIR/bin` would also work.  Or I wonder if maybe there is an easy way to just extract the MSI and run it from that directory it's extracted to?

contrib/ci/build_windows.sh
superm1: We do have other scripts that change one or two things based on CI=1 though.

.circleci/config.yml
hughsie: I've asked the msitools maintainer to do a new upstream release, so maybe we can just wait until the official release happens.

contrib/ci/build_windows.sh
hughsie: Or is it possible on Windows to set the DLL search path add /usr/x86_64-w64-mingw32/sys-root/mingw/bin`?

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       - persist_to_workspace:
           root: .
           paths:
-            - "dist/setup/*.exe"
+            - "dist/setup/*.msi"
```
superm1, line 57: AFAICT this MSI is ending up in `./setup` not `dist/setup`

contrib/ci/build_windows.sh
```diff
@@ -57,25 +76,91 @@ echo $VERSION > $DESTDIR/VERSION
...
+	-x share/doc/ \
+	-x lib/pkgconfig/ \
+	--directory-ref INSTALLDIR \
+	--var "var.DESTDIR" \
+	--component-group "CG.fwupd-files" | \
+	tee $build/contrib/fwupd-files.wxs
 
-export WINEPATH="/usr/x86_64-w64-mingw32/sys-root/mingw/bin/;$build/libfwupd/;$build/libfwupdplugin/;$build/subprojects/libxmlb/src/;$build/subprojects/gcab/libgcab/;$build/subprojects/libjcat/libjcat/;$build/subprojects/gusb/gusb/"
-ninja -C $build test
+MSI_FILENAME="setup/fwupd-$VERSION-setup-x86_64.msi"
```
superm1, line 150: needs `DESTDIR` prefix

hughsie: Ack fixed, thanks.

contrib/ci/build_windows.sh
superm1: Not easily that I can tell.   How about if you just run `msiextract` at the end of the script?  It should overwrite everything in dist and copy in the missing dependency libraries.

hughsie: That didn't work super-well. I've added a more useful script in ./contrib -- can you try that with WSL and Fedora please.

contrib/ci/build_windows.sh
hughsie: I think it actually makes sense to have different compile options for the CI and a local win32 build. I've added a script for local "getting up and going" use.

contrib/build-windows.sh
```diff
@@ -0,0 +1,114 @@
...
+        tree \
+        wine
+fi
+
+root=$(pwd)
+export DESTDIR=${root}/dist
+build=$root/build-win32
+mkdir -p "$build" && cd "$build"
+
+# try to keep this and ../contrib/ci/build_windows.sh in sync as much as makes sense
```
superm1, line 36: How about just source them from one another rather than have to keep in sync?

cwsystem76:
Via the [artifact build](https://app.circleci.com/pipelines/github/fwupd/fwupd/9892/workflows/11961248-0d10-4686-be25-d3822d24dd64/jobs/35847/artifacts) in the CI:

image

And the uninstaller works as intended from the "Apps & features" page and from the old Control Panel as well. One thing to note, though, it's being installed into `Program Files (x86)`, which is for 32-bit apps, instead of `Program Files` which is where 64bit apps usually go. There may be a reason for this but I thought I'd mention it.

contrib/build-windows.sh
hughsie: I did ponder putting them in `contrib/ci/dependencies.xml` -- e.g.

      <package variant="mingw64">mingw64-curl</package>

insane?

hughsie:
> it's being installed into Program Files (x86)

It seems that although we specify `Win64=true` in the `Component` and set `InstalledVersion=200` we're hitting https://gitlab.gnome.org/GNOME/msitools/-/issues/46

contrib/build-windows.sh
superm1: Not insane, makes perfect sense!

hughsie: Ya,concur, it kinda works.

superm1: Drop the manual install steps now from circleci too.

superm1: Drop the manual install steps now from circleci too.

superm1: On my phone maybe i missed it. I was meaning  in .circleci/config.yaml

.circleci/config.yml
```diff
@@ -34,18 +34,8 @@ jobs:
...
-                                  mingw64-gettext
-                                  mingw64-pkg-config
-                                  mingw64-glib2
-                                  mingw64-gnutls
-                                  mingw64-libusbx
-                                  mingw64-sqlite
-                                  mingw64-libarchive
-                                  mingw64-json-glib
-                                  mingw64-curl
+                                  tree
```
superm1, line 37: shouldn't everything else above here be in dependencies.xml?  The idea is supposed to be you don't have to manually install anything

hughsie: Yes, and done. There are a few more native things than I liked, but I've been filing bugs against other components all day. As they get merged and we update the subprojects we can remove a few BRs from the mingw64 build.

.circleci/config.yml
hughsie: I'll rip out the manual dnf install https:// when the update hits updates-testing in a few days.

contrib/ci/build_windows.sh
hughsie: I actually got WSL working with Fedora too. :)

contrib/build-windows.sh
```diff
@@ -0,0 +1,92 @@
...
+    --prefix=/ \
+    --sysconfdir="etc" \
+    --libexecdir="bin" \
+    --bindir="bin" \
+    -Dbuild=all \
+    -Ddocs=none \
+    -Dhsi=false \
+    -Dman=false \
+    -Dfish_completion=false \
+    -Dbash_completion=false \
```
superm1, line 26: what do you think about converting these two to tristate options that auto-disable for os == windows?

hughsie: do either *ever* make sense on Windows? If not, then yes please, although I think you're the expert at the auto-disable magic.

superm1 (approved):
I still do think running ./contrib/build-windows.sh from contrib/ci/build-windows.sh would make sense, but otherwise it LGTM.

contrib/build-windows.sh
superm1: The only time you run bash in Windows is WSL, and if you're in WSL you won't read bash-completion.  So I don't think they ever make sense in Windows

hughsie:
> still do think running ./contrib/build-windows.sh from contrib/ci/build-windows.sh would make sense, 

It's not a 1:1 mapping; i.e. some options make sense in CI, but not for local hacking and vice-versa.

contrib/ci/build_windows.sh
superm1: It's handy right!

PR #4602, opened 2022-05-09
scsi: Support updating firmware on UFS devices

hughsie:
Change-Id: I5573b1890135d1fd3b2f0cb8f95f74cbb3e21abe

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/scsi/README.md
+++ b/plugins/scsi/README.md
@@ -2,8 +2,19 @@
 
 ## Introduction
 
-This plugin adds support for SCSI storage hardware. Devices are only enumerated
-and are not updatable.
+This plugin adds support for SCSI storage hardware. Most SCSI devices are enumerated and some UFS
+devices may also be updatable.
+
+Firmware is sent in 4kB chunks and activated on next reboot only.
+
+## Firmware Format
+
+The daemon will decompress the cabinet archive and extract a firmware blob in
+an unspecified binary file format.
+
+This plugin supports the following protocol ID:
+
+* org.jedec.ufs
 
 ## GUID Generation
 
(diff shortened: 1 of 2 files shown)


plugins/scsi/README.md
```diff
@@ -2,8 +2,19 @@
 
 ## Introduction
 
-This plugin adds support for SCSI storage hardware. Devices are only enumerated
-and are not updatable.
+This plugin adds support for SCSI storage hardware. Most SCSI devices are are enumerated and UFS
```
daniil-lunev, line 5: s/are are/are/

plugins/scsi/fu-scsi-device.c
```diff
@@ -46,9 +70,36 @@ fu_scsi_device_probe(FuDevice *device, GError **error)
 				    "no assigned vendor");
 		return FALSE;
 	}
+
 	vendor_id = g_strdup_printf("SCSI:%s", fu_device_get_vendor(device));
 	fu_device_add_vendor_id(device, vendor_id);
 
+	/* the controller could really on any bus.. */
```
daniil-lunev, line 77: could really be*

 May be word it as a TODO to add better search for ufshci?

daniil-lunev (approved):
LGTM modulo small comment nits

One Q, not related to this CL - I opbserved update with fwupdtool doesn't actually update the history? i.e. I see need-update flag in verbose log of install-blob, but then it is not reflected in the get-devices, is that expected?

plugins/scsi/README.md
hughsie: Fixed, thanks.

plugins/scsi/fu-scsi-device.c
hughsie: Agree, fixed.

hughsie:
> fwupdtool doesn't actually update the history

Ya expected, you need to use `install` rather than `install-blob` if you want to use the history functionality.

PR #4615, opened 2022-05-10
steelseries: Add support for Rival 3 Wireless

gportay:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/steelseries/fu-steelseries-sonic.c
+++ b/plugins/steelseries/fu-steelseries-sonic.c
@@ -1,0 +1,1151 @@
+/*
+ * Copyright (C) 2022 Gaël PORTAY <gael.portay@collabora.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-steelseries-common.h"
+#include "fu-steelseries-sonic.h"
+
+#define STEELSERIES_TRANSACTION_TIMEOUT	       1000
+#define STEELSERIES_BUFFER_CONTROL_SIZE	       64
+#define STEELSERIES_BUFFER_FLASH_TRANSFER_SIZE 128
+#define STEELSERIES_BUFFER_RAM_TRANSFER_SIZE   48
+
+#define STEELSERIES_SONIC_WIRELESS_STATUS_OPCODE_OFFSET 0x0U
+#define STEELSERIES_SONIC_WIRELESS_STATUS_VALUE_OFFSET	0x0U
+
+#define STEELSERIES_SONIC_BATTERY_OPCODE_OFFSET	  0x0U
+#define STEELSERIES_SONIC_BATTERY_BAT_MODE_OFFSET 0x1U
+#define STEELSERIES_SONIC_BATTERY_VALUE_OFFSET	  0x0U
+
+#define STEELSERIES_SONIC_READ_FROM_RAM_OPCODE_OFFSET 0x0U
+#define STEELSERIES_SONIC_READ_FROM_RAM_OFFSET_OFFSET 0x2U
+#define STEELSERIES_SONIC_READ_FROM_RAM_SIZE_OFFSET   0x4U
+#define STEELSERIES_SONIC_READ_FROM_RAM_DATA_OFFSET   0x0U
+
+#define STEELSERIES_SONIC_READ_FROM_FLASH_OPCODE_OFFSET 0x0U
+#define STEELSERIES_SONIC_READ_FROM_FLASH_CHIPID_OFFSET 0x2U
+#define STEELSERIES_SONIC_READ_FROM_FLASH_OFFSET_OFFSET 0x4U
+#define STEELSERIES_SONIC_READ_FROM_FLASH_SIZE_OFFSET	0x8U
+
+#define STEELSERIES_SONIC_WRITE_TO_RAM_OPCODE_OFFSET 0x0U
+#define STEELSERIES_SONIC_WRITE_TO_RAM_OFFSET_OFFSET 0x2U
+#define STEELSERIES_SONIC_WRITE_TO_RAM_SIZE_OFFSET   0x4U
+#define STEELSERIES_SONIC_WRITE_TO_RAM_DATA_OFFSET   0x6U
+
+#define STEELSERIES_SONIC_WRITE_TO_FLASH_OPCODE_OFFSET 0x0U
+#define STEELSERIES_SONIC_WRITE_TO_FLASH_CHIPID_OFFSET 0x2U
+#define STEELSERIES_SONIC_WRITE_TO_FLASH_OFFSET_OFFSET 0x4U
+#define STEELSERIES_SONIC_WRITE_TO_FLASH_SIZE_OFFSET   0x8U
+
+#define STEELSERIES_SONIC_ERASE_OPCODE_OFFSET 0x0U
+#define STEELSERIES_SONIC_ERASE_CHIPID_OFFSET 0x2U
+
+#define STEELSERIES_SONIC_RESTART_CHIPID_OFFSET 0x0U
+
+typedef enum {
+	STEELSERIES_SONIC_CHIP_NORDIC = 0,
+	STEELSERIES_SONIC_CHIP_HOLTEK,
+	STEELSERIES_SONIC_CHIP_MOUSE,
+	/*< private >*/
+	STEELSERIES_SONIC_CHIP_LAST,
+} SteelseriesSonicChip;
+
+typedef enum {
+	STEELSERIES_SONIC_WIRELESS_STATE_OFF,	 /* WDS not initiated, radio is off */
+	STEELSERIES_SONIC_WIRELESS_STATE_IDLE,	 /* WDS initiated, dongle is transmitting beacon
+						  * (mouse will not have this state) */
+	STEELSERIES_SONIC_WIRELESS_STATE_SEARCH, /* WDS initiated, mouse is trying to synchronize to
+						  * dongle (dongle will not have this state) */
+	STEELSERIES_SONIC_WIRELESS_STATE_LOCKED, /* Dongle and mouse are synchronized, but not
+						  * necessarily connected. */
+	STEELSERIES_SONIC_WIRELESS_STATE_CONNECTED,  /* Dongle and mouse are connected. */
+	STEELSERIES_SONIC_WIRELESS_STATE_TERMINATED, /* Mouse has been disconnected from the dongle.
+						      */
+} SteelseriesSonicWirelessStatus;
+
+const guint16 STEELSERIES_SONIC_READ_FROM_RAM_OPCODE[] = {0x00c3U, 0x00c3U, 0x0083U};
+const guint16 STEELSERIES_SONIC_READ_FROM_FLASH_OPCODE[] = {0x00c5U, 0x00c5U, 0x0085U};
+const guint16 STEELSERIES_SONIC_WRITE_TO_RAM_OPCODE[] = {0x0043U, 0x0043U, 0x0003U};
+const guint16 STEELSERIES_SONIC_WRITE_TO_FLASH_OPCODE[] = {0x0045U, 0x0045U, 0x0005U};
+const guint16 STEELSERIES_SONIC_ERASE_OPCODE[] = {0x0048U, 0x0048U, 0x0008U};
+const guint16 STEELSERIES_SONIC_RESTART_OPCODE[] = {0x0041U, 0x0041U, 0x0001U};
+const guint16 STEELSERIES_SONIC_CHIP[] = {0x0002U, 0x0003U, 0x0002U};
+const guint32 STEELSERIES_SONIC_FIRMWARE_SIZE[] = {0x9000U, 0x4000U, 0x12000U};
+const gchar *STEELSERIES_SONIC_FIRMWARE_ID[] = {"app-nordic.bin",
+						"app-holtek.bin",
+						"mouse-app.bin"};
+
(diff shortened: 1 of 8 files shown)


plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+	/* success */
+	return TRUE;
+}
+
+static gboolean
+fu_steelseries_sonic_battery_state(FuDevice *device, guint16 *value, GError **error)
+{
+	guint8 data[STEELSERIES_BUFFER_CONTROL_SIZE] = {0};
+	const guint16 opcode = 0xAAU;
+	const guint8 bat_mode = 1U; /* percentage */
```
hughsie, line 182: the `U` is confusing me, maybe `0x01` instead?

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+
+	chunks =
+	    fu_chunk_array_mutable_new(buf, bufsz, 0x0, 0x0, STEELSERIES_BUFFER_RAM_TRANSFER_SIZE);
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_READ);
+	fu_progress_set_steps(progress, chunks->len);
+	for (guint i = 0; i < chunks->len; i++) {
+		FuChunk *chk = g_ptr_array_index(chunks, i);
+		const guint16 offset = fu_chunk_get_address(chk);
+		const guint16 size = fu_chunk_get_data_sz(chk);
```
hughsie, line 225: do we need these temp vars?

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+	chip = STEELSERIES_SONIC_CHIP_HOLTEK;
+	if (!fu_steelseries_sonic_restart(device, chip, fu_progress_get_child(progress), error)) {
+		g_prefix_error(error, "failed to restart chip %u: ", chip);
+		return FALSE;
+	}
+	fu_progress_step_done(progress);
+
+	/* the user has to do something */
+	msg = g_strdup_printf("%s needs to be manually restarted to complete the update. "
+			      "Please unplug the dongle and then re-plug it.",
```
hughsie, line 678: i think dongle is one of those words we're not supposed to use any more. perhaps "receiver" instead?

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+	}
+	g_debug("WirelessStatus: %u", wl_status);
+	if (wl_status != STEELSERIES_SONIC_WIRELESS_STATE_CONNECTED) {
+		g_autoptr(FwupdRequest) request = NULL;
+		g_autoptr(GTimer) timer = NULL;
+		g_autofree gchar *msg = NULL;
+
+		/* the user has to do something */
+		msg = g_strdup_printf("%s needs to be connected to start the update. "
+				      "Please press any button.",
```
hughsie, line 713: i think this needs to be a bit more specific. A button on _what_?

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+fu_steelseries_sonic_dump_firmware(FuDevice *device, FuProgress *progress, GError **error)
+{
+	SteelseriesSonicChip chip;
+	gsize blobsz;
+	guint32 bufsz;
+	guint8 *buf;
+	g_autofree guint8 *blob = NULL;
+
+	blobsz = fu_device_get_firmware_size_max(FU_DEVICE(device));
+	blob = g_malloc0(blobsz);
```
hughsie, line 763: this feels like the kind of thing you want to use `GByteArray` for.

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+	guint32 bufsz;
+	g_autoptr(GBytes) blob = NULL;
+	g_autofree guint8 *buf = NULL;
+
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_READ);
+	fu_progress_set_steps(progress, 1);
+
+	bufsz = STEELSERIES_SONIC_FIRMWARE_SIZE[chip];
+	buf = g_malloc0(bufsz);
```
hughsie, line 886: maybe GByteArray again?

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1238 @@
...
+	FuFirmware *firmware_chip;
+	SteelseriesSonicChip chip;
+	gsize offset;
+	gsize size;
+	g_autoptr(FuFirmware) firmware = fu_firmware_new();
+	g_autoptr(FuFirmware) firmware_nordic = NULL;
+	g_autoptr(FuFirmware) firmware_holtek = NULL;
+	g_autoptr(FuFirmware) firmware_mouse = NULL;
+
+	/* TODO: Use FuArchive */
```
hughsie, line 1089: still the plan?

hughsie:
Precommit has words:

    plugins/steelseries/fu-steelseries-sonic.c line 1046 got FALSE, which is not valid
    plugins/steelseries/fu-steelseries-sonic.c line 1059 got FALSE, which is not valid
    plugins/steelseries/fu-steelseries-sonic.c line 1046 got FALSE, which is not valid
    plugins/steelseries/fu-steelseries-sonic.c line 1059 got FALSE, which is not valid

plugins/steelseries/fu-steelseries-sonic.c
gportay: Indeed.

plugins/steelseries/fu-steelseries-sonic.c
gportay: Will check.

plugins/steelseries/fu-steelseries-sonic.c
gportay: Arf, do we? I was not aware about that word.

plugins/steelseries/fu-steelseries-sonic.c
gportay: Ah ah, I tried to be the most generic in case of the the plugin is reuse by device such as keyboard or headset.

So okay. I will be more specific, et even more if more products are added in the future.

plugins/steelseries/fu-steelseries-sonic.c
gportay: I gave a try and I failed.

How can I auto free the array and it's content?

plugins/steelseries/fu-steelseries-sonic.c
gportay: Oh, I would say yes.

I will replace the concatenated blob by a zip archive, and remove the `_dump_firmware()` because there is no write support yet for archive, and because it is wrong as it dumps blobs from different chips.

It should have been `_read_firmware()` right?

plugins/steelseries/fu-steelseries-sonic.c
hughsie: g_autoptr(GByteArray) buf = g_byte_array_new();

and then you can convert to a GBytes using `g_byte_array_free_to_bytes(g_steal_pointer(&buf))`

plugins/steelseries/fu-steelseries-sonic.c
hughsie: https://www.urbandictionary.com/define.php?term=dongle

plugins/steelseries/fu-steelseries-sonic.c
hughsie: > It should have been _read_firmware() right?

Ahh yes, indeed.

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1131 @@
...
+		g_autofree gchar *msg = NULL;
+
+		/* the user has to do something */
+		msg = g_strdup_printf("%s needs to be connected to start the update. "
+				      "Please put the switch button underneath to 2.4G, or "
+				      "click on any button to reconnect it.",
+				      fu_device_get_name(device));
+		request = fwupd_request_new();
+		fwupd_request_set_kind(request, FWUPD_REQUEST_KIND_IMMEDIATE);
+		fwupd_request_set_id(request, FWUPD_REQUEST_ID_PRESS_UNLOCK);
```
hughsie, line 716: I wonder if we want FWUPD_REQUEST_ID_PRESS_ANY_BUTTON

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1131 @@
...
+
+	/* mouse */
+	chip = STEELSERIES_SONIC_CHIP_MOUSE;
+	firmware_chip =
+	    fu_firmware_get_image_by_id(firmware, STEELSERIES_SONIC_FIRMWARE_ID[chip], error);
+	if (firmware_chip == NULL)
+		return NULL;
+	if (!fu_steelseries_sonic_parse_firmware(firmware_chip, flags, error))
+		return FALSE;
+	firmware_mouse = g_steal_pointer(&firmware_chip);
```
hughsie, line 1018: i'm unclear what you need firmware_mouse for... can we maybe do:

	g_autoptr(FuFirmware) firmware_chip1 = NULL;
	firmware_chip1 = fu_firmware_get_image_by_id(firmware, STEELSERIES_SONIC_FIRMWARE_ID[chip], error);
	if (firmware_chip1 == NULL)
		return NULL;

etc?

plugins/steelseries/fu-steelseries-sonic.c
gportay: This makes sense; but I do not know what it implies honestly; I mean how it is used then.

hughsie: Don't worry about it more; the idea is the frontend can display a *localised* version of FWUPD_REQUEST_ID_PRESS_ANY_BUTTON rather than the text specified here. I'm not 100% sure it's going to always work.

plugins/steelseries/fu-steelseries-sonic.c
gportay: Well, the idea have the exact same code for the three bloc of code.

I wanted to have:

``` c
g_autoptr(FuFirmware) firmwares[_LAST] = { NULL };

firmwares[chip] = fu_firmware_get_image_by_id(firmware, STEELSERIES_SONIC_FIRMWARE_ID[chip], error);
if (firmwares[chip] == NULL)
	return NULL;
```

I do no know if this is doable.

plugins/steelseries/fu-steelseries-sonic.c
gportay: Do you want me to add this?

plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -0,0 +1,1127 @@
...
+
+	/* dongle (nordic, holtek; same command) */
+	chip = STEELSERIES_SONIC_CHIP_HOLTEK;
+	if (!fu_steelseries_sonic_restart(device, chip, fu_progress_get_child(progress), error)) {
+		g_prefix_error(error, "failed to restart chip %u: ", chip);
+		return FALSE;
+	}
+	fu_progress_step_done(progress);
+
+	/* the user has to do something */
```
gportay, line 674: @hughsie Is that possible to not run that part if the update was cancel by `_prepare()`?

plugins/steelseries/fu-steelseries-sonic.c
hughsie: Nah, lets do it as a follow up.

plugins/steelseries/fu-steelseries-sonic.c
hughsie: Can you have a private flag perhaps? e.g. NEEDS_CLEANUP or something?

gportay: Ok, so there is nothing I can use from the internals.

My initial though was (re?)using an internal flag (`_UPDATE_SUCCEED`?); but I think I may want to ask the user to re-plug the dongle if the upgrade has failed. dunno.

hughsie: > My initial though was (re?)using an internal flag

I think relying on the device status might be too fragile. Can you do fu_device_to_string() on  failed update and a successful update when it gets to cleanup() and see what's different?

gportay:
@hughsie do you want I squash the commits?

hughsie:
pls

gportay:
squashed and rebased!

hughsie (approved):

gportay:
@hughsie Can we merge it?

hughsie:
@gportay sure, feel free to press the button -- any approval with all the CI checks green is good.

gportay:
> @gportay sure, feel free to press the button -- any approval with all the CI checks green is good.

Oh, I did not know that I am allowed to merged :)

To be sure, and to not screwed up the git history, I should run rebase and merged right? or squash and merge is fair?

hughsie:
> rebase and merged

This is preferred, thanks. I'd like to avoid merge commits if I can, but it's not the end of the world.

PR #4618, opened 2022-05-11
Allow specifying a dbus socket of tcp:host=localhost

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -826,6 +826,10 @@ fwupd_client_connect_async(FwupdClient *self,
 		} else {
 			socket_address = g_strdup(socket_filename);
 		}
+	} else {
+#ifdef _WIN32
+		socket_address = g_strdup("tcp:host=localhost,port=1341");
+#endif
 	}
 
 	/* use peer-to-peer only if the env variable is set */
--- a/src/fu-main.c
+++ b/src/fu-main.c
@@ -2381,6 +2381,10 @@ main(int argc, char *argv[])
 		} else {
 			socket_address = g_strdup(socket_filename);
 		}
+	} else {
+#ifdef _WIN32
+		socket_address = g_strdup("tcp:host=localhost,port=1341");
+#endif
 	}
 
 	/* own the object */


superm1 (approved):

hughsie:
Screenshot from 2022-05-11 15-35-58

libfwupd/fwupd-client.c
```diff
@@ -818,10 +819,18 @@ fwupd_client_connect_async(FwupdClient *self,
 		return;
 	}
 
-	/* use peer-to-peer only if the env variable is set */
+	/* convert from filename to address, if required */
```
superm1, line 822: I think you want to do something special here for
`#ifdef _WIN32` actually to make fwupd clients work more "automagically".  Maybe look if it's NULL then hardcode to `tcp:...`?

hughsie: That's a good idea. what port do we use tho? we have to set the same port on the client and the server, and I don't think there's any way to "discover" what port the daemon chose if you leave it to choose something unused.

src/fu-main.c
```diff
@@ -2368,23 +2369,32 @@ main(int argc, char *argv[])
 	}
 #endif
 
-	/* own the object */
+	/* convert from filename to address, if required */
```
superm1, line 2372: similar to my other comment - I think hardcode the daemon to listen on TCP for Windows here.

libfwupd/fwupd-client.c
superm1: How about 1341?  It seems like https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?=&skey=2&page=22 it's only used for Qubes today.

superm1:
Can you see if registering a service is basically just this: https://itsiti.com/create-delete-windows-service-using-sc-command/ ?  If so, I think you should register it as part of the MSI.

libfwupd/fwupd-client.c
hughsie: Good idea, fixed.

superm1 (approved):

superm1:
> @superm1 I think https://wixtoolset.org/documentation/manual/v3/xsd/wix/serviceinstall.html is probably what we want -- no idea if it works. OUTOFTIME for today.

yeah thoughts for the next PR.  This is awesome, great job!

PR #4620, opened 2022-05-12
Split out daemon startup into two files

hughsie:
The Win32 StartServiceCtrlDispatcher service startup is very different
from the usual Linux exec() mechanism.

Split out the D-Bus daemon into a new object so we can consume it from
a different binary in the future.

This looks like a big change but it's 99% just moving code around.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
It's hard to review with such a big diff, but yeah conceptually makes sense.

PR #4623, opened 2022-05-12
archive: Add writing support

gportay:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

This is an early work on implement the `_write()` support for `FuArchive` and `FuArchiveFirmware`.

### DIFF
--- a/libfwupdplugin/fu-archive-firmware.c
+++ b/libfwupdplugin/fu-archive-firmware.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2021 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2022 Gaël PORTAY <gael.portay@collabora.com>
  *
  * SPDX-License-Identifier: LGPL-2.1+
  */
@@ -10,6 +11,7 @@
 
 #include "fu-archive-firmware.h"
 #include "fu-archive.h"
+#include "fu-common.h"
 
 /**
  * FuArchiveFirmware:
@@ -19,7 +21,24 @@
  * See also: [class@FuFirmware]
  */
 
-G_DEFINE_TYPE(FuArchiveFirmware, fu_archive_firmware, FU_TYPE_FIRMWARE)
+typedef struct {
+	FuArchiveFormat format;
+	FuArchiveCompression compression;
+} FuArchiveFirmwarePrivate;
+
+G_DEFINE_TYPE_WITH_PRIVATE(FuArchiveFirmware, fu_archive_firmware, FU_TYPE_FIRMWARE)
+#define GET_PRIVATE(o) (fu_archive_firmware_get_instance_private(o))
+
+static void
+fu_archive_firmware_export(FuFirmware *firmware, FuFirmwareExportFlags flags, XbBuilderNode *bn)
+{
+	FuArchiveFirmware *self = FU_ARCHIVE_FIRMWARE(firmware);
+	FuArchiveFirmwarePrivate *priv = GET_PRIVATE(self);
+	fu_xmlb_builder_insert_kv(bn, "format", fu_archive_format_to_string(priv->format));
+	fu_xmlb_builder_insert_kv(bn,
+				  "compression",
+				  fu_archive_compression_to_string(priv->compression));
+}
 
 static gboolean
 fu_archive_firmware_parse_cb(FuArchive *self,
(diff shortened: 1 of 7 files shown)


gportay:
_Note_: I will revert the change about renaming the symbol `_new()` to `_new_with_bytes()`.

gportay:
I am looking for good names for the new symbols in `FuArchive`: especially for `_new()` -> `_{mutable,write,???}_new()`

libfwupdplugin/fu-archive-firmware.c
```diff
@@ -46,14 +51,45 @@ fu_archive_firmware_parse(FuFirmware *firmware,
 	g_autoptr(FuArchive) archive = NULL;
 
 	/* load archive */
-	archive = fu_archive_new(fw, FU_ARCHIVE_FLAG_IGNORE_PATH, error);
```
gportay, line 49: I will revert the renaming in the next iteration.

libfwupdplugin/fu-archive-firmware.c
```diff
@@ -46,14 +51,45 @@ fu_archive_firmware_parse(FuFirmware *firmware,
...
+			     GError **error)
+{
+	FuArchiveFirmwareIter *iter = (FuArchiveFirmwareIter *)user_data;
+	if (iter->index < iter->array->len) {
+		FuFirmware *img = g_ptr_array_index(iter->array, iter->index++);
+		*filename =  fu_firmware_get_id(img);
+		return fu_firmware_get_bytes(img, error);
+	}
+	*filename = NULL;
+	return g_bytes_new(NULL, 0);
```
gportay, line 76: I do no really like returning an empty `GBytes` to stop.

Any better idea?

libfwupdplugin/fu-archive.c
```diff
@@ -70,23 +72,23 @@ fu_archive_init(FuArchive *self)
 GBytes *
 fu_archive_lookup_by_fn(FuArchive *self, const gchar *fn, GError **error)
 {
-	GBytes *fw;
+	GBytes *bytes;
```
gportay, line 75: I will revert that change as well, and/or propose in another commit.

`fw` does not make sense in the `FuArchive`.

libfwupdplugin/fu-archive.c
```diff
@@ -162,7 +185,7 @@ fu_archive_load(FuArchive *self, GBytes *blob, FuArchiveFlags flags, GError **er
 		return FALSE;
 	}
 	while (TRUE) {
-		const gchar *fn;
+		const gchar *fn = NULL;
```
gportay, line 188: Will revert. the change was for the function `_save()`. Now I understand why it looped forever before I force `*filename = NULL` in the callback :facepalm:

libfwupdplugin/fu-archive.c
```diff
@@ -234,8 +257,136 @@ fu_archive_load(FuArchive *self, GBytes *blob, FuArchiveFlags flags, GError **er
...
 
+static GByteArray *
+fu_archive_save(FuArchive *self,
+		const gchar *compress,
+		FuArchiveEntryIterateFunc callback,
+		gpointer user_data,
+		GError **error)
+{
+#ifdef HAVE_LIBARCHIVE
+	const gchar *fn;
```
gportay, line 268: Reduce the scope to `while`; initialize to `= NULL`.

libfwupdplugin/fu-archive.c
```diff
@@ -234,8 +257,136 @@ fu_archive_load(FuArchive *self, GBytes *blob, FuArchiveFlags flags, GError **er
...
+	gsize blobsz;
+	size_t size;
+	int r;
+	g_autoptr(GByteArray) blob = NULL;
+	g_autoptr(_archive_write_ctx) arch = NULL;
+
+	blobsz = 0x20000;
+	blob = g_byte_array_sized_new(blobsz);
```
gportay, line 276: Find a way to resize buffer with libarchive.

libfwupdplugin/fu-archive-firmware.c
hughsie: Ya, I think fu_archive_new() is ABI -- although we could change `fw` to be NULL if you want to create an empty archive...

libfwupdplugin/fu-archive-firmware.c
hughsie: Maybe a gboolean in the `FuArchiveFirmwareIter` of `done` or something?

libfwupdplugin/fu-archive-firmware.c
```diff
@@ -46,14 +51,45 @@ fu_archive_firmware_parse(FuFirmware *firmware,
...
+{
+	FuArchiveFirmwareIter iter;
+	g_autoptr(FuArchive) archive = NULL;
+	g_autoptr(GPtrArray) images = fu_firmware_get_images(firmware);
+
+	/* save archive and compress each image to the archive */
+	iter.index = 0;
+	iter.array = images;
+	archive = fu_archive_new();
+	return fu_archive_get_bytes(archive, "gzip", fu_archive_firmware_write_cb, &iter, error);
```
hughsie, line 90: feels like we want something like `fu_archive_firmware_g/set_format` perhaps?

libfwupdplugin/fu-archive.c
hughsie: agree. `blob` is another name for a GBytes we use.

libfwupdplugin/fu-archive.c
```diff
@@ -70,23 +72,23 @@ fu_archive_init(FuArchive *self)
...
 	}
-	return fw;
+	return bytes;
 }
 
 /**
  * fu_archive_iterate:
  * @self: a #FuArchive
- * @callback: (scope call): a #FuArchiveIterateFunc.
+ * @callback: a #FuArchiveIterateFunc.
```
hughsie, line 91: ohh? is this not call scope?

libfwupdplugin/fu-archive.c
```diff
@@ -256,3 +407,29 @@ fu_archive_new(GBytes *data, FuArchiveFlags flags, GError **error)
...
+ * @callback: a #FuArchiveEntryIterateFunc.
+ * @user_data: user data
+ * @error: (nullable): optional return location for an error
+ *
+ * TODO
+ *
+ * Since: 1.8.1
+ **/
+GBytes *
+fu_archive_get_bytes(FuArchive *self,
```
hughsie, line 422: not sure about the name. perhaps fu_archive_write() or fu_archive_save_as_bytes()

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -204,7 +204,7 @@ fu_plugin_uefi_capsule_get_splash_data(guint width, guint height, GError **error
 	blob_archive = fu_common_get_contents_bytes(filename_archive, error);
 	if (blob_archive == NULL)
 		return NULL;
-	archive = fu_archive_new(blob_archive, FU_ARCHIVE_FLAG_NONE, error);
+	archive = fu_archive_new_from_bytes(blob_archive, FU_ARCHIVE_FLAG_NONE, error);
```
hughsie, line 207: not convinced with the API break :)

gportay: ah ah, for me it makes sense to append `_from_bytes` to the original symbol `_new`

libfwupdplugin/fu-archive.c
gportay: why not

libfwupdplugin/fu-archive.c
gportay: nah, just forget about it

libfwupdplugin/fu-archive-firmware.c
gportay: sounds interesting.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: right, and if it was a new API I'd agree, but I think we have to work with what we've got.

gportay: yes, that the pity part :)

hughsie:
> to dump only the firmware part from the command line

You mean write only the blob that fwupdtool install-blob would write for the subdevice? If so, I don't think so. If the device isn't exposing actual FuDevices as child devices then the daemon only "sees" one thing and only gets one blob out.

gportay:
> You mean write only the blob that fwupdtool install-blob would write for the subdevice? If so, I don't think so. If the device isn't exposing actual FuDevices as child devices then the daemon only "sees" one thing and only gets one blob out.

Yes, dumping exactly what `install-blob` writes.

Could we thing about a `read-blob`-ish command in `fwupdtool`? does it make sense for other developers?
 - `firmware-dump` -> `_dump_firmware()` -> `FILE`
 - `firmware-read` -> `_read_firmware()` -> `FILE`

Because `firmware-dump` uses `_dump_firmware()` is supposed to dump the full chip memory and I can read only the firmware blob memory.

hughsie:
> does it make sense for other developers

It seemed pretty easy to provide, do you mean something like https://github.com/fwupd/fwupd/pull/4647

libfwupdplugin/fu-archive-firmware.c
gportay, line 90: Leaking:

```
==339333== 40 bytes in 1 blocks are definitely lost in loss record 1,770 of 2,958
==339333==    at 0x4845888: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==339333==    by 0x4C13029: g_malloc (gmem.c:125)                               
==339333==    by 0x4C2AAE6: g_slice_alloc (gslice.c:1072)                       
==339333==    by 0x4BD71FA: g_bytes_new_with_free_func (gbytes.c:186)           
==339333==    by 0x491C810: fu_archive_firmware_write_cb (fu-archive-firmware.c:76)
==339333==    by 0x48DB8D3: fu_archive_save (fu-archive.c:306)                  
==339333==    by 0x48DBDAB: fu_archive_get_bytes (fu-archive.c:431)             
==339333==    by 0x491C88A: fu_archive_firmware_write (fu-archive-firmware.c:90)
==339333==    by 0x4904606: fu_firmware_write (fu-firmware.c:1124)              
==339333==    by 0x12798A: fu_util_firmware_read (fu-tool.c:1179)               
==339333==    by 0x132247: fu_util_cmd_array_run (fu-util-common.c:633)         
==339333==    by 0x12F2A7: main (fu-tool.c:3843)                                
```

libfwupdplugin/fu-archive.c
```diff
@@ -234,6 +257,137 @@ fu_archive_load(FuArchive *self, GBytes *blob, FuArchiveFlags flags, GError **er
 #endif
 }
 
+static GByteArray *
+fu_archive_save(FuArchive *self,
+		const gchar *compress,
```
hughsie, line 262: maybe `compression_type` or `compress_format` perhaps? We've used compress in other places as a boolean.

gportay: I am creating an enum

hughsie: With what contents? Isn't the underlying archive format pretty free text?

hughsie: If you push what you have now, I've got a few other ideas how the write stuff could work too.

libfwupdplugin/fu-archive.c
hughsie: I can do this if you want.

libfwupdplugin/fu-archive.c
gportay: well, I did not want to rely on their strings

libfwupdplugin/fu-archive.c
gportay: okay

gportay:
I will give a try to various combination.

I thing we should add some ~~automated~~ tests. Could you prepare the skeleton and I will do the rest?

libfwupdplugin/fu-archive.c
```diff
@@ -563,11 +563,11 @@ fu_archive_new(GBytes *data, FuArchiveFlags flags, GError **error)
 
 #ifdef HAVE_LIBARCHIVE
 static gssize
-fu_archive_write_cb(struct archive *arch, void *client_data, const void *buffer, gsize length)
+fu_archive_write_cb(struct archive *arch, void *user_data, const void *buf, gsize bufsz)
 {
-	GByteArray *blob = (GByteArray *)buffer;
```
gportay: Does it fix the segmentation fault?

I will test it tomorrow :)

gportay:
> @gportay something like ./src/fwupdtool firmware-build ../libfwupdplugin/tests/archive.builder.xml foo.bin seems to work rather well. I'm no sure how testable it is, as the exact binary output is going to be slightly different on each system. But it works!

Indeed, we check for the magic of the compressed algo.

libfwupdplugin/fu-archive.c
gportay: It did!

Segfault fixed! Thanks.

hughsie:
@gportay if you want to rebase and squash this it basically LGTM now.

gportay:
I will do it, later; I am thinking about renaming some compression constants and filter to compression. what do you think? or later maybe? at least before the next release :)

gportay:
I have renamed some formats and rename filter to compression; :crossed_fingers: I did break nothing...

hughsie (approved):

PR #4627, opened 2022-05-13
Use StartServiceCtrlDispatcherA for the daemon on Windows

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-daemon.c
+++ b/src/fu-daemon.c
@@ -2110,11 +2110,9 @@ fu_daemon_set_machine_kind(FuDaemon *self, FuDaemonMachineKind machine_kind)
 }
 
 gboolean
-fu_daemon_setup(FuDaemon *self, GError **error)
+fu_daemon_setup(FuDaemon *self, const gchar *socket_address, GError **error)
 {
 	const gchar *machine_kind = g_getenv("FWUPD_MACHINE_KIND");
-	const gchar *socket_filename = g_getenv("FWUPD_DBUS_SOCKET");
-	g_autofree gchar *socket_address = NULL;
 
 	g_return_val_if_fail(FU_IS_DAEMON(self), FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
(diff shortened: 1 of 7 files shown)


src/fu-main-windows.c
```diff
@@ -0,0 +1,138 @@
...
+
+	/* parse debugging args */
+	g_option_context_add_group(context, fu_debug_get_option_group());
+	if (!g_option_context_parse(context, &argc, &argv, &error)) {
+		g_printerr("Failed to parse command line: %s\n", error->message);
+		return EXIT_FAILURE;
+	}
+
+	/* set up the daemon, which includes coldplugging devices -- then run it */
+	if (!fu_daemon_setup(daemon, &error)) {
```
superm1, line 115: I wonder if you want to pass in here something about what socket to listen on. It's less ifdef code further on then.

hughsie: Like that^^?

superm1: Yeap that looks good

src/fu-daemon.c
```diff
@@ -2110,11 +2110,9 @@ fu_daemon_set_machine_kind(FuDaemon *self, FuDaemonMachineKind machine_kind)
 }
 
 gboolean
-fu_daemon_setup(FuDaemon *self, GError **error)
+fu_daemon_setup(FuDaemon *self, const gchar *socket_address, GError **error)
 {
 	const gchar *machine_kind = g_getenv("FWUPD_MACHINE_KIND");
-	const gchar *socket_filename = g_getenv("FWUPD_DBUS_SOCKET");
```
superm1, line 2116: I think this environment variable should be dropped in matching readme.

Also the client should probably use the same define.

hughsie: I think we still want it in Linux for the socket self tests, no?

superm1: Oh I guess still want to use unix socket in Linux, right ok.

Issue #4629, opened 2022-05-14
List updates that require AC power even if power isn't connected


nullr0ute:
**Is your feature request related to a problem? Please describe.**
When running "fwupdmgr get-updates" it doesn't list available updates that need a AC power connect. The get-updates option doesn't actually apply updates so there's not a risk of updating while low power.

**Describe the solution you'd like**
It's useful to be able to see all updates that are available.

**Describe alternatives you've considered**
None

**Additional context**

Output without power connected to a Lenovo X1 gen9
````
# fwupdmgr get-updates
Devices with no available firmware updates: 
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • WDC WDS100T2B0C
 • Prometheus IOTA Config
 • UEFI dbx
Devices with the latest available firmware version:
 • Prometheus
Devices not currently updatable:
 • Embedded Controller — Cannot install update when not on AC power
 • Intel Management Engine — Cannot install update when not on AC power
 • System Firmware — Cannot install update when not on AC power
 • UEFI Device Firmware — Cannot install update when not on AC power
No updates available
````

With power attached:
````
# fwupdmgr get-updates
Devices with no available firmware updates: 
 • Prometheus IOTA Config
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI Device Firmware
 • UEFI dbx
 • WDC WDS100T2B0C
Devices with the latest available firmware version:
 • Prometheus
20XXS3HC03
│
├─Embedded Controller:
│ │   Device ID:          5bfc5b91c18ef8b751d3d052af77b69ebf738038
│ │   Summary:            UEFI ESRT device
│ │   Current version:    0.1.31
│ │   Minimum Version:    0.1.31
│ │   Vendor:             Lenovo (DMI:LENOVO)
│ │   Update State:       Success
│ │   GUID:               61b65ccc-0116-4b62-80ed-ec5f089ae523
│ │   Device Flags:       • Internal device
│ │                       • Updatable
│ │                       • System requires external power source
│ │                       • Supported on remote server
[...]

PR #4632, opened 2022-05-16
Allow clients to opt-in to showing updates with inhibits

hughsie:
When typing 'fwupdmgr get-updates' show the updates that *could* be
installed if the inhibit was removed.

Do not unconditionally do this, as some clients such as gnome-software
assume that all updates returned by GetUpdates() are updatable with
Install() -- and in a GUI we only want to show the updates we can apply
*right now*. When the inhibit is removed (e.g. AC power is connected)
the GUI client will notify the user as required, unlike a CLI tool.

Fixes https://github.com/fwupd/fwupd/issues/4629

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -74,6 +74,7 @@ typedef enum {
  * @FWUPD_FEATURE_FLAG_REQUESTS:		Can show interactive requests
  * @FWUPD_FEATURE_FLAG_FDE_WARNING:		Can warn about full disk encryption
  * @FWUPD_FEATURE_FLAG_COMMUNITY_TEXT:		Can show information about community supported
+ * @FWUPD_FEATURE_FLAG_SHOW_PROBLEMS:		Can show problems when getting the update list
  *
  * The flags to the feature capabilities of the front-end client.
  **/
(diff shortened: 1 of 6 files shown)


libfwupd/fwupd-enums.h
```diff
@@ -86,6 +87,7 @@ typedef enum {
 	FWUPD_FEATURE_FLAG_REQUESTS = 1 << 4,	    /* Since: 1.6.2 */
 	FWUPD_FEATURE_FLAG_FDE_WARNING = 1 << 5,    /* Since: 1.7.1 */
 	FWUPD_FEATURE_FLAG_COMMUNITY_TEXT = 1 << 6, /* Since: 1.7.5 */
+	FWUPD_FEATURE_FLAG_SHOW_INHIBITS = 1 << 7,  /* Since: 1.7.5 */
```
superm1, line 90: shouldn't this be 1.8.x?

hughsie: It should indeed, fixed.

superm1:
It's also getting translated strings for the reason.  If the plugin has free-form text, then you can't get a translated string to the user about what to do to fix.

hughsie:
Maybe `fwupd_device_add_problem()` makes more sense; `inhibit->inhibit = inhibit` makes my head hurt a bit. On the flip side, maybe the same name makes sense, but with a qualifier; e.g. `FwupdDeviceInhibitKind` and `FWUPD_DEVICE_INHIBIT_KIND_SYSTEM_POWER_TOO_LOW`

hughsie:
> I think with that we don't need FWUPD_FEATURE_FLAG_SHOW_INHIBITS.

Nope, we still need this functionality. If this feature flag works then I'll add the bits to gnome-software and gnome-firmware to show the problems.

superm1:
> > I think with that we don't need FWUPD_FEATURE_FLAG_SHOW_INHIBITS.
> 
> Nope, we still need this functionality. If this feature flag works then I'll add the bits to gnome-software and gnome-firmware to show the problems.

Show problems probably instead of show inhibits.

hughsie:
> Show problems probably instead of show inhibits.

Yup, I've renamed it in the PR already. I'll do the gnome-firmware bits today and then if that works I guess this PR makes sense.

hughsie:
@superm1 this is the gnome-firmware MR; it's somewhat hypnotic to click on the UEFI entry and open and close the laptop lid and all the UI does the right thing... https://gitlab.gnome.org/World/gnome-firmware/-/merge_requests/87

superm1 (approved):

PR #4635, opened 2022-05-16
Specify an enumerated kind when inhibiting devices

hughsie:
This allows us to make smarter policy decisions in the future on when
to show unavailable updates.

Only inhibits the user can "fix" are enumerated. For example, opening
the laptop lid, or charging the device battery.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -83,6 +83,7 @@ typedef struct {
 } FuDeviceRetryRecovery;
 
 typedef struct {
+	FwupdDeviceInhibitKind inhibit_kind;
 	gchar *inhibit_id;
 	gchar *reason;
 } FuDeviceInhibit;
@@ -1262,7 +1263,10 @@ fu_device_add_child(FuDevice *self, FuDevice *child)
 		g_autoptr(GList) values = g_hash_table_get_values(priv->inhibits);
 		for (GList *l = values; l != NULL; l = l->next) {
 			FuDeviceInhibit *inhibit = (FuDeviceInhibit *)l->data;
-			fu_device_inhibit(child, inhibit->inhibit_id, inhibit->reason);
+			fu_device_inhibit_full(child,
+					       inhibit->inhibit_kind,
+					       inhibit->inhibit_id,
+					       inhibit->reason);
 		}
 	}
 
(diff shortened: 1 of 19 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -2676,16 +2686,18 @@ fu_device_ensure_inhibits(FuDevice *self)
...
  *
- * Since: 1.6.0
+ * Since: 1.8.1
  **/
 void
-fu_device_inhibit(FuDevice *self, const gchar *inhibit_id, const gchar *reason)
+fu_device_inhibit_full(FuDevice *self,
+		       FwupdDeviceInhibitKind inhibit_kind,
+		       const gchar *inhibit_id,
+		       const gchar *reason)
```
superm1, line 2695: do you really need the "reason" string still?

hughsie: I think we do; e.g. for `"Cannot install update when system battery is not at least %u%%"`

superm1: But you want that to be translatable, so shouldn't the client put it together?

hughsie: Urg, but that would mean exporting fu_device_get_battery_level() and fu_device_get_battery_threshold() into libfwupd -- plus fu_context_get_battery_level() and fu_context_get_battery_threshold() -- that seems like an awful lot of extra exports, no? If that's what we've got to do I guess that's fine, but urgh.

superm1: I mean; it comes down to how useful the information is.  If we're going to be showing a message in the client with accurate info, we should make sure the client can build the same accurate info.  The alternative is a generic `"Cannot install update until the system battery is charged"`.

Note: I don't feel strongly either way whether to show the number, but I do think either it needs to show it and translate it or not show it and translate the generic string.

hughsie: I don't even know *how* we'd wire up the information in FuContext into FwupdClient and then use the right thing in fu_util_device_to_string(). Maybe FuContext should subclass FwupdClient logically -- although there doesn't seem to be a ton of overlap and changes the ABI. Perhaps we should synthesize a FwupdClient for fwupdtool and use something like g_object_bind_property()?

hughsie: Also, if the description is going to be NULL for the enumerated inhibit kinds, perhaps it should be `s/fu_device_inhibit_full/fu_device_inhibit_kind` instead, and only take an enum rather than all three variables.

hughsie:
Closed in preference to https://github.com/fwupd/fwupd/pull/4650

PR #4647, opened 2022-05-18
Add a way to read the device firmware in fwupdtool

hughsie:
In most cases 'fwupdtool firmware-dump' and 'fwupdtool firmware-read'
are going to be the same. This isn't true in all cases, especially when
dealing with composite firmware like archives.

See https://github.com/fwupd/fwupd/pull/4623#issuecomment-1129227133

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


src/fu-tool.c
```diff
@@ -1117,6 +1117,71 @@ fu_util_firmware_dump(FuUtilPrivate *priv, gchar **values, GError **error)
...
+		if (device == NULL)
+			return FALSE;
+	}
+	priv->current_operation = FU_UTIL_OPERATION_READ;
+	g_signal_connect(FU_ENGINE(priv->engine),
+			 "device-changed",
+			 G_CALLBACK(fu_util_update_device_changed_cb),
+			 priv);
+
+	/* dump firmware */
```
gportay, line 1175: s,dump,read?

gportay, line 1175: s,dump,read?

gportay (approved):
LGTM, despite the nitpick.

I have tested locally with the changes for write support in FuArchiveFirmware on SS Rival 3 Wireless.

src/fu-tool.c
hughsie: agree, thanks, fixed.

PR #4650, opened 2022-05-19
Specify an enumerated problem when inhibiting devices

hughsie:
This allows us to make smarter policy decisions in the future on when
to show unavailable updates. It also means we can show translated
text in the frond-end clients.

Only problems the user can "fix" are enumerated. For example, opening
the laptop lid, or charging the device battery.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-device.c
+++ b/libfwupd/fwupd-device.c
@@ -34,6 +34,7 @@ typedef struct {
 	guint64 created;
 	guint64 modified;
 	guint64 flags;
+	guint64 problems;
 	GPtrArray *guids;
 	GPtrArray *vendor_ids;
 	GPtrArray *protocols;
@@ -86,6 +87,7 @@ enum {
 	PROP_UPDATE_IMAGE,
 	PROP_BATTERY_LEVEL,
 	PROP_BATTERY_THRESHOLD,
+	PROP_PROBLEMS,
 	PROP_LAST
 };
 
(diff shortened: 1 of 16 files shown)


libfwupd/fwupd-enums.h
```diff
@@ -539,6 +539,78 @@ typedef enum {
...
+ */
+#define FWUPD_DEVICE_PROBLEM_SYSTEM_POWER_TOO_LOW (1u << 0)
+/**
+ * FWUPD_DEVICE_PROBLEM_UNREACHABLE:
+ *
+ * The device is unreachable, or out of wireless range.
+ *
+ * Since 1.8.1
+ */
+#define FWUPD_DEVICE_PROBLEM_UNREACHABLE (1u << 1)
```
superm1, line 565: Thinking forward this can be something to use for transiently connected USB devices too. Like if there is an update available for a dongle connected to your work dock only.

libfwupd/fwupd-enums.h
```diff
@@ -539,6 +539,78 @@ typedef enum {
...
+ * This problem is not defined, this typically will happen from mismatched
+ * fwupd library and clients.
+ *
+ * Since 1.8.1
+ */
+#define FWUPD_DEVICE_PROBLEM_UNKNOWN G_MAXUINT64
+/**
+ * FwupdDeviceProblem:
+ *
+ * Inhibits are reasons why the device is not updatable.
```
superm1, line 610: Wrong description

libfwupd/fwupd-device.c
```diff
@@ -1662,6 +1664,107 @@ fwupd_device_has_flag(FwupdDevice *self, FwupdDeviceFlags flag)
...
+ * Since: 1.8.1
+ **/
+void
+fwupd_device_add_problem(FwupdDevice *self, FwupdDeviceProblem problem)
+{
+	FwupdDevicePrivate *priv = GET_PRIVATE(self);
+	g_return_if_fail(FWUPD_IS_DEVICE(self));
+	if (problem == 0)
+		return;
+	if ((priv->problems | problem) == priv->problems)
```
superm1, line 1721: Shouldn't this be &

libfwupd/fwupd-device.c
```diff
@@ -1662,6 +1664,107 @@ fwupd_device_has_flag(FwupdDevice *self, FwupdDeviceFlags flag)
...
+ * @problem: the #FwupdDeviceProblem
+ *
+ * Finds if the device has a specific device problem kind.
+ *
+ * Returns: %TRUE if the problem is set
+ *
+ * Since: 1.8.1
+ **/
+gboolean
+fwupd_device_has_problem(FwupdDevice *self, FwupdDeviceProblem problem)
```
superm1, line 1761: If you move this function up you could use it in add/remove.

src/fu-util-common.c
```diff
@@ -1304,6 +1304,55 @@ fu_util_update_state_to_string(FwupdUpdateState update_state)
...
+			return g_strdup(_("System power is too low to perform the update"));
+		}
+		return g_strdup_printf(
+		    /* TRANSLATORS: as in laptop battery power */
+		    _("System power is too low to perform the update (%u%%, requires %u%%)"),
+		    fwupd_client_get_battery_level(client),
+		    fwupd_client_get_battery_threshold(client));
+	}
+	if (problem == FWUPD_DEVICE_PROBLEM_UNREACHABLE) {
+		/* TRANSLATORS: for example, a BlueTooth mouse that is in powersave mode */
```
superm1, line 1327: I don't think Bluetooth needs the uppercase T

libfwupd/fwupd-enums.h
hughsie: Yup, but there be dragons -- in that you need a way to "forget" devices like this too.

libfwupd/fwupd-device.c
hughsie: No, as it's comparing it with itself, but it's a weird construct, and we can just use fwupd_device_has_problem instead....

libfwupd/fwupd-device.c
hughsie: No need to move it, it's in the header. But I've used it in both, like you suggested.

src/fu-util-common.c
hughsie: Indeed, all three fixed. thanks.

superm1 (approved):

PR #4656, opened 2022-05-21
steelseries: sonic: Allow to read firmware

gportay:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

This is an example on how to use the write support for an `FuArchiveFirmware`.

### DIFF
--- a/plugins/steelseries/fu-steelseries-sonic.c
+++ b/plugins/steelseries/fu-steelseries-sonic.c
@@ -612,6 +612,54 @@ fu_steelseries_sonic_restart(FuDevice *device,
 }
 
 static gboolean
+fu_steelseries_sonic_wait_for_device(FuDevice *device, GError **error)
+{
+	SteelseriesSonicWirelessStatus wl_status;
+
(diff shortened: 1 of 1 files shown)


gportay:
Ideally, I would like to remove the need for the flag `FWUPD_DEVICE_FLAG_CAN_VERIFY_IMAGE` when `firmware-read`-ing (`firmware-dump`-ing as well?) but keep it for `verify`-ing.

I need to investigate the code for that.

BTW, the `verify` function checks for the cheksum stored in the `.cab` file? quid for Archive Firmwares? is that possible to update LVFS for run checksum on internal files? or this is going too far?

hughsie:
> Ideally, I would like to remove the need for the flag FWUPD_DEVICE_FLAG_CAN_VERIFY_IMAGE

This tells the front-end that we can read the entire image; I don't think it's needed for fwpud**tool**, right?

> BTW, the verify function checks for the cheksum stored in the .cab file?

The *device checksum* yes, but only if CAN_VERIFY_IMAGE is set IIRC. Not sure how well that's going to work with an archive.

Richard

hughsie (approved):

PR #4661, opened 2022-05-22
Allow a plugin to only use the RemoveDelay when waiting for replug

hughsie:
Only opt-in plugins that have been tested -- unconditionally enabling
this may cause regressions on devices like docks.

Fixes https://github.com/fwupd/fwupd/issues/4378

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.h
+++ b/libfwupdplugin/fu-device.h
@@ -458,6 +458,16 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_AUTO_PAUSE_POLLING (1ull << 24)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_ONLY_WAIT_FOR_REPLUG:
+ *
+ * Only use the device removal delay when explicitly waiting for a replug, rather than every time
+ * the device is removed.
+ *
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-device.h
```diff
@@ -458,6 +458,15 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_AUTO_PAUSE_POLLING (1ull << 24)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_ONLY_WFR_REMOVE_DELAY:
```
superm1, line 462: don't like this name, any better?

hughsie: `NO_DELAYED_REMOVAL`? `NO_REMOVAL_DELAY`? `ONLY_WAIT_FOR_REPLUG`? `STRICT_WAIT_FOR_REPLUG`?

superm1: I like ONLY_WIAT_FOR_REPLUG

hughsie: Ack, fixed.

superm1 (approved):

PR #4666, opened 2022-05-24
Wip/gportay/trivial changes

gportay:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

I am doing lots of mistakes during development which ends in returning `FALSE` without setting any error; mostly because of wrong casting device.

There is may places in the code where the `error` pointer is left untested before getting de-referenced.

I have fixed the `SEGFAULT`s for most common vfuncs; `_probe()` ,`_setup()`, and `_attach()`, `_detach()`, `_write_firmware()`.

I can be more exhaustive is needed.


src/fu-engine.c
```diff
@@ -6402,6 +6402,12 @@ fu_engine_backend_device_added_cb(FuBackend *backend, FuDevice *device, FuEngine
 		if (plugin == NULL)
 			continue;
 		if (!fu_plugin_runner_backend_device_added(plugin, device, &error)) {
+			/* sanity check */
+			if (error == NULL) {
+				g_warning("failed to add device %s: exec failed but no error set!",
```
hughsie, line 6407: maybe g_critical?

hughsie:
> before getting de-referenced

So for me that crash is actually a feature; it means I can backtrace on exactly where the error isn't being set. Perhaps we should wrap these checks around `#ifdef SUPPORTED_BUILD` or something?

gportay:
> > before getting de-referenced
> 
> So for me that crash is actually a feature; it means I can backtrace on exactly where the error isn't being set. Perhaps we should wrap these checks around `#ifdef SUPPORTED_BUILD` or something?

Well, yes and no because it is hard to follow:

For example: returning `FALSE` in `_setup()` ends with:

```
sudo valgrind fwupdtool install-blob dongle.bin-1.93 291b0582-8d08-5714-88db-986911c744d9 
(...)
==2098738== Invalid read of size 8
==2098738==    at 0x1530AD: fu_engine_backend_device_added_cb (fu-engine.c:6413)
==2098738==    by 0x4B6AF8E: g_closure_invoke (gclosure.c:830)
==2098738==    by 0x4B86BA7: signal_emit_unlocked_R (gsignal.c:3743)
==2098738==    by 0x4B885C6: g_signal_emit_valist (gsignal.c:3496)
==2098738==    by 0x4B88823: g_signal_emit (gsignal.c:3553)
==2098738==    by 0x48DE0F5: fu_backend_device_added (fu-backend.c:65)
==2098738==    by 0x164E58: fu_usb_backend_device_added_cb (fu-usb-backend.c:57)
==2098738==    by 0x4B6AF8E: g_closure_invoke (gclosure.c:830)
==2098738==    by 0x4B86BA7: signal_emit_unlocked_R (gsignal.c:3743)
==2098738==    by 0x4B885C6: g_signal_emit_valist (gsignal.c:3496)
==2098738==    by 0x4B88823: g_signal_emit (gsignal.c:3553)
==2098738==    by 0x4E26EB4: g_usb_context_enumerate (in /usr/lib/libgusb.so.2.0.10)
```

Or ends with if returning `FALSE` in `_prepare()`, `_write_firmware()`, `_attach()`:

```
sudo valgrind fwupdtool install-blob dongle.bin-1.93 291b0582-8d08-5714-88db-986911c744d9 
(...)
==2101390== Invalid read of size 8
==2101390==    at 0x1302C1: main (fu-tool.c:3830)
```

I never reached the original code using valgrind.

hughsie (approved):

PR #4669, opened 2022-05-24
Wip/gportay/steelseries changes

gportay:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/steelseries/fu-steelseries-fizz.c
+++ b/plugins/steelseries/fu-steelseries-fizz.c
@@ -14,6 +14,9 @@
 #define STEELSERIES_FIZZ_FILESYSTEM_RECEIVER 0x01U
 #define STEELSERIES_FIZZ_FILESYSTEM_MOUSE    0x02U
 
+#define STEELSERIES_FIZZ_RESET_MODE_NORMAL     0x00U
+#define STEELSERIES_FIZZ_RESET_MODE_BOOTLOADER 0x01U
+
 #define STEELSERIES_FIZZ_RECEIVER_FILESYSTEM_MAIN_BOOT_ID	  0x01U
 #define STEELSERIES_FIZZ_RECEIVER_FILESYSTEM_FSDATA_FILE_ID	  0x02U
 #define STEELSERIES_FIZZ_RECEIVER_FILESYSTEM_FACTORY_SETTINGS_ID  0x03U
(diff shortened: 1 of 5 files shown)


plugins/steelseries/fu-steelseries-sonic.c
```diff
@@ -247,7 +253,10 @@ fu_steelseries_sonic_read_from_flash(FuDevice *device,
 					    0x0,
 					    STEELSERIES_BUFFER_FLASH_TRANSFER_SIZE);
 	fu_progress_set_id(progress, G_STRLOC);
-	fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_READ);
+	if (verify)
+		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_VERIFY);
+	else
+		fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_READ);
```
hughsie, line 259: rather than this, could you set the status on the "parent" FuProgress?

gportay: Could we imaging that the child inherits from the parent value at creation?

gportay: what is the function to get parent?

hughsie: Nah, I mean set the progress *before* calling fu_steelseries_sonic_read_from_flash

gportay: Hum, I need some time to convince myselt it is the way to go.

For now, my brain is blocking on doing:

```
fu_progress_set_status(progress, ...);
fu_progress_set_steps(progress, chunks->len);
```
where `...` is either `fu_progress_get_status(fu_progress_get_parent(progress)));` or `fu_progress_get_parent_status(progress)`.

so the code is following "almost" the same pattern.

Or, not having the `fu_progress_set_status(progress, ...);` because it is kind of inherited from its parent at creation.

gportay: Cool

gportay: [bff9c5b](https://github.com/fwupd/fwupd/pull/4669/commits/bff9c5b3188994525e2da478da9ca85e7085bf5a) does the trick...

I have drop the `READ`/`VERIFY` status, and I can go even more further! `ERASE/WRITE`. to be continued

plugins/steelseries/fu-steelseries-fizz.c
```diff
@@ -527,8 +529,13 @@ fu_steelseries_fizz_setup(FuDevice *device, GError **error)
 	guint32 stored_crc;
 	guint8 fs = STEELSERIES_FIZZ_FILESYSTEM_MOUSE;
 	guint8 id = STEELSERIES_FIZZ_MOUSE_FILESYSTEM_BACKUP_APP_ID;
+	g_autoptr(GError) error_local = NULL;
```
hughsie, line 532: used?

hughsie (dismissed):
One tiny nit, then LGTM.

plugins/steelseries/fu-steelseries-sonic.c
gportay: I could save one `_set _status(..., WRITE)` in the end.

plugins/steelseries/fu-steelseries-fizz.c
gportay: Oh, I have to check that

gportay: fixed

hughsie (approved):

PR #4675, opened 2022-05-25
trivial: Use API from flashrom >= 1.3 to show user progress

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Lets re-open this if the libflashrom API ever appears. At the moment my gut feeling is that we want to encourage projects just to push low-level plugins into fwupd rather than rely on flashrom in any meaningful way.

PR #4676, opened 2022-05-25
modem-manager: Fix one critical warning when updating

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -1192,6 +1192,7 @@ fu_mm_device_write_firmware_mbim_qdu(FuDevice *device,
 	g_autofree gchar *version = NULL;
 	g_autoptr(FuArchive) archive = NULL;
 	g_autoptr(FuDeviceLocker) locker = NULL;
+	g_autoptr(GArray) digest = NULL;
 	g_autoptr(XbBuilder) builder = xb_builder_new();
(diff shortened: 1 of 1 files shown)


aleksander0m:
Not sure I get what the change is trying to fix? Why was the MMManager object added in `fu_mm_device_udev_new()` not good?

aleksander0m:
I see, this is an attempt to fix https://github.com/fwupd/fwupd/discussions/4674#discussioncomment-2821300 right?
Couldn't it be that the `MMManager` pointer is actually a valid one, not NULL, and we're missing some ref somewhere? i.e. the warning happens because it was already disposed by the time we try to ref it again.

aleksander0m:
Oh, I see the manager pointer may indeed be NULL, all the new "incorporate" logic is new to my eyes.
Why not just `self->manager = g_object_ref (donor->manager);` in `fu_mm_device_incorporate()`?

hughsie:
> Why not just self->manager = g_object_ref (donor->manager); in fu_mm_device_incorporate()?

Ya, that's what I did first. g_set_object is slightly "safer" in that it will unref the destination if it's already somehow set, and also skip the operation if the source is the same as the destination.

clobrano:
Just tested, it fix https://github.com/fwupd/fwupd/discussions/4674#discussioncomment-2820919 indeed :+1:

Issue #4649, opened 2022-05-19
There is a  “Decompressing” progress bar when the first time to open Gnome-firmware software.


gaby-xie:
**Describe the bug**
There is a  “Decompressing” progress bar when the first time to open Gnome-firmware software.

**Steps to Reproduce**
1) Install Fedora 36
2)Update Fedora 36
   $sudo dnf update -y 
3)Install gnome-firmware
   $sudo dnf install gnome-firmware
4)Enable Testing State.
  $sudo dnf copr enable rhughes/fwupd 
  $sudo dnf update fwupd
  $sudo /sbin/shutdown -r now 
  $sudo /usr/bin/fwupdmgr --force refresh 
  $/usr/bin/fwupdmgr enable-remote lvfs-testing
5)Run gnome-firmware and test firmware update.
6)Open the gnome-firmware software by click Firmware icon after dock FW update successful. =>Find the “Decompressing” progress bar issue.
7)Then we launch gnome-firmware in terminal, can also duplicate this issue.

**Expected behavior**
No “Decompressing” progress bar show in gnome-firmware interface.

**fwupd version information**
fwupd version is 1.8.1.

gaby-xie:
![1 (1)](https://user-images.githubusercontent.com/105832528/169219969-bfb3e2bd-f763-40f0-8fb9-641ab9305043.png)

chlee75:
@hughsie 
Please check Gaby's issue. It seems to be a simple gnome-firmware UI issue and "Decompressing ..." message disappears when pressing "Refresh Metadata" menu in the gnome-firmware.

hughsie:
@gaby-xie hey, thanks for reporting this. Can you get me the output from `gnome-firmware --verbose` please when you can see the problem. i.e. install the firmware, but make sure gnome-firmware is closed and then open up gnome firmware on the console with the verbose flag, and then copy and paste the verbose log into this bug. I want to see if this is a fwupd bug where the status is being left incorrectly in the decompressing state, or whether the front-end client is not actually getting the status value at startup. Thanks.

gaby-xie:
![Screenshot from 2022-05-20 02-36-28](https://user-images.githubusercontent.com/105832528/169468900-38d0fc3b-aa73-435f-8d44-63dcb30a858d.png)

gaby-xie:
@hughsie Please check the verbose log info. Thanks!

chlee75:
@gaby-xie 
Could you share the log files to Richard?
[...]

PR #4678, opened 2022-05-26
Store the last-set status in the daemon

hughsie:
We emit to the client from the daemon, and also proxy the status from
the engine, i.e. we're setting the same property in two places.

When the client requests the last-set status, recover the value sent
from the lowest layer (the daemon) rather than the engine. The former
gets reset back to IDLE automatically, the latter does not.

Fixes https://github.com/fwupd/fwupd/issues/4649

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-daemon.c
+++ b/src/fu-daemon.c
@@ -58,6 +58,7 @@ struct _FuDaemon {
 #ifdef HAVE_POLKIT
 	PolkitAuthority *authority;
 #endif
+	FwupdStatus status; /* last emitted */
 	guint owner_id;
 	guint process_quit_id;
 	FuEngine *engine;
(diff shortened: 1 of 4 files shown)


hughsie:
@chlee75 I've updated the fwupd in the COPR with this fix.

chlee75:
@hughsie 

> I've updated the fwupd in the COPR with this fix.

I will check it and let you know in 30 min.

hughsie:
Cool, thanks. You need fwupd 1.8.1-0.855.20220526git.fc36 installed

chlee75:
@hughsie 

"Decompressing" issue has been fixed.

But HP G5 firmware shown in gnome-firmware has been changed. 
 Previous Testing : 1.0.11.0 , 1.0.13.0, 1.0.14.0, 1.0.15.0
 New Testing (fwupd 1.8.1-0.855.20220526git.fc36):  1.0.11.0

I will check whether HP moved G5 firmware from the testing state to another state to make it clear.

hughsie:
1.0.15.0 was moved to stable 2 hours ago; the metadata is probably still syncing.

chlee75:
@hughsie 
Has HP also been moving G4 dock firmware to stable?

hughsie:
I think that's in lvfs-testing still.

PR #4689, opened 2022-05-31
trivial: Do not use the unset MSR ucode version

hughsie:
This happens in VirtualBox.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/msr/fu-plugin-msr.c
+++ b/plugins/msr/fu-plugin-msr.c
@@ -203,7 +203,7 @@ fu_plugin_msr_backend_device_added(FuPlugin *plugin, FuDevice *device, GError **
 						G_LITTLE_ENDIAN,
 						error))
 			return FALSE;
-		if (ver_raw != 0) {
+		if (ver_raw != 0 && ver_raw != G_MAXUINT32) {
 			FwupdVersionFormat verfmt = fu_device_get_version_format(device_cpu);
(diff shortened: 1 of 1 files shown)


plugins/msr/fu-plugin-msr.c
```diff
@@ -203,7 +203,7 @@ fu_plugin_msr_backend_device_added(FuPlugin *plugin, FuDevice *device, GError **
 						G_LITTLE_ENDIAN,
 						error))
 			return FALSE;
-		if (ver_raw != 0) {
+		if (ver_raw != 0 && ver_raw != G_MAXUINT32) {
```
superm1, line 206: should this maybe be a fail that you return FALSE for this condition?

hughsie: Maybe. I do worry that returning with an error here is going to print something on the console -- it's not really noteworthy. The other thing I worried about is if maybe we wanted to read another offset other than PCI_MSR_IA32_BIOS_SIGN_ID in the future, we'd need to refactor this out and have to swallow the error anyway. I think just skipping the block is probably much simpler.

superm1: okay

superm1 (approved):

PR #4706, opened 2022-06-06
Add startup profiling functionality

hughsie:
With `--verbose` this will print a dump of startup times onto the
console so that we can debug which plugin is slowing down system
startup. For example:

    ../src/fu-engine.c:5409:plugins-coldplug [108.55ms]:
        ../plugins/amt/fu-plugin-amt.c:437:amt [27.14ms]:
            ../plugins/amt/fu-plugin-amt.c:437:create-context [12.66ms]
            ../plugins/amt/fu-plugin-amt.c:437:get-version [8.44ms]
            ../plugins/amt/fu-plugin-amt.c:437:add-device [3.95ms]

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-progress.c
+++ b/libfwupdplugin/fu-progress.c
@@ -11,6 +11,7 @@
 #include <math.h>
 
 #include "fu-progress.h"
+#include "fu-string.h"
 
 /**
  * FuProgress:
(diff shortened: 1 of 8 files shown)


libfwupdplugin/fu-progress.c
```diff
@@ -334,6 +330,43 @@ fu_progress_get_percentage(FuProgress *self)
 	return priv->percentage;
 }
 
+static void
+fu_progress_set_parent(FuProgress *self, FuProgress *parent)
+{
+	FuProgressPrivate *priv = GET_PRIVATE(self);
+	g_return_if_fail(FU_IS_PROGRESS(self));
```
superm1, line 337: I think you want to fail if parent is null too, no?

src/fu-tool.c
```diff
@@ -3880,6 +3880,22 @@ main(int argc, char *argv[])
 		return EXIT_FAILURE;
 	}
 
+	/* show traceback */
+	if (g_getenv("FWUPD_VERBOSE") != NULL) {
+		guint64 threshold_ms = 1;
+		g_autofree gchar *str = NULL;
+		if (!fu_common_strtoull_full(g_getenv("FWUPD_VERBOSE"),
```
superm1, line 3887: this new behavior for `FWUPD_VERBOSE` needs to be documented, and also why is it it inconsistent with the daemon?

hughsie: I wasn't sure which made more sense; I started with the env var, then moved it to a daemon config option, and now I'm thinking the env var makes more sense for both.

libfwupdplugin/fu-progress.c
hughsie: Agree, added.

superm1 (approved):

PR #4709, opened 2022-06-06
genesys: scaler: Remove FuGenesysMtkFooter and XOR encryption

gportay:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

The RSA MTK Footer is used by the vendor's tool.

The footer has to be regenerated to ship firmwares to LVFS, and it is error prone in packaging firmware pipeline; and it is much like bloat code for `fwupd`.

This PR aims to remove the footer entirely; and it is important to mention that no scaler firmwares was released up to now.


hughsie:
precommit has words, but in concept i approve.

gportay:
> precommit has words, but in concept i approve.

Thanks, I will fix it, and redo the tests as well.

gportay:
@hughsie; the only thing that is missing here is probably regenerating the [https://github.com/fwupd/fwupd/tree/main/plugins/genesys/tests].

I know nothing about fuzzing yet; I do not know which package I should install the tests on Arch Linux. Do you know which is the one I should install? or do we have a Docker image on which I can run the tests?

hughsie:
to regenerate the files first delete the checked in binary versions do `contrib/ci/oss-fuzz.py` -- that will build versions of the deps and create the binary files from the .builder.xml files

gportay:
> contrib/ci/oss-fuzz.py

Ouch, I am on Arch.

```
[gportay@archlinux fwupd]$ sudo contrib/ci/oss-fuzz.py
No provider of 'libcbor-devel' found.
No provider of 'liblzma-devel' found.
None
```

I have removed the `apt-get` thing. but nohting has changed. should I copy the blobs from another location?

hughsie:
> No provider of 'libcbor-devel' found.

Where's that coming from? You should have those libraries installed already, no?

hughsie:
> I expect the genesys scaler file to get updated

You need to delete the file, e.g. look here: https://github.com/fwupd/fwupd/blob/main/contrib/ci/oss-fuzz.py#L179

hughsie (approved):

PR #4710, opened 2022-06-06
Wip/hughsie/less common

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-byte-array.c
```diff
@@ -0,0 +1,167 @@
...
+ * fu_byte_array_set_size:
+ * @array: a #GByteArray
+ * @length: the new size of the GByteArray
+ *
+ * Sets the size of the GByteArray, expanding it with NULs if necessary.
+ *
+ * Since: 1.5.0
+ **/
+void
+fu_byte_array_set_size(GByteArray *array, guint length)
```
superm1, line 126: Why not just drop this with the abi bump recently?

hughsie: Yup, good idea.

hughsie: Done.

libfwupdplugin/README.md
```diff
@@ -4,6 +4,8 @@ This library is only partially API and ABI stable. Keeping unused, unsafe and
 deprecated functions around forever is a maintenance burden and so symbols are
 removed when branching for new minor versions.
 
+Use `./contrib/migrate.py` to migrate up out-of-tree plugins to the new API.
```
superm1, line 7: What about connecting this to pre commit?

hughsie: I did think about it, but it's a kinda dumb search-and-replace and I'm worried it's going to do bad things *tm* when run unconditionally. e.g. you still need to manually fix up the .map file for instance.

hughsie: I don't think running the script is in anyway helpful for an out-of-tree plugin; it would only be run when the source is changed (rather than treewide) and when just rebasing against master it won't do anything. Unless I'm missing the point somewhere.

hughsie:
@superm1 some feedback on the new names would be great -- I don't expect you to wade through all of that!

hughsie:
@superm1 last call on the new method names please; if you give this PR the nod I can rebase the others I have pending against it.

libfwupdplugin/fwupdplugin.map
```diff
@@ -70,7 +70,6 @@ LIBFWUPDPLUGIN_0.9.3 {
 
 LIBFWUPDPLUGIN_0.9.7 {
   global:
-    fu_common_extract_archive;
```
superm1, line 80: I guess since we just made the ABI change this is OK.

superm1 (approved):

PR #4722, opened 2022-06-08
Reorder plugins FuDevice->probe() to speed up daemon startup by 100ms

hughsie:
There's no point scanning the baseclass again if we can quickly check
if the device is valid.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):

hughsie:
Actually, maybe we don't need to call the superclass at all... if incorporate is set up correctly it's probably entirely redundant. Will experiment more tomorrow.

PR #4725, opened 2022-06-09
logitech-bulkcontroller: Only move the progressbar forwards when writing

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
+++ b/plugins/logitech-bulkcontroller/fu-logitech-bulkcontroller-device.c
@@ -853,9 +853,14 @@ fu_logitech_bulkcontroller_device_write_firmware(FuDevice *device,
 			g_usleep(G_USEC_PER_SEC);
 			continue;
 		}
-		fu_progress_set_percentage_full(fu_progress_get_child(progress),
-						self->update_progress,
-						100);
+
+		/* only update the child if the percentage is bigger -- which means the progressbar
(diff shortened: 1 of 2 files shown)


vcdmp:
Still seeing issues after applying this patch. FuProgress throws errors with both fu_progress_set_percentage_full() & fu_progress_set_percentage() APIs. Full log attached.
[console_output.txt](https://github.com/fwupd/fwupd/files/8875090/console_output.txt)

hughsie:
@vcdmp can you build and install with the extra 3 commits on this branch please, and then also add "FWUPD_LOGITECH_BULKCONTROLLER_VERBOSE=1" before the fwupdtool line please. Then can I have the entire debug log again please. Thanks!

vcdmp:
Thanks Richard. Here you go: 
[FuProgress_0610_10.txt](https://github.com/fwupd/fwupd/files/8880807/FuProgress_0610_10.txt)
.

hughsie:
Ahh! You're using fwupdtool! That explains why I couldn't reproduce using the daemon. I know where the bug is (we're re-using the FuProgress in the wrong way in fu-tool.c) and I'll fix on Monday.

vcdmp:
Thanks Richard for all your help. Have a nice weekend.

hughsie:
@vcdmp try that?

vcdmp:
No more errors.

vcdmp (approved):

PR #4726, opened 2022-06-09
scsi: Use the SCSI target to correctly set the physical ID

hughsie:
We don't want to include the SCSI LUN in the physical ID, as this is a
logical concept not relating to the hardware device.

Use the UDev devtype to find the 'correct' parent.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I've tested this on all the SCSI hardware I have access to.

daniil-lunev:
I will try to verify it by the end of the week on HW I have. (sorry for not responding earlier, we had a long weekend in AU)

PR #4727, opened 2022-06-09
Add a progress flag to indicate for when the child returns early

hughsie:
Using `fu_progress_finished(fu_progress_get_child(progress))` is IMHO
a hacky workaround.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-progress.h
+++ b/libfwupdplugin/fu-progress.h
@@ -64,6 +64,15 @@ typedef guint64 FuProgressFlags;
  */
 #define FU_PROGRESS_FLAG_NO_PROFILE (1ull << 1)
 
+/**
+ * FU_PROGRESS_FLAG_CHILD_FINISHED:
+ *
+ * The child completed all the expected steps.
+ *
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-progress.h
```diff
@@ -64,6 +64,15 @@ typedef guint64 FuProgressFlags;
  */
 #define FU_PROGRESS_FLAG_NO_PROFILE (1ull << 1)
 
+/**
+ * FU_PROGRESS_FLAG_CHILD_FINISHED:
+ *
+ * The child did not complete all the required steps, and that is expected.
```
superm1, line 70: this english reads funny.  I think it should just be "The child completed all expected steps"

superm1 (dismissed):
nit

libfwupdplugin/fu-progress.h
hughsie: Ack, thanks.

PR #4735, opened 2022-06-11
Remove gtkdoc build option

hughsie:
We've seen that gi-docgen is active and other projects are switching to
it. Remove the legacy (and slow) gtk-doc build support.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Shouldn't docs change to a boolean with this change (it can have a deprecated flag I think though).

Also - is gi docgen stable enough yet that it's in packaged distros? If not then I guess effectively a lot of distros will switch to no docs or we have to "package" the subproject. Not sure that's what we want yet.

hughsie:
It seems to be packaged in everything except Ubuntu lts and rhel 8, both of which won't do 1.8.x due to the new meson.

hughsie:
-Ddocs=true?

superm1:
> It seems to be packaged in everything except Ubuntu lts and rhel 8, both of which won't do 1.8.x due to the new meson.

We should default docs to disabled for debian packaging until it's in Debian/Ubuntu then IMO.

superm1:
> -Ddocs=true? 

Yeah

hughsie:
> We should default docs to disabled for debian packaging until it's in Debian/Ubuntu then IMO.

It seems to be in Debian, and Ubuntu latest -- just not the LTS. For the latter we just build it as a subproject instead. I guess we could use a non-LTS target in CI, but it seems a useful thing to check.

superm1:
> > We should default docs to disabled for debian packaging until it's in Debian/Ubuntu then IMO.
> 
> It seems to be in Debian, and Ubuntu latest -- just not the LTS. For the latter we just build it as a subproject instead. I guess we could use a non-LTS target in CI, but it seems a useful thing to check.

My preference would be to add as explicit build depends then. Ubuntu won't run the packaging until all deps are there. We might even be able to even get it backported to LTS (along with new meson too).

superm1:
Actually - it is in Ubuntu LTS too (Jammy is the LTS).  I think we should have it in build deps for debian/ubuntu then.
https://launchpad.net/ubuntu/+source/gi-docgen

hughsie:
Ohh it was ubuntu-20.04 that didn't have it. Is that not the LTS? Should we bump ubuntu-20.04 up to something else?

superm1:
> Ohh it was ubuntu-20.04 that didn't have it. Is that not the LTS? Should we bump ubuntu-20.04 up to something else?

LTS is every 2 years, so just picking latest LTS is fine I think.  22.04 is what we should track now.

hughsie:
@superm1 I think that's what you wanted. :)

meson_options.txt
```diff
@@ -2,7 +2,7 @@ option('build', type : 'combo', choices : ['all', 'standalone', 'library'], valu
 option('consolekit', type : 'feature', description : 'ConsoleKit support', deprecated: {'true': 'enabled', 'false': 'disabled'})
 option('static_analysis', type : 'boolean', value : false, description : 'enable GCC static analysis support')
 option('firmware-packager', type : 'boolean', value : true, description : 'enable firmware-packager installation')
-option('docs', type : 'combo', choices : ['none', 'gtkdoc', 'docgen'], value : 'docgen', description : 'developer documentation type')
+option('docs', type : 'boolean', value : true, description : 'Build developer documentation', deprecated: {'docgen': 'true', 'none': 'false'})
```
superm1, line 5: Do you want to keep it enabled by default? I would argue it's not needed for most people since we publish to a URI anyway.

Or in a related thought should it be a meson tristate feature instead? Then it wouldn't be a hard dep by default.

superm1 (commented):
Just a question on defaults/making a feature, otherwise it seems fine

meson_options.txt
hughsie: I don't think it wants to be a tristate-- it's not really an extra nice to have feature, it seems like something most people want to do; we're only syncing the docs at release time, and it might be the check-ed out copy is the most up to date docs available. That said, most people building a plugin a probably using other plugins and perhaps devhelp as well, rather than pointing firefox at the local docs. My gut instinct is it's an explicit boolean, but I could argue the default either way.

superm1: Well as another thought... We could have two copies of docs published to the web. Latest stable release and also "tip" from CI.

hughsie: Yes, sane. Not for this PR tho :)

superm1: Okay.. I say what you have is good for now and if/when we publish multiple docs copies it should flip default to off.

PR #4737, opened 2022-06-11
genesys: Do not claim interface 0

gportay:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -525,19 +525,6 @@ fu_genesys_usbhub_device_authenticate(FuGenesysUsbhubDevice *self, GError **erro
 }
 
 static gboolean
-fu_genesys_usbhub_device_probe(FuDevice *device, GError **error)
-{
-	/* FuUsbDevice->probe */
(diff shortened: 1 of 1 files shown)


hughsie:
So, just to check, we still do the FuUsbDevice:open() -- we just don't claim the interface. I assume you've tested firmware updating without claiming the interface? Thanks for looking at this!

gportay:
> So, just to check, we still do the FuUsbDevice:open() -- we just don't claim the interface. I assume you've tested firmware updating without claiming the interface? Thanks for looking at this!

I have tested firmware upgrades usbhub and scaler.

OOS: I have noticed a typo in the scaler code; the [_close()](https://github.com/fwupd/fwupd/blob/v1.8.1/plugins/genesys/fu-genesys-scaler-device.c#L1637) calls parent's `_open()` :facepalm:; fixing the typo throws an error.

hughsie (approved):

gportay:
> OOS: I have noticed a typo in the scaler code; the [_close()](https://github.com/fwupd/fwupd/blob/v1.8.1/plugins/genesys/fu-genesys-scaler-device.c#L1637) calls parent's `_open()` facepalm; fixing the typo throws an error.

Making these changes before #4738 fixes the error I saw.

PR #4742, opened 2022-06-13
plugins/intel-spi: Correct BIOS Control Device

Sean-StarLabs:
Correct the BIOS Control Device, which is should be 31.

This can be found on Intel doc numbers:
* 334819 (APL)
* 336561 (GLK)


hughsie:
I generated that file with `./generate-quirk.py ~/flashrom/chipset_enable.c` -- can you fix up the generator please.

hughsie (approved):

PR #4744, opened 2022-06-14
wip/3mdeb/qubes-fwupd-ci: Qubes OS fwupd CI

Asiderr:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [x] Documentation

Following commits add Qubes OS CI jobs. These jobs build Qubes packages as it is described in the [contrib/README.md](https://github.com/fwupd/fwupd/tree/main/contrib#readme). The jobs create binary artifacts with the Qubes packages.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,32 @@ jobs:
           echo $GITHUB_WORKSPACE
           docker run --privileged -e CI=true -t -v $GITHUB_WORKSPACE:/github/workspace docker.pkg.github.com/fwupd/fwupd/fwupd-${{matrix.os}}:latest
 
+  qubes-os:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [fedora, debian-x86_64]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Docker login
+        run: docker login docker.pkg.github.com -u $GITHUB_ACTOR -p $GITHUB_TOKEN
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+      - name: Build in container
+        env:
+          CI_NETWORK: true
+          CI: true
+        run: |
+          echo $GITHUB_WORKSPACE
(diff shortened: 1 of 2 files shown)


Asiderr:
@hughsie Yes

.github/workflows/main.yml
```diff
@@ -71,6 +71,32 @@ jobs:
...
+          echo $GITHUB_WORKSPACE
+          docker run --privileged -e CI=true -e QUBES=true -t -v $GITHUB_WORKSPACE:/github/workspace docker.pkg.github.com/fwupd/fwupd/fwupd-${{matrix.os}}:latest
+      - name: Upload fwupd binary artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: Binary package
+          path: |
+            dist/fwupd-qubes-*.rpm
+            fwupd-qubes-*.deb
+
```
superm1, line 99: A few comments:
1. Is it better to publish these to an apt and yum repository as part of CI?  Consuming a handful of deb/rpm packages can be pretty inconvenient.
2. By saving the artifacts does that mean these also end up on the release page?  If so; we need to be careful how we message them.

Asiderr: @superm1 @hughsie 

> Is it better to publish these to an apt and yum repository as part of CI? Consuming a handful of deb/rpm packages can be pretty inconvenient.

The main purpose of this CI is to create these packages, so it makes sense to move this job elsewhere. There is no need to create them at every push. IMO they should be built once per release.

> By saving the artifacts does that mean these also end up on the release page? If so; we need to be careful how we message them.

I agree. These are only intended for use in Qubes OS and it should be highlighted. IMO we should think about upstreaming these packages to apt and yum (if it's possible). It would really make easier life for Qubes OS users.

superm1: > The main purpose of this CI is to create these packages, so it makes sense to move this job elsewhere. There is no need to create them at every push. IMO they should be built once per release.

OK then we're aligned here.  I think you can change it so the job only runs on tagged events to accomplish that.
In terms of publishing elsewhere, I think if you want to host an apt and yum repo on the Qubes website it's feasible to scp them over (store a private SSH key in a github token).

> I agree. These are only intended for use in Qubes OS and it should be highlighted. IMO we should think about upstreaming these packages to apt and yum (if it's possible). It would really make easier life for Qubes OS users.

How receptive have distros been to taking other qubes specific packages in their repos?

PR #4746, opened 2022-06-14
Implementation of Verified Boot for Embedded (VBE)

sjg20:
This is not yet ready for prime time as there are a few things to resolve. I would like to get an initial review to figure out how to resolve a problem I mention at the end of the README.md as well as how to add and run more test code for the plugin.

Type of pull request:

Checklist
X  Fill out README.md with update protocol
X  Fill out README.md with any custom quirks and flags
X  Fill out README.md with the vendor ID security value
no Implement FuFirmware->write() and include at least one fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass

(cannot find src/fuzzing/firmware directory, but I did include fit_test.c and need to figure out how to run it, as well as how to write a test for my plugin)

no  CI run of the plugin for at least one target
no Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc).

Not currently run as I need to figure out how to run the fit_test code as well as how to add proper test code for the plugin.


meson.build
```diff
@@ -453,6 +453,9 @@ if build_standalone
                             required: get_option('plugin_flashrom').disable_auto_if(host_machine.system() != 'linux'))
 endif
 
+libfdt = cc.find_library('fdt', required: get_option('plugin_vbe'))
+libz = cc.find_library('z', required: get_option('plugin_vbe'))
```
superm1, line 457: For CI coverage add `zlib1g-dev` to `contrib/ci/dependencies.xml` (for Ubuntu/Debian).

And don't you want to be using `dependency('zlib', required: get_option('plugin_vbe')`?  Then you'll get all the CFLAGS and LDFLAGS associated with the pkgconfig for this library.

plugins/vbe/README.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Adding new VBE methods
+
+New VBE methods can be added in the plugin file `fu-plugin-vbe.c` in the
+`driver_list` array. See the documentation there for how to do this.
+
+## Finding out more
+
+See [fwupd site](https://github.com/fwupd/fwupd) for information about how
+to discuss or chat about fwupd.
```
superm1, line 225: not sure this one make sense.  Isn't that where they got this from? :)

plugins/vbe/README.md
```diff
@@ -0,0 +1,326 @@
...
+
+Some things you might need to install:
+
+```bash
+   sudo apt install build-essential git python3-pip pkg-config cmake \
+ libglib2.0-dev liblzma-dev python3.10-venv libgirepository1.0-dev \
+ gtk-doc-tools libgnutls28-dev libgpgme-dev gnutls-bin valac \
+ libusb-1.0-0-dev python3-jinja2 python3-toml python3-typogrify \
+ gcab gettext ninja-build libjson-glib-dev libxml2-utils libfdt-dev \
+ gir1.2-gusb-1.0 appstream-util
```
superm1, line 283: I would hope that none of this is necessary.  Shouldn't `./contrib/setup` do all this for you?  If not - some stuff is missing in `contrib/ci/dependencies.xml`.

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+	bfname = g_build_filename(priv->vbe_dir, SYSTEM_DT, NULL);
+	if (!g_file_get_contents(bfname, &buf, &len, error)) {
+		/* Check if we have a kernel device tree */
+		g_warning("Cannot find system DT '%s'", bfname);
+
+		/* Read in the system info */
+		g_free(bfname);
+		bfname = g_build_filename(KERNEL_DT, NULL);
+		if (!g_file_get_contents(bfname, &buf, &len, error)) {
+			g_warning("No kernel device tree '%s'", bfname);
```
superm1, line 209: you should be setting error here

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+			g_free(bfname);
+			return FALSE;
+		}
+	}
+	g_log(G_LOG_DOMAIN, G_LOG_LEVEL_INFO, "Processing system DT '%s'",
+	      bfname);
```
superm1, line 215: just use `g_debug`

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+	return TRUE;
+}
+
+static gboolean
+fu_plugin_vbe_startup(FuPlugin *plugin, FuProgress *progress, GError **error)
+{
+	FuPluginData *priv = fu_plugin_get_data(plugin);
+	g_autofree gchar *vbe_dir = NULL;
+	g_autofree gchar *state_dir = NULL;
+	gchar *buf, *bfname;
```
superm1, line 185: You should be using `g_autofree` for `bfname` and initialize this to NULL.

superm1 (changes requested):
A few initial comments.  I notice in a variety of places the plugin is way too noisy.  You're using `g_error`/`g_info`/`g_log/g_warning` for messaging.  You should by default not log anything, and instead use `g_debug`.  Then if the plugin is run in verbose mode it makes noise only.  I've commented out some examples of this but there are others.

Your error handling is a bit questionable.  I notice a bunch of places that you return FALSE but don't set `error`, or you don't pass in an error variable.  I've commented out some specific examples of this, but there are others.

In a few cases you're relying upon g_free for strings.  Instead, please use `g_autofree` to free the string.  We try to avoid manually freeing anything as it's error prone.

I notice a few cases that you're just using C types of data instead of glib types (like `int`/`gint`).  Please double check all these.

superm1:
### Thanks for your commit

**Clang Format**
 
**Pre-commit overview**

superm1:
For the Arch CI failure I don't think that fdt package is even available in their repos.  You'll probably want to disable the plugin by default for arch.

plugins/vbe/fit.c
```diff
@@ -0,0 +1,302 @@
...
+ *
+ * Copyright (C) 2022 Google LLC
+ * Written by Simon Glass <sjg@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include <libfdt.h>
```
hughsie, line 16: How much do you need to depend on this? Could we perhaps reimplement the fdt parsing in fwupd?

plugins/vbe/fit.h
```diff
@@ -0,0 +1,278 @@
...
+ * Library for U-Boot Flat Image Tree (FIT)
+ *
+ * Copyright (C) 2022 Google LLC
+ * Written by Simon Glass <sjg@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#ifndef __FU_PLUGIN_VBE_FIT_H
+#define __FU_PLUGIN_VBE_FIT_H
```
hughsie, line 11: Just #once please

plugins/vbe/fit.h
```diff
@@ -0,0 +1,278 @@
...
+ *
+ * this would return 2, since there are two images mentioned.
+ *
+ * @fit: FIT to check
+ * @cfg: Offset of configuration node to check
+ * @prop_name: Name of property to look up
+ * Returns: Number of images in the configuration, or -ve if the offset is
+ * invalid or the proprerty is not found
+ */
+int fit_cfg_img_count(struct fit_info *fit, int cfg, const char *prop_name);
```
hughsie, line 166: Normally the docs go in the .c for glib, not the .h

plugins/vbe/fit_test.c
```diff
@@ -0,0 +1,627 @@
...
+		fprintf(stderr, "line %d: %s: %d\n", __LINE__, #x, iret); \
+		return iret; \
+	}}
+#define CHECK(x)	{ \
+	int iret; \
+	iret = (x); if (!iret) { \
+	fprintf(stderr, "line %d: %s: %d\n", __LINE__, #x, iret); \
+	return 1; \
+	}}
+#define CHECKEQ(y, x)	{ \
```
hughsie, line 34: g_assert_cmpstr please

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+			meth = g_malloc(sizeof(struct FuVbeMethod));
+			meth->vbe_method = p + 1;
+			meth->node = node;
+			meth->driver = driver;
+			g_info("Update mechanism: %s", meth->vbe_method);
+			*methp = meth;
+			return TRUE;
+		}
+	}
+	g_error("No driver for VBE method '%s'", method_name);
```
hughsie, line 121: Seems a bit Draconian 🙂

plugins/vbe/fu-plugin-vbe.h
```diff
@@ -0,0 +1,21 @@
...
+ * VBE plugin for fwupd,mmc-simple
+ *
+ * Copyright (C) 2018 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2022 Google LLC
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#ifndef __FU_PLUGIN_VBE_H
+#define __FU_PLUGIN_VBE_H
```
hughsie, line 11: #once

plugins/vbe/fit.c
```diff
@@ -0,0 +1,302 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
```
hughsie, line 1: no ctags please, we have precommit.

plugins/vbe/README.md
```diff
@@ -0,0 +1,326 @@
...
+A typical FIT file looks like this:
+
+```fdt
+/ {
+    timestamp = <0x62a74c6c>;
+    description = "Firmware image with one or more FDT blobs";
+    creator = "U-Boot mkimage 2021.01+dfsg-3ubuntu0~20.04.4";
+    #address-cells = <0x00000001>;
+    images {
+        firmware-1 {
```
hughsie, line 32: This looks dangerously close to JSON, no?

plugins/vbe/example/com.Vbe.Board.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+  <name>VBE BoardFirmware</name>
+  <summary>System firmware for a board, for use with VBE</summary>
+  <description>
+    <p>
+      The board can be updated using Verified Boot for Embedded (VBE).
+    </p>
+  </description>
+  <provides>
+    <!-- this is a suitable HWID, found using `fwupdmgr hwids` -->
+    <firmware type="flashed">bb3b05a8-ebef-11ec-be98-d3a15278be95</firmware>
```
hughsie, line 14: I think you want to include the A/B into the instance ID, no?

plugins/vbe/fit.c
```diff
@@ -0,0 +1,302 @@
...
+{
+	int ret;
+
+	ret = fdt_check_header(buf);
+	if (ret)
+		return -FITE_BAD_HEADER;
+	fit->blob = buf;
+	fit->size = size;
+
+	return false;
```
hughsie, line 64: FALSE and TRUE pls

meson.build
sjg20: done

meson.build
```diff
@@ -453,6 +453,9 @@ if build_standalone
                             required: get_option('plugin_flashrom').disable_auto_if(host_machine.system() != 'linux'))
 endif
 
+libfdt = cc.find_library('fdt', required: get_option('plugin_vbe'))
```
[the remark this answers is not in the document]
sjg20: done

plugins/vbe/README.md
sjg20: removed

plugins/vbe/README.md
sjg20: I get this when running that script:

E: Unable to locate package gi-docgen

so I had to do various things manually

This is Ubuntu 20.04.4 LTS

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+	FuPluginData *priv = fu_plugin_get_data(plugin);
+	g_autofree gchar *vbe_dir = NULL;
+	g_autofree gchar *state_dir = NULL;
+	gchar *buf, *bfname;
+	gsize len;
+	int ret;
+
+	ret = fit_test();
+	if (ret) {
+		g_info("fit_test failed: %d", ret);
```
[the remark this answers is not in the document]
sjg20: done

plugins/vbe/fu-plugin-vbe.c
sjg20: done

plugins/vbe/fu-plugin-vbe.c
sjg20: done

sjg20:
> For the Arch CI failure I don't think that fdt package is even available in their repos. You'll probably want to disable the plugin by default for arch.

How do I do that? Is there an example?

plugins/vbe/fit.c
sjg20: No I don't think that is a good idea. This is a fairly substantial library, many years old with a huge arrange of security and other fixes and is widely used in other tools.

plugins/vbe/fit.c
```diff
@@ -0,0 +1,302 @@
...
+		return -FITE_UNKNOWN_ALGO;
+	}
+
+	return 0;
+}
+
+int fit_check_hashes(struct fit_info *fit, int img, const char *data, int size)
+{
+	int node;
+	int ret;
```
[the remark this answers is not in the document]
sjg20: I actually think I need to break this library out into a separate place, but for now I have not done this.

When I do, I need to follow this style rather than fwupd, avoid glib, etc.

Should I try to do that split-out now? It would involve cloning from the new repo and building it, I suppose.

plugins/vbe/fit.h
sjg20: See above re style

plugins/vbe/fit.h
sjg20: See above re style

plugins/vbe/fit_test.c
sjg20: See above re style (don't want to require glib)

plugins/vbe/fu-plugin-vbe.c
sjg20: I changed it to g_set_error(), but what is draconian about it? We cannot operate with an unknown method.

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+
+static gboolean
+fu_plugin_vbe_coldplug(FuPlugin *plugin, FuProgress *progress, GError **error)
+{
+	FuContext *ctx = fu_plugin_get_context(plugin);
+	FuPluginData *priv = fu_plugin_get_data(plugin);
+	struct FuVbeMethod *meth;
+	GList *entry;
+
+	/* Create a driver for each method */
```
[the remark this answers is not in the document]
sjg20: OK, fixed throughout

plugins/vbe/fu-plugin-vbe.h
sjg20: done

plugins/vbe/fu-plugin-vbe.c
sjg20: Then how do I make the messages appear?

I tried:

G_MESSAGES_DEBUG=Vbe sudo -E build/src/fwupdtool install ...

but see nothing

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,279 @@
...
+	const struct VbeDriver *driver;
+};
+
+static void
+fu_plugin_vbe_init(FuPlugin *plugin)
+{
+	FuPluginData *priv;
+	(void)fu_plugin_alloc_data(plugin, sizeof(FuPluginData));
+	priv = fu_plugin_get_data(plugin);
+	priv->vbe_dir = NULL;
```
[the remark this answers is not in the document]
sjg20: done

plugins/vbe/fit.c
sjg20: done

plugins/vbe/README.md
sjg20: Few quotes and support for binary data and fast parsing:

https://en.wikipedia.org/wiki/Devicetree#Device_Tree_formats

plugins/vbe/example/com.Vbe.Board.metainfo.xml
sjg20: What is A/B info? For now I have only implemented a simple update, not an A/B update

plugins/vbe/fit.c
sjg20: Want to use libfdt style for this library

plugins/vbe/README.md
hughsie: @superm1 is there any way to tell apt-get to ignore packages it knows nothing about? I don't really want to create an ubuntu-old-lts variant in deps.xml

plugins/vbe/fu-plugin-vbe.c
hughsie: Ohh, my comment was about `g_error()` -- that essentially stops all operation in the daemon, including all other plugins. It's effectively an assert() with a message.

plugins/vbe/README.md
superm1: Not that I can tell.  Two ideas:
1. We add a new optional field to dependencies.xml that is "minimum OS version" to install this dep?  Then the helper script can examine this extra field and match to lsb_release to decide.
2. Helper script can use python3-apt to match whether or not the package is plausible to install or not before populating the install command.

superm1: I think that's too antisocial; Chromium dev environment is still 20.04 for now.
Here's a stab at 2: https://github.com/fwupd/fwupd/pull/4748

superm1: @sjg20 that was merged.  If you rebase on latest main, you shouldn't encounter that on Ubuntu 20.04.

hughsie: (and use ./contrib/migrate.py -- then fix up any issues -- we broke internal ABI a bit)

contrib/ci/dependencies.xml
```diff
@@ -145,6 +145,21 @@
       <package variant="x86_64" />
     </distro>
   </dependency>
+  <dependency type="build" id="libfdt-dev">
+    <distro id="centos">
+      <package variant="x86_64" />
+    </distro>
+    <distro id="fedora">
+      <package variant="x86_64" />
```
superm1, line 153: Fedora's package name here is `libfdt-devel`

libfwupd/fwupd-device.c
```diff
@@ -74,22 +74,20 @@ typedef struct {
...
-	PROP_UPDATE_STATE,
-	PROP_UPDATE_MESSAGE,
-	PROP_UPDATE_ERROR,
-	PROP_UPDATE_IMAGE,
-	PROP_BATTERY_LEVEL,
-	PROP_BATTERY_THRESHOLD,
-	PROP_PROBLEMS,
-	PROP_LAST
-};
+enum { PROP_0,
```
superm1, line 77: this looks like unintended whitespace change in the PR

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,314 @@
...
+		if (!g_file_get_contents(bfname, &buf, &len, error)) {
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INTERNAL,
+				    "No kernel device tree '%s'",
+				    bfname);
+			return FALSE;
+		}
+	}
+	g_log(G_LOG_DOMAIN, G_LOG_LEVEL_INFO, "Processing system DT '%s'", bfname);
```
superm1, line 254: too noisy, downgrade this to `g_debug`

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,314 @@
...
+
+		version = fdt_getprop(priv->fdt, meth->node, "cur-version", NULL);
+		fu_device_set_version(dev, version);
+
+		version = fdt_getprop(priv->fdt, meth->node, "bootloader-version", NULL);
+		fu_device_set_version_bootloader(dev, version);
+		fu_device_add_icon(dev, "computer");
+		fu_device_add_flag(dev, FWUPD_DEVICE_FLAG_UPDATABLE);
+		fu_plugin_device_add(plugin, dev);
+		g_object_ref(dev);
```
superm1, line 300: I don't think you need to bump the refcounf here, it's a memory leak no?

plugins/vbe/fu-plugin-vbe.c
sjg20: thank you

plugins/vbe/README.md
sjg20: Yes, fixed, thanks!

Now I am getting an error with progress. Please see #4749

plugins/vbe/fit.c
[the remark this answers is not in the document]
sjg20: OK so I should keep the library in here for now? If so, should I put it in a different directory for now?

In any case, can I avoid adding glib to it? It would create an API that I don't want

plugins/vbe/fu-plugin-vbe.c
sjg20: OK, I changed that to g_set_error()

libfwupd/fwupd-client.c
```diff
@@ -99,23 +99,21 @@ enum {
...
+       PROP_TAINTED,
+       PROP_SOUP_SESSION, /* compat ABI, do not use! */
+       PROP_HOST_PRODUCT,
+       PROP_HOST_MACHINE_ID,
+       PROP_HOST_SECURITY_ID,
+       PROP_HOST_BKC,
+       PROP_INTERACTIVE,
+       PROP_ONLY_TRUSTED,
+       PROP_BATTERY_LEVEL,
+       PROP_BATTERY_THRESHOLD,
```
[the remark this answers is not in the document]
sjg20: done

libfwupd/fwupd-device.c
sjg20: done

plugins/vbe/fu-plugin-vbe.c
sjg20: done

plugins/vbe/fu-plugin-vbe.c
sjg20: I thought that when dev went out of scope it would remove the device we just created? I cannot test now due to my 'progress' issue

sjg20:
Re the whitespace change, this is caused by the commit hook

When I commit it always makes this change

sjg20:
I had to force push. I hope that didn't break anything for the review

hughsie:
> I had to force push. I hope that didn't break anything for the review

Nah, you're good. It's your branch, you can do as you please.

plugins/vbe/fu-plugin-vbe.c
hughsie: `fu_plugin_device_add()` takes a ref on the device.

plugins/vbe/fit.c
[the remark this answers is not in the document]
hughsie: > so I should keep the library in here for now

I'm not 100% comfortable shipping libraries in plugin directories; it kinda sets a bad precedent for the next vendor that wants to ship some vendor library but doesn't want to use the existing project style and all the existing GLib and fwupd helper code. @superm1 opinions welcome too.

plugins/vbe/example/com.Vbe.Board.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+          This release updates firmware to version 1.2.3 which includes support
+          for a mythical new feature.
+        </p>
+      </description>
+    </release>
+  </releases>
+
+  <!-- only newer versions of fwupd know how to write to this hardware -->
+  <requires>
+    <id compare="ge" version="1.0.1">org.freedesktop.fwupd</id>
```
hughsie, line 34: 1.8.2 at least, I'd guess

plugins/vbe/vbe-simple.c
```diff
@@ -0,0 +1,838 @@
...
+
+	G_OBJECT_CLASS(fu_vbe_simple_device_parent_class)->finalize(object);
+}
+
+static void
+fu_vbe_simple_device_class_init(FuVbeSimpleDeviceClass *klass)
+{
+	GParamSpec *pspec;
+	GObjectClass *objc = G_OBJECT_CLASS(klass);
+	FuDeviceClass *dev = FU_DEVICE_CLASS(klass);
```
hughsie, line 786: and `device_class`...

plugins/vbe/vbe-simple.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * VBE plugin for fwupd,mmc-simple
```
hughsie, line 2: I think this should be called `fu-vbe-simple-device.c` -- but also -- why the _simple_? Are there going to be other device types in the future? i.e. could this just be `FuVbeDevice` and `fu-vbe-device.c`?

plugins/vbe/fu-plugin-vbe.h
```diff
@@ -0,0 +1,19 @@
...
+#pragma once
+
+#include <fwupdplugin.h>
+
+enum { PROP_0, PROP_VBE_METHOD, PROP_VBE_FDT, PROP_VBE_NODE, PROP_LAST };
+
+typedef FuDevice *(*vbe_device_new_func)(FuContext *ctx,
+					 const gchar *vbe_method,
+					 const gchar *fdt,
+					 int node);
```
hughsie, line 19: I don't really like this, or the VbeDriver matching stuff. For other plugins, we create the correct FuDevice superclass directly -- and in this case there only seems to be one anyway, so it seems a ton of extra complexity for no real purpose.

plugins/vbe/fu-plugin-vbe.h
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (C) 2022 Google LLC
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#pragma once
+
+#include <fwupdplugin.h>
+
+enum { PROP_0, PROP_VBE_METHOD, PROP_VBE_FDT, PROP_VBE_NODE, PROP_LAST };
```
hughsie, line 14: What's this for?

plugins/vbe/meson.build
```diff
@@ -0,0 +1,36 @@
...
+   (get_option('plugin_vbe').auto() and libfdt.found())
+cargs = ['-DG_LOG_DOMAIN="Vbe"']
+
+shared_module('fu_plugin_vbe',
+  fu_hash,
+  sources : [
+    'vbe-simple.c',
+    'fu-plugin-vbe.c',
+    'fit.c',
+    'fit_test.c'
```
hughsie, line 11: I think this should be fu-vbe-test.c and actually run as a meson test -- and also using the GTest framework.

plugins/vbe/example/com.Vbe.Board.metainfo.xml
superm1: Also do you really want to be targeting an update to A or B by the HWID?  I feel like that detail is only relevant when flashing, not when targeting.

plugins/vbe/fit.c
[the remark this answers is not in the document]
superm1: If splitting out to a library that is carried in tree so this can be reused elsewhere is the direction this needs to go, can we consider making it a safer language like rust or go?

plugins/vbe/example/com.Vbe.Board.metainfo.xml
sjg20: done

plugins/vbe/vbe-simple.c
```diff
@@ -0,0 +1,838 @@
...
+		g_free(self->vbe_method);
+
+	G_OBJECT_CLASS(fu_vbe_simple_device_parent_class)->finalize(object);
+}
+
+static void
+fu_vbe_simple_device_class_init(FuVbeSimpleDeviceClass *klass)
+{
+	GParamSpec *pspec;
+	GObjectClass *objc = G_OBJECT_CLASS(klass);
```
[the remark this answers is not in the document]
sjg20: done

plugins/vbe/vbe-simple.h
sjg20: Yes there will be more. Once this is figured out the next one is a verified boot and then A/B/recovery. We may also have ones were we need to talk to some kernel service to initiate an update.

'simple' just means that it is a simple, single firmware image. It covers a large number of devices today, I believe, but these will move to a more secure setup over time.

plugins/vbe/vbe-simple.c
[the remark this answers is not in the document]
hughsie: ya, that works too.

plugins/vbe/vbe-simple.h
hughsie: Ack, fu-vbe-simple-device.c then please.

plugins/vbe/fu-plugin-vbe.h
sjg20: Well I need to start somewhere. I found that the docs in the readme are not visible, so I updated them.

Here is the main VBE idea. Every board potentially has its own way of doing updates, similar to how every USB device does. I am trying to implement common features, but I suspect there will be vendor-specific drivers too.

https://docs.google.com/document/d/e/2PACX-1vQjXLPWMIyVktaTMf8edHZYDrEvMYD_iNzIj1FgPmKF37fpglAC47Tt5cvPBC5fvTdoK-GA5Zv1wifo/pub

plugins/vbe/meson.build
sjg20: OK well we need to decide on this. If this fit stuff is going to live in fwupd and not use the external library for some years, it seems best to rewrite it.

But can I at least avoid adding glib stuff to it? That would change the API (header file) which I would very much like to keep the same between the different users.

plugins/vbe/vbe-simple.h
sjg20: done

plugins/vbe/fu-plugin-vbe.h
hughsie: Ahh, so is the idea that all vbe devices will support these properties? The way we've done this before is make another superclass, i.e. `FuSimpleVbeDevice -> FuVbeDevice -> FuDevice -> FwupdDevice -> GObject` and then all the common properties can go in `FuVbeDevice` and all the special _simple_ logic can go in the top superclass. This is also why we allow the quick file to set the device GType -- i.e. you can say in the quirk file `create a FuSimpleVbeDevice for GUID XXXYYYYY` and the right things happens.

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,316 @@
...
+}
+
+static gboolean
+process_system(FuPluginData *priv, gchar *fdt, gsize fdt_len, GError **error)
+{
+	gint ret, parent, node;
+	gint found;
+
+	ret = fdt_check_header(fdt);
+	if (ret) {
```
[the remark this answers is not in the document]
hughsie: I'd agree that GLib is halfway between "classic C" and "modern C++"...

plugins/vbe/fu-plugin-vbe.h
sjg20: Yes I expect these to be common properties. I was leaving that until we had another driver, but we don't need too. I will see if I can figure out what you have here.

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,838 @@
...
+
+	return out;
+}
+
+static void
+fu_vbe_simple_device_set_progress(FuDevice *self, FuProgress *progress)
+{
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED);
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 100, "write");
```
hughsie, line 678: I think we're expecting 4 `fu_progress_add_step()`'s here

hughsie: I'll add a test for this.

hughsie: Done in the other PR.

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,838 @@
...
+ * @cur_version: Currently installed version, NULL if none
+ * @status: State of the last update (always "completed"), NULL if none
+ */
+struct vbe_simple_state {
+	struct last_update {
+		time_t finish_time;
+		gchar *cur_version;
+		gchar *status;
+	} last;
+};
```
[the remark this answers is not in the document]
sjg20: The struct name is wrong, it should really be update_detail or something like that.

The idea is that we may record previous updates which the firmware can report.

But we can worry about this later - another option might be to read the fwupd log from firmware. I will squash this.

plugins/vbe/fu-vbe-simple-device.c
sjg20: Can you explain that a bit? In this case we only do a write.

sjg20: So are you saying we need to detach from the MMC? This is not a USB device so I don't think we can do that. It is really just a case of writing to the MMC.

As for the reload, is that to make the device reload its firmware? This is the main firmware for the device (potentially for AP, EC and even perhaps some secure bits), so the only way to reload it is to reboot.

hughsie: > So are you saying we need to detach from the MMC

sjg20: Ah OK, so there always must be these four steps. I'll update it.

sjg20: Thanks for all the info. Done

I filed #4755 too.

plugins/vbe/fu-plugin-vbe.h
sjg20: Done, but have not looked at quirks yet

plugins/vbe/example/com.Vbe.Board.metainfo.xml
sjg20: I don't think I understand what is meant by A or B.

plugins/vbe/fu-plugin-vbe.c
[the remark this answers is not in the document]
sjg20: $ git grep -w rc |wc -l
308
$ git grep -w ret |wc -l
1204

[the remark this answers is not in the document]
sjg20: Also I'd suggest that 'return code' should mean the return code from the program, e.g. normally a number from 0 to 255

'err' suggests that there is always an error, which is why I like 'ret' - the return value

plugins/vbe/fu-plugin-vbe.h
sjg20: I'm going to leave quirks for the future.

plugins/vbe/meson.build
sjg20: OK I figured it out. Done

plugins/vbe/fu-vbe-simple-device.c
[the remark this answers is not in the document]
sjg20: done

sjg20:
I handled all the comments, I think.

PTAL

hughsie:
Will do, thanks. I've got one big task this week, and will have time later this week. Sorry for the delay.

plugins/vbe/example/com.Vbe.Board.metainfo.xml
superm1: Slot A vs Slot B.  If you end up eventually splitting this into a two stage process across multiple boots (which is mentioned in the TODO section in README.md), the question was if there needs to have this information in the GUIDs.  I would argue no - the payload is the same no matter which slot you flash in, but it will come down to your (later) implementation for flashing both slots if that is true.

plugins/vbe/meson.build
```diff
@@ -0,0 +1,63 @@
...
+    'fit_test',
+    sources : [
+      'fit_test.c',
+      'fit.c',
+    ],
+    include_directories : [
+      root_incdir,
+    ],
+    dependencies : [
+      glib2,
```
superm1, line 53: Any particular reason to not use `plugin_deps` instead here?

plugins/vbe/meson.build
```diff
@@ -0,0 +1,63 @@
+if get_option('plugin_vbe').enabled() or \
+   (get_option('plugin_vbe').auto() and libfdt.found())
```
superm1, line 2: don't you also need a `libz.found()` for the second clause?

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
...
+ *
+ * Copyright (C) 2017 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2022 Google LLC
+ * Written by Simon Glass <sjg@chromium.org>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#undef G_LOG_DOMAIN
+#define G_LOG_DOMAIN "VbeSimple"
```
superm1, line 13: `FuPluginVbeSimple` perhaps

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
...
+/**
+ * fdt_get_u32() - Get a 32-bit integer value from the device tree
+ *
+ * @fdt: Device tree to read from
+ * @node: Node offset to read from
+ * @prop_name: Name of property to read
+ * Returns: integer value, if found and of the correct size, else -1
+ */
+static long
+fdt_get_u32(const gchar *fdt, gint node, const gchar *prop_name)
```
superm1, line 129: prefix with `fu_vbe_`?  Or if there is reason not to; maybe they should go into your "library" source file.

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
...
+/**
+ * fdt_get_u64() - Get a 64-bit integer value from the device tree
+ *
+ * @fdt: Device tree to read from
+ * @node: Node offset to read from
+ * @prop_name: Name of property to read
+ * Returns: integer value, if found and of the correct size, else -1
+ */
+static long
+fdt_get_u64(const gchar *fdt, gint node, const gchar *prop_name)
```
superm1, line 150: prefix with `fu_vbe_`?  Or if there is reason not to; maybe they should go into your "library" source file.

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
...
+fu_vbe_simple_device_probe(FuDevice *self, GError **error)
+{
+	struct _FuVbeSimpleDevice *dev = FU_VBE_SIMPLE_DEVICE(self);
+	FuVbeDevice *vdev;
+	const void *fdt;
+	gint devnum, len;
+	gint node;
+
+	vdev = FU_VBE_DEVICE(self);
+	g_return_val_if_fail(FU_IS_VBE_DEVICE(self), FALSE);
```
superm1, line 172: I think this needs to come before setting `vdev` in the line above

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
...
+	g_autoptr(GBytes) bytes = NULL;
+	struct fit_info fit;
+	GList *compat_list;
+	FuVbeDevice *vdev;
+	const guint8 *buf;
+	gsize size = 0;
+	gint ret;
+
+	vdev = FU_VBE_DEVICE(self);
+	g_return_val_if_fail(FU_IS_VBE_DEVICE(self), FALSE);
```
superm1, line 638: same problem I've seen elsewhere; you shouldn't set `vdev` until after the `g_return_val_if_fail`.

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
...
+
+	compat_list = fu_vbe_device_get_compat_list(vdev);
+
+	bytes = fu_firmware_get_bytes(firmware, error);
+	if (!bytes)
+		return FALSE;
+	buf = g_bytes_get_data(bytes, &size);
+
+	g_debug("Size of FIT: %#zx\n", size);
+	ret = fit_open(&fit, buf, size);
```
superm1, line 648: It's a bit odd you don't pass in the error here to `fit_open`.  I guess that's because you're trying to keep glib out of `fit.c` still?

plugins/vbe/fit_test.h
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#ifndef __FU_PLUGIN_VBE_FIT_TEST_H
+#define __FU_PLUGIN_VBE_FIT_TEST_H
+
+int
+fit_test(void);
+
+#endif /* __FU_PLUGIN_VBE_FIT_TEST_H */
```
superm1, line 16: this whole file isn't needed AFAICT.  It's just used by the test suite right?  Can't you just make `fit_test` static to that source file?

plugins/vbe/fit_test.c
```diff
@@ -0,0 +1,645 @@
...
+ */
+
+#include "config.h"
+
+#include <glib-2.0/glib.h>
+#include <libfdt.h>
+#include <stdio.h>
+
+#include "fit.h"
+#include "fit_test.h"
```
superm1, line 18: don't think you need this include (if you axe that file)

plugins/vbe/fu-vbe-device.c
```diff
@@ -0,0 +1,297 @@
...
+	fu_device_set_physical_id(FU_DEVICE(self), "vbe");
+	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_PAIR);
+	fu_device_add_icon(FU_DEVICE(self), "computer");
+}
+
+static gboolean
+fu_vbe_device_probe(FuDevice *fud, GError **error)
+{
+	FuVbeDevicePrivate *priv;
+	FuVbeDevice *self = FU_VBE_DEVICE(fud);
```
superm1, line 144: you shouldn't do this until after `g_return_val_if_fail`

superm1 (commented):
I notice fedora CI fails, did you lose the changes to `contrib/ci/dependencies.xml` in a rebase?

plugins/vbe/README.md
```diff
@@ -0,0 +1,324 @@
...
+This plugin requires access to system firmware, e.g. via a file or an eMMC
+device.
+
+## Documentation
+
+The following documents may help in understanding VBE:
+
+* [VBE](https://docs.google.com/document/d/e/2PACX-1vQjXLPWMIyVktaTMf8edHZYDrEvMYD_iNzIj1FgPmKF37fpglAC47Tt5cvPBC5fvTdoK-GA5Zv1wifo/pub)
+* [VBE Bootflows](https://docs.google.com/document/d/e/2PACX-1vQF7g5vjaoDzRDPWmbYOoOaNQsaPYDaCT0I2LZRNFIT0lD1hqANOcBmMicM-MACFQIjvE_VN9cSvCzy/pub)
+* [VBE Firmware update](https://docs.google.com/document/d/e/2PACX-1vTnlIL17vVbl6TVoTHWYMED0bme7oHHNk-g5VGxblbPiKIdGDALE1HKId8Go5f0g1eziLsv4h9bocbk/pub)
```
agraf, line 254: Neither of these 2 URLs open for me.

plugins/vbe/fit.c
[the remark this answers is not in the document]
sjg20: Well libfdt is written in C (as is fwupd) so at the least we must have a C library. We could add other versions later, of course, but I think C is a good starting point.

As to safety, we should be able to rely on tests to cover that.

plugins/vbe/example/com.Vbe.Board.metainfo.xml
sjg20: OK. That info is particular to the system itself. The update is just the update, but fwupd will need to decide which slot to use.

plugins/vbe/meson.build
sjg20: It isn't needed since this stands alone. I have added it.

plugins/vbe/meson.build
sjg20: Yes, done

plugins/vbe/meson.build
```diff
@@ -0,0 +1,63 @@
...
+cargs = ['-DG_LOG_DOMAIN="Vbe"']
```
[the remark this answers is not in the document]
sjg20: done

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,901 @@
+/*
+ *
```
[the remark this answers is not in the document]
sjg20: done

plugins/vbe/fu-vbe-simple-device.c
sjg20: done

plugins/vbe/fu-vbe-simple-device.c
sjg20: Good point, moved to fit as it will be needed there. Done

plugins/vbe/fu-vbe-simple-device.c
sjg20: Moved. done

plugins/vbe/fu-vbe-simple-device.c
sjg20: done

plugins/vbe/fu-vbe-simple-device.c
sjg20: done

plugins/vbe/fu-vbe-simple-device.c
sjg20: Yes that's right. I want the fit library to use usable by non-glib code, since it might be used in firmware, or other projects which use libfdt.

plugins/vbe/fu-plugin-vbe.c
```diff
@@ -0,0 +1,348 @@
...
+				    FWUPD_ERROR_INTERNAL,
+				    "No kernel device tree '%s'",
+				    bfname);
+			return FALSE;
+		}
+	}
+	g_debug("Processing system DT '%s'", bfname);
+	priv->fdt = buf;
+	if (!process_system(priv, buf, len, error)) {
+		g_debug("Failed: %s", (*error)->message);
```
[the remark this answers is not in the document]
sjg20: Er, yes, that was a bug. Fixed

plugins/vbe/fit_test.h
sjg20: Yes not needed now we use g_test. Done

plugins/vbe/fit_test.c
sjg20: done

plugins/vbe/fu-vbe-device.c
sjg20: done

plugins/vbe/fu-vbe-simple-device.c
sjg20: > I notice fedora CI fails, did you lose the changes to `contrib/ci/dependencies.xml` in a rebase?

Yes sadly I did. I just noticed it bringing up a new machine. Done

plugins/vbe/README.md
sjg20: I fixed the first one

The second one works OK for me. Can you try again please?

sjg20:
Something else is wrong with CI now

hughsie:
> Something else is wrong with CI now

Ya, that's not your fault, don't worry. It's tracked in https://github.com/fwupd/fwupd/pull/4763

superm1:
> > Something else is wrong with CI now
> 
> Ya, that's not your fault, don't worry. It's tracked in #4763

you can rebase on main to pickup the CI fixes.

plugins/vbe/README.md
```diff
@@ -0,0 +1,325 @@
...
+    chmod a+rwx /var/local/lib/fwupd/vbe
+    dtc /path/to/fwupd/plugins/vbe/example/test.dts -o \
+       /var/local/lib/fwupd/vbe/system.dtb
+```
+
+Some things you might need to install:
+
+```bash
+
+   pip3 install meson
```
superm1, line 282: hopefully with the fixes in main you don't need this line anymore either.

plugins/vbe/fit.c
```diff
@@ -0,0 +1,346 @@
...
+
+	return 0;
+}
+
+const char *
+fit_img_data(struct fit_info *fit, int img, int *sizep)
+{
+	const char *data;
+	int offset, size;
+	int ret;
```
superm1, line 264: I agree with the static analysis here; change the scope of this variable

plugins/vbe/fit.c
```diff
@@ -0,0 +1,346 @@
...
+			if (ret && ret != -FITE_MISSING_VALUE)
+				return ret;
+		}
+	}
+
+	return 0;
+}
+
+const char *
+fit_img_data(struct fit_info *fit, int img, int *sizep)
```
superm1, line 260: the start of this function needs a guard to make sure that `sizep` isn't NULL.

plugins/vbe/fu-plugin-vbe.c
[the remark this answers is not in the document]
superm1: I still see the spurious `g_debug` here

plugins/vbe/meson.build
superm1: You can drop the glib2 with `plugin_deps` added I'd expect.  Please confirm that though.

plugins/vbe/fit.c
sjg20: done

plugins/vbe/fit.c
sjg20: done

plugins/vbe/README.md
sjg20: ok

plugins/vbe/fit.c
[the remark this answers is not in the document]
hughsie: This isn't going to be popular, but I do think we need to be strict about the requirements here: Either the FIT stuff is useful to other projects and gets included in `libfdt` (and in an upstream release, and included in distros) or this code needs to be in GLib/GObject style. We can't have code that's simultaneously important enough to be used in multiple places, but also not important enough to be part of an existing library or new project.

I told Wacom they couldn't just copy and paste libwacom.cpp into a new plugin, told Synaptics they couldn't just put the existing crufty reference code into fwupd, that USI global couldn't just use their existing windows update code, and most recently Intel they couldn't just import metee and libigsc as source files into fwupd.

I know it seems draconian, but having a consistent codebase is the only way fwupd as a project remains maintainable. If we allow just one vendor to do it then fwupd becomes a dumping ground for legacy code -- and that's not a healthy place to be.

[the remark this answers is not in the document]
superm1: Getting code into libfdt, getting a new release, etc takes time.
So maybe the right answer is to open a PR into libfdt with `fit.c` as it exists today, and then to get this plugin merged update fit.c to glib based code.  When a new libfdt releases, there can be a dependency bump in fwupd and all the glib code torn out (or conditionally used depending on the use case).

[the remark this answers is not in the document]
hughsie: The thing is, I think that a lot of `fit.c` can go away if we use GHashTable and GPtrArray instead.

sjg20:
> > > Something else is wrong with CI now
> > 
> > 
> > Ya, that's not your fault, don't worry. It's tracked in #4763
> 
> you can rebase on main to pickup the CI fixes.

done

sjg20:
> I notice fedora CI fails, did you lose the changes to `contrib/ci/dependencies.xml` in a rebase?

Yes, fixed now

superm1:
> error: Installed (but unpackaged) file(s) found:
   /usr/lib/debug/usr/lib64/fwupd-plugins-7/libfu_plugin_vbe.so-1.8.2-0.1alpha.fc36.x86_64.debug
   /usr/lib64/fwupd-plugins-7/libfu_plugin_vbe.so

The Fedora spec file needs to have the new plugin explicitly added.

sjg20:
> > error: Installed (but unpackaged) file(s) found:
> > /usr/lib/debug/usr/lib64/fwupd-plugins-7/libfu_plugin_vbe.so-1.8.2-0.1alpha.fc36.x86_64.debug
> > /usr/lib64/fwupd-plugins-7/libfu_plugin_vbe.so
> 
> The Fedora spec file needs to have the new plugin explicitly added.

Hi, I cannot kick off these checks and don't really know what I am doing enough to get it right first time. I've made an attempt but perhaps you could take a look and save us some back and forth?

Thanks,
Simon

hughsie:
> don't really know what I am doing enough to get it right first time

Add the libfu_plugin_vbe.so line to the end of contrib/fwupd.spec.in line the other plugins.

> I cannot kick off these checks

I'll add you as a committer, and then would you mind pushing the branch to the origin remote, and opening a new PR? That way the CI will just run every time.

sjg20:
> > don't really know what I am doing enough to get it right first time
> 
> Add the libfu_plugin_vbe.so line to the end of contrib/fwupd.spec.in line the other plugins.
> 
> > I cannot kick off these checks
> 
> I'll add you as a committer, and then would you mind pushing the branch to the origin remote, and opening a new PR? That way the CI will just run every time.

OK thanks. I was able to kick it off and will do the new PR when I get back later.

superm1:
> I'll add you as a committer, and then would you mind pushing the branch to the origin remote, and opening a new PR? That way the CI will just run every time.

Actually once he has an accepted PR I'm pretty sure his account gets added to the allow list to do it every time too.

sjg20:
> > I'll add you as a committer, and then would you mind pushing the branch to the origin remote, and opening a new PR? That way the CI will just run every time.
> 
> Actually once he has an accepted PR I'm pretty sure his account gets added to the allow list to do it every time too.

Yes that worked, so shall I just leave it as a PR in my tree?

hughsie:
> so shall I just leave it as a PR in my tree

Yup; the only (biggest?) advantage is that Mario or I can just jump in and push fixups for nits on your branch directly rather than adding review comments.

sjg20:
I am going to close this in favour of one on the main tree

https://github.com/fwupd/fwupd/pull/4794

PR #4752, opened 2022-06-17
Add eMMC guids to include FW version

daniil-lunev:
Those are required if we need to re-map FW versions with quirks. Some
devices physical version number are not consecutively mapped to logical
version number.

Change-Id: Id6adf2ec6c54319b1e3d8b142eacac3e27db0ef0

Type of pull request:

- [ ] Feature: more guid identifications for emmc devices

### DIFF
--- a/plugins/emmc/fu-emmc-device.c
+++ b/plugins/emmc/fu-emmc-device.c
@@ -224,6 +224,7 @@ fu_emmc_device_probe(FuDevice *device, GError **error)
 	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "OEM", NULL);
 	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "OEM", "NAME", NULL);
 	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "NAME", "REV", NULL);
+	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "OEM", "NAME", "REV", NULL);
 
 	/* set the vendor */
 	tmp = g_udev_device_get_sysfs_attr(udev_parent, "manfid");
--- a/plugins/emmc/fu-emmc-device.c
+++ b/plugins/emmc/fu-emmc-device.c
@@ -205,6 +205,15 @@ fu_emmc_device_probe(FuDevice *device, GError **error)
 	fu_device_build_instance_id(device, NULL, "EMMC", "NAME", NULL);
 	fu_device_set_name(device, tmp);
 
+	/* firmware version */
+	tmp = g_udev_device_get_sysfs_attr(udev_parent, "fwrev");
+	if (tmp != NULL) {
+		fu_device_set_version_format(device, FWUPD_VERSION_FORMAT_NUMBER);
+		fu_device_set_version(device, tmp);
+	}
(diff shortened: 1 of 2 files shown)


plugins/emmc/fu-emmc-device.c
```diff
@@ -214,6 +223,7 @@ fu_emmc_device_probe(FuDevice *device, GError **error)
 	fu_device_add_instance_u16(device, "OEM", oemid);
 	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "OEM", NULL);
 	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "OEM", "NAME", NULL);
+	fu_device_build_instance_id(device, NULL, "EMMC", "MAN", "NAME", "REV", NULL);
```
hughsie, line 226: can we do `"EMMC", "MAN", "OEM", "NAME", "REV"` instead? In that way it's a complete superset of the line above.

plugins/emmc/fu-emmc-device.c
```diff
@@ -205,6 +205,15 @@ fu_emmc_device_probe(FuDevice *device, GError **error)
...
 	fu_device_set_name(device, tmp);
 
+	/* firmware version */
+	tmp = g_udev_device_get_sysfs_attr(udev_parent, "fwrev");
+	if (tmp != NULL) {
+		fu_device_set_version_format(device, FWUPD_VERSION_FORMAT_NUMBER);
+		fu_device_set_version(device, tmp);
+	}
+	fu_device_add_instance_strsafe(device, "REV", tmp);
+	fu_device_build_instance_id(device, NULL, "EMMC", "NAME", "REV", NULL);
```
hughsie, line 215: Works for me, but both will need adding to the README.md in the emmc plugin directory.

superm1:
> Some devices physical version number are not consecutively mapped to logical version number.

That is bonkers, but I guess this is the right way to handle it.

plugins/emmc/fu-emmc-device.c
daniil-lunev: Added as an additional. In fact, I am a bit unhappy about OEMID being a part of GUID in the first place - cuz that is not something that identifies the storage device per-ce, but not gonna remove it not to break potential dependency on it by those who added the parameter.

plugins/emmc/fu-emmc-device.c
daniil-lunev: done

plugins/emmc/fu-emmc-device.c
superm1: It's quite common to have custom firmware per OEM though. But usually this corresponds with a custom part number too..

daniil-lunev: True. Probably it is a good idea to have both then - in the case which I am concern with OEM ID should be ignored and only model + rev needs to be matched.
There is another complication (not specific to OEMID though) - if have same FW number for all part in a family, but different model tags for different sizes, currently I would need to mention each mapping for each model, is it right? should I thin add may be just REV tag as well?

hughsie (approved):

superm1 (approved):
Please squash down to one commit then LGTM.

PR #4753, opened 2022-06-17
Wip/hughsie/gresource quirks2

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@flanfly this was a ton less work to just build in the quirks into the daemon binary, rather than the plugin shared library. Comments welcome!

hughsie:
This needs https://github.com/mesonbuild/meson/issues/10367 for the last commit btw, so WIP.

superm1:
> This needs [mesonbuild/meson#10367](https://github.com/mesonbuild/meson/issues/10367) for the last commit btw, so WIP.

so bumping dep up to 0.63 meson?  can we maybe detect old vs new meson and special case it?

hughsie:
>  can we maybe detect old vs new meson and special case it

I'm still not 100% sure this is a good idea.

meson.build
```diff
@@ -566,12 +566,13 @@ if polkit.found()
...
 endif
 if build_standalone
+  plugin_quirks = []
   gcab = find_program('gcab', required: get_option('tests'))
   subdir('data')
   subdir('po')
   subdir('libfwupdplugin')
-  subdir('src')
   subdir('plugins')
+  subdir('src')
```
superm1, line 575: why this change?  I think if you're gonna move this around just make the whole list alphabetical

src/meson.build
```diff
@@ -149,12 +149,35 @@ fwupdoffline = executable(
...
 endif
 
-resources_src = gnome.compile_resources(
-  'fwupd-resources',
-  'fwupd.gresource.xml',
-  source_dir: '.',
-  c_name: 'fu'
-)
+# this requires a new enough meson version or for the fix to be backported
+if get_option('gresource_quirks')
```
superm1, line 153: Doesn't this need a guard on meson version?

meson.build
hughsie: the stuff plugins adds the gresource files into the array, then when in src:

 * in the non-gresource case, , and then they get installed into quirks.d
 * in the gresource case, they get built into the ./src/fwupd binary and ignored

contrib/generate-gresource-xml.py
```diff
@@ -0,0 +1,33 @@
...
+import sys
+import os
+
+if len(sys.argv) < 2:
+    print("not enough arguments")
+    sys.exit(1)
+
+xml = '<?xml version="1.0" encoding="UTF-8"?>\n'
+xml += "<gresources>\n"
+xml += '<gresource prefix="/org/freedesktop/fwupd">\n'
```
superm1, line 17: why not use an xml library?

src/meson.build
hughsie: or maybe an `error()`? It would be weird to specify the non-default option and then for it silently do nothing

meson_options.txt
```diff
@@ -1,6 +1,7 @@
 option('build', type : 'combo', choices : ['all', 'standalone', 'library'], value : 'all', description : 'build type')
 option('consolekit', type : 'feature', description : 'ConsoleKit support', deprecated: {'true': 'enabled', 'false': 'disabled'})
 option('static_analysis', type : 'boolean', value : false, description : 'enable GCC static analysis support')
+option('gresource_quirks', type : 'boolean', value : false, description : 'compile the quirk files into the fwupd binary')
```
superm1, line 4: I think this could be a good candidate for a feature.  Given it is only to be enabled on newer meson it can auto-disable on older etc.

meson.build
hughsie: meson always parses meson.build files in a predictable linear order. It only parallelizes when all the targets are known.

superm1: Then I think we need to leave a comment here that this order is important

contrib/generate-gresource-xml.py
hughsie: Done.

meson.build
hughsie: done.

meson_options.txt
hughsie: Done.

plugins/upower/fu-plugin-upower.c
```diff
@@ -27,6 +27,7 @@ typedef enum {
 static void
 fu_plugin_upower_init(FuPlugin *plugin)
 {
+	g_warning("HERE HERE HERE");
```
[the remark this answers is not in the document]
hughsie: Indeed, thanks.

src/meson.build
```diff
@@ -149,12 +149,36 @@ fwupdoffline = executable(
...
 
-resources_src = gnome.compile_resources(
-  'fwupd-resources',
-  'fwupd.gresource.xml',
-  source_dir: '.',
-  c_name: 'fu'
-)
+# this requires a new enough meson version
+if get_option('gresource_quirks').require(meson.version().version_compare('>=0.63.0'),
+    error_message: 'meson >= 0.63.0 is needed for -Dgresource_quirks=true').allowed()
```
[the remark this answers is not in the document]
hughsie: Fixed

src/meson.build
```diff
@@ -149,12 +149,36 @@ fwupdoffline = executable(
...
-  'fwupd.gresource.xml',
-  source_dir: '.',
-  c_name: 'fu'
-)
+# this requires a new enough meson version
+if get_option('gresource_quirks').require(meson.version().version_compare('>=0.63.0'),
+    error_message: 'meson >= 0.63.0 is needed for -Dgresource_quirks=enabled').allowed()
+  fwupd_gresource_xml = custom_target('fwupd-resources-xml',
+    input : [
+      'org.freedesktop.fwupd.xml',
```
superm1, line 157: why does this need to be compiled in?  It's installed either way isn't it?

superm1 (dismissed):
I think that `contrib/generate-gresource-xml.py` shouldn't be in `contrib` if it's part of the build process.  It should probably move into `src`.

Otherwise LGTM.

superm1:
Arch will require an explicit disable of the feature IIRC

src/meson.build
hughsie: ya, this was trying to make it work for older meson versions -- it didn't work properly and we now have the meson check so i ripped it out.

hughsie:
@flanfly was this what you had in mind?

flanfly:
@hughsie thanks for taking care of this.

PR #4756, opened 2022-06-20
amt: Modernize the plugin and split out common functionality

hughsie:
The MEI code is soon going to be used by a new plugin, so split it
out into libfwupdplugin like we did i2c.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-mei-device.c
```diff
@@ -0,0 +1,315 @@
...
+ *
+ * Since: 1.8.2
+ **/
+gboolean
+fu_mei_device_read(FuMeiDevice *self,
+		   guint8 *buf,
+		   gsize bufsz,
+		   gsize *bytes_read,
+		   guint timeout_ms,
+		   GError **error)
```
superm1, line 195: If this is coming into plugin ABI, do we need some sort of mutex to ensure two plugins haven't stepped on each other's data in multiple back to back reads?

libfwupdplugin/fu-mei-device.c
```diff
@@ -0,0 +1,315 @@
...
+ * Returns: %TRUE for success
+ *
+ * Since: 1.8.2
+ **/
+gboolean
+fu_mei_device_write(FuMeiDevice *self,
+		    const guint8 *buf,
+		    gsize bufsz,
+		    guint timeout_ms,
+		    GError **error)
```
superm1, line 237: same question as read

libfwupdplugin/fu-mei-device.c
```diff
@@ -0,0 +1,315 @@
...
+
+#include "fu-mei-device.h"
+#include "fu-string.h"
+
+#define FU_MEI_DEVICE_IOCTL_TIMEOUT 5000 /* ms */
+
+/**
+ * FuMeiDevice
+ *
+ * A MEI device.
```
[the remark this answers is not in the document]
hughsie: >  Maybe links to what kind of data can be passed in and out, expected return codes etc

Ack; I'm wading through Intel-provided LGPLv2+ code with all this kind of stuff; I think it's good for a new PR tho.

libfwupdplugin/fu-mei-device.c
hughsie: The way this will work in practice is that the AMT device is a superclass of FuMeiDevice -- and the possible plugins will create that specific instance, we run probe, open, setup, close, then move onto the next device. Each instance is namespaced by the connection GUID which seems to provide some kind of interface API (hence the GUID on the ioctl _connect_). Of course, two things can't use the same /dev/mei0 interface, but I don't think it's allowed to happen. If we started threading the plugin coldplug (which I don't think we should do) then we'd need some kind of access control, but in this case I'm tempted to KISS and then if if we do need two things writing to the same device at the same time create a `fu-mei-backend.c` which handles the device open/closes and does the locking there.

hughsie:
>  Maybe a direct read and write don't make sense

Unfortunately, the API for MEI seems to be much like ioctl, in that you write a blob of memory, and you read another blob of memory; i.e. what the blobs actually are is defined by the GUID protocol.

libfwupdplugin/fu-mei-device.c
superm1: Yeah I'm aware we're not planning to thread coldplug, just being cognizant that if we want to some day that we need to watch out for design problems.  Sounds like a good approach.

hughsie: I'll merge this, and then propose the new hardware support next week so we can see if there are any changes needed in the baseclass before the next tarball. Thanks for all the review.

PR #4758, opened 2022-06-20
trivial: Make meson.build indentation consistent

gportay:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

The indentation used for the parameter is not consistent across the `meson.build` files; even in the same files. I picked the most commonly used (aka space, colon, space)  even though it is not the one I prefer.

I can change it if needed.

### DIFF
--- a/contrib/PKGBUILD
+++ b/contrib/PKGBUILD
@@ -17,7 +17,7 @@ pkgver() {
     cd ${pkgname}
 
     VERSION=$(git describe | sed 's/-/.r/;s/-/./')
-    [ -z $VERSION ] && VERSION=$(head meson.build | grep ' version :' | cut -d \' -f2)
+    [ -z $VERSION ] && VERSION=$(head meson.build | grep ' version:' | cut -d \' -f2)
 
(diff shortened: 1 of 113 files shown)


superm1:
Thanks for taking a stab at improving this.  It's bugged me too, but we haven't had a nice reformatter for it, so stuff like this slips by all the time.

superm1 (dismissed):
I don't have a strong opinion of which way is better for spacing, but I like the consistency, so the flip of a coin or whichever was more common sounds good to me.

hughsie:
I believe meson upstream does `: ` i.e. no leading space.

hughsie (approved):

gportay:
@hughsie what was wrong with the ci?

I have check, and nothing jumped to my eyes :/

However, I want to make a single small change, with the ternary operator, spacing before and after in the meson.build of the bash completion directory.

hughsie:
@superm1 debian ideas please?

hughsie (approved):

PR #4763, opened 2022-06-22
Fix the CI failures

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build-ubuntu-x86_64:
     machine:
-      image: ubuntu-2204:edge
+      image: ubuntu-2004:edge
     steps:
       - checkout
       - run:
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     permissions:
       actions: read
       contents: read
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   pre-commit:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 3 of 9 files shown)


hughsie (approved):

superm1:
it seems the real problem is something with the repos, this bandaid doesn't seem to help. :/
But we also can't re-generate containers (they fail every night).

hughsie:
Concur; `libsystemd-dev : Depends: libsystemd0 (= 249.11-0ubuntu3) but 249.11-0ubuntu3.1 is to be installed` is probably something that we can't fix, right? i.e. we just wait it out and click retry every day until it's fixed? :/

superm1:
> Concur; `libsystemd-dev : Depends: libsystemd0 (= 249.11-0ubuntu3) but 249.11-0ubuntu3.1 is to be installed` is probably something that we can't fix, right? i.e. we just wait it out and click retry every day until it's fixed? :/

I don't see that on any of my other systems that take Ubuntu, I'm not sure why it's only happening on the CI ones.  I'm worried that CI needs a container regeneration, but containers don't want to publish for some reason still.

hughsie:
They seem to be failing with:

    error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount2454717565/Dockerfile: no such file or directory
Error: buildx failed with: error: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to read dockerfile: open /var/lib/docker/tmp/buildkit-mount2454717565/Dockerfile: no such file or directory

hughsie:
I think we're hitting https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1979579 -- right?

superm1:
Yeah. I've got a hunch it's caused by update phasing. I wonder if there's a way to turn it off.

superm1:
@hughsie for now let's go back to using Ubuntu 20.04 for CI.  I made various cleanups to make those jobs work better, and also make `contrib/setup` more friendly for them.  We can try to go back to 22.04 by reverting [c44b9dc](https://github.com/fwupd/fwupd/pull/4763/commits/c44b9dcc056ddb389b043cb60b8c3994ef495040) sometime in the future whenever the mirror/phasing/whatever issue is sorted.

hughsie (approved):

PR #4768, opened 2022-06-23
Fix running fwupdtool security with a plugin allowlist

superm1:
A number of plugins make assumptions that ->init() was called, but
when an allowlist is used this won't have been called.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Do you think we should call ->init in all cases? If not, then I think we want to skip calling those vfuncs in the daemon, i.e. one central place.

hughsie:
Alternate PR in https://github.com/fwupd/fwupd/pull/4771 -- which just makes the `data` and `priv` stuff all the same first and uses `== NULL`.

PR #4770, opened 2022-06-23
Add support for reading and writing CPD blobs

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-cpd-firmware.c
```diff
@@ -0,0 +1,426 @@
...
+#include "fu-byte-array.h"
+#include "fu-bytes.h"
+#include "fu-cpd-firmware.h"
+#include "fu-mem.h"
+#include "fu-string.h"
+
+/**
+ * FuCpdFirmware:
+ *
+ * A CPD often found in PCI OptionROM firmware images.
```
superm1, line 23: What is a CPD?

libfwupdplugin/fu-cpd-firmware.c
```diff
@@ -0,0 +1,426 @@
...
+	guint32 header_version;
+	guint32 flags;
+	guint32 vendor;
+	guint32 date;
+	guint32 size;
+	guint32 id;
+	guint32 rsvd;
+	guint64 version;
+	guint32 svn;
+	guint64 rsvd1;
```
superm1, line 68: if this and the next one are both reserved, why have separate variables?

hughsie: Agreed; we need neither, both removed.

hughsie:
I'm also wondering if this should be `FuIntelCpdFirmware`, `FuIntelIfwiCpdFirmware` or `FuIpuCpdFirmware`

hughsie:
> I'm also wondering if this should be FuIntelCpdFirmware, FuIntelIfwiCpdFirmware or FuIpuCpdFirmware

I went with `FuIfwiCpdFirmware` in the end...

src/fu-engine.c
```diff
@@ -7226,6 +7226,8 @@ fu_engine_load(FuEngine *self, FuEngineLoadFlags flags, FuProgress *progress, GE
 	fu_context_add_firmware_gtype(self->ctx, "raw", FU_TYPE_FIRMWARE);
 	fu_context_add_firmware_gtype(self->ctx, "dfu", FU_TYPE_DFU_FIRMWARE);
 	fu_context_add_firmware_gtype(self->ctx, "dfuse", FU_TYPE_DFUSE_FIRMWARE);
+	fu_context_add_firmware_gtype(self->ctx, "ifwi-cpd", FU_TYPE_IFWI_CPD_FIRMWARE);
+	fu_context_add_firmware_gtype(self->ctx, "oprom", FU_TYPE_OPROM_FIRMWARE);
```
superm1, line 7230: oprom is optionrom right?  Maybe just take the full name

libfwupdplugin/fu-self-test.c
```diff
@@ -2308,6 +2308,77 @@ fu_firmware_dfu_func(void)
...
+	g_assert_nonnull(data_bin);
+	g_assert_cmpint(g_bytes_get_size(data_bin), ==, 512);
+
+#if 0
+	img1 = fu_firmware_get_image_by_id(firmware, "cpd", &error);
+	g_assert_no_error(error);
+	g_assert_nonnull(img1);
+	g_assert_cmpint(fu_firmware_get_offset(img1), ==, 68);
+	g_assert_cmpint(fu_firmware_get_size(img1), ==, 11);
+#endif
```
superm1, line 2379: why this here still

hughsie: Heh, I ran out of time and this bug still exists :) Working on it now the kids are asleep.

libfwupdplugin/fu-oprom-firmware.c
```diff
@@ -0,0 +1,375 @@
...
+
+#include "fu-byte-array.h"
+#include "fu-bytes.h"
+#include "fu-ifwi-cpd-firmware.h"
+#include "fu-mem.h"
+#include "fu-oprom-firmware.h"
+#include "fu-string.h"
+
+/**
+ * FuOpromFirmware:
```
superm1, line 22: I'm wondering if this should be PciOptionRomFirmware

hughsie: That's a bit wordy, but I wasn't sure so did some rough research:

optrom: 50% of qemu
oprom: 50% of EDK/uefi,coreboot
optionrom: SeaBIOS, 50% EDK, 50% optionrom, most of Linux kernel

So maybe `PciOptionRomFirmware` is best? It's going to get wordy if we split out the sections even further, e.g. `PciEfiOptionRomFirmware`. I guess my vote is Oprom just because it's shorter and what the Intel documentation seems to use but it's not that important to me either way.

libfwupdplugin/fu-self-test.c
hughsie: Fixed, uncommented and values fixed to reflect reality.

hughsie:
@superm1 I've also added a super-handy firmware parser that knows how to unwrap any of the image formats that store the size inside, I'm not convinced about the name tho; `FuLinearFirmware` seems a bit clunky, but it's the terminology that RAID uses. Another option is `FuConcatFirmware` but other ideas welcome!

libfwupdplugin/fu-oprom-firmware.c
superm1: If it makes cross referencing documentation and other code easier, keep it shorter.

superm1:
> @superm1 I've also added a super-handy firmware parser that knows how to unwrap any of the image formats that store the size inside, I'm not convinced about the name tho; `FuLinearFirmware` seems a bit clunky, but it's the terminology that RAID uses. Another option is `FuConcatFirmware` but other ideas welcome!

sounds generally useful as this isn't the first time you'll find firmware within firmware within firmware!

hughsie:
> sounds generally useful

Agreed. Not sure about the Linear name tho.

PR #4776, opened 2022-06-27
dell-dock: cache hub device and add it until ec is added

CragW:
The usb3 hub 413c:b06f may be added to the plugin ahead of ec device
while dock type is yet known at that time. Cache the hub until ec can
tell the instance id, then add it to the plugin.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/dell-dock/fu-plugin-dell-dock.c
+++ b/plugins/dell-dock/fu-plugin-dell-dock.c
@@ -79,7 +79,7 @@ fu_plugin_dell_dock_probe(FuPlugin *plugin, FuDevice *proxy, GError **error)
 
 	/* create mst endpoint */
 	mst_device = fu_dell_dock_mst_new(ctx);
-	if (fu_dell_dock_get_ec_type(FU_DEVICE(ec_device)) == ATOMIC_BASE)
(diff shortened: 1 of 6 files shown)


plugins/dell-dock/fu-plugin-dell-dock.c
```diff
@@ -177,27 +179,32 @@ fu_plugin_dell_dock_backend_device_added(FuPlugin *plugin, FuDevice *device, GEr
...
-		guint8 ec_type = fu_dell_dock_get_ec_type(ec_device);
-		fu_dell_dock_hub_add_instance(FU_DEVICE(hub), ec_type);
+	if (ec_device == NULL) {
+		fu_plugin_cache_add(plugin, hub_cache_key, FU_DEVICE(hub));
+		return TRUE;
 	}
+
+	/* determine dock type by ec */
+	ec_type = fu_dell_dock_get_ec_type(ec_device);
+	fu_dell_dock_hub_add_instance(FU_DEVICE(hub), ec_type);
```
superm1, line 198: Should you guard to make sure that you can get a valid base type (`ec_type != 0`)?

superm1: It should be an integer I think not a pointer

superm1 (approved):

Issue #4783, opened 2022-06-29
Command get-details foo.cab errors out immediately if one components fails


CragW:
**Describe the bug**
Running command `get-details` to look-up what `components` are carried by foo.cab, results error out immediately if any of the release fails a requirement check or the physical device given by `guid` is not found. In the past, all the releases are listed and components are viewable by the same command. 

This issue happens especially for an unified cabinet file that consists a variety of devices that may or may not presents, for instance the dell dock's firmware cab file.
```
$ sudo ./dist/fwupdtool.sh get-details ~/mywork/salomon-firmware/DellDockFirmwareUpdateLinux_01.00.24.cab 
Loading…                 [***********************                ]
Loading…                 [-                                      ]
WARNING: This package has not been validated, it may not work properly.
Decompressing…           [ -                                     ]
Device Package level of Dell dock [(null)] has no firmware version
```

**Steps to Reproduce**
```
# fwupdmgr get-details foo.cab
# fwupdtool get-details foo.cab
```

**Expected behavior**
All components are listed by command `get-details` to understand what the cab file offers.

**fwupd version information**
```
$ ./dist/fwupdmgr.sh --version
Decompressing…           [-                                      ]
runtime   org.freedesktop.fwupd         1.8.2
runtime   org.freedesktop.fwupd-efi     1.0
compile   org.freedesktop.gusb          0.3.10
runtime   com.dell.libsmbios            2.4
runtime   org.kernel                    5.15.0-39-generic
compile   com.hughsie.libjcat           0.1.9
compile   org.freedesktop.fwupd         1.8.2
runtime   org.freedesktop.gusb          0.3.10
```

**Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):**
source upstream ac55012

**fwupd device information**
```
$ ./dist/fwupdmgr.sh get-devices
Decompressing…           [-                                      ]WARNING: This package has not been validated, it may not work properly.
XPS 13 9310
│
├─WD22TB4:
│ │   Device ID:          587f12d7d5d5e7a078a4d5124c4b3c2017502f3e
│ │   Summary:            High performance dock
│ │   Current version:    01.01.00.01
│ │   Vendor:             Dell Inc. (USB:0x413C)
│ │   Install Duration:   1 minute
│ │   Serial Number:      6R2X0L3/3107576664195875
[...]

PR #4784, opened 2022-06-29
trivial: finish all components from a cab and show errors

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-release.c
+++ b/src/fu-release.c
@@ -942,6 +942,7 @@ fu_release_load(FuRelease *self,
 
 	/* check requirements for device */
 	if (self->device != NULL && self->request != NULL &&
+	    (install_flags & FWUPD_INSTALL_FLAG_NO_REQUIREMENTS) == 0 &&
 	    fu_engine_request_get_kind(self->request) == FU_ENGINE_REQUEST_KIND_ACTIVE) {
 		if (!fu_release_check_requirements(self, component, rel, install_flags, error))
 			return FALSE;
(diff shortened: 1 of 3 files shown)


src/fu-release.c
```diff
@@ -940,6 +940,9 @@ fu_release_load(FuRelease *self,
 					   self);
 	}
 
+	/* don't exercise requirement checks for get-details */
+	return TRUE;
+
```
superm1, line 945: don't you need some sort of check here?

CragW:
How to understand it's called by `get-details` and don't have to run requirement checks? Or, a better way to fix #4783?

src/fu-release.c
CragW: Yes it should, though the load-releases is share code between install tasks and get-details / get-updates it seems harder to make a check here. Ideas is appreciated.

hughsie:
Can you add a self test that fails? i.e. with the exact but least complicated scenario?

src/fu-release.c
hughsie: I think the idea was we'd do something like:

	request = fu_engine_request_new(FU_ENGINE_REQUEST_KIND_ONLY_SUPPORTED);
	fu_engine_request_set_feature_flags(request, ~0);

but that's not happening for get-details. Will ponder.

src/fu-release.c
```diff
@@ -942,6 +942,7 @@ fu_release_load(FuRelease *self,
 
 	/* check requirements for device */
 	if (self->device != NULL && self->request != NULL &&
+	    (install_flags & FWUPD_INSTALL_FLAG_NONE) != 0 &&
```
superm1, line 945: I think I would feel better about this to have a new install flag for ignore requirements explicitly set by the get_details flow.  There may be multiple callers into this function that might want to evaluate requirements even if INSTALL_FLAG_NONE wasn't set.

CragW: I've added `FWUPD_INSTALL_FLAG_NO_REQUIREMENTS` install flag and introduced it from 1.8.2 onward.

PR #4786, opened 2022-06-30
Wip/hughsie/get details

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


CragW:
It works, I can see the reasons if they fail requirements, way more than only displaying the releases. It seems we don't need the 2nd time of `load_release`? And thanks for the `self test`, I was just started looking into this area.

PR #4794, opened 2022-07-01
Sjg vbe

sjg20:
This is a copy of https://github.com/fwupd/fwupd/pull/4746

New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))

Checklist
X Fill out README.md with update protocol
X Fill out README.md with any custom quirks and flags
X Fill out README.md with the vendor ID security value
no Implement FuFirmware->write() and include at least one fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass

(cannot find src/fuzzing/firmware directory, but I did include fit_test.c and need to figure out how to run it, as well as how to write a test for my plugin)

X CI run of the plugin for at least one target
X Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc).


hughsie:
Mind if we push some fixups onto your branch?

sjg20:
> Mind if we push some fixups onto your branch?

Please go ahead!

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -799,12 +794,12 @@ fu_vbe_simple_device_class_init(FuVbeSimpleDeviceClass *klass)
 	/**
 	 * FuVbeSimpleDevice:devname:
 	 *
-	 * device that contains firmware (e.g. '/dev/mmcblk1')
+	 * device that contains firmware (e.g. '/self/mmcblk1')
```
sjg20, line 797: this should stay as is

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -799,12 +794,12 @@ fu_vbe_simple_device_class_init(FuVbeSimpleDeviceClass *klass)
...
 	 */
 	pspec =
 	    g_param_spec_string("devname",
 				NULL,
-				"Device that contains firmware (e.g. '/dev/mmcblk1')",
+				"Device that contains firmware (e.g. '/self/mmcblk1')",
```
sjg20, line 802: and this

hughsie: Oopa, sorry, fixed.

hughsie:
I'd like to refactor some of the FIT code; I'm on it, but give me a few more days.

plugins/vbe/README.md
```diff
@@ -93,7 +89,7 @@ VBE requires the board firmware to provide information about the firmware within
 the device tree passed to the Operating System.
 
 For systems without device tree, currently the only option is to install a file
-for use by fwupd, typically in `/var/local/lib/fwupd/vbe/system.dtb`.
+for use by fwupd, typically in `/var/lib/fwupd/vbe/system.dtb`.
```
sjg20, line 92: These changes seem to make it necessary to install fwupd / the plugin before it can be tested. Am I missing something?

plugins/vbe/README.md
```diff
@@ -294,16 +273,6 @@ To build the example:
...
 ```
 
-To install the cab on your development computer:
-
-```bash
-   # Set up the test file
-   dd if=/dev/zero of=/tmp/testfw bs=1M count=3
-
-   sudo build/src/fwupdtool install plugins/vbe/example/Vbe-Board-1.2.4.cab \
-      bb3b05a8-ebef-11ec-be98-d3a15278be95
```
sjg20: Why drop this? It is hard to know how to test it otherwise.

Having said that, I think I need to create a proper test at some point.

plugins/vbe/README.md
hughsie: In general we don't advise users to install fwupd system-wide and instead encourage a prefix; in this case I thought it would be more obvious to show the distro-installed location rather than using something like `${PREFIX}/var/lib/fwupd/vbe/system.dtb` which might be even more confusing.

plugins/vbe/README.md
hughsie: In general, only plugin-specific stuff goes in each plugin README.d, and the general stuff can go in the toplevel; in this case it's no different to any of the other plugins.

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,844 @@
...
+ * This simply records successful updates at present. There is no log of
+ * failures.
+ *
+ * @finish_time: Time that the the last update happened, 0 if none
+ * @cur_version: Currently installed version, NULL if none
+ * @status: State of the last update (always "completed"), NULL if none
+ */
+struct vbe_simple_state {
+	/* information about the last update */
+	time_t finish_time;
```
hughsie, line 40: @sjg20 I'm a bit confused why we're setting `finish_time`; is it going to be used in the future?

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,844 @@
...
+
+	g_debug("Best configuration: '%s', priority %d, version %s", cfg_name, best_prio, version);
+
+	if (!process_config(self, fit, best_cfg, progress, error))
+		return FALSE;
+
+	g_free(state->cur_version);
+	g_free(state->status);
+	state->finish_time = time(NULL);
+	state->cur_version = g_strdup(version);
```
hughsie, line 569: @sjg20 I'm a bit confused why we're saving the version to a separate file here; can't it be read out from `cur-version` during the next-reboot coldplug?

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,844 @@
...
+			    "Cannot seek file '%s' (%d) to %#jx (%s)",
+			    self->devname,
+			    self->fd,
+			    (uintmax_t)seek_to,
+			    strerror(errno));
+		return FALSE;
+	}
+
+	ret = write(self->fd, buf + self->skip_offset, size - self->skip_offset);
+	if (ret < 0) {
```
hughsie, line 434: @sjg20 does this take a significant amount of time? Could we write it in lots of chunks so that we can advance the FuProgress bar in a smooth way or does the write have to be done all at once?

plugins/vbe/fu-vbe-simple-device.c
```diff
@@ -0,0 +1,844 @@
...
+	switch (prop_id) {
+	case PROP_DEVNAME:
+		g_value_set_string(value, self->devname);
+		break;
+	case PROP_AREA_START:
+		g_value_set_int64(value, self->area_start);
+		break;
+	case PROP_AREA_SIZE:
+		g_value_set_int64(value, self->area_size);
+		break;
```
hughsie, line 742: @sjg20 I'm a bit confused why these are GObject properties; you're not planning to subclass FuVbeSimpleDevice are you? If so, the device needs to be derivable, not final.

plugins/vbe/README.md
sjg20: What about the /tmp/testfw thing? I don't think others do something like that.

Anyway, this info is useful for me when I come back to it in a few months. Is there another place it could go?

plugins/vbe/README.md
sjg20: Fair enough, but this info is for development, not users. Perhaps we should have a separate file?

plugins/vbe/fu-vbe-simple-device.c
sjg20: No I am not planning to subclass it. I added these so that I can call the constructor with their values. Also I hope that these values might be visible to the fwupd client?

plugins/vbe/fu-vbe-simple-device.c
hughsie: > Using the filesystem is not very helpful

Yup, I really think we don't want to do the /var/lib/fwupd/vbe/simple.dtb thing -- it's storing the data in the wrong "layer". I like your `nvdata` proposal -- i.e. the version has to be read from somewhere in the flash image, not from a partition.

sjg20: Well nvdata is for firmware since it cannot always access simple.dtb - having said that, fwupd will eventually have to be able to update nvdata.

My thinking is that for user space / fwupd we need richer information about what needs doing and what has already been tried, e.g. to handle rollbacks and limits on how far to roll forward. We should be able to signal what needs doing from firmware, but fwupd needs to actually do it. Quite apart from that we need debugging and logging which is tricky to do across reboots, so recording recent updates and why VBE did what it did is helpful.

Let's say (in an A/B situation) we select an update and boot it, only to find that it does not boot into user space. We reboot and try the old firmware, which works. We now need to go back to the old firmware in both slots (A and B).

The problem with vbe-simple is that it oes not have to deal with these cases. On the other hand if we don't have the feature included, we are not defining how to do it with more complex VBE methods.

Still. I don't have a clear idea of exactly what is needed in this file and vbe-simple does not need it, so I am OK with dropping it for now.

plugins/vbe/fu-vbe-simple-device.c
sjg20: OK, so what should we do here?

hughsie: Do you *need* the area-start values in the client? If so, why?

plugins/vbe/fu-vbe-simple-device.c
sjg20: It allows the firmware to show when the last update was done. We can drop it for now if you like

plugins/vbe/fu-vbe-simple-device.c
sjg20: It allows people to see the status of things, in this case where the firmware is located. With the A/B/recovery version, when I do it, we'll want to show the current state of things in more detail, e.g. what versions are in slots A and B. I suppose that could be done by reading the settings .dtb file directly, but it seems better to use fwupd if possible.

hughsie: The way we've done this so far is to expose each of the A/B partitions as separate devices, with the "other" partition being marked as non-updatable and as a child to the "active" partition. In the Redfish server case it seems to work really well.

sjg20: OK that sounds good. Once there is a successful boot we need to copy the firmware from B to A and roll forward the TPM.

sjg20:
> Alternate proposal in #4811

That seems like something quite different. For example it doesn't seem to use FIT for the firmware format?

hughsie:
> it doesn't seem to use FIT for the firmware format

It sure does :) https://github.com/fwupd/fwupd/pull/4811/files#diff-42acb997b991a452069343af1581bd6f6b6be7ac4cf28681ca5c0e0da4511c90R171 is for loading the per-system fdt and https://github.com/fwupd/fwupd/pull/4811/files#diff-42acb997b991a452069343af1581bd6f6b6be7ac4cf28681ca5c0e0da4511c90R171 is for loading the FIT file we used as the update payload. I've reworked the FIT loading into the same kind of FuFirmware parsing and writing we do for all the other formats; both FIT and FDT are really well documented and much simpler than some of the other formats we already support in-tree.

sjg20:
> > it doesn't seem to use FIT for the firmware format
> 
> It sure does :) https://github.com/fwupd/fwupd/pull/4811/files#diff-42acb997b991a452069343af1581bd6f6b6be7ac4cf28681ca5c0e0da4511c90R171 is for loading the per-system fdt and https://github.com/fwupd/fwupd/pull/4811/files#diff-daf7cb050a1a4d074a1570922aaae4a1c20084a556ca7cbfd4f31fa3eded36cdR228 is for loading the FIT file we used as the update payload. I've reworked the FIT loading into the same kind of FuFirmware parsing and writing we do for all the other formats; both FIT and FDT are really well documented and much simpler than some of the other formats we already support in-tree.

OK well github strikes again. Only the first commit seems to be visible. I downloaded the PR and took a look.

I can see what you are doing, and having a more general-purpose (within fwupd) way of encoding and decoding firmware seems useful. But I'm not a big fan of rewriting libfdt, given all the sanity checks and tests that it has!

Minor point - 'Flattened Devicetree aka FDT firmware' - it isn't really a firmware format. It is only FIT which is that.

Anyway, if we go with this, how will we deal with configuration signing? That is a key feature but it does rely on being able to access the FIT at a binary level.

https://research.google/pubs/pub42038/
https://www.youtube.com/watch?v=kdpZC9jFzZA&t=26m
http://www.pacificsimplicity.ca/blog/verified-boot-%E2%80%93-introduction-u-boot%E2%80%99s-secure-boot/

hughsie:
> But I'm not a big fan of rewriting libfdt, given all the sanity checks and tests that it has!

Yup, I know. It's certainly a balance -- at least for me rewriting a few hundred lines of code is several orders of magnitude less work than dealing with the fallout of fwupd hard depending on a library that isn't already installed on per-distro IOT and server targets. Every time I add a dep like that there's a huge amount of work and security review (and actual paperwork!), and for an simple open format like FDT that is documented like it is then reimplementing is fine. It also means we can use the same building format as we do for all the other formats, which means testing and fuzzing becomes trivial.

> Anyway, if we go with this, how will we deal with configuration signing?

Good question. We do get the raw GBytes for the entire FIT blob, and all the properties of the FDT, but I'm happy to add the other signing bits if it's just sub-chunks of that. I suspect we'd want to re-use the existing verification code rather than putting it in the FIT parser directly. I'll watch your youtube talk tonight, many thanks for those links.

sjg20:
> > But I'm not a big fan of rewriting libfdt, given all the sanity checks and tests that it has!
> 
> Yup, I know. It's certainly a balance -- at least for me rewriting a few hundred lines of code is several orders of magnitude less work than dealing with the fallout of fwupd hard depending on a library that isn't already installed on per-distro IOT and server targets. Every time I add a dep like that there's a huge amount of work and security review (and actual paperwork!), and for an simple open format like FDT that is documented like it is then reimplementing is fine. It also means we can use the same building format as we do for all the other formats, which means testing and fuzzing becomes trivial.

I am surprised that libfdt is not a standard library. It has been around for more than a decade.It is required by u-boot-tools which is also very common, particularly in embedded distros. Perhaps it is time those distros accepted it?

> 
> > Anyway, if we go with this, how will we deal with configuration signing?
> 
> Good question. We do get the raw GBytes for the entire FIT blob, and all the properties of the FDT, but I'm happy to add the other signing bits if it's just sub-chunks of that. I suspect we'd want to re-use the existing verification code rather than putting it in the FIT parser directly. I'll watch your youtube talk tonight, many thanks for those links.

The key thing is that the parsing code lets you hash regions of the FDT. See it here:

https://github.com/u-boot/u-boot/blob/master/include/fdt_region.h

My plan is to upstream this to here:

https://github.com/devicetree-org/fdt-tools

hughsie:
Reworked as https://github.com/fwupd/fwupd/pull/4811

hughsie:
> The key thing is that the parsing code lets you hash regions of the FDT

It certainly looks very straightforward, but I'm a bit confused *why* fwupd needs to validate the signature at all.  Of course I understand why it's validated at startup, but validating it at update time seems "helpful" at best -- and means we'd have to teach fwupd how to load the "dev" key name loading location for example.

PR #4804, opened 2022-07-05
trivial: rename thunderbolt to linux-thunderbolt

superm1:
This plugin relies on a linux only interface. There isn't a good reason to keep
a separate configuration option as we have meson tri-state features for gudev.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/fedora.sh
+++ b/contrib/ci/fedora.sh
@@ -24,7 +24,6 @@ meson .. \
     -Dplugin_dummy=true \
     -Dplugin_flashrom=enabled \
     -Dplugin_modem_manager=disabled \
-    -Dplugin_thunderbolt=enabled \
     -Dplugin_uefi_capsule=enabled \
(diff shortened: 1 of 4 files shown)


hughsie:
I'm not sure I get the logic for the rename; I mean, anything using sysfs is linux-specific, no?

superm1:
> I'm not sure I get the logic for the rename; I mean, anything using sysfs is linux-specific, no?

We have a bunch of linux-* plugins when stuff relies on linux only concepts or interfaces.
This plugin doesn't have a line of sight to ever having a future outside Linux though the way it's built up.  It's built on assumptions of what the kernel driver provides and how it works which might not be true in other driver stacks.  Some other plugins I think that could be the case they _can_ be extended still.

But if you want to keep name the same, how about just dropping the configuration option like is in this PR?

hughsie (approved):

PR #4805, opened 2022-07-06
modem-manager: add Quectel secure boot status AT command

ivanmikh:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

Add Quectel's specific AT command to check secure boot status of the modem firmware:
AT+QSECBOOT="status"

Typical responses are:
+QSECBOOT: "status",1
+QSECBOOT: "status",0

Firmware versions comparison method should still be used for EM05 modems since it does not expose AT port to the system.

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -196,41 +196,6 @@ validate_firmware_update_method(MMModemFirmwareUpdateMethod methods, GError **er
 	return FALSE;
 }
 
-static void
-fu_mm_device_probe_quectel_flags(FuMmDevice *self)
-{
-	const gchar *version = fu_device_get_version(FU_DEVICE(self));
-	g_autofree gchar *name = NULL;
-	struct {
-		const gchar *name;
-		const gchar *version;
-	} secboot[] = {{"EM05GF", "EM05GFAR07A07M1G_01.005.01.005"},
-		       {"EM05CE", "EM05CEFCR08A16M1G_LNV"},
-		       {"EM120G", "EM120RGLAPR02A07M4G_18.018.18.018"},
-		       {"EM160G", "EM160RGLAPR02A07M4G_18.018.18.018"},
-		       {"EG25GG", "EG25GGBR07A07M2G_OCPU_30.001.30.001"},
-		       {NULL, NULL}};
-
-	/* find model name and compare with table from Quectel */
-	if (version == NULL)
-		return;
-	name = g_strndup(version, 6);
-	for (guint i = 0; secboot[i].name != NULL; i++) {
-		if (g_strcmp0(name, secboot[i].name) == 0) {
-			if (fu_version_compare(version,
-					       secboot[i].version,
-					       FWUPD_VERSION_FORMAT_PLAIN) >= 0) {
-				fu_device_add_flag(FU_DEVICE(self),
-						   FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD);
-			} else {
-				fu_device_add_flag(FU_DEVICE(self),
-						   FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);
-			}
-			return;
-		}
-	}
-}
-
 static gboolean
 fu_mm_device_probe_default(FuDevice *device, GError **error)
 {
@@ -347,6 +312,9 @@ fu_mm_device_probe_default(FuDevice *device, GError **error)
 				self->port_qcdm = g_strdup_printf("/dev/%s", ports[i].name);
 			else if (ports[i].type == MM_MODEM_PORT_TYPE_MBIM)
 				self->port_mbim = g_strdup_printf("/dev/%s", ports[i].name);
+			/* to read secboot status */
+			else if (ports[i].type == MM_MODEM_PORT_TYPE_AT)
+				self->port_at = g_strdup_printf("/dev/%s", ports[i].name);
 		}
 		fu_device_add_protocol(device, "com.qualcomm.firehose");
 	}
@@ -554,12 +522,6 @@ fu_mm_device_probe_default(FuDevice *device, GError **error)
 	if (g_strcmp0(fu_device_get_vendor(device), "QUALCOMM INCORPORATED") == 0)
 		fu_device_set_vendor(device, "Qualcomm");
 
-	/* Quectel secboot enabled */
-	if (fu_device_get_name(device) != NULL &&
-	    g_str_has_prefix(fu_device_get_name(device), "QUECTEL")) {
-		fu_mm_device_probe_quectel_flags(self);
-	}
-
 	/* success */
 	return TRUE;
 }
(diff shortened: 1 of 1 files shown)


plugins/modem-manager/fu-mm-device.c
```diff
@@ -555,10 +736,7 @@ fu_mm_device_probe_default(FuDevice *device, GError **error)
 		fu_device_set_vendor(device, "Qualcomm");
 
 	/* Quectel secboot enabled */
-	if (fu_device_get_name(device) != NULL &&
-	    g_str_has_prefix(fu_device_get_name(device), "QUECTEL")) {
-		fu_mm_device_probe_quectel_flags(self);
-	}
+	fu_mm_device_probe_quectel_flags(self);
```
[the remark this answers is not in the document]
superm1: Why not just look for the name like before as a pre-requisite?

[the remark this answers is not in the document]
ivanmikh: > Why not just look for the name like before as a pre-requisite?

Name is different for every module unfortunately, EG25G is `QUECTEL Mobile Broadband Module`, EM05-G is `EM05-G`, EM160R-GL is `EM160R GL`, etc.

But I agree that it is good to have a check for Quectel devices, I will add one based on VIDs for both `+GETFWBRANCH` and `+QSECBOOT`. There are only two Quectel VIDs for now: `PCI:1eac` and `USB:2c7c`

ivanmikh:
So `AT+GETFWBRANCH` is only being run for modems with `ModemManagerBranchAtCommand` quirk set. Currently its only EG25-G.

Also I moved secure boot status to `fu_mm_device_setup` and added a VID check.

@DylanVanAssche please take a look at the `fu_mm_device_setup_branch_at` changes. I changed the logic a little bit so it does not open AT port if branch_at command is not set.

DylanVanAssche:
> So `AT+GETFWBRANCH` is only being run for modems with `ModemManagerBranchAtCommand` quirk set. Currently its only EG25-G.
> 
> Also I moved secure boot status to `fu_mm_device_setup` and added a VID check.
> 
> @DylanVanAssche please take a look at the `fu_mm_device_setup_branch_at` changes. I changed the logic a little bit so it does not open AT port if branch_at command is not set.

Looks fine, if the self->at_branch variable is not set because the quirk is not available, it is not needed to run that method.

plugins/modem-manager/fu-mm-device.c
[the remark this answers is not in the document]
DylanVanAssche: Can this also be used to turn off secure boot? like `AT+QSECBOOT="secure_boot, 0"`? If secure boot blocks switching branches, we need to handle that part properly.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1742,45 +1730,112 @@ fu_mm_device_attach(FuDevice *device, FuProgress *progress, GError **error)
...
+		g_autoptr(FuDeviceLocker) locker = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		/* Create IO channel to send AT commands to the modem */
+		locker = fu_device_locker_new_full(self,
+						   (FuDeviceLockerFunc)fu_mm_device_io_open,
+						   (FuDeviceLockerFunc)fu_mm_device_io_close,
+						   &error_local);
+		if (locker == NULL)
+			g_debug("failed to open AT port: %s", error_local->message);
```
hughsie, line 1771: same here

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1742,45 +1730,112 @@ fu_mm_device_attach(FuDevice *device, FuProgress *progress, GError **error)
...
-	/* success */
-	return TRUE;
+static void
+fu_mm_device_setup_secboot_status(FuDevice *device)
+{
+	GPtrArray *vendors;
+	FuMmDevice *self = FU_MM_DEVICE(device);
+
+	vendors = fu_device_get_vendor_ids(device);
+	if (vendors != NULL && vendors->len != 0) {
```
hughsie, line 1813: you don't need this line; vendors has to be non-NULL and the loop does the len check for you

plugins/modem-manager/fu-mm-device.c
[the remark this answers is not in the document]
ivanmikh: > Can this also be used to turn off secure boot? like AT+QSECBOOT="secure_boot, 0"? If secure boot blocks switching branches, we need to handle that part properly.

No, unfortunately. Flashing secure boot firmware blows an efuse, so there is no way to disable it and modem can only be flashed with signed firmware after that.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1742,45 +1730,112 @@ fu_mm_device_attach(FuDevice *device, FuProgress *progress, GError **error)
...
+					   &error_local);
 
 	/* firmware branch AT command may fail if not implemented,
 	 * clear error if not supported */
-	if (self->branch_at != NULL) {
-		g_autoptr(GError) error_branch = NULL;
-		if (!fu_mm_device_at_cmd(self, self->branch_at, TRUE, &error_branch))
-			g_debug("unable to get firmware branch: %s", error_branch->message);
+	if (locker == NULL)
+		g_debug("failed to open AT port: %s", error_local->message);
```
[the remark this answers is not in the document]
ivanmikh: but it is followed by `else if`, so if locker is NULL it prints the error_local message and then function returns

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1742,45 +1730,112 @@ fu_mm_device_attach(FuDevice *device, FuProgress *progress, GError **error)
...
 static gboolean
 fu_mm_device_setup(FuDevice *device, GError **error)
 {
 	FuMmDevice *self = FU_MM_DEVICE(device);
 
-	if (self->port_at != NULL) {
-		if (!fu_mm_device_setup_branch_at(self, error))
-			return FALSE;
-	}
+	fu_mm_device_setup_branch_at(self);
```
DylanVanAssche, line 1831: @ivanmikh Since you said that you cannot disable secboot at all, branches should be blocked IMO since you can't switch at all. Or at least disable the `community` option if Quectel decides to use branches at some point.

DylanVanAssche: Or the community one should be able to sign their firmware?

hughsie: ack; if the device is signed then my definition it can't support branch-switching.

ivanmikh: okay, i added a check for signed firmware before running firmware branch AT command

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1755,31 +1756,94 @@ fu_mm_device_setup_branch_at(FuMmDevice *self, GError **error)
...
+		if (locker == NULL) {
+			g_debug("failed to open AT port: %s", error_local->message);
+			return;
+		}
+
+		if (!fu_mm_device_at_cmd(self, "AT+QSECBOOT=\"status\"", TRUE, &error_local))
+			g_debug("failed to send AT+QSECBOOT command: %s", error_local->message);
+	} else {
+		const gchar *version = fu_device_get_version(FU_DEVICE(self));
+		g_autofree gchar *name = NULL;
```
hughsie, line 1791: I think if AT+QSECBOOT fails, we want to fall back to the table; i.e. return TRUE if fu_mm_device_at_cmd succeeded, otherwise fall back to the lower parts. I've added you to the committers, so if you push your branch to the origin server I can fix it up for you.

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1755,31 +1756,94 @@ fu_mm_device_setup_branch_at(FuMmDevice *self, GError **error)
...
+}
+
+static void
+fu_mm_device_setup_secboot_status(FuDevice *device)
+{
+	GPtrArray *vendors;
+	FuMmDevice *self = FU_MM_DEVICE(device);
+
+	vendors = fu_device_get_vendor_ids(device);
+	for (guint i = 0; i < vendors->len; i++) {
```
hughsie, line 1828: Just use fu_device_has_vendor_id() here please.

plugins/modem-manager/fu-mm-device.c
ivanmikh: thanks! just pushed the latest edit to the origin

Biktorgj:
Just my two cents, the EG25-G, from version `EG25GGBR07A07M2G_30.003.30.003` and higher, has a different command to check secure boot:
`AT+QCFG="secbootstat"`
Possible responses seem to be:
`+QCFG: "secbootstat",0`
`+QCFG: "secbootstat",1`

ivanmikh:
> Just my two cents, the EG25-G, from version `EG25GGBR07A07M2G_30.003.30.003` and higher, has a different command to check secure boot: `AT+QCFG="secbootstat"` Possible responses seem to be: `+QCFG: "secbootstat",0` `+QCFG: "secbootstat",1`

Is this one available on community firmware?

Biktorgj:
No, this is only available in stock, but I can add it to the community firmware if needed

ivanmikh:
> No, this is only available in stock, but I can add it to the community firmware if needed (right now with the community firmware it will just return "OK")

Its fine. Thanks for bringing it up, will add it to the commit as well

ivanmikh:
Hey,
I have updated the logic, so if AT commands fails it falls back to the table comparison. Also added another AT command for EG25-G

hughsie (approved):
Thanks!

PR #4806, opened 2022-07-06
Add a title and long translated description for security attributes

hughsie:
We need these in at least two projects (gnome-control-center, cockpit)
and it makes sense to store these where we define the new IDs.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -6281,6 +6281,12 @@ fu_engine_security_attrs_depsolve(FuEngine *self)
 			}
 			fwupd_security_attr_set_name(attr, name_tmp);
 		}
+		if (fwupd_security_attr_get_title(attr) == NULL)
+			fwupd_security_attr_set_title(attr, fu_security_attr_get_title(attr));
+		if (fwupd_security_attr_get_description(attr) == NULL) {
+			fwupd_security_attr_set_description(attr,
+							    fu_security_attr_get_description(attr));
+		}
 	}
 
 	/* set the obsoletes flag for each attr */
--- a/src/fu-security-attr.c
+++ b/src/fu-security-attr.c
@@ -208,6 +208,315 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
 }
 
 const gchar *
+fu_security_attr_get_title(FwupdSecurityAttr *attr)
+{
+	const gchar *appstream_id = fwupd_security_attr_get_appstream_id(attr);
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BIOSWE) == 0) {
+		/* TRANSLATORS: Title: firmware refers to the flash chip in the computer */
+		return _("Firmware Write Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0) {
+		/* TRANSLATORS: Title: firmware refers to the flash chip in the computer */
+		return _("Firmware Write Protection Lock");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("Firmware BIOS Region");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_DESCRIPTOR) == 0) {
+		/* TRANSLATORS: Title: firmware refers to the flash chip in the computer */
+		return _("Firmware BIOS Descriptor");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PREBOOT_DMA_PROTECTION) == 0) {
+		/* TRANSLATORS: Title: DMA as in https://en.wikipedia.org/wiki/DMA_attack  */
+		return _("Pre-boot DMA Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ENABLED) == 0) {
+		/* TRANSLATORS: Title: BootGuard is a trademark from Intel */
+		return _("Intel BootGuard");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_VERIFIED) == 0) {
+		/* TRANSLATORS: Title: BootGuard is a trademark from Intel,
+		 * verified boot refers to the way the boot process is verified */
+		return _("Intel BootGuard Verified Boot");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ACM) == 0) {
+		/* TRANSLATORS: Title: BootGuard is a trademark from Intel,
+		 * ACM means to verify the integrity of Initial Boot Block */
+		return _("Intel BootGuard ACM Protected");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_POLICY) == 0) {
+		/* TRANSLATORS: Title: BootGuard is a trademark from Intel,
+		 * error policy is what to do on failure */
+		return _("Intel BootGuard Error Policy");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_OTP) == 0) {
+		/* TRANSLATORS: Title: BootGuard is a trademark from Intel */
+		return _("Intel BootGuard Fuse");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET_ENABLED) == 0) {
+		/* TRANSLATORS: Title: CET = Control-flow Enforcement Technology,
+		 * enabled means supported by the processor */
+		return _("Intel CET Enabled");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET_ACTIVE) == 0) {
+		/* TRANSLATORS: Title: CET = Control-flow Enforcement Technology,
+		 * active means being used by the OS */
+		return _("Intel CET Active");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_SMAP) == 0) {
+		/* TRANSLATORS: Title: SMAP = Supervisor Mode Access Prevention */
+		return _("Intel SMAP");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_ENCRYPTED_RAM) == 0) {
+		/* TRANSLATORS: Title: Memory contents are encrypted, e.g. Intel TME */
+		return _("Encrypted RAM");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_IOMMU) == 0) {
+		/* TRANSLATORS: Title:
+		 * https://en.wikipedia.org/wiki/Input%E2%80%93output_memory_management_unit */
+		return _("IOMMU Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_LOCKDOWN) == 0) {
+		/* TRANSLATORS: Title: lockdown is a security mode of the kernel */
+		return _("Linux Kernel Lockdown");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_TAINTED) == 0) {
+		/* TRANSLATORS: Title: if it's tainted or not */
+		return _("Linux Kernel Verification");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_SWAP) == 0) {
+		/* TRANSLATORS: Title: swap space or swap partition */
+		return _("Linux Swap");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_RAM) == 0) {
+		/* TRANSLATORS: Title: sleep state */
+		return _("Suspend To RAM");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_IDLE) == 0) {
+		/* TRANSLATORS: Title: a better sleep state */
+		return _("Suspend To Idle");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_UEFI_PK) == 0) {
+		/* TRANSLATORS: Title: PK is the 'platform key' for the machine */
+		return _("UEFI Platform Key");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_UEFI_SECUREBOOT) == 0) {
+		/* TRANSLATORS: Title: SB is a way of locking down UEFI */
+		return _("UEFI Secure Boot");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_TPM_EMPTY_PCR) == 0) {
+		/* TRANSLATORS: Title: PCRs (Platform Configuration Registers) shouldn't be empty */
+		return _("TPM Platform Configuration");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_TPM_RECONSTRUCTION_PCR0) == 0) {
+		/* TRANSLATORS: Title: the PCR is rebuilt from the TPM event log */
+		return _("TPM Reconstruction");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_TPM_VERSION_20) == 0) {
+		/* TRANSLATORS: Title: TPM = Trusted Platform Module */
+		return _("TPM v2.0");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MEI_MANUFACTURING_MODE) == 0) {
+		/* TRANSLATORS: Title: MEI = Intel Management Engine */
+		return _("Intel Management Engine Manufacturing Mode");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MEI_OVERRIDE_STRAP) == 0) {
+		/* TRANSLATORS: Title: MEI = Intel Management Engine, and the "override" is enabled
+		 * with a jumper -- luckily it is probably not accessible to end users on consumer
+		 * boards */
+		return _("Intel Management Engine Override");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MEI_VERSION) == 0) {
+		/* TRANSLATORS: Title: MEI = Intel Management Engine */
+		return _("Intel Management Engine Version");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_FWUPD_UPDATES) == 0) {
+		/* TRANSLATORS: Title: if firmware updates are available */
+		return _("Firmware Updates");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_FWUPD_ATTESTATION) == 0) {
+		/* TRANSLATORS: Title: if we can verify the firmware checksums */
+		return _("Firmware Attestation");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_FWUPD_PLUGINS) == 0) {
+		/* TRANSLATORS: Title: if the fwupd plugins are all present and correct */
+		return _("Firmware Updater Verification");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PLATFORM_DEBUG_ENABLED) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PLATFORM_DEBUG_LOCKED) == 0) {
+		/* TRANSLATORS: Title: Allows debugging of parts using proprietary hardware */
+		return _("Platform Debugging");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUPPORTED_CPU) == 0) {
+		/* TRANSLATORS: Title: if fwupd supports HSI on this chip */
+		return _("Processor Security Checks");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_ROLLBACK_PROTECTION) == 0) {
+		/* TRANSLATORS: Title: if firmware enforces rollback protection */
+		return _("AMD Rollback Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_REPLAY_PROTECTION) == 0) {
+		/* TRANSLATORS: Title: if hardware enforces control of SPI replays */
+		return _("AMD Firmware Replay Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_WRITE_PROTECTION) == 0) {
+		/* TRANSLATORS: Title: if hardware enforces control of SPI writes */
+		return _("AMD Firmware Write Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PLATFORM_FUSED) == 0) {
+		/* TRANSLATORS: Title: if the part has been fused */
+		return _("Fused Platform");
+	}
+	return NULL;
+}
+
+/* one line describing the attribute */
+const gchar *
+fu_security_attr_get_description(FwupdSecurityAttr *attr)
+{
+	const gchar *appstream_id = fwupd_security_attr_get_appstream_id(attr);
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BIOSWE) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_DESCRIPTOR) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_WRITE_PROTECTION) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Firmware Write Protection protects device firmware memory from being "
+			 "accessed and tampered with.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PREBOOT_DMA_PROTECTION) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Pre-boot DMA protection prevents devices from being connected to the "
+			 "computer and accessing system memory.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ENABLED) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ACM) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_OTP) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_VERIFIED) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Intel BootGuard prevents unauthorized device software from operating "
+			 "when the device is started.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_POLICY) == 0) {
+		/* TRANSLATORS: longer description */
+		return _(
+		    "Intel BootGuard Error Policy ensures the device does not continue to start if "
+		    "its device software has been tampered with.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET_ENABLED) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET_ACTIVE) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Intel Control-Flow Enforcement Technology detects and prevents certain "
+			 "methods for running malicious software on the device.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_SMAP) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Intel Supervisor Mode Access Prevention ensures critical parts of "
+			 "device memory are not accessed by less secure programs.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_ENCRYPTED_RAM) == 0) {
+		/* TRANSLATORS: longer description */
+		return _(
+		    "Encrypted RAM makes it impossible for information that is stored in device "
+		    "memory to be read, if the memory chip is removed and accessed.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_IOMMU) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("IOMMU Protection prevents connected devices from accessing unauthorized "
+			 "parts of system memory.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_KERNEL_LOCKDOWN) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Linux Kernel Lockdown mode prevents administrator (root) accounts from "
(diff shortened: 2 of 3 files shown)


hughsie:
@superm1 grab a `${beverage}` before looking at this.

libfwupd/fwupd-security-attr.c
```diff
@@ -445,6 +447,52 @@ fwupd_security_attr_set_name(FwupdSecurityAttr *self, const gchar *name)
...
+	priv->summary = g_strdup(summary);
+}
+
+/**
+ * fwupd_security_attr_set_description:
+ * @self: a #FwupdSecurityAttr
+ * @description: (nullable): the attribute description
+ *
+ * Sets the attribute description.
+ *
```
superm1, line 479: what's the difference with a summary and description?

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_UEFI_PK) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_UEFI_SECUREBOOT) == 0) {
+		/* TRANSLATORS: Title: the secure boot chain */
+		return _("Secure Boot");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET_ENABLED) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_CET_ACTIVE) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_SMAP) == 0) {
+		/* TRANSLATORS: Title: in this case means the main CPU */
+		return _("Processor Protection");
```
superm1, line 240: CET is more "process" protection isn't it?

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	const gchar *appstream_id = fwupd_security_attr_get_appstream_id(attr);
+	g_autoptr(GPtrArray) strs = g_ptr_array_new();
+	gboolean contact_hw_vendor = FALSE;
+	gboolean check_bios_settings = FALSE;
+
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BIOSWE) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_WRITE_PROTECTION) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			/* TRANSLATORS: longer success message (unlikely to see) */
+			g_ptr_array_add(strs, _("The system firmware is locked and not writable."));
```
superm1, line 289: Not writable by software.  This doesn't mitigate dediprog!

src/fu-security-attr.c
hughsie: Ya, I was going for more protection provided *by* the processor.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			/* TRANSLATORS: longer success message (unlikely to see) */
+			g_ptr_array_add(strs, _("The system firmware cannot be easily unlocked."));
+		} else {
+			g_ptr_array_add(
+			    strs,
+			    /* TRANSLATORS: longer failure message */
+			    _("The system firmware can be unlocked and written by any user."));
```
superm1, line 306: again, any user is too strong

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+			    strs,
+			    /* TRANSLATORS: longer failure message */
+			    _("The system firmware is unlocked and writable by any user."));
+			contact_hw_vendor = TRUE;
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			/* TRANSLATORS: longer success message (unlikely to see) */
+			g_ptr_array_add(strs, _("The system firmware cannot be easily unlocked."));
```
superm1, line 301: what does easily mean?

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			/* TRANSLATORS: longer success message (unlikely to see) */
+			g_ptr_array_add(strs, _("The system firmware cannot be unlocked."));
+		} else {
+			g_ptr_array_add(
+			    strs,
+			    /* TRANSLATORS: longer failure message */
+			    _("The system firmware could be unlocked and written by any user."));
```
superm1, line 318: again, any user is too strong

libfwupd/fwupd-security-attr.c
```diff
@@ -543,6 +591,48 @@ fwupd_security_attr_get_name(FwupdSecurityAttr *self)
...
+	return priv->summary;
+}
+
+/**
+ * fwupd_security_attr_get_description:
+ * @self: a #FwupdSecurityAttr
+ *
+ * Gets the attribute description.
+ *
+ * Interestingly, the fwupd client program may be able to get translations for this value using
```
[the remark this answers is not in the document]
hughsie: I'll drop that; interesting to me -- perhaps.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_DESCRIPTOR) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("Critical parts of the system firmware are protected."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("Critical parts of the system firmware could be changed "
+					  "by any user."));
```
superm1, line 331: again, any user is too strong

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+			    /* TRANSLATORS: longer failure message */
+			    _("The system firmware could be unlocked and written by any user."));
+			contact_hw_vendor = TRUE;
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_DESCRIPTOR) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("Critical parts of the system firmware are protected."));
```
superm1, line 326: what does critical part mean here?

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BIOSWE) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_WRITE_PROTECTION) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			/* TRANSLATORS: longer success message (unlikely to see) */
+			g_ptr_array_add(strs, _("The system firmware is locked and not writable."));
+		} else {
+			g_ptr_array_add(
+			    strs,
+			    /* TRANSLATORS: longer failure message */
+			    _("The system firmware is unlocked and writable by any user."));
```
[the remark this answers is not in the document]
hughsie: Agreed, I've added a `privileged` here.

src/fu-security-attr.c
hughsie: a ` at runtime` suffix?

src/fu-security-attr.c
hughsie: Fixed.

src/fu-security-attr.c
hughsie: Fixed

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					_("Critical parts of the system firmware could be changed "
+					  "by any user."));
+			contact_hw_vendor = TRUE;
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PREBOOT_DMA_PROTECTION) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("System memory is protected from malicious devices "
```
superm1, line 339: How about `System memory implemented protections from malicious devices...`

src/fu-security-attr.c
hughsie: Critical areas? There are a few attacks where the descriptor is unlocked and you can change the protected ranges, and I wanted to signify that this allowed other bad things to happen.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					_("System memory is not protected from malicious devices "
+					  "before starting the OS."));
+			contact_hw_vendor = TRUE;
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_BOOTGUARD_ENABLED) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("System firmware is being checked by the processor at "
```
superm1, line 353: This needs to talk about signature verification not just being checked

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("System firmware is being verified by the processor at "
+					  "system start."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("System firmware is not being verified by the processor "
+					  "at system start."));
+			contact_hw_vendor = TRUE;
+		}
```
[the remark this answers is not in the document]
hughsie: Ack, agree, fixed.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					_("The processor will not detect and ignore more attack "
+					  "attempts from malicious software."));
+			check_bios_settings = TRUE;
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_INTEL_SMAP) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("The processor can detect and ignore simple attack "
```
superm1, line 447: what's a simple attack?

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					/* TRANSLATORS: longer failure message */
+					_("The processor cannot detect and ignore simple attack "
+					  "attempts from malicious software."));
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_ENCRYPTED_RAM) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("Memory is being encrypted and cannot be read if removed "
```
superm1, line 460: it can be read, it just can't be interpreted

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		}
+	}
+
+	/* any postfix */
+	if (contact_hw_vendor) {
+		g_ptr_array_add(strs,
+				/* TRANSLATORS: OEM is the original end manufacturer, i.e. the logo
+				   on the case */
+				_("Ensure you have installed all available firmware updates and if "
+				  "still not fixed contact your OEM vendor."));
```
superm1, line 783: we know who their OEM is.  How about we replace "your OEM vendor" with a proper value from SMBIOS?

src/fu-security-attr.c
superm1: Yeah, NVM I don't like my proposed wording either.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("System firmware is being checked by the processor at "
+					  "system start."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("System firmware is not being checked by the processor "
+					  "at system start."));
+			contact_hw_vendor = TRUE;
```
superm1, line 360: This needs `check_bios_settings` too.

src/fu-security-attr.c
hughsie: How about `The processor can detect and ignore malicious attack attempts from malicious software`

src/fu-security-attr.c
superm1: That's good

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+			g_ptr_array_add(
+			    strs,
+			    /* TRANSLATORS: longer success message (unlikely to see) */
+			    _("Malicious peripherals cannot write to main system memory."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("Some malicious peripherals could potentially write to "
+					  "main system memory."));
+			check_bios_settings = TRUE;
```
superm1, line 480: I'm not sure I see the value of IOMMU and pre-boot IOMMU being a separate thing.  If pre-boot IOMMU is enabled the Linux kernel automatically turns on IOMMU.  The only way to avoid it is custom kernel command line then in Linux.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					_("The system sleep mode is set correctly so that files "
+					  "are not readable."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("The system sleep mode could allow an attacker to remove "
+					  "system memory and read the content."));
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_IDLE) == 0) {
```
superm1, line 540: S2I and S3 should combine.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("The system sleep mode is set correctly so that files "
+					  "are not readable."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("The system sleep mode could allow an attacker to remove "
+					  "system memory and read the content."));
```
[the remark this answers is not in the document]
hughsie: Correct, added.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("The system has been correctly configured when assembed "
+					  "by the vendor and firmware is being protected."));
+		} else {
+			g_ptr_array_add(
+			    strs,
+			    /* TRANSLATORS: longer failure message */
+			    _("The system has not been correctly configured when assembed by the "
+			      "vendor, which may allow an attacker to replace system firmware."));
```
superm1, line 633: maybe replace vendor with the name of the vendor

src/fu-security-attr.c
hughsie: Ack, agree. Fixed.

src/fu-security-attr.c
hughsie: I guess we could do `your OS` -> `Fedora` too, right?

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PLATFORM_DEBUG_ENABLED) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			/* TRANSLATORS: longer success message (unlikely to see) */
+			g_ptr_array_add(strs, _("Low-level hardware debugging is not enabled."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("Low-level hardware debugging has been enabled for this "
+					  "system and the firmware and OS are not secure."));
```
superm1, line 717: this should be contact vendor if they didn't properly lock the hardware down

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PLATFORM_DEBUG_LOCKED) == 0) {
```
superm1, line 720: combine with previous attribute

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("The system firmware cannot be downgraded to an older "
+					  "legitimate version with known security problems."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("The system firmware can be downgraded to an older "
+					  "legitimate version with known security problems."));
+			contact_hw_vendor = TRUE;
```
superm1, line 759: with potential security problems

src/fu-security-attr.c
superm1: Yes

src/fu-security-attr.c
superm1: but then that doesn't make sense when you compare to "memory protection".  Those aren't provided _by_ the memory.
Also, these are SOC protections not X86 core protections aren't they?

src/fu-security-attr.c
hughsie: Hmm. We can't do this unless we do the $OS$ substitution client side; otherwise the translation IDs don't match and `dgettext("fwupd",str)` doesn't work. I suppose we could export `fwupd_client_get_host_vendor` for this and maybe even `fwupd_client_get_os_vendor()` but it's getting a bit unwieldy then.

hughsie: Ahh; maybe we can use `fwupd_security_attr_add_metadata()` -- will ponder.

libfwupd/fwupd-security-attr.c
superm1: So maybe summary should actually be title?

src/fu-security-attr.c
superm1: but to a user do they "care" which is which?

src/fu-security-attr.c
superm1: How about qualifying it `The processor can detect and ignore some attacks from malicious software`

src/fu-security-attr.c
hughsie: What about `An important area of the system firmware could be changed by a privileged user.`

src/fu-security-attr.c
hughsie: WFM, thanks.

src/fu-security-attr.c
hughsie: I've heard of a vendor doing exactly that in the field -- but that was part of a BIOS update (i.e. `contact_hw_vendor`), not adding BIOS setting.

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+					  "provided by the vendor for this system."));
+			contact_hw_vendor = TRUE;
+		}
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_FWUPD_PLUGINS) == 0) {
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(
+			    strs,
+			    /* TRANSLATORS: longer success message (unlikely to see) */
+			    _("The fwupd project is set up correctly and all plugins are valid."));
```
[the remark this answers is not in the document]
hughsie: Ya, thanks. Fixed.

src/fu-security-attr.c
superm1: I think switching out host vendor is more important than OS vendor is FWIW.  I think there is less that you'll contact Red Hat or Canonical about than you would a HW vendor

src/fu-security-attr.c
```diff
@@ -203,6 +203,598 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		if (fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS)) {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer success message (unlikely to see) */
+					_("This system can be tested for firmware security."));
+		} else {
+			g_ptr_array_add(strs,
+					/* TRANSLATORS: longer failure message */
+					_("This system cannot be tested for firmware security as "
+					  "it does not have a supported processor or mainboard."));
+			contact_hw_vendor = TRUE;
```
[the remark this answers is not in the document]
hughsie: Ack, agree, removed.

src/fu-security-attr.c
hughsie: Agree, fixed, thx.

src/fu-engine.c
```diff
@@ -6236,6 +6236,26 @@ fu_engine_ensure_security_attrs(FuEngine *self)
 			}
 			fwupd_security_attr_set_name(attr, name_tmp);
 		}
+		if (fwupd_security_attr_get_title(attr) == NULL)
+			fwupd_security_attr_set_title(attr, fu_security_attr_get_title(attr));
+		if (fwupd_security_attr_get_description(attr) == NULL) {
+			g_autofree gchar *tmp = fu_security_attr_get_description(attr);
+#if 0
```
superm1, line 6243: why comment this section out?  seems sensible to me

hughsie: Ahh, because it breaks the gettext trick; changing the daemon locale at runtime just to return the correct translations seems kinda risky. I've removed this and added some notes on the client-side replacement action.

src/fu-security-attr.c
superm1: Ah I see- makes sense.

src/fu-security-attr.c
hughsie: I don't disagree, and I don't think it makes sense to show any status IOMMU if the Pre-boot IOMMU isn't enabled. Maybe we should use the obsoletes feature for that?

hughsie:
Right; we're doing a lot of UX thinking first. No panic.

hughsie:
@superm1 we've done quite a bit of back-and-forth with the GNOME design team on the strings and the general plan; I've pushed the latest -- any complaints merging something like this soonish so it can be included in the release on Monday?

src/fu-security-attr.c
```diff
@@ -207,6 +207,314 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI Lock");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI BIOS Region");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_DESCRIPTOR) == 0) {
+		/* TRANSLATORS: Title: SPI refers to the flash chip in the computer */
+		return _("SPI BIOS Descriptor");
```
superm1, line 228: All of these I know SPI is the correct term but I think you should go for "Flash Chip" because that's a lot more accessible of a description.

src/fu-security-attr.c
```diff
@@ -207,6 +207,314 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MEI_OVERRIDE_STRAP) == 0) {
+		/* TRANSLATORS: Title: MEI = Intel Management Engine, and the "override" is enabled
+		 * with a jumper -- luckily it is probably not accessible to end users on consumer
+		 * boards */
+		return _("MEI Override");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MEI_VERSION) == 0) {
+		/* TRANSLATORS: Title: MEI = Intel Management Engine */
+		return _("MEI Version");
+	}
```
superm1, line 334: Similar to my SPI comment, I think you should expand this to "Intel Management Engine Interface ..."

src/fu-security-attr.c
```diff
@@ -207,6 +207,314 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		/* TRANSLATORS: Title: if firmware enforces rollback protection */
+		return _("Rollback Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_REPLAY_PROTECTION) == 0) {
+		/* TRANSLATORS: Title: if hardware enforces control of SPI replays */
+		return _("SPI Replay Protection");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_WRITE_PROTECTION) == 0) {
+		/* TRANSLATORS: Title: if hardware enforces control of SPI writes */
+		return _("SPI Write Protection");
```
superm1, line 366: Same as my SPI comment above

src/fu-security-attr.c
```diff
@@ -207,6 +207,314 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+		/* TRANSLATORS: longer description */
+		return _("Kernel Swap temporarily saves information to disk as you work. This "
+			 "information which could be accessed by someone if they gained access to "
+			 "the disk.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_RAM) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SUSPEND_TO_IDLE) == 0) {
+		/* TRANSLATORS: longer description */
+		return _(
+		    "During system sleep the device’s memory chip could be removed, allowing the "
```
superm1, line 451: Since the same is used for S3/S2I you probably need to make it clear which is which

src/fu-security-attr.c
```diff
@@ -207,6 +207,314 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_TPM_VERSION_20) == 0) {
+		return _(
+		    /* TRANSLATORS: longer description */
+		    "TPM (Trusted Platform Module) v2 is a computer chip which detects when the "
+		    "hardware components have been tampered with.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_MEI_MANUFACTURING_MODE) == 0 ||
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PLATFORM_FUSED) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Manufacturing Mode is used during device assembly and disables security "
```
superm1, line 471: Technically it doesn't disable them, it's _before_ they have been enabled.

superm1 (commented):
You're a little inconsistent on the descriptions for how much of the title to include if any.  Knowing this will show up in a screen with both title and description together my suggestion is to remove all mentions of the "title" in the description field for all attributes

superm1:
> @superm1 we've done quite a bit of back-and-forth with the GNOME design team on the strings and the general plan; I've pushed the latest -- any complaints merging something like this soonish so it can be included in the release on Monday?

I think it's close, just a little more consistency and a few nits suggested.

src/fu-security-attr.c
hughsie: What about just `Firmware`? Most users won't know what flash is either.

src/fu-security-attr.c
hughsie: `Manufacturing Mode is used during device assembly when security features have yet to be enabled.` perhaps?

src/fu-security-attr.c
hughsie: The designers wanted to make the title the unique thing, and make the description more generic.

src/fu-security-attr.c
```diff
@@ -207,6 +207,314 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
...
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_BLE) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_SMM_BWP) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_SPI_DESCRIPTOR) == 0 &&
+	    g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_SPI_WRITE_PROTECTION) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Protects device firmware memory from being accessed and tampered with.");
+	}
+	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_PREBOOT_DMA_PROTECTION) == 0) {
+		/* TRANSLATORS: longer description */
+		return _("Pre-boot DMA protection prevents devices from being connected to the "
```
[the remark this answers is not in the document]
hughsie: Design rationale here: https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/1938

src/fu-security-attr.c
superm1: Great idea

src/fu-security-attr.c
superm1: Yup

src/fu-security-attr.c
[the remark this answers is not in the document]
superm1: But you're inconsistent is the problem then. If the title should be in the description it should be everywhere.

src/fu-security-attr.c
hughsie: Added to the google doc

src/fu-security-attr.c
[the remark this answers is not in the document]
hughsie: I think the aim is to include it everywhere it makes sense.

src/fu-security-attr.c
hughsie: Maybe a second sentence? `Using Suspend-to-Idle is better than Suspend to Disk` etc.

superm1: Yeah something like that would work.

src/fu-security-attr.c
superm1: I think the google doc is no longer maintained right?   Shouldn't it go into the markdown document now?

hughsie: I've sent you a link to the gdoc.

superm1: Wouldn't it have been easier to just review it all here in one place? :)

src/fu-security-attr.c
hughsie: I've made a lot of these nicer to read.

src/fu-security-attr.c
hughsie: Ya, I'm tempted to push what we have and then iterate on specific strings when we can see then in an actual UI.

src/fu-security-attr.c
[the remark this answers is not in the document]
hughsie: I've added it to all by two of them to be consistent.

src/fu-security-attr.c
hughsie: Done.

src/fu-security-attr.c
hughsie: Fixed.

PR #4811, opened 2022-07-08
vbe: Add a new plugin for writing Flat Image Trees

hughsie:
This adds the logic for the VBE plugin. It supports reading information
from the system to determine which VBE method (a device in fwupd) is in
use, then creating those devices to handle the actual update.

Verified Boot for Embedded (VBE) relies on FIT files to operate.
FIT is a way of packaging multiple images along with information about
them.

Signed-off-by: Simon Glass <sjg@chromium.org>

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@sjg20 I know this doesn't implement the per-device state, but if we want to do that we probably need to use the existing "history" database that the fwupd engine maintains. I still think the only way this is going to work well is writing the version reported by fwupd to the flash image itself. I've also left your name as the author as I've just rewritten things to use the new stuff in libfwupdplugin and made it all a bit more GObject-y. If you want me to use my name instead that's fine too, but I didn't want to step on any toes.

hughsie:
@sjg20 if this gets your approval I'd be happy to merge it into 1.8.2 -- and we can work on more for later releases.

sjg20 (approved):
If this is the way you wan to go, I am OK with it, as mentioned in more detail in https://github.com/fwupd/fwupd/pull/4794

sjg20:
> If this is the way you wan to go, I am OK with it, as mentioned in more detail in #4794

Should we avoid squashing, so we have separate commits? For some reason only one commit is shown here, but there are two more

hughsie:
> Should we avoid squashing, so we have separate commits

In this case I don't think it's needed, most of the code was re-written or moved around and so I don't think the other commit messages make a huge amount of sense. Lets merge this one, then we can start working on the nvram and A/B partition changes.

PR #4813, opened 2022-07-08
Do not make failing to parse the i2c bus number fatal

hughsie:
This acccidentally become more strict in bb548f15f0 and means there
should be no more false-positive daemon warnings at startup.

Fixes the 2nd half of https://github.com/fwupd/fwupd/discussions/4810

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@hugh712

PR #4817, opened 2022-07-11
elanfp: Add new protocol for CFU update

MichaelCheng04:
Add new protocol for CFU update.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/elanfp/fu-elanfp-device.c
+++ b/plugins/elanfp/fu-elanfp-device.c
@@ -30,6 +30,19 @@
 
 #define TAG_SEND_COMMAND	0xFA
 
+/**
+ * FU_ELAN_FP_DEVICE_FLAG_USB_CONTROL_TRANSFER:
+ *
+ * Use usb control transfer.
+ */
+#define FU_ELAN_FP_DEVICE_FLAG_USB_CONTROL_TRANSFER (1 << 0)
+/**
+ * FU_ELAN_FP_DEVICE_FLAG_USB_BULK_TRANSFER:
+ *
+ * Use usb bulk transfer.
+ */
+#define FU_ELAN_FP_DEVICE_FLAG_USB_BULK_TRANSFER (1 << 1)
+
 struct _FuElanfpDevice {
 	FuUsbDevice parent_instance;
 };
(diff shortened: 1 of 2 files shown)


plugins/elanfp/fu-elanfp-device.c
```diff
@@ -90,10 +127,22 @@ static gboolean
...
 {
 	GUsbDevice *usb_device = fu_usb_device_get_dev(FU_USB_DEVICE(self));
+	guint16 pid = (guint)fu_usb_device_get_pid(FU_USB_DEVICE(self));
+	guint8 endpoint = ELAN_EP_CMD_IN;
 	gsize actual = 0;
+	gboolean is_new_protocol = FALSE;
+
+
+	if ((pid != 0x0C7E) && (pid != 0x0C82)) {
+					is_new_protocol = TRUE;
```
hughsie, line 137: what does "new" mean? When the next protocol version is added will that be new_new? or new2? :) I think you want to use fu_device_register_private_flag() on a new flag called "use-bulk-transfer" and then set this in the quirk file, i.e.

    [USB\VID_04F3&PID_0C82]
    Flags = use-bulk-transfer

and then use `fu_device_has_private_flag` in the code.

plugins/elanfp/fu-elanfp-device.c
```diff
@@ -40,40 +42,75 @@ fu_elanfp_iap_send_command(FuElanfpDevice *self,
...
+					sendsz,
+					&actual,
+					BULK_SEND_TIMEOUT_MS,
+					NULL,
+					error)) {
+		g_prefix_error(error, "failed to send command (bulk): ");
+		return FALSE;
+		}
+	}
+	else {
```
hughsie, line 94: intending is weird, you need to use contrib/setup so that clang-format can auto-fix this for you.

plugins/elanfp/fu-elanfp-device.c
MichaelCheng04: Fixed in commit id e1fc9869b5b22a463de2ecc01a333aecad900d44.

plugins/elanfp/fu-elanfp-device.c
```diff
@@ -28,6 +28,21 @@
...
+ * Use usb control transfer.
+ */
+#define FU_ELAN_FP_DEVICE_FLAG_USB_CONTROL_TRANSFER (1 << 0)
+/**
+ * FU_ELAN_FP_DEVICE_FLAG_USB_BULK_TRANSFER:
+ *
+ * Use usb bulk transfer.
+ */
+#define FU_ELAN_FP_DEVICE_FLAG_USB_BULK_TRANSFER (1 << 1)
+
```
hughsie, line 45: why both?

plugins/elanfp/fu-elanfp-device.c
```diff
@@ -40,40 +55,72 @@ fu_elanfp_iap_send_command(FuElanfpDevice *self,
...
 {
 	GUsbDevice *usb_device = fu_usb_device_get_dev(FU_USB_DEVICE(self));
 	gsize actual = 0;
-	guint8 buftmp[61] = {request, 0};
+	gsize sendsz = bufsz + 1;
+	guint8 start_index = 0x01;
+	guint8 buftmp[64] = {request, 0};
+	gboolean is_bulk_transfer = FALSE;
+
+	if (fu_device_has_private_flag(FU_DEVICE(self), FU_ELAN_FP_DEVICE_FLAG_USB_BULK_TRANSFER)) {
```
hughsie, line 68: I think just use fu_device_has_private_flag() directly, there's no need for a is_bulk_transfer temp variable. It's really just a bitfield check of a uint64 under the covers, and it's really fast.

plugins/elanfp/fu-elanfp-device.c
MichaelCheng04: I ran the pre-commit run --hook-stage commit --all-files at local side and got all pass results. But this commit id e1fc9869b5b22a463de2ecc01a333aecad900d44 still got fail result on the item clang-format  on the web side. How to get relative fail messages to fix it.

plugins/elanfp/fu-elanfp-device.c
MichaelCheng04: Old protocol is using a control transfer to send the command, the new protocol is using bulk transfer. Did you mean we only just to set a flag to new device with new protocol?

hughsie: > Did you mean we only just to set a flag to new device with new protocol

plugins/elanfp/fu-elanfp-device.c
hughsie: if you squash the patches together (e.g. `git rebase main -i`) the CI should be happy.

plugins/elanfp/fu-elanfp-device.c
MichaelCheng04: Fixed in 7ee92a89640014327469a87cc9be9a48998b484c.

plugins/elanfp/fu-elanfp-device.c
MichaelCheng04: Fixed in 7ee92a89640014327469a87cc9be9a48998b484c.

hughsie (approved):

PR #4820, opened 2022-07-12
Wip/hughsie/fu firmware check magic

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-firmware.h
+++ b/libfwupdplugin/fu-firmware.h
@@ -62,8 +62,9 @@ struct _FuFirmwareClass {
 	gchar *(*get_checksum)(FuFirmware *self,
 			       GChecksumType csum_kind,
 			       GError **error)G_GNUC_WARN_UNUSED_RESULT;
+	gboolean (*check_magic)(FuFirmware *self, GBytes *fw, gsize offset, GError **error);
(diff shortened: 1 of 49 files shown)


libfwupdplugin/fu-firmware.h
```diff
@@ -122,6 +123,14 @@ struct _FuFirmwareClass {
...
  **/
 #define FU_FIRMWARE_FLAG_HAS_STORED_SIZE (1u << 5)
+/**
+ * FU_FIRMWARE_FLAG_SEARCH_FOR_MAGIC:
+ *
+ * Search for the magic anywhere in the firmware image.
+ *
+ * Since: 1.8.3
+ **/
+#define FU_FIRMWARE_FLAG_SEARCH_FOR_MAGIC (1u << 6)
```
superm1, line 133: I think about the fact that some firmware can be GB.  I wonder if you want to have flags to separately search header / footer / middle.

hughsie: I think this kind of searching only makes sense for small firmware. I've added a check for < 64MB and also fixed some of the failure messages to be more helpful.

PR #4822, opened 2022-07-13
Add 'action' resolution flags to each FwupdSecurityAttr

hughsie:
We tried to solve this by matching the org.fwupd.hsi AppStream ID, but
in some cases the resolution depends on what actually failed.

Add "the action the user is supposed to do" as flags so that the
front-end can translate this in the appropriate way, for instance,
using a different string for log events and HSI dialogs.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 some sanity checking please, especially on the AMD-specific ones.

plugins/acpi-dmar/fu-plugin-acpi-dmar.c
```diff
@@ -46,6 +46,7 @@ fu_plugin_acpi_dmar_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
 		return;
 	}
 	if (!fu_acpi_dmar_get_opt_in(dmar)) {
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_OEM_FW_UPDATE);
```
superm1, line 49: Not always; I've seen on Lenovo that this is fixable by BIOS config option.

plugins/acpi-facp/fu-plugin-acpi-facp.c
```diff
@@ -43,6 +43,7 @@ fu_plugin_acpi_facp_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
 	}
 	if (!fu_acpi_facp_get_s2i(facp)) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_FW);
```
superm1, line 46: Not always, not all systems offer it...

plugins/acpi-ivrs/fu-plugin-acpi-ivrs.c
```diff
@@ -48,6 +48,7 @@ fu_plugin_acpi_ivrs_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
 	}
 	if (!fu_acpi_ivrs_get_dma_remap(ivrs)) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_OEM_FW_UPDATE);
```
superm1, line 51: Same statement as the DMAR optin.  Lenovo ties it to "optimized BIOS defaults" IIRC

plugins/acpi-facp/fu-plugin-acpi-facp.c
hughsie: Ya, this would have some kind of "you may be able to" wording.

plugins/iommu/fu-plugin-iommu.c
```diff
@@ -48,6 +48,7 @@ fu_plugin_iommu_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
 
 	if (priv == NULL || !priv->has_iommu) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_FW);
```
superm1, line 51: This should be CONFIG_OS

plugins/acpi-facp/fu-plugin-acpi-facp.c
superm1: I think that you may be able to wording will need to apply to all 3 resolutions then.

plugins/iommu/fu-plugin-iommu.c
hughsie: Ahh, on the logic that it's in the blocklist?

plugins/linux-sleep/fu-plugin-linux-sleep.c
```diff
@@ -32,6 +32,7 @@ fu_plugin_linux_sleep_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attr
 	}
 	if (g_strstr_len(buf, bufsz, "[deep]") != NULL) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_ENABLED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_FW);
```
superm1, line 35: It's config OS too in some cases.  If the system offers both and you changed the default on the kernel command line or in sysfs for example.

plugins/acpi-facp/fu-plugin-acpi-facp.c
hughsie: on the logic that you might need a new OS kernel?

plugins/msr/fu-plugin-msr.c
```diff
@@ -357,6 +359,7 @@ fu_plugin_add_security_attr_amd_sme_enabled(FuPlugin *plugin, FuSecurityAttrs *a
 
 	if (!priv->amd64_syscfg.fields.sme_is_enabled) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENCRYPTED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_FW);
```
superm1, line 362: Should be contact vendor - they decide whether or not to offer SME.  TSME is changable in firmware

plugins/msr/fu-plugin-msr.c
```diff
@@ -369,6 +372,7 @@ fu_plugin_add_security_attr_amd_sme_enabled(FuPlugin *plugin, FuSecurityAttrs *a
 	if (!(fu_plugin_msr_kernel_enabled_sme(&error_local))) {
 		g_debug("%s", error_local->message);
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENCRYPTED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_FW);
```
superm1, line 375: Should be contact vendor - they decide whether or not to offer SME.  TSME is changable in firmware

plugins/pci-psp/fu-plugin-pci-psp.c
```diff
@@ -226,6 +231,7 @@ fu_plugin_add_security_attrs_rpmc(const gchar *path, FuSecurityAttrs *attrs)
 	if (!val) {
 		g_debug("no RPMC compatible SPI rom present");
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_OEM_FW_UPDATE);
```
superm1, line 234: I don't think this is something the OEM can fix.  They picked an RPMC part or they didn't.

libfwupd/fwupd-security-attr.h
```diff
@@ -47,6 +50,9 @@ typedef enum {
 	FWUPD_SECURITY_ATTR_FLAG_RUNTIME_UPDATES = 1 << 8,
 	FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ATTESTATION = 1 << 9,
 	FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE = 1 << 10,
+	FWUPD_SECURITY_ATTR_FLAG_ACTION_OEM_FW_UPDATE = 1 << 11,
```
superm1, line 53: Can you make this instead 
`FWUPD_SECURITY_ATTR_FLAG_ACTION_CONTACT_MANUFACTURER`

I think it would make it more clear.

plugins/iommu/fu-plugin-iommu.c
superm1: Yeah, you can compile the kernel without it, you can blacklist it, you can put it in soft mode.  All those things can cause this outcome.
Also - this points out to me that I don't think the check that we have for an IOMMU being present is a good enough test.  We probably should be making sure that it's actually being used for translation.  But that's a whole separate can of worms because I'm not sure the kernel exports enough on it right now.

plugins/acpi-facp/fu-plugin-acpi-facp.c
superm1: Think about how fwupd can be part of an embedded image, traditional distro, OS Tree, etc.  It will be a different solution depending on how it's packaged.  So the logic here is that you need to fix the software configuration, but you might not be able to actually accomplish this yourself and might have to contact whatever entity provides your software .

plugins/acpi-dmar/fu-plugin-acpi-dmar.c
hughsie: Ack, fixed.

plugins/msr/fu-plugin-msr.c
hughsie: So we have CONFIG_FW and OEM_FW_UPDATE?

plugins/pci-psp/fu-plugin-pci-psp.c
hughsie: Ack, thanks.

libfwupd/fwupd-security-attr.h
hughsie: FWUPD_SECURITY_ATTR_FLAG_ACTION_CONTACT_OEM makes it clear to contact the OEM rather than the vendor who's name is on the chip, e.g. Broadcom/Intel. Fixed.

plugins/linux-sleep/fu-plugin-linux-sleep.c
hughsie: Fixed.

plugins/msr/fu-plugin-msr.c
superm1: Well on second thought this is a little more complex.  SME needs OEM to set it in BIOS, but then also kernel needs to be configured to enable it.
It's more like "Fix OS" or "Contact OEM", but it's situational which one is the reason.

superm1: And this doesn't matter if TSME is enabled or available - you can just turn that on in in firmware steup yourself..

hughsie: So potentially all three actions then?

superm1: Well tsme is a separate attribute. So at least for this one it's only 2, but tsme it would be one.

hughsie: Gotcha, fixed.

plugins/msr/fu-plugin-msr.c
hughsie: Done.

plugins/msr/fu-plugin-msr.c
```diff
@@ -357,6 +359,8 @@ fu_plugin_add_security_attr_amd_sme_enabled(FuPlugin *plugin, FuSecurityAttrs *a
 
 	if (!priv->amd64_syscfg.fields.sme_is_enabled) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENCRYPTED);
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_OS);
```
superm1, line 362: I think this and the other SME one are mixed up.  If the MSR doesn't say enabled then it's contact OEM.  If the kernel doesn't say enabled it's turn on in the OS.

superm1 (approved):

Issue #4823, opened 2022-07-13
HP Dev One doesn't update firmware on restart, just boots operating system


maxamillion:
**Describe the bug**
```
$ fwupdmgr upgrade
Devices with no available firmware updates:
 • HP HD Camera
 • PC711 HFS001TDE9X073N
╔══════════════════════════════════════════════════════════════════════════════╗
║ Upgrade System Firmware from 251920384 to 251985920?                         ║
╠══════════════════════════════════════════════════════════════════════════════╣
║ This release contains the following changes:                                 ║
║                                                                              ║
║ • Correct erroneous 90b fan error reporting                                  ║
║                                                                              ║
║ Dev One Notebook PC must remain plugged into a power source for the          ║
║ duration of the update to avoid damage.                                      ║
╚══════════════════════════════════════════════════════════════════════════════╝

Perform operation? [Y|n]: Y
Downloading…             [***************************************]
Downloading…             [***************************************]
Decompressing…           [***************************************]
Authenticating…          [***************************************]
Waiting…                 [************************************** ]Successfully installed firmware
Waiting…                 [***************************************]

An update requires a reboot to complete. Restart now? [y|N]: y

```

on reboot, it just boots the operating system

**Steps to Reproduce**
Steps to reproduce the behavior.

listed above

**Expected behavior**
reboot into firmware update process

**fwupd version information**
Please provide the version of the daemon and client.

```shell
$ fwupdmgr --version
runtime   org.freedesktop.fwupd         1.8.1
runtime   org.freedesktop.fwupd-efi     1.3
compile   org.freedesktop.gusb          0.3.10
runtime   com.dell.libsmbios            2.4
runtime   org.kernel                    5.18.10-200.fc36.x86_64
compile   com.hughsie.libjcat           0.1.10
compile   org.freedesktop.fwupd         1.8.1
runtime   org.freedesktop.gusb          0.3.10
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
$ fwupdmgr get-devices --show-all
Dev One Notebook PC
│
├─Unknown Device:
│     Device ID:          f685512aa07369c9e77742acef941d779d31e766
│     GUID:               37b440a9-2473-5087-a39b-db84f32a8ed8 ← GPIO\ID_AMDI0030:00
│
├─HP HD Camera:
[...]

PR #4824, opened 2022-07-13
Do not use CoD even when advertized on non-aarch64 platforms

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/4823

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/README.md
+++ b/plugins/uefi-capsule/README.md
@@ -46,9 +46,14 @@ This plugin supports the following protocol ID:
 Described in  [UEFI specification](https://www.uefi.org/sites/default/files/resources/UEFI%20Spec%202_6.pdf)
 § 8.5.5 - Delivery of Capsules via file on Mass Storage device.
 
-If the firmware supports this, it will be the preferred method of updating. You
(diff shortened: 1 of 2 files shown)


hughsie:
I want to get confirmation from one OEM I've mailed a few hours ago before merging this; they are shipping test updates on the LVFS and I think that they *are* using CoD on x64

superm1:
> I want to get confirmation from one OEM I've mailed a few hours ago before merging this; they are shipping test updates on the LVFS and I think that they _are_ using CoD on x64

If they are indeed; can you please also ask them to try to use the traditional method too to see if it still works?
Also - please find out who is the IBV.  If need be we could try to suss out IBVs for opt-in/opt-out, but hopefully we don't go that direction.

hughsie:
> If they are indeed; can you please also ask them to try to use the traditional method too to see if it still works?

OEM has confirmed they actually don't *want* to use CoD.

superm1:
> OEM has confirmed they actually don't want to use CoD.

Can you note here who the IBV is?  We should be aware of this for if/when we try to enable it again on x86.

hughsie:
I'll ask.

PR #4831, opened 2022-07-14
Wip/hughsie/hsi

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Wait a minute. Won't this break AMD security? Haven't tried yet but would think it would.

hughsie:
> Won't this break AMD security

We still do the INTEL check, no? I can check on AMD in a few mins if that helps.

superm1:
> > Won't this break AMD security
> 
> We still do the INTEL check, no? I can check on AMD in a few mins if that helps.

What system do you have? Does it support the AMD stuff?

hughsie:
Lenovo T14, so no.

superm1:
OK just double checked it on a Z13.  It's fine, no worries.

PR #4834, opened 2022-07-15
Allow loading in test HSI profiles for debugging

hughsie:
This allows us to load sets of different host security attributes
for testing the various front end tools we have now. e.g.

    sudo FWUPD_TEST_PROFILE=thinkpad-p1-iommu.json.gz fwupd

or, using a non-compressed absolute path:

    sudo FWUPD_TEST_PROFILE=/tmp/test/thinkpad-p1-iommu.json fwupd

Test profiles can be created with `fwupdmgr security --json` and
manually modified if required.

See https://github.com/fwupd/fwupd/discussions/4832

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/contrib/minimize-test-profile.py
+++ b/contrib/minimize-test-profile.py
@@ -1,0 +1,33 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2017 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+#
+# pylint: disable=invalid-name,missing-docstring
+
+import json
+import sys
+
+from typing import Dict, List, Any
+
+if len(sys.argv) < 2:
+    print("specify filenames")
+    sys.exit(1)
+
+for fn in sys.argv[1:]:
+    with open(fn, "rb") as f:
+        attrs = json.loads(f.read().decode())
+    new_attrs: List[Dict[str, Any]] = []
+    for attr in attrs["SecurityAttributes"]:
+        new_attr: Dict[str, Any] = {}
+        for key in attr:
+            if key in ["AppstreamId", "HsiResult", "HsiLevel", "Flags", "Plugin"]:
+                new_attr[key] = attr[key]
+        new_attrs.append(new_attr)
+    with open(fn, "wb") as f:
+        f.write(
+            json.dumps(
+                {"SecurityAttributes": new_attrs}, indent=2, separators=(",", " : ")
+            ).encode()
+        )
--- a/docs/env.md
+++ b/docs/env.md
@@ -24,6 +24,7 @@ with a non-standard filesystem layout.
 ## daemon
 
 * `FWUPD_MACHINE_KIND` can be used to override the detected machine type, e.g. `physical`, `virtual`, or `container`
+* `FWUPD_TEST_PROFILE` can be used to load test data from `/usr/share/fwupd/test-profile.d`, e.g. `thinkpad-p1-no-iommu.json.gz`
 
 ## Self Tests
 
--- a/src/tests/test-profile/meson.build
+++ b/src/tests/test-profile/meson.build
@@ -1,0 +1,16 @@
+if build_standalone
+  gzip = find_program('gzip')
+  foreach input_file : [
+    'thinkpad-p1-no-iommu.json',
+    'thinkpad-p1-iommu.json',
+  ]
+    custom_target(input_file,
+      input: input_file,
+      output: '@0@.gz'.format(input_file),
+      capture: true,
+      command: [gzip, '--keep', '--stdout', '@INPUT@'],
+      install: true,
+      install_dir: join_paths(datadir, 'fwupd', 'test-profile.d'),
+    )
+  endforeach
+endif
--- a/src/tests/test-profile/thinkpad-p1-no-iommu.json
+++ b/src/tests/test-profile/thinkpad-p1-no-iommu.json
@@ -1,0 +1,272 @@
+{
+  "SecurityAttributes" : [
+    {
+      "AppstreamId" : "org.fwupd.hsi.Kernel.Tainted",
+      "HsiLevel" : 0,
+      "HsiResult" : "not-tainted",
+      "Plugin" : "linux_tainted",
+      "Flags" : [
+        "success",
+        "runtime-issue"
+      ]
+    },
+    {
+      "AppstreamId" : "org.fwupd.hsi.Kernel.Lockdown",
+      "HsiLevel" : 0,
+      "HsiResult" : "enabled",
+      "Plugin" : "linux_lockdown",
+      "Flags" : [
+        "success",
+        "runtime-issue"
+      ]
+    },
+    {
+      "AppstreamId" : "org.fwupd.hsi.Kernel.Swap",
+      "HsiLevel" : 0,
+      "HsiResult" : "encrypted",
+      "Plugin" : "linux_swap",
+      "Flags" : [
+        "success",
+        "runtime-issue"
+      ]
+    },
+    {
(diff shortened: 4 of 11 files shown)


src/tests/test-profile/meson.build
```diff
@@ -0,0 +1,8 @@
+if build_standalone
```
superm1, line 1: Maybe it's better to store these at https://github.com/fwupd/fwupd-test-firmware so they don't end up packaged and more samples can be added out of band

src/tests/test-profile/meson.build
```diff
@@ -0,0 +1,8 @@
...
+  install_data([
+      'thinkpad-p1-no-iommu.json.gz',
+      'thinkpad-p1-iommu.json.gz',
```
superm1, line 4: whichever tree they end up would much rather store in tree uncompressed so that we can treat them as text and see differences if test data needs to be refreshed from new plugins and such.

Even if they're installed into a package, they can be compressed at build time instead.

src/tests/test-profile/meson.build
hughsie: My initial thought is that's a good idea. I'm then worried about the implied "API" we have then; i.e. if we change attr AppStreamIDs or HSI levels.

src/tests/test-profile/meson.build
hughsie: Yes, that's a good idea. We also need that in the "rename an ID" case too.

docs/env.md
```diff
@@ -24,6 +24,7 @@ with a non-standard filesystem layout.
 ## daemon
 
 * `FWUPD_MACHINE_KIND` can be used to override the detected machine type, e.g. `physical`, `virtual`, or `container`
+* `FWUPD_TEST_PROFILE` can be used to load test data from `/usr/share/fwupd/test-profile.d`, e.g. `thinkpad-p1-no-iommu.json.gz`
```
superm1, line 27: This is begging to apply into unit tests for a variety of machines!  If we store it at fwupd-test-firmware CI already clones that and can be running the battery of tests to make sure that machines security data doesn't regress different levels.

src/tests/test-profile/meson.build
superm1: See my other comment; unit tests.  Then we can make sure that calculated value matches the given fwupd version.

docs/env.md
hughsie: Hilariously we have all that data already; i.e. all the HSI uploads to the LVFS.

superm1:
Once this concept is landed, we should encourage contribution of more test samples for unit tests too somewhere in documentation.  The more "in the wild" hardware that this information is captured the better.

docs/env.md
superm1: Ha. that's right!

src/tests/test-profile/meson.build
hughsie: Lets do this as phase 2 -- what I'm not convinced about is the *name* -- profile is what we use for "how long it takes" and so we need another name I think. Maybe `emulate`? i.e. `FWUPD_DEVICE_EMULATE` or `FWUPD_EMULATE` then the json files could be in `/usr/share/fwupd/emulate.d` -- other ideas welcome.

src/tests/test-profile/meson.build
```diff
@@ -0,0 +1,16 @@
...
+  foreach input_file : [
+    'thinkpad-p1-no-iommu.json',
+    'thinkpad-p1-iommu.json',
+  ]
+    custom_target(input_file,
+      input: input_file,
+      output: '@0@.gz'.format(input_file),
+      capture: true,
+      command: [gzip, '--keep', '--stdout', '@INPUT@'],
+      install: true,
```
superm1, line 12: I'm still not convinced we want this to be installed by default, especially as we add more data for other systems and it grows.

contrib/minimize-test-profile.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2017 Richard Hughes <richard@hughsie.com>
+#
```
superm1, line 4: I think in this header need to provide instruction on how this is intended to be used;

IE `fwupdtool security --force --json | tee ./contrib/minimize-test-profile.py > foo.json`

Another useful thing could be to add a capture option if nothing on stdin/arg1 that just calls `fwupdtool security --force --json` and does everything for you and spits out a file.

src/tests/test-profile/meson.build
hughsie: Maybe one of them? I wanted something I can reference in the commit and maybe the docs.

src/tests/test-profile/thinkpad-p1-no-iommu.json
```diff
@@ -0,0 +1,272 @@
...
+      "HsiResult" : "encrypted",
+      "Plugin" : "linux_swap",
+      "Flags" : [
+        "success",
+        "runtime-issue"
+      ]
+    },
+    {
+      "AppstreamId" : "org.fwupd.hsi.Uefi.SecureBoot",
+      "HsiLevel" : 0,
```
superm1, line 35: I think you should strip HsiLevel in this data, let fwupd calculate it by mapping HsiLevel over to AppstreamId internally.  This also lets us re-define which levels things should be tied to based on fwupd version.

hughsie: Okay, lets do this bit first; PR incoming.

hughsie: https://github.com/fwupd/fwupd/pull/4835

src/tests/test-profile/meson.build
hughsie: FWUPD_HOST_EMULATE might work. or FWUPD_EMULATE_HOST

hughsie: we've got FWUPD_SOLOKEY_EMULATE so maybe the former?

contrib/minimize-test-profile.py
hughsie: JsonBuilder can't remove nodes, so it's not trivial to do in C. Maybe we add a fwupd_security_attr_to_json_full() with some flags? I do like the stdout pipe thing.

src/tests/test-profile/meson.build
superm1: How about `FWUPD_HSI_EMULATE`

superm1: BTW I think you need to make sure the daemon can't upload data while it's emulating and also there should be a flag in place to say data is emulated that GUI or command line clients can use to warn it's under emulation.

superm1: I added something for that, take a look.  The other option is to taint the daemon.  I can see arguments both ways.

hughsie: Ya, we need the flag, and we'll need a fwupd_device flag for the same purposes too. Thanks.

hughsie:
Alternate proposal in https://github.com/fwupd/fwupd/pull/4840 -- this one has the ability to load *devices* too, which makes testing gnome-firmware so much easier.

contrib/minimize-test-profile.py
hughsie: I did the pipe thing. I think pushing this early feature into the C or CLI API is a bit early.

src/tests/test-profile/meson.build
hughsie: I did a new attribute; I think that's semantically most pleasing, and also means the attribute flags are the same for emulation and non-emulation.

PR #4836, opened 2022-07-15
WIP: implement bios updates for supermicro

flanfly:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

WIP and untested


plugins/redfish/fu-redfish-multipart-device.c
```diff
@@ -19,26 +19,72 @@ G_DEFINE_TYPE(FuRedfishMultipartDevice, fu_redfish_multipart_device, FU_TYPE_RED
...
+	json_builder_set_member_name(builder, "@Redfish.OperationApplyTime");
+	json_builder_add_string_value(builder, "Immediate");
+	json_builder_end_object(builder);
+}
 
-	/* create header */
+static void
+fu_redfish_multipart_device_get_parameters_smc(FuRedfishMultipartDevice *self, JsonBuilder *builder)
+{
+	/* https://supermicro.com/manuals/other/RedishRefGuide.pdf */
```
hughsie, line 42: My gut instinct here is that /redfish/v1/UpdateService/Actions/UpdateService.StartUpdate is sufficiently different that you want to create a `FuRedfishSupermicroDevice` device type and put all the differences there.

flanfly:
- [x] Separate device subclass for Supermicro.
- [x] Check whether we have the license and add a problem to the device if not.
- [ ] Check for pending/stuck update.

libfwupdplugin/fu-device.c
```diff
@@ -2711,6 +2711,8 @@ fu_device_problem_to_inhibit_reason(FuDevice *self, guint64 device_problem)
 		return g_strdup("Device requires AC power to be connected");
 	if (device_problem == FWUPD_DEVICE_PROBLEM_LID_IS_CLOSED)
 		return g_strdup("Device cannot be used while the lid is closed");
+	if (device_problem == FWUPD_DEVICE_PROBLEM_MISSING_LICENSE)
+		return g_strdup("Device does not have the necessary license installed");
```
hughsie, line 2715: I think you need a translated version in src/fu-util-common.c too.

libfwupd/fwupd-enums.h
```diff
@@ -597,6 +597,14 @@ typedef guint64 FwupdDeviceFlags;
  * Since 1.8.1
  */
 #define FWUPD_DEVICE_PROBLEM_LID_IS_CLOSED (1u << 5)
+/**
+ * FWUPD_DEVICE_PROBLEM_MISSING_LICENSE:
+ *
+ * The device cannot be updated due to missing vendor's license.
+ *
+ * Since TODO
```
hughsie, line 605: Stick `1.8.5` here; we can change it later if needed.

plugins/redfish/fu-redfish-backend.c
```diff
@@ -16,6 +16,7 @@
 #include "fu-redfish-multipart-device.h"
 #include "fu-redfish-request.h"
 #include "fu-redfish-smbios.h"
+#include "fu-redfish-smc.h"
```
hughsie, line 19: `fu-redfish-smc-device.h` perhaps -- or did you forget to `git add` a header?

plugins/redfish/fu-redfish-backend.c
```diff
@@ -250,7 +267,15 @@ fu_redfish_backend_coldplug(FuBackend *backend, FuProgress *progress, GError **e
 			return FALSE;
 		}
 	}
-	if (json_object_has_member(json_obj, "MultipartHttpPushUri")) {
+	if (self->push_uri_path == NULL) {
```
hughsie, line 270: does Supermicro have `MultipartHttpPushUri` too? If not it might be good to move this to the last case, so that we start using the standard device types if Supermicro ever exports *both*.

plugins/redfish/fu-redfish-device.c
```diff
@@ -360,6 +360,27 @@ fu_redfish_device_set_vendor(FuRedfishDevice *self, const gchar *vendor)
...
+	if (fu_redfish_request_perform(request,
+				       fu_redfish_backend_get_push_uri_path(backend),
+				       FU_REDFISH_REQUEST_PERFORM_FLAG_LOAD_JSON,
+				       &error_local))
+		return;
+
+	if (error_local->code == FWUPD_ERROR_NOT_SUPPORTED)
+		fu_device_add_problem(device, FWUPD_DEVICE_PROBLEM_MISSING_LICENSE);
+
+	return;
```
hughsie, line 381: it's more idiomatic to have `if (!fu_redfish_request_perform(` and then check the error with `g_error_matches`

plugins/redfish/fu-redfish-device.c
```diff
@@ -498,12 +519,22 @@ fu_redfish_device_probe(FuDevice *dev, GError **error)
 			if (json_object_has_member(related_item, "@odata.id")) {
 				const gchar *id =
 				    json_object_get_string_member(related_item, "@odata.id");
-				if (!fu_redfish_device_probe_related_item(self, id, error))
-					return FALSE;
+				g_autoptr(GError) error_local = NULL;
+				if (!fu_redfish_device_probe_related_item(self, id, &error_local)) {
+					if (error_local->code != FWUPD_ERROR_NOT_SUPPORTED) {
```
hughsie, line 524: !g_error_matches

plugins/redfish/fu-redfish-device.c
```diff
@@ -498,12 +519,22 @@ fu_redfish_device_probe(FuDevice *dev, GError **error)
...
+						*error = g_error_copy(error_local);
+						return FALSE;
+					} else {
```
hughsie, line 527: no need for the }else{ if you're `return`ing.

plugins/redfish/fu-redfish-smc-device.c
```diff
@@ -0,0 +1,217 @@
...
+
+static const gchar *
+fu_redfish_smc_device_get_task(JsonObject *json_obj)
+{
+	JsonObject *tmp_obj;
+	JsonArray *tmp_ary;
+	const gchar *msgid;
+
+	if (!json_object_has_member(json_obj, "Accepted"))
+		return FALSE;
```
hughsie, line 27: needs to be NULL if it's const gchar *

plugins/redfish/fu-redfish-smc-device.c
```diff
@@ -0,0 +1,217 @@
...
+
+	if (!fu_redfish_request_perform(
+		request,
+    fu_redfish_backend_get_push_uri_path(backend),
+		FU_REDFISH_REQUEST_PERFORM_FLAG_LOAD_JSON,
+		error))
+		return FALSE;
+	json_obj = fu_redfish_request_get_json_object(request);
+
+	if (!(location = fu_redfish_smc_device_get_task(json_obj))) {
```
hughsie, line 110: i think spitting this up would be much more readable; e.g. 

    location = fu_redfish_smc_device_get_task(json_obj)
    if (location == NULL) {

plugins/redfish/fu-redfish-smc-device.c
```diff
@@ -0,0 +1,217 @@
...
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INVALID_FILE,
+			    "failed to upload: %li",
+			    fu_redfish_request_get_status_code(request));
+		return FALSE;
+	}
+	json_obj = fu_redfish_request_get_json_object(request);
+
+	/* poll the verify task for progress */
+	if (!(location = fu_redfish_smc_device_get_task(json_obj))) {
```
hughsie, line 181: same

plugins/redfish/fu-redfish-backend.c
flanfly: It does. No idea if it works, though. Their documentation doesn't mention it.

plugins/redfish/fu-redfish-backend.c
flanfly: The former

plugins/redfish/fu-redfish-backend.c
hughsie: Okay, maybe just add a comment to this effect; i.e. why it's listed first please

plugins/redfish/fu-redfish-device.c
```diff
@@ -360,6 +360,25 @@ fu_redfish_device_set_vendor(FuRedfishDevice *self, const gchar *vendor)
...
+	g_autoptr(FuRedfishRequest) request = fu_redfish_backend_request_new(backend);
+	g_autoptr(GError) error_local = NULL;
+
+	/* see if we don't get an license error */
+	if (!fu_redfish_request_perform(request,
+					fu_redfish_backend_get_push_uri_path(backend),
+					FU_REDFISH_REQUEST_PERFORM_FLAG_LOAD_JSON,
+					&error_local))
+		if (g_error_matches(error_local, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED))
+			fu_device_add_problem(device, FWUPD_DEVICE_PROBLEM_MISSING_LICENSE);
```
hughsie, line 377: {} here please

plugins/redfish/fu-redfish-device.c
```diff
@@ -498,12 +517,23 @@ fu_redfish_device_probe(FuDevice *dev, GError **error)
...
 				const gchar *id =
 				    json_object_get_string_member(related_item, "@odata.id");
-				if (!fu_redfish_device_probe_related_item(self, id, error))
-					return FALSE;
+				g_autoptr(GError) error_local = NULL;
+				if (!fu_redfish_device_probe_related_item(self, id, &error_local)) {
+					if (!g_error_matches(error_local,
+							     FWUPD_ERROR,
+							     FWUPD_ERROR_NOT_SUPPORTED)) {
+						*error = g_error_copy(error_local);
```
hughsie, line 525: rather, please do (the slightly insane) `g_propagate_error(error, g_steal_pointer(&error_local))` -- then it doesn't explode if @error is ever NULL.

plugins/redfish/fu-redfish-smc-device.c
```diff
@@ -0,0 +1,217 @@
...
+
+	return fu_redfish_smc_device_start_update(device, progress, error);
+}
+
+static void
+fu_redfish_smc_device_set_progress(FuDevice *self, FuProgress *progress)
+{
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 50, "write");
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_RESTART, 50, "restart");
```
hughsie, line 202: I think you need 4 things here, no? don't you get a runtime warning?

flanfly: Before I got something like "error: progress reset to 0". Does the FuDevice code set the status or is it just my code?

hughsie: I think the docs are https://fwupd.github.io/libfwupdplugin/tutorial.html#set_progress -- but the redfish plugin is somewhat special as it wants to set the progress and status on it's own using the polling interface. Can you share a complete daemon log?

plugins/redfish/fu-redfish-backend.c
flanfly: Ok, I messed it up. Supermicro also uses `MultipartHttpPushUri` but needs an additional POST to `UpdateService.StartUpdate` to do the actual update. I'll update the code.

hughsie: Thanks!

plugins/redfish/fu-redfish-smc-device.c
flanfly: [log.txt](https://github.com/fwupd/fwupd/files/9234178/log.txt)

hughsie: Another drive-by comment:

>   Guid:                 cea310e9-acad-5573-98c3-3f0322bc1886 <- REDFISH\VENDOR_SMCI&ID_BIOS

That's super generic. Does SuperMicro have any kind of ID that identifies the firmware *stream* -- i.e. is specific to this model and device?

hughsie: Can you do the same log again please, but this time with `FWUPD_REDFISH_VERBOSE=1` set pls.

flanfly: AFAIK all updates are mainboard specific and they don't have different stream/trains/LTS versions. This all the BMC gives me about the current version:
```
{
  "@odata.type": "#SoftwareInventory.v1_3_0.SoftwareInventory",
  "@odata.id": "/redfish/v1/UpdateService/FirmwareInventory/BIOS",
  "Id": "BIOS",
  "Name": "BIOS",
  "Updateable": true,
  "Version": "BIOS Date: 04/14/2022 Ver 2.4",
  "SoftwareId": "BIOS Date: 04/14/2022 Ver 2.4",
  "Status": {
    "State": "Enabled",
    "Health": "OK"
  },
  "RelatedItem": [
    {
      "@odata.id": "/redfish/v1/Systems/1/Bios"
    }
  ],
  "Actions": {
    "Oem": {}
  },
  "Oem": {},
  "Manufacturer": "SMCI",
  "ReleaseDate": "2022-04-14T00:00:00Z",
  "WriteProtected": false
}
```
The `/redfish/v1/Systems/1/Bios` route gives me a license error.

hughsie: Ha,

    "Version": "BIOS Date: 04/14/2022 Ver 2.4",
    "SoftwareId": "BIOS Date: 04/14/2022 Ver 2.4",

the latter isn't a software ID. I wonder if they copy&pasted something wrong here :) 

I guess we should use some SMBIOS data; what in `sudo fwupdtool hwids` makes sense? Perhaps `REDFISH\VENDOR_SMCI&ID_$ProductName$`?

hughsie: So maybe it isn't a huge problem if we 100% use a hwid to handle this. We can fix async to this feature, other Redfish stuff is going to be affected.

hughsie: tl;dr: not your problem

flanfly: [log-verbose-redfish.txt](https://github.com/fwupd/fwupd/files/9235858/log-verbose-redfish.txt)

flanfly: Ah, I see. There are two tasks that are polled. First the upload, then the actual update (`fu_redfisk_smc_device_start_update()`) maybe create a child progress for both?

hughsie: Yes, if you're the one that's doing the actions. Thanks!

superm1 (commented):
Perhaps do you want to model how the license checking should work in the redfish self tests?

hughsie:
Also, clang-format has words: you used contrib/setup, right?

flanfly:
- [ ] Add test case for license check
- [ ] Check for pending/stuck update
- [ ] Fix formatting

flanfly:
Ok, after rebasing onto `main` the redfish plugin isn't there anymore even if I enable it with `-Dplugin_redfish=enabled`. The BMC device doesn't show up anymore and `fwupdtool get-plugins` doesn't list the redfish plugin. What Am I missing?

hughsie:
> What Am I missing

Got a build log?

flanfly:
```
meson build -Dintrospection=disabled \
  -D{libxmlb:,libjcat:,gcab:}introspection=false \
  -Dgcab:docs=false -Dlibxmlb:gtkdoc=false \
  -Dplugin_redfish=enabled -Dplugin_tpm=enabled \
  -Dgudev=enabled -Dcurl=enabled -Dgusb=enabled \
  --prefix /opt/fwupd
```
[meson-log.txt](https://github.com/fwupd/fwupd/files/9683693/meson-log.txt)

```
ninja -C build && sudo ninja -C build install
```
[ninja-build.log](https://github.com/fwupd/fwupd/files/9683694/ninja-build.log)
[ninja-install.log](https://github.com/fwupd/fwupd/files/9683696/ninja-install.log)

flanfly:
@superm1 
>Perhaps do you want to model how the license checking should work in the redfish self tests?

I tried to write some tests for the whole SMC specific logic but I hit a roadblock when extending `redfish.py`. The backend decides what device subclass to use based on 
 the response of `/v1/redfish/UpdateService`. Changing the response breaks the existing tests and I can't have two update services either.

hughsie:
> Changing the response breaks the existing tests

Can we change the response to v1/redfish/UpdateService based on some heuristic -- i.e. if password is password2 use normal redfish and if password3 then use the SMC version?

flanfly:
Sounds doable. Do I add this as another plugin in `fu_test_self_init()`?

hughsie:
Hmm, it's set in redfish.conf, so you might have to call into the redfish backend directly to change the password directly.

libfwupd/fwupd-enums.h
```diff
@@ -607,6 +607,14 @@ typedef guint64 FwupdDeviceFlags;
  * Since 1.8.3
  */
 #define FWUPD_DEVICE_PROBLEM_IS_EMULATED (1u << 6)
+/**
+ * FWUPD_DEVICE_PROBLEM_MISSING_LICENSE:
+ *
+ * The device cannot be updated due to missing vendor's license.
+ *
+ * Since 1.8.5
```
superm1, line 615: 1.8.6 now

plugins/redfish/fu-redfish-device.c
```diff
@@ -360,6 +360,26 @@ fu_redfish_device_set_vendor(FuRedfishDevice *self, const gchar *vendor)
...
+	g_autoptr(FuRedfishRequest) request = fu_redfish_backend_request_new(backend);
+	g_autoptr(GError) error_local = NULL;
+
+	/* see if we don't get an license error */
+	if (!fu_redfish_request_perform(request,
+					fu_redfish_backend_get_push_uri_path(backend),
+					FU_REDFISH_REQUEST_PERFORM_FLAG_LOAD_JSON,
+					&error_local)) {
+		if (g_error_matches(error_local, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED))
+			fu_device_add_problem(device, FWUPD_DEVICE_PROBLEM_MISSING_LICENSE);
```
superm1, line 377: For debugging purpose would it make sense to output the error the debug log if it _doesn't match_ here?

plugins/redfish/fu-redfish-device.c
```diff
@@ -498,12 +518,24 @@ fu_redfish_device_probe(FuDevice *dev, GError **error)
...
+						return FALSE;
+					}
+					fu_device_set_logical_id(dev, id);
+				}
 			}
 		}
 	}
 
+	/* for Supermicro check whether we have a proper Redfish license installed */
+	fu_redfish_backend_smc_license_check(dev);
```
superm1, line 537: shouldn't this only run if using the SMC type of device?

plugins/redfish/fu-redfish-plugin.c
```diff
@@ -126,6 +126,16 @@ fu_redfish_plugin_coldplug(FuPlugin *plugin, FuProgress *progress, GError **erro
 	return TRUE;
 }
 
+gboolean
+fu_redfish_plugin_set_credentials(FuPlugin *plugin, const gchar *username, const gchar *password)
```
superm1, line 130: you may as well make this void I think

src/fu-util-common.c
```diff
@@ -1374,6 +1374,10 @@ fu_util_device_problem_to_string(FwupdClient *client, FwupdDevice *dev, FwupdDev
 		/* TRANSLATORS: emulated means we are pretending to be a different model */
 		return g_strdup(_("Device is emulated"));
 	}
+	if (problem == FWUPD_DEVICE_PROBLEM_MISSING_LICENSE) {
+		/* TRANSLATORS: the update function is an add-on feature sold separately" */
+		return g_strdup(_("Device requires a software license to update"));
```
superm1, line 1379: This string is different than the other problem string in the PR.  Shouldn't they be the same?

superm1 (commented):
As an up-leveled question - how does a user resolve this?  Does it need some discoverable documentation?  Perhaps would it make sense to write up a markdown page that goes to fwupd.github.io and when the problem shows up point people to that page?

flanfly:
> As an up-leveled question - how does a user resolve this? Does it need some discoverable documentation? Perhaps would it make sense to write up a markdown page that goes to fwupd.github.io and when the problem shows up point people to that page?

The missing license? Sure. What is the process? Do I open a PR on the Gitlab project?

superm1:
> > As an up-leveled question - how does a user resolve this? Does it need some discoverable documentation? Perhaps would it make sense to write up a markdown page that goes to fwupd.github.io and when the problem shows up point people to that page?
> 
> The missing license? Sure. What is the process? Do I open a PR on the Gitlab project?

You can roll it into this PR or do a separate one after this merges.  Basically:
1. Add a markdown document to `docs/`
2. Reference the doc from `docs/index.html`
3. Add it to this [array](https://github.com/fwupd/fwupd/blob/main/docs/fwupdplugin.toml.in#L44) so that it gets rewritten into HTML
4. Modify the client so that when this problem is displayed there is a message saying to go there for more information.

plugins/redfish/fu-redfish-plugin.c
superm1: Ah; you're right.  I think it should still be void though, and you should probably leave a comment that it's only intended for self tests.

plugins/redfish/fu-redfish-device.c
```diff
@@ -504,6 +526,10 @@ fu_redfish_device_probe(FuDevice *dev, GError **error)
 		}
 	}
 
+	/* for Supermicro check whether we have a proper Redfish license installed */
+	if (g_strcmp0("SMCI", fu_device_get_vendor(dev)) == 0)
```
superm1, line 530: do you ever OEM systems that this wouldn't be valid?  If not, then this is probably fine.  If you do, I would wonder if it's better to set some metadata or check the device gtype.

superm1 (approved):
LGTM, thanks.

plugins/redfish/fu-redfish-device.c
```diff
@@ -360,6 +360,28 @@ fu_redfish_device_set_vendor(FuRedfishDevice *self, const gchar *vendor)
...
+					fu_redfish_backend_get_push_uri_path(backend),
+					FU_REDFISH_REQUEST_PERFORM_FLAG_LOAD_JSON,
+					&error_local)) {
+		if (g_error_matches(error_local, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED))
+			fu_device_add_problem(device, FWUPD_DEVICE_PROBLEM_MISSING_LICENSE);
+		else
+			g_debug("supermicro license check returned %s\n", error_local->message);
+	}
+
+	return;
```
hughsie, line 382: no need for this return

plugins/redfish/fu-self-test.c
```diff
@@ -36,17 +39,50 @@ fu_test_self_init(FuTest *self)
...
+		error = NULL;
+	} else {
+		g_assert_no_error(error);
+		g_assert_true(ret);
+		ret = fu_plugin_runner_coldplug(self->plugin, progress, &error);
+		g_assert_no_error(error);
+		g_assert_true(ret);
+	}
+
+	// supermicro BMC
```
hughsie, line 56: /* style please */

plugins/redfish/fu-self-test.c
```diff
@@ -36,17 +39,50 @@ fu_test_self_init(FuTest *self)
...
+	} else {
+		g_assert_no_error(error);
+		g_assert_true(ret);
+		fu_redfish_plugin_set_credentials(self->smc_plugin, "smc_username", "password2");
+		ret = fu_plugin_runner_coldplug(self->smc_plugin, progress, &error);
+		g_assert_no_error(error);
+		g_assert_true(ret);
+	}
+
+	// unlicensed supermicro BMC
```
hughsie, line 71: same

plugins/redfish/fu-self-test.c
```diff
@@ -36,17 +39,50 @@ fu_test_self_init(FuTest *self)
 	g_assert_no_error(error);
 	g_assert_true(ret);
 
+	// generic BMC
```
hughsie, line 42: same

plugins/redfish/fu-self-test.c
```diff
@@ -36,17 +39,50 @@ fu_test_self_init(FuTest *self)
...
 	self->plugin = fu_plugin_new_from_gtype(fu_redfish_plugin_get_type(), ctx);
 	ret = fu_plugin_runner_startup(self->plugin, progress, &error);
 	if (g_error_matches(error, FWUPD_ERROR, FWUPD_ERROR_INVALID_FILE)) {
 		g_test_skip("no redfish.py running");
-		return;
+		error = NULL;
```
hughsie, line 47: you need g_clear_error here, no?

PR #4840, opened 2022-07-17
Wip/hughsie/host emulate

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


contrib/generate-emulation.py
```diff
@@ -0,0 +1,51 @@
...
+    except KeyError:
+        pass
+    return json.dumps(
+        {"SecurityAttributes": new_attrs, "Devices": new_devices},
+        indent=2,
+        separators=(",", " : "),
+    )
+
+
+if len(sys.argv) < 2:
```
superm1, line 44: I have an idea here... How about you use the client API with gobject introspection to fetch the data instead of piping or arguments? Then this script can be "totally standalone".

I'm thinking a bit bigger...
I could see us installing it "with" fwupd and then when we want to reproduce any problem not specific to actual writing we can ask someone to run it and give us the json file it spits out.

hughsie: Can that be phase 2? There are some other API changes we'd need to do (e.g. dropping the _`Protocol`, but `Protocols` if more than 1 value_ compatibility hacks to do this in a clean way.

src/fu-engine.c
```diff
@@ -6194,19 +6196,146 @@ fu_engine_record_security_attrs(FuEngine *self, GError **error)
...
 {
 #ifdef HAVE_HSI
 	GPtrArray *plugins = fu_plugin_list_get_all(self->plugin_list);
 	g_autoptr(GPtrArray) devices = fu_device_list_get_all(self->device_list);
-	g_autoptr(GPtrArray) items = NULL;
 	g_autoptr(GError) error = NULL;
 
 	/* already valid */
-	if (self->host_security_id != NULL)
+	if (self->host_security_id != NULL || self->host_emulation)
```
superm1, line 6338: I'm not convinced you want to avoid all this.  Don't you want the engine to do depsolve on all the attributes?  The way that i'm thinking about emulation is that you skip certain vfuncs in plugins that probe the device or probe the attributes and that you avoid letting you call write vfuncs.  So you don't avoid  as much engine functionality tied to all of the corner case behavior.

hughsie: Depsolve should be being done already. I'd like to keep it simple for now, and either have a 100% emulated host or a 100% real host -- otherwise it gets really complicated fast with things like custom FuDevice GTypes.

contrib/generate-emulation.py
superm1: Yeah; then it's trivial changes to this script as the daemon revs and does other API changes.

hughsie: Added.

src/fu-util.c
```diff
@@ -3458,6 +3459,18 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 			g_print("%s", estr);
 	}
 
+	/* host emulation */
+	for (guint j = 0; j < attrs->len; j++) {
+		FwupdSecurityAttr *attr = g_ptr_array_index(attrs, j);
+		if (g_strcmp0(fwupd_security_attr_get_appstream_id(attr),
+			      FWUPD_SECURITY_ATTR_ID_HOST_EMULATION) == 0) {
+			host_emulation = TRUE;
```
superm1, line 3467: Why not just use `priv->no_unreported_check` instead of a new variable

superm1 (dismissed):
just a few nits.  I also think JSON dumps shouldn't be included here but should be at test-firmware repo, but otherwise OK.

src/fu-util.c
hughsie: Ohh yes, good idea. Fixed.

src/fu-util-common.c
```diff
@@ -1368,6 +1368,10 @@ fu_util_device_problem_to_string(FwupdClient *client, FwupdDevice *dev, FwupdDev
 		/* TRANSLATORS: lid means "laptop top cover" */
 		return g_strdup(_("Device cannot be used while the lid is closed"));
 	}
+	if (problem == FWUPD_DEVICE_PROBLEM_IS_EMULATED) {
+		/* TRANSLATORS: emulated means we are pretending to be a diffent model */
```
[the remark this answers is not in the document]
hughsie: Fixed, ta

hughsie:
>  I also think JSON dumps shouldn't be included here

I've deleted all but one; I think we need one to showcase the feature easily.

superm1 (approved):

PR #4850, opened 2022-07-20
Wait for the system to acquiesce after doing each update

hughsie:
We want to allow all the device hotplug events to be processed before
marking the update as completed. Otherwise, we might have a situation
where we have a child device attached to a parent, where we want to
update the parent, then the child. e.g.

 1. Add parent
 2. Add child
 3. Update parent
 4. Attach parent
 5. Wait for parent

...some time passes...

 6. Parent re-appears
 7. Update finishes, client indicates success

...child update is scheduled...
...which returns with failure as it does not exist...

 8. Add child

The child should have been added *before* the update completed to avoid
the caller from needing an unspecified delay as a *workaround*.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -235,6 +235,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "auto-pause-polling";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_ONLY_WAIT_FOR_REPLUG)
 		return "only-wait-for-replug";
+	if (flag == FU_DEVICE_INTERNAL_FLAG_REQUIRES_ACQUIESCE)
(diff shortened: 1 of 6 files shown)


d4s:
Definitely helpful for issues detection!
tested with steelseries mouse

superm1 (commented):
Why not set this as a general behavior?  It seems like it should probably be safer overall wouldn't it be?

hughsie:
>  It seems like it should probably be safer overall wouldn't it be

I've set it for all USB-based devices -- but it adds 2500ms to all updates otherwise -- and in the case of fingerprint devices that's 100x the actual update time. Plus for things like DBX updates it's not useful at all and a bit confusing.

superm1 (approved):

PR #4853, opened 2022-07-20
fu-smbios: Fix parsing smbios data

superm1:
When support was added for falling back to SMBIOS data from the kernel
in /sys/class/dmi, we inadvertently stopped caring about the data parsed
directly from DMI tables as first priority.  This caused a regression in
hwids from some OEMs that relied upon IDs that could only be properly built
from DMI tables, not the kernel /sys/class/dmi interface.

Link: https://bugs.launchpad.net/ubuntu/+source/fwupd/+bug/1982103
Fixes: 464425fb5 ("SMBIOS: try reading from /sys/class/dmi if direct access fails")
Signed-off-by: Mario Limonciello <mario.limonciello@amd.com>

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-smbios.c
+++ b/libfwupdplugin/fu-smbios.c
@@ -869,7 +869,8 @@ fu_smbios_setup(FuSmbios *self, GError **error)
 				return FALSE;
 			}
 			g_debug("ignoring %s", error_local->message);
-		}
+		} else
+			return TRUE;
 	}
 
 	/* the values the kernel parsed; these are world-readable */


libfwupdplugin/fu-smbios.c
```diff
@@ -863,13 +863,13 @@ fu_smbios_setup(FuSmbios *self, GError **error)
...
 	if (g_file_test(path, G_FILE_TEST_EXISTS)) {
 		g_autoptr(GError) error_local = NULL;
-		if (!fu_smbios_setup_from_path(self, path, &error_local)) {
-			if (!g_error_matches(error_local, G_FILE_ERROR, G_FILE_ERROR_ACCES)) {
-				g_propagate_error(error, g_steal_pointer(&error_local));
-				return FALSE;
-			}
-			g_debug("ignoring %s", error_local->message);
+		if (fu_smbios_setup_from_path(self, path, &error_local))
+			return TRUE;
```
hughsie, line 867: this makes my head hurt a bit. can we keep the ! return failure and just add a `} else { return TRUE;}`?

superm1: sure

hughsie (approved):
Thanks for hunting this one down.

superm1:
> Thanks for hunting this one down.

Sure.  It's kinda a bad one.  Between that and the one with the devone (and anything based off Insyde really) being broken, care for a new fixup release?

hughsie:
> care for a new fixup release

Planned for Monday.

PR #4854, opened 2022-07-21
Wip/hughsie/acquiesce composite

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -57,6 +57,7 @@ typedef struct {
 	GRWLock parent_guids_mutex;
 	GPtrArray *parent_physical_ids; /* (nullable) */
 	guint remove_delay;		/* ms */
+	guint acquiesce_delay;		/* ms */
 	guint request_cnts[FWUPD_REQUEST_KIND_LAST];
 	gint order;
 	guint priority;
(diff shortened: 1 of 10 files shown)


superm1 (approved):
The 2500 you came up with before; is it a fairly arbitrary guess?  I wonder if the same number should be used for all devices or if different types should have different numbers?

hughsie:
> is it a fairly arbitrary guess

Yup. We could certainly have something like `fu_device_set_acquiesce_delay()` and quirk if you think that's useful. Better names welcome too.

superm1:
> Yup. We could certainly have something like fu_device_set_acquiesce_delay() and quirk if you think that's useful. Better names welcome too.

Well just for future thought if this is too annoying/slow a value we find.  No rush on doing that now.

hughsie:
We could also specify this in the config file if you think that's a good idea.

superm1:
Config file /dropping the flag is a good idea.  Will make it easier to come up with the right values too.

hughsie:
> Config file /dropping the flag is a good idea

Thinking about it a bit more, I think we **do** want per-device values; e.g. synaptics_mst wants a few seconds, scsi needs a new ms.

hughsie:
Okay, reworked a bit ^^

superm1 (dismissed):
That looks good, but I think that you should leave in the generated documentation that the default for USB and Udev devices is 2500.

hughsie:
> that the default for USB and Udev devices is 2500

Ack, done. I also wondered about having a 50ms delay for *all* devices on the logic that it'll flush out any pending events and not actually make anything measurably slower.

superm1:
> > that the default for USB and Udev devices is 2500
> 
> Ack, done. I also wondered about having a 50ms delay for *all* devices on the logic that it'll flush out any pending events and not actually make anything measurably slower.

Not a bad idea.

superm1 (dismissed):

PR #4856, opened 2022-07-21
trivial: fixup setup helpers to install right version of meson

superm1:
if meson wasn't installed by pip it won't have a version string

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/fwupd_setup_helpers.py
+++ b/contrib/ci/fwupd_setup_helpers.py
@@ -58,9 +58,20 @@ def test_markdown(debug):
 
 
 def test_meson(debug):
-    from importlib.metadata import version
+    from importlib.metadata import version, PackageNotFoundError
 
-    new_enough = version("meson") >= MINIMUM_MESON
+    try:
+        new_enough = version("meson") >= MINIMUM_MESON
+    except PackageNotFoundError:
+        import subprocess
+
+        try:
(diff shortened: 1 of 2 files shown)


contrib/ci/fwupd_setup_helpers.py
```diff
@@ -58,9 +58,12 @@ def test_markdown(debug):
...
 
 def test_meson(debug):
-    from importlib.metadata import version
+    from importlib.metadata import version, PackageNotFoundError
 
-    new_enough = version("meson") >= MINIMUM_MESON
+    try:
+        new_enough = version("meson") >= MINIMUM_MESON
+    except PackageNotFoundError:
+        new_enough = False
```
hughsie, line 66: Could we call into the system meson and get the version? It seems a bit antisocial to just ignore the distro version.

superm1: I agree but I'm not sure how to. Any ideas?

hughsie: Would `meson --version` work?

superm1: Yeah I think I got that setup now, take a look.

hughsie (approved):

PR #4860, opened 2022-07-22
Wip/hughsie/1 7 x backports

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/arch.sh
+++ b/contrib/ci/arch.sh
@@ -6,6 +6,9 @@ shopt -s extglob
 #clone test firmware if necessary
 . ./contrib/ci/get_test_firmware.sh
 
+#refresh package cache and update image
+pacman -Syu --noconfirm
+
 #install anything missing from the container
 ./contrib/ci/fwupd_setup_helpers.py install-dependencies -o arch
 
@@ -19,11 +22,11 @@ popd
 chown nobody . -R
 
 # install and run the custom redfish simulator
-pacman -Syu --noconfirm python-flask
+pacman -S --noconfirm python-flask
 ../plugins/redfish/tests/redfish.py &
 
 # install and run TPM simulator necessary for plugins/uefi-capsule/uefi-self-test
-pacman -Syu --noconfirm swtpm tpm2-tools
+pacman -S --noconfirm swtpm tpm2-tools
 swtpm socket --tpm2 --server port=2321 --ctrl type=tcp,port=2322 --flags not-need-init --tpmstate "dir=$PWD" &
 trap 'kill $!' EXIT
 # extend a PCR0 value for test suite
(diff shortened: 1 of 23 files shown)


hughsie:
@superm1 anything else you think we're missing?

hughsie:
@superm1 I'm also a bit perplexed by the out-of-tree-link CI failure; any ideas?

superm1:
> Run-time dependency fwupd found: YES1.8.3
> Run-time dependency fwupdplugin found: YES1.8.3

I think that's why it's failing.  It's trying to link with a 1.8.3 library.

superm1:
> > Run-time dependency fwupd found: YES1.8.3
> > Run-time dependency fwupdplugin found: YES1.8.3
> 
> I think that's why it's failing. It's trying to link with a 1.8.3 library.

It seems to me when the just built package installs it happens to pull in the newer version from the arch repos.
>  fwupd-1.8.3-1-x86_64 downloading...

@ArchangeGabriel thoughts?

superm1:
> @superm1 anything else you think we're missing?

anything for the ccgx docks so that they don't need 1.8.x to update?  That's all that comes to mind to me.

ArchangeGabriel:
> > > Run-time dependency fwupd found: YES1.8.3
> > > Run-time dependency fwupdplugin found: YES1.8.3
> > 
> > 
> > I think that's why it's failing. It's trying to link with a 1.8.3 library.
> 
> It seems to me when the just built package installs it happens to pull in the newer version from the arch repos.
> 
> > fwupd-1.8.3-1-x86_64 downloading...
> 
> @ArchangeGabriel thoughts?

Yes, it happens when you run `pacman -Syu --noconfirm qt5-base` (https://github.com/fwupd/fwupd/blob/main/contrib/ci/arch.sh#L40). Given the time span between the first command and this last one, I’m not sure it’s relevant to be constantly upgrading the cache and the system. I initially added it to the first command to avoid caching issue w.r.t. to the database in the Docker image, but then this spread to other commands to. I’ll open two PR against 1_7_X and main to better fix this.

superm1:
> > > > Run-time dependency fwupd found: YES1.8.3
> > > > Run-time dependency fwupdplugin found: YES1.8.3
> > > 
> > > 
> > > I think that's why it's failing. It's trying to link with a 1.8.3 library.
> > 
> > It seems to me when the just built package installs it happens to pull in the newer version from the arch repos.
> > 
> > > fwupd-1.8.3-1-x86_64 downloading...
> > 
> > @ArchangeGabriel thoughts?
> 
> Yes, it happens when you run `pacman -Syu --noconfirm qt5-base` (https://github.com/fwupd/fwupd/blob/main/contrib/ci/arch.sh#L40). Given the time span between the first command and this last one, I’m not sure it’s relevant to be constantly upgrading the cache and the system. I initially added it to the first command to avoid caching issue w.r.t. to the database in the Docker image, but then this spread to other commands too. I’ll open two PR against 1_7_X and main to better fix this.

Thanks so much!

PR #4868, opened 2022-07-25
Export the generic Intel NVM firmware format

hughsie:
This is being used for other products, e.g. USB4 docks.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-intel-nvm.c
```diff
@@ -0,0 +1,880 @@
...
+			   {0x15EF, 3, FU_INTEL_NVM_FAMILY_TR, 2},   /* TR 4C device */
+			   {0x15EE, 3, FU_INTEL_NVM_FAMILY_BB, 0},   /* BB device */
+			   /* Maple ridge devices
+			    * NOTE: These are expected to be flashed via UEFI capsules *not*
+			    * Thunderbolt plugin Flashing via fwupd will require matching kernel
+			    * work. They're left here only for parsing the binaries
+			    */
+			   {0x1136, 4, FU_INTEL_NVM_FAMILY_MR, 2},
+			   {0x1137, 4, FU_INTEL_NVM_FAMILY_MR, 2},
+			   {0}};
```
superm1, line 520: probably missing entries for Goshen Ridge here

superm1 (commented):
the kernel provided NVM binary and NVM update binary are not identical. (update binary will have more data in the header).  shouldn't you still have two update types?

also this should be `fu-intel-thunderbolt-nvm`.  they use the term NVM for other technologies too and we have no confirmation it follows similar format (it shouldn't).

superm1:
CC @YehezkelShB and @westeri

libfwupdplugin/fu-intel-nvm.c
hughsie: Ya, I pinged @bleungatchromium for those

hughsie:
> shouldn't you still have two update types?

Yup, there are two types; `FuIntelNvm` (no header) and `FuIntelNvmFirmware` (the blob we ship as the update with the header).

superm1:
> > shouldn't you still have two update types?
> 
> Yup, there are two types; `FuIntelNvm` (no header) and `FuIntelNvmFirmware` (the blob we ship as the update with the header).

Yes; right.  I'm not crazy about these names.  What about `FuIntelThunderboltNvmKernel` and `FuIntelThunderboltNvmFirmware`.  The kernel is the one that we get the limited representation from.

hughsie:
> Yes; right. I'm not crazy about these names

What about now? I think this works okay.

superm1:
> libfwupdplugin/tests/intel-thunderbolt.bin

Can this be generated at the time of tests rather than the binary stored in the tree?

hughsie:
> Can this be generated at the time of tests rather than the binary stored in the tree?

Nah, for fuzzing it needs to be super quick; we don't have fwupdtool in place to generate them.

superm1 (commented):
Knowing that we have another vendor's USB4 firmware needing to be supported soon (#4913) I think this is missing an abstraction layer of the vendor's firmware rather than assuming all TBT3/USB4 firmware is Intel's.

Things like `fu_intel_thunderbolt_nvm_is_host` for example only make sense on Intel's platforms and won't make sense with ASMedia or AMD (although granted AMD doesn't use this plugin).

There are other assumptions where it is `FU_INTEL_THUNDERBOLT_NVM(foo)` when it should be something like `FU_USB4_NVM(foo)` and `FU_INTEL_THUNDERBOLT_NVM` happens to derive from `FU_USB4_NVM`.  Then when it comes time to add something for asmedia there can be various -EOPNOTSUPP returns for the non-applicable things rather than touch all the **exact same code** _again_.

hughsie:
So maybe we want another layer, e.g. `FuFirmware < FuThunderboltNvm < FuIntelThunderboltNvm < FuIntelThunderboltFirmware` -- what I'm not sure is what to move lower, e.g. for ASMedia. e.g. do all these make sense?

 * _get_device_id()
 * _get_flash_size()
 * _get_model_id()
 * _get_vendor_id()
 * _has_pd()
 * _is_host()
 * _is_native()

Ideas welcome.

superm1:
> * _get_device_id()
> * _get_flash_size()
> * _get_model_id()
> * _get_vendor_id()
> * _has_pd()
> * _is_host()

ASMedia may not be documenting their  NVM format.  IOW even if they support some of these they might not even tell us.  I think all of those should be Intel-NVMisms or if you put them a level higher default to a not supported error code so when we add ASMedia they aren't required. 

> * _is_native()
Definitely not this one - it's Intel specific.

hughsie:
> So maybe we want another layer

I spend an hour or so on this, and then came to the realisation -- I don't think we need the generic `FuThunderboltNvm` subclass format at all. The two users of `FuIntelThunderboltFirmware` would be the `thunderbolt` plugin and the `intel-usb4` plugin. I can't see the Intel USB4 plugin updating ASMedia devices (I'd be a different VID/PID and mailbox format - surely?) and the Thunderbolt plugin doesn't do anything else than the Intel-only TBT3 -- and if dell-dock was to use a different controller then we'd use something like asmedia-usb4 rather than deal with the device internally like before.

Unless I'm missing something i.e. the mbox format is part of USB4) and the `intel-usb4` plugin needs to be called just `usb4` or something.

superm1:
I think the part you're missing is the reason that I said that we should rename the plugin linux-usb4 or linux-thunderbolt before.

We don't care about the specifics of *how* it's flashed. That's abstracted by the kernel API to sysfs. It's also the reason that you can flash USB4 and TBT3 identically from userspace.

All of the stuff fetching from the NVM is used for building GUIDs for TBT3 and not applicable to USB4 really.

So this is why I think we should differentiate only on the firmware type we interacted with because we build GUIDs from it on older stuff.

hughsie:
Wait, so the ASMedia-USB4 device is going to be flashed using the `thunderbolt` plugin?!

superm1:
Yup

hughsie:
Ha! That's the nugget of information I was missing! :) Will reassess and respin.

hughsie:
It seems we can do without the intermediate format (and solve things like _is_native() in a nicer way) if we have a `FuFirmware->compare(other)` vfunc. Then each format can decide however it likes if `VID=0` means "compatible with everything" or whether the VID mas to match exactly. I'm not sold on the name tho, how about `FuFirmware->check_compatible(other)`?

superm1:
> It seems we can do without the intermediate format (and solve things like _is_native() in a nicer way) if we have a `FuFirmware->compare(other)` vfunc. Then each format can decide however it likes if `VID=0` means "compatible with everything" or whether the VID mas to match exactly. I'm not sold on the name tho, how about `FuFirmware->check_compatible(other)`?

Sure conceptually that makes sense to me I think.

hughsie:
https://github.com/fwupd/fwupd/pull/4951 means I can clean up this branch quite a bit.

hughsie:
Okay, I think this works much better.

plugins/thunderbolt/fu-thunderbolt-device.c
```diff
@@ -305,13 +304,13 @@ fu_thunderbolt_device_prepare_firmware(FuDevice *device,
 				       GError **error)
 {
 	FuThunderboltDevice *self = FU_THUNDERBOLT_DEVICE(device);
-	g_autoptr(FuThunderboltFirmwareUpdate) firmware = fu_thunderbolt_firmware_update_new();
-	g_autoptr(FuThunderboltFirmware) firmware_old = fu_thunderbolt_firmware_new();
+	g_autoptr(FuFirmware) firmware = fu_intel_thunderbolt_firmware_new();
```
superm1, line 307: I still fell like this shouldn't be defaulting to Intel FW type.  When it's parsed it looks specifically like an Intel one you subclass it to the Intel one, but it should start out as one that is generic.

hughsie: I put the thing about fu_firmware_new_from_gtypes() into the commit message -- I figured we'd switch from the explicit type to using `fu_firmware_new_from_gtypes` once we had a firmware parser for ASMedia implemented.

superm1: I guess i'm not convinced we're going to get a parser for ASMedia.  They indicated on the kernel mailing list that they're not disclosing their NVM format (no need to).  So I would rather see it that we match Intel's firmware we know some of the format or "generic" USB4 firmware with no details.  I don't know if we'll be seeing VIA come to the table either with their format.

hughsie: How about the fixup commit? That's what I had in mind.

superm1: Yes! That's the direction I had in mind too.  The only thing I would change with it is that there should be a type that will match when it's not Intel.  Maybe just "GENERIC_USB4".  Then ASMedia and Via and all will hopefully with no other changes, and if we get parser info we can add that later.

hughsie: > should be a type that will match when it's not Intel

What would it do that isn't covered by "_unknown binary blob_" of FuFirmware?

hughsie: FYI `fu_firmware_check_compatible` does the heavy lifting, and that's a vfunc that can be implemented by any future firmware subclass.

superm1: > What would it do that isn't covered by "unknown binary blob" of FuFirmware?

Oh perhaps.  I was getting caught up in the checking it's not `NULL`.  If that's not the common path, then yeah this is good.

superm1 (approved):

superm1 (approved):

PR #4877, opened 2022-07-28
Allow plugins and backends to print debugging information to the console

hughsie:
This is really useful for debugging.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -14,6 +14,7 @@ with a non-standard filesystem layout.
 * `FWUPD_DEVICE_TESTS_BASE_URI` sets the base URI when downloading firmware for the device-tests
 * `FWUPD_SUPPORTED` overrides the `-Dsupported_build` meson option at runtime
 * `FWUPD_VERBOSE` is set when running `--verbose`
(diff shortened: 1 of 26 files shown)


docs/env.md
```diff
@@ -14,6 +14,7 @@ with a non-standard filesystem layout.
 * `FWUPD_DEVICE_TESTS_BASE_URI` sets the base URI when downloading firmware for the device-tests
 * `FWUPD_SUPPORTED` overrides the `-Dsupported_build` meson option at runtime
 * `FWUPD_VERBOSE` is set when running `--verbose`
+* `FWUPD_PLUGIN_VERBOSE` can be used to show detailed plugin and backend debugging
```
superm1, line 17: I don't like this name.  It's too similar to just running --verbose for a given plugin.
How about  `FWUPD_BACKEND_VERBOSE`

hughsie: yup, works for me. Will fix when meetings done.

hughsie: Done.

superm1 (approved):

PR #4881, opened 2022-07-29
Add support to map BIOS attributes to Security Attributes

superm1:
This will allow clients to be able to suggest to users to change BIOS settings to try to fix security problems.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-bios-attr.c
+++ b/libfwupd/fwupd-bios-attr.c
@@ -22,10 +22,12 @@ fwupd_bios_attr_finalize(GObject *object);
 
 typedef struct {
 	FwupdBiosAttrKind kind;
+	gchar *id;
 	gchar *name;
 	gchar *description;
 	gchar *path;
 	gchar *current_value;
+	gchar *preferred_value;
 	guint64 lower_bound;
 	guint64 upper_bound;
 	guint64 scalar_increment;
@@ -35,6 +37,46 @@ typedef struct {
 
 G_DEFINE_TYPE_WITH_PRIVATE(FwupdBiosAttr, fwupd_bios_attr, G_TYPE_OBJECT)
 #define GET_PRIVATE(o) (fwupd_bios_attr_get_instance_private(o))
+
+/**
+ * fwupd_bios_attr_get_id
+ * @self: a #FwupdBiosAttr
+ *
+ * Gets the unique attribute identifier for this attribute/driver
+ *
+ * Returns: attribute ID if set otherwise NULL
+ *
+ * Since: 1.8.4
+ **/
+const gchar *
+fwupd_bios_attr_get_id(FwupdBiosAttr *self)
+{
+	FwupdBiosAttrPrivate *priv = GET_PRIVATE(self);
+	g_return_val_if_fail(FWUPD_IS_BIOS_ATTR(self), NULL);
+	return priv->id;
+}
+
+/**
+ * fwupd_bios_attr_set_id
+ * @self: a #FwupdBiosAttr
+ *
+ * Sets the unique attribute identifier for this attribute
+ *
+ * Since: 1.8.4
+ **/
+void
+fwupd_bios_attr_set_id(FwupdBiosAttr *self, const gchar *id)
+{
+	FwupdBiosAttrPrivate *priv = GET_PRIVATE(self);
+	g_return_if_fail(FWUPD_IS_BIOS_ATTR(self));
+
+	/* not changed */
+	if (g_strcmp0(priv->id, id) == 0)
+		return;
+
+	g_free(priv->id);
+	priv->id = g_strdup(id);
+}
 
 /**
  * fwupd_bios_attr_get_read_only:
@@ -412,6 +454,47 @@ fwupd_bios_attr_get_description(FwupdBiosAttr *self)
 }
 
 /**
+ * fwupd_bios_attr_get_preferred_value:
+ * @self: a #FwupdBiosAttr
+ *
+ * Gets the value that when written to an attribute would activate it or satisfy
+ * a security requirement.
+ *
+ * Returns: the preferred value of the attribute.
+ *
+ * Since: 1.8.4
+ **/
+const gchar *
+fwupd_bios_attr_get_preferred_value(FwupdBiosAttr *self)
+{
+	FwupdBiosAttrPrivate *priv = GET_PRIVATE(self);
+	g_return_val_if_fail(FWUPD_IS_BIOS_ATTR(self), NULL);
+	return priv->preferred_value;
+}
+
+/**
+ * fwupd_bios_attr_set_preferred_value:
+ * @self: a #FwupdBiosAttr
+ * @value: The string to set preferred value to
+ *
+ * Sets the string used for the preferred value of an attribute.
+ *
+ * Since: 1.8.4
+ **/
+void
+fwupd_bios_attr_set_preferred_value(FwupdBiosAttr *self, const gchar *value)
+{
+	FwupdBiosAttrPrivate *priv = GET_PRIVATE(self);
+
+	/* not changed */
+	if (g_strcmp0(priv->preferred_value, value) == 0)
+		return;
+
+	g_free(priv->preferred_value);
+	priv->preferred_value = g_strdup(value);
+}
+
+/**
  * fwupd_bios_attr_get_current_value:
  * @self: a #FwupdBiosAttr
  *
@@ -477,6 +560,12 @@ fwupd_bios_attr_to_variant(FwupdBiosAttr *self)
 			      "{sv}",
 			      FWUPD_RESULT_KEY_BIOS_ATTR_TYPE,
 			      g_variant_new_uint64(priv->kind));
+	if (priv->id != NULL) {
+		g_variant_builder_add(&builder,
+				      "{sv}",
+				      FWUPD_RESULT_KEY_BIOS_ATTR_ID,
+				      g_variant_new_string(priv->id));
+	}
 	if (priv->name != NULL) {
 		g_variant_builder_add(&builder,
 				      "{sv}",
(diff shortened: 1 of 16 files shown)


superm1:
Depending upon how this is used it may also make sense to try to include information about what value is the correct "positive" one to use in the security attr.

For some cases it's super obvious (enable/disable) but others you gotta know (sleep state of Linux or Windows 10).

libfwupdplugin/fu-self-test.c
```diff
@@ -2948,7 +2948,7 @@ fu_bios_attrs_load_func(void)
 	g_assert_false(ret);
 
 	/* check an attribute reads from kernel as expected by fwupd today */
-	attr = fu_context_get_bios_attr(ctx, "AMDMemoryGuard");
+	attr = fu_context_get_bios_attr(ctx, "com.thinklmi.AMDMemoryGuard");
```
hughsie, line 2951: I think this is a really good idea.

libfwupd/fwupd-bios-attr.c
```diff
@@ -477,6 +518,12 @@ fwupd_bios_attr_to_variant(FwupdBiosAttr *self)
 			      "{sv}",
 			      FWUPD_RESULT_KEY_BIOS_ATTR_TYPE,
 			      g_variant_new_uint64(priv->kind));
+	if (priv->attr_id != NULL) {
+		g_variant_builder_add(&builder,
+				      "{sv}",
+				      FWUPD_RESULT_KEY_BIOS_ATTR_ID,
```
hughsie, line 524: I think FWUPD_RESULT_KEY_APPSTREAM_ID works here, no?

libfwupd/fwupd-security-attr.c
```diff
@@ -223,6 +224,46 @@ fwupd_security_attr_flag_to_suffix(FwupdSecurityAttrFlags flag)
...
+ *
+ * Gets the #FwupdBiosAttr that can be used to improve this
+ * #FwupdSecurityAttr.
+ *
+ * Returns: The unique ID used for #FwupdBiosAttr or NULL
+ *
+ * Since: 1.8.4
+ **/
+const gchar *
+fwupd_security_attr_get_bios_attr_id(FwupdSecurityAttr *self)
```
hughsie, line 239: I think fwupd_security_attr_get_bios_id is better tbh.

plugins/acpi-facp/fu-plugin-acpi-facp.c
```diff
@@ -41,6 +41,12 @@ fu_plugin_acpi_facp_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
 		return;
 	}
 	if (!fu_acpi_facp_get_s2i(facp)) {
+		FwupdBiosAttr *bios_attr =
+		    fu_context_get_bios_attr(fu_plugin_get_context(plugin), "SleepState");
+		if (bios_attr != NULL) {
+			fwupd_security_attr_set_bios_attr_id(attr,
+							     fwupd_bios_attr_get_id(bios_attr));
+		}
```
hughsie, line 49: I think we do need to say what bios attr value we need to fix this. e.g. have something like `fwupd_security_attr_set_bios_value` making it clear that this value is the *desired* value, not the actual value.

plugins/acpi-facp/fu-plugin-acpi-facp.c
```diff
@@ -41,6 +41,12 @@ fu_plugin_acpi_facp_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
 		return;
 	}
 	if (!fu_acpi_facp_get_s2i(facp)) {
+		FwupdBiosAttr *bios_attr =
+		    fu_context_get_bios_attr(fu_plugin_get_context(plugin), "SleepState");
```
hughsie, line 45: I think we want to get this by AppStream ID, no? e.g. the desired value from `com.lenovo.SleepState` might be `idle` and the desired value for `com.dell.SleepState` might be `s2idle`.

libfwupd/fwupd-bios-attr.c
superm1: That was my original thought. But if you want to share with security attr you need to have two different key names.

plugins/acpi-facp/fu-plugin-acpi-facp.c
superm1: Yes. This is only for Lenovo. Will fix to add the ID.

libfwupd/fwupd-bios-attr.c
hughsie: Bleh, indeed. Sorry.

plugins/acpi-facp/fu-plugin-acpi-facp.c
superm1: Right. I think we need two keys then. Like a desired value key and which BIOS setting to change key.

hughsie: Agree.

libfwupd/fwupd-bios-attr.c
```diff
@@ -22,10 +22,12 @@ fwupd_bios_attr_finalize(GObject *object);
 
 typedef struct {
 	FwupdBiosAttrKind kind;
+	gchar *attr_id;
```
hughsie, line 25: hmm, i still think this one wants to be appstream_id.

hughsie: if not, then just `id`

libfwupd/fwupd-bios-attr.c
```diff
@@ -546,6 +645,10 @@ fwupd_bios_attr_from_key_value(FwupdBiosAttr *self, const gchar *key, GVariant *
 		fwupd_bios_attr_set_path(self, g_variant_get_string(value, NULL));
 		return;
 	}
+	if (g_strcmp0(key, FWUPD_RESULT_KEY_BIOS_ATTR_POSITIVE_VALUE) == 0) {
+		fwupd_bios_attr_set_positive_value(self, g_variant_get_string(value, NULL));
```
hughsie, line 649: not sure i like the `positive` noun. Got any other ideas?

libfwupd/fwupd-bios-attr.c
superm1: It shouldn't be `appstream_id`, I think `id` is fine.

libfwupd/fwupd-bios-attr.c
superm1: OK so if the heuristic to set it is dropped from `fu-bios-attrs`, how about `preferred`?  Then we can set preferred only with security attributes.

hughsie: `preferred` makes sense to me. That or `ideal` or `endorsed` or `recommended`

libfwupd/fwupd-bios-attr.c
```diff
@@ -22,10 +22,12 @@ fwupd_bios_attr_finalize(GObject *object);
 
 typedef struct {
 	FwupdBiosAttrKind kind;
+	gchar *id;
 	gchar *name;
 	gchar *description;
 	gchar *path;
 	gchar *current_value;
+	gchar *positive_value;
```
hughsie, line 30: preferred

libfwupd/fwupd-bios-attr.c
```diff
@@ -495,6 +584,12 @@ fwupd_bios_attr_to_variant(FwupdBiosAttr *self)
 				      FWUPD_RESULT_KEY_DESCRIPTION,
 				      g_variant_new_string(priv->description));
 	}
+	if (priv->positive_value != NULL) {
+		g_variant_builder_add(&builder,
+				      "{sv}",
+				      FWUPD_RESULT_KEY_BIOS_ATTR_POSITIVE_VALUE,
```
hughsie, line 590: same

libfwupd/fwupd-enums-private.h
```diff
@@ -508,11 +508,26 @@ G_BEGIN_DECLS
...
+ **/
+#define FWUPD_RESULT_KEY_BIOS_ATTR_ID "BiosAttrId"
+/**
+ * FWUPD_RESULT_KEY_BIOS_ATTR_POSITIVE_VALUE:
+ *
+ * Result key to represent the value that would enable this attribute.
+ *
+ * The D-Bus type signature string is 's' i.e. a string.
+ **/
+#define FWUPD_RESULT_KEY_BIOS_ATTR_POSITIVE_VALUE "BiosAttrPositiveValue"
```
hughsie, line 526: same

libfwupd/fwupd-bios-attr.c
superm1: gah, forgot a few thanks.

hughsie (approved):
Looks good. Do you have any ideas on how this could work on the CLI? e.g. do we want a --interactive set of prompts for `fwupdmgr security`? A different PR tho.

superm1:
> Looks good. Do you have any ideas on how this could work on the CLI? e.g. do we want a --interactive set of prompts for `fwupdmgr security`? A different PR tho.

I hadn't really given it thought until you asked.  I think at the end of showing security information there could be a prompt along the lines of:
````
fwupd could change some BIOS settings for you for the following items:
Foo
* bar->baz 
SecureBoot
*  Disabled -> Enabled

WARNING: Although changing BIOS settings may help improve your security score, it may cause problems with the operation of your system, please be prepared to enter your firmware setup to change settings back if any of these cause a problem for you.
```

hughsie:
Ya, something exactly like that.

hughsie:
Maybe allowlisting them for now too.

superm1:
> Maybe allowlisting them for now too.

What do you mean?  Only certain settings are tagged to a security attribute.

hughsie:
> What do you mean? Only certain settings are tagged to a security attribute.

Ohh, i mean we probably want bespoke translations and explanations for the super common ones, like turning on SB.

superm1:
> > What do you mean? Only certain settings are tagged to a security attribute.
> 
> Ohh, i mean we probably want bespoke translations and explanations for the super common ones, like turning on SB.

well that's why I want to have more values filled in for the Dell stuff and some hardcoded stuff for Lenovo.  I think if we can dump all of the known lenovo attributes into a google doc and then ask the internet to fill it in we should get all the strings we need (and then some).

PR #4883, opened 2022-07-29
Update POTFILES.in

piotrdrag:


superm1 (approved):

superm1:
Qq for you - 

I was thinking it would also make sense to do translations for all the descriptions we have for Dell, but it's a bunch of random text files. Can we just add those to a list somehow?

In libfwupdplugin/tests/bios-attrs/dell-wmi-sysman/attributes/*/display_name

I wonder if maybe we just need a helper that dumps them all into a C file perhaps.

hughsie:
Also can we add a precommit script that checks for translations in files that are not in POTFILES.in? It's something we seem to get wrong *a lot*.

superm1:
Yeah agree.
Follow up thought to my other comment; can we crowdsource some descriptions for all the Lenovo settings?  They don't provide them with their BIOS like Dell does, but I think it would be good to have, and do it early so we can get translations in.
I just wrote up one of them that's in that file that @piotrdrag caught us missing to kick it off.

mrhpearson:
> can we crowdsource some descriptions for all the Lenovo settings? They don't provide them with their BIOS like Dell does

What are you looking for? If you give me an example I'll see what I can find

superm1:
Basically all the attributes thinklmi exports. The driver has a few major problems:

1. It doesn't export the mandatory "type" attribute. Right now we're working around this in fwupd but it doesn't scale so driver needs to fix it in future.
2. Doesn't export possible_values attribute consistently. On p620 I can see this doesn't work.
3. display_name doesn't have a description but just duplicates the attribute name.

So I want all 3 of those fixed for the best experience.

If the kernel and BIOS can't get descriptions for 3 then we should crowd source descriptions or if you can get a whole list that's fine too.

mrhpearson:
OK - I'll talk to the FW team

 - Right now we don't have the type attribute to my knowledge. I think that's a Dell thing. I believe everything for us is just a string.
 - The possible_values should work - if it's not working on the P620 that's a bug. I'll flag it
 - Ack on the descriptions - be nice if that was exported somehow. I can go and ask for that but suspect it will be a slow one. Happy to be the first pass for a crowd-source implementation until we get something official implemented.

As a note for myself - tracked internally with  LO-1905

superm1:
> Right now we don't have the type attribute to my knowledge. I think that's a Dell thing. I believe everything for us is just a string.

It's a kernel construct not a firmware construct.  It's [mandatory](https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/sysfs-class-firmware-attributes#L15) per the spec.

> The possible_values should work - if it's not working on the P620 that's a bug. I'll flag it

Thanks

> Ack on the descriptions - be nice if that was exported somehow. I can go and ask for that but suspect it will be a slow one. Happy to be the first pass for a crowd-source implementation until we get something official implemented.

Yeah I'm fine if it's not exported by the BIOS.  That would be nice, but crowdsourced approach would be just fine for now.

If you can make a first pass, please populate this area with more attributes and descriptions:
https://github.com/fwupd/fwupd/blob/main/src/fu-util-bios-attr.c#L34

superm1:
Put that in https://github.com/fwupd/fwupd/issues/4899 for you to keep up on.

PR #4888, opened 2022-07-31
Set the preferred value on the security attribute

hughsie:
Semantically it's the desire of the security attribute, not the bios
attribute, i.e. you could imagine that a specific attribute would have
to be *foo or bar or baz* for HSI-1 and *only foo* for HSI-2

Also make it easier to add possible BIOS attribute preferred values
in plugin code.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
Yes I agree with this direction.

hughsie:
> Yes I agree with this direction.

Do you think all the plugins should use `fu_security_attr_new()` as well?

superm1:
> fu_security_attr_new() 

Don't see why not.

hughsie:
Okay, I think we can do a bit better still. How about this^^?

libfwupdplugin/fu-security-attr.c
```diff
@@ -0,0 +1,103 @@
...
+ * @id: a #FwupdBiosAttr ID or name
+ * @needle: The substring of a target value
+ *
+ * Checks all configured possible values of an enumeration attribute and
+ * if any match @needle then set as the target value.
+ *
+ * Since: 1.8.4
+ **/
+void
+fu_security_attr_add_bios_target_value(FwupdSecurityAttr *attr,
```
superm1, line 34: Maybe make the to_string use the fwupdbiosattr to_string as well for debugging?

hughsie: We don't have the object, we just have the ID string.

superm1: But you can fetch the object when doing to string. You're saving the context

hughsie: Maybe we want to keep the FwupdBiosAttr ref'd in F**u**SecurityAttr? Will experiment.

hughsie: Hmm, so I did this and it wasn't hugely useful; `fu_util_security_events_to_string()` is the only caller of `fu_security_attr_to_string` and we don't get `FwupdBiosAttr` objects on _events_. What did you have in mind? The Json output?

superm1: Yeah that's exactly what I had in mind. I was thinking that if you get it as part of the json you get a capture of there being a problem what can fix it and also what the current setting is.

hughsie: I think the only way that's possible is to make fwupd_security_attr_set_bios_attr_id() into fwupd_security_attr_set_bios_attr() -- i.e. pass the full object.

superm1: I was thinking you can still pass the id but if json or string are requested use the local reference to the context to look it up? But maybe that fails for clients from dbus doesn't it..

hughsie: Ya, it's the wrong layer; no context in libfwupd. I'm not against passing in the object, but it does feel a little circular.

superm1: It's a weak use case, probably not worth it.

Really long term we should have a full system information to json (hsi, devices, updates, bios attributes). The same stuff that we can use for emulating a system.

So it can just dump as part of that.

hughsie: > So it can just dump as part of that.

I'd forgotten emulation, we probably want that too. Phase 2.

superm1: Well emulation can already be accomplished from the environment variables and a sysfs tree. But yeah json emulation for everything is preferred.

superm1 (approved):

PR #4892, opened 2022-08-01
plugins/flashrom: Skip BC check for coreboot devices

Sean-StarLabs:
Set BcrAddr to 0x0 for all coreboot devices, so that the check of
BIOS Control is skipped as coreboot won't forcibly set this.

### DIFF
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -39,11 +39,13 @@ Plugin = flashrom
 [d33219e2-b84c-53a8-a624-27af9752dba6]
 Branch = coreboot
 Flags = reset-cmos
+PciBcrAddr = 0x0
 
 # StarLite Mk II (coreboot GUID)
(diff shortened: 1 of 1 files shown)


hughsie (approved):
If it works, LGTM. Coreboot not handling the SMM makes sense, i guess.

Sean-StarLabs:
> If it works, LGTM. Coreboot not handling the SMM makes sense, i guess.

That would be a no!

superm1:
Maybe should this quirk apply to all coreboot machines by default and then it can be inverted only when an implementation has it?

hughsie:
> Maybe should this quirk apply to all coreboot machines by default

Mayybe. The tricky bit is that PciBcrAddr is being set by multiple plugins.

Sean-StarLabs:
Works as intended. Not sure how one would disable it by default though...

superm1 (approved):

PR #4893, opened 2022-08-01
Prompt users to fix BIOS configuration issues

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -3387,6 +3387,55 @@ fu_util_sync_bkc(FuUtilPrivate *priv, gchar **values, GError **error)
 }
 
 static gboolean
+fu_util_security_modify_bios_attr(FuUtilPrivate *priv, FwupdSecurityAttr *attr, GError **error)
+{
+	g_autoptr(GString) body = g_string_new(NULL);
+	g_autoptr(GString) title = g_string_new(NULL);
+
+	g_string_append_printf(title,
+			       "%s: %s",
+			       /* TRANSLATORS: title prefix for the BIOS settings dialog */
+			       _("Configuration Change Suggested"),
+			       fwupd_security_attr_get_title(attr));
+
+	g_string_append(body, fwupd_security_attr_get_description(attr));
+	g_string_append(body, "\n\n");
+	g_string_append_printf(body,
+			       /* TRANSLATORS: the %1 is a BIOS setting name.  %2 and %3 are the
+				  values, e.g. "True" or "Windows10" */
+			       _("This tool can change the BIOS setting '%s' from '%s' to '%s' "
+				 "automatically, but it will only be active after restarting the "
+				 "computer."),
+			       fwupd_security_attr_get_bios_attr_id(attr),
+			       fwupd_security_attr_get_bios_attr_current_value(attr),
+			       fwupd_security_attr_get_bios_attr_target_value(attr));
+	g_string_append(body, "\n\n");
+	g_string_append_printf(body,
+			       /* TRANSLATORS: the user has to manually recover; we can't do it */
+			       _("You should ensure you are comfortable restoring the setting from "
+				 "the system firmware setup, as this change may cause the system "
+				 "to not boot into Linux or cause other system instability."));
+	fu_util_warning_box(title->str, body->str, 80);
+
+	/* ask for confirmation */
+	g_print("\n%s [y|N]: ",
+		/* TRANSLATORS: prompt to apply the update */
+		_("Perform operation?"));
+	if (!fu_util_prompt_for_boolean(FALSE))
+		return TRUE;
+	if (!fwupd_client_modify_bios_attr(priv->client,
+					   fwupd_security_attr_get_bios_attr_id(attr),
+					   fwupd_security_attr_get_bios_attr_target_value(attr),
+					   priv->cancellable,
+					   error))
+		return FALSE;
+
+	/* do not offer to upload the report */
+	priv->completion_flags |= FWUPD_DEVICE_FLAG_NEEDS_REBOOT;
+	return TRUE;
+}
+
+static gboolean
 fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 {
 	FuSecurityAttrToStringFlags flags = FU_SECURITY_ATTR_TO_STRING_FLAG_NONE;
@@ -3480,12 +3529,33 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 		}
 	}
 
-	/* opted-out */
-	if (priv->no_unreported_check)
-		return TRUE;
+	/* any things we can fix? */
+	for (guint j = 0; j < attrs->len; j++) {
+		FwupdSecurityAttr *attr = g_ptr_array_index(attrs, j);
+		if (!fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS) &&
+		    fwupd_security_attr_get_bios_attr_id(attr) != NULL &&
+		    fwupd_security_attr_get_bios_attr_current_value(attr) != NULL &&
+		    fwupd_security_attr_get_bios_attr_target_value(attr) != NULL) {
+			if (!fu_util_security_modify_bios_attr(priv, attr, error))
+				return FALSE;
+		}
+	}
 
 	/* upload, with confirmation */
-	return fu_util_upload_security(priv, attrs, error);
+	if (!priv->no_unreported_check) {
+		if (!fu_util_upload_security(priv, attrs, error))
+			return FALSE;
+	}
+
+	/* reboot is required? */
+	if (!priv->no_reboot_check &&
+	    (priv->completion_flags & FWUPD_DEVICE_FLAG_NEEDS_REBOOT) > 0) {
+		if (!fu_util_prompt_complete(priv->completion_flags, TRUE, error))
+			return FALSE;
+	}
+
+	/* success */
+	return TRUE;
 }
 
 static void


hughsie:
Screenshot from 2022-08-01 13-59-23

src/fu-util.c
```diff
@@ -3484,6 +3524,16 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 	if (priv->no_unreported_check)
 		return TRUE;
 
+	/* any things we can fix? */
+	for (guint j = 0; j < attrs->len; j++) {
+		FwupdSecurityAttr *attr = g_ptr_array_index(attrs, j);
+		if (fwupd_security_attr_get_bios_attr_id(attr) != NULL &&
+		    fwupd_security_attr_get_bios_attr_value(attr) != NULL) {
```
superm1, line 3531: I think you need to compare that the BIOS attr is not already set to this don't you?

src/fu-util.c
```diff
@@ -3484,6 +3524,16 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
...
+			if (!fu_util_security_modify_bios_attr(priv, attr, error))
+				return FALSE;
+		}
+	}
+
 	/* upload, with confirmation */
 	return fu_util_upload_security(priv, attrs, error);
```
superm1, line 3538: if `attrs->len > 0` and you made it this far (didn't say no) you should offer to reboot at the end of the func.

src/fu-util.c
```diff
@@ -3484,6 +3524,16 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
...
+	for (guint j = 0; j < attrs->len; j++) {
+		FwupdSecurityAttr *attr = g_ptr_array_index(attrs, j);
+		if (fwupd_security_attr_get_bios_attr_id(attr) != NULL &&
+		    fwupd_security_attr_get_bios_attr_value(attr) != NULL) {
+			if (!fu_util_security_modify_bios_attr(priv, attr, error))
+				return FALSE;
+		}
+	}
+
 	/* upload, with confirmation */
```
superm1, line 3537: the upload should come before the BIOS setting change.

src/fu-util.c
```diff
@@ -3386,6 +3386,46 @@ fu_util_sync_bkc(FuUtilPrivate *priv, gchar **values, GError **error)
...
+			       /* TRANSLATORS: title prefix for the BIOS settings dialog */
+			       _("Configuration Change Suggested"),
+			       fwupd_security_attr_get_title(attr));
+
+	g_string_append(body, fwupd_security_attr_get_description(attr));
+	g_string_append(body, "\n\n");
+	g_string_append_printf(body,
+			       /* TRANSLATORS: the % is the value, e.g. "True" or "Windows10" */
+			       _("This tool can change the BIOS setting to '%s' automatically, "
+				 "but it will only be active after restarting the computer."),
```
superm1, line 3406: how about if you tell them what it's current set to?  Not sure if it's useful so feel  free to waive.

src/fu-util.c
superm1: Yes that was my thought too.  We set the BIOS attr ID and target value no matter what.  The other thing you can do is only do this if the security attribute doesn't have a success flag.

hughsie: > if the security attribute doesn't have a success flag

Doh, we should definitely do that. Fixed.

superm1: OK.  A few other ideas
* You should check here whether or not it has the config_fw flag.  More below.
* In security attributes drop the config_fw flag if the target and current value are the same.

Then client doesn't need to check the values (or even know them).  It just needs to see if config_fw is set, and if so then offer to fix it.

src/fu-util.c
hughsie: Agreed, I've made it a bit more sensible.

src/fu-util.c
hughsie: The upload should come on the next reboot, i.e. not this boot tbh

superm1: But don't you want to see data that someone had a fail, and then see MORE data if they fixed it?

src/fu-util.c
hughsie: That's more of a designer question. I suppose if we're asking the user to potentially change it back we should tell then what it's set to. Aren't some of the BIOS attr strings localised too?

src/fu-util.c
hughsie: The new one would override the old one, for the same machine id.

src/fu-util.c
superm1: Yeah that was my thought; they'll need to know what to set it back to if it was changed.  In terms of localization of values that would be OEM dependent if they do that.

hughsie: Gah, so I tried this, and of course got the policykit prompt -- which is really jarring and confusing.

superm1: Hmm, in general we shouldn't be exposing what BIOS settings are for _everything_ to non-root.  But if we're already exposing that a particular current setting is what leads to a potential security vulnerability perhaps we should allow to check those particular settings?  Which I guess comes back to the question of storing some more information in the security attribute.

hughsie: Maybe `s\fwupd_security_attr_get_bios_attr_value\fwupd_security_attr_get_bios_attr_target_value` and then create `fwupd_security_attr_get_bios_attr_current_value()`? I'm nearly OUTATIME for today, but I can do it in a couple of weeks if needed.

superm1: Yeah that's what I'm thinking, and only because of the above that it's okay.

src/fu-util.c
```diff
@@ -3480,6 +3528,27 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
...
+		}
+	}
+
+	/* reboot required? */
+	if (priv->completion_flags & FWUPD_DEVICE_FLAG_NEEDS_REBOOT) {
+		if (priv->no_reboot_check) {
+			g_debug("skipping reboot check");
+			return TRUE;
+		}
+		return fu_util_prompt_complete(priv->completion_flags, TRUE, error);
```
superm1, line 3549: you might not get to prompt to upload results if you do it this way

superm1:
```suggestion
		if (!priv->no_reboot_check) {
			return fu_util_prompt_complete(priv->completion_flags, TRUE, error);
		}
		g_debug("skipping reboot check");
```

superm1:
Other than my comment about the reboot stuff, other comments:
1. This needs to be ported into `fwupdtool` too.
2. At least with UEFI secure boot attribute I think we need to do some more safety checking.  I have a system booting an unsigned personal kernel that it was offered to.  I knew it wouldn't work but wanted to see what exactly happens with an "Oh crap, let me fix that before I reboot moment."

UEFI secure boot is a "enable only option"
```
$ sudo fwupdtool set-bios-setting SecureBoot Disable
17:39:40:0249 FuBiosAttrs          KERNEL BUG: thinklmi doesn't export a 'type' attribute
failed to write 7 bytes to 17: Invalid argument
```

A few ideas on how we can approach it:
1. Modify how the attribute works, special case it and mark it `read_only` if it's set to enabled.
2. Don't offer the attribute if it's set to enabled.  It doesn't do the user ANY value, and we measure it in other ways.
3. Prefix the error when trying to set secure boot to disabled to make it more user friendly.

superm1:
> Modify how the attribute works, special case it and mark it read_only if it's set to enabled.

This is my favorite.  https://github.com/fwupd/fwupd/pull/4897

src/fu-util.c
```diff
@@ -3386,6 +3386,54 @@ fu_util_sync_bkc(FuUtilPrivate *priv, gchar **values, GError **error)
...
+			       /* TRANSLATORS: title prefix for the BIOS settings dialog */
+			       _("Configuration Change Suggested"),
+			       fwupd_security_attr_get_title(attr));
+
+	g_string_append(body, fwupd_security_attr_get_description(attr));
+	g_string_append(body, "\n\n");
+	g_string_append_printf(
+	    body,
+	    /* TRANSLATORS: the %1 and %2 are the values, e.g. "True" or "Windows10" */
+	    _("This tool can change the BIOS setting from '%s' to '%s' "
```
superm1, line 3406: I realized that we need to tell them _what_ BIOS setting we're changing.  The attribute title doesn't match the name of the string.

```suggestion
	    /* TRANSLATORS: the %1 is a BIOS setting name.  %2 and %3 are the values, e.g. "True" or "Windows10" */
	    _("This tool can change the BIOS setting '%s' from '%s' to '%s' "
	    "automatically, but it will only be active after restarting the "
	      "computer."),
	    fwupd_bios_attr_get_name(fwupd_security_attr_get_bios_attr_id(attr)),
	    fwupd_security_attr_get_bios_attr_current_value(attr),
	    fwupd_security_attr_get_bios_attr_target_value(attr));
```

src/fu-util.c
```diff
@@ -3386,6 +3386,54 @@ fu_util_sync_bkc(FuUtilPrivate *priv, gchar **values, GError **error)
...
+	      "automatically, but it will only be active after restarting the "
+	      "computer."),
+	    fwupd_security_attr_get_bios_attr_current_value(attr),
+	    fwupd_security_attr_get_bios_attr_target_value(attr));
```
superm1, line 3410: I realized that we need to tell them _what_ BIOS setting we're changing.  The attribute title doesn't match the name of the string.

```suggestion
	    /* TRANSLATORS: the %1 is a BIOS setting name.  %2 and %3 are the values, e.g. "True" or "Windows10" */
	    _("This tool can change the BIOS setting '%s' from '%s' to '%s' "
	    "automatically, but it will only be active after restarting the "
	      "computer."),
	    fwupd_bios_attr_get_name(fwupd_security_attr_get_bios_attr_id(attr)),
	    fwupd_security_attr_get_bios_attr_current_value(attr),
	    fwupd_security_attr_get_bios_attr_target_value(attr));
```

src/fu-util.c
```diff
@@ -3480,6 +3528,27 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 		}
 	}
 
+	/* any things we can fix? */
+	for (guint j = 0; j < attrs->len; j++) {
+		FwupdSecurityAttr *attr = g_ptr_array_index(attrs, j);
+		if (!fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS) &&
+		    fwupd_security_attr_get_bios_attr_id(attr) != NULL &&
+		    fwupd_security_attr_get_bios_attr_current_value(attr) != NULL &&
+		    fwupd_security_attr_get_bios_attr_target_value(attr) != NULL) {
```
superm1, line 3537: Assuming https://github.com/fwupd/fwupd/pull/4897 goes in, I think there should be a library change to not set the target value if the attribute is read only.

superm1: https://github.com/fwupd/fwupd/pull/4905

superm1:
With those other fixes landed, can you rebase this and take a look at my suggestions now?

src/fu-util.c
hughsie: I've added a more comprehensive version, thanks.

hughsie:
> With those other fixes landed, can you rebase this and take a look at my suggestions now?

Done.

superm1 (approved):

PR #4897, opened 2022-08-01
trivial: don't let people try to turn off UEFI secure boot

superm1:
The firmware from both Dell and Lenovo actually blocks this, but the
error message is pretty confusing.

```
$ sudo fwupdtool set-bios-setting SecureBoot Disable
17:39:40:0249 FuBiosAttrs          KERNEL BUG: thinklmi doesn't export a 'type' attribute
failed to write 7 bytes to 17: Invalid argument
```

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-bios-attrs.c
+++ b/libfwupdplugin/fu-bios-attrs.c
@@ -155,6 +155,31 @@ fu_bios_attr_set_current_value(FwupdBiosAttr *attr, GError **error)
 #define LENOVO_READ_ONLY_NEEDLE "[Status:ShowOnly]"
 #define LENOVO_EXCLUDED		"[Excluded from boot order:"
 
+static void
+fu_bios_attr_fixup_read_only(FwupdBiosAttr *attr)
+{
+	if (fwupd_bios_attr_get_kind(attr) == FWUPD_BIOS_ATTR_KIND_ENUMERATION) {
+		struct {
+			const gchar *id;
+			const gchar *value;
+		} read_only_map[] = {{"com.thinklmi.SecureBoot", "Enable"},
+				     {"com.dell-wmi-sysman.SecureBoot", "Enabled"},
+				     {NULL, NULL}};
+		const gchar *id = fwupd_bios_attr_get_id(attr);
+		const gchar *tmp = fwupd_bios_attr_get_current_value(attr);
+		for (guint i = 0; read_only_map[i].id != NULL; i++) {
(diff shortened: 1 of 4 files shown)


CragW (approved):

superm1:
@mrhpearson I notice that on Z13 the `TpmSelection` setting is acting similar.

```
$ sudo fwupdtool set-bios-setting TpmSelection PlutonTPM2.0
failed to write 12 bytes to 17: Invalid argument
```

Is that to be expected and we should make it read only too?  Or is there something special to be done for it?

superm1:
> @mrhpearson I notice that on Z13 the `TpmSelection` setting is acting similar.
> 
> ```
> $ sudo fwupdtool set-bios-setting TpmSelection PlutonTPM2.0
> failed to write 12 bytes to 17: Invalid argument
> ```
> 
> Is that to be expected and we should make it read only too? Or is there something special to be done for it?

same thing with 
```
$ sudo fwupdtool set-bios-setting Allow3rdPartyUEFICA Enable
failed to write 6 bytes to 17: Invalid argument
```
So I think we need a list of attributes that are expected to read only and under what circumstances.

libfwupdplugin/fu-bios-attrs.c
```diff
@@ -155,6 +155,25 @@ fu_bios_attr_set_current_value(FwupdBiosAttr *attr, GError **error)
...
 #define LENOVO_EXCLUDED		"[Excluded from boot order:"
 
+static void
+fu_bios_attr_fixup_read_only(FwupdBiosAttr *attr)
+{
+	if (fwupd_bios_attr_get_kind(attr) == FWUPD_BIOS_ATTR_KIND_ENUMERATION) {
+		const gchar *id = fwupd_bios_attr_get_id(attr);
+		const gchar *tmp = fwupd_bios_attr_get_current_value(attr);
+		if (g_strcmp0(id, "com.thinklmi.SecureBoot") == 0 &&
+		    g_strcmp0(tmp, "Enable") == 0) {
```
hughsie, line 165: I honestly think we need a struct here that's being looped; there's going to be quite a few of these I suspect.

superm1: Can you explain better?  I don't follow.

hughsie: I thought of something like https://github.com/fwupd/fwupd/blob/main/src/fu-udev-backend.c#L31

hughsie (approved):

PR #4904, opened 2022-08-03
fu-engine: Evalute HSI number when FWUPD_HOST_EMULATE is set in the VM

smallorange:
When fwupd runs in VM with FWUPD_MACHINE_KIND and FWUPD_HOST_EMULATE set,
the HSI value will not be computed since the FU_SMBIOS_CHASSIS_KIND is
OTHER. Consequence, a "HSI-INVALID:chassis[0x01]" is returned through
dbus. This should ensure when fwupd runs in a VM with FWUPD_HOST_EMULATE,
the HSI value can be properly returned according to the mockup data.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -6344,6 +6344,11 @@ fu_engine_attrs_calculate_hsi_for_chassis(FuEngine *self)
 	guint val =
 	    fu_context_get_smbios_integer(self->ctx, FU_SMBIOS_STRUCTURE_TYPE_CHASSIS, 0x05);
 
+	if (g_getenv("FWUPD_HOST_EMULATE") != NULL) {
+		return fu_security_attrs_calculate_hsi(self->host_security_attrs,
+						       FU_SECURITY_ATTRS_FLAG_ADD_VERSION);
+	}
+
 	switch (val) {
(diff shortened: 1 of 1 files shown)


superm1 (commented):
When doing host emulate shouldn't the chassis type be loaded as well from the json data?

I would think you want to be able to emulate different chassis types.

smallorange:
Maybe both "MACHINE_KIND" and "CHASSIS_TYPE" can be set in a JSON when "FWUPD_HOST_EMULATE" is set.

superm1:
If you need both my thought is:
1. we should publish both the machine kind and the chassis type as client properties.
2. the emulation JSON generation script should capture these properties
3. When the emulation json generation is loaded, set up those properties as environment variables, or pass them in otherwise.

If you want to take a stab at that, go ahead.
Otherwise for now you can use `FWUPD_MACHINE_KIND` environment variable and `/etc/fwupd/daemon.conf` key `EnclosureKind` to override it and unblock you.

smallorange:
I've filed a new pull request and planned to close it :)

PR #4907, opened 2022-08-04
trivial: map common positive or negative keys to possible enumeration…

superm1:
… values

Dell and Lenovo use Enable or Enabled and Disable or Disabled which is confusing
to an end user.

Set up some heuristics to map positive values and negative values when passed
into the client.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-bios-attr.c
+++ b/libfwupd/fwupd-bios-attr.c
@@ -326,6 +326,100 @@ fwupd_bios_attr_set_description(FwupdBiosAttr *self, const gchar *description)
 	priv->description = g_strdup(description);
 }
 
+/* determine if key is supposed to be positive */
+static gboolean
+fu_bios_attr_key_is_positive(const gchar *key)
+{
+	if (g_strrstr(key, "enable"))
+		return TRUE;
+	if (g_strcmp0(key, "true") == 0)
+		return TRUE;
+	if (g_strcmp0(key, "1") == 0)
+		return TRUE;
+	if (g_strcmp0(key, "on") == 0)
+		return TRUE;
+	return FALSE;
+}
+
+/* determine if key is supposed to be negative */
+static gboolean
+fu_bios_attr_key_is_negative(const gchar *key)
+{
+	if (g_strrstr(key, "disable"))
+		return TRUE;
+	if (g_strcmp0(key, "false") == 0)
+		return TRUE;
+	if (g_strcmp0(key, "0") == 0)
+		return TRUE;
+	if (g_strcmp0(key, "off") == 0)
+		return TRUE;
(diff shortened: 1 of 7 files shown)


CragW (approved):

libfwupd/fwupd-bios-attr.c
```diff
@@ -326,6 +326,94 @@ fwupd_bios_attr_set_description(FwupdBiosAttr *self, const gchar *description)
...
+		return TRUE;
+	return FALSE;
+}
+
+/**
+ * fwupd_bios_attr_map_possible_value:
+ * @self: a #FwupdBiosAttr
+ * @key: the string to try to map
+ * @error: (nullable): optional return location for an error
+ *
```
hughsie, line 364: i think you need some body here, i.e. what is the method *for* and why would you use it?

libfwupd/fwupd-bios-attr.c
```diff
@@ -326,6 +326,94 @@ fwupd_bios_attr_set_description(FwupdBiosAttr *self, const gchar *description)
...
+	for (guint i = 0; i < priv->possible_values->len; i++) {
+		const gchar *possible = g_ptr_array_index(priv->possible_values, i);
+		g_autofree gchar *lower_possible = g_utf8_strdown(possible, -1);
+		gboolean positive_possible;
+		gboolean negative_possible;
+
+		/* perfect match */
+		if (g_strcmp0(lower_possible, lower_key) == 0)
+			return possible;
+		/* fuzzy match */
```
hughsie, line 402: shouldn't we do all the perfect matches in pass-1, then the fuzzy ones in a 2nd pass?

src/fu-engine.c
```diff
@@ -767,7 +767,7 @@ fu_engine_update_bios_attr(FwupdBiosAttr *attr, const gchar *value, GError **err
  * error messages
  */
 static gboolean
-fu_engine_validate_bios_attr_input(FwupdBiosAttr *attr, const gchar *value, GError **error)
+fu_engine_validate_bios_attr_input(FwupdBiosAttr *attr, const gchar **value, GError **error)
```
hughsie, line 770: don't we want to check that **value and *value are non-NULL?

libfwupd/fwupd-bios-attr.c
superm1: Why? Just to have less string checking?

superm1: Yeah I can't think of one.

hughsie (approved):

hughsie (approved):

PR #4912, opened 2022-08-05
Allow running get-bios-settings without root or PK

superm1:
The only information that is secret is the `current_value`.
Augment the d-bus call to determine whether the caller needs this
information.

* If `fwupdmgr` is launched as root it will be provided.
* If `fwupdmgr` is launched with `--authenticate` it will be requested
  and PK will be engaged.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-bios-attr.c
+++ b/libfwupd/fwupd-bios-attr.c
@@ -587,9 +587,22 @@ fwupd_bios_attr_set_current_value(FwupdBiosAttr *self, const gchar *value)
 	priv->current_value = g_strdup(value);
 }
 
+static gboolean
+fwupd_bios_attr_trusted(FwupdBiosAttr *self, gboolean trusted)
+{
+	g_return_val_if_fail(FWUPD_IS_BIOS_ATTR(self), FALSE);
+
+	if (trusted)
+		return TRUE;
+	if (g_strcmp0(fwupd_bios_attr_get_name(self), "pending_reboot") == 0)
+		return TRUE;
+	return FALSE;
+}
+
 /**
  * fwupd_bios_attr_to_variant:
  * @self: a #FwupdBiosAttr
+ * @trusted: whether the caller should receive trusted values
  *
  * Serialize the bios attribute.
  *
@@ -598,7 +611,7 @@ fwupd_bios_attr_set_current_value(FwupdBiosAttr *self, const gchar *value)
  * Since: 1.8.4
  **/
 GVariant *
-fwupd_bios_attr_to_variant(FwupdBiosAttr *self)
+fwupd_bios_attr_to_variant(FwupdBiosAttr *self, gboolean trusted)
 {
 	FwupdBiosAttrPrivate *priv = GET_PRIVATE(self);
 	GVariantBuilder builder;
@@ -638,10 +651,12 @@ fwupd_bios_attr_to_variant(FwupdBiosAttr *self)
 			      "{sv}",
 			      FWUPD_RESULT_KEY_BIOS_ATTR_READ_ONLY,
 			      g_variant_new_boolean(priv->read_only));
-	g_variant_builder_add(&builder,
-			      "{sv}",
-			      FWUPD_RESULT_KEY_BIOS_ATTR_CURRENT_VALUE,
-			      g_variant_new_string(priv->current_value));
+	if (fwupd_bios_attr_trusted(self, trusted)) {
+		g_variant_builder_add(&builder,
+				      "{sv}",
+				      FWUPD_RESULT_KEY_BIOS_ATTR_CURRENT_VALUE,
+				      g_variant_new_string(priv->current_value));
+	}
 	if (priv->kind == FWUPD_BIOS_ATTR_KIND_INTEGER ||
 	    priv->kind == FWUPD_BIOS_ATTR_KIND_STRING) {
 		g_variant_builder_add(&builder,
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -75,19 +75,21 @@ typedef enum {
  * @FWUPD_FEATURE_FLAG_FDE_WARNING:		Can warn about full disk encryption
  * @FWUPD_FEATURE_FLAG_COMMUNITY_TEXT:		Can show information about community supported
  * @FWUPD_FEATURE_FLAG_SHOW_PROBLEMS:		Can show problems when getting the update list
+ * @FWUPD_FEATURE_FLAG_ALLOW_AUTHENTICATION:	Can authenticate with PolicyKit for requests
  *
  * The flags to the feature capabilities of the front-end client.
  **/
 typedef enum {
-	FWUPD_FEATURE_FLAG_NONE = 0,		    /* Since: 1.4.5 */
-	FWUPD_FEATURE_FLAG_CAN_REPORT = 1 << 0,	    /* Since: 1.4.5 */
-	FWUPD_FEATURE_FLAG_DETACH_ACTION = 1 << 1,  /* Since: 1.4.5 */
-	FWUPD_FEATURE_FLAG_UPDATE_ACTION = 1 << 2,  /* Since: 1.4.5 */
-	FWUPD_FEATURE_FLAG_SWITCH_BRANCH = 1 << 3,  /* Since: 1.5.0 */
-	FWUPD_FEATURE_FLAG_REQUESTS = 1 << 4,	    /* Since: 1.6.2 */
-	FWUPD_FEATURE_FLAG_FDE_WARNING = 1 << 5,    /* Since: 1.7.1 */
-	FWUPD_FEATURE_FLAG_COMMUNITY_TEXT = 1 << 6, /* Since: 1.7.5 */
-	FWUPD_FEATURE_FLAG_SHOW_PROBLEMS = 1 << 7,  /* Since: 1.8.1 */
+	FWUPD_FEATURE_FLAG_NONE = 0,			  /* Since: 1.4.5 */
+	FWUPD_FEATURE_FLAG_CAN_REPORT = 1 << 0,		  /* Since: 1.4.5 */
+	FWUPD_FEATURE_FLAG_DETACH_ACTION = 1 << 1,	  /* Since: 1.4.5 */
+	FWUPD_FEATURE_FLAG_UPDATE_ACTION = 1 << 2,	  /* Since: 1.4.5 */
+	FWUPD_FEATURE_FLAG_SWITCH_BRANCH = 1 << 3,	  /* Since: 1.5.0 */
+	FWUPD_FEATURE_FLAG_REQUESTS = 1 << 4,		  /* Since: 1.6.2 */
+	FWUPD_FEATURE_FLAG_FDE_WARNING = 1 << 5,	  /* Since: 1.7.1 */
+	FWUPD_FEATURE_FLAG_COMMUNITY_TEXT = 1 << 6,	  /* Since: 1.7.5 */
+	FWUPD_FEATURE_FLAG_SHOW_PROBLEMS = 1 << 7,	  /* Since: 1.8.1 */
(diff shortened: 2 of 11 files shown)


superm1:
Sample output:

```
$ fwupdmgr get-bios-setting pending_reboot MmioAbove4GLimit

pending_reboot:
  Current Value:        1
  Description:          Settings will apply after system reboots
  Read Only:            False


MmioAbove4GLimit:
  Setting type:         Enumeration
  Current Value:        Run with '--authenticate' to see
  Description:          MmioAbove4GLimit
  Read Only:            False
  Possible Values:
    0:                  Auto
    1:                  40
    2:                  42
    3:                  44
    4:                  46

$ fwupdmgr get-bios-setting pending_reboot MmioAbove4GLimit --authenticate
Authenticating…          [ -                                     ]==== AUTHENTICATING FOR org.freedesktop.fwupd.get-bios-attributes ===
Authentication is required to read BIOS attributes
Authenticating as: Mario Limonciello (supermario)
Authenticating…          [                       \               ]
==== AUTHENTICATION COMPLETE ===
Authenticating…          [                       \               ]

pending_reboot:
  Current Value:        1
  Description:          Settings will apply after system reboots
  Read Only:            False


MmioAbove4GLimit:
  Setting type:         Enumeration
  Current Value:        Auto
  Description:          MmioAbove4GLimit
  Read Only:            False
  Possible Values:
    0:                  Auto
    1:                  40
    2:                  42
    3:                  44
    4:                  46
```

I envision this as GNOME control center could have a `Firmware settings` page which will show you all the stuff your system supports.
You can see this without authenticating.
When you hit the unlock button, you authenticate with PK and then all the settings are fetched with `current_value`

Thinking through that use case we might also want to make `set-bios-setting` into `set-bios-settings`.  That is; make it accept arrays of settings pairs.  This would allow you to change a bunch of things in a GUI at one time instead.  Otherwise it's going to be pretty painful to do this in a GUI one at a time.

superm1:
The other thing this opens up is bash-completion for BIOS settings!
Example:

```
$ fwupdmgr set-
set-approved-firmware  set-bios-setting
130 $ fwupdmgr set-bios-setting
Display all 147 possibilities? (y or n)
130 $ fwupdmgr set-bios-setting W
WakeonLAN                  WakeUponAlarm              WindowsUEFIFirmwareUpdate
130 $ fwupdmgr set-bios-setting WindowsUEFIFirmwareUpdate
--allow-branch-switch  --assume-yes           Enable                 --ipfs                 --no-metadata-check    --no-safety-check      --show-all
--allow-older          Disable                --filter               --json                 --no-reboot-check      --no-unreported-check  --sign
--allow-reinstall      --disable-ssl-strict   --force                --no-history           --no-remote-check      --offline              --verbose
130 $ fwupdmgr set-bios-setting WindowsUEFIFirmwareUpdate Enable
```

libfwupd/fwupd-bios-attr.c
```diff
@@ -496,6 +496,7 @@ fwupd_bios_attr_set_current_value(FwupdBiosAttr *self, const gchar *value)
 /**
  * fwupd_bios_attr_to_variant:
  * @self: a #FwupdBiosAttr
+ * @trusted: whether the caller should receive trusted values
```
hughsie, line 499: we used FwupdDeviceFlags for this in other places, but maybe that's wrong here. What you have is probably fine.

libfwupd/fwupd-client-sync.c
```diff
@@ -958,6 +958,7 @@ fwupd_client_get_bios_attrs_cb(GObject *source, GAsyncResult *res, gpointer user
 /**
  * fwupd_client_get_bios_attrs:
  * @self: a #FwupdClient
+ * @authenticate: Whether to use PolicyKit to authorize
```
hughsie, line 961: this seems like the tail wagging the dog. Can't we use the interactive hint from the client?

libfwupd/fwupd-self-test.c
```diff
@@ -1468,6 +1470,25 @@ fwupd_bios_attrs_func(void)
 	ret = fu_test_compare_lines(str3, str1, &error);
 	g_assert_no_error(error);
 	g_assert_true(ret);
+
+	/* Make sure we filter CurrentValue if not trusted */
```
hughsie, line 1474: lowercase `Make`

src/fu-daemon.c
```diff
@@ -1990,33 +1990,43 @@ fu_daemon_daemon_method_call(GDBusConnection *connection,
 		return;
 	}
 	if (g_strcmp0(method_name, "GetBiosAttrs") == 0) {
-		g_autoptr(FuMainAuthHelper) helper = NULL;
+		g_autoptr(FuBiosAttrs) attrs = NULL;
+		gboolean authenticate = FALSE;
+
+		g_variant_get(parameters, "(b)", &authenticate);
```
hughsie, line 1996: This doesn't feel right; I kinda thought this kind of thing could be done with FwupdFeatureFlags; maybe using `FWUPD_FEATURE_FLAG_REQUESTS` -- or maybe adding a `FWUPD_FEATURE_FLAG_NO_AUTH` or `FWUPD_FEATURE_FLAG_ANON` type thing.

src/fu-util.c
```diff
@@ -4094,6 +4096,14 @@ main(int argc, char *argv[])
 					 /* TRANSLATORS: command line option */
 					 N_("Output in JSON format"),
 					 NULL},
+					{"authenticate",
+					 '\0',
+					 0,
+					 G_OPTION_ARG_NONE,
+					 &priv->authenticate,
```
hughsie, line 4103: I think this would be more natural the other way around, e.g. authenticate by default, but "opt out" for the case where you don't want the polkit auth.

libfwupd/fwupd-client-sync.c
superm1: Not if we want bash completion to work. That's an interactive terminal usage.

hughsie: Ahh, but not interactive as in "vt active" more a fwupd client hint.

superm1: So like a feature flag to indicate whether to use PK for a command?

hughsie: Is that sane? It *feels* per-client rather than per-request.

superm1: It's not per client though. You will want to have the ability to do it for fwupdmgr both ways. I also think in the GUI tools you would need to have the ability to get some information before you authenticate.

None of the gnome panels require you to unlock the page just to view it. It's only once you want to interact with the page you authenticate.

hughsie: > You will want to have the ability to do it for fwupdmgr both ways

Ya, maybe. I do think we're possibly painting ourselves into a corner with just a boolean tho. Maybe a `FwupdClientGetBiosAttrsFlags` guint64?

superm1: How would you feel about the client calling set feature flags multiple times?  Then thinking through the GUI case there can be a set feature flags call that says `FWUPD_FEATURE_FLAG_CAN_AUTHENTICATE` (or if you can think of a better name).  And then we can drop having an argument all together on this.  GUI (and bash completion case) can avoid setting that.

hughsie: > the client calling set feature flags multiple times

I think that works fine now, no?

> FWUPD_FEATURE_FLAG_CAN_AUTHENTICATE

I think it needs to be the other way around for compat; old gnome-software won't be setting that flag, and if any of the other methods get the same kind of filtering in the future then we don't want to break API. e.g. FWUPD_FEATURE_FLAG_NO_AUTHENTICATE

hughsie: Maybe that's not true; maybe we can fix it for old clients either-way.

src/fu-util.c
superm1: swapped it around

src/fu-daemon.c
superm1: added new flag

libfwupd/fwupd-client-sync.c
superm1: OK, take a look at what I've done now.  I agree this generally scales better.

src/fu-daemon.c
```diff
@@ -1990,33 +1990,43 @@ fu_daemon_daemon_method_call(GDBusConnection *connection,
 		return;
 	}
 	if (g_strcmp0(method_name, "GetBiosAttrs") == 0) {
-		g_autoptr(FuMainAuthHelper) helper = NULL;
+		g_autoptr(FuBiosAttrs) attrs = NULL;
```
hughsie, line 1993: i think that can go inside the {} scope

src/fu-util.c
```diff
@@ -3917,184 +3918,193 @@ main(int argc, char *argv[])
...
+	     /* TRANSLATORS: command line option */
+	     N_("Output in JSON format"),
+	     NULL},
+	    {"no-authenticate",
+	     '\0',
+	     0,
+	     G_OPTION_ARG_NONE,
+	     &no_authenticate,
+	     /* TRANSLATORS: command line option */
+	     N_("Don't authenticate the transaction (less details will be exposed)"),
```
hughsie, line 4105: `less detail may be shown` perhaps?

libfwupd/fwupd-enums.h
```diff
@@ -75,6 +75,7 @@ typedef enum {
  * @FWUPD_FEATURE_FLAG_FDE_WARNING:		Can warn about full disk encryption
  * @FWUPD_FEATURE_FLAG_COMMUNITY_TEXT:		Can show information about community supported
  * @FWUPD_FEATURE_FLAG_SHOW_PROBLEMS:		Can show problems when getting the update list
+ * @FWUPD_FEATURE_FLAG_AUTHENTICATE:		Can authenticate with PolicyKit for this request
```
hughsie, line 78: maybe ALLOW_AUTHENTICATION or something like that?

src/fu-daemon.c
```diff
@@ -1990,33 +1990,43 @@ fu_daemon_daemon_method_call(GDBusConnection *connection,
...
+		gboolean authenticate = fu_engine_request_get_feature_flags(request) &
+					FWUPD_FEATURE_FLAG_ALLOW_AUTHENTICATION;
+
+		g_debug("Called %s", method_name);
+		if (!authenticate) {
+			g_autoptr(FuBiosAttrs) attrs =
+			    fu_context_get_bios_attrs(fu_engine_get_context(self->engine));
+			val = fu_bios_attrs_to_variant(attrs,
+						       fu_engine_request_get_device_flags(request) &
+							   FWUPD_DEVICE_FLAG_TRUSTED);
```
hughsie, line 2002: I'm a bit confused with this; is this ever going to be true?

superm1: I think you're right.  This is a vestige from an earlier version of this patch that doesn't make sense anymore.

libfwupdplugin/fu-bios-attrs.c
```diff
@@ -493,9 +493,22 @@ fu_bios_attrs_get_pending_reboot(FuBiosAttrs *self, gboolean *result, GError **e
 	return TRUE;
 }
 
+static gboolean
+fu_bios_attr_trusted(FwupdBiosAttr *self, gboolean trusted)
```
hughsie, line 497: how come this is here and not in `libfwupd/fwupd-bios-attr.c`? I think it's a bit odd to have `fu_bios_attrs_to_variant` and `fwupd_bios_attr_to_variant` returning different results when !trusted.

hughsie (commented):
One nit then LGTM.

libfwupdplugin/fu-bios-attrs.c
superm1: moved it

hughsie (approved):

Issue #4915, opened 2022-08-10
Cros-ec plugin: add stay-in-ro command if EC is already in bootloader


the-sneaky-fox:
If EC is booting from reset when `fwupd` is invoked - it is necessary to issue a stay-in-ro command. 
Otherwise, EC will continue booting and might jump to RW in the middle of the firmware update, resulting in an unexpected behavior.

Implementation can be done either by calling the command from the `detach` function, or by setting the `FU_CROS_EC_USB_DEVICE_FLAG_REBOOTING_TO_RO` private flag.

the-sneaky-fox:
Fixed by pull request #4916

PR #4916, opened 2022-08-10
cros-ec: adding stay-in-ro command if EC is in bootloader

the-sneaky-fox:
Type of pull request:

- [ ] Code fix
- [x] Feature #4915 
- [ ] Documentation

### DIFF
--- a/plugins/cros-ec/fu-cros-ec-usb-device.c
+++ b/plugins/cros-ec/fu-cros-ec-usb-device.c
@@ -765,6 +765,29 @@ fu_cros_ec_usb_device_jump_to_rw(FuDevice *device)
 }
 
 static gboolean
+fu_cros_ec_usb_device_stay_in_ro(FuDevice *device, GError **error)
+{
+	gsize response_size = 1;
+	guint8 response;
+	guint16 subcommand = UPDATE_EXTRA_CMD_STAY_IN_RO;
+	guint8 command_body[2]; /* Max command body size. */
+	gsize command_body_size = 0;
+
+	if (!fu_cros_ec_usb_device_send_subcommand(device,
+						   subcommand,
+						   command_body,
+						   command_body_size,
+						   &response,
+						   &response_size,
+						   FALSE,
+						   error)) {
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static gboolean
 fu_cros_ec_usb_device_write_firmware(FuDevice *device,
 				     FuFirmware *firmware,
 				     FuProgress *progress,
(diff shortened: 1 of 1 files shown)


plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -778,22 +801,10 @@ fu_cros_ec_usb_device_write_firmware(FuDevice *device,
...
 		fu_device_remove_private_flag(device, FU_CROS_EC_USB_DEVICE_FLAG_REBOOTING_TO_RO);
-		if (!fu_cros_ec_usb_device_send_subcommand(device,
-							   subcommand,
-							   command_body,
-							   command_body_size,
-							   &response,
-							   &response_size,
-							   FALSE,
-							   error)) {
+		if (!fu_cros_ec_usb_device_stay_in_ro(device)){
```
hughsie, line 807: you need to pass `error` here too, otherwise you'll have nothing to prefix.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -764,6 +764,29 @@ fu_cros_ec_usb_device_jump_to_rw(FuDevice *device)
...
+	gsize command_body_size = 0;
+
+	if (!fu_cros_ec_usb_device_send_subcommand(device,
+						   subcommand,
+						   command_body,
+						   command_body_size,
+						   &response,
+						   &response_size,
+						   FALSE,
+						   NULL)) {
```
hughsie, line 783: needs `error` here.

plugins/cros-ec/fu-cros-ec-usb-device.c
```diff
@@ -764,6 +764,29 @@ fu_cros_ec_usb_device_jump_to_rw(FuDevice *device)
 	return TRUE;
 }
 
+static gboolean
+fu_cros_ec_usb_device_stay_in_ro(FuDevice *device, GError **error)
+{
+	gsize response_size = 1;
+	guint8 response;
+	guint16 subcommand = UPDATE_EXTRA_CMD_STAY_IN_RO;
```
superm1, line 772: does this cmd exist in all FW versions?  or do you need some guards for it?

bleungatchromium (approved):
This looks good to me.

plugins/cros-ec/fu-cros-ec-usb-device.c
superm1: Right; that was my concern if it's called from a second location if any extra guards are going to be needed from that second location calling.

PR #4919, opened 2022-08-11
Add instance IDs that include the kernel driver

hughsie:
Some interesting i2c devices have a generic name, e.g. 'Synopsys'.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -517,6 +517,12 @@ fu_udev_device_probe(FuDevice *device, GError **error)
 	fu_device_add_instance_str(device, "DRIVER", priv->driver);
 	fu_device_build_instance_id_quirk(device, NULL, subsystem, "DRIVER", NULL);
 
+	/* add the modalias */
+	fu_device_add_instance_strsafe(device,
+				       "MODALIAS",
+				       g_udev_device_get_property(priv->udev_device, "MODALIAS"));
+	fu_device_build_instance_id_quirk(device, NULL, subsystem, "MODALIAS", NULL);
+
 	/* add subsystem to match in plugins */
 	if (subsystem != NULL) {
 		fu_device_add_instance_id_full(device,
--- a/plugins/elantp/elantp.quirk
+++ b/plugins/elantp/elantp.quirk
@@ -26,6 +26,16 @@ Flags = elantp-recovery
 [513cde3d-d939-59bd-a634-5c1645ebb93b]
 Flags = elantp-recovery
 
+[I2C\MODALIAS_acpi:ELAN0000:]
+Plugin = elantp
+GType = FuElantpI2cDevice
+ElantpI2cTargetAddress = 0x15
+
+[I2C\MODALIAS_of:NtrackpadT(null)Celan,ekth3000]
+Plugin = elantp
+GType = FuElantpI2cDevice
+ElantpI2cTargetAddress = 0x15
+
 # recovery device
 [I2C\NAME_Synopsys-DesignWare-I2C-adapter]
 Plugin = elantp


superm1:
I think I'd like to see what this is used for before it's added.

hughsie:
> xxx

Before:


    01:49:47:0010 FuEngine             udev added FuI2cDevice:
      Guid:                 32729c6c-d8b8-5102-9003-8042e22146c1 ? I2C\NAME_Synopsys-DesignWare-I2C-adapter ?
      Flags:                internal
      PhysicalId:           DEVPATH=/devices/platform/AMD0010:01/i2c-1
      BackendId:            /sys/devices/platform/AMD0010:01/i2c-1/i2c-dev/i2c-1
      PossiblePlugin:       elantp
      BusNumber:            0x1


After (guessed, based on vendor report):

    01:49:47:0010 FuEngine             udev added FuI2cDevice:
      Guid:                 32729c6c-d8b8-5102-9003-8042e22146c1 ? I2C\NAME_Synopsys-DesignWare-I2C-adapter ?
      Guid:                 ???????????????????????????????????? ? I2C\DRIVER_elan_i2c ?
      Guid:                 ???????????????????????????????????? ? I2C\NAME_Synopsys-DesignWare-I2C-adapter&DRIVER_elan_i2c ?
      Flags:                internal
      PhysicalId:           DEVPATH=/devices/platform/AMD0010:01/i2c-1
      BackendId:            /sys/devices/platform/AMD0010:01/i2c-1/i2c-dev/i2c-1
      PossiblePlugin:       elantp
      BusNumber:            0x1

this lets the elantp bind to things that have already matched in the kernel. _I think_, 100% untested.

superm1:
I don't think it's going to work like you expect...
I think a safer extra GUID would be to use the udev key `MODALIAS`

hughsie:
I've asked the vendor for other ideas.

josh-chen-elan:
I'm Josh from the ELAN Touchpad team.

I list the devices on the x86_64 and ARM platforms separately through "udevadm info -e", I can see the device information I wish to match as follows:
x86_64:
```P: /devices/platform/AMD0010:00/i2c-0/i2c-ELAN0000:00
E: DEVPATH=/devices/platform/AMD0010:00/i2c-0/i2c-ELAN0000:00
E: DRIVER=elan_i2c
E: ID_VENDOR_FROM_DATABASE=ELAD srl
E: MODALIAS=acpi:ELAN0000:
E: SUBSYSTEM=i2c
E: USEC_INITIALIZED=6938697
```

ARM:
```P: /devices/platform/soc/11009000.i2c/i2c-2/2-0015
E: DEVPATH=/devices/platform/soc/11009000.i2c/i2c-2/2-0015
E: DRIVER=elan_i2c
E: MODALIAS=of:NtrackpadT<NULL>Celan,ekth3000
E: OF_COMPATIBLE_0=elan,ekth3000
E: OF_COMPATIBLE_N=1
E: OF_FULLNAME=/soc/i2c@11009000/trackpad@15
E: OF_NAME=trackpad
E: SUBSYSTEM=i2c
```

Their common attribute is ***DRIVER=elan_i2c***, if this method is not suitable, is it possible to use **[I2C\MODALIAS_ELAN0000]** and **[I2C\MODALIAS_ekth3000]** in elantp.quirk to match the device?

superm1:
Do both use the same firmware across architecture? Or there is one firmware for each ?

josh-chen-elan:
> Do both use the same firmware across architecture? Or there is one firmware for each ?

ELAN touchpads can be used across architecture, using the same firmware and update process

superm1:
Right but so the question here is if that is universal. Would you always want touchpad firmware to target multiple machines across architectures?

Or would you want to have the ability to target more narrowly go only apply to modalises from one architecture without applying to others?

I suspect we want to have the ability to do both across both architectures and target them individually by making guid for both cases.

josh-chen-elan:
Sorry, I can not get exactly what you mean.
Please let me explain the requirements first.

The current elantp plugin is for HID touchpads. However, the ones used on Chromebooks (both x86 and ARM) are I2C non-HID touchpads.

The clearest way to distinguish between HID and non-HID touchpads is to use the "DRIVER=elan_i2c" attribute to represent non-HID touchpads.  It looks like "MODALIAS=ELAN0000" (on x86) and "MODALIAS=ekth3000" (on ARM) can also be used to represent non-HID touchpads.

We expect to add a new flag in elantp.quirk while matching a non-HID touchpad, and then use this flag to generate a GUID which is different with the HID touchpad in FuElantpI2cDevice.

Either touchpad firmware can be used on x86 and ARM platforms, but HID touchpads can only be updated by HID firmware, and non-HID touchpads can only be updated by non-HID firmware.

josh-chen-elan:
By the way, [I2C\NAME_Synopsys-DesignWare-I2C-adapter] in elantp.quirk can not work on ARM platform. The I2C host names seems not to be consistent all the time.

As a result, using **"DRIVER=elan_i2c"** or **the combination of "MODALIAS=ELAN0000" & "MODALIAS=ekth3000"** to match is more suitable.

superm1:
Presumably this includes the prefix "acpi:". Is that preferred? I think yes

hughsie:
I wasn't sure. Is that the subsystem or something else?

superm1:
> I wasn't sure. Is that the subsystem or something else?

Modaliases are used by the kernel for loading driver automatically.  So yes "acpi" prefix means it's acpi subsystem, of means it's from devicetree.

libfwupdplugin/fu-i2c-device.c
```diff
@@ -128,6 +128,12 @@ fu_i2c_device_probe(FuDevice *device, GError **error)
 	if (!fu_device_build_instance_id(device, error, "I2C", "NAME", NULL))
 		return FALSE;
 
+	/* optional */
+	fu_device_add_instance_strsafe(device,
+				       "MODALIAS",
+				       g_udev_device_get_property(udev_device, "MODALIAS"));
```
superm1, line 134: You'll need to update README.md too

josh-chen-elan:
This patch does not seem to work,
even if I add [I2C\MODALIAS_acpi:ELAN0000:] & [I2C\MODALIAS_acpi:ELAN0000] & [I2C\MODALIAS_ELAN0000] in elantp.quirk, it cannot be matched.

I don't know the order in which fu_i2c_device_probe() and fu_udev_device_probe() are executed is first, but I found that the reason why the previous attempts would be matched (add [I2C\DRIVER_elan_i2c] in elantp.quirk) was because the Instance ID was added by fu_udev_device_probe(), rather than the previous patch.
 https://github.com/fwupd/fwupd/blob/main/libfwupdplugin/fu-udev-device.c#L518

Should we move the procedure of building Instance ID with MODALIAS to FuUdevDevice?

hughsie:
Can you use `FWUPD_PROBE_VERBOSE=1` -- the strsafe is probably converting the `:` to `_`

josh-chen-elan:
The log I executed on the ARM platform is as follows, and the MODALIAS is NULL:
```
03:17:13:0380 FuEngine             udev added FuI2cDevice:
  Flags:                none
  BackendId:            /sys/devices/platform/soc/11009000.i2c/i2c-2/i2c-dev/i2c-2
  BusNumber:            0x0

03:17:13:0382 GLib                 g_regex_new: assertion 'error == NULL || *error == NULL' failed
03:17:13:0382 FuEngine             failed to probe device /sys/devices/platform/soc/11009000.i2c/i2c-2/i2c-dev/i2c-2: no value for MODALIAS
```

I think FuI2cDevice only handles "subsystem=i2c-dev", and "subsystem=i2c" to which "/devices/platform/soc/11009000.i2c/i2c-2/2-0015" belongs is handled by FuUdevDevice

josh-chen-elan:
When enumerating the "/devices/platform/soc/11009000.i2c/i2c-2/2-0015" device, the relevant log of FuI2cDevice cannot be found:
```
03:17:13:0160 FuEngine             udev added FuUdevDevice:
  Flags:                none
  BackendId:            /sys/devices/platform/soc/11009000.i2c/i2c-2/2-0015
  SysfsPath:            /sys/devices/platform/soc/11009000.i2c/i2c-2/2-0015
  Subsystem:            i2c
  Driver:               elan_i2c

03:17:13:0161 FuEngine             udev added FuUdevDevice:
  Flags:                internal
  BackendId:            /sys/devices/platform/soc/11009000.i2c/i2c-2/2-0015
  PossiblePlugin:       parade_lspcon
  PossiblePlugin:       realtek_mst
  SysfsPath:            /sys/devices/platform/soc/11009000.i2c/i2c-2/2-0015
  Subsystem:            i2c
  Driver:               elan_i2c

03:17:13:0162 FuEngine             parade_lspcon ignoring: device name (null) is not supported by this plugin
03:17:13:0163 FuEngine             realtek_mst ignoring: device name (null) is not supported
```

superm1:
I tend to think you're right, it's probably better to set a modalias key for all udev devices in fu-udev-device.c.

josh-chen-elan:
Glad to hear that.

I have tried adding this patch here: https://github.com/fwupd/fwupd/blob/main/libfwupdplugin/fu-udev-device.c#L519
The elantp plugin can be matched when I add [I2C\MODALIAS_acpi:ELAN0000:] and [I2C\MODALIAS_of:NtrackpadT(null)Celan,ekth3000] in elantp.quirk

In this condition, I'm trying to modify FuI2cDevice to support Chromebook's non-HID touchpads.

superm1:
OK thanks!  I tried to land that in the PR directly.  Can you make sure I got it right?  (i'm mostly worried about the ACPI one ending in a `:`)

plugins/elantp/elantp.quirk
```diff
@@ -26,6 +26,16 @@ Flags = elantp-recovery
 [513cde3d-d939-59bd-a634-5c1645ebb93b]
 Flags = elantp-recovery
 
+[I2C\MODALIAS_acpi:ELAN0000:]
```
superm1, line 29: need to double check the ending `:`

libfwupdplugin/fu-udev-device.c
```diff
@@ -517,6 +517,12 @@ fu_udev_device_probe(FuDevice *device, GError **error)
 	fu_device_add_instance_str(device, "DRIVER", priv->driver);
 	fu_device_build_instance_id_quirk(device, NULL, subsystem, "DRIVER", NULL);
 
+	/* add the modalias */
+	fu_device_add_instance_strsafe(device,
+				       "MODALIAS",
+				       g_udev_device_get_property(priv->udev_device, "MODALIAS"));
+	fu_device_build_instance_id(device, NULL, subsystem, "MODALIAS", NULL);
```
josh-chen-elan, line 524: I use fu_device_build_instance_id_quirk().
If I use fu_device_build_instance_id() here, it always fails to compile in my chromeOS development environment.

superm1: Fixed.
Is the quirk file correct?

plugins/elantp/elantp.quirk
josh-chen-elan: > need to double check the ending `:`

Yes, the `:` at the end is necessary after the test result

superm1: thanks!

libfwupdplugin/fu-udev-device.c
josh-chen-elan: > Fixed. Is the quirk file correct?

Thanks, that's right.

hughsie:
LGTM, thanks guys!

PR #4925, opened 2022-08-18
Add support to set BIOS settings using a JSON payload

superm1:
The JSON payload is ideally generated from `fwupdmgr get-bios-settings --json`
with all attributes expected to be included as additional arguments.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client-sync.c
+++ b/libfwupd/fwupd-client-sync.c
@@ -899,8 +899,7 @@ fwupd_client_modify_bios_attr_cb(GObject *source, GAsyncResult *res, gpointer us
 /**
  * fwupd_client_modify_bios_attr
  * @self: a #FwupdClient
- * @key: the name of the BIOS attribute, e.g. `SleepMode`
- * @value: the value to set, e.g. 'S3'
+ * @settings: (transfer container): BIOS settings
  * @cancellable: (nullable): optional #GCancellable
  * @error: (nullable): optional return location for an error
  *
@@ -913,16 +912,14 @@ fwupd_client_modify_bios_attr_cb(GObject *source, GAsyncResult *res, gpointer us
  **/
 gboolean
 fwupd_client_modify_bios_attr(FwupdClient *self,
-			      const gchar *key,
-			      const gchar *value,
+			      GHashTable *settings,
 			      GCancellable *cancellable,
 			      GError **error)
 {
 	g_autoptr(FwupdClientHelper) helper = NULL;
 
 	g_return_val_if_fail(FWUPD_IS_CLIENT(self), FALSE);
-	g_return_val_if_fail(key != NULL, FALSE);
-	g_return_val_if_fail(value != NULL, FALSE);
+	g_return_val_if_fail(settings != NULL, FALSE);
 	g_return_val_if_fail(cancellable == NULL || G_IS_CANCELLABLE(cancellable), FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
 
@@ -933,8 +930,7 @@ fwupd_client_modify_bios_attr(FwupdClient *self,
 	/* call async version and run loop until complete */
 	helper = fwupd_client_helper_new(self);
 	fwupd_client_modify_bios_attr_async(self,
-					    key,
-					    value,
+					    settings,
 					    cancellable,
 					    fwupd_client_modify_bios_attr_cb,
 					    helper);
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -1106,8 +1106,7 @@ fwupd_client_modify_bios_attr_cb(GObject *source, GAsyncResult *res, gpointer us
 /**
  * fwupd_client_modify_bios_attr_async:
  * @self: a #FwupdClient
- * @key: the name of the BIOS attribute, e.g. `SleepMode`
- * @value: the value to set, e.g. 'S3'
+ * @settings: (transfer container): BIOS settings
  * @cancellable: (nullable): optional #GCancellable
  * @callback: the function to run on completion
  * @callback_data: the data to pass to @callback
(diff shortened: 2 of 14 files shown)


src/fu-engine.c
```diff
@@ -852,29 +852,102 @@ fu_engine_validate_bios_attr_input(FwupdBiosAttr *attr, const gchar **value, GEr
...
  *
  * Use the kernel API to set a BIOS attribute.
+ * - This can be done with a key/value pair
+ *   or
+ * - This can be done with key set to "json" and value being a JSON blob
  *
  * Returns: %TRUE for success
  **/
 gboolean
 fu_engine_modify_bios_attr(FuEngine *self, const gchar *key, const gchar *value, GError **error)
```
hughsie, line 938: not convinced we want to handle this inside this function -- maybe create a fu_engine_modify_bios_attrs and handle the "json" key in the engine/tool?

superm1: How about we just leave all the JSON parsing to the tool and send a set of pairs every time?  Then we don't have a special "json" key.

hughsie: I think that's even better, although isn't "atomic" -- which is probably fine. If we need atomic we can have a different method call.

libfwupd/fwupd-client-sync.c
```diff
@@ -913,16 +913,16 @@ fwupd_client_modify_bios_attr_cb(GObject *source, GAsyncResult *res, gpointer us
  **/
 gboolean
 fwupd_client_modify_bios_attr(FwupdClient *self,
-			      const gchar *key,
-			      const gchar *value,
+			      GPtrArray *keys,
+			      GPtrArray *values,
```
hughsie, line 917: hmm, this doesn't feel very idiomatic. Could we do a GHashTable of key/values? Was there a huge overhead with `n`xDBus requests from the client? The single key/value sure seems a lot easier to use.

superm1: The problem with multiple requests from the client is that you need to authenticate each one of them.  So you need some sort of way to send all the pairs atomically to avoid that.

I think a hashtable still will have the same problem - at the end of the day you need to convert to string to pass over dbus don't you?

superm1:
I just force pushed again after your comment and expanded the command line tool (and bash completion for it) to take multiple pairs of arguments.

I think if another transport over dbus makes more sense will need your guidance which one to use.

src/org.freedesktop.fwupd.xml
```diff
@@ -866,28 +866,28 @@
     </method>
 
     <!--***********************************************************-->
-    <method name='SetBiosAttr'>
+    <method name='SetBiosAttrs'>
```
hughsie, line 869: One sec; would it makes sense to use SetBiosAttr() and *also* have CommitBiosAttrs() -- i.e. have an API where you can set A=1, B=2, C=3 and then atomically set all three at the same time? Otherwise I think semantically we need `a(ss)` as a parameter rather than `asas`

superm1: I don't see a reason to have separate commands for setting and committing. That's not how the kernel works or anything.

libfwupd/fwupd-client.c
```diff
@@ -1119,26 +1118,34 @@ fwupd_client_modify_bios_attr_cb(GObject *source, GAsyncResult *res, gpointer us
...
 	FwupdClientPrivate *priv = GET_PRIVATE(self);
+	GVariantBuilder builder;
+	GHashTableIter iter;
+	gpointer key, value;
 	g_autoptr(GTask) task = NULL;
 
 	g_return_if_fail(FWUPD_IS_CLIENT(self));
-	g_return_if_fail(key != NULL);
-	g_return_if_fail(value != NULL);
+	g_return_if_fail(settings != NULL);
```
hughsie, line 1133: maybe check that `g_hash_table_size() > 0` too?

src/fu-engine.c
```diff
@@ -852,29 +859,59 @@ fu_engine_validate_bios_attr_input(FwupdBiosAttr *attr, const gchar **value, GEr
...
+	FwupdBiosAttr *attr = fu_context_get_bios_attr(self->ctx, key);
+	const gchar *tmp = value;
+
+	if (!fu_engine_validate_bios_attr_input(attr, &tmp, error))
+		return FALSE;
+
+	if (!fu_engine_update_bios_attr(attr, tmp, error))
+		return FALSE;
+
+	return TRUE;
```
hughsie, line 877: maybe just return the func retval?

src/fu-util-bios-attr.c
```diff
@@ -199,3 +200,50 @@ fu_util_bios_attr_to_string(FwupdBiosAttr *attr, guint idt)
...
+
+	if (g_strv_length(input) == 0 || g_strv_length(input) % 2) {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_ARGS,
+				    /* TRANSLATORS: error message */
+				    _("Invalid arguments"));
+		return NULL;
+		/* json input */
+	} else if (g_strv_length(input) == 1) {
```
hughsie, line 218: no need for else if returning NULL above.

src/fu-util-bios-attr.h
```diff
@@ -16,3 +16,5 @@ gboolean
 fu_util_bios_attr_matches_args(FwupdBiosAttr *attr, gchar **values);
 gboolean
 fu_util_get_bios_attr_as_json(gchar **values, GPtrArray *attrs, GError **error);
+GHashTable *
+fu_util_bios_attr_args_to_hash_table(gchar **input, GError **error);
```
hughsie, line 20: can we const this? then it's not looking like it's a `gchar *` that's being returned.

src/fu-util.c
```diff
@@ -3865,30 +3866,30 @@ fu_util_show_plugin_warnings(FuUtilPrivate *priv)
...
-		g_autofree gchar *msg =
-		    g_strdup_printf(_("Set BIOS attribute '%s' using '%s'."), values[0], values[1]);
-		g_print("\n%s\n", msg);
+		const gchar *key;
+		const gchar *value;
+		GHashTableIter iter;
+
+		g_hash_table_iter_init(&iter, settings);
+		while (g_hash_table_iter_next(&iter, &key, &value)) {
+			/* TRANSLATORS: Configured a BIOS setting to a value */
```
hughsie, line 3888: need to move this down one line, it needs to be immediately above the `_(`

src/fu-util.c
```diff
@@ -3865,30 +3866,30 @@ fu_util_show_plugin_warnings(FuUtilPrivate *priv)
...
-fu_util_set_bios_attr(FuUtilPrivate *priv, gchar **values, GError **error)
+fu_util_set_bios_attr(FuUtilPrivate *priv, gchar **input, GError **error)
 {
-	if (g_strv_length(values) < 2) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_INVALID_ARGS,
-				    /* TRANSLATORS: error message */
-				    _("Invalid arguments"));
+	g_autoptr(GHashTable) settings = fu_util_bios_attr_args_to_hash_table(input, error);
```
hughsie, line 3871: maybe `fu_util_bios_attrs_parse_argv()` ?

src/fu-util-bios-attr.h
hughsie: sorry, i meant `fu_util_bios_attr_args_to_hash_table(const gchar **input, GError **error);` or `fu_util_bios_attr_args_to_hash_table(const gchar * const *input, GError **error);` if you want to be super correct.

superm1: Nope. I tried but `g_strv_length` is very unhappy with this.

hughsie (approved):

PR #4928, opened 2022-08-19
Convert HSI into a meson tristate-feature

superm1:
This allows us to disable it automatically on architectures that
aren't supported and OSes that aren't supported.

Link: https://bugs.launchpad.net/ubuntu/+source/fwupd/+bug/1987067

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/acpi-dmar/meson.build
+++ b/plugins/acpi-dmar/meson.build
@@ -1,6 +1,4 @@
-if get_option('hsi') and \
-   host_machine.system() == 'linux' and \
-   (host_cpu == 'x86' or host_cpu == 'x86_64')
+if hsi and (host_cpu == 'x86' or host_cpu == 'x86_64')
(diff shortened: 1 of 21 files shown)


plugins/acpi-dmar/meson.build
```diff
@@ -1,6 +1,4 @@
-if get_option('hsi') and \
-   host_machine.system() == 'linux' and \
-   (host_cpu == 'x86' or host_cpu == 'x86_64')
+if hsi and (host_cpu == 'x86' or host_cpu == 'x86_64')
```
superm1, line 1: the reason for this pattern in addition to at top level `meson.build` is so that if/when HSI is expanded to other architectures we make sure that X86 specific plugins don't compile for those architectures.

hughsie (approved):

PR #4935, opened 2022-08-22
Use ID_VENDOR_ID and ID_MODEL_ID fallbacks

hughsie:
Some udev subsystems run probers to populate the values, rather than
just the kernel providing the sysfs files. Support reading these too.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -325,6 +325,21 @@ fu_udev_device_probe_serio(FuUdevDevice *self, GError **error)
 	return TRUE;
 }
 
+static guint16
+fu_udev_device_get_property_as_uint16(GUdevDevice *udev_device, const gchar *key)
+{
+	const gchar *tmp = g_udev_device_get_property(udev_device, key);
+	guint64 value = 0;
+	g_autofree gchar *str = NULL;
+
+	if (tmp == NULL)
+		return 0x0;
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Do we have to worry about those racing with the daemon now though?

hughsie:
I don't think so; the probers are run sync before the udev message.

libfwupdplugin/fu-udev-device.c
```diff
@@ -325,6 +325,21 @@ fu_udev_device_probe_serio(FuUdevDevice *self, GError **error)
 	return TRUE;
 }
 
+static guint16
+fu_udev_device_get_property_as_uint16(GUdevDevice *udev_device, const gchar *key)
+{
+	const gchar *tmp = g_udev_device_get_property(udev_device, key);
+	guint64 value = 0;
```
superm1, line 332: if you're capping the stoull call at G_MAX_UINT16, this should probably be a guint16.

superm1 (approved):
one nit; otherwise LGTM

libfwupdplugin/fu-udev-device.c
hughsie: nah, we pass by ref, and casting a different uint type isn't always portable.

hughsie:
> one nit; otherwise LGTM

I'll check this works with Logitech and if so will merge, thanks.

PR #4937, opened 2022-08-23
HSI emulation profile for machine_kind and chassis_type

smallorange:
When fwupd runs in VM with FWUPD_MACHINE_KIND and FWUPD_HOST_EMULATE set,
the HSI value will not be computed since the FU_SMBIOS_CHASSIS_KIND is
OTHER. Consequence, a "HSI-INVALID:chassis[0x01]" is returned through
dbus. If both machine_kind and chassis_type are not properly set, the HSI
result and security attribute can't be evaluated and presented.

Two JSON items- MachineKind and ChassisType were introduced to set the
mockup values. Therefore, fwupd can get these values from a single JSON
file to evaluate the HSI value.

Related to #4904 

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/contrib/generate-emulation.py
+++ b/contrib/generate-emulation.py
@@ -8,6 +8,7 @@
 
 import json
 import sys
+import argparse
 
 from typing import Dict, List, Any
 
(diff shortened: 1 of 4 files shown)


superm1:
### Thanks for your commit

**Clang Format**
 
**Pre-commit overview**

src/fu-engine.c
```diff
@@ -118,6 +118,9 @@ struct _FuEngine {
 	GMainLoop *acquiesce_loop;
 	guint acquiesce_id;
 	guint acquiesce_delay;
+
+	gchar *machine_kind_str_mockup;
+	FuSmbiosChassisKind chassis_type_mockup;
```
superm1, line 123: we shouldn't be storing override variables in `struct _FuEngine` most likely.  I would rather see an override in fu-smbios related to this so that all the other callers get this information correctly too.

contrib/generate-emulation.py
```diff
@@ -121,16 +163,35 @@ def _get_host_devices_and_attrs() -> str:
...
+    if len(sys.argv) < 2:
+        sys.stdout.write(_minimize_json(sys.stdin.read()))
+    else:
+        parser = argparse.ArgumentParser(description='Generate HSI emulation json.')
+        parser.add_argument('json_file_list', metavar='json_file_list', type=str, nargs='+',
+                    help='An list of HSI emulation json file.')
+        parser.add_argument('--machine', default="physical",
+                    help='Machines kind (default: physical)')
+        parser.add_argument('--chassis', default="laptop",
+                    help='Chassis type (laptop and desktop. default: physical)')
```
superm1, line 176: Can't we get this information out of fwupd so you don't need to override it on the command line for generation of emulation data?  I would think it makes more sense to populate it through the daemon daemon so it can export it.

contrib/generate-emulation.py
```diff
@@ -21,7 +22,46 @@
 from gi.repository import Json  # pylint: disable=wrong-import-position
 
 
-def _minimize_json(json_str: str) -> str:
+fu_chassis_kind = {
+    "FU_SMBIOS_CHASSIS_KIND_OTHER": 0x01,
```
superm1, line 26: Maybe this enum should be exported from `libfwupd` instead.

superm1:
your python code is failing formatting with black.

contrib/generate-emulation.py
smallorange: Okay, I'll import the shared library to replace this.

src/fu-engine.c
smallorange: I can look into fu-smbios first.  machine_kind is placed in fu-daemon. I have to find a way to set it.

contrib/generate-emulation.py
smallorange: I think most developers will run fwupd in a VM to develop and test their HSI-related software. They need to set both to emulate the environment, for example, HSI hypervisor exception, HSI level 0-4...etc. These arguments can actually simplify the process to generate emulation data without editing the JSON again.

hughsie:
I don't think we need this now, right? If we do, please reopen.

PR #4938, opened 2022-08-23
ELANTP: Added support for absolute reporting (non-HID) devices

josh-chen-elan:
These non-HID devices are bind with the elan_i2c driver, and quirks uses ACPI/DeviceTree information for matching, same as the driver method.
After the firmware update, it needs to switch back to absolute report mode and rebind the driver to update the information.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/elantp/fu-elantp-i2c-device.c
+++ b/plugins/elantp/fu-elantp-i2c-device.c
@@ -8,6 +8,7 @@
 
 #include <fwupdplugin.h>
 
+#include <fcntl.h>
 #include <linux/i2c-dev.h>
 
 #include "fu-elantp-common.h"
@@ -24,6 +25,8 @@ struct _FuElantpI2cDevice {
 	guint16 module_id;
 	guint16 fw_page_size;
 	guint8 pattern;
+	gchar *bind_path;
+	gchar *bind_id;
 };
 
 G_DEFINE_TYPE(FuElantpI2cDevice, fu_elantp_i2c_device, FU_TYPE_UDEV_DEVICE)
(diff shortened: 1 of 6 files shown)


plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -620,13 +720,26 @@ fu_elantp_i2c_device_attach(FuDevice *device, FuProgress *progress, GError **err
...
+
 	if (!fu_elantp_i2c_device_ensure_iap_ctrl(self, error))
 		return FALSE;
 
+	if (g_strcmp0(fu_device_get_custom_flags(device), "elantp-absolute") == 0) {
+		if (!fu_elantp_i2c_device_write_cmd(self, 0x0300, 0x001, error)) {
+			g_prefix_error(error, "cannot switch to TP ABS mode: ");
+			return FALSE;
+		}
+		if (getuid() == 0 && geteuid() == 0) {
```
hughsie, line 732: ?? why would this be running as non-root?

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -269,7 +357,19 @@ fu_elantp_i2c_device_setup(FuDevice *device, GError **error)
 	/* define the extra instance IDs (ic_type + module_id) */
 	fu_device_add_instance_u8(device, "ICTYPE", ic_type);
 	fu_device_build_instance_id(device, NULL, "ELANTP", "ICTYPE", NULL);
-	fu_device_build_instance_id(device, NULL, "ELANTP", "ICTYPE", "MOD", NULL);
+	if (g_strcmp0(fu_device_get_custom_flags(device), "elantp-absolute") == 0) {
```
hughsie, line 360: I'd much rather you use fu_device_register_private_flag() and fu_device_has_private_flag()

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -44,12 +47,97 @@ fu_elantp_i2c_device_to_string(FuDevice *device, guint idt, GString *str)
...
+	}
+
+	if (!fu_elantp_i2c_device_writeln(g_build_filename(self->bind_path, "unbind", NULL),
+					  self->bind_id,
+					  error))
+		return FALSE;
+	if (!fu_elantp_i2c_device_writeln(g_build_filename(self->bind_path, "bind", NULL),
+					  self->bind_id,
+					  error))
+		return FALSE;
```
hughsie, line 93: can you use fu_device_unbind_driver() here?

plugins/elantp/fu-elantp-i2c-device.c
josh-chen-elan: Because the permissions of bind and unbind are 200, and chromeOS uses a separate user/group to execute fwupd.
When updating firmware via fwupdmgr & fwupd, there will be no permission to rebind.

plugins/elantp/fu-elantp-i2c-device.c
josh-chen-elan: OK, I will try to modify it.

plugins/elantp/fu-elantp-i2c-device.c
hughsie: @bleungatchromium is this by design? We should probably handle this better in fwupd itself.

plugins/elantp/fu-elantp-i2c-device.c
josh-chen-elan: I'm not sure, maybe unbind can work, but elan_i2c is a builtin driver and can't be found in /sys/module/, it can't be bind by similar method
https://github.com/fwupd/fwupd/blob/main/libfwupdplugin/fu-udev-device.c#L730

plugins/elantp/fu-elantp-i2c-device.c
superm1: But so if there is no permissions to rebind will the update be able to actually run?

josh-chen-elan: > But so if there is no permissions to rebind will the update be able to actually run?

Yes, fwupd has enough permissions to access the I2C device.
The rebind action is only used to update the device information in the driver.

superm1: IMO rather than guarding for this you should use a local error (IE `error_local`) and then catch the error type you get when failing to write due to permissions and emit a debug message then.  Something like this:

```
if (!fu_elantp_i2c_device_rebind_driver(self, &error_local)) {
	if(g_error_matches(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED))
		g_debug("%s", error_local->message);
	else {
		g_propagate_error(error, g_steal_pointer(&error_local));
		return FALSE;
	}
}
```

Switch out the error codes for the right ones.

josh-chen-elan: Many thanks, updated as suggested.

plugins/elantp/fu-elantp-i2c-device.c
josh-chen-elan: Many thanks, updated as suggested.

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -269,7 +365,19 @@ fu_elantp_i2c_device_setup(FuDevice *device, GError **error)
...
+		fu_device_build_instance_id(device,
+					    NULL,
+					    "ELANTP",
+					    "ICTYPE",
+					    "MOD",
+					    "DRIVER",
+					    NULL);
+	} else {
+		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
+		fu_device_build_instance_id(device, NULL, "ELANTP", "ICTYPE", "MOD", NULL);
```
hughsie, line 379: i think it's valuable to have this for both kinds, i.e. for the DRIVER one you'd just ignore the extra instance ID. If you need to target the opposite, then it's worth setting `DRIVER` to `none` or something and be explicit rather than "hiding" the original ID. With the  `fu_device_add_instance_str(device, "DRIVER", "none");` then you also don't need to worry about accidentally installing on the wrong device on an old fwupd version.

plugins/elantp/fu-elantp-i2c-device.c
```diff
@@ -620,13 +728,33 @@ fu_elantp_i2c_device_attach(FuDevice *device, FuProgress *progress, GError **err
...
+
+		if (!fu_elantp_i2c_device_write_cmd(self, 0x0300, 0x001, error)) {
+			g_prefix_error(error, "cannot switch to TP ABS mode: ");
+			return FALSE;
+		}
+
+		if (!fu_elantp_i2c_device_rebind_driver(self, &error_local)) {
+			if (g_error_matches(error_local, FWUPD_ERROR, FWUPD_ERROR_PERMISSION_DENIED))
+				g_debug("%s", error_local->message);
+			else {
```
hughsie, line 746: braces here please

plugins/elantp/fu-elantp-i2c-device.c
josh-chen-elan: Do you mean like this:
```
if (fu_device_has_private_flag(device, FU_ELANTP_I2C_DEVICE_ABSOLUTE)) {
	fu_device_add_instance_str(device, "DRIVER", "ELAN_I2C");
} else {
	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_NEEDS_REBOOT);
	fu_device_add_instance_str(device, "DRIVER", "nonoe");
	fu_device_build_instance_id(device, NULL, "ELANTP", "ICTYPE", "MOD", NULL);
}
fu_device_build_instance_id(device, NULL, "ELANTP", "ICTYPE", "MOD", "DRIVER", NULL);
```

As far as I know, instance ID of existing HID devices are matched using "ELANTP", "ICTYPE", "MOD", so I avoid generating this instance ID on non-HID devices.

hughsie: > are matched using "ELANTP", "ICTYPE", "MOD",

Ahh okay, so maybe doing what you are doing makes sense. Sorry for the noise.

josh-chen-elan: After internal discussions, no updated firmware for HID devices has been uploaded yet.
Maybe the instance ID of **ELANNTP\ICTYPE_XX&MOD_YYYY&DRIVER_HID** can be used on the HID device, but the FuElantpHidDevice needs to be modified.
Do you recommend this change?

hughsie: > but the FuElantpHidDevice needs to be modified

I think that makes sense. Can you also update README.md too please.

josh-chen-elan: Thanks, updated as suggested.

plugins/elantp/fu-elantp-i2c-device.c
josh-chen-elan: Thanks, updated as suggested.

hughsie (approved):

hughsie:
@josh-chen-elan there are a couple of precommit issues; can you fix those up please?

josh-chen-elan:
> @josh-chen-elan there are a couple of precommit issues; can you fix those up please?

Actually, I don't know what happened...

hughsie:
It's probably a formatting issue -- can you try clang-format on your plugin?

josh-chen-elan:
> It's probably a formatting issue -- can you try clang-format on your plugin?

Modified, hope it works.
Thanks.

hughsie (approved):

PR #4940, opened 2022-08-24
trivial: Unify ambiguity between bios-attrs and bios-settings

superm1:
These are currently used interchangeably since there was indecisiveness
    which to use as the feature was being developed.
    
    As outward facing it will be named with "settings", change all uses
    in the code to match this.

This is on top of https://github.com/fwupd/fwupd/pull/4925.  If that changes this will need to be rebased.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):

hughsie:
Okay, rebase and fix the self tests trivial issue and then LGTM, thanks.

PR #4941, opened 2022-08-24
Only fail the kernel tainted HSI test for specific taint reasons

hughsie:
For example, working around a firmware bug in the kernel isn't really
what the HSI attribute was intended to measure.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/linux-tainted/fu-plugin-linux-tainted.c
+++ b/plugins/linux-tainted/fu-plugin-linux-tainted.c
@@ -13,6 +13,26 @@ struct FuPluginData {
 	GFileMonitor *monitor;
 };
 
+#define KERNEL_TAINT_FLAG_PROPRIETARY_MODULE	      (1 << 0)
+#define KERNEL_TAINT_FLAG_MODULE_FORCE_LOAD	      (1 << 1)
+#define KERNEL_TAINT_FLAG_KERNEL_OUT_OF_SPEC	      (1 << 2)
+#define KERNEL_TAINT_FLAG_MODULE_FORCE_UNLOAD	      (1 << 3)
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Which ones are you thinking of excluding explicitly? The argument for taint on staging drivers for example means they don't have the same rigor as other drivers.

hughsie:
> Which ones are you thinking of excluding explicitly

I was thinking of excluding:

	KERNEL_TAINT_FLAG_KERNEL_OUT_OF_SPEC
	KERNEL_TAINT_FLAG_PROCESSOR_MCE
	KERNEL_TAINT_FLAG_BAD_PAGE
	KERNEL_TAINT_FLAG_REQUESTED_BY_USERSPACE
	KERNEL_TAINT_FLAG_KERNEL_DIED
	KERNEL_TAINT_FLAG_ACPI_OVERRIDDEN
	KERNEL_TAINT_FLAG_KERNEL_ISSUED_WARNING
	KERNEL_TAINT_FLAG_FIRMWARE_WORKAROUND_APPLIED
	KERNEL_TAINT_FLAG_SOFT_LOCKUP_OCCURRED
	KERNEL_TAINT_FLAG_KERNEL_LIVE_PATCHED
	KERNEL_TAINT_FLAG_STRUCT_RANDOMIZATION_PLUGIN
	KERNEL_TAINT_FLAG_IN_KERNEL_TEST

superm1:
I guess I see the argument for those; but I think you should still not allow an acpi table to be overridden.  Allowing an ACPI table to be overridden is like having free reign of 32 bit register access to anything.

hughsie:
> but I think you should still not allow an acpi table to be overridden

Ack, agreed. Modified.

superm1 (approved):

PR #4944, opened 2022-08-24
Add support for default BIOS settings policy

superm1:
With this change, system administrators can place JSON files in `/etc/fwupd/bios-settings.d` which will set up fwupd to enforce their BIOS settings policy.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

Some open discussion questions:

1. Should settings modified this way also be marked "read only"?
2. Should the default directory be created at fwupd startup?
3. Should the empty directory be shipped with packages?
4. Should an empty JSON file be shipped?
5. Should we make a wiki page for how to use BIOS settings (they're getting a lot more complex!), or `fwupd.github.io` page for it?

### DIFF
--- a/libfwupdplugin/fu-bios-settings.c
+++ b/libfwupdplugin/fu-bios-settings.c
@@ -508,7 +508,7 @@ gboolean
 fu_bios_settings_get_pending_reboot(FuBiosSettings *self, gboolean *result, GError **error)
 {
 	FwupdBiosSetting *attr;
-	const gchar *data;
+	g_autofree gchar *data = NULL;
 	guint64 val = 0;
 
 	g_return_val_if_fail(result != NULL, FALSE);
@@ -532,7 +532,10 @@ fu_bios_settings_get_pending_reboot(FuBiosSettings *self, gboolean *result, GErr
 		return FALSE;
 	}
 
-	data = fwupd_bios_setting_get_current_value(attr);
+	/* refresh/re-read */
+	if (!fu_bios_setting_get_key(attr, NULL, &data, error))
+		return FALSE;
+	fwupd_bios_setting_set_current_value(attr, data);
 	if (!fu_strtoull(data, &val, 0, G_MAXUINT32, error))
 		return FALSE;
 
--- a/libfwupdplugin/fu-bios-settings.c
+++ b/libfwupdplugin/fu-bios-settings.c
@@ -574,6 +574,8 @@ fu_bios_settings_to_variant(FuBiosSettings *self, gboolean trusted)
 /**
  * fu_bios_settings_from_json:
  * @self: a #FuBiosSettings
+ * @json_node: a Json node to parse from
+ * @error: (nullable): optional return location for an error
  *
  * Loads #FwupdBiosSetting objects from a JSON node.
  *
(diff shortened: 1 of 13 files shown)


hughsie:
> Should settings modified this way also be marked "read only"?

Yes, I think so; otherwise they'll just be changed back post-reboot.

> Should the default directory be created at fwupd startup?

I think it should have a README.md in it :)

> Should the empty directory be shipped with packages?

As above.

> Should an empty JSON file be shipped?

Nope.

> Should we make a wiki page for how to use BIOS settings (they're getting a lot more complex!), or fwupd.github.io page for it?

I'd prefer the latter.

superm1:
> > Should settings modified this way also be marked "read only"?
> 
> Yes, I think so; otherwise they'll just be changed back post-reboot.
> 
> > Should the default directory be created at fwupd startup?
> 
> I think it should have a README.md in it :)
> 
> > Should the empty directory be shipped with packages?
> 
> As above.
> 
> > Should an empty JSON file be shipped?
> 
> Nope.
> 
> > Should we make a wiki page for how to use BIOS settings (they're getting a lot more complex!), or fwupd.github.io page for it?
> 
> I'd prefer the latter.

OK did all the above except documentation, that will be saved for another PR.

src/fu-engine.c
```diff
@@ -905,8 +912,24 @@ fu_engine_modify_bios_settings(FuEngine *self, GHashTable *settings, GError **er
 				    (const gchar *)key);
 			return FALSE;
 		}
-		if (!fu_engine_modify_single_bios_setting(self, key, value, error))
-			return FALSE;
+		if (!fu_engine_modify_single_bios_setting(self, key, value, &error_local)) {
+			if (g_error_matches(error_local, FWUPD_ERROR, FWUPD_ERROR_NOTHING_TO_DO)) {
+				g_debug("%s", error_local->message);
+				continue;
+			} else {
```
hughsie, line 934: i don't think you need the }else{ here

hughsie: also, maybe cherry pick this one to main

libfwupdplugin/fu-bios-settings.c
```diff
@@ -582,7 +584,7 @@ fu_bios_settings_to_variant(FuBiosSettings *self, gboolean trusted)
  * Since: 1.8.4
  **/
 gboolean
-fu_bios_settings_from_json(FuBiosSettings *self, JsonNode *json_node, GError **error)
+fu_bios_settings_from_json_node(FuBiosSettings *self, JsonNode *json_node, GError **error)
```
hughsie, line 587: I think the original name was better -- it matches the others, e.g. fwupd_security_attr_from_json

libfwupdplugin/fu-bios-settings.c
```diff
@@ -615,6 +617,60 @@ fu_bios_settings_from_json(FuBiosSettings *self, JsonNode *json_node, GError **e
...
+ * @self: A #FuBiosSettings
+ *
+ * Creates a #GHashTable with the name and current value of
+ * all BIOS settings.
+ *
+ * Returns: (transfer full): name/value pairs
+ * Since: 1.8.4
+ **/
+GHashTable *
+fu_bios_settings_to_hash_table(FuBiosSettings *self)
```
hughsie, line 658: I think `fu_bios_settings_to_hash_kv` might be better.

superm1: ack

libfwupdplugin/fu-bios-settings.c
superm1: ack

src/fu-engine.c
superm1: ack

src/fu-engine.c
```diff
@@ -6915,8 +6929,14 @@ fu_engine_check_firmware_attributes(FuEngine *self, FuDevice *device)
 	subsystem = fu_udev_device_get_subsystem(FU_UDEV_DEVICE(device));
 	if (g_strcmp0(subsystem, "firmware-attributes") == 0) {
 		g_autoptr(GError) error = NULL;
-		if (!fu_context_reload_bios_settings(self->ctx, &error))
+		if (!fu_context_reload_bios_settings(self->ctx, &error)) {
 			g_debug("%s", error->message);
+			return;
+		}
+		if (!fu_engine_apply_default_bios_settings_policy(self, &error)) {
+			g_debug("%s", error->message);
```
superm1, line 6937: This I wonder if it should actually be `g_warning` so an administrator sees in their logs that some policies weren't loading properly.

hughsie: MSTM.

hughsie (approved):

PR #4949, opened 2022-08-25
Add a read-only view of BIOS authentication settings

superm1:
This isn't incredibly useful right now for anything other than a building block to design how to interact with these settings.

By design the kernel will not let you read the current password for anything.
This means that the password (or certificate) would need to be stored somewhere to be used.  By reading these other attributes we can discover whether a password is needed and at least prevent fwupd from trying to set an attribute in this case.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

If we ignore certificate authentication for now and focus on password, my big open question is where/how to get the password into the daemon (for another PR).  Here's the ideas i've come up with.
1. Pass it over dbus from the client.  I'm not personally crazy about it going in clear text.
2. Require a user to populate in clear text in a file `/etc/fwupd/bios-settings.d/bios_settings.conf` only accessible by root.
3. Use the client to encrypt the password, pass it over dbus.
4. Attempt to store and fetch it from the TPM.

My current leaning is conffile.  If a user went out of their way to set a BIOS setting let them populate that file.  When the daemon detects the need to set a password (from `is_enabled` attribute in `bios-admin` type) then it would use the value fetched from that file.

### DIFF
--- a/libfwupd/fwupd-bios-setting.h
+++ b/libfwupd/fwupd-bios-setting.h
@@ -35,10 +35,10 @@ struct _FwupdBiosSettingClass {
 /**
  * FwupdBiosSettingKind:
  * @FWUPD_BIOS_SETTING_KIND_UNKNOWN:		BIOS setting type is unknown
- * @FWUPD_BIOS_SETTING_KIND_ENUMERATION:		BIOS setting that has enumerated possible
- *values
+ * @FWUPD_BIOS_SETTING_KIND_ENUMERATION:	BIOS setting that has enumerated possible values
  * @FWUPD_BIOS_SETTING_KIND_INTEGER:		BIOS setting that is an integer
  * @FWUPD_BIOS_SETTING_KIND_STRING:		BIOS setting that accepts a string
+ * @FWUPD_BIOS_SETTING_KIND_AUTH:		BIOS setting used for managing authentication
  *
  * The type of BIOS setting.
  **/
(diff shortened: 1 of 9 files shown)


src/fu-util-bios-setting.c
```diff
@@ -187,6 +195,73 @@ fu_util_bios_setting_to_string(FwupdBiosSetting *setting, guint idt)
...
+				tmp = _("NVME");
+				break;
+			case FWUPD_BIOS_AUTH_ROLE_SYSTEM:
+				/* TRANSLATORS: BIOS Authentication purpose */
+				tmp = _("System");
+				break;
+			default:
+				tmp = NULL;
+			}
+			if (tmp) {
```
hughsie, line 261: != NULL

hughsie:
I've pulled in the first patch to master, and the `settingibute` bits, but do you think the auth stuff should be left until after the next release? I think we should probably try and *use* this a bit before we set the API in stone.

superm1:
> I've pulled in the first patch to master, and the `settingibute` bits, but do you think the auth stuff should be left until after the next release? I think we should probably try and _use_ this a bit before we set the API in stone.

Sure that's fine.  We can wait to land it until there is more clarity on how it needs to work.

superm1:
I'm waiting to hear more from Mark about how all this stuff is *supposed* to work so we don't come up with bad assumptions.

libfwupd/fwupd-bios-setting.h
```diff
@@ -116,4 +155,17 @@ fwupd_bios_setting_get_id(FwupdBiosSetting *self);
 void
 fwupd_bios_setting_set_id(FwupdBiosSetting *self, const gchar *id);
 
+FwupdBiosAuthRole
+fwupd_bios_setting_get_auth_role(FwupdBiosSetting *self);
+void
+fwupd_bios_setting_set_auth_role(FwupdBiosSetting *self, FwupdBiosAuthRole role);
```
hughsie, line 161: the bit I'm struggling with is that these auth settings are not really per-attribute -- as I understand it you have one auth mechanism that allows you to "unlock" all the attributes -- right?

libfwupdplugin/fu-bios-settings.c
```diff
@@ -440,6 +511,39 @@ fu_bios_settings_setup(FuBiosSettings *self, GError **error)
...
+		if (driver_auth_dir == NULL)
+			return FALSE;
+		do {
+			const gchar *name = g_dir_read_name(driver_auth_dir);
+			g_autofree gchar *full_path = NULL;
+			g_autoptr(GError) error_local = NULL;
+			if (name == NULL)
+				break;
+			full_path = g_build_filename(authentication_path, name, NULL);
+			if (!fu_bios_settings_populate_authentication(self,
```
hughsie, line 531: is the logic here that you can have two "active" authentication mechanisms, e.g. "password" **OR** "certificate"?

PR #4950, opened 2022-08-25
Force the chassis type to be valid if emulating

hughsie:
The logic being that if we're emulating another system we're possibly
on a VM, and that's possibly going to report that it isn't a desktop.

I also think it makes sense to serialize out the FuSmbios object
to-and-from JSON, but that's a patch for another day.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
I still think we need something like https://github.com/fwupd/fwupd/pull/4937 but this patch is much more the kind of thing I want to include on Monday for the release.

superm1:
For where we are with emulation and how widely HSI is supported this seems pragmatic to me as a short term solution.

superm1 (approved):

PR #4952, opened 2022-08-25
Move BIOS Settings descriptions database into the daemon

superm1:
We'll want all clients to be able to get translations from updated descriptions.  So move this into the daemon so at least the right English string is set by default.

Clients can then look up from there.

While making this change, also prepare for better lookups for read-only as well.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-bios-settings.c
+++ b/libfwupdplugin/fu-bios-settings.c
@@ -8,6 +8,8 @@
 
 #include "config.h"
 
+#include <glib/gi18n.h>
+
 #include "fwupd-bios-setting-private.h"
 #include "fwupd-error.h"
 
@@ -17,6 +19,7 @@
 
 struct _FuBiosSettings {
 	GObject parent_instance;
+	GHashTable *descriptions;
 	GPtrArray *attrs;
 };
 
@@ -27,6 +30,7 @@ fu_bios_settings_finalize(GObject *obj)
 {
 	FuBiosSettings *self = FU_BIOS_SETTINGS(obj);
 	g_ptr_array_unref(self->attrs);
+	g_hash_table_unref(self->descriptions);
 	G_OBJECT_CLASS(fu_bios_settings_parent_class)->finalize(obj);
 }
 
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-bios-settings.c
```diff
@@ -174,27 +176,15 @@ fu_bios_setting_set_current_value(FwupdBiosSetting *attr, GError **error)
...
-
-			if (read_only_map[i].value == NULL ||
-			    g_strcmp0(tmp, read_only_map[i].value) == 0) {
-				fwupd_bios_setting_set_read_only(attr, TRUE);
-				return;
-			}
-		}
+		if (!g_hash_table_contains(self->read_only, id))
+			return;
+		if (g_strcmp0(g_hash_table_lookup(self->read_only, id),
```
hughsie, line 185: seems inefficient to look this up twice; just do `const gchar *value = g_hash_table_lookup(self->read_only, id)` then check `value` for NULL.

libfwupdplugin/fu-bios-settings.c
```diff
@@ -55,12 +59,24 @@ fu_bios_setting_get_key(FwupdBiosSetting *attr, const gchar *key, gchar **value_
...
 	g_autofree gchar *data = NULL;
 
 	g_return_val_if_fail(FWUPD_IS_BIOS_SETTING(attr), FALSE);
 
+	/* Try ID, then name, and then key */
+	if (g_hash_table_contains(self->descriptions, fwupd_bios_setting_get_id(attr))) {
+		fwupd_bios_setting_set_description(
+		    attr,
+		    g_hash_table_lookup(self->descriptions, fwupd_bios_setting_get_id(attr)));
+		return TRUE;
```
hughsie, line 75: one lookup i think.

superm1:
Something else I'm wondering - since we're also going to be adding categories eventually should this "database" maybe be punted off into a file that is loaded by the daemon?
I could see it living in a CSV, JSON or XML file that is distributed with the daemon instead.

Otherwise that is going to become unwieldy as we end up with 100+ translations eventually.

Not sure it's 1.8.4 to move it into another file or $later though too.

hughsie:
> eventually should this "database" maybe be punted off into a file that is loaded by the daemon

I don't think we can if we want the client to use `gettext` to load it from our translation domain.

libfwupdplugin/fu-bios-settings.c
```diff
@@ -158,27 +176,14 @@ fu_bios_setting_set_current_value(FwupdBiosSetting *attr, GError **error)
...
-			if (read_only_map[i].value == NULL ||
-			    g_strcmp0(tmp, read_only_map[i].value) == 0) {
-				fwupd_bios_setting_set_read_only(attr, TRUE);
-				return;
-			}
-		}
+		const gchar *value =
+		    g_hash_table_lookup(self->read_only, fwupd_bios_setting_get_id(attr));
+		if (value != NULL &&
+		    g_strcmp0(value, fwupd_bios_setting_get_current_value(attr)) == 0)
```
hughsie, line 185: I don't think you need the value != NULL -- g_strcmp0 is NULL safe.

hughsie (approved):
One nit, then LGTM.

superm1:
> > eventually should this "database" maybe be punted off into a file that is loaded by the daemon
> 
> I don't think we can if we want the client to use `gettext` to load it from our translation domain.

If we did something like this, we could easily dump the strings out of the database into something `po/POTFILES.in` could read.  Heck it even could be part of the build process that we generate an xml/json file and keep the flat file in source form.

PR #4954, opened 2022-08-25
trivial: don't allow turning on secure boot if 3rd party CA is off

superm1:
Turning on UEFI secure boot on Lenovo systems with 3rd party CA turned off will cause a boot loop.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

@mrhpearson FYI

### DIFF
--- a/libfwupdplugin/fu-bios-settings.c
+++ b/libfwupdplugin/fu-bios-settings.c
@@ -409,6 +409,25 @@ fu_bios_settings_populate_read_only(FuBiosSettings *self)
 			    g_strdup(_("Enabled")));
 }
 
+static void
+fu_bios_settings_combination_fixups(FuBiosSettings *self)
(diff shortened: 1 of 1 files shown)


hughsie (approved):
Woah, good catch.

mrhpearson:
Great idea - thank you!

superm1:
sure!
If the 3rd party CA default doesn't change in the future as a related suggestion can you bring to your firmware team some feedback to allow controlling the 3rd party CA in that boot device selection screen you get stuck in a loop and perhaps to show a proper message of why you can't boot?

I've found that when you're in this state the page has no way to get into settings AT ALL.  You have to manually power off, and then get into BIOS setup and find the specific setting yourself.  I think it could go a long way to to improving this experience.

PR #4962, opened 2022-08-28
tpm: Don't require uefi capsule updates for checking TPM PCR0

superm1:
There isn't any reason that we need to have a system advertising a
GUID with `main-system-firmware` in order to check that PCR0 reconstruction
matches what the TPM actually has.

Fixes: #4961

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm/fu-plugin-tpm.c
+++ b/plugins/tpm/fu-plugin-tpm.c
@@ -159,7 +159,7 @@ fu_plugin_tpm_add_security_attr_eventlog(FuPlugin *plugin, FuSecurityAttrs *attr
 	fu_security_attrs_append(attrs, attr);
 
 	/* check reconstructed to PCR0 */
-	if (priv->ev_items == NULL || priv->bios_device == NULL) {
+	if (priv->ev_items == NULL) {
 		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_FOUND);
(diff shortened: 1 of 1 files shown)


superm1:
@Tibladar can you have a try with this?  If you can and it works for you I think we should try to squeeze it in 1.8.4 before we tag it.

Tibladar:
> @Tibladar can you have a try with this? If you can and it works for you I think we should try to squeeze it in 1.8.4 before we tag it.

Works fine, thanks!

```
$ sudo fwupdmgr security --force
WARNUNG: UEFI capsule updates not available or enabled in firmware setup
  Siehe https://github.com/fwupd/fwupd/wiki/PluginFlag:capsules-unsupported für weitere Informationen.
WARNUNG: Dieses Paket wurde nicht validiert und könnte nicht richtig funktionieren.
Host Security ID: HSI:INVALID:missing-data

HSI-1
✔ TPM empty PCRs:                Gültig
✔ TPM v2.0:                      Gefunden
✔ UEFI-Plattformschlüssel:       Gültig
✘ Supported CPU:                 Fehlgeschlagen

HSI-2
✔ IOMMU:                         Aktiviert
✔ TPM PCR0 reconstruction:       Gültig
```

hughsie (approved):
Good catch, thanks.

PR #4964, opened 2022-08-29
Documentation improvements for 1.8.4

superm1:
I think we want our documentation more accessible as fwupd 1.8.4 will have more scrutiny of HSI and BIOS settings and such.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [x] Documentation

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -40,7 +40,7 @@ if gidocgen_dep.found() and docs_python_deps.allowed() and gidocgen_app.found() 
     ],
     build_by_default: true,
     install: true,
-    install_dir: join_paths(datadir, 'doc'),
+    install_dir: join_paths(datadir, 'doc', 'fwupd'),
   )
 
   custom_target('doc-fwupdplugin',
@@ -66,15 +66,24 @@ if gidocgen_dep.found() and docs_python_deps.allowed() and gidocgen_app.found() 
     ],
     build_by_default: true,
     install: true,
-    install_dir: join_paths(datadir, 'doc'),
+    install_dir: join_paths(datadir, 'doc', 'fwupd'),
   )
  install_data(['index.html', 'hsi.html'],
     install_dir : join_paths(datadir, 'doc', 'fwupd')
   )
  install_data(['urlmap_fwupd.js'],
-    install_dir: join_paths(datadir, 'doc', 'libfwupd')
+    install_dir: join_paths(datadir, 'doc', 'fwupd', 'libfwupd')
   )
  install_data(['urlmap_fwupdplugin.js'],
-    install_dir: join_paths(datadir, 'doc', 'libfwupdplugin')
+    install_dir: join_paths(datadir, 'doc', 'fwupd', 'libfwupdplugin')
+  )
+  #make devhelp work
+  install_symlink('libfwupd',
+    install_dir: join_paths(datadir, 'doc',),
+    pointing_to: join_paths('fwupd', 'libfwupd'),
+  )
+  install_symlink('libfwupdplugin',
+    install_dir: join_paths(datadir, 'doc',),
+    pointing_to: join_paths('fwupd', 'libfwupdplugin'),
   )
 endif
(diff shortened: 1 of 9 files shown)


src/fu-tool.c
```diff
@@ -2935,8 +2935,9 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 		g_set_error_literal(error,
 				    FWUPD_ERROR,
 				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "Not enough data was provided to make an HSI calculation. "
-				    "To ignore this warning, use --force.");
+				    "Not enough data was provided to make an HSI calculation.\n"
+				    " » https://fwupd.github.io/hsi.html#not-enough-info\n"
+				    "To ignore this warning, use --force");
```
hughsie, line 2940: I think we should translate these strings.

docs/bios-settings.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>Redirecting to https://fwupd.github.io/libfwupdplugin/bios-settings.html</title>
```
hughsie, line 3: eigh? Not sure I get the *why*

superm1: I was thinking we don't need to include libfwupdplugin in our URLs then.  We just do a front landing URL that will redirect us.  That's how we do HSI too.

docs/meson.build
```diff
@@ -71,10 +71,21 @@ if gidocgen_dep.found() and docs_python_deps.allowed() and gidocgen_app.found()
...
     install_dir: join_paths(datadir, 'doc', 'libfwupdplugin')
   )
+ install_symlink('libfwupd',
+    install_dir: join_paths(datadir, 'doc', 'fwupd'),
+    pointing_to: join_paths('..', 'libfwupd')
+ )
+ install_symlink('libfwupdplugin',
+    install_dir: join_paths(datadir, 'doc', 'fwupd'),
+    pointing_to: join_paths('..', 'libfwupdplugin')
+ )
```
hughsie, line 90: I don't get it...?

docs/bios-settings.html
hughsie: Ahh, gotcha.

hughsie:
I've cherry picked some of these to master too.

docs/meson.build
superm1: When you install into the filesystem, neither of these are put in the right place relative to our index pages.  They are in the right place for what we do with Circle and fwupd.github.io though.  This makes sure both work.

hughsie: Ahh, and we need to use libfwupdplugin for devhelp to work.

src/fu-tool.c

docs/bios-settings.html
superm1: decided to drop it

docs/meson.build
superm1: Oh I think I might get what's going on.  You might have broken doc deployment with 06a2dcf62b.

superm1:
I think I've fixed it, take another look.

docs/meson.build
superm1: > Ahh, and we need to use libfwupdplugin for devhelp to work.

Wait, so you're saying the directory needs to be `/usr/share/doc/libfwupdplugin` for devhelp to work?  If that can take a symlink I would rather we do it as the way I committed and then add a symlink.  Otherwise our doc deployment will still be broken too it seems to me because it expect we get everything from `DESTDIR/doc/fwupd/`

hughsie: Ahh, no, I'm wrong: /usr/share/doc/fwupd/libfwupdplugin/ is what's needed

superm1: Okay so then the PR as it is now should work then.

hughsie: Nope, I'm double wrong, it has to be /usr/share/doc/libfwupd for it to work with devhelp -- i.e. no parent dir.

superm1: Can you see if a symlink to fwupd/libfwupd works then for devhlp.

hughsie: Yes, that works. I'd be fine with the meson bump in this PR if that helps.

superm1: Can we just pull that other PR then too perhaps?

hughsie: I have to sleep, but you have my blessing :)

PR #4966, opened 2022-08-29
Only suggest BIOS settings changes with --force

hughsie:
We're getting a bit of pushback with this feature, so lets test it for one more
release.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -3534,14 +3534,16 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 	}
 
 	/* any things we can fix? */
-	for (guint j = 0; j < attrs->len; j++) {
-		FwupdSecurityAttr *attr = g_ptr_array_index(attrs, j);
-		if (!fwupd_security_attr_has_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS) &&
-		    fwupd_security_attr_get_bios_setting_id(attr) != NULL &&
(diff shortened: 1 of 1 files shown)


superm1 (commented):
Can you share more on the pushback?

hughsie:
> Can you share more on the pushback?

Well my Twitter and https://www.phoronix.com/news/Ubuntu-No-GNOME-Device-Security -- although there seems to be genuine concerns with how well switching sleep to s2idle is going to work.

superm1:
> > Can you share more on the pushback?
> 
> Well my Twitter and https://www.phoronix.com/news/Ubuntu-No-GNOME-Device-Security -- although there seems to be genuine concerns with how well switching sleep to s2idle is going to work.

Can we just not offer that one attribute for release perhaps until we dig in further to the specifics instead of all of them?

hughsie:
> Can we just not offer that one attribute

Ya, that'd work too. Maybe we leave them all in and see how well this works. I'm open for advice.

superm1:
Without some solid details of a system and kernel that breaks I expect it's a lot of "but I like s3 better ".

 Or " I don't care about the security implications of s3".

My preference is keep it all in and let's get the details out of the woodwork if there really is breakage from this.

The only AMD Lenovo systems that are going to offer this already only do s2idle...

hughsie:
Sure, point taken, thanks.

PR #4968, opened 2022-08-30
trivial: disable `security` arg when HSI disabled

superm1:
There is no point to offering the security argument to the tools if HSI was disabled
at compile time.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util.c
+++ b/src/fu-util.c
@@ -3449,6 +3449,15 @@ fu_util_security(FuUtilPrivate *priv, gchar **values, GError **error)
 	g_autoptr(GError) error_local = NULL;
 	g_autofree gchar *str = NULL;
 
+#ifndef HAVE_HSI
+	g_set_error(error,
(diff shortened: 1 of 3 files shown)


src/fu-util.c
```diff
@@ -4395,12 +4399,14 @@ main(int argc, char *argv[])
 			      /* TRANSLATORS: command description */
 			      _("Switch the firmware branch on the device"),
 			      fu_util_switch_branch);
+#ifdef HAVE_HSI
 	fu_util_cmd_array_add(cmd_array,
```
hughsie, line 4403: is this what we do for other commands? I think not hiding the argument and printing something like "is not supported" is probably more helpful.

superm1: adjusted

hughsie (approved):

Issue #4878, opened 2022-07-28
Set up Windows MSI to install fwupd daemon service


superm1:
Currently the daemon on Windows needs to be manually launched.  At some point this should be set up as a proper Windows service.

WIX documentation on this topic:
https://wixtoolset.org/documentation/manual/v3/howtos/general/install_windows_service.html

We may also want to setup the `fwupdmgr.exe` process to show a nice error message about the service not running and how to start it (IE `net start fwupd`).

CC @cwsystem76.

PR #4972, opened 2022-08-30
Add a Service for Windows

superm1:
This makes `fwupdmgr.exe` "work".  We still don't have the ability to pass CAB files or metadata between client and daemon, but it's a step in that direction.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client-sync.c
+++ b/libfwupd/fwupd-client-sync.c
@@ -518,7 +518,7 @@ fwupd_client_get_details(FwupdClient *self,
 	g_set_error_literal(error,
 			    FWUPD_ERROR,
 			    FWUPD_ERROR_NOT_SUPPORTED,
-			    "Not supported as <glib-unix.h> is unavailable");
+			    "Get Details only supported on Linux");
 	return NULL;
 #endif
 }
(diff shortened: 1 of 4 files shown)


superm1:
This should close #4878 as well.

libfwupd/fwupd-client-sync.c
```diff
@@ -515,10 +515,11 @@ fwupd_client_get_details(FwupdClient *self,
...
 #else
-	g_set_error_literal(error,
-			    FWUPD_ERROR,
-			    FWUPD_ERROR_NOT_SUPPORTED,
-			    "Not supported as <glib-unix.h> is unavailable");
+	g_set_error_literal(
+	    error,
+	    FWUPD_ERROR,
+	    FWUPD_ERROR_NOT_SUPPORTED,
+	    "fwupd_client_get_details not supported as <glib-unix.h> is unavailable");
```
hughsie, line 522: I think these should be nicer to the eye, e.g. "Getting details not supported as <glib-unix.h> is unavailable"

src/fu-util.c
```diff
@@ -4604,7 +4604,13 @@ main(int argc, char *argv[])
 
 	/* show a warning if the daemon is tainted */
 	if (!fwupd_client_connect(priv->client, priv->cancellable, &error)) {
+#ifdef _WIN32
+		/* TRANSLATORS: Error message for Windows */
```
hughsie, line 4608: lowercase E

hughsie (approved):

PR #4975, opened 2022-08-30
pci-psp: Only compile on x86_64

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
> Why not build for 32 bit too?

I guessed it would be a bit pointless; i.e. it's never going to exist...

superm1:
But it's not the actual CPU architecture it's the package I thought. If someone builds a 32 bit x86 kernel it wouldn't run on something modern?

I'm wondering if maybe we should turn off HSI entirely on 32 bit x86 instead/in addition.

PR #4980, opened 2022-08-30
Install the symlink the other way around

hughsie:
RPM based distros cannot replace a directory with a symlink without a giant hack.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
       - persist_to_workspace:
           root: .
           paths:
-            - "dist/share/doc/fwupd"
+            - "dist/share/doc"
 
   build-windows:
     docker:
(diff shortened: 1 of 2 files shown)


superm1:
I think if we do this we're going to need to do some changes to how Circle CI job saves the workspace and publishes the docs too.

hughsie:
I'm not really sure how the workspace save stuff works; how do you debug the deploy please?

superm1:
> I'm not really sure how the workspace save stuff works; how do you debug the deploy please?

I pushed what I think will work, but generally this has been painful to change.  If it fails at 1.8.5 then IMO we'll need to do a manual fixup.

superm1 (approved):

PR #4981, opened 2022-08-31
Fix lenovo-thinklmi and add a new case for it

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-context.c
+++ b/libfwupdplugin/fu-context.c
@@ -590,6 +590,7 @@ fu_context_load_hwinfo(FuContext *self, GError **error)
 	GPtrArray *guids;
 	g_autoptr(GError) error_smbios = NULL;
 	g_autoptr(GError) error_hwids = NULL;
+	g_autoptr(GError) error_bios_settings = NULL;
 
 	g_return_val_if_fail(FU_IS_CONTEXT(self), FALSE);
 	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
@@ -614,8 +615,9 @@ fu_context_load_hwinfo(FuContext *self, GError **error)
 		}
 	}
 
-	/* set it up so that udev will load firmware attributes */
 	fu_context_add_udev_subsystem(self, "firmware-attributes");
+	if (!fu_context_reload_bios_settings(self, &error_bios_settings))
+		g_debug("%s", error_bios_settings->message);
 
 	/* always */
 	return TRUE;
--- a/plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
+++ b/plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
@@ -14,13 +14,13 @@
 static gboolean
 fu_plugin_lenovo_thinklmi_startup(FuPlugin *plugin, FuProgress *progress, GError **error)
 {
-	FuContext *ctx = fu_plugin_get_context(plugin);
-
-	if (fu_context_get_bios_setting(ctx, BOOT_ORDER_LOCK) == NULL) {
+	const gchar *hwid =
+	    fu_context_get_hwid_value(fu_plugin_get_context(plugin), FU_HWIDS_KEY_MANUFACTURER);
+	if (g_strcmp0(hwid, "LENOVO") != 0) {
 		g_set_error_literal(error,
 				    FWUPD_ERROR,
 				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "unable to access needed attribute");
+				    "unsupported manufacturer");
 		return FALSE;
 	}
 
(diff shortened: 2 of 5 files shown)


src/fu-engine.c
```diff
@@ -6937,6 +6937,13 @@ fu_engine_check_firmware_attributes(FuEngine *self, FuDevice *device)
 	subsystem = fu_udev_device_get_subsystem(FU_UDEV_DEVICE(device));
 	if (g_strcmp0(subsystem, "firmware-attributes") == 0) {
 		g_autoptr(GError) error = NULL;
+		if (add) {
+			FuBiosSettings *settings = fu_context_get_bios_settings(self->ctx);
+			if (fu_bios_settings_get_count(settings) > 0) {
```
hughsie, line 6942: this seems like a single-use API call in the slow path -- can't we get the entire array and just check the length ourselves?

src/fu-engine.c
```diff
@@ -6926,7 +6926,7 @@ fu_engine_apply_default_bios_settings_policy(FuEngine *self, GError **error)
 }
 
 static void
-fu_engine_check_firmware_attributes(FuEngine *self, FuDevice *device)
+fu_engine_check_firmware_attributes(FuEngine *self, FuDevice *device, gboolean add)
```
hughsie, line 6929: maybe `add` to `added` just to make it a bit clearer

libfwupdplugin/fu-context.c
```diff
@@ -590,6 +590,7 @@ fu_context_load_hwinfo(FuContext *self, GError **error)
 	GPtrArray *guids;
 	g_autoptr(GError) error_smbios = NULL;
 	g_autoptr(GError) error_hwids = NULL;
+	g_autoptr(GError) error_fw_attributes = NULL;
```
hughsie, line 593: fw_settings, surely :)

plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
```diff
@@ -27,14 +28,29 @@ fu_plugin_lenovo_thinklmi_startup(FuPlugin *plugin, FuProgress *progress, GError
 }
 
 static void
-fu_plugin_lenovo_thinklmi_device_registered(FuPlugin *plugin, FuDevice *device)
+fu_plugin_lenovo_thinklmi_cpu_registered(FuPlugin *plugin, FuDevice *device)
+{
+	/* Ryzen 6000 doesn't support S3 even if the BIOS offers it */
+	if (fu_device_has_guid(device, "CPUID\\PRO_0&FAM_19&MOD_44")) {
```
hughsie, line 34: has_instance_id, no?

plugins/lenovo-thinklmi/fu-plugin-lenovo-thinklmi.c
```diff
@@ -55,6 +66,16 @@ fu_plugin_lenovo_thinklmi_device_registered(FuPlugin *plugin, FuDevice *device)
...
 }
 
+static void
+fu_plugin_lenovo_thinklmi_device_registered(FuPlugin *plugin, FuDevice *device)
+{
+	if (g_strcmp0(fu_device_get_plugin(device), "uefi_capsule") == 0)
+		fu_plugin_lenovo_thinklmi_uefi_capsule_registered(fu_plugin_get_context(plugin),
+								  device);
+	else if (g_strcmp0(fu_device_get_plugin(device), "cpu") == 0)
+		fu_plugin_lenovo_thinklmi_cpu_registered(fu_plugin_get_context(plugin), device);
```
hughsie, line 76: braces here perhaps

src/fu-engine.c
hughsie: Added too?

hughsie (approved):

superm1:
@mrhpearson FYI on this MR.  (I'm pretty sure this is a BIOS bug you'll want to raise that the setting is exporting anyway)

hughsie:
Yup, I was going to say too; we don't want fwupd papering over this kind of thing forever.

mrhpearson:
Just as a note - BIOS team have done the fix so you can't adjust sleep setting using WMI. It should be in the next release

superm1:
> Just as a note - BIOS team have done the fix so you can't adjust sleep setting using WMI. It should be in the next release

It's probably an infinitesimally small audience, but do they move the sleep setting out of S3 if someone happened to have set it to this using the WMI interface but upgrades to the new BIOS?

We mitigate it from fwupd due to this PR, but not to say someone couldn't do it from Windows over WMI or Linux directly in the kernel.

mrhpearson:
No idea - I can't imagine anybody did it but if they did...I guess they'd need to go back to the previous BIOS and fix it). 
I'll check and see what they say.

PR #4990, opened 2022-09-01
corsair: remove support for some devices

dushko-devx:
Remove these mice from the plugin:
    * KATAR PRO WIRELESS
    * HARPOON RGB WIRELESS

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
Ohh? How come? You don't need to remove the VID/PIDs even if you decide not to update firmware on the LVFS.

dushko-devx:
Yes I know but I cannot change vendor's decision to do it like this

dushko-devx:
It was vendor's decision to remove the unsupported devices from the plugin

hughsie:
> It was vendor's decision to remove the unsupported devices from the plugin

hughsie (approved):
> It was vendor's decision to remove the unsupported devices from the plugin

It's 100% easier to add all the devices to the quirk files and then just upload the LVFS updates for the ones you want to update than to keep adding and removing things from the quirk file. That said, your plugin, your rules.

PR #4993, opened 2022-09-01
plugins: android-boot: new plugin

DylanVanAssche:
Add a plugin for supporting Android bootloaders which are used on
all Qualcomm-based Android devices. These bootloaders are stored on
their own partitions and the partition table cannot be altered on
Qualcomm devices. This plugin supports any block device, but only
exposes the ones defined in the quirk file as updatable.

New plugin checklist

- [x] Implement FuFirmware->write() and include at least one fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass
- [x] Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc).

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


DylanVanAssche:
>  CI run of the plugin for at least one target

Do I need to adjust anything to the CI here? I think it just 'works'?

superm1:
> >  CI run of the plugin for at least one target
> 
> Do I need to adjust anything to the CI here? I think it just 'works'?

Fix the fedora build by adding to spec file

plugins/android_boot/README.md
```diff
@@ -0,0 +1,51 @@
...
+This plugin is used to update hardware that use partitions to store their firmware on.
+
+## Firmware Format
+
+The daemon will decompress the cabinet archive and extract a firmware blob as a Raw Disk File
+in the IMG format. The firmware blob will be flashed to the partition.
+
+This plugin supports the following protocol ID:
+
+* be.dylanvanassche.android_boot
```
hughsie, line 14: Is `be.dylanvanassche` appropriate? Not that I have any better ideas, but using your name for essentially a dd seems a little over the top :)

plugins/android_boot/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## GUID Generation
+
+The GUID is generated by combining the partition UUID of the block device, its label and optionally boot slot
+when using an Android A/B partitioning scheme, e.g.
+
+* `DRIVE\UUID_c49183ed-dddc-sd55-960b-64450fdsfnsd&LABEL_label&SLOT_a`
```
hughsie, line 21: 64450fdsfnsd isn't a GUID...

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+#include <string.h>
+
+#include "fu-android-boot-device.h"
+
+#define PROC_CMDLINE	 "/proc/cmdline"
+#define MAX_CMDLINE_SIZE 4096
+#define BOOT_SLOT_ARG	 "androidboot.slot_suffix"
+#define ABL_VERSION_ARG	 "androidboot.abl.revision"
+#define SERIAL_ARG	 "androidboot.serialno"
+#define UNKNOWN_VERSION	 "0.0.UNKNOWN"
```
hughsie, line 21: 0.0.0 would be more regular perhaps?

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+#include <fcntl.h>
+#include <string.h>
+
+#include "fu-android-boot-device.h"
+
+#define PROC_CMDLINE	 "/proc/cmdline"
+#define MAX_CMDLINE_SIZE 4096
+#define BOOT_SLOT_ARG	 "androidboot.slot_suffix"
+#define ABL_VERSION_ARG	 "androidboot.abl.revision"
+#define SERIAL_ARG	 "androidboot.serialno"
```
hughsie, line 20: maybe prefix these with ANDROID_BOOT_ or something?

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+#define UNKNOWN_VERSION	 "0.0.UNKNOWN"
+#define SECTOR_SIZE	 512
+
+struct _FuAndroidBootDevice {
+	FuUdevDevice parent_instance;
+	const char *label;
```
hughsie, line 26: `gchar` please

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+	gint fd;
+	char pcmd[MAX_CMDLINE_SIZE];
+	g_auto(GStrv) args = NULL;
+	g_auto(GStrv) key_value = NULL;
+
+	fd = open(PROC_CMDLINE, O_RDONLY);
+	rc = read(fd, pcmd, MAX_CMDLINE_SIZE);
+	close(fd);
+
+	if (rc > 0) {
```
hughsie, line 48: if rc < 0 g_warning then return, then you can unindent all the other bits.

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+	const char *boot_slot;
+	const char *uuid;
+	const char *version;
+	guint64 max_size;
+};
+
+G_DEFINE_TYPE(FuAndroidBootDevice, fu_android_boot_device, FU_TYPE_UDEV_DEVICE)
+
+static char *
+fu_android_boot_extract_cmdline_arg_value(const char *key)
```
hughsie, line 36: src/fu-engine.c parses cmdline too, maybe you can copy that?

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+		return FALSE;
+
+	/* set the physical ID */
+	if (!fu_udev_device_set_physical_id(FU_UDEV_DEVICE(device), "block", error))
+		return FALSE;
+
+	/* extract label and check if it matches boot slot*/
+	if (g_udev_device_has_property(udev_device, "ID_PART_ENTRY_NAME")) {
+		self->label = g_udev_device_get_property(udev_device, "ID_PART_ENTRY_NAME");
+		g_debug("Partition label: %s", self->label);
```
hughsie, line 103: you don't need this, it's in to_string()

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+		/* If the device has A/B partitioning, compare boot slot to only expose partitions
+		 * in-use */
+		if (self->boot_slot != NULL && g_str_has_suffix(self->label, self->boot_slot))
+			matches_boot_slot = TRUE;
+	}
+
+	/* extract partition UUID */
+	if (g_udev_device_has_property(udev_device, "ID_PART_ENTRY_UUID")) {
+		self->uuid = g_udev_device_get_property(udev_device, "ID_PART_ENTRY_UUID");
+		g_debug("Partition UUID: %s", self->uuid);
```
hughsie, line 117: same

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+	}
+
+	/* set max firmware size, required to avoid writing firmware bigger than partition */
+	if (g_udev_device_has_property(udev_device, "ID_PART_ENTRY_SIZE")) {
+		sectors = g_udev_device_get_property_as_uint64(udev_device, "ID_PART_ENTRY_SIZE");
+		size = sectors * SECTOR_SIZE;
+		g_debug("Partition size: %" G_GUINT64_FORMAT " bytes", size);
```
hughsie, line 124: same

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+		self->max_size = size;
+	} else {
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "device does not expose its size");
+		return FALSE;
```
hughsie, line 131: do the error case first, then unindent the success case

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_NOT_SUPPORTED,
+				    "device is on a different bootslot");
+		return FALSE;
+	}
+
+	/* set serial */
+	serial = fu_android_boot_extract_cmdline_arg_value(SERIAL_ARG);
+	fu_device_set_serial(FU_DEVICE(self), serial);
```
hughsie, line 154: you can just use device here without the FU_DEVICE(self)

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+	/* GUID based on UUID */
+	fu_device_build_instance_id(device, error, "DRIVE", "UUID", NULL);
+
+	/* GUID based on label, and UUID */
+	fu_device_build_instance_id(device, error, "DRIVE", "UUID", "LABEL", NULL);
+
+	/* GUID based on label, boot slot, and UUID */
+	fu_device_build_instance_id(device, error, "DRIVE", "UUID", "LABEL", "SLOT", NULL);
+
+	/* Success */
```
hughsie, line 174: lowercase s please

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+
+	/*
+	 * Fallback for ABL without version reporting, fwupd will always provide an upgrade in this
+	 * case. Once upgraded, the version reporting will be available and the update notification
+	 * will disappear. If version reporting is available, the reported version is set.
+	 */
+	fu_device_set_version(FU_DEVICE(self),
+			      self->version == NULL ? UNKNOWN_VERSION : self->version);
```
hughsie, line 189: i think an `if` would be clearer than a trigraph.

meson_options.txt
```diff
@@ -18,6 +18,7 @@ option('lzma', type: 'feature', description : 'LZMA support', deprecated: {'true
 option('cbor', type: 'feature', description : 'CBOR support for coSWID and uSWID')
 option('plugin_amt', type : 'feature', description : 'Intel AMT support', deprecated: {'true': 'enabled', 'false': 'disabled'})
 option('plugin_acpi_phat', type : 'feature', description : 'ACPI PHAT support', deprecated: {'true': 'enabled', 'false': 'disabled'})
+option('plugin_android_boot', type : 'feature', description : 'Android Boot plugin', deprecated: {'true': 'enabled', 'false': 'disabled'})
```
superm1, line 21: no need to include deprecated options for something new.

plugins/android_boot/README.md
superm1, line 14: If this plugin is just basically "dd", then maybe  protocol should be `org.fwupd.android_boot` (or maybe even it shouldn't be called `android_boot` but `dd`?)

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include <fcntl.h>
+#include <string.h>
+
+#include "fu-android-boot-device.h"
+
+#define PROC_CMDLINE	 "/proc/cmdline"
```
superm1, line 16: needing to read `/proc/cmdline` probably means this plugin should only run when the OS is linux (which I guess you've implicitly covered by gudev requirement)

plugins/android_boot/fu-android-boot-device.c
superm1: Or preferably - make a helper in libfwupdplugin.

plugins/android_boot/fu-android-boot-device.c
superm1: I say axe it.  It's a pointless comment

plugins/android_boot/fu-android-boot-device.c
superm1: What happens if you just don't set the version?  Is it not upgradable I guess?

plugins/android_boot/android-boot.quirk
```diff
@@ -0,0 +1,16 @@
+[BLOCK]
+Plugin = android_boot
+
+# SHIFT6mq ABL A
+[DRIVE\UUID_c49183ed-aaec-9bf5-760a-66330fbcffc1&LABEL_abl-a]
+Flags = updatable,signed-payload
```
superm1, line 6: This will only run for the supported GUIDs, but always be true where it runs right?  So perhaps these flags should just hardcode in the source in that case.

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+}
+
+static gboolean
+fu_android_boot_device_open(FuDevice *device, GError **error)
+{
+	g_autoptr(GError) error_local = NULL;
+
+	/* FuUdevDevice->open */
+	if (!FU_DEVICE_CLASS(fu_android_boot_device_parent_class)->open(device, &error_local)) {
+		if (g_error_matches(error_local, G_IO_ERROR, G_IO_ERROR_PERMISSION_DENIED)) {
```
superm1, line 204: how exactly does this happen?

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+	fu_progress_step_done(progress);
+
+	/* verify */
+	g_debug("Verifying partition");
+	if (!fu_android_boot_device_verify(self, chunks, fu_progress_get_child(progress), error))
+		return FALSE;
+	fu_progress_step_done(progress);
+
+	/* success */
+	return TRUE;
```
superm1, line 370: I think you're too verbose in both your debug messages and comments.  If you need to show these messages for debugging then you should do prefixed errors

plugins/android_boot/README.md
superm1: Or maybe even `com.google.android_boot`?

plugins/android_boot/README.md
DylanVanAssche: I took mine and random changed characters as I wasn't sure it was safe to publish this information? I can update it with the real one I have for the SHIFT6mq (see quirks)

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,441 @@
...
+				    const char *key,
+				    const char *value,
+				    GError **error)
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+	guint64 tmp = 0;
+
+	/* load from quirks */
+	if (g_strcmp0(key, "AndroidBootVersionProperty") == 0) {
+		self->version = fu_android_boot_extract_cmdline_arg_value(value);
```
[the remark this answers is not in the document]
DylanVanAssche: It is for handling the unknown version case.
If the version is still NULL after quirks are loaded, it is set to unknown.
Maybe set the unknown one as default one in init?

plugins/android_boot/README.md
DylanVanAssche: It does `dd` with some additional stuff like Android A/B bootslots.
I rather name it then `org.fwupd.android_boot` or `org.postmarketos.android_boot` --> https://postmarketos.org/

plugins/android_boot/android-boot.quirk
DylanVanAssche: Well depending on the previous bootloader stages, ABL might be signed or not.
I haven't seen much ABLs which are *not* signed, but in theory it is possible.

Updatable is just added here to only expose partitions which should be updatable for a given device, otherwise you get all partitions as devices in fwupd (which we really should avoid).

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: Most of the writing, reading, erasing comes from the MTD plugin.
I re-used this part as well, in theory you can get a permission denied for devices which are not readable or something something with AppArmor/Flatpak/SELinux stuff. This is mainly a precaution I think?

DylanVanAssche:
I replied to some comments, will resolve the others locally here as I agree with them :)

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: Hmmm I can drop these as the progress logging already prints 'write', 'erase', and 'verify' in the logs.

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: I was wondering how I can easily read this to_string() information, is there a fwupdtool/fwupdmgr command for that? Maybe I just missed it

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: If you don't set the version it cannot be upgraded.

plugins/android_boot/fu-android-boot-device.c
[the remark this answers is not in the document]
DylanVanAssche: Defaulting to unknown is a good approach.

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: Probably yes, I didn't add a check here for that, but using the internal cmdline parsing of fwupd might solve this already?

DylanVanAssche: Probably yes, I didn't add a check here for that, but using the internal cmdline parsing of fwupd might solve this already?

plugins/android_boot/README.md
hughsie: I think com.google.android_boot is probably best IMHO.

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+	return NULL;
+}
+
+static void
+fu_android_boot_device_to_string(FuDevice *device, guint idt, GString *str)
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+
+	if (self->boot_slot != NULL)
+		fu_string_append(str, idt, "BootSlot", self->boot_slot);
```
hughsie, line 73: I don't think you actually need to check for NULL here, I think fu_string_append skips when the value is NULL. Pls check tho, no coffee yet.

DylanVanAssche: Just checked, indeed NULL immediately returns, will fix.

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+fu_android_boot_device_erase(FuAndroidBootDevice *self, FuProgress *progress, GError **error)
+{
+	g_autoptr(GPtrArray) chunks = NULL;
+	gsize bufsz = fu_device_get_firmware_size_max(FU_DEVICE(self));
+	g_autofree guint8 *buf = g_malloc0(bufsz);
+
+	chunks = fu_chunk_array_new(buf, bufsz, 0x0, 0x0, 10 * 1024);
+
+	if (g_getenv("FWUPD_ANDROID_BOOT_VERBOSE") != NULL)
+		fu_dump_bytes(G_LOG_DOMAIN, "erase", g_bytes_new(buf, bufsz));
```
hughsie, line 244: you leak the GBytes, just use fu_dump_raw

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+			    FWUPD_ERROR_INVALID_FILE,
+			    "firmware too large, got 0x%x, expected <= 0x%x",
+			    (guint)g_bytes_get_size(fw),
+			    (guint)fu_device_get_firmware_size_max(device));
+		return FALSE;
+	}
+	chunks = fu_chunk_array_new_from_bytes(fw, 0x0, 0x0, 10 * 1024);
+
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED);
```
hughsie, line 322: does the daemon not print out the values to use? then you can remove GUESSED

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+
+	if (g_bytes_get_size(fw) > fu_device_get_firmware_size_max(device)) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INVALID_FILE,
+			    "firmware too large, got 0x%x, expected <= 0x%x",
+			    (guint)g_bytes_get_size(fw),
+			    (guint)fu_device_get_firmware_size_max(device));
+		return FALSE;
+	}
```
hughsie, line 318: doesn't fu_device_write_firmware do exactly this check? again coffee...

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+	gchar pcmd[ANDROID_BOOT_MAX_CMDLINE_SIZE];
+	g_auto(GStrv) args = NULL;
+	g_auto(GStrv) key_value = NULL;
+
+	fd = open(ANDROID_BOOT_PROC_CMDLINE, O_RDONLY);
+	rc = read(fd, pcmd, ANDROID_BOOT_MAX_CMDLINE_SIZE);
+	close(fd);
+
+	if (rc < 0) {
+		g_warning("Failed to read %s (%d)", ANDROID_BOOT_PROC_CMDLINE, rc);
```
hughsie, line 48: i think this should be set as a GError to let the caller handle it

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+	fu_udev_device_set_flags(FU_UDEV_DEVICE(self),
+				 FU_UDEV_DEVICE_FLAG_OPEN_READ | FU_UDEV_DEVICE_FLAG_OPEN_WRITE |
+				     FU_UDEV_DEVICE_FLAG_OPEN_SYNC);
+	fu_device_add_icon(FU_DEVICE(self), "computer");
+
+	self->label = NULL;
+	self->uuid = NULL;
+	self->version = ANDROID_BOOT_UNKNOWN_VERSION;
+	self->max_size = 0;
+	self->boot_slot = fu_android_boot_extract_cmdline_arg_value(ANDROID_BOOT_BOOT_SLOT_ARG);
```
hughsie, line 394: this is failable, so should probably go in ->probe

plugins/android_boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,410 @@
...
+static void
+fu_android_boot_device_finalize(GObject *obj)
+{
+	G_OBJECT_CLASS(fu_android_boot_device_parent_class)->finalize(obj);
+}
+
+static void
+fu_android_boot_device_init(FuAndroidBootDevice *self)
+{
+	fu_device_set_summary(FU_DEVICE(self), "Android Bootloader Device");
```
hughsie, line 378: remove Device I think

plugins/meson.build
```diff
@@ -18,6 +18,7 @@ subdir('acpi-ivrs')
 subdir('acpi-phat')
 subdir('amt')
 subdir('analogix')
+subdir('android_boot')
```
hughsie, line 21: maybe a dash not underline? I think all the others are dashed.

plugins/android_boot/README.md
DylanVanAssche: Renamed locally already to `com.google.android_boot` as this suggestion came up twice already.

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: Well it prints it out several times and complains about wrong values while point at GUESSED on the wiki. Not sure how to fix it. Any suggestions?

hughsie: Change the numbers to the ones it suggests and then take out the GUESSED line :)

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: I have seen this in other places like the MTD plugin as well. That's why I have it. Not sure either.

hughsie: If the MDT plugin implements ->prepare_firmware() then it indeed needs to do it, otherwise fu_device_prepare_firmware() does it for you.

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: This is done in other plugins like this, it expects a FuDevice and not FuAndroidBootDevice.

hughsie: Ohh sorry for being unclear; i meant `Android Bootloader` should be the summary! :)

plugins/android_boot/fu-android-boot-device.c
DylanVanAssche: Will have a look to adapt it to such a thing.

plugins/android-boot/README.md
```diff
@@ -0,0 +1,51 @@
...
+This plugin is used to update hardware that use partitions to store their firmware on.
+
+## Firmware Format
+
+The daemon will decompress the cabinet archive and extract a firmware blob as a Raw Disk File
+in the IMG format. The firmware blob will be flashed to the partition.
+
+This plugin supports the following protocol ID:
+
+* be.dylanvanassche.android_boot
```
superm1, line 14: update this here too

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,374 @@
...
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include <fcntl.h>
+#include <string.h>
+
+#include "fu-android-boot-device.h"
+
+#define ANDROID_BOOT_PROC_CMDLINE     "/proc/cmdline"
```
superm1, line 16: no longer needed

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,374 @@
...
+#define ANDROID_BOOT_MAX_CMDLINE_SIZE 4096
```
superm1, line 17: no longer needed

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,374 @@
...
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+
+	fu_string_append(str, idt, "BootSlot", self->boot_slot);
+	fu_string_append(str, idt, "Label", self->label);
+	fu_string_append(str, idt, "UUID", self->uuid);
+	fu_string_append_kx(str, idt, "MaxSize", self->max_size);
+}
+
+static gboolean
+fu_android_boot_device_probe(FuDevice *device, GError **error)
```
superm1, line 46: In this function you use a lot of parsing gudev attributes directly.  I think most of this should move to fu-udev-device.c for this device type.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,374 @@
...
+
+	/* extract boot slot if available */
+	self->boot_slot = g_hash_table_lookup(cmdline, ANDROID_BOOT_BOOT_SLOT_ARG);
+
+	/* extract label and check if it matches boot slot*/
+	if (g_udev_device_has_property(udev_device, "ID_PART_ENTRY_NAME")) {
+		self->label = g_udev_device_get_property(udev_device, "ID_PART_ENTRY_NAME");
+
+		/* Use label as device name */
+		fu_device_set_name(device, self->label);
```
superm1, line 74: like this could be done in fu-udev-device.c.  Then you would use `fu_device_get_name` and no longer need to store to `self->label`.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,374 @@
...
+#define ANDROID_BOOT_SERIAL_ARG	      "androidboot.serialno"
+#define ANDROID_BOOT_UNKNOWN_VERSION  "0.0.0"
+#define ANDROID_BOOT_SECTOR_SIZE      512
+
+struct _FuAndroidBootDevice {
+	FuUdevDevice parent_instance;
+	const gchar *label;
+	const gchar *boot_slot;
+	const gchar *uuid;
+	const gchar *version;
```
superm1, line 28: I don't think you have any need to store this here.  Just use `fu_device_set_version()` whenever you write it and `fu_device_get_version` whenever you read it.

plugins/android-boot/README.md
```diff
@@ -0,0 +1,51 @@
+# Android Bootloaders
+
+## Introduction
+
+This plugin is used to update hardware that use partitions to store their firmware on.
```
superm1, line 5: I think it would be good here to clarify why and how this is different from fastboot

plugins/android-boot/fu-android-boot-device.c
DylanVanAssche: It is to simply the `UNKNOWN_VERSION` assignment. It is set at init and only if the value is reported through the cmdline, it is assigned.

superm1: Maybe just `fu_device_set_version(dev, UNKNOWN_VERSION)` at the start then

plugins/android-boot/fu-android-boot-device.c
hughsie: I'm not so worried; if there was two plugins reading the same values perhaps, but for one I don't mind so much.

plugins/android-boot/fu-android-boot-device.c
hughsie: I also thought of this, but the daemon actually does modify the name quite a bit, e.g. it might remove the vendor name -- which isn't what you want if you are going to use it as an instance ID.

DylanVanAssche: self->label is only stored for the `to_string`, if we remove it there, it can be a local variable in `_probe`.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+
+#include <fcntl.h>
+#include <string.h>
+
+#include "fu-android-boot-device.h"
+
+#define ANDROID_BOOT_BOOT_SLOT_ARG    "androidboot.slot_suffix"
+#define ANDROID_BOOT_SERIAL_ARG	      "androidboot.serialno"
+#define ANDROID_BOOT_UNKNOWN_VERSION  "0.0.0"
+#define ANDROID_BOOT_SECTOR_SIZE      512
```
hughsie, line 19: maybe just put these values where they're used? They only seem to be used once afterall.

plugins/android-boot/fu-android-boot-device.c
hughsie: It's used in `fu_device_add_instance_strsafe(device, "LABEL", self->label);` -- I think it's safer to store tbh.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+
+	/* get kernel cmdline */
+	cmdline = fu_kernel_get_cmdline(error);
+
+	/* set the physical ID */
+	if (!fu_udev_device_set_physical_id(FU_UDEV_DEVICE(device), "block", error))
+		return FALSE;
+
+	/* extract boot slot if available */
+	self->boot_slot = g_hash_table_lookup(cmdline, ANDROID_BOOT_BOOT_SLOT_ARG);
```
hughsie, line 64: you need to g_strdup() the return values from g_hash_table_lookup() as when the cmdline is unrefd these are no longer valid.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+
+static gboolean
+fu_android_boot_device_probe(FuDevice *device, GError **error)
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+	GUdevDevice *udev_device = fu_udev_device_get_dev(FU_UDEV_DEVICE(device));
+	g_autofree gchar *serial = NULL;
+	guint64 sectors = 0;
+	guint64 size = 0;
+	GHashTable *cmdline = NULL;
```
hughsie, line 50: g_autoptr(GHashTable) pls

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+	/* failed */
+	g_set_error_literal(error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED, "quirk key not supported");
+	return FALSE;
+}
+
+static void
+fu_android_boot_device_finalize(GObject *obj)
+{
+	G_OBJECT_CLASS(fu_android_boot_device_parent_class)->finalize(obj);
+}
```
hughsie, line 329: Needs at least `g_free(self->boot_slot)` here

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+	 * Fallback for ABL without version reporting, fwupd will always provide an upgrade in this
+	 * case. Once upgraded, the version reporting will be available and the update notification
+	 * will disappear. If version reporting is available, the reported version is set.
+	 */
+	fu_device_set_version(FU_DEVICE(self), ANDROID_BOOT_UNKNOWN_VERSION);
+
+	self->label = NULL;
+	self->uuid = NULL;
+	self->max_size = 0;
+	self->boot_slot = NULL;
```
hughsie, line 356: you don't need any of these, it's malloc0'd

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+static gboolean
+fu_android_boot_device_set_quirk_kv(FuDevice *device,
+				    const gchar *key,
+				    const gchar *value,
+				    GError **error)
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+	const gchar *version = NULL;
+	guint64 tmp = 0;
+	GHashTable *cmdline = NULL;
```
hughsie, line 302: g_autoptr too

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+
+	/* load from quirks */
+	if (g_strcmp0(key, "AndroidBootVersionProperty") == 0) {
+		cmdline = fu_kernel_get_cmdline(error);
+		version = g_hash_table_lookup(cmdline, value);
```
hughsie, line 307: g_strdup too

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,372 @@
...
+
+#define ANDROID_BOOT_BOOT_SLOT_ARG    "androidboot.slot_suffix"
+#define ANDROID_BOOT_SERIAL_ARG	      "androidboot.serialno"
+#define ANDROID_BOOT_UNKNOWN_VERSION  "0.0.0"
+#define ANDROID_BOOT_SECTOR_SIZE      512
+
+struct _FuAndroidBootDevice {
+	FuUdevDevice parent_instance;
+	const gchar *label;
+	const gchar *boot_slot;
```
hughsie, line 24: not const

plugins/android-boot/fu-android-boot-device.c
DylanVanAssche: I like defines to make them self explanatory, otherwise I feel the need to write a comment for them why these values are used. I rather keep them like this.

plugins/android-boot/fu-android-boot-device.c
DylanVanAssche: Is that necessary? Once you call set_version, it `g_strdup` internally...

hughsie: Ohh no, indeed, sorry! Maybe just check the other users of g_hash_table_lookup() to be safe.

DylanVanAssche: No problem :) I checked the other usages already when I saw the first g_strdup. Only needed for the bootslot.

plugins/android-boot/fu-android-boot-device.c
hughsie: I think `androidboot.serialno` is way more descriptive and `grep`-able than `ANDROID_BOOT_SERIAL_ARG`

DylanVanAssche: But keeping the SECTOR and UNKNOWN_VERSION ones then?

hughsie: yup, that's a fine compromise :)

hughsie: Ohh, you might also want to set the version format -- talking about version strings.

DylanVanAssche: Should be already be the case: `fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_PLAIN);` :)

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,367 @@
...
+static gboolean
+fu_android_boot_device_set_quirk_kv(FuDevice *device,
+				    const gchar *key,
+				    const gchar *value,
+				    GError **error)
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+	const gchar *version = NULL;
+	guint64 tmp = 0;
+	g_autoptr(GHashTable) cmdline = NULL;
```
hughsie, line 300: i think you can reduce the scope of these, i.e. move them into the if blocks...

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,367 @@
...
+	guint64 sectors = 0;
+	guint64 size = 0;
+	g_autoptr(GHashTable) cmdline = NULL;
+
+	/* FuUdevDevice->probe */
+	if (!FU_DEVICE_CLASS(fu_android_boot_device_parent_class)->probe(device, error))
+		return FALSE;
+
+	/* get kernel cmdline */
+	cmdline = fu_kernel_get_cmdline(error);
```
hughsie, line 55: need to check for NULL and if so return FALSE.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,367 @@
...
+				    GError **error)
+{
+	FuAndroidBootDevice *self = FU_ANDROID_BOOT_DEVICE(device);
+	const gchar *version = NULL;
+	guint64 tmp = 0;
+	g_autoptr(GHashTable) cmdline = NULL;
+
+	/* load from quirks */
+	if (g_strcmp0(key, "AndroidBootVersionProperty") == 0) {
+		cmdline = fu_kernel_get_cmdline(error);
```
hughsie, line 304: check for NULL and return FALSE

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -0,0 +1,367 @@
...
+	G_OBJECT_CLASS(fu_android_boot_device_parent_class)->finalize(obj);
+	g_free(self->boot_slot);
+}
+
+static void
+fu_android_boot_device_init(FuAndroidBootDevice *self)
+{
+	fu_device_set_summary(FU_DEVICE(self), "Android Bootloader");
+	fu_device_add_protocol(FU_DEVICE(self), "com.google.android_boot");
+	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_PLAIN);
```
hughsie, line 336: it's triplet, no?

DylanVanAssche: Triplet can also be MAJOR-MINOR-DATE ?

DylanVanAssche: I thought triplet is only 3 integers?

DylanVanAssche:
(not sure why the pre-commit CI job fails, the formatting is 'fine' locally with my pre-commit hooks)

hughsie:
> (not sure why the pre-commit CI job fails, the formatting is 'fine' locally with my pre-commit hooks)

Try `xargs clang-format -i plugins/android-boot/*.c` -- the precommit clang format integration is good, but not perfect when `--amend` gets involved.

hughsie:
FYI, I think with passing CI we should approve and merge this!

hughsie (approved):

DylanVanAssche:
Debian CI seems to be broken here as it behave similar in other PRs, auto merge won't work then :P

PR #4994, opened 2022-09-01
Adjust how we check for python and markdown

superm1:
Adopt some of what was demonstrated in https://gitlab.gnome.org/GNOME/gi-docgen/-/merge_requests/166 here.  We can probably drop some of this when gi-docgen release bumps up include that.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
It all still seems a bit Rube Goldberg, but I think this is better than what we had before, thanks.

PR #4995, opened 2022-09-02
Add shared functionality to get the parsed kernel cmdline

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-kernel.c
+++ b/libfwupdplugin/fu-kernel.c
@@ -17,6 +17,7 @@
 #include "fu-common.h"
 #include "fu-kernel.h"
 #include "fu-path.h"
+#include "fu-string.h"
 #include "fu-version-common.h"
 
 /**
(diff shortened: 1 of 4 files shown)


libfwupdplugin/fu-kernel.c
```diff
@@ -218,3 +219,40 @@ fu_kernel_reset_firmware_search_path(GError **error)
...
+ * @error: (nullable): optional return location for an error
+ *
+ * Loads all the kernel /proc/cmdline key/values into a hash table.
+ *
+ * Returns: (transfer container) (element-type utf8 utf8): keys from the kernel command line
+ *
+ * Since: 1.8.5
+ **/
+GHashTable *
+fu_kernel_get_proc_cmdline(GError **error)
```
superm1, line 234: I think this should have some `#ifndef _WIN32` guard

hughsie: Does macos have /proc/cmdline? The one in fu-engine.c has #ifdef __linux__

superm1: I'm pretty sure `/proc/cmdline` is a linux only concept.
Windows has https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getcommandlinea, but that's not really the same thing as `/proc/cmdline`, it's more like `/proc/self/cmdline`.

hughsie: Gotcha, thanks. Fixed.

superm1 (approved):

DylanVanAssche:
Will use this in #4993 once merged.

hughsie:
Done!

DylanVanAssche (approved):
LGTM

PR #5001, opened 2022-09-05
Wip/hughsie/tbt only digital

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 more `fwupd: library-not-linked-against-libc` warnings I think.

superm1:
> @superm1 more `fwupd: library-not-linked-against-libc` warnings I think.

Can you just add some more exceptions?I think we should ignore em.

hughsie:
> Can you just add some more exceptions

I tried, but IIUC, we're already matching all of them:

   fwupd: library-not-linked-against-libc usr/lib/*/fwupd-plugins-*/*

PR #5002, opened 2022-09-06
Add tests that emulate a USB device

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
Needs https://github.com/hughsie/libgusb/pull/69

docs/env.md
```diff
@@ -27,6 +27,7 @@ with a non-standard filesystem layout.
 
 * `FWUPD_MACHINE_KIND` can be used to override the detected machine type, e.g. `physical`, `virtual`, or `container`
 * `FWUPD_HOST_EMULATE` can be used to load test data from `/usr/share/fwupd/host-emulate.d`, e.g. `thinkpad-p1-no-iommu.json.gz`
+* `FWUPD_BACKEND_SAVE` can be used to save fake backend devices to a file. Use `sudo fwupdtool get-devices` to generate.
```
superm1, line 30: I would think it's better to just add a new argument to fwupdtool.

Maybe generate-emulation? And it would then generate as much emulation as possible.

hughsie: > I would think it's better to just add a new argument to fwupdtool.

The tricky bit is that you want to capture the events of things like dumping the firmware or updating the device (that's the interesting bit to emulate, not just the enumeration) -- so it really needs to be a flag to say "save the log for the action" rather than just dumping the coldplugged list. Maybe something like --save-events=PLUGIN-NAME might be better than an env var tho.

superm1: Yeah I like that better

hughsie: Done; will push when all the CI stuff is fixed.

superm1:
> Needs https://github.com/hughsie/libgusb/pull/69

Considering this, I think it's worth bumping `subprojects/gusb.wrap` too.

libfwupdplugin/fu-usb-device.c
```diff
@@ -267,13 +269,352 @@ fu_usb_device_open(FuDevice *device, GError **error)
...
+			dsinfo->total_length,
+			dsinfo->vendor_code,
+			dsinfo->alt_code);
+	}
+
+	/* sort by platform_ver, highest first */
+	g_ptr_array_sort(dsinfos, fu_usb_device_sort_by_dsinfo_platform_version_cb);
+
+	/* use Windows 8.1 as a baseline */
+	if (g_strcmp0(guid, USB_DEVICE_PLATFORM_CAPABILITY_UUID_MICROSOFT) == 0) {
```
superm1, line 509: I don't think you want to mix and match with the Microsoft one.  If this is going to be a program tied specifically to WWCB/LVFS certifications then it needs to be all in on a fwupd GUID.

superm1:
`docs/ds20.md` should probably percolate up to libfwupdplugin docs and index.html docs too.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+    bRequest: {value of bVendorCode in the BOS descriptor)
+    wValue: 0x00
+    wIndex: 0x07
+    wLength: {value of wLength in the BOS descriptor)
+
+The device should return something like:
+
+    50 6c 75 67 69 6e 3d 64 66 75 0a 49 63 6f 6e 3d 63 6f 6d 70 75 74 65 72 0a 00 00 00 00 00 00 00
+
+...which is the UTF-8 quirk data, e.g.
```
superm1, line 104: This means _USB devices_ have a new attack surface area of the quirk parser which otherwise was just reading files distributed with fwupd.

I think we'll want to be doing something in the code to ensure that the plugins that can be used are only USB plugins.

I also think that there needs to be a required comparison that a USB device's `idVendor` can only be used with _certain_  USB plugins.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+    Plugin=dfu
+    Icon=computer
```
superm1, line 107: The spec is missing information about how someone would "come up with" the fields to put into their descriptor.  Like a development flow.  I would think everyone should "start out" with a flat quirk file while under development and when they prove it's all working, move the quirk content into the descriptor.  Which begs a `contrib/` script that turns a quirk file into this descriptor format!

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+* The VID/PID is not necessarily the information that we match to assign the firmware stream, and so this would have to be a separate facet of metadata -- which may have to include quirks too.
+
+Matching devices to plugins should be thought of as *orthogonal* to matching firmware to devices.
+To simplify deployment we should allow the device itself to specify the plugin to use and optionally additional quirk data.
+
+## Prior Art
+
+### Hardcoded Class & Subclass
+
+We already support two plugins that use the USB class code, rather than the exact instance ID with the VID/PID.
```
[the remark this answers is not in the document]
hughsie: Fixed, thanks.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+* We don't want to only **enumerate** hardware that has updates on the main LVFS remote.
+* A *lot* of machines using fwupd now have never connected to the internet and we still want to enumerate hardware for audit and verification.
+* Re-enumerating all physical hardware (to get the latest quirks) when updating the metadata isn't simple or quick.
+* The VID/PID is not necessarily the information that we match to assign the firmware stream, and so this would have to be a separate facet of metadata -- which may have to include quirks too.
+
+Matching devices to plugins should be thought of as *orthogonal* to matching firmware to devices.
+To simplify deployment we should allow the device itself to specify the plugin to use and optionally additional quirk data.
+
+## Prior Art
```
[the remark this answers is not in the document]
hughsie: It's less prior art and more of a _thought process_ or much as I hate the word, a _journey_. Maybe it should just be part of the introduction?

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+The [Microsoft OS Descriptors 2.0 Specification](https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/microsoft-os-2-0-descriptors-specification) is more helpful as it defines a new device capability that can return a variable length chunk of data, using a UUID as a key.
+We do need to use the `BOS` *“Binary Object Store”* descriptor, which is only supported for devices using USB 2.1 and newer, although that’s probably the vast majority of devices now.
+Usually the BOS descriptor is used for Wireless USB details, USB 2.0 extensions, SuperSpeed USB connection details and a Container ID.
+
+Usefully, the BOS descriptor does give us is the ability to use the platform capability descriptor, which is `bDevCapabilityType=0x05`. For UUID `D8DD60DF-4589-4CC7-9CD2-659D9E648A9F`, this is identified as a structured blob of data Windows usually uses to put workarounds like the suspend mode of the device and that kind of thing.
+
+We could certainly create a new `bDevCapabilityType` which would allow us to store a binary blob (e.g. `Plugin=foobarbaz\nFlags=QuirkValueHere\n`) but that would be out-of-specification and difficult for vendors to implement.
+
+## Proposal
```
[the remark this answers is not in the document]
hughsie: Fixed, ta.

docs/ds20.md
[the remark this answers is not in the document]
superm1: Yeah, I think it's useful to draw allusions to and explain how you got to where you are.  I just don't think 99% of the people who consume this spec need to see it immediately.  I am a fan of put the meat in front.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+Devices create a new platform device capability BOS descriptor with `bDevCapabilityType=0x05` and a fwupd-specific UUID, specifically `010aec63-f574-52cd-9dda-2852550d94f0`.
```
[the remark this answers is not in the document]
hughsie: I did exactly that. What about adding `This UUID is generated type-5 SHA-1 hash of the word fwupd with a DNS namespace.` or something like that?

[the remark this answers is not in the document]
superm1: Yup, that's perfect to add.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+    1C 10 05 00 63 ec 0a 01 74 f5 cd 52 9d da 28 52 55 0d 94 f0 05 08 01 00 20 00 2a 00
+    |  |  |  |  \---------------------------------------------/ \--------/  \---/ |  |
+    |  |  |  |                ^-PlatformCapabilityUUID            |   wLength-^   |  |
+    |  |  |  ^-bReserved                                 dwVersion^   bVendorCode-^  |
+    |  |  ^-bDevCapability                                                bAltEnumCmd^
+    |  ^-bDescriptorType
+    ^-bLength
+
+The `dwVersion` encoded here is fwupd `1.8.5`, which is also the first version that supports BOS DS20 descriptors.
```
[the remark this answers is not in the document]
hughsie: Done.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+Using this custom UUID will ensure that Microsoft Windows does not try to parse the capability descriptor as a *Microsoft OS 2.0 descriptor set*.
+
+## Implementation
+
+Create a BOS DS20 descriptor such as:
+
+    1C 10 05 00 63 ec 0a 01 74 f5 cd 52 9d da 28 52 55 0d 94 f0 05 08 01 00 20 00 2a 00
+    |  |  |  |  \---------------------------------------------/ \--------/  \---/ |  |
+    |  |  |  |                ^-PlatformCapabilityUUID            |   wLength-^   |  |
+    |  |  |  ^-bReserved                                 dwVersion^   bVendorCode-^  |
```
superm1, line 76: I noticed you didn't explain `bVendorCode`.  What about using `bVendorCode` to designate a numerical value corresponding to a specific USB plugin or group of plugins?  We are of course limited to 255 USB plugins, but that _seems_ reasonable to me.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+    |  |  ^-bDevCapability                                                bAltEnumCmd^
+    |  ^-bDescriptorType
+    ^-bLength
+
+The `dwVersion` encoded here is fwupd `1.8.5`, which is also the first version that supports BOS DS20 descriptors.
+A `dwVersion` of 0x0 will match any fwupd version, but should probably not be used unless required.
```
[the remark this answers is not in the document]
hughsie: Good idea, fixed.

docs/ds20.md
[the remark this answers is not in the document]
hughsie: Done.

libfwupdplugin/fu-usb-device.c
hughsie: Right, this was added for the future; there are a few fields in the MS descriptor that mean we don't have to assign the composite devices (e.g. docks) in such a manual way. i.e. it's useful, but not for this PR.

docs/ds20.md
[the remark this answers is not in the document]
hughsie: Rearranged like you suggest.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+    50 6c 75 67 69 6e 3d 64 66 75 0a 49 63 6f 6e 3d 63 6f 6d 70 75 74 65 72 0a 00 00 00 00 00 00 00
+
+...which is the UTF-8 quirk data, e.g.
+
+    Plugin=dfu
+    Icon=computer
+
+## Conclusion
+
+New devices that implement a fwupd BOS DS20 descriptor can be updated without having to update the runtime version of fwupd, assuming the device is updatable by the existing vendor plugin.
```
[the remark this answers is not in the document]
hughsie: Done.

docs/ds20.md
hughsie: It's there:

    bRequest: {value of bVendorCode in the BOS descriptor}

It's explained in much more detail in https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/microsoft-os-2-0-descriptors-specification which is probably some good bedtime reading. It's way more flexible as it defines the bRequest so you can return almost unlimited data if you want to.

superm1: Yeah I saw that, but that's how the "client" responds, not the value that the vendor should be placing _into_ the descriptor.  So I was thinking we could prescribe the values there.

The Microsoft one is tangential, remember it's a fwupd spec!

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+One suggestion would be to put this quirk information into the existing update metadata provided by the configured remote, but this has a few problems:
+
+* We don't want to only **enumerate** hardware that has updates on the main LVFS remote.
+* A *lot* of machines using fwupd now have never connected to the internet and we still want to enumerate hardware for audit and verification.
+* Re-enumerating all physical hardware (to get the latest quirks) when updating the metadata isn't simple or quick.
+* The VID/PID is not necessarily the information that we match to assign the firmware stream, and so this would have to be a separate facet of metadata -- which may have to include quirks too.
+
+Matching devices to plugins should be thought of as *orthogonal* to matching firmware to devices.
+To simplify deployment we should allow the device itself to specify the plugin to use and optionally additional quirk data.
```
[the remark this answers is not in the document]
hughsie: I've made it a whole lot less colloquial, I think.

docs/ds20.md
hughsie: Right, it's tangential -- but I think it makes a lot of sense to copy the API design -- we need to make this as least weird as possible for vendors. They're already familiar with the _control transfer with buffer of data_ thing and it's useful to have similar version semantics.

superm1: In that case we need to specify what exactly we **want** them to put in that field in this spec, even if it's similar to the other.

docs/ds20.md
hughsie: How about `Vendors using this API should use an existing fwupd quirk file to validate the fields needed in the DS20 control transfer buffer`. I'm not 100% sure about a ./contrib/generate-descriptor.py as it's going to need almost everything specified -- and in some firmware you just want `00 63 ec 0a` i.e. not the first three bytes as the descriptor is set up in a common way.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+We could certainly create a new `bDevCapabilityType` which would allow us to store a binary blob (e.g. `Plugin=foobarbaz\nFlags=QuirkValueHere\n`) but that would be out-of-specification and difficult for vendors to implement.
+
+## Proposal
+
+Devices create a new platform device capability BOS descriptor with `bDevCapabilityType=0x05` and a fwupd-specific UUID, specifically `010aec63-f574-52cd-9dda-2852550d94f0`.
+Using this custom UUID will ensure that Microsoft Windows does not try to parse the capability descriptor as a *Microsoft OS 2.0 descriptor set*.
+
+## Implementation
+
+Create a BOS DS20 descriptor such as:
```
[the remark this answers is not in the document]
hughsie: Done.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+    |  |  ^-bDevCapability                                                bAltEnumCmd^
+    |  ^-bDescriptorType
+    ^-bLength
+
+The `dwVersion` encoded here is fwupd `1.8.5`, which is also the first version that supports BOS DS20 descriptors.
+A `dwVersion` of 0x0 will match any fwupd version, but should probably not be used unless required.
+
+The BOS descriptors are sorted by the requester and can appear in any order.
+The descriptor with the highest `dwVersion` that is not newer than the current running daemon version is used.
+This allows devices to set different quirks depending on the fwupd version, although in practice this should not be required.
```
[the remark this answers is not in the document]
hughsie: > What happens when a fix lands in 1.8.4 and backports to 1.7.8?

It's a good question, and one that's pretty common in everything that uses semvers and has multiple supported parallel development streams. The same can be said about RHEL 9.2, RHEL 8.7, etc. My gut instinct is we keep it simple and in the same way we don't backport plugins (in general) and we codify we don't backport features, only fixes.

docs/ds20.md
hughsie: I've done the UTF-8 thing. I've also reviewed all the quick parser vfuncs, and we're doing a really good job with things like `fu_strtoull` rather than doing unsafe parsing of value. I'm not super worried about it tbh.

docs/ds20.md
```diff
@@ -0,0 +1,111 @@
...
+
+The device should return something like:
+
+    50 6c 75 67 69 6e 3d 64 66 75 0a 49 63 6f 6e 3d 63 6f 6d 70 75 74 65 72 0a 00 00 00 00 00 00 00
+
+...which is the UTF-8 quirk data, e.g.
+
+    Plugin=dfu
+    Icon=computer
+
```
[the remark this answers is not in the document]
hughsie: > fwupdtool get-fwupd-bos-d20

What about if we included it in the `->to_string()` output? i.e. it would display in the --verbose output of `fwupdtool get-devices`.

docs/ds20.md
superm1: Even putting it into UTF-8 can be messed up if someone tries to do it by hand.  I think it's also a good litmus test to have something in fwupd source that is a unit test like this:
1. Do something needing a quirk file.
2. Use helper script that converts into a device replay buffer
3. Remove the quirk file
4. Use the device replay buffer to interact with it
5. verify it works
6. Modify the replay buffer to break it in all the common ways.

The common ways that come to mind for me from this discussion:

- Lower version than ours
- Pass in a bad GUID
- Pass in invalid quirk data

docs/ds20.md
superm1: > I've done the UTF-8 thing. I've also reviewed all the quick parser vfuncs, and we're doing a really good job with things like `fu_strtoull` rather than doing unsafe parsing of value. I'm not super worried about it tbh.

OK sounds good.  Hopefully the fuzzers can catch anything remaining then.

docs/ds20.md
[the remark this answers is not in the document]
superm1: But so even if you only backport fixes you get caught up in this mess.  If there is a specific **fix** that you're depending upon for your device to upgrade then you can't be using semver on the backport anymore.

docs/ds20.md
hughsie: > Use the device replay buffer to interact with it

I think we're doing that now in the new self tests.

> Lower version than ours

You mean lower fwupd version?

>  Pass in a bad GUID

You mean like a GUID specified like `Guid = FOOBARBAZ`?

>  Pass in invalid quirk data

Got any ideas on what `invalid` might be?

docs/ds20.md
hughsie: Fuzzers, hmm, indeed. Wilfix.

docs/ds20.md
hughsie: Back a bit; the descriptor just says *hit this control transfer to get the raw data` -- it's literally just the UTF-8 data rather than any specific encoding or framing.

docs/ds20.md
superm1: > You mean lower fwupd version?

Yeah

> You mean like a GUID specified like Guid = FOOBARBAZ?

I mean like not the fwupd GUID.  If there is only one GUID used and we don't find it in the list but expected to see it throw an error up.

> Got any ideas on what invalid might be?

A plugin name that doesn't exist is probably the easiest to model.

docs/ds20.md
superm1: But who decides the value should be `0x2a`?  I'm Mario's new dongle shop and I make a new device.  How do I know what to put in that field?
That's what is missing from the spec.

hughsie: Gotcha. What about `A suitable bVendorCode should be chosen that is not used for existing device operation`?

superm1: Yeah that's fine

docs/ds20.md
```diff
@@ -0,0 +1,112 @@
+# fwupd BOS DS20 Specification
+
+## Introduction
+
+When `fwupd` starts it enumerates all PCI and USB hardware and generates *Instance IDs* based on the reported vendor and product so that it can match the device to a plugin.
+The plugin knows how to communicate with the device (for instance using vendor-specific USB control transfers) and also knows how to parse the firmware and deliver it to the device.
+
+Before a vendor can ship a firmware on the LVFS to a machine using Linux (or ChromeOS) the fwupd package must be updated so that it knows what plugin to use for that specific device.
```
superm1, line 8:
```suggestion
Before a vendor can ship a firmware on the LVFS to a machine using Linux (or ChromeOS) the fwupd package often must be updated so that it knows what plugin to use for that specific device.
```

docs/ds20.md
```diff
@@ -0,0 +1,112 @@
...
+At this point other overridden *quirk* data can also be set. For instance, the icon, long summary or even per-plugin flags that change device behavior.
+For example `colorhug.quirk`:
+
+    [USB\VID_273F&PID_1001]
+    Plugin = colorhug
+    Flags = self-recovery
+    Icon = colorimeter-colorhug
```
superm1, line 15: You might send a link out to https://fwupd.github.io/libfwupdplugin/class.Quirks.html

docs/ds20.md
hughsie: Done.

docs/ds20.md
hughsie: Fixed.

docs/ds20.md
hughsie: Agreed, added.

docs/ds20.md
```diff
@@ -0,0 +1,112 @@
...
+    └──────────bLength
+
+The `dwVersion` encoded here is fwupd `1.8.5`, which is also the first version that supports BOS DS20 descriptors.
+
+The BOS descriptors are sorted by the requester and can appear in any order.
+The descriptor with the highest `dwVersion` that is not newer than the current running daemon version is used.
+This means that `dwVersion` is effectively the minimum version of fwupd that should read the descriptor.
+This allows devices to set different quirks depending on the fwupd version, although in practice this should not be required.
+
+* The `dwVersion` parameter **must not** be zero.
```
superm1, line 57: how about must be 1.8.5 or newer (to make it clear that we don't support it until such a version)

docs/ds20.md
hughsie: Fixed.

docs/ds20.md
hughsie: Fixed, ta.

src/fu-self-test.c
```diff
@@ -3176,6 +3178,162 @@ _plugin_device_register_cb(FuPlugin *plugin, FuDevice *device, gpointer user_dat
...
+		g_autofree gchar *str2 = NULL;
+		g_autofree gchar *str = NULL;
+		g_autoptr(GByteArray) buf2 = g_byte_array_new();
+		g_autoptr(GByteArray) buf = g_byte_array_new();
+
+		/* bReserved */
+		fu_byte_array_append_uint8(buf, 0x0);
+
+		/* PlatformCapabilityUUID */
+		ret = fwupd_guid_from_string("010aec63-f574-52cd-9dda-2852550d94f0",
```
superm1, line 3215: wait, isn't it supposed to be `fwupd_guid_from_string("fwupd"`?

src/fu-self-test.c
```diff
@@ -3176,6 +3178,162 @@ _plugin_device_register_cb(FuPlugin *plugin, FuDevice *device, gpointer user_dat
...
+	g_autoptr(GPtrArray) possible_plugins = NULL;
+	g_autoptr(JsonParser) parser = json_parser_new();
+
+#if !G_USB_CHECK_VERSION(0, 4, 0)
+	g_test_skip("GUsb version too old");
+	return;
+#endif
+
+	/* generate the fake base64 chunks in the usb-devices.json file */
+	if (g_getenv("FWUPD_GENERATE_DS20") != NULL) {
```
superm1, line 3203: why is this gated by the env variable?

src/fu-self-test.c
```diff
@@ -3176,6 +3178,162 @@ _plugin_device_register_cb(FuPlugin *plugin, FuDevice *device, gpointer user_dat
...
+					     &guid,
+					     FWUPD_GUID_FLAG_MIXED_ENDIAN,
+					     &error);
+		g_assert_no_error(error);
+		g_assert_true(ret);
+		g_byte_array_append(buf, (const guint8 *)&guid, sizeof(guid));
+
+		/* CapabilityData */
+		fu_byte_array_append_uint32(buf, fwupd_version, G_LITTLE_ENDIAN);
+		fu_byte_array_append_uint16(buf, 32, G_LITTLE_ENDIAN); /* TotalLength */
```
superm1, line 3225: can you instead build `buf2` first and do `strlen(buf2)`?

libfwupdplugin/fu-usb-device.c
superm1: Shouldn't it be pulled out then?

libfwupdplugin/fu-usb-device.c
```diff
@@ -267,13 +269,381 @@ fu_usb_device_open(FuDevice *device, GError **error)
...
+				    G_IO_ERROR,
+				    G_IO_ERROR_NOT_SUPPORTED,
+				    "invalid platform version 0x%08x, expected >= 0x%08x",
+				    dsinfo->platform_ver,
+				    (guint)USB_DEVICE_PLATFORM_CAPABILITY_VERSION_FWUPD_LOWEST);
+				return FALSE;
+			}
+
+			/* dwVersion is effectively the minimum version of fwupd */
+			if (dsinfo->platform_ver <= fwupd_version) {
```
superm1, line 568: rather than run through all of them and parse each one - why not find the one with the biggest matching version and just parse once?

superm1: Or at least only fail on errors if they ALL fail to parse.  I think it's painting into a corner if you try to parse even older versions and say we have a quirk schema change in newer versions.

src/fu-self-test.c
hughsie: the 32 is usually the control transfer buffer size -- it's normally just padded out with NULs.

src/fu-self-test.c
hughsie: I think all this can do away if/when we have the *generate-ds20-blob.py* script.

src/fu-self-test.c
hughsie: fwupd_guid_from_string() takes a GUID and parses it into a `fwupd_guid_t` blob. Be we certainly could do `fwupd_guid_hash_string()` to get the former.

libfwupdplugin/fu-usb-device.c
hughsie: > why not find the one with the biggest matching version and just parse once?

Isn't that what it's doing now with the g_ptr_array_sort and the `return TRUE` in the middle of the loop?

superm1: Oh! I missed that up above, yeah carry on.

docs/ds20.md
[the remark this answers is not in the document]
hughsie: I think in that case I'd rather support firmware level requirements, e.g. `fwupd > 1.8.2 OR (fwupd >= 1.7.9 AND fwupd < 1.8.0)`. The worst case for quirks is that the daemon will spit out a g_warning() I think. If it's worse than that (e.g. device hangs on custom control transfer) then we tell the vendor to only have the requirement on the new branch.

docs/ds20.md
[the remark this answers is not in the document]
hughsie: Here's a thought; we move the BOS descriptor parsing to a new FuUsbDeviceFwPlatform class that subclasses FuUsbDevicePlatform (that subclasses FuFirmware), and the FuUsbDeviceMsPlatform for the MS parts. We add all of them to a GPtrArry in FuUsbDevice->priv->platform_descriptors, and then call ->to_string() on them when dumping the device. This is something I could have used last night for the fuzzing experiment....

I think that breaks up that massive file a bit too. The other way would be to create a FuUsbDevice child (of the original FuUsbDevice) that parses the file on write_firmware() or something, but that's going to make incorporate break, and the debug output look weird -- and it's weird to parse in a fake write.

libfwupdplugin/fu-usb-device.c
hughsie: I split it out into a new file.

src/fu-self-test.c
hughsie: `fwupdtool firmware-build` to the rescue.

src/fu-self-test.c
hughsie: I've removed this whole chunk thanks to `fwupdtool firmware-build`

docs/ds20.md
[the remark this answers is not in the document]
superm1: Wow cool! That sounds like a workable flow.

I would still suggest not to use dd though for that last step however. For development purposes don't you want to be able to populate the quirk file with a header to indicate the GUID?

So if you leave a contrib script that "converts" the quirk file it can strip the header, or even take an argument of the GUID in the header to pull out a "single" device from a quirk file...

superm1:
> > Needs https://github.com/hughsie/libgusb/pull/69
> 
> Considering this, I think it's worth bumping `subprojects/gusb.wrap` too.

As it's merged please put a subproject on the commit. It will make it easier to pull it down and run this all through CI.

Before this next fwupd release tags you'll of course want to tag a new libgusb and update to a tag tho

docs/ds20.md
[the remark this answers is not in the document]
hughsie: I just wrote `contrib/generate-ds20.py fw-ds20.quirk --bufsz 32` which does all that you say :)

docs/ds20.md
superm1: Sounds good!

docs/ds20.md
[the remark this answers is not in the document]
superm1: OK so then if it's purely firmware metadata level requirements then I think this needs to be positioned differently.
Right now it's only probe if you're dwVersion 1.8.5 or later, but it should be able to probe no _differently_ 1.8.5, 1.9.0, or 2.0.0.

The key is whether it can match to a device and accept an update and that is tied entirely to the CAB then.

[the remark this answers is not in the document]
hughsie: I think the idea is that if you have a `foobarbaz` plugin added in 1.8.6 that then added a `FooBarBaz` quirk entry in 1.8.9, you'd have:

    dwVersion: 1.8.9
    data: Plugin=dfu\nFooBarBaz=whatever

    dwVersion: 1.8.5
    data: Plugin=foobarbaz

That way if you're using 1.8.9 or above then you'd use the first ds, and if you're using 1.8.6 you'd be using the second ds. If you're running 1.8.5 then you'd match neither.

contrib/generate-ds20.py
```diff
@@ -0,0 +1,75 @@
...
+    sys.exit(return_code)
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-b",
+        "--bufsz",
+        type=int,
```
superm1, line 37: you should add a help string for this

contrib/generate-ds20.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+    parser.add_argument("-i", "--instance-id", type=str)
```
superm1, line 39: you should add a help string for this

contrib/generate-ds20.py
hughsie: Done.

contrib/generate-ds20.py
hughsie: Fixed, and I've made it a a bit more resistant to doing the wrong thing.

contrib/generate-ds20.py
```diff
@@ -0,0 +1,73 @@
...
+        config.read(args.filename)
+    except configparser.MissingSectionHeaderError:
+        print("not a quirk file")
+        sys.exit(1)
+
+    # fall back to the default if there is only one device in the quirk file
+    if not args.instance_id:
+        sections = config.sections()
+        if len(sections) != 1:
+            print("multiple devices found, use --instance-id to choose")
```
superm1, line 46: What do you think about printing them in the case of  this error message?

contrib/generate-ds20.py
```diff
@@ -0,0 +1,73 @@
...
+            sys.exit(1)
+        args.instance_id = sections[0]
+
+    # create the smallest kv store possible
+    lines: List[str] = []
+    try:
+        for key in config[args.instance_id]:
+            if key in ["Children", "Inhibit", "Issue", "Children"]:
```
superm1, line 54: you have Children twice, but also why don't you want this key?  I would think that composite devices could still benefit just the same with this.

hughsie: Yup, I guess you're right. Both children removed.

contrib/generate-ds20.py
hughsie: Something like:

    multiple devices found, use --instance-id to choose between USB\VID_273F&PID_1000, USB\VID_04D8&PID_F8DA, USB\VID_273F&PID_1001, USB\VID_273F&PID_1004, USB\VID_273F&PID_1005, USB\VID_273F&PID_1007, USB\VID_273F&PID_1006

-- or a bullet list? That might be easier to copy-paste.

superm1: Bullet list all the way. What I'm thinking here was all about the copy paste indeed!

hughsie: $ ./contrib/generate-ds20.py plugins/colorhug/colorhug.quirk  --bufsz 235
	Multiple devices found, use --instance-id to choose between:
	 • USB\VID_273F&PID_1000
	 • USB\VID_04D8&PID_F8DA
	 • USB\VID_273F&PID_1001
	 • USB\VID_273F&PID_1004
	 • USB\VID_273F&PID_1005
	 • USB\VID_273F&PID_1007
	 • USB\VID_273F&PID_1006

superm1: That looks good.  I'll leave one more usability comment separate

contrib/generate-ds20.py
```diff
@@ -0,0 +1,73 @@
...
+import base64
+from typing import List
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-b",
+        "--bufsz",
```
superm1, line 22: How about making this argument optional and default to a rounded number based on quirk data?
Then you can print at the end that a buffer of X bytes was used, if you need it to be bigger, use the `--buffsz` argument

contrib/generate-ds20.py
```diff
@@ -0,0 +1,73 @@
...
+    # pad to the buffer size
+    buf: bytes = "\n".join(lines).encode()
+    if len(buf) > args.bufsz:
+        print("quirk data is larger than bufsz")
+        sys.exit(1)
+    buf = buf.ljust(args.bufsz, b"\0")
+
+    # success
+    print("DS20 descriptor control transfer data:")
+    print(" ".join(["{:02x}".format(val) for val in list(buf)]))
```
superm1, line 72: Knowing a lot of people do their development on Windows and go back and forth what happens if you throw Windows line endings in here?  Does it get upset?  It might be worth stripping them first as a precaution.

docs/ds20.md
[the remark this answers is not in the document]
superm1: But aren't unknown quirks just ignored by older versions?

docs/ds20.md
```diff
@@ -0,0 +1,132 @@
...
+The device should return something like:
+
+    50 6c 75 67 69 6e 3d 64 66 75 0a 49 63 6f 6e 3d 63 6f 6d 70 75 74 65 72 0a 00 00 00 00 00 00 00
+
+...which is the UTF-8 quirk data, e.g.
+
+    Plugin=dfu
+    Icon=computer
+
+## See Also
```
superm1, line 81:
```suggestion
## Workflow
```

libfwupdplugin/fu-usb-device-ds20.c
```diff
@@ -0,0 +1,318 @@
...
+		fu_dump_raw(G_LOG_DOMAIN, "PlatformCapabilityOs20", buf, actual_length);
+
+	/* FuUsbDeviceDs20->parse */
+	blob = g_bytes_new_take(g_steal_pointer(&buf), actual_length);
+	return klass->parse(self, blob, device, error);
+#else
+	g_set_error_literal(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "Not supported as <gudev.h> is unavailable");
```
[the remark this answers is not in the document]
hughsie: It's GUsb frustratingly

docs/ds20.md
```diff
@@ -0,0 +1,132 @@
...
+
```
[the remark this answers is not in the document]
hughsie: Normally it's the buffer size for the microcontroller, i.e specified in a datasheet somewhere

contrib/generate-ds20.py
hughsie: I think we need to strip on the python tool, and in the ds20 parser tbh

docs/ds20.md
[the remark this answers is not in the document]
superm1: I guess make that explicitly mentioned then

libfwupdplugin/fu-usb-device-ds20.c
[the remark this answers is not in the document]
hughsie: Fixed.

docs/ds20.md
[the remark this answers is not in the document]
hughsie: How about `The maximum buffer size is typically hardcoded for the device microcontroller and specified in a datasheet.`?

src/fu-tool.c
```diff
@@ -3532,6 +3561,15 @@ main(int argc, char *argv[])
 	     N_("Filter with a set of device flags using a ~ prefix to "
 		"exclude, e.g. 'internal,~needs-reboot'"),
 	     NULL},
+	    {"save-backends",
```
superm1, line 3564: don't forget to add bash completion for this

hughsie: Good catch, fixed.

docs/ds20.md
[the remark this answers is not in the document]
superm1: Sounds good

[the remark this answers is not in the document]
hughsie: Fixed, ta.

contrib/generate-ds20.py
hughsie: The python tool didn't need anything, `configparser.ConfigParser()` just does the right thing. I've added a check to fail to load the descriptor if win32 line endings are detected; we want the descriptor to be as small as possible, and the parser to be as strict as possible. I've also added this as a note in the spec too. Something like `The UTF-8 quirk data must **not** contain Windows *CRLF-style* line endings.`

superm1: Ah good.

> we want the descriptor to be as small as possible

So on this thought process, there is a few more possibilities that come to mind that can make it smaller:
1) Make the keys accept a shortened version as well.  IE instead of `Plugin` use `P`
2) Add a numerical assignment to all USB plugins and use that.  IE instead of `P=android_boot` use `P=2`
3) Use a compression algorithm for the buffer.  (With such small buffers this might not make a lot of sense.)
3) Use a numerical assignment for common patterns.  For example Logitech has identical `Flags`  for most devices.  So maybe there could be a number that represents that which gets encoded.

Any/all of the magic for that stuff I think could go into the `contrib` script.  So as a consumer of this you would write your quirk as normal, and the contrib script would do the best it could to shrink it.

docs/ds20.md
```diff
@@ -0,0 +1,132 @@
+# fwupd BOS DS20 Specification
```
superm1, line 1: Don't forget to add a reference in `libfwupdplugin` docs and `index.html` docs too to this

hughsie: Added, thanks. Good catch.

contrib/generate-ds20.py
hughsie: > Use a compression algorithm for the buffer

I actually tried zlib for this, but you're correct because the data is not repeated and because it's pretty tiny it actually increased the size by 34%!

I think we always want to support the unminimized quirk data, but I agree we can do something. I was thinking about something like fu_plugin_add_alias(plugin, "cap") where we could have a 3 letter acronym. Certainly flags could be specified by integer too, but that gets a bit messy as then the quirk "compiler" needs to know lots of grotty details about what the internal and plugin private flag values are. In most cases I think the control transfer is going to be of size 0x100 and that's a *lot* of data that can be included. I think we could also make flags "inherit" in the common case, so that setting one flag actually sets a bitmask of other flags -- e.g. `fu_device_alias_private_flag(FuDevice *self, guint64 value, const gchar *value_str)`

hughsie: I was talking to a vendor about this last week, and their consensus was it wasn't worth compressing and adding complexity in the docs. Basically their logic was they're already doing a 1024 byte transfer, and they don't care if it's `Plugin=tacom-usb` or `P=tac` as the other bytes will all be 0x0 and transferred anyway.

superm1: OK, sounds good.  If we decide to change it later, it can always be a new minimum version requirement to use the minified version.

hughsie: Ack, agreed.

docs/ds20.md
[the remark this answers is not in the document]
hughsie: > But aren't unknown quirks just ignored by older versions

Nah, there's a vocal g_warning() on the CLI -- which is probably the right thing to do to avoid typos.

hughsie:
Lets merge this and we can work on the next part, true emulation of devices. Thanks for all the help as always @superm1 -- it's really appreciated.

PR #5003, opened 2022-09-06
Ensure the context is set for all backends

hughsie:
This isn't required right now, but new functionality is much easier to add if
we know the context has always been set.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):

superm1 (commented):
Technically this is an abi bump tho right?

hughsie:
> Technically this is an abi bump tho right?

Nope, it's all in src/

PR #5006, opened 2022-09-06
Disable the snap build until the systemd deps are fixed

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
According to Zygmunt we have a few things to do: https://twitter.com/zygoon/status/1567198817979613185 -- we can revert this when it's working again, but right now every snap failure is spamming my inbox and causing the auto-merge actions to fail.

PR #5010, opened 2022-09-07
trivial: only build thunderbolt for x86 architectures

superm1:
It's true that the USB4 spec won't require x86, but a recent autopkgtest failure on powerpc doesn't make sense to debug.  You won't ever see TBT in a powerpc machine.

We can loosen this to any non-x86 architectures later that DO prove to have support for USB4 add-in cards.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):

bleungatchromium:
Is this a safe assumption?

Just off the top of my head, Apple M1 macs (which can run Asahi Linux) have a USB4 and Thunderbolt controller, and would benefit from this plugin.

Those are arm64.

superm1:
> Is this a safe assumption?
> 
> Just off the top of my head, Apple M1 macs (which can run Asahi Linux) have a USB4 and Thunderbolt controller, and would benefit from this plugin.
> 
> Those are arm64.

Actually I'm not convinced they would benefit from it.  The reason is that the kernel needs to know enough about the NVM format before it exports the ability to upgrade.  As it stands today release kernels only support Intel, and kernel 6.1 will support ASMedia.

bleungatchromium:
Got it, so the Asahi folks have some work to do in the kernel before this is relevant.

superm1:
> Got it, so the Asahi folks have some work to do in the kernel before this is relevant.

Not only that - Apple might not even support upgrading the NVM for the USB4 controller independent of other firmware for integrated devices.  Neither Intel or AMD support this either.

PR #5018, opened 2022-09-09
synaptics-mst: Check the firmware payload size when updating cayenne

hughsie:
PVS: Parameter 'payload_len' is always rewritten in function body before being used.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/synaptics-mst/fu-synaptics-mst-device.c
+++ b/plugins/synaptics-mst/fu-synaptics-mst-device.c
@@ -42,6 +42,8 @@
 
 #define FLASH_SETTLE_TIME 5000000 /* us */
 
+#define CAYENNE_FIRMWARE_SIZE 0x50000 /* bytes */
+
 /**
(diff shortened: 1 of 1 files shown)


plugins/synaptics-mst/fu-synaptics-mst-device.c
```diff
@@ -860,6 +860,15 @@ fu_synaptics_mst_device_update_cayenne_firmware(FuSynapticsMstDevice *self,
 	guint32 offset = 0;
 	guint32 write_loops = 0;
 
+	/* sanity check */
+	if (payload_len < 0x50000) {
```
superm1, line 864: 3 places using a hardcoded `0x50000` tells me this should be a `#define`

hughsie: Fair point, updated.

superm1 (approved):

PR #5019, opened 2022-09-09
Wip/hughsie/pvs

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-security-attrs.c
+++ b/libfwupdplugin/fu-security-attrs.c
@@ -204,7 +204,7 @@ fu_security_attrs_calculate_hsi(FuSecurityAttrs *self, FuSecurityAttrsFlags flag
 {
 	guint hsi_number = 0;
 	FwupdSecurityAttrFlags attr_flags = FWUPD_SECURITY_ATTR_FLAG_NONE;
-	GString *str = g_string_new("HSI:");
+	g_autoptr(GString) str = g_string_new("HSI:");
 	const FwupdSecurityAttrFlags hpi_suffixes[] = {
 	    FWUPD_SECURITY_ATTR_FLAG_RUNTIME_ISSUE,
 	    FWUPD_SECURITY_ATTR_FLAG_NONE,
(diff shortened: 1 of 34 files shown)


libfwupdplugin/fu-security-attrs.c
```diff
@@ -204,7 +204,7 @@ fu_security_attrs_calculate_hsi(FuSecurityAttrs *self, FuSecurityAttrsFlags flag
 {
 	guint hsi_number = 0;
 	FwupdSecurityAttrFlags attr_flags = FWUPD_SECURITY_ATTR_FLAG_NONE;
-	GString *str = g_string_new("HSI:");
+	g_autoptr(GString) str = g_string_new("HSI:");
```
superm1, line 207: is this right?  I thought that how `g_string_free` was used is proper.

plugins/logitech-hidpp/fu-logitech-hidpp-hidpp.c
```diff
@@ -14,10 +14,10 @@
 static gchar *
 fu_logitech_hidpp_msg_to_string(FuLogitechHidPpHidppMsg *msg)
 {
-	GString *str = g_string_new(NULL);
 	const gchar *tmp;
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GString) flags_str = g_string_new(NULL);
+	g_autoptr(GString) str = g_string_new(NULL);
```
superm1, line 20: this one also seems weird to me

superm1 (approved):
everything but the string handling ones seems fine.  If those are just done to appease the static analyzer I guess that's fine though too.

libfwupdplugin/fu-security-attrs.c
hughsie: Ya, we need to use the slightly weird `g_string_free(g_steal_pointer(&str), FALSE)` to avoid a double-free. The PVS linter was getting upset that we could return_val_if_fail and leak the string -- although we probably have bigger problems at that point...

superm1: I guess you could do g_string_free in `g_return_val_if_fail`, but yeah that seems weird too.

PR #5020, opened 2022-09-09
android-boot: fix segfault

DylanVanAssche:
Store strings during the lifetime of the plugin and free them when destroyed.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/android-boot/fu-android-boot-device.c
+++ b/plugins/android-boot/fu-android-boot-device.c
@@ -18,8 +18,8 @@
 
 struct _FuAndroidBootDevice {
 	FuUdevDevice parent_instance;
-	const gchar *label;
-	const gchar *uuid;
+	gchar *label;
+	gchar *uuid;
 	gchar *boot_slot;
 	guint64 max_size;
 };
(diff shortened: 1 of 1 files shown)


hughsie (dismissed):

DylanVanAssche:
@hughsie I added a `g_strdup` to all strings we store in the object and free them when the object is destroyed.
The biggest problem seems to be in the `serial` variable, It had an autofree but it doesn't own things.
Since we just want to pass it to the `fu_device_set_serial` method, we can just put it in the argument (that method does a `g_strdup` anyway internal). Benefit: we don't need to have it as a local var and does not need to be freed.

hughsie:
@DylanVanAssche needs less const:

../plugins/android-boot/fu-android-boot-device.c:336:20: error: passing argument 1 of ‘g_free’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]

DylanVanAssche:
@hughsie fixed!
`pmbootstrap build` seems to be less strict than the CI :)

hughsie (approved):

PR #5026, opened 2022-09-12
Allow adding backend tags to devices

hughsie:
This allows the backend to identify the specific device for a specific phase.

For instance, there might be a pre-update runtime, a bootloader and a post-update runtime and allowing tags to be saved to the backend object allows us to identify each version of the same physical device.

This takes us one step closer to emulating a complete byte-perfect end-to-end update without actual hardware installed.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -81,6 +81,7 @@ typedef struct {
 	gchar *custom_flags;
 	gulong notify_flags_handler_id;
 	GHashTable *instance_hash;
+	GPtrArray *backend_tags; /* of utf-8 */
 } FuDevicePrivate;
 
 typedef struct {
@@ -103,6 +104,7 @@ enum {
 	PROP_CONTEXT,
 	PROP_PROXY,
 	PROP_PARENT,
+	PROP_BACKEND_TAGS,
 	PROP_LAST
 };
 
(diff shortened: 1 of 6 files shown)


superm1:
> This takes us one step closer to emulating a complete byte-perfect end-to-end update without actual hardware installed.

TBH I think emulating a "good" update is not as valuable as emulating the behaviors of a "bad" update.  There are many more interesting things you can do there.

If you pull the USB device mid-update.  If you've got a policy forbiding downgrades and you make sure it fails right.  Try to install bad firmware to it and see how it behaves.  It's an exciting prospect to codifiy all of that!

superm1 (commented):
Is this actually needed?  You can't tell from the device state?

hughsie:
> You can't tell from the device state?

Nah, we're resetting to idle most of the time, and some plugins do things like WRITE in detach and things. Having multiple tags also makes it easy to identify the devices in the big JSON dump. All agreed about the broken updates tho

superm1:
OK..  You probably also want a probe/setup phase then.

hughsie:
>  You probably also want a probe/setup phase then

I don't think so; for two reasons:

 * Every phase gets called probe and setup, although some are skipped if it's already been run on the same FuDevice
 * I think the tag should be used to describe _what we're using the device for_ rather than _what was done to the device_

PR #5027, opened 2022-09-13
trivial: convert pci-psp plugin to create and use a device

superm1:
In general this is a cleanup, means less vfuncs that get called when devices are added. Beyond the optimization functionally should be a no-op for fwupd.  It will also prepare for potentially storing the version to the newly created secure processor device in the future.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


plugins/pci-psp/fu-pci-psp-device.c
```diff
@@ -0,0 +1,277 @@
...
+		return FALSE;
+	}
+	if (!fu_strtoull(buf, &val, 0, G_MAXUINT32, error))
+		return FALSE;
+	*out = val ? TRUE : FALSE;
+	return TRUE;
+}
+
+static void
+fu_pci_psp_add_security_attrs_tsme(FuDevice *device, const gchar *path, FuSecurityAttrs *attrs)
```
hughsie, line 43: these should all be `fu_pci_psp_device_` prefixes.

plugins/pci-psp/fu-pci-psp-device.c
```diff
@@ -0,0 +1,277 @@
...
+}
+
+static void
+fu_pci_psp_add_security_attrs(FuDevice *device, FuSecurityAttrs *attrs)
+{
+	const gchar *sysfs_path = NULL;
+	g_autofree gchar *test_file = NULL;
+
+	/* ccp not loaded */
+	if (device) {
```
hughsie, line 243: != NULL

plugins/pci-psp/fu-plugin-pci-psp.c
```diff
@@ -8,279 +8,18 @@
...
-
-static void
-fu_plugin_pci_psp_add_security_attrs(FuPlugin *plugin, FuSecurityAttrs *attrs)
-{
-	FuDevice *psp_device = fu_plugin_cache_lookup(plugin, "pci-psp-device");
-	const gchar *sysfs_path = NULL;
-	g_autofree gchar *test_file = NULL;
-
-	/* only AMD */
-	if (fu_cpu_get_vendor() != FU_CPU_VENDOR_AMD)
```
hughsie, line 258: we lost this, but I think that's okay

superm1: Yeah I didn't think it was needed anymore since CCP only probes when right PCI device present.

hughsie (approved):

PR #5028, opened 2022-09-13
cpu: Add the microcode version to the CPU device

superm1:
I swear this used to do it at one point, but I noticed it was missing recently.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/msr/fu-plugin-msr.c
+++ b/plugins/msr/fu-plugin-msr.c
@@ -262,6 +262,10 @@ fu_plugin_msr_backend_device_added(FuPlugin *plugin, FuDevice *device, GError **
 	/* get microcode version */
 	if (device_cpu != NULL) {
 		guint32 ver_raw;
+		guint8 offset;
+
+		if (!fu_cpuid(0x1, NULL, NULL, NULL, NULL, error))
+			return FALSE;
 		if (!fu_udev_device_pread(FU_UDEV_DEVICE(device),
 					  PCI_MSR_IA32_BIOS_SIGN_ID,
 					  buf,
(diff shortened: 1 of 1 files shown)


hughsie:
Shouldn't this be done using CPUID? I think mixing and matching sysfs-provided and cpuid-provided is probably a recipe for confusion.

superm1:
> Shouldn't this be done using CPUID? I think mixing and matching sysfs-provided and cpuid-provided is probably a recipe for confusion.

Yeah, I double checked how it's supposed to be working.  The kernel issues a CPUID of 0x1 then reads the MSR.  I think we should be doing the same as a precaution.

Also it seems the kernel doesn't fetch 4 bytes in, but right afterwards.  The fixup I pushed seems to work.

hughsie:
In main I get:

    Current version:    0x000000f0

with this branch I get nothing on my Lenovo Intel CPU :(

plugins/msr/fu-plugin-msr.c
```diff
@@ -273,7 +276,7 @@ fu_plugin_msr_backend_device_added(FuPlugin *plugin, FuDevice *device, GError **
 		fu_dump_raw(G_LOG_DOMAIN, "IA32_BIOS_SIGN_ID", buf, sizeof(buf));
 		if (!fu_memread_uint32_safe(buf,
 					    sizeof(buf),
-					    0x4,
+					    0x0,
```
hughsie, line 279: ya, i need 0x0 on INTEL...

hughsie: On AMD is IA32_BIOS_SIGN_ID even valid?

superm1: Hmm.  It's the same MSR for both AMD and Intel (0x8b).
* AMD code in the kernel: https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpu/amd.c#L610
* Intel code in the kernel: https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/microcode_intel.h#L56.

But yeah it looks like in the kernel it's off by 4 for Intel and AMD.  I'll push a fixup.

hughsie (approved):
Thanks!

PR #5031, opened 2022-09-13
Allow firmware files to apply quirks FIXME

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/firmware.quirk
+++ b/data/firmware.quirk
@@ -1,0 +1,5 @@
+# Gatkex EVB & HP Thunderbolt Dock G4
+[INTEL_THUNDERBOLT_NVM\DEV_0B26]
+IntelThunderboltNvmFamily=goshen-ridge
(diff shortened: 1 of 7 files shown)


hughsie:
@superm1 before I invest time in this, is the idea crazy or sane?

superm1:
IMO overkill.  Quirks are compiled into the resource files, so this type of stuff is going to save just a few lines of C code.

hughsie:
> IMO overkill

Ya, it feels *heavy*. It turns out that most of the GR usb4 docks all have the same device-id anyway....

superm1:
> > IMO overkill
> 
> Ya, it feels _heavy_. It turns out that most of the GR usb4 docks all have the same device-id anyway....

well they should.  The custom stuff is ID the OEM uses that goes in the NVM.

PR #5032, opened 2022-09-13
Add new plugin to display SMU firmware version on AMD APU/CPU

superm1:
This plugin is read-only, but can be used to tell what version of SMU firmware is installed.

This working is dependent upon:
https://patchwork.kernel.org/project/platform-driver-x86/patch/20220914141850.259-1-mario.limonciello@amd.com/

@jwrdegoede FYI so you can see the other half of this and how it's used.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


plugins/amd-pmc/fu-amd-pmc-device.c
```diff
@@ -0,0 +1,73 @@
...
+	version =
+	    fu_udev_device_get_sysfs_attr(FU_UDEV_DEVICE(device), "smu_fw_version", &error_local);
+	if (version == NULL) {
+		if (g_error_matches(error_local, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+			g_set_error_literal(error,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_NOT_SUPPORTED,
+					    "unsupported kernel version");
+		else
+			g_propagate_error(error, g_steal_pointer(&error_local));
```
hughsie, line 36: needs braces

hughsie: i also think it'd be clearer to just return FALSE early and unindent the g_propagate_error

hughsie: Maybe

    if (g_error_matches(error_local, G_IO_ERROR, G_IO_ERROR_NOT_FOUND)) {
        g_set_error_literal()
        return FALSE;
    }
    g_propagate_error()
    return FALSE;

seems more glibby.

superm1: Got it, thanks!

hughsie (approved):
LGTM, but maybe get the kernel ack from Hans before merging.

superm1:
> LGTM, but maybe get the kernel ack from Hans before merging.

Yeah.  Will leave it here until it's in his review tree.

superm1:
The kernel stuff is in hans review tree now. I'll merge this.

PR #5033, opened 2022-09-15
G14t compatibility

flying-elephant:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


plugins/wacom-raw/fu-wacom-aes-device.c
```diff
@@ -253,6 +269,14 @@ fu_wacom_aes_device_write_firmware(FuDevice *device,
 						(gsize)i,
 						(gsize)chunks->len);
 	}
+
+	/* finalize the device */
+	if (!fu_wacom_aes_device_finalize_firmware(device, error)) {
+		return FALSE;
+	}
```
hughsie, line 276: no {} needed here

plugins/wacom-raw/fu-wacom-aes-device.c
```diff
@@ -220,6 +220,22 @@ fu_wacom_aes_device_write_block(FuWacomAesDevice *self,
 	return TRUE;
 }
 
+static gboolean
+fu_wacom_aes_device_finalize_firmware(FuDevice *device, GError **error)
```
hughsie, line 224: shouldn't this be fu_wacom_aes_device_detach() i.e. implementing the FuDevice->detach() vfunc?

plugins/wacom-raw/fu-wacom-device.c
```diff
@@ -132,16 +132,20 @@ fu_wacom_device_attach(FuDevice *device, FuProgress *progress, GError **error)
...
-		g_debug("already in runtime mode, skipping");
-		return TRUE;
-	}
-	if (!fu_wacom_device_set_feature(self, (const guint8 *)&req, sizeof(req), error)) {
-		g_prefix_error(error, "failed to switch to runtime mode: ");
-		return FALSE;
+	gboolean emr_device = fu_device_has_guid(device, "WacomEMR");
+
+	/* only EMR requires way back to runtime mode */
+	if (emr_device) {
```
hughsie, line 138: this should move to fu-wacom-emr-device.c -- if the aes and emr devices have different attach behaviours this should be moved into the subclassed functions.

plugins/wacom-raw/wacom-raw.quirk
```diff
@@ -134,4 +134,4 @@ GType = FuWacomEmrDevice
 GType = FuWacomAesDevice
 WacomI2cFlashBlockSize = 128
 WacomI2cFlashBaseAddr = 0x8000
-WacomI2cFlashSize = 0x24000
+WacomI2cFlashSize = 0x28000
```
hughsie, line 137: is this *all* AES panels?

plugins/wacom-raw/fu-wacom-aes-device.c
```diff
@@ -253,6 +269,14 @@ fu_wacom_aes_device_write_firmware(FuDevice *device,
 						(gsize)i,
 						(gsize)chunks->len);
 	}
+
+	/* finalize the device */
+	if (!fu_wacom_aes_device_finalize_firmware(device, error)) {
+		return FALSE;
+	}
+
+	g_usleep(500 * 1000);
```
superm1, line 278: what's this sleeping after the firmware is done writing for?  If it's real and not debugging, it needs a comment explaining it.

plugins/wacom-raw/fu-wacom-aes-device.c
flying-elephant: I"ll work on this.

plugins/wacom-raw/fu-wacom-aes-device.c
flying-elephant: Sure, that's possible.

plugins/wacom-raw/fu-wacom-device.c
flying-elephant: Yes, now they should have different process to get back to their boot loader. I'll separate out the functions for each EMR or AES; I'll create wacom_aes_attach and wacom_emr_attach subclasses.

plugins/wacom-raw/fu-wacom-aes-device.c
flying-elephant: I'll leave the comment for it.
The sleep is required for our AES firmwrae to get back to runtime mode.

superm1: I think you might want to be using `fu_device_get_acquiesce_delay` then instead

plugins/wacom-raw/wacom-raw.quirk
flying-elephant: available address for the latest AES, G14T, is expanded to 0x4000, so the maximum size is also required a change.
If this plug-in writes the firmware up to this address, the other generations than G14T, sends back an  error in order not to overwrite the address on them.

Is this current wacom-raw plugin to write up to "WacomI2CFlashSize" ? If so, I guess I have to modify the write part as well.
Then, I'd like to know where I should make modification on because chunks->len seems to be set somewhere else.

plugins/wacom-raw/fu-wacom-aes-device.c
flying-elephant: Thank you for your advice. 
I looked up a sample to use that function, but don't know how to set desired time on it.
Would you like to explain how I wait for 500msec with fu_device_get_acquiesce_delay?
Only I found an example was in fu-engine.c.

hughsie: I think the hardcoded delay is fine -- if the hardware cannot actually respond for that time. I think the acquiesce delay is only really useful for composite devices - it's certainly harder to explain for the single-device update. :)

flying-elephant (commented):
I changed the files on the previous review.
Please take a look at them again.

hughsie:
Closing in favour of https://github.com/fwupd/fwupd/pull/5048 -- @flying-elephant you have commit permission on that branch too.

Issue #5035, opened 2022-09-16
Cannot update on Fedora 36 wirh Secure Boot dbx


thorrrr:
Cannot update today on Fedora 36 with Secure Boot dbx
❯ sudo dsync --refresh
[sudo] password for dale:
sudo: dsync: command not found
❯ sudo dsync --refresh
sudo: dsync: command not found
❯
❯ sudo dnf dsync --refresh
Copr repo for fedy owned by kwizart                                                           7.0 kB/s | 3.0 kB     00:00
Copr repo for PyCharm owned by phracek                                                         11 kB/s | 3.6 kB     00:00
Fedora 36 - x86_64                                                                            8.5 kB/s |  21 kB     00:02
Fedora 36 openh264 (From Cisco) - x86_64                                                      2.8 kB/s | 989  B     00:00
Fedora Modular 36 - x86_64                                                                     29 kB/s |  20 kB     00:00
Fedora 36 - x86_64 - Updates                                                                   23 kB/s |  12 kB     00:00
Fedora Modular 36 - x86_64 - Updates                                                           32 kB/s |  13 kB     00:00
google-chrome                                                                                 9.6 kB/s | 1.3 kB     00:00
microsoft-edge                                                                                 17 kB/s | 3.0 kB     00:00
created by dnf config-manager from https://packages.microsoft.com/yumrepos/edge                19 kB/s | 3.0 kB     00:00
RPM Fusion for Fedora 36 - Free                                                                34 kB/s | 7.7 kB     00:00
RPM Fusion for Fedora 36 - Free - Updates                                                      41 kB/s | 7.0 kB     00:00
RPM Fusion for Fedora 36 - Nonfree                                                             23 kB/s | 7.8 kB     00:00
RPM Fusion for Fedora 36 - Nonfree - NVIDIA Driver                                             44 kB/s | 7.3 kB     00:00
RPM Fusion for Fedora 36 - Nonfree - Steam                                                     20 kB/s | 7.0 kB     00:00
RPM Fusion for Fedora 36 - Nonfree - Updates                                                   40 kB/s | 7.1 kB     00:00
Dependencies resolved.
Nothing to do.
Complete!
❯
❯ sudo fwupdmgr update
Devices with no available firmware updates:
 • THNSN5512GPUK NVMe TOSHIBA 512GB
Devices with the latest available firmware version:
 • System Firmware
╔══════════════════════════════════════════════════════════════════════════════╗
[...]

PR #5036, opened 2022-09-16
Always check the BDP partitions when getting all the possible ESPs

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/5035

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-volume.c
+++ b/libfwupdplugin/fu-volume.c
@@ -767,3 +767,60 @@ fu_volume_new_esp_for_path(const gchar *esp_path, GError **error)
 		    esp_path);
 	return NULL;
 }
+
+/**
+ * fu_volume_new_by_esp:
+ * @error: (nullable): optional return location for an error
+ *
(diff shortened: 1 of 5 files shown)


hughsie:
@superm1 I'm not sure on the *name*....

superm1:
What exactly is BDP?  I don't remember what that GUID represents and our comment is equally helpful.
https://fwupd.github.io/libfwupdplugin/const.VOLUME_KIND_BDP.html

hughsie:
BDP: https://en.wikipedia.org/wiki/Microsoft_basic_data_partition

superm1:
> BDP: https://en.wikipedia.org/wiki/Microsoft_basic_data_partition

ah we should have that link in the documentation too!  Will look at this with that in mind.

superm1 (approved):

hughsie:
> we should have that link in the documentation too

Done.

PR #5038, opened 2022-09-19
android-boot: support eMMC boot partitions

DylanVanAssche:
eMMC boot partitions are commonly found in eMMCs to write platform firmware on like Tow-Boot.
This way, users can install their distro without having to deal with bootloaders embedded into the distro image.

- These partitions do not contain a partition table so we need to identify them differently, added a new GUID for this.
- Such partitions can only be written if `force_ro` is set to 0 in sysfs, do this when flashing firmware.
- Tow-Boot leverages uSWID for version information, add version extraction for this to the plugin based upon the MTD plugin.

Would be great to have this in before the next release so it lands before the freeze of the next Alpine stable.

```
pine64-pinephone:~$ fwupdmgr get-devices
pine64 Pine64 PinePhone (1.2)
│
├─QUECTEL Mobile Broadband Module:
│     Device ID:          976c4a39e87f61e6940ea6a8d39c583cfa99615f
│     Summary:            Quectel EG25-G modem
│     Current version:    0.6.8
│     Vendor:             Qualcomm (USB:0x2C7C)
│     Release Branch:     community
│     GUIDs:              db379a33-254f-5140-b37e-d36ae7e5c039
│                         d18f31f1-a3fa-55a2-b4ed-decfbc1e004d ← USB\VID_2C7C&PID_0125&REV_0318&NAME_EG25GGB
│                         1a2996cb-f86e-5583-a464-e1b96e1c6ae9 ← USB\VID_2C7C&PID_0125&REV_0318
│                         587bf468-6859-5522-93a7-6cce552a0aa3 ← USB\VID_2C7C&PID_0125
│                         22ae45db-f68e-5c55-9c02-4557dca238ec ← USB\VID_2C7C
│     Device Flags:       • Updatable
│                         • System requires external power source
│   
└─Tow-Boot platform firmware:
      Device ID:          10dd409c2799697c90ebca1a12c460ad0cbbf1eb
      Summary:            Tow-Boot platform firmware
      Current version:    2021.10-005
      Vendor:             Tow-Boot (org.tow-boot)
      GUID:               c96f594a-8b9e-5ee8-bae2-74c1b1355877 ← DRIVE\DEVNAME_dev-mmcblk2boot0&ID_PATH_platform-1c11000.mmc&ID_NAME_DA4032
      Device Flags:       • Internal device
                          • Updatable
                          • System requires external power source
                          • Needs a reboot after installation
```

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/plugins/android-boot/fu-android-boot-device.c
+++ b/plugins/android-boot/fu-android-boot-device.c
@@ -81,8 +81,12 @@ fu_android_boot_device_probe(FuDevice *device, GError **error)
 		}
 	}
 
-	/* set max firmware size, required to avoid writing firmware bigger than partition */
-	if (!g_udev_device_has_property(udev_device, "ID_PART_ENTRY_SIZE")) {
+	/*
+	 * Set max firmware size, required to avoid writing firmware bigger than partition.
+	 * Some partitions like eMMC boot partitions do not expose their size, a quirk must
+	 * be added to these partitions to define their maximum size.
+	 */
+	if (g_udev_device_has_property(udev_device, "ID_PART_ENTRY_SIZE") && self->max_size <= 0) {
 		g_set_error_literal(error,
 				    FWUPD_ERROR,
 				    FWUPD_ERROR_NOT_SUPPORTED,
@@ -95,13 +99,6 @@ fu_android_boot_device_probe(FuDevice *device, GError **error)
 	self->max_size = size;
 
 	/* extract partition UUID and require it for supporting a device */
-	if (!g_udev_device_has_property(udev_device, "ID_PART_ENTRY_UUID")) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_NOT_SUPPORTED,
-				    "device does not have a UUID");
-		return FALSE;
-	}
 	self->uuid = g_strdup(g_udev_device_get_property(udev_device, "ID_PART_ENTRY_UUID"));
 
 	/* extract serial number and set it */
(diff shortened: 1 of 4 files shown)


plugins/android-boot/fu-android-boot-device.c
```diff
@@ -116,10 +113,27 @@ fu_android_boot_device_probe(FuDevice *device, GError **error)
...
+	 * partition table. These devices are typically used to store platform or bootloader
+	 * firmware on it at offset 0. Combine their exposed device name, path and hardware name for
+	 * a GUID. The GUID depends on the device DT and therefore unique.
+	 */
+	fu_device_add_instance_strsafe(device,
+				       "DEVNAME",
+				       g_udev_device_get_property(udev_device, "DEVNAME"));
+	fu_device_add_instance_strsafe(device,
+				       "ID_PATH",
+				       g_udev_device_get_property(udev_device, "ID_PATH"));
```
hughsie, line 127: This isn't a great GUID as it's not part of the kernel API; the exact name/path isn't going to be predictable from the kernel. Is there anything else we can use?

DylanVanAssche: I'm afraid not... All the information to uniquely identify these partitions is not available because it has no partition table (no UUIDs, no labels, etc.). Setting these doesn't even work here as these are not 'regular' partitions. They always have `boot` in their devname if they exist on a device though.

hughsie: Could we read back the block data and read any header?

DylanVanAssche: Well we have uSWID data for Tow-Boot (already for the version information), we could extract maybe something else from that?

hughsie: That does feel a bit circular, but yes, that's probably a better idea than relying on the kernel naming scheme.

DylanVanAssche: - So extending the firmware version retrieval to find `Tow-Boot` there and use that to build GUIDs?
- I would use this as a fallback, so try to build GUIDs with the ones we have and if no GUIDs are generated then, fallback to this approach?
- For performance reasons I would do the version retrieval then also in the probe method to read the firmware only once.

hughsie: Did that fly? What's the status of this PR please?

DylanVanAssche: I haven't been able to test this out I'm afraid. Currently, I swamped with $DAYJOB. I hope to pick this up soon, but it might take several weeks.

hughsie: Got it, thanks. No panic.

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -154,6 +230,26 @@ fu_android_boot_device_open(FuDevice *device, GError **error)
 		return FALSE;
 	}
 
+	/* Allow writing to eMMC boot partitions */
```
hughsie, line 233: lowercase allow please

plugins/android-boot/fu-android-boot-device.c
```diff
@@ -154,6 +230,26 @@ fu_android_boot_device_open(FuDevice *device, GError **error)
...
+	return TRUE;
+}
+
+static gboolean
+fu_android_boot_device_close(FuDevice *device, GError **error)
+{
+	const gchar *force_ro = "force_ro";
+
+	/* Disallow writing to eMMC boot partitions */
+	if (fu_udev_device_get_sysfs_attr(FU_UDEV_DEVICE(device), force_ro, NULL) != NULL &&
```
hughsie, line 248: lowercase disallow please

plugins/android-boot/fu-android-boot-device.c
DylanVanAssche: Yes, let's close this one for now. I'm still interested in this, but upstream stuff stalled (need Tow-Boot etc. on board). We can re-open later.

Issue #5042, opened 2022-09-20
unable to install update with ESP on raid1 


ausil:
```
[root@kasuf ~]# fwupdmgr update
Devices with no available firmware updates: 
 • BCM20702A0
 • SSD 850 EVO M.2 500GB
 • SSD 970 EVO Plus 1TB
 • SSD 970 EVO Plus 1TB
 • System Firmware
╔══════════════════════════════════════════════════════════════════════════════╗
║ Upgrade UEFI dbx from 77 to 217?                                             ║
╠══════════════════════════════════════════════════════════════════════════════╣
║ This updates the dbx to the latest release from Microsoft which adds         ║
║ insecure versions of grub and shim to the list of forbidden signatures due   ║
║ to multiple discovered security updates.                                     ║
║                                                                              ║
║ UEFI dbx and all connected devices may not be usable while updating.         ║
╚══════════════════════════════════════════════════════════════════════════════╝

Perform operation? [Y|n]: 
Downloading…             [***************************************]
Downloading…             [***************************************]
Downloading…             [***************************************]
Decompressing…           [***************************************]
Authenticating…          [***************************************]
Waiting…                 [***************************************]
Writing…                 [***************************************]
Decompressing…           [                                       ]Blocked executable in the ESP, ensure grub and shim are up to date: no volumes of type c12a7328-f81f-11d2-ba4b-00a0c93ec93b
[root@kasuf ~]# lsblk
NAME                                            MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS
sda                                               8:0    0 465.8G  0 disk  
└─sda1                                            8:1    0 465.8G  0 part  /run/media/dennis/252690c6-5dde-471a-abe2-2b59dd3ca718
zram0                                           252:0    0     8G  0 disk  [SWAP]
nvme1n1                                         259:0    0 931.5G  0 disk  
├─nvme1n1p1                                     259:5    0     1G  0 part  
│ └─md127                                         9:127  0  1023M  0 raid1 /boot
├─nvme1n1p2                                     259:6    0   601M  0 part  
│ └─md125                                         9:125  0 600.9M  0 raid1 /boot/efi
└─nvme1n1p3                                     259:7    0 929.7G  0 part  
  └─md126                                         9:126  0 929.6G  0 raid1 
    └─luks-a4d3f8ab-8545-40d9-be42-c960666ec138 253:0    0 929.6G  0 crypt 
[...]

PR #5050, opened 2022-09-20
Move the getting the ESP to the context

hughsie:
We now have two plugins getting the ESP values, and we only allow hardcoding the ESP in uefi_capsule.conf. We also have *another* override in the `FWUPD_UEFI_ESP_PATH` environment variable.

Make all this a lot simpler by moving the ESP+BDP code to `FuContext`, which also means we can handle the override (via the config file) in the engine, and the override (in the command line tools) using the same mechanism.

Also, automate the migration of the `OverrideESPMountPoint` -> `EspLocation` when loading the engine.

Fixes https://github.com/fwupd/fwupd/issues/5042

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-volume.c
+++ b/libfwupdplugin/fu-volume.c
@@ -647,7 +647,9 @@ fu_volume_new_by_devnum(guint32 devnum, GError **error)
  * fu_volume_new_esp_default:
  * @error: (nullable): optional return location for an error
  *
- * Gets the platform default ESP
+ * Gets the platform default ESP.
+ *
+ * NOTE: This has been deprecated, please use fu_context_get_esp_volumes() instead.
  *
  * Returns: (transfer full): a volume, or %NULL if the ESP was not found
  *
@@ -755,8 +757,11 @@ fu_volume_new_esp_for_path(const gchar *esp_path, GError **error)
 	basename = g_path_get_basename(esp_path);
 	for (guint i = 0; i < volumes->len; i++) {
 		FuVolume *vol = g_ptr_array_index(volumes, i);
-		g_autofree gchar *vol_basename =
-		    g_path_get_basename(fu_volume_get_mount_point(vol));
+		const gchar *mount_point = fu_volume_get_mount_point(vol);
+		g_autofree gchar *vol_basename = NULL;
+		if (mount_point == NULL)
+			continue;
+		vol_basename = g_path_get_basename(mount_point);
 		if (g_strcmp0(basename, vol_basename) == 0)
 			return g_object_ref(vol);
 	}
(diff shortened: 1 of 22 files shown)


libfwupdplugin/fu-volume.c
```diff
@@ -445,7 +445,16 @@ fu_volume_locker(FuVolume *self, GError **error)
 					 error);
 }
 
-/* private */
+/**
+ * fu_volume_new_from_mount_path:
+ * @mount_path: a mount path, typically a string
+ *
+ * Creates a volume with a specified mount path. Usually used internally or from the self tests.
```
superm1, line 452: even though this is a "public" symbol, I want to discourage it's use.  How about if we kick it over to a header only to be used by self tests?  Or instead let the self tests set an environment variable and the daemon picks up that environment variable to use this symbol?

libfwupdplugin/fu-volume.c
```diff
@@ -647,7 +656,9 @@ fu_volume_new_by_devnum(guint32 devnum, GError **error)
  * fu_volume_new_esp_default:
  * @error: (nullable): optional return location for an error
  *
- * Gets the platform default ESP
+ * Gets the platform default ESP.
+ *
+ * NOTE: This has been deprecated, please use fu_context_get_esp_volumes() instead.
```
superm1, line 661: Didn't we need to add this to a list somewhere to say we need to drop it on the next ABI change?

src/fu-engine.c
```diff
@@ -7698,6 +7720,17 @@ fu_engine_load(FuEngine *self, FuEngineLoadFlags flags, FuProgress *progress, GE
 	}
 	fu_progress_step_done(progress);
 
+	/* migrate per-plugin settings into daemon.conf */
+	plugin_uefi = fu_plugin_list_find_by_name(self->plugin_list, "uefi_capsule", NULL);
+	if (plugin_uefi != NULL) {
+		const gchar *tmp = fu_plugin_get_config_value(plugin_uefi, "OverrideESPMountPoint");
+		if (tmp != NULL && g_strcmp0(tmp, fu_config_get_esp_location(self->config)) != 0) {
+			g_debug("migrating OverrideESPMountPoint=%s to EspLocation", tmp);
```
superm1, line 7728: I think we should probably change the original config too, right?

superm1: Also, I think this message should _not_ be debug.  It should be an information level, we want people to know it happened in their system journal.

libfwupdplugin/fu-volume.c
hughsie: Indeed! Added to `readme.md` alongside the others.

plugins/uefi-capsule/uefi_capsule.conf
```diff
@@ -9,6 +9,8 @@
 
 # the EFI system partition (ESP) path used if UDisks is not available
 # or if this partition is not mounted at /boot/efi, /boot, or /efi
+#
+# NOTE: this is now unused; use the ESPMountPoint value in daemon.conf
 #OverrideESPMountPoint=
```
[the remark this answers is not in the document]
hughsie: Good point; removed.

src/fu-engine.c
hughsie: Agreed, fixed.

libfwupdplugin/fu-volume.c
hughsie: I agree the export is ugly. Will have a rethink.

hughsie: I've pushed a fixup that adds back the env-variable. Maybe in the two self test cases it's actually fine.

ausil:
I applied this and did a build locally, I can confirm that #5042 is fixed by this

libfwupdplugin/fu-volume.c
hughsie: @superm1 do you like the export more or the env-var more?

superm1: env variable, definitely.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
```diff
@@ -500,6 +500,16 @@ fu_plugin_uefi_capsule_load_config(FuPlugin *plugin, FuDevice *device)
...
 }
 
+static FuVolume *
+fu_plugin_uefi_capsule_get_default_esp(FuPlugin *plugin, GError **error)
+{
+	g_autoptr(GPtrArray) esp_volumes = NULL;
+	esp_volumes = fu_context_get_esp_volumes(fu_plugin_get_context(plugin), error);
+	if (esp_volumes == NULL)
+		return NULL;
+	return g_object_ref(g_ptr_array_index(esp_volumes, 0));
```
superm1, line 510: This is now a heuristic buried in the plugin source (if more than one ESP pick the first one).
Consequently I think it's a good idea to do something.
Ideas:
1) emit a message to the journal when more than one ESP was detected here.
2) An error saying we don't know "which" to pick, and don't use one.
3) If more than one, but one is ESP type and the other is BDP, pick the ESP type one.

plugins/uefi-capsule/fu-uefi-tool.c
```diff
@@ -234,19 +236,23 @@ main(int argc, char *argv[])
...
+		fu_context_add_esp_volume(ctx, volume);
+	}
+
+	/* get the default ESP only */
+	esp_volumes = fu_context_get_esp_volumes(ctx, &error);
+	if (esp_volumes == NULL) {
+		g_printerr("%s\n", error->message);
+		return EXIT_FAILURE;
 	}
+	esp = g_object_ref(g_ptr_array_index(esp_volumes, 0));
```
superm1, line 255: same heuristic comment applies here.  Which means maybe we want the heuristic/message/whatever decision in a common symbol used by both places.

superm1: At least for this tool; the other option is to prompt for ESP if multiple are found.

plugins/uefi-capsule/fu-plugin-uefi-capsule.c
hughsie: > If more than one, but one is ESP type and the other is BDP, pick the ESP type one.

That's the current ordering of the fu_context_get_esp_volumes() i.e. ESP first, BDP next. I think a g_warning() is appropriate here. Fixed.

plugins/uefi-capsule/fu-uefi-tool.c
hughsie: I think in general we want to deal with multiple ESPs properly (i.e. your prompt idea) -- I don't think any heuristic is a good idea when we have an interactive user. Fixed.

superm1 (approved):

PR #5058, opened 2022-09-21
redfish: Only create users using IPMI when we know it's going to work

hughsie:
Make the IPMI auto-account feature allow-listed on specific vendors as some IPMI implementations are not specification compliant and do entirely the wrong thing.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
The back story here is that some old versions of Supermicro firmware actually change the ADMIN password when you ask them to change the OPERATOR password. :exploding_head:

superm1 (approved):

PR #5060, opened 2022-09-22
usi-dock: Add initialization retry to ensure flash ready

victor-cheng:
sometimes initialization will report flash busy.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/usi-dock/fu-usi-dock-mcu-device.c
+++ b/plugins/usi-dock/fu-usi-dock-mcu-device.c
@@ -524,6 +524,37 @@ fu_usi_dock_mcu_device_wait_for_spi_ready_cb(FuDevice *device, gpointer user_dat
 }
 
 static gboolean
+fu_usi_dock_mcu_device_wait_for_spi_initial_ready_cb(FuDevice *device,
+						     gpointer user_data,
+						     GError **error)
+{
+	FuUsiDockMcuDevice *self = FU_USI_DOCK_MCU_DEVICE(device);
+	guint8 buf[] = {USBUID_ISP_DEVICE_CMD_FWBUFER_INITIAL};
+	guint8 val = 0;
+
+	if (!fu_usi_dock_mcu_device_txrx(self,
+					 TAG_TAG2_CMD_SPI,
+					 buf,
+					 sizeof(buf),
+					 &val,
+					 sizeof(val),
+					 error))
+		return FALSE;
+	if (val != SPI_STATE_READY) {
+		g_set_error(error,
+			    G_IO_ERROR,
+			    G_IO_ERROR_BUSY,
+			    "SPI state is %s [0x%02x]",
+			    fu_usi_dock_spi_state_to_string(val),
+			    val);
+		return FALSE;
+	}
+
+	/* success */
+	return TRUE;
+}
+
+static gboolean
 fu_usi_dock_mcu_device_wait_for_checksum_cb(FuDevice *device, gpointer user_data, GError **error)
 {
 	FuUsiDockMcuDevice *self = FU_USI_DOCK_MCU_DEVICE(device);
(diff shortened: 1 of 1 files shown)


plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -572,14 +603,15 @@ fu_usi_dock_mcu_device_write_firmware_with_idx(FuUsiDockMcuDevice *self,
 					 sizeof(val),
 					 error))
 		return FALSE;
```
hughsie, line 605: I think you can get rid of this initial one as well, no?

hughsie: @victor-cheng I'm prepping the new fwupd release for tomorrow, so if you want this in the tarball then can you fix this up today please. If timing is not important then no panic.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -568,18 +599,19 @@ fu_usi_dock_mcu_device_write_firmware_with_idx(FuUsiDockMcuDevice *self,
 					 TAG_TAG2_CMD_SPI,
 					 &cmd,
 					 sizeof(cmd),
-					 &val,
-					 sizeof(val),
+					 NULL,
+					 0x0,
 					 error))
```
hughsie, line 604: nah, i mean this whole thing -- it's always called at least once with retry.

victor-cheng: Do you mean to omit this command? I had to make sure that the order of these commands would behave the same as on other platforms, and we would only write to the external flash once. We need this retry mechanism.

hughsie: i mean you could omit the:

	/* initial external flash */
	cmd = USBUID_ISP_DEVICE_CMD_FWBUFER_INITIAL;
	if (!fu_usi_dock_mcu_device_txrx(self,
					 TAG_TAG2_CMD_SPI,
					 &cmd,
					 sizeof(cmd),
					 &val,
					 sizeof(val),
					 error))
		return FALSE;
	if (val != SPI_STATE_READY) {
		g_set_error(error,
			    G_IO_ERROR,
			    G_IO_ERROR_INVALID_DATA,
			    "invalid state for CMD_FWBUFER_INITIAL, got 0x%02x",
			    val);
		return FALSE;
	}

and just rely on it being called in the fu_usi_dock_mcu_device_wait_for_spi_initial_ready_cb() loop.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -563,23 +594,15 @@ fu_usi_dock_mcu_device_write_firmware_with_idx(FuUsiDockMcuDevice *self,
...
-			    val);
+	if (!fu_device_retry(FU_DEVICE(self),
+			     fu_usi_dock_mcu_device_wait_for_spi_initial_ready_cb,
+			     30,
+			     NULL,
+			     error)) {
+		g_prefix_error(error, "failed to wait for initial: ");
 		return FALSE;
 	}
+
```
hughsie, line 605: no extra newline here, and then LGTM.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -563,23 +594,15 @@ fu_usi_dock_mcu_device_write_firmware_with_idx(FuUsiDockMcuDevice *self,
 	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 42, "internal-flash");
```
hughsie, line 594: also, there's still FU_PROGRESS_FLAG_GUESSED here -- doesn't the daemon spit out some big red text on what the values are supposed to be? See https://github.com/fwupd/fwupd/wiki/Daemon-Warning:-FuProgress-steps

hughsie (approved):

hughsie (dismissed):

hughsie:
@victor-cheng 

    ../plugins/usi-dock/fu-usi-dock-mcu-device.c:582:16: error: unused variable ‘val’ [-Werror=unused-variable]
      582 |         guint8 val = 0x0;

hughsie (approved):

PR #5062, opened 2022-09-22
snap: Use strict confinement

valentindavid:
In order to run fwupd in Ubuntu Core, the snap has to use strict confinement. There is a snap called `uefi-fw-tools` which contained fwupd which could work with strict confinement. But it would be better if the upstream snap was strictly confined in the first place.

This PR is experimental and will need a lot of testing. I would like some suggestion of hardware to test.

Here are PRs in snapd to improve the required interfaces
 - https://github.com/snapcore/snapd/pull/12169
 - https://github.com/snapcore/snapd/pull/12149
 - https://github.com/snapcore/snapd/pull/12144

In order to run on Ubuntu Core, upower and udisks2 snaps are also needed. polkit is not yet available. I have some `snapcraft.yaml` files for those snaps at:
- https://github.com/valentindavid/snap-udisks2
- https://github.com/valentindavid/snap-upower

On other systems, it should work without those snaps.

Type of pull request:

- [ ] Code fix
- [*] Feature
- [ ] Documentation


src/fu-util.c
```diff
@@ -3650,7 +3650,7 @@ fu_util_check_daemon_version(FuUtilPrivate *priv, GError **error)
 static gboolean
 fu_util_check_polkit_actions(GError **error)
 {
-#ifdef HAVE_POLKIT
+#ifdef VERIFY_POLKIT_ACTIONS
```
hughsie, line 3653: I think we need to have PolicyKit working -- this allows any user to do anything.

superm1:
Some of these changes make sense " generally " even until the experimental strict confinement is ready.

Could you perhaps split off the ones that make sense generically into their own PR and propose merging sooner? We can then keep this one as a draft until everything else is landed and we're confident on it.

snap/snapcraft.yaml
```diff
@@ -44,11 +70,25 @@ parts:
     plugin: nil
     stage-packages:
       - python3-gi
-      - python3-cairo
+      - python3-gi-cairo
+    override-stage: |
+      craftctl default
```
superm1, line 75: I think with this change we might be able to turn on UEFI capsule splash in the build arguments, right?

snap/snapcraft.yaml
```diff
@@ -71,13 +113,6 @@ parts:
                        "-Dlibjcat:introspection=false",
                        "-Dlibjcat:tests=false"]
     source: .
-    source-type: git
-    override-build: |
-      snapcraftctl build
-      echo $(git describe HEAD --always) > $SNAPCRAFT_STAGE/version
```
superm1, line 77: As the snap builds on every single release and goes to edge channel we had wanted this.

When you get to tagged release then it includes the tags. When you don't you at least know what commit you build from.

src/fu-util.c
valentindavid: It is intended to work. But the check of the policy file does not work when strictly confined.  Snapd is responsible of installing the policy with the fwupd interface. The path would be `/var/lib/snapd/hostfs/usr/share/polkit-1/actions/snap.fwupd.interface.polkit.org.freedesktop.fwupd.policy` and we would need to give permission to read that file.

snap/snapcraft.yaml
```diff
@@ -6,30 +6,56 @@ description: |
...
+    plugs: [bluez, udisks2, modem-manager, upower-observe, network]
+    slots: [fwupd]
     completer:
       share/bash-completion/completions/fwupdtool
   fwupd:
     command: fwupd.wrapper
-    daemon: simple
+    daemon: dbus
+    slots: [fwupd]
+    plugs: [udisks2, modem-manager, upower-observe, polkit, network]
```
superm1, line 47: don't you need bluez here too?

snap/snapcraft.yaml
superm1: never mind, I see you moved it.

snap/snapcraft.yaml
hughsie: Yes, and that's an internal network not the internet. Not sure if snap can confine to internal adapters or not.

src/fu-util.c
superm1: By default can the confined `fwupdmgr` not access that file then?  

I follow the intent of the new config option, but any other way we can dynamically verify that we don't need to check for it?  Perhaps can we check that we're in a confined snap environnment in some way?  environment variable?

snap/snapcraft.yaml
valentindavid: I feel this part should not be merged. It was to build the local files rather than from git. I can make it a separate commit not to be merged.

snap/snapcraft.yaml
valentindavid: Yes. I think we need it too. I will test that again and verify there are errors.

snap/snapcraft.yaml
superm1: Got it; don't worry.  This is a draft, we won't merge anything until it's not a draft.  But please pull in anything you want merged sooner into separate PR's so this draft PR stays lean.

superm1:
> https://github.com/snapcore/snapd/pull/12169

I made a pass at all the plugins to identify a few more missing things for this one, thanks!

valentindavid:
> > In order to run on Ubuntu Core, upower and udisks2 snaps are also needed. polkit is not yet available. I have some snapcraft.yaml files for those snaps at:
> 
> How do you mark these as "dependencies" snaps? Is that a thing?

Do you mean dependency to install automatically? Or to connect installed fwupd to installed usdisks2 and upower?

I do not think there is a way to automatically install other snaps.

For the connections it is done with `snap connect`. Once it will be published in store, I believe we can have it automatically connected.

superm1:
> Do you mean dependency to install automatically? Or to connect installed fwupd to installed usdisks2 and upower?
> I do not think there is a way to automatically install other snaps.

Yeah if they can't install automatically, we probably need to also think about how we can show some messaging to users that they need to do something themselves to install the other snaps to enable features.  We have a way to show plugin level messages to users, and I think this makes good sense to bring over in that case.

valentindavid:
> Yeah if they can't install automatically, we probably need to also think about how we can show some messaging to users that they need to do something themselves to install the other snaps to enable features. We have a way to show plugin level messages to users, and I think this makes good sense to bring over in that case.

Note that these snaps are only needed on Ubuntu Core which targets mostly IoT. For desktop distributions, udisks, upower and polkit installed onfrom host used from host.

snap/snapcraft.yaml
valentindavid: The snapd interfaces has no netfilter backend. So I do not think we could easily make an interface for specific networks.

However, we could add more restricted network access in the `fwupd` slot. For example, allow just tcp, and not allow resolve.

superm1: @hughsie do the internal networks ever use DNS that you've seen though?  It's probably better not to back ourselves into a corner if they expect to use mDNS or something like `redfish.local`

hughsie: I've only ever seen ipv4 addresses, but it's not inconceivable to see a hostname there.

superm1: I'd say we should avoid locking this down any further for now.

src/fu-util.c
valentindavid: Yes, we can use an environment variable. We can define it in the wrapper.

snap/snapcraft.yaml
```diff
@@ -121,6 +156,11 @@ parts:
...
       - uuid-dev
     stage-packages:
+      - libunbound8
+      - libevent-2.1-7
+      - libnghttp2-14
+      - libpsl5
+      - librtmp1
       - libgcab-1.0-0
       - libarchive13
       - libcbor0.8
```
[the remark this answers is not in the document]
valentindavid: I have disabled it.

valentindavid:
As a transition, we could have two `snapcraft.yaml`, one confined and one not. And they can be published in different channels. For example `confined/edge`.

superm1:
> As a transition, we could have two `snapcraft.yaml`, one confined and one not. And they can be published in different channels. For example `confined/edge`.

Sure; I'm fine with that.  But how do you get Launchpad to trigger "two" different builds then?

Also; generally I think several changes in this file we can start doing for the unconfined build too if you want to do some smaller byte size PRs for them.

src/fu-util.c
superm1: OK, please make a new environment variable and document it in in `docs/env.md`.  I don't think it needs to be snap specific.  Maybe `FWUPD_IGNORE_POLKIT`?

superm1:
Recently I overhauled our CI for snap and now use Github actions.  As part of the change I've also merged some of your changes that were applicable to classic snap too into main.  Feel free to rebase your branch on that.

contrib/snap/fwupdmgr.wrapper
```diff
@@ -1,2 +1,4 @@
 #!/bin/sh
+LD_LIBRARY_PATH="$SNAP/usr/local/lib${LD_LIBRARY_PATH:+:}${LD_LIBRARY_PATH-}"
```
superm1, line 2: What is being installed into `/snap/usr/local/lib/` now?  I don't believe fwupd does, and I would think that all the dependent libraries for this snap are in `/snap/usr/lib`.

snap/snapcraft.yaml
```diff
@@ -171,12 +210,44 @@ parts:
...
+      - libmbim-glib4
+      # pkttyagent
+      - polkitd
     prime:
       # we explicitly don't want /usr/bin/gpgconf
       # this will cause gpgme to error finding it
       # but that also avoids trying to use non-existent
       # /usr/bin/gpg2
-      - -usr/bin
+      - -usr/bin/dirmngr
```
superm1, line 226: why not take out of all `/usr/bin` like before?  I figured by putting the snap into a different prefix (`/`) this is a sane way to go.

contrib/snap/fwupdmgr.wrapper
valentindavid: Might not be needed. I do not remember why I needed that.

snap/snapcraft.yaml
```diff
@@ -239,6 +322,11 @@ parts:
       - -usr/share/lintian
       - -usr/share/pkgconfig
       - -usr/share/GConf
+      - -usr/lib/dconf
+      - -usr/lib/x86_64-linux-gnu/gio/modules/libdconfsettings.so
```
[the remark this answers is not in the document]
valentindavid: I agree.

valentindavid:
I have noticed that fwupd tries to read `/etc/fwupd/redfish.conf`. Is this user configurable? If so we will have to generate it with a configure hook for `snap set`.

superm1:
> I have noticed that fwupd tries to read `/etc/fwupd/redfish.conf`. Is this user configurable? If so we will have to generate it with a configure hook for `snap set`.

Yeah; all of the stuff traditionally in `/etc/fwupd/` is user configurable.  That particular one though is supposed to be locked down to only readable by root because it may have credentials that could interact with a BMC.

snap/snapcraft.yaml
valentindavid: Done.

src/fu-util.c
valentindavid: I am afraid `FWUPD_IGNORE_POLKIT` is not really the right thing. We do not ignore polkit, we just want to not check if the policies were installed. Would `FWUPD_NOCHECK_POLKIT` be ok?

valentindavid: I have pushed changes using `FWUPD_POLKIT_NOCHECK`

superm1 (commented):
These content of these changes look good to me.  Please split it across 3 commits though and into multiple PR's.  The `FWUPD_POLKIT_NOCHECK` and `FWUPD_LOCKDIR` can go in now, the snapcraft yaml should wait until the snapd changes are done.

hughsie:
> FWUPD_POLKIT_NOCHECK

That worries me slightly... What's the reasoning?

superm1:
> This worries me slightly... What's the reasoning?

It was [mentioned above](https://github.com/fwupd/fwupd/pull/5062#discussion_r977603673):

>It is intended to work. But the check of the policy file does not work when strictly confined. Snapd is responsible of installing the policy with the fwupd interface. The path would be /var/lib/snapd/hostfs/usr/share/polkit-1/actions/snap.fwupd.interface.polkit.org.freedesktop.fwupd.policy and we would need to give permission to read that file.

src/fu-util.c
```diff
@@ -3651,6 +3651,9 @@ static gboolean
 fu_util_check_polkit_actions(GError **error)
 {
 #ifdef HAVE_POLKIT
+	if (g_getenv("FWUPD_POLKIT_NOCHECK") != NULL) {
+		return TRUE;
+	}
```
superm1, line 3656: this causes fedora CI failure.  It needs to go after the variable declarations.

```
../src/fu-util.c: In function ‘fu_util_check_polkit_actions’:
/usr/include/glib-2.0/glib/gmacros.h:1204:43: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 1204 | #define _GLIB_CLEANUP(func)               __attribute__((cleanup(func)))
      |                                           ^~~~~~~~~~~~~
/usr/include/glib-2.0/glib/gmacros.h:1242:20: note: in expansion of macro ‘_GLIB_CLEANUP’
 1242 | #define g_autofree _GLIB_CLEANUP(g_autoptr_cleanup_generic_gfree)
      |                    ^~~~~~~~~~~~~
../src/fu-util.c:3657:9: note: in expansion of macro ‘g_autofree’
 3657 |         g_autofree gchar *directory = fu_path_from_kind(FU_PATH_KIND_POLKIT_ACTIONS);
      |         ^~~~~~~~~~
```

snap/snapcraft.yaml
```diff
@@ -6,45 +6,92 @@ description: |
...
       share/bash-completion/completions/fwupdtool
   fwupd:
     command: fwupd.wrapper
-    daemon: simple
+    daemon: dbus
+    slots: [fwupd]
+    plugs: [bluez, udisks2, modem-manager, upower-observe, polkit, network]
+    daemon-scope: system
+    activates-on:
+      - fwupd-dbus
```
superm1, line 50: CI seems to not like this:

```
Bad snapcraft.yaml content:
- extra field 'activates-on' not permitted in 'apps.fwupd' configuration
- extra field 'daemon-scope' not permitted in 'apps.fwupd' configuration
Full execution log: '/home/runner/.cache/snapcraft/log/snapcraft-20221026-1[24](https://github.com/fwupd/fwupd/actions/runs/3328017580/jobs/5505931179#step:3:26)838.008704.log'
```

src/fu-util.c
hughsie: Could we rename FU_PATH_KIND_POLKIT_ACTIONS to FU_PATH_KIND_POLKIT_POLICY and return the file path? Otherwise there's not much point in FU_PATH_KIND_POLKIT_ACTIONS existing...

valentindavid:
Doing some testing, I have noticed that the paths in `remotes.d/vendor-directory.conf` for example is not pointing to something useful. We should make it point to something within `/var/snap/fwupd/common`. But I will open a PR in the future when this one is done.

valentindavid:
@superm1 The required changes in snapd have been merged so now it can be easily tested. Also the snapcraft has the needed fixes. Remember to use "latest/stable" of snapcraft.

First, update snapd to edge (soon, the fixes will be released but for now it is in edge):
```
snap refresh --channel=latest/edge snapd
```

After installing the built snap, you need to connect the interfaces:
```
sudo snap connect fwupd:bluez
sudo snap connect fwupd:fwupdmgr fwupd:fwupd
sudo snap connect fwupd:hardware-observe
sudo snap connect fwupd:modem-manager
sudo snap connect fwupd:polkit
sudo snap connect fwupd:udisks2
sudo snap connect fwupd:upower-observe
```

Hopefully this will be done by the store in the future.

superm1:
>The required changes in snapd have been merged so now it can be easily tested

That's great, thanks!

>Also the snapcraft has the needed fixes. Remember to use "latest/stable" of snapcraft.

Is that what CI does today?  Or do we need to fixup CI too?  CI is what is responsible for building and publishing to the store.

> First, update snapd to edge (soon, the fixes will be released but for now it is in edge):

Do you have a "rough" ETA on this?  If it's not within the next month (approximately when we would do the next fwupd release), I think we should only merge the pieces of this that don't introduce the strict confinement.

> After installing the built snap, you need to connect the interfaces:
>Hopefully this will be done by the store in the future.

Do you know how we go about that auto-connecting?  I guess a request with the store people?

superm1 (changes requested):
The actual code changes look fine to me, but can you please split out the changes into individual commits so that if we need to revert something for a problem we can revert a smaller chunk.

valentindavid:
> The actual code changes look fine to me, but can you please split out the changes into individual commits so that if we need to revert something for a problem we can revert a smaller chunk.

I have split it.

valentindavid:
> > Also the snapcraft has the needed fixes. Remember to use "latest/stable" of snapcraft.
> 
> Is that what CI does today? Or do we need to fixup CI too? CI is what is responsible for building and publishing to the store.

It looks like it.

> > First, update snapd to edge (soon, the fixes will be released but for now it is in edge):
> 
> Do you have a "rough" ETA on this? If it's not within the next month (approximately when we would do the next fwupd release), I think we should only merge the pieces of this that don't introduce the strict confinement.

> > After installing the built snap, you need to connect the interfaces:
> > Hopefully this will be done by the store in the future.
> 
> Do you know how we go about that auto-connecting? I guess a request with the store people?

We will need to open a thread on the snapcraft.io forum to request it. I think upload will not succeed until it has been cleared.

superm1 (dismissed):
LGTM.  @hughsie can you please review?

src/fu-util.c
```diff
@@ -3695,9 +3695,15 @@ static gboolean
...
 #ifdef HAVE_POLKIT
-	g_autofree gchar *directory = fu_path_from_kind(FU_PATH_KIND_POLKIT_ACTIONS);
-	g_autofree gchar *filename =
-	    g_build_filename(directory, "org.freedesktop.fwupd.policy", NULL);
+	g_autofree gchar *directory = NULL;
+	g_autofree gchar *filename = NULL;
+
+	if (g_getenv("FWUPD_POLKIT_NOCHECK") != NULL) {
+		return TRUE;
+	}
```
hughsie, line 3703: no braces please, then LGTM.

hughsie (changes requested):

superm1 (approved):

superm1:
I've filed https://forum.snapcraft.io/t/classic-strict-fwupd/33731 for the request to allow the upload and connect those.

PR #5066, opened 2022-09-22
trivial: Fix the Pre-boot DMA protection HSI long description

hughsie:
The original string is ambiguous to translate and logically incorrect.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):

superm1:
BTW - what is up with the fuzzing job?  I feel like it times out more often than not.

hughsie:
> BTW - what is up with the fuzzing job

It's doing quite a bit; I'll have a poke and see if we can do less upfront. I fear compiling Glib is kinda heavy.

PR #5071, opened 2022-09-23
Add a helper to get a GUsbDevice from a FuUdevDevice

hughsie:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -2081,6 +2081,75 @@ fu_udev_device_get_children_with_subsystem(FuUdevDevice *self, const gchar *cons
 	return g_steal_pointer(&out);
 }
 
+/**
+ * fu_udev_device_find_usb_device:
+ * @FuUdevDevice: a #FuUdevDevice
+ * @error: (nullable): optional return location for an error
+ *
+ * Gets the matching #GUsbDevice for the #GUdevDevice.
+ *
+ * NOTE: This should never be stored in the device class as an instance variable, as the lifecycle
+ * for `GUsbDevice` may be different to the `FuUdevDevice`. Every time the `GUsbDevice` is used
+ * this function should be called.
+ *
+ * Returns: (transfer full): a #GUsbDevice, or NULL if unset or invalid
+ *
+ * Since: 1.8.7
+ **/
+GUsbDevice *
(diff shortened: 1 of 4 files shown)


hughsie:
100.0% untested.

superm1 (commented):
I would think you want to return the FuUSBDevice associated not the FuUSBDevice, no?

hughsie:
> I would think you want to return the FuUSBDevice associated not the GUSBDevice, no?

I think that would be misleading and possibly lead to trouble; what happens if we added that FuUsbDevice to the engine with a different context? The life-cycle stuff is all wrong.

superm1:
I guess it depends on what you want to do with the helper :)

superm1 (commented):
I'd still like to see how this is to be used.  Normally I don't think we really want to have plugins _directly_ interacting with gusb.

hughsie:
Ack, agree. I don't want it to be a workaround for wait-for-replug for instance.

hughsie:
In the example given by Logitech, I think it makes sense to add this.

superm1:
But so what does the plugin do with this?

I am still not convinced we want to normalize plugins to directly interact with `GUsbDevice` instead of `FuUsbDevice`.
If the lifecycle is wrong, should the relationship be created by the engine between the two and put into metadata?  Parent/child or sibling or something?

hughsie:
>  If the lifecycle is wrong

So my thinking here is that we should never add either the `GUdevDevice` from `fu_usb_device_find_udev_device()` or the `GUsbDevice` from `fu_udev_device_find_usb_device()` as a `FuDevice`-derived device. i.e. we'd have the FuUdevDevice as the lifecycled device and then we'd get the `GUsbDevice` **every single time** we needed to do any kind of bulk/control transfer and not store it in any kind of private state, i.e. it's only valid for the lifetime of the current function, not matching the lifecycle of the FuUdevDevice.

Would you be more comfortable if we enforced that you couldn't use a `fu_udev_device_find_usb_device` as the input to `fu_usb_device_new()`? Maybe we could check if the context is different, and do a critical warning or assert or something?

hughsie:
> if we enforced that you couldn't use a fu_udev_device_find_usb_device as the input to fu_usb_device_new

I added that.

superm1 (approved):

PR #5082, opened 2022-09-26
Only compress one version of the builtin-quirks

hughsie:
Saving the quirks in the GResource section worked well, but it made the build system very complicated and also meant the .data section was duplicated in both `fwupd` and `fwupdtool` -- negating a lot of the hard-fought savings.

Simplify this feature so that we just `cat` all the quirk files together, then gzip them into a single file. This means that at startup fwupd only needs to check the mtime of one file, and weirdly it's actually faster to load a smaller compressed file from disk that it is to load multiple uncompressed files.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-quirks.c
```diff
@@ -248,7 +248,7 @@ fu_quirks_add_quirks_for_path(FuQuirks *self, XbBuilder *builder, const gchar *p
 	if (dir == NULL)
 		return FALSE;
 	while ((tmp = g_dir_read_name(dir)) != NULL) {
-		if (!g_str_has_suffix(tmp, ".quirk")) {
+		if (!g_str_has_suffix(tmp, ".quirk") && !g_str_has_suffix(tmp, ".quirk.gz")) {
```
superm1, line 251: Presumably you still want to allow uncompressed quirks too. So perhaps want both checks.

meson.build
```diff
@@ -571,16 +571,38 @@ if build_standalone
...
+  # append all the quirks into one big file
+  builtin_quirk = custom_target('builtin-quirk',
+    input: plugin_quirks,
+    output: 'builtin.quirk',
+    capture: true,
+    command: ['cat', '@INPUT@'],
+  )
+
+  # gzip it
+  gzip = find_program('gzip')
```
superm1, line 591: Did you profile lzma vs gzip?

meson.build
```diff
@@ -571,16 +571,38 @@ if build_standalone
...
-  # * for -Dgresource_quirks=enabled, they get built into ./src/fwupd and the files are ignored
-  # * for -Dgresource_quirks=disabled, the files get installed into quirks.d
   subdir('plugins')
   subdir('src')
-
 endif
 subdir('docs')
 
+# install the quirk files
+if get_option('compressed_quirks')
```
superm1, line 580: I think you need to fail or disable the feature if gzip command not found.

meson_options.txt
```diff
@@ -1,7 +1,7 @@
 option('build', type : 'combo', choices : ['all', 'standalone', 'library'], value : 'all', description : 'build type')
 option('consolekit', type : 'feature', description : 'ConsoleKit support', deprecated: {'true': 'enabled', 'false': 'disabled'})
 option('static_analysis', type : 'boolean', value : false, description : 'enable GCC static analysis support')
-option('gresource_quirks', type : 'feature', description : 'compile the quirk files into the fwupd binary')
+option('compressed_quirks', type : 'boolean', description : 'compile the quirk files into a compressed blob')
```
superm1, line 4: I think this should stay as a feature. It can be auto enabled when gzip is found.

superm1:
> Simplify this feature so that we just cat all the quirk files together, then gzip them into a single file. This means that at startup fwupd only needs to check the mtime of one file, and weirdly it's actually faster to load a smaller compressed file from disk that it is to load multiple uncompressed files.

I had another thought here.  Does the mtime of the directory change when a file within it changes?  If so, can't you just ignore the directory and use the already cached `quirks.xmlb`?

libfwupdplugin/fu-quirks.c
hughsie: ohh it's saying "if not .quirk and not .quirk.gz" then skip this file. Is my logic wonky?

meson.build
hughsie: lzma was smaller, but needs libxmlb 0.3.3 -- and we only BR on 0.1.13 for RHEL and Ubuntu LTS.

hughsie:
>  Does the mtime of the directory change when a file within it changes

It's a good question, but it's not the mtime checking that's the slow bit, it's the stat and file read io that's slow.

meson.build
superm1: Okay, maybe a future optimization some day we can bump that. Leave a comment for future selves the numbers in difference to justify looking into it someday

meson.build
hughsie: We already have a hard requirement of `find_program('gzip')` for `host-emulate.d` -- and tbh, i think gzip is a just fine hard dep.

superm1:
> >  Does the mtime of the directory change when a file within it changes
> 
> It's a good question, but it's not the mtime checking that's the slow bit, it's the stat and file read io that's slow.

But can you use the mtime or the stat to proxy to tell whether it needs changing?

hughsie:
I think we do have to stat all the files that make up the silo to make sure that the silo is still valid. Making one large file means there's only one thing to stat, not ~90.

meson.build
hughsie: Fixed.

meson_options.txt
hughsie: My thinking here is that the compressed_quirks might make it slightly harder to do development (as in you have to run from a checkout, rather than just dumping files into quirks.d to *replace* the existing quirk data -- maybe. In the common case (adding an extra VID/PID) you can still add quirks as before. As in it's not so much related to the availability of gzip as to the behaviour the system builder wanted.

Maybe we just rip off the sticky plaster and make compressed_quirks the only way to do it? i.e. if there's a (in my head, theoretical) problem with overriding compressed builtin data then we fix the bug then and just don't have an option to disable it. Certainly KISS is easier to debug... Certainly using one compressed file is two order of magnitude less space for quirks on disk, as at the moment even the smallest 12 byte quirk files still take up one block (e.g.4kb) on disk, and 100x4kb is kind big.

hughsie: I've made it simpler.

superm1 (approved):
this approach lgtm

superm1 (approved):

PR #5094, opened 2022-09-28
Remove references to the plugin library version

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


libfwupdplugin/fu-path.c
```diff
@@ -317,16 +317,16 @@ fu_path_from_kind(FuPathKind path_kind)
 
 	/* /usr/lib/<triplet>/fwupd-plugins-3 */
 	case FU_PATH_KIND_PLUGINDIR_PKG:
```
superm1, line 319: hmm, I guess we might want to rename this too.

src/fwupd.conf
```diff
@@ -1 +1 @@
-@FWUPD_PLUGINDIR@
+@FWUPD_INTERNAL_LIBDIR@
```
superm1, line 1: I actually realize I don't know why we have this conffile, am I missing something?

libfwupdplugin/fu-path.h
```diff
@@ -53,7 +53,7 @@ typedef enum {
 	FU_PATH_KIND_EFIAPPDIR,
 	FU_PATH_KIND_LOCALSTATEDIR,
 	FU_PATH_KIND_LOCALSTATEDIR_PKG,
-	FU_PATH_KIND_PLUGINDIR_PKG,
+	FU_PATH_KIND_INTERNAL_LIBDIR_PKG,
```
hughsie, line 56: Maybe `FU_PATH_KIND_LIBDIR_PKG`?

libfwupdplugin/fu-self-test.c
```diff
@@ -3638,7 +3638,7 @@ main(int argc, char **argv)
 
 	testdatadir = g_test_build_filename(G_TEST_DIST, "tests", NULL);
 	(void)g_setenv("FWUPD_DATADIR", testdatadir, TRUE);
-	(void)g_setenv("FWUPD_PLUGINDIR", testdatadir, TRUE);
+	(void)g_setenv("FWUPD_INTERNAL_LIBDIR", testdatadir, TRUE);
```
hughsie, line 3641: and `FWUPD_LIBDIR_PKG` perhaps?

superm1: fixed

libfwupdplugin/fu-path.h
superm1: fixed

libfwupdplugin/meson.build
```diff
@@ -259,7 +259,7 @@ fwupdplugin = library(
   ],
   link_args: cc.get_supported_link_arguments([vflag]),
   link_depends: fwupdplugin_mapfile,
-  install_dir: plugin_dir,
+  install_dir: lib_dir_pkg,
```
hughsie, line 262: I think `lib_dir_pkg` would be a closer match to the existing `libdir`

hughsie: Gah, `libdir_pkg` late here.

hughsie (approved):
looks great, thanks.

jtojnar:
Would not it still be nicer to distinguish plug-ins from other internal libraries like `libfwupdplugin.so`, `libfwupdengine.so` and `libfwupdutil.so`? The former could go to `lib/fwupd-1.8.7/plug-ins`  while the latter to `lib/fwupd-1.8.7/plug-ins`. 

I am interested in this because in Nixpgks, we are splitting fwupd into library and daemon part but since the output path contains a hash that is fixed at build time, we cannot just split it post hoc like FHS-based distros do – we need to patch the build system. Previously, we had a [single variable for plug-ins we could patch](https://github.com/NixOS/nixpkgs/blob/51944000638cb69c69649cfb56a8fb6e56bd2380/pkgs/os-specific/linux/firmware/fwupd/install-fwupdplugin-to-out.patch#L34) but this is no longer possible.

jtojnar:
> > we are splitting fwupd into library and daemon part
> 
> why? What use is one without the other?

Nix makes dependency tree for C libraries completely static so since e.g. ~~gnome-control-center~~ gnome-software depends on libfwupd, not splitting would pull in the daemon with all the plug-ins, even the though user might not want to have the fwupd daemon for some reason.

> libfwupdengine.so` is required for the daemon to run, and completely useless without the daemon.

Then we can explicitly patch it to move it to `daemon` package like we have been previously doing with [`libfwupdplugin`](https://github.com/NixOS/nixpkgs/blob/efa60e11cf4f5e8f6c2857616057f71875ee020e/pkgs/os-specific/linux/firmware/fwupd/install-fwupdplugin-to-out.patch#L11) and for [all plug-ins by changing `plugin_dir` variable in Meson](https://github.com/NixOS/nixpkgs/blob/efa60e11cf4f5e8f6c2857616057f71875ee020e/pkgs/os-specific/linux/firmware/fwupd/install-fwupdplugin-to-out.patch#L34).

After this PR, we can have few options what to patch:

- [`libdir`](https://github.com/fwupd/fwupd/blob/716f41b7b28771e73af7969154efe51de75d7054/meson.build#L175) – that would misplace `libfwupd.so` & co.
- [`libdir_pkg`](https://github.com/fwupd/fwupd/blob/716f41b7b28771e73af7969154efe51de75d7054/meson.build#L507) – that would probably work correctly at the moment but I am not sure how future proof it would be – it would break if new library belonging to the `libfwupd` package were to be introduced.
- individual targets’s `install_dir` – those would be most precise but also labour intensive (especially when rebasing our patches for new versions)

I think the state before this PR where all the plug-ins used a single variable we could patch and additional files were moved individually fits best our workflow.

superm1:
I think individual targets install_dir is really what you should patch.

Going forward the only external library is libfwupd. Everything else that we introduce should go into the daemon package.

PR #5095, opened 2022-09-29
Build FuEngine functionality into libfwupdengine

hughsie:
This speeds up the build, and further reduces the on disk size by 340kB.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -213,6 +213,24 @@ fwupdpluginbuiltin = library(
   ],
 )
 
+fwupdengine = library(
+  'fwupdengine',
+  resources_src,
+  sources: daemon_src,
+  install: true,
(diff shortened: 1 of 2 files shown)


src/meson.build
```diff
@@ -240,6 +257,7 @@ fwupdtool = executable(
     fwupdplugin,
     fwupdpluginbuiltin,
     fwupdutil,
+    fwupdengine,
```
superm1, line 260: Alphabetical

superm1 (approved):
Just an ordering nit, LGTM

src/meson.build
hughsie: Fixed, thanks.

PR #5101, opened 2022-09-29
trivial: Make the build simpler by not using libfwupdengine

hughsie:
Just compile daemon sources into libfwupdpluginbuiltin.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -34,7 +34,7 @@ if polkit.found()
   client_src += 'fu-polkit-agent.c'
 endif
 
-daemon_src = [
+fwupd_engine_src = [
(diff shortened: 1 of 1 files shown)


hughsie:
internal is a bit non-specific. It's only the stuff we need for loading the engine. Maybe it should be libfwupdengine, and we kill libfwupdbuiltin?

superm1:
> internal is a bit non-specific. It's only the stuff we need for loading the engine. Maybe it should be libfwupdengine, and we kill libfwupdbuiltin?

Fine by me

superm1 (approved):

PR #5103, opened 2022-09-29
modem-manager: check driver name instead of bus type

ivanmikh:
In case of Firehose, firmware update mechanism depends on the device bus (PCI or USB) and driver type.

Currently `fu_device_get_physical_id` is used to determine PCI devices. This approach is not perfect since USB devices can be sitting on the PCI bus as well.
Also, loading of the firehose programmer is implemented in the MHI driver, so we should only use this update mechanism for the devices that use `mhi-pci-generic` driver.

This commit adds `mm_modem_get_drivers` to determine devices that use `mhi-pci-generic` driver.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/modem-manager/fu-mm-device.c
+++ b/plugins/modem-manager/fu-mm-device.c
@@ -1509,6 +1509,7 @@ fu_mm_device_write_firmware_firehose(FuDevice *device,
 				     GError **error)
 {
 	FuMmDevice *self = FU_MM_DEVICE(device);
+	MMModem *modem = mm_object_peek_modem(self->omodem);
 	GBytes *firehose_rawprogram;
 	GBytes *firehose_prog;
 	g_autoptr(FuDeviceLocker) locker = NULL;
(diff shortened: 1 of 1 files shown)


hughsie:
Do we need to check the build time mm version?

plugins/modem-manager/fu-mm-device.c
```diff
@@ -1550,10 +1552,10 @@ fu_mm_device_write_firmware_firehose(FuDevice *device,
 	/* Firehose program needs to be loaded to the modem before firehose update process can
 	 * start. Generally, modems use Sahara protocol to load the firehose binary.
 	 *
-	 * In case of MHI PCI modems, the mhi_wwan driver reads the firehose binary from the
+	 * In case of MHI PCI modems, the mhi-pci-generic driver reads the firehose binary from the
 	 * firmware-loader and writes it to the modem.
 	 **/
-	if (g_strrstr(fu_device_get_physical_id(device), "pci") && self->port_qcdm != NULL) {
+	if (g_strrstr(driver, "mhi-pci-generic") && self->port_qcdm != NULL) {
```
hughsie, line 1558: Does g_strv_contains (iirc) work here?

ivanmikh: oh nice, didn't know about that. updated it in the commit

hughsie (approved):

PR #5110, opened 2022-09-30
Do not check the build hash of modular plugins

hughsie:
We removed the abilty to easily build modular plugins, and so we don't need to worry so much about random 3rd party modules being loaded.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Debian's fwupd-tests.install needs changing.
Also daemon.conf needs to stop referencing this.

hughsie:
@superm1 but you're broadly in favour of the proposal?

superm1:
> @superm1 but you're broadly in favour of the proposal?

I oscillate. This is the only thing that taints the daemon isn't it? So that should mean all tainting goes away too.

It also means that someone can drop a plugin binary in /usr/lib/*/fwupd-RELEASE and it works without tainting.

So it's possibly a much bigger change

hughsie:
> This is the only thing that taints the daemon isn't it? So that should mean all tainting goes away too.

Yup, I've got a local version here that does that too.

> It also means that someone can drop a plugin binary in /usr/lib/*/fwupd-RELEASE and it works without tainting.

I'm not sure how you would build a `.so` targetting the exact fwupd release; it's nether easy or obvious now. IMHO it's in the realms of theoretical, not likely.

superm1:
> Yup, I've got a local version here that does that too.

Isn't it a security attribute too?  So we'll have to rev HSI for it and worry about implications for any GUI clients.

hughsie:
> Isn't it a security attribute too

Ya, but we need it for the disabled plugins case so we need to keep it.

superm1:
> > Isn't it a security attribute too
> 
> Ya, but we need it for the disabled plugins case so we need to keep it.

We're always going to have disabled plugins though.  `test` is disabled by default but builtin.

hughsie:
Yup, we allow-list those:

		if (!g_str_has_prefix(name_tmp, "test") && g_strcmp0(name_tmp, "invalid") != 0) {
			disabled_plugins = TRUE;
			break;
		}

superm1:
> Yup, we allow-list those:
> 
> ```
> 	if (!g_str_has_prefix(name_tmp, "test") && g_strcmp0(name_tmp, "invalid") != 0) {
> 		disabled_plugins = TRUE;
> 		break;
> 	}
> ```

I guess that logic needs to change for dropping the invalid plugin.

I suppose I'm fine with this PR now then once you change that logic.

PR #5115, opened 2022-10-03
Support an enumerated release action for the post-update warning

hughsie:
The hardware OEMs have chosen all very different and difficult to understand post-install warnings -- it was perhaps a mistake to allow freeform text. Additionally these warnings are important, and not translated, and so many users are going to be blindly clicking 'Ok' to make the dialog they can't understand go away.

After analysing a few dozens of the warnings on the public LVFS firmwares we can see there's only really two different cases we need to support -- and one already has a FwupdRequest ID.

Add an enumerated type that can be set by the LVFS so that we can translate the user-visible message text. The LVFS will continue to set the fallback text and image and so this is forwards and backwards compatible.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -698,6 +698,26 @@ typedef guint64 FwupdDeviceProblem;
  */
 #define FWUPD_RELEASE_FLAG_IS_COMMUNITY (1u << 7)
 /**
+ * FWUPD_RELEASE_FLAG_UPDATE_ACTION_DO_NOT_UNPLUG_POWER:
+ *
+ * Show the user a message not to unplug the machine from the AC power, e.g.
+ *
+ * "Do not turn off your computer or remove the AC adaptor while the update is in progress."
+ *
+ * Since: 1.8.6
+ */
+#define FWUPD_RELEASE_FLAG_UPDATE_ACTION_DO_NOT_UNPLUG_POWER (1u << 8)
+/**
+ * FWUPD_RELEASE_FLAG_UPDATE_ACTION_REMOVE_REPLUG:
+ *
+ * Show the user a message to replug the USB cable, e.g.
+ *
+ * "The update will continue when the device USB cable is unplugged and then re-inserted."
+ *
+ * Since: 1.8.6
+ */
+#define FWUPD_RELEASE_FLAG_UPDATE_ACTION_REMOVE_REPLUG (1u << 9)
+/**
  * FWUPD_RELEASE_FLAG_UNKNOWN:
  *
  * The release flag is unknown, typically caused by using mismatched client and daemon.
--- a/libfwupd/fwupd-request.h
+++ b/libfwupd/fwupd-request.h
@@ -59,10 +59,59 @@ typedef enum {
  */
 #define FWUPD_REQUEST_ID_PRESS_UNLOCK "org.freedesktop.fwupd.request.press-unlock"
 
+/**
+ * FWUPD_REQUEST_ID_PRESS_DO_NOT_UNPLUG_POWER:
+ *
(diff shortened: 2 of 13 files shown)


hughsie:
Requires https://gitlab.com/fwupd/lvfs-website/-/merge_requests/1371 also

libfwupd/fwupd-request.h
```diff
@@ -59,11 +59,60 @@ typedef enum {
...
+ * No flags are set.
+ *
+ * Since: 1.8.6
+ */
+#define FWUPD_REQUEST_FLAG_NONE (0u)
+
+/**
+ * FWUPD_REQUEST_FLAG_FORCE_GENERIC:
+ *
+ * The payload binary is trusted.
```
superm1, line 84: Huh?

libfwupd/fwupd-enums.h
```diff
@@ -697,6 +697,26 @@ typedef guint64 FwupdDeviceProblem;
...
+ *
+ * Since: 1.8.6
+ */
+#define FWUPD_RELEASE_FLAG_UPDATE_ACTION_DO_NOT_UNPLUG_POWER (1u << 8)
+/**
+ * FWUPD_RELEASE_FLAG_UPDATE_ACTION_REMOVE_REPLUG:
+ *
+ * Show the user a message to replug the USB cable, e.g.
+ *
+ * "The update will continue when the device USB cable is unplugged and then re-inserted."
```
superm1, line 715: This is from Dell dock right? I don't think it technically needs to be reinserted. The update happens whether the computer returned or not.

superm1 (commented):
As a general comment - shouldn't these be device flags not release flags? They're historically release because of how they were set and delivered in the past.

But if it's being changed anyways why not reflect it properly independent of the remote?

libfwupd/fwupd-request.h
hughsie: Fixed.

libfwupd/fwupd-enums.h
hughsie: This is actually for CCGX docks like the HP one and various other reference designs from Synaptics.

superm1: If that's the case then isn't there two different statements needed since Dell and HP ones work slightly differently?

hughsie:
> As a general comment - shouldn't these be device flags not release flags?

Hmm, indeed. There are (were?) two reasons for putting this data in the `LVFS::UpdateMessage` field.

* The update requires the unplug/replug action to deploy the firmware
* The vendor policy requires a warning for UEFI capsule updates as removing the power during update might brick the system

For the first point you could argue that it's really a missing `fwupd_request_new()` in the plugin code -- possibly triggered by a plugin private flag or possibly better a `FuDeviceInternalFlags` in the daemon -- and I'd probably agree with you that it would be better as a device flag.

For the second point you could argue that it's a bad corporate policy and that updates should always be safe regardless of the dumb things the user does -- or that we should warn for _all UEFI updates_, or even _all updates_ full stop. I'm not sure on how safer that makes things. At the moment the vendors are putting this kind of "Do not remove AC" message in the update details text which is **entirely** the wrong place and I want to give them a standard way to do it that's translated.

superm1:
> > I'd probably agree with you that it would be better as a device flag.
> 
> Glad we agree :)
> 
> > or that we should warn for all UEFI updates, or even all updates full stop. I'm not sure on how safer that makes things.
> 
> I think warning for all UEFI updates is reasonable.

Another idea - rather than warn for all UEFI updates, how about warn to keep stuff plugged in on everything that reboots?

hughsie:
Yes, that's probably sane. Will ponder. Lets close this PR and I'll have a regroup and rethink.

PR #5116, opened 2022-10-03
Add FwupdRequestFlags for future usage

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-request.c
+++ b/libfwupd/fwupd-request.c
@@ -21,13 +21,23 @@
 typedef struct {
 	gchar *id;
 	FwupdRequestKind kind;
+	FwupdRequestFlags flags;
 	guint64 created;
 	gchar *device_id;
 	gchar *message;
 	gchar *image;
 } FwupdRequestPrivate;
 
-enum { PROP_0, PROP_ID, PROP_KIND, PROP_MESSAGE, PROP_IMAGE, PROP_DEVICE_ID, PROP_LAST };
+enum {
+	PROP_0,
+	PROP_ID,
+	PROP_KIND,
+	PROP_FLAGS,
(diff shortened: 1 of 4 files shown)


libfwupd/fwupd-request.c
```diff
@@ -403,6 +457,100 @@ fwupd_request_set_kind(FwupdRequest *self, FwupdRequestKind kind)
...
+{
+	FwupdRequestPrivate *priv = GET_PRIVATE(self);
+	g_return_val_if_fail(FWUPD_IS_REQUEST(self), 0);
+	return priv->flags;
+}
+
+/**
+ * fwupd_request_set_flags:
+ * @self: a #FwupdRequest
+ * @flags: request flags, e.g. %FWUPD_REQUEST_FLAG_FORCE_GENERIC
```
superm1, line 481: This flag `FWUPD_REQUEST_FLAG_FORCE_GENERIC` doesn't exist in the PR

hughsie: Ta, fixed.

superm1 (approved):
The code looks fine, but I suggest holding off to merge it based on the comments from #5115 since I think the design changes quite a bit from those.  Maybe this is superfluous.

hughsie:
I think we need this anyway for the daemon to signal that it's okay to substitute the plugin-generated update text with translatable text. It seems a useful thing to add regardless IMHO.

superm1:
> I think we need this anyway for the daemon to signal that it's okay to substitute the plugin-generated update text with translatable text. It seems a useful thing to add regardless IMHO.

OK, sounds good.

PR #5118, opened 2022-10-03
Allow optionally translating request messages

hughsie:
Also add two more request IDs to make these specific enough to be useful.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-request.c
+++ b/libfwupd/fwupd-request.c
@@ -101,6 +101,10 @@ fwupd_request_flag_to_string(FwupdRequestFlags flag)
 {
 	if (flag == FWUPD_REQUEST_FLAG_NONE)
 		return "none";
+	if (flag == FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGE)
+		return "allow-generic-message";
+	if (flag == FWUPD_REQUEST_FLAG_ALLOW_GENERIC_IMAGE)
+		return "allow-generic-image";
 	return NULL;
 }
 
@@ -117,6 +121,10 @@ fwupd_request_flag_to_string(FwupdRequestFlags flag)
 FwupdRequestFlags
 fwupd_request_flag_from_string(const gchar *flag)
 {
+	if (g_strcmp0(flag, "allow-generic-message") == 0)
+		return FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGE;
+	if (g_strcmp0(flag, "allow-generic-image") == 0)
+		return FWUPD_REQUEST_FLAG_ALLOW_GENERIC_IMAGE;
 	return FWUPD_REQUEST_FLAG_NONE;
 }
 
--- a/libfwupd/fwupd-request.h
+++ b/libfwupd/fwupd-request.h
@@ -44,7 +44,8 @@ typedef enum {
 /**
  * FWUPD_REQUEST_ID_REMOVE_REPLUG:
  *
- * The user needs to remove and reinsert the device.
+ * The user needs to remove and reinsert the device to complete the update, e.g.
+ * "The update will continue when the device USB cable has been unplugged and then re-inserted."
  *
  * Since 1.6.2
  */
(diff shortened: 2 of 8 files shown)


libfwupd/fwupd-request.c
```diff
@@ -101,6 +101,8 @@ fwupd_request_flag_to_string(FwupdRequestFlags flag)
 {
 	if (flag == FWUPD_REQUEST_FLAG_NONE)
 		return "none";
+	if (flag == FWUPD_REQUEST_FLAG_ALLOW_GENERIC)
```
superm1, line 104: if you don't feel strongly, can this be `FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGES`?

hughsie: We could, but wouldn't we want a generic icon/image at the same time?

libfwupd/fwupd-request.h
```diff
@@ -53,12 +54,33 @@ typedef enum {
...
- * The user needs to press unlock on the device.
+ * The user needs to press unlock on the device to continue, e.g.
+ * "Press unlock on the device to continue the update process."
  *
  * Since 1.6.2
  */
 #define FWUPD_REQUEST_ID_PRESS_UNLOCK "org.freedesktop.fwupd.request.press-unlock"
 
+/**
+ * FWUPD_REQUEST_ID_REMOVE_DEVICE:
```
superm1, line 65: Maybe make this `REQUEST_ID_REMOVE_USB_DEVICE` or change the string to drop the word USB.
Otherwise we have a problem some day with some other sort of PCIe device or such.

libfwupd/fwupd-request.c
superm1: Oh...  OK never mind.

superm1: How about we instead have two flags, one for message and one for icon?  Then the vendor can prefer the translated message but allow their own logo.

libfwupd/fwupd-request.h
```diff
@@ -53,12 +54,33 @@ typedef enum {
...
+ *
+ * Since 1.8.6
+ */
+#define FWUPD_REQUEST_ID_REMOVE_DEVICE "org.freedesktop.fwupd.request.remove-device"
+
+/**
+ * FWUPD_REQUEST_ID_DO_NOT_POWER_OFF:
+ *
+ * Show the user a message not to unplug the machine from the AC power, e.g.
+ * "Do not turn off your computer or remove the AC adaptor while the update is in progress."
```
superm1, line 78: Knowing some vendors have black screens as part of their update process how about instead "Until the update is complete".

src/fu-util-common.c
```diff
@@ -2164,6 +2164,33 @@ fu_util_remote_to_string(FwupdRemote *remote, guint idt)
 	return g_string_free(g_steal_pointer(&str), FALSE);
 }
 
+const gchar *
+fu_util_request_get_message(FwupdRequest *req)
```
superm1, line 2168: as a thought, I wonder if these should be emitted from libfwupd?  Then it's easier on GUI clients?

libfwupd/fwupd-request.h
hughsie: `Do not turn off your computer or remove the AC adaptor until you are sure the update is complete.` -- maybe?

src/fu-util-common.c
hughsie: I'd really like frontends to use the gnome/kde specific themed icons where applicable.

libfwupd/fwupd-request.h
superm1: That's good

libfwupd/fwupd-request.h
superm1: That's good

src/fu-util-common.c
superm1: But for the message string, wouldn't a helper be good for them to pick up from?

libfwupd/fwupd-request.c
hughsie: Works for me, good idea.

src/fu-util-common.c
superm1: OK

libfwupd/fwupd-request.h
hughsie: Fixed.

libfwupd/fwupd-request.h
hughsie: Fixed.

superm1 (approved):

PR #5130, opened 2022-10-07
Make the CoSWID parser more full-featured

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-coswid-common.h
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#pragma once
+
+#include <glib.h>
+
+#define FU_COSWID_GLOBAL_MAP_TAG_ID		       0
+#define FU_COSWID_GLOBAL_MAP_SOFTWARE_NAME	       1
```
superm1, line 12: Could all these be an enum instead?

superm1 (dismissed):

libfwupdplugin/fu-coswid-common.h
hughsie: Good idea, that also cleans up the code randomly too. Fixed.

PR #5133, opened 2022-10-07
Add some more profiling to debug custom device GType creation

hughsie:
It turns out using tss2_esys is slow (200ms) -- and now we know.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1 (approved):
That's more expensive than I expected. I wonder how badly we really need those GUIDs and information it gathers. Could we defer it's init to the first time the information it provides is really accessed?

hughsie:
> That's more expensive than I expected

Agree. I'm going to dig into the specifics and find out what's causing the slowdown. I can only see it on this one Chromebox I was working on, on my desktop it's invisible.

PR #5135, opened 2022-10-07
add an rpath to things not compiled into a library

superm1:
A harmless error shows up in debian packages at build time:
```
dpkg-shlibdeps: warning: cannot find library libfwupdplugin.so needed by debian/fwupd/usr/lib/x86_64-linux-gnu/fwupd-1.8.6/libfu_plugin_flashrom.so (ELF format: 'elf64-x86-64' abi: '0201003e00000000'; RPATH: '')
```

This doesn't cause a functional problem because libfwupdplugin has already been loaded by the daemon by the time these libraries are loaded.

In case the `dpkg-shlibdeps` checker becomes more stringent in the future fix the warning.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
If CI is fine, fine with me.

Issue #5138, opened 2022-10-08
snap fails to run


Overcraft90:
**Describe the bug**
When opened for the first time after booting into the system the `Flatpack Software App` shows an error message related to _**fwupd**_.

**Steps to Reproduce**
Log into the account, press on the app icon Software (Flatpack).

**Expected behavior**
A clear screen which allows me to navigate the app and search for system updates.

**fwupd version information**
Please provide the version of the daemon and client.

```shell
/snap/fwupd/4397/bin/fwupdmgr: error while loading shared libraries: libfwupdplugin.so: cannot open shared object file: No such file or directory
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):
I installed _**fwupd**_ on my machine using the following link: https://flathub.org/apps/details/org.gnome.Firmware. 

However, today for some reason it stopped working properly. Noted, the app has been installed long ago; today, it simply happened to have this malfunction.
<details>

<summary>**fwupd device information**</summary>

Please provide the output of the fwupd devices recognized in your system.

```shell
/snap/fwupd/4397/bin/fwupdmgr: error while loading shared libraries: libfwupdplugin.so: cannot open shared object file: No such file or directory
```

</details>

**Additional questions**

- Operating system and version: UBUNTU 22.04 "Jammy Jellyfish"
- Have you tried rebooting? Yes
- Is this a regression? No

See attachment
![Screenshot from 2022-10-07 21-10-24](https://user-images.githubusercontent.com/98895614/194687889-7934babf-bb5a-44fa-bb8c-cc0d99311328.png)

superm1:
This is actually a bug with the snap it seems to me not the flatpak.

Overcraft90:
@superm1 Thanks for the clarification, thing is how can I address it?

superm1:
I think we need to dig into it to see what went wrong with the snap build. If you don't have a need for the snap I suggest you uninstall it and use the distro package instead.

Overcraft90:
Thanks again, how can I do so? In fact if I remove the Flatpak app the one installed in Snap gets deleted too and vice-versa..., or at least that is what happened during my 'experimentation' in order to fix this issue. 
Let me know, thank you!

Overcraft90:
@superm1 Never mind I did it! Just wonder when the actual app from your side will be ready. It is great to have it and very convenient.

superm1:
[...]

PR #5142, opened 2022-10-09
trivial: snap: add a workaround to the build process to ignore rpath

superm1:
snapcraft overrides the rpath for the snap to make it most compatible. Our rpath decisions make this not work properly though.  Since everything is in /snap/fwupd anyway, just move the libraries to this place.

Fixes: #5138

Link: https://snapcraft.io/blog/the-new-classic-confinement-in-snaps-even-the-classics-need-a-change

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


meson.build
```diff
@@ -18,9 +18,12 @@ conf.set('MICRO_VERSION', fwupd_micro_version)
 conf.set_quoted('PACKAGE_VERSION', fwupd_version)
 
 # get source version, falling back to package version
+fs = import('fs')
 git = find_program('git', required: false)
 tag = false
-if git.found()
+if fs.is_file('VERSION')
+  source_version = fs.read('VERSION', encoding: 'utf-8').strip()
```
hughsie, line 25: not a huge fan; how come you can't use git here?

superm1: I think it doesn't have .git at this point. I'll double check.

superm1: seems it works actually, I must have had some transient problem.

hughsie (approved):

PR #5145, opened 2022-10-10
trivial: use github actions to build snaps

superm1:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie (dismissed):
If it works, LGTM, thanks.

superm1:
OK.  My hope is that if this works it's step 1, and step 2 is that we can publish to the store using this instead of snapcraft.io.

superm1:
Let me know if you don't follow the PR, I can better explain it.

hughsie (approved):
Lets see what breaks. Thanks!

PR #5147, opened 2022-10-11
Add a new plugin to get more Intel ME data

hughsie:
Medium term I want to fold the AMT plugin into this one.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


plugins/intel-me/README.md
```diff
@@ -0,0 +1,19 @@
+# Intel ME
```
superm1, line 1: Doesn't Intel call this CSME these days?

plugins/intel-me/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+These devices use the existing GUIDs provided by the ME host interfaces.
+
+## Vendor ID Security
+
+The devices are not upgradable and thus require no vendor ID set.
+
+## External Interface Access
+
+This plugin requires read only access to `/dev/mei0`.
```
superm1, line 19: presumably it also needs an ioctl to the kernel too?

plugins/intel-me/meson.build
```diff
@@ -0,0 +1,15 @@
+if get_option('plugin_amt').disable_auto_if(host_machine.system() != 'linux').allowed()
+cargs = ['-DG_LOG_DOMAIN="FuPluginIntelMkhi"']
```
superm1, line 2: this seems inconsistent with rest of the naming

plugins/intel-me/fu-intel-me-mkhi-device.c
```diff
@@ -0,0 +1,146 @@
...
+	msg.hdr.command = 0x0A;
+	msg.flags = (1 << 29);
+	msg.data_size = 0x40;
+	msg.offset_maybe = 0x23;
+	msg.section = section;
+
+	if (!fu_mei_device_write(FU_MEI_DEVICE(self),
+				 (const guint8 *)&msg,
+				 sizeof(msg),
+				 2000,
```
superm1, line 75: timeout should probably be a define

plugins/intel-me/README.md
hughsie: As I understand it, CSME is a type of ME, and there are a few other types too.

plugins/intel-me/meson.build
hughsie: Fixed.

plugins/intel-me/README.md
superm1: sure

hughsie: Fixed.

docs/hsi.md
```diff
@@ -350,6 +350,33 @@ To meet HSI-1 on systems that run this test, the result must be `locked`. *[v1.5
 
 - [Chromium documentation for Intel ME](https://chromium.googlesource.com/chromiumos/third_party/flashrom/+/master/Documentation/mysteries_intel.txt)
 
+<a id="org.fwupd.hsi.Mei.BootGuardPlatformKey"></a>
+
+### [ME BootGuard Platform Key](#org.fwupd.hsi.Mei.BootGuardPlatformKey)
+
+The BootGuard Platform Key is fused into the CPU PCH during manufacturing by the OEM.
```
superm1, line 357: `s,CPU,platform,` perhaps??

plugins/intel-me/fu-intel-me-mca-device.c
```diff
@@ -0,0 +1,256 @@
...
+}
+
+static void
+fu_intel_me_mca_device_add_security_attrs(FuDevice *device, FuSecurityAttrs *attrs)
+{
+	FuIntelMeMcaDevice *self = FU_INTEL_ME_MCA_DEVICE(device);
+	g_autoptr(FwupdSecurityAttr) attr = NULL;
+	const gchar *leaked_sha256[] = {
+	    "WAITING-FOR-DATA-FROM-LENOVO0a3f3be5464e04bc3423e8648860a8c9d9b2",
+	    "WAITING-FOR-DATA-FROM-LENOVO6f2ef3296fc9eca86bff3539801e9e265b22",
```
superm1, line 222: This information isn't available yet from the leak?

plugins/intel-me/fu-intel-me-mca-device.c
```diff
@@ -0,0 +1,256 @@
...
+	fu_security_attrs_append(attrs, attr);
+
+	/* no checksums */
+	if (fu_device_get_checksums(device)->len == 0) {
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_MISSING_DATA);
+		return;
+	}
+
+	/* check for any of the leaked keys */
+	for (guint i = 0; leaked_sha256[i] != NULL; i++) {
```
superm1, line 237: I realize this ups the complexity; but what do you think about providing the IDs to check for in LVFS metadata?  This can make the plugin a lot more responsive in the future.

plugins/intel-me/meson.build
```diff
@@ -0,0 +1,16 @@
+if get_option('plugin_amt').disable_auto_if(host_machine.system() != 'linux').allowed()
```
superm1, line 1: presumably you want to be renaming the plugin option from `plugin_amt` now

hughsie: I wasn't sure; can we include it as a "deprecated" name to not break things like openbmc?

superm1: I don't know if deprecated works for renaming things.  However openbmc won't break.  They compile with `-Dauto_features=disabled`

hughsie: Agree, fixed.

plugins/intel-me/fu-intel-me-mca-device.c
hughsie: Maybe not into metadata, but we probably should use a quirk file. Lets do this as phase 2, I'd like to get the initial code into users hand fairly quickly.

hughsie: We also need a CVE/VINCE ID for this; I'll start that process.

plugins/intel-me/fu-intel-me-mca-device.c
hughsie: I've added the keys which I've calculated.

superm1 (dismissed):

PR #5150, opened 2022-10-11
Do not use GKeyFile to parse quirk files

hughsie:
There are two reasons for this. First is that GKeyFile is quite inefficient, using a large amount of heap memory when loading. Given we don't actually use the merge and replace functionality of GKeyFile and we only need line-by-line access we can parse this ourselves and reduce the peak RSS considerably.

This also accidentally fixes another bug. Moving from multiple quirk files to a single builtin.quirk meant that multiple subsystem plugins were deduped -- which isn't really what we wanted. For example, this now works:

    [MEI]
    Plugin = one
    [MEI]
    Plugin = two

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-quirks.c
+++ b/libfwupdplugin/fu-quirks.c
@@ -119,79 +119,128 @@ fu_quirks_validate_flags(const gchar *value, GError **error)
 	return TRUE;
 }
 
+typedef struct {
+	GString *group;
+	XbBuilderNode *bn;
+	XbBuilderNode *root;
+} FuQuirksConvertHelper;
+
+static void
+fu_quirks_convert_helper_free(FuQuirksConvertHelper *helper)
+{
+	g_string_free(helper->group, TRUE);
+	g_object_unref(helper->root);
+	if (helper->bn != NULL)
+		g_object_unref(helper->bn);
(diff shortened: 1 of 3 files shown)


libfwupdplugin/fu-quirks.c
```diff
@@ -124,70 +124,92 @@ fu_quirks_convert_keyfile_to_xml(FuQuirks *self, GBytes *bytes, GError **error)
...
-				       error))
-		return NULL;
-
-	/* add each set of groups and keys */
-	groups = g_key_file_get_groups(kf, NULL);
-	for (guint i = 0; groups[i] != NULL; i++) {
-		g_auto(GStrv) keys = NULL;
-		g_autofree gchar *group_id = NULL;
+	/* split into lines */
+	lines = fu_strsplit((const gchar *)g_bytes_get_data(bytes, NULL),
```
hughsie, line 133: actually, we can do better still; fu_strsplit_full() reduces RSS peak much more.

hughsie: Fixed.

superm1 (approved):

PR #5151, opened 2022-10-11
Match the MEI UUID in quirk files

hughsie:
This allows us to be more specific when matching devices, and also means we get more attributes 'for free' from the FuUdevDevice->probe().

This would allow us to have multiple device GTypes handling multiple MEI interfaces in the same plugin., for instance, PTHI and MKHI.

The slight fly in the ointment is that the kernel does not set the 'dev' for the mei_me devices, but it's always going to be just /dev/mei0, so hardcode it.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupdplugin/fu-mei-device.c
```diff
@@ -65,11 +67,25 @@ fu_mei_device_to_string(FuDevice *device, guint idt, GString *str)
...
 fu_mei_device_probe(FuDevice *device, GError **error)
 {
-	/* only open the main device */
-	if (fu_udev_device_get_device_file(FU_UDEV_DEVICE(device)) == NULL) {
-		g_set_error_literal(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED, "no device");
+	FuMeiDevice *self = FU_MEI_DEVICE(device);
+	FuMeiDevicePrivate *priv = GET_PRIVATE(self);
+	const gchar *uuid;
+
+	/* the kernel is missing `dev` on mei_me children */
```
superm1, line 74: is this a kernel bug that should be filed somewhere?

superm1 (approved):

libfwupdplugin/fu-mei-device.c
hughsie: I'm not sure; I can certainly file one if you think it's worthwhile.

superm1: I don't feel strongly on it.

PR #5163, opened 2022-10-17
Only count the Microsoft hashes when getting the dbx version

hughsie:
HP include extra keys that means the version is higher than expected.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
For environment variable enabled verbose mode, do you maybe want to emit the owners of skipped keys?

hughsie:
> For environment variable enabled verbose mode

Just FWUPD_VERBOSE or something EFI specific?

superm1:
> > For environment variable enabled verbose mode
> 
> Just FWUPD_VERBOSE or something EFI specific?

None of them seem to match this stuff, I think new variable.

hughsie:
Something like that?? ^^

superm1 (approved):

PR #5164, opened 2022-10-17
Consider the release priority when returning results

hughsie:
This means we might be able to offer two versions of firmware for the same device, where one has additional requirement such as a CHID.

The idea here is to allow OEMs to distribute thier own superset dbx updates on the LVFS without having to invent an anti-CHID requirement type.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -5052,19 +5052,24 @@ static gint
 fu_engine_sort_releases_cb(gconstpointer a, gconstpointer b, gpointer user_data)
 {
 	FuDevice *device = FU_DEVICE(user_data);
-	FwupdRelease *rel_a = FWUPD_RELEASE(*((FwupdRelease **)a));
-	FwupdRelease *rel_b = FWUPD_RELEASE(*((FwupdRelease **)b));
+	FuRelease *rel_a = FU_RELEASE(*((FuRelease **)a));
+	FuRelease *rel_b = FU_RELEASE(*((FuRelease **)b));
 	gint rc;
 
 	/* first by branch */
-	rc = g_strcmp0(fwupd_release_get_branch(rel_b), fwupd_release_get_branch(rel_a));
+	rc = g_strcmp0(fu_release_get_branch(rel_b), fu_release_get_branch(rel_a));
 	if (rc != 0)
 		return rc;
 
 	/* then by version */
-	return fu_version_compare(fwupd_release_get_version(rel_b),
-				  fwupd_release_get_version(rel_a),
-				  fu_device_get_version_format(device));
+	rc = fu_version_compare(fu_release_get_version(rel_b),
+				fu_release_get_version(rel_a),
+				fu_device_get_version_format(device));
+	if (rc != 0)
+		return rc;
+
+	/* then by priority */
+	return fu_release_compare(rel_a, rel_b);
 }
 
 static gboolean
--- a/src/fu-release.c
+++ b/src/fu-release.c
@@ -33,6 +33,7 @@ struct _FuRelease {
 	gboolean is_downgrade;
 	GPtrArray *soft_reqs; /* nullable, element-type XbNode */
 	GPtrArray *hard_reqs; /* nullable, element-type XbNode */
+	guint64 priority;
 };
 
 G_DEFINE_TYPE(FuRelease, fu_release, FWUPD_TYPE_RELEASE)
@@ -618,6 +619,13 @@ fu_release_check_requirements(FuRelease *self,
 	return TRUE;
 }
 
+void
+fu_release_set_priority(FuRelease *self, guint64 priority)
+{
+	g_return_if_fail(FU_IS_RELEASE(self));
+	self->priority = priority;
+}
+
 /**
  * fu_release_load:
  * @self: a #FuRelease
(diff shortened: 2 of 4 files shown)


superm1:
OEMs are shipping their own signed additive dbx files that are different than the one distributed by Microsoft?  That makes me question whether the design for dbx updates from fwupd needs to be changing at all for some different assumptions.

hughsie:
> OEMs are shipping their own signed additive dbx files that are different than the one distributed by Microsoft

Ya, I only found out recently too. All ideas welcome.

superm1:
> > OEMs are shipping their own signed additive dbx files that are different than the one distributed by Microsoft
> 
> Ya, I only found out recently too. All ideas welcome.

I think you need:
1. A way to tell whether there are additive dbx entries in place to start compared to what is in the Microsoft one. Which I think you can do with the recent PR.
2. To make two dbx "devices".  One of them for Microsoft, one of them for the "OEM".  The OEM one if there are no dbx entries it should be version "0" but marked updatable.
3. OEMs that want to target their own updates should use their own device, and the Microsoft entries should come from the Microsoft device.
4. The "Microsoft" device maybe use the name of the CA in the KEK?  And build a GUID based on who the provider is.  Then if the KEK is changed Microsoft updates don't flow to it.

hughsie:
> should be version "0"

Why not count the number of OEM keys as the version?

>  And build a GUID based on who the provider is

I think we still want to allow people running with the OEM dbx to apply the latest updates from Microsoft -- the OEM might not be planning an update, and its certainly not going to be released on the same day.

But a good idea, will prototype.

superm1:
> Why not count the number of OEM keys as the version?

Exactly, if there are none it should be zero.  But you should make the device exist whether there are any keys or not so there is flexibility to add them later.

> I think we still want to allow people running with the OEM dbx to apply the latest updates from Microsoft -- the OEM might not be planning an update, and its certainly not going to be released on the same day.

Yes exactly.  That's why I think two devices.

hughsie:
>  But you should make the device exist whether there are any keys or not so there is flexibility to add them later

Aha, indeed. How do we know the GUID used in the dbx entry? Does it **have** to be related to the gnutls_x509_crt_t we parsed from the PK?

hughsie:
> So use the PK

    17:10:00.542 FuPluginUefiPk       PK issuer: C=JP,ST=Kanagawa,L=Yokohama,O=Lenovo Ltd.,CN=Lenovo Ltd. PK CA 2012
    17:10:00.542 FuPluginUefiPk       PK subject: C=JP,ST=Kanagawa,L=Yokohama,O=Lenovo Ltd.,CN=Lenovo Ltd. PK CA 2012

Is all we get... Or do you mean the key ID or something?

superm1:
> > So use the PK
> 
> ```
> 17:10:00.542 FuPluginUefiPk       PK issuer: C=JP,ST=Kanagawa,L=Yokohama,O=Lenovo Ltd.,CN=Lenovo Ltd. PK CA 2012
> 17:10:00.542 FuPluginUefiPk       PK subject: C=JP,ST=Kanagawa,L=Yokohama,O=Lenovo Ltd.,CN=Lenovo Ltd. PK CA 2012
> ```
> 
> Is all we get... Or do you mean the key ID or something?

At the end of the day it's going to be a hash of something in the PK.  Maybe a fingerprint of the PK?

hughsie:
>  Maybe a fingerprint of the PK?

Yup. I think my plan is to switch the uefi-pk plugin into creating a non-UPDATABLE device that we can grab in the uefi-dbx plugin and then add a duplicate FuUefiDbxDevice as a child to the existing FuUefiDbxDevice with just the OEM entries. I also think it makes sense to stop setting the version in FuEfiSignatureList and instead have something like `gchar *fu_efi_signature_list_get_version(FuEfiSignatureList *self, const gchar *owner);` so that we can get the version for each different owner in the siglist.

The other way around it is to move the `fu_firmware_parse(KEK)` to `fu_uefi_dbx_plugin_coldplug()` and create a `FuUefiDbxDevice` for each different owner. This is more straightforward, but means we can't copy the data from the PK and would probably need a quirk like:

    [77fa9abd-0359-4d32-bd60-28f4e78f784b]
    Vendor = Microsoft

 This is probably needed to do the dbx description in a good way and so it's probably not the end of the world -- I'm also not sure using the PK `issuer` is always going to be the kind of thing we want to show in the UI.

Ideas welcome.

superm1:
> 

We've got a few big companies that distribute firmware and they don't change PKs very much right?  I don't think it's a big deal to have a handful of quirks for those.

hughsie:
Hang on a minute; do we want to add the vendor-specific dbx depend on the contents of the PK or **KEK**?

superm1:
> Hang on a minute; do we want to add the vendor-specific dbx depend on the contents of the PK or **KEK**?

Whichever is actually signed from

hughsie:
> Whichever is actually signed from

According to HP that's the KEK.. but that doesn't make sense as otherwise they wouldn't be getting the MS updates either.

superm1:
> > Whichever is actually signed from
> 
> According to HP that's the KEK.. but that doesn't make sense as otherwise they wouldn't be getting the MS updates either.

Yeah... Exactly. Maybe can we gather a dump from a system setup like this to introspect?

Maybe they're making false assertions.

hughsie:
Hah, so there's a custom KEK. HP can just upload an update to the LVFS which targets that GUID instead.

hughsie:
@superm1 I still think this patch makes sense logically; okay to merge it anyway -- even if now unneeded?

superm1:
> @superm1 I still think this patch makes sense logically; okay to merge it anyway -- even if now unneeded?

Sure

PR #5168, opened 2022-10-18
Split out FuUefiPkDevice as a device for future use

hughsie:
This does not serve much purpose now, but would be useful if we need to know more about the installed PK from other plugins. If nothing else it makes the `--verbose` output more helpful.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -25,6 +25,8 @@
 #define FU_DEVICE_RETRY_OPEN_COUNT 5
 #define FU_DEVICE_RETRY_OPEN_DELAY 500 /* ms */
 
+#define FU_DEVICE_GUID_MAIN_SYSTEM_FIRMWARE "230c8b18-8d9b-53ec-838b-6cfc0383493a"
+
 /**
  * FuDevice:
  *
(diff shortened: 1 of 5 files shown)


libfwupdplugin/fu-device.c
```diff
@@ -5555,6 +5557,21 @@ fu_device_security_attr_new(FuDevice *self, const gchar *appstream_id)
 	attr = fu_security_attr_new(priv->ctx, appstream_id);
 	fwupd_security_attr_set_plugin(attr, fu_device_get_plugin(FU_DEVICE(self)));
 	fwupd_security_attr_add_guids(attr, fu_device_get_guids(FU_DEVICE(self)));
+
+	/* if the device has a parent of main-system-firmware then add those GUIDs too */
+	if (fu_device_has_parent_guid(self, FU_DEVICE_GUID_MAIN_SYSTEM_FIRMWARE)) {
```
superm1, line 5562: This is going to match a lot of things, are you sure about it?

hughsie: Less than you think; grep for fu_device_security_attr_new

superm1: Oh, I see.. OK.

plugins/uefi-pk/fu-uefi-pk-device.c
```diff
@@ -0,0 +1,250 @@
...
+	/* success */
+	fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_SUCCESS);
+	fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_VALID);
+}
+
+static void
+fu_uefi_pk_device_init(FuUefiPkDevice *self)
+{
+	fu_device_set_physical_id(FU_DEVICE(self), "pk");
+	fu_device_set_name(FU_DEVICE(self), "UEFI pk");
```
superm1, line 233: Although correct this is going to mean squat to most people.  How about `UEFI Platform Key`?  And then describe in the description what it is

superm1 (approved):
Just a nit

PR #5170, opened 2022-10-19
dbx: check only the efi executables from boot menu

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

This is purposed to check only the efi exectuables in the boot menu rather all the efi files in the ESP.

### DIFF
--- a/plugins/uefi-dbx/fu-uefi-dbx-common.c
+++ b/plugins/uefi-dbx/fu-uefi-dbx-common.c
@@ -8,6 +8,9 @@
 
 #include <fwupdplugin.h>
 
+#include <efivar/efiboot.h>
+#include <efivar/efivar-dp.h>
+
 #include "fu-efi-image.h"
 #include "fu-uefi-dbx-common.h"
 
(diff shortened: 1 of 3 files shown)


CragW:
Is there a better way to extract the efi image file path? Now it parses the text path to replace slashes.

hughsie:
What's the context; what files are we detecting that we should ignore? I agree using the targets in `BootXXXX` is probably a sane way to check that we're not going to stop the machine booting, but I'm not sure we want to to do it exactly the way it's done here.

CragW:
> What's the context; what files are we detecting that we should ignore? I agree using the targets in `BootXXXX` is probably a sane way to check that we're not going to stop the machine booting, but I'm not sure we want to to do it exactly the way it's done here.

For now it finds [all files](https://github.com/fwupd/fwupd/blob/5d38e0aeea250ee2d4fdf347a7a8a90da4203356/plugins/uefi-dbx/fu-uefi-dbx-common.c#L46) in the ESP. I've chatted with @superm1 about making a change in the way it prevents no-boot, and maybe implement it in the `libfwupdplugin` for future use.

The change at present uses `efivars` for the `device path` object, but there is no direct access to the image's file name field. As you can see the [text path](https://github.com/fwupd/fwupd/pull/5170/files#diff-052ae5f441191e7cc257e7a148d8044aa5091919004e68e9eafc8eff7ebec11fR145) in return makes me worry.

superm1 (commented):
Does the dbx blocklist currently block both shim and grub or just shims?

If it needs to cover grub too, this might need some further heuristic of if there is a shim in same directory check the grub too.

Perhaps even just check all EFI binaries in the directory the boot#### points to?

superm1:
> What's the context; what files are we detecting that we should ignore? I agree using the targets in `BootXXXX` is probably a sane way to check that we're not going to stop the machine booting, but I'm not sure we want to to do it exactly the way it's done here.

This BootXXXX detection also can be used to figure out whether you need shim in the UEFI capsule plugin (based on whether BootCurrent has a shim or not in the filename)

So I think we should have this new code in libfwupdplugin so we can make that change after this one is done.

CragW:
> Does the dbx blocklist currently block both shim and grub or just shims?

shim is included too. 
https://uefi.org/sites/default/files/resources/dbx_info_2020_2022_uefiorg_v2.csv

> Perhaps even just check all EFI binaries in the directory the boot#### points to?

Okay, will include all files along the EFI loader.

superm1:
> I'm not super happy depending on efiboot in libfwupdplugin; I wonder how hard it would be to reimplement the common case of parsing?

Yeah good point.  I think for the purpose outlined in this PR we can just do it like this.  If we introduce a generic parsing method that doesn't use efivar later we can tear this out.

plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -31,6 +34,135 @@ fu_uefi_dbx_get_authenticode_hash(const gchar *fn, GError **error)
...
+		 */
+		tp_toks = g_strsplit(text_path, "/", -1);
+		tp_toks_len = g_strv_length(tp_toks);
+		if (tp_toks_len > 0) {
+			g_autoptr(GString) efip = g_string_new(tp_toks[tp_toks_len - 1]);
+			fu_string_replace(efip, "File(", "");
+			fu_string_replace(efip, ")", "");
+			fu_string_replace(efip, "\\", "/");
+
+			g_ptr_array_add(efi_paths, g_strdup(efip->str));
```
superm1, line 159: As suggested, you should probably strip the last part of this so you get a directory instead.  Then you can list all the files in that directory and return a list of all the files.  That will make sure even if bootloaders are chained that it works properly.

plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -43,17 +175,24 @@ fu_uefi_dbx_signature_list_validate_volume(FuEfiSignatureList *siglist,
 	esp_path = fu_volume_get_mount_point(esp);
 	if (esp_path == NULL)
 		return TRUE;
-	files = fu_path_get_files(esp_path, error);
+	files = fu_uefi_dbx_bootable_efi_path(FU_EFIVAR_GUID_EFI_GLOBAL, error);
```
superm1, line 178: I think you're going to also want to explicitly add to this list _everything_ ending in `*.efi` in the efi removable path (`/efi/boot/*`)

plugins/uefi-dbx/fu-uefi-dbx-common.h
```diff
@@ -8,6 +8,8 @@
 
 #include <fwupdplugin.h>
 
+#include <efivar.h>
```
superm1, line 11: not actually needed is it?

superm1: Or if it is, maybe it should go to `fu-uefi-dbx-common.c` instead.

plugins/uefi-dbx/meson.build
```diff
@@ -12,7 +12,11 @@ plugin_builtin_uefi_dbx = static_library('fu_plugin_uefi_dbx',
   include_directories: plugin_incdirs,
   link_with: plugin_libs,
   c_args: cargs,
-  dependencies: plugin_deps,
+  dependencies: [
+    plugin_deps,
+    efiboot,
+    efivar,
```
superm1, line 18: just as a nit to avoid the same thing in 3 places in one file, can you instead do something like:

`uefi_dbx_deps = [plugin_deps, efiboot, efivar]`

And then use `dependencies: uefi_dbx_deps`?

hughsie: I don't think we want to use more of efiboot. Do we know the format of Boot####? Is it simple to parse using glib in this common case?

CragW: Trying to get them with glib.

hughsie: @CragW how did you get on? Is this PR still on your radar?

hughsie:
I'm actually working on doing this properly in the wip/hughsie/efiboot branch.

hughsie:
> I'm actually working on doing this properly in the wip/hughsie/efiboot branch.

https://github.com/fwupd/fwupd/pull/5941 might be interesting to you -- I'm still not sure about the chainloaded-to binaries. Lets close this one now we have DP parsing in libfwupdplugin.

Issue #5173, opened 2022-10-19
Failed to update metadata for lvfs: checksum failure: failed to verify data, expected 4705a9d0cacec299a079523d0fcd3ac7ef9b6221


Locutus64:
When I use Topgrade to do my system udates and it gets to the firmware update I get the below. Have tried a couple of things to fix it from another issue thread but nothing works. 

`Failed to update metadata for lvfs: checksum failure: failed to verify data, expected 4705a9d0cacec299a079523d0fcd3ac7ef9b6221`

On the lastest XeroLinux release (Arch based).

hughsie:
I get the same! debugging!

hughsie:
I get:

    wget https://fwupd.org/downloads/firmware.xml.gz.jcat
$ jcat-tool info firmware.xml.gz.jcat 
JcatFile:
  Version:               0.1
  JcatItem:
    ID:                  firmware-04761-stable.xml.gz
    AliasId:             firmware.xml.gz
    JcatBlob:
      Kind:              sha1
      Flags:             is-utf8
      Timestamp:         2022-10-19T12:47:27Z
      Size:              0x28
      Data:              c65dbc3fe6ef8b8f559d0fe6e00c3fad3f24645e
    JcatBlob:
      Kind:              sha256
      Flags:             is-utf8
      Timestamp:         2022-10-19T12:47:27Z
      Size:              0x40
      Data:              2072aec885473ed1afeb5f425b526b5a1b517954f363ee580784639065ad01d1
    JcatBlob:
      Kind:              gpg
      Flags:             is-utf8
      Timestamp:         2022-10-19T12:47:27Z
      Size:              0x1e8
      Data:              -----BEGIN PGP SIGNATURE-----
                         
                         iQEzBAABCAAdFiEEP8a4BEEO0IQNjy+XSKbYDkU4usIFAmNP8d8ACgkQSKbYDkU4
                         coGq/R4JAHJZLTceBMOe/MVLw+caPQ==
                         =Nx1T
                         -----END PGP SIGNATURE-----
                         
    JcatBlob:
      Kind:              pkcs7
      Flags:             is-utf8
      Timestamp:         2022-10-19T12:47:27Z
      Size:              0x8c0
      Data:              -----BEGIN PKCS7-----
                         MIIGUgYJKoZIhvcNAQcCoIIGQzCCBj8CAQExDTALBglghkgBZQMEAgEwCwYJKoZI
[...]

PR #5174, opened 2022-10-19
Allow parsing metadata more than 1MB in size

hughsie:
The LVFS crept over this limit yesterday. I've put some emergency commits in place that take it back down to 800KB, and I'll focus next week on getting it much lower than that.

The real problem is that we thought that g_input_stream_read_bytes() was using `count` as the chunk size, not the total size. Raise the total size to 32MB and chunk in 32kB blocks to reduce the RSS peak when loading metadata.

Fixes https://github.com/fwupd/fwupd/issues/5173

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -84,6 +84,9 @@
 
 #define MINIMUM_BATTERY_PERCENTAGE_FALLBACK 10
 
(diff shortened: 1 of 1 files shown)


hughsie:
A bit more headroom: 532K /var/lib/fwupd/metadata/lvfs/metadata.xml.gz

PR #5177, opened 2022-10-20
Use xz-compressed metadata to reduce bandwidth used by ~25%

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/remotes.d/lvfs-testing.conf
+++ b/data/remotes.d/lvfs-testing.conf
@@ -3,7 +3,7 @@
 # this remote provides metadata and firmware marked as 'testing' from the LVFS
 Enabled=false
 Title=Linux Vendor Firmware Service (testing)
-MetadataURI=https://cdn.fwupd.org/downloads/firmware-testing.xml.gz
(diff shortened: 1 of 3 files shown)


hughsie:
@superm1 if you want to test this you need to use qa.fwupd.org in the repo not cdn.fwupd.org as the server changes are not live yet

superm1:
Can we have a fallback in case we 404?  That would let potentially old local LVFS instances work better too.

hughsie:
>  That would let potentially old local LVFS instances work better too

I don't think we actually have any other instances; it's not something I want to encourage...

superm1:
OK, LGTM then

superm1 (approved):

gdamjan:
`lvfs-testing.conf` is in `install_data()` in meson.build, and needs to be in `configure()` to have the string replacement.

hughsie:
https://github.com/fwupd/fwupd/pull/5253 -- will look at the lvfs-website bits now

hughsie:
> pps. seems that url is invalid anyway:

Ohh it's a generated URL we get from the jcat file. With the fixed remote it all works.

PR #5178, opened 2022-10-20
Check system integrity when installing UEFI updates

hughsie:
Recently we had an update that changed the system-defined Platform Key, and we've certainly had updates in the past that changed the Boot#### variables.

Store some core ACPI and UEFI system integrity state and make sure that it is exactly the same when the system comes back from the requried reboot.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-capsule-plugin.c
+++ b/plugins/uefi-capsule/fu-uefi-capsule-plugin.c
@@ -997,6 +997,7 @@ static void
 fu_uefi_capsule_plugin_init(FuUefiCapsulePlugin *self)
 {
 	self->bgrt = fu_uefi_bgrt_new();
+	fu_plugin_add_flag(FU_PLUGIN(self), FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY);
 }
 
 static void
--- a/src/fu-engine-helper.c
+++ b/src/fu-engine-helper.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2020 Mario Limonciello <mario.limonciello@dell.com>
+ * Copyright (C) 2022 Richard Hughes <richard@hughsie.com>
  *
  * SPDX-License-Identifier: LGPL-2.1+
  */
@@ -186,4 +187,107 @@ fu_engine_update_devices_file(FuEngine *self, GError **error)
 	directory = fu_path_from_kind(FU_PATH_KIND_CACHEDIR_PKG);
 	target = g_build_filename(directory, "devices.json", NULL);
 	return g_file_set_contents(target, data, (gssize)len, error);
+}
+
+static void
+fu_engine_integrity_add_measurement(GHashTable *self, const gchar *id, GBytes *blob)
+{
+	g_autofree gchar *csum = g_compute_checksum_for_bytes(G_CHECKSUM_SHA256, blob);
+	g_hash_table_insert(self, g_strdup(id), g_steal_pointer(&csum));
+}
+
+static void
+fu_engine_integrity_measure_acpi(GHashTable *self)
+{
+	g_autofree gchar *path = fu_path_from_kind(FU_PATH_KIND_ACPI_TABLES);
+	const gchar *tables[] = {"SLIC", "MSDM", "TPM2", NULL};
+
+	for (guint i = 0; tables[i] != NULL; i++) {
+		g_autofree gchar *fn = g_build_filename(path, tables[i], NULL);
+		g_autoptr(GBytes) blob = NULL;
+
+		blob = fu_bytes_get_contents(fn, NULL);
+		if (blob != NULL && g_bytes_get_size(blob) > 0) {
+			g_autofree gchar *id = g_strdup_printf("ACPI:%s", tables[i]);
+			fu_engine_integrity_add_measurement(self, id, blob);
+		}
+	}
+}
+
+static void
+fu_engine_integrity_measure_uefi(GHashTable *self)
+{
+	struct {
+		const gchar *guid;
+		const gchar *name;
+	} keys[] = {{FU_EFIVAR_GUID_EFI_GLOBAL, "BootOrder"},
+		    {FU_EFIVAR_GUID_EFI_GLOBAL, "BootCurrent"},
+		    {FU_EFIVAR_GUID_EFI_GLOBAL, "KEK"},
+		    {FU_EFIVAR_GUID_EFI_GLOBAL, "PK"},
+		    {FU_EFIVAR_GUID_SECURITY_DATABASE, "db"},
+		    {FU_EFIVAR_GUID_SECURITY_DATABASE, "dbx"},
+		    {NULL, NULL}};
+
+	/* important keys */
+	for (guint i = 0; keys[i].guid != NULL; i++) {
+		g_autoptr(GBytes) blob =
+		    fu_efivar_get_data_bytes(keys[i].guid, keys[i].name, NULL, NULL);
+		if (blob != NULL) {
+			g_autofree gchar *id = g_strdup_printf("UEFI:%s", keys[i].name);
+			fu_engine_integrity_add_measurement(self, id, blob);
+		}
+	}
+
+	/* Boot#### */
+	for (guint i = 0; i < 0xFF; i++) {
+		g_autofree gchar *name = g_strdup_printf("Boot%04X", i);
+		g_autoptr(GBytes) blob =
+		    fu_efivar_get_data_bytes(FU_EFIVAR_GUID_EFI_GLOBAL, name, NULL, NULL);
+		if (blob != NULL && g_bytes_get_size(blob) > 0) {
+			g_autofree gchar *id = g_strdup_printf("UEFI:%s", name);
+			fu_engine_integrity_add_measurement(self, id, blob);
+		}
+	}
+}
+
+GHashTable *
+fu_engine_integrity_new(GError **error)
+{
+	g_autoptr(GHashTable) self = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+
+	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
+
+	fu_engine_integrity_measure_uefi(self);
+	fu_engine_integrity_measure_acpi(self);
+
+	/* nothing of use */
+	if (g_hash_table_size(self) == 0) {
+		g_set_error_literal(error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND, "no measurements");
+		return NULL;
+	}
+
+	/* success */
+	return g_steal_pointer(&self);
+}
+
+gchar *
+fu_engine_integrity_to_string(GHashTable *self)
+{
+	GHashTableIter iter;
+	gpointer key, value;
+	g_autoptr(GPtrArray) array = g_ptr_array_new_with_free_func(g_free);
+
+	g_return_val_if_fail(self != NULL, NULL);
+
+	/* sanity check */
+	if (g_hash_table_size(self) == 0)
+		return NULL;
+
+	/* build into KV array */
+	g_hash_table_iter_init(&iter, self);
+	while (g_hash_table_iter_next(&iter, &key, &value)) {
+		g_ptr_array_add(array,
+				g_strdup_printf("%s=%s", (const gchar *)key, (const gchar *)value));
+	}
+	return fu_strjoin("\n", array);
 }
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2405,6 +2405,16 @@ fu_engine_install_releases(FuEngine *self,
 	return TRUE;
 }
 
+static void
+fu_engine_update_release_integrity(FuEngine *self, FwupdRelease *release, const gchar *key)
+{
+	g_autoptr(GHashTable) integrity = fu_engine_integrity_new(NULL);
+	if (integrity != NULL) {
+		g_autofree gchar *str = fu_engine_integrity_to_string(integrity);
+		fwupd_release_add_metadata_item(FWUPD_RELEASE(release), key, str);
+	}
+}
+
 static gboolean
 fu_engine_add_release_metadata(FuEngine *self, FuRelease *release, FuPlugin *plugin, GError **error)
 {
@@ -2447,6 +2457,14 @@ fu_engine_add_release_metadata(FuEngine *self, FuRelease *release, FuPlugin *plu
 			}
 		}
 	}
+
+	/* measure the "old" system state */
+	if (fu_plugin_has_flag(plugin, FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY)) {
+		fu_engine_update_release_integrity(self,
+						   FWUPD_RELEASE(release),
+						   "SystemIntegrityOld");
+	}
+
 	return TRUE;
 }
 
@@ -7307,6 +7325,13 @@ fu_engine_update_history_device(FuEngine *self, FuDevice *dev_history, GError **
 		}
 	}
 
+	/* measure the "new" system state */
+	plugin = fu_plugin_list_find_by_name(self->plugin_list, fu_device_get_plugin(dev), error);
+	if (plugin == NULL)
+		return FALSE;
+	if (fu_plugin_has_flag(plugin, FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY))
+		fu_engine_update_release_integrity(self, rel_history, "SystemIntegrityNew");
+
 	/* the system is running with the new firmware version */
 	if (fu_version_compare(fu_device_get_version(dev),
 			       fwupd_release_get_version(rel_history),
@@ -7330,9 +7355,6 @@ fu_engine_update_history_device(FuEngine *self, FuDevice *dev_history, GError **
 	}
 
 	/* does the plugin know the update failure */
-	plugin = fu_plugin_list_find_by_name(self->plugin_list, fu_device_get_plugin(dev), error);
-	if (plugin == NULL)
-		return FALSE;
 	if (!fu_plugin_runner_get_results(plugin, dev, error))
 		return FALSE;
 
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -1702,6 +1702,10 @@ fu_util_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		/* TRANSLATORS: the plugin was created from a .so object, and was not built-in */
 		return _("Loaded from an external module");
 	}
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY) {
+		/* TRANSLATORS: check various UEFI and ACPI tables are unchanged after the update */
+		return _("Will measure elements of system integrity around an update");
+	}
 	if (plugin_flag == FWUPD_PLUGIN_FLAG_EFIVAR_NOT_MOUNTED) {
 		/* TRANSLATORS: the user is using Gentoo/Arch and has screwed something up */
 		return _("Required efivarfs filesystem was not found");
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -448,6 +448,8 @@ fwupd_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		return "secure-config";
 	if (plugin_flag == FWUPD_PLUGIN_FLAG_MODULAR)
 		return "modular";
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY)
+		return "measure-system-integrity";
 	return NULL;
 }
 
@@ -496,6 +498,8 @@ fwupd_plugin_flag_from_string(const gchar *plugin_flag)
 		return FWUPD_PLUGIN_FLAG_SECURE_CONFIG;
 	if (g_strcmp0(plugin_flag, "modular") == 0)
 		return FWUPD_PLUGIN_FLAG_MODULAR;
+	if (g_strcmp0(plugin_flag, "measure-system-integrity") == 0)
+		return FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY;
 	return FWUPD_DEVICE_FLAG_UNKNOWN;
 }
 
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -880,6 +880,14 @@ typedef enum {
  */
 #define FWUPD_PLUGIN_FLAG_MODULAR (1u << 14)
 /**
+ * FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY:
+ *
+ * The plugin will be checked that it preserves system state such as `KEK`, `PK`, `BOOT####` etc.
+ *
+ * Since: 1.8.7
+ */
+#define FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY (1u << 15)
+/**
  * FWUPD_PLUGIN_FLAG_UNKNOWN:
  *
  * The plugin flag is Unknown.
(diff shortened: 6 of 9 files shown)


src/fu-engine-integrity.c
```diff
@@ -0,0 +1,212 @@
...
+void
+fu_engine_integrity_add_checksum(GHashTable *self, const gchar *id, const gchar *csum)
+{
+	g_hash_table_insert(self, g_strdup(id), g_strdup(csum));
+}
+
+static void
+fu_engine_integrity_add_measurement(GHashTable *self, const gchar *id, GBytes *blob)
+{
+	g_autofree gchar *csum = g_compute_checksum_for_bytes(G_CHECKSUM_SHA256, blob);
```
superm1, line 23: If we're starting from scratch, any reason to pick SHA256 rather than 384?

src/fu-engine-integrity.c
```diff
@@ -0,0 +1,212 @@
...
+	for (guint i = 0; keys[i].guid != NULL; i++) {
+		g_autoptr(GBytes) blob =
+		    fu_efivar_get_data_bytes(keys[i].guid, keys[i].name, NULL, NULL);
+		if (blob != NULL) {
+			g_autofree gchar *id = g_strdup_printf("UEFI:%s", keys[i].name);
+			fu_engine_integrity_add_measurement(self, id, blob);
+		}
+	}
+
+	/* Boot#### */
```
superm1, line 69: `BootOrder` too maybe?  Not sure.

src/fu-engine-integrity.c
```diff
@@ -0,0 +1,212 @@
...
+	struct {
+		const gchar *guid;
+		const gchar *name;
+	} keys[] = {{FU_EFIVAR_GUID_EFI_GLOBAL, "BootOrder"},
+		    {FU_EFIVAR_GUID_EFI_GLOBAL, "BootCurrent"},
+		    {FU_EFIVAR_GUID_EFI_GLOBAL, "KEK"},
+		    {FU_EFIVAR_GUID_EFI_GLOBAL, "PK"},
+		    {FU_EFIVAR_GUID_SECURITY_DATABASE, "db"},
+		    {FU_EFIVAR_GUID_SECURITY_DATABASE, "dbx"},
+		    {NULL, NULL}};
```
superm1, line 57: How about having the plugin do the measurement?  That way it's only measure when you do uefi-capsule updates.

src/fu-engine-integrity.c
```diff
@@ -0,0 +1,212 @@
...
+{
+	g_autofree gchar *csum = g_compute_checksum_for_bytes(G_CHECKSUM_SHA256, blob);
+	fu_engine_integrity_add_checksum(self, id, csum);
+}
+
+static void
+fu_engine_integrity_measure_acpi(GHashTable *self)
+{
+	g_autofree gchar *path = fu_path_from_kind(FU_PATH_KIND_ACPI_TABLES);
+	const gchar *tables[] = {"SLIC", NULL};
```
superm1, line 31: You want MSDM and TPM2 as well I think.

src/fu-util-common.c
```diff
@@ -1702,6 +1702,10 @@ fu_util_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		/* TRANSLATORS: the plugin was created from a .so object, and was not built-in */
 		return _("Loaded from an external module");
 	}
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_VERIFY_SYSTEM_INTEGRITY) {
+		/* TRANSLATORS: the plugin was created from a .so object, and was not built-in */
```
superm1, line 1706: wrong translator message

src/fu-engine-integrity.c
hughsie: I don't think it matters; I think MD5 would have been fine too. SHA384 would mean upping our gio dep to 2.51 and SHA1 means I have to explain why I'm using it to Red Hat security.

src/fu-engine-integrity.c
hughsie: Oh hah, BootOrder is already there.

src/fu-engine-integrity.c
hughsie: Added.

src/fu-util-common.c
hughsie: Fixed.

src/fu-engine-integrity.c
hughsie: I think we probably ought to do this for other plugins too -- e.g. flashrom.

superm1: Then maybe should it just be a libfwupdplugin helper that those plugins that change it can call?  I guess I see it either way.

src/fu-engine-integrity.c
superm1: OK

src/fu-engine-integrity.c
hughsie: My logic was that the daemon should probably define what "system integrity" was -- rather than the plug-ins deciding for themselves. I guess the PK and KEK could be set from the uefi-pk plugin, but I couldn't split the others up into anywhere sane without adding a lot more wiring -- and I couldn't see what else I'd want to add -- e.g. TPM measurements are not relevant here I think.

superm1: OK, I guess I'm fine with this.  So basically plugins say they might influence integrity data, and daemon measures to see if data changed.

hughsie: Yup; I'm not against polling the plugins for this data, I just don't think we need the extra complexity. If that changes in the future, I'm happy to add it.

hughsie:
@CragW @mrhpearson any worries from your sides?

bluca:
MSFT's KEK is about to expire, and I'm told some vendors do not have access to their PK anymore, which means the new KEK will likely have to come down via firmware updates, since on those platforms a signed KEK update is no longer possible. How will this change affect that workflow?

hughsie:
> the new KEK will likely have to come down via firmware updates

This is my worry too. I wonder if we should allow a PK change from say `Lenovo Ltd. PK CA 2012` to `Lenovo Ltd. PK CA 2022` -- but that's very hard to codify without actually seeing the new cert first.

hughsie:
@superm1 -- maybe we should remove TPM2?

Screenshot 2022-10-21 at 10-37-12 Morten Linderud on Twitter

Foxboron:
Some vendors ship faulty TPM2 ACPI tables and patch these after shipping. This has happened to at least the Framework laptops and I have some guesses that Lenovo has made the same mistake before.

https://lore.kernel.org/linux-integrity/20220608123109.678343-1-morten@linderud.pw/T/#u

Foxboron:
Ohno, beaten by my own tweet :)

Foxboron:
You could include custom ACPI tables during boot, either for debugging or local patching. How would `fwupdmgr` react if that was the case?

hughsie:
> How would fwupdmgr react if that was the case

The update would "fail". Could we detect the user using custom ACPI tables?

Foxboron:
I'm not sure sadly. The documentation on how to do it is here: https://www.kernel.org/doc/html/latest/admin-guide/acpi/initrd_table_override.html

I see the appopriate kernel config is enabled for Arch Linux. But I'm not sure how often this is done or if it is a legitimate concern.

superm1:
> I'm not sure sadly. The documentation on how to do it is here: https://www.kernel.org/doc/html/latest/admin-guide/acpi/initrd_table_override.html
> 
> I see the appopriate kernel config is enabled for Arch Linux. But I'm not sure how often this is done or if it is a legitimate concern.

It can be done via grub too. When done this way you have no custom kernel command line.

mrhpearson:
From my point of view - if it catches when we have issues like raised in the original feature request it's all good. I'll have to ask the FW team for their opinion on any unexpected side-effects (as flagged in the discussion above). Just to confirm that you're checking BootOrder, BootCurrent, DB, DBX, PK and KEK 

There's a few corporate customers who do their own key handling that I'd like to run this by too if that's OK - afraid it might take a bit of time to get answers though.

Just to check my reading of the code - if something unexpected changes that it doesn't auto-fix and override changes, but just flags it as an error that will get reported to our FW teams for review?

Mark

hughsie:
> Just to confirm that you're checking BootOrder, BootCurrent, DB, DBX, PK and KEK

And the ACPI tables `SLIC` and `MSDM`, and *all* the Boot#### keys.

> There's a few corporate customers who do their own key handling

To be honest I think that's the most common failure we're going to see; a firmware update that "resets" the corporate-set PK back to the Microsoft one.

> If something unexpected changes that it doesn't auto-fix and override changes, but just flags it as an error that will get reported to our FW teams for review?

Exactly that. My worry is that it breaks if Lenovo updates the Microsoft PK from the 2012 to 2022 one in a firmware update -- although I don't know if doing that is a very good idea in existing devices...

bluca:
> > If something unexpected changes that it doesn't auto-fix and override changes, but just flags it as an error that will get reported to our FW teams for review?
> 
> Exactly that. My worry is that it breaks if Lenovo updates the Microsoft PK from the 2012 to 2022 one in a firmware update -- although I don't know if doing that is a very good idea in existing devices...

I think this is exactly what is planned for the KEK update for many vendors for existing devices, so as a use case it needs to be supported, I think. I will try and confirm with the UEFI/SB team at Microsoft.

superm1:
> To be honest I think that's the most common failure we're going to see; a firmware update that "resets" the corporate-set PK back to the Microsoft one.

To be clear - the idea here was not explicitly to make the failures and demotions automatic by any fwupd client, but rather let fwupd gather the data and report it to LVFS and then let the policy decisions be on the LVFS side.

On the LVFS policy side some of the ideas:

1. Don't allow promotion from testing to stable without a report showing that the measured data was the same.  If it's intended to change (IE KEK rollover) then allow the ability to waive that.
2. "1" but only for the first time a new system is uploaded to LVFS.
3. If X failures, send emails to people to let them investigate.
4. Automatic demotion after Y failures.

superm1:
> > Just to confirm that you're checking BootOrder, BootCurrent, DB, DBX, PK and KEK
> 
> And the ACPI tables `SLIC` and `MSDM`, and _all_ the Boot#### keys.
> 
> > There's a few corporate customers who do their own key handling
> 
> To be honest I think that's the most common failure we're going to see; a firmware update that "resets" the corporate-set PK back to the Microsoft one.

Mmm - you mean OEM-set PK right?  Microsoft doesn't set the PK.

das-menschy:
I think we have to differentiate: 
- SLIC and MSDM probably should not change, because then the Microsoft Product License Key would get lost. 
- TPM2 can change, to fix TPM problems, as MortenLinderud said. 
- Boot#### should only change if it is concerning the firmware-provided EFI boot options (like Boot menu, UEFI Setup, Recovery, ...); NOT the operating system itself. In my case: 
```
[root@linux-on-20hr002mmx ~]$ efibootmgr 
[sudo] password for conbold: 
BootCurrent: 0001
Timeout: 0 seconds
BootOrder: 0001,0002,0017,0018,0019,001A,001B,001C,001D,001E,0000
Boot0000* Linux-Firmware-Updater	HD(1,GPT,88a96d13-381e-4dc1-8eea-9ac7f7fb5bc3,0x800,0x800000)/File(\EFI\arch\fwupdx64.efi)
Boot0001* Linux Boot Manager	HD(1,GPT,88a96d13-381e-4dc1-8eea-9ac7f7fb5bc3,0x800,0x800000)/File(\EFI\systemd\systemd-bootx64.efi)
Boot0002* Windows Boot Manager	HD(1,GPT,88a96d13-381e-4dc1-8eea-9ac7f7fb5bc3,0x800,0x800000)/File(\EFI\Microsoft\Boot\bootmgfw.efi)
Boot0010  Setup	FvFile(721c8b66-426c-4e86-8e99-3457c46ab0b9)
Boot0011  Boot Menu	FvFile(126a762d-5758-4fca-8531-201a7f57f850)
Boot0012  Diagnostic Splash Screen	FvFile(a7d8d9a6-6ab0-4aeb-ad9d-163e59a7a380)
Boot0013  Lenovo Diagnostics	FvFile(3f7e615b-0d45-4f80-88dc-26b234958560)
Boot0014  Startup Interrupt Menu	FvFile(f46ee6f4-4785-43a3-923d-7f786c3c8479)
Boot0015  Rescue and Recovery	FvFile(665d3f60-ad3e-4cad-8e26-db46eee9f1b5)
Boot0016  MEBx Hot Key	FvFile(ac6fd56a-3d41-4efd-a1b9-870293811a28)
Boot0017* USB CD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,86701296aa5a7848b66cd49dd3ba6a55)
Boot0018* USB FDD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,6ff015a28830b543a8b8641009461e49)
Boot0019* NVMe0	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,001c199932d94c4eae9aa0b6e98eb8a400)
Boot001A* ATA HDD0	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,91af625956449f41a7b91f4f892ab0f600)
Boot001B* USB HDD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,33e821aaaf33bc4789bd419f88c50803)
Boot001C* PCI LAN	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,78a84aaf2b2afc4ea79cf5cc8f3d3803)
Boot001D  Other CD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,aea2090adfde214e8b3a5e471856a35406)
Boot001E  Other HDD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,91af625956449f41a7b91f4f892ab0f606)
Boot001F* IDER BOOT CDROM	PciRoot(0x0)/Pci(0x16,0x2)/Ata(0,1,0)
Boot0020* IDER BOOT Floppy	PciRoot(0x0)/Pci(0x16,0x2)/Ata(0,0,0)
Boot0021* ATA HDD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,91af625956449f41a7b91f4f892ab0f6)
Boot0022* ATAPI CD	VenMsg(bc7838d2-0f82-4d60-8316-c068ee79d25b,aea2090adfde214e8b3a5e471856a354)
```
... only the Boot#### entries after Boot0010 should change (so Boot0010, Boot0011, Boot0012, Boot0013, etc.); not the ones before (Boot0000, Boot0001, Boot0002, ...). But these numbers are different for every firmware; so I guess we can't make a general rule. 
- Secure Boot certs should only be changed if the manufacturer/OEM-provided Platform key is in use ("Standard Mode"). If a custom, personal Platform key is in use ("Custom Mode"), then the PK, KEK, DB keys should probably not be changed. In the firmware on my Thinkpad laptop, the difference can be seen from these two pictures - the first one is from before enrolling my personal Secure Boot keys, the second one is from after: 
https://github.com/das-menschy/ThinkpadX1Carbon5thGen-CustomSecureBootSetup/blob/main/10_DISABLE_SECURE_BOOT_ON_NEW_PC/BIOS_Setup_-_04_Security_-_10_Secure_Boot_Configuration_-_01_Platform_Mode_User_Standard_mode%2C_Secure_Boot_Enabled_-_01_Secure_Boot_Enabled_-_00.png
https://github.com/das-menschy/ThinkpadX1Carbon5thGen-CustomSecureBootSetup/blob/main/20_ENABLE_SECURE_BOOT_AGAIN_ON_NEW_PC/BIOS_Setup_-_04_Security_-_10_Secure_Boot_Configuration_-_04_Platform_Mode_User_Custom_mode%2C_Secure_Boot_Enabled_-_01_Secure_Boot_Enabled.png . 
But I don't know how we can detect the kind of Secure Boot mode...

baloo:
>     * Secure Boot certs should only be changed if the manufacturer/OEM-provided Platform key is in use (`Standard Mode`). If a custom, personal Platform key is in use (`Custom Mode`), then the `PK`, `KEK`, `DB` keys should probably not be changed. In the firmware on my Thinkpad laptop, the difference can be seen from these two pictures - the first one is from before enrolling my personal Secure Boot keys, the second one is from after:

It could still be worth raising a flag if the PK/KEK/DB are changed in the non-custom mode. As those are factored in PCR7 and secrets are often tied to it on the TPM (even if that's indirectly via a TPM2 enhanced policy).

mrhpearson:
No feedback from the FW team yet, but a few notes/questions from some other feelers I sent out:

 - The check would be done over a FW update only right? e.g. measure/update/measure/check. Or is the scope wider than that?I really want to avoid false positives - I have enough of a fight getting FW teams to put FW on LVFS in a timely manner without having to deal with churn if FW gets bounced back and forth between states due to false positives.
 - Believe DBX shouldn't be included as that does deliberately get updated by FW update (I think we released a round of this just recently)
 - Can there be a flag to disable the checking? Some customers will be doing debug/setup and wouldn't want to trigger this.

Mark

hughsie:
> The check would be done over a FW update only right? 

Correct.

> Believe DBX shouldn't be included as that does deliberately get updated by FW update

EEEK! IIRC @vathpela told all OEMs "_don't do that_" as if you do it without checking all the things in the ESP then you risk making the machine unbootable. I think it's unwise to update the dbx on any system that's already deployed to the field.

> Can there be a flag to disable the checking?

Ya, I need to rework all this PR to push the data to the LVFS where the uploader can just waive the failure as expected.

superm1:
> But I don't know how we can detect the kind of Secure Boot mode...

Perhaps a heuristic is good enough here.  Can we just look at who is in the KEK?  If it's not Microsoft it's "probably" custom mode.

> EEEK! IIRC @vathpela told all OEMs "don't do that" as if you do it without checking all the things in the ESP then you risk making the machine unbootable. I think it's unwise to update the dbx on any system that's already deployed to the field.

+1

hughsie:
I've made this a lot simpler, and how we just record the system state before the update, and after the update, and then pass it all to the LVFS to actually compare it. The LVFS will be able to process this data and fail the (waivable) test if the data is changed. The data-collection bit is needed first, which is this PR.

src/fu-engine-helper.c
```diff
@@ -187,3 +188,106 @@ fu_engine_update_devices_file(FuEngine *self, GError **error)
...
+
+	for (guint i = 0; tables[i] != NULL; i++) {
+		g_autofree gchar *fn = g_build_filename(path, tables[i], NULL);
+		g_autoptr(GBytes) blob = NULL;
+
+		blob = fu_bytes_get_contents(fn, NULL);
+		if (blob != NULL && g_bytes_get_size(blob) > 0) {
+			g_autofree gchar *id = g_strdup_printf("ACPI:%s", tables[i]);
+			fu_engine_integrity_add_measurement(self, id, blob);
+		}
```
superm1, line 213: There is a problem here.  If one of these is present before and disappears the measurement is gone.  I think you want this measurement for each of these even if not present so that it's explicitly clear it was missing.

src/fu-engine-helper.c
```diff
@@ -187,3 +188,106 @@ fu_engine_update_devices_file(FuEngine *self, GError **error)
...
+	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
+
+	fu_engine_integrity_measure_uefi(self);
+	fu_engine_integrity_measure_acpi(self);
+
+	/* nothing of use */
+	if (g_hash_table_size(self) == 0) {
+		g_set_error_literal(error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND, "no measurements");
+		return NULL;
+	}
```
superm1, line 267: if you follow my other suggestion this block won't matter since you'll always have a measurement for ACPI entries even if not present.

src/fu-engine.c
```diff
@@ -2447,6 +2447,18 @@ fu_engine_add_release_metadata(FuEngine *self, FuRelease *release, FuPlugin *plu
 			}
 		}
 	}
+
+	/* ensure system state stays the same */
```
superm1, line 2451: this comment should change; it's just measuring now, not verifying.

src/fu-engine.c
```diff
@@ -7307,6 +7320,18 @@ fu_engine_update_history_device(FuEngine *self, FuDevice *dev_history, GError **
 		}
 	}
 
+	/* check nothing changed */
```
superm1, line 7323: this isn't checking nothing changed, it's just re-measuring.

src/fu-engine.c
```diff
@@ -7307,6 +7320,18 @@ fu_engine_update_history_device(FuEngine *self, FuDevice *dev_history, GError **
...
+	if (integrity_str != NULL) {
+		g_autoptr(GHashTable) integrity = fu_engine_integrity_new(NULL);
+		if (integrity != NULL) {
+			g_autofree gchar *str = fu_engine_integrity_to_string(integrity);
+			fwupd_release_add_metadata_item(FWUPD_RELEASE(rel_history),
+							"SystemIntegrityNew",
+							str);
+		}
+	}
+
```
superm1, line 7334: I would argue that you should include the data for new even if old wasn't present.  It's more data points that LVFS can sift through on a given platform.

src/fu-util-common.c
```diff
@@ -1702,6 +1702,10 @@ fu_util_plugin_flag_to_string(FwupdPluginFlags plugin_flag)
 		/* TRANSLATORS: the plugin was created from a .so object, and was not built-in */
 		return _("Loaded from an external module");
 	}
+	if (plugin_flag == FWUPD_PLUGIN_FLAG_VERIFY_SYSTEM_INTEGRITY) {
+		/* TRANSLATORS: check various UEFI and ACPI tables are unchanged after the update */
+		return _("Will verify system integrity after update");
```
superm1, line 1707: Will capture elements of system integrity around an update.

src/fu-engine-helper.c
superm1: Will you detect that they're added too from the same code?

hughsie: Yup, it's a straight port of the compare code I had before to python.

superm1: OK that's fine to me then, thanks.

src/fu-engine.c
hughsie: Fixed

src/fu-engine.c
hughsie: Fixed.

src/fu-util-common.c
hughsie: How about `FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY` too?

hughsie: I did that :)

plugins/uefi-capsule/fu-uefi-capsule-plugin.c
```diff
@@ -997,6 +997,7 @@ static void
 fu_uefi_capsule_plugin_init(FuUefiCapsulePlugin *self)
 {
 	self->bgrt = fu_uefi_bgrt_new();
+	fu_plugin_add_flag(FU_PLUGIN(self), FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY);
```
superm1, line 1000: I think you also should add this flag to coreboot plugin.

superm1 (approved):
just two nits; lgtm otherwise

plugins/uefi-capsule/fu-uefi-capsule-plugin.c
hughsie: Good idea, done.

src/fu-engine.c
```diff
@@ -7307,6 +7319,20 @@ fu_engine_update_history_device(FuEngine *self, FuDevice *dev_history, GError **
...
+		return FALSE;
+	if (fu_plugin_has_flag(plugin, FWUPD_PLUGIN_FLAG_MEASURE_SYSTEM_INTEGRITY)) {
+		g_autoptr(GHashTable) integrity = fu_engine_integrity_new(NULL);
+		if (integrity != NULL) {
+			g_autofree gchar *str = fu_engine_integrity_to_string(integrity);
+			fwupd_release_add_metadata_item(FWUPD_RELEASE(rel_history),
+							"SystemIntegrityNew",
+							str);
+		}
+	}
```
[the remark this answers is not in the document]
hughsie: Done.

PR #5182, opened 2022-10-21
Make the HSI specification machine readable

hughsie:
This allows us to convert it to other forms, e.g. OPAL.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation


hughsie:
@superm1 if you think this makes sense I'll convert the others.

superm1:
Sure seems fine.

docs/meson.build
```diff
@@ -43,6 +43,17 @@ if gidocgen_dep.found() and docs_python_deps.allowed() and gidocgen_app.found()
     install_dir: join_paths(datadir, 'doc'),
   )
 
+  hsi_md = custom_target('generate-hsi-spec',
+    input : 'hsi.md',
+    output : 'hsi.md',
+    command : [
+      join_paths(meson.current_source_dir(), 'generate-hsi-spec.py'),
```
superm1, line 50: I think you shouldn't assume it's executable and use the default interpreter from the shebang.  You should use python as found by meson to execute it

hughsie: Got it, thanks. Fixed.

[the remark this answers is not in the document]
superm1: Yeah, but does the correct ordering work?  Does hsi.md get made before the docs get made?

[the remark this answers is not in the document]
hughsie: Jup, I added it in `depends`.

hughsie:
@smallorange does this give you enough data for OVAL? Is there anything missing you can think of?

docs/meson.build
```diff
@@ -43,6 +43,18 @@ if gidocgen_dep.found() and docs_python_deps.allowed() and gidocgen_app.found()
     install_dir: join_paths(datadir, 'doc'),
   )
 
+  hsi_md = custom_target('generate-hsi-spec',
+    input : 'hsi.md',
```
superm1, line 47: I know it's not required with meson, but I think it makes sense to rename this `hsi.md.in` so that this doesn't render in github U/I anymore.  Since it's missing all the chunks of data it's not very useful in this form.

Another idea is to have a comment that is here and removed by the python script:

"Some information in this specification is dynamically built.  Visit https://fwupd.github.io/libfwupdplugin/hsi.html for the complete specification."

hughsie: I think `hsi.md.in` makes sense. Fixed.

docs/meson.build
eli-schwartz: Truthfully, since this code doesn't seem to care what version of python it runs using, you could specify the shebang as `/usr/bin/env python3`.

Random aside: "use python as found by meson" -- if you ***remove*** the executable bit from the script, Meson notices that it cannot be run as a command, so it reads the shebang itself and runs the interpreter with the script as an argument. For scripts with a shebang ending in python3, it then always uses Meson's own internal python (might be /usr/bin/python3, might be `meson.exe runpython` on Windows).

superm1 (approved):

PR #5192, opened 2022-10-24
usi-dock: set correct chip type before update.

victor-cheng:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -707,6 +708,30 @@ fu_usi_dock_mcu_device_write_firmware(FuDevice *device,
...
+fu_usi_dock_mcu_device_prepare(FuDevice *device,
+			       FuProgress *progress,
+			       FwupdInstallFlags flags,
+			       GError **error)
+{
+	FuUsiDockMcuDevice *self = FU_USI_DOCK_MCU_DEVICE(device);
+	guint8 inbuf[] = {USBUID_ISP_DEVICE_CMD_SET_CHIP_TYPE, 1, 1};
+
+	if (fu_device_has_guid(device, USI_DOCK_NON_IOT_INSTANCE_ID) &&
+	    g_strcmp0(fu_device_get_version(device), "10.10") == 0) {
```
hughsie, line 721: is this only one model and version combination, or could there be more in the future? I'm wondering if this is better suited as a private flag.

superm1: That's why I was looking for a comment for more context too..

victor-cheng: Only such a combination requires additional commands to avoid some problems after the update.

hughsie: Okay, lets merge this, but if we need any more we can move it to a quirk.

hughsie (approved):

PR #5195, opened 2022-10-25
elantp: solve the problem that I2C device cannot be matched

josh-chen-elan:
revert some files of faa7871

After fu-i2c-device.c is reverted, I2C\MODALIAS_* of elantp.quirk can be matched.
After fu-elantp-i2c-device.c is reverted, the elantp plugin can correctly read the value of the "ElantpI2cTargetAddress" key.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-i2c-device.c
+++ b/libfwupdplugin/fu-i2c-device.c
@@ -118,6 +118,10 @@ fu_i2c_device_probe(FuDevice *device, GError **error)
 	g_autoptr(GUdevDevice) udev_parent = NULL;
 #endif
 
+	/* FuUdevDevice->probe */
+	if (!FU_DEVICE_CLASS(fu_i2c_device_parent_class)->probe(device, error))
+		return FALSE;
+
 	/* set physical ID */
 	if (!fu_udev_device_set_physical_id(FU_UDEV_DEVICE(device), "i2c", error))
 		return FALSE;
--- a/plugins/elantp/fu-elantp-i2c-device.c
+++ b/plugins/elantp/fu-elantp-i2c-device.c
@@ -110,6 +110,10 @@ fu_elantp_i2c_device_probe(FuDevice *device, GError **error)
 {
 	FuElantpI2cDevice *self = FU_ELANTP_I2C_DEVICE(device);
 
+	/* FuUdevDevice->probe */
+	if (!FU_DEVICE_CLASS(fu_elantp_i2c_device_parent_class)->probe(device, error))
+		return FALSE;
+
 	/* check is valid */
 	if (g_strcmp0(fu_udev_device_get_subsystem(FU_UDEV_DEVICE(device)), "i2c") == 0) {
 		g_autoptr(GPtrArray) i2c_buses = NULL;


hughsie:
Hmm, I'm sorry for causing this; I sent the laptop I used for testing back to Jingle Wu, so I have no way to test this device now. Out of interest, can you get me the `sudo FU_UDEV_DEVICE_DEBUG=1 fwupdtool --plugins elantp --verbose --no-timestamp get-devices` logs from before the fix and after the fix please.

josh-chen-elan:
This issue only occurs with ABS touchapd on chromebooks, not with the HID touchapd you tested earlier.
The following are the log files before and after the fix.
[before.txt](https://github.com/fwupd/fwupd/files/9858784/before.txt)
[after.txt](https://github.com/fwupd/fwupd/files/9858787/after.txt)

If there is a better suggestion to fix, please let me know.

hughsie:
That's super weird. Just to debug this a bit more (I can't explain the failure, and I'd like to understand the issue rather than papering over the issue) could you please add something like:

    g_debug("BEFORE: %s", fu_device_to_string(device));
    if (!FU_DEVICE_CLASS(fu_i2c_device_parent_class...
    g_debug("AFTER: %s", fu_device_to_string(device));

...and...

the same thing for `FU_DEVICE_CLASS(fu_elantp_i2c_device_parent_class)->probe` -- and then grab the `sudo FU_UDEV_DEVICE_DEBUG=1 fwupdtool --plugins elantp --verbose --no-timestamp get-devices` output again.

Something is failing to be copied using incorporate and I'm just a bit confused. Thanks!

hughsie:
> Something is failing to be copied using incorporate and I'm just a bit confused. Thanks!

Actually less confused now. I'd be interested if https://github.com/fwupd/fwupd/pull/5198 fixes your issue instead of your fix.

josh-chen-elan:
Thanks for the suggestion.

I add some debug logs, the `g_debug("AFTER: %s", fu_device_to_string(device));` will be printed when `FU_DEVICE_CLASS(fu_XXXX_parent_class)->probe` fails. It seems that the probe() are executed normally,since I can't see these messages in the log file [debug.txt](https://github.com/fwupd/fwupd/files/9865400/debug.txt)

Then I removed these reverts, and added the fixes #5198, the MODALIAS from elantp.quirk is also not being matched from the log file [fix.txt](https://github.com/fwupd/fwupd/files/9865430/fix.txt)

hughsie:
> the MODALIAS from elantp.quirk is also not being matched from the log

I think the fixes in https://github.com/fwupd/fwupd/pull/5198 make the first chunk of your fix unnecessary, but as with all good bugs this one seems to be two different things intersecting. I've pushed a `wip/hughsie/elan-debugging` branch; can you get me a debug trace of `sudo FU_UDEV_DEVICE_DEBUG=1 fwupdtool --plugins elantp --verbose --no-timestamp get-devices` from that branch please? It should enumerate fine, but it's the BEFORE and AFTER text I'm after.

josh-chen-elan:
After porting `wip/hughsie/elan-debugging` to ChromeOS, the result is in file [elan-debugging.txt](https://github.com/fwupd/fwupd/files/9874908/elan-debugging.txt).
Without executing `FU_DEVICE_CLASS(fu_elantp_i2c_device_parent_class)->probe`, the result is the same as the previous fix.txt, the MODALIAS from elantp.quirk is also not being matched.
BTW, I found that `fu_i2c_device_probe()` will build instance ID `I2C\NAME_ELAN0000:00`, should I change the matching method from `MODALIAS` to `NAME`?
Many thanks.

hughsie:
No, don't use `NAME_` as FuElantpI2cDevice derives from FuUdevDevice, not FuI2cDevice. It seems we're not getting to `fu_elantp_i2c_device_probe()` at all.

So we can see what is detected with the (updated) `wip/hughsie/elan-debugging` branch, could you please do:

    sudo FU_UDEV_DEVICE_DEBUG=1 FWUPD_PROBE_VERBOSE=1 fwupdtool --plugins elantp --verbose --no-timestamp get-devices

...that should be a lot more verbose. It'll dump to the console the FuUdevDevice we build for quirk matching. I think we're aiming for `I2C\MODALIAS_of:NtrackpadT(null)Celan,ekth3000` -- correct? I think we are matching as I see `already set GType to FuElantpI2cDevice, ignoring FuElantpI2cDevice` which seems to suggest we hit the `GType = FuElantpI2cDevice` bit. So then I'm confused why we don't hit the `g_debug("BEFORE")`

josh-chen-elan:
I'm testing on an x86 (ACPI) platform, so it should match `[I2C\MODALIAS_acpi:ELAN0000:]`,
but I can't find this instance ID in the latest log file [elan-debugging2.txt](https://github.com/fwupd/fwupd/files/9877471/elan-debugging2.txt), maybe that's why `fu_elantp_i2c_device_probe()` is not called.

hughsie:
> I'm testing on an x86 (ACPI) platform

I thought you were using an ABS touchpad on a chromebook. That's not going to match `I2C\MODALIAS_of:NtrackpadT(null)Celan,ekth3000` -- are you actually trying to match `I2C\NAME_Synopsys-DesignWare-I2C-adapter` instead? If so you need to derive `FuElantpI2cDevice` from `FuI2cDevice` instead of `FuUdevDevice`.

josh-chen-elan:
Sorry to misunderstand you.
I'm using the ABS touchpad on a chromebook, but it has two different platforms, **x86 (ACPI)** and **ARM (OF)**. On the x86 platform I expect to use `I2C\MODALIAS_acpi:ELAN0000:` to match, on the ARM platform it will be `I2C\MODALIAS_of:NtrackpadT(null)Celan,ekth3000`.

On the x86 platform, it is also possible to use the `I2C\NAME_Synopsys-DesignWare-I2C-adapter` match, but it is definitely suitable for the ARM platform. So I tend to use a unified approach (MODALIAS) to match chromebooks.

superm1:
> Sorry to misunderstand you. I'm using the ABS touchpad on a chromebook, but it has two different platforms, **x86 (ACPI)** and **ARM (OF)**. On the x86 platform I expect to use `I2C\MODALIAS_acpi:ELAN0000:` to match, on the ARM platform it will be `I2C\MODALIAS_of:NtrackpadT(null)Celan,ekth3000`.
> 
> On the x86 platform, it is also possible to use the `I2C\NAME_Synopsys-DesignWare-I2C-adapter` match, but it is definitely suitable for the ARM platform. So I tend to use a unified approach (MODALIAS) to match chromebooks, while also avoiding the HID I2C recovery touchpad on general Linux machines.

If the goal is to be uniform; why not make an ACPI device for ELAN000 on the ARM platform too?

josh-chen-elan:
> If the goal is to be uniform; why not make an ACPI device for `ELAN0000` device on the ARM platform too?

This is good advice, but this mechanism has worked for several years, is not easily modify on older machines, and needs to be modified by OEM rather than IHV (ELAN).

superm1:
> > If the goal is to be uniform; why not make an ACPI device for `ELAN0000` device on the ARM platform too?
> 
> This is good advice, but this mechanism has worked for several years, is not easily modify on older machines, and needs to be modified by OEM rather than IHV (ELAN).

I suppose maybe when you introduce the next model touchpad you can do it this way.

josh-chen-elan:
> > > If the goal is to be uniform; why not make an ACPI device for `ELAN0000` device on the ARM platform too?
> > 
> > 
> > This is good advice, but this mechanism has worked for several years, is not easily modify on older machines, and needs to be modified by OEM rather than IHV (ELAN).
> 
> I suppose maybe when you introduce the next model touchpad you can do it this way.
OK, I will notify the project manager.

Issue #5199, opened 2022-10-25
fwupdmgr "Host Security Events" seems to get kernel taint status wrong


apollo13:
**Describe the bug**
Executing `sudo fwupdmgr security --force` yields this on my laptop:

```
Host Security Events
  2022-10-25 07:44:49:  ✘ Kernel is no longer tainted
  2022-10-10 17:30:03:  ✔ Kernel lockdown enabled
  2022-10-10 17:30:03:  ✔ Secure Boot enabled
  2022-09-09 06:24:39:  ✔ Kernel is tainted
```

I do not think a tainted kernel should be shown as ✔-- this should be the icon when it is not longer tainted I think.

**fwupd version information**

```shell
runtime   org.freedesktop.fwupd         1.8.6
runtime   org.freedesktop.fwupd-efi     1.3
compile   org.freedesktop.gusb          0.4.1
runtime   com.hughsie.libjcat           0.1.12
runtime   org.kernel                    5.19.15-301.fc37.x86_64
runtime   com.dell.libsmbios            2.4
compile   com.hughsie.libjcat           0.1.12
compile   org.freedesktop.fwupd         1.8.6
runtime   org.freedesktop.gusb          0.4.1
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): dnf

**Additional questions**

- Operating system and version: Fedora 37
- Have you tried rebooting? Yes
- Is this a regression? Can't tell, first time I noticed it due to the gnome device security panel which is rather new…

PR #5203, opened 2022-10-26
trivial: fix a host security event message

superm1:
the structure is old->new, so from tained to not tainted should be: "Kernel is no longer tainted" not "Kernel is tainted"

Fixes: #5199

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
Ha! I was just pondering that issue, you ticked off a TODO without me even opening an editor :)

PR #5218, opened 2022-10-31
qsi-dock: create a new plugin

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/qsi-dock/README.md
```diff
@@ -0,0 +1,32 @@
+# QSI Dock
+
+## Introduction
+
+This plugin uses the MCU to write all the dock firmware components. The MCU version
+is provided by the DMC bcdDevice.
```
superm1, line 6: MCU and DMC should probably be explained

plugins/qsi-dock/qsi-dock.quirk
```diff
@@ -0,0 +1,3 @@
+[USB\VID_047D&PID_808D]
```
superm1, line 1: probably worth the discussion; should all QSI stuff be matched or case by case?

plugins/qsi-dock/fu-qsi-dock-common.h
```diff
@@ -0,0 +1,67 @@
...
+#define FU_QSI_DOCK_FIRMWARE_IDX_DMC_PD 0x01
+#define FU_QSI_DOCK_FIRMWARE_IDX_DP	0x02
+#define FU_QSI_DOCK_FIRMWARE_IDX_TBT4	0x04
+#define FU_QSI_DOCK_FIRMWARE_IDX_USB3	0x08
+#define FU_QSI_DOCK_FIRMWARE_IDX_USB2	0x10
+#define FU_QSI_DOCK_FIRMWARE_IDX_AUDIO	0x20
+#define FU_QSI_DOCK_FIRMWARE_IDX_I225	0x40
+#define FU_QSI_DOCK_FIRMWARE_IDX_MCU	0x80
+
+typedef struct {
```
superm1, line 24: doesn't this need to be packed?

plugins/qsi-dock/qsi-dock.quirk
hughsie: case by case; i want vendors to implement the BOS descriptors longer term.

plugins/qsi-dock/fu-qsi-dock-mcu-device.c
```diff
@@ -271,7 +271,7 @@ fu_qsi_dock_mcu_device_write_chunk(FuQsiDockMcuDevice *self,
 	g_autoptr(GPtrArray) chunks = NULL;
 
 	chunks = fu_chunk_array_new(fu_chunk_get_data(chk_page),
-				    fu_chunk_get_data_sz(chk_page),
+				    FU_QSI_DOCK_EXTERN_FLASH_PAGE_SIZE,
```
hughsie, line 274: I don' understand this change; won't this be exactly the same value? If it's not exactly the same value then we're over-reading the buffer which is going to make valgrind and asan explode.

Is the issue perhaps that fu_firmware_get_bytes() isn't aligned to FU_QSI_DOCK_EXTERN_FLASH_PAGE_SIZE? In which case we need to use fu_bytes_align() rather than just overread a C buffer which probably has junk.

plugins/qsi-dock/README.md
hughsie: Done.

plugins/qsi-dock/fu-qsi-dock-common.h
hughsie: Yes, good catch! Fixed.

plugins/qsi-dock/fu-qsi-dock-mcu-device.c
hughsie: I've switched this to use `fu_bytes_align()` instead.

hughsie:
@hssinf can you verify this still works on actual hardware please - then we're good to commit I think.

hssinf:
> @hssinf can you verify this still works on actual hardware please - then we're good to merge I think.

@hughsie It works fine.

hssinf:
> branch from [ `67762c8` ](/fwupd/fwupd/commit/67762c82b2fd5fbb1b65228e45aef32b132a724f) to

@hughsie  It works fine.

PR #5220, opened 2022-10-31
Wip/hughsie/dbx ignore

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
I would think for the purpose of VM testing of secure boot stuff you don't want dbx updates turned off by default

hughsie:
> you don't want dbx updates turned off by default

You mean we should disable it when under emulation in all cases? Or we should instead nuke the extra entries and try to narrow it down to old "bios" versions with the new metadata and some extra time?

superm1:
I mean we should leave it for emulation. So we can very easily validate dbx updates in an environment that can be controlled more easily.

hughsie:
Okay, lets drop this one and have a rethink when we've got the data we need: https://github.com/fwupd/fwupd/pull/5224

PR #5228, opened 2022-11-02
elantp - Support ic_type 0x12 & 0x13, iap_ver 0x3 & 0x5 iap flow.

jinglewu:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -192,19 +237,77 @@ fu_elantp_hid_device_read_hatpic_enable(FuElantpHidDevice *self, GError **error)
...
+		self->force_table_addr = 0xFF40 * 2;
+		return TRUE;
+	}
+	if (!fu_elantp_hid_device_read_cmd(self, ETP_CMD_FORCE_ADDR, buf, sizeof(buf), error)) {
+		g_prefix_error(error, "failed to read force table address cmd: ");
+		return FALSE;
+	}
+	addr_wrds = fu_memread_uint16(buf, G_LITTLE_ENDIAN);
+	if (addr_wrds % 32 != 0) {
+		g_prefix_error(error, "illgeal force table address (%x): ", addr_wrds);
```
hughsie, line 256: g_set_error

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -192,19 +237,77 @@ fu_elantp_hid_device_read_hatpic_enable(FuElantpHidDevice *self, GError **error)
...
+				       guint16 iap_ver,
+				       GError **error)
+{
+	guint8 buf[2] = {0x0};
+	guint16 pw;
+	guint16 value;
+	if (iap_ver == 0x5 && ic_type == 0x13)
+		pw = ETP_I2C_IC13_IAPV5_PW;
+	else
+		return TRUE;
```
hughsie, line 278: can you reverse this please, i.e. only return when it's not 0x5 and 0x13 -- then you can set `guint16 pw = ETP_I2C_IC13_IAPV5_PW`

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -192,19 +237,77 @@ fu_elantp_hid_device_read_hatpic_enable(FuElantpHidDevice *self, GError **error)
...
+		return FALSE;
+	}
+
+	if (!fu_elantp_hid_device_read_cmd(self, ETP_CMD_I2C_FW_PW, buf, sizeof(buf), error)) {
+		g_prefix_error(error, "failed to read fw password cmd: ");
+		return FALSE;
+	}
+	value = fu_memread_uint16(buf, G_LITTLE_ENDIAN);
+	if (value != pw) {
+		g_set_error(error, FWUPD_ERROR, FWUPD_ERROR_NOT_SUPPORTED, "can't set fw password");
```
hughsie, line 291: maybe include what we got and also what we expected in the error?

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -305,6 +408,26 @@ fu_elantp_hid_device_setup(FuDevice *device, GError **error)
 	if (!fu_elantp_hid_device_ensure_iap_ctrl(self, error))
 		return FALSE;
 
+	self->force_table_support = FALSE;
```
hughsie, line 411: i don't think you need this, it's false by default.

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -338,11 +463,122 @@ fu_elantp_hid_device_prepare_firmware(FuDevice *device,
...
+
+	if (self->force_table_addr < force_table_addr) {
+		g_prefix_error(error,
+			       "forcetable address wrong (%x,%x): ",
+			       force_table_addr,
+			       self->force_table_addr);
+		return FALSE;
+	}
+
+	memset(buf, 0xFF, 64);
```
hughsie, line 542: can you do guint8 buf[64] = {0xFF}; instead?

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -338,11 +463,122 @@ fu_elantp_hid_device_prepare_firmware(FuDevice *device,
...
+	const guint8 signature[] = {0xAA, 0x55, 0xCC, 0x33, 0xFF, 0xFF};
+	guint8 buf[64];
+	guint16 block_checksum;
+	guint16 filling_value;
+
+	if (self->force_table_addr == force_table_addr)
+		return TRUE;
+
+	if (self->force_table_addr < force_table_addr) {
+		g_prefix_error(error,
```
hughsie, line 535: g_set_error

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -382,8 +619,48 @@ fu_elantp_hid_device_write_firmware(FuDevice *device,
...
+				    FWUPD_ERROR_WRITE,
+				    "filling forcetable failed");
+			return FALSE;
+		}
+		chunks = fu_chunk_array_new(buf2 + iap_addr,
+					    bufsz - iap_addr,
+					    0x0,
+					    0x0,
+					    self->fw_page_size);
+		total_pages = (self->force_table_addr - iap_addr - 1) / self->fw_page_size + 1;
```
hughsie, line 654: i'm not sure I understand this; what's going on?

plugins/elantp/fu-elantp-firmware.c
```diff
@@ -130,6 +192,53 @@ fu_elantp_firmware_parse(FuFirmware *firmware,
...
+	if (!fu_memread_uint16_safe(buf,
+				    bufsz,
+				    offset + ETP_IAP_VER_ADDR_WRDS * 2,
+				    &self->iap_ver,
+				    G_LITTLE_ENDIAN,
+				    error))
+		return FALSE;
+
+	self->force_table_addr = 0;
+	self->force_table_support = FALSE;
```
hughsie, line 211: both these are set to 0x0 by default, so I don't think this is needed.

plugins/elantp/fu-elantp-firmware.c
```diff
@@ -72,6 +108,30 @@ fu_elantp_firmware_check_magic(FuFirmware *firmware, GBytes *fw, gsize offset, G
 		}
 	}
 
+	if (self->force_table_addr == 0)
+		return TRUE;
+
+	for (gsize i = 0; i < sizeof(elantp_signature); i++) {
```
hughsie, line 114: I think I'd rather this loop was in the `if` block in case we have other signatures in the future.

plugins/elantp/fu-elantp-hid-device.c
hughsie: @jinglewu

jinglewu:
I can't write any comments at codereview. (show pending, Collaborator)
1. The forcetable data is start the force_table_addr to the end of the firmware file.
2. The firmware update process does not contain the forcetable data.
3. The forcetable size will be different, needs to be filled the data to current firmware.

plugins/elantp/fu-elantp-hid-device.c
josh-chen-elan: @jinglewu

plugins/elantp/fu-elantp-hid-haptic-device.c
```diff
@@ -159,11 +159,18 @@ fu_elantp_hid_haptic_device_ensure_iap_ctrl(FuDevice *parent,
...
 	else
 		fu_device_remove_flag(FU_DEVICE(parent), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
 
+	if (self->tp_iap_ver <= 5) {
+		if ((self->tp_iap_ctrl & ETP_I2C_MAIN_MODE_ON2) == 0)
+			fu_device_add_flag(FU_DEVICE(parent), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+		else
+			fu_device_remove_flag(FU_DEVICE(parent), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+	}
+
```
hughsie, line 173: The logic here is a bit funky; I think you want something like `if (self->tp_iap_ver <= 5) { CHECK ON2 } else { CHECK ON }` -- no?

plugins/elantp/fu-elantp-hid-device.c
hughsie: so is total_pages less than chunks->len?

jinglewu: yes.

hughsie: Can you add that into the code please; it's certainly not obvious.

jinglewu: Do you mean to add this 3 three comments to the code?

hughsie: I think doing something like `if (total_pages > chunks->len) g_set_error()` etc would be a good idea, but you can do this as a follow up PR if that helps.

hughsie (approved):

hughsie:
@jinglewu you need to fix the two clang-format precommit issues. In the future if you push a `wip/elan/whatever` branch to the origin remote then I can fix up these kinds of nits on my own.

Issue #5077, opened 2022-09-24
'KERNEL BUG: "type" attribute not exported'


LinuxOnTheDesktop:
**Describe the bug**
Some (all?) fwupd commands produce output that includes the following.
`
KERNEL BUG: 'type' attribute not exported: (failed to load type: Failed to open file “/sys/class/firmware-attributes/thinklmi/attributes/SecureBoot/type”: No such file or directory)`

**Steps to Reproduce**
```
$ sudo -i flatpak run org.freedesktop.fwupd refresh
[sudo] password for [me]: 
Loading…                 [**************************             ]00:03:02:0020 FuBiosSettings       KERNEL BUG: 'type' attribute not exported: (failed to load type: Failed to open file “/sys/class/firmware-attributes/thinklmi/attributes/SecureBoot/type”: No such file or directory)
Loading…                 [-                                      ]
WARNING: This package has not been validated, it may not work properly.
Downloading…             [***************************************]
Downloading…             [***************************************]
Downloading…             [***************************************]
```
(As I have said before, the warning about validation is unfortunate, because it is ambiguous between the problem pertaining to (1) fwupd, (2) some particular update. With time I inferred that _1_ is meant. The user should not have to make that inference.)

**Expected behavior**
No cryptic error message.

**fwupd version information**
Please provide the version of the daemon and client.

```shell
$ fwupdmgr --version
runtime   org.freedesktop.fwupd         1.7.5
compile   org.freedesktop.gusb          0.3.10
runtime   com.dell.libsmbios            2.4
runtime   org.kernel                    5.15.0-48-generic
compile   org.freedesktop.fwupd         1.7.5
runtime   org.freedesktop.gusb          0.3.10
$ flatpak list | grep fwup
fwupd	org.freedesktop.fwupd	1.8.4	stable	system

```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): flatpak, but I have an apt version installed too.

**fwupd device information**

```
$ fwupdmgr get-devices --show-all-devices
20KH001UKP
│
├─Core™ i7-8550U CPU @ 1.80GHz:
│     Device ID:          4bde70ba4e39b28f9eab1628f9dd6e6244c03027
│     Current version:    0x000000f0
│     Vendor:             Intel
│     GUIDs:              b9a2dd81-159e-5537-a7db-e7101d164d3f
│                         30249f37-d140-5d3e-9319-186b1bd5cac3
[...]

PR #5230, opened 2022-11-02
Only include the 'attribute not exported' warning on debug builds

hughsie:
End users are not really capable of fixing the root issue, and people seem overly worried that something bad has gone wrong.

Fixes https://github.com/fwupd/fwupd/issues/5077

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Can we please fix the kernel issue first? This is never going to get fixed if there is no noise.

hughsie:
The problem is that the noise is being shown to customers who are filing bugs when the noise is shown. I think practically @mrhpearson is the person on the hook for fixing it, and I'm pretty sure it's already on his TODO.

superm1:
I mean anyone can fix it. It's not rocket science, it's just breaking up some strings and exporting another sysfs file.

superm1 (changes requested):
Although this is would "work", we're going to have to change the test suite to reflect the behavior of the fixed driver vs the non-fixed driver.

That's why I want to see this fixed in the kernel first.

hughsie:
Ahh I see, i.e. we need a system with the kernel feature added. @mrhpearson any idea on who we need to poke to move this a long a bit? Regardless of when it's fixed upstream it'll be in RHEL 8/9 unfixed for a very long time...

superm1:
FYI - a test patch was uploaded to the kernel Bugzilla by a community member but it only seems to help some platforms.  I have a P620 that still fails with it.

PR #5231, opened 2022-11-02
Wip/hughsie/esp recurse

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-context.c
+++ b/libfwupdplugin/fu-context.c
@@ -988,6 +988,9 @@ fu_context_get_esp_volumes(FuContext *self, GError **error)
 	} else {
 		for (guint i = 0; i < volumes_esp->len; i++) {
 			FuVolume *vol = g_ptr_array_index(volumes_esp, i);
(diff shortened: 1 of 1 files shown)


superm1 (commented):
As a different heuristic what do you think about checking if the filesystem is a Linux filesystem? Or Linux partition type? Those can't be ESPs anyway.

hughsie:
> Or Linux partition type? Those can't be ESPs anyway.

IIRC I think there are quite a few different things that *could* be ESPs -- e.g. you can use NTFS with some firmware. EXT3/4 would be unusual, but wouldn't be madness.

superm1:
How about if there are multiple ESPs then ignore the Linux ones then?

hughsie:
Lets keep it simple; we'll do what you suggest and just filter out ext4 as a type from the ESPs. If someone wants to do it we can say "not supported". Updated.

superm1 (approved):

PR #5243, opened 2022-11-04
wacom-usb: remove the use-runtime-version flag from IPS

skomra:
The use-runtime-version flag was initially used with the Intuos Pro Small (2nd-gen USB v2) to force the use of the legacy code path. The legacy code path was required because of a bug in the identification of the Bluetooth (I6) firmware.

It has been decided that the bug will be fixed before any fimware changes are released, removing the need to use the legacy code path.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/wacom-usb/wacom-usb.quirk
+++ b/plugins/wacom-usb/wacom-usb.quirk
@@ -86,4 +86,3 @@ GType = FuWacAndroidDevice
 [USB\VID_056A&PID_03DC]
 GType = FuWacDevice
 Plugin = wacom_usb
-Flags = use-runtime-version


superm1:
But are there devices in the wild with this bug? Or will it be fixed before first devices to market?

skomra:
The fw bug is in the identification of a sub-module, but the bug won't prevent us from updating the MCU. The bug (which will be fixed by updating the MCU fw) would prevent us from updating the touch submodule. Once the MCU is updated though, we will be able to update the submodules correctly.  If we leave this flag in, the touch module won't be recognized by the legacy path and thus can't be updated.

I'm new to our wacom-usb plugin and didn't fully understand that I was putting in the flag because of a firmware bug, I primarily put it in because the Bluetooth firmware wouldn't update without it. I'm a software engineer and hadn't considered a FW bug.

superm1:
> The fw bug is in the identification of a sub-module, but the bug won't prevent us from updating the MCU. The bug (which will be fixed by updating the MCU fw) would prevent us from updating the touch submodule. Once the MCU is updated though, we will be able to update the submodules correctly.  If we leave this flag in, the touch module won't be recognized by the legacy path and thus can't be updated.
> 
> I'm new to our wacom-usb plugin and didn't fully understand that I was putting in the flag because of a firmware bug, I primarily put it in because the Bluetooth firmware wouldn't update without it. I'm a software engineer and hadn't considered a FW bug. 

Thanks for explaining, agree with your direction.

superm1 (approved):

skomra:
Thank you!

PR #5252, opened 2022-11-09
trivial: Fix the tests on s390x

hughsie:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ plugins/acpi-facp/tests/
 plugins/ata/tests/
 plugins/dfu/tests/
 plugins/nvme/tests/
-plugins/synaptics-mst/tests/
 plugins/synaptics-prometheus/tests/
 plugins/tpm-eventlog/tests/
 plugins/uefi-dbx/tests/
--- a/plugins/synaptics-mst/fu-self-test.c
+++ b/plugins/synaptics-mst/fu-self-test.c
@@ -203,6 +203,7 @@ main(int argc, char **argv)
 	/* only critical and error are fatal */
 	g_log_set_fatal_mask(NULL, G_LOG_LEVEL_ERROR | G_LOG_LEVEL_CRITICAL);
 	testdatadir = g_test_build_filename(G_TEST_DIST, "tests", NULL);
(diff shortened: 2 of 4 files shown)


superm1:
Can I see the actual failure?

hughsie:
Sure! https://koji.fedoraproject.org/koji/getfile?taskID=93975142&volume=DEFAULT&name=build.log&offset=-4000

superm1:
Hmm, wouldn't it be better to expect that warning and run more of the test?

hughsie:
> Hmm, wouldn't it be better to expect that warning and run more of the test?

I don't think it's going to work. I don't think we're going to find many s390x machines running on Synaptics MST-capable hardware...

superm1:
> > Hmm, wouldn't it be better to expect that warning and run more of the test?
> 
> I don't think it's going to work. I don't think we're going to find many s390x machines running on Synaptics MST-capable hardware...

That's a good point.  Maybe we shouldn't be compiling the plugin even for that architecture then too..

hughsie:
>  Maybe we shouldn't be compiling the plugin even for that architecture then too

I think the firmware parsing tests are maybe useful, but otherwise I don't mind.

hughsie:
I'm not sure s390x even has working USB FWIW. Maybe it doesn't make sense at all.

superm1:
> I'm not sure s390x even has working USB FWIW. Maybe it doesn't make sense at all.

oh you mean like axe fwupd for s390x all together?

hughsie:
> oh you mean like axe fwupd for s390x all together?

Maybe; my worry is that it might get a bit political inside Red Hat. I'm erring on merging this, and testing the waters in Fedora.

superm1:
> Maybe; my worry is that it might get a bit political inside Red Hat. I'm erring on merging this, and testing the waters in Fedora.

If you're gonna keep s390x I'd say axe this whole plugin, we can nickle and dime it away on breakages.

hughsie:
Lets just use the DMI data we use in the libfwupdplugin self tests. Fixed!

superm1 (approved):

PR #5259, opened 2022-11-14
genesys: If the transfer failed then do not retry the flash status

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/genesys/fu-genesys-usbhub-device.c
+++ b/plugins/genesys/fu-genesys-usbhub-device.c
@@ -368,10 +368,12 @@ fu_genesys_usbhub_device_wait_flash_status_register_cb(FuDevice *device,
 		return FALSE;
 	}
 	if (status != helper->expected_val) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_INTERNAL,
-				    "wrong value in flash status register");
+		g_set_error(error,
+			    G_IO_ERROR,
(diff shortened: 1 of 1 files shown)


gportay:
I was not aware about that callback.

This trick is about dropping the retry traces "failed on try x of y"?

If so, this is very nice !

hughsie:
> This trick is about dropping the retry traces "failed on try x of y"?

Yes and no. Genesys told us that we shouldn't be retrying the request if anything *other* than the register value is incorrect. e.g. a USB transfer failure should be fatal without retries. Not sure if it fixes the bug they saw tho.

adamgene:
Hi all, the old issue was fixed by new hub FW. But for the new device, it has to support a new flash. Please add below into `cfi.quirk`. Very thanks.

[CFI\FLASHID_EF3011]
Name = W25X10CL
FirmwareSizeMax = 0x20000

hughsie:
@adamgene I've added this in 0eee4d7faa0195b823865789e123851bbdab21a8 -- are there any other SPI chips you've used in other projects you'd like to add too?

adamgene:
@hughsie Can you create a PR that I can commit these changes?

hughsie:
@adamgene I've pushed directly to the main branch with that new W25X10CL CFI data, but if you have any more please push to a new wip/genesys/cfi or something -- you have commit ACLs.

PR #5264, opened 2022-11-16
Adjust rollback protection

superm1:
Fedora 37 coming out has exposed more people to security attributes and it seems that some decisions made for rollback protection don't scale as well as expected.

1. Move rollback protection to level 4.
2. Add UEFI based rollback protection for Lenovo

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


docs/hsi-tests.d/org.fwupd.hsi.Amd.PlatformRollbackProtection.json
```diff
@@ -1,22 +1,22 @@
...
-  },
-  "success-results" : {
-    "enabled" : "rollback protection enabled"
-  },
-  "hsi-level" : 1,
-  "references" : {
-    "https://www.psacertified.org/blog/anti-rollback-explained/" : "Rollback protection"
-  },
-  "fwupd-version" : "1.8.0"
+	"id": "org.fwupd.hsi.Amd.PlatformRollbackProtection",
```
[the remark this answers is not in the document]
superm1: yeah clang-format told me that the json needs whitespace.

superm1:
@mrhpearson FYI
I think it's self-explanatory but if you have more questions on this feel free to ask here or hit me up on email.

libfwupdplugin/fu-security-attrs.c
```diff
@@ -348,6 +348,7 @@ static struct {
     {FWUPD_SECURITY_ATTR_ID_TPM_VERSION_20, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL},
     {FWUPD_SECURITY_ATTR_ID_UEFI_PK, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL},
     {FWUPD_SECURITY_ATTR_ID_UEFI_SECUREBOOT, FWUPD_SECURITY_ATTR_LEVEL_CRITICAL},
+    {FWUPD_SECURITY_ATTR_ID_UEFI_ROLLBACK_PROTECTION, FWUPD_SECURITY_ATTR_LEVEL_SYSTEM_PROTECTION},
```
hughsie, line 351: doesn't this need a .xml file so it makes it into the spec? I'm also not sure about the name; it's not UEFI specific.

superm1: Oh yeah; I'll write up a separate JSON for it.  In terms of the name, any better ideas?  It's for FAR on UEFI firmware, which is why I came up with the name I did.

superm1: added new JSON file.

plugins/lenovo-thinklmi/fu-lenovo-thinklmi-plugin.c
```diff
@@ -83,6 +84,38 @@ fu_lenovo_thinklmi_plugin_device_registered(FuPlugin *plugin, FuDevice *device)
...
+
+	if (g_strcmp0(fwupd_bios_setting_get_current_value(bios_attr), "Disable") == 0) {
+		fwupd_security_attr_add_flag(attr, FWUPD_SECURITY_ATTR_FLAG_ACTION_CONFIG_FW);
+		fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_NOT_ENABLED);
+		return;
+	}
+
+	/* success */
+	fwupd_security_attr_set_result(attr, FWUPD_SECURITY_ATTR_RESULT_ENABLED);
+	fwupd_security_attr_add_obsolete(attr, "pci_psp");
```
superm1, line 115: Wanted to discuss this specifically...

Given the AMD solution is hardware based but the OEM ones might not be; I was thinking maybe the OEM ones maybe we should put at a lower level, and **not obsolete** the AMD one.  They're mutually exclusive.  You can have the OEM protection and the AMD one simultaneously.

This has an implication that on AMD you can't get to HSI 4 unless you have the AMD one at least then, but you can still get the OEM ones turned on for lower levels (whatever level we decide)

superm1: (I've mocked up the PR accordingly)

src/fu-security-attr-common.c
```diff
@@ -189,7 +189,7 @@ fu_security_attr_get_name(FwupdSecurityAttr *attr)
 	}
 	if (g_strcmp0(appstream_id, FWUPD_SECURITY_ATTR_ID_AMD_ROLLBACK_PROTECTION) == 0) {
 		/* TRANSLATORS: Title: if firmware enforces rollback protection */
-		return g_strdup(_("Rollback protection"));
+		return g_strdup(_("Security Processor Rollback protection"));
```
superm1, line 192: I suspect this string is too long....
```
HSI-4
✔ Encrypted RAM:                 Encrypted
✔ Security Processor Rollback protection:Enabled
```

libfwupdplugin/fu-security-attrs.c
superm1: Renamed the attribute

hughsie (approved):

PR #5266, opened 2022-11-17
synaptics-rmi: Fix the issue when updating non-secure devices

vhuag:
Correct the register parsing, properties2 only exists when the bit in properties1 is set.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Can you reformat with clang-format and force push the branch please? CI is failing.

vhuag:
> 

I fixed the failure and push one more commit to fix the problem displaying the incorrect BL version.
Thanks

hughsie (approved):

PR #5267, opened 2022-11-17
Do not query PolicyKit when requester is trusted

valentindavid:
In the context of snap, we want to be able to distribute the same build of fwupd for both desktop and IoT. However IoT images do not typically run PolicyKit.

So if we detect requested in trusted, we skip using PolicyKit.

When compiled against polkit we also make 0 (root) untrusted by default so that the new behavior of not using polkit is only used when 0 is explictly added to `TrustedUids`.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-daemon.c
+++ b/src/fu-daemon.c
@@ -455,7 +455,8 @@ fu_daemon_authorization_is_valid(PolkitAuthorizationResult *auth, GError **error
 	/* success */
 	return TRUE;
 }
-#else
+#endif /* HAVE_POLKIT */
+
 static gboolean
 fu_daemon_authorization_is_trusted(FuEngineRequest *request, GError **error)
 {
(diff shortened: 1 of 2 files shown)


superm1:
This would have an implication outside of snapped packages too; if you call `fwupdmgr` as root (ie `sudo fwupdmgr foo`) then it would no longer query PK either.

Rather than making that bold assertion; could we instead have the daemon make a runtime check for PK in the environment it's running and only apply this behavior if PK isn't found?

superm1:
also it looks like this broke in CI in various ways.

> 17:48:11.073 FIXME                polkit_authority_check_authorization_finish: assertion 'error == NULL || *error == NULL' failed

hughsie:
> if you call fwupdmgr as root (ie sudo fwupdmgr foo) then it would no longer query PK either.

I know it's changing the status quo, but ideologically, is `sudo fwupdmgr fooo` not asking polkit auth any different from running `sudo fwupdtool fooo` which also doesn't ask for permission either?

valentindavid:
Note the change in `fu-engine.c`. The point here was to make 0 not trusted by default so it would always try PK unless it was configured.

But now I see there are methods like `Quit` which do not use PK but rather use trusted uids. So maybe that approach was not working properly.

I can test the idea of testing polkit is installed (that is either running or can be activated), and passing around the context.

valentindavid:
I have made this alternative implementation that checks if polkit is either running or activatable: https://github.com/valentindavid/fwupd/commit/0c124c5447e9a132a58c87ddab29eb7c2d896d84

Please tell me what approach is best, then I will go through all the tests and make it ready for proper review.

superm1:
> I have made this alternative implementation that checks if polkit is either running or activatable: [valentindavid@0c124c5](https://github.com/valentindavid/fwupd/commit/0c124c5447e9a132a58c87ddab29eb7c2d896d84)
> 
> Please tell me what approach is best, then I will go through all the tests and make it ready for proper review.

Yes I prefer that alternative approach.

hughsie:
@valentindavid another idea in https://github.com/fwupd/fwupd/pull/5269 perhaps?

hughsie:
Closing in favour of https://github.com/fwupd/fwupd/pull/5269

PR #5270, opened 2022-11-19
Generate OVAL rules

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
What we generated: [oval.xml.txt](https://github.com/fwupd/fwupd/files/10046766/oval.xml.txt)

superm1 (approved):

hughsie:
Ohh it massively fails validation at the moment.

smallorange:
There are some errors in the generated OVAL file. The fixed version can be found in the attachment.
[kate-oval.xml.txt](https://github.com/fwupd/fwupd/files/10081708/kate-oval.xml.txt)

docs/generate-oval.py
```diff
@@ -0,0 +1,147 @@
...
+
+        red_def = ET.SubElement(tests, "red-def:fwupdsecattr_test")
+        red_def.set("check", "at least one")
+        red_def.set("comment", item["name"])
+        red_def.set("id", "oval:{}:tst:1".format(item["id"]))
+        red_def.set("version", "1")
+
+        red_def_object = ET.SubElement(red_def, "red-def:object")
+        red_def_object.set("red-def:object", "oval:{}:obj:1".format(item["id"]))
+        red_def_state = ET.SubElement(red_def, "red-def:state")
```
smallorange, line 117: The "state" in "tests" should look like as follows.
<red-def:state **state_ref**="oval:org.fwupd.hsi.Amd.PlatformRollbackProtection:ste:1" />

not
<red-def:state **red-def:state**="oval:org.fwupd.hsi.Amd.SpiReplayProtection:ste:1" />

docs/generate-oval.py
```diff
@@ -0,0 +1,147 @@
...
+    tests = ET.SubElement(oval_definitions, "tests")
+    for item in items:
+
+        red_def = ET.SubElement(tests, "red-def:fwupdsecattr_test")
+        red_def.set("check", "at least one")
+        red_def.set("comment", item["name"])
+        red_def.set("id", "oval:{}:tst:1".format(item["id"]))
+        red_def.set("version", "1")
+
+        red_def_object = ET.SubElement(red_def, "red-def:object")
```
smallorange, line 115: The "object" in "tests" should be
<red-def:object **object_ref**="oval:org.fwupd.hsi.Amd.PlatformRollbackProtection:obj:1" />

not
<red-def:object **red-def:object**="oval:org.fwupd.hsi.Amd.PlatformRollbackProtection:obj:1" />

docs/generate-oval.py
```diff
@@ -0,0 +1,147 @@
...
+        red_def_state.set("red-def:state", "oval:{}:ste:1".format(item["id"]))
+
+    objects = ET.SubElement(oval_definitions, "objects")
+    for item in items:
+
+        red_def = ET.SubElement(objects, "red-def:fwupdsecattr_object")
+        red_def.set("id", "oval:{}:obj:1".format(item["id"]))
+        red_def.set("version", "1")
+
+        red_def_stream_id = ET.SubElement(red_def, "red-def:fwupdsecattr_object")
```
smallorange, line 127: The input (object) value is in key:value format so the tag name should be "ref-def:**key-name**"

For example:
 <red-def:**stream-id** datatype="string">org.fwupd.hsi.Amd.PlatformRollbackProtection</red-def:stream-id>

I only made one input (stream-id) and one output value (security-attr) for the HSI test item.

docs/generate-oval.py
```diff
@@ -0,0 +1,147 @@
...
+        red_def_stream_id.text = format(item["id"])
+
+    states = ET.SubElement(oval_definitions, "states")
+    for item in items:
+
+        red_def = ET.SubElement(states, "red-def:fwupdsecattr_state")
+        red_def.set("id", "oval:{}:ste:1".format(item["id"]))
+        red_def.set("version", "1")
+
+        red_def_security_attr = ET.SubElement(red_def, "red-def:fwupdsecattr_object")
```
smallorange, line 138: same here. red-def should be "red-def:**output-key-name**" in the "state" element.

for example:
<red-def:**security-attr** datatype="string" operation="pattern match">not-found|not-enabled</red-def:security-attr>

docs/generate-oval.py
hughsie: Fixed, thanks.

docs/generate-oval.py
hughsie: Fixed, thanks.

docs/generate-oval.py
hughsie: Fixed, thanks.

docs/generate-oval.py
hughsie: Fixed, thanks.

docs/generate-oval.py
```diff
@@ -0,0 +1,151 @@
...
+        "http://oval.mitre.org/XMLSchema/oval-definitions-5 oval-definitions-schema.xsd "
+        "http://oval.mitre.org/XMLSchema/oval-definitions-5#unix unix-definitions-schema.xsd "
+        "http://oval.mitre.org/XMLSchema/oval-definitions-5#linux linux-definitions-schema.xsd",
+    )
+
+    generator = ET.SubElement(oval_definitions, "generator")
+    for key, value in {
+        "oval:product_name": "fwupd",
+        "oval:product_version": args.version,
+        "oval:schema_version": args.version,
```
smallorange, line 69: "schema_version" should address to OVAL schema version. For now, the HSI feature is implemented in version 5.11.3. We need a new argument to set OVAL schema version.

I only fixed this and then the generated XML file pass the "oscap oval validate"

hughsie: Fixed, thanks!

smallorange:
I've tested it on my system and OVAL implementation. It's ok with me.

smallorange (approved):
I've tested it on my system and OVAL implementation. It's ok with me.

Issue #5281, opened 2022-11-24
04F3:311C ELAN touchpad unusable after upgrade


dequis:
Laptop: Dell Precision 5560
Touchpad: `DLL0945:00 04F3:311C` via i2c_designware

```
runtime   org.freedesktop.fwupd         1.8.7
runtime   org.freedesktop.fwupd-efi     1.3
runtime   com.hughsie.libjcat           0.1.12
runtime   com.dell.libsmbios            2.4
compile   com.hughsie.libjcat           0.1.12
runtime   org.kernel                    5.15.77-1-lts
compile   org.freedesktop.gusb          0.4.2
compile   org.freedesktop.fwupd         1.8.7
runtime   org.freedesktop.gusb          0.4.2
```

Hey there!

I think I bricked the touchpad of my laptop.

Title inspired by a very similar issue, https://github.com/fwupd/fwupd/issues/2926 (which is actually about both the touchpad and the fingerprint reader)

I was trying to fix the issue described [in this issue from the libinput tracker](https://gitlab.freedesktop.org/libinput/libinput/-/issues/618) using the steps described [in this comment](https://gitlab.freedesktop.org/libinput/libinput/-/issues/618#note_1379374) to extract and flash the firmware from the official updater, which i downloaded from [dell](https://www.dell.com/support/home/en-eu/drivers/driversdetails?driverid=4w96v&lwp=rt).

I realize this probably isn't a supported workflow since there's no ELAN touchpad firmware on LVFS but I figured I might have a chance if I ask for help here.

It no longer shows up in get-devices (which used to mention ELAN more explicitly) or dmesg, I got the IDs from logs of previous boots:

```
kernel: input: DLL0945:00 04F3:311C Mouse as /devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-14/i2c-DLL0945:00/0018:04F3:311C.0009/input/input51
kernel: input: DLL0945:00 04F3:311C Touchpad as /devices/pci0000:00/0000:00:15.1/i2c_designware.1/i2c-14/i2c-DLL0945:00/0018:04F3:311C.0009/input/input52
kernel: hid-generic 0018:04F3:311C.0009: input,hidraw2: I2C HID v1.00 Mouse [DLL0945:00 04F3:311C] on i2c-DLL0945:00
```

Interestingly only the touchpad is missing in current boots, and the "mouse" is 04F3:0400 instead of the usual 04F3:311C (but it reports no input events) 

```
[...]

PR #5283, opened 2022-11-25
elantp: Include 04F3:0400 (bootloader) as a valid i2c-hid touchpad

dequis:
ELAN Touchpads with product ids between 0x3000 and 0x4000 can sometimes end up in bootloader mode (due to not having a valid firmware, for example)

The bootloader has a product id outside of the range checked by fu_elantp_hid_device_probe, resulting in the device disappearing from get-devices when it's in that mode, which makes it seem dead.

This expands the range of that check to include 0x400 so that it is possible to upload a valid firmware to it.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

----

As requested by this comment https://github.com/fwupd/fwupd/issues/5281#issuecomment-1327298367

(We could alternatively keep it as `< 0x3000` and add another condition to allow `0x400` only)

Fixes: https://github.com/fwupd/fwupd/issues/5281

See also: https://github.com/fwupd/fwupd/issues/2926

### DIFF
--- a/plugins/elantp/fu-elantp-hid-device.c
+++ b/plugins/elantp/fu-elantp-hid-device.c
@@ -53,7 +53,7 @@ fu_elantp_hid_device_to_string(FuDevice *device, guint idt, GString *str)
 static gboolean
 fu_elantp_hid_device_probe(FuDevice *device, GError **error)
 {
-	guint16 did;
+	guint16 device_id = fu_udev_device_get_model(FU_UDEV_DEVICE(device));
 
 	/* check is valid */
 	if (g_strcmp0(fu_udev_device_get_subsystem(FU_UDEV_DEVICE(device)), "hidraw") != 0) {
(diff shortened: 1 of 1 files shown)


jinglewu (commented):
PID 1000-2FFFF are not elan touchpad PID. PID 3XXX & 0400 are elan Touchpad PID.

hughsie:
@dequis can you allow-list 0x400 please instead of increasing the range. Thanks!

dequis:
There!

plugins/elantp/fu-elantp-hid-device.c
```diff
@@ -63,9 +65,10 @@ fu_elantp_hid_device_probe(FuDevice *device, GError **error)
 		return FALSE;
 	}
 
+	did = fu_udev_device_get_model(FU_UDEV_DEVICE(device));
```
hughsie, line 68: you can move this up, e.g. `guint16 device_id = fu_udev_device_get_model(FU_UDEV_DEVICE(device));;`

dequis: Done

PR #5286, opened 2022-11-28
Add more metadata to the report attributes

hughsie:
This can be used in gnome-control-center when building the security report.

Fixes https://github.com/fwupd/fwupd/discussions/5262

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2006,6 +2006,28 @@ fu_engine_get_boot_time(void)
 	return NULL;
 }
 
+static void
+fu_engine_get_report_metadata_cpu_device(FuEngine *self, GHashTable *hash)
+{
+	g_autofree gchar *guid = fwupd_guid_hash_string("cpu");
+	g_autoptr(FuDevice) device = NULL;
+	g_autoptr(GError) error_local = NULL;
+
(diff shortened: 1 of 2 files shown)


hughsie:
Urh, contrib/codespell.cfg is being fixed by... codespell...

src/fu-engine.c
```diff
@@ -2006,6 +2006,28 @@ fu_engine_get_boot_time(void)
 	return NULL;
 }
 
+static void
+fu_engine_get_report_metadata_cpu_device(FuEngine *self, GHashTable *hash)
+{
+	g_autofree gchar *guid = fwupd_guid_hash_string("cpu");
```
superm1, line 2012: This feels like it's getting too specifically into the current situation of some older CPUs don't export the needed information.

I feel we should be sticking to whether ANY plugin doesn't export enough information to better scale for future measurements.

hughsie: This is getting the device with that attribute, not the security attribute. i.e. it's the data from CPUID.

superm1: Oh right okay. No worry then.

PR #5289, opened 2022-11-29
remove errant % in spec file

garybuhrmaster:
addresses the build error "sh: line 1: noreplace: command not found"

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

When performing a source build on rpm based systems, one may get messages of the form:
   `sh: line 1: noreplace: command not found`
(seen in fedora koji build logs such as: https://kojipkgs.fedoraproject.org//packages/fwupd/1.8.7/3.fc38/data/logs/x86_64/build.log )

line 362 of fwupd.spec.in file is
   ` %config%(noreplace).....`
which appears to be an error, and should be
    `%config(noreplace).....`
as elsewhere in the spec.

### DIFF
--- a/contrib/fwupd.spec.in
+++ b/contrib/fwupd.spec.in
@@ -359,7 +359,7 @@ done
 %{_bindir}/fwupdagent
 %dir %{_sysconfdir}/fwupd
 %dir %{_sysconfdir}/fwupd/bios-settings.d
-%config%(noreplace)%{_sysconfdir}/fwupd/bios-settings.d/README.md
+%{_sysconfdir}/fwupd/bios-settings.d/README.md
 %dir %{_sysconfdir}/fwupd/remotes.d
 %if 0%{?have_dell}
 %config(noreplace)%{_sysconfdir}/fwupd/remotes.d/dell-esrt.conf


hughsie (changes requested):
I wondered where that was coming from! I actually think we could remove the noreplace completely; it's not designed to be changed by the admin.

garybuhrmaster:
> I wondered where that was coming from!

I was preparing for something else, saw the error, and since I hate errors I don't understand[0], looked further.

> I actually think we could remove the noreplace completely; it's not designed to be changed by the admin.

I went back and forth a bit, but figured that if some admin decided to change the README that preserving it on upgrade (by creating a new README.rpmnew file) would hurt no one (admins sometimes do the strangest things[1]), as any normal upgrade would just replace it should the text get updated by the package at some future point as long as the admin did not change the file.

I'll go back and remove the (noreplace) entirely later when I have some (free) time in front of a real keyboard, or I am also fine with you just closing this PR and making the (few) character change yourself with your own commit.  The PR was mostly to get the projects attention and serve as a placeholder, as I did not know how long reviews typically take (some projects take many months even for trivial issues).


[0] Errors I understand I can often tolerate for years.
[1] Guilty as charged

hughsie (approved):
Thanks!

PR #5291, opened 2022-11-29
build: don't record timestamp in builtin.quirk.gz

jelly:
On Arch Linux fwupd is almost reproducible except for the timestamp recorded in the generated gzip file.

Type of pull request:

- [x] Code fix

This can be observed in this [diffoscope report](https://reproducible.archlinux.org/api/v0/builds/371445/diffoscope). Arch Linux differs from other distributions in reproducibility in that it does not strip non-determinism during it's build (like for example Debian).


hughsie (approved):
Thanks!

PR #5295, opened 2022-12-01
Load report data from /etc/lsb-release

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2055,6 +2055,30 @@ fu_engine_get_report_metadata_os_release(GHashTable *hash, GError **error)
 	return TRUE;
 }
 
+static gboolean
+fu_engine_get_report_metadata_lsb_release(GHashTable *hash, GError **error)
+{
+	const gchar *fn = "/etc/lsb-release";
+	g_autoptr(GHashTable) os_release = NULL;
+	struct {
+		const gchar *key;
+		const gchar *val;
+	} distro_kv[] = {{"CHROMEOS_RELEASE_TRACK", "DistroReleaseTrack"},
+			 {"CHROMEOS_RELEASE_BOARD", "DistroReleaseBoard"},
+			 {NULL, NULL}};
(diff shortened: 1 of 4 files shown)


src/fu-engine.c
```diff
@@ -2055,6 +2055,30 @@ fu_engine_get_report_metadata_os_release(GHashTable *hash, GError **error)
...
+static gboolean
+fu_engine_get_report_metadata_lsb_release(GHashTable *hash, GError **error)
+{
+	const gchar *fn = "/etc/lsb-release";
+	g_autoptr(GHashTable) os_release = NULL;
+	struct {
+		const gchar *key;
+		const gchar *val;
+	} distro_kv[] = {{"CHROMEOS_RELEASE_TRACK", "DistroReleaseTrack"},
+			 {"CHROMEOS_RELEASE_BOARD", "DistroReleaseBoard"},
```
superm1, line 2067: Track makes sense to me but is the board here much different than what's in `/sys/class/dmi/id/`?

hughsie: @bleungatchromium ?

bleungatchromium: Hi @superm1 and @hughsie 

Just to confirm: CHROMEOS_RELEASE_BOARD is not the same as /sys/class/dmi/id.

The board here is a statement of the target board class of the OS image itself, since different Chromebook hardware  may reuse the same OS image.

Just as an example, Google builds a single OS image for all Chromebooks with a 12th Gen Intel mobile processor. The CHROMEOS_RELEASE_BOARD for those is: `brya-signed-mp-v9keys`

However, that "brya" doesn't match the dmi id for all possible hardware variants that can run that OS image. The specific one I have in front of me (HP Dragonfly Chromebook) has a `/sys/class/dmi/id/board_name` of `Redrix`.

hughsie: Gotcha, thanks for the info.

PR #5299, opened 2022-12-06
Add the X-Gpu category for new hardware support

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-util-common.c
+++ b/src/fu-util-common.c
@@ -817,6 +817,11 @@ fu_util_release_get_name(FwupdRelease *release)
 			 * SATA or NVMe disk */
 			return g_strdup_printf(_("%s SSD Update"), name);
 		}
+		if (g_strcmp0(cat, "X-Gpu") == 0) {
+			/* TRANSLATORS: GPU refers to a Graphics Processing Unit, e.g.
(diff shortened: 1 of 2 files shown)


src/fu-util-common.c
```diff
@@ -817,6 +817,11 @@ fu_util_release_get_name(FwupdRelease *release)
 			 * SATA or NVMe disk */
 			return g_strdup_printf(_("%s SSD Update"), name);
 		}
+		if (g_strcmp0(cat, "X-Gpu") == 0) {
+			/* TRANSLATORS: GPU refers to a Graphics Processing Unit, e.g.
+			 * the "video card" */
+			return g_strdup_printf(_("%s GPU Update"), name);
```
superm1, line 823: is this for VBIOS?  or what firmware is the GPU exactly updating?  Because GPUs tend to load firmware from runtime too.

hughsie: This is the persistent firmware, a.k.a. FW_DATA and FW_CODE -- I needed this for the Intel Arc firmware on the LVFS.

superm1 (approved):

PR #5300, opened 2022-12-07
wistron-dock: Add a new plugin to update several dock models

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@a999153 @bradyc9449

plugins/wistron-dock/fu-wistron-dock-device.c
```diff
@@ -0,0 +1,944 @@
...
+	FuWistronDockDevice *self = FU_WISTRON_DOCK_DEVICE(device);
+	fu_string_append_kx(str, idt, "Status", self->status);
+	fu_string_append_kx(str, idt, "ImgMode", self->imgmode);
+	fu_string_append(str, idt, "IcpBbInfo", self->icp_bbinfo);
+	fu_string_append(str, idt, "IcpUserInfo", self->icp_userinfo);
+}
+
+typedef struct {
+	guint8 *cmd;
+	gsize cmdsz;
```
superm1, line 86: aren't all the commands just a byte?

plugins/wistron-dock/fu-wistron-dock-device.c
```diff
@@ -0,0 +1,944 @@
...
+
+	/* the user has to do something */
+	fwupd_request_set_kind(request, FWUPD_REQUEST_KIND_IMMEDIATE);
+	fwupd_request_set_id(request, FWUPD_REQUEST_ID_REMOVE_REPLUG);
+	fwupd_request_add_flag(request, FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGE);
+	fwupd_request_set_message(request,
+				  "The update will continue when the device USB cable has been "
+				  "unplugged and then re-inserted.");
+	fu_device_emit_request(device, request);
+	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
```
superm1, line 896: If this is the behavior it uses; then I think you should be adding the flag that it's usable while it updates.

plugins/wistron-dock/fu-wistron-dock-device.c
```diff
@@ -0,0 +1,944 @@
...
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_BUSY, 43, "reload");
+}
+
+static void
+fu_wistron_dock_device_init(FuWistronDockDevice *self)
+{
+	fu_device_add_protocol(FU_DEVICE(self), "com.wistron.dock");
+	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_QUAD);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	//	fu_usb_device_add_interface(FU_USB_DEVICE(self), FU_WISTRON_DOCK_USB_INTERFACE);
```
superm1, line 919: ?

plugins/wistron-dock/fu-wistron-dock-device.c
```diff
@@ -0,0 +1,944 @@
...
+		const gchar *fn = fu_firmware_get_id(img);
+		if (g_str_has_suffix(fn, ".wdfl.sig")) {
+			if (fw_wdfl_sig != NULL) {
+				g_set_error_literal(error,
+						    G_IO_ERROR,
+						    G_IO_ERROR_INVALID_DATA,
+						    "multiple WDFL signatures found in archive");
+				return NULL;
+			}
+			if (fu_firmware_get_size(img) != FU_WISTRON_DOCK_WDFL_SIG_SIZE) {
```
superm1, line 381: As a future proofing - perhaps this shouldn't be checked.  It would allow newer (longer) detached signature types in the future.

plugins/wistron-dock/wistron-dock.quirk
```diff
@@ -0,0 +1,8 @@
+# Wistron Travel Dock
+[USB\VID_0FB8&PID_0010]
+Plugin = wistron_dock
+
+# Wistron USB-C Dock
+[USB\VID_17EF&PID_30EF]
```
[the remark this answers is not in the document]
hughsie: Ya, it's a future _maybe Lenovo_ product we can't name yet :)

plugins/wistron-dock/fu-wistron-dock-device.c
hughsie: Indeed, good catch, fixed.

plugins/wistron-dock/fu-wistron-dock-device.c
hughsie: Good idea. I've changed it to `<` as any newer version isn't going to be any smaller.

plugins/wistron-dock/wistron-dock.quirk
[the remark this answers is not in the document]
superm1: The README should probably changed to something along the lines of multiple vendors can be supported, or something more generic.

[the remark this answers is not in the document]
hughsie: Maybe add something like `The wistron-dock protocol may be used by docks branded as other OEMs.`?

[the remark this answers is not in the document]
superm1: I was mostly fixating on the USB vendor ID bit.
But yeah a statement like that should be good too.

plugins/wistron-dock/fu-wistron-dock-device.c
hughsie: Removed as not required.

PR #5310, opened 2022-12-09
Do not filter in fwupdtool for ->attach() if specifying --force

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (dismissed):
This makes sense but I think it probably needs to be documented somewhere too doesn't it?

hughsie:
In fwupdtool man page perhaps?

superm1:
> In fwupdtool man page perhaps?

I couldn't remember where we documented what --force can do. So wherever it is.

And if it's missing add it :)

PR #5321, opened 2022-12-12
Remove data only useful for ->probe() when the device has been added

hughsie:
This might include snapshot data like GUdevDevice that is only very rarely useful after ->probe has been completed.

This reduces the idle RSS by ~200kB on my laptop.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-udev-device.c
+++ b/libfwupdplugin/fu-udev-device.c
@@ -36,6 +36,7 @@
 
 typedef struct {
 	GUdevDevice *udev_device;
+	gboolean udev_device_cleared;
 	guint32 class;
 	guint16 vendor;
 	guint16 model;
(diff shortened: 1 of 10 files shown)


libfwupdplugin/fu-udev-device.c
```diff
@@ -697,6 +698,18 @@ fu_udev_device_get_slot_depth(FuUdevDevice *self, const gchar *subsystem)
...
 
+static void
+fu_udev_device_probe_complete(FuDevice *device)
+{
+	FuUdevDevice *self = FU_UDEV_DEVICE(device);
+	FuUdevDevicePrivate *priv = GET_PRIVATE(self);
+
+	if (priv->udev_device != NULL) {
+		priv->udev_device_cleared = TRUE;
+		g_clear_object(&priv->udev_device);
```
superm1, line 709: how much resource are we really talking about here?  I would think it's not a worthwhile trade off.

hughsie: 200kB -- about 3% of the idle RSS of fwupd on my system...

superm1: Okay that's more than I expected. I guess it's worth it.

superm1 (dismissed):

hughsie:
Don't merge this yet; there is a potential for regression and I want to test it with a ton of HW first.

hughsie:
Maybe making this opt-in would be more sensible?

superm1:
> Maybe making this opt-in would be more sensible?

With the critical message perhaps more clairvoyance on the right direction will be found.  I'm fine either way.

hughsie:
> perhaps more clairvoyance on the right direction

I've switched this to just warning, and we can flip the switch after a few months of grace. I'm a lot more comfortable with this approach.

hughsie:
@superm1 for .11 or punt?

superm1 (commented):
this i would think punt, risk for probing

PR #5324, opened 2022-12-13
Add a 'get-report-metadata' command to allow debugging metadata collection

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -22,6 +22,7 @@ _fwupdtool_cmd_list=(
 	'get-history'
 	'get-plugins'
 	'get-remotes'
+	'get-report-metadata'
 	'get-topology'
 	'hwids'
 	'update'
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -24,6 +24,7 @@
 #include "fwupd-bios-setting-private.h"
 #include "fwupd-common-private.h"
 #include "fwupd-device-private.h"
+#include "fwupd-enums-private.h"
 #include "fwupd-plugin-private.h"
 
 #include "fu-bios-settings-private.h"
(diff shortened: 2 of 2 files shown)


data/bash-completion/fwupdtool
```diff
@@ -22,6 +22,7 @@ _fwupdtool_cmd_list=(
 	'get-history'
 	'get-plugins'
 	'get-remotes'
+	'get-report-metadata'
```
superm1, line 25: should this be get-device-report-metadata maybe since it's focused on one device?

superm1: Yeah that's exactly my worry.

hughsie: `get-device-report-metadata` breaks the `get-dev<tab>` autocomplete for me -- what about just `device-metadata` or `device-report-metadata`? Or `report-metadata-device` perhaps?

superm1: Or how about you just build a report for all metadata instead of just one device?

hughsie: Actually that's not insane; most of the time I'm using --plugins to filter anyway.

hughsie: Something like ^^?

superm1 (dismissed):

src/fu-tool.c
```diff
@@ -3712,6 +3789,13 @@ main(int argc, char *argv[])
 			      /* TRANSLATORS: command description */
 			      _("Attach to firmware mode"),
 			      fu_util_attach);
+	fu_util_cmd_array_add(cmd_array,
+			      "get-report-metadata",
+			      /* TRANSLATORS: command argument: uppercase, spaces->dashes */
+			      _("DEVICE-ID|GUID"),
```
superm1, line 3795: this is no long necessary

hughsie: Fixed, thanks.

PR #5326, opened 2022-12-13
Remove HSI-5 as a level

hughsie:
We have no tools for detecting out-of-band attestation, and no tests that would fit a higher level than 4. It's easier to explain as a '5 level scale' that is logically numbered from 0 to 4.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/docs/hsi.md.in
+++ b/docs/hsi.md.in
@@ -107,14 +107,8 @@ This security level corresponds to out-of-band protection of the system firmware
 
 The system is in a robust secure state.
 
-The system is corresponding several kind of encryption and execution protection for the system firmware.
-
-<a id="hsi-level5"></a>
-
-### [HSI:5 (Secure Proven State)](#hsi-level5)
-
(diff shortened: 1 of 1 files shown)


superm1:
You don't think we're ever going to get out of band attestation?

hughsie:
I don't think we're going to be able to detect it from the host. Discussion welcome, I've not got my heart set yet.

superm1:
> I don't think we're going to be able to detect it from the host. Discussion welcome, I've not got my heart set yet.

I mean I had pie in the sky envisioned that we could connect to a remote attestation server that communicated with the client out of band.  TBH I haven't given serious thought the architecture.

But for example it could be working with a BMC or something like a DASH feature on a WLAN card that connects to the PSP for certain reporting.

hughsie:
@superm1 do we think merge this or close this? I guess HSI-5 is nice to *aspire to* but maybe not so useful to measure.

superm1:
I think we should close it.

Issue #5314, opened 2022-12-12
Thunderbolt being disabled causes IOMMU device protection alert


BryanQuigley:
**Describe the bug**

When running sudo fwupdmgr security:
2022-12-12 06:52:59:  ✔ IOMMU device protection enabled - Thunderbolt is enabled in BIOS.
2022-11-16 02:40:06:  ✘ IOMMU device protection disabled - Thunderbolt is disabled in BIOS.

This seems to encourage me to enable Thunderbolt to get IOMMU protection even though I don't think it's needed without Thunderbolt.

**Steps to Reproduce**

1. Set Thunderbolt to disabled in BIOS
2. Note warning that IOMMU device protection is disabled.

**Expected behavior**

IOMMU device protection should only show up if it would provide some protection.  Disabling unnecessary capabilities should generally be considered more secure.

Obviously there are other possibilities:
 * The BIOS ties together IOMMU and Thunderbolt even though IOMMU protection does help with other threats.
 * The kernel ties together IOMMU and Thunderbolt even though IOMMU protection does help with other threats.
 * Thunderbolt still poses a DMA threat even when disabled.
 * I'm mistaken in some of my assumptions

**fwupd version information**

```shell
runtime   org.freedesktop.fwupd         1.8.7
runtime   org.freedesktop.fwupd-efi     1.3
runtime   com.dell.libsmbios            2.4
runtime   com.hughsie.libjcat           0.1.12
runtime   org.kernel                    6.0.12-300.fc37.x86_64
compile   com.hughsie.libjcat           0.1.12
compile   org.freedesktop.gusb          0.4.1
compile   org.freedesktop.fwupd         1.8.7
runtime   org.freedesktop.gusb          0.4.2
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc):

Fedora default / dnf.

<summary>**fwupd device information**</summary>

```shell
fwupdmgr get-devices --show-all-devices
LENOVO 20RYCTO1WW
│
├─USB2.0 Hub:
│     Device ID:          301046452a49d84af6356d23e43a684b8f10660f
│     Summary:            USB 3.x hub
│     Current version:    4.124
│     Vendor:             VIA Labs, Inc. (USB:0x17EF)
│     Install Duration:   15 seconds
│     GUIDs:              fd36ff26-79e6-51bc-a9f2-558b7fe9ee34 ← USB\VID_17EF&PID_7229
[...]

PR #5331, opened 2022-12-14
iommu: Suggest to turn on `ThunderboltAccess` for Lenovo systems

superm1:
Lenovo systems that offer a BIOS setting for `ThunderboltAccess` will use this option to control whether the IOMMU is enabled by default or not.

It may be counter-intuitive; but as there are other more physically difficult to attack PCIe devices it's better to have the IOMMU enabled pre-boot even if it enables access to Thunderbolt/USB4.

Fixes: #5314

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Can we put some of that rationale into the code too please.

superm1:
> Can we put some of that rationale into the code too please.

Do you mean a code comment? Or do you mean a message someone sees?

hughsie:
comment, sorry.

hughsie (approved):

PR #5332, opened 2022-12-14
Add X-Dock and X-UsbDock update categories

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -3978,6 +3978,10 @@ fu_common_device_category_to_name(const gchar *cat)
 		return "USB Receiver";
 	if (g_strcmp0(cat, "X-Gpu") == 0)
 		return "GPU";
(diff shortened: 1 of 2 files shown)


superm1 (commented):
Wait, what are we using for the other docks already supported?

hughsie:
Either X-Device or X-Controller

hughsie:
e.g. ` Lenovo ThinkPad Universal USB-C Dock Device Update` :/

superm1:
That's too bad :/
I guess the new categories make sense but it would be nice to fix those up too.

hughsie:
> but it would be nice to fix those up too.

Yes, I could do a LVFS fsck for sure. I have to be a little careful as some dock updates only target one part of it, and X-Device is probably fine for that. If it's "one component" for the dock _or composite_ then I think X-UsbDock makes sense.

PR #5336, opened 2022-12-14
Fix the regression to find all device GUID

hughsie:
This regressed in 000e986dc8cd0c7e49bdb5d177dcfc91749351aa but we did not notice because devices with only one GUID (e.g. ESRT devices) still worked as before. Fix the error handling to only abort if none of the GUIDs are found.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
> we did not notice because

And because it's only a few days old!

superm1 (approved):

PR #5341, opened 2022-12-16
uefi-dbx: Ignore failure to mount the ESP if unsupported

hughsie:
I think this is only possible when a raw partition with no filesystem has the ESP or BDP UEFI GUID set in the partition table.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-dbx/fu-uefi-dbx-common.c
+++ b/plugins/uefi-dbx/fu-uefi-dbx-common.c
@@ -89,9 +89,16 @@ fu_uefi_dbx_signature_list_validate(FuContext *ctx, FuEfiSignatureList *siglist,
 	for (guint i = 0; i < volumes->len; i++) {
 		FuVolume *esp = g_ptr_array_index(volumes, i);
 		g_autoptr(FuDeviceLocker) locker = NULL;
-		locker = fu_volume_locker(esp, error);
-		if (locker == NULL)
-			return FALSE;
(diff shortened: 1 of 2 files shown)


plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -89,9 +89,16 @@ fu_uefi_dbx_signature_list_validate(FuContext *ctx, FuEfiSignatureList *siglist,
...
-		if (locker == NULL)
-			return FALSE;
+		g_autoptr(GError) error_local = NULL;
+		locker = fu_volume_locker(esp, &error_local);
+		if (locker == NULL) {
+			if (!g_error_matches(error_local, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED)) {
+				g_propagate_error(error, g_steal_pointer(&error_local));
+				return FALSE;
+			}
+			g_debug("failed to mount ESP: %s", error_local->message);
```
superm1, line 99: This is a pretty uncommon situation; shouldn't this be a warning?

superm1 (approved):

plugins/uefi-dbx/fu-uefi-dbx-common.c
hughsie: Yup, agreed. I've upgraded to a g_warning().

PR #5342, opened 2022-12-16
Add fu_archive_firmware_get_image_fnmatch() for future use

hughsie:
It's useful to get images from archives by a specific filename extension.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
I expect CI to explode due to missing `zip` -- other ideas welcome for a different archive format that we can use without adding deps.

superm1:
How about build it with python and gzip? I think that's more common

hughsie:
> build it with python

Good idea! Fixed.

superm1 (approved):

PR #5349, opened 2022-12-19
Make no-conffile behavior more expected

superm1:
Right now when someone removes the `daemon.conf` conffile for whatever reason, we end up with the test and test_ble plugins enabled and also some defaults don't match.

Adjust the handling of this file such that:

    1) All keys are configured with their default value.
    2) Commenting out a key should keep the default value.
    3) Modifying a key sets a new value.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/daemon.conf
+++ b/data/daemon.conf
@@ -60,11 +60,16 @@ TrustedUids=
 # config level. e.g. `vendor-factory-2021q1`
 HostBkc=
 
-# The EFI system partition (ESP) path used if UDisks is not available
-# or if this partition is not mounted at /boot/efi, /boot, or /efi
-#EspLocation=
+# Override the location used for the EFI system partition (ESP) path.
+# This is typically used if UDisks is not available, or was not able to
+# automatically identify the location for any reason.
+EspLocation=
 
-# these are only required when the SMBIOS or Device Tree data is invalid or missing
+# Override values for SMBIOS / Device Tree data on a system.
+# These are only required when the SMBIOS or Device Tree data is invalid,
+# missing, or to simulate running on another system.
+# * Uncomment them to populate blank entries
+# * Uncomment them and add values to populate specific entries
 #Manufacturer=
 #ProductName=
 #ProductSku=
--- a/src/fu-config.c
+++ b/src/fu-config.c
@@ -192,7 +192,7 @@ fu_config_reload(FuConfig *self, GError **error)
 	}
 
 	/* get idle timeout */
-	idle_timeout = g_key_file_get_int64(keyfile, "fwupd", "IdleTimeout", &error_timeout);
+	idle_timeout = g_key_file_get_uint64(keyfile, "fwupd", "IdleTimeout", &error_timeout);
 	if (idle_timeout > 0)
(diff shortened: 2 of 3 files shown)


data/daemon.conf
```diff
@@ -62,13 +62,13 @@ HostBkc=
...
-#EnclosureKind=
-#BaseboardProduct=
-#BaseboardManufacturer=
+Manufacturer=
+ProductName=
+ProductSku=
+Family=
+EnclosureKind=
+BaseboardProduct=
+BaseboardManufacturer=
```
superm1, line 74: this might need an associated code change I'm realizing; let me double check it.

superm1: added matching code change for it.

src/fu-config.c
```diff
@@ -188,6 +194,8 @@ fu_config_reload(FuConfig *self, GError **error)
 	idle_timeout = g_key_file_get_uint64(keyfile, "fwupd", "IdleTimeout", NULL);
 	if (idle_timeout > 0)
 		self->idle_timeout = idle_timeout;
+	else
+		self->idle_timeout = 7200;
```
superm1, line 198: hmm, this might break the no timeout case.  I'll add more code for it too.

superm1: fixed

libfwupdplugin/fu-hwids.c
```diff
@@ -379,7 +379,7 @@ fu_hwids_setup_overrides(FuHwids *self, GError **error)
 	for (guint i = 0; i < keys->len; i++) {
 		const gchar *key = g_ptr_array_index(keys, i);
 		g_autofree gchar *value = g_key_file_get_string(kf, "fwupd", key, NULL);
-		if (value != NULL)
+		if (value != NULL && strlen(value))
 			fu_hwids_add_smbios_override(self, key, value);
```
hughsie, line 383: how do we set keys as zero length strings? e.g. setting the Family to `""`?

superm1: I thought about that; but is that something that we often need to do though?

hughsie:
Maybe we can think a bit bigger? e.g. can we use GSettings as a system daemon?

superm1:
> Maybe we can think a bit bigger? e.g. can we use GSettings as a system daemon?

I think that it's nice to be able to have a flat conffile actually.  It makes it a lot easier for non GNOME to stomach too..

libfwupdplugin/fu-hwids.c
superm1: OK, I guess these can't be done that way then.  I'll revert this part of the series.

data/daemon.conf
```diff
@@ -62,9 +62,13 @@ HostBkc=
 
 # The EFI system partition (ESP) path used if UDisks is not available
 # or if this partition is not mounted at /boot/efi, /boot, or /efi
-#EspLocation=
+EspLocation=
```
hughsie, line 65: Maybe add a comment here? Then lgtm.

superm1: Added

PR #5351, opened 2022-12-20
trivial: Add new convenience API to make plugins simpler

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1 (commented):
What do you think about including the word "from" in the new helper func names?

Ie `set_version_from_raw16`

hughsie:
Maybe; what about counter-counter-proposal `fu_device_set_version_from_uint16`

superm1:
> Maybe; what about counter-counter-proposal `fu_device_set_version_from_uint16`

Sure

hughsie:
Updated^^

superm1 (approved):

PR #5352, opened 2022-12-20
Skip the fwupdx64.efi BootXXXX entry when measuring system integrity

hughsie:
This is created before the UEFI capsule plugin adds the new BootXXXX entry on NVRAM systems.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
I think you can make the same argument about modifications to BootNext no?

hughsie:
You can, but we don't record BootNext :)

superm1 (approved):

PR #5353, opened 2022-12-20
mtd-recovery: Add a plugin that knows how to MUX and load an MTD device

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/mtd-recovery/README.md
+++ b/plugins/mtd-recovery/README.md
@@ -1,0 +1,64 @@
+---
+title: Plugin: MTD Recovery
+---
+
+This plugin allows a system without an exported MTD device to be updated by matching the DMI
+details of the platform. This allows updating SPI devices that are hidden behind a MUX which is
+very common when running fwupd on a BMC.
+
+This plugin supports the following protocol ID:
+
+* `org.infradead.mtd`
+
+## GUID Generation
+
+These devices are generated from the baseboard vendor ID:
+
+* `MTD\VEN_ASPEED&DEV_AST2500EVB`
+
+## Quirk Use
+
+This plugin uses the following plugin-specific quirks:
+
+### MtdRecoveryGpioNumber
+
+Optional GPIO device number to use as a MUX, e.g. `123`.
+
+Since: 1.8.9
+
+### MtdRecoveryKernelDriver
+
+Kernel driver name, e.g. `npcm-fiu`.
+
+Since: 1.8.9
+
+### MtdRecoveryBindId
+
+Optional kernel ID for binding the device, e.g. `10000000.spi`.
+
+Since: 1.8.9
+
+## Update Behavior
+
+On detach the platform driver for the MTD device is loaded, optionally also using a GPIO MUX.
+The discovered MTD device is used as a proxy and the image it written, then on detach the MUX is
+optionally disabled and the MTD device unloaded.
+
+To test this plugin you can use a quirk file with the following contents:
+
+    # a GUID from `fwupdtool hwids`
+    [536c0c4c-e739-5bc7-8f93-042c73d961ef]
+    Plugin = mtd_recovery
+
+    [MTD\VEN_LENOVO&DEV_20TJS2F44A]
+    MtdRecoveryGpioNumber = 123
+    MtdRecoveryKernelDriver = modulename
+    MtdRecoveryBindId = device.spi
+
+## Vendor ID Security
+
+The vendor ID is set from the baseboard vendor, for example `DMI:ASPEED`
+
+## External Interface Access
+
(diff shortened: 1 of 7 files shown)


plugins/mtd-recovery/README.md
```diff
@@ -0,0 +1,63 @@
...
+This plugin supports the following protocol ID:
+
+* org.infradead.mtd
+
+## GUID Generation
+
+These devices are generated from the baseboard vendor ID:
+
+* `MTD\VEN_Lenovo&DEV_ThinkPad`
+* `MTD\VEN_Lenovo`
```
superm1, line 16: If the most common case is from behind a BMC then I think this is very unlikely to ever show up in a laptop...  Maybe a different example from a server makes sense.

plugins/mtd-recovery/fu-mtd-recovery-device.c
```diff
@@ -0,0 +1,263 @@
...
+{
+	FuMtdRecoveryDevice *self = FU_MTD_RECOVERY_DEVICE(device);
+	g_autoptr(FuUdevDevice) device_tmp = NULL;
+
+	/* already available */
+	if (fu_device_get_proxy(device) != NULL)
+		return TRUE;
+
+	/* create GPIO device and disable write lock */
+	if (self->gpio_number != 0) {
```
superm1, line 133: why can't it be GPIO 0?

hughsie: Agreed, fixed.

hughsie:
> Is this to enable a "new" design or an "existing" design?

In this case, an existing, but non-public design deployed all over the place.

>  it's begging a spec to specify in the ACPI/DT tables that a certain GPIO is connected to a mux that can do this.

Yes, this is what I'm lobbying for too. Another security company wants to know the same information and it's pretty crap that we'd have to hardcode it per-model, perhaps in more than one project.

> presumably other things can potentially be exposed from a mux controlled by GPIO?

Hmm, perhaps -- you mean like a USB device?

> Swapping the GPIO to expose the MTD device has to have some implication for whatever is not connected anymore...

Agreed, also a concern. Apparently the "host OS" gets writes denied, i.e. a kernel write error rather than anything electrically bad. I believe in the demo hardware that we were developing against the host OS has to be powered down, although I'm not sure of the mechanism that enforces this.

superm1:
> have to hardcode it per-model, perhaps in more than one project.

What other projects?  If it's coreboot or an open source BSP firmware then it would be better to potentially introduce the mechanism in that other project to introspect what GPIO it is (advertise to a table, a mailbox message, _anything really_)

> Hmm, perhaps -- you mean like a USB device?

Yeah I mean I don't know the details of what actually happens when you trigger this GPIO.  If it's really a mux then it has something connected when set one way and something else when connected another way.  If it's not really a mux, it might be toggling a write protect pin for something.

> Apparently the "host OS" gets writes denied, i.e. a kernel write error rather than anything electrically bad

Which jives more to the GPIO being on a _write protect pin not a mux_. 

> host OS has to be powered down, although I'm not sure of the mechanism that enforces this.

Eek!

hughsie:
> What other projects?

In this example it'd be some kind of security firmware agent, so not the kind of thing we expect to be installed by default. I think this kind of data really belongs in DT or ACPI somewhere. That's what I'm advocating for future hardware.

>  open source BSP firmware

I'll investigate what OpenBMC could provide; the medium term plan is for OpenBMC to *use* fwupd to update stuff, so I'm kinda expecting *not our problem, it's yours!* :)

> If it's really a mux then it has something connected when set one way and something else when connected another way. 

This is what I was lead to believe.

> it might be toggling a write protect pin for something.

Nah, without the GPIO set the kernel driver just refuses to load, so I actually think it's controlling the clock and data.

> Eek!

Yes, I should dive more into this -- e.g. setting an inhibit on the device if the host is powered on. That might be phase #2.

superm1:
> Nah, without the GPIO set the kernel driver just refuses to load, so I actually think it's controlling the clock and data.

I bet it's a mux that pulls the clock into a pull down or clock to a real clock pin then.  

> Yes, I should dive more into this -- e.g. setting an inhibit on the device if the host is powered on. That might be phase 2

Yeah, I hope that there can be an indication from the BMC to make this design safe.

plugins/mtd-recovery/README.md
hughsie: Yup, changed.

hughsie:
Lets close this PR but keep the branch -- if the vendor has interest in doing this we can ask them to make it "just work" at the same time.

PR #5358, opened 2022-12-23
Wip/hughsie/meson deprecated

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,7 +39,7 @@ jobs:
       run: |
         mkdir -p $GITHUB_WORKSPACE/build
         cd $GITHUB_WORKSPACE/build
-        meson .. -Dman=false -Defi_binary=false -Dplugin_uefi_capsule_splash=false --prefix=$GITHUB_WORKSPACE/dist
+        meson setup .. -Dman=false -Defi_binary=false -Dplugin_uefi_capsule_splash=false --prefix=$GITHUB_WORKSPACE/dist
         ninja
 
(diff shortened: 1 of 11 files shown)


hughsie:
Also relevant: https://github.com/mesonbuild/meson/issues/9300

.github/workflows/codeql-analysis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       run: |
         mkdir -p $GITHUB_WORKSPACE/build
         cd $GITHUB_WORKSPACE/build
-        meson .. -Dman=false -Defi_binary=false -Dplugin_uefi_capsule_splash=false --prefix=$GITHUB_WORKSPACE/dist
+        meson setup .. -Dman=false -Defi_binary=false -Dplugin_uefi_capsule_splash=false --prefix=$GITHUB_WORKSPACE/dist
```
superm1, line 42: While touching those lines can we perhaps also drop the efi binary and uedi capsule splash stuff? I would think meson features does that for us now.

meson.build
```diff
@@ -435,7 +435,7 @@ if build_standalone and efiboot.found() and efivar.found()
   endif
 
   if get_option('plugin_uefi_capsule_splash')
-    r = run_command([python3, 'po/test-deps'])
+    r = run_command([python3, 'po/test-deps'], check: false)
```
superm1, line 438: If you make this check true can probably drop the next few lines too.

superm1 (approved):

PR #5366, opened 2022-12-29
dfu: Add some wisdom for vendors implementing devices

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/plugins/dfu/README.md
+++ b/plugins/dfu/README.md
@@ -32,6 +32,42 @@ the device again re-enumerates back to the runtime mode.
 For this reason the `REPLUG_MATCH_GUID` internal device flag is used so that
 the bootloader and runtime modes are treated as the same device.
 
+## Implementation Notes
+
+The runtime mode is just as important as the DFU mode from the point of view of
+fwupd and should be included if firmware updates are to "just work". Without a
+DFU runtime interface we can match the device with `Flags = no-dfu-runtime` but
+will need a suitably new fwupd version before the device is recognized.
+
+The USB interface revision (`REV`) is used as the BCD version number, as DFU has
+no way of representing a firmware version number. A new firmware version should
+always increment the USB REV of the *runtime* interface as fwupd will **not** switch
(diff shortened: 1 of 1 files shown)


plugins/dfu/README.md
```diff
@@ -32,6 +32,42 @@ the device again re-enumerates back to the runtime mode.
 For this reason the `REPLUG_MATCH_GUID` internal device flag is used so that
 the bootloader and runtime modes are treated as the same device.
 
+## Implementation Notes
```
superm1, line 35: As a general note; wouldn't this make more sense to end up on fwupd.github.io?  I know we don't really do a good job segementing documentation between README.md and what ends up there, but I think this is the type of thing we would want to end up there.  Maybe we even want to split up the `README.md` across all plugins to automagically build a table of information on fwupd.github.io.

In an ideal world I think someone should only need to go there to find out info and it would link out to other documentation at fwupd.org and readthedocs etc.

plugins/dfu/README.md
```diff
@@ -32,6 +32,42 @@ the device again re-enumerates back to the runtime mode.
...
+no way of representing a firmware version number. A new firmware version should
+always increment the USB REV of the *runtime* interface as fwupd will **not** switch
+the device into *DFU mode* during enumeration to read the version number.
+The version number of the DFU mode should represent the *bootloader version* and
+this should not change as the firmware is updated.
+
+The runtime USB interface should have a unique vendor ID and product ID for the
+specific firmware stream. A different version of software should have a unique
+VID/PID USB descriptor pair. The microcontroller example VID/PID should **never**
+be used in the runtime mode.
```
superm1, line 52: maybe add a comment here "because ...."

hughsie: Agreed, added.

plugins/dfu/README.md
hughsie: Agreed, but in a different pull request :) https://github.com/fwupd/fwupd/pull/5393

PR #5369, opened 2023-01-03
mtd: Use dummy SMBIOS data to fix self test on s390x

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/linux-swap/meson.build
+++ b/plugins/linux-swap/meson.build
@@ -14,6 +14,9 @@ plugin_builtin_linux_swap = static_library('fu_plugin_linux_swap',
 plugin_builtins += plugin_builtin_linux_swap
 
 if get_option('tests')
+  env = environment()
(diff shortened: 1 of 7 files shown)


superm1:
Does s390x even support `mtd` devices?  Maybe we should just skip it.  Also - would really love if we could get s390x building in CI again :/

hughsie:
tbh, it's no s390 specific at all, it's just the only architecture we test that doesn't have either SMBIOS or DT exposed on the runner.

superm1:
> 

Ah got it.  Then yeah this approach makes sense.  Really we probably should be setting that for all the test suites to give a known environment.

hughsie:
>  Really we probably should be setting that for all the test suites

Agreed, added.

superm1 (approved):

Issue #5370, opened 2023-01-04
1.8.9: test suite is failing in `linux-swap-self-test` and `redfish-self-test` units


kloczek:
**Describe the bug**
Looks like new version is failing in two units.

**Steps to Reproduce**
meson settings:

```spec
%meson \
        -D bluez=enabled \
        -D build=all \
        -D consolekit=enabled \
        -D curl=enabled \
        -D docs=enabled \
        -D firmware-packager=true \
        -D gudev=enabled \
        -D gusb=enabled \
        -D introspection=enabled \
        -D libarchive=enabled \
        -D lvfs=true \
        -D lzma=enabled \
        -D man=true \
        -D plugin_acpi_phat=enabled \
        -D plugin_bcm57xx=enabled \
        -D plugin_cfu=enabled \
        -D plugin_cpu=enabled \
        -D plugin_dell=enabled \
        -D plugin_dummy=enabled \
        -D plugin_dummy=true \
        -D plugin_emmc=enabled \
        -D plugin_ep963x=enabled \
        -D plugin_fastboot=enabled \
        -D plugin_flashrom=enabled \
        -D plugin_gpio=enabled \
        -D plugin_igsc=enabled \
        -D plugin_intel_me=enabled \
        -D plugin_intel_spi=enabled \
        -D plugin_intel_spi=true \
        -D plugin_logitech_bulkcontroller=enabled \
        -D plugin_modem_manager=enabled \
        -D plugin_msr=enabled \
        -D plugin_mtd=enabled \
        -D plugin_nitrokey=enabled \
        -D plugin_nvme=enabled \
        -D plugin_parade_lspcon=enabled \
        -D plugin_pixart_rf=enabled \
        -D plugin_powerd=enabled \
        -D plugin_realtek_mst=enabled \
        -D plugin_redfish=enabled \
        -D plugin_scsi=enabled \
        -D plugin_synaptics_mst=enabled \
        -D plugin_synaptics_rmi=enabled \
        -D plugin_tpm=enabled \
        -D plugin_uefi_capsule=enabled \
        -D plugin_uefi_capsule_splash=enabled \
        -D plugin_uefi_capsule_splash=true \
        -D plugin_uefi_pk=enabled \
        -D plugin_uf2=enabled \
        -D plugin_upower=enabled \
        -D polkit=enabled \
        -D soup_session_compat=true \
        -D supported_build=enabled \
        -D systemd=enabled \
        -D tests=true \
        %{nil}
```
[...]

PR #5371, opened 2023-01-04
trivial: skip self tests that fail with G_IO_ERROR_NOT_DIRECTORY

superm1:
fixes: #5370

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/linux-swap/fu-self-test.c
+++ b/plugins/linux-swap/fu-self-test.c
@@ -35,6 +35,7 @@ fu_linux_swap_plain_func(void)
 			      0,
 			      &error);
 	if (g_error_matches(error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND) ||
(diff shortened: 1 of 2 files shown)


hughsie:
This seems like it's working around a GLib bug; shouldn't the GDBus error domain/code be returned in this case?

superm1:
We also skip some other `G_IO_ERROR`s too already.  Maybe should just combine with those.
```
        if (g_error_matches(error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND) ||
            g_error_matches(error, G_IO_ERROR, G_IO_ERROR_INVALID_ARGUMENT)) {
```

hughsie (approved):

kloczek:
> This seems like it's working around a GLib bug; shouldn't the GDBus error domain/code be returned in this case?

May I ask about ref to that bug?

PR #5375, opened 2023-01-04
Do not allow LZX compressed cabinet archives

hughsie:
The GCab decompression code is *scary* low level C that has not had any security auditing other than by the fuzzer (which found plenty of memory-safety bugs).

Supporting LZX is useless at best, and could be a security exploit at worst.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
Depends on https://gitlab.gnome.org/GNOME/gcab/-/merge_requests/15

superm1 (approved):

PR #5389, opened 2023-01-09
emmc: Add back a legacy GUID to fix a regression

hughsie:
I had no idea anyone was using this, so apologies. The instance ID is completely wrong, but it is effectively D-Bus ABI and we need to support old firmware in the field.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Should also go into readme

hughsie:
> Should also go into readme

Ack, fixed.

superm1 (approved):

PR #5390, opened 2023-01-09
synaptics-rmi: add support to write signature

vhuag:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation
When the firmware partition contains signature, firmware would require an additional step to write signature.

### DIFF
--- a/plugins/synaptics-rmi/fu-synaptics-rmi-v7-device.c
+++ b/plugins/synaptics-rmi/fu-synaptics-rmi-v7-device.c
@@ -22,6 +22,7 @@ typedef enum {
 	RMI_FLASH_CMD_ERASE,
 	RMI_FLASH_CMD_ERASE_AP,
 	RMI_FLASH_CMD_SENSOR_ID,
+	RMI_FLASH_CMD_SIGNATURE,
 } RmiFlashCommand;
 
 typedef enum {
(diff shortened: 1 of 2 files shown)


plugins/synaptics-rmi/fu-synaptics-rmi-v7-device.c
```diff
@@ -331,6 +416,14 @@ fu_synaptics_rmi_v7_device_write_partition(FuSynapticsRmiDevice *self,
 			return FALSE;
 		fu_progress_step_done(progress);
 	}
+	if (!fu_synaptics_rmi_v7_device_write_partition_signature(self,
+								  firmware,
+								  id,
+								  partition_id,
+								  fu_progress_get_child(progress),
```
hughsie, line 423: I don't think this is going to work, I think you need to split up the phases with a new `fu_progres_set_id()` block and some new steps. If you push this branch to the origin remote I can push a fixup if that helps.

vhuag: Hi do I need permission to push to origin?

hughsie: You're in the commit list, so just push something like wip/vhuag/linux-add-signature-20230109

vhuag: Hi I can't push due to permission so I remove the progress tracking inside the signature function, normally the chunks->len is 1.

hughsie (approved):

PR #5401, opened 2023-01-10
Add a Flat Device Tree backend

hughsie:
This is needed because multiple plugins will soon be consuming the system FDT, and we don't want to either parse this multiple times, or duplicate the loading logic.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/vbe/fu-vbe-plugin.c
+++ b/plugins/vbe/fu-vbe-plugin.c
@@ -13,14 +13,12 @@
 
 struct _FuVbePlugin {
 	FuPlugin parent_instance;
-	FuFirmware *fdt;
-	gchar *vbe_dir;
 };
 
 G_DEFINE_TYPE(FuVbePlugin, fu_vbe_plugin, FU_TYPE_PLUGIN)
 
 static gboolean
-fu_vbe_plugin_coldplug_img(FuPlugin *plugin,
+fu_vbe_plugin_coldplug_img(FuVbePlugin *self,
 			   FuFdtImage *fdt_root,
 			   FuFdtImage *fdt_node,
 			   GError **error)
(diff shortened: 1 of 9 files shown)


hughsie:
@superm1 the alternative is that we have a fdt plugin and then hook `->registered` but I don't know if that's any better.

plugins/vbe/fu-vbe-plugin.c
```diff
@@ -79,33 +77,41 @@ fu_vbe_plugin_coldplug_img(FuPlugin *plugin,
...
 	g_autoptr(GPtrArray) fdt_imgs = NULL;
 
+	/* not interesting */
+	if (g_strcmp0(fu_device_get_backend_id(device), "fdt") != 0)
+		return TRUE;
+
 	/* get compatible from root node */
-	fdt_root =
-	    fu_fdt_firmware_get_image_by_path(FU_FDT_FIRMWARE(self->fdt), "/chosen/fwupd", error);
+	fdt = fu_device_read_firmware(device, progress, error);
```
hughsie, line 104: this is a hack to avoid moving the FuFdtDevice to libfwupdplugin and having a `fu_fdt_device_get_firmware()` export -- if you think that's preferable let me know and I can move it.

hughsie: Ooooor, we could just have a fu_context_get_fdt() and return the FuFdtFirmware there; loading it with all the other hwinfo context stuff. The disadvantage with that if you can't do instance ID quirk matches but maybe that's not the end of the world.

hughsie: I like https://github.com/fwupd/fwupd/pull/5407 more

PR #5405, opened 2023-01-11
Respect user requested paths for ESP even if they aren't volumes

superm1:
These paths may be used for bind mounts, which we can't accurately discover. Leave a message in the logs accordingly, but allow these paths.

fixes: #5185

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-volume.c
+++ b/libfwupdplugin/fu-volume.c
@@ -725,6 +725,10 @@ fu_volume_new_esp_for_path(const gchar *esp_path, GError **error)
 		if (g_strcmp0(basename, vol_basename) == 0)
 			return g_object_ref(vol);
 	}
+	if (g_file_test(esp_path, G_FILE_TEST_IS_DIR)) {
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-volume.c
```diff
@@ -725,6 +725,10 @@ fu_volume_new_esp_for_path(const gchar *esp_path, GError **error)
 		if (g_strcmp0(basename, vol_basename) == 0)
 			return g_object_ref(vol);
 	}
+	if (g_file_test(esp_path, G_FILE_TEST_IS_DIR)) {
+		g_info("Using user requested path %s for ESP", esp_path);
```
hughsie, line 729: is the info message required? It seems like if the user explicitly set it we should just use it.

superm1: OK will adjust to debug level.

hughsie (approved):

PR #5409, opened 2023-01-11
Add System76 launch_heavy_1

jackpot51:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation


hughsie:
@jackpot51 for future hardware / firmware you could use a DS20 BOS descriptor to make fwupd magically work without submitting a patch to the quirk file: https://fwupd.github.io/libfwupdplugin/ds20.html -- it basically ships the quirk in the USB descriptor. If you want to keep the descriptor small and simple then updating the quirk file is of course fine too.

jackpot51:
@hughsie that looks awesome! I'll see if we can do that.

PR #5414, opened 2023-01-14
Wip/hughsie/hwids

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 or, if fu-context.c is getting a bit long, maybe we re-introduce fu-hwids.c and then add a fu_context_get_hwid() call that allows us to call `fu_hwid_get_value(hwid)` -- although the indirection kinda sucks. I think this PR is probably about right to avoid big plugin changes, but I'm open for other ideas.

hughsie:
> but I'm open for other ideas

I've respun this with the fu-hwids.c fix and it's a much smaller thing to review. Once https://github.com/fwupd/fwupd/pull/5417 goes in I'll replace this PR.

hughsie:
A less dramatic version in https://github.com/fwupd/fwupd/pull/5418

PR #5417, opened 2023-01-16
trivial: Add FuContextHwidFlags for future usage

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-context.c
+++ b/libfwupdplugin/fu-context.c
@@ -130,6 +130,24 @@ fu_context_get_fdt(FuContext *self, GError **error)
 }
 
 /**
+ * fu_context_get_smbios:
+ * @self: a #FuContext
+ *
+ * Gets the SMBIOS store.
+ *
+ * Returns: (transfer none): a #FuSmbios
+ *
+ * Since: 1.8.10
+ **/
+FuSmbios *
+fu_context_get_smbios(FuContext *self)
+{
+	FuContextPrivate *priv = GET_PRIVATE(self);
+	g_return_val_if_fail(FU_IS_CONTEXT(self), NULL);
+	return priv->smbios;
+}
+
+/**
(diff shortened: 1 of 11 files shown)


plugins/mtd/fu-self-test.c
```diff
@@ -32,7 +32,7 @@ fu_test_mtd_device_func(void)
 	ret = fu_context_load_quirks(ctx, FU_QUIRKS_LOAD_FLAG_NO_CACHE, &error);
 	g_assert_no_error(error);
 	g_assert_true(ret);
-	ret = fu_context_load_hwinfo(ctx, &error);
+	ret = fu_context_load_hwinfo(ctx, FU_CONTEXT_HWID_FLAG_NONE, &error);
```
superm1, line 35: In the self tests don't we actually want it to match our expected environment?   Or is your intention that our environment is "empty"?

hughsie: mtd doesn't care about any specific hwids -- so i figured we should just do as little as possible.

libfwupdplugin/fu-context.c
```diff
@@ -703,19 +776,28 @@ fu_context_security_changed(FuContext *self)
  * Since: 1.6.0
```
superm1, line 776: this should be bumped up as we're (essentially) changing ABI for this symbol now.

hughsie: Agree, fixing.

plugins/mtd/fu-self-test.c
superm1: But that's how SMBIOS tables get loaded, and how we fixed the recent error on s390x because it didn't have them.   So perhaps we should back some of that out now?

Issue #5424, opened 2023-01-18
grub2-probe No path or device is specified.


hramrach:
**Describe the bug**
Error printed when generating grub configuration

```
No path or device is specified.
Usage: grub2-probe [OPTION...] [OPTION]... [PATH|DEVICE]
Try 'grub2-probe --help' or 'grub2-probe --usage' for more information.
```

**Steps to Reproduce**
`grub2-mkconfig`

**Expected behavior**
no error

**fwupd version information**

https://github.com/fwupd/fwupd/commit/684bc038132977ff28e54550f95941cff5e4ee84

This line https://github.com/fwupd/fwupd/blob/main/plugins/uefi-capsule/fwupd.grub.conf.in#L17 invokes grub-probe without arguments

OS: openSUSE

hughsie:
@superm1 is it missing a `${ESP}` perhaps?

superm1:
Can I see the generated grub.conf for the distro?

hramrach:
```
#! /bin/sh
# SPDX-License-Identifier: LGPL-2.1+
set -e

[ -d ${pkgdatadir:?} ]
# shellcheck source=/dev/null
. "$pkgdatadir/grub-mkconfig_lib"

if [ -f /var/lib/fwupd/uefi_capsule.conf ] &&
   ls /sys/firmware/efi/efivars/fwupd-*-0abba7dc-e516-4167-bbf5-4d9d1c739416 1>/dev/null 2>&1; then
      . /var/lib/fwupd/uefi_capsule.conf
      if [ "${EFI_PATH}" != "" ] && [ "${ESP}" != "" ]; then
      echo "Adding Linux Firmware Updater entry" >&2
cat << EOF
menuentry 'Linux Firmware Updater' \$menuentry_id_option 'fwupd' {
EOF
      ${grub_probe:?}
      prepare_grub_to_access_device '`${grub_probe} --target=device \${ESP}` | sed -e "s/^/\t/"'
cat << EOF
	chainloader ${EFI_PATH}
}
EOF
      fi
fi

```

hramrach:
hm, this version works, there must be something different on the broken machine

hramrach:
I see, the working machine does not have `/var/lib/fwupd/uefi_capsule.conf` which skips execution the broken code.

bzoltan1:
Here is the full grub2-mkconfig output with -x added to the script on the broken machine

`$ sudo grub2-mkconfig
[sudo] password for root: 
[...]

PR #5426, opened 2023-01-18
uefi-capsule: Do not call grub2-probe without arguments

hramrach:
commit 684bc0381 ("trivial: fix various shellcheck warnings") adds call to ${grub_probe} without any argument which causes grub2-probe to return an error, and generating grub configuration fails.

Add --version argument to avoid the error, and redierct the output to /dev/null so that it's not included in the grub configuration file.

The commit also adds superfluous single quotes causing another error:

/usr/sbin/grub2-probe: error: cannot find a GRUB drive for `${grub_probe} --target=device \${ESP}` | sed -e "s/^/\t/".  Check your device.map.

Remove the single quotes again.

Fixes: #5424
Fixes: 684bc0381 ("trivial: fix various shellcheck warnings")
Signed-off-by: Michal Suchanek <msuchanek@suse.de>

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fwupd.grub.conf.in
+++ b/plugins/uefi-capsule/fwupd.grub.conf.in
@@ -14,8 +14,8 @@ if [ -f @localstatedir@/lib/fwupd/uefi_capsule.conf ] &&
 cat << EOF
 menuentry 'Linux Firmware Updater' \$menuentry_id_option 'fwupd' {
 EOF
-      ${grub_probe:?}
-      prepare_grub_to_access_device '`${grub_probe} --target=device \${ESP}` | sed -e "s/^/\t/"'
+      ${grub_probe:?} --version > /dev/null
+      prepare_grub_to_access_device "$(${grub_probe} --target=device ${ESP})" | sed -e "s/^/\t/"
 cat << EOF
 	chainloader ${EFI_PATH}
 }


superm1:
This looks correct, but it looks like shellcheck must waive this file.

```
In plugins/uefi-capsule/fwupd.grub.conf.in line 18:
      prepare_grub_to_access_device `${grub_probe} --target=device \${ESP}` | sed -e "s/^/\t/"
                                    ^-- SC2046: Quote this to prevent word splitting.
```

hramrach:
I think that shellcheck is wrong here. It may or may not be desirable to split word depending on context. Even if you do quote  there is a number of ways you can quote.

superm1:
Yeah, can you please add an [exclusion ](https://[pre-commit](https://pre-commit.com/#config-exclude).com/#config-exclude)to `~/.pre-commit-config.yaml` for this changed file so shellcheck calms down?

hramrach:
I suppose prepare_grub_to_access_device only accepts one device but may accept more in the future if needed.

`${grub_probe} --target=device \${ESP}` should return only one device anyway, and a device name should not include whitespace.

If you really wanted to quote you could do "$(${grub_probe} --target=device \${ESP})"

hramrach:
and definitely not single quotes, and not around the pipe

superm1 (approved):

superm1:
Thanks a lot for fixing it!

PR #5427, opened 2023-01-18
Add initial optional Rust support

hughsie:
Do not get too excited, this is only to see what breaks.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/dependencies.xml
+++ b/contrib/ci/dependencies.xml
@@ -1755,4 +1755,19 @@
       <package variant="mingw64">msitools</package>
     </distro>
   </dependency>
+  <!-- optional -->
+  <dependency id="rustc">
+    <distro id="arch">
+      <package variant="x86_64">rust</package>
+    </distro>
+    <distro id="fedora">
+      <package variant="x86_64" />
+    </distro>
+    <distro id="debian">
+      <package variant="x86_64" />
+    </distro>
+    <distro id="ubuntu">
+      <package variant="x86_64" />
+    </distro>
+  </dependency>
 </dependencies>
--- a/contrib/ci/rustfmt.py
+++ b/contrib/ci/rustfmt.py
@@ -1,0 +1,46 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2023 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+import glob
+import os
+import sys
+import subprocess
+
+
+def __get_license(fn: str) -> str:
+    with open(fn, "r") as f:
+        for line in f.read().split("\n"):
+            if line.find("SPDX-License-Identifier:") > 0:
+                return line.split(":")[1]
+    return ""
+
+
+def test_files() -> int:
+
+    rc: int = 0
+    for fn in glob.glob("**/*.rs", recursive=True):
+        if "meson-private" in fn:
+            continue
+        if "venv" in fn:
+            continue
+        if fn.startswith("subprojects"):
+            continue
+        if fn.startswith("dist"):
+            continue
+        try:
+            subprocess.check_output(["rustfmt", "--quiet", "--check", fn])
+        except FileNotFoundError as e:
+            break
+        except subprocess.CalledProcessError as e:
+            print(str(e))
+            rc = 1
+    return rc
+
+
+if __name__ == "__main__":
+
+    # all done!
+    sys.exit(test_files())
--- a/libfwupdplugin/fu-crc.c
+++ b/libfwupdplugin/fu-crc.c
@@ -23,6 +23,7 @@
  *
  * Since: 1.8.2
  **/
+#ifndef HAVE_RUST
 guint8
 fu_crc8_full(const guint8 *buf, gsize bufsz, guint8 crc_init, guint8 polynomial)
 {
@@ -37,6 +38,7 @@ fu_crc8_full(const guint8 *buf, gsize bufsz, guint8 crc_init, guint8 polynomial)
 	}
 	return ~((guint8)(crc >> 8));
 }
+#endif
 
 /**
  * fu_crc8:
@@ -68,6 +70,7 @@ fu_crc8(const guint8 *buf, gsize bufsz)
  *
  * Since: 1.8.2
  **/
+#ifndef HAVE_RUST
 guint16
 fu_crc16_full(const guint8 *buf, gsize bufsz, guint16 crc, guint16 polynomial)
 {
@@ -83,6 +86,7 @@ fu_crc16_full(const guint8 *buf, gsize bufsz, guint16 crc, guint16 polynomial)
 	}
 	return ~crc;
 }
+#endif
 
 /**
  * fu_crc16:
@@ -114,6 +118,7 @@ fu_crc16(const guint8 *buf, gsize bufsz)
  *
  * Since: 1.8.2
  **/
+#ifndef HAVE_RUST
 guint32
 fu_crc32_full(const guint8 *buf, gsize bufsz, guint32 crc, guint32 polynomial)
 {
@@ -127,6 +132,7 @@ fu_crc32_full(const guint8 *buf, gsize bufsz, guint32 crc, guint32 polynomial)
 	}
 	return ~crc;
 }
+#endif
 
 /**
  * fu_crc32:
--- a/libfwupdplugin/fu-crc.rs
+++ b/libfwupdplugin/fu-crc.rs
@@ -1,0 +1,78 @@
+// Copyright (C) 2022 Richard Hughes <richard@hughsie.com>
+//
+// SPDX-License-Identifier: LGPL-2.1+
+
+use std::slice;
+
+fn crc8_full(buf: &[u8], crc_init: u8, polynomial: u8) -> u8 {
+    let mut crc = crc_init as u32;
+    for tmp in buf.iter() {
+        crc ^= (*tmp as u32) << 8;
+        for _ in 0..8 {
+            if crc & 0x8000 > 0 {
+                crc ^= ((polynomial as u32 | 0x100) << 7) as u32;
+            }
+            crc <<= 1;
+        }
+    }
+    !(crc >> 8) as u8
+}
+
+fn crc16_full(buf: &[u8], crc_init: u16, polynomial: u16) -> u16 {
+    let mut crc = crc_init;
+    for tmp in buf.iter() {
+        crc = crc ^ *tmp as u16;
+        for _ in 0..8 {
+            if crc & 0x1 > 0 {
+                crc = (crc >> 1) ^ polynomial;
+            } else {
+                crc >>= 1;
+            }
+        }
+    }
+    !crc
+}
+
+fn crc32_full(buf: &[u8], crc_init: u32, polynomial: u32) -> u32 {
+    let mut crc = crc_init;
+    for tmp in buf.iter() {
+        crc = crc ^ *tmp as u32;
+        for _ in 0..8 {
+            let mask = {
+                if crc & 0b1 > 0 {
+                    0xFFFFFFFF
+                } else {
+                    0x0
+                }
+            };
+            crc = (crc >> 1) ^ (polynomial & mask);
+        }
+    }
+    !crc
+}
+
+#[no_mangle]
+pub extern "C" fn fu_crc8_full(buf: *mut u8, bufsz: usize, crc_init: u8, polynomial: u8) -> u8 {
+    let slice = unsafe { slice::from_raw_parts(buf, bufsz) };
+    crc8_full(&slice, crc_init, polynomial)
+}
+
+#[no_mangle]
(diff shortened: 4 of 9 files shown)


superm1 (commented):
Something I genuinely don't understand about rust is with distro packages how do you add extra crates?  buildds aren't supposed to have access to the web.  So I feel like we should probably be cautious which crates are used and if at all possible configure the build process for distro packages in CI to fail if a crate isn't already installed rather than try to install it.

meson.build
```diff
@@ -139,6 +139,14 @@ add_project_link_arguments(
   language: 'c'
 )
 
+# rust is optional
+if add_languages('rust', required: false)
+  have_rust = true
+  add_project_arguments('-DHAVE_RUST', language: 'c')
+else
+  have_rust = false
+endif
```
superm1, line 148: do we want this guarded to a `meson_options.txt` option?

libfwupdplugin/meson.build
```diff
@@ -236,6 +236,20 @@ library_deps = [
   platform_deps,
 ]
 
+if have_rust
+  fwupdpluginrs = static_library(
+    'fwupdpluginrs',
+    sources: [
+      'fu-crc.rs',
+    ],
```
superm1, line 244: is it not possible to link the rusty object files directly to fwupdplugin instead of a new static library?

meson.build
superm1: I sorta lean on yes as a meson tri-state auto feature where `required` will be controlled by it.  That will let us test builds with either way.

libfwupdplugin/meson.build
```diff
@@ -236,6 +236,20 @@ library_deps = [
...
+    rust_crate_type: 'staticlib',
+  )
+  rust = import('unstable-rust')
```
superm1, line 247: any minimum meson version we need to consider?

libfwupdplugin/fu-crc.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright (C) 2022 Richard Hughes <richard@hughsie.com>
+//
+// SPDX-License-Identifier: LGPL-2.1+
```
superm1, line 3: do we have clang rules already for rust files?  If not, probably a good time to add them.

libfwupdplugin/fu-crc.rs
```diff
@@ -0,0 +1,85 @@
...
+                crc = (crc >> 1) ^ polynomial;
+            } else {
+                crc >>= 1;
+            }
+        }
+    }
+    !crc
+}
+
+fn crc32_full(buf: &[u8], crc_init: u32, polynomial: u32) -> u32 {
```
superm1, line 37: this is probably missing comments about the function documentation and parameters.  Not sure if the documentation generator will parse the rust files, but another thing worth checking.

libfwupdplugin/fu-crc.rs
```diff
@@ -0,0 +1,85 @@
...
+#[no_mangle]
+pub extern "C" fn fu_crc32_full(buf: *mut u8, bufsz: usize, crc_init: u32, polynomial: u32) -> u32 {
+    let slice = unsafe { slice::from_raw_parts(buf, bufsz) };
+    let ret = crc32_full(&slice, crc_init, polynomial);
+    mem::forget(slice);
+    ret
+}
+
+#[test]
+fn fu_crc32() {
```
superm1, line 80: this should rename, it's testing all 3 of the CRC functions in one run

superm1:
> Something I genuinely don't understand about rust is with distro packages how do you add extra crates?

Answered my own question at least for Debian.
https://wiki.debian.org/Teams/RustPackaging

I think we're still going to have some pain though in getting people setup depending upon what crates we use/depend upon.

contrib/ci/dependencies.xml
```diff
@@ -1754,4 +1754,16 @@
...
+  <!-- optional -->
+  <dependency id="rust">
+    <distro id="fedora">
+      <package variant="x86_64" />
+    </distro>
+    <distro id="debian">
+      <package variant="x86_64" />
+    </distro>
+    <distro id="ubuntu">
+      <package variant="x86_64" />
```
superm1, line 1766: for these I think it's [rustc](https://tracker.debian.org/pkg/rustc) that you're looking for.

hughsie:
> how do you add extra crates

At the moment I'm not planning to use any crates -- all the stuff here is from std.

libfwupdplugin/meson.build
hughsie: I don't *think* so; libfwupdplugin is `library()` which can't have type `staticlib`, and if you leave it out you get `ERROR: Rust target fwupdplugin contains a non-rust source file.`

libfwupdplugin/meson.build
hughsie: this is the meson rust module, not the rustc compiler version. I think in 1.0.0 the rust module went *stable* and I'll see if the import name needs changing.

libfwupdplugin/fu-crc.rs
hughsie: Fixed.

contrib/ci/dependencies.xml
hughsie: I think I can use rustc for Fedora too; Fixed.

meson.build
hughsie: Added, good idea.

libfwupdplugin/fu-crc.rs
hughsie: rust has it's own formatter unfortunately. I'll look into setting up precommit.

hughsie: Because we're not using cargo the built-in one fails -- but I've added a tiny python version in contrib/ci

libfwupdplugin/meson.build
superm1: OK

libfwupdplugin/meson.build
superm1: Yeah, I'd be fine bumping the meson version to the project to 1.0.0 as part of this PR.

superm1:
> At the moment I'm not planning to use any crates -- all the stuff here is from std.

Right; but it's somethign we need to think about.  Once the cat is out of the bag, it's a matter of time until there are rust based plugins and those might need to take crates.

meson.build
superm1: I guess this will fix a lot of the CI.

contrib/ci/rustfmt.py
```diff
@@ -0,0 +1,45 @@
...
+            continue
+        if "venv" in fn:
+            continue
+        if fn.startswith("subprojects"):
+            continue
+        if fn.startswith("dist"):
+            continue
+        print(fn)
+        try:
+            subprocess.check_output(["rustfmt", "--quiet", "--check", fn])
```
superm1, line 35: So what's going to be interesting here is that not everyone will have `rustfmt` installed by default.  So two things this means:
1. Pre-commit CI job needs to explicitly install `rustc`/`rustfmt` if it's not doing this.
2. This test/script should be skipped if `rustfmt` isn't available.

libfwupdplugin/fu-crc.c
```diff
@@ -37,6 +38,7 @@ fu_crc8_full(const guint8 *buf, gsize bufsz, guint8 crc_init, guint8 polynomial)
 	}
 	return ~((guint8)(crc >> 8));
 }
+#endif
```
superm1, line 41: Couldn't you just have this one `ifndef` wrap all 3?  Or is the concern the documentation?

meson.build
```diff
@@ -139,6 +139,11 @@ add_project_link_arguments(
   language: 'c'
 )
 
+# rust is optional
+if add_languages('rust', native: false, required: get_option('rust').enabled())
```
superm1, line 143: I think you got this wrong from how I expected.  See if this makes more sense:
```suggestion
rustc= dependency('rustc', required: get_option('rust'))
if add_languages('rust', native: false, required: rustc.found())
```

libfwupdplugin/fu-crc.c
hughsie: I was worried about the docs. I'll play -- I didn't like the idea of duplicating them in two source files.

contrib/ci/rustfmt.py
hughsie: I did option 2.

libfwupdplugin/meson.build
```diff
@@ -360,7 +375,8 @@ if get_option('tests')
     ],
     link_with: [
       fwupd,
-      fwupdplugin
+      fwupdplugin,
```
superm1, line 378: there's a lot more places that link with `fwupdplugin`, this might need to be done more ubiquitously up front to make sure we don't have weird bugs later as more is converted to rust or goes in as rust.

hughsie: Ohh, it's linked *into* fwupdplugin -- so anything linking fwupdplugin gets fwupdpluginrs "for free"

libfwupdplugin/meson.build
```diff
@@ -236,7 +236,7 @@ library_deps = [
   platform_deps,
 ]
 
-if get_option('rust').allowed()
+if rustc.found()
```
hughsie, line 239: shouldn't this be something like `if get_option('rust').disable_auto_if(not rustc.found()).allowed()` ?

superm1: Yeah I think you're right

contrib/ci/rustfmt.py
```diff
@@ -0,0 +1,47 @@
...
+        if "venv" in fn:
+            continue
+        if fn.startswith("subprojects"):
+            continue
+        if fn.startswith("dist"):
+            continue
+        try:
+            subprocess.check_output(["rustfmt", "--quiet", "--check", fn])
+        except FileNotFoundError as e:
+            print("ignoring {}".format(e))
```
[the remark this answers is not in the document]
hughsie: It is noisy -- I've removed the print

libfwupdplugin/fu-crc.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright (C) 2022 Richard Hughes <richard@hughsie.com>
```
superm1, line 1: 2023 now

meson.build
```diff
@@ -139,6 +139,12 @@ add_project_link_arguments(
   language: 'c'
 )
 
+# rust is optional
+rustc = find_program('rustc', required: get_option('rust'))
+if add_languages('rust', native: false, required: rustc.found())
```
superm1, line 144: Shouldn't required be get option rust?

Nyabinary:
What the status on this?

hughsie:
> What the status on this?

It's work in progress. It's something I still want to do, but I don't want to make it more difficult to build fwupd.Being super-honest, I'm also somewhat unimpressed with both the binary size and the compile speed when enabling the rust support. You might have also noticed that slowly other parts of the codebase are actually using rust structures and enums, although we transpile them to C before compiling. e.g. see https://fwupd.github.io/libfwupdplugin/tutorial.html#the-rustgen-helper

libfwupdplugin/meson.build
hughsie: Done.

libfwupdplugin/fu-crc.rs
hughsie: I think this should be the date we wrote the code, not committed the code

hughsie:
@superm1 can you explain what's wrong with `rustc = find_program('rustc', required: get_option('rust').enabled())` -- I don't understand the failure! Thanks.

superm1:
I've got the meson stuff sorted now at least between this and https://github.com/fwupd/fwupd/pull/6271

superm1:
I'm obviously not experienced in rust but given the error the compiler is throwing, it's just saying there is no point to dropping the ref because it's not allocating the memory, right?  so just dropping `mem::drop` is probably the right action?

hughsie:
>  so just dropping mem::drop is probably the right action

Nah, you have to tell rust explicitly -- otherwise it leaks. I'll have a look after the weekend -- thanks for the buildsystem stuff, it's much appreciated.

hughsie:
> Nah, you have to tell rust explicitly -- otherwise it leaks

Turns out I was 100% wrong; fixed.

hughsie:
Note: this raises the meson dep to 1.0 -- this might be a problem for ChromeOS.

superm1 (approved):

hughsie:
@superm1 lets branch first -- I don't want to open the gates just yet.

superm1:
Sure no problem.

nyabinary:
Any updates to this? (been close to a year :P)

hughsie:
I'll close this, but keep the branch alive. We can re-open when it's ready to merge in the future. I'm really waiting for better rust support in meson, and also the ability to use rust in all the supported distros I support (e.g. RHEL 8/9).

PR #5434, opened 2023-01-21
Speed up startup by a huge amount when using directory remotes

hughsie:
By using libxmlb correctly we can avoid rebuilding the metadata if the directory is unchanged. This means the initial 'generation' is unchanged in speed, but subsequent loads are almost instant as the archives are not decompressed.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@bleungatchromium you might want this for ChromeOS :) If it helps, I'm planning an upstream tarball next week.

superm1 (approved):

PR #5436, opened 2023-01-21
Allow desktop software to inhibit devices to prevent updates

hughsie:
On edge hardware a process may want to disable firmware updates as it might be a bad time to allow an upgrade.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdmgr
+++ b/data/bash-completion/fwupdmgr
@@ -20,6 +20,8 @@ _fwupdmgr_cmd_list=(
 	'get-updates'
 	'get-upgrades'
 	'get-plugins'
+	'inhibit'
+	'uninhibit'
 	'install'
 	'local-install'
 	'modify-config'
--- a/libfwupd/fwupd-client-sync.c
+++ b/libfwupd/fwupd-client-sync.c
@@ -673,6 +673,113 @@ fwupd_client_unlock(FwupdClient *self,
 	}
 	return TRUE;
 }
+
+static void
+fwupd_client_inhibit_cb(GObject *source, GAsyncResult *res, gpointer user_data)
+{
+	FwupdClientHelper *helper = (FwupdClientHelper *)user_data;
+	helper->str = fwupd_client_inhibit_finish(FWUPD_CLIENT(source), res, &helper->error);
+	g_main_loop_quit(helper->loop);
+}
+
+/**
+ * fwupd_client_inhibit:
+ * @self: a #FwupdClient
+ * @reason: (not nullable): the inhibit reason, e.g. `user active`
+ * @cancellable: (nullable): optional #GCancellable
+ * @error: (nullable): optional return location for an error
+ *
+ * Marks all devices as unavailable for update. Update is only available if there is no other
+ * inhibit imposed by other applications or by the system (e.g. low power state).
+ *
+ * The same application can inhibit the system multiple times.
+ *
+ * Returns: (transfer full): a string to use for [method@FwupdClient.uninhibit_async],
+ * or %NULL for failure
+ *
+ * Since: 1.8.11
+ **/
+gchar *
+fwupd_client_inhibit(FwupdClient *self,
+		     const gchar *reason,
+		     GCancellable *cancellable,
+		     GError **error)
+{
+	g_autoptr(FwupdClientHelper) helper = NULL;
+
+	g_return_val_if_fail(FWUPD_IS_CLIENT(self), NULL);
+	g_return_val_if_fail(reason != NULL, NULL);
+	g_return_val_if_fail(cancellable == NULL || G_IS_CANCELLABLE(cancellable), NULL);
+	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
+
+	/* connect */
+	if (!fwupd_client_connect(self, cancellable, error))
+		return NULL;
+
+	/* call async version and run loop until complete */
+	helper = fwupd_client_helper_new(self);
+	fwupd_client_inhibit_async(self, reason, cancellable, fwupd_client_inhibit_cb, helper);
+	g_main_loop_run(helper->loop);
+	if (helper->str == NULL) {
+		g_propagate_error(error, g_steal_pointer(&helper->error));
+		return NULL;
+	}
+	return g_steal_pointer(&helper->str);
+}
+
+static void
+fwupd_client_uninhibit_cb(GObject *source, GAsyncResult *res, gpointer user_data)
+{
+	FwupdClientHelper *helper = (FwupdClientHelper *)user_data;
+	helper->ret = fwupd_client_uninhibit_finish(FWUPD_CLIENT(source), res, &helper->error);
+	g_main_loop_quit(helper->loop);
+}
+
+/**
+ * fwupd_client_uninhibit:
+ * @self: a #FwupdClient
+ * @uninhibit_id: (not nullable): the inhibit ID
+ * @cancellable: (nullable): optional #GCancellable
+ * @error: (nullable): optional return location for an error
+ *
+ * Removes the inhibit token added by the application.
+ *
+ * Returns: %TRUE for success
+ *
+ * Since: 1.8.11
+ **/
+gboolean
+fwupd_client_uninhibit(FwupdClient *self,
+		       const gchar *inhibit_id,
+		       GCancellable *cancellable,
+		       GError **error)
+{
+	g_autoptr(FwupdClientHelper) helper = NULL;
+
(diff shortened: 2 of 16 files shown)


superm1 (commented):
Why inhibit individual devices?  Shouldn't the inhibit be for all?

hughsie:
In this case it was intermittently polling a device, don't suggest firmware update -- but another usecase was "using webcam, don't update webcam" -- I don't think we'd want to inhibit all devices for this -- it would also be to easy to make everything "stop getting security updates" from an app point of view.

libfwupd/fwupd-client-sync.c
```diff
@@ -673,6 +673,120 @@ fwupd_client_unlock(FwupdClient *self,
...
+ * by other applications or by the system (e.g. low power state).
+ *
+ * The same application can inhibit a device multiple times.
+ *
+ * Returns: (transfer full): a string, or %NULL for failure
+ *
+ * Since: 1.8.10
+ **/
+gchar *
+fwupd_client_inhibit(FwupdClient *self,
```
hughsie, line 703: @superm1 maybe `fwupd_client_device_inhibit()` would be clearer?

superm1:
> In this case it was intermittently polling a device, don't suggest firmware update -- but another usecase was "using webcam, don't update webcam" -- I don't think we'd want to inhibit all devices for this -- it would also be to easy to make everything "stop getting security updates" from an app point of view.

I feel like it's a lot to expect software to know what devices to inhibit.

I guess walk me through the webcam case. What software would identify it's in use? How would it link it to the fwupd device?

hughsie:
> I guess walk me through the webcam case. What software would identify it's in use? 

The video-conferencing app running in usespace.

> How would it link it to the fwupd device?

 I guessed it would use the InstanceID/GUID, i.e. like `USB\VID_0BDA&PID_5850`

superm1:
> > I guess walk me through the webcam case. What software would identify it's in use? 
> 
> The video-conferencing app running in usespace.

I'm still not convinced. Userspace video conferencing apps will use stuff like /dev/video0 and it means they need to link that to a fwupd device.

> 
> > How would it link it to the fwupd device?
> 
>  I guessed it would use the InstanceID/GUID, i.e. like `USB\VID_0BDA&PID_5850`

So the other worry I have is you don't know the topology of devices. What if my webcam is connected to a USB hub connected to my USB4 dock and I decide to update my host controller's retimer.

The retimer update takes out the whole chain but the software can't possibly know that - we only inhibited the webcam.

We don't show a relationship for retimers today iirc.

hughsie:
Fair point. Do you think it makes sense to support both per-device and whole-device inhibits, or just the latter?

superm1:
> Fair point. Do you think it makes sense to support both per-device and whole-device inhibits, or just the latter?

I was thinking whole device only. 
The other thing to think about is that we have a usable while updating flag. Clients can be smarter on what to update - ie only apply updates for devices without they flag

hughsie:
How about this version?

libfwupd/fwupd-client-sync.c
```diff
@@ -673,6 +673,112 @@ fwupd_client_unlock(FwupdClient *self,
...
+ * @reason: (not nullable): the inhibit reason, e.g. `user active`
+ * @cancellable: (nullable): optional #GCancellable
+ * @error: (nullable): optional return location for an error
+ *
+ * Marks all devices as unavailable for update. Update is only available if there is no other
+ * inhibit imposed by other applications or by the system (e.g. low power state).
+ *
+ * The same application can inhibit a device multiple times.
+ *
+ * Returns: (transfer full): a string, or %NULL for failure
```
superm1, line 697: So this presumably returns a string that you use to undo inhibit, right?

You'll probably want to document it.

libfwupd/fwupd-client.c
```diff
@@ -5390,6 +5390,167 @@ fwupd_client_upload_bytes_finish(FwupdClient *self, GAsyncResult *res, GError **
...
+ * @self: a #FwupdClient
+ * @reason: (not nullable): the inhibit reason, e.g. `user active`
+ * @cancellable: (nullable): optional #GCancellable
+ * @callback: the function to run on completion
+ * @callback_data: the data to pass to @callback
+ *
+ * Marks all devices as unavailable for update. Update is only available if there is no other
+ * inhibit imposed by other applications or by the system (e.g. low power state).
+ *
+ * The same application can inhibit a device multiple times.
```
superm1, line 5424: Not applicable

src/fu-util-common.c
```diff
@@ -1402,6 +1402,10 @@ fu_util_device_problem_to_string(FwupdClient *client, FwupdDevice *dev, FwupdDev
 		/* TRANSLATORS: The device cannot be updated due to missing vendor's license." */
 		return g_strdup(_("Device requires a software license to update"));
 	}
+	if (problem == FWUPD_DEVICE_PROBLEM_SYSTEM_INHIBIT) {
+		/* TRANSLATORS: an application is preventing system updates */
+		return g_strdup(_("System is inhibited"));
```
superm1, line 1407: I think this string should be expanded.

src/fu-util.c
```diff
@@ -977,6 +977,23 @@ fu_util_device_test_filename(FuUtilPrivate *priv,
...
+static gboolean
+fu_util_inhibit(FuUtilPrivate *priv, gchar **values, GError **error)
+{
+	const gchar *reason = "not set";
+	g_autofree gchar *inhibit_id = NULL;
+
+	if (g_strv_length(values) > 0)
+		reason = values[1];
+
+	/* inhibit, wait, uninhibit */
```
superm1, line 989: This doesn't make sense to me. It should be two separate commands, or if combined command needs to have the delay as an argument.

superm1 (commented):
I think generally this direction makes sense. But if it's going to be enabled by command line there should probably be two separate commands or at least an argument for how long to inhibit.

data/bash-completion/fwupdmgr
```diff
@@ -20,6 +20,7 @@ _fwupdmgr_cmd_list=(
 	'get-updates'
 	'get-upgrades'
 	'get-plugins'
+	'inhibit'
```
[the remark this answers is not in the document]
hughsie: Done.

libfwupd/fwupd-client-sync.c
```diff
@@ -673,6 +673,112 @@ fwupd_client_unlock(FwupdClient *self,
...
+ * fwupd_client_inhibit:
+ * @self: a #FwupdClient
+ * @reason: (not nullable): the inhibit reason, e.g. `user active`
+ * @cancellable: (nullable): optional #GCancellable
+ * @error: (nullable): optional return location for an error
+ *
+ * Marks all devices as unavailable for update. Update is only available if there is no other
+ * inhibit imposed by other applications or by the system (e.g. low power state).
+ *
+ * The same application can inhibit a device multiple times.
```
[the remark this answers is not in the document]
hughsie: Fixed.

libfwupd/fwupd-client-sync.c
hughsie: Fixed.

libfwupd/fwupd-client.c
hughsie: Fixed.

src/fu-util-common.c
hughsie: Got any ideas? `All devices are prevented from update by inhibit` perhaps?

src/fu-util.c
hughsie: Agreed, two separate commands makes a lot more sense if we put in place a cancellable GMainLoop. Fixed.

src/fu-util-common.c
superm1: sounds good

src/fu-util.c
```diff
@@ -977,6 +978,40 @@ fu_util_device_test_filename(FuUtilPrivate *priv,
...
+	g_autofree gchar *inhibit_id = NULL;
+
+	if (g_strv_length(values) > 0)
+		reason = values[1];
+
+	/* inhibit then wait */
+	inhibit_id = fwupd_client_inhibit(priv->client, reason, priv->cancellable, error);
+	if (inhibit_id == NULL)
+		return FALSE;
+	g_print("Inhibit ID is: %s, use ctrl-c to abort\n", inhibit_id);
```
superm1, line 994: presumably this should be translatable

hughsie: How about this:

Screenshot from 2023-01-23 09-32-17

superm1: Lgtm

Issue #5402, opened 2023-01-10
s390x/ppc64el fail autopkgtest


superm1:
W/ 1.8.9 + 6c18cc0f66b10dd63ae2947f56ba9c3a078fc4d8 I'm finding that both Debian and Ubuntu are failing autopkgtest specific to s390x and ppc64el.

```
# Start of mtd tests
# GLib-GIO-DEBUG: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ?gio-vfs?
# XbSilo-DEBUG: attempting to load /tmp/.WWNEY1
# XbSilo-DEBUG: failed to load silo: blob too small
Bail out! FuContext-FATAL-WARNING: Failed to load SMBIOS: neither SMBIOS or DT found
```

I guess it's because that environment change we made isn't set for running installed tests.  I suspect it's because we forgot to install the symlink for `plugins/mtd/tests/dmi`

superm1:
Here's one link to the failure: https://autopkgtest.ubuntu.com/results/autopkgtest-lunar/lunar/ppc64el/f/fwupd/20230105_182402_168df@/log.gz.
The other 3 look the same.

hughsie:
Isn't the symlink set up in 6c18cc0f66b10dd63ae2947f56ba9c3a078fc4d8?

superm1:
It's not installed though into installed-tests.  Neither is the DMI data from libfwupdplugin.  So we either need to fail the test if missing data or install the data too.

hughsie:
Ahh, installed-tests. Which would you rather do?

superm1:
I think it's better to install the dmi data.  We can have more reproducible tests even in installed test environment then too.

superm1:
With W/ 1.8.9 + 6c18cc0f66b10dd63ae2947f56ba9c3a078fc4d8 + c2a494e8711dd4d5813d72aed126091fe6a3434f it's [still failing](https://ci.debian.net/data/autopkgtest/testing/ppc64el/f/fwupd/30480472/log.gz) unfortunately.

```
# Start of mtd tests
# GLib-GIO-DEBUG: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ?gio-vfs?
# XbSilo-DEBUG: attempting to load /tmp/.RPTZY1
# XbSilo-DEBUG: failed to load silo: blob too small
Bail out! FuContext-FATAL-WARNING: Failed to load SMBIOS: neither SMBIOS or DT found

[...]

PR #5443, opened 2023-01-24
Do not make any of the HWIDs setup failures fatal

hughsie:
It's perfectly okay to have no HWIDs defined.

Should help with https://github.com/fwupd/fwupd/issues/5402

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-context.c
+++ b/libfwupdplugin/fu-context.c
@@ -767,6 +767,8 @@ fu_context_security_changed(FuContext *self)
 	g_signal_emit(self, signals[SIGNAL_SECURITY_CHANGED], 0);
 }
 
+typedef gboolean (*FuContextHwidsSetupFunc)(FuContext *self, FuHwids *hwids, GError **error);
+
 /**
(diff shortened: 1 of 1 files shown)


superm1 (approved):
This looks fine to me, but I'm not convinced to understand why it helps.

superm1:
> This looks fine to me, but I'm not convinced to understand why it helps.

I mean why it didn't fail ci before.

hughsie:
> I mean why it didn't fail ci before.

I'm wondering if the debian build machines provide *just enough* DT support to start parsing, but but enough to actually be valid. Any way you can exfiltrate the `fdt` file from a build server?

superm1:
> > I mean why it didn't fail ci before.
> 
> I'm wondering if the debian build machines provide _just enough_ DT support to start parsing, but not enough to actually be valid. Any way you can exfiltrate the `fdt` file from a build server?

No; I'm not sure how to do that.  I'll see if the new patch on top of 1.8.10 works though.

PR #5445, opened 2023-01-24
wacom-usb: Retry set_report on failure

jigpu:
Sometimes the flash process will randomly hang and time-out when sending data to the device. We currently do not use any retry logic, so if this happens the flash attempt is treated as a failure. This can be a source of worry or frustration, especially if subsequent manual retries fail in a similar way.

Adding FU_HID_DEVICE_FLAG_RETRY_FAILURE to the list of flags used when calling fu_hid_device_set_report allows fwupd to try sending a block multiple times if such a time-out (or other error) occurs. This makes the flash process less prone to failure.

Signed-off-by: Jason Gerecke <jason.gerecke@wacom.com>

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/wacom-usb/fu-wac-device.c
+++ b/plugins/wacom-usb/fu-wac-device.c
@@ -182,7 +182,8 @@ fu_wac_device_set_feature_report(FuWacDevice *self,
 					buf,
 					bufsz,
 					FU_WAC_DEVICE_TIMEOUT,
-					flags | FU_HID_DEVICE_FLAG_IS_FEATURE,
+					flags | FU_HID_DEVICE_FLAG_IS_FEATURE |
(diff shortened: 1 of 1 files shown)


hughsie (dismissed):

hughsie:
Ohh, you need to run clang-format on the file.

jigpu:
Ah, thanks for the reminder. Been too long since I opened a pull request! :embarrased:

hughsie:
Certainly no need to be embarrassed -- we're the _anally retentive_ ones who insist on such strict coding style :) If it helps, `./contrib/setup` will set up precommit for you -- which makes git match the same rules as CI.

hughsie (approved):

PR #5452, opened 2023-01-26
Remove duplicate calls to fu_device_list_wait_for_replug()

hughsie:
The fu_engine_get_device() function is called at the start of each phase. Duplicate calls are at best confusing, and at worst could cause a 'device failed to come back' failure.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -3153,11 +3153,6 @@ fu_engine_prepare(FuEngine *self,
 			return FALSE;
 	}
 
-	/* wait for device to disconnect and reconnect */
-	if (!fu_device_list_wait_for_replug(self->device_list, error)) {
-		g_prefix_error(error, "failed to wait for prepare replug: ");
-		return FALSE;
-	}
 	return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


fdanis-oss:
I'm not sure to understand the commit comment, it refers to `fu_engine_get_device()` but the commit removes calls to `fu_device_list_wait_for_replug()`

hughsie:
Right, maybe it's confusing -- the fu_engine_get_device() function gets called every time we need to convert a device-id to a FuDevice. Maybe it's the wrong name entirely -- e.g. we should rename fu_engine_get_device() to be fu_engine_wait_for_replug_get_device()

fdanis-oss:
maybe changing comment to:
```
The fu_engine_get_device() function is called at the start of each phase, which calls fu_device_list_wait_for_replug(). Duplicate calls are at best confusing, and at worst could cause a 'device failed to come back' failure.
```

hughsie:
@fdanis-oss agreed, fixed.

fdanis-oss (approved):

PR #5454, opened 2023-01-26
Add a device problem for 'update-in-progress'

hughsie:
We already had this as an inhibit, but this was not translated client-side.

We also need to propagate the problem to the bootloader device if the device replugs during firmware update.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@fdanis-oss this also allows us to ask the device "are you in the middle of a firmware update" when recording the emulation data.

fdanis-oss (approved):

PR #5467, opened 2023-01-30
Use a more compact JSON format using precommit

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 is it fair to ignore the precommit failure for this PR as the "commit" includes the precommit fix? Or am I missing something?

hughsie:
> Or am I missing something?

The bit I was missing was ignore-words-list has weird case matching semantics. Fixed.

superm1 (approved):

PR #5475, opened 2023-01-31
Add API to wait for a device

hughsie:
This allows us to ignore all the delays when the device is emulated, with the idea being to do dozens of device emulations in the CI tests.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -635,8 +635,8 @@ fu_device_retry_full(FuDevice *self,
 		g_autoptr(GError) error_local = NULL;
 
 		/* delay */
-		if (i > 0 && delay > 0)
-			g_usleep(delay * 1000);
+		if (i > 0)
+			fu_device_sleep(self, delay);
 
 		/* run function, if success return success */
 		if (func(self, user_data, &error_local))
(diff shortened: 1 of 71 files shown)


hughsie:
@fdanis-oss you have keen eyes; spot any potential regressions? :)

superm1 (dismissed):

libfwupdplugin/fu-device.c
```diff
@@ -718,6 +718,24 @@ fu_device_retry(FuDevice *self,
...
+ *
+ * Delays program execution, unless the device is emulated.
+ *
+ * Since: 1.8.11
+ **/
+void
+fu_device_sleep(FuDevice *self, guint delay_ms)
+{
+	g_return_if_fail(FU_IS_DEVICE(self));
+	g_return_if_fail(delay_ms < 1000);
```
fdanis-oss, line 734: `fu_ti_tps6598x_device_wait_for_command()` is some times called with a 1000 ms delay.  
Should this be:
```suggestion
	g_return_if_fail(delay_ms <= 1000);
```
or the delays replaced by 999 in plugins/ti-tps6598x/fu-ti-tps6598x-device.c ?

hughsie: eek! I got my ms confused by my us! The idea there was to do a critical warning if the programmer tried to do `fu_device_sleep(100*1000)` i.e. in us. I think if we're waiting more than 100 seconds for something then it's probably better we use `fu_progress_sleep()` to get some progress indication. I've fixed up the constant and also added this in the docs. I said you had keen eyes! :)

hughsie: Okay fixed, this time also dealing with fu_progress_sleep too.

fdanis-oss (approved):

PR #5476, opened 2023-01-31
trivial: use github mirror for gcab

superm1:
[GNOME git](https://gitlab.gnome.org/GNOME/gcab.git) seems to be down a few days. As this breaks Windows CI for us and we run fwupd on Github let's use the Github mirror instead.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/mingw64.cross
+++ b/contrib/mingw64.cross
@@ -1,2 +1,3 @@
 [binaries]
 windmc = '/usr/bin/x86_64-w64-mingw32-windmc'
+exe_wrapper = 'wine'
--- a/meson.build
+++ b/meson.build
@@ -574,7 +574,7 @@ if polkit.found()
 endif
 if build_standalone
   plugin_quirks = []
-  gcab = find_program('gcab', required: get_option('tests'))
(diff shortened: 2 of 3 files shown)


superm1:
Hmm given the CI failure, I have to wonder why are we using the windows gcab if it's a cross compile and just building a .cab?  We may as well just use the host one, no?

hughsie:
> why are we using the windows gcab if it's a cross compile

We still need libgcab for decompressing archives in the daemon, no?

superm1:
> > why are we using the windows gcab if it's a cross compile
> 
> We still need libgcab for decompressing archives in the daemon, no?

But gcab.exe during build. That can be the x86 one

hughsie:
> But gcab.exe during build. That can be the x86 one

Yes, indeed, agreed.

hughsie:
@superm1 still want this pr?

superm1:
So the real problem is we can't upgrade to a newer gcab.  I'd like to move mirrors too, but that's secondary to actually being able to adopt newer software.

PR #5478, opened 2023-02-01
Don't wait for replug of emulated devices

fdanis-oss:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Does this make the TI dock work?

src/fu-device-list.c
```diff
@@ -905,6 +905,13 @@ fu_device_list_wait_for_replug(FuDeviceList *self, GError **error)
 		return TRUE;
 	}
 
+	/* emulated devices don't need to wait for replug */
+	for (guint i = 0; i < devices_wfr1->len; i++) {
+		FuDevice *device_tmp = g_ptr_array_index(devices_wfr1, i);
+		if (fu_device_has_flag(device_tmp, FWUPD_DEVICE_FLAG_EMULATED))
+			fu_device_remove_flag(device_tmp, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
```
hughsie, line 912: I think get rid of the `if fu_device_has_flag(device_tmp, FWUPD_DEVICE_FLAG_EMULATED` -- just remove it unconditionally.

hughsie: Nope, ignore me, i'm being crazy.

fdanis-oss: if I remove the test it will break the behavior when waiting for a real device re-plug, no?

fdanis-oss:
Yes, it allows to complete TI dock update, but there's still another issue as it doesn't get the right firmware version

fdanis-oss:
Close as it's better handled by #5479

PR #5488, opened 2023-02-04
trivial: Only use clang-format on C source files

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Don't we also use it for protobuf?

hughsie:
> Don't we also use it for protobuf?

We do, fixed.

PR #5490, opened 2023-02-04
ccgx: Add some more plugin debugging

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/ccgx/ccgx.quirk
+++ b/plugins/ccgx/ccgx.quirk
@@ -141,3 +141,13 @@ Summary = Dock Management Controller Device
 CcgxDmcTriggerCode = 0x01
 CcgxImageKind = dmc-composite
 RemoveDelay = 732000
+
+# Caldigit ElementHub
+[USB\VID_2188&PID_0035]
+Plugin = ccgx
+GType = FuCcgxDmcDevice
+CcgxDmcTriggerCode = 0x01
+CcgxImageKind = dmc-composite
+RemoveDelay = 732000
+[USB\VID_2188&PID_0035&CID_05&VER_3.3.1.69]
+CcgxDmcCompositeVersion = 15
--- a/plugins/ccgx/fu-ccgx-dmc-devx-device.c
+++ b/plugins/ccgx/fu-ccgx-dmc-devx-device.c
@@ -1,0 +1,283 @@
+/*
+ * Copyright (C) 2023 Richard Hughes <richard@hughsie.com>
+ * Copyright (C) 2020 Cypress Semiconductor Corporation.
+ *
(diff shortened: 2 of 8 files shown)


hughsie:
@bleungatchromium this might be interesting to you too; I can share more details when Infineon have finished talking internally.

plugins/ccgx/ccgx.quirk
```diff
@@ -141,3 +141,13 @@ Summary = Dock Management Controller Device
 CcgxDmcTriggerCode = 0x01
 CcgxImageKind = dmc-composite
 RemoveDelay = 732000
+
+# Caldigit ElementHub
```
superm1, line 145: This should be a separate commit

plugins/ccgx/fu-ccgx-dmc-devx-device.c
```diff
@@ -0,0 +1,275 @@
...
+	/* add GUIDs */
+	fu_device_add_instance_strup(device,
+				     "TYPE",
+				     fu_ccgx_dmc_devx_device_type_to_string(status->device_type));
+	fu_device_add_instance_u8(device, "CID", status->component_id);
+	fu_device_add_instance_u16(device, "VID", fu_usb_device_get_vid(FU_USB_DEVICE(proxy)));
+	fu_device_add_instance_u16(device, "PID", fu_usb_device_get_pid(FU_USB_DEVICE(proxy)));
+	fu_device_build_instance_id(device, NULL, "USB", "VID", "PID", "CID", NULL);
+	fu_device_build_instance_id_quirk(device, NULL, "USB", "VID", "PID", "CID", "TYPE", NULL);
+	fu_device_build_instance_id_quirk(device, NULL, "USB", "VID", "PID", "CID", "VER", NULL);
```
superm1, line 244: This should only be if VER isn't null right?

hughsie: `fu_device_build_instance_id_quirk()` just skips the instance ID if any one of the keys is unset.

plugins/ccgx/ccgx.quirk
hughsie: Agreed, done.

superm1 (approved):
This LGTM.  As part of reviewing it though, I noticed that the `RemoveDelay` key is the same for a lot of devices.  Should that maybe be the codified default too?

hughsie:
> I noticed that the RemoveDelay key is the same for a lot of devices. 

I'll ask Infineon -- maybe we can set it automatically according to the CCG version or something.

PR #5497, opened 2023-02-06
trivial: Fix Fedora CI when it cannot autostart udisks

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdmgr
+++ b/data/bash-completion/fwupdmgr
@@ -26,6 +26,7 @@ _fwupdmgr_cmd_list=(
 	'local-install'
 	'modify-config'
 	'modify-remote'
+	'quit'
 	'reinstall'
 	'refresh'
 	'report-history'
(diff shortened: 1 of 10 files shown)


.github/workflows/matrix.yml
```diff
@@ -7,6 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [fedora, debian-x86_64, arch, debian-i386]
+      fail-fast: false
```
superm1, line 10: Why want to change this? The idea is to save CI minutes.

superm1 (commented):
Besides the fail fast rest seems ok

.github/workflows/matrix.yml
hughsie: Ohh it was just a hack to workaround all the debian builds failing due to a network issue. I've dropped that one now.

PR #5501, opened 2023-02-07
snap: read the host os-release instead of base snap

### LINKED ISSUE #5485: edge snap sets up "ubuntu-core" path for `fwupdx64.efi`
Experimenting with the edge snap (1.8.10-67-gaef162223 rev 4646) with an edge snapd (2.58.2+git450.g4241ddc rev 18423) on Ubuntu 22.04.

I find that running an update that the path built for UEFI is `\efi\ubuntu-core\fwupdx64.efi`.  As the snap can run on Ubuntu desktop, or anything else that runs a snap I think we should change this, but I invite discussion.

I would propose the default for the snap should be `\efi\fwupd\fwupx64.efi`

valentindavid:
This should fix #5485. Depends on https://github.com/snapcore/snapd/pull/12545

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -80,6 +80,7 @@ for details.
 * `FWUPD_DATADIR_QUIRKS`
 * `FWUPD_EFIAPPDIR`
 * `FWUPD_FIRMWARESEARCH`
+* `FWUPD_HOSTDIR` looks for host OS `os-release` in this sysroot, default is /
 * `FWUPD_LIBDIR_PKG`
 * `FWUPD_LOCALSTATEDIR`
 * `FWUPD_LOCALSTATEDIR_QUIRKS`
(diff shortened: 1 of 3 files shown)


docs/env.md
```diff
@@ -80,6 +80,7 @@ for details.
 * `FWUPD_DATADIR_QUIRKS`
 * `FWUPD_EFIAPPDIR`
 * `FWUPD_FIRMWARESEARCH`
+* `FWUPD_HOSTDIR`
```
hughsie, line 83: i think this need docs

valentindavid: Done.

superm1 (approved):
This looks good, thanks.

PR #5506, opened 2023-02-09
Ensure composite devices all share the same acquiesce delay

hughsie:
We go to great pains to set the parent to the MAX of the child delays, but we don't actually set the children to have the new parent value.

This means if you update the child device without the parent in the transaction we don't always wait for all the USB devices to come back -- which is what the acquiesce delay is designed to do.

Fixes an intermittant device test for the Wacom USB tablet.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
Will make some composite updates take longer than "necessary" but I believe it's safe.

PR #5508, opened 2023-02-09
Install and use shim in the confined snap

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

cc @valentindavid 
cc @xnox

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-common.c
+++ b/plugins/uefi-capsule/fu-uefi-common.c
@@ -85,8 +85,21 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
 	return g_strdup_printf("%s/%s%s.efi", base, cmd, suffix);
 }
 
+/**
+ * fu_uefi_get_built_app_path:
+ * @basename: the prefix for the binary
+ * @error: (nullable): optional return location for an error
+ *
+ * Gets the path intended to be used for an EFI binary on the local system.
+ * The binary is matched against the correct architecture and if secure
+ * boot is enabled.
+ *
+ * Returns: The full path to the binary, or %NULL if not found
+ *
+ * Since: 1.8.1
+ **/
 gchar *
-fu_uefi_get_built_app_path(GError **error)
+fu_uefi_get_built_app_path(const gchar *binary, GError **error)
 {
 	const gchar *suffix;
 	g_autofree gchar *prefix = NULL;
@@ -100,7 +113,7 @@ fu_uefi_get_built_app_path(GError **error)
 		return NULL;
 	prefix = fu_path_from_kind(FU_PATH_KIND_EFIAPPDIR);
 
-	source_path = g_strdup_printf("%s/fwupd%s.efi", prefix, suffix);
+	source_path = g_strdup_printf("%s/%s%s.efi", prefix, binary, suffix);
 	source_path_signed = g_strdup_printf("%s.signed", source_path);
 
 	source_path_exists = g_file_test(source_path, G_FILE_TEST_EXISTS);
--- a/plugins/uefi-capsule/fu-uefi-device.h
+++ b/plugins/uefi-capsule/fu-uefi-device.h
@@ -67,12 +67,6 @@ typedef enum {
  */
 #define FU_UEFI_DEVICE_FLAG_SUPPORTS_BOOT_ORDER_LOCK (1 << 3)
 /**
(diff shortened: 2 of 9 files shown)


plugins/uefi-capsule/fu-uefi-device.h
```diff
@@ -66,30 +66,24 @@ typedef enum {
...
- * Try to fallback to use UEFI removable path if the shim path doesn't exist.
- */
-#define FU_UEFI_DEVICE_FLAG_FALLBACK_TO_REMOVABLE_PATH (1 << 4)
 /**
  * FU_UEFI_DEVICE_FLAG_USE_SHIM_FOR_SB:
  *
  * Use shim to load fwupdx64.efi when SecureBoot is turned on.
  */
-#define FU_UEFI_DEVICE_FLAG_USE_SHIM_FOR_SB (1 << 5)
+#define FU_UEFI_DEVICE_FLAG_USE_SHIM_FOR_SB (1 << 4)
```
hughsie, line 74: I think these can stay - a "hole" is fine

plugins/uefi-capsule/fu-uefi-common.c
```diff
@@ -86,7 +86,7 @@ fu_uefi_get_esp_app_path(FuDevice *device, const gchar *esp_path, const gchar *c
 }
 
 gchar *
-fu_uefi_get_built_app_path(GError **error)
+fu_uefi_get_built_app_path(const gchar *binary, GError **error)
```
hughsie, line 89: can you add a comment to this function please -- for our future selves. Otherwise LGTM.

hughsie (approved):

superm1:
@valentindavid I had a try with the artifacts produced from CI.  It seems that the apparmor rules would need to be changed for this if this is the direction we go.

This is what I get:
```
Failed to copy /snap/fwupd/x1/libexec/fwupd/efi/shimx64.efi.signed to /boot/efi/EFI/fwupd/shimx64.efi: Error opening file ?/boot/efi/EFI/fwupd/shimx64.efi?: Permission denied
```

valentindavid:
@superm1 Can you look at https://github.com/snapcore/snapd/pull/12557 and tell me if this is what you want? I have tested it, and it fixed it.

Also I have noticed that fwupdmgr tries to reboot, but it does not have the permission. So I have opened https://github.com/fwupd/fwupd/pull/5513

superm1:
> @superm1 Can you look at https://github.com/snapcore/snapd/pull/12557 and tell me if this is what you want? I have tested it, and it fixed it.
> 
> Also I have noticed that fwupdmgr tries to reboot, but it does not have the permission. So I have opened https://github.com/fwupd/fwupd/pull/5513

Yup, thanks!

And the os-release thing we did for previous solution is still useful, please don't back it out. We'll have more accurate reports when the snap is used and a user opts into reporting.

xnox:
For Ubuntu, we have multiple shim signatures. To work on both Core and Classic Ubuntu it is best to use the signature that is signed by the Core interim key and chained to the Classic CA as it is trusted by all Ubuntu installations. I will propose a change to the snapcraft packaging.

superm1:
> For Ubuntu, we have multiple shim signatures. To work on both Core and Classic Ubuntu it is best to use the signature that is signed by the Core interim key and chained to the Classic CA as it is trusted by all Ubuntu installations. I will propose a change to the snapcraft packaging.

I wasn't aware there were different signed binaries.

Could you do it "soon"? We're targeting the next release on Feb 23.

valentindavid:
@superm1 I have noticed this PR can brick. We are installing a shim that is potentially newer than in the distro. That may revoke old bootloaders (grub) and brick the current OS installation. The only way to get to boot it is to clear shim UEFI variables or manually update the bootloader offline.

valentindavid:
One way to work around this is to not take the latest shim from `shim-signed`. Maybe we can provide an older shim from a PPA. And  we update this shim only when all major distros have updated the latest shim.

valentindavid:
Rejecting the update would be fine.

We could query UEFI variable `SbatLevelRT` to check what has been already revoked. And if the revocations do not match what we expect, this is when we should bock.

superm1:
How do you know what you expect though?

valentindavid:
We can read section `.sbatlevel` of the shim we want to install. https://github.com/rhboot/shim/blob/main/sbat_var.S
Then we should have at least `SBAT_VAR_PREVIOUS`, I think.

PR #5510, opened 2023-02-09
Adjust family view for a few devices 

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/thunderbolt/fu-thunderbolt-controller.c
+++ b/plugins/thunderbolt/fu-thunderbolt-controller.c
@@ -288,6 +288,7 @@ fu_thunderbolt_controller_setup(FuDevice *device, GError **error)
 
 		} else {
 			device_id = g_strdup("TBT-fixed");
+			fu_device_add_parent_guid(device, "cpu");
 		}
 		fu_device_add_instance_id(device, device_id);
 		if (domain_id != NULL)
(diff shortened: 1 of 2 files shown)


plugins/thunderbolt/fu-thunderbolt-controller.c
```diff
@@ -288,6 +288,7 @@ fu_thunderbolt_controller_setup(FuDevice *device, GError **error)
 
 		} else {
 			device_id = g_strdup("TBT-fixed");
+			fu_device_add_parent_guid(device, "cpu");
```
hughsie, line 291: i know this is true electrically, but is this what the user is going to think?

superm1: Nothing really - you only see it in `--show-all-devices` output.  Another argument can be to just not populate the device.  It has nothing valuable in fwupd.

hughsie (approved):

plugins/thunderbolt/fu-thunderbolt-controller.c
hughsie: > Nothing really - you only see it in --show-all-devices output

Ah, fair point. Got it.

PR #5513, opened 2023-02-10
snap: allow fwupdmgr to shutdown

valentindavid:
When installing a capsule, optional reboot is done by fwupdmgr instead of fwupd. This means fwupdmgr needs the permission to do that.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -54,7 +54,7 @@ apps:
       FWUPD_HOSTDIR: /var/lib/snapd/hostfs
   fwupdmgr:
     command: fwupdmgr.wrapper
-    plugs: [fwupdmgr, network, polkit]
+    plugs: [fwupdmgr, network, polkit, shutdown]
     completer:
       share/bash-completion/completions/fwupdmgr
   fwupdagent:


hughsie:
It sometimes also needs reboot instead, is that the same permission?

superm1 (approved):

valentindavid:
`shutdown` interface includes reboot. Here is the AppArmor profile for it.

```
dbus (send)
    bus=system
    path=/org/freedesktop/systemd1
    interface=org.freedesktop.systemd1.Manager
    member={Reboot,PowerOff,Halt}
    peer=(label=unconfined),

dbus (send)
    bus=system
    path=/org/freedesktop/login1
    interface=org.freedesktop.login1.Manager
    member={PowerOff,Reboot,Suspend,Hibernate,HybridSleep,CanPowerOff,CanReboot,CanSuspend,CanHibernate,CanHybridSleep,ScheduleShutdown,CancelScheduledShutdown,SetWallMessage}
    peer=(label=unconfined),
```

PR #5516, opened 2023-02-10
Split out the console handling to a new module

hughsie:
This simplifies a lot of confusion.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


src/fu-util.c
```diff
@@ -356,51 +361,60 @@ fu_util_perhaps_show_unreported(FuUtilPrivate *priv, GError **error)
...
+						   fwupd_device_get_version(dev),
+						   fwupd_release_get_version(rel));
 			}
 		}
 
 		/* success */
 		if (devices_success->len > 0) {
-			/* TRANSLATORS: a list of successful updates */
-			g_print("\n%s\n\n", _("Devices that have been updated successfully:"));
+			fu_console_println_literal(
```
hughsie, line 385: @superm1 the other way to do this is have fu_console_print("%s\n", _("Devices that h..` but I don't know if that's better or worse.

src/fu-util.c
```diff
@@ -356,51 +361,60 @@ fu_util_perhaps_show_unreported(FuUtilPrivate *priv, GError **error)
 	/* show the success and failures */
 	if (!priv->assume_yes && !all_automatic) {
 		/* delimit */
-		g_print("________________________________________________\n");
+		fu_console_println_literal(priv->console,
+					   "________________________________________________");
```
superm1, line 365: This seems like something that a helper should exist for to me.  Create a line that is "X" characters long

src/fu-util.c
superm1: I like this the current way

src/fu-util.c
```diff
@@ -2564,14 +2611,14 @@ fu_util_update(FuUtilPrivate *priv, gchar **values, GError **error)
 						 &error_local);
 		if (rels == NULL) {
 			if (!latest_header) {
-				g_printerr(
-				    "%s\n",
+				fu_console_println_literal(
+				    priv->console,
```
superm1, line 2615: this used to be on stderr but is now on stdout right?  Is that intended?

src/fu-util.c
```diff
@@ -2564,14 +2611,14 @@ fu_util_update(FuUtilPrivate *priv, gchar **values, GError **error)
...
 				    /* TRANSLATORS: message letting the user know no device upgrade
 				     * available */
 				    _("Devices with the latest available firmware version:"));
 				latest_header = TRUE;
 			}
-			g_printerr(" • %s\n", fwupd_device_get_name(dev));
+			fu_console_println(priv->console, " • %s", fwupd_device_get_name(dev));
```
superm1, line 2621: same stderr comment

src/fu-tool.c
```diff
@@ -665,20 +665,23 @@ fu_util_get_updates(FuUtilPrivate *priv, gchar **values, GError **error)
...
-		/* TRANSLATORS: message letting the user know no device upgrade available */
-		g_printerr("%s\n", _("Devices with the latest available firmware version:"));
+		fu_console_println_literal(
+		    priv->console,
+		    /* TRANSLATORS: message letting the user know no device upgrade available */
+		    _("Devices with the latest available firmware version:"));
 		for (guint i = 0; i < devices_no_upgrades->len; i++) {
 			FwupdDevice *dev = g_ptr_array_index(devices_no_upgrades, i);
-			g_printerr(" • %s\n", fwupd_device_get_name(dev));
+			fu_console_println(priv->console, " • %s", fwupd_device_get_name(dev));
```
superm1, line 684: same stderr comment

src/fu-tool.c
```diff
@@ -665,20 +665,23 @@ fu_util_get_updates(FuUtilPrivate *priv, gchar **values, GError **error)
...
 		for (guint i = 0; i < devices_no_support->len; i++) {
 			FwupdDevice *dev = g_ptr_array_index(devices_no_support, i);
-			g_printerr(" • %s\n", fwupd_device_get_name(dev));
+			fu_console_println(priv->console, " • %s", fwupd_device_get_name(dev));
 		}
 	}
 	if (devices_no_upgrades->len > 0) {
-		/* TRANSLATORS: message letting the user know no device upgrade available */
-		g_printerr("%s\n", _("Devices with the latest available firmware version:"));
+		fu_console_println_literal(
```
superm1, line 678: same stderr comment

src/fu-console.c
```diff
@@ -0,0 +1,753 @@
...
+#endif
+
+struct _FuConsole {
+	GObject parent_instance;
+	GMainContext *main_ctx;
+	FwupdStatus status;
+	gboolean spinner_count_up; /* chars */
+	guint spinner_idx;	   /* chars */
+	guint length_percentage;   /* chars */
+	guint length_status;	   /* chars */
```
superm1, line 28: I realize this was lifted from fu-progressbar.c, but what exactly is chars supposed to actually _mean_?

src/fu-console.c
```diff
@@ -0,0 +1,753 @@
...
+	guint percentage;
+	GSource *timer_source;
+	gint64 last_animated; /* monotonic */
+	GTimer *time_elapsed;
+	gdouble last_estimate;
+	gboolean interactive;
+	gboolean tainted;
```
superm1, line 35: this is a weird adjective to indicate it's got contents to clear.  Any better ideas?

src/fu-console.c
```diff
@@ -0,0 +1,753 @@
...
+
+/**
+ * fu_console_set_progress:
+ * @self: A #FuConsole
+ * @status: A #FwupdStatus
+ * @percentage: unsigned integer
+ *
+ * Refreshes a console
+ *
+ * Since: 0.9.7
```
superm1, line 620: this is wrong

src/fu-console.c
```diff
@@ -0,0 +1,753 @@
...
+	g_return_if_fail(FU_IS_CONSOLE(self));
+	self->interactive = interactive;
+}
+
+/**
+ * fu_console_set_status_length:
+ * @self: A #FuConsole
+ * @len: unsigned integer
+ *
+ * Sets the length of the console status
```
superm1, line 685: Probably needs something here to explain why len needs to be less than 3

src/fu-console.c
```diff
@@ -0,0 +1,753 @@
...
+	g_return_if_fail(len > 3);
+	self->length_status = len;
+}
+
+/**
+ * fu_console_set_percentage_length:
+ * @self: A #FuConsole
+ * @len: unsigned integer
+ *
+ * Sets the length of the progressba percentage
```
superm1, line 700: progressbar
also explain why limit is less than 3

src/fu-console.c
```diff
@@ -0,0 +1,753 @@
...
+	g_return_if_fail(FU_IS_CONSOLE(self));
+	g_return_if_fail(len > 3);
+	self->length_percentage = len;
+}
+
+static void
+fu_console_init(FuConsole *self)
+{
+	self->length_percentage = 40;
+	self->length_status = 25;
```
superm1, line 714: these seems arbitrary, shouldn't they be set by the caller?

src/fu-util.c
hughsie: Good idea, fixed. I've also used the UTF-8 ─ char rather than _.

src/fu-util.c
hughsie: I guess I'm torn on what we should do with stderr -- on one hand if we're doing anything thats going to be captured or shown in logs then using --json is probably strongly advised. On the other hand it gets a bit handwavy on what goes to stderr and what goes to stdout. Perhaps a better API would be to have:

 * fu_console_print_literal() -- with a newline
 * fu_console_print() -- with a newline

for the common case, and then have a

 * fu_console_print_literal_full()
 * fu_console_print_full()

With FU_CONSOLE_PRINT_FLAG_NEWLINE and FU_CONSOLE_PRINT_FLAG_STDERR

That's probably going to cover all usecases.

src/fu-util.c
hughsie: I think this one actually wants to be on stdout.

src/fu-tool.c
hughsie: I think this is legit stdout, no?

src/fu-tool.c
hughsie: I think this is legit stdout, no?

src/fu-console.c
hughsie: Fixed.

src/fu-console.c
hughsie: I think you do :) `contents_to_clear` works for me, fixed.

src/fu-util.c
hughsie: I did that, and it seems to work okay.

src/fu-console.c
hughsie: it's the width in chars, which can be different to the number of bytes for UTF-8 output. I've added a comment.

src/fu-console.c
hughsie: Done.

src/fu-console.c
hughsie: done, and done.

src/fu-console.c
superm1: sure makes sense.

superm1 (approved):

PR #5528, opened 2023-02-13
Wip/hughsie/412

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -321,7 +321,16 @@ fwupd_client_set_host_security_id(FwupdClient *self, const gchar *host_security_
 	fwupd_client_object_notify(self, "host-security-id");
 }
 
-static void
+/**
+ * fwupd_client_set_daemon_version:
+ * @self: a #FwupdClient
+ * @daemon_version: A semantic version, e.g. "1.2.3"
+ *
+ * Sets the daemon version number.
+ *
+ * Since: 1.8.11
+ **/
+void
 fwupd_client_set_daemon_version(FwupdClient *self, const gchar *daemon_version)
 {
 	FwupdClientPrivate *priv = GET_PRIVATE(self);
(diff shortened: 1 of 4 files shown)


superm1 (approved):

AdamWill:
note the additional check here breaks KDE's "Discover" tool: https://bugzilla.redhat.com/show_bug.cgi?id=2173022 and https://bugs.kde.org/show_bug.cgi?id=466318 . Probably needs addressing in Discover, though, as the check does make sense.

hughsie:
@AdamWill It wouldn't have worked before this PR either -- now the LVFS is filtering out insecure fwupd user agents less than 0.8 it would refuse to provide updates for `fwupd/(null)` which is what this PR fixes. It definitely needs fixing as you describe n discover, and TBH I completely forgot it was using libfwupd. tl;dr: call `fwupd_client_set_user_agent_for_package()` after connect has completed when we know the daemon version.

Both bugs are correct, well done for understanding the root cause.

AdamWill:
Roger. Note the [API doc for connect](https://fwupd.github.io/libfwupd/method.Client.connect.html) says "Sets up the client ready for use. Most other methods call this for you, and do you only need to call this if you are just watching the client." Seems like that's not really correct, right? If I was writing an fwupd client from scratch based on the API docs, I probably wouldn't grok that I need to do `connect()` and `set_user_agent()` or `set_user_agent_for_package()` before anything will really work, which seems to be the case, right? Perhaps the API docs could use updating?

hughsie:
So maybe it's best to just make it work both ways? i.e. *save* the package name and version and then only set the user_agent when the client has connected? That seems a whole less magic, and also means the sync-auto-connect stuff works easier.

AdamWill:
anything you can do on the fwupd side to make life easier for clients sounds like a win to me!

PR #5534, opened 2023-02-15
Fix -Dbuild=library target

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/fedora.sh
+++ b/contrib/ci/fedora.sh
@@ -16,7 +16,16 @@ if [ "$QUBES" = "true" ]; then
     QUBES_MACRO=(--define "qubes_packages 1")
 fi
 
+# smoke test something small with no auto-deps
 meson setup .. \
+    -Dbuild=library \
+    -Dauto_features=disabled \
+    -Dtests=false
+ninja-build
+rm -rf *
+
+# do the full-fat build
+meson setup \
     -Ddocs=disabled \
     -Dman=true \
     -Dtests=true \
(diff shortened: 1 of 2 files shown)


superm1 (commented):
Can we please add a ci target for library to make sure it doesn't break again?

hughsie:
> add a ci target

As in one more into the matrix? Or edit the Fedora one to re-do the build with `meson configure -Dbuild=library` perhaps?

hughsie:
> I think a new target is probably fine for this.

Actually, if we use the existing fedora container we can do a 23-target initial build as a smoke-test before doing all the other stuff -- that would quickly fail in the case we forgot to add a source file or something drastic.

contrib/ci/fedora.sh
```diff
@@ -16,7 +16,16 @@ if [ "$QUBES" = "true" ]; then
     QUBES_MACRO=(--define "qubes_packages 1")
 fi
 
+# smoke test something small with no auto-deps
 meson setup .. \
+    -Dbuild=library \
+    -Dauto_features=disabled \
+    -Dtests=false $@
```
superm1, line 23: do we actually need to pass arguments to this?

superm1 (dismissed):
one nit, otherwise LGTM

contrib/ci/fedora.sh
hughsie: Nope; removed, thanks.

PR #5535, opened 2023-02-15
uefi-capsule: Require at least 80MB free in the ESP

hughsie:
See https://github.com/fwupd/firmware-lenovo/issues/257

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
TBH, I think it would be better to make this dynamic based on the capsule size itself rather than SPI size.

hughsie:
> make this dynamic based on the capsule size

That's a much better idea. How about something like this? ^^

superm1 (approved):
Looks great, thanks!

PR #5540, opened 2023-02-16
trivial: Only call fu_engine_md_refresh_device_from_component once per device add

hughsie:


hughsie:
@superm1 i couldn't remember any good reason why we'd check twice.

superm1:
I do remember a need to do it twice, but I don't remember why.  Does the commit log give any insight?

hughsie:
@superm1 I guess it was e0384192400c83545cbe365667321e6bc83cec13 we're thinking about, but then that implies we're doing it wrong at the moment. I've fixed up the commit to set the verfmt before getting supported.

superm1 (dismissed):

PR #5541, opened 2023-02-16
Use a heuristic to set ignore-system-power for all Lenovo batteries

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -4512,6 +4512,23 @@ fu_engine_md_refresh_device_name_category(FuEngine *self, FuDevice *device, XbNo
 		fu_device_remove_internal_flag(device,
 					       FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME_CATEGORY);
 	}
+
+	/* batteries updated using capsules should ignore the system power restriction */
+	if (g_strcmp0(fu_device_get_plugin(device), "uefi_capsule") == 0) {
+		gboolean is_battery = FALSE;
+		for (guint i = 0; i < cats->len; i++) {
+			XbNode *n = g_ptr_array_index(cats, i);
+			if (g_strcmp0(xb_node_get_text(n), "X-Battery") == 0) {
+				is_battery = TRUE;
+				break;
+			}
+		}
+		if (is_battery) {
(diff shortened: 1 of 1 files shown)


src/fu-engine.c
```diff
@@ -4512,6 +4512,23 @@ fu_engine_md_refresh_device_name_category(FuEngine *self, FuDevice *device, XbNo
 		fu_device_remove_internal_flag(device,
 					       FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME_CATEGORY);
 	}
+
+	/* using GUIDs to set internal flag ignore-system-power doesn't scale well */
+	if (fu_device_has_vendor_id(device, "DMI:LENOVO")) {
```
superm1, line 4517: why restrict this to Lenovo?

src/fu-engine.c
```diff
@@ -4512,6 +4512,23 @@ fu_engine_md_refresh_device_name_category(FuEngine *self, FuDevice *device, XbNo
 		fu_device_remove_internal_flag(device,
 					       FU_DEVICE_INTERNAL_FLAG_MD_SET_NAME_CATEGORY);
 	}
+
+	/* using GUIDs to set internal flag ignore-system-power doesn't scale well */
```
superm1, line 4516: I think this comment should be adjusted

src/fu-engine.c
hughsie: I guess what i'm trying to do is restrict it to laptop batteries, but we don't have enough metadata to know the difference between a battery in a laptop and a battery in a wireless mouse for instance.

superm1: you could just only do it with the uefi-capsule plugin instead then perhaps?

src/fu-engine.c
hughsie: Fixed.

src/fu-engine.c
hughsie: Yup, that's a good heuristic I guess. Fixed.

hughsie:
> I don't recall if we have documentation for the X-Foo stuff in tree

LVFS yes, fwupd no. Thanks.

PR #5544, opened 2023-02-17
Add pcap file converter

fdanis-oss:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/pcap2emulation.py
+++ b/contrib/pcap2emulation.py
@@ -1,0 +1,611 @@
+#!/usr/bin/python3
+# pylint: disable=invalid-name,missing-docstring
+#
+# Copyright (C) 2023 Collabora Ltd
+# Author: Frédéric Danis <frederic.danis@collabora.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+import argparse
+import base64
+import json
+import os
+import subprocess
+import sys
+from typing import Any, Dict, List, Optional, Tuple
+from zipfile import ZipFile, ZIP_DEFLATED
+
+URB_INTERRUPT = 1
+URB_CONTROL = 2
+URB_BULK = 3
+
+DESCRIPTOR_DEVICE = "1"
+DESCRIPTOR_CONFIGURATION = "2"
+DESCRIPTOR_STRING = "3"
+DESCRIPTOR_INTERFACE = "4"
+DESCRIPTOR_ENDPOINT = "5"
+DESCRIPTOR_HID = "33"
+
+INTERFACE_CLASS_HID = 3
+
+
+def get_int(data: str) -> int:
+    if data[:2] == "0x":
+        return int(data, 16)
+    return int(data)
+
+
+class Pcap2Emulation:
+    def __init__(self, device_ids: str):
+        self.device: Dict[str, Any] = {}
+        self.platform_id = ""
+        self.phases: List[Any] = []
+        self.device_ids: List[List[str]] = []
+        self.interface_index = 0
+        self.endpoint_index = 0
+        self.previous_data: Optional[str]
+        self.bulk_incoming_lens: Dict[str, int] = {}
+        self.usb_port = None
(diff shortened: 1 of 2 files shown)


contrib/pcap2emulation.py
```diff
@@ -0,0 +1,361 @@
...
+
+if __name__ == "__main__":
+    if {"-?", "-h", "--help", "--usage"}.intersection(set(sys.argv)):
+        usage(0)
+
+    if len(sys.argv) < 4:
+        usage(1)
+
+    path = os.path.abspath(os.path.expanduser(os.path.expandvars(sys.argv[1])))
+    parser = Pcap2Emulation(sys.argv[2], sys.argv[3:])
```
superm1, line 360: why not use `argparse`?  It's a lot more flexible to expand arguments later and re-order and such.

docs/device-emulation.md
```diff
@@ -69,3 +69,33 @@ For example:
     Decompressing…           [***************************************]
     Waiting…                 [***************************************]
     Hughski ColorHug2: OK!
+
+## Pcap file conversion
+
+Emulation can also be used during the development phase of the plugin if the hardware is not
+available or to reduce the testing time. But this requires a way to create the emulation file while
```
superm1, line 76: probably also helps with the fact you don't need to do a write cycle with the device while writing the code too.   Nice!

contrib/pcap2emulation.py
```diff
@@ -0,0 +1,361 @@
...
+                                    self._save_event(event_bytes)
+
+                            if descriptor_type == DESCRIPTOR_INTERFACE:
+                                interface = {
+                                    "Length": get_int(layers["usb_usb_bLength"][-1]),
+                                    "DescriptorType": 4,
+                                }
+                                val = get_int(layers["usb_usb_bInterfaceNumber"])
+                                if val != 0:
+                                    interface["InterfaceNumber"] = val
```
superm1, line 259: stuff like this seems to beg a dictionary matching table.    Something like this:
```
table = {
"usb_usb_bInterfaceNumber" : "InterfaceNumber"
}
for o in table:
    val = get_int(layers[o])
    if val != 0:
        interface[o[table]] = val
```

contrib/pcap2emulation.py
```diff
@@ -0,0 +1,349 @@
...
+def get_int(str):
+    if str[:2] == "0x":
+        return int(str, 16)
+    return int(str)
+
+
+class Pcap2Emulation:
+    def __init__(self, prefix, device_ids):
+        self.device = None
+        self.prefix = prefix
```
hughsie, line 35: i'd feel a lot more warm and fuzzy if these had mypy annotations -- not a blocker by any means, but it's certainly saved me quite a few headaches when I wrote the LVFS...

hughsie:
Can you squash these down into one commit please?

hughsie (approved):
This looks great!

fdanis-oss:
@hughsie I'm not sure of the following change: https://github.com/fwupd/fwupd/compare/8e97a7c1c6f55098df39630b107dacd1801990d8..f0fdde5e0bdf58183f4082433552a513b19e4b7c

This works well for the emulation file generated for the Synaptics-Prometheus, allowing the firmware update to complete.

But for the emulation file generated for the CalDigit, even if the firmware update completes, this patch prevents the version update because the device is not re-enumerated during the reload phase. 

do you have any suggestion on how to detect the need for re-enumeration from the USB packets?

fdanis-oss:
I updated the re-enumeration requirement detection by tracking the `C_PORT_CONNECTION` clear feature packet.

It allows me to complete firmware update with emulation files generated from `CalDigit`, `ColorHug` and `Synaptics-Prometheus` pcap files.

hughsie (approved):

hughsie:
@fdanis-oss feel free to merge whenever you're happy.

Issue #5553, opened 2023-02-19
fwupd accepts metadata.xml.gz but saves it as metadata.xml.xz and crashes


marmarek:
**Describe the bug**
With `fwupdmgr refresh`, one can provide correct firmware.xml.gz with matching jcat file and it gets accepted. But fwupd then tries to load it as xz-compressed file and crashes.

**Steps to Reproduce**

1. Take fwupd-1.8.10-2.fc37.x86_64
2. Download `https://cdn.fwupd.org/downloads/firmware.xml.gz` and matching jcat (note it is **gz**, not **xz**, as an attempt to workaround https://gitlab.com/fwupd/lvfs-website/-/issues/860)
3. Call `fwupdagent refresh firmware.xml.gz firmware.xml.gz.jcat lvfs`

**Expected behavior**
Either metadata.xml.gz is correctly saved as gz file and interpreted as such, or rejected early.

**fwupd version information**
Please provide the version of the daemon and client.

```shell
fwupdmgr --version
compile   org.freedesktop.fwupd         1.8.10
compile   com.hughsie.libjcat           0.1.12
runtime   org.freedesktop.fwupd-efi     1.4
compile   org.freedesktop.gusb          0.4.3
runtime   com.hughsie.libjcat           0.1.12
runtime   com.dell.libsmbios            2.4
runtime   org.freedesktop.gusb          0.4.5
runtime   org.kernel                    6.1.12-1.qubes.fc37.x86_64
runtime   org.freedesktop.fwupd         1.8.10
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): default package in Fedora 37

The crash:
```
fwupd[8171]: ERROR:../src/xb-lzma-decompressor.c:132:xb_lzma_decompressor_convert: code should not be reached
fwupd[8171]: Bail out! ERROR:../src/xb-lzma-decompressor.c:132:xb_lzma_decompressor_convert: code should not be reached

#0  0x000075a420da1e5c __pthread_kill_implementation (libc.so.6 + 0x8ce5c)
#1  0x000075a420d51a76 raise (libc.so.6 + 0x3ca76)
#2  0x000075a420d3b7fc abort (libc.so.6 + 0x267fc)
#3  0x000075a420f64ef8 g_assertion_message.cold (libglib-2.0.so.0 + 0x1fef8)
#4  0x000075a420fc58ae g_assertion_message_expr (libglib-2.0.so.0 + 0x808ae)
[...]

PR #5555, opened 2023-02-19
Only accept application/x-xz compression for the metadata payload

hughsie:
Fixes some of https://github.com/fwupd/fwupd/issues/5553

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -4946,6 +4946,7 @@ fu_engine_update_metadata_bytes(FuEngine *self,
 	FwupdKeyringKind keyring_kind;
 	FwupdRemote *remote;
 	JcatVerifyFlags jcat_flags = JCAT_VERIFY_FLAG_REQUIRE_SIGNATURE;
+	g_autofree gchar *content_type = NULL;
 	g_autoptr(JcatFile) jcat_file = jcat_file_new();
 
 	g_return_val_if_fail(FU_IS_ENGINE(self), FALSE);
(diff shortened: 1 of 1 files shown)


superm1 (commented):
why not also support gzip? for flexibility?

hughsie:
> why not also support gzip

I think it makes it a lot less complicated to only support the one format; I don't think we want to get into the game of decompressing the gzip and compressing it in into a xz when updating the metadata. Using local files for refresh is kinda a debug function and I'd rather just fix the LVFS issue.

superm1:
> > why not also support gzip
> 
> I think it makes it a lot less complicated to only support the one format; I don't think we want to get into the game of decompressing the gzip and compressing it in into a xz when updating the metadata. Using local files for refresh is kinda a debug function and I'd rather just fix the LVFS issue.

OK sounds good.

hughsie:
The other point is we're kinda using the format as an API version -- e.g. in the gz we don't include <artifact> and in the xz we include the tested_by stuff -- i.e. optimizing the metadata contents for the fwupd version we know will support it. e.g. if we ever moved to zstd we could drop some of the artifact source fixups in that.

superm1 (approved):

marmarek:
How can I get from fwupd which format it expects (for `refresh` call)?

hughsie:
We only support xz now. You can grab that from https://cdn.fwupd.org/downloads/firmware.xml.xz

marmarek:
that I know, but given arbitrary fwupd version, how can I ask it which format it expects?

hughsie:
Hm, now that is a good question. I guess give it xz first, and if there's an error give it gz instead. I'm open for ideas.

marmarek:
Add some explicit info in `fwupd --version`? Or at least declare which version switched the accepted/default compressions? It was a build option for some time, right? if so, just declaring version might not be enough...
How many versions were accepting gz but really expecting xz?

marmarek:
For xz/gz it's probably too late anyway, but could help with future changes (zstd?).

hughsie:
I think for now, using `fwupdmgr --version --json` and then checking the fwupd version is the only thing you can do. 1.8.7 and newer use xz. Lets go back a bit; I think you're the only person other than me to use the _load local files as metadata_ functionality -- are you trying to work around something else we should fix?

marmarek:
I'm finishing/updating qubes integration. fwupd runs in dom0 where you don't have direct network access, only dedicated VM (aka "updatevm") can download stuff. So, a wrapper script downloads stuff in that dedicated VM, verifies its integrity (as extra precaution) and only then passes it to fwupd. It's done for both metadata and actual updates.

hughsie:
I'm hesitant to just say "we can just convert the .gz to .xz in fwupd" as the different metadata files have subtly different contents. Could you look at the MetadataURI in /etc/fwupd/remotes.d/lvfs.conf and use that perhaps?

PR #5560, opened 2023-02-20
Wip/hughsie/unsigned payload

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -6829,9 +6829,9 @@ fu_engine_add_device(FuEngine *self, FuDevice *device)
 	    !fu_device_has_flag(device, FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD) &&
 	    !fu_device_has_flag(device, FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD) &&
 	    !fu_device_has_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_MD_SET_SIGNED)) {
-		g_warning("%s device does not define payload %s [%s]",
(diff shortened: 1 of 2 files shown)


src/fu-engine.c
```diff
@@ -6829,9 +6829,9 @@ fu_engine_add_device(FuEngine *self, FuDevice *device)
 	    !fu_device_has_flag(device, FWUPD_DEVICE_FLAG_SIGNED_PAYLOAD) &&
 	    !fu_device_has_flag(device, FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD) &&
 	    !fu_device_has_internal_flag(device, FU_DEVICE_INTERNAL_FLAG_MD_SET_SIGNED)) {
-		g_warning("%s device does not define payload %s [%s]",
+		g_warning("%s [%s] does not declare signed/unsigned payload -- perhaps add "
+			  "fu_device_add_flag(device, FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);",
```
superm1, line 6833: Who does the warning actually help?  I would argue it's not really useful to an end user.  Maybe:

```suggestion
		g_warning("%s [%s] does not declare if it accepts unsigned payloads",
```

hughsie: I think we need to tell the programmer what do to.

superm1: But an end user will see this.  Can we maybe only show that warning when supported build isn't set?

superm1: Oh wait, this block is in that already.  Carry on :)

superm1 (approved):

Issue #3386, opened 2021-06-17
upower plugin / FWUPD_DEVICE_FLAG_REQUIRE_AC functionality : Support explicit Mains, or USB* power supply types as an indication of AC present


bleungatchromium:
**Is your feature request related to a problem? Please describe.**
The upower plugin uses the system's battery state to determine whether the system is on AC power. If the battery state is FU_BATTERY_STATE_DISCHARGING, it sets an error that we are not on AC power.

```
        /* not charging */
        if (fu_context_get_battery_state (ctx) == FU_BATTERY_STATE_DISCHARGING ||
            fu_context_get_battery_state (ctx) == FU_BATTERY_STATE_EMPTY) {
                g_set_error_literal (error,
                                     FWUPD_ERROR,
                                     FWUPD_ERROR_AC_POWER_REQUIRED,
                                     "Cannot install update "
                                     "when not on AC power unless forced");
                return FALSE;
        }

```

There are some systems that, even while attached to some sort of AC mains, will decide to allow the battery to go into a DISCHARGING state (either for battery health, or some other sort of system policy, or as a transient condition when ac is attached, and the battery object hasn't been polled for the new state yet, or possibly some combination of a weaker charger + high system consumption could result in DISCHARGING state).

This could lead to some cases where the AC power error is set while the AC is actually plugged in. 

**Describe the solution you'd like**

The power supply subsystem has different device types, including non-battery ones Mains, USB, etc. Instead of determining whether the system is on AC power by inference (ie, the battery is not discharging) we should use those more reliable power supplies, which gives a binary online = 1 or online = 0.

**Describe alternatives you've considered**

**Additional context**

See the comment in chromiumos's power supply properties.proto which addresses this possibility: https://chromium.googlesource.com/chromiumos/platform/system_api/+/HEAD/dbus/power_manager/power_supply_properties.proto#13

```
// For any of these power sources, the system may be consuming power at a high
// enough rate that the battery is discharging rather than charging; see
// BatteryState. See the PowerSource submessage's \|type\| field for more
// details about the active power source.
```

I'm unfamiliar with how upower presents other power supply information through dbus, so I don't know how easy it would be to add this in the upower plugin, but for the new chromiumos power manager plugin being worked on by @binaryNewts and @g2pop, this is something to consider. 

@djcampello as a fyi.

superm1:
[...]

PR #5561, opened 2023-02-20
Be more nuanced when considering if the system is on AC power

hughsie:
This should be no beahviour change, but it allows the powerd and upower plugins to provide more information in the future.

Fixes https://github.com/fwupd/fwupd/issues/3386

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-common.c
+++ b/libfwupdplugin/fu-common.c
@@ -219,34 +219,58 @@ fu_common_align_up(gsize value, guint8 alignment)
 }
 
 /**
- * fu_battery_state_to_string:
- * @battery_state: a battery state, e.g. %FU_BATTERY_STATE_FULLY_CHARGED
+ * fu_power_state_to_string:
+ * @power_state: a power state, e.g. %FU_POWER_STATE_AC_FULLY_CHARGED
  *
  * Converts an enumerated type to a string.
  *
  * Returns: a string, or %NULL for invalid
  *
- * Since: 1.6.0
+ * Since: 1.8.11
  **/
 const gchar *
-fu_battery_state_to_string(FuBatteryState battery_state)
+fu_power_state_to_string(FuPowerState power_state)
 {
-	if (battery_state == FU_BATTERY_STATE_UNKNOWN)
+	if (power_state == FU_POWER_STATE_UNKNOWN)
 		return "unknown";
-	if (battery_state == FU_BATTERY_STATE_CHARGING)
-		return "charging";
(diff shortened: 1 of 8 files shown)


libfwupdplugin/fu-common.c
```diff
@@ -219,34 +219,58 @@ fu_common_align_up(gsize value, guint8 alignment)
...
+	if (power_state == FU_POWER_STATE_BATTERY_DISCHARGING)
+		return "battery-discharging";
+	if (power_state == FU_POWER_STATE_BATTERY_EMPTY)
+		return "battery-empty";
+	if (power_state == FU_POWER_STATE_AC)
+		return "ac";
+	if (power_state == FU_POWER_STATE_AC_CHARGING)
+		return "ac-charging";
+	if (power_state == FU_POWER_STATE_AC_FULLY_CHARGED)
+		return "ac-fully-charged";
```
superm1, line 247: technically I think you're missing a state here of `ac-discharging` when you have a power adapter plugged in but it's small and battery is going down

superm1 (commented):
Conceptually looks good, but I think you're missing one state (which I later noticed you mentioned in a comment)

libfwupdplugin/fu-common.c
hughsie: Yup, I'm struggling with this one; if the AC adapter is plugged in, but the battery is actually _going down_ is that conceptually "on AC power"? I'd argue it's **relying** on the battery being present, and no longer on AC power.

superm1: you would see a little AC adapter plug in your DE and everywhere else in the kernel would report it to be plugged in.  So I'd say `ac-discharging`

hughsie: ...but would you expect it to allow firmware updates to start? I'm guessing the UEFI updater is only really checking the AC input ACPI input.

superm1: > I'm guessing the UEFI updater is only really checking the AC input ACPI input.

Yeah exactly.

hughsie: So in the AC plugged in case, but battery actually slowly discharging case - I'd argue you should set just FU_POWER_STATE_AC.

hughsie:
@superm1 you think include in 1.8.11 or punt to 1.8.12?

superm1 (approved):
It's mostly just string and symbol changing, I think include it.

PR #5568, opened 2023-02-21
Allow enabling low-level debugging of fwupdx64.efi from uefi_capsule.…

hughsie:
…conf

Short-to-medium-term we want to stop installing fwupdate and this was the last useful part we never migrated.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-device.h
+++ b/plugins/uefi-capsule/fu-uefi-device.h
@@ -84,6 +84,12 @@ typedef enum {
  * Do not prepend a plausible missing capsule header.
  */
 #define FU_UEFI_DEVICE_FLAG_NO_CAPSULE_HEADER_FIXUP (1 << 7)
+/**
+ * FU_UEFI_DEVICE_FLAG_ENABLE_EFI_DEBUGGING:
+ *
+ * Enable debugging the EFI binary.
+ */
+#define FU_UEFI_DEVICE_FLAG_ENABLE_EFI_DEBUGGING (1 << 8)
 
 FuUefiDeviceKind
 fu_uefi_device_kind_from_string(const gchar *kind);
@@ -130,3 +136,5 @@ void
 fu_uefi_device_set_status(FuUefiDevice *self, FuUefiDeviceStatus status);
 void
 fu_uefi_device_set_require_esp_free_space(FuUefiDevice *self, gsize require_esp_free_space);
+gboolean
+fu_uefi_device_perhaps_enable_debugging(FuUefiDevice *self, GError **error);
--- a/plugins/uefi-capsule/fu-uefi-nvram-device.c
+++ b/plugins/uefi-capsule/fu-uefi-nvram-device.c
@@ -89,11 +89,11 @@ fu_uefi_nvram_device_write_firmware(FuDevice *device,
 	if (!fu_bytes_set_contents(fn, fixed_fw, error))
 		return FALSE;
 
-	/* delete the logs to save space; use fwupdate to debug the EFI binary */
-	if (fu_efivar_exists(FU_EFIVAR_GUID_FWUPDATE, "FWUPDATE_VERBOSE")) {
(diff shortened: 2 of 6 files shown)


plugins/uefi-capsule/uefi_capsule.conf
```diff
@@ -1,5 +1,8 @@
 [uefi_capsule]
 
+# enable the low-level debugging of fwupdx64.efi to the FWUPDATE_DEBUG_LOG EFI variable
```
superm1, line 3: Perhaps we want a warning here that this is going to fill up the NVRAM store more quickly and should only be enabled for working on an issue?

plugins/uefi-capsule/fu-uefi-nvram-device.c
```diff
@@ -40,6 +73,10 @@ fu_uefi_nvram_device_get_results(FuDevice *device, GError **error)
 		return TRUE;
 	}
 
+	/* capture EFI binary debug output */
+	if (fu_device_has_private_flag(device, FU_UEFI_DEVICE_FLAG_ENABLE_DEBUGGING))
+		fu_uefi_nvram_device_capture_debugging(device);
```
superm1, line 78: I would think we want this even when GRUB is used too not just NVRAM.

plugins/uefi-capsule/uefi_capsule.conf
superm1: Also we need to mention it's not used for CoD

plugins/uefi-capsule/fu-uefi-device.h
```diff
@@ -84,6 +84,12 @@ typedef enum {
  * Do not prepend a plausible missing capsule header.
  */
 #define FU_UEFI_DEVICE_FLAG_NO_CAPSULE_HEADER_FIXUP (1 << 7)
+/**
+ * FU_UEFI_DEVICE_FLAG_ENABLE_DEBUGGING:
+ *
+ * Increase debugging level for the device.
+ */
+#define FU_UEFI_DEVICE_FLAG_ENABLE_DEBUGGING (1 << 8)
```
superm1, line 92: `FU_UEFI_DEVICE_FLAG_ENABLE_EFI_DEBUGGING` perhaps

plugins/uefi-capsule/fu-uefi-device.h
```diff
@@ -84,6 +84,12 @@ typedef enum {
  * Do not prepend a plausible missing capsule header.
  */
 #define FU_UEFI_DEVICE_FLAG_NO_CAPSULE_HEADER_FIXUP (1 << 7)
+/**
+ * FU_UEFI_DEVICE_FLAG_ENABLE_DEBUGGING:
+ *
+ * Increase debugging level for the device.
```
superm1, line 90: `Enable debugging the EFI binary` perhaps

hughsie: WFM, thanks.

superm1:
> Short-to-medium-term we want to stop installing fwupdate and this was the last useful part we never migrated.

I think as part of this PR you should tear out the logging functionality from `fwupdate` so that it can't conflict with doing it here.

Either this PR or another mark `fwupdate` as noinst.

plugins/uefi-capsule/fu-uefi-device.h
hughsie: Yup, that's better.

plugins/uefi-capsule/fu-uefi-nvram-device.c
hughsie: Ack, good catch. Fixed.

plugins/uefi-capsule/uefi_capsule.conf
hughsie: Something like this?

superm1 (approved):

PR #5569, opened 2023-02-21
Stop shipping compat_cli by default

superm1:
This just flips the default.  Distros that need it for real compatibility reason (like RHEL?) could still -`Dcompat_cli=true`

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/void.sh
+++ b/contrib/ci/void.sh
@@ -17,5 +17,6 @@ meson build \
  -Dconsolekit=disabled \
  -Dsystemd=disabled \
  -Doffline=disabled \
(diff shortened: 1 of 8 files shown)


hughsie (approved):

hughsie:
@superm1 you need to remove the binary from the .spec.in file too.

superm1:
> @superm1 you need to remove the binary from the .spec.in file too.

The issue was actually more places we weren't checking for `get_option('compat_cli')`.  I've fixed the rest and it passes now.

hughsie (approved):

PR #5572, opened 2023-02-22
trivial: Add some wisdom about self-signing updates for testing

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/docs/only-trusted.md
+++ b/docs/only-trusted.md
@@ -1,0 +1,110 @@
+---
+title: Signing Test Firmware Payloads
+---
+
+## Introduction
+
+In the normal vendor update flow, firmware is optionally signed and encrypted by the vendor, and
+then uploaded to the LVFS wrapped in a cabinet archive with a small XML metadata file to describe
+how the firmware should be matched to hardware.
+
+Upon uploading, the LVFS signs the firmware and metadata XML contained in the archive and adds it to
+a `.jcat` file also included in the image.
+The original firmware is never modified, which is why the [Jcat](https://github.com/hughsie/libjcat)
+file exists as both a detached checksum (SHA-1, SHA-256 and SHA-512) and a detached signature.
+The LVFS can add either GPG or PKCS#7 signatures in the Jcat file, and currently does *both* for
+maxmum compatibility with how client systems have been configured.
+
+The keys in `/etc/pki/fwupd` and `/etc/pki/fwupd-metadata` are used for per-system trust and are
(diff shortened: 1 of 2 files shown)


docs/only-trusted.md
```diff
@@ -0,0 +1,110 @@
...
+a `.jcat` file also included in the image.
+The original firmware is never modified, which is why the [Jcat](https://github.com/hughsie/libjcat)
+file exists as both a detached checksum (SHA-1, SHA-256 and SHA-512) and a detached signature.
+The LVFS can add either GPG or PKCS#7 signatures in the Jcat file, and currently does *both* for
+maxmum compatibility with how client systems have been configured.
+
+The keys in `/etc/pki/fwupd` and `/etc/pki/fwupd-metadata` are used for per-system trust and are
+currently used for "did the firmware update *come from* somewhere I trust" rather than "verify the
+vendor signed the update" -- on the logic the "signed the update" is already covered by the signature
+on the EFI capsule payload.
```
superm1, line 21: this is an assumption if this document is oriented for EFI capsules only.  Also if this is a valid assumption then perhaps we should be applying such a flag by `uefi-capsule` too.

docs/only-trusted.md
```diff
@@ -0,0 +1,110 @@
...
+an unmodified fwupd package (without uploading to the LVFS) then the following instructions can be
+modified to suit.
+
+First, lets verify that an existing firmware binary and metainfo file without a Jcat signature
+refuses to install when packaged into a cabinet archive:
+
+    $ gcab -c firmware.cab firmware.bin firmware.metainfo.xml 
+    $ fwupdmgr install firmware.cab --allow-reinstall
+    Decompressing…           [ -                                     ]
+    firmware signature missing or not trusted; set OnlyTrusted=false in /etc/fwupd/daemon.conf ONLY if you are a firmware developer
```
superm1, line 40: another way to do is is just `fwupdtool install-blob firmware.bin`.  Should that be a way to suggest too?

docs/only-trusted.md
hughsie: I've reworded it to talk about device signing and removed the reference to capsule updates. Better?

docs/only-trusted.md
superm1: Yeah I'm mostly thinking to mention it because sometimes you don't need the signing infrastructure but rather just the ability to install payloads.

superm1 (approved):
LGTM

PR #5580, opened 2023-02-23
Do not check for at least one loaded backend

hughsie:
If we're building for OpenBMC then we might not have USB -- and it's perfectly valid to just use plugin methods and not backends.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -33,8 +33,7 @@ jobs:
           python3 -m pip install --user "meson >= 0.60.0"
       - name: Build
         run: |
-          ./contrib/build-openbmc.sh
-          ninja -C ..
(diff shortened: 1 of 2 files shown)


superm1 (dismissed):

superm1:
Btw shouldn't we have caught this in open BMC CI test?

hughsie:
> Btw shouldn't we have caught this in open BMC CI test?

We only configure and build it, we don't actually run the daemon. Lets try and fix that.

hughsie:
@superm1 help me out with the exit code stuff please!

superm1:
> @superm1 help me out with the exit code stuff please!

above should work, go ahead and squash that together

hughsie:
Thanks @superm1, many thanks.

PR #5583, opened 2023-02-24
Allow setting the package user agent before the client has connected

hughsie:
This fixes a regression introduced in https://github.com/fwupd/fwupd/pull/5528 where we added a critical warning if setting the package user agent before we retrieved the fwupd daemon version.

If we rebuild the user agent whenever we get new information it does not matter in which order these are set.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-client.c
+++ b/libfwupd/fwupd-client.c
@@ -72,6 +72,8 @@ typedef struct {
 	GMutex proxy_mutex; /* for @proxy */
 	GDBusProxy *proxy;
 	GProxyResolver *proxy_resolver;
+	gchar *package_name;
+	gchar *package_version;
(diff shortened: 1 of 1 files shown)


superm1 (approved):

hughsie:
@superm1 any complaints if i do a new brown-paper-bag release with current git master?

superm1:
> @superm1 any complaints if i do a new brown-paper-bag release with current git master?

I think the stuff that we meant to slip for a while comes in sooner than expected.  IE
https://github.com/fwupd/fwupd/commit/a18deb7081f602e7a7f99150fa00bd3f30fb5c19
But hopefully we can sort that out.

Issue #5579, opened 2023-02-23
Snap shows error on first run


superm1:
When running snap for the first time due to the permissions for `redfish.conf` we have a failure occur.

```
$ fwupd.fwupdtool install-blob foo.bin
cp: cannot open '/snap/fwupd/4600/etc/fwupd/redfish.conf' for reading: Permission denied
```

We need to look at the files we copy on first run to make sure they all make sense.  Perhaps we shouldn't package `redfish.conf` by default in the snap.

cc @valentindavid

PR #5584, opened 2023-02-24
trivial: snap: only copy /etc and /var for tools that run as root

superm1:
we don't need to make a copy of /etc in ~ for fwupdmgr

fixes: #5579

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


hughsie (approved):
Makes sense, I guess.

PR #5586, opened 2023-02-24
fixed translation spacing

alba4k:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

Fixes how some translations in different languages were sometimes missing a space after a colon.

example:
`title:content` would have been shown to the user instead of `title: content`.

real-world example:
immagine

Should be safely mergable

### DIFF
--- a/po/ca.po
+++ b/po/ca.po
@@ -845,7 +845,7 @@ msgstr "Els dispositius que no s'han actualitzat correctament:"
 #. TRANSLATORS: message letting the user know no
 #. * device upgrade available due to missing on LVFS
 msgid "Devices with no available firmware updates: "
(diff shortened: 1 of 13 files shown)


hughsie:
I don't think this is going to work unfortunately; the .po files are authgenerated from Transifex.

alba4k:
What would be the best way to proceed to fix this issue? Should I maybe just open an issue?

hughsie:
I think translators have to use the transifex web UI I'm afraid.

alba4k:
@hughsie could you maybe share a link to that? I'd like to fix that since it's been bothering me for a while

hughsie:
@alba4k sure thing, here you go: https://www.transifex.com/freedesktop/fwupd/dashboard/

PR #5587, opened 2023-02-26
Don't mask autopkgtest failures

superm1:
autopkgtest is failing with 1.8.12.  I don't think it's a config problem (we set it in `contrib/debian/tests/ci`) but rather getting masked.  This should help get the correct error passed up.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Failure log: https://ci.debian.net/data/autopkgtest/testing/amd64/f/fwupd/31668401/log.gz

hughsie (approved):

PR #5589, opened 2023-02-27
trivial: use `Wants=` in fwupd-refresh unit

superm1:
systemd documentation indicates to use this for units starting after network is up. Whether that fixes the problem at hand I don't know but we should at least match the suggestion.

Link: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/
Link: https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/2008699

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/motd/fwupd-refresh.service.in
+++ b/data/motd/fwupd-refresh.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=Refresh fwupd metadata and update motd
 Documentation=man:fwupdmgr(1)
+Wants=network-online.target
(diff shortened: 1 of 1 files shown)


hughsie (approved):
If it fixes the bug, then LGTM.

superm1:
> If it fixes the bug, then LGTM.

Even if it doesn't, I think we should still do it to be consistent with systemd suggestions.

PR #5591, opened 2023-02-27
Actually use GLib logging correctly

hughsie:
In the old code we only used `g_warning()`, `g_critical()` and then `g_debug()` -- where the latter level was then filtered by a range of per-log-domain, per-plugin and other tangentially related environment variables.

Knowing how to make each debug line actually show up meant either knowing the secret incantation, or looking up and pasting from `docs/env.md` and adding enough keys until the line actually showed up. This made debugging with end users frustrating.

Rather than using env variables to filter debug, show messages and warnings by default, using `--verbose` to enable info messages, and then using `--verbose` **again** to enable all verbose debugging. Also promote useful messages to the INFO level so they show in the single `-v` output.

Additionally, filtering on the log domain with `--daemon-verbose` still works when you know exactly what you're looking for.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -803,12 +803,8 @@ device will only be updatable when all of them are removed.
 
 ## Debugging tips
 
-The most important rule when debugging is using the `--verbose` flag
-when running fwupd or fwupdtool. Besides that, there are many
-environment variables that allow some debug traces to be printed
-conditionally, for example: `FWUPD_PROBE_VERBOSE`,
-`FU_HID_DEVICE_VERBOSE`, `FWUPD_DEVICE_LIST_VERBOSE` and many other
-plugin-specific envvars.
+The most important rule when debugging is using the `--verbose` and
+duplicate `--verbose` flag when running fwupd or fwupdtool.
 
 ### Adding debug messages
 
--- a/plugins/lenovo-thinklmi/fu-self-test.c
+++ b/plugins/lenovo-thinklmi/fu-self-test.c
@@ -32,6 +32,15 @@ _plugin_device_added_cb(FuPlugin *plugin, FuDevice *device, gpointer user_data)
 }
 
 static gboolean
+fu_test_fatal_handler_cb(const gchar *log_domain,
+			 GLogLevelFlags log_level,
+			 const gchar *message,
+			 gpointer user_data)
+{
+	return log_level >= G_LOG_LEVEL_MESSAGE;
+}
+
+static gboolean
 fu_test_self_init(FuTest *self, GError **error_global)
 {
 	gboolean ret;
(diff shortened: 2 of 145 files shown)


docs/tutorial.md
```diff
@@ -803,12 +803,8 @@ device will only be updatable when all of them are removed.
...
 ## Debugging tips
 
-The most important rule when debugging is using the `--verbose` flag
-when running fwupd or fwupdtool. Besides that, there are many
-environment variables that allow some debug traces to be printed
-conditionally, for example: `FWUPD_PROBE_VERBOSE`,
-`FU_HID_DEVICE_VERBOSE`, `FWUPD_DEVICE_LIST_VERBOSE` and many other
-plugin-specific envvars.
+The most important rule when debugging is using the `--verbose` and
+duplicate `--verbose` flag when running fwupd or fwupdtool.
```
superm1, line 807: This feels pretty weird to have two long arguments make a different behavior.  Can we instead do a verbosity level as an optional argument?
```
--verbose
--verbose 1
--verbose 2
```

superm1:
Presumably fwupdtool/fwupdmgr should be adjusted to match this too.

superm1 (commented):
I think we might have something in /etc/fwupd/daemon.conf that needs changing for this too.

docs/tutorial.md
hughsie: I think repeating arguments is pretty common in low level UNIX tools -- the other way is like `libusb -vvv` but I don't know how to do single-dash multiple args. I'll experiment with an optional argument, but it's not quite as _catchy_ imho.

hughsie:
> Presumably fwupdtool/fwupdmgr should be adjusted to match this too.

No need; this is common code in fu-debug.c

hughsie:
> I think we might have something in /etc/fwupd/daemon.conf that needs changing for this too.

I forgot all about that, fixed.

hughsie:
Let's see what the world says. (non binding) :) https://mastodon.social/@hughsie/109938492009061390

docs/tutorial.md
hughsie: Ha, so GLib treats `-vv` the same as `-v -v` which means it works like lsusb.

superm1: > Ha, so GLib treats -vv the same as -v -v which means it works like lsusb.

Great!

plugins/lenovo-thinklmi/fu-self-test.c
```diff
@@ -39,7 +39,9 @@ fu_test_self_init(FuTest *self, GError **error_global)
 	g_autoptr(FuProgress) progress = fu_progress_new(G_STRLOC);
 	g_autoptr(GError) error = NULL;
 
+#if GLIB_CHECK_VERSION(2, 64, 0)
```
superm1, line 42: did the level of this message change?

hughsie: Nope, but I couldn't see how the test would pass with old versions...

superm1: Yeah I guess makes sense.

src/fu-debug.c
```diff
@@ -239,8 +260,11 @@ fu_debug_pre_parse_hook(GOptionContext *context, GOptionGroup *group, gpointer d
 	     "DOMAIN"},
 	    {NULL}};
 
-	/* add main entry */
-	g_option_context_add_main_entries(context, main_entries, NULL);
+	/* set from FuConfig */
+	if (g_strcmp0(g_getenv("FWUPD_VERBOSE"), "*") == 0)
```
superm1, line 264: I'd actually question if `VerboseDomains=` is useful.  It's a glorified grep.  Maybe we should just axe it all together?

superm1 (approved):
LGTM

src/fu-debug.c
hughsie: I've used it for profiling in the past -- i.e. not wasting cycles syncing everything, just the profiling.

hughsie: I'll re-profile and see if it makes it on the profile graph.

PR #5595, opened 2023-02-28
uefi-capsule: Use a CapsuleOnDisk filename supported by InsydeH2O

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-capsule/fu-uefi-cod-device.c
+++ b/plugins/uefi-capsule/fu-uefi-cod-device.c
@@ -134,6 +134,39 @@ fu_uefi_cod_device_get_results(FuDevice *device, GError **error)
 	return TRUE;
 }
 
+static gchar *
+fu_uefi_cod_device_get_indexed_filename(FuUefiDevice *self, GError **error)
+{
+	g_autofree gchar *esp_path = fu_volume_get_mount_point(fu_uefi_device_get_esp(self));
+	for (guint i = 0; i < 0xFFFF; i++) {
+		g_autofree gchar *basename = g_strdup_printf("CapsuleUpdateFile%04X.bin", i);
+		g_autofree gchar *cod_path =
+		    g_build_filename(esp_path, "EFI", "UpdateCapsule", basename, NULL);
+		if (!g_file_test(cod_path, G_FILE_TEST_EXISTS))
+			return g_steal_pointer(&cod_path);
+	}
+	g_set_error_literal(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INTERNAL,
+			    "all potential CapsuleUpdateFile file names are taken");
+	return NULL;
+}
+
+static gchar *
+fu_uefi_cod_device_get_filename(FuUefiDevice *self, GError **error)
+{
+	g_autofree gchar *esp_path = fu_volume_get_mount_point(fu_uefi_device_get_esp(self));
+	g_autofree gchar *basename = NULL;
+
+	/* InsydeH2O */
+	if (fu_device_has_private_flag(FU_DEVICE(self), FU_UEFI_DEVICE_FLAG_COD_INDEXED_FILENAME))
+		return fu_uefi_cod_device_get_indexed_filename(self, error);
+
+	/* fallback */
+	basename = g_strdup_printf("fwupd-%s.cap", fu_uefi_device_get_guid(self));
+	return g_build_filename(esp_path, "EFI", "UpdateCapsule", basename, NULL);
+}
+
 static gboolean
 fu_uefi_cod_device_write_firmware(FuDevice *device,
 				  FuFirmware *firmware,
@@ -142,7 +175,6 @@ fu_uefi_cod_device_write_firmware(FuDevice *device,
 				  GError **error)
 {
 	FuUefiDevice *self = FU_UEFI_DEVICE(device);
-	g_autofree gchar *basename = NULL;
 	g_autofree gchar *cod_path = NULL;
 	g_autofree gchar *esp_path = fu_volume_get_mount_point(fu_uefi_device_get_esp(self));
 	g_autoptr(GBytes) fw = NULL;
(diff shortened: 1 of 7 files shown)


hughsie:
100% untested.

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -147,6 +147,39 @@ fu_uefi_cod_device_get_results(FuDevice *device, GError **error)
 	return TRUE;
 }
 
+static gchar *
+fu_uefi_cod_device_get_filename_insyde(FuUefiDevice *self, GError **error)
+{
+	g_autofree gchar *esp_path = fu_uefi_device_get_esp_path(self);
+	for (guint i = 0; i < 0xFFFF; i++) {
+		g_autofree gchar *basename = g_strdup_printf("CapsuleUpdateFile%04X.bin", i);
```
superm1, line 155: Before making this type of change we need to confirm if this is an Insyde behavior or a Huawei behavior.  If it's a Huawei behavior we should quirk against something different.

superm1: Or maybe even go off a combination of Insyde and Huawei as Huawei may use another IBV in the future.

hughsie: For my future self:

    # HUAWEI MateStation S [BIOS]
    [1b7f560d-2646-4232-b526-b0b1897e6dcd]
    Flags = is-insyde

which maybe we should use instead? Also, is-insyde s a bad flag name.

vit9696: We tested this mechanism on Acer and Huawei machines. We found the mechanism present in Lenovo machines as well. This is not Huawei-specific but Insyde-specific.

superm1: These were with the fixed Insyde stack that we quirk around requiring a specific bit?  Or is this part of that original bug?

superm1:
```
# HUAWEI MateStation S [BIOS]
[1b7f560d-2646-4232-b526-b0b1897e6dcd]
Flags = is-insyde
```

How about CoDprefix = `foo`

hughsie: I do wonder if we should add those as quirk guids to the UEFI device.

superm1:
```
<firmware gtype="FuAcpiUefi">
  <flags>has-checksum</flags>
  <id>UEFI</id>
  <size>0x236</size>
  <data size="0x236">UEFI6.</data>
  <revision>0x1</revision>
  <oem_id>HUAWEI</oem_id>
  <oem_table_id>H2O BIOS</oem_table_id>
  <oem_revision>0x1</oem_revision>
  <is_insyde>true</is_insyde>
  <guid>9d4bf935-a674-4710-ba02-bf0aa1758c7b</guid>
</firmware>
```
Where is the `insyde_cod_status` key?
IE - are we trying to work around the Insyde bug?

hughsie: > are we trying to work around the Insyde bug

Nah, what maybe I'm trying to do is say "for **any** Insyde implementation of CoD use the XXXX.bin filename" -- but do you know if Lenovo is working with the fwupd-GUID.cap format too? i..e we got that from *somewhere*.

superm1: I've never seen a Lenovo system that uses Insyde first hand.  All the ones that i've seen use Phoenix.

hughsie: Do you know what glob Phoenix uses? e.g. I know a few of the aarch64 ones use `*.cap` for example.

superm1: Phoenix doesn't support it on those systems that I've seen

hughsie: I think we're going to need these heuristics; probably adding them when we discover them rather than in any planned way. You don't have any insight into AMI, do you?

superm1: I've only ever seen it on Insyde.

hughsie: @apalos anything you can add from an aarch64 point of view?

plugins/uefi-capsule/fu-uefi-cod-device.c
```diff
@@ -134,6 +134,39 @@ fu_uefi_cod_device_get_results(FuDevice *device, GError **error)
...
+		g_autofree gchar *cod_path =
+		    g_build_filename(esp_path, "EFI", "UpdateCapsule", basename, NULL);
+		if (!g_file_test(cod_path, G_FILE_TEST_EXISTS))
+			return g_steal_pointer(&cod_path);
+	}
+	g_set_error_literal(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INTERNAL,
+			    "cannot find capsule update file idx that does not exist");
```
superm1, line 151: If this ever happens; this is a really confusing message.
```suggestion
			    "all potential CapsuleUpdateFile file names are taken");
```

superm1 (commented):
I don't personally like all this leaking this into `fu-uefi-capsule-plugin.c`.  Can we just have a helper we call into instead to tell if we need the private flag?

hughsie:
@superm1 we could the `detected InsydeH2O` bit out to the COD device if we provided a getter for `self->acpi_uefi` -- or do you want something more drastic?

superm1:
> @superm1 we could the `detected InsydeH2O` bit out to the COD device if we provided a getter for `self->acpi_uefi` -- or do you want something more drastic?

I looked again, and I think a lot of it is fine, I just want to see this bit moved out into a helper

```
	if (self->acpi_uefi != NULL &&
	    g_strcmp0(fu_acpi_uefi_get_guid(FU_ACPI_UEFI(self->acpi_uefi)), FU_EFI_INSYDE_GUID) ==
		0) {
		fu_device_add_private_flag(FU_DEVICE(dev),
					   FU_UEFI_DEVICE_FLAG_COD_INDEXED_FILENAME);
```

Similar to how we have a helper like: `fu_acpi_uefi_cod_functional()`

superm1:
I pushed my suggestion to the branch. If it looks good to you, please squash everything and it's fine by me.

PR #5598, opened 2023-03-01
Optionally retain firmware in a backup remote

hughsie:
This allows the device to set the flag and then 'fwupdmgr reinstall' works correctly without internet access. This is probably a good idea for network devices that have different loadable firmwares depending on configuration, e.g. inserted SIM card.

Fixes https://github.com/fwupd/fwupd/discussions/5592

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.h
+++ b/libfwupdplugin/fu-device.h
@@ -493,6 +493,20 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_NO_PROBE_COMPLETE (1ull << 27)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_SAVE_INTO_BACKUP_REMOTE:
+ *
+ * Save the cabinet archive to persistent storage remote before starting the update process.
+ *
+ * This is useful when the network device is being updated, and different blobs inside the archive
+ * could be required in different scenarios. For instance, if the user installs a firmware update
+ * for a specific network device and then changes the SIM -- it might be they need the archive
+ * again and have no internet access.
+ *
+ * Since: 1.8.13
+ */
+#define FU_DEVICE_INTERNAL_FLAG_SAVE_INTO_BACKUP_REMOTE (1ull << 28)
+
 /* accessors */
 gchar *
 fu_device_to_string(FuDevice *self);
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -2999,6 +2999,47 @@ fu_engine_schedule_update(FuEngine *self,
 	return fu_engine_offline_setup(error);
 }
 
+static gboolean
+fu_engine_save_into_backup_remote(FuEngine *self, GBytes *fw, GError **error)
+{
+	FwupdRemote *remote_tmp = fu_remote_list_get_by_id(self->remote_list, "backup");
+	g_autofree gchar *localstatepkg = fu_path_from_kind(FU_PATH_KIND_LOCALSTATEDIR_PKG);
+	g_autofree gchar *backupdir = g_build_filename(localstatepkg, "backup", NULL);
+	g_autofree gchar *backupdir_uri = g_strdup_printf("file://%s", backupdir);
+	g_autofree gchar *remotes_path = fu_path_from_kind(FU_PATH_KIND_LOCALSTATEDIR_REMOTES);
+	g_autofree gchar *remotes_fn = g_build_filename(remotes_path, "backup.conf", NULL);
+	g_autofree gchar *archive_checksum = g_compute_checksum_for_bytes(G_CHECKSUM_SHA256, fw);
+	g_autofree gchar *archive_basename = g_strdup_printf("%s.cab", archive_checksum);
+	g_autofree gchar *archive_fn = g_build_filename(backupdir, archive_basename, NULL);
+	g_autoptr(FwupdRemote) remote = fwupd_remote_new();
+
+	/* save archive if required */
+	if (!g_file_test(archive_fn, G_FILE_TEST_EXISTS)) {
+		g_info("saving archive to %s", archive_fn);
+		if (!fu_bytes_set_contents(archive_fn, fw, error))
+			return FALSE;
+	}
+
+	/* already exists as an enabled remote */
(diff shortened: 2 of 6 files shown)


hughsie:
Notably, this adds the files:

    /var/lib/fwupd/remotes.d/backup.conf
    /var/lib/fwupd/backup/foo.cab

fwupd-backup might be a better prefix. Ideas welcome.

libfwupdplugin/fu-device.h
```diff
@@ -493,6 +493,20 @@ typedef guint64 FuDeviceInternalFlags;
...
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_RETAIN_FIRMWARE_ARCHIVE:
+ *
+ * Save the cabinet archive to persistent storage before starting the update process.
+ *
+ * This is super useful when the network device is being updated, and different blobs inside the
+ * archive could be required in different scenarios. For instance, if the user installs a firmware
+ * update for a specific network device and then changes the SIM -- it might be we need the archive
+ * again and have no internet access.
```
superm1, line 504: not we, they

libfwupdplugin/fu-device.h
```diff
@@ -493,6 +493,20 @@ typedef guint64 FuDeviceInternalFlags;
...
+ *
+ * Since: 1.8.13
+ */
+#define FU_DEVICE_INTERNAL_FLAG_RETAIN_FIRMWARE_ARCHIVE (1ull << 28)
```
superm1, line 508: How about `CACHE_FIRMWARE_ARCHIVE`

src/fu-engine.c
```diff
@@ -2999,6 +2999,47 @@ fu_engine_schedule_update(FuEngine *self,
...
+			return FALSE;
+	}
+
+	/* already exists as an enabled remote */
+	if (remote_tmp != NULL && fwupd_remote_get_enabled(remote_tmp))
+		return TRUE;
+
+	/* just enable */
+	if (remote_tmp != NULL) {
+		g_debug("enabling remote %s", fwupd_remote_get_id(remote_tmp));
```
superm1, line 3029: with the new debugging stuff shouldn't this be `g_info`?

superm1:
Rather than making this a property of the device how about making it a property of the daemon and set some sort of cache expiration policy?

My thought process is that all firmware that is downloaded by the daemon may need to be reinstalled if a problem happens.  
Hypothetically what if I upgraded the firmware for my dock and my LAN in the dock broke?

libfwupdplugin/fu-device.h
```diff
@@ -493,6 +493,20 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_NO_PROBE_COMPLETE (1ull << 27)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_RETAIN_FIRMWARE_ARCHIVE:
+ *
+ * Save the cabinet archive to persistent storage before starting the update process.
+ *
+ * This is super useful when the network device is being updated, and different blobs inside the
```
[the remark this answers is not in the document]
hughsie: Done.

libfwupdplugin/fu-device.h
hughsie: Fixed.

libfwupdplugin/fu-device.h
hughsie: It's not a cache -- which implies you can just nuke it without consequence. Maybe `BACKUP_FIRMWARE_ARCHIVE`?

src/fu-engine.c
hughsie: It should indeed. Hoisted by my own petard. Fixed.

hughsie:
> Rather than making this a property of the device how about making it a property of the daemon and set some sort of cache expiration policy?

I considered this, but in the 99% of cases we just download it from the LVFS again. It's only really mobile broadband where this "firmware per carrier" exists -- although I've heard something of the same for headsets that have a firmware per-locale. The problem with a cache expiration policy is that you'd want mobile broadband to be "never" so you'd need a flag to say "never drop this" and in the common case we don't want to add 32MB to the mutable FS every firmware update we do.

> Hypothetically what if I upgraded the firmware for my dock and my LAN in the dock broke?

I think in that case we'd want to save the *old* archive, not the current one, no? It's a related problem, but not for this PR IMHO.

superm1:
> I considered this, but in the 99% of cases we just download it from the LVFS again. It's only really mobile broadband where this "firmware per carrier" exists -- although I've heard something of the same for headsets that have a firmware per-locale. The problem with a cache expiration policy is that you'd want mobile broadband to be "never" so you'd need a flag to say "never drop this" and in the common case we don't want to add 32MB to the mutable FS every firmware update we do.

Got it.

> I think in that case we'd want to save the old archive, not the current one, no? It's a related problem, but not for this PR IMHO.

Yeah I guess this case is really a "download my current version" to a cache.

src/fu-engine.c
```diff
@@ -2999,6 +2999,47 @@ fu_engine_schedule_update(FuEngine *self,
...
+		fwupd_remote_set_enabled(remote_tmp, TRUE);
+		return fwupd_remote_save_to_filename(remote_tmp, remotes_fn, NULL, error);
+	}
+
+	/* create a new remote we can use for re-installing */
+	g_debug("creating new backup remote");
```
superm1, line 3035: also `g_info` here

hughsie: Fixed.

libfwupdplugin/fu-device.h
hughsie: Maybe _BACKUP_ARCHIVE_REMOTE or _ARCHIVE_BACKUP_REMOTE or _USE_BACKUP_REMOTE or _SAVE_BACKUP_REMOTE perhaps?

superm1: I think _SAVE_BACKUP_REMOTE is my favorite

hughsie: _SAVE_INTO_BACKUP_REMOTE is even better, thanks. Fixed.

superm1 (approved):

PR #5600, opened 2023-03-03
Revert "Convert the man pages to MarkDown format"

hughsie:
This reverts commit a18deb7081f602e7a7f99150fa00bd3f30fb5c19.

I've since found that pandoc is unavailable on two of the distros we care about, and has a pretty huge and complicated dep chain on another two.

Man pages are not an important enough source of documenation for this dep, and better would be if the man pages just pointed to other local or online resources that may be easier to consume.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


plugins/dfu/dfu-tool.1
```diff
@@ -0,0 +1,31 @@
...
+libdfu supports DFU 1.0, DFU 1.1 and the ST DfuSe vendor extension, and
+handles many device `quirks' necessary for the real-world implementations
+of DFU\&.
+.SH OPTIONS
+The dfu-tool command takes various options depending on the action.
+Run \fBdfu-tool --help\fR for the full list.
+.SH SEE ALSO
+fwupdtool(1), fwupdmgr(1)
+.SH BUGS
+No known bugs.
```
superm1, line 29: I think when reverting it at least you should pick up the newly generated text from the commit

hughsie: Done, phew.

superm1 (approved):

PR #5604, opened 2023-03-07
Wip/hughsie/trusted vendor local

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


mangelajo:
Works well on my side

data/remotes.d/vendor.conf
```diff
@@ -3,6 +3,10 @@
 # @datadir@/fwupd/remotes.d/vendor and firmware in @datadir@/fwupd/remotes.d/vendor/firmware
 Enabled=false
 Title=Vendor
+# using `Keyring=none` is only designed for local firmware installed to an immutable
+# location, and should not be used when the metadata can be written by an untrusted user
```
superm1, line 7: This can probably be codified as a warning by checking if filesystem is RW.

superm1 (approved):

data/remotes.d/vendor.conf
hughsie: immutable as root, or as the user? my /usr/share is owned by root and mutable, but I still think directory remotes should work. If someone has root on my machine there's a lot more interesting things to do than install unsigned firmware -- which they can do with `fwupdtool install-blob` anyway...

PR #5614, opened 2023-03-09
tpm: Mark the device as updatable if it supports Field Upgrade

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/tpm/fu-tpm-v2-device.c
+++ b/plugins/tpm/fu-tpm-v2-device.c
@@ -223,6 +223,55 @@ fu_tpm_v2_device_setup_pcrs(FuTpmV2Device *self, ESYS_CONTEXT *ctx, GError **err
 }
 
 static gboolean
+fu_tpm_v2_device_ensure_commands(FuTpmV2Device *self, ESYS_CONTEXT *ctx, GError **error)
+{
+	gboolean seen_upgrade_data = FALSE;
+	gboolean seen_upgrade_start = FALSE;
+	TSS2_RC rc;
+	g_autofree TPMS_CAPABILITY_DATA *capability = NULL;
+	g_autoptr(GString) str = g_string_new(NULL);
+
+	rc = Esys_GetCapability(ctx,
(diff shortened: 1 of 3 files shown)


zaolin:
Add ReadFirmware commands for verification as well

plugins/tpm/fu-tpm-v2-device.c
```diff
@@ -222,6 +222,55 @@ fu_tpm_v2_device_setup_pcrs(FuTpmV2Device *self, ESYS_CONTEXT *ctx, GError **err
...
+			seen_upgrade_start = TRUE;
+		} else if (cap_cmd == TPM2_CC_FieldUpgradeData) {
+			seen_upgrade_data = TRUE;
+		}
+	}
+	g_debug("CAP_COMMANDS: %s", str->str);
+
+	/* both available */
+	if (seen_upgrade_start && seen_upgrade_data)
+		fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
```
superm1, line 268: presumably we actually need a routine that can do updates to add this flag

hughsie:
> Add ReadFirmware commands for verification as well

Done.

plugins/tpm/fu-tpm-v2-device.c
hughsie: Yup, I've just added this although it's 100.00% untested.

hughsie:
So this PR gets us 70% the way there -- the vendors would still need to tell us how to set up the transfer. I can't find any vendor that does TPM updates using the standardized API, and so this code is more of a "this is how it should be implemented" rather than with any expectation it's going to work. Unless that is the vendor allows flashing unsigned firmware, which for a TPM would be **bad**.

hughsie:
> so this code is more of a "this is how it should be implemented" rather

It's also to show Infineon how to do it with their vendor update flow...

PR #5626, opened 2023-03-15
Add support for deserializing smart structs like Rust

hughsie:
This makes it much easier to use..

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 any comments on the concept? I'd like to convert a few more plugins but don't want to waste my time if you hate it.

superm1:
> @superm1 any comments on the concept? I'd like to convert a few more plugins but don't want to waste my time if you hate it.

Sorry, I been out, and then seeing big PR while catching up in stuff it's been at the bottom of my list.

libfwupdplugin/fu-struct.c
```diff
@@ -0,0 +1,1465 @@
...
+}
+
+/**
+ * fu_struct_get_id_size:
+ * @self: a #FuStruct
+ * @id: struct ID, e.g. `vendor_id`
+ *
+ * Gets the size of the struct ID.
+ *
+ * NOTE: It is a programmer to attempt to get a structure ID that does not exist.
```
superm1, line 834: programmer error

libfwupdplugin/fu-struct.c
```diff
@@ -0,0 +1,1465 @@
...
+}
+
+/**
+ * fu_struct_get_u8:
+ * @self: a #FuStruct
+ * @id: struct ID, e.g. `vendor_id`
+ *
+ * Gets a #guint8 value.
+ *
+ * NOTE: It is a programmer to attempt to get a structure ID that does not exist, or to try to read
```
superm1, line 860: programmer error

hughsie:
Ya, no panic at all, sorry.

libfwupdplugin/fu-struct.c
```diff
@@ -0,0 +1,1465 @@
...
+ *
+ * A smart structure that supports endian conversion, arrays, GUIDs, strings, default and constant
+ * data of variable size.
+ *
+ * In most cases the smart structure will be defined in the #GObject `_init()` function:
+ *
+ *    fu_struct_register(self,
+ *                       "StructureName {"
+ *                       "    member_id: u16le,"
+ *                       "}");
```
superm1, line 31: I don't really like the defining how it looks in a big string and parsing that.  Can we not instead just have a series of calls that define the name of the structure and the endiannness and type of each member?

libfwupdplugin/fu-struct.c
```diff
@@ -0,0 +1,1465 @@
...
+ *
+ * ..and binary writing functions like this:
+ *
+ *    FuStruct *st = fu_struct_lookup(self, "StructureName");
+ *    fu_struct_set_u16(st, "member_id", value);
+ *    fu_struct_pack_into(st, buf);
+ *
+ * NOTE: is is very important to only read and write the correct member ID type and non-constant
+ * values. For instance, it is invalid to read a `u16le` type into a `guint32` value using
+ * `fu_struct_get_u32()`.
```
superm1, line 48: It's going to be harder to catch bugs because of this I think.

superm1 (commented):
I guess it's a poor man's rust?  Why not instead implement more rust stuff in the codebase?

libfwupdplugin/fu-struct.c
hughsie: I actually tried that; it gets very verbose. You need to essentially have FuStructItem exported, and then have getters and setters for the id, the type, the endian, the default value, the flags to indicate it's constant. It's at least 3 lines per struct field. Plus, I'd quite like to keep them in Rust format so that when we do start using it in core code we can just use the struct without big changes.

Even one line it's `fu_struct_add_member(st, "member_id", FU_STRUCT_FMT_U16, G_LITTLE_ENDIAN, NULL, FU_STRUCT_FLAG_NONE)` which gets split by clang-format into 6 lines....

libfwupdplugin/fu-struct.c
```diff
@@ -0,0 +1,1465 @@
...
+}
+
+/**
+ * fu_struct_get_id_offset:
+ * @self: a #FuStruct
+ * @id: struct ID, e.g. `vendor_id`
+ *
+ * Gets the offset to the struct ID.
+ *
+ * NOTE: It is a programmer to attempt to get a structure ID that does not exist.
```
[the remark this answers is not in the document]
hughsie: Fixed, thanks.

libfwupdplugin/fu-struct.c
hughsie: Fixed, and a few dozen more.

libfwupdplugin/fu-struct.c
hughsie: The other thing we could do is put the blobs of struct into a `fu-fmap-firmware.struct` file and then process them at build time to have `guint16 fu_fmap_header_get_nareas(st);`, `fu_fmap_header_parse(st, buf, bufsz, error)`, `GByteArray fu_fmap_header_pack(st)` in generated fu-fmap-header.c and fu-fmap-header..h files. That's dangerously close to just re-inventing a subset of protobuf but without any of the chaos. If you think that makes sense say and I can prototype something.

superm1: Why not just use protobuf then?

hughsie: I spent the morning trying to use protobuf. I've opened https://github.com/fwupd/fwupd/pull/5639 with a similar idea that might work better for us.

superm1:
I think we're leaning upon #5639 instead, closing this one.

PR #5630, opened 2023-03-16
Add the latest dbx version version fixups

hughsie:
This is required because again Microsoft removed hashes in the x64 DBXUpdate 20230314 update.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
As this keeps happening, I wonder if we need a way to distribute this from LVFS as well.

hughsie:
I'm not entirely sure how...

superm1:
> I'm not entirely sure how...

Publishing something with the metadata to describe how many entries should be there for this and consuming that.

PR #5634, opened 2023-03-17
Wip/hughsie/cleanup

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-tool.c
+++ b/src/fu-tool.c
@@ -1789,8 +1789,9 @@ fu_util_attach(FuUtilPrivate *priv, gchar **values, GError **error)
 static void
 fu_util_report_metadata_to_string(GHashTable *metadata, guint idt, GString *str)
 {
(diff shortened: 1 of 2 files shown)


src/fu-tool.c
```diff
@@ -1853,6 +1855,17 @@ fu_util_get_report_metadata(FuUtilPrivate *priv, gchar **values, GError **error)
 			fu_util_report_metadata_to_string(metadata_post, 3, str);
 		}
 	}
+
+	/* plugin metadata */
+	plugins = fu_engine_get_plugins(priv->engine);
```
superm1, line 1860: If you're getting all the plugins metadata, I wonder if you should also get all the enabled plugins for the report too.

superm1 (approved):

src/fu-tool.c
hughsie: Ya, but not for this PR. Thx.

PR #5635, opened 2023-03-18
wacom-usb: add touch id7 support to wacom-usb

Joshua-Dickens:
Support for updating new Wacom touch id7 firmware that was introduced with the Cintiq Pro 27.

Type of pull request:

- [ ] Code fix
- [x] Feature (new submodule)
- [ ] Documentation


plugins/wacom-usb/fu-wac-device.c
```diff
@@ -784,6 +785,14 @@ fu_wac_device_add_modules(FuWacDevice *self, GError **error)
 			fu_device_set_name(FU_DEVICE(module), name);
 			fu_device_set_version_from_uint16(FU_DEVICE(module), ver);
 			break;
+		case FU_WAC_MODULE_FW_TYPE_TOUCH_ID7:
+			module = fu_wac_module_touch_id7_new(FU_DEVICE(self));
+			name = g_strdup_printf("%s [Touch Module (ID7)]",
```
superm1, line 790: I think the stuff in [] probably should be in description rather than title.

plugins/wacom-usb/fu-wac-module-touch-id7.c
```diff
@@ -0,0 +1,362 @@
...
+#include <string.h>
+
+#include "fu-wac-device.h"
+#include "fu-wac-module-touch-id7.h"
+
+struct _FuWacModuleTouchId7 {
+	FuWacModule parent_instance;
+};
+
+struct WTAInfo {
```
hughsie, line 21: this should be a `typedef struct` with name `WtaInfo`

plugins/wacom-usb/fu-wac-module-touch-id7.c
```diff
@@ -0,0 +1,362 @@
...
+ *   u32:    Segment Size
+ *   u8:     IC_ID
+ *   u8:     MA_ID
+ *   u8[]:   2 Bytes padding to bring IC_ID/MA_ID to a multiple of 4 bytes
+ *   u32:    Block Count
+ * }
+ * 
+ */
+static void
+readWTARecordHeader(struct WTARecordHeader *header, struct WTAInfo *info) {
```
hughsie, line 99: can these all be `camel_case` with the correct `fu_whatever` prefixes please.

plugins/wacom-usb/fu-wac-module-touch-id7.c
```diff
@@ -0,0 +1,362 @@
...
+				       FU_WAC_MODULE_ERASE_TIMEOUT,
+				       error))
+		return FALSE;
+	fu_progress_step_done(progress);
+
+	/* Set basic info */
+	info.data = g_bytes_get_data(fwRead, &info.len);
+	info.op_id = 1;
+
+	readWTAFileHeader(&file_hdr, &info);
```
hughsie, line 313: this really isn't in the GLib style. Please return boolean, and accept a GError as the last parameter. You can pass `WTAFileHeader` by ref. I also wonder if this makes sense to split off this as a `FuWacModuleTouchId7Firmware` object -- and then we can unit test parsing it.

superm1:
### Thanks for your commit

**Clang Format**
 
**Pre-commit overview**

Joshua-Dickens:
Thanks for the feedback, I'll fix all the styling issues and follow the instructions provided to get it up to spec with the styling standards.

plugins/wacom-usb/fu-wac-device.c
Joshua-Dickens: I had used the Bluetooth ID6 as an example for the naming part of this submodule, if this is something we want to change should we change the other examples too?

superm1: Yeah I think so. Generally we want it to be useful and not confusing to people that see it.

plugins/wacom-usb/fu-wac-module-touch-id7.c
Joshua-Dickens: For the read functions would it warrant returning a boolean or passing GError? The most they do is move the data object forward or call fu_memread_uintXX methods. There's not much to catch in terms of failures from what I can think of.

hughsie: I think moving it to subclass FuFirmware will allow the` ->parse()` vfunc to do a lot of the heavy lifting.

Joshua-Dickens: Is there a good example of a FuFirmware subclass I can look at to better understand it and how it's best implemented?

hughsie: I think any of the existing FuFirmware subclasses that parse binary data would be fine -- e.g. FuFmapFirmware or FuFdtFirmware. There are a ton of developer docs here: https://fwupd.github.io/libfwupdplugin/tutorial.html#creating-a-new-firmware-type and https://fwupd.github.io/libfwupdplugin/class.Firmware.html

hughsie: @Joshua-Dickens would you like me to add you some template code?

Joshua-Dickens: Quick before that, looking through what you sent parse is meant to go over the entire file first, then write takes care of the various images gathered by the parse. 

But the way we settled on writing this firmware code to be the simplest/most readable was to parse as we go such as.
(File Header Parse) -> Individual Firmware -> (Firmware Header Parse) -> Individual Blocks -> (Block Header Parse) -> Write

We initially were parsing it all at once then writing but because the new file type we're handling (wtg) can have any number of firmware and any number of blocks in each firmware along with the commands being sent to the tablet relying on information from the firmware header and the block header + data, we found that parsing alongside the writing of each block made for cleaner overall code.

I can work it back to parsing it all at once and then writing it, but I first want to ask if there is a way to keep the process of how we're doing it right now while fitting it into your suggested changes.

hughsie: What we normally do is parse the entire file (to verify it's complete and compatible) and then write the prepared object to the device. If you're asking the device for the next block that's harder, as we still need to do an accurate percentage-complete signal.

Joshua-Dickens: With the new firmware it's a bit harder to verify things are correct with how the new firmware type (.wtg) is structured.

How the file is structured means we only get that information as we go, so when we read the file header it tells us how many firmware images we have but not how many blocks we have, it's not until we get to the header of the individual firmware that we learn how many blocks are in that specific firmware image, and it isn't until we go to the individual block header that we get the size for the block that we have no real way of verifying since there's no distinct symbol or mark of an end for the block and the start of a new block or firmware. And since there's no distinct symbol/data to identify the firmware images/blocks there's no good way to check if we have the proper number of firmware/blocks in the file in the first place. 

The only static symbol in the file that will always be constant is a string that reads "WTA" at the very beginning of the file. Everything else is values we have no static expected values for meaning we have to trust the data we get during parse.

For verification that the data is correct we can only really rely on the start and end commands we're sending to the tablet for each individual firmware image with which the tablet verifies if the firmware image wrote correctly and give a response if it did, or does not respond and cause a timeout which has fwupd abort the update properly and the tablet will boot into a safe mode which disables touch until a clean touch firmware install happens but otherwise doesn't affect functionality of the tablet.

Originally we wanted to parse everything at first, but since there wasn't any real way to truly verify the data is good/complete it made less sense to us to keep the process separate and instead to make the code simpler and more readable we would write the blocks as we got to them and let the tablet verify using the additional start and end commands for each firmware image. That way if for some reason the data within the block got corrupted or if the data in the header was wrong, the tablet would catch it and abort safely.

If you want to preserve the styling of parsing everything first then writing I absolutely can swap it over to that style. But I just wanted to quick give you an overview of why we swapped away from that and the difficulties in verifying the data so that you have the full context of why we made the decisions we did in how we structured the code.

hughsie: Okay -- it's your plugin and you can do what you want. But -- I still think a custom FuFirmware subclass makes sense as then you can pre-parse the format and get everything in easy-to-push packets of data for the actual write -- and you can also test and fuzz the parser easily as we need to be careful of malicious data being fed to fwupd.

superm1:
It seems like it's still failing the pre-commit style hook.  What's your host environment you're working from?

Joshua-Dickens:
Fedora 37 (6.2.9-200.fc37.x86_64)

Is there anything I can do locally to see why it failed in the CI?
I setup the pre commit hook as instructed and made sure it passed for this commit before pushing so I'm confused on why the CI test for it failed.

superm1:
> Fedora 37 (6.2.9-200.fc37.x86_64)
> 
> Is there anything I can do locally to see why it failed in the CI?
> I setup the pre commit hook as instructed and made sure it passed for this commit before pushing so I'm confused on why the CI test for it failed.
> 

Hmm. Try to reset your branch to main and recommit.

# git reset origin/main
# git add foo
# git commit

Hopefully it triggers the same thing.

Joshua-Dickens:
> > Fedora 37 (6.2.9-200.fc37.x86_64)
> > Is there anything I can do locally to see why it failed in the CI?
> > I setup the pre commit hook as instructed and made sure it passed for this commit before pushing so I'm confused on why the CI test for it failed.
> 
> Hmm. Try to reset your branch to main and recommit.
> 
> ```
> # git reset origin/main
> # git add foo
> # git commit
> ```
> 
> Hopefully it triggers the same thing.

It resulted in the same thing, after git adding the modifications I assume are form the pre commit reformatting the files it passes fine for the pre commit check.

superm1:
Great, push the result and I'll retrigger CI.

Joshua-Dickens:
> Great, push the result and I'll retrigger CI.

Done, if you need me to do anything else let me know.

plugins/wacom-usb/fu-wac-module-touch-id7.c
```diff
@@ -0,0 +1,435 @@
...
+				   FuFirmware *firmware,
+				   FuProgress *progress,
+				   FwupdInstallFlags flags,
+				   GError **error)
+=======
+				       FuFirmware *firmware,
+				       FuProgress *progress,
+				       FwupdInstallFlags flags,
+				       GError **error)
+>>>>>>> 2550fcb48 (wacom-usb: add touch id7 support to wacom-usb)
```
superm1, line 348: bad merge

superm1 (approved):

PR #5637, opened 2023-03-20
trivial: Move Fedora CI container from 37 to 39

superm1:
Needed for https://github.com/fwupd/fwupd-efi/pull/40

Will trigger a build before merging this to make sure that works rather than waiting 24 hours.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/Dockerfile-fedora.in
+++ b/contrib/ci/Dockerfile-fedora.in
@@ -1,4 +1,4 @@
-FROM fedora:37
+FROM fedora:39
 %%%OS%%%
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
--- a/subprojects/fwupd-efi.wrap
+++ b/subprojects/fwupd-efi.wrap
@@ -1,4 +1,4 @@
 [wrap-git]
 directory = fwupd-efi
 url = https://github.com/fwupd/fwupd-efi
-revision = 1.4
+revision = 189f7637f423cb2983aec9bdebd9e9a20765ab85


superm1:
Container built successfully, re-running CI jobs.

superm1:
Hmm...

> ../subprojects/fwupd-efi/efi/meson.build:52:6: ERROR: Problem encountered: Cannot find elf_x64_efi.lds

@hughsie thoughts?

superm1:
It seems like from the logic it might not be finding `efibind.h` in the right place?

> Well, my general thoughts about gnu-efi are probably unprintable. From a what are we going to do point of view, I guess my medium term goal is https://github.com/hughsie/fwupd-efirs/tree/wip/hughsie/poc

Can't see the link - I guess it's fwupd-efi ported to rust?

hughsie:
> It seems like from the logic it might not be finding efibind.h in the right place?

Yup, Fedora isn't using the old compat paths anymore.

> Can't see the link - I guess it's fwupd-efi ported to rust?

Ahh yes -- exactly that, it's 30% complete and has embarrassing Rust code. It's my plan to flesh it out and see it it's more maintainable than using gnu-efi.

superm1:
OK, so it works if called with `-Defi-libdir=/usr/lib`.  So the call to `gcc -print-multi-os-directory` is what breaks things.

superm1:
OK I put a w/a to get us by in https://github.com/fwupd/fwupd-efi/pull/42.

hughsie:
That looks like the new GLib introduced some leaks. I think ignore those.

PR #5638, opened 2023-03-22
added WIP JabraGnp plugin, meant for draft PR

gdpcastro:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

New plugin checklist:

- [ ] Implement FuFirmware->write() and include at least one fuzzer testcase in src/fuzzing/firmware for any custom FuFirmware subclass -- Not applicable yet
- [x] CI run of the plugin for at least one target -- pending creation of PR I believe
- [ ] Document targets CI isn't currently run and the reasons (i.e. minimum versions needed or distribution limitations etc). -- Not applicable


gdpcastro:
Draft PR for the newly created -- still WIP -- JabraGnp plugin. 
Noteworthy missing components: Creation of FuFirmware subclass.

hughsie:
@gdpcastro I've done a first round of review fixups, please ask if you don't understand what I've done or why.

plugins/jabra-gnp/README.md
```diff
@@ -0,0 +1,23 @@
+# Jabra Gnp
```
hughsie, line 1: `GNP` perhaps?

hughsie:
> How come the removal of the ->open and ->close ? 

It's already being done by the baseclass, in this case by FuUsbDevice. We just tell the baseclass what interface we care about with fu_usb_device_add_interface() and it magically works.

> And regarding the reading of the device's version, wouldn't it make sense to decouple the transfer functionality from the actual setting of the version?

Not sure I understand the question. The device version is stored on the FuDevice baseclass (the one that FuUsbDevice inherits from) and it's more suitable to be stored there.

> I would assume it's the endianness of the OS that might create problems, is that correctly understood?

Yes, correct. In the most case it doesn't matter, as you're not likely to plug in a USB device into a mainframe.

Richard

gdpcastro:
Hello again! This took a while, but we finally have a working plugin for our device. I still need some pointers on a couple of things, but I believe they should be minor, as the firmware update of the device, given a correct cabinet file, goes through (most of the time, there seems to be some instability, but that may be due to my development environment, we will be running tests on this to figure it out!). 
The main doubt I have is regarding the "state change" of the progress, as for the whole fw update process, fwupd is reporting the state is "Decompressing" at around 50%. I assumed this would be handled automatically, but I'm probably wrong. 
And of course any other comments and changes are more than welcome

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -0,0 +1,859 @@
...
+		return FALSE;
+	if (!fu_jabra_gnp_device_read_dfu_pid(self, error))
+		return FALSE;
+	dfu_pids = xb_silo_query(silo, "buildVector/targetUsbPids", 0, error);
+	if (dfu_pids == NULL)
+		return FALSE;
+
+	for (guint i = 0; i < dfu_pids->len; i++) {
+		n = g_ptr_array_index(dfu_pids, i);
+		tmp = xb_node_query_text(n, "usbPid", NULL);
```
hughsie, line 732: is there ever more than one usbPid in the targetUsbPids block?

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -0,0 +1,859 @@
...
+	if (!fu_jabra_gnp_device_write_dfu_from_squif(self, error))
+		return FALSE;
+	return TRUE;
+}
+
+static void
+fu_jabra_gnp_device_init(FuJabraGnpDevice *self)
+{
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_protocol(FU_DEVICE(self), "org.jabra.gnp");
```
hughsie, line 846: you mean `com.jabra.gnp` -- right?

hughsie:
I've fixed up a few of the nits, refactored the firmware loading, and rebased against main. Once you answer the two questions above I think it's basically good to be force pushed, retested, and merged.

plugins/jabra-gnp/fu-jabra-gnp-device.c
gdpcastro: I think you are right, `com.jabra.gnp` makes way more sense. I will need to update our cabinet files, but that shouldn't be an issue

plugins/jabra-gnp/fu-jabra-gnp-device.c
gdpcastro: As far as I am aware, there may be. I'm trying to get confirmation internally.

hughsie:
> fwupd is reporting the state is "Decompressing" at around 50%. I assumed this would be handled automatically,

Not automatically; we need to tell fwupd how long each part is going to take as a preportion. I've added some code to do that, although I'll need the actual values from someone with hardware. The `FU_PROGRESS_FLAG_GUESSED` means you get the values right in the console with a big red warning.

plugins/jabra-gnp/fu-jabra-gnp-device.c
hughsie: Okay, lets push this with one supported, and if we need to support both we can make it more complicated.

plugins/jabra-gnp/fu-jabra-gnp-device.c
hughsie: I mean the footer-us_en-ev275.gnv file -- is that signed by some Jabra private key or could I flash modified content onto the device?

hughsie: I mean app_p0-us_en-ev275.gnv -- but you get the idea.

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -0,0 +1,859 @@
...
+		return FALSE;
+	if (!fu_device_retry_full(FU_DEVICE(self),
+				  fu_jabra_gnp_device_receive_with_sequence_cb,
+				  FU_JABRA_GNP_MAX_RETRIES,
+				  FU_JABRA_GNP_RETRY_DELAY,
+				  (gpointer)&receive_data,
+				  error))
+		return FALSE;
+	version = fu_strsafe((const gchar *)receive_data.rxbuf + 8, FU_JABRA_GNP_BUF_SIZE - 8);
+	fu_device_set_version(FU_DEVICE(self), version);
```
hughsie, line 290: Is this always going to be a "triplet"? e.g. X.Y.Z? If you use PLAIN verfmt that's fine, but it gets nasty when you compare `1.12.0` and `1.2.0` as ASCII strings.

gdpcastro: I wasn't entirely sure if `1.12.0` would classify was a triplet! As we can have firmware in which the minor or micro can go above 9. If that's okay in the triplet format, it should cover our version formatting.

hughsie:
Some more questions please:

 * Does this device have dual images? i.e. FWUPD_DEVICE_FLAG_DUAL_IMAGE is for "The device firmware update architecture uses a redundancy mechanism such as A/B partitions for updates"
 * FWUPD_DEVICE_FLAG_SELF_RECOVERY is for "device will only accept intended payloads and will revert back to a valid firmware image if an invalid or incomplete payload was sent"

plugins/jabra-gnp/fu-jabra-gnp-device.c
hughsie: yup, triplet can be XXXXX.YYYYYY.ZZZZZZ i.e. pretty big -- and it'll massively simplify version comparisons. Thanks.

plugins/jabra-gnp/fu-jabra-gnp-device.c
gdpcastro: Our firmware files are signed, I'm just going to get some confirmation on how exactly. If knowing they are signed is enough, the answer should be yes, if you need to know a bit more I'll have to get back to you at a later point

gdpcastro:
> Some more questions please:
> 
> * Does this device have dual images? i.e. FWUPD_DEVICE_FLAG_DUAL_IMAGE is for "The device firmware update architecture uses a redundancy mechanism such as A/B partitions for updates"
> * FWUPD_DEVICE_FLAG_SELF_RECOVERY is for "device will only accept intended payloads and will revert back to a valid firmware image if an invalid or incomplete payload was sent"

Yes and yes

plugins/jabra-gnp/fu-jabra-gnp-device.c
hughsie: "signed" is enough for fwupd -- but the (confidential) detail is needed by the LVFS. You can sent that to me privately if you like.

gdpcastro: Perfect, I should get confirmation from the team about it and will reach out to you privately.
Just a note that Thursday the 18th and Friday the 19th (for the most part) are public holidays in Denmark, so I won't be working, and will get back to you early next week most likely.

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -0,0 +1,859 @@
...
+	    0xB9887C5B, 0x62F97A1A, 0xD41B7698, 0x0F6A70D9};
+
+	guint64 t = acc >> 24;
+	guint32 lookup = (guint32)(t & 0xFF);
+	acc = ((acc) << 8) ^ crcLookupTable[lookup] ^ (delta);
+	return acc & 0x00000000FFFFFFFF;
+}
+
+guint64
+fu_jabra_gnp_device_calculate_crc(const guint8 *data, gsize *data_len)
```
hughsie, line 152: @gdpcastro can you tell me any more about this? e.g. can we use fu_crc32_full() with a known polynomial perhaps?

hughsie:
@gdpcastro i've force pushed a new branch, can you try this on real hardware please and check I've not broken anything. I'm especially interested in the big red warning that say "steps were XXXXXX but should have been YYYYYYY" -- thanks.

plugins/jabra-gnp/jabra-gnp.quirk
```diff
@@ -0,0 +1,11 @@
+# Jabra Evolve2 75 [runtime]
+[USB\VID_0B0E&PID_24D9]
+Plugin = jabra_gnp
+
+# Jabra Evolve2 75 [runtime]
```
d4s, line 5: Just to be sure -- really the same device in the same state with different PIDs?

plugins/jabra-gnp/fu-jabra-gnp-image.c
```diff
@@ -0,0 +1,141 @@
...
+	const gchar *language;
+	const gchar *name;
+	const gchar *part_str = NULL;
+	guint64 crc_expected = 0;
+	guint64 partition = 0;
+	g_autoptr(FuFirmware) img_archive = NULL;
+	g_autoptr(GBytes) blob = NULL;
+
+	/* only match on US English */
+	language = xb_node_query_text(n, "language", NULL);
```
d4s, line 44: Is there any technical difference in update sequence or firmware format related to language?
Or it is related only to "partition", "crc", "name" node names below?

If I correctly got the idea -- in the ZIP archive might be several firmwares aimed to different languages (?) and only English variant could be used for flashing. Is that a correct assumption?

hughsie: Yes, there are multiple languages in the archive. The thing is; fwupd does know the users locale -- do we don't *have* to flash English. I'm also guessing we don't want to update a device from *1.2.3 French* to *1.2.4 English* so maybe we can query the device and add the current device language as an instance ID?

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -0,0 +1,707 @@
...
+						  error))
+				return FALSE;
+			if (rx_data.rxbuf[5] != match_buf[5] || rx_data.rxbuf[6] != match_buf[6]) {
+				g_set_error(error,
+					    FWUPD_ERROR,
+					    FWUPD_ERROR_INTERNAL,
+					    "internal error, buf did not match");
+				return FALSE;
+			}
+			if (0) {
```
d4s, line 388: Forgot?

plugins/jabra-gnp/fu-jabra-gnp-device.c
gdpcastro: Unfortunately I don't know much about this. It is provided to my team from a different team, I could try to dig a bit deeper if it would be possible to use the option you mentioned.

plugins/jabra-gnp/jabra-gnp.quirk
gdpcastro: Yes, we have more than 1 PID for the same device depending on some of it's features.

plugins/jabra-gnp/fu-jabra-gnp-firmware.c
```diff
@@ -0,0 +1,191 @@
...
+	for (guint i = 0; i < dfu_pids->len; i++) {
+		XbNode *c = g_ptr_array_index(dfu_pids, i);
+		const gchar *dfu_pid_str = xb_node_query_text(c, "usbPid", NULL);
+		guint64 val = 0;
+		if (!fu_strtoull(dfu_pid_str, &val, 0x0, 0xFFFF, error)) {
+			g_prefix_error(error, "cannot parse usbPid of %s: ", dfu_pid_str);
+			return FALSE;
+		}
+		self->dfu_pid = (guint16)val;
+		break;
```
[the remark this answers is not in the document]
gdpcastro: It is an OK assumption for the first device we are supporting, and most likely for most of the devices we will support.

plugins/jabra-gnp/fu-jabra-gnp-image.c
gdpcastro: We had a brief talk internally about this. I will follow-up and see what final solution we would like. If I have to query the device for it's language I will need to make a bit of a change to the code, including the parsing.

plugins/jabra-gnp/fu-jabra-gnp-device.c
gdpcastro: Not exactly. That is an extra check we can use to validate the firmware bytes being sent to the device, but it may be a bit too restrictive, so I left it in for review, but disabled the check until we start testing the code internally.

gdpcastro:
I tried to pull the new code, but I'm getting an error when running fwupdtool that I figured I'd just ask for some help with rather than bang my head against the wall all day. If I compile either main or this branch, I get the following error: 
image
I have a slight lead that it may be that I need to update GLib, so I will try that in the meantime

plugins/jabra-gnp/fu-jabra-gnp-image.c
hughsie: If it helps, we could do the English-only update as a first cut -- but I do think we want to query the current language and match on do avoid *changing* the users language.

hughsie:
> did not write:

What version Glib were you on before, and what are you on now please?

gdpcastro:
> What version Glib were you on before, and what are you on now please?
I believe I was on version 2.64.6, upgraded to 2.76.2 which created a compilation issue in fwupd, so I reverted back. If the issue really is the glib version I will try to upgrade again

hughsie:
> upgraded to 2.76.2 which created a compilation issue in fwupd

Can you upgrade, and then `meson configure --clearcache` then rebuild please?

gdpcastro:
> Can you upgrade, and then `meson configure --clearcache` then rebuild please?

Ok, upgrading properly did the trick. Thanks!

gdpcastro:
Hi, just wanted to inform that after fixing 2 small things, the code is working again on actual hardware.

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -0,0 +1,704 @@
...
+	if (!fu_jabra_gnp_device_write_dfu_from_squif(self, error))
+		return FALSE;
+	return TRUE;
+}
+
+static void
+fu_colorhug_device_set_progress(FuDevice *self, FuProgress *progress)
+{
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED); // FIXME, remove
```
hughsie, line 675: @gdpcastro do you get warnings in your console like https://github.com/fwupd/fwupd/wiki/Daemon-Warning:-FuProgress-steps -- if so can you submit the real values and remove GUESSED please.

gdpcastro: I do not get those warnings (even tried running with --verbose), which seems a bit odd, as I'm quite sure writing of chunks takes most of the time for most images -- also to note that since images have different sizes the steps would be different for each image I believe. 
I do however get a "Less than one minute remaining" message on the console when there is definitely more than one minute remaining, not sure how problematic that is.

gdpcastro:
I wanted to ask if, as part of CI, packages are built for this PR, and if so, where could I find the packages?

hughsie:
> I wanted to ask if, as part of CI, packages are built for this PR, and if so, where could I find the packages?

In some cases. If you look at the `ci/circleci: build-windows — Your tests passed on CircleCI! ` details you should find some *artifacts* although I don't know if you need to be a member to download them.

plugins/jabra-gnp/fu-jabra-gnp-device.c
hughsie: > I do however get a "Less than one minute remaining" message on the console when there is definitely more than one minute remaining

Right, this is probably because the values are wildly inaccurate. Can you try using `meson configure -Dsupported_build=disabled`? That's normally only set by distro packages. that have had QA.

gdpcastro:
> In some cases. If you look at the `ci/circleci: build-windows — Your tests passed on CircleCI! ` details you should find some _artifacts_ although I don't know if you need to be a member to download them.

I can see and download the Windows .msi, but I can't see anything under Ubuntu, which is the one I would be interested in. I also tried to build it locally using `./contrib/ci/debian.sh`, and while that does build, it doesn't seem to generate any .deb packages.

hughsie:
> I can't see anything under Ubuntu, which is the one I would be interested in

FWIW, I think it would be much better to use a local checkout at this stage.

superm1:
> > In some cases. If you look at the `ci/circleci: build-windows — Your tests passed on CircleCI! ` details you should find some _artifacts_ although I don't know if you need to be a member to download them.
> 
> I can see and download the Windows .msi, but I can't see anything under Ubuntu, which is the one I would be interested in. I also tried to build it locally using `./contrib/ci/debian.sh`, and while that does build, it doesn't seem to generate any .deb packages.

If you look at the logs, it's probably because you're missing something in your environment to generate the deb files and it should tell you what that is.

> FWIW, I think it would be much better to use a local checkout at this stage.

Also FWIW they're super easy to clean up.  Everything is plopped in `/usr/local` and you can just delete that.

plugins/jabra-gnp/fu-jabra-gnp-device.c
gdpcastro: I tried this, it didn't seem to help. I'm trying to fidget a bit with the progress, but it doesn't seem like I know how to fix it. I entered some somewhat more accurate values for the progress of each image (generalized, as images vary in size), but I still get the "less than one minute remaining" console output way too early.

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -102,7 +104,24 @@ fu_jabra_gnp_device_rx_cb(FuDevice *device, gpointer user_data, GError **error)
 					     error)) {
 		g_prefix_error(error, "failed to read from device: ");
 		return FALSE;
+	} else {
+		if (rx_data->rxbuf[5] == match_buf[5] && rx_data->rxbuf[6] == match_buf[6]) {
+			/* Battery report, ignore and rx again*/
```
hughsie, line 109: Lowercase single line comment please

hughsie: Also space before the *

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -385,17 +405,14 @@ fu_jabra_gnp_device_write_chunks(FuJabraGnpDevice *self,
...
-					g_set_error(error,
-						    FWUPD_ERROR,
-						    FWUPD_ERROR_INTERNAL,
-						    "internal error, buf did not match");
-					return FALSE;
-				}
-			}
+			if (fu_memread_uint16(rxbuf + 7, G_LITTLE_ENDIAN) != chunk_number) {
+				chunk_number--;
+				failed_chunk = TRUE;
```
hughsie, line 410: Do you need to check for underflow?

plugins/jabra-gnp/fu-jabra-gnp-device.c
```diff
@@ -576,12 +593,12 @@ fu_jabra_gnp_device_write_image(FuJabraGnpDevice *self,
 	/* progress */
 	fu_progress_set_id(progress, G_STRLOC);
 	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED); // FIXME, remove
-	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 10, "write-partition");
```
hughsie: I think we can loose the fixme too. Isn't there also one other guessed progress steps to fill in too?

gdpcastro: The other guessed was quite accurate, so I have just removed the guessed flag now. However I'm still getting a "Less than one minute remaining..." console output when executing with fwupdtool, when there is more than one minute remaining. It appears at the start of the writing phase, and I'm not entirely sure how to fix it, or how important it is to fix it.

hughsie: I don't think it's terribly important FWIW. I'd be curious to know if this also happens with fwupd/fwupdmgr too.

hughsie:
FWIW, I'd be happy to rebase and merge this if you'd like me to. We can always do more fixups later in a future PR. I'll run Coverity and PVS this morning.

gdpcastro:
I'm happy we're at the stage of almost merging! :) 
Let me get the final confirmation from the team, there was 1 thing we were discussing adding to this PR

hughsie:
PVS found: `fu-jabra-gnp-firmware.c:89	Medium	V612	An unconditional 'break' within a loop.`

-- I think it's fine -- but we should really consider using xb_silo_query_first() if we're only going to support **one** usbPid

gdpcastro:
Ok, I got confirmation from the team to add something to this PR! It's only on the quirk file, so I should be done very soon, and the review of it is probably minor. 

> if we're only going to support one usbPid
I think we're okay to do so, actually. I'll make the change

gdpcastro:
From our side we're ready for the merge!

hughsie (approved):

PR #5639, opened 2023-03-22
Generate packed struct getters and setters like protobuf

hughsie:
At the moment we use two styles, both broken in subtly ways:

 * `memcpy()` a packed struct, then `GUINTXX_TO_LE` and `GUINTXX_FROM_LE`
 * `fu_memread_uintXX` and `fu_memwrite_uintXX` from known buffer offsets

In some cases, both were being used in the same plugin. Using the same method to parse and write the structure makes it much easier to read and write plugins whilst still being memory safe.

The actual struct definition is taken from the source file itself, and the code generated with the correct types and checks.

We didn't use protobuf for this for a few reasons:

 * No support for default or constant values
 * Clucky to use from GLib-style parsers and writers
 * No support for uint24 or GUID types
 * Needing to wrap the protobuf_c generated class in a GByteArray wrapper
 * The protobuf_c library is not a hard dep and not available in RHEL

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/contrib/ci/oss-fuzz.py
+++ b/contrib/ci/oss-fuzz.py
@@ -156,6 +156,27 @@ class Builder:
             sys.exit(1)
         return os.path.join(self.builddir, "{}".format(dst))
 
+    def generate_struct(self, src: str) -> str:
+
+        fn_root = os.path.basename(src).replace(".struct", "")
+        fulldst_c = os.path.join(self.builddir, f"{fn_root}-struct.c")
+        fulldst_h = os.path.join(self.builddir, f"{fn_root}-struct.h")
+        try:
+            subprocess.run(
+                [
+                    "fwupd/libfwupdplugin/generate-struct.py",
+                    src,
+                    fulldst_c,
+                    fulldst_h,
+                ],
+                cwd=self.srcdir,
+                check=True,
+            )
+        except subprocess.CalledProcessError as e:
+            print(e)
+            sys.exit(1)
+        return fulldst_c
+
     def link(self, objs: List[str], dst: str) -> None:
         """link multiple objects into a binary"""
         argv = [self.cxx] + self.cxxflags
@@ -345,7 +366,10 @@ def _build(bld: Builder) -> None:
     for path in ["fwupd/libfwupd", "fwupd/libfwupdplugin"]:
         bld.add_src_includedir(path)
         for src in bld.grep_meson(path):
-            built_objs.append(bld.compile(src))
+            if src.endswith(".c"):
+                built_objs.append(bld.compile(src))
+            elif src.endswith(".struct"):
+                built_objs.append(bld.compile(bld.generate_struct(src)))
 
     # dummy binary entrypoint
     if "LIB_FUZZING_ENGINE" in os.environ:
@@ -415,7 +439,10 @@ def _build(bld: Builder) -> None:
     ]:
         fuzz_objs = []
         for obj in bld.grep_meson("fwupd/plugins/{}".format(fzr.srcdir)):
-            fuzz_objs.append(bld.compile(obj))
+            if obj.endswith(".c"):
+                fuzz_objs.append(bld.compile(obj))
+            elif obj.endswith(".struct"):
+                fuzz_objs.append(bld.compile(bld.generate_struct(obj)))
         src = bld.substitute(
             "fwupd/libfwupdplugin/fu-fuzzer-firmware.c.in",
             {
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -24,6 +24,7 @@
 #include "fu-device-progress.h"
 #include "fu-plugin-private.h"
 #include "fu-security-attrs-private.h"
+#include "fu-self-test-struct.h"
 #include "fu-smbios-private.h"
 
 static GMainLoop *_test_loop = NULL;
@@ -3767,6 +3768,56 @@ fu_progress_child_finished(void)
 	fu_progress_step_done(progress);
 }
 
+static void
+fu_plugin_struct_func(void)
+{
+	gboolean ret;
+	g_autoptr(GByteArray) st = fu_struct_self_test_new();
+	g_autoptr(GByteArray) st2 = NULL;
+	g_autoptr(GByteArray) st3 = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autofree gchar *str1 = NULL;
+	g_autofree gchar *oem_table_id = NULL;
+
+	/* size */
+	g_assert_cmpint(st->len, ==, 51);
+
+	/* getters and setters */
+	fu_struct_self_test_set_revision(st, 0xFF);
+	fu_struct_self_test_set_length(st, 0xDEAD);
+	ret = fu_struct_self_test_set_oem_table_id(st, "X", &error);
+	g_assert_no_error(error);
+	g_assert_true(ret);
+	g_assert_cmpint(fu_struct_self_test_get_revision(st), ==, 0xFF);
+	g_assert_cmpint(fu_struct_self_test_get_length(st), ==, 0xDEAD);
+
+	/* pack */
+	str1 = fu_byte_array_to_string(st);
+	g_assert_cmpstr(str1,
+			==,
+			"12345678adde0000ff000000000000000000000000000000004142434445465800000000"
+			"000000000000000000000000000000");
+
+	/* parse */
+	st2 = fu_struct_self_test_parse(st->data, st->len, 0x0, &error);
+	g_assert_no_error(error);
+	g_assert_nonnull(st2);
+	g_assert_cmpint(fu_struct_self_test_get_revision(st2), ==, 0xFF);
+	g_assert_cmpint(fu_struct_self_test_get_length(st2), ==, 0xDEAD);
+	oem_table_id = fu_struct_self_test_get_oem_table_id(st2);
+	g_assert_cmpstr(oem_table_id, ==, "X");
+
+	/* parse failing signature */
+	st->data[0] = 0xFF;
+	st3 = fu_struct_self_test_parse(st->data, st->len, 0x0, &error);
+	g_assert_error(error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA);
+	g_assert_null(st3);
+	g_clear_error(&error);
+	ret = fu_struct_self_test_validate(st->data, st->len, 0x0, &error);
+	g_assert_error(error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA);
+	g_assert_false(ret);
+}
+
 int
 main(int argc, char **argv)
 {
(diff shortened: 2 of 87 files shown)


libfwupdplugin/meson.build
```diff
@@ -8,6 +8,31 @@ fwupdplugin_version_h = configure_file(
...
 )
 
+fwupdplugin_generated_srcs = []
+foreach generated: [
+  'fu-acpi-table',  # fuzzing
+  'fu-self-test',   # fuzzing
+]
+  fwupdplugin_generated_srcs += custom_target('@0@generated'.format(generated),
+    input: '@0@.c'.format(generated),
+    output: '@0@-generated.c'.format(generated),
```
superm1, line 18: rather than parsing comments, how about we specify in a separate XML, JSON, or other directly parsable format?

contrib/generate-struct.py
```diff
@@ -0,0 +1,385 @@
+#!/usr/bin/python3
+# pylint: disable=invalid-name,missing-docstring
+#
+# Copyright (C) 2022 Richard Hughes <richard@hughsie.com>
```
superm1, line 4: 2023

libfwupdplugin/meson.build
hughsie: We could do; it was just handy having the definition in the c file where it was being used. Something like `fu-acpi-table.struct` perhaps?

contrib/generate-struct.py
```diff
@@ -0,0 +1,385 @@
...
+            return "G_BIG_ENDIAN"
+        return "G_BYTE_ORDER"
+
+    @property
+    def type_glib_mut(self) -> str:
+        tmp_type = ""
+        if self.type == Type.U8:
+            tmp_type = "guint8"
+        elif self.type == Type.STRING:
+            tmp_type = "gchar"
```
superm1, line 73: just as a nit I'd say swap the order of these two so all the integers are together

hughsie: Fixed.

contrib/generate-struct.py
```diff
@@ -0,0 +1,385 @@
...
+#
+# Copyright (C) 2022 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+
+import os
+import sys
+import argparse
+
+from enum import StrEnum
```
superm1, line 12: CI doesn't seem to like this.  Why not just use a dict?
```
    from enum import StrEnum
ImportError: cannot import name 'StrEnum' from 'enum' (/usr/lib/python3.8/enum.py)
```

superm1:
(And BTW I like this approach better)

contrib/generate-struct.py
hughsie: Ahh, StrEnum was only added in 3.11 -- will do something simpler.

contrib/ci/oss-fuzz.py
```diff
@@ -156,6 +156,22 @@ def compile(self, src: str) -> str:
...
+        fulldst_h = os.path.join(self.builddir, f"{os.path.basename(src)}-struct.h")
+        try:
+            subprocess.run(
+                ["fwupd/contrib/generate-struct.py", fullsrc, fulldst_c, fulldst_h],
+                cwd=self.srcdir,
+                check=True,
+            )
+        except subprocess.CalledProcessError as e:
+            print(e)
+            sys.exit(1)
```
superm1, line 172: you can combine these.

```
$ python3
Python 3.8.10 (default, Jun  2 2021, 10:49:15)
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.exit("foo the bar")
foo the bar
$ echo $?
1
```

contrib/generate-struct.py
superm1, line 1: As this is core to the build process, I think this file shouldn't live in `contrib`.  I think of stuff in contrib as optional packaging things and CI things that could be entirely stripped out for someone that chooses to compile by hand and doens't use CI.

libfwupdplugin/meson.build
```diff
@@ -8,6 +8,33 @@ fwupdplugin_version_h = configure_file(
   configuration: conf
 )
 
+fwupdplugin_struct = []
+fwupdplugin_struct_srcs = []
+foreach generated: [
+  'fu-acpi-table',  # fuzzing
+  'fu-self-test',   # fuzzing
+]
```
superm1, line 16: thinking through how plugins will probably want to be providing these as well, probably need to make the array additive.

contrib/generate-struct.py
```diff
@@ -0,0 +1,381 @@
...
+        # useful constants
+        name_snake = _camel_to_snake(name)
+
+        # header
+        str_h = "/* auto-generated, do not modify */\n"
+        str_h += "#include <glib.h>\n"
+        str_h += f"GByteArray* {name_snake}_new(void);\n"
+        str_h += f"GByteArray* {name_snake}_unpack(const guint8 *buf, gsize bufsz, gsize offset, GError **error);\n"
+        for item in items:
+            if item.element_id.startswith("_"):
```
superm1, line 169: probably want to log a loud warning on this

contrib/generate-struct.py
```diff
@@ -0,0 +1,381 @@
...
+                continue
+            if item.constant:
+                continue
```
superm1, line 172: really all of these should log something about how it's not conformant, otherwise we're going to see weird bugs

contrib/generate-struct.py
```diff
@@ -0,0 +1,381 @@
...
+        str_c += f"                        G_IO_ERROR,\n"
+        str_c += f"                        G_IO_ERROR_INVALID_DATA,\n"
+        str_c += f'                        "cannot parse buffer of size 0x%x at offset 0x%x for struct {name} of size 0x%x",\n'
+        str_c += f"                        (guint) bufsz,\n"
+        str_c += f"                        (guint) offset,\n"
+        str_c += f"                        (guint) {size});\n"
+        str_c += f"            return NULL;\n"
+        str_c += f"    }}\n"
+        str_c += f"    g_byte_array_append(st, buf + offset, {size});\n"
+        str_c += f"    /* FIXME: check constants */\n"
```
superm1, line 298: presumably don't want to be commenting in the C code to check things, eh?

contrib/ci/oss-fuzz.py
hughsie: TIL, thanks.

contrib/generate-struct.py
hughsie: agreed, but I'm struggling to work out where it *does* belong. `data` perhaps?

libfwupdplugin/meson.build
hughsie: I've switched to using a meson generator and it's much much nicer.

contrib/generate-struct.py
hughsie: this was deliberate, i.e. the leading underscore is "don't bother creating getters and setters as this isn't used" -- i.e. like `_reserved0` where we're never going to set to get the value.

contrib/generate-struct.py
hughsie: aha, so the logic here is that the user never needs getters or setters for something that's a constant. It's either set by `_new()` or `_parse()` fails to run.

hughsie: I've added some docs

contrib/generate-struct.py
superm1: How about right in `libfwupdplugin`?

meson.build
```diff
@@ -565,6 +565,13 @@ else
   run_sanitize_unsafe_tests = true
 endif
 
+# take foo.struct and generate foo-struct.c and foo-struct.h files like protobuf_c
+structgenpy = find_program(join_paths(meson.project_source_root(), 'contrib', 'generate-struct.py'))
+structgen = generator(structgenpy,
```
superm1, line 570: for CI purpose, what do you think about having a few bad structs that should fail to process that we test for?

libfwupdplugin/meson.build
```diff
@@ -8,32 +8,9 @@ fwupdplugin_version_h = configure_file(
...
-  fwupdplugin_struct_srcs += fwupdplugin_struct_src
-  fwupdplugin_struct += static_library('@0@struct2'.format(generated),
-    fwupdplugin_struct_src,
-    dependencies: [
-      gio,
-      libxmlb,
-    ],
-  )
-endforeach
+fwupdplugin_structs = structgen.process(
```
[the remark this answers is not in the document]
hughsie: I think it _should_ be as easy as:

     plugin_builtin_redfish = static_library('fu_plugin_redfish',
    +     structgen.process('fu-redfish-smbios.struct'),
       sources: [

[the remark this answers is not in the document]
superm1: Great 👍

libfwupdplugin/fu-self-test.c
```diff
@@ -3489,6 +3490,14 @@ fu_firmware_builder_round_trip_func(void)
 		g_assert_true(ret);
 		csum1 = fu_firmware_get_checksum(firmware1, G_CHECKSUM_SHA1, &error);
 		g_assert_no_error(error);
+		if (g_strcmp0(csum1, map[i].checksum) != 0) {
+			const gchar *filename_out = "/tmp/foo.bin";
+			g_autoptr(GFile) file = g_file_new_for_path(filename_out);
+			ret = fu_firmware_write_file(firmware1, file, &error);
+			g_assert_no_error(error);
+			g_assert_true(ret);
+			g_debug("did not build %s into %s", filename, filename_out);
```
superm1, line 3499: eh?

contrib/generate-struct.py
```diff
@@ -0,0 +1,392 @@
...
+                str_h += f"void {name_snake}_set_{item.element_id}(GByteArray *st, {item.type_glib_const} value);\n"
+
+        # body
+        str_c = "/* auto-generated, do not modify */\n"
+        str_c += f'#include "{self.basename}"\n'
+        str_c += (
+            '#include "fu-byte-array.h"\n'  # FIXME: need override for libfwupdplugin.h
+        )
+        str_c += '#include "fu-mem.h"\n'  # FIXME: need override for libfwupdplugin.h
+        str_c += '#include "fu-string.h"\n'  # FIXME: need override for libfwupdplugin.h
```
superm1, line 196: maybe the checker for the headers should ignore files with `struct.c` in the name?

hughsie: nah, there's a cute meson trick you can do to modify the local generator -- i'll do that in libfwupdplugin only -- atm I'm converting the tree like the FuStruct branch and finding out any missing parts.

contrib/generate-struct.py
hughsie: i guess that makes sense. Fixed.

meson.build
hughsie: Maybe not for CI -- but i certainly found a few converting tons of code ;) The benefit with this generator approach is that any failures to compile are compile **errors**, nor runtime warnings.

libfwupdplugin/fu-self-test.c
```diff
@@ -3489,6 +3490,14 @@ fu_firmware_builder_round_trip_func(void)
...
+		}
```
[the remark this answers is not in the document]
hughsie: Done.

superm1 (approved):

hughsie:
1.8.14 tho, right?

superm1:
Yeah

superm1:
FYI  this broke fuzzing CI:
> FileNotFoundError: [Errno 2] No such file or directory: 'fwupd/libfwupdplugin/generate-struct.py'

hughsie:
yup, but i'm trying to work out why. Maybe we need an explicit "python" in the argv[0]?

hughsie:
Fixed now -- but the fuzzer found a valid issue, so I'll fix that too.

PR #5644, opened 2023-03-23
Wip/hughsie/cape timeouts

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


d4s:
@hughsie sorry, have no access to the device. I found that suspicious place after the log analysis.

d4s (commented):
The patch looks good for me but I'd prefer f somebody with the HW on hands check that.
Preferable Epos 260 -- catched on downgrade/upgrade.

plugins/synaptics-cape/fu-synaptics-cape-device.c
```diff
@@ -14,10 +14,11 @@
 #include "fu-synaptics-cape-firmware.h"
 
 /* defines timings */
-#define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_WRITE_TIMEOUT	20000 /* us */
-#define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_READ_TIMEOUT	30000 /* us */
+#define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_WRITE_TIMEOUT	20    /* ms */
+#define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_READ_TIMEOUT	30    /* ms */
 #define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_RETRY_INTERVAL 10    /* ms */
 #define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_RETRY_TIMEOUT	300   /* ms */
+#define FU_SYNAPTICS_CAPE_DEVICE_USB_CMD_INTR_TIMEOUT	300000 /* ms */
```
CNXT-Simon, line 21: Usually, device can respond the fw update request within 300 ms. I think it could be the FW was too busy to respond the request. According to FW API spec, the maximum of timeout here should be 2 seconds.

CNXT-Simon:
@hughsie could you point me where can I get the pre-compiled fwupd tool to repro this issue?

hughsie:
Okay, I've changed this to a maximum of 5000ms which should be easily big enough for 300ms x a few retries. I've build the package here, does that help: https://copr.fedorainfracloud.org/coprs/rhughes/fwupd/ -- please update fwupd to 1.8.13-0.990.20230324git.fc38

@CNXT-Simon would it be possible to be sent a device [with updates on the LVFS] for validation please? We can use it to prevent this kind of regression.

d4s:
I did the test with different EPOS device: `ID 1395:0298 Sennheiser Communications EPOS ADAPT 1x5`

1. On `main` I reproduced the issue:
```
╔══════════════════════════════════════════════════════════════════════════════╗
║ Upgrade EPOS ADAPT 1x5 from 8.41.34.0 to 8.41.38.0?                          ║
╠══════════════════════════════════════════════════════════════════════════════╣
Perform operation? [Y|n]: y
Restarting device…       [**************************             ] Less than one minute remaining…
reset command is not supported: failed to get IN_REPORT: failed to get report over interrupt ep: transfer timed out
```
2. With the patch it works well and I got `Successfully installed firmware` without errors.

d4s (approved):
LGTM

hughsie:
Thanks @d4s -- much appreciated.

PR #5646, opened 2023-03-26
Add some more docs on Best Known Configuration

hughsie:
Fixes some of https://github.com/fwupd/fwupd/discussions/5641

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/docs/best-known-configuration.md
+++ b/docs/best-known-configuration.md
@@ -1,0 +1,102 @@
+# Best Known Configuration
+
+## Introduction
+
+Component `<tag>`s are used both by OEMs and by customers to identify a *known-working* (or
+commercially supported) set of firmware on the machine.
+
+This allows two things:
+
+* Factory recovery where a system in the field has been upgraded
+* Ensuring a consistent set of vendor-tested firmware for a specific workload
+
+The tags are either assigned in the firmware cabinet archive (the `.metainfo.xml` file) or added
+post-upload on the LVFS and are then included in the public AppStream metadata.
+A single firmware can be marked with multiple tags, and tags can be duplicated for different
+firmwares.
+This would allow an OEM to say *this set of firmware has been tested as a set for workload A,
+and this other set of firmware has been tested for workload B* which is fairly typical for
+enterprise deployments.
+
+## LVFS
+
+The LVFS added support for “vendor-defined” component `<tag>`s in 2021, which are also supported in
+fwupd since version 1.7.3.
+These tags are typically used by OEMs to identify a manifest of firmware for a specific machine SKU.
+This is opt-in for each vendor, and so if you are a vendor reading this and want to use this feature,
+let us know by [opening an issue](https://gitlab.com/fwupd/lvfs-website/-/issues).
+
+## Client
+
+When provisioning the client machine, we can set the BKC by setting `HostBkc=vendor-2021q1` in
+`/etc/fwupd/daemon.conf`.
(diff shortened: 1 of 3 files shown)


docs/best-known-configuration.md
```diff
@@ -0,0 +1,96 @@
+# Best Known Configuration
+
+Component <tag>s are used both by OEMs and by customers to identify a *known-working* (or
```
superm1, line 3:
```suggestion
`<tag>`s is used in components used both by OEMs and by customers to identify a *known-working* (or
```

superm1 (commented):
The first section needs to be grouped by something to indicate it's for the concept, the second for LVFS and third for local (I only see heading for local)

docs/best-known-configuration.md
```diff
@@ -0,0 +1,96 @@
...
+
+..and can be synced on the command line:
+
+    $ fwupdmgr sync-bkc
+    ╔══════════════════════════════════════════════════════════════════════════════╗
+    ║ Downgrade System Firmware from 225.52.1521 to 225.53.1649?                   ║
+    ╠══════════════════════════════════════════════════════════════════════════════╣
+    ║ This release sets the number of HBAs supported by the system to 1024.        ║
+    ╚══════════════════════════════════════════════════════════════════════════════╝
+    Perform operation? [Y|n]:
```
superm1, line 96: this should be in a conceptual section in section 1

hughsie:
> The first section needs to be grouped by something

I've rearranged things a bit, how does that look?

docs/best-known-configuration.md
```diff
@@ -0,0 +1,102 @@
...
+firmwares.
+This would allow an OEM to say *this set of firmware has been tested as a set for workload A,
+and this other set of firmware has been tested for workload B* which is fairly typical for
+enterprise deployments.
+
+## LVFS
+
+The LVFS added support for “vendor-defined” component `<tag>`s in 2021, which are also supported in
+fwupd since version 1.7.3.
+These tags are typically used by server vendors to identify a manifest of firmware for a specific
```
superm1, line 25: still don't think this should have server in it

docs/best-known-configuration.md
```diff
@@ -0,0 +1,102 @@
...
+## Client
+
+When provisioning the client machine, we can set the BKC by setting `HostBkc=vendor-2021q1` in
+`/etc/fwupd/daemon.conf`.
+Then any invocation of `fwupdmgr sync-bkc` will install or downgrade firmware on all compatible
+devices (UEFI, RAID, network adapter, & SAS HBA etc.) to make the system match a compatible set.
+
+Updating or downgrading firmware away from the *Best Known Configuration* is allowed, but the UI
+shows a warning.
+Using `fwupdmgr sync-bkc` will undo any manual changes and bring the machine back to the BKC.
```
superm1, line 38: I feel like this should be the last section

docs/best-known-configuration.md
hughsie: Oopa, removed.

docs/best-known-configuration.md
hughsie: Maybe Local metadata should be H3? I think it's more common that you set a HostBkc in daemon.conf than define local metadata -- hence why i put it in this order.

superm1: Got it, makes sense then.

superm1 (approved):

PR #5652, opened 2023-03-27
ch341a: Add important note about the 5V MOSI and MISO pins

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation


plugins/ch341a/README.md
```diff
@@ -18,6 +18,16 @@ The assumed map between UIO command bits, pins on CH341A chip and pins on SPI ch
     6    D6/21   unused  DIN2
     7    D7/22   SO/2    DIN
 
+IMPORTANT NOTE:
+
+You must perform the 3.3V signal output modification if you are using the CH341A with 3.3V SPI
+chips. The CH341A has a design flaw that outputs 5V on the MISO and MOSI pins even when VCC is 3V
+which will almost certainly be out-of-specification for the device you are trying to program.
+
+![CH341A Signal Output Modification](ch341a-vmod.png)
```
superm1, line 27: where did this image come from?  I don't think we should be including random images in our source tree without attribution.  It might be better to just link to it.

hughsie: That's my device on my desk :)

hughsie: If you've got any ideas on how to attribute the image I'm all ears.

superm1: Heh; OK then that's probably fine as long as you are OK having it the same as the project license.

PR #5653, opened 2023-03-27
ch347: Add support for the CH347 SPI programmer

hughsie:
Move the generic SPI functionality from the CH341a plugin to libfwupdplugin.

The ->send_command() vfunc is inspired by flashrom, and it seems to be most suitable for USB programmers with odd quirks.

NOTE: I've not switched the CH341a plugin to use the generic command as there are two quirks (JEDEC 4 byte read, junk in initial read buffer) which need further reverse engineering to unpuzzle.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-cfi-device.c
+++ b/libfwupdplugin/fu-cfi-device.c
@@ -8,7 +8,11 @@
 
 #include "config.h"
 
+#include "fu-byte-array.h"
+#include "fu-bytes.h"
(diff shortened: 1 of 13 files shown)


hughsie:
@superm1 any concerns? no detailed review needed IMHO, it's a new SPI flashing device.

superm1 (approved):

PR #5657, opened 2023-03-29
Add a note about when each plugin was added

hughsie:
This is something I'm asked all the time by end-users and vendors.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation


superm1 (approved):
Didn't verify values, but I guess you did `git describe --contains`, right?

PR #5659, opened 2023-03-29
Adjust thinklmi handling

superm1:
The plan for this is that we offer a compile option to turn on compatibility with older kernels.
Proposal:
* In ~March 2024 we'll change default for this option to false.
* In ~September 2024 we'll drop all the compatibility code

This PR is currently marked draft because it's pending one more kernel fix to adjust the `current_value` sysfs file to behave properly.  It will be adjusted and marked ready after that is landed in mainline.

@mrhpearson  FYI.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -3146,11 +3146,10 @@ fu_bios_settings_load_func(void)
 	test_dir = g_test_build_filename(G_TEST_DIST, "tests", "bios-attrs", "lenovo-p620", NULL);
 	(void)g_setenv("FWUPD_SYSFSFWATTRIBDIR", test_dir, TRUE);
 
-	g_test_expect_message("FuBiosSettings", G_LOG_LEVEL_WARNING, "*BUG*");
 	ret = fu_context_reload_bios_settings(ctx, &error);
+#ifdef FU_THINKLMI_COMPAT
 	g_assert_no_error(error);
 	g_assert_true(ret);
-	g_test_assert_expected_messages();
 
 	p620_settings = fu_context_get_bios_settings(ctx);
 	p620_items = fu_bios_settings_get_all(p620_settings);
@@ -3201,6 +3200,13 @@ fu_bios_settings_load_func(void)
 		g_debug("%s", tmp);
 		g_assert_null(g_strrstr(tmp, "[Status"));
 	}
+#else
+	g_assert_error(error, G_FILE_ERROR, G_FILE_ERROR_NOENT);
+	g_assert_false(ret);
+	g_clear_error(&error);
+#endif
+	g_free(test_dir);
+
 
 	g_free(test_dir);
 
@@ -3209,6 +3215,7 @@ fu_bios_settings_load_func(void)
 	    g_test_build_filename(G_TEST_DIST, "tests", "bios-attrs", "lenovo-p14s-gen1", NULL);
 	(void)g_setenv("FWUPD_SYSFSFWATTRIBDIR", test_dir, TRUE);
 	ret = fu_context_reload_bios_settings(ctx, &error);
+#ifdef FU_THINKLMI_COMPAT
 	g_assert_no_error(error);
 	g_assert_true(ret);
 
@@ -3241,7 +3248,11 @@ fu_bios_settings_load_func(void)
 	g_assert_nonnull(setting);
 	ret = fwupd_bios_setting_get_read_only(setting);
 	g_assert_true(ret);
-
+#else
+	g_assert_error(error, G_FILE_ERROR, G_FILE_ERROR_NOENT);
+	g_assert_false(ret);
+	g_clear_error(&error);
+#endif
 	g_free(test_dir);
 
 	/* load BIOS settings from a Dell XPS 9310 */
--- a/libfwupdplugin/fu-self-test.c
+++ b/libfwupdplugin/fu-self-test.c
@@ -3136,10 +3136,12 @@ fu_bios_settings_load_func(void)
 	g_autoptr(FuContext) ctx = fu_context_new();
 	g_autoptr(GError) error = NULL;
 	g_autoptr(FuBiosSettings) p620_settings = NULL;
+	g_autoptr(FuBiosSettings) p620_6_3_settings = NULL;
 	g_autoptr(FuBiosSettings) p14s_settings = NULL;
 	g_autoptr(FuBiosSettings) xp29310_settings = NULL;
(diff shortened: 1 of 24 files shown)


libfwupdplugin/fu-self-test.c
```diff
@@ -3201,6 +3202,74 @@ fu_bios_settings_load_func(void)
...
+	setting = fu_context_get_bios_setting(ctx, "com.thinklmi.AMDMemoryGuard");
+	g_assert_nonnull(setting);
+	tmp = fwupd_bios_setting_get_name(setting);
+	g_assert_cmpstr(tmp, ==, "AMDMemoryGuard");
+	tmp = fwupd_bios_setting_get_description(setting);
+	g_assert_cmpstr(tmp, ==, "AMDMemoryGuard");
+	// Waiting for kernel patch
+	//	tmp = fwupd_bios_setting_get_current_value(setting);
+	//	g_assert_cmpstr(tmp, ==, "Disable");
+	values = fwupd_bios_setting_get_possible_values(setting);
```
superm1, line 3239: ( TODO )

libfwupdplugin/tests/bios-attrs/lenovo-p620-6.3/thinklmi/attributes/AutomaticBootSequence/display_name
```diff
@@ -0,0 +1 @@
+AutomaticBootSequence
```
hughsie, line 1: can you symlink any of the existing tree contents?

plugins/lenovo-thinklmi/fu-self-test.c
```diff
@@ -41,44 +41,42 @@ fu_test_fatal_handler_cb(const gchar *log_domain,
...
+	g_assert_no_error(*error);
 	g_assert_true(ret);
-	ret = fu_context_load_hwinfo(ctx, progress, FU_CONTEXT_HWID_FLAG_LOAD_CONFIG, &error);
-	g_assert_no_error(error);
+	ret = fu_context_load_hwinfo(ctx, progress, FU_CONTEXT_HWID_FLAG_LOAD_CONFIG, error);
+	g_assert_no_error(*error);
 	g_assert_true(ret);
-	ret = fu_context_reload_bios_settings(ctx, &error);
+	ret = fu_context_reload_bios_settings(ctx, error);
+#ifdef THINKLMI_COMPAT
```
hughsie, line 61: I think this needs a FU_ prefix

libfwupdplugin/tests/bios-attrs/lenovo-p620-6.3/thinklmi/attributes/AutomaticBootSequence/display_name
superm1: Certainly can, but that's a lot of manual work and intend to drop the old folder eventually.

hughsie: Ohh, I thought you could just point `fdupes` at it...

libfwupdplugin/fu-self-test.c
```diff
@@ -3201,6 +3202,74 @@ fu_bios_settings_load_func(void)
...
+	}
+
+	/* try to read an BIOS setting known to have ][Status] to make sure we worked
+	 * around the thinklmi bug sufficiently
+	 */
+	setting = fu_context_get_bios_setting(ctx, "com.thinklmi.StartupSequence");
+	g_assert_nonnull(setting);
+	// Waiting for kernel patch
+	//	tmp = fwupd_bios_setting_get_current_value(setting);
+	//	g_assert_cmpstr(tmp, ==, "Primary");
```
superm1, line 3255: (TODO)

libfwupdplugin/tests/bios-attrs/lenovo-p620-6.3/thinklmi/attributes/AutomaticBootSequence/display_name
superm1: 6.6M source to 4.9M source by making this change!

superm1:
The patches to fix the `current_value` handling on P620 were submitted.
https://lore.kernel.org/platform-driver-x86/20230403013120.2105-1-mpearson-lenovo@squebb.ca/T/#t

I confirmed these work properly and so I've adjusted the unit tests accordingly and this PR is ready to go.

libfwupdplugin/fu-self-test.c
```diff
@@ -3201,6 +3202,72 @@ fu_bios_settings_load_func(void)
...
+			g_assert_cmpstr(possible, ==, "Automatic");
+	}
+
+	/* check no BIOS settings have [Status in them */
+	for (guint i = 0; i < p620_6_3_items->len; i++) {
+		setting = g_ptr_array_index(p620_6_3_items, i);
+		tmp = fwupd_bios_setting_get_current_value(setting);
+		g_debug("%s", tmp);
+		// Waiting for kernel patch
+		//		g_assert_null(g_strrstr(tmp, "[Status"));
```
hughsie, line 3269: ??

superm1: /sigh 

There's one more place the unit tests are still failing it seems and I missed it, thanks for catching it.
These 3 attributes have a `ShowOnly` appended.

```
thinklmi/attributes/AlarmTime/current_value:[00:00:00][Status:ShowOnly]
thinklmi/attributes/AlarmDate/current_value:[01/01/2019][Status:ShowOnly]
thinklmi/attributes/UserDefinedAlarmTime/current_value:[00:00:00][Status:ShowOnly]
```

@mrhpearson the kernel should be stripping that too.

superm1: You're right I guess since they're read only it probably doesn't really matter too much.  Let me adjust the test suite accordingly.

hughsie:
Is there any way we can put these thousands of small files in an archive and load from that? Or maybe decompress it to /tmp in the self-tests? Just looking at the PR locks up my browser tab and each file is going to be ~4k on disk even if just a few bytes.

superm1:
> How about if we move it to the fwupd test firmware repo?

Or... we could reduce all of them down to a handful (2-5) for each of the types of tests and keep them here.  There really are only a few things that are different about different types of attributes that matter.  We can just catch those and ignore the other 100+ attributes for the purpose of testing.

hughsie:
Yup, either is fie -- thanks.

hughsie (approved):

Issue #5656, opened 2023-03-28
fwupd 1.8.13 core dumps


bulletmark:
fwupd updated on Arch today: `2023-03-29 08:01:38 fwupd 1.8.12-1 -> 1.8.13-1` so I did a refresh and attempted update of my devices:
```
$ sudo fwupdmgr update
Devices with no available firmware updates: 
 • Integrated Webcam HD
 • Internal SPI Controller
 • PC611 NVMe SK hynix 512GB
 • TPM
 • TPM 2.0
Devices with the latest available firmware version:
 • Package level of Dell dock
 • RTS5413 in Dell dock
 • RTS5487 in Dell dock
 • VMM5331 in Dell dock
 • WD19TB
╔══════════════════════════════════════════════════════════════════════════════╗
║ Upgrade System Firmware from 1.15.0 to 1.16.0?                               ║
╠══════════════════════════════════════════════════════════════════════════════╣
║ This stable release fixes the following issues:                              ║
║                                                                              ║
║ • Firmware updates to address security vulnerabilities.                      ║
║                                                                              ║
║ XPS 13 9300 must remain plugged into a power source for the duration of the  ║
║ update to avoid damage.                                                      ║
╚══════════════════════════════════════════════════════════════════════════════╝
Perform operation? [Y|n]: 
Authenticating…          [***************************************]
Message recipient disconnected from message bus without replying
```

Last message is because fwupd core dumps:

```
           PID: 2070 (fwupd)
           UID: 0 (root)
           GID: 0 (root)
        Signal: 6 (ABRT)
     Timestamp: Wed 2023-03-29 09:28:19 AEST (1min 23s ago)
  Command Line: /usr/lib/fwupd/fwupd
    Executable: /usr/lib/fwupd/fwupd
 Control Group: /system.slice/fwupd.service
          Unit: fwupd.service
         Slice: system.slice
       Boot ID: b3b4b7369e204189ae33f5ce9b00d3fb
    Machine ID: 9eabba1a60b9408c970e3e88cfc74c46
      Hostname: lt
       Storage: /var/lib/systemd/coredump/core.fwupd.0.b3b4b7369e204189ae33f5ce9b00d3fb.2070.1680046099000000.zst (inaccessibl>
       Message: Process 2070 (fwupd) of user 0 dumped core.
                
                Stack trace of thread 2070:
                #0  0x00007fd0a45ea8ec n/a (libc.so.6 + 0x878ec)
[...]

PR #5662, opened 2023-03-30
uefi-capsule: Fix possible crash when getting UEFI report metadata

hughsie:
The fix of 0eec0c4a66a5e8a5957172e34fb675cfce646617 was incorrect as the hash table takes allocated strings, and mount_point is g_autofree.

Fixes https://github.com/fwupd/fwupd/issues/5656

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Do you think we need to re-spin?

superm1:
> Do you think we need to re-spin?

Could potentially branch at 1.8.13 and add the fix

hughsie:
I was going to fix the fuzzing issues and do a 1.8.14...

PR #5671, opened 2023-03-31
nvme: Add support for CA3 activation

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/plugins/nvme/README.md
+++ b/plugins/nvme/README.md
@@ -61,6 +61,7 @@ Since: 1.1.3
 ### Flags
 
 * `force-align` if image should be padded, since 1.2.4
+* `commit-ca3` download and activate immediately, since 1.8.15
 
 ## Vendor ID Security
 
--- a/plugins/nvme/fu-nvme-device.c
+++ b/plugins/nvme/fu-nvme-device.c
@@ -22,12 +22,24 @@ struct _FuNvmeDevice {
 	guint64 write_block_size;
 };
 
+#define FU_NVME_COMMIT_ACTION_CA0 0b000 /* replace only */
+#define FU_NVME_COMMIT_ACTION_CA1 0b001 /* replace, and activate on next reset */
+#define FU_NVME_COMMIT_ACTION_CA2 0b010 /* activate on next reset */
+#define FU_NVME_COMMIT_ACTION_CA3 0b011 /* replace, and activate immediately */
+
 /**
  * FU_NVME_DEVICE_FLAG_FORCE_ALIGN:
  *
  * Force alignment of the firmware file.
  */
 #define FU_NVME_DEVICE_FLAG_FORCE_ALIGN (1 << 0)
+
+/**
+ * FU_NVME_DEVICE_FLAG_COMMIT_CA3:
+ *
+ * Replace, and activate immediately rather than on next reset.
+ */
+#define FU_NVME_DEVICE_FLAG_COMMIT_CA3 (1 << 1)
 
 G_DEFINE_TYPE(FuNvmeDevice, fu_nvme_device, FU_TYPE_UDEV_DEVICE)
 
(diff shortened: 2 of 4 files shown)


plugins/nvme/README.md
```diff
@@ -58,6 +58,15 @@ The block size used for NVMe writes
...
 Since: 1.1.3
 
+### NvmeCommitAction
+
+The action for firmware commit, defaulting to `1`. Typical values:
+
+* `1`: Download and activate on next reboot
+* `3`: Download and activate immediately
+
+Since: 1.8.14
```
superm1, line 68: 15

superm1 (commented):
I'd like to understand bigger picture on this - I'm not aware of drives that support CA3 typically for safety and security.

hughsie:
This is a request from SMI -- and I challenged them if CA3 is always safe. I'm waiting for more details, and a list of affected drives.

plugins/nvme/README.md
hughsie: Fixed.

Homerrhsu:
> I'd like to understand bigger picture on this - I'm not aware of drives that support CA3 typically for safety and security.

After the CA1 FFU process, the drive can continue to receive instructions, but some behaviors of the hardware host may affect the boot-up process of the drive during the next boot up. However, after the CA3 FFU process, the next host request instruction will be operated by the new firmware.

Typically, SSDs can support both types, and it can be identified which FFUs are supported by the drive through the "identify controller" command to inform the host.

hughsie:
> which FFUs are supported by the drive through the "identify controller" command to inform the host

Oh, so can we detect a drive that only supports CA3 automatically? That would remove the need for the quirk to a specific PCI ID in the common case.

hughsie:
Answering my own question -- the CNS has: `FRMW.ActivationWithoutReset` which tells the client if the drive *supports* CA3, not if the device *does not support* CA1.

Homerrhsu:
> I'd like to understand bigger picture on this - I'm not aware of drives that support CA3 typically for safety and security.

It's true that not all drives support CA3 due to safety and security concerns. 
However, in our case, our drive does support CA3 and it is more stable than CA1.

Homerrhsu:
> Answering my own question -- the CNS has: `FRMW.ActivationWithoutReset` which tells the client if the drive _supports_ CA3, not if the device _does not support_ CA1.

You are right.
image

hughsie:
@superm1 I'm not sure we should enable CA3 if that bit is set -- I think from a regression-avoiding point of view I'd rather stick to CA1 as a default even if the drive supports CA3. Ideally we could have a new bit in the CNS to say *Drive does not support CA1* or *drive prefers CA3 over CA1* which means we don't use the quirk -- but that's a spec change.

superm1:
How about starting to capture data about whether CA1 or CA3 supported to understand how wide CA3 support actually is?

hughsie:
And what CAx would be useful to log too.

superm1:
> And what CAx would be useful to log too.

I had another idea.  How about we have a GUID created for NVME's that support CA3 that when F/W is targeted to specifically uses CA3.  Then no need to quirk it, and every vendor can decide on their own what's best for them by what they build the metadata against.

hughsie:
I think that's a slightly slippery slope. I think I'd rather have MD_SET_FLAGS on the proviso the the flags only affect device write_firmware rather than enumeration.

Homerrhsu:
@hughsie @superm1 
Thank you for your work on this! 
This PR is crucial for us, and I was wondering if there's an estimated timeline for when it might be merged. 
Any updates would be greatly appreciated!

superm1:
> I think that's a slightly slippery slope. I think I'd rather have MD_SET_FLAGS on the proviso the the flags only affect device write_firmware rather than enumeration.

A quirk list is an expensive way to do this that doesn't really scale IMO. How about in the LVFS release a new flag that can be used to say use CA3 for this disk?

superm1:
Or maybe overload `LVFS::UpdateProtocol` to have `org.nvmexpress` and also `org.nvmexpress.ca3`

hughsie:
I think https://github.com/fwupd/fwupd/pull/5694 is cleaner -- otherwise we're going to get `org.nvmexpress.ca3`, `org.nvmexpress.ca1.needs-shutdown` etc.

hughsie:
>  I was wondering if there's an estimated timeline for when it might be merged.

I've tagged it for 1.8.15 which we're aiming for end of the month.

PR #5674, opened 2023-04-01
synaptics-cxaudio: Convert to FuStruct which makes the code a lot less magic

hughsie:
This also makes it endian-clean.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Included in https://github.com/fwupd/fwupd/pull/5677

PR #5678, opened 2023-04-03
Reduce the amount of debugging when using --plugins

hughsie:
When we switched all the plugins to GObjectClass->constructed this meant that they all added firwmare GTypes and udev subsystems unconditionally, rather than only when the plugin is enabled. Switch to the almost-identical FuPluginClass->constructed vfunc which is only called if the plugin is enabled.

This makes debugging way easier as there's less debug output.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-plugin.c
+++ b/libfwupdplugin/fu-plugin.c
@@ -134,7 +134,20 @@ fu_plugin_get_name(FuPlugin *self)
 void
 fu_plugin_set_name(FuPlugin *self, const gchar *name)
 {
+	FuPluginPrivate *priv = GET_PRIVATE(self);
+
 	g_return_if_fail(FU_IS_PLUGIN(self));
+	g_return_if_fail(!priv->done_init);
+
+	if (g_strcmp0(name, fwupd_plugin_get_name(FWUPD_PLUGIN(self))) == 0) {
+		g_critical("plugin name set to original value: %s", name);
+		return;
+	}
+	if (fwupd_plugin_get_name(FWUPD_PLUGIN(self)) != NULL) {
+		g_debug("overwriting plugin name %s -> %s",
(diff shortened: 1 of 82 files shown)


libfwupdplugin/fu-plugin.c
```diff
@@ -134,7 +134,20 @@ fu_plugin_get_name(FuPlugin *self)
...
+
 	g_return_if_fail(FU_IS_PLUGIN(self));
+	g_return_if_fail(!priv->done_init);
+
+	if (g_strcmp0(name, fwupd_plugin_get_name(FWUPD_PLUGIN(self))) == 0) {
+		g_critical("plugin name set to original value: %s", name);
+		return;
+	}
+	if (fwupd_plugin_get_name(FWUPD_PLUGIN(self)) != NULL) {
+		g_debug("overwriting plugin name %s -> %s",
```
superm1, line 147: `g_print`, no?

plugins/wacom-usb/fu-wac-plugin.c
```diff
@@ -39,10 +39,16 @@ fu_wac_plugin_init(FuWacPlugin *self)
 }
 
 static void
-fu_wac_plugin_constructed(GObject *obj)
+fu_wac_plugin_constructed_early(GObject *obj)
```
superm1, line 42: rather than `plugin_constructed_early` could this be `object_constructed` perhaps?

hughsie: I think they should all have the fu_xxx_plugin prefix -- but fu_xxx_plugin_object_constructed would be okay too?

hughsie: Fixed.

PR #5685, opened 2023-04-05
Tag releases with an extra flag if they have a report we trust

hughsie:
For instance, this could be used to only show a desktop notification if the report has been verified to work on ChromeOS, or to only auto-install the update on an automotive system if tested by Red Hat QA.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/daemon.conf
+++ b/data/daemon.conf
@@ -58,6 +58,23 @@ AllowEmulation=false
 # UIDs that should marked as trusted
 TrustedUids=
 
+# Vendor reports matching these globs will have releases marked as `trusted-report`, e.g.
+#
+# * `DistroId=chromeos`
+# * `DistroId=fedora&VendorId=19`
+# * `DistroId=fedora&VendorId=$OEM`
+# * `DistroId=fedora;DistroId=rhel&DistroVersion=9`
+#
+# NOTE: a `VendorId` of `$OEM` represents the OEM vendor ID of the vendor that owns the firmware,
+# for example, where Lenovo QA has generated a signed report for a Lenovo laptop.
+#
+# There are also three os-release values available, `$ID`, `$VERSION_ID` and `$VARIANT_ID`, which
+# allow globs like:
+#
+# * `DistroId=$ID`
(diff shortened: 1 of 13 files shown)


data/daemon.conf
```diff
@@ -58,6 +58,12 @@ AllowEmulation=false
 # UIDs that should marked as trusted
 TrustedUids=
 
+# Vendor reports matching these globs will have releases marked as `trusted-report`, e.g.
+# * `DistroId:chromeos`
+# * `DistroId:fedora,VendorId:19`
+# * `DistroId:fedora;DistroId:rhel,DistroVersion:9`
+TrustedReports=
```
superm1, line 65: Pandoras box.  So package maintainers will get to decide which reports are trusted?  That seems weird to me.

How about instead we have some level of opt in for the thing we're running on for trusted reports?

superm1: > Indirectly, yes. For ChromeOS I see them just hardcoding that, and possibly the same for RHEL. For other corporate deployments I figured this would be set in an Ansible script somewhere.

I think the problem with reports being trusted just because they were reports run on a particular OS.  It's entirely possible to game this system so this seems incongruent to me.

They should be trusted because they were run by a trusted person, or trusted software stack that **signed** the report.

hughsie: Ohh it's both. Only signed reports end up in the metadata. I understand what you mean about "running on" now, i.e. we want something like $distro_id$ available as a placeholder.

superm1: Yup!

hughsie: Okay, I've added that -- but one thing I think we also want is something like `VendorId:$self` i.e. I trust the signed report if it's been uploaded by the vendor that owns the firmware rather than trusting the any signed report from any QA team.

hughsie: Okay, done, but it depends on something like https://github.com/fwupd/fwupd/pull/5705 first.

hughsie: Okay, done.

superm1 (approved):

PR #5692, opened 2023-04-06
nordic-hid: Add nRF52 Desktop Keyboard to the quirk

MarekPieta:
Change adds Product ID of nRF52 Desktop Keyboard to the quirk file.
This is needed to support nRF52 Desktop Keyboard in the plugin.


MarekPieta:
Hi @hughsie, @d4s,
We would like to introduce support for an additional device in the plugin (nRF52 Desktop keyboard reference design).
Can we append its PID to the quirk file to achieve this ?

hughsie (approved):

PR #5693, opened 2023-04-07
plugins: nordic-hid: Allow board to override bootloader from quirk

MarekPieta:
Change allows board to override bootloader defined in quirk file. The value defined in the quirk file can still be used for boards that do not provide information about the bootloader through the configuration channel.

### DIFF
--- a/plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
+++ b/plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
@@ -513,16 +513,13 @@ fu_nordic_hid_cfg_channel_get_bl_name(FuNordicHidCfgChannel *self, GError **erro
 			return FALSE;
 
 		/* check if not set via quirk */
-		if (self->bl_name != NULL &&
-		    strncmp(self->bl_name, (const char *)res->data, res->data_len) != 0) {
-			g_set_error(
-			    error,
-			    G_IO_ERROR,
-			    G_IO_ERROR_INVALID_DATA,
-			    "bootloader in quirk file is '%s' while the board is supporting '%s'",
-			    self->bl_name,
-			    g_strndup((const gchar *)res->data, res->data_len));
-			return FALSE;
+		if (self->bl_name != NULL) {
+			g_autofree gchar *tmp = g_strndup((const gchar *)res->data, res->data_len);
+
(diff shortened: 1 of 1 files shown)


MarekPieta:
Hi @hughsie, @d4s,
Can we allow board to override the bootloader name defined in quirk, so that the quirk value would be used only as a fallback?

hughsie:
I'm a bit confused; why would the quirk entry be *wrong*?

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -513,16 +513,11 @@ fu_nordic_hid_cfg_channel_get_bl_name(FuNordicHidCfgChannel *self, GError **erro
...
-			    error,
-			    G_IO_ERROR,
-			    G_IO_ERROR_INVALID_DATA,
-			    "bootloader in quirk file is '%s' while the board is supporting '%s'",
-			    self->bl_name,
-			    g_strndup((const gchar *)res->data, res->data_len));
-			return FALSE;
+		if (self->bl_name != NULL) {
+			g_debug("Bootloader readout '%s' overrides bootloader from quirk '%s'",
+				g_strndup((const gchar *)res->data, res->data_len),
```
hughsie, line 518: yes, this is fine -- but g_strndup leaks the string. you'll need to do something like `g_autofree gchar *tmp = g_strndup...` to use it in the g_debug.

MarekPieta: Fixed. Thanks for pointing this out.
btw. Is the leak present on in the current implementation from `main` branch? We pass the pointer as format argument in the `g_set_error`, is it automatically freed in that case?

hughsie: Nah, the leak is there on the `main` branch too -- we missed that on initial review. :)

hughsie (approved):

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
MarekPieta: Thanks a lot for the review :)

hughsie: I've invited you as a contributor -- you can merge this now if you want.

PR #5694, opened 2023-04-07
Allow setting device flags in the metadata

hughsie:
NOTE: This can only be used to affect device write, not device enumeration.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.c
+++ b/libfwupdplugin/fu-device.c
@@ -245,6 +245,8 @@ fu_device_internal_flag_to_string(FuDeviceInternalFlags flag)
 		return "no-probe-complete";
 	if (flag == FU_DEVICE_INTERNAL_FLAG_SAVE_INTO_BACKUP_REMOTE)
 		return "save-into-backup-remote";
+	if (flag == FU_DEVICE_INTERNAL_FLAG_MD_SET_FLAGS)
+		return "md-set-flags";
 	return NULL;
 }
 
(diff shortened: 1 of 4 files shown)


hughsie:
This needs a lot of supporting LVFS code too, and a lot of server-side validation on each of the valid flags for each protocol.

hughsie:
LVFS side: https://gitlab.com/fwupd/lvfs-website/-/merge_requests/1403

libfwupdplugin/fu-device.c
```diff
@@ -5519,6 +5534,8 @@ fu_device_ensure_from_component(FuDevice *self, XbNode *component)
 		fu_device_ensure_from_component_signed(self, component);
 	if (fu_device_has_internal_flag(self, FU_DEVICE_INTERNAL_FLAG_MD_SET_VERFMT))
 		fu_device_ensure_from_component_verfmt(self, component);
+	if (fu_device_has_internal_flag(self, FU_DEVICE_INTERNAL_FLAG_MD_SET_FLAGS))
```
superm1, line 5537: I guess you're missing to add this flag to NVME plugin still

superm1 (approved):
approach looks good to me

libfwupdplugin/fu-device.c
hughsie: Yup, will do in the other PR.

PR #5705, opened 2023-04-16
Add a report flag if generated by the OEM

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupd/fwupd-report.h
```diff
@@ -25,6 +25,41 @@ struct _FwupdReportClass {
...
+ * The report was generated by the OEM.
+ *
+ * Since: 1.9.1
+ */
+#define FWUPD_REPORT_FLAG_FROM_OEM (1llu << 0)
+
+/**
+ * FWUPD_REPORT_FLAG_UNKNOWN:
+ *
+ * The plugin flag is unknown.
```
superm1, line 49: Report

libfwupd/fwupd-report.h
```diff
@@ -25,6 +25,41 @@ struct _FwupdReportClass {
...
+ * This is usually caused by a mismatched libfwupdplugin and daemon.
+ *
+ * Since: 1.9.1
+ */
+#define FWUPD_REPORT_FLAG_UNKNOWN G_MAXUINT64
+
+/**
+ * FwupdReportFlags:
+ *
```
superm1, line 58: Report

src/fu-release.c
```diff
@@ -323,6 +323,10 @@ fu_release_load_test_result(FuRelease *self, XbNode *n, GError **error)
 	if (custom != NULL) {
 		for (guint i = 0; i < custom->len; i++) {
 			XbNode *c = g_ptr_array_index(custom, i);
+			if (g_strcmp0(xb_node_get_attr(c, "key"), "FromOEM") == 0) {
```
superm1, line 326: Isn't this named improperly? It's more like from a trusted source. For example if Dell shipped a dock update but it was validated by say Google in ChromeOS and Google happened to push a signed report it should be given some weight just like this I would think.

libfwupd/fwupd-report.h
hughsie: Fixed, ta.

libfwupd/fwupd-report.h
hughsie: Fixed.

src/fu-release.c
hughsie: The logic I was aiming for was to have something like `DistroId:chromeos,VendorId:$SELF;DistroId:chromeos,VendorId:19` i.e. a trusted report is anything by the original OEM, or by the Google QA team.

superm1 (approved):

PR #5706, opened 2023-04-16
wacom-usb: Use FuStruct and slightly safer parsing code

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@Joshua-Dickens can you verify this still works please.

Joshua-Dickens:
Running the same test I used for my initial code it seems to work just fine. Ran the test firmware update which cleared through without error, and then tested touch functionality on the tablet and it works as expected.

PR #5710, opened 2023-04-17
Wip/hughsie/build archive

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdtool
+++ b/data/bash-completion/fwupdtool
@@ -1,5 +1,6 @@
 _fwupdtool_cmd_list=(
 	'activate'
+	'build-cabinet'
 	'clear-history'
 	'esp-list'
 	'esp-mount'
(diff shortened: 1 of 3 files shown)


data/bash-completion/fwupdtool
```diff
@@ -1,6 +1,6 @@
 _fwupdtool_cmd_list=(
 	'activate'
-	'build-firmware'
+	'build-archive'
```
superm1, line 3: if no strong feeling, how about changing this to `build-cab-archive`?

superm1 (dismissed):

data/bash-completion/fwupdtool
superm1: That's good

superm1 (approved):

PR #5711, opened 2023-04-17
trivial: Remove and unexport some unused code

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
This is great!  How did you discover all of this?  If it's a tool that can be friendly to CI maybe we should have it in CI so we don't introduce more dead code in the future.

PR #5712, opened 2023-04-17
Add a helper script that checks for unused symbols

hughsie:
This isn't plumbed into precommit as it operates on the locally built source objects, not the source code. There are also some exceptions we want to ignore.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 other ideas welcome!

superm1:
How about just tying it to the end of a CI job that does compile?

superm1 (approved):

PR #5714, opened 2023-04-18
dell-dock: trigger passive flow for usb4 from intel-usb4 plugin

CragW:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

The USB4 dock requires `usable-during-update` flag to enable updating the NVM firmware in a cycle of the Type-C cable is disconnected and re-connected.


hughsie (approved):
Did I break this? If so, very sorry.

CragW:
> Did I break this? If so, very sorry.

No worries, I should have remembered to add the check for the new plugin when it's moved. :)  I will send another for 1_8_X.

PR #5723, opened 2023-04-19
mtd: Allow specifying the metadata offset and size

hughsie:
Some MTD partitions are huge, and take a long time to dump. If we know the offset the metadata is going to be located at, allow it to be specified in the quirk file.

This speeds up reading uSWID metadata for the BMC partition by several orders of magnitude.

Also, don't fail to add the MTD device if the metdata is not found.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@dylanmtaylor this might be interesting for you too.

superm1 (approved):

PR #5730, opened 2023-04-20
Show the default config values in 'man fwupd.conf'

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation


superm1 (commented):
Any reason not to use jinja2? This feels like a hand grown version of it.

hughsie:
> Any reason not to use jinja2

Wouldn't that add it to the list of hard build-requires on all distros? I suppose I could make it *look* like jinja2, i.e. 

    -The `@SYSCONFDIR@/fwupd/fwupd.conf` file is the main configuration file for the fwupd daemon.
   +The `{{SYSCONFDIR}}/fwupd/fwupd.conf` file is the main configuration file for the fwupd daemon.

superm1:
> > Any reason not to use jinja2
> 
> Wouldn't that add it to the list of hard build-requires on all distros? I suppose I could make it _look_ like jinja2, i.e.
> 
> ```
> -The `@SYSCONFDIR@/fwupd/fwupd.conf` file is the main configuration file for the fwupd daemon.
> +The `{{SYSCONFDIR}}/fwupd/fwupd.conf` file is the main configuration file for the fwupd daemon.
> ```

I mean man pages are optional and also Jinja2 is readily available.  Even if it's not in the repos for some reason it's in PyPI.  We could make man a tri state feature if it's available too.

src/fu-engine-config.c
```diff
@@ -29,6 +29,19 @@ struct _FuEngineConfig {
...
+#define FU_DAEMON_CONFIG_DEFAULT_URI_SCHEMES	       "file;https;http;ipfs"
+#define FU_DAEMON_CONFIG_DEFAULT_UPDATE_MOTD	       TRUE
+#define FU_DAEMON_CONFIG_DEFAULT_IGNORE_POWER	       FALSE
+#define FU_DAEMON_CONFIG_DEFAULT_ONLY_TRUSTED	       TRUE
+#define FU_DAEMON_CONFIG_DEFAULT_SHOW_DEVICE_PRIVATE   TRUE
+#define FU_DAEMON_CONFIG_DEFAULT_ALLOW_EMULATION       FALSE
+#define FU_DAEMON_CONFIG_DEFAULT_ENUMERATE_ALL_DEVICES TRUE
+#define FU_DAEMON_CONFIG_DEFAULT_TRUSTED_UIDS	       NULL
+#define FU_DAEMON_CONFIG_DEFAULT_HOST_BKC	       NULL
+#define FU_DAEMON_CONFIG_DEFAULT_TRUSTED_REPORTS       "VendorId=$OEM"
```
superm1, line 43: can we perhaps put all the defaults for everything into a single source of truth header?  `fu-engine-config-defaults.h`.  Then the thing that parses it looks at one file and the plugins and engine that use it just have it included.

hughsie:
> I mean man pages are optional and also Jinja2 is readily available

I'm thinking of RHEL where there's no Jinja2 in the buildroot, we can't use pip when building and we **have** to have manpages :(

superm1:
> > I mean man pages are optional and also Jinja2 is readily available
> 
> I'm thinking of RHEL where there's no Jinja2 in the buildroot, we can't use pip when building and we **have** to have manpages :(

But can't you specify to have an extra package in the buildroot?  It seems that jinja2 exists in RHEL, otherwise pandas wouldn't work (and https://bugzilla.redhat.com/show_bug.cgi?id=2153843 wouldn't exist either)

hughsie:
>  It seems that jinja2 exists in RHEL

Ahh, it's been dragged in via gi-docgen so maybe it's fine now. Will play.

superm1 (approved):
Looks great!  I do have one more suggestion though...
Now that the defaults can be kept in one place and written/use "everywhere" they can also be added to fwupd.conf in commented out for the same way.

hughsie:
> they can also be added to fwupd.conf

I think I'd rather keep this as minimal as possible -- otherwise we have to work out a way to _keep it up to date_ when we add new options.

superm1:
> > they can also be added to fwupd.conf
> 
> I think I'd rather keep this as minimal as possible -- otherwise we have to work out a way to _keep it up to date_ when we add new options.

Okay

Issue #5738, opened 2023-04-21
plugins: nordic-hid: Invalid VID and PID of devices connected via dongle


MarekPieta:
**Describe the bug**
`nordic-hid` plugin reports VID and PID of the nRF Desktop dongle for peripherals connected via the dongle.
That  breaks our DFU mechanism (the *.cab packages cannot be properly matched because of unexpected GUID value).
Previously such peripherals used hardcoded VID and PID of `0x0000`, which was taken into account in GUIDs used by *.cab file.
Every peripheral was assigned an additional GUID with VID and PID of `0x0000` instead of peripheral's actual VID and PID.
With the current implementation the workaround cannot be applied - assigning every peripheral GUID related to each supported dongle the peripheral can be connected through would significantly limit interoperability.

**Steps to Reproduce**
Use `fwupdtool` with an nRF Desktop HID peripheral device connected through nRF Desktop dongle.
Check GUID strings reported for the device.

**Expected behavior**
The devices should report their own VID and PID.
Since this may not be possible for devices connected through nRF Desktop dongle (host has no direct access to the VID and PID values), we can either pass VID and PID over configuration channel protocol or use a fixed, hardcoded value (like 0x0000 which was used before).
Keep in mind that passing VID and PID over configuration channel would not be possible for devices with legacy firmware.

**fwupd version information**
Tested with `fwupdtool` built locally from sources on d15eab455e288e84603ca77d9f267c4fa58d571f.

**Additional details**
The `fwupd` regression seems introduced by the following commit c4ca02629487d3f26278f47e4720167b1ae04eb4.
So that `fwupd` versions since `1.7.5` seems affected.

MarekPieta:
Issue is fixed by the following PR: https://github.com/fwupd/fwupd/pull/5699. The PR is already merged.

We still need to introduce a way to bring back reporting VID and PID of `0x0000` for devices that do not report their VID and PID over configuration channel (devices with legacy firmware would not support the new functionality).

hughsie:
@MarekPieta okay to close this now?

MarekPieta:
Yes, since https://github.com/fwupd/fwupd/pull/5742 is merged I think we can close this one.
Thanks a lot for help with fixing this one

PR #5739, opened 2023-04-21
nordic-hid: Fix a regression where the InstanceIDs accidentally changed

hughsie:
We really want to use the VID/PID of 0x0000 if there is no devinfo section.

Fixes https://github.com/fwupd/fwupd/issues/5738

Regressed in c4ca026.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
+++ b/plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
@@ -1034,10 +1034,9 @@ fu_nordic_hid_cfg_channel_setup(FuDevice *device, GError **error)
 	fu_device_add_instance_strsafe(device, "BL", self->bl_name);
 	fu_device_add_instance_strsafe(device, "GEN", self->generation);
 
-	if (self->vid != 0x00 && self->pid != 0x00) {
-		fu_device_add_instance_u16(device, "VEN", self->vid);
-		fu_device_add_instance_u16(device, "DEV", self->pid);
-	}
+	/* peripherals without devinfo really do use hardcoded VID and PID of 0x0000 */
+	fu_device_add_instance_u16(device, "VEN", self->vid);
+	fu_device_add_instance_u16(device, "DEV", self->pid);
 
 	/* For the default generation, generate GUID without the generation parameter.
 	 * Required for compatibility with already released application images.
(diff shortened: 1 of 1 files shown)


hughsie:
@MarekPieta does this kind of thing work?

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -1034,10 +1034,9 @@ fu_nordic_hid_cfg_channel_setup(FuDevice *device, GError **error)
 	fu_device_add_instance_strsafe(device, "BL", self->bl_name);
 	fu_device_add_instance_strsafe(device, "GEN", self->generation);
 
-	if (self->vid != 0x00 && self->pid != 0x00) {
-		fu_device_add_instance_u16(device, "VEN", self->vid);
-		fu_device_add_instance_u16(device, "DEV", self->pid);
-	}
+	/* peripherals without devinfo really do use hardcoded VID and PID of 0x0000 */
```
MarekPieta, line 1037: Can we use the hardcoded values only for device that is connected via a dongle (in such case, the peripheral is a child device of the dongle)? For directly connected devices, reported PID and VID should be proper, because host can get information about VID and PID.

MarekPieta (changes requested):

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
hughsie: isn't what this patch does? i.e. we don't *set* the VID to 0x0 we just use whatever udev told us. It's kinda hard for me to debug this without actual hardware tbh.

hughsie: Can you commit to this branch what works please? i.e. with a device without a devinfo section attached to a dongle.

hughsie: Thanks! :+1:

MarekPieta: I checked locally that the proposed fix sets VID and PID to 0x0000 also for directly connected devices (which is not intended). I will try to slightly update it.

MarekPieta: Fix is available here: https://github.com/fwupd/fwupd/pull/5742 (very small change :) )

PR #5759, opened 2023-04-25
trivial: Document how to use rustgen

hughsie:
We can't very well ask plugin authors to use it if the docs are in my head.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -839,3 +839,132 @@ The `fu_dump_full` and `fu_dump_raw` functions implement the
 printing of a binary buffer to the console as a stream of bytes in
 hexadecimal. See `libfwupdplugin/fu-common.c` for their definitions, you
 can find many examples of how to use them in the plugins code.
+
+## The rustgen Helper
+
+The rustgen script generates C source files that allow parsing, modifying and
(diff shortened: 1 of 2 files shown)


docs/tutorial.md
```diff
@@ -839,3 +839,132 @@ The `fu_dump_full` and `fu_dump_raw` functions implement the
...
+and constant data of variable size.
+The generated code is endian safe and if used correctly, is also safe
+against malicious data.
+
+In most cases the structure or enumeration will be defined in a `.rs`
+file -- which is the usual file extension of Rust programs.
+This was done as the format is heavily inspired by Rust, and it makes editor
+highlighting support work correctly.
+Although these files *look like* Rust files they're *not actually compiled by
+rustc*, so small differences may be noticeable.
```
superm1, line 863: I had assumed the eventual intention was to swap out rustc for these. Is that not true?

superm1 (approved):

docs/tutorial.md
hughsie: Eventually; but I don't want to scare anybody off with a rust dep right now.

PR #5764, opened 2023-04-27
Enable the retimer offline mode depending on HWIDs

hughsie:
One config value is not granular enough; for the same build of fwupd on RHEL or ChromeOS we only want to target the hardware where the EC is going to actually do what we tell it to -- rather than crash...

Type of pull request:

- [X] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/docs/fwupd.conf.md
+++ b/docs/fwupd.conf.md
@@ -258,10 +258,6 @@ The `[thunderbolt]` section can contain the following parameters:
 
   Forces delaying activation until shutdown/logout/reboot.
 
-**RetimerOfflineMode={{FU_THUNDERBOLT_CONFIG_DEFAULT_RETIMER_OFFLINE_MODE}}**
-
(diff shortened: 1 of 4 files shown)


superm1:
Given this is for stuff with a more malleable BIOS than most could we have a device property that comes from cros_ec perhaps?  It would mean a lot less lists of hwids for every single generation of chromebook.

superm1:
Couldn't such a case be accomplished with a quirk?  As an alternative, is there a _DSD that indicates this all should work that we can detect?

hughsie:
> is there a _DSD that indicates this all should work that we can detect

@djcampello any ideas?

djcampello:
Not sure about that. Maybe @bleungatchromium have some insights

hughsie:
Lets merge this as it unblocks those two platforms (without risking crashing everything else!), and then I'll talk to Benson this week about asking the EC for more info.

PR #5769, opened 2023-04-28
Wip/hughsie/remote depsolve

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/remotes.d/lvfs-testing.conf
+++ b/data/remotes.d/lvfs-testing.conf
@@ -7,6 +7,6 @@ MetadataURI=https://cdn.fwupd.org/downloads/firmware-testing.xml.@compression@
 ReportURI=https://fwupd.org/lvfs/firmware/report
 #Username=
 #Password=
-OrderBefore=lvfs,dell-esrt
+OrderBefore=lvfs
 AutomaticReports=false
 ApprovalRequired=false
--- a/data/remotes.d/lvfs.conf
+++ b/data/remotes.d/lvfs.conf
@@ -6,7 +6,6 @@ Title=Linux Vendor Firmware Service
 MetadataURI=https://cdn.fwupd.org/downloads/firmware.xml.@compression@
 ReportURI=https://fwupd.org/lvfs/firmware/report
 SecurityReportURI=https://fwupd.org/lvfs/hsireports/upload
-OrderBefore=dell-esrt
 AutomaticReports=false
 AutomaticSecurityReports=false
 ApprovalRequired=false
--- a/docs/fwupd.conf.md
+++ b/docs/fwupd.conf.md
@@ -116,6 +116,10 @@ The `[daemon]` section can contain the following parameters:
 
   A host best known configuration is used when using `fwupdmgr sync` which can downgrade firmware to factory versions or upgrade firmware to a supported config level. e.g. **vendor-factory-2021q1**
 
+**ReleaseDedupe={{FU_DAEMON_CONFIG_DEFAULT_RELEASE_DEDUPE}}**
+
+  Deduplicate duplicate releases by the archive checksum are available from more than one source.
+
 **EspLocation=**
 
   Override the location used for the EFI system partition (ESP) path.
--- a/docs/fwupd.conf.md
+++ b/docs/fwupd.conf.md
@@ -120,6 +120,15 @@ The `[daemon]` section can contain the following parameters:
 
   Deduplicate duplicate releases by the archive checksum are available from more than one source.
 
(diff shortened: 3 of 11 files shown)


data/remotes.d/lvfs-testing.conf
```diff
@@ -7,6 +7,7 @@ MetadataURI=https://cdn.fwupd.org/downloads/firmware-testing.xml.@compression@
 ReportURI=https://fwupd.org/lvfs/firmware/report
 #Username=
 #Password=
-OrderBefore=lvfs,fwupd
+OrderBefore=lvfs,dell-esrt
```
superm1, line 10: dell-esrt isn't installed in all configs.  This might need to be dynamic.

data/remotes.d/lvfs.conf
```diff
@@ -6,7 +6,8 @@ Title=Linux Vendor Firmware Service
 MetadataURI=https://cdn.fwupd.org/downloads/firmware.xml.@compression@
 ReportURI=https://fwupd.org/lvfs/firmware/report
 SecurityReportURI=https://fwupd.org/lvfs/hsireports/upload
-OrderBefore=fwupd
+OrderBefore=dell-esrt
+OrderAfter=backup,vendor,vendor-directory
```
superm1, line 10: Isn't vendor-directory just an example?  I think this makes it a lot more confusing for someone who is using it as a _template_ and drops a new file in.  They would have no idea that they need to replace it in these other remote conffiles.

superm1 (commented):
As an alternative; could you just set file remotes as higher priority than web remotes?

hughsie:
>  As an alternative; could you just set file remotes as higher priority than web remotes?

Yes, I thought about that; but do we *always* want to prefer the local one? You could argue that if I'm shipping the exact same version in a remote and there is internet access that might be better because it's got the latest update description for instance. Maybe it makes sense to prefer local if there is no explicit OrderAfter or OrderBefore?

data/remotes.d/lvfs.conf
hughsie: > Isn't vendor-directory just an example

Google is using it "as-in" on ChromeOS :)

> and drops a new file in

I don't know if it makes more sense to edit the "worse" file to the "better" one. I guess whatever policy you have for download v.s. local remotes is probably going to be the same, i.e. lvfs probably wants the same dep order as say dell-embargo -- and vendor-directory-another-one wants the same dep order as vendor-directory.

data/remotes.d/lvfs-testing.conf
hughsie: If it's not installed you get a g_debug message -- hidden by default.

superm1:
> >  As an alternative; could you just set file remotes as higher priority than web remotes?
> 
> Yes, I thought about that; but do we *always* want to prefer the local one? You could argue that if I'm shipping the exact same version in a remote and there is internet access that might be better because it's got the latest update description for instance. Maybe it makes sense to prefer local if there is no explicit OrderAfter or OrderBefore?

If they're identical how about just merging entries?

hughsie:
> If they're identical how about just merging entries?

Then we have to define identical. I guess container [cab] checksum is probably the best way.

data/remotes.d/lvfs.conf
hughsie: Maybe this is a fwupd.conf thing? i.e. `DuplicateReleasePolicy = local` for the common case?

data/remotes.d/lvfs-testing.conf
superm1: Yeah but can't we also just use a template to build it dynamically?

data/remotes.d/lvfs.conf
superm1: That seems reasonable.

hughsie: Okay, I've got a prototype that works -- adding ReleaseDedupe and ReleasePriority as different conf keys -- as they're different things and trying to squash them into one thing gets messy. I'll split it into a proper patch series tomorrow.

hughsie:
> Then we have to define identical. I guess container [cab] checksum is probably the best way.

Added.

data/remotes.d/lvfs.conf
hughsie: Done; it's much better IMHO.

data/remotes.d/lvfs-testing.conf
hughsie: It's removed now :)

superm1 (approved):

PR #5775, opened 2023-04-29
dfu: Use rustgen to remove a lot of boilerplate

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
You lose a lot of the code comments by moving but otherwise no concerns.

hughsie:
The comments were really for when it was exported as libdfu; I can add them to the .rs file if you think they were useful.

superm1:
> The comments were really for when it was exported as libdfu; I can add them to the .rs file if you think they were useful.

no big deal, let 'em go.

PR #5792, opened 2023-05-07
Delete the obsolete .gz files if the remote is now using .xz

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/src/fu-remote-list.c
+++ b/src/fu-remote-list.c
@@ -9,6 +9,7 @@
 #include "config.h"
 
 #include <glib/gi18n.h>
+#include <glib/gstdio.h>
 #include <xmlb.h>
 
 #ifdef HAVE_INOTIFY_H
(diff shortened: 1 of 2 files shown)


src/fu-remote-list.c
```diff
@@ -185,6 +186,47 @@ _fwupd_remote_build_component_id(FwupdRemote *remote)
...
+	g_autoptr(GPtrArray) files = NULL;
+
+	/* sanity check */
+	if (fn_cache == NULL)
+		return TRUE;
+	if (!g_str_has_suffix(fn_cache, ".xz"))
+		return TRUE;
+	if (fwupd_remote_get_kind(remote) != FWUPD_REMOTE_KIND_DOWNLOAD)
+		return TRUE;
+
```
d4s, line 203: Shouldn't be here additional check if remote is enabled?
At the moment it is cleaning old files for remotes which are disabled in configuration files.

d4s:
LGTM.
If the cleanup of all available old *.gz is expected regardless the remote is enabled or not please feel free to merge.

src/fu-remote-list.c
hughsie: Good catch fixed.

Issue #4259, opened 2022-02-08
Vulnerable firmware can't get upgraded: "failed to get device after detach: failed to wait for detach replug: device did not come back" or reverts back to old version after restart


mYnDstrEAm:
**Describe the bug**
The issue is described in more detail and partly in chronological order in my question about it [here](https://unix.stackexchange.com/questions/689284/how-to-solve-failed-to-get-device-after-detach-failed-to-wait-for-detach-replu).

I was trying to upgrade the vulnerable Unifying Receiver firmware with `fwupdmgr update` but I'm getting this output on Debian11/KDE:
```
Unifying Receiver and all connected devices may not be usable while updating. Continue with update? [Y|n]: Y
Downloading…             [***************************************]
Decompressing…           [***************************************]
Authenticating…          [***************************************]
Authenticating…          [***************************************]
failed to get device after detach: failed to wait for detach replug: device {id} did not come back
```
I then installed `solaar` (and I was never prompted to install it on Debian which is probably another issue) and disconnected one of the two receivers and could run the updates (it was two updates) after doing many retries.

Even stranger than this error is that it displayed that it has the latest version afterwards (and I checked this multiple times) but it's back to the old firmware after a restart. (Like so:)

```
Devices with no available firmware updates: 
 [...]
 • System Firmware
Devices with the latest available firmware version:
 • Unifying Receiver
```

After rebooting I had to run the command a few times until it got so far as to prompt me about the root password but I still get the same error or (far less often) this error instead: `Message recipient disconnected from message bus without replying`.

When I unplug the dongle during the time it waits for a few seconds after the second Authenticating is displayed and plug it back in right after it shows this error instead: `not supported`.

Maybe the problem is connected to how integrity is ensured for firmware updates [discussion about it **here**](https://github.com/fwupd/fwupd/discussions/4261), but I suspect the cause to be something else. I'm having this problem with both of the Unifying Receivers.

**Steps to Reproduce**
On Debian11/KDE with Wayland try to update the vulnerable Unifying Receiver firmware.

**Expected behavior**
It should successfully update at the first run of `fwupdmgr update`.

**fwupd version information**

```
client version: 1.5.7
compile-time dependency versions
        gusb:   0.3.5

daemon version: 1.5.7
```

Please note how you installed it (`apt`, `dnf`, `pacman`, source, etc): apt (Apper)

**fwupd device information**
Please provide the output of the fwupd devices recognized in your system.

```shell
[...]

PR #5793, opened 2023-05-07
logitech-hidpp: Add support for more TI Unifying receiver variants

Rongronggg9:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

The patch enables fwupd to work with two new variants of Logitech Unifying receiver (TI).
I have tested the patch on my `C-U0012` Pico receiver and it works totally fine. Differing from previous `C-U0012`s whose PID in DFU mode is `AAAD`, mine is `AAE5`.
There is another variant get supported in this patch, whose PID in DFU mode is `AAF8`. I do not have such a receiver and the PID is documented in https://github.com/Logitech/fw_updates/blob/1ddde61310573ecea54c4204b401393a90f4bbae/RQR24/RQR24.11/RQR24.11_B0036.txt#L22. I guess it is `C-U0016`, a signal-enhanced Unifying receiver bundled with `K600` (ref: https://post.smzdm.com/p/aqnlx78p), which is known to be driven by a TI CC2544 (ref: https://fccid.io/JNZCU0016/Internal-Photos/Internal-Photos-3182497.pdf). I've managed to buy one. Once I receive it, I will update more information here.

See also https://github.com/fwupd/fwupd/issues/4259#issuecomment-1537361667

Note: I did not update macros in `plugins/logitech-hidpp/fu-logitech-hidpp-common.h` as no C code needs to reference the new PIDs.

### DIFF
--- a/plugins/logitech-hidpp/logitech-hidpp.quirk
+++ b/plugins/logitech-hidpp/logitech-hidpp.quirk
@@ -60,6 +60,23 @@ FirmwareSizeMin = 0x4000
 CounterpartGuid = HIDRAW\VEN_046D&DEV_C52B
 InstallDuration = 18
 
+# Texas Pico (another variant) / C-U0016
+[USB\VID_046D&PID_AAE5]
+Plugin = logitech_hidpp
+GType = FuLogitechHidPpBootloaderTexas
+FirmwareSizeMin = 0x4000
+CounterpartGuid = HIDRAW\VEN_046D&DEV_C52B
+InstallDuration = 18
+
+# Texas Unknown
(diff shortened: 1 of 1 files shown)


hughsie:
I've asked Logitech to verify this. Thanks.

Rongronggg9:
> I've managed to buy one. Once I receive it, I will update more information here.

The DFU mode PID of `C-U0016` is `AAE5`.

hughsie:
> The DFU mode PID of C-U0016 is AAE5.

Thanks for checking that; I assume the update works? If so, can you remove the `perhaps` in the quirk entry please.

Rongronggg9:
> I assume the update works?

Yes, upgraded with no error.

> If so, can you remove the perhaps in the quirk entry please.

Done.

hughsie (approved):

PR #5798, opened 2023-05-09
Fix behavior of EspLocation if UDisks is not available

nicholasbishop:
When an ESP mount path is set in the daemon config via the `EspLocation` setting, UDisks should not be required for finding the ESP path. In `fu_volume_new_esp_for_path`, move up the check of `esp_path` so that it returns that mount immediately if it's a valid directory. This needs to happen _before_ the call to `fu_volume_new_by_kind`, as that function requires UDisks to be available.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-volume.c
+++ b/libfwupdplugin/fu-volume.c
@@ -757,11 +757,12 @@ fu_volume_new_esp_for_path(const gchar *esp_path, GError **error)
 	g_return_val_if_fail(esp_path != NULL, NULL);
 	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
 
+	/* check if it's a valid directory already */
+	if (g_file_test(esp_path, G_FILE_TEST_IS_DIR))
+		return fu_volume_new_from_mount_path(esp_path);
+
 	volumes = fu_volume_new_by_kind(FU_VOLUME_KIND_ESP, &error_local);
 	if (volumes == NULL) {
-		/* check if it's a valid directory already */
-		if (g_file_test(esp_path, G_FILE_TEST_IS_DIR))
-			return fu_volume_new_from_mount_path(esp_path);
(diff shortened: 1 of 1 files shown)


superm1 (commented):
What happens if the directory already exists but isn't mounted?  I guess my preference to solve this problem would be to change the error handling path instead.  If `fu_volume_new_by_kind` returns an error because udisks isn't around and that path exists then do all the other code later in the function (which eventually also calls `fu_volume_new_from_mount_path`).  Otherwise propagate the error.

nicholasbishop:
> What happens if the directory already exists but isn't mounted? 

Perhaps I'm misunderstanding, but I thought the point of `EspLocation` is to manage mounting the ESP outside of fwupd -- it's up to the user to make sure they pass in a valid path. I don't think the rest of the code there can check that it's a valid ESP mount without UDisks anyway, or am I missing something?

superm1:
It's dual purpose - it could also be used to disambiguate among a system with multiple ESP's, which one to use.  Udisks looks at `/etc/fstab` to identify the mount location.

So for example if I have 2 ESP's on my system and the my ESP I want to use is  specified at `/boot/efi` in `/etc/fstab` and also in `/etc/fwupd/daemon.conf` but not mounted before running a firmware update the code today will try to mount `/boot/efi`.  After your change I expect that this fails because it's no longer attempted to be mounted.

nicholasbishop:
Thanks, that makes sense.

Curiously, I can no longer reproduce the problem I was trying to solve. I'm not running udisks, and previously I was seeing fwupd spend a long time waiting for the dbus response, which would time out. But now for some reason this delay is no longer happening, I set `EspLocation` and it works fine.

I'm quite confused as to what changed in my test environment, but for now I'll go ahead and close this.

Thanks again for your help!

PR #5799, opened 2023-05-09
Add the latest dbx fixups for BlackLotus

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
I still thing it's better to come up with a way to distribute this info from LVFS, but this makes sense with what exists today.

hughsie:
My thoughts exactly. I guess we want something like we did inheriting flags from the metadata, but I'm drowning in MSI right now.

Issue #5817, opened 2023-05-14
Flatpak packaging bug: Incorrect permissions on read-only mounted /app/etc/fwupd/fwupd.conf prevents running fwupdtool


mihailim:
**Describe the bug**

The `org.freedesktop.fwupd` flatpak package provides `/app/etc/fwupd/fwupd.conf` with mode `0644`. When run, `fwupdtool` attempts to change it to `0640` (`FU_CONFIG_FILE_MODE_SECURE`) as per [this code section](https://github.com/fwupd/fwupd/blob/218037263454dac4892c6bec49fff58de192b014/libfwupdplugin/fu-config.c#L104), however that fails inside the flaptak sandbox since `/app` is obviously read-only.

The flatpak-ed fwupdtool used to work perfectly before the latest update.

**Steps to Reproduce**

```sh
root@host:~# flatpak install flathub org.freedesktop.fwupd
<snip>
root@host:~# flatpak run org.freedesktop.fwupd --version
Loading…                 [ -                                     ]
Failed to load config: failed to change permission of /app/etc/fwupd/fwupd.conf
```

And, indeed, checking inside the sandbox it looks like this:

```sh
root@host:~# flatpak run --command=sh org.freedesktop.fwupd
[📦 org.freedesktop.fwupd ~]# ls -la /app/etc/fwupd/fwupd.conf
-rw-r--r-- 2 root root 119 ian  1  1970 /app/etc/fwupd/fwupd.conf
[📦 org.freedesktop.fwupd ~]# /app/bin/fwupdtool --version
Loading…                 [ -                                     ]
Failed to load config: failed to change permission of /app/etc/fwupd/fwupd.conf
```

**Expected behavior**

fwupdtool should run.

**fwupd version information**

`org.freedesktop.fwupd` version 1.9.1 at flatpak commit `964c2ffa191e29` installed from Flathub as per the above.

It's not possible to provide the details asked for in the issue template, since fwupdtool will no longer run anything, including `--version`. Fortunately, the system details aren't really pertinent to the issue.

- Operating system and version: Debian bullseye, however it's immaterial.
- Have you tried rebooting?: Yes.
- Is this a regression?: Yes, used to work in 1.8.x.

hughsie:
Weird, and somewhat unexpected -- meson should be installing:

	  install_data(['fwupd.conf'],
	    install_dir: join_paths(sysconfdir, 'fwupd'),
	    install_mode: 'rw-r-----',
	  )

[...]

PR #5819, opened 2023-05-14
Ignore the immutable flatpak fwupd.conf file permissions

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/5817

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-config.c
+++ b/libfwupdplugin/fu-config.c
@@ -26,9 +26,15 @@ static guint signals[SIGNAL_LAST] = {0};
 #define FU_CONFIG_FILE_MODE_SECURE 0640
 
 typedef struct {
+	gchar *filename;
+	GFile *file;
+	GFileMonitor *monitor; /* nullable */
+	gboolean is_writable;
+} FuConfigItem;
+
+typedef struct {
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-config.c
```diff
@@ -104,6 +104,8 @@ fu_config_reload(FuConfig *self, GError **error)
 	/* ensure the config files are set to the correct permissions */
 	for (guint i = 0; i < priv->filenames->len; i++) {
 		const gchar *fn = g_ptr_array_index(priv->filenames, i);
+		if (g_str_has_prefix(fn, "/app"))
```
superm1, line 107: We probably have the same problem for snap too? Maybe we should use environment variables instead.

hughsie: Can we ask the FS "is this immutable"? Is the chatter bit just set?

hughsie: Seems to work, will respin tomorrow.

hughsie: Done, all fixed up.

superm1 (approved):

PR #5823, opened 2023-05-15
intel-gsc: update official Intel product names

vlubart:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [x] Documentation

### DIFF
--- a/plugins/intel-gsc/igsc.quirk
+++ b/plugins/intel-gsc/igsc.quirk
@@ -61,22 +61,22 @@ Flags = has-aux,has-oprom
 Name = DG2_SOC1
 Flags = has-aux,has-oprom
 [MEI\VEN_8086&DEV_5690]
-Name = DG2_SOC1
+Name = Arc A770M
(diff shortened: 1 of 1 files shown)


plugins/intel-gsc/igsc.quirk
```diff
@@ -61,22 +61,22 @@ Flags = has-aux,has-oprom
 Name = DG2_SOC1
 Flags = has-aux,has-oprom
 [MEI\VEN_8086&DEV_5690]
-Name = DG2_SOC1
+Name = Intel(R) Arc(TM) A770M Graphics
```
hughsie, line 64: No, Intel is the *vendor* name, not the device name. If you do this, you'll have "Intel Intel(R)" in the UI. Also no "(TM)" please, it's not required; all vendor and device names are assumed to be trademarks. You also really don't need Graphics too; I'd say this entire line should just be `Arc A77M`.

hughsie (approved):

Issue #5802, opened 2023-05-10
Candidate snap shows `WARNING: This package has not been validated, it may not work properly.`


superm1:
When running `fwupdmgr` from the candidate snap (`1.8.13-5-g33cc09899`) the following is displayed:
```
WARNING: This package has not been validated, it may not work properly.
```

This is unexpected as it's built from a tagged release.
```
$ git describe 33cc09899
1.8.14
```

Tagged releases are _supposed_ to [not show that message](https://github.com/fwupd/fwupd/blob/main/meson.build#L484).

The way this all is supposed to work is:
1. A release is tagged and this causes a GitHub `published` event: https://github.com/fwupd/fwupd/blob/main/.github/workflows/snap-candidate.yml
2. This causes snap workflow to be called with `deploy` to `true`, and `channel` set to `candidate`
3. The job checks out with `fetch-depth: 0`, so all tags should be checked out.

So I expect it all works.  The issue might have been related to branching and not changing https://github.com/fwupd/fwupd/commit/e3f18203d395a1e3e785220ce018740eaf9060ba when branching.

Once 1.9.0 releases, we should debug this if it doesn't work again.

superm1:
Here is the release for 1.9.1 https://github.com/fwupd/fwupd/actions/runs/4966294117/jobs/8887649235

It happened again. Can see it checked out on the tag but it's not referring to the tag ref in the snap job properly.

superm1:
OK to make matters even weirder, there was another job that triggered that got the tagging right!
https://github.com/fwupd/fwupd/actions/runs/4966291986

This wasn't on the same releasing event though, it was just a regular CI job.

hughsie:
Is this an artefact of how I did the release? I'm a bit confused what I need to do :)

superm1:
I'm not sure right now what the root cause actually is.  I thought it was a nuance of timing that the ref didn't exist from internal Github stuff , but this doesn't seem to be the case.

This is the one with the 1.9.1 ref, triggered by a push of a commit: https://github.com/fwupd/fwupd/actions/runs/4966291986
Notice:
```
Checking out the ref
  /usr/bin/git checkout --progress --force -B main refs/remotes/origin/main
  Switched to a new branch 'main'
  branch 'main' set up to track 'origin/main'.
```
And then later it's got 1.9.1.
```
:: + craftctl default
:: ++ git describe HEAD --always
:: + craftctl set version=1.9.1
```
[...]

PR #5826, opened 2023-05-15
trivial: snap: use `git describe --exact-match` to decide promotions

superm1:
Rather than running on a release event, just examine every build job whether it should go to candidate channel or not.

I think this would fix https://github.com/fwupd/fwupd/issues/5802, but it also means that if a job happens when the tags "aren't yet pushed" it won't promote properly.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
It it works, LGTM, thanks.

PR #5831, opened 2023-05-18
Wip/hughsie/goodix tp

xulinkun:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/goodix-tp/fu-goodixtp-brlb-firmware.c
+++ b/plugins/goodix-tp/fu-goodixtp-brlb-firmware.c
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2023 Goodix.inc <xulinkun@goodix.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#include "fu-goodixtp-brlb-firmware.h"
+#include "fu-goodixtp-common.h"
+#include "fu-goodixtp-struct.h"
+
+struct _FuGoodixtpBrlbFirmware {
+	FuGoodixtpFirmware parent_instance;
+};
+
+G_DEFINE_TYPE(FuGoodixtpBrlbFirmware, fu_goodixtp_brlb_firmware, FU_TYPE_GOODIXTP_FIRMWARE)
+
+#define FW_HEADER_SIZE 512
+
+gboolean
+fu_goodixtp_brlb_firmware_parse(FuGoodixtpFirmware *self,
+				GBytes *fw,
+				guint8 sensor_id,
+				GError **error)
+{
+	guint32 version;
+	gsize offset_hdr;
+	gsize offset_payload = FW_HEADER_SIZE;
+	guint32 checksum = 0;
+	guint32 firmware_size;
+	guint8 subsys_num;
+	guint8 cfg_ver = 0;
+	gsize bufsz = 0;
+	const guint8 *buf = g_bytes_get_data(fw, &bufsz);
+	g_autoptr(GByteArray) st = NULL;
+
+	st = fu_struct_goodix_brlb_hdr_parse(buf, bufsz, 0x0, error);
+	if (st == NULL)
+		return FALSE;
+	firmware_size = fu_struct_goodix_brlb_hdr_get_firmware_size(st);
+	firmware_size += 8;
+
+	/* [payload][64 bytes padding?][config] */
+	if ((gsize)firmware_size < bufsz) {
+		g_autoptr(GBytes) fw_img = NULL;
+		g_autoptr(FuFirmware) img = fu_firmware_new();
+		fu_firmware_set_idx(img, 4);
+		fu_firmware_set_addr(img, 0x40000);
+		fw_img = fu_bytes_new_offset(fw,
+					     firmware_size + 64,
+					     bufsz - (gsize)firmware_size - 64,
+					     error);
+		if (fw_img == NULL)
+			return FALSE;
+		fu_firmware_set_bytes(img, fw_img);
+		fu_firmware_add_image(FU_FIRMWARE(self), img);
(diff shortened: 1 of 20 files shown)


xulinkun:
Hi, @hughsie ,here I go again... please review.

plugins/goodix-tp/fu-goodixtp-firmware.c
```diff
@@ -0,0 +1,108 @@
...
+			   guint8 type,
+			   guint32 addr,
+			   const guint8 *buf,
+			   gint bufsz)
+{
+	FuGoodixtpFirmwarePrivate *priv = GET_PRIVATE(self);
+	if (bufsz > RAM_BUFFER_SIZE)
+		return;
+
+	memcpy(priv->fw_data + priv->fw_len, buf, bufsz);
```
hughsie, line 79: I'd rather a fu_memcpy_safe() here

plugins/goodix-tp/fu-goodixtp-brlb-firmware.c
```diff
@@ -0,0 +1,146 @@
...
+		if (tmp_type == 0x0B || tmp_type == 0x01) {
+			g_debug("skip type[%x] subsystem", tmp_type);
+			continue;
+		}
+
+		for (gint j = 0; j < (gint)(tmp_size / RAM_BUFFER_SIZE); j++) {
+			fu_goodixtp_add_chunk_data(self,
+						   tmp_type,
+						   tmp_addr + offset,
+						   tmp_data + offset,
```
hughsie, line 111: this looks a bit risky; i think you need to check fw_offset+offset to see if it's in-range.

plugins/goodix-tp/fu-goodixtp-firmware.c
```diff
@@ -0,0 +1,108 @@
...
+#include "config.h"
+
+#include <fwupdplugin.h>
+
+#include "fu-goodixtp-common.h"
+#include "fu-goodixtp-firmware.h"
+
+#define MAX_CHUNK_NUM 80
+
+struct FuGoodixChunkInfo {
```
hughsie, line 16: typedef struct pelase

plugins/goodix-tp/fu-goodixtp-firmware.c
```diff
@@ -0,0 +1,108 @@
...
+	guint32 flash_addr;
+};
+
+typedef struct {
+	FuFirmwareClass parent_instance;
+	guint32 version;
+	guint8 *fw_data;
+	gint fw_len;
+	gint index;
+	struct FuGoodixChunkInfo chunk_info[MAX_CHUNK_NUM];
```
hughsie, line 27: this seems like a big thing to put on the stack; can you make this a GPtrArray please.

hughsie:
@xulinkun i've pushed another fixup patch; can you verify i've not broken anything please. There are a few more things I think we need to fix -- and then I think we're nearly complete.

hughsie:
PVS Studio also picked up: 

    General Analysis	fu-goodixtp-hid-device.c:96	Medium	V1086	A call of the 'memcpy' function will lead to underflow of the buffer 'priv->patch_pid'.

xulinkun:
I see, I'll submit again tomorrow. Google will use the fwupd instead of their own way on Chrome OS, so we need to hurry up.

hughsie:
> Google will use the fwupd instead of their own way on Chrome OS

Yes, I hear the upstream maintainer is really strict about code quality :)

xulinkun:
Hi, @hughsie , I've optimized goodix_firmware.c, please review it.

plugins/goodix-tp/fu-goodixtp-hid-device.c
```diff
@@ -0,0 +1,254 @@
...
+#include <linux/input.h>
+#endif
+
+#include "fu-goodixtp-common.h"
+#include "fu-goodixtp-firmware.h"
+#include "fu-goodixtp-hid-device.h"
+
+typedef struct {
+	guint8 patch_pid[9];
+	guint8 patch_vid[4];
```
hughsie, line 22: Question: what do we actually need the patch_pid and patch_vid _for_?

xulinkun: Yes, patch_pid storage chip name such as GT7863, GT9966.... and vid storage firmware version. We need to print them.

hughsie:
@xulinkun coverity was really upset with the meory safety in FuGoodixtpBrlbFirmware -- specifically the ownership of the data of the FuGoodixtpChunkInfo -- and I had very little idea of what the code was actually doing. I rewrote it using the standard primitives we have in fwupd -- which means the right thing should happen if you do `sudo fwupdtool firmware-parse gtx8-firmware.bin`

Can you check what I've done please, and verify the code still works on both hardware types. Feel free to add the missing data in the .rs file if that's helpful to you.

plugins/goodix-tp/fu-goodixtp.rs
```diff
@@ -0,0 +1,48 @@
...
+    Berlinb,
+    Yellowstone,
+}
+
+#[derive(Parse)]
+struct GoodixBrlbHdr {
+    firmware_size: u32le,
+    checksum: u32le,
+    _unknown: [u8; 19],
+    vid: u16be,
```
hughsie, line 19: @xulinkun is this definitely big endian? I thought all of brlb was LE?

xulinkun:
Hi,@hughsie, thanks for your help. Writing flash must to align with 4K, I changed a little bit base your code.

plugins/goodix-tp/fu-goodixtp.rs
xulinkun: En... except vid is BE, the others is LE in brlb.

hughsie (approved):
LGTM.

PR #5834, opened 2023-05-19
WIP ccgx: Implement pure HID update for Framework HDMI/DP cards

JohnAZoidberg:
Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

---

Status:

- [x] Get version, bootloader version
  - [ ] Can we display versions of both images?
- [x] Write one firmware image
- [ ] Automatically reset and write the other image as well
- [x] Get Silicon ID
- [ ] Does `set_quirk_kv` work? What does it do?
- [ ] Check if need to enable in CI
- [ ] Decode version information from FW binary
- [ ] Figure out binary format. Currently just a raw dump. See if I can wrap it like other CCGX images
- [ ] See what documentation I have to add
- [ ] Make sure code follows style guide

Interaction with the device looks like this:

- Get Version
  1. Set "Magic Unlock" Feature Report
  2. Set "Enable I2C bridge mode" Report
  3. Get "Info" Feature Report
- Flash Firmware
  1. Get Version (see above)
  2. Depending on which image is active, flash the other
  3. Get Version again (see above, probably not required)
  4. Write each firmware row to the offset of the selected image (Set Report)
  5. Write metadata firmware row to the offset of the selected image (Set Report)
  6. Send some sort of switch command param with opposite image index (Set Report)
  7. Send reset command (Set Report)
  8. Wait 5s for reset and repeat

Currently you need to run fwuptool twice in order to update both images.
I haven't been able to make the device reset into the opposite image without flashing.
I assume that's disabled in the SDK.

Here's what firmware information looks like:

```
> ninja -C build && sudo ./build/src/fwupdtool --plugins ccgx get-devices
ninja: Entering directory `build'
ninja: no work to do.
Loading…                 [************************************** ]
Framework Laptop (12th Gen Intel Core)
│
└─DisplayPort Expansion Card:
      Device ID:          083dd03eb720e673d23eee0bb9293ea1cf990996
      Current version:    3.0.17.101
      Bootloader Version: 3.0.2.900
      Vendor:             Framework (USB:0x32AC)
      Serial Number:      11AD1D00C9D73E0223190B00
      GUIDs:              7e137108-1fe0-5820-b962-6603fa7b2070 ← USB\VID_32AC&PID_0003
                          a671c02b-adb7-5f2d-b7b6-818cb56c9deb ← USB\VID_32AC&PID_0003&REV_0000
      Device Flags:       • Updatable
                          • Device stages updates
                          • Unsigned Payload
```

This is a snippet of the verbose output:

```
01:20:31.778 FuPluginCcgx         Report ID:       0xe0
01:20:31.778 FuPluginCcgx         Signature:       0x5943
01:20:31.778 FuPluginCcgx         Operating Mode:  fw1 (0x01)
01:20:31.778 FuPluginCcgx         Bootloader Info: 0x00
01:20:31.778 FuPluginCcgx         Bootmode Reason: 0x00
01:20:31.778 FuPluginCcgx         Silicon ID:      0x00001d00
01:20:31.778 FuPluginCcgx         BL Version:      3.0.2 Build 900 (0x30020384)
01:20:31.778 FuPluginCcgx         Image 1 Ver:     3.0.17 Build 101 (0x30110065)
01:20:31.778 FuPluginCcgx         Image 2 Ver:     3.0.17 Build 101 (0x30110065)
01:20:31.778 FuPluginCcgx         Image 1 Row:     0x00001800
01:20:31.778 FuPluginCcgx         Image 2 Row:     0x00010000
01:20:31.778 FuPluginCcgx         Device UID:      C9D73E022319
```

Btw. How can I make just a single plugin do verbose output?
`--plugin-verbose=ccgx` doesn't work with `fwupdtool` and `export FWUPD_CCGX_VERBOSE=1`  doesn't either.

Firmware update works but it'll only install one image at a time.
```
> ninja -C build && sudo ./build/src/fwupdtool --plugins ccgx install-blob dump101-combined.bin b4f18862d2175a5d490a0fb7427a19d43ecbb913
```

I can't quite figure out how to make it wait for device reset and then attempt another firmware update.
I don't quite understand the setup/attach/detach/... phases.
Why does `fu-ccgx-hpi-device.c` issue a reset in detach AND in attach?

Another question: Is it expected that the device GUID changes after firmware update (or at least when switching to the other image?)


JohnAZoidberg:
I'm in discussion with @hughsie about implementing this. He's got some cards to test as well.

hughsie:
> I can't quite figure out how to make it wait for device reset and then attempt another firmware update.

Question: Do we want to? I thought the idea was there was an A/B partition -- so if you're running A, you update B and switch to B -- and then if B fails to "boot" it falls back to the known-working A. is there a reason why you want to flash both partitions?

plugins/ccgx/fu-ccgx-plugin.c
```diff
@@ -36,6 +37,7 @@ fu_ccgx_plugin_constructed(GObject *obj)
 	fu_plugin_add_firmware_gtype(plugin, NULL, FU_TYPE_CCGX_FIRMWARE);
 	fu_plugin_add_firmware_gtype(plugin, NULL, FU_TYPE_CCGX_DMC_FIRMWARE);
 	fu_plugin_add_device_gtype(plugin, FU_TYPE_CCGX_HID_DEVICE);
+	fu_plugin_add_device_gtype(plugin, FU_TYPE_CCGX_PURE_HID_DEVICE);
```
hughsie, line 40: what's Pure in reference to? Is there a better name?

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+	//guint32 image_1_row;
+	//guint32 image_2_row;
+	//device_uid: // [u8; 6],
+	guint32 flash_row_size;
+	guint32 flash_size;
+};
+G_DEFINE_TYPE(FuCcgxPureHidDevice, fu_ccgx_pure_hid_device, FU_TYPE_HID_DEVICE)
+
+
+typedef struct __attribute__((packed)) {
```
hughsie, line 48: this looks perfect for the rustgen stuff in https://github.com/fwupd/fwupd/blob/main/docs/tutorial.md#the-rustgen-helper -- it'll make the code a lot simpler too.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+				      FU_CCGX_PURE_HID_DEVICE_TIMEOUT,
+				      FU_HID_DEVICE_FLAG_IS_FEATURE,
+				      error)) {
+		g_prefix_error(error, "magic enable command error: ");
+		return FALSE;
+	}
+
+
+	g_debug("bridge mode");
+	err = fu_ccgx_pure_hid_command(device, CCGX_HID_CMD_MODE, CY_PD_BRIDGE_MODE_CMD_SIG, NULL);
```
hughsie, line 126: just do a g_debug() with a g_autoptr(GError) error_local -- it might be useful for debugging.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+		return FALSE;
+	}
+
+	g_debug("Operating Mode:  0x%02x", self->operating_mode);
+	switch (self->operating_mode) {
+		case FU_CCGX_FW_MODE_BOOT:
+		case FU_CCGX_FW_MODE_FW2:
+			// Update Image 1
+			g_debug("Flashing Image 1");
+			fu_ccgx_pure_hid_flash_firmware_image(device, fw1_binary, fw_size, FW1_START, FW1_METADATA, 1, error);
```
hughsie, line 409: if you're using the GError passed in you need to return FALSE on failure

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+struct _FuCcgxPureHidDevice {
+	FuHidDevice parent_instance;
+	FuCcgxFwMode operating_mode;
+	guint32 versions[FU_CCGX_FW_MODE_LAST];
+	//guint8 bootloader_info;
+	//guint8 bootmode_reason;
+	guint32 silicon_id;
+	//guint32 image_1_row;
+	//guint32 image_2_row;
+	//device_uid: // [u8; 6],
```
superm1, line 41: why are these commented out?  Trying to decide whether to make them properties of the device or not?

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+#define FW1_METADATA 0x03FF
+#define FW2_METADATA 0x03FE
+#define WRITE_HEADER_SIZE 4
+
+/**
+ * FU_CCGX_PURE_HID_DEVICE_IS_IN_RESTART:
+ *
+ * Device is in restart and should not be closed manually.
+ *
+ * Since: 1.7.0
```
superm1, line 28: 1.9.2

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+
+static gboolean
+fu_ccgx_pure_hid_flashing_mode(FuDevice *device, gpointer user_data, GError **error)
+{
+	g_debug("before flashing mode");
+	if (!fu_ccgx_pure_hid_command(device, CCGX_HID_CMD_FLASH, CY_PD_ENTER_FLASHING_MODE_CMD_SIG, error)) {
+		g_prefix_error(error, "flashing enable command error: ");
+		return FALSE;
+	}
+	g_debug("After flashing mode");
```
superm1, line 102: this and similar comments are probably needlessly noisy once you've got it all working

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+	g_debug("Silicon ID:      0x%08x", info->silicon_id);
+	bl_ver = fu_version_from_uint32(info->bl_version, fu_device_get_version_format(self));
+	g_debug("BL Version:      %s (0x%08x)", bl_ver, info->bl_version);
+	ver1 = fu_version_from_uint32(info->image_1_ver, fu_device_get_version_format(self));
+	g_debug("Image 1 Ver:     %s (0x%08x)", ver1, info->image_1_ver);
+	ver2 = fu_version_from_uint32(info->image_2_ver, fu_device_get_version_format(self));
+	g_debug("Image 2 Ver:     %s (0x%08x)", ver2, info->image_2_ver);
+	g_debug("Image 1 Row:     0x%08x", info->image_1_row);
+	g_debug("Image 2 Row:     0x%08x", info->image_2_row);
+	g_debug("Device UID:      %02X%02X%02X%02X%02X%02X", info->device_uid[0], info->device_uid[1], info->device_uid[2], info->device_uid[3], info->device_uid[4], info->device_uid[5]);
```
superm1, line 178: this is begging to be moved into a `to_string` function

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+
+static void
+fu_ccgx_pure_hid_device_init(FuCcgxPureHidDevice *self)
+{
+	fu_device_add_protocol(FU_DEVICE(self), "com.cypress.ccgx");
+	fu_device_add_protocol(FU_DEVICE(self), "com.infineon.ccgx");
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_DUAL_IMAGE);
+	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_INTEL_ME2);
```
superm1, line 464: is this the right version format?

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_RESTART, 25, "attach");
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_BUSY, 50, "reload");
+}
+
+static void
+fu_ccgx_pure_hid_device_init(FuCcgxPureHidDevice *self)
+{
+	fu_device_add_protocol(FU_DEVICE(self), "com.cypress.ccgx");
+	fu_device_add_protocol(FU_DEVICE(self), "com.infineon.ccgx");
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);
```
superm1, line 461: Are they payloads really unsigned?  :(

plugins/ccgx/ccgx.quirk
```diff
@@ -1,3 +1,13 @@
+# Framework HDMI Expansion Card
+[USB\VID_32AC&PID_0002]
+Plugin = ccgx
+GType = FuCcgxPureHidDevice
+
+# Framework DisplayPort Expansion Card
+[USB\VID_32AC&PID_0003]
+Plugin = ccgx
+GType = FuCcgxPureHidDevice
```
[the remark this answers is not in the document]
superm1: Oh but in your PR message I saw the sample output `└─DisplayPort Expansion Card:`, so I guess the core stuff in this plugin is already getting it for you.  You can ignore this comment then.

superm1:
### Thanks for your commit

**Clang Format**
 
**Pre-commit overview**

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+static gboolean
+fu_ccgx_pure_hid_get_fw_info(FuDevice *device, gpointer user_data, GError **error)
+{
+	FuCcgxPureHidDevice *self = FU_CCGX_PURE_HID_DEVICE(device);
+	gsize bufsz = 0x40;
+	g_autofree guint8 *buf = g_malloc0(bufsz);
+	HidFwInfo *info;
+	g_autofree gchar *bl_ver;
+	g_autofree gchar *ver1;
+	g_autofree gchar *ver2;
```
superm1, line 142: `g_autofree gchar*` needs to be initialized to `NULL`

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+
+	return TRUE;
+}
+
+static gboolean
+fu_ccgx_pure_hid_get_fw_info(FuDevice *device, gpointer user_data, GError **error)
+{
+	FuCcgxPureHidDevice *self = FU_CCGX_PURE_HID_DEVICE(device);
+	gsize bufsz = 0x40;
+	g_autofree guint8 *buf = g_malloc0(bufsz);
```
superm1, line 138: as it's a fixed size, does it really need dynamic memory allocation?  Just put it on the stack.
I would think you can just do something like this:

```
guint8 buf[bufsz] = { 0 };
```

superm1:
> Another question: Is it expected that the device GUID changes after firmware update (or at least when switching to the other image?)

I wouldn't think so.  What GUID was it before and what was it after?

superm1:
> > I can't quite figure out how to make it wait for device reset and then attempt another firmware update.
> 
> Question: Do we want to? I thought the idea was there was an A/B partition -- so if you're running A, you update B and switch to B -- and then if B fails to "boot" it falls back to the known-working A. is there a reason why you want to flash both partitions?

I completely agree FWIW.  This could also give you an out if there is some horribly fatal problem with the new update and you need to revert back to the old update.  It should be a simple command to the device to switch partitions on the next boot.

JohnAZoidberg:
> Question: Do we want to? I thought the idea was there was an A/B partition -- so if you're running A, you update B and switch to B -- and then if B fails to "boot" it falls back to the known-working A. is there a reason why you want to flash both partitions?

Yes, exactly. When running A, you update B and switch to B. Only if B succeeds in booting you can update A. If B doesn't succeed booting, we'll boot back into A and can only try updating B again.

Developing this plugin I actually have flashed broken firmware quite a few times, resulting in the version being reported as 0x00000000 and the image being marked as invalid. Or I've flashed HDMI firmware on the DP card accidentally.
Only one image was ever broken because I can't update the currently running image. And I think bootloader mode is disabled in our device.

plugins/ccgx/fu-ccgx-plugin.c
JohnAZoidberg: As in it uses HID reports for everything. Unlike the FU_TYPE_CCGX_HID_DEVICE which uses HID only to switch to another mode.
Not sure of a better name.

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: That looks awesome, I'll try that.

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: Yep. Probably won't keep them. Maybe device_uid is worth putting here? Will these feed into the GUID?

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: Yes, will remove them.

plugins/ccgx/fu-ccgx-pure-hid-device.c
superm1: Can you use uid to better differentiate what firmware goes with what device? If so then you can probably parse the header and match it to ensure no one ever flashes a DP firmware to the HDMI for example.

plugins/ccgx/fu-ccgx-plugin.c
superm1: How about native hid?

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: Right, thanks!

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: No, UID is different for each device. We can avoid flashing the wrong firmware just by looking at the PID.
I mean...once you flash the wrong one, the PID will change. But that shouldn't happen, I think

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: Yes

plugins/ccgx/ccgx.quirk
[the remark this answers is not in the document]
JohnAZoidberg: Yes, the device name is set properly.
I noticed on Windows it's not, some driver seems to be hijacking the name.

JohnAZoidberg:
> The fwupd project has had a style change starting with fwupd 1.6.x. The style is summarized in main/CONTRIBUTING.md and it appears your current pull request is inconsistent with the new style.

Thanks. I did in fact base my commit on an older version of the main branch.

>> Question: Do we want to? I thought the idea was there was an A/B partition -- so if you're running A, you update B and switch to B -- and then if B fails to "boot" it falls back to the known-working A. is there a reason why you want to flash both partitions?

> Yes, exactly. When running A, you update B and switch to B. Only if B succeeds in booting you can update A. If B doesn't succeed booting, we'll boot back into A and can only try updating B again.

What I'm trying to say is that it's fine updating both images because you can't brick it.

hughsie:
> What I'm trying to say is that it's fine updating both images because you can't brick it.

But what's the benefit of updating the currently running image? I think traditionally we've only written when we needed to. i.e.:

    A=1.2.3 (running)
    B=1.2.0

A then updates B to 1.2.4 and switches to B, then we have:

    A=1.2.3
    B=1.2.4 (running)

..why do we need to update A then?

plugins/ccgx/fu-ccgx-pure-hid-device.c
hughsie: > I've made a script to generate the cyacd file like fwupd expects it:

I don't think you should modify the firmware just for fwupd; I think the plugin needs to adapt - in this case just using the BaseVersion for display, not the AppVersion. e.g. have a private flag that is called something like `prefer-base-version` or something.

I can play with this code tomorrow; sorry for nor playing with your branch already -- I'm juggling vendors like crazy.

plugins/ccgx/fu-ccgx-plugin.c
hughsie: I guess we want to know if it's specific to anything. e.g. is this Bizlink specific? would FU_TYPE_CCGX_BIZLINK_DEVICE be more representative?

plugins/ccgx/fu-ccgx-pure-hid-device.c
hughsie: You can do; but lets see what we can do by modifying fu-ccgx-firmware.c: What offset is base-version? app-version is 228 if that helps.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,538 @@
...
+
+			// TODO: Wait 5s
+			// TODO: Refresh HID devices
+			//fu_ccgx_pure_hid_magic_unlock(device, user_data, error);
+			// Update Image 1
+			//flash_firmware_image(device, fw1_binary, fw_size, FW1_START, FW1_METADATA, 1);
+			break;
+		default:
+			// Invalid value
+			g_assert(self->operating_mode < FU_CCGX_FW_MODE_LAST);
```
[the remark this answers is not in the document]
hughsie: I've removed this.

plugins/ccgx/ccgx.quirk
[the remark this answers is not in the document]
hughsie: I get:

	└─HDMI Expansion Card:
	      Device ID:          a0df02b6c3f24d131fd352868c89b4d6ea3bf9e4
	      Current version:    48.16.0.5
	      Bootloader Version: 48.2.3.132
	      Vendor:             Framework (USB:0x32AC)

plugins/ccgx/fu-ccgx-pure-hid-device.c
hughsie: Also, can you upload the HDMI-6-1.cyacd file somewhere (or send me an email please)

plugins/ccgx/fu-ccgx-pure-hid-device.c
hughsie: I've removed then now I've shifted to rustgen.

plugins/ccgx/fu-ccgx-pure-hid-device.c
hughsie: I've pushed this to wip/framework-hdmi-dp -- feel free to push there yourself too -- I'll add you on the commit ACL now.

plugins/ccgx/fu-ccgx-plugin.c
hughsie: Yup, I think we definitely need a better name now. This looks ODM specific -- so ideas very welcome.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,559 @@
...
+
+	fw_bytes = g_bytes_get_data(fw, NULL);
+	fw_total_size = g_bytes_get_size(fw);
+	fw_size = fw_total_size / 2;
+
+	// TODO: Do this in prepare_firmware
+	// Make sure there are two firmware images composed of full rows
+	g_assert(fw_total_size % 2 * ROW_SIZE == 0);
+	fw1_binary = fw_bytes;
+	fw2_binary = &fw_bytes[fw_size];
```
hughsie, line 401: is this how you got the firmware from the ODM, or did you manually "stick together" two files into one binary payload?

plugins/ccgx/fu-ccgx-plugin.c
hughsie: I'm actually wondering if this belongs in a new directory too. It doesn't seem to use much from the CCGX plugin at all!

JohnAZoidberg: I'll be using the ccgx firmware parsing. And some of the constants.

> Yup, I think we definitely need a better name now. This looks ODM specific -- so ideas very welcome.

The HID protocol is actually included in the cypress/infineon CCG3 SDK, though slightly different.

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: That's awesome!

JohnAZoidberg:
> ..why do we need to update A then?

Ok yes, I've done some testing and it seems the device never swaps firmwares by itself.
It'll always boot from the one that was most recently flashed.
So we can do that.

plugins/ccgx/fu-ccgx-pure-hid-device.c
JohnAZoidberg: I manually concatenated them. Will switch to CYACD format.

hughsie: Cool. This is why each bank has it's own guid btw; so you can match the correct firmware for each bank in the metainfo file.

plugins/ccgx/fu-ccgx-plugin.c
hughsie: Maybe VendorHid instead then? Or NativeHid?

JohnAZoidberg: I renamed it to NativeHid

JohnAZoidberg:
Closing in favor of the new PR.

PR #5843, opened 2023-05-22
Increase the GLib requirement to 2.68.0

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
@superm1 do you understand the snap failure? Thanks.

superm1:
My suspicion is the the snap environment doesn't have polkit 0.114 which became a harder requirement with this set of changes.

Ubuntu 22.04 (which I believe is what the core snap is built from) has [0.105](https://launchpad.net/ubuntu/+source/policykit-1).

So I'd say you have to back out the polkit hunk that bumps the minimum requirement.

hughsie:
@superm1 aha, got it, thanks.

superm1 (approved):

Issue #5830, opened 2023-05-17
Capture used space in efivarfs


superm1:
A kernel patch that has been queued up for kernel 6.5 exposes the used and total size of efivarfs.  We should capture this as part of the the report for UEFI capsules and also use this in fwupd to identify some failure scenarios.

https://lkml.org/lkml/2023/5/17/957

hughsie:
Just for the metadata report, or to check *before* to do things like dbx and capsule updates?

superm1:
I think metadata report initially, we need to see how well vendors report this information.

PR #5850, opened 2023-05-22
uefi-dbx: Record the EFI filesystem used in metadata reports

hughsie:
Fixes: https://github.com/fwupd/fwupd/issues/5830

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-efivar-linux.c
+++ b/libfwupdplugin/fu-efivar-linux.c
@@ -329,6 +329,23 @@ fu_efivar_space_used_impl(GError **error)
 	guint64 total = 0;
 	g_autoptr(GDir) dir = NULL;
 	g_autofree gchar *path = fu_efivar_get_path();
+	g_autoptr(GFile) file_fs = g_file_new_for_path(path);
+	g_autoptr(GFileInfo) info_fs = NULL;
+	g_autoptr(GError) error_local = NULL;
+
+	/* this is only supported in new kernels */
+	info_fs = g_file_query_info(file_fs,
+				    G_FILE_ATTRIBUTE_FILESYSTEM_USED,
+				    G_FILE_QUERY_INFO_NONE,
+				    NULL,
+				    &error_local);
+	if (info_fs == NULL) {
+		g_debug("failed to get efivar used space: %s", error_local->message);
+	} else {
+		total = g_file_info_get_attribute_uint64(info_fs, G_FILE_ATTRIBUTE_FILESYSTEM_USED);
+		if (total > 0)
+			return total;
+	}
 
 	/* stat each file */
 	dir = g_dir_open(path, 0, error);
--- a/plugins/uefi-dbx/fu-uefi-dbx-device.c
+++ b/plugins/uefi-dbx/fu-uefi-dbx-device.c
@@ -143,6 +143,17 @@ fu_uefi_dbx_device_probe(FuDevice *device, GError **error)
 }
 
 static void
+fu_uefi_dbx_device_report_metadata_pre(FuDevice *device, GHashTable *metadata)
+{
+	guint64 nvram_total = fu_efivar_space_used(NULL);
+	if (nvram_total != G_MAXUINT64) {
+		g_hash_table_insert(metadata,
+				    g_strdup("EfivarNvramUsed"),
+				    g_strdup_printf("%" G_GUINT64_FORMAT, nvram_total));
+	}
+}
+
+static void
 fu_uefi_dbx_device_set_progress(FuDevice *self, FuProgress *progress)
 {
 	fu_progress_set_id(progress, G_STRLOC);
@@ -188,6 +199,7 @@ fu_uefi_dbx_device_class_init(FuUefiDbxDeviceClass *klass)
 	klass_device->write_firmware = fu_uefi_dbx_device_write_firmware;
 	klass_device->prepare_firmware = fu_uefi_dbx_prepare_firmware;
 	klass_device->set_progress = fu_uefi_dbx_device_set_progress;
+	klass_device->report_metadata_pre = fu_uefi_dbx_device_report_metadata_pre;
 }
 
 FuUefiDbxDevice *


hughsie:
100.0% untested with that kernel patch.

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -144,6 +144,31 @@ fu_uefi_dbx_device_probe(FuDevice *device, GError **error)
 	return fu_uefi_dbx_device_set_version_number(device, error);
 }
 
+static void
+fu_uefi_dbx_device_report_metadata_pre(FuDevice *device, GHashTable *metadata)
+{
+	guint64 filesystem_used;
+	g_autoptr(GError) error_local = NULL;
+	g_autoptr(GFile) file = g_file_new_for_path("/sys/firmware/efi/efivars");
```
superm1, line 152: FU_PATH_KIND_SYSFSDIR_FW

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -144,6 +144,31 @@ fu_uefi_dbx_device_probe(FuDevice *device, GError **error)
...
+	g_autoptr(GFileInfo) info = NULL;
+
+	info = g_file_query_info(file,
+				 G_FILE_ATTRIBUTE_FILESYSTEM_USED,
+				 G_FILE_QUERY_INFO_NONE,
+				 NULL,
+				 &error_local);
+	if (info == NULL) {
+		g_debug("failed to get efivar used space: %s", error_local->message);
```
superm1, line 161: this is the path you end up on without the kernel patch right?

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -144,6 +144,31 @@ fu_uefi_dbx_device_probe(FuDevice *device, GError **error)
 	return fu_uefi_dbx_device_set_version_number(device, error);
 }
 
+static void
+fu_uefi_dbx_device_report_metadata_pre(FuDevice *device, GHashTable *metadata)
```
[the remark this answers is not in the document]
hughsie: Yup, agree. I just want to see if it *works* at all first :)

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: nah, you actually fall through to the ==0 case.

plugins/uefi-dbx/fu-uefi-dbx-device.c
```diff
@@ -144,6 +144,31 @@ fu_uefi_dbx_device_probe(FuDevice *device, GError **error)
...
+				 G_FILE_ATTRIBUTE_FILESYSTEM_USED,
+				 G_FILE_QUERY_INFO_NONE,
+				 NULL,
+				 &error_local);
+	if (info == NULL) {
+		g_debug("failed to get efivar used space: %s", error_local->message);
+		return;
+	}
+	filesystem_used = g_file_info_get_attribute_uint64(info, G_FILE_ATTRIBUTE_FILESYSTEM_USED);
+	if (filesystem_used > 0) {
```
superm1, line 165: Hmm, I was really hoping we could have a case that we know patch is here or patch isn't.  I think it's best to record even the zero.  As the patch lands, the reports should hopefully start showing people flip from 0 to non-zero, and then we can know it's working.

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: Yup, thanks, fixed.

plugins/uefi-dbx/fu-uefi-dbx-device.c
hughsie: Makes sense to me. Fixed.

plugins/uefi-dbx/fu-uefi-dbx-device.c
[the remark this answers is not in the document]
hughsie: So we have this: https://github.com/fwupd/fwupd/blob/main/libfwupdplugin/fu-efivar-linux.c#L326 -- but that's really capturing the wrong thing entirely -- that's space free -- not space used. Maybe all this belongs there?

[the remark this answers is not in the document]
superm1: Sure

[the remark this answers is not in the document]
hughsie: Ignore me; that was capturing the right thing as it was stat'ing each file. I've left that in as a fallback for when the kernel is old.

libfwupdplugin/fu-efivar-linux.c
```diff
@@ -329,6 +329,23 @@ fu_efivar_space_used_impl(GError **error)
...
+				    G_FILE_ATTRIBUTE_FILESYSTEM_USED,
+				    G_FILE_QUERY_INFO_NONE,
+				    NULL,
+				    &error_local);
+	if (info_fs == NULL) {
+		g_debug("failed to get efivar used space: %s", error_local->message);
+	} else {
+		total = g_file_info_get_attribute_uint64(info_fs, G_FILE_ATTRIBUTE_FILESYSTEM_USED);
+		if (total > 0)
+			return total;
```
superm1, line 347: at least until it's proved out, can we see if the numbers for adding each file and using the API are different?

hughsie: I think yes as a g_warning -- but I'm hesitant to record it to the LVFS otherwise we'll either need to return two numbers or have two APIs -- that need win32/macos stubs too. Was a g_warning what you thought?

hughsie: Done.

PR #5856, opened 2023-05-24
Limit the number of file objects in the EFI filesystem

hughsie:
When fuzzing a 700KB FuEfiFirmwareFilesystem object we used 12MB of RSS when printing the XML -- which isn't insane, but is too much for oss-fuzz.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=59249

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


superm1:
Is 700KB reasonable for this object?  Maybe it's better to just cap the size all the time instead of just for the fuzzer?

hughsie:
> Is 700KB reasonable for this object

Quite reasonable; a UEFI BIOS is ~16MB and the rest.

superm1 (approved):

PR #5865, opened 2023-05-28
Reformat the meson files with muon

hughsie:
Use this command to reformat all the meson files:

    find . -name meson.build -exec muon fmt -c contrib/muon_fmt.ini -i {} +

I've deliberately not included muon into the precommit scripts as it is not (yet?) available in most ditributions.

Maybe meson should grow a fmt script too?

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (commented):
Which distros is muon missing for? Is it at least available for fedora and Ubuntu?

I'd say add a script for it but "return 0" if it's missing.

data/meson.build
```diff
@@ -82,7 +87,7 @@ if build_daemon
     endif
     if systemd.version().version_compare('>= 231')
       dynamic_options += 'RestrictRealtime=yes'
-#     dynamic_options += 'MemoryDenyWriteExecute=yes'
+      #     dynamic_options += 'MemoryDenyWriteExecute=yes'
```
superm1, line 90: I think this should be `#dynamic_options`

hughsie:
> Which distros is muon missing for?

All of them! It's in neither Ubuntu or Fedora -- it's a c++ re-implementation of meson -- but it's quite incomplete and can't compile fwupd for instance. It does however have a fmt command which is kinda useful. I'm guessing meson will inherit the same feature at some point too.

data/meson.build
hughsie: Agree, fixed.

superm1:
> > Which distros is muon missing for?
> 
> All of them! It's in neither Ubuntu or Fedora -- it's a c++ re-implementation of meson -- but it's quite incomplete and can't compile fwupd for instance. It does however have a fmt command which is kinda useful. I'm guessing meson will inherit the same feature at some point too.
> 

Is it hard to install? We could just run it in pre commit CI only or if it's installed.

hughsie:
> We could just run it in pre commit CI only or if it's installed.

Done.

hughsie:
Lets close this, and revisit when meson gets it's own fmt command.

PR #5876, opened 2023-06-05
plugins: nordic-hid: Update quirk content

MarekPieta:
Allow to use the plugin with any Nordic HID device that supports the configuration channel (check only Vendor ID).

Define common default values of generation and bootloader in quirk. The values are used for all of the Nordic HID devices. Values other than "B0" bootloader and "default" generation must be explicitly reported by device. This is needed to apply the quirk content to peripherals connected via dongle (in this scenario we do not know Product ID in advance).

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/nordic-hid/README.md
+++ b/plugins/nordic-hid/README.md
@@ -53,7 +53,9 @@ This plugin uses the following plugin-specific quirks:
 
 ### NordicHidBootloader
 
-Explicitly set the expected bootloader type: "B0", "MCUBOOT" or "MCUBOOT+XIP".
+Explicitly set the expected bootloader type.
+Only the `B0` bootloader can be set by the quirk file.
+Other values can only be provided by device.
 This quirk must be set for devices without support of `bootloader variant` DFU option.
 
 ### NordicHidGeneration
--- a/plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
+++ b/plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
@@ -471,6 +471,9 @@ fu_nordic_hid_cfg_channel_add_peers(FuNordicHidCfgChannel *self, GError **error)
 					    FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_OPEN);
 		fu_device_add_child(FU_DEVICE(self), FU_DEVICE(peer));
 
+		/* ensure that the general quirk content for Nordic HID devices is applied */
+		fu_device_add_instance_id(FU_DEVICE(peer), "HIDRAW\\VEN_1915");
+
 		/* remove child that does not support config channel DFU */
 		if (!fu_device_setup(FU_DEVICE(peer), &error_peer)) {
 			g_debug("failed to discover peer 0x%02x: %s",
@@ -1294,19 +1297,14 @@ fu_nordic_hid_cfg_channel_set_quirk_kv(FuDevice *device,
 	FuNordicHidCfgChannel *self = FU_NORDIC_HID_CFG_CHANNEL(device);
 
 	if (g_strcmp0(key, "NordicHidBootloader") == 0) {
-		if (g_strcmp0(value, "B0") == 0)
-			self->bl_name = g_strdup("B0");
-		else if (g_strcmp0(value, "MCUBOOT") == 0)
-			self->bl_name = g_strdup("MCUBOOT");
-		else if (g_strcmp0(value, "MCUBOOT+XIP") == 0)
-			self->bl_name = g_strdup("MCUBOOT+XIP");
-		else {
+		if (g_strcmp0(value, "B0") != 0) {
 			g_set_error_literal(error,
 					    G_IO_ERROR,
 					    G_IO_ERROR_INVALID_DATA,
-					    "must be 'B0', 'MCUBOOT' or 'MCUBOOT+XIP'");
+					    "can be only 'B0' in quirk");
 			return FALSE;
 		}
+		self->bl_name = g_strdup(value);
 		return TRUE;
 	}
 
@@ -1318,7 +1316,7 @@ fu_nordic_hid_cfg_channel_set_quirk_kv(FuDevice *device,
 					    "can be only 'default' in quirk");
 			return FALSE;
 		}
-		self->generation = g_strdup("default");
+		self->generation = g_strdup(value);
 		return TRUE;
 	}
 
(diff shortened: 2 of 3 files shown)


mkapala-nordic (dismissed):

hughsie:
Is this going to break any hardware in the field, so to speak?

MarekPieta:
Hi @hughsie,
The change is intended to be backwards compatible with existing devices. In case a device does not support the configuration channel we would end up with an `ioctl` error and device will be silently ignored. Also, for devices connected over an nRF Desktop dongle we do not know PID in advance so unfortunately we cannot rely on it.

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -471,6 +471,9 @@ fu_nordic_hid_cfg_channel_add_peers(FuNordicHidCfgChannel *self, GError **error)
 					    FU_DEVICE_INTERNAL_FLAG_USE_PARENT_FOR_OPEN);
 		fu_device_add_child(FU_DEVICE(self), FU_DEVICE(peer));
 
+		/* Ensure that the general quirk content for Nordic HID devices is applied. */
+		fu_device_add_guid(FU_DEVICE(peer), "HIDRAW\\VEN_1915");
```
hughsie, line 475: That's not a GUID -- what are you trying to do here?

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
```diff
@@ -1294,19 +1297,14 @@ fu_nordic_hid_cfg_channel_set_quirk_kv(FuDevice *device,
...
-		else {
+		if (g_strcmp0(value, "B0") != 0) {
 			g_set_error_literal(error,
 					    G_IO_ERROR,
 					    G_IO_ERROR_INVALID_DATA,
-					    "must be 'B0', 'MCUBOOT' or 'MCUBOOT+XIP'");
+					    "can be only 'B0' in quirk");
 			return FALSE;
 		}
+		self->bl_name = g_strdup("B0");
```
hughsie, line 1307: maybe `g_strdup(value);` for future proofing?

plugins/nordic-hid/nordic-hid.quirk
```diff
@@ -1,19 +1,5 @@
...
-# Nordic Semiconductor ASA Keyboard nRF52 Desktop
-[HIDRAW\VEN_1915&DEV_52DD]
-Plugin = nordic_hid
-GType = FuNordicHidCfgChannel
-NordicHidGeneration = default
-
-# Nordic Semiconductor ASA Dongle nRF52 Desktop
-[HIDRAW\VEN_1915&DEV_52DC]
+# Nordic Semiconductor ASA HID Devices
+[HIDRAW\VEN_1915]
```
hughsie, line 2: you want to probe every device past and future as a CFG channel? This seems a bit risky.

plugins/nordic-hid/fu-nordic-hid-cfg-channel.c
MarekPieta: Changed. Also updated generation handling for consistency

plugins/nordic-hid/README.md
```diff
@@ -53,7 +53,9 @@ This plugin uses the following plugin-specific quirks:
 
 ### NordicHidBootloader
 
-Explicitly set the expected bootloader type: "B0", "MCUBOOT" or "MCUBOOT+XIP".
+Explicitly set the expected bootloader type.
+Only the 'B0' bootloader can be set by the quirk file.
```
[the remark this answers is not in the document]
MarekPieta: Changed

plugins/nordic-hid/nordic-hid.quirk
MarekPieta: That would affect only HID devices with Nordic's Vendor ID, right?
As the first operation during `setup` we simply exchange a HID feature report with predefined report ID and data. We perform HID feature report set and get sequence using `fu_udev_device_ioctl` (and underlying `ioctl`). If a device does not support the HID feature report ID used by the configuration channel, we observe `ioctl` error - in that case we assume that device does not support configuration channel and error is not propagated further eventually resulting in ignoring the device. Unless I missed something, the main risk here is the case of Nordic's HID device that supports the same HID feature report ID and uses it for other purposes. The spurious HID feature report operation may accidentally trigger an operation on device (an unexpected report data would be rejected by the firmware in most cases anyway I think).

On the other hand, using this try&check approach allows us easily handle new devices (no need to update plugin's quirk - new HID devices related to nRF Desktop reference design are supported out-of-the-box).
Also we do not know PIDs of HID peripherals connected via dongle (old version of firmware used not to support reporting them over configuration channel at all). That leads to problems with applying quirks (they need to be applied before the `setup` is done). Apart from this, the fact that PIDs of dongle connected peripherals are not known in advance may lead to inconsistent behavior (e.g. a device may be handled in LVFS when when connected via dongle but it would not work when connected directly - case when dongle's PID is part of quirk, but device's PID is not).

MarekPieta: I was thinking about a solution for this for a while, but nothing better comes to my mind. Do you have a suggestion, how to handle this use-case, @hughsie ?

hughsie: Does https://github.com/fwupd/fwupd/blob/main/docs/ds20.md help?

MarekPieta: nRF Desktop devices may be connected via USB, Bluetooth LE or through an nRF Desktop dongle. Using the `BOS DS20` requires to connect the device directly over USB which may lead to inconsistent behavior that depends on selected connection type (generally not intended in our HID use-case). Configuration channel uses HID feature reports to communicate with the device in very similar way no matter how it's connected to the host, using transport-specific features may not work well with this assumption.

mkapala-nordic (approved):

hughsie:
@MarekPieta -- can you give me a list of commits to include into 1_8_X please? Or even better, file another PR with the ones you want included.

MarekPieta:
Hi @hughsie,
Sure, I will collect the fixes and open a PR.
Please let me know when is the deadline. I still have a few more issues that I would like to address to improve user experience. Knowing the deadline, we will be able to decide what more can we include.

hughsie:
No deadline; Do what you can and I can always tag a new release -- tarballs are cheap.

PR #5887, opened 2023-06-08
Use a helper script to built builtin.quirk.gz

hughsie:
Stripping out the spaces and comments makes the file 33% smaller.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


src/meson.build
```diff
@@ -102,10 +101,19 @@ fwupdutil = library(
...
 )
 
-if build_daemon
-install_data(['org.freedesktop.fwupd.xml'],
-  install_dir: join_paths(datadir, 'dbus-1', 'interfaces')
+dbus_interface = custom_target('fwupd-generate-dbus-interface',
+  input : 'org.freedesktop.fwupd.xml',
+  output : 'org.freedesktop.fwupd.xml',
+  command : [
+    join_paths(meson.project_source_root(), 'contrib', 'generate-dbus-interface.py'),
```
superm1, line 108: This is assuming that the shebang in the top works which I thought is a no-no from past issues.  So aren't you missing a python launcher for this?

superm1: Also - we shouldn't be putting python scripts that are mandatory for the build process into `contrib`.

hughsie: >  So aren't you missing a python launcher for this?

100% yes, good catch.

hughsie: > python scripts that are mandatory for the build

I couldn't see a better option, and I think we need to move quite a few around now. libfwupd or libfwupdplugin don't feel right -- maybe `data`?

contrib/generate-dbus-interface.py
```diff
@@ -0,0 +1,69 @@
...
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-r",
+        "--releases",
+        type=int,
+        default=5,
+    )
+    parser.add_argument(
+        "filename_src", action="store", type=str, help="metainfo source"
```
superm1, line 30: just as a personal preference, why not make this `src` and `dst`?  It seems needlessly long.

contrib/generate-dbus-interface.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-r",
+        "--releases",
+        type=int,
+        default=5,
```
superm1, line 27: eh?

contrib/generate-dbus-interface.py
hughsie: Agreed, done.

contrib/generate-dbus-interface.py
hughsie: Removed.

contrib/generate-dbus-interface.py
```diff
@@ -0,0 +1,69 @@
...
+        root_new = ET.Element('node')
+    #    root_new = tree_new.getroot()
+        root_new.tag = "node"
+        tree_new._setroot(root_new)
+
+        root_new.set('name', "/")
+        n_interface = root.findall("interface")[0]
+        print(n_interface)
+        root_new.append(n_interface)
+    """
```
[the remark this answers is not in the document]
hughsie: oopa, removed.

contrib/generate-quirk-builtin.py
superm1, line 1: move this from contrib

src/meson.build
hughsie: not `build` -- maybe a `meson-helpers` toplevel directory?

meson.build
```diff
@@ -604,19 +604,15 @@ if build_standalone
...
-  gzip = find_program('gzip')
+  # append all the quirks into one big file and gzip it
   custom_target('builtin-quirk-gz',
-    input: builtin_quirk,
+    input: plugin_quirks,
     output: 'builtin.quirk.gz',
-    capture: true,
-    command: [gzip, '-n', '-k', '--stdout', '@INPUT@'],
+    command: [
+      join_paths(meson.project_source_root(), 'contrib', 'generate-quirk-builtin.py'),
```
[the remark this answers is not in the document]
hughsie: Fixed.

src/meson.build
hughsie: the problem with `src` is that it's visited _late_ in the build process; if we had something visited before libfwupd we could just do a `mesonhelperdir = dir(''.')` and then refer to the generate python scripts with `file`

hughsie: Okay, fixed.

contrib/generate-quirk-builtin.py
hughsie: Done.

superm1 (approved):

PR #5889, opened 2023-06-08
Move the scripts needed to generate build artifacts to `generate-build`

hughsie:
The `contrib` directory isn't the right place at all.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Note, this needs the jcat fix copied to gusb too otherwise the mac-os build will fail too.

po/fix_translations.py
superm1, line 1: even though this is translations specific, I say move it to the new directory.  We could eventually aim for code duplication in that directory and have `common.py` that all the scripts import or something too.

hughsie: makes sense to me

superm1: To avoid having to bump subprojects now, can you leave a compat symlink for those projects?

hughsie: It's okay, we're overdue for tarballs anyway.

hughsie:
@superm1 if you like the directory name then this works for me -- and then I can rebase the other PR on top.

superm1:
> @superm1 if you like the directory name then this works for me -- and then I can rebase the other PR on top.

everything looks good to me, but maybe `generators` or `build-scripts` if those sound good to you?  I'm not too strongly feeling on the names.

hughsie:
> maybe generators

I thought this too; but I couldn't get this working as actual meson generators and I think it might be confusing.

> or build-scripts

I wanted to avoid the build prefix -- it'll screw up my tab completion completely :)

superm1:
> > maybe generators
> 
> I thought this too; but I couldn't get this working as actual meson generators and I think it might be confusing.
> 
> > or build-scripts
> 
> I wanted to avoid the build prefix -- it'll screw up my tab completion completely :)

Okay I don't feel strong.

PR #5890, opened 2023-06-08
trivial: Don't run emulation tests without CI_NETWORK set

superm1:
autopkgtest in Ubuntu has a proxy configuration issue on some autopkgtest hosts that is causing it to sometimes fail.

This needs to get fixed on the server end but it's actually surprising it was being run in the first place because we snuck tests in with new artifacts that depended on LVFS now.

We should gate this running behind CI_NETWORK, just like we do for other test data.

Link: https://bugs.launchpad.net/ubuntu/+source/fwupd/+bug/2021908

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/data/installed-tests/fwupd.sh
+++ b/data/installed-tests/fwupd.sh
@@ -12,7 +12,7 @@ run_test()
 
 run_device_tests()
 {
-	if [ -d @devicetestdir@ ]; then
+	if [ -n "$CI_NETWORK" ] && [ -d @devicetestdir@ ]; then
 		# grab device tests from the CDN to avoid incrementing the download counter
 		export FWUPD_DEVICE_TESTS_BASE_URI=http://cdn.fwupd.org/downloads
 		for f in `grep --files-with-matches -r emulation-url @devicetestdir@`; do


data/installed-tests/fwupd.sh
```diff
@@ -12,7 +12,7 @@ run_test()
 
 run_device_tests()
 {
-	if [ -d @devicetestdir@ ]; then
+	if [ -n "$CI_NETWORK" ] && [ -d @devicetestdir@ ]; then
```
hughsie, line 15: We're sure CI_NETWORK is set for PRs?

superm1: It's set in the docker containers that everything runs:
```
$ grep CI_NETWORK contrib/ -R
contrib/ci/Dockerfile-ubuntu.in:ENV CI_NETWORK true
contrib/ci/Dockerfile-void.in:ENV CI_NETWORK true
contrib/ci/get_test_firmware.sh:if [ "$CI_NETWORK" = "true" ]; then
contrib/ci/Dockerfile-arch.in:ENV CI_NETWORK true
contrib/ci/Dockerfile-debian.in:ENV CI_NETWORK true
```

superm1: So there are "some" tests (not launched in docker) that won't run with it set this means, but there will definitely be coverage.

hughsie (approved):

PR #5892, opened 2023-06-09
usi-dock: Adjust the display message according to the customer's requ…

victor-cheng:
…irements.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/usi-dock/fu-usi-dock-mcu-device.c
+++ b/plugins/usi-dock/fu-usi-dock-mcu-device.c
@@ -705,6 +705,23 @@ fu_usi_dock_mcu_device_attach(FuDevice *device, FuProgress *progress, GError **e
 }
 
 static gboolean
+fu_usi_dock_mcu_device_insert_cb(gpointer user_data)
+{
+	g_autoptr(FwupdRequest) request = fwupd_request_new();
+	/* interactive request to start the SPI write */
+	fwupd_request_set_kind(request, FWUPD_REQUEST_KIND_IMMEDIATE);
+	fwupd_request_set_id(request, FWUPD_REQUEST_ID_INSERT_USB_CABLE);
+	fwupd_request_add_flag(request, FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGE);
+	fwupd_request_set_message(
+	    request,
+	    "The update will continue when the device USB cable has been re-inserted.");
+	fu_device_emit_request(user_data, request);
+
+	/* success */
+	return G_SOURCE_REMOVE;
+}
+
+static gboolean
 fu_usi_dock_mcu_device_cleanup(FuDevice *device,
 			       FuProgress *progress,
 			       FwupdInstallFlags install_flags,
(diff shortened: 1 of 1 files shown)


plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -718,7 +718,10 @@ fu_usi_dock_mcu_device_cleanup(FuDevice *device,
 	fwupd_request_add_flag(request, FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGE);
 	fwupd_request_set_message(
 	    request,
-	    "The update will continue when the device USB cable has been re-inserted.");
+	    "Wait 30 seconds after unplugging the device USB cable, the Phase 2 update will "
+	    "continue. Otherwise, it will continue to update the next time the USB cable is "
+	    "unplugged. The new FW will take effect automatically when the dock cable is "
+	    "disconnected from and connected to the computer again.");
```
hughsie, line 724: That's really hard to understand if you don't know english. Can we just ask the user to disconnect the cable, set a timer for 30 seconds and then ask the user to plug it back in?

victor-cheng: Sure, are you referring to the approach used in fu_wistron_dock_device_attach? I have a question: if the user does not follow the first instruction to unplug the cable and the instruction to plug in the cable still appears from the callback function, it seems that there is no longer a purpose to set the timer if the user goes on to unplug and plug the cable again?

hughsie: > fu_wistron_dock_device_attach

if we need to do that, yes. If I'm honest I'm really unimpressed with how the resets and physical actions are performed -- it seems really really fragile and easy to go wrong. We can't ask the user to wait 30 seconds -- user studies have shown that it's really hard for users to follow multi-stage processes with intervals.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -718,12 +735,15 @@ fu_usi_dock_mcu_device_cleanup(FuDevice *device,
 	fwupd_request_add_flag(request, FWUPD_REQUEST_FLAG_ALLOW_GENERIC_MESSAGE);
```
hughsie, line 735: this has the wrong id -- you need to replace `FWUPD_REQUEST_ID_INSERT_USB_CABLE` with `FWUPD_REQUEST_ID_REMOVE_USB_CABLE`

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -718,12 +735,15 @@ fu_usi_dock_mcu_device_cleanup(FuDevice *device,
...
-	    "The update will continue when the device USB cable has been re-inserted.");
+	    "Please unplug the USB cable. The update will continue in 30 seconds from now.");
 	fu_device_emit_request(device, request);
 
 	/* success */
 	fu_device_set_remove_delay(device, 900000);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+
+	fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_BUSY);
+	g_timeout_add_seconds(40, fu_usi_dock_mcu_device_insert_cb, device);
```
hughsie, line 746: this is much better, but I think you should do `g_timeout_add_seconds_full(....` with a `g_object_ref(device) so that the ref on FuDevice stays valid until the callback fires.

plugins/usi-dock/fu-usi-dock-mcu-device.c
```diff
@@ -714,16 +731,24 @@ fu_usi_dock_mcu_device_cleanup(FuDevice *device,
...
 	/* success */
 	fu_device_set_remove_delay(device, 900000);
 	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+
+	fu_progress_set_status(progress, FWUPD_STATUS_DEVICE_BUSY);
+	g_timeout_add_seconds_full(G_PRIORITY_DEFAULT,
+				   40,
+				   fu_usi_dock_mcu_device_insert_cb,
+				   g_object_ref(device),
+				   NULL);
```
hughsie, line 750: You need g_object_unref here, no?

hughsie (approved):

PR #5895, opened 2023-06-11
trivial: Do not use versioned symbols for libfwupdplugin

hughsie:
This isn't exported, and doesn't have an API or ABI so there's little point being strict about the exports.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation


superm1 (approved):
Yeah this matters less now

PR #5897, opened 2023-06-12
ccgx: Add support for Framework HDMI/DP cards

JohnAZoidberg:
Superceding #5834. Thanks for @superm1 and @hughsie for the reviews so far.

- [ ] Code fix
- [x] Feature
- [ ] Documentation

---

Adding support to the ccgx plugin for Framework HDMI and DisplayPort Expansion cards.
They main chip inside of them is a CCG3 PD controller.
This method of getting firmware version and updating uses only HID report read/writes.
This is unlike the current HID support in this plugin which only uses HID to switch the device into HPI mode.

The protocol supported by these Framework Expansion cards is slightly different from the standard CCG3 SDK.
I haven't encountered other devices that use this API, so I'm not 100% sure what all is different. When we want to add support for those other devices, we can customize it.

Another difference is that the Framework HDMI/DP Expansion cards update the "BaseVersion", not the "AppVersion".
This is not what the other implementations of this plugin expect. So you'll see the wrong version listed when trying to parse Framework cyacd files. That's okay, we'll just specify the right version in the metadata XML.

Like many other CCGX devices, the cards have a main firmware and a backup firmware. AppPriority is not implemented, so both images are equal in functionality and priority. If one is running, only the other one can be updated. And the card will keep booting to the one that was most recently updated. If an update fails or the flashed firmware is not bootable, the card boots back into the original firmware. This makes it impossible for an update to brick the card.

The images are different, though. They know whether they are firmware 1 (backup) or firmware 2 (main). So you have to flash the right one. That's why the device GUIDs include this. The metainf file can filter by that to assign the right firmware file to the device only when in the right mode.

Once the firmware is updated the card needs to reboot into the new firmware. It'll come back in the other mode.

I have validated that this plugin can update and downgrade in both modes. And it works properly when multiple cards are attached - fwupd waits for the right one to reboot until it reports success.
If both cards need an update it will properly update both of them.

Required TODO:

The currently shipping HDMI cards SHOULD have [some rework](https://guides.frame.work/Guide/HDMI+Expansion+Card+power+saving+rework+(Beta)/193) done before they can be updated. So we need some way to ask the user for confirmation if they have done that. Otherwise abort the flash. If they update anyways, it won't render the card inoperable, they can still downgrade.
I have only found how to attach a picture: https://lvfs.readthedocs.io/en/latest/metainfo.html#client-features
But it also doesn't let the user cancel the update.
Is there something like this?

Optional enhancements:

- [ ] Could check if it reboots into the same firmware and signal update failure


plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+ */
+#define FU_CCGX_NATIVE_HID_DEVICE_IS_IN_RESTART (1 << 0)
+
+struct _FuCcgxNativeHidDevice {
+	FuHidDevice parent_instance;
+	FuCcgxFwMode fw_mode;
+	guint32 versions[FU_CCGX_FW_MODE_LAST];
+	guint32 silicon_id;
+	guint32 flash_row_size; // FIXME: unused?
+	guint32 flash_size;	// FIXME: unused?
```
hughsie, line 32: still unused?

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	fu_device_remove_private_flag(device, FU_CCGX_NATIVE_HID_DEVICE_IS_IN_RESTART);
+	return TRUE;
+}
+
+static gboolean
+fu_ccgx_native_hid_device_close(FuDevice *device, GError **error)
+{
+	/* do not close handle when device restarts */
+	if (fu_device_has_private_flag(device, FU_CCGX_NATIVE_HID_DEVICE_IS_IN_RESTART))
+		return TRUE;
```
hughsie, line 181: do we need this? i think the daemon just ignores the error if we close a device that doesn't exist.

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+			     const guint8 *row,
+			     guint16 rowsize,
+			     GError **error)
+{
+	g_autoptr(GByteArray) st_hdr = fu_struct_ccgx_native_hid_write_hdr_new();
+
+	fu_struct_ccgx_native_hid_write_hdr_set_pd_resp(st_hdr,
+							FU_CCGX_PD_RESP_FLASH_READ_WRITE_CMD_SIG);
+	fu_struct_ccgx_native_hid_write_hdr_set_addr(st_hdr, address);
+	//	buf[0] = FU_CCGX_NATIVE_HID_REPORT_ID_WRITE;
```
hughsie, line 240: delete?

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED);
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_BUSY, 5, "unlock");
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 90, NULL);
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_BUSY, 5, "bootswitch");
+
+	if (!fu_ccgx_native_hid_magic_unlock(self, error))
+		return FALSE;
+	// TODO: Probably not required
```
hughsie, line 280: yup, i think it's better to remove

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+		fu_progress_set_percentage_full(fu_progress_get_child(progress),
+						(gsize)i + 1,
+						(gsize)records->len);
+	}
+	fu_progress_step_done(progress);
+
+	// TODO: Doing this and reset by themselves (with magic unlock) doesn't
+	// switch to the alternative image. Seems we always need to flash in
+	// order to switch.
+	g_debug("before bootswitch");
```
hughsie, line 307: remove

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	// TODO: fw_img_no should just be
+	// fw_img_no = fu_ccgx_firmware_get_fw_mode(FW_CCGX_FIRMWARE(firmware))
+	if (!fu_ccgx_native_hid_command(self, CCGX_HID_CMD_SET_BOOT, fw_img_no, error)) {
+		g_prefix_error(error, "bootswitch command error: ");
+		return FALSE;
+	}
+	g_debug("After bootswitch");
```
hughsie, line 314: remove etc

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+
+	g_debug("before reset");
+	if (!fu_ccgx_native_hid_command(self,
+					CCGX_HID_CMD_JUMP,
+					FU_CCGX_PD_RESP_DEVICE_RESET_CMD_SIG,
+					error)) {
+		g_prefix_error(error, "reset command error: ");
+		return FALSE;
+	}
+	g_debug("After reset");
```
hughsie, line 325: same

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+static gboolean
+fu_ccgx_native_hid_device_write_firmware(FuDevice *device,
+					 FuFirmware *firmware,
+					 FuProgress *progress,
+					 FwupdInstallFlags flags,
+					 GError **error)
+{
+	FuCcgxNativeHidDevice *self = FU_CCGX_NATIVE_HID_DEVICE(device);
+
+	g_debug("Operating Mode:  0x%02x", self->fw_mode);
```
hughsie, line 391: no need for this, it should be in to_string() already.

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	switch (self->fw_mode) {
+	case FU_CCGX_FW_MODE_BOOT:
+	case FU_CCGX_FW_MODE_FW2:
```
hughsie, line 394: can you just do `if (self->fw_mode == FU_CCGX_FW_MODE_BOOT || self->fw_mode == FU_CCGX_FW_MODE_FW2)` instead please and then move out the common code.

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+		g_debug("Flashing Image 2");
+		if (!fu_ccgx_native_hid_flash_firmware_image(self, firmware, progress, 2, error))
+			return FALSE;
+
+		g_debug("Add wait for replug");
+		fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+		fu_device_add_private_flag(device, FU_CCGX_NATIVE_HID_DEVICE_IS_IN_RESTART);
+		break;
+	default:
+		return FALSE;
```
hughsie, line 415: i don't think we need to handle this if we abort on setup() if it's not FW1, FW2 or BOOT.

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	return TRUE;
+}
+
+// TODO: Need this? Why are there two?
```
hughsie, line 421: we need this for whatever subclass implements write() -- so I think we do need this.

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_RESTART, 5, "detach");
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_WRITE, 45, "write");
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_RESTART, 5, "attach");
+	fu_progress_add_step(progress, FWUPD_STATUS_DEVICE_BUSY, 45, "reload");
+}
+
+static void
+fu_ccgx_native_hid_device_init(FuCcgxNativeHidDevice *self)
+{
+	// FIXME: I don't think this is try anymore
```
hughsie, line 436: we can remove the FIXME i think

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	fu_device_add_protocol(FU_DEVICE(self), "com.infineon.ccgx");
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UNSIGNED_PAYLOAD);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_UPDATABLE);
+	fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_DUAL_IMAGE);
+	fu_device_set_version_format(FU_DEVICE(self), FWUPD_VERSION_FORMAT_INTEL_ME2);
```
hughsie, line 441: `quad` or `triplet` perhaps?

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,505 @@
...
+	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_ONLY_WAIT_FOR_REPLUG);
+	fu_device_add_internal_flag(FU_DEVICE(self), FU_DEVICE_INTERNAL_FLAG_REPLUG_MATCH_GUID);
```
hughsie, line 443: do we need this flag?

plugins/ccgx/fu-ccgx.rs
```diff
@@ -87,3 +90,42 @@ enum CcgxPdResp {
     SenderResponseTimerTimeout,
     NoVdmResponseReceived,
 }
+
+#[derive(Parse)]
+struct CcgxNativeHidFwInfo {
+    report_id: u8: const=0xE0,
+    _reserved_1: u8,
```
hughsie, line 97: i don't think it matters, but _reserved1 would be more standard

plugins/ccgx/fu-ccgx-native-hid-device.c
JohnAZoidberg: No, this is actually the correct format for the ccgx base version. The name is confusing but quad format isn't what we want.

plugins/ccgx/fu-ccgx-native-hid-device.c
JohnAZoidberg: Yes, because the firmware mode changes between fw1 and fw2.
Otherwise when the device comes back after the update fwupd won't recognize that it's the same.

I'm not quite sure how it figures out that it's the same device. Because the "device id" does change. But it seems to work.

hughsie: Ahh of course, you're 100% right to use the GUID to identify the "same" device.

plugins/ccgx/fu-ccgx-native-hid-device.c
hughsie: If you really do want `A.B.CC.DDDD` then I agree the name is confusing :)

plugins/ccgx/fu-ccgx-native-hid-device.c
JohnAZoidberg: It won't work when I remove this.
But what values should I put here?

hughsie: if you run the daemon with --verbose it should spit out some numbers. Make sure you're compiling with `-Dsupported=false`

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,466 @@
...
+	bl_ver = fu_version_from_uint32(fu_struct_ccgx_native_hid_fw_info_get_bl_version(st_info),
+					fu_device_get_version_format(self));
+	fu_device_set_version_bootloader(FU_DEVICE(self), bl_ver);
+	return TRUE;
+}
+
+static gboolean
+fu_ccgx_native_hid_device_attach(FuDevice *device, FuProgress *progress, GError **error)
+{
+	fu_device_remove_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
```
hughsie, line 169: I don't think you need to do this; the daemon will do it for you

plugins/ccgx/fu-ccgx-native-hid-device.c
```diff
@@ -0,0 +1,466 @@
...
+	fu_device_remove_private_flag(device, FU_CCGX_NATIVE_HID_DEVICE_IS_IN_RESTART);
```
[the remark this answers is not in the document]
JohnAZoidberg: It seem fine, yeah.
I just defensively copied it from the HPI code because I wasn't sure how to make fwupd wait for the device to come back.

plugins/ccgx/fu-ccgx-native-hid-device.c
hughsie: @JohnAZoidberg want me to create a `com.infineon.ccgx.native-hid` on the LVFS?

hughsie:
@JohnAZoidberg do you want this in 1.9.3? If so, we've got a couple of weeks to fixup all the last few TODOs.

hughsie:
@JohnAZoidberg hey, can we get this over the finish line? it's so nearly there.

superm1:
> @JohnAZoidberg hey, can we get this over the finish line? it's so nearly there.

likewise.  If we want to land this in things that move at a glacial pace like Ubuntu's next LTS we need it in 1.9.8 at the latest IMO.

superm1:
I've rebased this on the API changes we've made in main (since we're going to have an ABI break with 2.0.0).  I've also cleaned up a few of the TODOs like adding a `prepare_firmware` callback and fixed the 32 bit CI build.

The remaining 10 TODOs / FIXMEs I think need @JohnAZoidberg to look them over though to decide.  Also, make sure the `prepare_firmware` callback I added actually works :)

superm1:
Also; caught a few cases of things that should have been rs enums that I've transitioned over.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,514 @@
...
+#define FW2_START	  0x0200
+#define FW1_METADATA	  0x03FF
+#define FW2_METADATA	  0x03FE
+
+/**
+ * FU_CCGX_PURE_HID_DEVICE_IS_IN_RESTART:
+ *
+ * Device is in restart and should not be closed manually.
+ *
+ * Since: 1.9.2
```
superm1, line 28: this should bump up now

hughsie: Ohh, I think @d4s has a better branch than this one now.

d4s: ohh... not yet. Just did some cleanup and minor fixes.
Let me fix the rest of TODO/FIXME tomorrow and test updates.

d4s:
Update is not work for me yet.

superm1:
> The original utility for windows is flashing non-active bank, then reboot the device to it and flash the other bank. So both banks have the same versions. It's a bit strange to me since cards are using symmetric banks. From my experiments with HDMI and DP cards, the second step looks unnecessary and potentially dangerous. Both devices I have, properly booted from the last flashed bank and I don't see the reason why we need to flash the same version to other bank immediately.

Completely agree with you.  We've adopted a similar policy in fwupd for a variety of other devices too.

> Unable to read the target device from the image -- that might be a potential issue, since the target device isn't protected from flashing the image from the other device. So, I'm able to flash the HDMI card with the image for DP card. The device has been flashed well and after restart it is detected as DP card ;) Able to reflash it back to HDMI variant.

There is at least a USB VID difference though, right?  So specifically the GUIDs will be different, but if someone tried to use install-blob they can do damage.

d4s:
> There is at least a USB VID/PID difference though, right? So specifically the GUIDs will be different, but if someone tried to use install-blob they can do damage.

right, however I don't know if VID:PID place is a part of the firmware format for all devices -- i.e. should I read that for all firmwares or only for Framework cards?
Here I **suspect**  it is placed on:
```
:0000360080 AC320300....
```
and for the 2-nd bank:
```
:0002060080 AC320300...
```
at least in all images I have on hands.

superm1:
> right, however I don't know if VID:PID place is a part of the firmware format for all devices -- i.e. should I read that for all firmwares or only for Framework cards?

If the spec is unclear; quirks are cheap and later reversible.  I suggest we quirk the two framework cards to have something like `DevIdOffset` and then the plugin can will validate that offset only if it's specified in a quirk.

d4s:
@superm1 sorry, which spec you are talking about? Would be appreciated for the link

superm1:
I don't have one, I was figuring you did and were cross referencing one :)

d4s:
@superm1 I've added the simple VID:PID protection. But since I have no proper documentation, it is based on images quick analysis and some guessing. However, it works for cards I have on hands.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+	fu_device_set_version_bootloader_raw(
+	    device,
+	    fu_struct_ccgx_pure_hid_fw_info_get_bl_version(st_info));
+	bl_ver = fu_version_from_uint32(fu_struct_ccgx_pure_hid_fw_info_get_bl_version(st_info),
+					fu_device_get_version_format(self));
+	fu_device_set_version_bootloader(FU_DEVICE(self), bl_ver);
+
+	/* FIXME: d4s: do we need that for querying??? */
+	if (!fu_ccgx_pure_hid_enter_flashing_mode(self, error))
+		return FALSE;
```
superm1, line 154: Can you try to drop this?  It's better we don't swap modes every time the daemon starts up IMO.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+				      FU_CCGX_PURE_HID_COMMAND_JUMP,
+				      FU_CCGX_PD_RESP_DEVICE_RESET_CMD_SIG,
+				      error)) {
+		g_prefix_error(error, "reset command error: ");
+		return FALSE;
+	}
+
+	g_debug("wait for replug");
+	fu_device_add_flag(device, FWUPD_DEVICE_FLAG_WAIT_FOR_REPLUG);
+
```
superm1, line 390: all these `g_debug` lines probably can go, they're more bringup related than valuable to debug IMO

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+		g_debug("expected HID report bridge mode failure: %s", error_local->message);
+	}
+
+	return TRUE;
+}
+
+static gboolean
+fu_ccgx_pure_hid_ensure_fw_info(FuCcgxPureHidDevice *self, GError **error)
+{
+	FuDevice *device = FU_DEVICE(self);
```
hughsie, line 100: not keen on this, just use FU_DEVICE(self) pls

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+
+	fu_device_add_instance_u16(FU_DEVICE(self), "SID", self->silicon_id);
+	if (!fu_device_build_instance_id_full(FU_DEVICE(self),
+					      FU_DEVICE_INSTANCE_FLAG_QUIRKS,
+					      error,
+					      "CCGX",
+					      "SID",
+					      NULL))
+		return FALSE;
+	g_debug("got silicon ID: 0x%04x", self->silicon_id);
```
hughsie, line 192: can this be in _to_string() instead please

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+		g_set_error_literal(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_FILE,
+				    "unable to read VID and PID from the image");
+		return NULL;
+	}
+	vidpid_rcd = g_ptr_array_index(records, VIDPID_BLOCK_ID);
+	bytes = g_bytes_get_data(vidpid_rcd->data, NULL);
+	vid = fu_memread_uint16(bytes, G_LITTLE_ENDIAN);
+	pid = fu_memread_uint16(bytes + 2, G_LITTLE_ENDIAN);
```
hughsie, line 284: fu_memread_uint16_safe() would be safer here.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+{
+	FuCcgxPureHidDevice *self = FU_CCGX_PURE_HID_DEVICE(device);
+	guint8 fw_mode = 1;
+	GPtrArray *records = fu_ccgx_firmware_get_records(FU_CCGX_FIRMWARE(firmware));
+
+	g_debug("operating mode: 0x%02x", self->operating_mode);
+
+	if (self->operating_mode != FU_CCGX_PURE_HID_FW_MODE_FW2) {
+		fw_mode = 2;
+	}
```
hughsie, line 354: no braces pls

plugins/ccgx/fu-ccgx.rs
```diff
@@ -22,15 +22,13 @@ enum FuCcgxImageType {
     DualAsymmetric,         // A=bootloader (fixed) B=runtime
     DualAsymmetricVariable, // A=bootloader (variable) B=runtime
 }
-
```
hughsie, line 25: nah, i like the spaces :)

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+
+	for (guint i = 0; i < records->len; i++) {
+		FuCcgxFirmwareRecord *record = g_ptr_array_index(records, i);
+		gsize record_size = g_bytes_get_size(record->data);
+		if (record_size != self->flash_row_size) {
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_FILE,
+				    "expected block length %lu, got %lu: array id=0x%02x, "
+				    "row=0x%04x (:%02x%04x%04lx)",
```
superm1, line 258: this isn't the correct format string for architecture independent, look at the CI error.

plugins/ccgx/fu-ccgx-pure-hid-device.c
```diff
@@ -0,0 +1,472 @@
...
+				    FWUPD_ERROR_INVALID_FILE,
+				    "unable to read VID and PID from the image");
+		return NULL;
+	}
+	vidpid_rcd = g_ptr_array_index(records, VIDPID_BLOCK_ID);
+	bytes = g_bytes_get_data(vidpid_rcd->data, NULL);
+	vid = fu_memread_uint16(bytes, G_LITTLE_ENDIAN);
+	pid = fu_memread_uint16(bytes + 2, G_LITTLE_ENDIAN);
+
+	g_debug("image is for %04X:%04X", vid, pid);
```
superm1, line 286: given you show an error when they don't match I don't know I see the value for the printing in debug log.

plugins/ccgx/fu-ccgx-pure-hid-device.c
hughsie: ya, just coerce into (guint) to print %x

plugins/ccgx/fu-ccgx-pure-hid-device.c
d4s: According to Wireshark, it is used for querying with Windows utility from Framework.
However the detection works without it.
I moved it to `prepare_firmware()` since I don't want to drop this command from the write sequence and I have no any documentation to check if it is really required.

plugins/ccgx/fu-ccgx-pure-hid-device.c
d4s: not safer, that's the reason I put it here -- in the loop above I do the check of each row if the block have a proper size. and I remember that warning "_You don't need to use this function in "obviously correct" cases_" ;)
Do you still think I need `_safe()` variant here?
@hughsie do you think I this loop is needed at all?

plugins/ccgx/fu-ccgx.rs
d4s: sorry, overlooked that change

plugins/ccgx/fu-ccgx-pure-hid-device.c
d4s: removed debug. it is already in `_to_string()`

plugins/ccgx/fu-ccgx-pure-hid-device.c
superm1: Yeah it makes sense in the write sequence, but don't think we should have it otherwise.

BTW - since you captured with wireshark in Windows, can you do device installation emulation data too?  We can add it to the installed tests then to keep this thing working.

superm1 (approved):

superm1:
Squash it all together once @hughsie is happy, otherwise it LGTM now.

hughsie (approved):
Look great, thanks for getting this over the line!

Issue #5898, opened 2023-06-12
1.9.2: build fails in generate-dbus-interface.py script


kloczek:
Looks like something is wrong.
```console
[tkloczko@pers-jacek x86_64-redhat-linux-gnu]$ ninja
[76/243] Generating src/fwupd-generate-dbus-interface with a custom command
FAILED: src/org.freedesktop.fwupd.xml
/usr/bin/python3 ../generate-build/generate-dbus-interface.py ../src/org.freedesktop.fwupd.xml src/org.freedesktop.fwupd.xml
Traceback (most recent call last):
  File "../generate-build/generate-dbus-interface.py", line 42, in <module>
    ET.indent(tree, space=" ", level=0)
AttributeError: module 'xml.etree.ElementTree' has no attribute 'indent'
```
I'm using libxml2 2.10.4.

hughsie:
What python are you using? You need 3.9 at least.

kloczek:
I'm using 3.8.17 (still it is not EOSed).
Nothing on meson run warned that python >=3.9 needs to be used.

PR #5899, opened 2023-06-12
Do not accidentally depend on python 3.9

hughsie:
Fixes https://github.com/fwupd/fwupd/issues/5898

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/generate-build/generate-dbus-interface.py
+++ b/generate-build/generate-dbus-interface.py
@@ -5,6 +5,7 @@
 #
 # SPDX-License-Identifier: LGPL-2.1+
 
+import sys
 import argparse
 import xml.etree.ElementTree as ET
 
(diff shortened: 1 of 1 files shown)


hughsie:
@kloczek -- can you verify please

kloczek:
Instead checking that in custom script IMO better would be use meson `python` module to check python version.
Really it is not possible to use 3.8? 🤔

hughsie:
> Really it is not possible to use 3.8?

Try this branch on 3.8 please.

superm1 (approved):

kloczek:
Still I think that better would be just check python version by use `dependency('python3 >= 3.9')`

superm1:
> Still I think that better would be just check python version by use `dependency('python3 >= 3.9')`

Why? It's not a hard dependency and this will allow it to work on more machines still.

kloczek:
Seems it is hard dependency because now without python >=3.9 it is not possible to generate dbus interface file.

superm1:
It generates, it's just ugly.

PR #5900, opened 2023-06-12
Ignore small ESP block devices if there are multiple choices

hughsie:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2214191

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-volume.c
+++ b/libfwupdplugin/fu-volume.c
@@ -198,6 +198,31 @@ fu_volume_get_id(FuVolume *self)
 }
 
 /**
+ * fu_volume_get_size:
+ * @self: a @FuVolume
+ *
+ * Gets the size of the block device pointed to by the volume.
+ *
+ * Returns: size in bytes, or 0 on error
+ *
+ * Since: 1.9.3
+ **/
+guint64
+fu_volume_get_size(FuVolume *self)
+{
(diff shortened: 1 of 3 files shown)


superm1 (commented):
What if they're both (or all) small? Seems unlikely but I think this would make them all get skipped. So don't you want to make sure that there is at least one bigger before you skip the first small?

hughsie:
What about if we sorted them by size, and just chose the biggest one?

libfwupdplugin/fu-volume.c
```diff
@@ -197,6 +197,31 @@ fu_volume_get_id(FuVolume *self)
...
 
+/**
+ * fu_volume_get_size:
+ * @self: a @FuVolume
+ *
+ * Gets the size of the block device pointed to by the volume.
+ *
+ * Returns: size in bytes, or 0 on error
+ *
+ * Since: 1.6.3
```
superm1, line 208: 1.9.3?

superm1 (approved):

libfwupdplugin/fu-volume.c
hughsie: Indeed, fixed, thanks.

PR #5910, opened 2023-06-16
Allow devices to only accept explicitly specified release versions

hughsie:
For some devices, a 'newer' release may be no better than the version they have right now, or is may be unexpected or disallowed to update the firmware automatically.

In other cases it might be that per-device user-set presets would be cleared on update, or that only Best Known Configuration is allowed for regulatory reasons.

Provide a per-device flag that allows the explicit install action, but hides the device from the 'simple' GetUpgrades API call.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-enums.c
+++ b/libfwupd/fwupd-enums.c
@@ -213,6 +213,8 @@ fwupd_device_flag_to_string(FwupdDeviceFlags device_flag)
 		return "emulated";
 	if (device_flag == FWUPD_DEVICE_FLAG_EMULATION_TAG)
 		return "emulation-tag";
+	if (device_flag == FWUPD_DEVICE_FLAG_ONLY_EXPLICIT_UPDATES)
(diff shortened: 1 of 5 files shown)


superm1:
I don't really like the precedent for this. Can you better explain the why?

hughsie:
It's mostly for the regulatory requirements for some devices -- e.g. if you have some medical device the requirement is that you can't just plug in a random ChromeBook and update the firmware automatically -- but you do want the technician to be able to update the firmware during servicing.

superm1:
> It's mostly for the regulatory requirements for some devices -- e.g. if you have some medical device the requirement is that you can't just plug in a random ChromeBook and update the firmware automatically -- but you do want the technician to be able to update the firmware during servicing.

Got it. So how will policy be configured for such devices?

superm1 (approved):

Issue #5907, opened 2023-06-15
Jabra Evolve 65 not detected


rodehoed:
**Describe the bug**
I have a Jabra Evolve 65 which is not getting detected with get-devices. As stated in the release notes the Jabra Evolve line should be supported in 1.9.2. 

**Steps to Reproduce**
- Plugin USB cable in headset
- run _fwupdmgr get-devices_

**Expected behavior**
The device should be shown

**fwupd version information**

```
compile   org.freedesktop.fwupd         1.9.2
compile   com.hughsie.libxmlb           0.3.11
compile   com.hughsie.libjcat           0.1.13
runtime   org.freedesktop.fwupd-efi     1.4
compile   org.freedesktop.gusb          0.4.5
runtime   com.hughsie.libjcat           0.1.13
runtime   org.freedesktop.gusb          0.4.5
runtime   org.freedesktop.fwupd         1.9.2
runtime   org.kernel                    6.3.5-200.fc38.x86_64
```
RPM packages:
```
fwupd-efi-1.4-1.fc38.x86_64
fwupd-plugin-flashrom-1.9.2-1.fc38.x86_64
fwupd-plugin-modem-manager-1.9.2-1.fc38.x86_64
fwupd-plugin-uefi-capsule-data-1.9.2-1.fc38.x86_64
fwupd-1.9.2-1.fc38.x86_64
```

fwupdmgr get-plugins Shows me that _jabra_ and _jabra_gnp_ are enabled.

<details>

<summary>**fwupd device information**</summary>

ASUSTeK COMPUTER INC. M11AD
│
├─Core™ i5-4460S CPU @ 2.90GHz:
│     Device ID:          4bde70ba4e39b28f9eab1628f9dd6e6244c03027
│     Current version:    0x00000028
│     Vendor:             Intel
│     GUIDs:              b9a2dd81-159e-5537-a7db-e7101d164d3f ← cpu
│                         30249f37-d140-5d3e-9319-186b1bd5cac3 ← CPUID\PRO_0&FAM_06
│                         dd2cdaeb-2b08-599d-aaec-5a8374add8d6 ← CPUID\PRO_0&FAM_06&MOD_3C
│                         112c1ba3-2c29-5ab2-bd72-ea65bd10dd67 ← CPUID\PRO_0&FAM_06&MOD_3C&STP_3
│     Device Flags:       • Internal device
│   
├─Lexa PRO [Radeon 540/540X/550/550X / RX 540X/550/550X]:
[...]

PR #5912, opened 2023-06-16
Document which Jabra devices are supported

ljrk0:
Addresses confusions in #5907 regarding the newly supported devices from #5638.

The mentioned PR adds support for Jabra Evolve and Evolve2 devices, including Jabra Evolve 65e/t and SE (second generation) devices, but excluding first generation devices.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [x] Documentation

I'm not sure whether this can be phrased better, but I think it'd be helpful to document that in the READMEs & Changelog instead of being buried in the GitHub issue.  Feedback welcome :)

### DIFF
--- a/plugins/jabra-gnp/README.md
+++ b/plugins/jabra-gnp/README.md
@@ -2,7 +2,11 @@
 
 ## Introduction
 
-This plugin is used to firmware update some Jabra devices.
+This plugin is used to firmware update for some Jabra devices
+(refer to the `jabra-gnp.quirk` file for more information).
+Notably this excludes devices supported by the `jabra` plugin,
+as well as 1st edition Jabra Evolve (non-SE) devices and the
+corresponding Jabra Link connectors.
 
 ## GUID Generation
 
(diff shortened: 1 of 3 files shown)


plugins/jabra-gnp/README.md
```diff
@@ -2,7 +2,8 @@
 
 ## Introduction
 
-This plugin is used to firmware update some Jabra devices.
+This plugin is used to firmware update some Jabra devices,
+notably the Evolve SE (2nd Edition), Evolve2 and Jabra Link connectors.
```
gdpcastro, line 6: I think I would phrase this slightly differently. 
The way I see it we can either:
Start the list of devices with "some" (some Evolve, Evolve2 etc.) and then refer the user to the quirk file, if that's acceptable.
Otherwise list all devices supported. 
The reason is the same stated in the issue, for example we have a Link connector that uses a different protocol and is therefore not supported by this plugin. 
Also, this is missing the Speak2 line

hughsie: @ljrk0 can you fix up the commit and force push your branch to update this PR please.

ljrk0: Yep, will do!  Maybe it's actually more helpful if we note by example what lines are not supported and refer to the quirks file?  I've drafted something I'll be pushing now, I does sound more clear from a docs perspective, but I'm interested in your feedback (gosh, writing good and concise docs ain't easy).

hughsie (approved):

PR #5921, opened 2023-06-21
added device name read for jabra-gnp devices

gdpcastro:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

With these changes fwupd is able to report the name of jabra-gnp devices, instead of displaying "(null)" or "Unknown device"

### DIFF
--- a/plugins/jabra-gnp/fu-jabra-gnp-device.c
+++ b/plugins/jabra-gnp/fu-jabra-gnp-device.c
@@ -147,6 +147,35 @@ fu_jabra_gnp_device_rx_with_sequence_cb(FuDevice *device, gpointer user_data, GE
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


hughsie (approved):

gdpcastro:
Hi @hughsie , sorry for the maybe silly question: What merge strategy should I use? Squash and merge, or rebase and merge?

hughsie:
> Squash and merge, or rebase and merge?

That's not a silly question; wars have been waged on less. In the general case, rebase and merge please. I only use squash and merge when I want to turn the possibly-tens of commits into one commit. e.g. you might see some PRs with `plugin: nice commit message` and then commits like `fixup` or even `f` -- squash allows you yo squash all the commits into the parent commit rather than cluttering the tree with junk messages.

In this case you could Squash and Merge, and change the commit message to something like `jabra-gnp: Get the device name from the device"

PR #5923, opened 2023-06-21
Read and write EFI DEVICE_PATH structures without libefiboot

hughsie:
This allows us to drop a significant dep and replace it with simpler code that uses higher-level primitives that are easier to fuzz.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


hughsie:
@superm1 this is the same code as in https://github.com/fwupd/fwupd/pull/5918 -- but merging it allows us to dribble in the feature and also let the fuzzer have a go before dropping the dep and making all the build changes.

PR #5926, opened 2023-06-21
trivial: Update SECURITY.md to reflect reality

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -5,22 +5,23 @@ If you have any concerns please let us know.
 
 ## Supported Versions
 
-The `main`, and `1.7.x`, branches are fully supported by the upstream authors.
-Additionally, the `1.5.x` and `1.6.x` branches are supported for security fixes.
+The `main`, and `1.8.x`, branches are fully supported by the upstream authors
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -5,22 +5,23 @@ If you have any concerns please let us know.
 
 ## Supported Versions
 
-The `main`, and `1.7.x`, branches are fully supported by the upstream authors.
-Additionally, the `1.5.x` and `1.6.x` branches are supported for security fixes.
+The `main`, and `1.8.x`, branches are fully supported by the upstream authors
+with all unstable code belonging in `wip/{USERANME}/{TOPIC}` branches.
```
superm1, line 9: Username spelled wrong and I think it's better to just say topic branches.

superm1 (dismissed):
One nit and then looks good

SECURITY.md
hughsie: How about this?

superm1 (approved):

PR #5927, opened 2023-06-21
Update TUXEDO quirks

Matombo:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

Updating quirks for TUXEDO devices:
- TUXEDO InfinityBook S 14/15 Gen6 GUID where based on older DMI strings that didn't actually ship
- New GUIDs for TUXEDO InfinityBook Pro 13 v3 (only for BIOS/flashrom plugin as EC/superio plugin is not yet tested on this device

I noticed that the fn-m-me-unlock flag doesn't actually do anything? At least from the command line. Is the promt to reboot with FN+M pressed only for GUI or am I encountering a bug?

### DIFF
--- a/plugins/flashrom/flashrom.quirk
+++ b/plugins/flashrom/flashrom.quirk
@@ -133,14 +133,16 @@ Plugin = flashrom
 Branch = dasharo
 VersionFormat = triplet
 
-# Tuxedo InfinityBook S14 Gen6 (HwId)
-[6c80d85b-d0b6-5ee2-99d4-ec28dd32febd]
+# TUXEDO InfinityBook S 14 Gen6 (HwId)
+[9366407e-433e-51e9-bd79-a517e3f31536]
 Plugin = flashrom
-[FLASHROM\GUID_6c80d85b-d0b6-5ee2-99d4-ec28dd32febd]
+[FLASHROM\GUID_9366407e-433e-51e9-bd79-a517e3f31536]
 Flags = fn-m-me-unlock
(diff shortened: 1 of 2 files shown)


hughsie:
> I noticed that the fn-m-me-unlock flag doesn't actually do anything

Hmm, that's news to me. Can you debug it a bit and see what you can find please?

hughsie (approved):

Matombo:
> > I noticed that the fn-m-me-unlock flag doesn't actually do anything
> 
> Hmm, that's news to me. Can you debug it a bit and see what you can find please?

Looking in the code, the flag only does something if the region is recognized as a FU_IFD_REGION_ME, but do i interpret the get-device output correctly that the whole chip is recognized as jsut a singular BIOS region?

[get-devices_dump.txt](https://github.com/fwupd/fwupd/files/11837553/get-devices_dump.txt)

hughsie:
> that the whole chip is recognized as just a singular BIOS region

Looks like it, indeed.

Matombo:
can i fix this via quirks or does this need work on the flashrom plugin?

hughsie:
If i'm honest, I don't really understand the IFD flashrom stuff -- sorry

Matombo:
Getting late here now, but as far as I got today: The intel me region is marked as such by the intel-spi plugin it seems, but that seems to be missing in the ubuntu deb and the deb I build myself from the debian docker container. I will dig further to see if I can fix that ^^.

PR #5932, opened 2023-06-23
Ensure that BootService-only variables cannot be read in runtime mode

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


superm1 (approved):
Sure makes sense.

hughsie:
@superm1 I did wonder about what GUIDs to check; I think those two are probably the biggest hitters.

superm1:
> @superm1 I did wonder about what GUIDs to check; I think those two are probably the biggest hitters.

Yeah when I was thinking about it I don't think you'll be able to cover a list of all private guid that might be in use. So this seems pragmatic.

PR #5941, opened 2023-06-26
uefi-dbx: Check only the efi executables from boot menu

hughsie:
Inspired from a patch by Crag Wang <crag.wang@dell.com>, many thanks.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/uefi-dbx/fu-uefi-dbx-common.c
+++ b/plugins/uefi-dbx/fu-uefi-dbx-common.c
@@ -29,13 +29,86 @@ fu_uefi_dbx_get_authenticode_hash(const gchar *fn, GError **error)
 	return g_strdup(fu_efi_image_get_checksum(img));
 }
 
+static GPtrArray *
+fu_uefi_dbx_get_basenames_bootxxxx(void)
+{
+	g_autoptr(GPtrArray) files = g_ptr_array_new_with_free_func(g_free);
+
+	/* hardcoded, as we chainload from shim to these */
+	g_ptr_array_add(files, g_strdup_printf("fwupd%s.efi", EFI_MACHINE_TYPE_NAME));
+	g_ptr_array_add(files, g_strdup_printf("grub%s.efi", EFI_MACHINE_TYPE_NAME));
+
+	for (guint i = 0; i <= G_MAXUINT16; i++) {
+		g_autofree gchar *basename = NULL;
+		g_autofree gchar *basename_down = NULL;
+		g_autofree gchar *fullpath = NULL;
+		g_autofree gchar *name = g_strdup_printf("Boot%04X", i);
+		g_autoptr(FuEfiLoadOption) loadopt = NULL;
+		g_autoptr(FuFirmware) dp_buf = NULL;
+		g_autoptr(FuFirmware) dp_file = NULL;
+		g_autoptr(GBytes) blob = NULL;
+		g_autoptr(GError) error_local = NULL;
+
+		/* load EFI load option */
+		blob = fu_efivar_get_data_bytes(FU_EFIVAR_GUID_EFI_GLOBAL, name, NULL, NULL);
+		if (blob == NULL)
+			continue;
+		loadopt = fu_efi_load_option_new();
+		if (!fu_firmware_parse(FU_FIRMWARE(loadopt),
+				       blob,
+				       FWUPD_INSTALL_FLAG_NONE,
+				       &error_local)) {
+			g_warning("failed to parse %s: %s", name, error_local->message);
+			continue;
+		}
+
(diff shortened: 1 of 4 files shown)


plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -29,13 +29,86 @@ fu_uefi_dbx_get_authenticode_hash(const gchar *fn, GError **error)
...
 
+static GPtrArray *
+fu_uefi_dbx_get_basenames_bootxxxx(void)
+{
+	g_autoptr(GPtrArray) files = g_ptr_array_new_with_free_func(g_free);
+
+	/* hardcoded, as we chainload from shim/grub to these */
+	g_ptr_array_add(files, g_strdup("fwupdx64.efi"));
+	g_ptr_array_add(files, g_strdup("fwupdaa64.efi"));
+	g_ptr_array_add(files, g_strdup("fwupdia32.efi"));
```
superm1, line 40: we know the arch we're running on via `EFI_MACHINE_TYPE_NAME`, use that to build the string.

hughsie:
Right, also my worry. The current code in main just looks at *everything* and that's never going to miss anything. This code would miss anything being chainloaded-to. I assume shim is also respecting the dbx too.

superm1:
> Right, also my worry. The current code in main just looks at _everything_ and that's never going to miss anything. This code would miss anything being chainloaded-to. I assume shim is also respecting the dbx too.

I say build a heuristic of the "common" stuff that is chainloaded.  By looking at that heuristic and all the `BootXXXX` entries you'll get enough stuff I think to prevent common problems.

hughsie:
> I say build a heuristic of the "common" stuff that is chainloaded

My worry is that it makes something that was very definite now very fuzzy.

hughsie:
> parse grub.cfg

That just gave me the shivers. :)

superm1:
> > parse grub.cfg
> 
> That just gave me the shivers. :)

Of course it could be crazy complex but at the end of the day it's really just look for a line with `chainload` in it and then any strings on that line match against files on the partition.

hughsie:
I can see downsides with this patch -- and not a lot of upsides. Is the bug we're really trying to fix basically "ignore stuff left in the ESP from previous Linux installs"

superm1:
> I can see downsides with this patch -- and not a lot of upsides. Is the bug we're really trying to fix basically "ignore stuff left in the ESP from previous Linux installs"

How about we add a config option to ignore selected files in `/etc/fwupd/daemon.conf`

hughsie:
> How about we add a config option to ignore selected files in /etc/fwupd/daemon.conf

I don't think that helps much; if it's an old no-longer used EFI binary the user is better off just deleting it.

superm1:
OK - then I think the remaining options are either:
1. Offer a config option to enable this PR behavior (only match executables in boot entries and extra special case ones - ignore chainloading)
2. Cancel this PR.

I personally lean on the second.  I think it's safer for anything we can't envision about how executables are used.

hughsie:
> Offer a config option to enable this PR behavior

My trouble with this is -- why would the user **need** to ignore something? I guess it's less scary to set something in a config file than to delete something you don't know.

plugins/uefi-dbx/fu-uefi-dbx-common.c
hughsie: Fixed.

hughsie:
Okay, I think a great solution here is to use --force to "downgrade" the checking. How about this?

plugins/uefi-dbx/fu-uefi-dbx-common.c
```diff
@@ -29,13 +29,85 @@ fu_uefi_dbx_get_authenticode_hash(const gchar *fn, GError **error)
 	return g_strdup(fu_efi_image_get_checksum(img));
 }
 
+static GPtrArray *
+fu_uefi_dbx_get_basenames_bootxxxx(void)
+{
+	g_autoptr(GPtrArray) files = g_ptr_array_new_with_free_func(g_free);
+
+	/* hardcoded, as we chainload from shim/grub to these */
+	g_ptr_array_add(files, g_strdup_printf("fwupd%s.efi", EFI_MACHINE_TYPE_NAME));
```
superm1, line 38: maybe we should also hardcode the grub prefixed ones because shim chainloads too.

superm1 (approved):
I think the new approach makes a lot of sense.

plugins/uefi-dbx/fu-uefi-dbx-common.c
hughsie: good idea, added.

superm1 (approved):

PR #5942, opened 2023-06-26
genesys_gl32xx: new plugin for Genesys SD readers

d4s:
New plugin for Genesys SD readers.
    
This plugin allow to update and dump firmware for GL3232 and GL3224 families SD readers.
Supported devices based on flash IDs:
- `PM25LD010` (tested)
- `MX25L1006E` (not tested)
- `GD25D10B` (not tested)

Type of pull request:

- [ X] New plugin (Please include [new plugin checklist](https://github.com/fwupd/fwupd/wiki/New-plugin-checklist))
- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
+++ b/plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
@@ -1,0 +1,850 @@
+/*
+ * Copyright (C) 2023 Denis Pynkin <denis.pynkin@collabora.com>
+ * Copyright (C) 2023 Richard Hughes <richard@hughsie.com>
+ *
+ * SPDX-License-Identifier: LGPL-2.1+
+ */
+
+#include "config.h"
+
+#ifdef HAVE_SCSI_SG_H
+#include <scsi/sg.h>
+#endif
+
+#include "fu-genesys-gl32xx-device.h"
+#include "fu-genesys-gl32xx-firmware.h"
+
+#define FU_GENESYS_GL32XX_FW_START_ADDR	    0x0
+#define FU_GENESYS_GL32XX_FW_SIZE	    0x00010000
+#define FU_GENESYS_GL32XX_CLEAR_WP_SLEEP_MS 800
+
+struct _FuGenesysGl32xxDevice {
+	FuUdevDevice parent_instance;
+	gchar *chip_name;
+	guint32 packetsz;
+};
+
+G_DEFINE_TYPE(FuGenesysGl32xxDevice, fu_genesys_gl32xx_device, FU_TYPE_UDEV_DEVICE)
+
+#define FU_GENESYS_GL32XX_BUFFER_LEN	   32
+#define FU_GENESYS_GL32XX_IOCTL_TIMEOUT_MS 20000
+
+/* based on function from scsi plugin */
+static gboolean
+fu_genesys_gl32xx_device_send_scsi_cmd_v3(FuGenesysGl32xxDevice *self,
+					  const guint8 *cdb,
+					  guint8 cdbsz,
+					  const guint8 *buf,
+					  guint32 bufsz,
+					  FuGenesysGl32xxDirection dir,
+					  GError **error)
+{
+#ifdef HAVE_SCSI_SG_H
+	guint8 sense_buffer[FU_GENESYS_GL32XX_BUFFER_LEN] = {0};
+	struct sg_io_hdr io_hdr = {.interface_id = 'S'};
+	gint rc = 0;
+
+	io_hdr.cmd_len = cdbsz;
+	io_hdr.mx_sb_len = sizeof(sense_buffer);
+
+	switch (dir) {
+	case FU_GENESYS_GL32XX_DIRECTION_IN:
+		io_hdr.dxfer_direction = SG_DXFER_FROM_DEV;
+		break;
+	case FU_GENESYS_GL32XX_DIRECTION_OUT:
+		io_hdr.dxfer_direction = SG_DXFER_TO_DEV;
+		break;
+	case FU_GENESYS_GL32XX_DIRECTION_NO_DATA:
+	default:
+		io_hdr.dxfer_direction = SG_DXFER_NONE;
+		break;
+	}
+
+	io_hdr.dxfer_len = bufsz;
+	io_hdr.dxferp = (guint8 *)buf;
+	/* pointer to command buf */
+	io_hdr.cmdp = (guint8 *)cdb;
+	io_hdr.sbp = sense_buffer;
+	io_hdr.timeout = FU_GENESYS_GL32XX_IOCTL_TIMEOUT_MS;
+	io_hdr.flags = SG_FLAG_DIRECT_IO;
+
+	fu_dump_raw(G_LOG_DOMAIN, "cmd", cdb, cdbsz);
+	if (!fu_udev_device_ioctl(FU_UDEV_DEVICE(self),
+				  SG_IO,
+				  (guint8 *)&io_hdr,
+				  &rc,
+				  5 * FU_GENESYS_GL32XX_IOCTL_TIMEOUT_MS,
+				  error))
+		return FALSE;
+
+	if (io_hdr.status) {
+		g_set_error(error,
+			    G_IO_ERROR,
+			    G_IO_ERROR_FAILED,
+			    "Command fail with status %x, senseKey 0x%02x, asc 0x%02x, ascq 0x%02x",
+			    io_hdr.status,
+			    sense_buffer[2],
+			    sense_buffer[12],
+			    sense_buffer[13]);
+		return FALSE;
+	}
+
+	if (bufsz > 0)
+		fu_dump_raw(G_LOG_DOMAIN, "cmd data", buf, bufsz);
+
+	/* success */
+	return TRUE;
+#else
+	g_set_error_literal(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_NOT_SUPPORTED,
+			    "unsupported as scsi/sg.h not found");
+	return FALSE;
+#endif
+}
+
+static void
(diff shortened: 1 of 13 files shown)


plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,882 @@
...
+
+	g_free(self->chip_name);
+
+	/* check for 3230, 3231, 3232, 3230S, 3231S, 3232S first */
+	cmd[3] = 0xDC;
+	/* ignore error here */
+	if (!fu_genesys_gl32xx_device_check_rom_mode(self, cmd, sizeof(cmd), &error_local)) {
+		g_debug("ignoring: %s", error_local->message);
+	} else {
+		self->chip_name = g_strdup("GL323x");
```
hughsie, line 357: i think it's better to call the static fu_genesys_gl32xx_device_set_chip_name() here.

d4s: ok, let's return it bak

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,882 @@
...
+static gboolean
+fu_genesys_gl32xx_device_ensure_rom_mode(FuGenesysGl32xxDevice *self, GError **error)
+{
+	guint8 cmd[] = {0xE4, 0x01, 0x00, 0x00, 0x04, 0x00};
+	g_autoptr(GError) error_local = NULL;
+
+	g_free(self->chip_name);
+
+	/* check for 3230, 3231, 3232, 3230S, 3231S, 3232S first */
+	cmd[3] = 0xDC;
```
hughsie, line 352: maybe better would be:

     	const guint8 cmd_legacy[] = {0xE4, 0x01, 0x00, 0x03, 0x04, 0x00};
     	const guint8 cmd_3224[] = {0xE4, 0x01, 0x00, 0xFC, 0x04, 0x00};

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,882 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+static GBytes *
+fu_genesys_gl32xx_device_dump_bytes(FuDevice *device, FuProgress *progress, GError **error)
+{
+	FuGenesysGl32xxDevice *self = FU_GENESYS_GL32XX_DEVICE(device);
+	g_autoptr(GPtrArray) chunks = NULL;
```
hughsie, line 617: g_autoptr is normally last in the variable block.

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,882 @@
...
+	} const flash_type[3] = {
+	    {
+		.name = "gd25d10b",
+		.id = {0xC8, 0x10},
+		.idsz = 2,
+		.cmd_req = {0xF3, 0x02, 0x00, 0x04, 0x00, 0x02},
+		.data_req = {0x90, 0x00, 0x00, 0x00},
+		.data_reqsz = 4,
+		.cmd_get = {0xF3, 0x04, 0x00, 0x00, 0x00, 0x02},
+	    }, /* gd25d10b */
```
hughsie, line 390: i don't think you need the /* comment */

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,882 @@
...
+		const gchar *name;
+		guint8 id[3];
+		guint32 idsz;
+		guint8 cmd_req[6];
+		guint8 data_req[4];
+		guint32 data_reqsz;
+		guint8 cmd_get[6];
+	} const flash_type[3] = {
+	    {
+		.name = "gd25d10b",
```
hughsie, line 383: maybe use the uppercase chip name here?

d4s: It's only visible in debug message but ok

hughsie: it's slightly frustrating we can't use https://github.com/fwupd/fwupd/blob/main/data/cfi.quirk for any of these. :/

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,879 @@
...
+		.id = {0xC8, 0x10},
+		.idsz = 2,
+		.cmd_req = {0xF3, 0x02, 0x00, 0x04, 0x00, 0x02},
+		.data_req = {0x90, 0x00, 0x00, 0x00},
+		.data_reqsz = 4,
```
hughsie, line 385: if you change this to 0x90 and `.data_reqsz = 1,` does it still work? I've never seen a 4 byte ID request before.

hughsie (changes requested):

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,882 @@
...
+
+	/* success */
+	return TRUE;
+}
+
+static void
+fu_genesys_gl32xx_device_set_progress(FuDevice *self, FuProgress *progress)
+{
+	fu_progress_set_id(progress, G_STRLOC);
+	fu_progress_add_flag(progress, FU_PROGRESS_FLAG_GUESSED);
```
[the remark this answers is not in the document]
d4s: nope, fixing

hughsie:
@d4s I think you need a manual clang-format to fix CI.

plugins/genesys-gl32xx/genesys-gl32xx.quirk
```diff
@@ -0,0 +1,7 @@
+[BLOCK\VEN_05E3&DEV_0749]
+Plugin = genesys_gl32xx
+FirmwareSize = 0x10000
+
+[BLOCK\VEN_05E3&DEV_0764]
+Plugin = genesys_gl32xx
+FirmwareSize = 0x1C000
```
hughsie, line 7: does the size depend on the *device type* or the *SPI chip*?

d4s: No answer here.
Due implicit signs I think it is for the device, but I'll better check.

d4s:
> I think you need a manual clang-format to fix CI.

indeed. Strange, why the hook missed that

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
d4s: from publicly available manual of `gd25d10b` chip:
_The command is initiated by driving the CS# pin low and shifting the command code “90H” followed by a 24-bit
address (A23-A0) of 000000H._

however it's a bit strange why `90H` is used for `gd25d10b` while the doc shows it is supporting `9F` command as well :thinking:

hughsie: Sometimes 90H returns different data to 9F-- but everything produced in the last 5 years seems to support 9F. If 9F works I'd really like to keep things simple.

plugins/genesys-gl32xx/genesys-gl32xx.quirk
hughsie: Seems device specific; sorry for the noise.

plugins/genesys-gl32xx/fu-genesys-gl32xx-device.c
```diff
@@ -0,0 +1,877 @@
...
+						       error))
+		return FALSE;
+	if (fu_memcmp_safe(int_rom_mode, sizeof(int_rom_mode), data, sizeof(data), NULL)) {
+		fu_device_add_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+		return TRUE;
+	}
+	if (fu_memcmp_safe(ext_rom_mode, sizeof(ext_rom_mode), data, sizeof(data), NULL)) {
+		fu_device_remove_flag(FU_DEVICE(self), FWUPD_DEVICE_FLAG_IS_BOOTLOADER);
+		return TRUE;
+	}
```
hughsie, line 330: Heads up; this API changed in 3b0a55fe4facf3dd3bc6c2eae81f01008a3731d7 -- you'll need to add some `0x0`'s when you next rebase against main. I'd be fine with you merging this now and then doing the fixups as different commits if that helps.

d4s: ooc, why api change instead of adding `fu_memcmp_safe_full()`?

hughsie (approved):
LGTM, we can do any fixups or extra checks in follow up PRs if that's easier.

hughsie:
Ooh! Now we can use cfi.quirk! Give me a few mins and i'll push a fixup, okay?

hughsie (approved):

hughsie:
@d4s if those CFI changes work, please merge. Thanks!

d4s:
> Now we can use cfi.quirk!

@hughsie but with this change literally any flash id from the quirk file is allowed for the device while the initial idea was to limit with allowed ones only.

hughsie:
@d4s from experience, vendors tend to just choose the least expensive SPI chip at manufacturing time. I think it's safer to accept anything. In the chip shortage last year vendors were even adding bigger than required SPI chips as it's all they could get their hands on. Maybe we can check the CFI size is >= 0x20000 as a sanity check?

d4s:
@hughsie ok, let's push as is -- thank you for the fix!
keeping in mind that the F/W sizes are 0x10000 and 0x1C000, I think the size sanity check would be useless

PR #5944, opened 2023-06-27
Only add the little-used _REV instance IDs by request

hughsie:
This reduces the number of calls to fu_quirks_lookup_by_id_iter() by 15% and speeds up engine coldplug by about the same amount.

Add the flag to anything we know depends on the _REV GUIDs, and now we can add flags from LVFS metadata we worry less if we miss one device.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-device.h
+++ b/libfwupdplugin/fu-device.h
@@ -538,6 +538,15 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_MD_ONLY_CHECKSUM (1ull << 31)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_ADD_INSTANCE_ID_REV:
+ *
+ * Add the `_REV` instance ID suffix.
+ *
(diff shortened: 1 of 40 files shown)


libfwupdplugin/fu-device.h
```diff
@@ -538,6 +538,15 @@ typedef guint64 FuDeviceInternalFlags;
  */
 #define FU_DEVICE_INTERNAL_FLAG_MD_ONLY_CHECKSUM (1ull << 31)
 
+/**
+ * FU_DEVICE_INTERNAL_FLAG_ADD_INSTANCE_ID_REV:
+ *
+ * Add the `_REV` instance ID suffix.
+ *
+ * Since: 1.9.1
```
superm1, line 546: 1.9.3

superm1 (approved):

libfwupdplugin/fu-device.h
hughsie: Fixed, thanks.

PR #5945, opened 2023-06-27
Do not use quirk lookups to match udev subsystems to plugins

hughsie:
This speeds up the engine startup by another 35%. We only need to track what plugin wants what udev subsystem and we can avoid 91 lookups for `BLOCK`, 65 for `PCI`, 57 for `PLATFORM`, 26 for `MEI` etc.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-context-private.h
+++ b/libfwupdplugin/fu-context-private.h
@@ -47,9 +47,13 @@ fu_context_get_firmware_gtypes(FuContext *self);
 GType
 fu_context_get_firmware_gtype_by_id(FuContext *self, const gchar *id);
 void
-fu_context_add_udev_subsystem(FuContext *self, const gchar *subsystem);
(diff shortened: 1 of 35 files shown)


superm1 (commented):
I think you should be dropping `FU_QUIRKS_PLUGIN` as well with this change too

hughsie:
> I think you should be dropping FU_QUIRKS_PLUGIN as well with this change too

I think the flashrom plugins use of the quirk is legitimate, no?

superm1:
> > I think you should be dropping FU_QUIRKS_PLUGIN as well with this change too
> 
> I think the flashrom plugins use of the quirk is legitimate, no?

Oh I forgot about that use. Yeah I guess it should stay then.

superm1 (approved):

PR #5954, opened 2023-06-29
Allow filtering by release flags from fwupdtool and fwupdmgr

hughsie:
For instance, filtering on releases that have metadata trust and also have a trusted report can be done using:

    fwupdmgr get-releases --filter-release=trusted-metadata,trusted-report

...this also works when doing `fwupdmgr update --filter-release=trusted-report`

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/data/bash-completion/fwupdmgr
+++ b/data/bash-completion/fwupdmgr
@@ -67,6 +67,7 @@ _fwupdmgr_opts=(
 	'--show-all'
 	'--sign'
 	'--filter'
+	'--filter-release'
 	'--disable-ssl-strict'
 	'--ipfs'
 	'--json'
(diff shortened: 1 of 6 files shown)


hughsie:
@superm1 we could also export `gboolean fwupd_release_match_flags(FwupdRelease *rel, FwupdReleaseFlags include, FwupdReleaseFlags exclude)` and `GPtrArray *fwupd_release_array_filter_flags(GPtrArray *rels, FwupdReleaseFlags include, FwupdReleaseFlags exclude)` if you think that makes sense.

hughsie:
> we could also export

This tidied a lot of things up too. Comments very welcome on the method names.

superm1 (approved):

superm1:
I think it's great as is.

PR #5957, opened 2023-06-30
uefi-capsule: Do not install a bricking shim

### LINKED ISSUE #5803: candidate snap doesn't validate enclosed shim is safe to install
Installing a firmware update from confined snap can break secure boot. This is because we are installing a shim that is potentially newer than in the distro. That may revoke old bootloaders (grub) and brick the current OS installation. The only way to get to boot it is to clear shim UEFI variables or manually update the bootloader offline.

Proposed solution:
```
              We can read section `.sbatlevel` of the shim we want to install. https://github.com/rhboot/shim/blob/main/sbat_var.S
Then we should have at least `SBAT_VAR_PREVIOUS`, I think.
```

_Originally posted by @valentindavid in https://github.com/fwupd/fwupd/issues/5508#issuecomment-1513340999_

valentindavid:
The snap contains a signed shim to install uefi capsules. However if this shim is newer than the host shim (or grub), it may brick the current OS to boot by revoking its version of the shim.

Now in the snap we keep track of the previous sbatlevel of the shim. This is what the shim will install to the sbatlevel on the machine. We compare it to the current value of the machine. If this the shim has any new or newer component, that means booting that shim would update the current sbatlevel. In this case we fail and prevent installation.

Fixes #5803

Type of pull request:

- [*] Code fix
- [ ] Feature
- [ ] Documentation


valentindavid:
@superm1 I promised to do it. Sorry I did not take the time to do that before.

snap/snapcraft.yaml
```diff
@@ -117,6 +117,11 @@ parts:
               ${CRAFT_PART_INSTALL}/share/dbus-1/system-services/org.freedesktop.fwupd.service
       cp -R /usr/libexec/fwupd/efi ${CRAFT_PART_INSTALL}/libexec/fwupd
       cp /usr/lib/shim/shimx64.efi.signed ${CRAFT_PART_INSTALL}/libexec/fwupd/efi/
+
+      objcopy -O binary -j .sbatlevel /usr/lib/shim/shimx64.efi.signed sbatlevel.section
+      python3 "${CRAFT_PROJECT_DIR}/snap/local/get_sbat_var_previous.py" sbatlevel.section >shimx64.efi.signed.sbatlevel
+      cp shimx64.efi.signed.sbatlevel /usr/lib/shim/shimx64.efi.signed ${CRAFT_PART_INSTALL}/libexec/fwupd/efi/
```
hughsie, line 123: this seems kinda fragile. Could the fwupd code use libelf to get the specific section data?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,189 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	/*
+	 * For every new shim entry, we need a matching entry in the
+	 * current sbatlevel. That is the entry of the shim is not
+	 * newer than current sbatlevel.
+	 *
+	 * The opposite way might work (for example shim's latest
+	 * sbatlevel matches) or not (shim is too old), but it will
+	 * not brick the current OS.
+	 */
+	for (size_t shim_entry = 0; sbatlevel_content[shim_entry];) {
```
hughsie, line 309: can we maybe make this a lot safer and split by line, and then split each line by comma? There's a lot of scary pointer accesses going on.

hughsie (changes requested):

snap/snapcraft.yaml
valentindavid: If it is ok to add dependencies, yes.

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: OK, will do that.

snap/snapcraft.yaml
hughsie: Maybe; ELF isn't super tricky to parse. Give me a few hours and I'll see what I can do.

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: This is done.

snap/snapcraft.yaml
valentindavid: I can do it with just the linux headers (`linux/elf.h`). It is not difficult. It is just a bit annoying.

hughsie: Hang on -- isn't the fwupdx64.efi going to be in PE format? we already have a parser for that.

valentindavid: I forgot, it is a PE/COFF....

valentindavid: So probably with `linux/coff.h`.

valentindavid: Nice. I will look at that.

hughsie: Nah! You can do `fu_pefile_firmware_new()`, then `fu_firmware_parse()`, then `fu_firmware_get_image_by_id_bytes(.sbat)`

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_FILE,
+				    "shim sbatlevel for %s has a bricking update for entry %s "
+				    "(missing entry in current UEFI variable)",
+				    source_shim,
+				    shim_entry);
+
+			return FALSE;
+		} else if (*current_version < *shim_version) {
```
hughsie, line 377: you don't need the else here

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+				(guint8 **)&current_sbatlevel,
+				&current_sbatlevel_size,
+				NULL,
+				NULL))
+		return FALSE;
+
+	current_hash = fu_uefi_bootmgr_parse_csv(current_sbatlevel, current_sbatlevel_size, error);
+	if (current_hash == NULL) {
+		return FALSE;
+	}
```
hughsie, line 347: no need for braces

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+		version = g_ascii_strtoll(columns[1], &endptr, 10);
+		if ((endptr == columns[1]) || (*endptr != '\0')) {
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_FILE,
+				    "invalid version in sbat level for entry %s",
+				    columns[0]);
+		}
+		version_bucket = g_malloc(sizeof(gint64));
+		g_assert(version_bucket != NULL);
```
hughsie, line 295: no assert -- if malloc fails we're sunk well before this -- also maybe `g_new0()`?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+		columns = g_strsplit(lines[i], ",", 3);
+		if (g_strv_length(columns) < 2) {
+			g_set_error(error,
+				    FWUPD_ERROR,
+				    FWUPD_ERROR_INVALID_FILE,
+				    "invalid CSV data: not enough columns");
+			return NULL;
+		}
+
+		version = g_ascii_strtoll(columns[1], &endptr, 10);
```
hughsie, line 286: Can you use fu_strtoull here?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	for (guint i = 0; lines[i] != NULL; i++) {
+		gchar *endptr = NULL;
+		g_auto(GStrv) columns = NULL;
+		gint64 version;
+		g_autofree gint64 *version_bucket = NULL;
+
+		if ((lines[i][0] == '\0') && (lines[i + 1] == NULL)) {
+			/* last line may be empty */
+			break;
+		}
```
hughsie, line 275: maybe just ignore all empty lines?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	g_auto(GStrv) lines = NULL;
+	g_autoptr(GHashTable) hash = NULL;
+	g_autofree gchar *nt_data = NULL;
+
+	hash = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, g_free);
+
+	/* null terminate the file */
+	nt_data = g_strndup(data, size);
+
+	lines = g_strsplit(nt_data, "\n", -1);
```
hughsie, line 264: you can make this a bit simpler with fu_strsplit

snap/snapcraft.yaml
hughsie: In case you did need ELF for some strange reason, I just rather enjoyed the last hour :) https://github.com/fwupd/fwupd/pull/5959

plugins/uefi-capsule/fu-uefi-bootmgr.c
hughsie: hang on-- the version can be a semver no? e.g. fwupd has `1.9.2` -- i think you should store the string rather than try an parse as a uint -- and then use fu_version_compare() to compare them.

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	gsize sbatlevel_size = 0;
+	g_autofree gchar *current_sbatlevel = NULL;
+	gsize current_sbatlevel_size = 0;
+	g_autoptr(GHashTable) shim_hash = NULL;
+	g_autoptr(GHashTable) current_hash = NULL;
+	GHashTableIter iter;
+	gpointer key, value;
+
+	sbatlevel_path = g_strdup_printf("%s.sbatlevel", source_shim);
+	if (!g_file_test(sbatlevel_path, G_FILE_TEST_EXISTS)) {
```
superm1, line 317: Once you start natively parsing the shim I think this block can go; the checks will happen even in traditional distros.

snap/local/get_sbat_var_previous.py
```diff
@@ -0,0 +1,31 @@
+import sys
```
superm1, line 1: This file is missing a shebang and header.

I do think even with moving parsing into fwupd you can keep this script in contrib as it could be useful for debugging purposes.

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: Actually I should call it "generation" rather than version. That is the correct name. It cannot be a semantic version. It is a number.

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: Done.

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: Done.

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,147 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	if (!g_file_get_contents(sbatlevel_path, &sbatlevel_content, &sbatlevel_size, error)) {
+		return FALSE;
+	}
+
+	shim_hash = fu_uefi_bootmgr_parse_csv(sbatlevel_content, sbatlevel_size, error);
+	if (shim_hash == NULL) {
+		return FALSE;
+	}
+
+	/* Not safe if variable is not set but new shim would set it */
```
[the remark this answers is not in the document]
valentindavid: Done

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: Done

snap/snapcraft.yaml
valentindavid: Done. But I had to fix the long names. It is `.sbatlevel` section we need to read. not `.sbat`.

snap/local/get_sbat_var_previous.py
valentindavid: File was removed

libfwupdplugin/fu-pefile-firmware.c
```diff
@@ -118,28 +139,82 @@ fu_pefile_firmware_parse(FuFirmware *firmware,
...
+
+			sect_id = fu_strsafe((const gchar *)buf + string_table + str_idx,
+					     bufsz - string_table);
+			if (sect_id == NULL) {
+				g_set_error_literal(error,
+						    FWUPD_ERROR,
+						    FWUPD_ERROR_INVALID_FILE,
+						    "no section name");
+				return FALSE;
+			}
```
hughsie, line 175: Is this COFF? I didn't see it https://learn.microsoft.com/en-us/windows/win32/debug/pe-format here

hughsie: Ohh it's PE COFF -- i just wondered where you got that spec. It might be correct!

libfwupdplugin/fu-pefile-firmware.c
```diff
@@ -118,33 +140,89 @@ fu_pefile_firmware_parse(FuFirmware *firmware,
...
+					return FALSE;
+				}
+			}
 		}
 
 		/* create new firmware */
 		if (g_strcmp0(sect_id, ".sbom") == 0) {
 			fw_sect = fu_coswid_firmware_new();
+		} else if (g_strcmp0(sect_id, ".sbatlevel") == 0) {
+			fw_sect = fu_sbatlevel_section_firmware_new();
```
hughsie, line 225: I don't have any files with a .sbatlevel section -- i only have .sbat -- why?

valentindavid: On the shim binary? Probably because you have an old one. `.sbatlevel` is only on shim (as well as `.sbat`). Other binaries have `.sbat` only. Older versions of shim should not have `.sbatlevel`. This was introduced last year.

valentindavid: It is only for introspection. Originally to calculate PCR values I think. But the "previous" entry gives us what shim will install as sbatlevel. Which is the value we want to compare to. `.sbat` will usually be newer that the current sbatlevel.

valentindavid: Also note that this PR allows installation if shim has no `.sbatlevel`. Because it will not brick the system.

hughsie: Hmm, on Fedora shim I don't see it at all:

	[hughsie@hughsie-work build (wip/hughsie/pefile %)]$ ninja src/fwupdtool && sudo ./src/fwupdtool firmware-parse /boot/efi/EFI/fedora/shimx64.efi pefile 
	ninja: no work to do.
	[sudo] password for hughsie: 
	Loading…                 [************************************** ]
	<firmware gtype="FuPefileFirmware">
	  <data size="0xe7218">MZ.</data>
	  <firmware>
	    <id>.eh_frame</id>
	    <offset>0x400</offset>
	    <size>0x1db1c</size>
	    <data size="0x1db1c">
	    </data>
	  </firmware>
	  <firmware>
	    <id>.text</id>
	    <idx>0x1</idx>
	    <offset>0x1e000</offset>
	    <size>0x5e663</size>
	    <data size="0x5e663">H...QRH.=....H.5</data>
	  </firmware>
	  <firmware>
	    <id>.reloc</id>
	    <idx>0x2</idx>
	    <offset>0x7c800</offset>
	    <size>0xa</size>
	    <data size="0xa">
	    </data>
	  </firmware>
	  <firmware>
	    <id>.data.ident</id>
	    <idx>0x3</idx>
	    <offset>0x7ca00</offset>
	    <size>0x49</size>
	    <data size="0x49">UEFI SHIM.$Versi</data>
	  </firmware>
	  <firmware>
	    <id>.data</id>
	    <idx>0x4</idx>
	    <offset>0x7cc00</offset>
	    <size>0x2d5b4</size>
	    <data size="0x2d5b4">S</data>
	  </firmware>
	  <firmware>
	    <id>.vendor_cert</id>
	    <idx>0x5</idx>
	    <offset>0xaa200</offset>
	    <size>0x37c</size>
	    <data size="0x37c">l.</data>
	  </firmware>
	  <firmware>
	    <id>.dynamic</id>
	    <idx>0x6</idx>
	    <offset>0xaa600</offset>
	    <size>0x100</size>
	    <data size="0x100">
	    </data>
	  </firmware>
	  <firmware>
	    <id>.rela</id>
	    <idx>0x7</idx>
	    <offset>0xaa800</offset>
	    <size>0x1b468</size>
	    <data size="0x1b468">
	    </data>
	  </firmware>
	  <firmware>
	    <id>.sbat</id>
	    <idx>0x8</idx>
	    <offset>0xc5e00</offset>
	    <size>0xdf</size>
	    <data size="0xdf">sbat,1,SBAT Vers</data>
	  </firmware>
	</firmware>

libfwupdplugin/fu-pefile-firmware.c
hughsie: Ahh, so this alternate zeros encoding is from https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#coff-symbol-table -- but that's only for the symbol table, not the section table. I'll fix things up.

libfwupdplugin/fu-pefile-firmware.c
valentindavid: I have checked. Fedora 38 has shim 15.6. Both .sbatlevel and the new revocations have been added to 15.7. shim 15.6 is always safe to install (might not boot though).

So what you see is expected. And this branch will not prevent from installing shim from Fedora 38.

hughsie: Aha, gotcha. Mind if I convert your parsing to a generic `FuCsvFirmware` object and then we layer on top of that too? I think it'll be useful in other places too.

valentindavid: Sure. We could even use that for `.sbat` if one day we want to test a binary is revoked by current sbatlevel.

hughsie: Exactly my thoughts too.

valentindavid: Should I convert the .sbatlevel parsing to use rustgen generated headers?

hughsie: sure, makes sense.

valentindavid: (the part that extract the "previous" and "latest" entries, I mean, not csv parsing)

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,162 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
 	return TRUE;
 }
 
+static GHashTable *
+fu_uefi_bootmgr_parse_csv(const gchar *data, gsize size, GError **error)
```
hughsie, line 253: can we use FuCsvFirmware to parse here?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,162 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	shim = fu_pefile_firmware_new();
+	blob = fu_bytes_get_contents(source_shim, error);
+	if (!blob)
+		return FALSE;
+	if (!fu_firmware_parse(shim, blob, FWUPD_INSTALL_FLAG_NONE, error))
+		return FALSE;
+
+	sbatlevel_section = fu_firmware_get_image_by_id(shim, ".sbatlevel", &error_local);
+	if (sbatlevel_section == NULL) {
+		g_debug("No sbatlevel section was found.");
```
hughsie, line 328: lowercase text with no dot at the end please

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,162 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+		g_propagate_error(error, g_steal_pointer(&error_local));
+	}
+
+	previous_sbatlevel = fu_firmware_get_image_by_id(sbatlevel_section, "previous", error);
+	if (previous_sbatlevel == NULL)
+		return FALSE;
+
+	sbatlevel_bytes = fu_firmware_get_bytes(previous_sbatlevel, error);
+	if (sbatlevel_bytes == NULL)
+		return FALSE;
```
hughsie, line 341: can you use `fu_firmware_get_image_by_id_bytes` here?

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,162 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	sbatlevel_content = g_bytes_get_data(sbatlevel_bytes, &sbatlevel_size);
+
+	shim_hash = fu_uefi_bootmgr_parse_csv(sbatlevel_content, sbatlevel_size, error);
+	if (shim_hash == NULL)
+		return FALSE;
+
+	/* not safe if variable is not set but new shim would set it */
+	if (!fu_efivar_get_data(FU_EFIVAR_GUID_SHIM,
+				"SbatLevelRT",
+				(guint8 **)&current_sbatlevel,
```
hughsie, line 352: do you need the cast?

libfwupdplugin/fu-sbatlevel-section.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2023 Canonical Ltd
+// SPDX-License-Identifier: LGPL-2.1+
+
+#[derive(Parse, Validate)]
+struct SbatLevelSectionHeader {
+    version: u32: const=0x0,
+    previous: u32: default=0x0,
+    latest: u32: default=0x0,
```
hughsie, line 8: helpfully i've changed the format in https://github.com/fwupd/fwupd/commit/1d16c78c912f504fd0d0e371f4e63fe67132e0b5 -- const is ` == ` default is ` =` - and you don't need a default of 0x0 as that's the _default default_ anyway :)

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,124 @@
...
+	gsize header_offset = offset + FU_STRUCT_SBAT_LEVEL_SECTION_HEADER_OFFSET_PREVIOUS;
+	guint32 previous_addr;
+	guint32 latest_addr;
+	g_autoptr(GByteArray) st = NULL;
+
+	buf = g_bytes_get_data(fw, &bufsz);
+
+	if (!fu_struct_sbat_level_section_header_validate(buf, bufsz, offset, error)) {
+		return FALSE;
+	}
```
hughsie, line 71: validate is done by parse automatically, so delete pls

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,124 @@
...
+
+	st = fu_struct_sbat_level_section_header_parse(buf, bufsz, offset, error);
+	if (st == NULL) {
+		return FALSE;
+	}
```
hughsie, line 76: no braces here

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,124 @@
...
+			       GError **error)
+{
+	g_autoptr(FuFirmware) entry_fw = NULL;
+	g_autoptr(GBytes) entry_blob = NULL;
+	gsize entry_size;
+
+	for (entry_size = 0; ((entry_offset + entry_size) < bufsz); ++entry_size) {
+		if (buf[entry_offset + entry_size] == 0)
+			break;
+	}
```
hughsie, line 36: i don't really get what this is doing

libfwupdplugin/fu-pefile-firmware.c
```diff
@@ -80,6 +81,8 @@ fu_pefile_firmware_parse_section(FuFirmware *firmware,
 	/* create new firmware */
 	if (g_strcmp0(sect_id, ".sbom") == 0) {
 		img = fu_coswid_firmware_new();
+	} else if (g_strcmp0(sect_id, ".sbatlevel") == 0) {
+		img = fu_sbatlevel_section_firmware_new();
```
hughsie, line 85: i think this is a bit verbose. Maybe fu_sbatlevel_firmware_new() or fu_sbatlevel_section_new() perhaps?

valentindavid: The `.sbatlevel` section contains 2 sbatlevel. I will call it `fu_sbatlevel_section_new`.

libfwupdplugin/fu-sbatlevel-section-firmware.c
valentindavid: Looking for the null character terminator.

plugins/uefi-capsule/fu-uefi-bootmgr.c
valentindavid: This part had to be rewritten.

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,151 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	if (generation_str == NULL) {
+		g_set_error(error,
+			    FWUPD_ERROR,
+			    FWUPD_ERROR_INVALID_FILE,
+			    "missing component generation");
+
+		return FALSE;
+	}
+
+	if (!fu_strtoull(generation_str, generation, 0, G_MAXUINT, error)) {
```
hughsie, line 267: maybe just `return fu_strtoull(......` here

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,151 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	g_autoptr(FuFirmware) previous_sbatlevel = NULL;
+	g_autoptr(FuFirmware) current_sbatlevel = NULL;
+	g_autoptr(GError) error_local = NULL;
+	g_autoptr(GBytes) blob = NULL;
+	g_autoptr(GBytes) sbatlevel_bytes = NULL;
+	g_autoptr(GPtrArray) shim_entries = NULL;
+
+	shim = fu_pefile_firmware_new();
+	blob = fu_bytes_get_contents(source_shim, error);
+	if (!blob)
```
hughsie, line 290: == NULL for pointers please

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,151 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+	previous_sbatlevel = fu_firmware_get_image_by_id(sbatlevel_section, "previous", error);
+	if (previous_sbatlevel == NULL)
+		return FALSE;
+
+	shim_entries = fu_firmware_get_images(previous_sbatlevel);
+
+	current_sbatlevel_bytes =
+	    fu_efivar_get_data_bytes(FU_EFIVAR_GUID_SHIM, "SbatLevelRT", NULL, error);
+	if (current_sbatlevel_bytes == NULL) {
+		/* not safe if variable is not set but new shim would set it */
```
hughsie, line 313: can you move the comment to before `current_sbatlevel_bytes =` and remove the braces pls

plugins/uefi-capsule/fu-uefi-bootmgr.c
```diff
@@ -249,6 +249,151 @@ fu_uefi_setup_bootnext_with_loadopt(FuEfiLoadOption *loadopt, GError **error)
...
+			g_prefix_error(error,
+				       "sbatlevel entry %s for shim %s : ",
+				       entry_id,
+				       source_shim);
+			return FALSE;
+		}
+		if (!fu_uefi_bootmgr_get_generation(current_entry, &current_generation, error)) {
+			g_prefix_error(error, "current sbatlevel entry %s: ", entry_id);
+			return FALSE;
+		}
```
hughsie, line 379: maybe we could have a `guint64 fu_csv_entry_get_value_by_column_id_uint64(FuCsvEntry *self, const gchar *column_id)` instead of this?

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,125 @@
...
+fu_sbatlevel_section_firmware_class_init(FuSbatlevelSectionFirmwareClass *klass)
+{
+	FuFirmwareClass *klass_firmware = FU_FIRMWARE_CLASS(klass);
+
+	klass_firmware->parse = fu_sbatlevel_section_firmware_parse;
+}
+
+FuFirmware *
+fu_sbatlevel_section_new(void)
+{
```
hughsie, line 123: i think drop the _FIRMWARE, Firmware and -firmware everywhere.

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,125 @@
...
+				    GError **error)
+{
+	const guint8 *buf = NULL;
+	gsize bufsz;
+	gsize header_offset = offset + FU_STRUCT_SBAT_LEVEL_SECTION_HEADER_OFFSET_PREVIOUS;
+	guint32 previous_addr;
+	guint32 latest_addr;
+	g_autoptr(GByteArray) st = NULL;
+
+	buf = g_bytes_get_data(fw, &bufsz);
```
hughsie, line 73: I think move this up `const guint8 *buf = g_bytes_get_data(fw, &bufsz);` -- you'll have to have `gsize bufsz = 0;` before of  course

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,125 @@
...
+#include "fu-sbatlevel-section-firmware.h"
+#include "fu-sbatlevel-section-struct.h"
+
+G_DEFINE_TYPE(FuSbatlevelSectionFirmware, fu_sbatlevel_section_firmware, FU_TYPE_FIRMWARE);
+
+static gboolean
+fu_sbatlevel_section_add_entry(FuFirmware *firmware,
+			       GBytes *fw,
+			       const guint8 *buf,
+			       gsize bufsz,
```
hughsie, line 23: you don't need these if you have `fw` -- just do `const guint8 *buf = g_bytes_get_data(fw, &bufsz);` again

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,125 @@
...
+			       gsize entry_offset,
```
hughsie, line 24: maybe just `offset` if it's an offset into `fw`

libfwupdplugin/fu-sbatlevel-section-firmware.c
```diff
@@ -0,0 +1,125 @@
...
+	g_autoptr(GByteArray) st = NULL;
+
+	buf = g_bytes_get_data(fw, &bufsz);
+
+	st = fu_struct_sbat_level_section_header_parse(buf, bufsz, offset, error);
+	if (st == NULL)
+		return FALSE;
+
+	previous_addr = fu_struct_sbat_level_section_header_get_previous(st);
+	latest_addr = fu_struct_sbat_level_section_header_get_latest(st);
```
hughsie, line 80: i think move this line to just above the 2nd fu_sbatlevel_section_add_entry block -- in fact you could just use fu_struct_sbat_level_section_header_get_latest() directly rather than a temp variable.

valentindavid: I have moved it. The name looked long when inline.

libfwupdplugin/fu-csv-entry.c
```diff
@@ -91,6 +91,16 @@ fu_csv_entry_get_value_by_column_id(FuCsvEntry *self, const gchar *column_id)
...
 }
 
+gboolean
+fu_csv_entry_get_value_by_column_id_uint64(FuCsvEntry *self, const gchar *column_id, guint64 *value)
+{
+	const gchar *str_value = fu_csv_entry_get_value_by_column_id(self, column_id);
+
+	g_return_val_if_fail(str_value != NULL, FALSE);
+
+	return fu_strtoull(str_value, value, 0, G_MAXUINT64, NULL);
```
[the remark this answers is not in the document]
valentindavid: Done

hughsie (approved):

superm1 (approved):

PR #5963, opened 2023-07-03
Add fu_memstrsafe() as a slightly safer fu_strsafe()

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-mem.c
+++ b/libfwupdplugin/fu-mem.c
@@ -11,6 +11,7 @@
 #include "fwupd-error.h"
 
 #include "fu-mem-private.h"
+#include "fu-string.h"
 
 /**
  * fu_memwrite_uint16:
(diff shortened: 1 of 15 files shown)


hughsie:
@superm1 i need to use the untrusted offset in future code. better names welcome on the method.

hughsie:
@valentindavid i think this will make our lives easier.

libfwupdplugin/fu-mem.c
```diff
@@ -941,3 +942,38 @@ fu_memwrite_uint64_safe(guint8 *buf,
...
+ *
+ * Since: 1.9.3
+ **/
+gchar *
+fu_memstrsafe(const guint8 *buf, gsize bufsz, gsize offset, gsize maxsz, GError **error)
+{
+	g_autofree gchar *str = NULL;
+
+	g_return_val_if_fail(buf != NULL, NULL);
+	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
```
superm1, line 966: do you want a check that `maxsz` isn't bigger than `bufsz`?

superm1 (approved):

libfwupdplugin/fu-mem.c
hughsie: I think setting the GError from fu_memchk_read() is probably better than the critical warning -- the general idea is that you can just throw any fuzzing random values into fu_memstrsafe() and get a failure rather than a critical warning.

PR #5964, opened 2023-07-04
Load PE/COFF long section names from string table

hughsie:
Also, use rustgen to validate PE files.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-pefile-firmware.c
+++ b/libfwupdplugin/fu-pefile-firmware.c
@@ -10,9 +10,9 @@
 
 #include "fu-bytes.h"
 #include "fu-coswid-firmware.h"
-#include "fu-dump.h"
 #include "fu-mem.h"
 #include "fu-pefile-firmware.h"
+#include "fu-pefile-struct.h"
 #include "fu-string.h"
 
 /**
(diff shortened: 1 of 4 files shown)


hughsie:
@valentindavid I've invited you to the project, can you review please.

libfwupdplugin/fu-pefile-firmware.c
```diff
@@ -62,124 +103,49 @@ fu_pefile_firmware_parse(FuFirmware *firmware,
...
+	if (st_coff == NULL)
 		return FALSE;
-	if (signature != 0x4550) {
-		g_set_error_literal(error,
-				    FWUPD_ERROR,
-				    FWUPD_ERROR_INVALID_FILE,
-				    "invalid signature for file");
-		return FALSE;
-	}
 	hdr_offset += 4;
```
valentindavid, line 120: I think that offset has become more confusing. We add 4 and `FU_STRUCT_PE_COFF_HEADER64_OFFSET_SIZE_OF_OPTIONAL_HEADER`. But what we want is the offset after the coff header (before optional header)

libfwupdplugin/fu-pefile.rs
```diff
@@ -0,0 +1,147 @@
...
+struct PeCoffHeader64 {
+    signature: u32le: const=0x4550, // "PE\0\0"
+    machine: PeCoffMachine: default=0x8664,
+    number_of_sections: u16le: default=0,
+    _time_date_stamp: u32le: default=0,
+    pointer_to_symbol_table: u32le: default=0,
+    number_of_symbols: u32le: default=0,
+    size_of_optional_header: u16le: default=0xf0,
+    characteristics: u16le: default=0x2022,
+    magic: PeCoffMagic: default=0x020b,
```
valentindavid, line 95: I would split the structure here to make it less confusing. That would also make `FU_STRUCT_PE_COFF_HEADER64_SIZE` the right size to add to `hdr_offset` (see my other comment).

Also in theory, some files do not have the optional header. So if one day we need that (though unlikely), it will be less confusing.

valentindavid (commented):
I looks good. I have double checked the format. And I have manually tested with #5957. I have just some comments about the way the offset to section headers is calculated. See my comments in the code.

libfwupdplugin/fu-pefile-firmware.c
hughsie: Right, agreed. Fixed.

libfwupdplugin/fu-pefile.rs
hughsie: Good idea, split. Is this version easier to understand?

valentindavid: Perfect!

valentindavid (approved):

PR #5990, opened 2023-07-12
cfu: Provide a script to convert .inf entries to a .quirk block

hughsie:
Type of pull request:

- [ ] Code fix
- [X] Feature
- [X] Documentation

### DIFF
--- a/contrib/cfu-inf-to-quirk.py
+++ b/contrib/cfu-inf-to-quirk.py
@@ -1,0 +1,56 @@
+#!/usr/bin/python3
+#
+# Copyright (C) 2017 Richard Hughes <richard@hughsie.com>
+#
+# SPDX-License-Identifier: LGPL-2.1+
+#
+# pylint: disable=invalid-name,missing-docstring
+
+import sys
+
+
+def _convert_inf_to_quirk(fn: str) -> None:
+    with open(fn, "rb") as f:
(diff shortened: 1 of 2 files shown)


plugins/cfu/inf-to-quirk.py
superm1, line 1: Maybe this should be in `contrib/`?

plugins/cfu/inf-to-quirk.py
```diff
@@ -0,0 +1,56 @@
...
+            if line.find(inf_key) != -1:
+                data[quirk_key] = line.split(",")[4].strip()
+    if is_cfu:
+        if comment:
+            print(f"# {comment}")
+        print(f"[{instance_id}]")
+        for key, value in data.items():
+            print(f"{key} = {value}")
+    else:
+        print("this is not a CFU firmware.inf")
```
superm1, line 52:
```suggestion
        sys.exit("this is not a CFU firmware.inf")
```

superm1 (dismissed):

plugins/cfu/inf-to-quirk.py
hughsie: maybe renaming it contrib/cfu-inf-to-quirk.py?

superm1: Sure just update readme for new location and name

PR #6007, opened 2023-07-17
Add a triplet-minor version format

hughsie:
Because, of course, the CFU protocol uses a unique version encoding.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation


libfwupd/fwupd-enums.h
```diff
@@ -1077,6 +1077,7 @@ typedef enum {
  * @FWUPD_VERSION_FORMAT_SURFACE:		Microsoft Surface 8b.16b.8b
  * @FWUPD_VERSION_FORMAT_DELL_BIOS:		Dell BIOS BB.CC.DD style
  * @FWUPD_VERSION_FORMAT_HEX:			Hexadecimal 0xAABCCDD style
+ * @FWUPD_VERSION_FORMAT_TRIPLET_MINOR:		CFU-style AA.BBCC.DD version numbers
```
superm1, line 1080: Is this not the same as `FWUPD_VERSION_FORMAT_SURFACE`?

hughsie: Ohh yes! Good catch, thanks!

PR #6008, opened 2023-07-17
trivial: Parse the CFU firwmare as surface verfmt

hughsie:
Type of pull request:

- [ ] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-cfu-offer.c
+++ b/libfwupdplugin/fu-cfu-offer.c
@@ -13,6 +13,7 @@
 #include "fu-cfu-offer.h"
 #include "fu-common.h"
 #include "fu-string.h"
+#include "fu-version-common.h"
 
 /**
  * FuCfuOffer:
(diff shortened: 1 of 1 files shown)


superm1 (approved):

superm1:
When we have next ABI bump can we add to list to rename all these to something more descriptive rather than the first time they were used?

Ie VERSION_FORMAT_AA_BBCC_DD

hughsie:
> When we have next ABI bump can we add to list to rename all these to something more descriptive

Yes, I think so too. Some are more tricky as they have bitshifts.

hughsie:
I'm not against it -- I just struggled to some up with anything good last time. e.g. you'd need `FWUPD_VERSION_FORMAT_AAAPLUS11_BBBBB_CCCCCCCC_DDDDDDDDDDDDDDDD`

superm1:
> I'm not against it -- I just struggled to some up with anything good last time. e.g. you'd need `FWUPD_VERSION_FORMAT_AAAPLUS11_BBBBB_CCCCCCCC_DDDDDDDDDDDDDDDD`

How about just for the easy ones then?

PR #6009, opened 2023-07-17
Add some compatibility version format definitions

superm1:
Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-enums.h
+++ b/libfwupd/fwupd-enums.h
@@ -1102,6 +1102,44 @@ typedef enum {
 } FwupdVersionFormat;
 
 /**
+ * FWUPD_VERSION_FORMAT_AABB_CCDD:
+ *
+ * A "pair" version format where a word is split into two sections
+ */
+#define FWUPD_VERSION_FORMAT_AABB_CCDD FWUPD_VERSION_FORMAT_PAIR
+
+/**
+ * FWUPD_VERSION_FORAMT_AA_BB_CC_DD:
+ *
+ * A "quad" version format where a word is split into four sections
+ */
+#define FWUPD_VERSION_FORAMT_AA_BB_CC_DD FWUPD_VERSION_FORMAT_QUAD
+
+/**
+ * FWUPD_VERSION_FORMAT_AA_BBCC_DD:
+ *
+ * A "triplet" version format where a word is split into three sections
+ * and the two middle sections are combined.
+ */
+#define FWUPD_VERSION_FORMAT_AA_BBCC_DD FWUPD_VERSION_FORMAT_SURFACE
+
+/**
+ * FWUPD_VERSION_FORMAT_AA_BB_CCDD:
+ *
+ * A "triplet" version format where a word is split into three sections
+ * and the two final sections are combined.
+ */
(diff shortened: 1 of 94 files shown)


libfwupd/fwupd-enums.h
```diff
@@ -1101,6 +1101,44 @@ typedef enum {
...
+ * A "pair" version format where a word is split into two sections
+ */
+#define FWUPD_VERSION_FORMAT_AABB_CCDD FWUPD_VERSION_FORMAT_PAIR
+
+/**
+ * FWUPD_VERSION_FORAMT_AA_BB_CC_DD:
+ *
+ * A "quad" version format where a word is split into four sections
+ */
+#define FWUPD_VERSION_FORAMT_AA_BB_CC_DD FWUPD_VERSION_FORMAT_QUAD
```
hughsie, line 1116: FORMAT

libfwupd/fwupd-enums.h
```diff
@@ -1101,6 +1101,44 @@ typedef enum {
...
+ */
+#define FWUPD_VERSION_FORMAT_AA_BB_CCDD FWUPD_VERSION_FORMAT_TRIPLET
+
+/**
+ * FWUPD_VERSION_FORMAT_AABBCCDD:
+ *
+ * A "number" version format where a word is rendered as a hexadecimal
+ * number.
+ */
+#define FWUPD_VERSION_FORMAT_AABBCCDD FWUPD_VERSION_FORMAT_HEX
```
hughsie, line 1140: is this one true? I thought for HEX we prepended a 0x too?

libfwupd/fwupd-enums.h
```diff
@@ -1101,6 +1101,44 @@ typedef enum {
 	FWUPD_VERSION_FORMAT_LAST
 } FwupdVersionFormat;
 
+/**
+ * FWUPD_VERSION_FORMAT_AABB_CCDD:
+ *
+ * A "pair" version format where a word is split into two sections
+ */
+#define FWUPD_VERSION_FORMAT_AABB_CCDD FWUPD_VERSION_FORMAT_PAIR
```
hughsie, line 1109: my worry with this too is that for the LVFS we make the vendors specify what the version format is in the metainfo file. Would they specify `LVFS::VersionFormat=="AABB_CCDD"` or `LVFS::VersionFormat=="aabb.ccdd"` etc?

hughsie:
As it stands, I think this PR makes things harder to understand, not easier. It might be worth doing something like:

    PAIR -> FWUPD_VERSION_FORMAT_N_N
    TRIPLET -> FWUPD_VERSION_FORMAT_N_N_N
    QUAD -> FWUPD_VERSION_FORMAT_N_N_N_N
    HEX -> FWUPD_VERSION_FORMAT_0XN

...but it kinda falls down for the other version formats -- and we also have other unofficial formats like quad-hex (e.g. `%x.%x.%x.%x` rather than `%i.%i.%i.%i` which could be ``FWUPD_VERSION_FORMAT_X_X_X_X`` or even `%02x.%02x.%02x.%02x` as `FWUPD_VERSION_FORMAT_02X_02X_02X_02X` although I do wonder at what point we turn to the vendor and tell them to stop being ridiculous.

PR #6015, opened 2023-07-19
Allow using varargs to find HID reports

hughsie:
This allows us to combine two functions into one, and means we can also filter on other report attributes as rquired.

Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation


hughsie:
Verified by the "customer", so to speak.

PR #6018, opened 2023-07-20
cros-ec: Handle boards with '_v' in their name

ghost:
servo_v4p1 firmware was identified as being version "4p1_v2.0.8584" which doesn't make much sense.

Since _v isn't supposed to appear in the version triplet, look for the last occurrence.

Type of pull request:

- [x] Code fix

### DIFF
--- a/plugins/cros-ec/fu-cros-ec-common.c
+++ b/plugins/cros-ec/fu-cros-ec-common.c
@@ -13,7 +13,8 @@
 gboolean
 fu_cros_ec_parse_version(const gchar *version_raw, struct cros_ec_version *version, GError **error)
 {
-	g_auto(GStrv) v_split = NULL;
+	gchar *ver = NULL;
+	g_autofree gchar *board = g_strdup(version_raw);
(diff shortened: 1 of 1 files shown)


plugins/cros-ec/fu-cros-ec-common.c
```diff
@@ -26,21 +27,23 @@ fu_cros_ec_parse_version(const gchar *version_raw, struct cros_ec_version *versi
 	}
 
 	/* sample version string: cheese_v1.1.1755-4da9520 */
-	v_split = g_strsplit(version_raw, "_v", 2);
-	if (g_strv_length(v_split) < 2) {
+	ver = g_strrstr(board, "_v");
+	if (ver == NULL) {
```
hughsie, line 31: I'm a bit confused why this wouldn't work for `4p1_v2.0.8584`

hughsie: Aha, got it -- thanks.

PR #6026, opened 2023-07-24
trivial: make permission failures for loading config non-fatal

superm1:
We already made the configuration files optional, so it's incongruent to make failures to read a non-essential file fatal.

This fixes running fwupdtool firmware-parse as non-root.

Type of pull request:

- [x] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-config.c
+++ b/libfwupdplugin/fu-config.c
@@ -163,22 +163,32 @@ fu_config_reload(FuConfig *self, GError **error)
 	for (guint i = 0; i < priv->items->len; i++) {
 		FuConfigItem *item = g_ptr_array_index(priv->items, i);
 		g_autofree gchar *dirname = g_path_get_dirname(item->filename);
+		g_autoptr(GError) error_load = NULL;
+		g_autoptr(GBytes) blob_item = NULL;
 		g_debug("trying to load config values from %s", item->filename);
-		if (g_file_query_exists(item->file, NULL)) {
-			g_autoptr(GBytes) blob = fu_bytes_get_contents(item->filename, error);
-			if (blob == NULL)
-				return FALSE;
-			fu_byte_array_append_bytes(buf, blob);
+		blob_item = fu_bytes_get_contents(item->filename, &error_load);
+		if (blob_item == NULL) {
+			if (g_error_matches(error_load, G_FILE_ERROR, G_FILE_ERROR_ACCES)) {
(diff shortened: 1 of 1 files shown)


libfwupdplugin/fu-config.c
```diff
@@ -165,9 +165,16 @@ fu_config_reload(FuConfig *self, GError **error)
...
+			if (blob == NULL) {
+				if (g_error_matches(error_load, G_FILE_ERROR, G_FILE_ERROR_ACCES)) {
+					g_debug("%s", error_load->message);
+					continue;
+				}
+				g_propagate_error(error, g_steal_pointer(&error_load));
 				return FALSE;
+			}
 			fu_byte_array_append_bytes(buf, blob);
 		}
```
hughsie, line 179: maybe we drop the g_file_query_exists() and handle the extra failure case?

superm1: changed

libfwupdplugin/fu-config.c
```diff
@@ -163,22 +163,29 @@ fu_config_reload(FuConfig *self, GError **error)
...
+		g_autoptr(GBytes) blob_item = NULL;
 		g_debug("trying to load config values from %s", item->filename);
-		if (g_file_query_exists(item->file, NULL)) {
-			g_autoptr(GBytes) blob = fu_bytes_get_contents(item->filename, error);
-			if (blob == NULL)
-				return FALSE;
-			fu_byte_array_append_bytes(buf, blob);
+		blob_item = fu_bytes_get_contents(item->filename, &error_load);
+		if (blob_item == NULL) {
+			if (g_error_matches(error_load, G_FILE_ERROR, G_FILE_ERROR_ACCES)) {
```
hughsie, line 171: is it EACCESS for not-found as well? my UNIX-foo is poor.

libfwupdplugin/fu-config.c
```diff
@@ -163,22 +163,29 @@ fu_config_reload(FuConfig *self, GError **error)
...
+				g_debug("%s", error_load->message);
```
hughsie, line 172: maybe "ignoring config file: %s"

libfwupdplugin/fu-config.c
superm1: no it should be `G_FILE_ERROR_EXIST``, added another case for it.

hughsie (approved):

PR #6037, opened 2023-07-27
Add support for Passim

hughsie:
Using the Passim project allows fwupd to download metadata from other machines on your local network using the SHA-256 hash.

Using Passim should reduce the bandwidth needed by fwupd from offices and homes considerably.

Type of pull request:

- [ ] Code fix
- [X] Feature
- [ ] Documentation

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -19,6 +19,10 @@ engine_dep = [
   cbor,
 ]
 
+if get_option('passim').allowed()
+  engine_dep += passim
+endif
+
 client_dep = [
   gudev,
   gusb,
--- a/subprojects/passim.wrap
+++ b/subprojects/passim.wrap
@@ -1,0 +1,4 @@
+[wrap-git]
+directory = passim
+url = https://github.com/hughsie/passim.git
+revision = 0.1.0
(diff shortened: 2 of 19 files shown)


subprojects/passim.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+directory = passim
+url = https://github.com/hughsie/passim.git
+revision = main
```
superm1, line 4: I think we should gate merging this in fwupd until it's at least a tagged version.

src/meson.build
```diff
@@ -19,6 +19,10 @@ daemon_dep = [
   cbor,
 ]
 
+if get_option('passim').allowed()
+  daemon_dep += passim
```
superm1, line 23: this seems like wrong place to put it; you want it in fwupdtool too.

subprojects/passim.wrap
hughsie: 100% yes.

src/meson.build
hughsie: Ohh, the variable is badly named; it should be engine_dep -- it's used in both. I'll rename it in this PR.

subprojects/passim.wrap
hughsie: Fixed, as I did a 0.1.0 release just now.

hughsie:
I think this probably depends on https://github.com/fwupd/fwupd/pull/6094

superm1 (approved):

PR #6042, opened 2023-07-28
Update BIOS setting through notifying fu-engine

smallorange:
The original way to update BIOS settings is to call an updating function which was placed in fu-engine. That increased the complexity when the other components need to update the BIOS. The fu-engine should be passed to the objects and then this object can use the services from fu-engine. That leads the code much more complicated and difficult to manage.

In this work, we would keep every low-level task in fu-engine. If any components would like to update the BIOS, it can put the pending values into the corresponding data structure and then notifies the fu-engine to process the corresponding task. The components, for example, just only update the BIOS setting pending value and then notify the engine to finish the request. It keeps the code become simpler and clean and also keeps its architecture.

Type of pull request:

- [ ] Code fix
- [x] Feature
- [ ] Documentation

### DIFF
--- a/libfwupdplugin/fu-bios-settings.c
+++ b/libfwupdplugin/fu-bios-settings.c
@@ -556,6 +556,33 @@ fu_bios_settings_get_all(FuBiosSettings *self)
 }
 
 /**
+ * fu_bios_settings_get_unsettled_values:
+ * @self: a #FuBiosSettings
+ *
+ * ets all BIOS unsettled values.
+ *
+ * Returns: (transfer container) (element-type FwupdBiosSetting): attributes
+ *
+ * Since: 1.9.4
+ **/
+GPtrArray *
+fu_bios_settings_get_unsettled_values(FuBiosSettings *self)
+{
+	GPtrArray *result = NULL;
+	g_return_val_if_fail(FU_IS_BIOS_SETTINGS(self), NULL);
+
+	result = g_ptr_array_new();
+
+	for (guint i = 0; i < self->attrs->len; i++) {
+		FwupdBiosSetting *attr = g_ptr_array_index(self->attrs, i);
+		if (fwupd_bios_setting_get_unsettled_value(attr))
+			g_ptr_array_add(result, g_object_ref(attr));
+	}
+
+	return g_ptr_array_ref(self->attrs);
+}
+
+/**
  * fu_bios_settings_get_pending_reboot:
  * @self: a #FuBiosSettings
  * @result: (out): Whether a reboot is pending
--- a/src/fu-engine.c
+++ b/src/fu-engine.c
@@ -1134,6 +1134,61 @@ fu_engine_modify_bios_settings(FuEngine *self,
 	return TRUE;
 }
 
+gboolean
+fu_engine_update_bios_unsettled_settings(FuEngine *self,
+					 GHashTable *settings,
+					 gboolean force_ro,
+					 GError **error)
+{
+	gboolean changed = FALSE;
+	GHashTableIter iter;
+	gpointer key, value;
+
+	g_return_val_if_fail(FU_IS_ENGINE(self), FALSE);
+	g_return_val_if_fail(settings != NULL, FALSE);
+	g_return_val_if_fail(error == NULL || *error == NULL, FALSE);
+
+	g_hash_table_iter_init(&iter, settings);
+	while (g_hash_table_iter_next(&iter, &key, &value)) {
+		changed = fu_context_get_bios_setting_unsettled_value(self->ctx, key, value, error);
+	}
+
+	if (!changed)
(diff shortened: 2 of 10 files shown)


libfwupdplugin/fu-bios-settings.c
```diff
@@ -555,6 +555,33 @@ fu_bios_settings_get_all(FuBiosSettings *self)
...
 
+/**
+ * fu_bios_settings_get_pending_values:
+ * @self: a #FuBiosSettings
+ *
+ * Gets all the attributes in the object.
+ *
+ * Returns: (transfer container) (element-type FwupdBiosSetting): attributes
+ *
+ * Since: 1.8.4
```
superm1, line 566: 1.9.4, no?

src/fu-engine.c
```diff
@@ -1135,6 +1135,64 @@ fu_engine_modify_bios_settings(FuEngine *self,
...
+		if (pending_value == NULL)
+			continue;
+
+		g_info("BIOS settings: %s is updating to %s", id, pending_value);
+		g_hash_table_insert(bios_settings_hash, g_strdup(id), g_strdup(pending_value));
+		fwupd_bios_setting_set_pending_value(attr, NULL);
+	}
+	fu_engine_modify_bios_settings(self, bios_settings_hash, FALSE, &local_error);
+
+	g_info("BIOS setting is triggered.\n");
```
superm1, line 1193: this seems needlessly noisy I'd think, do you really still need it?

src/fu-engine.c
```diff
@@ -1135,6 +1135,64 @@ fu_engine_modify_bios_settings(FuEngine *self,
...
+		FwupdBiosSetting *attr = g_ptr_array_index(pending_bios_settings, i);
+		const gchar *id = NULL;
+		const gchar *pending_value = NULL;
+
+		id = fwupd_bios_setting_get_id(attr);
+		pending_value = fwupd_bios_setting_get_pending_value(attr);
+		if (pending_value == NULL)
+			continue;
+
+		g_info("BIOS settings: %s is updating to %s", id, pending_value);
```
superm1, line 1187: this seems needlessly noisy I'd think, do you really still need it?

libfwupdplugin/fu-bios-settings.c
smallorange: Thank you for reviewing.

It should be 1.9.4. I'll correct this.

src/fu-engine.c
smallorange: Sure. It needs to be removed.

libfwupdplugin/fu-bios-settings.c
```diff
@@ -555,6 +555,33 @@ fu_bios_settings_get_all(FuBiosSettings *self)
 	return g_ptr_array_ref(self->attrs);
 }
 
+/**
+ * fu_bios_settings_get_pending_values:
+ * @self: a #FuBiosSettings
+ *
+ * Gets all the attributes in the object.
```
superm1, line 562: This description seems wrong.

src/fu-engine.c
smallorange: Okay. I'll remove it.

libfwupdplugin/fu-bios-settings.c
```diff
@@ -555,6 +555,33 @@ fu_bios_settings_get_all(FuBiosSettings *self)
 	return g_ptr_array_ref(self->attrs);
 }
 
+/**
+ * fu_bios_settings_get_pending_values:
```
superm1, line 559: I'm a little apprehensive about the use of the word `pending` in this context.  There is also a `pending_reboot` sysfs file, and I think by using `pending` the two can get mixed up.  How about `unsettled`?

smallorange: Sounds good. I'll replace them :)

hughsie:
Hey Kate -- I was thinking about this functionality on PTO -- I've had a alight change in heart in how I think this should work -- mind if I take your branch, fork it and prototype something different?

hughsie:
@smallorange I've prototyped something completely different based on this branch -- does https://github.com/fwupd/fwupd/pull/6082 work or have I missed some vital information?

smallorange:
Looks good, Let me try this with the "object notifiy". This looks like, the implementation of BIOS setting in the engine can be removed. So, I can only get the attr from the plugin and then call fwupd_bios_setting_write_value(). The setting updating is done. Cool.

smallorange:
> Hey Kate -- I was thinking about this functionality on PTO -- I've had a alight change in heart in how I think this should work -- mind if I take your branch, fork it and prototype something different?

Sorry for the late response. sure, it's ok :)

hughsie:
@smallorange do we still need this?

PR #6049, opened 2023-07-31
Wip/hughsie/macos

hughsie:
Type of pull request:

- [X] Code fix
- [ ] Feature
- [ ] Documentation

### DIFF
--- a/libfwupd/fwupd-common.c
+++ b/libfwupd/fwupd-common.c
@@ -245,8 +245,51 @@ fwupd_get_os_release_full(const gchar *filename, GError **error)
 GHashTable *
 fwupd_get_os_release(GError **error)
 {
+#ifdef HOST_MACHINE_SYSTEM_DARWIN
+	g_autofree gchar *stdout = NULL;
+	g_autofree gchar *sw_vers = g_find_program_in_path("sw_vers");
+	g_auto(GStrv) split = NULL;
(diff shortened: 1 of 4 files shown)


superm1 (approved):

hughsie:
@superm1 one more -- this actually gives us os-release data.

libfwupd/fwupd-common.c
```diff
@@ -245,7 +281,14 @@ fwupd_get_os_release_full(const gchar *filename, GError **error)
...
 fwupd_get_os_release(GError **error)
 {
+	g_autofree gchar *sw_vers = g_find_program_in_path("sw_vers");
 	g_return_val_if_fail(error == NULL || *error == NULL, NULL);
+
+	/* macOS */
+	if (sw_vers != NULL)
+		return fwupd_get_os_release_for_macos(sw_vers, error);
+
+	/* Linux and *BSD */
```
superm1, line 291: I think we should be #ifdef for different OS here - there is no value to having a fallback for Linux do this.  It's just going to be an extra check on each OS that is unnecessary.

hughsie: TARGET_OS_MAC isn't super easy to use. Maybe like this instead? ^^

superm1 (approved):

PR #6063, opened 2023-08-09
trivial: Add some Threat Modeling notes

hughsie:
I have to do this for a new Red Hat process, and I think it's easier to keep up to date if it's upstream.

Type of pull request:

- [ ] Code fix
- [ ] Feature
- [X] Documentation

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -39,3 +39,135 @@ Failing that, please report the issue against the `fwupd` component in Red Hat
 bugzilla, with the security checkbox set. You should get a response within 3
 days. We have no bug bounty program, but we're happy to credit you in updates
 if this is what you would like us to do.
+
+## Threat Modeling
+
+### Who We Trust
+
+In this diagram, the arrow shows the flow of *information* from one entity to another.
+
+Important things to note:
+
+* OEMs and ODMs have to apply for a LVFS account and the website and email domain is verified
+* OEMs and ODMs can only upload for devices that match their device-supplied vendor-id
+* The relationship between the OEM/ODM and affiliate ISV/IBV is implemented using per-task ACLs
+* The LVFS is assumed to be managed by a vendor-neutral trusted team
+* Signing of the metadata and firmware is implemented using PKCS#7 and GPG
+* End users only trust the LVFS signing signatures by default
+* Metadata contains SHA-1 and SHA-256 hashes of the firmware archive
+* Access to the embargo remotes requires knowing the secret vendor hash, but not a token
+* The firmware archive internal metadata and firmware payload are both signed
+* Reports have to be signed by the user machine key to be attributable to an OEM or QA team
+* Signed reports are uploaded using a username and access token
+* SBoM metadata is extracted from the payload by the LVFS and formatted into HTML/SWID formats
+* Security researchers can only run FwHunt/Yara scans on public firmware
+
+```mermaid
+  flowchart TD
+      LVFS((LVFS Webservice)) -- "SBoM.html" --> User(End User)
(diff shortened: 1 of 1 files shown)


hughsie:
@superm1 I think when we merge this the diagrams should appear -- you can use https://mermaid.live/ if you want a sneak preview.

superm1:
Cool tool! TIL.

It renders here too.
https://github.com/fwupd/fwupd/blob/wip/hughsie/mermaid/SECURITY.md

You're missing entities for the kernel and firmware as data exchanges with both

hughsie:
@superm1 how about that? I added a lot more words too.

superm1:
> @superm1 how about that? I added a lot more words too.

So every threat model I've done includes a few things that I think are missing.  The tool might not support them but I'll mention them in case they can be added:
1. Extra dotted lines to show boundaries, for example a root / kernel boundary, a user / root boundary, web / app boundary.
2. ALL data flows.  This is very UEFI centric right now, but the daemon exchanges a lot of data with the kernel via IOCTLs, sysfs files, etc.  These could potentially be tampered with.  So either more flows needed or more assumptions needed (for example unless you assume UEFI SB trusted by MS + distro kernel under lockdown)

hughsie:
> ALL data flows

I agree, but I don't think we need to enumerate everything on that diagram -- we already have it spelt out in the various plugin README.md files.

> Extra dotted lines to show boundaries

Will ponder. I think I can do a subgraph -- or maybe we just use different colors for different things? Will play -- this is iteration 1 afterall

superm1:
Ah, that's great!  I think the "root" one needs to be split though.  Especially when you think about kernel lockdown; root isn't all powerful.  I think `root` should split into these categories.

`root`
`kernel`
`bios`
`platform`

Within platform you can then have these:
`bios`, `disk`

Within disk you can have the `efi system partition` as a data store.

Within bios you can have
`nvram` as a data store
`runtime services` as an API

hughsie:
> Ah, that's great! I think the "root" one needs to be split though

Hmm, that gets very murky -- I'm trying to show *what* we trust, rather than explicitly show the different privilege layers. What I'm trying to explain is that fwupd-efi is trusted, and the user stuff is untrusted -- rather than showing the interfaces per-se.

I think the traditionally layers of rings diagram is useful, I just think it's an additional diagram. I have added a few of your ideas tho:

Screenshot 2023-08-10 at 19-05-34 Online FlowChart   Diagrams Editor - Mermaid Live Editor

superm1:
Hmm, but the daemon does more to the ESP.  It puts the capsule loader onto the ESP too.  I really think the ESP should be a data store.   I think the problem is that this graph is compounding multiple steps that the daemon does.  The actual act of running the EFI application and what happens from that maybe it should be a separate graph.

hughsie:
>  It puts the capsule loader onto the ESP too

Added

> I really think the ESP should be a data store

I've changed the icon, as I did the statedir and nvram.

>  The actual act of running the EFI application and what happens from that maybe it should be a separate graph

Once I've got the hang of this I was thinking of putting some of this in the plugin readme.md files.

superm1:
I have an idea how to make it obvious that fwupdtool and fwupd support all the same stuff. How about you make an abstraction for the engine and everything that talks to the daemon instead talks to the engine.

hughsie:
> make an abstraction for the engine

Done.

superm1:
In the most recent push I was looking at it and saw what looks like an arrow to nothing.

image

Otherwise I'm happy with this.  I think we can add "low" level graphs for the individual plugins that show the data flows from the engine to the device/kernel as we go forward.

superm1 (approved):

